PR #1189, opened 2021-01-09
Cypress test calc top toolbar

Ezinnem:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
+++ b/cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
@@ -17,6 +17,21 @@ describe('Top toolbar tests.', function() {
 		helper.afterAll(testFileName, this.currentTest.state);
 	});
 
+	it('Merge cells', function() {
+
+		// Select the full column
+		calcHelper.selectFirstColumn();
+	
+		// Despite the selection is there, merge cells needs more time here.
+		cy.wait(1000);
+	
+		cy.get('.w2ui-tb-image.w2ui-icon.togglemergecells')
+			.click();
+
+		desktopHelper.checkDialogAndClose('Merge Cells');
+	});
+
+
 	it('Remove cell border', function() {
 		cy.get('#tb_editbar_item_setborderstyle')
 			.click();
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -146,9 +146,25 @@ function selectEntireSheet(removeSelection = true) {
 	cy.log('Selecting entire sheet - end.');
 }
 
+function selectFirstColumn() {
+	cy.get('.spreadsheet-header-columns')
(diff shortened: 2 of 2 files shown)


cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
```diff
@@ -17,17 +17,61 @@ describe('Top toolbar tests.', function() {
...
+
+		helper.getCursorPos('left', 'currentTextEndPos');
+	}
 
 	it('Merge cells', function() {
-		calcHelper.selectEntireSheet();
 
-		cy.get('#tb_editbar_item_togglemergecells')
+		// Select the full row
+		calcHelper.selectFirstColumn();
```
tzolnai, line 29: The comment above says different thing about what the code actually does.

cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
```diff
@@ -17,17 +17,61 @@ describe('Top toolbar tests.', function() {
...
+		// Despite the selection is there, merge cells needs more time here.
+		cy.wait(1000);
+	
+		cy.get('.w2ui-tb-image.w2ui-icon.togglemergecells')
 			.click();
-		
-		cy.get('.lokdialog_canvas')
+	
+		cy.get('.ui-dialog.ui-corner-all.ui-widget.ui-widget-content.ui-front.lokdialog_container')
 			.should('exist');
```
tzolnai, line 38: There is a helper method for dialogs: checkDialogAndClose() in helper.js. It needs the dialog title as a parameter, so it can check whether the right dialog is opened. Can you use that here?

tzolnai (commented):
You modify the merge cells test case in both commits. It would be better to merge these two commits into one.

cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
```diff
@@ -17,6 +17,60 @@ describe('Top toolbar tests.', function() {
...
+		cy.get('@currentTextEndPos')
+			.as('originalTextEndPos');
+
+		cy.get('@currentTextEndPos')
+			.should('be.greaterThan', 0);
+
+		calcHelper.selectFirstColumn();
+
+		cy.get('.w2ui-tb-image.w2ui-icon.wraptext')
+			.click();
```
tzolnai, line 55: If I comment out these lines above, where the test applies the text wrapping, the test still passes. That means we don't actually verify that the text wrapping toolbar item does work. It would be good to find out a way to test it properly.

tzolnai (commented):
The commit message should be consistent with the commit. In this commit, you do not add only a merge cells test case.
Maybe you can create a separate PR for the Merge Cells test case. That one nicely works, so I can push that and so you can focus on the other test case, to find out what to do there.

PR #1192, opened 2021-01-09
check if we run in a container

timar:
Change-Id: Idc9db40c00d41c95160db130eb324c487f14de17

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/security.h
+++ b/common/security.h
@@ -32,6 +32,31 @@ inline int hasUID(const char *userId)
     return 0;
 }
 
+inline int isInContainer()
+{
+#ifdef __linux__
+    FILE *cgroup;
+    char line[80];
+    const char *docker = ":/docker/";
+    cgroup = fopen("/proc/self/cgroup", "r");
+    if(!cgroup)
+    {
+        fprintf(stderr, "Error: cannot open /proc/self/cgroup\n");
+        return 0;
+    }
+    while (fgets(line, sizeof(line), cgroup) != NULL)
+    {
+        if (strstr(line, docker) != NULL)
+        {
+            fclose(cgroup);
+            return 1;
+        }
+    }
+    fclose(cgroup);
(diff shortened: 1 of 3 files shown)


common/security.h
```diff
@@ -32,6 +32,30 @@ inline int hasUID(const char *userId)
...
+    cgroup = fopen("/proc/self/cgroup", "r");
+    if(!cgroup)
+    {
+        fprintf(stderr, "Error: cannot open /proc/self/cgroup\n");
+        return 0;
+    }
+    if (fgets(line, sizeof(line), cgroup) == NULL)
+    {
+        fprintf(stderr, "Error: cannot read from /proc/self/cgroup\n");
+        return 0;
```
kendy, line 50: Shouldn't we fclose(cgroup) here too?

common/security.h
```diff
@@ -32,6 +32,30 @@ inline int hasUID(const char *userId)
...
+    }
+    fclose(cgroup);
```
kendy, line 52: I guess we should line[sizeof(line)-1] = 0; here before the strstr to be safe?

common/security.h
kendy: I mean, one line up, before the return obviously :-)

common/security.h
timar: fgets() inserts the terminating \0

common/security.h
timar: I fixed this in latest version of PR.

mmeeks (approved):
Looks great to me =) thanks for the feedback etc. - good to have lots of eyes on this sort of thing.

common/security.h
kendy: Ah you are right, thanks :-)

PR #1210, opened 2021-01-12
wsd: disable ssl by default, enable ssl termination by default

vmiklos:
Change-Id: I7fd725a83b0e9ca1012f2c0e0c3bf038e5fa0059

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -387,6 +387,7 @@ run: all @JAILS_PATH@
 	./loolwsd --o:sys_template_path="@SYSTEMPLATE_PATH@" \
 			  --o:security.capabilities="$(CAPABILITIES)" \
 			  --o:child_root_path="@JAILS_PATH@" --o:storage.filesystem[@allow]=true \
+			  --o:ssl.enable=true \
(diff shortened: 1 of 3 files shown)


vmiklos:
Let's see if CI has any problems with this. A normal Linux make check passes; the intention is to have defaults which are close to how people typically use Online in production.

mmeeks:
Looks good to me, but is 'make run' happy with this ? =)

vmiklos:
> Looks good to me, but is 'make run' happy with this ? =)

Yes, it is, however there is an interesting trade-off here. On one hand, we want to use the non-ssl codepath in wsd for make run if the production scenario uses non-ssl as well. On the other hand, we want to use the ssl codepath for make run, so the browser gets https, similar to the production scenario.

The above version does non-ssl codepath for both cases, but now that you asked, it's probably more important to have https in the browser for make run, to be able to hit the tricky "js problem with ssl" cases with make run. Let me update the patch accordingly. (I.e. the config change is still the same, but explicitly request ssl for make run and disable ssl termination for cypress.)

mmeeks:
Sounds fine - I think the SSL code-paths are more 'hairy' than the plain socket ones - so we should prolly be testing those as developers too =) thanks !

PR #1211, opened 2021-01-12
Notebookbar: disable commands without gray background

Andreas-Kainz:
Change-Id: I539d70ab1ba07d27cf34b61a8bf1a333e0bc4531

Before
2021-01-12 19_40_24-ws-ea5b92be-06de-4305-a5ec-72e82ce16980_0 - noVNC
After
2021-01-12 19_39_28-ws-ea5b92be-06de-4305-a5ec-72e82ce16980_0 - noVNC

not available commands shouldn't be "highlighted"  with a gray background. The icon and string are gray (like in the menu bar) so there is no reason for additional colors.

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -329,6 +329,10 @@ div[id*='Row'].notebookbar, div[id*='Column'].notebookbar, #SendToBack.notebookb
 	color: var(--blue1-txt-primary-color) !important;
 }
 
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
without feedback from @pedropintosilva I wan't submit this patch.

pedropintosilva:
^ I think we really need to have something else for those disabled actions because otherwise the contrast between enabled and disabled actions gets extremely fuzzy. Thus I have iterated upon your changes and have added opacity attribute.: 
image

Andreas-Kainz:
> image

I'm fine with the bottom layout. Look good.

PR #1212, opened 2021-01-12
.ui-content.unobutton.selected selected color is larger

Andreas-Kainz:
Change-Id: Ic45b68778d8adef375ad0cb2495879c860e986cd

Before
2021-01-12 21_03_21-text docx - Dateien - Nextcloud
After
2021-01-12 21_27_53-text docx - Dateien - Nextcloud

How it look in classic toolbar mode
2021-01-12 21_02_50-text docx - Files - Nextcloud

**with the patch the selected commands look similar to how they look in classic toolbar mode**

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -227,6 +227,7 @@
 
 .hasnotebookbar .ui-content.unobutton.selected {
 	border-radius: 2px;
+	box-shadow: 0px 0px 0px 3px var(--gray-bg-color);
 	background-color: var(--gray-bg-color);
 }
 


Andreas-Kainz:
Feedback @pedropintosilva

pedropintosilva:
hm I think it would be slightly tidier if we do the other way around so, instead of increasing in the notebookbar I would decrease in the classic mode, this would also have the advantage to leave us with better and more consistent border radius (because we are not so heavily dependent on box-shadow to increase size). I have been iterating locally and it would result in something like this this,
for classic:
after-classic

and for notebookbar mode:
after-notebookbar

This I think achieved your end goal, the elements become more similar between different modes (similar not equal since the size for the icons are different in each mode) What do you think?

Andreas-Kainz:
2021-01-13 12_48_59-SimpleTemplate2 odt - LibreOfficeDev Writer

this is how it look like on desktop (sidebar is similar) 

I would prefer to have more space between icon and selection frame.

pedropintosilva:
I understand but space and size are not one and the same, for example my simply tweaking the space: 
image

Andreas-Kainz:
> After
> 2021-01-12 21_27_53-text docx - Dateien - Nextcloud

I prefer this layout maybe we can use something else than box-shadow but the result is good in my point of view.

pedropintosilva:
understood and I ok either way but if we go with that then we probably need to find a better way to do this without affecting too much the radius

Andreas-Kainz:
any idea how to make do it with another setting?

loleaflet/css/notebookbar.css
```diff
@@ -227,6 +227,7 @@
 
 .hasnotebookbar .ui-content.unobutton.selected {
 	border-radius: 2px;
+	box-shadow: 0px 0px 0px 3px var(--gray-bg-color);
```
pedropintosilva, line 230: instead of adding this

pedropintosilva: add after the following selector `.unotoolbutton.notebookbar:not(.disabled):hover, #clearFormatting.notebookbar div img:hover` block :

```css
.hasnotebookbar .ui-content.unotoolbutton.selected:not(.has-label):not(.inline) {
	background-color: #e6e6e6b0;
	border-radius: 4px; /*same as .w2ui-toolbar table.w2ui-button*/
}
```

pedropintosilva (changes requested):

loleaflet/css/notebookbar.css
Andreas-Kainz: Didn't work also the box-shadow stuff didn't work now.

PR #1213, opened 2021-01-12
Cypress Test on Enable text wrapping on Calc

Ezinnem:
Created a cypress test for the Text wrap feature on the Calc top toolbar

Signed-off-by: Ezinne Nnamani <nnamani.ezinne@collabora.com>
Change-Id: I3d2e59b94e23e73b5980e049fe5c6e99178b8d3f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
+++ b/cypress_test/integration_tests/desktop/calc/top_toolbar_spec.js
@@ -16,7 +16,45 @@ describe('Top toolbar tests.', function() {
 	afterEach(function() {
 		helper.afterAll(testFileName, this.currentTest.state);
 	});
+	function getTextEndPosForFirstCell() {
+		calcHelper.dblClickOnFirstCell();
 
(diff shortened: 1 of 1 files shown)


tzolnai:
Hi Ezinne,

My previous comment on this test code still stands. If I remove these code lines:
```
cy.get('.w2ui-tb-image.w2ui-icon.wraptext')
			.click();
```
, then the test passes. This means the test won't catch if the tested feature is broken, non-working. So you should fix this thing. I checked the similar mobile test (calc/alignment_options_spec.js) and there the test fails if I remove "pushing the text wrapping button" part. There must be some difference between mobile and desktop behavior. So I would suggest trying to find out what is that difference here doing the test steps and adapt the code for desktop accordingly.

tzolnai (changes requested):
Let's mark this, it has requested changes. See previous comment.

PR #1217, opened 2021-01-13
wsd: new NetUtil file for network utilities

Ashod:
Move the connect function into the NetUtil
translation unit to aid using it for the
upcoming async socket logic.

The NetUtil should also come in handy for
the miscellaneous network helpers we have.

Change-Id: I2ee0c6e3e1769fd87572d7407d3b4979b59ffe6a

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -1,0 +1,88 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <config.h>
+
+#include "NetUtil.hpp"
+
+#include "Socket.hpp"
+#if ENABLE_SSL && !MOBILEAPP
+#include "SslSocket.hpp"
+#endif
+
+#include <netdb.h>
+
+namespace net
+{
+std::shared_ptr<StreamSocket>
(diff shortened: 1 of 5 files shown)


net/NetUtil.cpp
```diff
@@ -0,0 +1,88 @@
...
+#if ENABLE_SSL && !MOBILEAPP
+#include "SslSocket.hpp"
+#endif
+
+#include <netdb.h>
+
+namespace net
+{
+std::shared_ptr<StreamSocket>
+connect(const std::string& host, const std::string& port, const bool encrypted,
```
kendy, line 22: Would be great not to rename the "isSSL" to "encrypted" if possible - "encrypted" is too general in my view, making the code reading a bit harder (at least for me) :-)  Can we keep "isSSL" (or "useSSL" if that does not fit) please?

test/Makefile.am
```diff
@@ -89,6 +89,7 @@ unittest_SOURCES = \
     ../common/Unit.cpp \
     ../common/StringVector.cpp \
     ../net/Socket.cpp \
+	../net/NetUtil.cpp \
```
kendy, line 92: Cosmetic - indentation...

kendy (commented):
@Ashod: Thanks for that!  Looks good to me in general, commented about some small things; would be good to fix if possible, otherwise good to go in from my point of view.  Thank you!

net/NetUtil.cpp
Ashod: Yes, no good reason to change it. I typed `encrypted` separately, but now changed to isSSL.

test/Makefile.am
Ashod: Ironic that I had noticed all the tabs and made sure (failed, of course) to use spaces (I blame my editor). I replaced the tabs to spaces.

PR #1218, opened 2021-01-13
WOPITest improvements

Ashod:
These commits fix 4 WOPI Tests so they wouldn't rely on the timing of the SocketPoll interval, instead they now properly wait for the different events from Core before they move to the next test phase.

Better logging, comments, and overall readability is also improved. 

More info in the individual commit messages (but the review is best done over all changes together).

- wsd: test: remove redundant wakeupWorld and log the exit status
- wsd: test: move inherited filterSendMessage to onFilterSendMessage
- wsd: test: make DocumentConflict test robus and simpler
- wsd: test: make UnitWOPI test robust and simpler
- wsd: test: make owner-termination test robust and simpler
- wsd: test: make Version Restore test robust and simpler
- wsd: test: simplify sending commands
- wsd: test: shorter poll interval during tests
- wsd: test: support SSL in classic tests

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -210,9 +210,10 @@ void UnitBase::exitTest(TestResult result)
     _setRetValue = true;
     _retValue = result == TestResult::Ok ? EX_OK : EX_SOFTWARE;
 #if !MOBILEAPP
-    SigUtil::requestShutdown();
-#endif
+    SigUtil::requestShutdown(); // And wakupWorld.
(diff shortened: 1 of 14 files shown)


vmiklos (approved):
I see nothing obviously problematic with these changes. The net/ and wsd/ pieces could break something, but those are fine. For the rest, doing more event-based tests and less time-based tests sounds like a good direction.

PR #1221, opened 2021-01-13
CanvasTileLayer: Use same canvas for row & column headers.

gokaysatir:
Change-Id: Ia612c25b0b9e8be96daf157fdcc1b9afc979c679

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/mobile/calc/hamburger_menu_spec.js
+++ b/cypress_test/integration_tests/mobile/calc/hamburger_menu_spec.js
@@ -511,104 +511,44 @@ describe('Trigger hamburger menu options.', function() {
 	it('Data: grouping / ungrouping.', function() {
 		before('hamburger_menu.ods');
 
-		// Use columns header height as indicator
-		helper.initAliasToNegative('origHeaderHeight');
-
-		cy.get('.spreadsheet-header-columns')
-			.invoke('height')
-			.as('origHeaderHeight');
-
-		cy.get('@origHeaderHeight')
-			.should('be.greaterThan', 0);
-
 		// Group first
 		calcHelper.selectFirstColumn();
 
 		mobileHelper.selectHamburgerMenuItem(['Data', 'Group and Outline', 'Group...']);
 
-		cy.get('@origHeaderHeight')
-			.then(function(origHeaderHeight) {
-				cy.get('.spreadsheet-header-columns')
-					.should(function(header) {
-						expect(header.height()).to.be.greaterThan(origHeaderHeight);
-					});
-			});
+		cy.get('[id="test-div-column group"]').should('exist');
 
 		// Then ungroup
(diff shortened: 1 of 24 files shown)


mmeeks (commented):
Looks good - would love to get rid of dpiscale thrash on that canvas though =) I assume we'll move all the rendering to corePx for row/column ? =)

gokaysatir:
> Looks good - would love to get rid of dpiscale thrash on that canvas though =) I assume we'll move all the rendering to corePx for row/column ? =)

     Of course, i will eliminate them as much as i can. We just need to keep some of them for the variables created on the client side. Like borders etc. We can multiply them with dpiScale before the drawing phase. I will check it again.

gokaysatir:
Fonts, canvas position on mobile and some refreshing issues will be adjusted.

cypress_test/integration_tests/mobile/calc/hamburger_menu_spec.js
```diff
@@ -528,9 +528,9 @@ describe('Trigger hamburger menu options.', function() {
 
 		cy.get('@origHeaderHeight')
 			.then(function(origHeaderHeight) {
-				cy.get('.spreadsheet-header-columns')
+				cy.get('[id="test-div-column header"]')
 					.should(function(header) {
-						expect(header.height()).to.be.greaterThan(origHeaderHeight);
+						expect(header.height()).to.be.equal(origHeaderHeight);
```
tzolnai, line 533: Here, it would be better to use something else than origHeaderHeight. Earlier, grouping was drawn inside the column header DOM element, so the header's height was a good indicator for the grouping visibility. However, checking equality here means, that the header's height does not change anymore, so this won't actually test that the grouping is there or not.
As I see with this PR, you introduced a new canvas section for grouping. So you should use that section to check whether enabling / disabling grouping has a result on the document. It might be enough to check the existence of the grouping section or the size of it, whatever indicates that the grouping is visible.
There are more test cases with the same issue: 'Data: grouping / ungrouping.', 'Data: remove grouping outline.', 'Data: show / hide grouping details.'.

tzolnai (changes requested):
Hi, It looks like this new test div method works nicely. I pointed out only one issue related to the grouping test cases. Otherwise, it looks great wrt. cypress.

gokaysatir:
> Hi, It looks like this new test div method works nicely. I pointed out only one issue related to the grouping test cases. Otherwise, it looks great wrt. cypress.

I updated related parts. Thank you for the review :)

tzolnai (approved):
Looks good to me wrt cypress tests.

kendy:
Looks great, thank you! :-)  Some minor nitpicks - will send them separately to you...

PR #1225, opened 2021-01-13
wsd: disable ssl by default, enable ssl termination by default

vmiklos:
The intention is to have defaults which are close to how people
typically use Online in production.

However, keep using ssl for 'make run', so that the https environment in
the browser is unchanged.

(cherry picked from commit f6bf6f49ed11c53e2f90ab1169d5fd64804a06ac)

Change-Id: I7fd725a83b0e9ca1012f2c0e0c3bf038e5fa0059


vmiklos:
@timar could you please review this one? Thanks. And once this is in, the install guide could be updated, now that you don't need to customize ssl.enable and ssl.termination in the proxy case.

vmiklos:
See c43562c9a765fd6f667e0101ea818f78155702c6, this is not a good idea after all.

PR #1226, opened 2021-01-13
add support for changing ui mode dynamically

gmasei11:
Change-Id: Id84a4feb8f37357d7de8f9cfebea3612751709b1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/control/Control.Menubar.js
+++ b/loleaflet/src/control/Control.Menubar.js
@@ -968,6 +968,12 @@ L.Control.Menubar = L.Control.extend({
 		this._initialized = false;
 		this._hiddenItems = [];
 		this._menubarCont = L.DomUtil.get('main-menu');
+		// In case it contains garbage
+		if (this._menubarCont)
+			this._menubarCont.remove();
+		// Use original template as provided by server
+		this._menubarCont = map.mainMenuTemplate.cloneNode(true);
+		$('#main-menu-state').after(this._menubarCont);
 		this._initializeMenu(this.options.initial);
 
 		map.on('doclayerinit', this._onDocLayerInit, this);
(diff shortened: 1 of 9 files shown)


wsd/LOOLWSD.cpp
```diff
@@ -1021,9 +1020,6 @@ void LOOLWSD::initialize(Application& self)
     // Allow UT to manipulate before using configuration values.
     UnitWSD::get().configure(config());
 
-    // Setup user interface mode
-    UserInterface = getConfigValue<std::string>(conf, "user_interface.mode", "classic");
```
eszkadev, line 1025: wait, this one goes from loolwsd.xml? as we discussed earlier also with @kendy: user_interface.mode should stay there. It setups the default mode if 'uidefaults' are not set for the request by integrator.

gmasei11: In that place it is useless after we decided to activate notebookbar anyway. It was used only for activating notebookbar in core. The default mode is set here: https://github.com/CollaboraOnline/online/blob/private/gmasei/change_ui_mode/wsd/FileServer.cpp#L774 in case 'uiDefaults' are not set.

eszkadev: ok then

loleaflet/src/control/Control.Menubar.js
```diff
@@ -1422,7 +1439,7 @@ L.Control.Menubar = L.Control.extend({
 			self._executeAction(item);
 		}
 
-		if (!window.mode.isMobile() && $(item).data('id') !== 'insertcomment')
+		if (!window.mode.isMobile() && $(item).data('id') !== 'insertcomment' && self && self.map)
 			self._map.focus();
```
eszkadev, line 1443: in condition: `self.map`
below: `self._map`

gmasei11: Good catch! Pushed fix.

eszkadev (changes requested):
I tested it. Switching components works nicely. But unfortunately we can't merge it currently.
There are some problems:
- classic mode has cleared font and style lists (after switch nb -> classic)
- notebookbar is not properly initialized (style previews doesn't work until we switch to other tab and back to Home , in case: nb -> classic -> nb)

eszkadev:
@gmasei11 hi, I fixed font and font size selectors. paragraph style listbox left to fix

eszkadev:
@gmasei11 hi, please rebase as PR has conflicts. you should also remove my commit: 'notebookbar: reuse font selectors ' as it is already merged to master.

eszkadev:
the last nitpick: when switching from original mode to the another there is a vertical difference between document content and sidebar top border. about 5x off.   when swithcing back to original it is ok, then when switching again it appears too.  Reproducible in both modes as a original one: classic and notebookbar.

gmasei11:
> the last nitpick: when switching from original mode to the another there is a vertical difference between document content and sidebar top border. about 5px off. when swithcing back to original it is ok, then when switching again it appears too. Reproducible in both modes as a original one: classic and notebookbar.

Fixed

gmasei11:
Closed by mistake :D Reopened.

eszkadev (approved):
tested just now, works perfectly! :)

PR #1236, opened 2021-01-13
Enable Draw

merttumer:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


loleaflet/src/control/Parts.js
```diff
@@ -279,7 +279,7 @@ L.Map.include({
 		var docLayer = this._docLayer;
 
 		// At least for Impress, we should not fire this. It causes a circular reference.
-		if (this.getDocType() !== 'presentation') {
+		if (this.isPresentationOrDrawing()) {
```
tzolnai, line 282: Here you rewrote this incorrectly. It should be `if (!this.isPresentationOrDrawing()) {` I assume.

loleaflet/src/control/Parts.js
```diff
@@ -298,14 +298,14 @@ L.Map.include({
...
-		if (this.getDocType() !== 'presentation') {
+		if (!this.isPresentationOrDrawing()) {
 			return;
 		}
 		this._socket.sendMessage('uno .uno:DuplicatePage');
 		var docLayer = this._docLayer;
 
 		// At least for Impress, we should not fire this. It causes a circular reference.
-		if (this.getDocType() !== 'presentation') {
+		if (this.isPresentationOrDrawing()) {
```
tzolnai, line 308: Here too.

loleaflet/src/control/Parts.js
```diff
@@ -345,7 +345,7 @@ L.Map.include({
 		}
 
 		// At least for Impress, we should not fire this. It causes a circular reference.
-		if (this.getDocType() !== 'presentation') {
+		if (this.isPresentationOrDrawing()) {
```
tzolnai, line 348: And here.

tzolnai (changes requested):
I marked some places, where I think the condition was changed incorrectly. I think that's why some cypress tests are failing on this PR. The test cases related to adding / removing slides are crashing because of this.

loleaflet/src/control/Parts.js
kendy: @tzolnai: Thanks for that, @merttumer has fixed that in an additional commit.

kendy:
I have no idea how to get rid of the "Changes requested", but the changes were done, so I'm just mering this.  @tzolnai , thanks again for your help!

tzolnai (approved):
Cypress tests pass now, so PR is OK for me.

PR #1239, opened 2021-01-14
wsd: asynchronous HTTP/1.1 implementation

Ashod:
This implements HTTP/1.1 per RFC 7230, partially.

Unit-tests are provided with documentation on usage.

This is desgined to serve as the http implementation
throughout loolwsd, for both synchronous and
asynchronous requests.

Change-Id: Iaf1b8c5fcb8cec032445e27c9f70d2fb807aa4dc

### DIFF
--- a/test/HttpRequestTests.cpp
+++ b/test/HttpRequestTests.cpp
@@ -1,0 +1,371 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <chrono>
+#include <condition_variable>
+#include <config.h>
+
+#include <Poco/Net/HTTPClientSession.h>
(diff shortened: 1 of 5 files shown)


kendy:
@Ashod : Thanks for this! :-)

test/HttpRequestTests.cpp
```diff
@@ -0,0 +1,371 @@
...
+    httpSession->asyncRequest(httpRequest, pollThread);
+
+    std::unique_lock<std::mutex> lock(mutex);
+    cv.wait_for(lock, std::chrono::seconds(1));
+
+    const std::shared_ptr<const http::Response> httpResponse = httpSession->response();
+    LOK_ASSERT(httpResponse->state() == http::Response::State::Complete);
+    LOK_ASSERT(!httpResponse->statusLine().httpVersion().empty());
+    LOK_ASSERT(!httpResponse->statusLine().reasonPhrase().empty());
+    LOK_ASSERT(httpResponse->statusLine().statusCode() == 200);
```
vmiklos, line 287: Hmm, this worked yesterday, but now it returns 403. Do you think you could hide these tests (that rely on httpbin.org accepting our requests) behind some flag, so tests don't start failing just because an external service is down? At least both me and @tml1024 get failures now. Thanks.

tml1024:
Also the example.com tests seem a bit fragile.

PR #1242, opened 2021-01-14
make run: don't enable ssl by default

vmiklos:
This has the following advantages:

1) Now make run uses the non-ssl codepath by default, which matches the
non-ssl + reverse proxy setup that is recommended in production.

2) This way editing loolwsd.xml makes it possible to opt-in for SSL. (It
doesn't work the other way around.)

It has the disadvantage that browser gets http by default, and certain
js errors only show up with https.

Overall this is meant to improve the default developer experience.

Change-Id: If5567a95753d6191bdb8af6ed649f64ed5638078


vmiklos:
This is a follow-up. My understanding was that moving away from SSL by default is wanted, but the way #1210 affected make run was a step backwards, as pointed out by @timar . This PR is meant to fix that aspect.

vmiklos:
See c43562c9a765fd6f667e0101ea818f78155702c6, this is not a good idea after all.

PR #1247, opened 2021-01-14
Fix crash in iOS app: Don't bother with UnitKit

tml1024:
I don't recall if simply ifdeffing is the proper way to bypass UnitKit
stuff (that I don't understand) for the mobile apps, but at least it
helps.

Signed-off-by: Tor Lillqvist <tml@iki.fi>
Change-Id: I85d477e4ee7d11c597d95a9b0b42af7e5f7ad122

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -2404,9 +2404,10 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
     LOG_TRC("ChildSession::loKitCallback [" << getName() << "]: " <<
             typeName << " [" << payload << "].");
 
+#if !MOBILEAPP
     if (UnitKit::get().filterLoKitCallback(type, payload))
         return;
-
+#endif
     if (isCloseFrame())
     {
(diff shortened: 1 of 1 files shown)


tml1024:
Fixes a crash for one trivial text document in the iOS app that then shows up normally. (But other documents fail to show up, because of other problems.)

mmeeks (commented):
Hmm:

/// LOKit (and some synthetic internal) callbacks
common/Unit.hpp:    virtual bool filterLoKitCallback(const int /* type */, const std::string& /* payload */)
common/Unit.hpp-    {
common/Unit.hpp-        return false;
common/Unit.hpp-    }

That crashes ? I'm surprised.

mmeeks:
Lots of other call-sites have:

#ifndef KIT_IN_PROCESS

around the UnitKit:: thing - which seems wasteful, and noisy for #ifdefs - I would suggest we just compile in a dummy UnitKit for the mobile case, and if we fear performance foo add some 'final's there to in-line and evaporate those.

Anyhow - seems fine =)

PR #1260, opened 2021-01-16
Special case opacity to avoid arithmetic where possible.

mmeeks:
Change-Id: I37a72bc966424ca1b4c30b31d9fcf88d15c11789

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Png.hpp
+++ b/common/Png.hpp
@@ -92,20 +92,28 @@ unpremultiply_data (png_structp /*png*/, png_row_infop row_info, png_bytep data)
     for (i = 0; i < row_info->rowbytes; i += 4)
     {
         uint8_t *b = &data[i];
-        uint32_t pixel;
+        uint32_t pix;
         uint8_t  alpha;
 
-        std::memcpy (&pixel, b, sizeof (uint32_t));
-        alpha = (pixel & 0xff000000) >> 24;
-        if (alpha == 0)
+        std::memcpy (&pix, b, sizeof (uint32_t));
+
+        alpha = (pix & 0xff000000) >> 24;
+        if (alpha == 255)
+        {
+            b[0] = ((pix & 0xff0000) >> 16);
+            b[1] = ((pix & 0x00ff00) >>  8);
+            b[2] = ((pix & 0x0000ff) >>  0);
+            b[3] = 255;
+        }
+        else if (alpha == 0)
         {
             b[0] = b[1] = b[2] = b[3] = 0;
         }
         else
         {
-            b[0] = (((pixel & 0xff0000) >> 16) * 255 + alpha / 2) / alpha;
-            b[1] = (((pixel & 0x00ff00) >>  8) * 255 + alpha / 2) / alpha;
-            b[2] = (((pixel & 0x0000ff) >>  0) * 255 + alpha / 2) / alpha;
+            b[0] = (((pix & 0xff0000) >> 16) * 255 + alpha / 2) / alpha;
+            b[1] = (((pix & 0x00ff00) >>  8) * 255 + alpha / 2) / alpha;
+            b[2] = (((pix & 0x0000ff) >>  0) * 255 + alpha / 2) / alpha;
             b[3] = alpha;
         }
     }


common/Png.hpp
```diff
@@ -92,20 +92,28 @@ unpremultiply_data (png_structp /*png*/, png_row_infop row_info, png_bytep data)
...
-        alpha = (pixel & 0xff000000) >> 24;
-        if (alpha == 0)
+        std::memcpy (&pix, b, sizeof (uint32_t));
+
+        alpha = (pix & 0xff000000) >> 24;
+        if (alpha == 255)
+        {
+            b[0] = ((pix & 0xff0000) >> 16);
+            b[1] = ((pix & 0x00ff00) >>  8);
+            b[2] = ((pix & 0x0000ff) >>  0);
```
Ashod, line 105: I would imagine 'shit and truncate' to be the more natural form, because the `&` is done way with in the `mov` byte instruction (it's only there to avoid warnings of size-mismatch implicit casting) ...

`b[0] = ((pix >> 16) & 0xff);`
`b[1] = ((pix >> 8) & 0xff);`
`b[2] = (pix & 0xff);`

Although I imagine a good compiler will generate the same code (but isn't guaranteed), anyway, technically this is the more explicit form for what the code is doing (copying a specific byte from the int) and it almost certainly will produce the most optimal code.

Notice that the code below is not using this form because we don't narrow from int to char. So the below code is correct.

common/Png.hpp
```diff
@@ -92,20 +92,28 @@ unpremultiply_data (png_structp /*png*/, png_row_infop row_info, png_bytep data)
...
-        uint32_t pixel;
+        uint32_t pix;
         uint8_t  alpha;
 
-        std::memcpy (&pixel, b, sizeof (uint32_t));
-        alpha = (pixel & 0xff000000) >> 24;
-        if (alpha == 0)
+        std::memcpy (&pix, b, sizeof (uint32_t));
+
+        alpha = (pix & 0xff000000) >> 24;
```
Ashod, line 100: Same comment as for the color bytes.

`alpha = (pix >> 24) & 0xff;`

common/Png.hpp
```diff
@@ -92,20 +92,28 @@ unpremultiply_data (png_structp /*png*/, png_row_infop row_info, png_bytep data)
...
             b[0] = b[1] = b[2] = b[3] = 0;
         }
         else
         {
-            b[0] = (((pixel & 0xff0000) >> 16) * 255 + alpha / 2) / alpha;
-            b[1] = (((pixel & 0x00ff00) >>  8) * 255 + alpha / 2) / alpha;
-            b[2] = (((pixel & 0x0000ff) >>  0) * 255 + alpha / 2) / alpha;
+            b[0] = (((pix & 0xff0000) >> 16) * 255 + alpha / 2) / alpha;
+            b[1] = (((pix & 0x00ff00) >>  8) * 255 + alpha / 2) / alpha;
+            b[2] = (((pix & 0x0000ff) >>  0) * 255 + alpha / 2) / alpha;
```
Ashod, line 116: While the `+ alpha/2` is technically correct (for proper rounding), I wonder if it's worth paying the cost. I mean, who/what can detect 1/256 difference on the screen (in the worst case)? And who cares? Indeed, most all screens (except for the prosumer-grade) have 6-bit pixels anyway, and can't display anything below 1/64 delta.

Does this rounding add any visible value?

Ashod (commented):
Good stuff. I've reviewed the whole code and left some notes.

As Kendy is the author of this as much as anyone else, I think it's worth requesting his participation. I'll add him.

common/Png.hpp
Ashod: Another possible improvement is to replace the `* 255` with `* 256 - x`. 

```
uint8_t val = (pix >> 16) & 0xff;
b[0] = (val * 256 - val) / alpha; // The compiler will generate `<< 8` for the multiply. This is just more readable.
```

While most modern CPUs (e.g. Zen and Haswel or newer) can multiply in a few cycles (typical latency is 3-4 cycles), there is only *one* multiply (and divide) ALU per core, but there are at least *three* ALUs that can do shifts and add/sub.

Not only do we win by doing all 3 pixel calculations in parallel, but we free the multiply/divide ALU to do the final `/ alpha`.

So, I expect by replacing the `* 255` with this proposal we will shave off quite a bit of the latency of this code (even if we leave the `+ alpha/2`, which as I say above seems superfluous`. 

Of course doing a quick `perf` run on my proposal is both desirable and reasonably easy.

mmeeks: Well correctness is helpful, and that's the correct way to round the alpha - it's also what was there before. Cosmetics of shift order - I agree with you, I'd personally optimize that myself, but its not my code =)

mmeeks: As for performance - that code path is now almost never hit - so there is little point in investing more programming time and review time optimizing it. We either have 0 alpha - or 255 and a very small transition around the edge of the page border I expect. I'd like to get this in fast, un-molested - particularly since the remaining code doesn't show on the profile - and move onto the next problem.

Ashod (approved):
Let's get it in.

kendy (approved):
So I think we want to get the obvious simple case in ASAP; and the patch does the right thing, so let's get it there :-)

@ashod: For your comments, you can file Easy Hacks as follow-ups maybe?

tml1024:
One thing I don't get in this patch is why "pixel" had to be renamed to "pix"... Sorry.

PR #1276, opened 2021-01-19
Switch to the correct view before calling getSelectionType() method.

tzolnai:
This method needs to be called with having the correct view activated.

It caused a crash in writer/simultaneous_typing multi-user cypress test.

Signed-off-by: Tamás Zolnai <tamas.zolnai@collabora.com>
Change-Id: I57dd58a635ce7ea127200a6a3f0cadc8c0861aaf
(cherry picked from commit 218e47e65a5e59871b9670e4bad020cf389d17fc)


mmeeks:
Good catch & backport - thanks ! =) +1 from me etc.

PR #1280, opened 2021-01-20
leaflet: blacklist language selection on misspelled words in mobile

lpranam:
Change-Id: I149539983897d20828dc6e32ed053c9b7e279ad4

* Target version: master 

### Checklist


tzolnai (commented):
Better to remove those cypress tests. They are not needed anymore if this feature is disabled.

PR #1288, opened 2021-01-20
Example how easy an tabbed NB will switch to tabbed compact NB

Andreas-Kainz:
Change-Id: Icd3b2f90b35921052267c4de5444ae3485cc1a54

How easy you convert an tabbed notebookbar into tabbed compact notebookbar you will see in this example PR (I wan't push it to master). And as the idea is to reuse the JSON notebookbar stuff, you can reuse even more like all the context toolbars will have the same commands than the context tab.

**Tabbed**
Screenshot_20210120_163316

**Tabbed Compact**
Screenshot_20210120_163953

### DIFF
--- a/loleaflet/src/control/Control.NotebookbarWriter.js
+++ b/loleaflet/src/control/Control.NotebookbarWriter.js
@@ -140,10 +140,11 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 							'type': 'toolbox',
 							'text': '',
 							'enabled': 'true',
+							'vertical': 'false',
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
4 values changed between tabbed and tabbed compact. That's great and simple!

eszkadev:
yes, we can use this trick (automaticallly, programmed in the code to make it 'one line').
thanks, when I will rework notebookbar to JSON i will try to do something like this as a variant.

I assume this is only to show the mechanism right? can we close this PR?

PR #1289, opened 2021-01-20
Fix build with libc++: std::min() parameters should be of the same type.

arrowd:
Change-Id: Icdb67f0576992cd54c8e15c85fbaf2745aba218c

### DIFF
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -426,7 +426,7 @@ public:
             const auto timeSincePingMicroS
                 = std::chrono::duration_cast<std::chrono::microseconds>(now - _lastPingSentTime);
             timeoutMaxMicroS
-                = std::min(timeoutMaxMicroS, (PingFrequencyMicroS - timeSincePingMicroS).count());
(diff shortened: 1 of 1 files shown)


arrowd:
Bump.

arrowd:
Another bump.

mrkara:
@arrowd Thank you for your contribution! Looking forward to your next PR. ;)

Ashod (commented):
Thanks @arrowd for the patch. I wonder if we should remove the `.count()`. Can you please try and see if it compiles (and without the cast, of course) ?

arrowd:
Nope, doesn't work too:

```
./net/WebSocketHandler.hpp:436:19: error: no matching function for call to 'min'
                = std::min(timeoutMaxMicroS, PingFrequencyMicroS - timeSincePingMicroS);
                  ^~~~~~~~
/usr/include/c++/v1/algorithm:2560:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('long' vs. 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> >')
min(const _Tp& __a, const _Tp& __b)
```

Ashod:
> Nope, doesn't work too:

Thanks for trying!

PR #1303, opened 2021-01-22
Make table markers clickable on desktop

pedropintosilva:
- Currently the markers are not touchable on ipad (both via web and app) so
- Revert back listener change introduce in ebaca16c4f28ae7a499c632b70b24eaaa1083439

Change-Id: Ia2ff20ed4ddf810c50b11c49600709489f07785f


tml1024:
Does not seem to cause any regression in the iOS app.

tml1024 (approved):
Does not seem to cause any regression in the iOS app, OK with me.

PR #1310, opened 2021-01-25
Document State

Ashod:
This PR introduces a new DocumentStatus class to encapsulate the document lifecycle states
and the related members to track it.

- wsd: introduce DocumentState class
- wsd: move the closeRequested flag into DocumentStatus
- wsd: move the markToDestroy flag into DocumentStatus
- wsd: move the loaded flag into DocumentStatus

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -174,7 +174,6 @@ DocumentBroker::DocumentBroker(ChildType type,
     _lastStorageUploadSuccessful(true),
     _lastSaveTime(std::chrono::steady_clock::now()),
     _lastSaveRequestTime(std::chrono::steady_clock::now() - std::chrono::milliseconds(COMMAND_TIMEOUT_MS)),
-    _markToDestroy(false),
     _isLoaded(false),
     _isModified(false),
     _cursorPosX(0),
@@ -451,7 +450,7 @@ void DocumentBroker::pollThread()
             }
         }
 #endif
-        else if (_sessions.empty() && (isLoaded() || _markToDestroy))
+        else if (_sessions.empty() && (isLoaded() || _docState.isMarkedToDestroy()))
         {
             // If all sessions have been removed, no reason to linger.
             LOG_INF("Terminating dead DocumentBroker for docKey [" << getDocKey() << "].");
@@ -574,7 +573,7 @@ bool DocumentBroker::download(const std::shared_ptr<ClientSession>& session, con
             return result;
     }
 
-    if (_markToDestroy)
+    if (_docState.isMarkedToDestroy())
     {
         // Tearing down.
         LOG_WRN("Will not load document marked to destroy. DocKey: [" << _docKey << "].");
@@ -980,11 +979,11 @@ void DocumentBroker::uploadToStorage(const std::string& sessionId, bool success,
 
     // If marked to destroy, or session is disconnected, remove.
     const auto it = _sessions.find(sessionId);
-    if (_markToDestroy || (it != _sessions.end() && it->second->isCloseFrame()))
+    if (_docState.isMarkedToDestroy() || (it != _sessions.end() && it->second->isCloseFrame()))
         disconnectSessionInternal(sessionId);
 
     // If marked to destroy, then this was the last session.
-    if (_markToDestroy || _sessions.empty())
+    if (_docState.isMarkedToDestroy() || _sessions.empty())
     {
         // Stop so we get cleaned up and removed.
         _stop = true;
@@ -1461,7 +1460,7 @@ std::size_t DocumentBroker::addSession(const std::shared_ptr<ClientSession>& ses
         if (_sessions.empty())
         {
             LOG_INF("Doc [" << _docKey << "] has no more sessions. Marking to destroy.");
-            _markToDestroy = true;
+            _docState.markToDestroy();
         }
 
         throw;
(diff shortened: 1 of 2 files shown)


wsd/DocumentBroker.hpp
```diff
@@ -433,6 +433,64 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
...
+    /// This regulates all other primary operations.
+    class DocumentState
+    {
+    public:
+        /// Strictly speaking, these are phases that are directional.
+        /// A document starts as New and progresses towards Unloaded.
+        /// Upon error, intermediary states may be skipped.
+        enum class Status
+        {
+            New, //< Doesn't exist, pending downloading.
```
kendy, line 446: I wonder if None instead of New, Destroying instead of Unloading and Destroyed instead of Unloaded wouldn't be a bit more clear (?)

wsd/DocumentBroker.hpp
```diff
@@ -484,8 +485,13 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
             _status = newStatus;
         }
 
+        /// Flag close-requested. Cannot be reset.
+        void setCloseRequested() { _closeRequested = true; }
+        bool isCloseRequested() const { return _closeRequested; }
+
     private:
         Status _status;
+        bool _closeRequested; //< Owner-Termination flag.
```
kendy, line 515: Not std::atomic any more (?)

wsd/DocumentBroker.cpp
```diff
@@ -1539,7 +1538,10 @@ std::size_t DocumentBroker::removeSession(const std::string& id)
         std::shared_ptr<ClientSession> session = it->second;
 
         // Last view going away, can destroy.
-        _markToDestroy = (_sessions.size() <= 1);
+        if (_sessions.size() <= 1)
+            _docState.markToDestroy();
+        else
+            assert(!_docState.isMarkedToDestroy());
```
kendy, line 1543: I am not fan at all of a pattern like "if (x) { blah(); } else assert(...) ;-)

Let's rather do a clear "assert(condition && other condiotion)" that is not a lone thing in an else branch please...

wsd/DocumentBroker.hpp
```diff
@@ -485,7 +485,14 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
             _status = newStatus;
         }
 
-        /// Flag close-requested. Cannot be reset.
+        /// True iff the document is fully loaded and available for viewing/editing.
+        bool isLive() const { return _status == Status::Live; }
```
kendy, line 493: Not used anywhere, will that be used later?

wsd/DocumentBroker.hpp
```diff
@@ -499,6 +513,7 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
     private:
         Status _status;
         bool _closeRequested; //< Owner-Termination flag.
+        bool _loaded; //< If the document ever loaded (check isLive to see if it still is).
```
kendy, line 516: Again not std::atomic any more - is it not needed?

wsd/DocumentBroker.hpp
Ashod: Yes, those were the other options. Happy to adopt them.

wsd/DocumentBroker.hpp
Ashod: This is set from ClientSession, which is invoked from the poll. There are no other threads to change this value (it's also set only, never reset). So my expectation is that there is no more concern with threading or visibility across threads.

Will triple check to make sure.

wsd/DocumentBroker.cpp
Ashod: Yes, I suppose that could be easier to read in some cases. Let me change it.

wsd/DocumentBroker.hpp
Ashod: Right. The poll function will be simplified to check for specific states. isLive() is the steady-state, for most of the normal operation and !isLive() will be for the cases we do at startup and at shutdown of the document.

wsd/DocumentBroker.hpp
Ashod: Same as _closeRequested, _loaded is set when we get the "status: " message from Kit. We read it from the socket in poll and it's a set-only (no resetting). I don't see any thread contentions on these two variables, but I'll triple check to make sure we don't introduce any regressions.

wsd/DocumentBroker.hpp
Ashod: I restored atomic because this is invoked from Admin and also to make this as identical to the original code as possible.

wsd/DocumentBroker.hpp
Ashod: Restored atomic for the same reason as for _closeRequested.

wsd/DocumentBroker.hpp
```diff
@@ -491,8 +492,13 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
             _status = status;
         }
 
+        /// Flag close-requested. Cannot be reset.
+        void setCloseRequested() { _closeRequested = true; }
+        bool isCloseRequested() const { return _closeRequested; }
```
kendy, line 518: Shouldn't we actually then have const std::atomic<bool>& as the return value here?

Ashod: That won't be necessary, because these flags change only once (at most) in the lifetime of DocBroker, and we don't repeatedly check the value to see they are changed. Alternatively, we call this function every time we need the latest value, so, again, no need to take a reference: the next call is guaranteed to give us the last value.

PR #1315, opened 2021-01-25
Bring kubernetes proof-of-concept setup with just collabora-online

yunusem:
Change-Id: I5daeeea3746bb107b3c4e0b3b68efa0adacb773f

* Target version: master 

### Checklist


kubernetes/helm/README.md
```diff
@@ -1,20 +1,50 @@
...
+example output :
+```
+|-----------|------------------|-------------|------------------------|
+| NAMESPACE |       NAME       | TARGET PORT |          URL           |
+|-----------|------------------|-------------|------------------------|
+| collabora | collabora-online |             | http://127.0.0.1:55064 |
+|-----------|------------------|-------------|------------------------|
+http://127.0.0.1:55064
+```
+so for this example http:// 127.0.0.1:55064 address should only response as "OK" on a plain html page, it means that it is ready to be used by nextcloud or owncloud of your local setup
```
timar, line 46: it is ready to be used by your WOPI-like host of your local setup

kubernetes/helm/collabora-online/values.yaml
```diff
@@ -1,10 +1,10 @@
-# Default values for libreoffice-online.
+# Default values for collabora-online.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 global:
   app:
-    name: libreoffice-online
-    version: 6.5.0
+    name: collabora-online
+    version: 6.4.4
```
timar, line 7: Do we have to change version number here always, when we bump version of the release?

kubernetes/helm/collabora-online/values.yaml
```diff
@@ -18,25 +18,26 @@ global:
...
+  tag: latest
 
 probes:
   urlScheme: HTTP # either HTTP or HTTPS
   port: 9980
   readiness:
     activate: true
     path: /
-    initialDelaySeconds: 180
+    initialDelaySeconds: 45
```
timar, line 40: why 180 -> 45 ?

kubernetes/helm/collabora-online/values.yaml
```diff
@@ -61,17 +62,17 @@ environmentVariablesLool:
   - name: extra_params
     value: --o:ssl.enable=false
   - name: username
-    value: 
+    value: peter
```
timar, line 65: default user/password? I'd not use them.

timar (changes requested):
See inline comments

kubernetes/helm/collabora-online/values.yaml
yunusem: I don't think so, I will test the setup without ^this version specification.

kubernetes/helm/collabora-online/values.yaml
yunusem: well I thought 3 minutes would be too much to wait just for loolwsd service to initialize

kubernetes/helm/collabora-online/values.yaml
yunusem: i will remove them

kubernetes/helm/collabora-online/values.yaml
yunusem: and in fact it was while testing

timar (approved):

kubernetes/helm/collabora-online/values.yaml
```diff
@@ -18,17 +17,18 @@ global:
...
-      name: libreoffice-online
-      type: ClusterIP
+      name: collabora-online
+      type: NodePort # Use ClusterIP instead of NodePort if you are using standalone cluster structure 
       externalPort: 9980
       internalPort: 9980
   ingress: {}
 
-replicaCount: 3
+replicaCount: 3 # This is the amount of how many instances of loolwsd will run
```
fabianhick, line 26: How do three replicas work properly? 
I thought the application is stateful when multiple users are editing the same file simultaneously?

PR #1316, opened 2021-01-25
Add caching of tunnelled dialog images

tml1024:
The same cache size is used in server and client. The caches use the
invalidation algorithm. Pass the hash of the pixmap in the
windowpaint: message. The client stores dialog images in its cache.
The server stores hashes of the images. When the server knows that the
client already has an image cached, it sends just its hash and the
client will use the cached image.

Pass the size of the cache to the client so that we don't have to keep
the the cache size synchronised in two places in the code.

Change-Id: Ie6cbfca79a9ede48fcc115e3acc669b925bb624e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1665,13 +1665,60 @@ bool ChildSession::renderWindow(const char* /*buffer*/, int /*length*/, const St
             << " and rendered in " << totalTime
             << "ms (" << area / elapsed << " MP/s).");
 
+    uint64_t pixmapHash = Png::hashSubBuffer(pixmap.data(), 0, 0, width, height, bufferWidth, bufferHeight);
+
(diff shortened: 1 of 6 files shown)


mmeeks (approved):
Code looks great to me Tor - thanks =) I think we generally have a convention of CamelCase for global variables in the C++ though rather than foo_baa_baz, perhaps best to be consistent there - but a minor nit.
Thanks !

tml1024:
> I think we generally have a convention of CamelCase for global variables in the C++ though rather than foo_baa_baz, 

You are right, it was accidentally left that way because I originally added it to the configuration thingie (where the names are in_that_form), but it seemed hard to get at those values from the kit process. Will submit a follow-up rename.

PR #1351, opened 2021-01-28
Created cypress test for the Clone Formatting feature on the Calc top…

Ezinnem:
…-toolbar

Signed-off-by: Ezinne <nnamani.ezinne@collabora.com>
Change-Id: Ic4a41c20e07672226f3296a6557a32a1e9afb2c2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tzolnai (changes requested):
You should check what this test does in interactive test runner:
`make run-desktop`
First, it does not select the first character of a word as the comment suggests and from that point, other steps also do different things that the comments say and what make sense. You should try to make sure the code you copy&paste-d from other places actually does, what is the intention. It's not enough if a test passes.

Issue #1322, opened 2021-01-25
iOS App: Impress: Sidebar is empty


smirta:
**Describe the bug**
Sidebar is empty. Select and de-select the sidebar doesn't change anything.

**To Reproduce**
1. Open/Create a presentation
2. Select sidebar top right

**Expected behavior**
Sidebar with tools appears.

**Actual behavior**
Sidebar appears but is empty.

**Screenshots**
![image](https://user-images.githubusercontent.com/6649852/105742401-c0f62c80-5f3b-11eb-8a20-7ad6f47dd157.png)

**iPad Pro**
 - iPad OS: 14.0.1
 - Collabora Office 6.4.5

thebearon:
Thanks for reporting, this is a regression from the following commit: d37bad53f6258a1ad0bcf75bb45cdc85cfa477ec
@pedropintosilva, what are your thoughts?

thebearon:
This is fixed in in build 6.4.5 (2). Thanks, @pedropintosilva!

smirta:
I can confirm this as resolved in 6.4.5 (2). Thanks, guys!

PR #1353, opened 2021-01-28
Sidebar: panelContainer is being set without height #1322

pedropintosilva:
Change-Id: I2c6421b34cd6c2345981008dd1bbee494bc13ca2

### DIFF
--- a/loleaflet/src/control/Control.LokDialog.js
+++ b/loleaflet/src/control/Control.LokDialog.js
@@ -1110,6 +1110,7 @@ L.Control.LokDialog = L.Control.extend({
 
 		var panelContainer = L.DomUtil.create('div', 'panel', L.DomUtil.get('sidebar-panel'));
 		panelContainer.id = strId;
+		this._setCanvasWidthHeight(panelContainer, width, height);
 
 		// Create the panel canvas.
 		var panelCanvas = L.DomUtil.create('canvas', 'panel_canvas', panelContainer);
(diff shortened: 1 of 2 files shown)


loleaflet/src/control/Control.LokDialog.js
```diff
@@ -765,10 +765,14 @@ L.Control.LokDialog = L.Control.extend({
 		}
 
 		var newHeight = height * scale;
+		// console.log('newHeight = height * scale: ' + newHeight + ' = ' + height + ' * ' + scale);
+		// console.log('if (canvas.height != newHeight) : if(' + canvas.height + ' != ' + newHeight + ')');
+		// console.debug(canvas);
```
thebearon, line 770: You probably didn't mean to commit commented out debug code. ;)

Otherwise, the patch works well for me, thanks!

pedropintosilva: Well It was I who added those in the first place, I just thought it would be better to commented them out and if someone wants to check those values can always uncommented but on the 2nd thought I will be removing those

thebearon: Yes, that's what I meant, it was added with the commit, and was added as commented out. I think providing hints for debugging is a great idea, the way I would probably consider doing it is in a more descriptive and explanatory way rather than raw debug code. But those are just my two cents, and removing the code is good, too, thanks for updating the pull request!

PR #1359, opened 2021-01-28
In Cypress tests throw error if dialog image cache is inconsistent wr…

tml1024:
…t server

Change-Id: I439263c7923d985679fafa0125cd2d7f635e0995

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tzolnai:
Hi Tor,

Two things.
1. As I see the test document has the selected cell at A22. It will cause the test to fail sometimes, because Collabora Online sometimes jumps to the cell selection by document load. If it jumps to A22, then `calcHelper.clickOnFirstCell();` selects A11 cell instead of the intended A1 cell, because that is at the top-left corner. So better to select A2 cell for example and save that to the test document, so the view won't jump.
2. Cypress can be very fast. So if you just push ENTER here 30 times it might happen that it happens very fast and core might not actually send that many sidebar updates. So I think it might be better to slow down the test a bit. In general, I would suggest checking somehow that the sidebar is updated after every enter, but since the sidebar is just an image, it's hard to test, so a `cy.wait(1000)` can be also good. It adds 1 sec waiting time after every step, so the sidebar has time to be updated.

tml1024:
@tzolnai Thanks for the comment, did what you suggested. I used 200 ms for the wait time, hopefully that is enough. At least interactively it was.

tzolnai:
> @tzolnai Thanks for the comment, did what you suggested. I used 200 ms for the wait time, hopefully that is enough. At least interactively it was.

Well, interactive test runner is much slower than the headless run, which is called by `make check`. You can use this:
`ENABLE_VIDEO_REC="1" make check-desktop spec=calc/sidebar_dialog_image_caching_spec.js`
to enable video recording and so cypress will create a video of the headless run. It's much closer to the headless build wrt. timing. The video is created under cypress_test/cypress/videos folder.

tml1024:
Ah, good point. Will experiment with that.

tml1024:
The video did show the sidebar updating at each step, but I increased the time to 500 ms anyway to be on the safe side. Ideally there could be some way to signal the Cypress JS code that the sidebar has updated from the Online JS code.

tzolnai (approved):
Looks good to me now.
I think, in the long term, the plan is to have the sidebar built by JS code (similar to mobile wizard), instead of getting images of the sidebar from core. If that happens it would be very easy to check the state of the sidebar from cypress test. For now, waiting a constant time is good enough.

PR #1369, opened 2021-01-29
Add support for split APKs

evelikov:
* Resolves: # 
* Target version: master 

### Summary

Currently GooglePlay builds are nicely optimised, as they require bundles which they split and optimise. At the same time, [standalone](https://www.collaboraoffice.com/downloads/Collabora-Office-Android-Snapshot/) or [fdroid](https://www.collaboraoffice.com/releases-en/collabora-office-on-mobiles-supporting-password-protected-documents-and-available-on-f-droid/)  builds are huge since they include all supported ABIs in the same APK. Essentially wasting bandwidth and device storate.

Add support for split APKs to address this issue.

### TODO

- [x] Test resulting APKs
- [ ] The fdroid repo will need a tweak to match the new APK names

### Checklist


evelikov:
Going through the checklist:
 - Change-Id - how does one do this with Github?
 - `make check` and `make run` - haven't built or tested the APK since i'm short on CPU cycles
 - Documentation - from quick search I didn't see anything relevant in-tree

kendy:
@evelikov : Hi Emil, thanks a lot for your patch!  Regarding the checklist:

> Going through the checklist:
> 
>     * Change-Id - how does one do this with Github?

Normally it is created via git hooks that are normally installed when you ./configure; but you can trigger their installation manually via ./scripts/refresh-git-hooks

And then just git commit --amend should hopefully do it :-)

>     * `make check` and `make run` - haven't built or tested the APK since i'm short on CPU cycles

Hopefully the CI will do this for you, no problem.

>     * Documentation - from quick search I didn't see anything relevant in-tree

Yep, don't worry about this either.

android/app/build.gradle
```diff
@@ -26,8 +26,9 @@ android {
             enable true
 
             reset ()
-            include = []
-            include.addAll("${liboAndroidAbi}".split(' ').collect{it as String})
+            def includedAbis = []
+            includedAbis.addAll("${liboAndroidAbi}".split(' ').collect{it as String})
+            include includedAbis
```
kendy, line 31: If it's just a fixup, best to git rebase -i --autosquash, and force-push :-)

evelikov: Agreed. I'm aiming to minimise the force-pushes.

evelikov:
Thanks for the tips and reassurance Jan.

For anyone wondering, I'm having a wrestle with gradle - it doesn't like the dynamically generated include list.
I am trying that locally and once that is working, I'll update the MR - don't want to churn CI cycles for nothing ;-)

mrkara:
@evelikov Hi! Do you need help on this, or have any questions?

FYI: We had a short period of time in the last weeks with CI failing for all changes. So if CI also fails for your change, with some error unrelated to your change, rebasing on top of master might help.

evelikov:
@mrkara it took me a while since a couple of issues came up:
 - CPU architecture variables, need particular quoting which was tricky to set due to M4 quoting rules
 - Similar commit was reverted for another Android app - turns out, the fdroid automatic updater (which the team does not use), doesn't work with split apks

With the above resolved, I'm just fixing up the final cmake glitches.

evelikov:
Woohoo it's done \o/

We massage `ANDROID_ABI` and shove it in the Android `build.gradle` files. This allows for per ABI splitting - four APKs, each roughly 1/4 of the original size. Hence less bandwidth is used and nearly¹ no wasted space on consumer devices.

Minor annoyance - one needs to copy four files instead of one and ensure all four are in the custom fdroid repo manifest. Pretty sure that can be automated if needed.

[1] There might still be a bit of "waste" since non CPU ABI artefacts, think resources for older/newer Android version, are still part of the APK.

mrkara:
@evelikov Great! :)

Yet a cypress unit test is failing. Could you please rebase your changes on top of current master in case it is just a bad commit unrelated to your change?

Please let me know if you need help with rebasing. ;)

evelikov:
Done:
 - rebased on top of master
 - tested the armv8 APK
 - `make check` and `make run` were executed by the CI - plus MR does nothing to affect those

More than happy to help with the fdroid repo (aka tweak to match the new names) or Android bundle for the PlayStore.
I couldn't see the tooling for those in this git repo.

PR #1374, opened 2021-02-01
Save and Storage Managers

Ashod:
This introduces DocBroker's SaveManager and StorageManager, internal classes to manage the save and upload states.

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -426,7 +426,7 @@ void DocumentBroker::pollThread()
         if (isLoaded() && getIdleTimeSecs() >= IdleDocTimeoutSecs)
         {
             // Don't hammer on saving.
-            if (_saveManager.timeSinceLastSaveRequest() >= std::chrono::seconds(5))
+            if (_saveManager.timeSinceLastRequest() >= std::chrono::seconds(5))
             {
                 // Stop if there is nothing to save.
                 LOG_INF("Autosaving idle DocumentBroker for docKey [" << getDocKey()
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -173,7 +173,6 @@ DocumentBroker::DocumentBroker(ChildType type,
     _documentChangedInStorage(false),
     _lastStorageUploadSuccessful(true),
     _lastSaveTime(std::chrono::steady_clock::now()),
-    _lastSaveRequestTime(std::chrono::steady_clock::now() - std::chrono::milliseconds(COMMAND_TIMEOUT_MS)),
     _isModified(false),
     _cursorPosX(0),
     _cursorPosY(0),
@@ -380,9 +379,8 @@ void DocumentBroker::pollThread()
             refreshLock();
 #endif
 
-        if (isSaving() &&
-            std::chrono::duration_cast<std::chrono::milliseconds>
-                    (now - _lastSaveRequestTime).count() <= COMMAND_TIMEOUT_MS)
+        //TODO: Review if we need this here.
+        if (_saveManager.isSaving() && !_saveManager.hasSavingTimedOut())
         {
             // We are saving, nothing more to do but wait (until we save or we timeout).
             continue;
@@ -430,8 +428,7 @@ void DocumentBroker::pollThread()
         if (isLoaded() && getIdleTimeSecs() >= IdleDocTimeoutSecs)
         {
             // Don't hammer on saving.
-            if (std::chrono::duration_cast<std::chrono::seconds>(now - _lastSaveRequestTime).count()
-                >= 5)
+            if (_saveManager.timeSinceLastSaveRequest() >= std::chrono::seconds(5))
             {
                 // Stop if there is nothing to save.
                 LOG_INF("Autosaving idle DocumentBroker for docKey [" << getDocKey()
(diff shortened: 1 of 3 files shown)


wsd/DocumentBroker.hpp
```diff
@@ -540,6 +540,46 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
...
+            : _lastAutosaveCheckTime(std::chrono::steady_clock::now())
+            , _isAutosaveEnabled(std::getenv("LOOL_NO_AUTOSAVE") == nullptr)
+        {
+        }
+
+        /// Return true iff auto save is enabled.
+        bool isAutosaveEnabled() const { return _isAutosaveEnabled; }
+
+        /// Returns true if we should issue an auto-save.
+        bool needAutosaveCheck() const
```
kendy, line 563: Maybe shouldAutosave() or shouldPerformAutosave() would be a bit better name?

wsd/DocumentBroker.hpp
```diff
@@ -540,6 +540,46 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
...
+        {
+            return isAutosaveEnabled()
+                   && std::chrono::duration_cast<std::chrono::seconds>(now()
+                                                                       - _lastAutosaveCheckTime)
+                          >= std::chrono::seconds(30);
+        }
+
+        /// Marks autosave check done.
+        void autosaveChecked() { _lastAutosaveCheckTime = now(); }
```
kendy, line 572: And setAutosaveChecked() maybe here?

wsd/DocumentBroker.cpp
```diff
@@ -430,8 +428,7 @@ void DocumentBroker::pollThread()
         if (isLoaded() && getIdleTimeSecs() >= IdleDocTimeoutSecs)
         {
             // Don't hammer on saving.
-            if (std::chrono::duration_cast<std::chrono::seconds>(now - _lastSaveRequestTime).count()
-                >= 5)
+            if (_saveManager.timeSinceLastSaveRequest() >= std::chrono::seconds(5))
```
kendy, line 431: I wonder - at one occasion there is a specific check method (hasSavingTimedOut()), yet here we expose the time, and check explicitly in the if here; looks a bit inconsistent (?)

wsd/DocumentBroker.hpp
kendy: Actually in a later patch, you use "mar", as in markLastSaveResponseTime(), that would make sense here too maybe, like markAutosaveChecked()?

wsd/DocumentBroker.cpp
```diff
@@ -992,10 +992,9 @@ void DocumentBroker::uploadAsToStorage(const std::string& sessionId,
 }
 
 void DocumentBroker::uploadToStorageInternal(const std::string& sessionId, bool success,
-                                             const std::string& result,
-                                             const std::string& saveAsPath,
-                                             const std::string& saveAsFilename, const bool isRename,
-                                             const bool force)
+                                           const std::string& result, const std::string& saveAsPath,
+                                           const std::string& saveAsFilename, const bool isRename,
+                                           const bool force)
```
kendy, line 996: Looks misaligned after the change (?)

wsd/DocumentBroker.cpp
```diff
@@ -1058,14 +1057,15 @@ void DocumentBroker::uploadToStorageInternal(const std::string& sessionId, bool
 
     const std::string uriAnonym = LOOLWSD::anonymizeUrl(uri);
 
-    // If the file timestamp hasn't changed, skip uploading.
+    // If the file timestamp hasn't changed, skip saving.
     const std::chrono::system_clock::time_point newFileModifiedTime
         = FileUtil::Stat(_storage->getRootFilePath()).modifiedTimepoint();
-    if (!isSaveAs && newFileModifiedTime == _lastFileModifiedTime && !isRename && !force)
+    if (!isSaveAs && newFileModifiedTime == _saveManager.getLastFileModifiedTime() && !isRename
```
kendy: Are you sure the change to "saving" is correct?  Like, if I get the split between the Upload manager and Save manager - the Save manager should track the core state; while the Upload manager the state of the upload.

So how the timestamp of the _save_ could affect the skipping of the _save_ operation?  Conceptually, I even wonder why the Save manger should have any kind of remembering the timestamp (?)  Isn't that responsibility of the Upload?

wsd/DocumentBroker.hpp
```diff
@@ -638,7 +632,7 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
         std::chrono::steady_clock::time_point _lastSaveResponseTime;
 
         /// The jailed file last-modified time.
-        std::chrono::system_clock::time_point _lastFileModifiedTime;
```
kendy: For the purpose of git-grep-ability, I wouldn't be offended if the LastFileModifiedTime didn't go as far as ModifiedTime only ;-) - let's avoid this large rename please...

wsd/DocumentBroker.cpp
Ashod: I didn't want to change the logic in this PR. The helper you mention checks that there is a request active, here the comparison is unconditional regardless of the request finished or not.

In a later patch this logic will be reworked to centralize the decision to save and upload.

wsd/DocumentBroker.hpp
Ashod: Actually, this is only to *check* if we need to autosave. It doesn't decide to autosave, it only decides if we should check the conditions to autosave or not.

But I never liked the name either! I'll rework it when I centralize the save/upload decision logic.

wsd/DocumentBroker.cpp
Ashod: Right, it must be uploading. Fixed.

This is the decision to upload, hence the comparison of the timestamp of the file.

Save manager remembers the the last timestamp of the file on disk. Storage manager remembers the last timestamp of the file we uploaded. We need to track both separately to detect changes on disk (after saving in Core) and changes in Storage (for example, to detect conflicts when Storage changes the files behind our back).

PR #1375, opened 2021-02-01
loleaflet: server recycling: Improve message

pedropintosilva:
Change-Id: I4cecb72acc927707b68711478c5474326d456d82


pedropintosilva:
trying to be 
* Explicit (not hiding that is down);
* Human-readable (no recycling);
* Polite;
* Precise (is restarting automatically [we don't know more]);
* Constructive advice since there is not more granular info there is nothing more that can be said apart from "Please wait" [denotes that it will be brief as opposed to something more abstract in time such as "available shortly" ]);

Ashod (approved):
Looks good, thanks!

PR #1376, opened 2021-02-01
loleaflet: do not update cursor from other views

hcvcastro:
Change-Id: I5e0cd6cfa6c5a5db61dcfbfa557c9ec3bc4c8977


kendy:
Looks reasonable, let's get it in, thank you!

PR #1390, opened 2021-02-02
loleaflet: update cursor position if belong to the document

hcvcastro:
Unfortunately the Server Side sends cursor position for the
annotation window with different coordinates.
I am not sure why it is sending if are useless here,
I guess is used for gtktiledviewer.

Change-Id: Ibb0e2d70786b136c9aad2a3f9f6328503aed60c2


tzolnai:
Hi,
The writer/insert_field_spec.js test fails, because of a commit in the cp-6-4 branch. The cypress test was already updated in the meantime, so a rebase would make this test failure go away.

PR #1392, opened 2021-02-02
In Cypress tests throw error if dialog image cache is inconsistent wr…

tml1024:
…t server

We have a dummy spreadsheet document with a column of cells each in a
different font size. The font size of the current cell is displayed in
the sidebar.

In the test we step through the cells first downwards and then back.
Thus the sidebar image that contains the font size will be updated as
we step through the cells. At first nothing will be found in cache,
but when returning, the images should be found in the cache. If the
cache in the JS client is not consistent with that in the server we
will throw an exception.

Change-Id: I439263c7923d985679fafa0125cd2d7f635e0995

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tzolnai:
Hi,
The writer/insert_field_spec.js test fails, because of a commit in the cp-6-4 branch. The cypress test was already updated in the meantime, so a rebase would make this test failure go away.

PR #1397, opened 2021-02-03
Also save cellviewcursor: messages for handling when we are ready

tml1024:
Without this, when joining an editing session of a spreadsheet, you
don't see the cell cursors of another participant until they move the
cell cursor.

Change-Id: I87b8ad584d57bce52b39706fa9f0bbf56a59082a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great; thanks ! =)

PR #1405, opened 2021-02-04
make clone formating context menu string shorter

Andreas-Kainz:
Change-Id: I8224c8351bd5528986302ae9df09090d8393c3f4

before
2021-02-04 13_02_17-ws-6046668b-b82e-4909-bd71-c9f38d8e79be_0 - noVNC

after
2021-02-04 13_09_16-ws-6046668b-b82e-4909-bd71-c9f38d8e79be_0 - noVNC

same string than in LibreOffice
2021-02-04 13_09_45-Untitled 1 - LibreOfficeDev Writer


eszkadev (commented):
look at 1st line of this file :)
It is automatically generated by the script. 
we can't change the original strings this way because we will not find correct translations for other languages

PR #1443, opened 2021-02-09
Add toString()

tml1024:
Please do not remove even if unused. It can be useful in temporary
console.log() etc.

Change-Id: I2734efa3cffb8dd5e4a9923672ec179fcae3e12f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
The Cypress tests seem to be failing, surely can't be caused by this change, which adds an unused function... Will ignore.

PR #1458, opened 2021-02-11
wsd: userName is optional

Ashod:
This resolves a dependency on userName
(a.k.a. UserFriendlyName) that was a source
of issues when missing.

It turns out that when it's missing but
spellOnline is set, an edge-case caused
an exception that failed loading.

The spellOnline value, in its turn, was
incorrectly set to "null" when in fact
it was missing. This resulted in online
spell-checking being active by default.
Perhaps this was intentional, but here
we don't change this behavior at all.
Instead, we avoid sending "null" when
it's missing, and by default we always
enable online spell-checking, unless
it is explicitly set to "false".

This way, the exception is not thrown
and spellOnline doesn't have invalid
values, while preserving the current
behavior.

Change-Id: I4a09ac44ff5d6147c715afa0fb34af9650da4afd

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1419,7 +1419,7 @@ private:
             Poco::Dynamic::Var var = parser.parse(renderOpts);
             renderOptsObj = var.extract<Object::Ptr>();
         }
-        else if (!userName.empty())
+        else
         {
             renderOptsObj = new Object();
         }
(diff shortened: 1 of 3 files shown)


mcecchetti:
I tested it locally and it fix the problem. Now the 'UserId' field is no more mandatory for running the php example.
Thanks Ash, nice investigation!

mcecchetti (approved):
It's good for me

Ashod:
> I tested it locally and it fix the problem. Now the 'UserId' field is no more mandatory for running the php example.

Thanks @mcecchetti !

PR #1473, opened 2021-02-13
wsd: download templates by wsd

Ashod:
Templates were downloaded by Core
upon loading. This works fine, as
long as there is no special network
setup in loolwsd. However, when
loolwsd has a complex network setup,
such as when using reverse proxies,
Core wouldn't know about the details
and would likely fail to download
the template.

Luckily, there is no reason to rely
on Core for downloading templates.
Instead, we download it in loolwsd,
just like any other document, and
load it in Core as normal. The
remaining post-load saving of
templates remain unchanged.

Change-Id: Ib22ada4ae469863d5e5c8baeee27f667f7cd40ff

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -628,7 +628,7 @@ bool ChildSession::loadDocument(const char * /*buffer*/, int /*length*/, const S
     }
 #endif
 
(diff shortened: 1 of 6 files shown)


mmeeks:
Oooh - exciting =) I love the concept much cleaner and patch looks nice too.

kendy:
Great stuff - thank you! :-)

PR #1480, opened 2021-02-16
add cypress checks for canvas-overlay items

dennisfrancis:
Currently only the cell-cursor and document selections are part of the
canvas overlay. The others are still part of svg layer.

For now only the bounds of the canvas-overlay items are asserted. The
canvas overlay items update their bounds data as JSON text on cypress
only divs inside the overlay section div every time there is any change.

Unlike cell cursors, the selections need not be always a single
rectangle. To add tests for the more general cases, the json format of
test data is helpful.

Change-Id: I70e6328116c44317f31deb9c8bea62e781d8c421

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


cypress_test/integration_tests/common/helper.js
```diff
@@ -1031,6 +1031,39 @@ function textSelectionShouldNotExist() {
...
+// itemName - The name of the overlay item that helps identify the
+//            corresponding test div element.
+// boundsString - A valid JSON string representing the expected bounds of
+//                the overlay item. It must have the keys: top, left, width and height
+//                for a valid bound.
+function overlayItemHasBounds(itemName, boundsString) {
+	cy.log('The overlay item = ' + itemName + ' is expected to have the bounds : ' + boundsString);
+
+	var itemDivId = '#test-div-overlay-' + itemName;
+	return cy.get(itemDivId)
```
tzolnai, line 1056: I don't think this `return` here have any purpose.

cypress_test/integration_tests/common/helper.js
```diff
@@ -1031,6 +1031,39 @@ function textSelectionShouldNotExist() {
...
+		.should('exist')
+		.then(function (elem) {
```
tzolnai, line 1058: You should use `.should()` here instead of `.then()`. The difference is that `then` is called only once and so the assertion is checked once, while with `should` the assertion is checked until it passes or reaches a timeout. That latter one is what we generally use because the selection overlay for example is not updated instantly, so with `then` the assertion can fail because it runs before the selection overlay is updated. However, with `should` selection overlay has time to be updated. The test case you commented out, might be fixed by this change.

cypress_test/integration_tests/common/helper.js
```diff
@@ -1031,6 +1031,39 @@ function textSelectionShouldNotExist() {
...
+//            corresponding test div element.
+// boundsString - A valid JSON string representing the expected bounds of
+//                the overlay item. It must have the keys: top, left, width and height
+//                for a valid bound.
+function overlayItemHasBounds(itemName, boundsString) {
+	cy.log('The overlay item = ' + itemName + ' is expected to have the bounds : ' + boundsString);
+
+	var itemDivId = '#test-div-overlay-' + itemName;
+	return cy.get(itemDivId)
+		.should('exist')
```
tzolnai, line 1057: This is not needed. If there is a `then` or `should` after `cy.get()` it will check existence anyway.

tzolnai (commented):
Hi, see my comments related to the new helper method.
In general, I'm not a big fan of hard-coded positions, because it needs to be updated every time some small things change either in core or in online code. Now, you are working on canvas rendering with Gokay, so I'm sure you will need to update this test code more times because of these hard-coded values.  But it's just my opinion, I won't stop it happen. Fortunately, somebody else will maintain this code in the long term.

tzolnai:
On second thought, can you add these assertions to separate test cases? You can create a new test suite like canvas_overlay_spec.js or something and add some simple test cases for cellcursor and selection overlay.
In general, I think it's better to keep test cases as short as possible and test one thing in one test case. These existing test cases check formula bar features, so it's not a good idea to increase these test cases's length and test more things at once in them. Now, you intend to test cellcursor and selection overlay, so you should create separate test cases for these, which test cases have the focus on these overlay items. Later it will be easier to extend the overlay-related test suite with even more related test cases, instead of finding existing, unrelated tests to extend with additional assertions.

dennisfrancis:
@tzolnai Thanks for the review! I've made the amends you suggested.

tzolnai:
Thanks! Looks good to me!
Earlier I avoided using local variables because they were not set as expected because of the asynchronicity, but it seems to work the way you used them here. So great to see this works!

PR #1485, opened 2021-02-17
wsd: revisit log messages DocumentBroker.cpp

hcvcastro:
other cases LOG_WRN, LOG_INF

Change-Id: I174a86f48e6b4107d0f846fd9af28a7ab1f62afc

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1772,7 +1772,7 @@ std::shared_ptr<ClientSession> DocumentBroker::createNewClientSession(
     }
     catch (const std::exception& exc)
     {
-        LOG_WRN("Exception while preparing session [" << id << "]: " << exc.what());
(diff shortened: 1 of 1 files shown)


Ashod (commented):
I'm not convinced the DocBroker exit cases are FTL. The issue is that FTL is used elsewhere to signal that an unrecoverable error happened, and we will exit the process (there is nothing we can do but exit). But here we are reporting either an unexpected situation (child terminated prematurely) or we are aborting the document loading (because it's taking too long), or we failed to spawn a child. All of these are errors, but they aren't fatal in any sense. We continue to run, and we are most likely going to be succeed in loading other documents (and even the one that just failed).

I suggest we reserve FTL for unrecoverable error, to be used only before exiting the process (or to flag for process shutdown and exit), but not for errors that we recover from and continue working normally.

PR #1487, opened 2021-02-17
wsd: revisit log messages ClientSession.cpp

hcvcastro:
other cases LOG_WRN, LOG_INF

Change-Id: I4ab8d04099a432e7fd3fb6810d7a9aa038cbb57d

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -1387,7 +1387,7 @@ bool ClientSession::handleKitToClientMessage(const char* buffer, const int lengt
         {
             if (errorCommand == "load")
             {
-                LOG_WRN("Document load failed: " << errorKind);
+                LOG_ERR("Document load failed: " << errorKind);
                 if (errorKind == "passwordrequired:to-view" ||
                     errorKind == "passwordrequired:to-modify" ||
                     errorKind == "wrongpassword")
@@ -1415,7 +1415,7 @@ bool ClientSession::handleKitToClientMessage(const char* buffer, const int lengt
             }
             else
             {
-                LOG_WRN(errorCommand << " error failure: " << errorKind);
+                LOG_ERR(errorCommand << " error failure: " << errorKind);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


wsd/ClientSession.cpp
```diff
@@ -1170,7 +1170,7 @@ bool ClientSession::handleKitToClientMessage(const char* buffer, const int lengt
         {
             if (errorCommand == "load")
             {
-                LOG_WRN("Document load failed: " << errorKind);
+                LOG_ERR("Document load failed: " << errorKind);
```
Ashod, line 1173: Careful, this isn't necessarily an error that we want to report! When a document is password protected, loading will fail, and we notify the user and prompt for the password. This isn't an error in the sense that something went wrong.

I think warning here is reasonable. But we can *add* an ERR log in the `else` case, when errorKind is not the three cases we recognize. (That is, add `else LOG_ERR` in line 1197 and leave this one as WRN.)

wsd/ClientSession.cpp
```diff
@@ -1113,7 +1113,7 @@ bool ClientSession::handleKitToClientMessage(const char* buffer, const int lengt
     const std::shared_ptr<DocumentBroker> docBroker = _docBroker.lock();
     if (!docBroker)
     {
-        LOG_ERR("No DocBroker to handle kit-to-client message: " << firstLine);
+        LOG_FTL("No DocBroker to handle kit-to-client message: " << firstLine);
```
Ashod, line 1116: This can happen when the user closes a document, but we still have some messages from Core queued up. Certainly not fatal. It's an error, because this is unexpected, and because there might be some important notifications worth logging, but nothing fatal here.

wsd/ClientSession.cpp
```diff
@@ -319,7 +319,7 @@ bool ClientSession::_handleInput(const char *buffer, int length)
     std::shared_ptr<DocumentBroker> docBroker = getDocumentBroker();
     if (!docBroker || docBroker->isMarkedToDestroy())
     {
-        LOG_ERR("No DocBroker found, or DocBroker marked to be destroyed. Terminating session " << getName());
+        LOG_FTL("No DocBroker found, or DocBroker marked to be destroyed. Terminating session " << getName());
```
Ashod, line 322: Nothing fatal. It's very common that a document being unloaded gets some activity that we reject (such as loading a new view). We continue working just fine, and this is reasonably common, but it's not FTL.

Ashod (changes requested):
Good, thanks. Left some notes.

PR #1495, opened 2021-02-17
Fix compilation with libc++

arrowd:
```
wsd/DocumentBroker.cpp:772:57: error: constexpr variable cannot have non-literal type 'const std::chrono::system_clock::time_point' (aka 'const time_point<std::__1::chrono::system_clock>')
        constexpr std::chrono::system_clock::time_point Zero;
                                                        ^
/usr/include/c++/v1/chrono:1355:28: note: 'time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000>>>' is not literal because it is not an aggregate and has no constexpr constructors other than copy or move con
class _LIBCPP_TEMPLATE_VIS time_point
                           ^
```

Change-Id: I88c5466b8ad104178cf7fa4101f5f5265ccf5e86

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -769,7 +769,7 @@ bool DocumentBroker::download(const std::shared_ptr<ClientSession>& session, con
     {
         // Check if document has been modified by some external action
         LOG_TRC("Document modified time: " << fileInfo.getModifiedTime());
-        constexpr std::chrono::system_clock::time_point Zero;
+        std::chrono::system_clock::time_point Zero;
         if (_storageManager.getLastModifiedTime() != Zero && fileInfo.getModifiedTime() != Zero
             && _storageManager.getLastModifiedTime() != fileInfo.getModifiedTime())
         {


mrkara:
@arrowd Thank you for your contribution! We're looking forward to your next pull request. :)

Ashod:
Thanks @arrowd. Curious what compiler and version is the error from.

arrowd:
I'm compiling on FreeBSD, which uses Clang as its default compiler and libc++ as standard C++ library.

Are you interested in integrating a CirrusCI check that runs FreeBSD?

Ashod:
> I'm compiling on FreeBSD, which uses Clang as its default compiler and libc++ as standard C++ library.

While Clang isn't officially supported, some devs do compile using Clang (especially for fuzzing, for example). Which version are you using? (I'm trying to minimize the effort of fixing these trivial issues and avoid regressions in the future.)

> Are you interested in integrating a CirrusCI check that runs FreeBSD?

That wouldn't be up to me, but any CI setup that improves the coverage footprint is surely welcome!

(Btw, I just noticed a couple of your PRs that were buried in my notifications. Apologies for that.)

arrowd:
> Which version are you using? (I'm trying to minimize the effort of fixing these trivial issues and avoid regressions in the future.)

I'm using LLVM/Clang 11, which is shipped with FreeBSD 13+.

> That wouldn't be up to me, but any CI setup that improves the coverage footprint is surely welcome!

All right, I'll get to it once I make the build finish successfully.

PR #1497, opened 2021-02-18
Stopgap measure to avoid iterating over rows to 500001 and columns to…

tml1024:
… 1024

When the ticks parameter passed to the
L.Control.Header.GapTickMap.initialize() function has a last entry for
"all the rest", don't actually create elements in the knownTicks array
for all of those, but set a flag called _toTheLimit that indicates
that the last element in the array is to be interpreted to cover all
the rest up to the maximum for the data in question. (Row index 500001
or column index 1024.)

For now, that flag is not checked anywhere, but this patch seems to
help avoiding pointless and very time consuming iteration anyway
without any huge negative impact?

Change-Id: I81b6c1175aebb51069cb33d9b136166b06d03459

* Resolves: # 
* Target version: co-4-2 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/control/Control.ColumnHeader.js
+++ b/loleaflet/src/control/Control.ColumnHeader.js
@@ -408,7 +408,7 @@ L.Control.ColumnHeader = L.Control.Header.extend({
 		}
 
 		// create data structure for column widths
-		this._tickMap = new L.Control.Header.GapTickMap(this._map, columns);
+		this._tickMap = new L.Control.Header.GapTickMap(this._map, columns, 1024);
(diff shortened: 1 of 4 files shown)


tml1024:
(In co-6-4 and master the code is completely different, this is for co-4-2 only.)

dennisfrancis (approved):
Looks good. Thanks!

tml1024:
In case the "Main chain for co-4-2" failure was a random fluke, did a minor edit of the commit message and pushed again, to cause a new check...

PR #1498, opened 2021-02-18
CanvasSectionContainer: Event propagation.

gokaysatir:
Now, bound sections propagate events to the sections those are beneath
them. StopPropagation feature is also implemented.

Change-Id: I54c886fea0e254d4b97c4d06585db8ed28d40ab9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CanvasSectionContainer.ts
+++ b/loleaflet/src/layer/tile/CanvasSectionContainer.ts
@@ -7,6 +7,25 @@
  	Supports multi touch with 2 fingers.
  	This class uses only native events. There is: No timer for anything & no longpress for desktop & no doubleclick for touch screen & no wheel for touchscreen etc.
 
+	Propagated events:  All events are propagated between "bound sections".
+						If 2 sections overlap but not bound, only top section gets the event.
+
+	lowestPropagatedBoundSection:
+			This property applies to bound sections.
+			Bound sections are like layers. They overlap entirely.
+			Every bound section should share same zIndex unless there is a very good reason for different zIndex values.
+			To make a section "bound" to another, one can use "boundToSection" property (section name - string).
+			When 2 or more sections overlap, the top section will get the event first.
+			If a section handles the event and calls stopPropagating function, the sections after that section won't get the event.
+			Example scenario:
+				Event: drag
+				Top section handles event.
+				Middle section handles event then calls "stopPropagating" function. From this point, top section still handles the event until the end of the event.
+				Bottom section doesn't get the event.
+
+			If stopPropagating function is called, it is valid until the end of the event. So for events like dragging, calling
+			the function only once is enough. When the event is lasted, lowestPropagatedBoundSection variable is set to null automatically.
+
 	New section's options:
 	name: 'tiles' | 'row headers' | 'column headers' | 'markers & cursor' | 'shapes' | 'scroll' etc.
 	anchor: 'top left' | 'top right' | 'bottom left' | 'bottom right' (Examples: For row & column headers, anchor will be 'top left'; If we would want to draw something sticked to bottom, it would be 'bottom left' or 'bottom right').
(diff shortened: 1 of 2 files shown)


loleaflet/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -285,10 +311,205 @@ class CanvasSectionContainer {
...
+			this.updateBoundSectionList(tempSectionList[i], sectionList);
+		}
+
+		return sectionList;
+	}
+
+	private orderBoundsList(section: CanvasSectionObject) {
+		// According to zIndex.
+		for (var i: number = 0; i < section.boundsList.length - 1; i++) {
+			for (var j = i + 1; j < section.boundsList.length; j++) {
```
dennisfrancis, line 349: These two can be done together in a double for loop (or better using a comparison function passed to array builtin sort()) so that it is easier to understand and yet shorter code in my opinion.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -285,10 +311,205 @@ class CanvasSectionContainer {
 		}
 	}
 
+	private updateBoundSectionList(section: CanvasSectionObject, sectionList: Array<CanvasSectionObject> = null): Array<CanvasSectionObject>{
```
dennisfrancis, line 314: Can we have a non recursive version of this please, which will be much easier to understand and write too. A tweaked double for-loop can achieve this I think.

dennisfrancis:
Looks good in general. If you could flatten the recursive definition that would be great. Thanks!

loleaflet/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: Hmm, i think we can't achieve this functionality with double for loop. Think complicated cases, there may be many sections bound to each other.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: İf you mean two ordering parts may be merged, i think it might cause buggy behaviour. Second one relies on the first's results. For comparison function, this is a small bits of code. Do we real need that:)

Thank you for the comments and your time.

dennisfrancis: It might seem so, but it is equivalent to a single sort with a comparison function (or its reverse depending on how you want to sort) :

if (one.zIndex > two.zIndex || (one.zIndex == two.zIndex && one.drawingOrder > two.drawingOrder) )
   return -1; // one less than two, so in ascending sort(), one is placed before two.
if (one.zIndex == two.zIndex && one.drawingOrder == two.drawingOrder)
  return 0; // equal (no swap)
return 1; // one greater than two.

(The second double-for is basically moving around just the ones that have same zindex.)

dennisfrancis: a single sort() with comparison function will save the trouble of writing a double for loop every time we want to sort :)

loleaflet/src/layer/tile/CanvasSectionContainer.ts
dennisfrancis: Actually, we can do even better, in one linear pass we use a map to get the "section->parent_of_boundsection_if_any" relationship and use this map to create lists of bound sections if it is really needed.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: Oh, this is a smart solution:) Thanks.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: I see you mean something like std::map right?
I put these lists because it would need to be re-calculated for every event, so lists are created performance in mind.
boundsList is an ordered list.
Bound sections may be complicated, i really think that we need recursive function to achieve this goal.
parent_of_boundsection does not have to be on top of that section, and sections do not need to be bound to one single section. There may be a view like a family tree. Current implementation allows this.
I think this function is fine.

dennisfrancis: I see, so "binding" of sections are bidirectional and don't have a parent child relationship. So to generalize this situation am I correct with this statement ? If we have a N sections, the job is to find disjoint sets of sections such that each section in a set is bound to every other section in that set. After finding those sets we need to order them by zindex and drawing order. If this is true, then there is a super fast way to achieve this using a famous data structure called "union-find" (Refer https://en.wikipedia.org/wiki/Disjoint-set_data_structure). This data structure is dynamic, so we can add/remove sections or change bindings of any section. For our purpose, we have to do the sorting of members of the sets ourselves.

As this requires more time implementing lets get this in now and hopefully come back to replace that costly recursive function called in a loop some day :)

Some side comments:
The lists/arrays in JS may have efficient implementations, but an algorithm >= O(N^2) can be made to crawl very easily (much easily for an exponential case algo) even if it uses super optimized data structure implementations and/or run on super fast hardware. Recursive functions may look neat and compact but it can get incredibly hard to debug them when we hit a problem with it. I used to write recursive functions until I had to debug them :)

PR #1506, opened 2021-02-19
Include the view id in the tunnelled dialog image hash

tml1024:
Otherwise this will break badly in the collaborative editing (multiple
views) case.

Change-Id: I63d391cb5a7d118c34564c4e1a4df7710b13e90e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tzolnai (approved):
The test failure goes away with this change. So looks good to me.

PR #1517, opened 2021-02-22
Port loolwsd-systemplate-setup to FreeBSD.

arrowd:
Change-Id: I9944e79d0fe220a5f2fe40d2ba98619c22d25a77

* Target version: master 

### DIFF
--- a/loolwsd-systemplate-setup
+++ b/loolwsd-systemplate-setup
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # set -x
 
 test $# -eq 2 || { echo "Usage: $0 <chroot template directory for system libs to create> <LO installation directory>"; exit 1; }
@@ -20,6 +20,16 @@ mkdir -p $CHROOT || exit 1
 CHROOT=`cd $CHROOT && /bin/pwd`
 INSTDIR_LOGICAL=`cd $INSTDIR && /bin/pwd -L`
 INSTDIR=`cd $INSTDIR && /bin/pwd -P`
+
+if [ ! `uname -s` = "FreeBSD" ]; then
+    CP=cp
+    REALPATH=realpath
+    LOCALBASE=usr/
+else
+    CP=gcp
+    REALPATH=grealpath
+    LOCALBASE=usr/local/
+fi
 
 cd / || exit 1
 
@@ -57,7 +67,7 @@ find $INSTDIR -name 'xpdfimport' |
 while read file; do
     ldd $file 2>/dev/null
 done |
-grep -v dynamic | cut -d " " -f 3 | grep -E '^(/lib|/usr)' | sort -u | sed -e 's,^/,,'
+grep -v dynamic | cut -s -d " " -f 3 | grep -E '^(/lib|/usr)' | sort -u | sed -e 's,^/,,'
 
 ) |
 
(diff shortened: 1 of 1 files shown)


loolwsd-systemplate-setup
```diff
@@ -21,6 +21,16 @@ CHROOT=`cd $CHROOT && /bin/pwd`
 INSTDIR_LOGICAL=`cd $INSTDIR && /bin/pwd -L`
 INSTDIR=`cd $INSTDIR && /bin/pwd -P`
 
+if [ ! `uname -s` = "FreeBSD" ]; then
+    CP=cp
+    REALPATH=realpath
+    LOCALBASE=usr/
+else
+    CP=gcp
```
Ashod, line 29: Freebsd doesn't have `cp`? (I find it hard to believe.)

loolwsd-systemplate-setup
```diff
@@ -55,7 +65,11 @@ find etc/fonts \
 # they link to.
 find $INSTDIR -name 'xpdfimport' |
 while read file; do
-    ldd $file 2>/dev/null
+    if [ ! `uname -s` = "FreeBSD" ]; then
+        ldd $file 2>/dev/null
+    else
+        ldd $file 2>/dev/null | tail -n +2
```
Ashod, line 71: A comment to explain what's going on?

Ashod (commented):
Thanks a lot Gleb! A couple of questions inline. (Also, we need to get Jenkins happy. A rebase perhaps will help.)

loolwsd-systemplate-setup
arrowd: It does, but it has different flags. Some like `--dereference` are missing. So I thought it is simpler to GNU `cp`.

Ashod:
@timar would be great if you can take a look and see if you have concerns. I'd personally prefer to avoid (or minimize) the conditionals on system name as much as possible.

Thanks.

loolwsd-systemplate-setup
```diff
@@ -55,7 +65,16 @@ find etc/fonts \
 # they link to.
 find $INSTDIR -name 'xpdfimport' |
 while read file; do
-    ldd $file 2>/dev/null
+    if [ ! `uname -s` = "FreeBSD" ]; then
+        ldd $file 2>/dev/null
+    else
+        # lld on FreeBSD prints extraneous line, so chop it
```
timar, line 71: lld > ldd

timar: but why is this necessary? The grep in line 79 won't match the empty line.

timar (changes requested):

loolwsd-systemplate-setup
arrowd: It is not empty but

    /usr/local/lib/libreoffice/program/xpdfimport:

timar:
> @timar would be great if you can take a look and see if you have concerns. I'd personally prefer to avoid (or minimize) the conditionals on system name as much as possible.
> 
> Thanks.

@ash I tend to agree. We could use autotools to generate this script. But that'd be an overkill. :)

loolwsd-systemplate-setup
timar: > It is not empty but
> 
> ```
> /usr/local/lib/libreoffice/program/xpdfimport:
> ```

Oh... Would  `cut -s -d " " -f 3` help then?

mmeeks:
Oooh - I suspect we don't use xpdfimport anymore - with the built-in pdfium code. Probably dunging that out would be useful (with some testing around PDF of course).

loolwsd-systemplate-setup
arrowd: Yep, seems to be working!

timar (approved):
Thanks!

timar:
> Oooh - I suspect we don't use xpdfimport anymore - with the built-in pdfium code. Probably dunging that out would be useful (with some testing around PDF of course).

@mmeeks Collabora's releases do not include xpdfimport (wrapper for the GPL-ed libpoppler). But technically it's possible to configure it.

PR #1522, opened 2021-02-23
kit: revisit log messages ChildSession.cpp

hcvcastro:
other cases LOG_WRN, LOG_INF

Change-Id: I76b4b066db59acd2e7dc45bf8606021104f79cae

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -754,7 +754,7 @@ bool ChildSession::sendFontRendering(const char* /*buffer*/, int /*length*/, con
     }
     catch (Poco::SyntaxException& exc)
     {
-        LOG_DBG(exc.message());
+        LOG_ERR(exc.message());
         sendTextFrameAndLogError("error: cmd=renderfont kind=syntax");
         return false;
     }


kit/ChildSession.cpp
```diff
@@ -631,7 +631,7 @@ bool ChildSession::loadDocument(const char * /*buffer*/, int /*length*/, const S
     const bool loaded = _docManager->onLoad(getId(), getJailedFilePathAnonym(), renderOpts);
     if (!loaded || _viewId < 0)
     {
-        LOG_ERR("Failed to get LoKitDocument instance for [" << getJailedFilePathAnonym() << "].");
+        LOG_FTL("Failed to get LoKitDocument instance for [" << getJailedFilePathAnonym() << "].");
```
Ashod, line 634: Loading can fail for many reasons. It's an error, but not fatal. We can load other documents and even the same document again. FTL signals unrecoverable error, where we exit the process because we cannot continue running. This is nothing like that. ERR is reasonable here.

Ashod (changes requested):
Thanks @hcvcastro. Looks good, but one case needs rethinking. Let's discuss as necessary.

PR #1523, opened 2021-02-23
common: revisit log messages misc files

hcvcastro:
other cases LOG_WRN, LOG_INF

Change-Id: I15bc840f5a6eaf9a74a1ee1caa57d447baa2afce

### DIFF
--- a/common/Authorization.cpp
+++ b/common/Authorization.cpp
@@ -80,7 +80,7 @@ void Authorization::authorizeRequest(Poco::Net::HTTPRequest& request) const
             break;
         }
         default:
-            LOG_TRC("No HTTP Authorization type detected. Assuming no authorization needed. "
+            LOG_ERR("No HTTP Authorization type detected. Assuming no authorization needed. "
                     "Specify access_token to set the Authorization Bearer header.");
             break;
     }
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -281,7 +281,7 @@ namespace FileUtil
         catch (const std::exception&e)
         {
             // Already removed or we don't care about failures.
-            LOG_DBG("Failed to remove [" << path << "] " << (recursive ? "recursively: " : "only: ")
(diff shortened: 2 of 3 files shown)


common/Authorization.cpp
```diff
@@ -80,7 +80,7 @@ void Authorization::authorizeRequest(Poco::Net::HTTPRequest& request) const
             break;
         }
         default:
-            LOG_TRC("No HTTP Authorization type detected. Assuming no authorization needed. "
+            LOG_ERR("No HTTP Authorization type detected. Assuming no authorization needed. "
```
Ashod, line 83: Interesting. It's hard to know if there is no http authorization intentionally or it's a bug. If it's intentional, then this will be a very annoying ERR that will fill the logs (it will be logged for each and every URI request we process!). But if it's a bug, it's a helpful hint, although without the authorization documents won't load anyway, so it's not going to be unnoticed.

I think we won't be helpful if we make this an ERR, because there is no requirement for http authorization (this error will make it a requirement, implicitly and without documentation, because there is no way to silence this error without having http authorization). So, we either make a requirement and change this to ERR, or, if it's nor required, then we can change this to INF, so that it's more visible when troubleshooting.

common/FileUtil.cpp
```diff
@@ -318,7 +318,7 @@ namespace FileUtil
         if (access(path, W_OK) == 0)
             return true;
 
-        LOG_DBG("Cannot access path [" << path << "]: " << strerror(errno));
+        LOG_ERR("Cannot access path [" << path << "]: " << strerror(errno));
```
Ashod, line 321: This isn't an error. The purpose of this utility helper is to check whether a directory is writable or not. There is no expectation that the directory is writable. The DBG is just for that, debugging (to help devs understand which directories are writable and which aren't, to understand the decision made by the code).

Ashod (changes requested):
Good, thanks @hcvcastro. See notes please. Let's discuss as necessary.

PR #1525, opened 2021-02-24
wsd: perform post-loading actions only once

Ashod:
Since now we consider "statusindicatorfinish:" as
indication of document loading (in addition to
"status:") we have to avoid performing the
post-loading actions twice.

Specifically, we need to save the document
immediately after loading if it had been
loaded from a template.

Change-Id: I269bab3b76feddaa9e65a51c8d23d7c1f8ad3e02


hcvcastro:
Thanks, I have another solution to avoid regressions

https://github.com/CollaboraOnline/online/pull/1544

PR #1526, opened 2021-02-24
Online Draw Features

merttumer:
* Resolves: # 
* Target version: master 

### Summary
The core patches are on only cp-6.4 at the moment. It wont work against master of core.

### TODO

- [ ] ...

### Checklist


dennisfrancis (commented):
@merttumer @kendy I think these changes look good. As it extends the current svg based marker/handlers, we might need some changes when we move them to canvas like using core-pixels instead of latlng, event handling with canvas sections etc. Thanks!

PR #1531, opened 2021-02-25
Private/gokay/canvasscroll4

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/TileLayer.js
+++ b/loleaflet/src/layer/tile/TileLayer.js
@@ -1979,7 +1979,8 @@ L.TileLayer = L.GridLayer.extend({
 					center = center.subtract(this._map.getSize().divideBy(2));
 					center.x = Math.round(center.x < 0 ? 0 : center.x);
 					center.y = Math.round(center.y < 0 ? 0 : center.y);
-					this._map.fire('scrollto', {x: center.x, y: center.y});
+					if (!this._map.wholeColumnSelected && !this._map.wholeRowSelected)
+						this._map.fire('scrollto', {x: center.x, y: center.y});
 				}
 			}
 		}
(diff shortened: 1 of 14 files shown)


gokaysatir:
@dennisfrancis i am not sure if it will compile. If not, i will solve the cypess test issue and send a commit again.

loleaflet/src/layer/tile/TileLayer.js
```diff
@@ -1979,7 +1979,6 @@ L.TileLayer = L.GridLayer.extend({
 					center = center.subtract(this._map.getSize().divideBy(2));
 					center.x = Math.round(center.x < 0 ? 0 : center.x);
 					center.y = Math.round(center.y < 0 ? 0 : center.y);
-					this._map.fire('scrollto', {x: center.x, y: center.y});
```
dennisfrancis, line 1982: It seems all the above code just computes the new center but does nothing with it. Is it possible to adjust the scroll to that new position in the scrollbar or just set the map's center to that so that everything automatically change to that location ?

dennisfrancis (commented):
Looks good to me except for the 'scrollto' firing removal, perhaps you were already going to amend that :) Please merge when you are happy with everything and the CI passes. Thanks a lot!

tzolnai:
I checked those tests. I don't like them, how they work (independently from this change). I think it's better to remove them in this form. The cursor position is not a good way of testing the page layout. So can you remove those failing tests? Later somebody might find a better way to test these things.

gokaysatir:
> I checked those tests. I don't like them, how they work (independently from this change). I think it's better to remove them in this form. The cursor position is not a good way of testing the page layout. So can you remove those failing tests? Later somebody might find a better way to test these things.

Okay, thank you for the review:) I will remove them and merge.

loleaflet/src/layer/tile/TileLayer.js
gokaysatir: By the way, later i saw that i shouldn't update that part. Instead, i used map.wholeColumnSelected and map.wholeRowSelected variables. FYI, thanks :)

PR #1537, opened 2021-02-25
Revert "kit: enable send LOK_CALLBACK_STATUS_INDICATOR_FINISH"

Ashod:
The commit in question breaks two unit-tests: wopi-template
and copy-paste. Temporarily reverting until the regressions
it introduces are resolved.

This reverts commit be68f06f70ce0acdec688abf1d201b98a367d376.

Change-Id: I1fafd73fdea57077eea05a146b7df0b013e6e8ba

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -792,8 +792,7 @@ public:
             self->setDocumentPassword(type);
             return;
         }
-        else if (type == LOK_CALLBACK_STATUS_INDICATOR_SET_VALUE ||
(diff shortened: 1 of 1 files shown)


Ashod:
The revert fixes the two regression C++ unit-tests. However there seems to be other failures in the Cypress tests. Since this PR does improve the situation, and since the failure is further ahead of the previous C++ failures, this seems like an improvement, so I will merge to facilitate fixing the Cypress failures.

PR #1544, opened 2021-02-26
wsd: add interactive state

hcvcastro:
Add a new state before load the document,
when Macro Security dialog popup, and avoid
to send the event load timeout.

Change-Id: I5973c5205e90e5447e5478cbab895709a68606f6

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -298,7 +298,7 @@ void DocumentBroker::pollThread()
 #endif
         LOOLWSD::getConfigValue<int>("per_document.limit_load_secs", 100);
 
-    const auto loadDeadline = std::chrono::steady_clock::now() + std::chrono::seconds(limit_load_secs);
+    auto loadDeadline = std::chrono::steady_clock::now() + std::chrono::seconds(limit_load_secs);
 #endif
 
     // Main polling loop goodness.
@@ -312,6 +312,12 @@ void DocumentBroker::pollThread()
         // a tile's data is ~8k, a 4k screen is ~128 256x256 tiles
         if (_tileCache)
             _tileCache->setMaxCacheSize(8 * 1024 * 128 * _sessions.size());
+
+        if (isInteractive())
+        {
+            loadDeadline = now + std::chrono::seconds(limit_load_secs);
+            continue;
+        }
 
         if (!isLoaded() && (limit_load_secs > 0) && (now > loadDeadline))
         {
@@ -1281,6 +1287,15 @@ void DocumentBroker::setLoaded()
         _loadDuration = std::chrono::duration_cast<std::chrono::milliseconds>(
                                 std::chrono::steady_clock::now() - _threadStart);
         LOG_TRC("Document loaded in " << _loadDuration);
+    }
+}
+
+void DocumentBroker::setInteractive(bool value)
+{
+    if (isInteractive() != value)
+    {
+        _docState.setInteractive(value);
+        LOG_TRC("Document has interactive dialogs before load");
     }
 }
 
--- a/wsd/DocumentBroker.hpp
+++ b/wsd/DocumentBroker.hpp
@@ -294,6 +294,9 @@ public:
     /// Notify that the load has completed
     virtual void setLoaded();
 
+    /// Notify that the document has dialogs before load
+    virtual void setInteractive(bool value);
+
     /// If not yet locked, try to lock
     bool attemptLock(const ClientSession& session, std::string& failReason);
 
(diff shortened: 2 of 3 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -313,6 +313,12 @@ void DocumentBroker::pollThread()
         if (_tileCache)
             _tileCache->setMaxCacheSize(8 * 1024 * 128 * _sessions.size());
 
+        if (isInteractive())
+        {
+            loadDeadline = now + std::chrono::seconds(limit_load_secs);
+            continue;
```
Ashod, line 319: How do we get out of interactive mode?

I expect we want to set interactive to false at some point so that we don't always `continue` here and execute the remaining conditions in this loop.

wsd/DocumentBroker.hpp
```diff
@@ -892,10 +897,14 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
...
         bool isCloseRequested() const { return _closeRequested; }
 
+        void setInteractive() { _interactive = true; }
+        bool isInteractive() const { return _interactive; }
+
     private:
         Status _status;
         std::atomic<bool> _closeRequested; //< Owner-Termination flag.
         std::atomic<bool> _loaded; //< If the document ever loaded (check isLive to see if it still is).
+        std::atomic<bool> _interactive; //< If the document has interactive dialogs before load
```
Ashod, line 907: Doesn't need to be atomic.

Ashod (changes requested):
Thanks Henry. Good usage of the state class, but I expect that we want to set the interactive flag to false at some point so that we let the normal processing take place. It's not clear to me how the full cycle should work. Can you add some comment about the use-case? When do we get interactive dialog before load? What happens with the dialog (what are the outcomes)? and how does the code behave for each case?

Thanks!

hcvcastro:
> Thanks Henry. Good usage of the state class, but I expect that we want to set the interactive flag to false at some point so that we let the normal processing take place. It's not clear to me how the full cycle should work. Can you add some comment about the use-case? When do we get interactive dialog before load? What happens with the dialog (what are the outcomes)? and how does the code behave for each case?

Sure,  this is the workflow

1) The document to load has an embedded VBA macro
2) When LO Core finalizes the progress status, it enters to execute an interactive handler
3) It checks the existence of macros and popup a Macro Warning dialog to enable/disable
4) Continue normal loading processing the document, depending on the dialog response

Ashod:
> Sure, this is the workflow
> 
> 1. The document to load has an embedded VBA macro
> 2. When LO Core finalizes the progress status, it enters to execute an interactive handler
> 3. It checks the existence of macros and popup a Macro Warning dialog to enable/disable

Thanks!

> 4. Continue normal loading processing the document, depending on the dialog response

Right, but we don't do anything to continue normal loading.

```
if (isInteractive())
  ... 
  continue;
```

If `isInteractive()` is always true, we will never execute the remaining of the poll loop.

So the question is: what do we do when the dialog is dismissed to get out of the interactive state in DocumentBroker ?

hcvcastro:
> So the question is: what do we do when the dialog is dismissed to get out of the interactive state in DocumentBroker ?
> Merge state
> 

when the "status:" message arrives

docBroker->setInteractive(false);

Ashod:
> > So the question is: what do we do when the dialog is dismissed to get out of the interactive state in DocumentBroker ?
> > Merge state
> 
> when the "status:" message arrives
> 
> docBroker->setInteractive(false);

Thanks.

Is there any reason why we don't consider the interactive dialog as a mark that the document loaded?

My concern is that we are introducing a new state that logically is the same as being loaded, which makes maintenance difficult. What happens during the interactive state that we can't do in the loaded state? How is Interactive different from Loaded?

Thanks again for your help and this patch. I'm trying to understand the consequences of this change, and if we can simplify it.

hcvcastro:
> Is there any reason why we don't consider the interactive dialog as a mark that the document loaded?

> My concern is that we are introducing a new state that logically is the same as being loaded, which makes maintenance difficult. What happens during the interactive state that we can't do in the loaded state? How is Interactive different from Loaded?
> 
> Thanks again for your help and this patch. I'm trying to understand the consequences of this change, and if we can simplify it.
> Merge state
> 

Interactive Handler
* it does not exist Model/Controller/View yet 
 
setLoaded
* The Model/Controller/View was created and ready to use

Ashod:
After discussing with @kendy, it appears that the interactive state might come in handy for other "during loading" interaction with users. 

Thanks @hcvcastro for the patch!

PR #1548, opened 2021-02-28
Simulate socket errors more reliably

Ashod:
We avoid changing the SSL wantsTo state and instead set EAGAIN in SSL just like we do for the non-SSL, because the error-handling is the same for both cases. We also have correct (and more informative) logging now.

- wsd: set errno when simulating socket errors for SSL
- wsd: socket error simulation to maintain the ssl state


mmeeks (commented):
Looks plausible =)

PR #1560, opened 2021-03-02
wsd: perform socket writes more efficiently

Ashod:
Since it's most efficent to send full blocks at
once (especially so when SSL encryption is done),
we now perform write when the buffer has fewer
than the send-buffer-size bytes. Previously,
we were only getting more data to write to the
socket when the buffer was completely empty,
potentially wasting valuable opportunity to
write more data to the socket while we're at it.

Change-Id: I69a18a042dc2e551ebc5e1af40dae091da3f3d13

### DIFF
--- a/common/Session.hpp
+++ b/common/Session.hpp
@@ -100,7 +100,7 @@ public:
     }
 
     // By default rely on the socket buffer.
-    void writeQueuedMessages() override
+    void writeQueuedMessages(std::size_t) override
     {
         assert(false);
     }
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -529,7 +529,7 @@ public:
             // Get the data to write into the socket
             // from the client's callback. This is
             // used to upload files, or other data.
-            char buffer[16 * 1024];
+            char buffer[64 * 1024];
             std::size_t wrote = 0;
             do
             {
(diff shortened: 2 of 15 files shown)


mmeeks:
Oooh - this is a really interesting one =) I like the change in so far as it goes, but ... a couple of things.

* Reading the code wsd/ClientSession.cpp (writeQueuedMessages) - appears to only de-queue and append a single item.
* Wow that is far from ideal - that is very noisy in terms of overhead, and means we presumably end up sending a ton of messages one by one - when we could send big blocks of data. That prolly creates extra race conditions in the JS as we process things then spin. So - great catch =)
* I would prefer to have 'writeQueuedMessages' - take a size_t - which is the size we should write up to (and beyond) if we can: "Give me as much as you can up to fill up to and beyond <N> bytes"
*  I wonder if we need the wsd/ClientSession.cpp deduplicate behavior for mobile though from only popping one message at a time (?) possibly we'd want to have a new 'getFillSendBufferSize()' - which un-conditionally returns the actual _sendBufferSize rather than MAX_INT for mobile.
 
Otherwise - looks very promising; I look forward to 'feeling' the result ;-)
Thanks !

Ashod:
@mmeeks Great! Thanks for the notes. Yes, we should really communicate how much data to write into the buffer and do this more efficiently.

Let me implement some of your suggestions and update this PR.

Thanks!

PR #1561, opened 2021-03-02
wsd: cap the socket read buffer size

Ashod:
Previously we read as much data as we could
from the socket. This is good if we only have
a single socket and unlimited memory. However
in practice we may end up hogging resources
for one fast connection, filling megabytes
of buffer in memory instead of tending to
other sockets and keeping the buffers from
growing too much.

We break the read when we hit 512 KB of data.
This limit seems reasonable, mostly due to
diminishing returns after some point.

Change-Id: I7de8f202a0d43872501de42d2f77dab8e0f2aaf0

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -970,7 +970,6 @@ public:
         do
         {
             // Drain the read buffer.
-            // TODO: Cap the buffer size, lest we grow beyond control.
             do
             {
                 len = readData(buf, sizeof(buf));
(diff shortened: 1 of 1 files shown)


mmeeks (changes requested):
I can't see that stopping us consuming memory for very verbose producers. The poll will spin, we always have POLLIN set - so we'll get the next half-megabyte on the next poll. Also - as a general view of the data-flow, there are almost never large reads from our remote clients (except from the Kit process) - the event flow should be mostly small and textual. When it comes to the Kit - we want to get that data across as fast as possible - but even so, I'd expect us to saturate a kernel write buffer, context switch and read anyway.

So - not sure that this gives any real benefit as of now; and not a fan of adding those random numbers =) In the longer term re-working our socket code to stream large chunks of data to files - particularly as we are parsing large incoming stream would make sense, but practically I don't think this is a real issue (or effective fix) as of now.

Sorry ! =)

Ashod:
@mmeeks Thanks. I suppose in practice it won't make a difference (we agree on the analysis). My attempt was to cap the vector growth (the data will have to be consumed before the next poll).

Anyway, I'll remove the TODO and instead say we want to drain the HW buffer asap and leave it at that.

Thanks!

PR #1572, opened 2021-03-03
Draw: Display Presentationbar for Insert/Duplicate/Delete Pages

merttumer:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Would be great to actually remove the "Presentation" button completely if possible; but otherwise let's get it in - thank you! :-)

PR #1579, opened 2021-03-03
private/michael/firefox fix

mmeeks:
- Firefox 42 ESR workaround.
- Tolerate a null this.dialog on the map at start.


mmeeks:
Abandoning duplicate.

PR #1585, opened 2021-03-04
wsd: terminate Kit when disconnected while loading

Ashod:
When loading hasn't completed yet, DocBroker can't
simply send a message to Kit that the session is
disconnected, because there would be no UI or
handlers for events if that were the only session.

Of course letting the document load, only to close
it, is also less than helpful to resource consumption.

Here, we recognize this case and simply terminate the
loading by killing the Kit process altogether to
avoid any potential hangs or resource wastage.

Change-Id: Ia72de715cc6238831c244444bb47b417b9b1e1a4

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1553,9 +1553,26 @@ void DocumentBroker::disconnectSessionInternal(const std::string& id)
             {
                 hardDisconnect = it->second->disconnectFromKit();
 
-                // Let the child know the client has disconnected.
-                const std::string msg("child-" + id + " disconnect");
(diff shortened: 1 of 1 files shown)


kendy:
Makes sense to me, let's get it in, thank you @Ashod !

mmeeks:
Looks good to me - but I'd expect another fix too =) what was the root cause of the 100% CPU burn when dis-connecting during load ? is it a dead socket in a poll or ? ...

mmeeks:
I lie, I read the detailed comment =) possibly that belongs in the commit msg rather than the code though =)

PR #1597, opened 2021-03-04
Allow jwt token expiry to be configurable.

mmeeks:
Change-Id: Ifd2b078adfa3106765f5b54c0397e04ab66fc804

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/Auth.cpp
+++ b/wsd/Auth.cpp
@@ -30,6 +30,7 @@
 #include <Log.hpp>
 #include <Util.hpp>
 #include <Protocol.hpp>
+#include "LOOLWSD.hpp"
 
 using Poco::Base64Decoder;
 using Poco::Base64Encoder;
@@ -154,6 +155,8 @@ bool JWTAuth::verify(const std::string& accessToken)
 
         std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
         std::time_t curtime = std::chrono::system_clock::to_time_t(now);
+
+        LOG_TRC("JWT: cur time " << curtime << " vs. " << decodedExptime);
         if (curtime > decodedExptime)
         {
             LOG_INF("JWTAuth:verify: JWT expired; curtime:" << curtime << ", exp:" << decodedExptime);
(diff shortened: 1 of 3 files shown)


wsd/Auth.cpp
```diff
@@ -191,14 +194,15 @@ const std::string JWTAuth::createPayload()
...
+    int expirySeconds = LOOLWSD::getConfigValue<int>("security.jwt_expiry_secs", 1800);
+    const std::string exptime = std::to_string(curtime + expirySeconds);
 
     // TODO: Some sane code to represent JSON objects
     const std::string payload = "{\"iss\":\"" + _iss + "\",\"sub\":\"" + _sub
                               + "\",\"aud\":\"" + _aud + "\",\"nme\":\"" + _name
                               + "\",\"exp\":\"" + exptime + "\"}";
 
-    LOG_INF("JWT Payload: " << payload);
+    LOG_INF("JWT Payload: " << payload << " expires in " << expirySeconds << "seconds");
```
Ashod, line 205: Intentional to have "seconds" stuck to the numbers (i.e. no space) ?

wsd/Auth.cpp
```diff
@@ -191,14 +194,15 @@ const std::string JWTAuth::createPayload()
 {
     std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
     std::time_t curtime = std::chrono::system_clock::to_time_t(now);
-    const std::string exptime = std::to_string(curtime + 1800);
+    int expirySeconds = LOOLWSD::getConfigValue<int>("security.jwt_expiry_secs", 1800);
```
Ashod, line 197: I wonder if this timeout gets extended when the admin actively interacts and moves between the different views (I guess not), because it would encourage them to extend the default timeout to avoid re-logging in every 30 minutes, regardless of how active they had been (negating the purpose of a timeout).

Ashod (approved):
Looks good, thanks!

wsd/Auth.cpp
mmeeks: Interesting; the websocket should be persistent and give interactive updates over time - so a session doesn't expire; but I forget if switching views in the UI re-connects, or gets a new websocket or something: hopefully not. Anyhow - it's no regression for now.

PR #1604, opened 2021-03-05
cypress: NC: disable failing Save as test.

tzolnai:
The feature it now broken. We can reenable these tests,
when it's fixed again.

Signed-off-by: Tamás Zolnai <tamas.zolnai@collabora.com>
Change-Id: I7cca152b4d19e1881bfcc195757605ffaa5b0f3c


eszkadev:
hi, I think I have fix for the issue: https://github.com/CollaboraOnline/online/pull/1605

PR #1605, opened 2021-03-05
Don't allow to switch ui mode on mobile

eszkadev:
Changing ui mode on mobile causes problems eg. after save as
or reloading the document.
We don't support notebookbar on mobile also

Change-Id: Iaa8d80dbdd139833c0505354be870f33b4f95413


tzolnai (approved):
Looks good to me! It fixes the issue.

Issue #1432, opened 2021-02-08
iOS App: Writer, Calc & Impress: comment function does not work reliably while keyboard is already active


Gianni154:
**Describe the bug**
If you insert a comment in any document (whether writer, calc or impress) via Notebookbar > Insert > Comment and then you click on the comment field while the keyboard is already active, it causes problems. It can cause the Collabora app to crash. When reproducing, my iPad even rebooted once.

**To Reproduce**

1. Open the Collabora iOS App
2. Open any document
3. Make sure that the keyboard is already displayed
3. On the notebookbar click on Insert
4. Click on Comment to open the comment dialog
5. Click on the comment field

**Expected behavior**
The comment field can be filled in and you can create a comment.

**Actual behavior**
Creating a comment is not possible. The keyboard has a strange behavior. It can crash the iOS Collabora app as well as the device itself (iPad). 

**Screenshots**

![comment](https://user-images.githubusercontent.com/78140364/107212482-50eda900-6a07-11eb-828d-b164061a0409.gif)

Unfortunately, you can't see in this GIF that the keyboard sometimes disappears. Below I have linked another video Google Drive. 
Link: https://drive.google.com/file/d/1Osk-BACANK5AZ0xg7CYIt0yN_LTTugK5/view

**Device**
 - iPad OS 14.4
 - iPad 6th generation
 - iOS Collabora App Version 6.4.5 (2)

tml1024:
Yep, can reproduce in a fresh build. Keyboard-related bugs are in general complicated, so this might be hard to fix.

tml1024:
Note to self: If the on-screen keyboard is displayed when you tap the "Comment" button, you can't type into it. If the on-screen keyboard is not displayed when you tap the button, you can.

tml1024:
This bug is related to the use of CollaboraOnlineWebViewKeyboardManager to handle on-screen keyboard for "normal" input, but then for this input, which is into a plain HTML dialog or some such, we apparently *don't* want to do that, and the state gets all confused. Hacking on this and trying to come up with some not too insane solution. (But I expect that will be impossible, and even more insanity will have to be added...)

tml1024:
[...]

PR #1607, opened 2021-03-05
Don't use CollaboraOnlineWebViewKeyboardManager after all

tml1024:
Unclear whether it is actually useful or not any longer. Quite
possibly cross-platform changes in the JavaScript since
CollaboraOnlineWebViewKeyboardManager was introduced have fixed the
problems that it was solving.

Fixes https://github.com/CollaboraOnline/online/issues/1432

Change-Id: Ifa000d0a7d0a3dc87527536719063b368266f038

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Seems reasonable if it helps =) we re-tested the older issues it addressed ?

PR #1614, opened 2021-03-06
leaflet: cursor: avoid first character not updating cursor position

lpranam:
Change-Id: I9cc00e5d50d21384155b16b60318090e4de70ac2

* Target version: master 

### Summary
problem:
In the android keyboard when you try to erase in an empty area
and then enter some character,
The first character will likely travel with the cursor,
And that is caused because after entering the first character
cursor position is never updated by keyboard (I know it is strange)
so here we manually correct the position

steps to reproduce:
1. type "=sum" in formula bar or an empty odt (or even just "=s" is okay)
2. now tap on backspace for total character +1 times (in this case 5 or 3 times)
3. now type "=s" it will become "s="
but this can be reproduced with any special characters (i.e: }, + etc...)

### Checklist

### DIFF
--- a/loleaflet/src/layer/marker/TextInput.js
+++ b/loleaflet/src/layer/marker/TextInput.js
@@ -701,6 +701,15 @@ L.TextInput = L.Layer.extend({
 		// remove leading & tailing spaces.
 		content = content.slice(1, -1);
 
+		// In the android keyboard when you try to erase in an empty area
+		// and then enter some character,
+		// The first character will likely travel with the cursor,
+		// And that is caused because after entering the first character
+		// cursor position is never updated by keyboard (I know it is strange)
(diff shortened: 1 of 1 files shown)


mmeeks (changes requested):
Strange and scary - but then, all text input is. The Android keyboard acts in an asynchronous fashion - and we have to fight it in various ways to try to force it to get its cursor in the right place. This looks plausible and reasonably risk free to me. Since we have had problem on Firefox with the cursor movement when the widget is not focused or has just been created - it would be good to have a setSelectionRange wrapper say 'setCursorPosition?' that catches the exception and is used in both places =)

Otherwise great.

mmeeks:
Looks great - thanks Pranam ! =)

PR #1633, opened 2021-03-08
fixes related to canvas overlays

dennisfrancis:
* Resolves: # 
* Target version: master 

### Summary
Fixes related to canvas overlays.

### Checklist


dennisfrancis:
@gokaysatir I think it is best to have this reviewed in master and amend here before merge to avoid follow-up fixes backported to co-6-4. So a review from you would be great.

PR #1644, opened 2021-03-09
config: add "enable_macros_execution" item

hcvcastro:
- config: add "enable_macros_execution" item
- config: add "macro_security_level" item


eszkadev (approved):
looks good, only small issue with xml comment

PR #1660, opened 2021-03-10
mobilewizard: check if "_docLayer" object exists

hcvcastro:
Change-Id: I875ebed7d2e8d0f8813c39a221e945988d511043
Signed-off-by: Henry Castro <hcastro@collabora.com>

PR #1663, opened 2021-03-10
Mobile: Edit btn: show only if no initial dialog is present

pedropintosilva:
in sequence of PR #1660

Change-Id: I542c9092f88196c9e59d675ff71a7f44dfb8f67a


pedropintosilva:
@eszkadev @hcvcastro : This seems to work for me locally but I'm not sure if it's the best way to:
- show `mobile-edit-button` button only if no initial dialog is present at loading time (as it happens with macro security dialog).

pedropintosilva:
Tested :star2:  , thanks!

PR #1671, opened 2021-03-10
loleaflet: fix "offsetWidth" property value of the sidebar panel

hcvcastro:
Unfortunately when show/hide sidebar the "offsetWidth" property
works only if the element is visible otherwise it returns parent
empty size.

To fix add temporary a visibility property, then obtain the sidebar
"offsetWidth" property and update document container size.

Change-Id: I8e1c821518a4c155340edcfc8f2313098edff8b0

### DIFF
--- a/loleaflet/src/control/Control.LokDialog.js
+++ b/loleaflet/src/control/Control.LokDialog.js
@@ -1642,9 +1642,22 @@ L.Control.LokDialog = L.Control.extend({
 		}
 
 		var wrapper = L.DomUtil.get('sidebar-dock-wrapper');
-		if (wrapper && wrapper.offsetWidth)
-			this._map.options.documentContainer.style.right = wrapper.offsetWidth + 'px';
-		else
+		if (wrapper) {
+			var offsetWidth;
+			var panel = L.DomUtil.get('sidebar-panel');
+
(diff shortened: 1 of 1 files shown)


loleaflet/src/control/Control.LokDialog.js
```diff
@@ -1642,9 +1642,22 @@ L.Control.LokDialog = L.Control.extend({
...
-			this._map.options.documentContainer.style.right = wrapper.offsetWidth + 'px';
-		else
+		if (wrapper) {
+			var offsetWidth;
+			var panel = L.DomUtil.get('sidebar-panel');
+
+			if (panel.style.display !== 'none')
+				offsetWidth = wrapper.offsetWidth;
+			else {
+				panel.style.visibility = 'hidden';
```
eszkadev, line 1652: I don't understand something. If panel has `display == none`, shouldn't be the final result always equal to ~0px? do we need to calculate that value?

hcvcastro: > I don't understand something. If panel has display == none, shouldn't be the final result always equal to ~0px? do we need to calculate that value?


Setting the hidden visibility we force the browser to calculate the layout panel so the offsetWidth is not 0px.

PR #1677, opened 2021-03-11
avoid incorrect overlay painting due to big document coordinates

dennisfrancis:
canvasRenderingContext2D.translate() and friends only seem to function
with numbers < 2^24, at least on Chrome 89.

So when the pane is being drawn where this limit is hit, lets do the
translation ourselves.

But in the case of zoom-animation more complicated transformations are
needed. Till we implement the manual matrix transformations correctly,
lets not animate overlay objects when the view area has hit this limit.

Change-Id: Ib48d6029ba1c6f1d7ade35060baf4ebe3e0a5fd3

### TODO

- [ ] ...

### Checklist


kendy:
@dennisfrancis: Let's do this, thank you! :-)

PR #1684, opened 2021-03-11
fixes: ghost pages/tiles in writer

dennisfrancis:
The issue is: On creating a page break on a empty writer file and
hitting backspace, the deleted page or parts of it is still rendered.

We get a 'status' message containing the document size every time it
changes. So avoid the issue do not draw tiles that are not within the
document area.

Change-Id: I88d1715e861f068326d28d2930d01e3b5355be02

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir:
I tested it and it is working, thanks for the solution.

PR #1685, opened 2021-03-11
Optimize for overlayfs by forcing an initial copy to linkable/

mmeeks:
built-in hard-link support performs extremely badly otherwise.

Change-Id: I26d5080590538cb6342d64e2e0a4023bbc84c427

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -16,6 +16,8 @@
 #include <dlfcn.h>
 #ifdef __linux__
 #include <ftw.h>
+#include <sys/vfs.h>
+#include <linux/magic.h>
 #include <sys/capability.h>
 #include <sys/sysmacros.h>
 #endif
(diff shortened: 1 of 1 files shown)


mmeeks:
Gabriel - it's not totally obvious to me that this catches your use-case; I'd love to see what MAGIC we get from fstype on your system, and check that we catch it - my comment is prolly inaccurate too - but ... at least this is how I'd do it.

I would like to keep that code reasonably condition free there - it's already complex; I have some evil plans to have separate jails and forkits for different WOPI-like servers that interact with us in future for one reason and another so want to foul it up myself later =)

Thoughts appreciated.

mmeeks:
Andras - what are we using for our app-images and so on ? do we need a configuration setting to turn this off ? thoughts appreciated.

timar:
> Andras - what are we using for our app-images and so on ? do we need a configuration setting to turn this off ? thoughts appreciated.

File system of AppImage is read-only. We create jails in /tmp, and presumably we copy. 
For docker we prefer bind mounting. But it requires privileges that are not available in all use cases. So I think it's great that Gabriel actually measured that this patch helped. I have no objections.

timar (approved):

timar:
I'm almost sure that we can ignore the "with gcc-4.9.2 — FAILURE". I'll double check with @yunusem but I think that check is obsolete now.

gmasei11:
Basically, the detection of OverlayFS was working but fixed some issues:
1. A copy is performed inside jail every time a kit is spawned. Now it copies to linkable.
2. Change of owner was skipped. Now it performs it.

Also, it seems that the initial owner of the systemplate and lokit folders is still lool. I thought that this was fixed (@timar). Am I missing something ? If not then needs to be fixed because it is important.

gmasei11:
> I don't think we want this one - it looks like it forces us to copy for every jail, and we only want to copy once - surely?

I think you're partially right. I think that it is useless exactly because of the call to link :D Right ?

gmasei11 (approved):

PR #1690, opened 2021-03-11
macro dialog: hide from menus if disabled

eszkadev:


hcvcastro:
Unfortunately, I have to reload 2 times to take effect, I think it is by cache browser.
What do you think about sending uno:RunMacro status back to client side?

eszkadev:
>  What do you think about sending uno:RunMacro status back to client side?

What do you mean? uno:RunMacro sfx slot state (disabled/enabled) ?
It's always enabled as we can open it even in LibreOffice when execution is disabled - so would need to implement status method for it in the core. I think it will not be changed dynamically during run-time as setting is in the loolwsd.xml so no need to use slot state.

hcvcastro:
> It's always enabled as we can open it even in LibreOffice when execution is disabled - so would need to implement status method for it in the core. I think it will not be changed dynamically during run-time as setting is in the loolwsd.xml so no need to use slot state.

Right, no change dynamically, so it looks like the minor development issue.

Looks good in general

hcvcastro (approved):

PR #1693, opened 2021-03-12
wsd: avoid using invalid DocBroker for clipboard

Ashod:
The clipboard content is handled by the DocBroker,
if it it's available, otherwise, we fallback to the
cached clipboard entries.

When DocBroker has already terminated polling and
is no longer valid, we may end up using it for
the clipboard. In this case, we don't send anything
back, since there is no SocketPoll thread to
process the request or the socket data plumbing.

We avoid this by simply using the fallback (cache)
when we don't have a valid DocBroker.

Change-Id: Idf7f27b453605d173dd8e40d4a5d101f50355468

### DIFF
--- a/wsd/LOOLWSD.cpp
+++ b/wsd/LOOLWSD.cpp
@@ -2767,7 +2767,14 @@ private:
             if (it != DocBrokers.end())
                 docBroker = it->second;
         }
-        if (docBroker)
+
+        // If we have a valid docBroker, use it.
+        // Note: there is a race here as DocBroker may
+        // have already exited its SocketPoll, but we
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks very sensible to me and certainly much better. I worry a bit about the childProcess->isAlive() in isAlive(); ideally we would use _stop directly I guess (?) and not the marked for destruction - since we slurp the clipboard data as our last act before death.

I wonder if our DocumentBroker ::poll should have:

    // Flush socket data first.
...
    while (_poll->getSocketCount() || _poll->hasCallback)

ie. adding some way to detect if there is a callback pending we should process (?)

Anyhow - fine for now.

PR #1697, opened 2021-03-12
Backport of zoom animation fixes

dennisfrancis:
* Target version: co-6-4

### Summary
* Fix overlay drawing on headers when animating zoom when view is tied to cell A1.
* Patch to avoid 2 branches of code to do coordinate translation/scaling for easier testing of overlays.

### Checklist


gokaysatir (approved):
Tested it, couldn't see any issues. Thanks for the patch :)

PR #1704, opened 2021-03-12
private/pedro/porting NB vertical align content n partial revert

pedropintosilva:
- NB: vertical-align content
- Partly revert cb27d2c ..


pedropintosilva:
Shouldn't be ported at all

PR #1708, opened 2021-03-12
private/pedro/porting sidebar kitconfig

### LINKED ISSUE #786: some sidebar icon are blured
compare to the Character icons, the paragraph icons are all blured. I think the issue is that collabora desktop edition use collibre_svg icon theme. I recommend to use the .png icons.

2020-11-26 14_51_55-Nextcloud

pedropintosilva:
- Sidebar should use .png icons instead of .svg fix #786
- change default sidebar icon size to use 24px icons (like in toolbars)

### DIFF
--- a/loolkitconfig.xcu
+++ b/loolkitconfig.xcu
@@ -25,11 +25,8 @@
 <!-- Enable thumbnail generation by default (disabling saves CPU time) -->
 <item oor:path="/org.openoffice.Office.Common/Save/Document"><prop oor:name="GenerateThumbnail" oor:op="fuse"><value>true</value></prop></item>
 
-<!-- Use the collabora_svg theme for the sidebar -->
-<item oor:path="/org.openoffice.Office.Common/Misc"><prop oor:name="SymbolStyle" oor:op="fuse"><value>collabora_svg</value></prop></item>
-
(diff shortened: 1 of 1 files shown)


kendy:
@pedropintosilva: IIRC the motivation to use SVG was due to hi-dpi, can you please check with hi-dpi turned on?

pedropintosilva:
So have tested this and I don't think we can merge this to the stable branch without both further testing on hiDPI monitor and a condition just for that case. What we know:

- Relaying on the on svg folder from core side is not ideal, the assets arrive with some poor render (both on Lo and HiDPI)
- The initial merged PR (on master) by @Andreas-Kainz solves this situation for LoDPI 
- I have also advised to play with `SidebarIconSize` in the ` loolkitconfig.xcu` (successfully done by @Andreas-Kainz )

With this PR it looks better than before on LoDPI
image

But it looks worse on HiDPI
Screenshot_2021-04-27 Online Editor

I wonder if we can choose to load svgs or other png folder only for HiDPI

Andreas-Kainz:
As there is the new sidebar implementation, I hoped with the new implementation, cool will render direct the lc_*.svg files so we don't have an HiDiPi or blur issue any more. Isn't it @eszkadev ?

PR #1714, opened 2021-03-13
Calc: Improve "cell cursor jump"-Online part.

gokaysatir:
Cell cursor shouldn't jump to end when whole row or column selected.
"undo delete row" and "undo delete column"  actions shouldn't cause cell cursor to jump to the end.

Change-Id: I51f077c5b3fdd5c7a610631f1e0259775278ab53

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/TileLayer.js
+++ b/loleaflet/src/layer/tile/TileLayer.js
@@ -2003,6 +2003,62 @@ L.TileLayer = L.GridLayer.extend({
 		this._updateReferenceMarks();
 	},
 
+	_getStringPart: function (string) {
+		var code = '';
+		var i = 0;
+		while (i < string.length) {
+			if (string.charCodeAt(i) < 48 || string.charCodeAt(i) > 57) {
+				code += string.charAt(i);
+			}
+			i++;
+		}
+		return code;
+	},
+
+	_getNumberPart: function (string) {
+		var number = '';
+		var i = 0;
+		while (i < string.length) {
+			if (string.charCodeAt(i) >= 48 && string.charCodeAt(i) <= 57) {
+				number += string.charAt(i);
+			}
+			i++;
+		}
+		return parseInt(number);
+	},
+
+	_isWholeColumnSelected: function (cellAddress) {
+		var startEnd = cellAddress.split(':');
+		if (startEnd.length === 1)
+			return false; // Selection is not a range.
+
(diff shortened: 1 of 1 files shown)


dennisfrancis:
I have a few comments regarding the celladdress message fix (but both of these are addressable as follow-up later when there is time), please merge when CI is done. Thanks.

1. When I debug the problem without the patch, I see that the issue is due to a wrong cellselectionarea message just before it sends the correct one of full row/column after doing 'undo'. So I'd say it is better see why that happens and fix it which is the root cause (which is in core) and we won't need a fix in online. 
2. If the point 1 is harder to do, then we need this workaround in online. Regarding the current online patch, it is better to have a cell address parser TS class or function for that which can potentially be mocha tested. This way we can avoid having generic functions/methods clutter the *Layer classes.

gokaysatir:
> I have a few comments regarding the celladdress message fix (but both of these are addressable as follow-up later when there is time), please merge when CI is done. Thanks.
> 
>     1. When I debug the problem without the patch, I see that the issue is due to a wrong cellselectionarea message just before it sends the correct one of full row/column after doing 'undo'. So I'd say it is better to see why that happens and fix it which is the root cause (which is in core) and we won't need a fix in online.
> 
>     2. If the point 1 is harder to do, then we need this workaround in online. Regarding the current online patch, it is better to have a cell address parser TS class or function for that which can potentially be mocha tested. This way we can avoid having generic functions/methods clutter the *Layer classes.

Thanks for the comment, i think we can add a cell-address parser. It will be useful:) I will add it when i have time.

loleaflet/src/core/Socket.js
```diff
@@ -929,7 +929,6 @@ L.Socket = L.Class.extend({
 	_tryToDelayMessage: function(textMsg) {
 		var delayed = false;
 		if (textMsg.startsWith('window:') ||
-			textMsg.startsWith('celladdress:') ||
```
tzolnai, line 932: I checked the cypress test failure and it looks like it is related to this line of change. This _tryToDelayMessage' purpose is to keep messages until the doc layer is initialized so we don't lose information during document load. It seems to me, after this change, we miss a celladdress message and so the address input field is not set correctly. It's empty as the error message also indicates.

gokaysatir: Thank you very much for this good detail :) I am trying it now.

PR #1715, opened 2021-03-13
Avoid static objects in LOOLWSD

Ashod:
Large and complex objects such as SocketPoll and its derivatives and users need controlled lifetime management to avoid destroying them out-of-order and especially after other statics are destroyed. One such external static is the Logging (and other) Poco instances that we can't control. A static object that logs anything at all can potentially crash if it runs after Poco is destroyed (partially or completely).

We hit such an issue on one of the CI boxes. The issue had been there, but a small change in the logging and other structures in the code resulted in consistent segfaults on said box everytime loolwsd exited. This failed the build, of course, which was enough to catch our notice.

Even though the issue was now in the limelight, it was still not reproducible elsewhere. It wasn't reproducible on my dev box either, and neither clang's ASAN, nor gcc's ASAN complained of any issues. In fact, it wasn't crashing anywhere else except for that one CI box. gdb on the broken CI box revealed the problem immediately. Stacktraces and other details in the commit logs.

This PR contains the patches that convert the global statics in loolwsd to unique_ptr instances and the lifetime management associated with it. We always make sure that the objects in question are created before use and that they are destroyed before we exit (at least for the normal workflow).

- wsd: capture and log exceptions during un/initialization
- wsd: move TerminatingPoll to Socket.hpp
- wsd: avoid static SocketPoll
- wsd: avoid static LOOLWSDServer instance
- wsd: convert static objects to unique_ptr
- wsd: move srv instance into LOOLWSD and rename
- make: delete loolwsd binary if it fails to cleanup

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -335,7 +335,10 @@ EXTRA_DIST = discovery.xml \
 if ENABLE_LIBFUZZER
 CLEANUP_COMMAND=true
 else
-CLEANUP_COMMAND=if test -s ./loolwsd; then echo "Cleaning up..." && ./loolwsd --disable-lool-user-checking --cleanup --o:logging.level=trace; fi
+# Use loolwsd to cleanup jails, if any. If it fails, we may have a broken/old loolwsd binary, remove it to rebuild.
+# A CI box may switch branches without cleaning up the binaries, if loolwsd from a broken branch is used here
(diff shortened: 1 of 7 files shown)


Ashod:
@mmeeks: follow-up to move them back into LOOLWSD.cpp: https://github.com/CollaboraOnline/online/pull/1727

vmiklos:
If you backport this PR later, please also include https://github.com/CollaboraOnline/online/pull/1735. Thanks!

PR #1718, opened 2021-03-15
Backport: touch: do not show onscreen keyboard post zoom

dennisfrancis:
Fix description:
On setting the view area/zoom in map after the pinch has finished the
textarea for the text input is getting focus and the keyboard pops up.
This is prevented by temporarily disabling the textarea while setting
the view area/zoom in the map.

Change-Id: I193d36c7366c9c0a200c00b2966c5d31c4f2df2c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/marker/TextInput.js
+++ b/loleaflet/src/layer/marker/TextInput.js
@@ -133,6 +133,14 @@ L.TextInput = L.Layer.extend({
 		this._map.removeLayer(this._cursorHandler);
 	},
 
+	disable: function () {
+		this._textArea.setAttribute('disabled', true);
+	},
+
+	enable: function () {
+		this._textArea.removeAttribute('disabled');
+	},
+
 	_onPermission: function(e) {
 		if (e.perm === 'edit') {
 			this._textArea.removeAttribute('disabled');
--- a/loleaflet/src/map/Map.js
+++ b/loleaflet/src/map/Map.js
@@ -540,6 +540,23 @@ L.Map = L.Evented.extend({
 		return this._inZoomViewPanning;
 	},
 
+	setZoomViewPanning: function (value) {
+		var docLayer = this._docLayer;
+		if (!docLayer)
+			return;
+		this._inZoomViewPanning = value;
+
+		if (!docLayer.isCursorVisible())
+			return;
+
+		if (this._inZoomViewPanning) {
+			this._textInput.disable();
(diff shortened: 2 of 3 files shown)


mmeeks (commented):
Odd; is there really not an event from the core caused by something that we send to it that triggers the focus change ? =)
What happens if you have the keyboard popped up already - and you zoom in in the area that you can see to better see the text - does the keyboard appear / disappear ? If w ehad the K/B before and zoomed it shoudl not disappear; but if it was not there - it should not appear post pinch zoom =)

Thanks !

gokaysatir:
Hello,
   I tested it. It works. It hides the keyboard even if it is visible before the zoom in/out as @mmeeks mentioned.
   Thank you for the fix.

mmeeks:
Hmm - so to check this is what I expect:
   * starting pinch-zoom with keyboard up -> the keyboard stays visible afterwards
   * starting pinch-zoom without the keyboard up -> the keyboard does not pop-up.
Both - when not in read-only mode.
Same needs testing for a 'pan' action.
If so - good, lets merge it =)

dennisfrancis:
@mmeeks: With the current patch 
"starting pinch-zoom without the keyboard up -> the keyboard does not pop-up." is working. 
but not the other requirement :
"starting pinch-zoom with keyboard up -> the keyboard stays visible afterwards"
the situation is the same for pan action.

This is because the keyboard state is not maintained correctly and returned by canAcceptKeyboardInput() in the already complex TextInput class. It is easy to get this boolean state get off sync with the keyboard state for example by focusing and closing the keyboard forcibly. Any fix using the 'canAcceptKeyboardInput()' this runs the risk of keyboard/focus regressions because we can't as far as I'm aware cannot directly check the state of the on screen keyboard in chrome browser.

mmeeks:
Interesting; well if the behavior is not a regression - fine; how did it work pre-canvas ? and what non-tile events are sent to/from the core during this interaction (pan eg.) =)

thebearon:
> Interesting; well if the behavior is not a regression - fine; how did it work pre-canvas ?

In a 6.4.3 mobile app, and in 4.2 online, zooming always brought up the virtual keyboard, and panning always hid it.

mmeeks:
Sounds good - lets merge it then - thanks ! =)

PR #1738, opened 2021-03-16
private/pedro/mobile macro selector layout style

pedropintosilva:
- CSS var: Add one more gray and replace instances
- Mobile: mobilewizard: use outline style for cancel btn
- Mobile: Macro Selector: Layout improvements and style..


loleaflet/src/control/Control.JSDialogBuilder.js
```diff
@@ -213,6 +213,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		];
 
 		this._currentDepth = 0;
+		this.helpmacroInnerText = '';
```
eszkadev, line 216: please revert changes in this file I can do the same in the core. we should avoid changing structure like this in JS code.

eszkadev (changes requested):

eszkadev (approved):

PR #1741, opened 2021-03-16
IE11 support

eszkadev:
Added polyfills needed for IE11

### DIFF
--- a/loleaflet/src/layer/tile/CanvasSectionContainer.ts
+++ b/loleaflet/src/layer/tile/CanvasSectionContainer.ts
@@ -265,6 +265,13 @@ class CanvasSectionContainer {
 		return this.clearColor;
 	}
 
+	/**
+	 * IE11 doesn't support Array.includes, use replacement
+	*/
+	private arrayIncludes<T> (array: Array<T>, element: T) {
+		return array.indexOf(element) >= 0;
+	};
+
 	private clearMousePositions () {
 		this.positionOnClick = this.positionOnDoubleClick = this.positionOnMouseDown = this.positionOnMouseUp = this.dragDistance = this.sectionOnMouseDown = null;
 		this.touchCenter = null;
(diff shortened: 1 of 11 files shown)


mmeeks:
Had a play. It is a lot better and at least renders something and I can type into writer =)

When I try to select text or an object in writer - I get an exception. In impress selecting a chart in a sheet gives me:
if (grandParentID.includes('sidebar-panel'))
"object doesn't support includes"
and a similar error from 'dragStart' trying to move slides in the slide sorter

Calc startup (on PC) gives:

L.Control.JSDialogBuilder.getMenuStructureForMobileWizard = function(menu, mainMenu, itemCommand) {
	if (itemCommand.includes('sep'))

exception on that- and no row/column headers.

So a bit more work required I think.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -265,6 +265,37 @@ class CanvasSectionContainer {
...
+			return false;
+		}
+		var n = parseInt(arguments[1]) || 0;
+		var k;
+		if (n >= 0) {
+			k = n;
+		} else {
+			k = len + n;
+			if (k < 0) {k = 0;}
+		}
```
gokaysatir, line 286: Hello, so IE doesn't support "includes" :) Thank you for the fix.

I think there is something wrong here. We can set n = k = 0 directly. You have 1 more variable in parameters. arguments[1] is not the "fromIndex" variable. We may not need the variable "n".

Also, TypeScript has type "array". We don't need to cast it to object.

We can also write something like:

var len: number = array.length; // without "|| 0" part.
var k: number = 0;

With your code, behaviour will seem fine. Because search objects are not numbers. If we would search a number in an array, there would be bugs.

I also read the mozilla page. Thank you for the link.

If you want to add "fromIndex" functionality, we need one more variable. I think it's okay without it. These are all your's to decide:)

Thanks :)

eszkadev:
> Had a play. It is a lot better and at least renders something and I can type into writer =)
> 
> When I try to select text or an object in writer - I get an exception. In impress selecting a chart in a sheet gives me:

This is the first part. Just wanted to get some feedback from @gokaysatir wrt typescript polyfill code.
This PR covers only loading Writer.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
eszkadev: @gokaysatir thanks for review. Now I found out that we can replace the `includes` with just `indexOf` :) So no need for implementing on my own.

mmeeks:
Makes perfect sense; typing in writer was working surprisingly nicely =) the canvas stuff looked crisp & joined up too which was great.

loleaflet/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: Okay, thank you for the fix :)

gokaysatir:
Thank you for the fixes :)

PR #1750, opened 2021-03-17
Mobile: Macro Selector: Only show ellipsis when in empty state

pedropintosilva:
- Display ellipsis inside of #command element when a library is still
  not selected
	- To communicate clearly to the user that there is still stuff to be
	selected
	- But also to hide that visual (ellipsis) when there is content to show

Change-Id: I5d0a1e627de9e83908c2d79e0d52abd428f0280c


loleaflet/src/control/Control.JSDialogBuilder.js
```diff
@@ -1893,6 +1893,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				var header = L.DomUtil.create('th', builder.options.cssClass, headers);
 				header.innerText = data.headers[h].text;
 			}
+		} else if (parentContainer.id == 'MacroSelectorDialog' && table.id == 'commands') {
+			L.DomUtil.addClass(table, 'empty');
```
eszkadev, line 1897: maybe better to make it for all treeviews?

eszkadev: I mean add class empty always. and when you decide that it should be only for MacroSelectorDialog put it in the css selector.  (this will make this code generic, without exceptions for specific dialogs which makes it harder to maintain later)

eszkadev (approved):

PR #1755, opened 2021-03-17
TileLayer: Update view cursor.

gokaysatir:
Even if the modifier view is different.

Change-Id: I32d0c12ae3c9c61da95f311dd8471cc4b7be5769

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Let's get it in, thank you for the fix! :-)

PR #1760, opened 2021-03-18
canvas: avoid flicker when zooming

lpranam:
Change-Id: I9fb8e463b69288fb3e9e6033eb04fd4b85351ea1

* Target version: master 

### Summary
flicker was caused because when we zoom we cleared the entire canvas at once
this patch would make sure instead of clearing the entire canvas,
we clear the area for a tile as we receive a tile

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CanvasSectionContainer.ts
+++ b/loleaflet/src/layer/tile/CanvasSectionContainer.ts
@@ -105,6 +105,8 @@
 */
 
 // This class will be used internally by CanvasSectionContainer.
+
+declare var L: any;
 class CanvasSectionObject {
 	context: CanvasRenderingContext2D = null;
 	myTopLeft: Array<number> = null;
@@ -1159,9 +1161,17 @@ class CanvasSectionContainer {
 	private drawSections () {
 		this.context.setTransform(1, 0, 0, 1, 0, 0);
 		this.context.fillStyle = this.clearColor;
-		this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
-		this.context.fillRect(0, 0, this.canvas.width, this.canvas.height);
 
+		// When we zoom in we try to avoid the entire canvas repaint
+		// instead we just clear the surrounding of the doc (remnant from previous zoom)
+		if (L.Map.THIS.getTileSectionMgr() && L.Map.THIS.getTileSectionMgr()._zoomChanged) {
+			this.context.clearRect(0, 0, -this.documentTopLeft[0], this.canvas.height);
+			this.context.fillRect(0, 0, -this.documentTopLeft[0], this.canvas.height);
+		}
+		else {
+			this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
+			this.context.fillRect(0, 0, this.canvas.width, this.canvas.height);
+		}
 		this.context.font = String(20 * this.dpiScale) + "px Verdana";
 		for (var i: number = 0; i < this.sections.length; i++) {
 			if (this.sections[i].isLocated) {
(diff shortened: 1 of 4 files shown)


loleaflet/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -1159,9 +1161,15 @@ class CanvasSectionContainer {
 	private drawSections () {
 		this.context.setTransform(1, 0, 0, 1, 0, 0);
 		this.context.fillStyle = this.clearColor;
-		this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
-		this.context.fillRect(0, 0, this.canvas.width, this.canvas.height);
 
+		if (L.Map.THIS.getTileSectionMgr() && L.Map.THIS.getTileSectionMgr()._zoomChanged) {
+			this.context.clearRect(0, 0, -this.documentTopLeft[0], this.canvas.height);
```
dennisfrancis, line 1166: Best to add a comment here about clearing just the left margin area in case of non-calc app instead of the entire canvas when zoom has changed.

loleaflet/src/layer/tile/TilesSection.ts
```diff
@@ -164,7 +164,10 @@ class TilesSection {
 			this.context.fillStyle = this.containerObject.getClearColor();
 			this.context.fillRect(offset.x, offset.y, ctx.tileSize.x, ctx.tileSize.y);
 		}
-
+		if (L.Map.THIS.getTileSectionMgr() && L.Map.THIS.getTileSectionMgr()._zoomChanged) {
+			this.context.clearRect(offset.x, offset.y, ctx.tileSize.x, ctx.tileSize.y);
+			this.context.fillRect(offset.x, offset.y, ctx.tileSize.x, ctx.tileSize.y);
```
dennisfrancis, line 169: This function "paintSimple" only takes care of tile painting for non-Calc apps. For calc it uses paintWithPanes() which handles calc case with freeze panes. So you need a similar change there too.

dennisfrancis (commented):
For desktop we have no zoom animation yet. For mobile, the zoom animation uses upto 4 (most general freeze panes case) offscreen canvases to work. See the zoom-animation code in
L.TileSectionManager methods - _zoomAnimation(), _calcZoomFrameScale(), zoomStep()
zoomStepEnd(). For avoiding the flicker in mobile, we can use the same approach when we call setView() after the call to zoomEnd (clear just the tilearea as they arrive). 

For animating in desktop, I'd say lets call zoomStep for a few intermediate zoom levels in a setInterval() callback when zoom change is activated and in that when it does the last frame, call setView() on the map with the final zoom. The center for the zoom frames needs to be picked based on the final center we want. For example in PC online calc, the final center is determined in a particular way in L.Map's setDesktopCalcViewOnZoom().

loleaflet/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -1159,9 +1161,17 @@ class CanvasSectionContainer {
...
+		// When we zoom in we try to avoid the entire canvas repaint
+		// instead we just clear the surrounding of the doc (remnant from previous zoom)
+		if (L.Map.THIS.getTileSectionMgr() && L.Map.THIS.getTileSectionMgr()._zoomChanged) {
+			this.context.clearRect(0, 0, -this.documentTopLeft[0], this.canvas.height);
+			this.context.fillRect(0, 0, -this.documentTopLeft[0], this.canvas.height);
+		}
+		else {
+			this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
+			this.context.fillRect(0, 0, this.canvas.width, this.canvas.height);
+		}
```
gokaysatir, line 1174: This is a specific approach for a generic class. I will want to update this part in the future.

gokaysatir (approved):
I tested it. Works as expected. Thank you for the fix.

Issue #1285, opened 2021-01-20
Remove the ... at NB Labels


Andreas-Kainz:
In LibreOffice the three dots (...) for an dialog wasn't shown in cool they will be shown. 

I suggest to don't show the ... in cool.

Cool
![2021-01-20 13_09_34-ws-a3df5ab9-a090-4043-8293-148f004bb2e7_0 - noVNC](https://user-images.githubusercontent.com/8517736/105173056-c957fd00-5b20-11eb-8ba3-db47234127ce.png)
LibreOffice tab
![2021-01-20 13_11_07-ws-a3df5ab9-a090-4043-8293-148f004bb2e7_0 - noVNC](https://user-images.githubusercontent.com/8517736/105173204-fad0c880-5b20-11eb-8771-a0b2fdeee149.png)

Andreas-Kainz:
@pedropintosilva any idea how to remove them?

pedropintosilva:
> I suggest to don't show the ... in cool.

Yes, I agree and I have suggested the same some time ago on Telegram and I think in some meeting.

I also have discussed this with @eszkadev and I have inspect it. It seems it's not as easy as remove (hardcoded) because it's coming already like that from translated strings. Now the question is where there are coming from or that maybe we are getting the original strings not from the tabs but from the menu items on core. 

Nevertheless @eszkadev said he thinks it will be easier to make sure that no ellipsis are there programmatically without affecting any current strings from the weblate side etc . If he does beat us, maybe you or I can come up with a PR (I will keep you posted)

Andreas-Kainz:
If you check both screenshots come from cool. The insert tab screenshot is made by .JS and JSON and the layout tab come direct from core.

However I know that we already discussed, I only want to have an real issue for tracking.

pedropintosilva:
Yeah definitely is important to have this in some tracking issue, thanks : ) and true the structure is done there but the translated strings are then applied according to user's idiom (at least with the majority if not all the labels)

kendy:
@Andreas-Kainz: There are more variants of texts in the .xcu files for various commands, I suspect we are using a wrong one for the Notebookbar (?)  @eszkadev ?

pedropintosilva:
as seen in the my commits I have fixed at the time before the release the more flagrant cases, but it would be great to have it fixed in all other places.

It seems we are pulling from core the wrong kind of string from those uno commands (menu instead of tab bar). Fix: get the right type of string (from tab bar)

Online:
[...]

PR #1767, opened 2021-03-18
Notebookbar: File tab: Remove ellipsis from Print and _saveAs

pedropintosilva:
In sequence of issue #1285:
* https://github.com/CollaboraOnline/online/issues/1285#issuecomment-764556056
  * @kendy we are actually using the right variant
  which is the one used also in core tabbed layout but it also has those
  ellipsis
* Removing the ellipsis in the most visible case (File tab) for Print...
  and Save As... shouldn't be problematic as it does not had any extra
  condition and instead just slices the string under their respective
  object's property

Change-Id: I04b01d8fb8b3ecb242fa684019a293cb5aaf51b8

### DIFF
--- a/loleaflet/src/control/Control.NotebookbarBuilder.js
+++ b/loleaflet/src/control/Control.NotebookbarBuilder.js
@@ -667,6 +667,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 	},
 
 	_saveAsControl: function(parentContainer, data, builder) {
+		data.text = data.text.slice(0, -3);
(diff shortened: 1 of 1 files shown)


loleaflet/src/control/Control.NotebookbarBuilder.js
```diff
@@ -667,6 +667,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 	},
 
 	_saveAsControl: function(parentContainer, data, builder) {
+		data.text = data.text.slice(0, -3);
```
eszkadev, line 670: better: `data.text = data.text.replace('...', '');`

PR #1768, opened 2021-03-18
fix: undo + sheet switch breaks cursors

dennisfrancis:
Fix description:
Doing an undo causes core to do switch sheets if last change was in some
other sheet. But core sends status message with changed sheet number, so
ask for sheetGeometryData for this selected sheet if the sheetGeometry
we have is not for that sheet.

Change-Id: I6870bed86e0ffa234d659b8853a9ef9a33e02bc5

* Target version: co-6-4

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CalcTileLayer.js
+++ b/loleaflet/src/layer/tile/CalcTileLayer.js
@@ -520,6 +520,10 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this._docType = command.type;
 			this._parts = command.parts;
 			this._selectedPart = command.selectedPart;
+			if (this.sheetGeometry && this._selectedPart != this.sheetGeometry.getPart()) {
+				// Core initiated sheet switch, need to get full sheetGeometry data for the selected sheet.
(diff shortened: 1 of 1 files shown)


kendy:
Great - thanks a lot :-)

kendy:
After the fact, I recalled one thing; could this code actually lead to situation where users editing in other parts would change geometry, which in turn would then trigger (unwanted) updates for the given user?  If yes, is there a possibility to fix the undo case directly in the core instead, so that the update is triggered for the right part from the start?

dennisfrancis:
@kendy I believe the status messages of view A will only contain the part number of the view A and in this patch we check if that is same as the part of sheet-geometry data we have, so if other users in views B, C, D etc. edit other parts it will not change the part of view A hence cannot the trigger the new "request sheet geometry" call. I did a manual test to assert that too.

But if other user in view B changes the geometry of the part the view A is currently in, then view A will be notified but this is not done via new code in this patch. In this case the core initiates the sheetGeometry invalidation.

PR #1776, opened 2021-03-18
canvas: avoid flicker when zooming

lpranam:
Change-Id: I5e3335daf9b573384d2305442dfd4332bc50d15e

### Summary
flicker was caused because when we zoom we cleared the entire canvas at once
this patch would make sure instead of clearing the entire canvas,
we clear the area for a tile as we receive a tile

### Checklist


gokaysatir (approved):
Thanks for the fix :)

kendy:
Thank you for this! - though let's not merge it yet, I have a video showing some problems, not sure if it is specific to this patch or not, but needs further investigation I fear...

PR #1777, opened 2021-03-18
canvas: fix sizing on tablets

eszkadev:
Fixes issues with incorrect position and size of the canvas on tablets.
Fixes some errors (using null reference) and initializes properly the view
in readonly mode on tablets so it shows rows and columns headers


gokaysatir (commented):
Hello,

Thanks for the fix:)

When i un-group rows, there occurs a space on the right of the view.

eszkadev:
> When i un-group rows, there occurs a space on the right of the view.
> 
fixed

gokaysatir (approved):
I tested it. Thank you for the fix:)

PR #1782, opened 2021-03-19
view cursor/selection fixes

dennisfrancis:
* Target version: co-6-4

### Summary
fixes for view cursor/selection issues :
* cell selector of another user not always shown
* crash when user with selection leaves
* don't crash if view message don't exist

### Checklist


gokaysatir (approved):
I tested and problems seem gone. Thank you for the fixes.

PR #1789, opened 2021-03-19
Update chart overlay coordinates on activation

eszkadev:
The new message contains coordinates so apply them
before creating the grey overlay.

Change-Id: I3593ce6a9ae9a1bb95962ecf9127f5055f3e8beb


gokaysatir (approved):
Thanks for the fix:)

PR #1799, opened 2021-03-19
FreeBSD/libc++ fixes

arrowd:
* Target version: master 

With these changes I'm finally able to run `make && make check` and it finished with exit code 0!

### DIFF
--- a/test/WhiteBoxTests.cpp
+++ b/test/WhiteBoxTests.cpp
@@ -910,11 +910,11 @@ void WhiteBoxTests::testTime()
 {
     std::ostringstream oss;
 
-    std::chrono::system_clock::time_point t(std::chrono::nanoseconds(1567444337874777375));
+    std::chrono::system_clock::time_point t(std::chrono::duration_cast<std::chrono::system_clock::duration>(std::chrono::nanoseconds(1567444337874777375)));
     LOK_ASSERT_EQUAL(std::string("2019-09-02T17:12:17.874777Z"),
                          Util::getIso8601FracformatTime(t));
 
-    t = std::chrono::system_clock::time_point(std::chrono::nanoseconds(0));
+    t = std::chrono::system_clock::time_point(std::chrono::system_clock::duration::zero());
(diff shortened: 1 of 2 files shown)


arrowd:
Bump.

test/WhiteBoxTests.cpp
```diff
@@ -945,7 +951,7 @@ void WhiteBoxTests::testTime()
     t = std::chrono::system_clock::time_point();
     LOK_ASSERT_EQUAL(std::string("Thu, 01 Jan 1970 00:00:00"), Util::getHttpTime(t));
 
-    t = std::chrono::system_clock::time_point(std::chrono::nanoseconds(1569592993495336798));
+    t = std::chrono::system_clock::time_point(std::chrono::duration_cast<std::chrono::system_clock::duration>(std::chrono::nanoseconds(1569592993495336798)));
```
Ashod, line 954: I imagine there is a shortcut to this, but I haven't figured it out.

Btw, worth using clang-format here (and everywhere, really).

test/WhiteBoxTests.cpp
```diff
@@ -963,8 +969,16 @@ void WhiteBoxTests::testTime()
 
         const std::string s = Util::getIso8601FracformatTime(t);
         t = Util::iso8601ToTimestamp(s, "LastModifiedTime");
-        LOK_ASSERT_EQUAL(std::to_string(t_in_micros),
-                             std::to_string(t.time_since_epoch().count()));
+
+        std::string t_in_micros_str = std::to_string(t_in_micros);
```
Ashod, line 973: Would have been good to add examples of the values these strings might have in a comment, so it would be easier to follow what's going on. 

My guess is that on some systems there is no nano-second precision (microseconds only), but it's not clear why we are trimming the last 3 digits from *both* values. I assume it's because one has "000" at the end, and the other has some non-zero three digits. Again, I'm guessing and can't validate (because on my system I have nanosecond precision), hence my suggestion for a comment with explanation/example.

Ashod (approved):
@arrowd: Thanks Gleb! Good stuff. I left a couple of notes, but no blockers. Feel free to follow up as necessary.

PR #1805, opened 2021-03-22
fix: avoid view jump after pinch zoom animation

dennisfrancis:
Fix description:

The view jump is because after the zoom animation has ended, the map
center is set to the pinch center and not the last frame's center. The
patch computes the last frame's center and uses that to set the map
view's center. This only fixes the view jump and not the flicker after
the animation which is an independent issue.

Caveats:

Since we show only one discontinuous frame at the "final allowable zoom"
at the end of the zoom animation, there will be perceived view jump at
the end of the animation itself. This is apparent if the user pays
attention to the contents of the document at the corners of the view
before and after the is pinch stopped. This jump is more pronounced if
the pinch zoom factor is much different from the "final-zoom".
Note: The final zoom is determined from the final pinch zoom as
var finalZoom = this._map._limitZoom(zoomDelta > 0 ? Math.ceil(this._zoom) : Math.floor(this._zoom));

A possible solution for this problem is to render a few frames at the
end of the animation that are closer to the "final zoom" from the where
the pinch left off.

Change-Id: I0a6989248970aeb60acc451d9b0bc7a08a0dbf06

* Target version: co-6-4

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CanvasTileLayer.js
+++ b/loleaflet/src/layer/tile/CanvasTileLayer.js
@@ -349,7 +349,7 @@ L.TileSectionManager = L.Class.extend({
 		var splitPos = ctx.splitPos;
 		var canvasOverlay = this._layer._canvasOverlay;
 
-		var rafFunc = function () {
+		var rafFunc = function (timeStamp, final) {
 			painter._sectionContainer.setPenPosition(painter._tilesSection);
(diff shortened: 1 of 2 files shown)


kendy (approved):
@dennisfrancis: Real pleasure with this one - let's get it in please! :-)

PR #1808, opened 2021-03-22
Draw: NotebookBar: Fix Print control and Add download as PNG

pedropintosilva:
- Make it as bigtoolitem instead (to be consistent with the other apps)
- Make use of command

Change-Id: I801802d81ba558f3778ea31a95d183c28bc70302


loleaflet/src/control/Control.NotebookbarDraw.js
```diff
@@ -191,9 +191,9 @@ L.Control.NotebookbarDraw = L.Control.NotebookbarImpress.extend({
 						'children': [
 							{
 								'id': 'print',
-								'type': 'menubartoolitem',
+								'type': 'bigtoolitem',
 								'text': _UNO('.uno:Print', 'presentation'),
-								'command': ''
+								'command': '.uno:Print'
```
eszkadev, line 196: hi, we shouldn't execute Print uno command. we need to execute JS code which will trigger browser's print dialog. (we don;t want to print on a printer connected to the collabora server, but using client's printer :) )

eszkadev: nevermind, I see this command is overriden in NotebookbarBuilder :)

pedropintosilva: yes it is that's how I get rid of ellipsis :) thanks anyhow for checking it, I'm adding now one more commit to this PR

PR #1813, opened 2021-03-22
leaflet: set German bold shortcut to ctrl+shift+f

lpranam:
Change-Id: I97ff4d95250aad326b7d48635ab425045959ca5d

* Target version: master 

### Checklist

### DIFF
--- a/loleaflet/src/map/handler/Map.Keyboard.js
+++ b/loleaflet/src/map/handler/Map.Keyboard.js
@@ -475,7 +475,7 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


thebearon:
I'd consider limiting the Ctrl-F handling [here](https://github.com/CollaboraOnline/online/blob/1fd0f7302bf8a7071831572b02e03f79abac255c/loleaflet/src/map/handler/Map.Keyboard.js#L478) so it doesn't react if other modifiers (Shift, Alt) are also pressed. Then core can handle the key combination, and there's no need to complicate that further on the online side.

thebearon (approved):
Works for me, thanks!

PR #1823, opened 2021-03-24
removed repeated if block

ayhanyalcinsoy:
Signed-off-by: Ayhan Yalçınsoy <ayhanyalcinsoy@pisilinux.org>
Change-Id: Ia0bb4b83f0247b95f2a0a6efc1bf8e095460a180
Hİ @mmeeks , I think this if condition block has never execute and will never execute because of previous conditon block. Therefore I removed it by this patch. Would you like to check and merge it? 

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/js/global.js
+++ b/loleaflet/js/global.js
@@ -281,12 +281,9 @@
 					else if (b <= 0xdf) {
 						code = b & 0x1f;
 						seqLen = 2;
-					} else if (b <= 0xdf) {
-						code = b & 0x0f;
-						seqLen = 3;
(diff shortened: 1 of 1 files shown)


mmeeks:
Tor - IIRC you did some work here recently ? =) I guess we got some numbers wrong in the conditions ? - really looking forward to dropping IE11 in master soon ;-)

tml1024:
The code is broken, sure, but this doesn't fix it.

tml1024:
Closing this PR as the fix is wrong.

PR #1827, opened 2021-03-24
Calc: Prevent unnecessary jumping while editing a cell.

gokaysatir:
When autocomplete feature is on, and user starts to edit a cell, view
jumps to the beginning and it quickly jumps again. With the second jump,
edited cell is centered. At the end, view moves. Even though the user
can still see the cell subject to edit, this is uncomfortable.

Sometimes Core side fires events unordered. While this doesn't cause any
problems for it, it may cause some for Online side. Because Online side
processes data immediately. If we can have a stronger state structure on
Online side, we can process this unordered data safely. At least for
some situations like this one.

Change-Id: I84d129ffa2bd06f4fd55505605f514ac5ca57c51

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/TileLayer.js
+++ b/loleaflet/src/layer/tile/TileLayer.js
@@ -885,6 +885,12 @@ L.TileLayer = L.GridLayer.extend({
 			message = message.split(' ');
 			if (message.length > 1) {
 				this._map.context = {context: message[1]};
+
+				if (this._docType === 'spreadsheet') {
+					// We need to know the current context in Calc.
+					this.currentContext = message[1];
+				}
+
 				this._map.fire('contextchange', {context: message[1]});
 			}
 		}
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
If it works - great; I'd feel much more comfortable with getting the event emissions correct in the core really. I suspect the input-bar and its use of EditEngine of causing many of these problems.

kendy:
@gokaysatir: Thanks for this, but please let's not merge it (at least not yet). Originally the idea was to disable the autocomplete; which seems that the original code was not doing well - the feature that does this is actually "AutoInput" - accessible as Tools -> AutoInput toggle in the desktop version.

I'll attempt to disable it there & see if it fixes the issue.

mmeeks:
Marco re-enabled the autocomplete - and it seemed to mostly work (and is -really- useful ;-) I'd hate to disable it having shipped it. You saw my core patch around this trying to further reduce bogus events from the input area EditEngine ? - doesn't do the trick but ... ;-)

kendy:
@mmeeks: It is problematic in at least 3 ways; let's disable it for now please, see the details here:

https://gerrit.libreoffice.org/c/core/+/113047

kendy:
@gokaysatir: Sorry about this - but let's not use this patch...  Thanks for the work here though! :-)

PR #1829, opened 2021-03-24
Cmd+K on macOS should open the JS dialog, not the tunnelled dialog fr…

timar:
…om core

Change-Id: Iaa1b57076800c155c0fbfc5f1c386f9ed705c928

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
#1828 is better

PR #1835, opened 2021-03-24
Socket.js: Refresh the page after a server restart.

gokaysatir:
Change-Id: I81eec6d40368768ca1a53d57bfe0eb6136464b90

* Resolves: 
When server is restarted, some data between server and client may have inconsistencies. So we refresh the page and get fresh data from the server.

* Target version: master 

### TODO

- [ ] ...

### Checklist


dennisfrancis (commented):
The logic looks good to me. But I'd let Kendy comment as I'm not very familiar with this part of online.

PR #1836, opened 2021-03-24
Scrollbar: Fix flickering.

gokaysatir:
Scroll bar animation was flickering after mouse leaves its area.

Change-Id: Ie4d96d98dee806e6b02c8c8a23dd4eaccabbf7b0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


dennisfrancis (approved):
Nice one! Tested - works well too. Noticed a related issue though - When I drag on scrollbar in desktop calc (Chrome) it flickers a lot. This seems a harder problem to solve since both tile section and scroll section are both changing and redrawn.

PR #1855, opened 2021-03-25
Remove incorrect assertion in iOS-only code

tml1024:
It caused a crash when opening some PDF files, for instance.

Change-Id: I85515e2e14ffac8928714d218cd2353df228ff4b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
This is an iOS-only one-liner, and I see that it compiles and works for me, so I don't see the point in waiting for builds for other platforms.

PR #1857, opened 2021-03-25
Client side grids and zoom animation improvements

dennisfrancis:
* Target version: co-6-4

### Summary
Client side grids and zoom animation improvements.
1. Prepare calc online grid/tile rendering for transparent tiles (without background and grids). Merged cell border clearing will be done by core. The feature of client side grid rendering is not yet enabled, it needs a core patch before that.
2. Zoom animation improvements: Avoid flicker, avoid swipe-pinch zoom overlap, fixes thicker overlay line width on zoom out.

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CanvasSectionContainer.ts
+++ b/loleaflet/src/layer/tile/CanvasSectionContainer.ts
@@ -144,7 +144,7 @@ class CanvasSectionObject {
 	onMultiTouchEnd: Function; // Parameters: e (native event object)
 	onResize: Function; // Parameters: null (Section's size is up to date when this callback is called.)
 	onDraw: Function; // Parameters: null || (frameCount, elapsedTime)
-	onDrawArea: Function; // Optional Parameters: (area, canvasContext) - area is the area to be painted using canvasContext.
+	onDrawArea: Function; // Optional Parameters: (area, paneTopLeft, canvasContext) - area is the area to be painted using canvasContext.
 	onAnimationEnded: Function; // frameCount, elapsedTime. Sections that will use animation, have to have this function defined.
 	onNewDocumentTopLeft: Function; // Parameters: Size [x, y]
 	onRemove: Function; // This Function is called right before section is removed.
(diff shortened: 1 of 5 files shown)


gokaysatir (commented):
Thanks for the commits.

While i test it on Writer. It is freezing and tiles are not rendered after a pinch zoom on mobile device. Desktop works fine. Can you double check it?

Also, instead of adding a section specific function into CanvasSectionObject and/or Container, we could directly call that function with an appropriate pointer. Those classes are generic and if we can keep them like that (without specific cases), it will be easier to maintain. We can try to add new features in a way that every section can benefit from it. We can refactor this later if you wish. It's not urgent of course.

dennisfrancis:
@gokaysatir On testing I found a situation when the animation frame of swipe scroll goes on forever hence it will not allow pinch zoom to start, so in such cases pinch zoom was not working. I think I have fixed the swipe action with the last commit. Can you test again ?

Regarding your canvas section container comment, I think drawing a specific area within a section is pretty generic. It will be needed for all sections sooner or later especially for performant and flicker free rendering. I would also like that to be done for tiles/overlay section in future. Right now it just feels like wrong to me to implement such selective painting as part of section manager. Ideally all sections must have their own module and hence avoid that confusion. (I had to write onDrawArea's implementer function in section manager as I wanted to reuse its onDraw's implementation.)

Thanks.

gokaysatir (commented):
Hello,

Thank you for the fixes.

I tested it several times with and without your patch, in order to be sure that i am not pointing to a wrong thing. I think we have a problem there.

In Writer:
* I open a document on mobile device.
* Zoom out several times.
* View is gone.

Without your patch:
* Open a document (Writer).
* Zoom out several times.
* View is not very good, it should be improved even without your patch, but it can turn back to normal with a tap.

Other than this, i use animation without partial drawing and it is fast. i can't agree with your ideas about partial drawing. Since it's not an issue, we can wait for the time, it will let us see & understand better.

dennisfrancis:
> In Writer:
> 
> * I open a document on mobile device.
> * Zoom out several times.
> * View is gone.
> 
> Without your patch:
> 
> * Open a document (Writer).
> * Zoom out several times.
> * View is not very good, it should be improved even without your patch, but it can turn back to normal with a tap.
> 

It would have been helpful if these reproduction steps were present in the previous review comment.
The problem is that writer can have views with negative coordinates, so the setInterval loop only needs to wait for tiles with valid coordinates, else it will never end. This is fixed in the last commit

gokaysatir (approved):
Looks better, thanks for the fixes:)

PR #1859, opened 2021-03-25
Dump versions when we have a signal.

mmeeks:
Change-Id: I6b1f5ac1cebdd05dd9ff9541a5418059f55841ed

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -225,6 +225,7 @@ namespace SigUtil
         sigaction(SIGHUP, &action, nullptr);
     }
 
+    static char *VersionInfo = nullptr;
     static char FatalGdbString[256] = { '\0' };
 
     static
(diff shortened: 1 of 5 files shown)


mmeeks:
Would love a review - hopefully this helps with backtraces from unknown versions and lets us unwind symbols properly =)

common/SigUtil.cpp
```diff
@@ -283,10 +289,19 @@ namespace SigUtil
         }
     }
 
-    void setFatalSignals()
+    void setVersionInfo(const std::string &versionInfo)
+    {
+        if (VersionInfo)
+            free (VersionInfo);
+        VersionInfo = strdup(versionInfo.c_str());
```
Ashod, line 296: I wonder if ASAN or some similar tool will complain about the loss of this allocation. I suppose we can whitelist it.

Ashod (approved):
Very useful stuff. Thanks!

common/SigUtil.cpp
mmeeks: Possibly; but we can ignore that I think. Ultimately we could prolly use a std::string, and probably c_str() is signal safe - but, better to leak than to hide undefined behaviour in the code I think.

PR #1868, opened 2021-03-26
leaflet: set cursor to text if blinking cursor visible

lpranam:
Change-Id: Ie27783cab0edfa3e264df41d584d81379e0afe79

* Target version: master 

### Summary
problem:
when hovered over the scroll bar and then moving away from scrollbar,
mouse cursor would not be set to text even if cursor visible

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/ScrollSection.ts
+++ b/loleaflet/src/layer/tile/ScrollSection.ts
@@ -383,6 +383,11 @@ class ScrollSection {
 			this.sectionProperties.mouseIsOnVerticalScrollBar = false;
 			this.sectionProperties.mapPane.style.cursor = this.sectionProperties.defaultCursorStyle;
 			this.fadeOutVerticalScrollBar();
+
+			// just in case if we have blinking cursor visible
+			// we need to change cursor from default style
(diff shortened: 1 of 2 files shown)


loleaflet/src/layer/tile/ScrollSection.ts
```diff
@@ -383,6 +383,11 @@ class ScrollSection {
 			this.sectionProperties.mouseIsOnVerticalScrollBar = false;
 			this.sectionProperties.mapPane.style.cursor = this.sectionProperties.defaultCursorStyle;
 			this.fadeOutVerticalScrollBar();
+
+			// just in case if we have blinking cursor visible
+			// we need to change cursor from default style
+			if (this.map._docLayer._cursorMarker)
+					this.map._docLayer._cursorMarker.setMouseCursor();
```
thebearon, line 390: A bit of nitpicking, here and in line 416 there is an extra tab.

Apart from these, I can see the correct pointer, thanks!

gokaysatir (approved):
Thank you for the fix:) I also tested it, it works:)

PR #1870, opened 2021-03-26
admin: add support for lost kits cleanup

gmasei11:
Change-Id: I31969196aa138b6da964c867679ec23f34c89fd0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -274,17 +274,17 @@ bool runClientTests(bool standalone, bool verbose)
 // Standalone tests don't really use WSD
 #ifndef STANDALONE_CPPUNIT
 
-std::vector<int> getKitPids()
+std::set<pid_t> getKitPids()
 {
     return LOOLWSD::getKitPids();
 }
 
 /// Get the PID of the forkit
-std::vector<int> getForKitPids()
+std::set<pid_t> getForKitPids()
 {
-    std::vector<int> pids;
+    std::set<pid_t> pids;
     if (LOOLWSD::ForKitProcId >= 0)
-        pids.push_back(LOOLWSD::ForKitProcId);
+        pids.emplace(LOOLWSD::ForKitProcId);
     return pids;
 }
 
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -491,6 +491,8 @@ void Admin::pollingThread()
             if (cleanupWait <= MinStatsIntervalMs / 2) // Close enough
             {
                 cleanupResourceConsumingDocs();
+                if (_defDocProcSettings.getCleanupSettings().getLostKitGracePeriod())
+                    cleanupLostKits();
 
                 cleanupWait += _cleanupIntervalMs;
                 lastCleanup = now;
@@ -714,6 +716,11 @@ void Admin::addSegFaultCount(unsigned segFaultCount)
     addCallback([=]{ _model.addSegFaultCount(segFaultCount); });
 }
 
+void Admin::addLostKitsTerminated(unsigned lostKitsTerminated)
+{
+    addCallback([=]{ _model.addLostKitsTerminated(lostKitsTerminated); });
+}
+
 void Admin::notifyForkit()
 {
     std::ostringstream oss;
@@ -781,6 +788,57 @@ void Admin::notifyDocsMemDirtyChanged()
 void Admin::cleanupResourceConsumingDocs()
 {
     _model.cleanupResourceConsumingDocs();
+}
+
+void Admin::cleanupLostKits()
+{
+    static std::map<pid_t, std::time_t> mapKitsLost;
+    std::set<pid_t> internalKitPids;
+    std::vector<int> kitPids;
+    int pid;
+    unsigned lostKitsTerminated = 0;
+    size_t gracePeriod = _defDocProcSettings.getCleanupSettings().getLostKitGracePeriod();
+
+    internalKitPids = LOOLWSD::getKitPids();
+    AdminModel::getKitPidsFromSystem(&kitPids);
+
+    for (auto itProc = kitPids.begin(); itProc != kitPids.end(); itProc ++)
+    {
+        pid = *itProc;
+        if (internalKitPids.find(pid) == internalKitPids.end())
+        {
+            // Check if this is our kit process (forked from our ForKit process)
+            if (Util::getStatFromPid(pid, 3) == (size_t)_forKitPid)
+                mapKitsLost.insert(std::pair<pid_t, std::time_t>(pid, std::time(nullptr)));
+        }
+        else
+            mapKitsLost.erase(pid);
+    }
+
+    for (auto itLost = mapKitsLost.begin(); itLost != mapKitsLost.end();)
+    {
+        if (std::time(nullptr) - itLost->second > (time_t)gracePeriod)
+        {
+            pid = itLost->first;
+            if (::kill(pid, 0) == 0)
+            {
+                if (::kill(pid, SIGKILL) == -1)
+                    LOG_ERR("Detected lost kit [" << pid << "]. Failed to send SIGKILL.");
+                else
+                {
+                    lostKitsTerminated ++;
+                    LOG_ERR("Detected lost kit [" << pid << "]. Sent SIGKILL for termination.");
+                }
+            }
+
+            itLost = mapKitsLost.erase(itLost);
+        }
+        else
+            itLost ++;
+    }
+
+    if (lostKitsTerminated)
+        Admin::instance().addLostKitsTerminated(lostKitsTerminated);
 }
 
 void Admin::dumpState(std::ostream& os)
--- a/wsd/AdminModel.cpp
+++ b/wsd/AdminModel.cpp
@@ -846,6 +846,11 @@ void AdminModel::addSegFaultCount(unsigned segFaultCount)
     _segFaultCount += segFaultCount;
 }
 
+void AdminModel::addLostKitsTerminated(unsigned lostKitsTerminated)
+{
+    _lostKitsTerminatedCount += lostKitsTerminated;
+}
+
 int filterNumberName(const struct dirent *dir)
 {
     return !fnmatch("[0-9]*", dir->d_name, 0);
@@ -862,9 +867,6 @@ int AdminModel::getPidsFromProcName(const std::regex& procNameRegEx, std::vector
 
     std::string comm;
     char line[256] = { 0 }; //Here we need only 16 bytes but for safety reasons we use file name max length
-
-    if (pids != NULL)
-        pids->clear();
 
     while (n--)
     {
@@ -893,6 +895,24 @@ int AdminModel::getPidsFromProcName(const std::regex& procNameRegEx, std::vector
     free(namelist);
 
     return pidCount;
+}
+
+int AdminModel::getAssignedKitPids(std::vector<int> *pids)
+{
+    return getPidsFromProcName(std::regex("kitbroker_.*"), pids);
+}
+
+int AdminModel::getUnassignedKitPids(std::vector<int> *pids)
+{
+    return getPidsFromProcName(std::regex("kit_spare_.*"), pids);
+}
+
+int AdminModel::getKitPidsFromSystem(std::vector<int> *pids)
+{
+    int count = getAssignedKitPids(pids);
+    count += getUnassignedKitPids(pids);
+
+    return count;
 }
 
 class AggregateStats
@@ -1026,8 +1046,8 @@ void AdminModel::CalcDocAggregateStats(DocumentAggregateStats& stats)
 void CalcKitStats(KitProcStats& stats)
 {
     std::vector<int> childProcs;
-    stats.unassignedCount = AdminModel::getPidsFromProcName(std::regex("kit_spare_[0-9]*"), &childProcs);
-    stats.assignedCount = AdminModel::getPidsFromProcName(std::regex("kit_[0-9]*"), &childProcs);
+    stats.unassignedCount = AdminModel::getUnassignedKitPids(&childProcs);
+    stats.assignedCount = AdminModel::getAssignedKitPids(&childProcs);
     for (int& pid : childProcs)
     {
         stats.UpdateAggregateStats(pid);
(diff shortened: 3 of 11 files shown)


wsd/LOOLWSD.cpp
```diff
@@ -4254,21 +4254,25 @@ int LOOLWSD::getClientPortNumber()
...
-    std::vector<int> pids;
+    std::map<pid_t, pid_t> pids;
+    pid_t pid;
     {
         std::unique_lock<std::mutex> lock(NewChildrenMutex);
         for (const auto &child : NewChildren)
-            pids.push_back(child->getPid());
+        {
+            pid = child->getPid();
+            pids.emplace(pid, pid);
```
Ashod, line 4266: std::set ? 

std::map is for mapping, but since both key and value are the same, std::set is more fitting.

Ashod (commented):
Thanks Gabriel for the patch!

I'm assuming this was prompted by the issue we saw with documents failing to load (due to macros). Because otherwise I don't know of any case where we had lost kit processes. We have a fairly good hold on the lifetime of the kits and we *do* kill them at the end of their lifetime (see class ChildProcess for details). So the likelihood of having lost kits is fairly small (again, the issue with the macros is a special case, not common).

Anyway, would be good to hear the use-case. I expect you have some interesting backstory.

Thanks!

wsd/LOOLWSD.cpp
gmasei11: > std::map is for mapping, but since both key and value are the same, std::set is more fitting.

You're right. I'll replace map wit set.

gmasei11:
Thanks for your review!

> I'm assuming this was prompted by the issue we saw with documents failing to load (due to macros). Because otherwise I don't know of any case where we had lost kit processes. We have a fairly good hold on the lifetime of the kits and we _do_ kill them at the end of their lifetime (see class ChildProcess for details). So the likelihood of having lost kits is fairly small (again, the issue with the macros is a special case, not common).
> 
> Anyway, would be good to hear the use-case. I expect you have some interesting backstory.

It was prompted by the fact that there are cases when there are kits that are not managed by loolwsd and they still run and consume resources. One of these cases is the one that you mentioned with failing to load. I can't give you right now other specific cases because some of them are very rare and is very difficult to catch and study them in live environment, but the idea is as follows: 
As I already explained at the COOL meeting we need that mechanism because of practical reasons. As you know, we deploy COOL in a Kubernetes environment. A single lost kit process which heavily consumes resources could jeopardize an entire pod in a matter of minutes which means that the service will be very slow or even could be denied to thousands of requests. We experienced that already and this mechanism was crucial to solve the problem in a reasonable way. In Kubernetes we have the option to restart the pod but that will mean to destroy the existing sessions. Obvious, the ideal way of solving it is to solve the underlying issues but this is not something that can be done easily and fast and can't guarantee for future issues.

In conclusion: this is a safe mechanism whenever an issue of the kind explained above arises. From our perspective we already have it and I don't think we will remove it: we need it even if the cases are very rare because a single case could create big problems. However, if you consider that it is not useful for you then we can discard this pull request. I wanted to share it with you because I was thinking that maybe it will be useful for you in the same way it is useful for us.

Thanks!

mmeeks:
Code looks quite nice - then again - I'd love to:
-  get the re-factor and switch to std:set patch in first by itself - it seems to confuse the work here quite a bit.
-  ensure that we only kill children ultimately inherited from our process / forkit - we run two loolwsd's in the same container for testing locally.
-  log vigorously with some nasty ERR in the logs when we have to kill one of these: it is always a bug
-  have a setting to turn this off (code is quite nice so can default to 'on') in loolwsd.xml if we can (?)
Thanks Gabriel !

gmasei11:
Thanks for your review!

>     * have a setting to turn this off (code is quite nice so can default to 'on') in loolwsd.xml if we can (?)

How do you think is the best way to do this ? Some suggestions below:
1. Add an entry inside current per_document.cleanup section that enables/disables it (and maybe a setting for grace period which is now hard coded to 120 seconds).
2. Create two separate subsections of per_document.cleanup section: one for resource consuming docs (e.g. resource_consuming_docs) that will contain everything is now inside per_document.cleanup and one for lost kits cleanup (lost_kits_cleanup). This will mean probably a separate handling inside Admin::pollingThread with separate time intervals.
3. Add it in a totally different place inside loolwsd.xml. This will also mean a separate handling in side Admin::pollingThread.

mmeeks:
per_document cleanup sounds a perfect section - and we can have a special value for the grace period I suppose that turns the feature off (?) =) up to you really. Would love to get the std::set in first though =)

gmasei11:
> per_document cleanup sounds a perfect section - and we can have a special value for the grace period I suppose that turns the feature off (?) =) up to you really. Would love to get the std::set in first though =)

Sounds good!

Related to std::set I'm not sure what you mean. The replace of std::map with std::set was already made in the second commit.

mmeeks:
> Related to std::set I'm not sure what you mean. The replace of std::map with std::set was already made in the second commit.

Sure - but this seems then to involve touching lots of the same code places in both commits - looked like it could be avoided if we switched to std::set in a first pure re-factoring commit, and then put the feature on the top (instead of the reverse =)

gmasei11:
> Sure - but this seems then to involve touching lots of the same code places in both commits - looked like it could be avoided if we switched to std::set in a first pure re-factoring commit, and then put the feature on the top (instead of the reverse =)

I understand. Wouldn't be better to squash all commits in a single one ? I was planning to do it anyway before merge.

mmeeks:
squash would be perfect =)

test/test.cpp
```diff
@@ -274,17 +274,17 @@ bool runClientTests(bool standalone, bool verbose)
...
 
-std::vector<int> getKitPids()
+std::set<pid_t> getKitPids()
 {
     return LOOLWSD::getKitPids();
 }
 
 /// Get the PID of the forkit
-std::vector<int> getForKitPids()
+std::set<pid_t> getForKitPids()
```
Ashod, line 283: Interesting. We can ever have a single ForKit instance, by design. Do we really need to create and return a set here? (Not sure why we had a vector in the first place.)

Looks like a nice cleanup to do, I think, to simply return a single PID. Perhaps can be done separately, and not in this PR. Up to you.

wsd/Admin.cpp
```diff
@@ -783,6 +790,57 @@ void Admin::cleanupResourceConsumingDocs()
...
+void Admin::cleanupLostKits()
+{
+    static std::map<pid_t, std::time_t> mapKitsLost;
+    std::set<pid_t> internalKitPids;
+    std::vector<int> kitPids;
+    int pid;
+    unsigned lostKitsTerminated = 0;
+    size_t gracePeriod = _defDocProcSettings.getCleanupSettings().getLostKitGracePeriod();
+
+    internalKitPids = LOOLWSD::getKitPids();
```
Ashod, line 802: Nitpick: it's more readable to declare variables at their first use, if possible. Also, allows for making them `const` where possible. Fewer lines of code and more information on each line = win-win.

Notice also that in the case of heap objects (like vector and set) the performance is much better with a single `const std::set<pid_t> internalKitPids = LOOLWSD::getKitPids();` than by having a declaration and then assignment.  And since there is no gain of having them on two lines, the performance win is an added bonus!

wsd/Admin.cpp
```diff
@@ -783,6 +790,57 @@ void Admin::cleanupResourceConsumingDocs()
...
+    AdminModel::getKitPidsFromSystem(&kitPids);
+
+    for (auto itProc = kitPids.begin(); itProc != kitPids.end(); itProc ++)
```
Ashod, line 805: `for (const auto pid : kitPids)` 

No need to declare pid separately (but it's needed for the loop below, which isn't a bad thing, since sharing variables across loops without there being a dependency between them is not a good idea).

wsd/AdminModel.cpp
```diff
@@ -895,6 +897,24 @@ int AdminModel::getPidsFromProcName(const std::regex& procNameRegEx, std::vector
     return pidCount;
 }
 
+int AdminModel::getAssignedKitPids(std::vector<int> *pids)
+{
+    return getPidsFromProcName(std::regex("kitbroker_.*"), pids);
```
Ashod, line 902: The string `kitbroker` and `kit_spare` should come from some common place, with a named constant. Common.hpp is a reasonable place to have it there. Perhaps we can do that later.

wsd/AdminModel.cpp
```diff
@@ -895,6 +897,24 @@ int AdminModel::getPidsFromProcName(const std::regex& procNameRegEx, std::vector
...
+{
+    return getPidsFromProcName(std::regex("kit_spare_.*"), pids);
+}
+
+int AdminModel::getKitPidsFromSystem(std::vector<int> *pids)
+{
+    int count = getAssignedKitPids(pids);
+    count += getUnassignedKitPids(pids);
+
+    return count;
```
Ashod, line 915: `return getAssignedKitPids(pids) + getUnassignedKitPids(pids);` ?

wsd/AdminModel.cpp
```diff
@@ -895,6 +897,24 @@ int AdminModel::getPidsFromProcName(const std::regex& procNameRegEx, std::vector
...
+{
+    return getPidsFromProcName(std::regex("kitbroker_.*"), pids);
+}
+
+int AdminModel::getUnassignedKitPids(std::vector<int> *pids)
+{
+    return getPidsFromProcName(std::regex("kit_spare_.*"), pids);
+}
+
+int AdminModel::getKitPidsFromSystem(std::vector<int> *pids)
```
Ashod, line 910: Any good reason to have a pointer here? Also, check for nullptr?

wsd/LOOLWSD.cpp
```diff
@@ -4254,21 +4256,25 @@ int LOOLWSD::getClientPortNumber()
...
-    std::vector<int> pids;
+    std::set<pid_t> pids;
+    pid_t pid;
     {
         std::unique_lock<std::mutex> lock(NewChildrenMutex);
         for (const auto &child : NewChildren)
-            pids.push_back(child->getPid());
+        {
+            pid = child->getPid();
+            pids.emplace(pid);
```
Ashod, line 4268: Check that `pid > 0` before `emplace`?

wsd/AdminModel.hpp
```diff
@@ -384,6 +388,9 @@ class AdminModel
     void setDefDocProcSettings(const DocProcSettings& docProcSettings) { _defDocProcSettings = docProcSettings; }
 
     static int getPidsFromProcName(const std::regex& procNameRegEx, std::vector<int> *pids);
+    static int getAssignedKitPids(std::vector<int> *pids);
+    static int getUnassignedKitPids(std::vector<int> *pids);
+    static int getKitPidsFromSystem(std::vector<int> *pids);
```
Ashod, line 393: Maybe for a follow up patch, but this pointer here is not necessary at all. Especially that getPidsFromProcName makes no sense to call with a nullptr (it will just waste some CPU cycles).

Best to take reference: `std::vector<int>& pids`. No pointers, no null, no issues.

Ashod (approved):
@gmasei11 Thanks Gabriel for the explanation and patch + improvements! It's really looking good.

I left some feedback and notes. I don't see any blockers, but would be good to have some of my suggestions as a follow up, if you have the bandwidth. If not, just let me know and we can have some of them as easy-hacks that others may pick up.

I'll let @mmeeks have a look before I merge as I think he didn't get to see the last update you did a few hours ago.

Thanks again!

mmeeks:
Looks nice to me - lets get it in - some of Ash's suggestions sound great if you're interested =) One thought - our forkit can crash and re-start (which is far from ideal) - and then we struggle onwards =), not sure if we cope with the cleanup in this circumstance - so quite possibly we want to check for the last PID of forkit as well as the current one as a parent (no idea - it's a nasty corner case - but a useful one for availability).

gmasei11:
> Looks nice to me - lets get it in - some of Ash's suggestions sound great if you're interested =) One thought - our forkit can crash and re-start (which is far from ideal) - and then we struggle onwards =), not sure if we cope with the cleanup in this circumstance - so quite possibly we want to check for the last PID of forkit as well as the current one as a parent (no idea - it's a nasty corner case - but a useful one for availability).

I was aware of this scenario when forkit can crash and was thinking to introduce a std::set with all forkit pids that existed during the lifetime of a loolwsd and use it in the sense you mentioned. However, we know that the OS reuses process pids and there is a very small chance that a pid of a once valid forkit associated with loolwsd instance 1 could become pid of a new valid forkit associated with loolwsd instance 2. In this case all kit processes spawned by forkit of loolwsd instance 2 will be killed by loolwsd instance 1 which means that the  loolwsd instance 2 will become practically unusable. I've chose to avoid this scenario because of its consequences.

I'm aware that this scenario has very small chances to become real. On the other hand the scenario that you mentioned also has small chances to become real: a forkit crash followed by one of the kits forked from it and still alive becoming lost.

A possible solution is to check that the kit process is a child of our current forkit process or it is an orphan process. I don't know a safe way to check that a process is orphan. It seems that is not enough to check that its parent pid is 1. But I'm not an expert on this. What you think ? Is this something easy to implement ?

test/test.cpp
gmasei11: I've already seen this but decided not to change it. It returns a set (previously vector) probably because its return value is passed to killPids: https://github.com/CollaboraOnline/online/blob/de62c0f359ecb589d8480889092c0b6592fe56ed/test/httpcrashtest.cpp#L252

wsd/Admin.cpp
gmasei11: Agree.

wsd/Admin.cpp
gmasei11: You can see that 'pid' variable is used later in the second loop: https://github.com/CollaboraOnline/online/blob/de62c0f359ecb589d8480889092c0b6592fe56ed/wsd/Admin.cpp#L822
 I don't think it's nice to declare twice in the same function a variable with the same purpose.

wsd/AdminModel.cpp
gmasei11: Agree.

wsd/AdminModel.cpp
gmasei11: Well, I prefer to do like this just for debug purpose. Just to be able to put a breakpoint at second call.

wsd/AdminModel.hpp
gmasei11: It makes sense to call getPidsFromProcName with nullptr. The function returns the number of found processes. It is used here: https://github.com/CollaboraOnline/online/blob/de62c0f359ecb589d8480889092c0b6592fe56ed/wsd/AdminModel.cpp#L1070
and here: https://github.com/CollaboraOnline/online/blob/de62c0f359ecb589d8480889092c0b6592fe56ed/wsd/AdminModel.cpp#L1076

As a consequence, all dependent function will have the same type for this parameter.

wsd/LOOLWSD.cpp
gmasei11: Agree, for safety reasons.

PR #1887, opened 2021-03-30
Fix background color in which document page is placed

pedropintosilva:
Avoid pure white and instead use another color for canvas background

Change-Id: I4cc9dffd19932bc631f5b880e53f8d413357d662


kendy:
@pedropintosilva: Thanks for the patch! Would be really good to have this configurable via css_variables, like https://sdk.collaboraonline.com/docs/theming.html?highlight=ui_defaults#content-of-hidden-field-css-variables before we integrate this...

gokaysatir (approved):
Thank you for the patch :) I tested it, i had to set bg to black to see it is working :) Anyway, if it is ok for @kendy, then we can merge this one i think.

PR #1893, opened 2021-04-01
Fix UTF-8 decoding and fix handling of non-BMP codepoints for IE11

tml1024:
The UTF-8 decoder here was buggy, and additionally it tried to feed
the Unicode codepoints it had constructed directly to
String.fromCharCode() even if that function takes UTF-16 units, not
Unicode codepoints.

The function in question is used only on IE11, apparently, so
presumably the problems had not been noticed.

Change-Id: I78f8ecea6ce6349e4121066774ae5c3fcfa87363

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/js/global.js
+++ b/loleaflet/js/global.js
@@ -285,10 +285,10 @@ window.app.definitions = {};
 					else if (b <= 0xdf) {
 						code = b & 0x1f;
 						seqLen = 2;
-					} else if (b <= 0xdf) {
+					} else if (b <= 0xef) {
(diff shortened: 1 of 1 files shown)


tml1024:
This is a correct fix for the problem noticed in https://github.com/CollaboraOnline/online/pull/1823 . I did actually verify manually using the js command-line interpreter on Linux that the decoder function now works correctly, even for non-BMP code points.

Issue #1461, opened 2021-02-11
iOS App: Calc: Text "Background Color" is visible when selecting a color for coloring a cell


Gianni154:
**Describe the bug**
If you want to color a cell in a table document (calc), this works without problems. However, I noticed that when selecting the background color, the text "Background Color" is displayed in the foreground. Unfortunately, this covers some colors during the selection, such as the color black.

**To Reproduce**

1. Open the Collabora iOS App
2. Open a new or an existing spreadsheet (calc)
3. Go to Edit Mode and click on any cell
4. Select "Home" in the notebookbar
5. Click on the "Background Color" icon 

**Expected behavior**
The text "Background Color" is no longer visible when selecting colors. 

**Actual behavior**
When selecting the color, the text "Background Color" hides the selection of some colors. 

**Screenshots**

![File_000](https://user-images.githubusercontent.com/78140364/107613394-82a18280-6c48-11eb-932c-a31bf9731b6f.jpeg)

**Device**
 - Device: iPad 6th generation 
 - OS: iPad OS 14.4
 - Version: iOS Collabora App Version 6.4.5 (2)

timar:
@pedropintosilva or @eszkadev would it be possible to move the tooltip out of the way when someone touches those dropdowns? There are other examples. Backgorund color, Font color, Conditional...

tml1024:
No tooltips should ever be displayed in the iOS app, right? Presumably this will be easy to fix.

thebearon:
> No tooltips should ever be displayed in the iOS or Android apps, right? Presumably this will be easy to fix.

@kendy back then thought tooltips on mobile could be shown for a short amount of time (eg 5s), and then disappear, but I agree it's probably fine without tooltips, too.

On the other hand, now the behavior diverges between web-based online on tablet and the mobile app, which is less desired.

tml1024:
> now the behavior diverges between web-based online on tablet and the mobile app, which is less desired

See the commit message. We should probably just use the window.mode.isDesktop() condition for all tooltip creation?

tml1024:
> thought tooltips on mobile could be shown for a short amount of time (eg 5s)

But how would we know when to display a "tooltip" when there is no cursor that would be kept over a UI element while the user wants to figure out (or tries to remember) what it does (to which the tooltip helpfully gives the answer)?

If there was proximity sensing and we would know that the user is "hovering" their finger above a certain spot, but not touching, the situation would be different. But no touch device does that, I think.

[...]

PR #1909, opened 2021-04-04
leaflet: unified the tooltip bindings

lpranam:
Change-Id: Idc9bb0ac12b849871292293286f9019449cc9851

* Target version: master 

### Summary
additionally, tooltips in tablet will be visible on long press now
#1461

### Checklist

### DIFF
--- a/loleaflet/src/control/Control.FormulaBar.js
+++ b/loleaflet/src/control/Control.FormulaBar.js
@@ -42,8 +42,7 @@ L.Control.FormulaBar = L.Control.extend({
 				$('#addressInput').off('keyup', that.onAddressInput.bind(that)).on('keyup', that.onAddressInput.bind(that));
 			}
 		});
-		if (window.mode.isDesktop())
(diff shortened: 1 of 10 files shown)


pedropintosilva:
@lpranam I have tested and it seems to work quite good when using `notebookbar` but not so when using `classic` mode:

- Device: iPad mini
- Mode: Notebookbar: shows the tooltip while long pressing it
- Mode: Classic: Long pressing does not show tooltip; but upon release it shows tooltip plus acts as a click. Example:
  - Long pressing Align Centre will do nothing (it stays in the default Align Left)
  - Upon releasing the finger: it shows the tooltip plus changes the alignment to centre
  - This ^ also happens when working with drop downs (it opens them)

pedropintosilva:
Sorry to have waited so long, I think we need to merge this to master and have more people testing it, because it seems that I'm the only one getting that weird behaviour and just in one device

PR #1930, opened 2021-04-07
test helpers: fix -Werror,-Wreturn-type

vmiklos:
./test/helpers.hpp:211:1: error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]

Perhaps it could go further and even call LOK_ASSERT_FAIL().

Change-Id: I4f68f6d290bcb0a832ea71153d5f699d5366def9


vmiklos:
@Ashod I didn't add an assert here, but perhaps you are more brave and you want to :-)

PR #1939, opened 2021-04-07
leaflet: enable comments for PDF in mobile

lpranam:
Change-Id: Ie80f9a808d0234f6c58b768fa27673c9fe50b122

* Target version: master 

### Summary
in addition, allow users to work with comments in read only mode
wizard z index increased to avoid mobile-edit-button overlapping it

### Checklist

### DIFF
--- a/loleaflet/src/control/Permission.js
+++ b/loleaflet/src/control/Permission.js
@@ -146,7 +146,10 @@ L.Map.include({
 	},
 
 	isPermissionEditForComments: function() {
-		return true;
+		// Currently we allow user to perform comment operations
+		// even in the view/readonly mode(initial mobile mode)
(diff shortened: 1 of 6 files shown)


loleaflet/src/control/Permission.js
```diff
@@ -146,7 +146,7 @@ L.Map.include({
 	},
 
 	isPermissionEditForComments: function() {
-		return true;
```
kendy, line 149: Ideally it would be great to have a comment here to explain what is going on (that we allow commenting even when in the view mode on the mobile) - can you do please?

kendy (approved):
Small thing in the review; otherwise looks great to me, thank you!

PR #1940, opened 2021-04-08
fix: tiles not rendered on calls from setZoom()

dennisfrancis:
where we set "zoomChanged". So we should have another state variable
waitForTiles to wait for tiles instead of re-using zoomChanged.

Change-Id: I3bd71fb97694a56be5f49a9e7b7ac7b9c49caa88

* Resolves: # 
* Target version: co-6-4

### Checklist


kendy:
Sounds like a good idea to differentiate the state - let's do, thank you!

PR #1941, opened 2021-04-08
fix: tiles not rendered on calls from setZoom() (master)

dennisfrancis:
where we set "zoomChanged". So we should have another state variable
waitForTiles to wait for tiles instead of re-using zoomChanged.

Conflicts:
	loleaflet/src/layer/tile/CanvasTileLayer.js

Change-Id: I3bd71fb97694a56be5f49a9e7b7ac7b9c49caa88
(cherry picked from commit c9d6d9f569fcee3516a55d3f3a8719b7338687a4)

* Target version: master 

### Checklist


dennisfrancis:
Kendy reviewed the co-6-4 version of the patch. Merging...

Issue #1751, opened 2021-03-17
iOS App: Calc: It is not possible to delete a created chart (diagram) through the "long press menu"


Gianni154:
**Describe the bug**
In spreadsheet document it is not possible to delete a created chart through the "long press menu". It is only possible to cut it out.

**To Reproduce**

1. Open the Collabora iOS App
2. Open a new or an existing spreadsheet (calc)
3. Go to Edit Mode and select "Insert" in the notebookbar
4. Select "Charts Wizard" to create a chart
5. Press on the created diagram until the Long-Press Menu appears. No Delete option is displayed.

**Expected behavior**
It is possible to delete the created diagram. The Delete option is available in the long press menu.

**Actual behavior**
It is not possible to delete the created diagram. Delete option is not present in the long press menu. 

**Screenshots**

![File_000](https://user-images.githubusercontent.com/78140364/111472084-ec221e80-8729-11eb-9f40-7364481a98a9.jpeg)

**Device**
 - Device: iPad 6th generation 
 - OS: iPad OS 14.4
 - Version: iOS Collabora App Version 6.4.6 (1)

thebearon:
Technically, cutting the chart has the same effect (except it'd be added to the clipboard in that case, too), but indeed, a separate Delete option would be reasonable on a device where there's normally no Delete button to press.

Since normally the desktop menu is shown, and adding the item on desktop would be unadvised, I wonder if it could be added only for mobile devices. In fact, I can see phones already have a Delete item for charts. @eszkadev, can you share how that is handled, and if there'd be a simple way to have that on tablets as well?

PR #1943, opened 2021-04-08
Add "Delete" to the context menu for graphic selection in the mobile …

tml1024:
…apps, too

This fixes https://github.com/CollaboraOnline/online/issues/1751

Change-Id: I2f61c43b310a32f5b5d883610b169d2ab47d554d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/control/Control.ContextMenu.js
+++ b/loleaflet/src/control/Control.ContextMenu.js
@@ -169,8 +169,8 @@ L.Control.ContextMenu = L.Control.extend({
 	},
 
 	_amendContextMenuData: function(obj) {
-		// Add a 'delete' entry for mobile for graphic selection.
-		if (this._map._docLayer.hasGraphicSelection() && window.mode.isMobile()) {
+		// Add a 'delete' entry  for graphic selection when on mobile browser or in a mobile app.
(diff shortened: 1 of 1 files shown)


loleaflet/src/control/Control.ContextMenu.js
```diff
@@ -169,8 +169,8 @@ L.Control.ContextMenu = L.Control.extend({
 	},
 
 	_amendContextMenuData: function(obj) {
-		// Add a 'delete' entry for mobile for graphic selection.
-		if (this._map._docLayer.hasGraphicSelection() && window.mode.isMobile()) {
+		// Add a 'delete' entry  for graphic selection when on mobile browser or in a mobile app.
+		if (this._map._docLayer.hasGraphicSelection() && (window.ThisIsAMobileApp || window.mode.isMobile())) {
```
thebearon, line 173: Tor, is there a particular reason why this check has to be done against whether it's a mobile app, and not against whether it's a tablet? I presume this fix won't work in web-based online on tablets.

tml1024:
@thebearon No other reason than that I don't want to affect what the code does in other contexts except the exact one I am fixing;) After all, there might be a good reason why web-based online on tablets is supposed to be different here, it has a different UI than web-based online on phones.

PR #1947, opened 2021-04-08
leaflet: set mouse cursor on cursor show

lpranam:
Change-Id: I6adda739dc26fb93857bf8dcf70348b07f8eab33

* Target version: master 

### Summary
This is absolutely necessary for impress,
if cursor visible state is not changed,
we do not add or remove the cursor,
which handles the mouse cursor

so we explicitly set mouse cursor

i.e: in impress try to edit in one text box,
now click on the text on another text box
the mouse cursor will be incorrectly set

### Checklist


dennisfrancis:
Looks good. Might be a good idea to update PR title/summary as well. Please merge when CI passes.

Issue #1297, opened 2021-01-21
Select Table has strange gray elements


Andreas-Kainz:
see screenshot

![Screenshot_20210121_155525](https://user-images.githubusercontent.com/8517736/105367866-35aa2d80-5c01-11eb-80d7-bb7158d272d7.png)

Andreas-Kainz:
how should it look like?

pedropintosilva:
Very good catch @Andreas-Kainz ! thanks to you report, even though not directly related, I found out that those markers are not working they are supposed to be clickable. I already found where the regression came from and I will try either to revert back completely or partly with some adjustments. 

Meanwhile and answering to your question, these should definitely be improved, They were suppose to be a quick way to select a whole row or column and the same commit that added the regression also enlarged that svg to be more touch friendly. This is how it looked before:
![image](https://user-images.githubusercontent.com/65948705/105496580-f4c41e80-5cbd-11eb-84ff-8ca1b7244afe.png)

Either way that svg needs 1. improvement/redraw for sure. And on top of that needs more 2. status, at very least is missing hover and focus status as you see in the image. Do you want to take that task?

Andreas-Kainz:
I don't think that the gray elements are easy enough to understood.

Andreas-Kainz:
some ideas maybe you like one

![Screenshot_20210123_004622](https://user-images.githubusercontent.com/8517736/105560578-7181e780-5d14-11eb-9282-77521fed0a4a.png)

PR #1954, opened 2021-04-09
Writer: Improve table controls (markers)

pedropintosilva:
- Do not use big rectangles as markers
  - it hides surrounding content
	- too intrusive (since it appears every time we select a table)
- Use smaller discreeter elements WHILE maintaining the existing
  active zone size (so it's still easy to click it, no need to be precise)
- Add missing hover status
  - Create new SVG plus add rounded corners
- Treat column and row marker as different elements
  - Use one generic class + add another for each type (element modifier)
	- So we can target each one and have a nice center line
	no matter the disposition (vertical, horizontal)

note: motivated by the reported issue:
https://github.com/CollaboraOnline/online/issues/1297

Change-Id: I4d997864e99582f74411a12c33381a926fd47a05

### DIFF
--- a/loleaflet/css/loleaflet.css
+++ b/loleaflet/css/loleaflet.css
@@ -131,14 +131,38 @@
 	cursor: row-resize;
 }
 
-.table-row-or-column-select-marker {
+.table-select-marker {
 	margin: 0px;
 	width: 24px;
 	height: 24px;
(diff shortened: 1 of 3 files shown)


pedropintosilva:
https://user-images.githubusercontent.com/65948705/114168812-0cac5580-9931-11eb-8aa0-101492e72a10.mp4

PR #1960, opened 2021-04-10
fix typescript error due to build/ dir if any

dennisfrancis:
Change-Id: Ibed6facc18a83ae2d7d3ef59be06ddedd5142cad

* Target version: master 

### Summary
Since typescript project compilation is done just for the source directory loleaflet/src/ and tsconfig.json sits in loleaflet/ we need to to add other dirs that contain js files to tsconfig.json's exclude list.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/tsconfig.json
+++ b/loleaflet/tsconfig.json
@@ -26,6 +26,7 @@
       "typescript_js",
       "dist",
       "release",
-      "debug"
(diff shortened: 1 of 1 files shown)


dennisfrancis:
This is not needed anymore.

PR #1984, opened 2021-04-13
notebookbar: reorganize sheet and data tabs

eszkadev:
there was no space in the data tab with some languages

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -699,7 +699,7 @@ div[id*='Row'].notebookbar, div[id*='Column'].notebookbar, #SendToBack.notebookb
 /* Sheet Tab */
 
 #table-Sheet-Section.notebookbar {
-	margin-top: 10px;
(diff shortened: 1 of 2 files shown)


pedropintosilva:
Nice! As discussed I think this is a great improvement :+1: 

I was also testing with lengthy languages (DE, I think with Mongolian it's the one with longer words on average ) and on a 27'' screen there is not so much space left. Maybe we can still decrease the space occupied in Data tab by removing the labels from sorting options, what do you think?

image

eszkadev:
> I was also testing with lengthy languages (DE, I think with Mongolian it's the one with longer words on average ) and on a 27'' screen there is not so much space left. Maybe we can still decrease the space occupied in Data tab by removing the labels from sorting options, what do you think?

yes, good idea, done in last commit

pedropintosilva (commented):
:+1:

pedropintosilva (approved):

PR #1992, opened 2021-04-14
Calc: Filter with split panes improvement.

gokaysatir:
Change-Id: I866edb95f55d8b16a6ce3ba4e0cb4bbaa79e182d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Before this patch, when the Autofilter was activated, the splitter between panes went up. And when the Autofilter was de-activated, the splitter didn't return to the same position, but way lower.

PR #1994, opened 2021-04-14
Server restart follow up. Don't refresh the page if file name is diff…

gokaysatir:
…erent.

Change-Id: I4174dade3ff334687b485264ed8480ec3509cc4a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
@gokaysatir: Thank you for the patch!  It seem to me that the BaseFileName could be unreliable as a check - is the wopiSrc the same when when do the save-as?  I think not - can you please dobule check if it's a more reliable way how to check we are connecting to a different server after the save-as?

kendy:
Looks good to me, let's do it - thank you! :-)

PR #1997, opened 2021-04-14
loolkitconfig: introduce config to show Non-printing Characters

lpranam:
Change-Id: Ia28fdcf76a899a6844d367bc6fb81e289abc12f6

* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
@lpranam: Thanks for finding them!  Please do comment them out though, so that they don't become new default, we want to have them in only as an example what to do when the admins want to enable that in their instances.

kendy:
Actually I see you've put "false" there explicitly; that's fine too I guess...  But can you check why the tests have failed please?

PR #2014, opened 2021-04-15
Logging improvements

tml1024:
Related to changes in core to produce Chrome Trace Event format log files for performance investigation.

### DIFF
--- a/scripts/perftrace.pl
+++ b/scripts/perftrace.pl
@@ -96,7 +96,7 @@ my @event_pairs = (
     { # Save - to storage
       name => 'save to storage',
       emitter => '^docbroker',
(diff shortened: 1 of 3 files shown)


mmeeks:
Hmm - I'd prefer to have newlines in the logs - particularly the easy to copy/paste lines to setup your proof-key are really helpful. Can't we log profiling to our own / different profile / log-file in C++ avoiding the perl-ness ? =)

tml1024:
@mmeeks Sure; this change is not necessary if I continue to the next steps, where that Perl tooling isn't necessarily needed any more. Will close this now, this was just an intermediate step.

PR #2023, opened 2021-04-16
Fix tscompile.done dependencies, and remove wildcards for ts/js files.

mmeeks:
Get dependencies right for tscompile - to point to dist/ and not src/

Change-Id: Ida6cd1a505376099551821e86144410ff75eca6b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Abandon this - we have a much better solution in master.

Issue #2006, opened 2021-04-15
Mobile: toolbar-down: swipe indicator not visisble at the start


pedropintosilva:
**Describe the bug**
1. Load a document
2. Tap FAB Edit
3. Swipe indicator does not show
4. Swipe to the right
5. Indicator appears

https://user-images.githubusercontent.com/65948705/114837364-e88ac180-9dd3-11eb-8f9f-1e8280e2dff7.mp4

PR #2025, opened 2021-04-16
Fix: mobile: toolbar-down: swipe indicator now visible at the start

Rash419:
Change-Id: Ib28ce57172f9c93e49eab4eeca3abeaa783a1fd4

* Resolves: #2006 
* Target version: master 

image


pedropintosilva (approved):
Thanks! :1st_place_medal:

PR #2028, opened 2021-04-16
test: tolerate missing minor core version

vmiklos:
ProductVersion used to be e.g. "6.4", but co-2021 sets it to just
"2021".

Change-Id: I08c1204c42bba21c8036975138e3d999b7357bdf


vmiklos:
Michael says let's get this in, it can't make things worse. :-)

PR #2029, opened 2021-04-16
Restore JS / TS file ordering.

mmeeks:
We badly need to preserve the ordering of files in the WEBORDER,
but also to be able to treat them differently.

Allow both the m4 concatenation for bundlifying and also the
HTML script substitution to use the files in the right order.

Change-Id: If2a1b32548e1086b18395385fb512e9d39978406

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Obsoleted by subsequent pull request.

PR #2032, opened 2021-04-16
loleaflet: fix error messages when a document is not loaded

hcvcastro:
It happens when reconnecting, a Macro Warning dialog is opened,
but the document is not loaded yet.

Change-Id: If98cf4d68ac8ef71d7c0ad3d8aa5151a77adbe4b


eszkadev (commented):
good in general, but would be nice to not make conditions more complex

eszkadev (approved):

PR #2033, opened 2021-04-16
loleaflet: fix activate messages

hcvcastro:
The activate() function assumes that the document
is loaded when the socket is connected.
This is not true until a status message arrives,
then a document layer is created.
If the document has an interactive dialog (i.e. Macro Warning)
the document formally is not loaded and it should not send
messages that will receive "nodocloaded error".

Change-Id: I01ceb156376712234ab4c12226d080446aa9279d


eszkadev (approved):
I checked if after one user goes back from idle - will get updated tiles and everything seems to be ok. Thanks.

PR #2035, opened 2021-04-16
Avoid ERANGE poisoning numerical conversions.

mmeeks:
Behave more similarly to before 65c245eab0.

Change-Id: I0156310257caf7c578fb273393566a0970b3bb1f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Good catch! Thanks.

mmeeks:
Seems some of the tests run vs. the original push rather than updated ones which is annoying; eg. DCO - no idea why.

PR #2040, opened 2021-04-18
HTTP Backported to 6-4

Ashod:
All the HTTP and async saving relevant patches.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,7 @@ endif
 
 else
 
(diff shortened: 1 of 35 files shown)


Ashod:
@kendy, this includes everything, but I think we want to reduce this. I left out anything unrelated to http, unless I got conflicts and even then I only took what was relevant.

mmeeks:
Lots of reading / review and follow-on cleanups from Ash - thanks Ash ! =)

PR #2066, opened 2021-04-20
Dialogdots

Andreas-Kainz:
* Resolves: #1285 
* Target version: master 

### Summary
No ... any more at the Notebookbar labels
Screenshot_20210421_003932

As there is already an _cleanText function to remove the ~ I added the '...' for replacement and add _cleanText to the NB labels.

### DIFF
--- a/loleaflet/src/control/Control.NotebookbarBuilder.js
+++ b/loleaflet/src/control/Control.NotebookbarBuilder.js
@@ -415,7 +415,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 			button.id = buttonId;
 			button.setAttribute('alt', id);
 
-			if (data.id === 'FontColor' || data.id === 'BackColor' || data.id === 'BackgroundColor') {
+			if (data.id === 'FontColor' || data.id === 'BackColor' || data.id === 'BackgroundColor' || data.id === 'Color' || data.id === 'CharBackColor' || data.id === 'XLineColor' || data.id === 'FillColor') {
 				L.DomUtil.create('i', 'unoarrow', div);
 			}
 
--- a/loleaflet/src/control/Control.NotebookbarWriter.js
+++ b/loleaflet/src/control/Control.NotebookbarWriter.js
@@ -124,21 +124,6 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 								}
 							]
 						} : {},
-					hasPrint ?
-						{
-							'id': 'Section4',
-							'type': 'toolbox',
-							'text': '',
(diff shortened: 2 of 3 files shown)


loleaflet/src/control/Control.NotebookbarBuilder.js
```diff
@@ -415,7 +415,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 			button.id = buttonId;
 			button.setAttribute('alt', id);
 
-			if (data.id === 'FontColor' || data.id === 'BackColor' || data.id === 'BackgroundColor') {
+			if (data.id === 'FontColor' || data.id === 'BackColor' || data.id === 'BackgroundColor' || data.id === 'Color' || data.id === 'CharBackColor' || data.id === 'XLineColor' || data.id === 'FillColor') {
```
eszkadev, line 418: please remove this condition at all.
We want dropdown in all color pickers, no sense to explicitly add list of them here.

loleaflet/src/control/Control.NotebookbarWriter.js
```diff
@@ -124,33 +124,88 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
...
 							'enabled': 'true',
 							'children': [
 								{
-									'id': 'print',
+									'id': 'rev-history',
 									'type': 'bigtoolitem',
-									'text': _UNO('.uno:Print', 'text'),
-									'command': '.uno:Print'
+									'text': _('See history'),
+									'command': '.uno:rev-history'
```
eszkadev, line 138: shareas and rev-history are now duplicated: we have 2 bigtoolitems and 2 small items

eszkadev (changes requested):
dupl

loleaflet/src/control/Control.NotebookbarWriter.js
Andreas-Kainz: strange. But I can use the small items instead of the big one isn't it? or is there some technical issue?

cause than I will make a separate request for ALL NB home tabs.

Andreas-Kainz (commented):
@eszkadev I think add the ... replace stuff to _cleanText is a good idea, sure as @pedropintosilva suggest it is always good if the correct label come from core, but anyhow also if the correct label come from core, replace ... isn't wrong, isn't it?

loleaflet/src/control/Control.NotebookbarWriter.js
eszkadev: it is because you added new items but not deleted old ones from the JSON :)

PR #2070, opened 2021-04-21
unit-crash: avoid socket deletion before _socketPoll completes the async shutdown

vmiklos:
Change-Id: Ibb9f7d7681d502d06de92c2f5d8305c3f27a76c0


vmiklos:
Seems https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_android/ is red currently, so that's a false alarm.

PR #2079, opened 2021-04-21
fix #1285 remove elipsis ... from NB strings

### LINKED ISSUE #1285: Remove the ... at NB Labels

Andreas-Kainz:
Change-Id: I9d9d70c292772b349687d14fc3b92cd3fd5d86b7

* Resolves: #1284 
* Target version: master 

### Summary
No ... any more at the Notebookbar labels
Screenshot_20210421_003932

As there is already an _cleanText function to remove the ~ I added the '...' for replacement and add _cleanText to the NB labels.

If NB use the wrong strings from core this PR isn't wrong anyway cause it's remove the elipsis elements in addition to ~ when the string from core don't has ... it's also well.

### DIFF
--- a/loleaflet/src/control/Control.JSDialogBuilder.js
+++ b/loleaflet/src/control/Control.JSDialogBuilder.js
@@ -357,7 +357,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_cleanText: function(text) {
 		if (!text)
 			return '';
-		return text.replace('~', '');
+		return text.replace('~', '').replace('...', '');
 	},
 
 	_cleanValueFromUnits: function(text) {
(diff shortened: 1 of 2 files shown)


pedropintosilva:
yup, Thanks!
_I have re-opened the #1285 issue so we can keep track of the stringy-problem in the long run._

kendy:
@Andreas-Kainz, @pedropintosilva: I am sorry, but I have to revert this one :-(  This is causing a real cypress test failure, and I believe the failure is real - it removes the 3 dots from far more than it should, which then shows in the test...

Andreas-Kainz:
sorry that I broke something.

PR #2087, opened 2021-04-22
NB Calc update Home Tab

Andreas-Kainz:
Change-Id: Icfcae2e856f0724395eef35e861826062e4f5d24

* Resolves: # 
* Target version: master 

### Summary
Review of the Calc Home Tab.

*before*
2021-04-22 11_44_03-ws-98adfbea-8ab4-4524-85e9-30b032202165_0 - noVNC
*after*
2021-04-22 11_41_00-ws-98adfbea-8ab4-4524-85e9-30b032202165_0 - noVNC

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -430,6 +430,7 @@ div[id*='Row'].notebookbar, div[id*='Column'].notebookbar, #SendToBack.notebookb
 }
 
 #Copy.notebookbar,
+#DataFilterAutoFilter.notebookbar,
 #clearFormatting.notebookbar {
 	margin-top: 10px;
 }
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
next time I will check the git command from pedro

pedropintosilva:
Thanks @Andreas-Kainz! Could you please explain here better your changes? Otherwise it gets very easy to forget what and why when reviewing or backporting. Maybe it helps if I give you a small list of things to aim for(?):

- The commit message should tell what the commit does, in imperative mood. If necessary, you can mention how the code used to work, too. But what actually is done by the commit should be in present tense. Not what you did. Not "Fixed hang in foo()" but "Fix hang in foo()"
- The first line should be clearly less than 80 characters. It is supposed to be a minimal summary. It can be called the subject of the commit. Do not end it with a period.
- If the change is very simple conceptually, that one line might be enough.
- The rest of the commit message describes the change in greater detail (body). The body is written after the subject line with one blank line in between. The blank line is used by various tools (such as 'git log', 'git show', etc.) to differentiate subject from body. 
- Explain the problem being solved by this commit. More importantly, explain **why** these changes are being made, as opposed to how. The 'why' part is your responsibility, the 'how' part is code's responsibility.

I hope this helps : )
more info: https://chris.beams.io/posts/git-commit/

PR #2103, opened 2021-04-22
annotation: fix vex dialog losing focus on comment insert

lpranam:
Change-Id: Ifc13261ddd58c17acbe1af65bff6a49bd4a0bb4a

### Checklist

### DIFF
--- a/loleaflet/src/layer/marker/TextInput.js
+++ b/loleaflet/src/layer/marker/TextInput.js
@@ -8,7 +8,7 @@
  * text area itself.
  */
 
-/* global */
+/* global isAnyVexDialogActive */
 
 L.TextInput = L.Layer.extend({
 	initialize: function() {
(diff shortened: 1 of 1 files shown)


loleaflet/src/layer/marker/TextInput.js
```diff
@@ -209,7 +209,8 @@ L.TextInput = L.Layer.extend({
 			// on clicking focus is important
 			// specially in chrome once document loses focus it never gets it back
 			// which causes shortcuts to stop working (i.e: print, search etc...)
-			this._map.getContainer().focus();
+			if (!isAnyVexDialogActive()) // required because now comments can be inserted in mobile veiw mode
```
thebearon, line 212: There's a typo here (veiw). ;)

eszkadev (commented):
shouldn't we just use:
`if (isAnyVexDialogActive)
    return`
on the beginning of the `focus` method? to not overcomplicate code with additional conditions

where this `focus` is called? maybe we can avoid calling it at all?

eszkadev (approved):

PR #2104, opened 2021-04-22
annotation: fix vex dialog losing focus on comment insert

lpranam:
Change-Id: Ifc13261ddd58c17acbe1af65bff6a49bd4a0bb4a

* Target version: master 

### Checklist


lpranam:
for now, closing this because this problem is not reproducible on master but only on co-6.4.

6.4 patch: #2103

PR #2109, opened 2021-04-23
clang-format: turn off BreakBeforeBinaryOperators

vmiklos:
Improve readability by not suggesting changes like:

-                    if (JsonUtil::findJSONValue(object, "Url", url) &&
-                        JsonUtil::findJSONValue(object, "Name", filename))
+                    if (JsonUtil::findJSONValue(object, "Url", url)
+                        && JsonUtil::findJSONValue(object, "Name", filename))

Note that clang-format is never enforced (we never re-format existing
code), but your editor may respect it for new code.

Change-Id: Id6679220f83b77a4c63ca9a9429052535c75939a


vmiklos:
unit-crash fails, but passes for me locally...

mmeeks:
Thanks Miklos ! =)

PR #2110, opened 2021-04-23
StatusLine::parse: only feed integers to std::atoi()

vmiklos:
Change-Id: I7591fd14fbdc7c1210ac8141cb78854e3f20fffd


mmeeks (commented):
a band-aid I'm afraid; fundamentally atoi works on a cstring which needs a \0 termination and not a length. Ultimately we need a better atoi that can take a length internally anyway - perhaps we should just write that (and transiently copy the string into a stack buffer I guess) =) that could speedup a number of sillies we have in parsing around the place.

vmiklos:
> a band-aid I'm afraid; fundamentally atoi works on a cstring which needs a \0 termination and not a length. Ultimately we need a better atoi that can take a length internally anyway - perhaps we should just write that (and transiently copy the string into a stack buffer I guess) =)

Indeed, I even found an other sample that needs this, done in a 2nd commit, so we now check for both problems.

mmeeks (commented):
Urk - that could be an extremely expensive operation, and on the heap too =)
We really need a proper std::atoi that does a limited length, informed copy of the data up to it not being white-space, 0-9 '.' 'E' 'e' +/- etc. and then operates on that I think.

mikekaganski:
Can't std::from_chars be used here?

vmiklos:
> Urk - that could be an extremely expensive operation, and on the heap too =)
> We really need a proper std::atoi that does a limited length, informed copy of the data up to it not being white-space, 0-9 '.' 'E' 'e' +/- etc. and then operates on that I think.

Ah, I see. We know that the max valid length is MaxStatusLineLen, so we can error out before allocated a null-terminated string. Done that now in a follow-up commit. Does that look better? As far as I see, that's the only std::atoi() call in net/HttpRequest.cpp, but sure, we can extract that to a separate safe_atoi() or something like that.

@mikekaganski at the moment we're limited to C++11 in online.git.

vmiklos:
Eh, the last commit is not correct, I got confused about the length of that whole buffer vs the length of the line. Let me fix that.

vmiklos:
>  Happy to have this in for now though as a quick fix.

OK, thanks. I'll get back to a better atoi() in a separate PR.

cypress says only calc/formulabar_spec.js failed (which seems to be better than the baseline at https://cpci.cbg.collabora.co.uk:8080/job/tinderbox_online_master_debug/1199/console which does not get to cypress tests)

PR #2117, opened 2021-04-23
Use the accepted socket, not the listening one.

mmeeks:
Change-Id: Iecaa90f4c3030b5b2e20bb0e33ef9b75717bb578

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -743,9 +743,9 @@ std::shared_ptr<Socket> LocalServerSocket::accept()
         // Sanity check this incoming socket
         struct ucred creds;
         socklen_t credSize = sizeof(struct ucred);
-        if (getsockopt(getFD(), SOL_SOCKET, SO_PEERCRED, &creds, &credSize) < 0)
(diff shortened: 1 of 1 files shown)


net/Socket.cpp
```diff
@@ -743,7 +743,7 @@ std::shared_ptr<Socket> LocalServerSocket::accept()
         // Sanity check this incoming socket
         struct ucred creds;
         socklen_t credSize = sizeof(struct ucred);
-        if (getsockopt(getFD(), SOL_SOCKET, SO_PEERCRED, &creds, &credSize) < 0)
+        if (getsockopt(_socket->getFD(), SOL_SOCKET, SO_PEERCRED, &creds, &credSize) < 0)
         {
             LOG_ERR("Failed to get peer creds on " << getFD() << ' ' << strerror(errno));
```
kevans91, line 748: `_socket->getFD()` as well?

kevans91: I guess the listening fd might actually be more useful here.

mmeeks:
Good point =)

Ashod (approved):
Looks good, thanks.

PR #2122, opened 2021-04-23
Add a fuzzer for http::Response::readData()

vmiklos:
And remove the httpheader one, which is not useful, since it uses Poco
for the actual parsing, it did not find anything. (If we switch away
from Poco there in the future, it's easy enough to restore it.)

Also fix some problems found by the fuzzer.

Change-Id: I254247c46ecc78c9c3e75aac4f10c441b0e10fb3


vmiklos:
Just the cypress calc/formulabar_spec.js test fails, as far as I see.

Issue #2007, opened 2021-04-15
Notebookbar: use the same margins and paddings across all tabs


pedropintosilva:
**Describe the bug**
Currently we have different layouts (different html) with different
- top margins
- bottom margins 
- inner margins

![image](https://user-images.githubusercontent.com/65948705/114837938-81b9d800-9dd4-11eb-85b1-35aa4a2f8aca.png)

Andreas-Kainz:
My idea is to have everywhere, or at the first step, at the json tabs the same structure, so that if something in the css file was change, it will be changed everywhere and will be consistent. Also if the structure is everywhere the same, an issue from the code has to be changed only one time.

### Notebookbar Tab JSON Structure

![Section](https://user-images.githubusercontent.com/8517736/115877322-98e96d00-a447-11eb-8625-6d8f125a5ee4.png)

````
getInsertTab: function() {
		var content = [
			{
				'type': 'bigtoolitem',
				'text': _UNO('.uno:InsertPage', 'presentation'),
				'command': '.uno:InsertPage'
			},
			{
				'id': 'Insert-Section-Pages',
				'type': 'container',
				'children': [
					{
						'id': 'Insert-DoplicatePage',
						'type': 'toolbox',
						'children': [
							{
								'type': 'toolitem',
								'text': _UNO('.uno:DuplicatePage', 'presentation'),
								'command': '.uno:DuplicatePage'
							}
						]
					},
					{
						'id': 'Insert-SelectBackground',
						'type': 'toolbox',
						'children': [
							{
								'id': 'selectbackground',
								'type': 'menubartoolitem',
								'text': _UNO('.uno:SelectBackground', 'presentation'),
								'command': ''
							}
						]
					}
				],
				'vertical': 'true'
			},
		];

		return this.getTabPage('Insert', content);
	},
````
This structure as the benefit that it also work if there are more than one command and as @eszkadev  has in mind to simplify the json structure, it will be also very usefull, when the structure is everywhere the same.

**Comment by pedropintosilva** [comment]
Thanks @Andreas-Kainz for pasting that here! as I said in our meeting this sounds very promising and a good direction to aim for. (this, hopefully, all those quirkinesses will be easier to target/fix

**Comment by Andreas-Kainz** [comment]
what would be nice is, that the tab content (all togheter) can be vertical center align. Now there is something like in notebookbar.css, which should be removed.

```
#table-File.notebookbar
{
	margin-top: 10px;
}
```

[...]

PR #2125, opened 2021-04-23
Notebookbar Writer Layout Tab update #2007

Andreas-Kainz:
Update the structure of the Writer Layout Notebookbar Tab

Use the JSON Layout from #2007

### writer:
- Page Dialog is the most importend command, so it's the first one (before last one). In LibreOffice it's the last command, cause there are dropdown menues for aspect/ratio, size, ... which are not available for cool. So have the dialogue is ok but as bigtoolitem
- Title Page and Watermark switch to bigtoolitem, cause there are not that much commands in the Layout tab
- Wrap, Object Align and Forward/Back Object use the layout from the context Draw tab so it's everywhere the same layout.

### DIFF
--- a/loleaflet/src/control/Control.NotebookbarWriter.js
+++ b/loleaflet/src/control/Control.NotebookbarWriter.js
@@ -918,11 +918,16 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 	getLayoutTab: function() {
 		var content = [
 			{
-				'id': 'Layout-Section-File5',
+				'type': 'bigtoolitem',
+				'text': _UNO('.uno:PageDialog'),
+				'command': '.uno:PageDialog'
+			},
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
### Layout Tab Before
2021-04-23 23_45_22-ws-cdb0e9e9-e710-4615-863b-cac4f19fb659_0 - noVNC

### Layout Tab After
2021-04-23 23_53_46-ws-cdb0e9e9-e710-4615-863b-cac4f19fb659_0 - noVNC

Andreas-Kainz:
checked classic mode and in the menubar there was FormatColumns so I added it to the TitlePageDialog Watermark section.

2021-04-24 00_27_29-ws-cdb0e9e9-e710-4615-863b-cac4f19fb659_0 - noVNC

PR #2140, opened 2021-04-26
Add version of std::atoi() that needs no extra allocation

vmiklos:
std::atoi() assumes a null-terminated string and our strings are not
always null-terminated. So add a version that takes a length parameter,
this way we don't have to copy strings around.

Also switch to this in http::StatusLine::parse().

Change-Id: I449b356c1b9948c562434618596e8e3b38656088

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -44,6 +44,7 @@
 #include <sstream>
 #include <string>
 #include <thread>
+#include <limits>
 
 #include <Poco/Base64Encoder.h>
 #include <Poco/HexBinaryEncoder.h>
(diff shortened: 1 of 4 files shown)


common/Util.cpp
```diff
@@ -1042,6 +1043,52 @@ namespace Util
...
+                multiplier = -1;
+                ++offset;
+                break;
+            case '+':
+                ++offset;
+                break;
+        }
+        if (offset >= len)
+        {
+            return multiplier * ret;
```
mikekaganski, line 1074: This is 0, and integers don't have a concept of negative zero (and returning a negative zero for this would be wrong even if there was such a thing, because the input string is not a valid number) - why multiply here?

common/Util.cpp
```diff
@@ -1042,6 +1043,52 @@ namespace Util
...
+        }
+
+        while (isdigit(p[offset]))
+        {
+            long next = ret * 10 + (p[offset] - '0');
+            if (next >= std::numeric_limits<int>::max())
```
mikekaganski, line 1080: This relies on long being longer than int - so this would fail in a hypothetical Windows port.

A safer would be to check if ret is greater than (max / 10 - newDigit) before addition.

common/Util.cpp
vmiklos: You're right, that's indeed not needed, removed.

common/Util.cpp
vmiklos: Hmm, but you need to store the "next" value if you don't want to calc it twice. :-) So I just used an explicit std::int64_t here instead of long.

test/WhiteBoxTests.cpp
```diff
@@ -1955,6 +1957,45 @@ void WhiteBoxTests::testParseUrl()
                      net::parseUrl("https://sub.domain.com:80/some/path"));
 }
 
+void WhiteBoxTests::testSafeAtoi()
```
Ashod, line 1960: Worth testing nullptr and 0 length combinations, also, null-terminated string with wrong length (should work because '\0' is not a digit).

Ashod (approved):
Looks good, thanks!

test/WhiteBoxTests.cpp
vmiklos: Good idea, done. :-)

common/Util.cpp
mikekaganski: Sorry, missed the response - thanks, of course explicit 64-bit value works fine. Just FTR, I didn't suggest to calculate 'next' twice - I was thinking about something like

        while (isdigit(p[offset]))
        {
            int digit = p[offset] - '0';
            if (ret >= std::numeric_limits<int>::max() / 10 - digit) // this only does subtraction, since std::numeric_limits<int>::max() / 10 is compile-time constant
                return multiplier * std::numeric_limits<int>::max();
            std::int64_t next = ret * 10 + digit;
            ...

PR #2145, opened 2021-04-26
Private/dennisf/header resize fix co64

dennisfrancis:
* Resolves: # 
* Target version: co-6-4

### Summary

Fix :

1. header handle flicker on resize and cell cursor move.
2. update resize line on view move (arrow keys)
3. wrong sized header just after resize + mouse move (but goes away with another mousemove).

### Checklist


kendy:
Thank you for these Dennis! :-)

PR #2147, opened 2021-04-26
fix:formulabar_spec failing

Rash419:
Change-Id: I4b5eaa5294177b116d793ed8fc0d8925ee29f447


lpranam (approved):

lpranam:
this is a regression from a41ab8f. let's wait for a while this is not the correct solution.

lpranam (changes requested):
I think we should revert a41ab8f. what do you suggest @pedropintosilva

kendy:
Let's rather use https://github.com/CollaboraOnline/online/pull/2153 if it passes :-)

lpranam:
closing this in favour of #2153

Issue #2152, opened 2021-04-26
Fix insert line and shape icons


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @Andreas-Kainz 

![image](https://user-images.githubusercontent.com/65948705/116062511-41841080-a684-11eb-8b9d-b8cf579056a7.png)

- Insert Line icon: is really really tiny (rendering issues)
- Insert Shape icon: some rendering issues as well

location: `loleaflet/images/`

pedropintosilva:
Thanks  @Andreas-Kainz !
Does this https://github.com/CollaboraOnline/online/pull/2159 fixes it? 

![image](https://user-images.githubusercontent.com/65948705/116227179-2385e100-a754-11eb-88b4-6b75eb403f7b.png)

It seems so, maybe the shape outline could get better render when increasing outline but anyhow I think it's better, and if it's ok with you I can close this issue

Andreas-Kainz:
Instead of an ellipse I can use an rectangular shape, which will have less troubles with rendering.

PR #2159, opened 2021-04-26
Fix insert line and shape icon #2152

Andreas-Kainz:
Change-Id: I327dc32c22ee156a967fea3c6012d6bd0b677482

line icon is now a horizontal line, cause when click on line icon an horizontal line was drawn.
shape icon is now pixel align and use highlight color (blue), cause by default the shapes are blue.


pedropintosilva:
thanks and thanks for the short but very clear commit message! (I noticed that somehow the `Signed-off-by:` got moved to the top before the message) but anyhow nice to see that it passed all the tests : )

PR #2177, opened 2021-04-27
StringVector: add a way to get a number out of this without copying the token

vmiklos:
And use it in TileDesc::parse(), which is known to be a hot path.

Change-Id: I20375d7a1c31f61662446979e4d6799fd45b49d3

### DIFF
--- a/wsd/TileDesc.hpp
+++ b/wsd/TileDesc.hpp
@@ -215,9 +215,9 @@ public:
         TileWireId wireId = 0;
         for (std::size_t i = 0; i < tokens.size(); ++i)
         {
-            if (LOOLProtocol::getTokenUInt32(tokens[i], "oldwid", oldWireId))
+            if (tokens.getUInt32(i, "oldwid", oldWireId))
                 ;
-            else if (LOOLProtocol::getTokenUInt32(tokens[i], "wid", wireId))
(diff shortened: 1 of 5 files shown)


vmiklos:
Forgot to cover what's the actually returned value (number) in the test, let me fix that.

wsd/TileDesc.hpp
```diff
@@ -215,9 +215,9 @@ class TileDesc final
...
         for (std::size_t i = 0; i < tokens.size(); ++i)
         {
-            if (LOOLProtocol::getTokenUInt32(tokens[i], "oldwid", oldWireId))
+            if (tokens.getUInt32(i, "oldwid", oldWireId))
                 ;
-            else if (LOOLProtocol::getTokenUInt32(tokens[i], "wid", wireId))
+            else if (tokens.getUInt32(i, "wid", wireId))
                 ;
             else
             {
```
mmeeks, line 223: Awesome - but I think you missed the hottest call in the line below (?) =)

vmiklos: Yes, I did. I've now replaced LOOLProtocol::parseNameIntegerPair() with StringVector::getNameIntegerPair() in a follow-up commit that attempts to avoid copying in that case as well.

PR #2183, opened 2021-04-27
Restore missing rule to populate DIST_FOLDER with js.

mmeeks:
Used by android to build the JS it desires. We lost that in c6f2e2211d

Change-Id: I4b790a300aee66628baa2e5a2997685c42f709fa

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/Makefile.am
+++ b/loleaflet/Makefile.am
@@ -575,6 +575,15 @@ tscompile.done: $(LOLEAFLET_JS_SRC) $(LOLEAFLET_TS_SRC)
 $(LOLEAFLET_TS_JS_DST): tscompile.done
 	@echo -n ""
 
+$(DIST_FOLDER)/src/%.js: $(srcdir)/src/%.js
+	@mkdir -p $(dir $@)
+	@if test -z '$(ENABLE_BROWSERSYNC)'; then \
+		`cp $< $@`; \
+	else \
+		`echo $< $@`; \
+		`ln -sf $(abs_srcdir)/$< $@`; \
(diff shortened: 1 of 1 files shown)


mmeeks:
Dennis - any thoughts - not thinking so carefully here at night; but the Android code does essentially:

mkdir /tmp/dist_folder
make -C /opt/libreoffice/online/loleaflet DIST_FOLDER=/tmp/dist_folder

for some value of /tmp/dist_folder; and this fails with:

make: *** No rule to make target '/tmp/dist_folder/src/Leaflet.js', needed by '/opt/libreoffice/online/loleaflet/debug/loleaflet-src.js'.  Stop.
make: Leaving directory '/data/opt/libreoffice/online/loleaflet'

without this hunk of rules restored - cf.

commit c6f2e2211d76f744003d926063b74d905195ae5d
Author: Dennis Francis <dennis.francis@collabora.com>
Date:   Wed Apr 7 16:52:50 2021 +0530

    typescript: do project level compilation for ts files in src/

Of course perhaps this breaks something else ;-) Really there is a big tangle here - we should I suspect make everything work every time cleanly into the intermediate folder and then just do some copies into whatever state we want it at the end but ... ;-)

Thoughts appreciated; hopefully it fixes the Android tinderbox; lets see.

timar (approved):
I tried the Android build and it was successful with this patch.

mmeeks:
Absent any further opinion from CI =) lets push then =)

PR #2186, opened 2021-04-28
async save fuzzer findings

vmiklos:
- unit-crash: avoid socket deletion before _socketPoll completes the async shutdown
- Add an initial libfuzzer based fuzzer for http::Header::parse()
- Add a fuzzer for http::StatusLine::parse()
- clang-format: turn off BreakBeforeBinaryOperators
- StatusLine::parse: only feed integers to std::atoi()
- StatusLine::parse: handle non-null-terminated buffer with std::atoi()
- StatusLine::parse: limit size of allocated string
- Add a fuzzer for http::Response::readData()
- Add version of std::atoi() that needs no extra allocation
- StringVector: add a way to get a number out of this without copying the token
- StringVector: add a way to get a string-number pair out of this ...


vmiklos:
@Ashod could you please take a look? Thanks!

mmeeks:
Thanks Miklos ! =)

PR #2189, opened 2021-04-28
fix: Other user's caret is shown when adding new sheet

dennisfrancis:
Don't crash if marker object is not defined for a view.

Change-Id: If8f13da684ebf4028bff888c9edf4bac3ab359bf

* Resolves: # 
* Target version: master 

### Summary

Problem summary:
Open eg. the sample hello world ODS in two views,
With view A, create and switch to a new sheet, and start editing a cell (don't finish it),
With view B, switch to the other sheet, and then add a new sheet.
-> View A's caret can be seen on the new sheet.

Fix: Add check for existence of marker for a view before accessing its methods.

### Checklist


kendy:
Makes perfect sense, thank you! :-)

PR #2193, opened 2021-04-29
Fix view jump on autoinput

dennisfrancis:
* Resolves: # 
* Target version: master 

### Summary
This fixes two different kinds of view jumps on autoinput/autocomplete.

### Checklist


gokaysatir (approved):
Thanks for the fix:)

PR #2195, opened 2021-04-29
Fix Debian 8 / gcc-4.9 warnings

vmiklos:
net/HttpRequest.cpp:126:19: error: declaration of 'end' shadows a member of 'this' [-Werror=shadow]

kit/Kit.cpp:163:29: error: missing initializer for member 'statfs::f_bsize' [-Werror=missing-field-initializers]

kit/Kit.cpp:171:14: error: 'OVERLAYFS_SUPER_MAGIC' was not declared in this scope

Change-Id: Ibbd35ab5af3adad403ed22a0aeb70917b9e21970

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -303,6 +303,7 @@ kit_headers = kit/ChildSession.hpp \
               kit/Watermark.hpp
 
 noinst_HEADERS = $(wsd_headers) $(shared_headers) $(kit_headers) \
+                 test/HttpTestServer.hpp \
                  test/WopiTestServer.hpp \
                  test/countloolkits.hpp \
                  test/lokassert.hpp \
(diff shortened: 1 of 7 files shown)


Ashod (approved):
Thanks Miklos for the fixes. I wonder if we should support the same minimum gcc version on all targets. This seems fragile and likely to happen again because, as far as I know, 4.9 is not used by any dev.

vmiklos:
> This seems fragile and likely to happen again because, as far as I know, 4.9 is not used by any dev.

Indeed, if we have the power for it, then having a Debian9 builder as part of CI would be nice. Note that one change was specific to the kernel headers version, not specific to gcc-4.9 that Debian9 has.

@yunusem I've added a second commit that fixes the build from tarballs for me.

PR #2201, opened 2021-04-29
android: Fix build fail and crash on doc load

merttumer:
Signed-off-by: merttumer <mert.tumer@collabora.com>
Change-Id: Id998e2bdf7051d54784543015c7d34c0bdd64543

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -176,10 +176,13 @@ void UnitWSD::lookupTile(int part, int width, int height, int tilePosX, int tile
                          int tileWidth, int tileHeight,
                          std::shared_ptr<std::vector<char>> &tile)
 {
-    if (tile)
-        onTileCacheHit(part, width, height, tilePosX, tilePosY, tileWidth, tileHeight);
-    else
-        onTileCacheMiss(part, width, height, tilePosX, tilePosY, tileWidth, tileHeight);
+    if (isUnitTesting())
+    {
+        if (tile)
+            onTileCacheHit(part, width, height, tilePosX, tilePosY, tileWidth, tileHeight);
+        else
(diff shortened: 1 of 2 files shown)


net/Socket.cpp
```diff
@@ -77,6 +77,7 @@ bool StreamSocket::simulateSocketError(bool read)
     return false;
 }
 
+#if !MOBILEAPP && ENABLE_SSL
```
Ashod, line 80: ENABLE_SSL shouldn't be needed here (because SslStreamSocket is only defined with ENABLE_SSL).

But I'm surprised we need this condition here at all, because the declaration of simulateSocketError and its usage is not guarded with `#if !MOBILEAPP`. So I expect we can remove this condition and it wouldn't change anything.

common/Unit.cpp
```diff
@@ -176,10 +176,13 @@ void UnitWSD::lookupTile(int part, int width, int height, int tilePosX, int tile
                          int tileWidth, int tileHeight,
                          std::shared_ptr<std::vector<char>> &tile)
 {
-    if (tile)
-        onTileCacheHit(part, width, height, tilePosX, tilePosY, tileWidth, tileHeight);
-    else
-        onTileCacheMiss(part, width, height, tilePosX, tilePosY, tileWidth, tileHeight);
+    if (isUnitTesting())
```
Ashod, line 179: Would have been better to find out what was segfaulting. I suspect something not initialized on Android or something silly like that. What does the debugger show at the point of segfault ?

Ashod (commented):
Thanks @merttumer! Left some notes/questions.

Ashod (approved):
Good. Also needs backporting to master.

PR #2211, opened 2021-04-29
Avoid touching the main docLayer canvas size pieces unless necessary.

mmeeks:
Improves calc editing performance and interactivity significantly.

Change-Id: I1442d253ac3f449f7b9e7b5e7ac6dc01365e8dd3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thank you for the catch. My bad i couldn't think that this function would be called very often.

Issue #2130, opened 2021-04-24
Notebookbar Calc Sheet Tab Break commands


Andreas-Kainz:
Can someone test if the break commands from the Calc Sheet Tab will work? 

I can't test, cause there is no visual information.

![Sheeet-Before](https://user-images.githubusercontent.com/8517736/115973804-16040780-a558-11eb-80ef-1b55531185c3.png)

Andreas-Kainz:
@pedropintosilva or @eszkadev can someone please check if the commands will work? 

Cause if the Break commands want work, the Insert/Delete commands are already in Home Tab, Group/Ungroup commands I moved to Data #2131 and than there is only Freeze commands which are in the menubar related to View. So maybe we can remove the Sheet tab or make at least an Layout tab.

pedropintosilva:
- [ ] Test on master (didn't seem to work)
  - [ ] will  pull online and master and re-check it
- [ ] Test co-6-4 and compare

pedropintosilva:
As far I could test I do not see any reason to have any of these on CO:

![image](https://user-images.githubusercontent.com/65948705/116548606-c4130700-a8f4-11eb-9649-210d0c8cd0a6.png)

Or is there a way to use column break there? Isn't that supposed to be only for writer?

Andreas-Kainz:
This commands are available in calc (via right click, in the notebookbar only via Layout Tab -> Drop Down Menu at Layout Button on the right.

![2021-04-29 14_17_30-Unbenannt 1 - LibreOffice Calc](https://user-images.githubusercontent.com/8517736/116549385-b27e2f00-a8f5-11eb-9df6-5a8892104c12.png)

As you can see in the screenshot it's only usefull if there is an additional view where you can see all the breaks.

Andreas-Kainz:
As long as they are not visible, I would drop the commands from the Sheet Tab.

pedropintosilva:
yup but since we do not have that here (define print region, we cannot see the layout) we should probably drop it? will ask @kendy about this and if there is no other reason we can then drop the whole tab ; )

Andreas-Kainz:
The freeze command's will be moved to somewhere else, cause they work. But sure the Sheet Tab can than be dropped.

PR #2215, opened 2021-04-29
Notebookbar Clac Sheet tab was renamed to Layout #2130

Andreas-Kainz:
In writer there is a layout tab and also in LibreOffice calc. There isn't really a lot of content only Page property Insert and Delete Row/Column Freeze and Alignment stuff like in writer.

Row/Column Break stuff was removed cause it didn't work, but can be added when available.

Tested also to add Freeze command into Data tab, but than it's way to much commands and as there are not that much tabs, I use Layout tab.

Screenshot_20210429_235638

Change-Id: I2f7f4aeb1bd4f2a2b351e581d20589d37a20d87f


pedropintosilva:
Strange is not passing the tests, I will test this out and check also the tests locally.

PR #2227, opened 2021-04-30
Tile load slurping.

gokaysatir:
The browsers - unfortunately appear to round-robin the asynchronous
unload of an individual tile - which causes a _tileReady callback
with updating the canvas.

This means that we re-render the canvas (at whatever speed) for each
tile we load - guarenteeing a nice animated effect of loading tiles
one by one, and refreshing the view: not so cool.

So - instead, as per WebSocket messages, slurp them - and then
refresh the view in one shot after 1ms; since it is fairly bogus
to convert a base64 encoded URL image to a real image asynchronously
anyway, this gives some nice results eg.

TileLayer.js:2464 Slurp tiles 34
TileLayer.js:2464 Slurp tiles 25
TileLayer.js:2464 Slurp tiles 53
TileLayer.js:2464 Slurp tiles 5
TileLayer.js:2464 Slurp tiles 3

It also significantly improves pageing down in calc, and presumably
a number of other operations.

Change-Id: Ib2a7daf8aa35835edd78dc5da187b0cb76b76ac3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/TileLayer.js
+++ b/loleaflet/src/layer/tile/TileLayer.js
@@ -2168,7 +2168,24 @@ L.TileLayer = L.GridLayer.extend({
 	},
 
 	_tileOnLoad: function (done, tile) {
-		done(null, tile);
+		// slurp multiple async tile loads into one render:
+		// FIXME: probably we should nicely combine
+		// async image decoding into Socket.js' _slurpMessage
+		// to save another 1ms timer.
+		var that = this;
+		if (!that._slurpQueue || !that._slurpQueue.length) {
+			setTimeout(function() {
+				for (var i = 0; i < that._slurpQueue.length; i += 2) {
+					var slurpTile = that._slurpQueue[i];
+					var slurpCallback = that._slurpQueue[i+1];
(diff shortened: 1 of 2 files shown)


loleaflet/src/layer/tile/TileLayer.js
```diff
@@ -2168,7 +2168,25 @@ L.TileLayer = L.GridLayer.extend({
...
 	_tileOnLoad: function (done, tile) {
-		done(null, tile);
+		// slurp multiple async tile loads into one render:
+		// FIXME: probably we should nicely combine
+		// async image decoding into Socket.js' _slurpMessage
+		// to save another 1ms timer.
+		var that = this;
+		if (!that._slurpQueue || !that._slurpQueue.length) {
+			setTimeout(function() {
+				console.log2('Slurp tiles ' + that._slurpQueue.length / 2);
```
mmeeks, line 2178: we should comment that out on all branches, after verifying that it is getting similar numbers to the above =) otherwise great. Thanks Gokay do merge on your own.

PR #2235, opened 2021-04-30
private/andreas/private/akainz/notebookbar text dropdown

Andreas-Kainz:
- Notebookbar.css cleanup unrelated values was removed
- Text dropdown fit into notebookbar row height


Andreas-Kainz:
### Home Tab Before
home-before

### Home Tab After
home-after

Issue #2088, opened 2021-04-22
Alignment of Color Picker preview icons


Andreas-Kainz:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @eszkadev @pedropintosilva @Andreas-Kainz 

The commands Color and Background color show an color preview, but as the svg icon wasn't hidded by the preview the icons are larger and non align.
![2021-04-22 11_35_05-ws-98adfbea-8ab4-4524-85e9-30b032202165_0 - noVNC](https://user-images.githubusercontent.com/8517736/115694120-d9ba8680-a360-11eb-8fbd-7baabcb5412b.png)

code pointer: https://github.com/CollaboraOnline/online/blob/a41ab8ff354900ade97b75a2ed8641140964c2a6/loleaflet/src/control/Control.NotebookbarBuilder.js#L420

pedropintosilva:
Minutes (part just the ones related to this) from our meeting, @Andreas-Kainz :

Elements that needs some adjusting in size:

- [x] Style previews
- [ ] Text dropdown
- [ ] Button dropdown

Let's focus item by item
- Button dropdown
  - We need to add one more CSS class for it `has-dropdown` (Add class to parent of `hasDropdownArrow` element on JS)
  - Use class to adjust:
    - `selected-color` top position 
    - Update icon so it's still 24px but visually is shorter (accounting for `selected-color`)
    - Parent element can then have a height of 24px ensuring that everything is aligned

pedropintosilva:
Unfortunately https://github.com/CollaboraOnline/online/commit/93bc5f6083291f28460ccc7fd6c4e6b6ff55112f add regression in classic mode:

PR #2237, opened 2021-05-01
Notebookbar height 72px instead of 84px #2088

Andreas-Kainz:
with #2235 and the fix of the home tab height it's now possible to reduce the notebookbar height to 72px in height instead of 84px.
Sure there would be 68 px theoretical possible, but the additional 4px in height are good for the home tab.

Change-Id: Id28bcc63db142d235986cb5449d750808b684c4f

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -206,7 +206,7 @@
 	scrollbar-width: none; /* Firefox */
 	-ms-overflow-style: none;  /* Internet Explorer 10+ */
 	display: flex;
-	height: 84px;
+	height: 72px;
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
### writer before
writer-d-01

### writer after
writer-d-02

### writer compact before
writer-c-01

### writer compact after
writer-c-02

## calc before
calc-d-01

# calc after
calc-d-02

### impress before
impress-d-01

### impress after
impress-d-02

Andreas-Kainz:
with this patch the notebookbar layout is done. Hey we reduce **12px** in height !

pedropintosilva (approved):

PR #2239, opened 2021-05-03
Fixes to properly close connections and not access released socket

Ashod:
- wsd: http: complete the http response only once
- wsd: http: add isConnected getter to http::Session
- wsd: http: disconnect if the server closed the socket

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -858,8 +858,8 @@ private:
         , _timeout(getDefaultTimeout())
         , _connected(false)
     {
-        assert(!_host.empty() && portNumber > 0 && !_port.empty()
-               && "Invalid hostname and portNumber for http::Sesssion");
+        assert(!_host.empty() && portNumber > 0 && !_port.empty() &&
+               "Invalid hostname and portNumber for http::Sesssion");
 #ifdef ENABLE_DEBUG
         std::string scheme;
         std::string hostString;
@@ -948,6 +948,7 @@ public:
     const std::string& port() const { return _port; }
     Protocol protocol() const { return _protocol; }
     bool isSecure() const { return _protocol == Protocol::HttpSsl; }
+    bool isConnected() const { return _connected; };
 
     /// Set the timeout, in microseconds.
     void setTimeout(const std::chrono::microseconds timeout) { _timeout = timeout; }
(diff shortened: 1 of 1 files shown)


net/HttpRequest.hpp
```diff
@@ -1140,12 +1149,18 @@ class Session final : public ProtocolHandlerInterface
         {
             // Remove consumed data.
             data.erase(data.begin(), data.begin() + read);
+            close = !isConnected();
```
mmeeks, line 1152: Why do we introduce this extra method / channel here to show that the socket is closed ? what is wrong with returning -1 from readData?
Why do we need to handle the disconnect this time around ? we can always spin the loop again if we read a chunk of data, and ? - I'm really missing a rationale here. I guess it is at least not some std::pair<> we're returning ;->

mmeeks: Still - it's all isolated to the http code at least ;-)

Ashod: Returning -1 will imply an error (we won't remove the read data and will simply close the connection). That's not what we want here. We have consumed the full data and there is nothing more to do. We just need to close the connect and release the socket, but *only* if the other side does not want to reuse the socket. So this is a condition that we learn about from the header, which is something that we get after parsing it.

Sorry if the commit log is didn't expound on all this. Perhaps it's not obvious that we should maintain the connection in non-error cases by default (it's a requirement of the RFC for HTTP/1.1).

mmeeks: Ah! - ok, makes perfect sense =) Of course - we can store up the completion information and just return that state on the next readData - surely (?) there's no need for another flag. Just to check - we are coping with the case where the poll returns not only a chunk of data, but also a connection-closed flag - so we process all the data before closing (?) =)

Ashod: @mmeeks Yes, that is what we do: we tell poll to close the connection (and remove our socket etc.). Unfortunately, it seems our poll is prone to invoking handleIncomingMessage even after we have set the disposition to Closed. That was the issue fixed here. The `if (!isConnected())` at the top of the function is to protect against rentrancy after closing and releasing the socket member object (which was segfaulting).

Also worth mentioning that there is *no* new state: the `_connected` member had been there because it's used to reconnect upon re-use (or indeed to connect the very first time an http::Session is used).

PR #2253, opened 2021-05-03
Use Util::getThreadId() instead of Linux system call.

arrowd:
Change-Id: I5cab82f0d869cc9f2ea97ac9320d8f4bbf466c37

* Target version: master 

### Checklist


mmeeks:
Tor? why not use the helper ?

mmeeks:
Thanks Gleb ! =)

tml1024:
Because then the included test program won't compile;) But no problem.

PR #2270, opened 2021-05-05
Message: cleanup message parsing of payloads.

mmeeks:
Change-Id: Id11a702590539ff4f20fa56272beeb6d91b9e7bd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Message.hpp
+++ b/common/Message.hpp
@@ -28,7 +28,7 @@ public:
     Message(const std::string& message,
             const enum Dir dir) :
         _forwardToken(getForwardToken(message.data(), message.size())),
-        _data(skipWhitespace(message.data() + _forwardToken.size()), message.data() + message.size()),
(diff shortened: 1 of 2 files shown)


common/Message.hpp
```diff
@@ -161,14 +159,21 @@ class Message
...
+        for (i = fromOffset; i < len; ++i)
         {
-            ++p;
+            if (p[i] != ' ')
+                break;
         }
-
-        return p;
+        if (i < len)
+            return std::vector<char>(p + i, p + len - i);
```
Ashod, line 174: The second argument is the last item pointer, not the length. So should be `p + len`.

Ashod (commented):
Thanks!

Ashod (approved):
Great, thanks!

PR #2279, opened 2021-05-05
Asynchronous Uploading

Ashod:
These patches enable async uploading and add tests.

- wsd: add async API to Storage
- wsd: support asynchronous upload requests to storage
- wsd: handle async upload callback
- wsd: upload after saving only if necessary
- wsd: better logging of upload state
- wsd: better logging of autosave, upload, and modified states
- wsd: test: add async upload test on closing
- wsd: test: add async upload test to modify after fail


mmeeks:
Will send some comments by E-mail on the delta from end to end.

PR #2280, opened 2021-05-05
Introduce a TRACEEVENT message from client to server

tml1024:
Introduce a TRACEEVENT message from client to server
    
    Add a function emitInstantTraceEvent() to leaflet that sends an
    "Instant" Trace Event to the server for logging.
    
    Add a function createAsyncTraceEvent() that creates an object that
    records the timestamp of its creation and sends the 'b' event to the
    server for logging, and when you call finish() on the object, sends
    the corresponding 'e' event.
    
    Finally, add a function createCompleteTraceEvent() that creates an
    object for a "Complete" Trace Event that includes the end timestamp
    *and* the duration. The event is sent to the server when you call
    finish() on the object.
    
    Loleaflet sends timestamps in the above messages from
    performance.now(). To enable the server to turn those into absolute
    timestamps, the loolclient message is amended to include the current
    Date.now() and performance.now() values.
    
    Note that the intent is that when generating Trace Event logs, the
    server and the web browser run on the same machine, so there is no
    wall-clock synchronisation issues between server and client.

Change-Id: Ie9e68b093b769cc942e1e1d17083febeb07ccf5e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/control/Toolbar.js
+++ b/loleaflet/src/control/Toolbar.js
@@ -1,4 +1,4 @@
-/* -*- js-indent-level: 8 -*- */
+/* -*- js-indent-level: 8; fill-column: 100 -*- */
 /*
  * Toolbar handler
  */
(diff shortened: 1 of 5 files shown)


tml1024:
I wonder if that failing Cypress test for some Impress thing is systemic and happens for everybody at the moment, or actually caused by this PR?

tml1024:
Oh wow, it finally passed.

mmeeks:
Looks good - I'd be inclined to jam multiple trace events into a single web socket message for performance reasons =) I expect browsers are also rather bad at not sending one message-per-packet and so on.

PR #2283, opened 2021-05-05
Read Only Mode: Allow changing selected part with PageUp & PageDown k…

gokaysatir:
…eys.

Change-Id: Ifb2451c627e984063efe6d6d17b2c8e76231e1d0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/map/handler/Map.Keyboard.js
+++ b/loleaflet/src/map/handler/Map.Keyboard.js
@@ -387,8 +387,12 @@ L.Map.Keyboard = L.Handler.extend({
 				ev.preventDefault();
 			}
 		}
-		else if (!this.modifier && (keyCode === 33 || keyCode === 34)) {
-			// let the scrollbar handle page up / page down when viewing
+		else if (!this.modifier && (keyCode === 33 || keyCode === 34) && ev.type === 'keydown') {
+			if (this._map._docLayer._docType === 'presentation' || this._map._docLayer._docType === 'drawing') {
+				var partToSelect = this._map._docLayer._selectedPart;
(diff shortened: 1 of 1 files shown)


loleaflet/src/map/handler/Map.Keyboard.js
```diff
@@ -387,9 +387,12 @@ L.Map.Keyboard = L.Handler.extend({
...
 		}
-		else if (!this.modifier && (keyCode === 33 || keyCode === 34)) {
-			// let the scrollbar handle page up / page down when viewing
-			return;
+		else if (!this.modifier && (keyCode === 33 || keyCode === 34) && ev.type === 'keydown') {
+			if (this._map._docLayer._docType === 'presentation' || this._map._docLayer._docType === 'drawing') {
+				var partToSelect = this._map._docLayer._selectedPart;
+				partToSelect = keyCode === 33 ? partToSelect - 1 : partToSelect + 1;
+				this._map.setPart(partToSelect);
+			}
```
kendy, line 395: Now we don't do a 'return', and stop propagation that way; which was not what was there previously...  Is that intentional, or not, please? :-)

dennisfrancis (commented):
I tested in chrome's iPad emulation to get read-only mode for a pdf, pgdn/pgup works there. I guess when you load pdf in chrome-desktop without emulation it is not in read-only mode, so there it does not work at least when one of the pages in slide sorter is in focus - it scrolls the sorter view which is a reasonable thing to do. Other than Kendy's question, I find no issues with this patch. Thanks.

gokaysatir:
Thanks for the reviews:)

I added return statement. Will merge it when it compiles.

PR #2285, opened 2021-05-05
Notebookbar Tab Content vertical-align

Andreas-Kainz:
Change-Id: Iaf876e786e8bc24dc6c03c412d12dc5e4700119c

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -18,9 +18,10 @@
 	font-family: var(--loleaflet-font);
 	line-height: normal;
 	color: dimgray;
-	height: 24px;
+	height: 32px;
+	display: flex;
+	align-items: center;
 	padding: 0px 1em;
-	padding-top: 7px;
 }
 
 .ui-tab.hidden.notebookbar {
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
### Alignment master
Screenshot_20210505_224807

### with Nextcloud Integration
Screenshot_20210505_224736

Andreas-Kainz:

checked with chrome and firefox on desktop.

loleaflet/css/notebookbar.css
```diff
@@ -79,7 +80,7 @@
 	display: inline-block;
 	position: relative;
 	z-index: 1000;
-	height: 100%;
+	height: 32px;
```
pedropintosilva, line 83: I don't think this change is necessary. Can you please check and if it's not remove it from the commit?

pedropintosilva (changes requested):
also please change the commit subject and add a message as well.:

- Subject: what the commit does in an imperative and succinct way (minimal summary and never ending with a period). Example: `Notebookbar: Tab: Fix vertical alginment`
- Blank line
- Message: 
  - What it fixes (present tense)
  - If the change is very simple conceptually, that one line might be enough
  -  Explain the problem being solved by this commit. More importantly, '''explain ''why'' these changes are being made''', as opposed to ''how''
  - Example: 
```
Center align tab titles (vertically) while removing when
possible hard-coded measurements from nearby elements:

- element a
- element b
- element c

Remove unnecessary padding from cyx
```

loleaflet/css/notebookbar.css
Andreas-Kainz: checked and isn't necessary.

PR #2292, opened 2021-05-07
wsd: create the unittest SocketPoll thread only when testing

Ashod:
This avoids having the test socket poll thread running when not testing (which was recently introduced with the async sockets in tests).

Also, initialize SSL when either termination is set or it's enabled outright.


mmeeks (approved):
Wow - I missed that ...

mmeeks:
Has a +1 from me - in concept - please do merge as/when CI is happy. Thanks Ash!

PR #2294, opened 2021-05-07
ClientSession: fix -Wdangling-gsl

vmiklos:
wsd/ClientSession.cpp:411:31: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
            const char* str = tokens[2].data();
                              ^~~~~~~~~

Change-Id: I67e4aa5f61e049ad13663dc1e11ae2bd67399bea


vmiklos:
@tml1024 the trap here is that StringVector has a single underlying (non-null-terminated) std::string, so operator[] returns an std::string, not an std::string&, so tokens[2].data() is not safe to use in later lines. If you backport the original commit, could you please backport this one as well? Thanks.

PR #2326, opened 2021-05-11
improve cell cursors and selections

dennisfrancis:
* No AA bleeding: render pixel aligned (ie. 0.5 offset) for all hair-lines
* Two pixel wide line, around the cell - black; with the inner pixels /
  line exactly on top of the cell border.
* inside this breaking the outline - an internal white border which is
  exactly 1 pixel inside that black border - for a high contrast look.

The change introduces CPathGroup which is just an array of CPaths that
allows separate styling. CCellCursor and CCellSelection classes
specialize CPathGroup and implements the above three guidelines.

The view cell-cursors and view-selection also use the same
implementation with different styling.

The cypress tests are updated accordingly for the name change in the
test-div.

Change-Id: I8881337df6cf8d543bd48e8de4560d9aab681dff

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -56,7 +56,7 @@ function clickOnFirstCell(firstClick = true, dblClick = false) {
 		});
 
 	if (firstClick && !dblClick) {
-		cy.get('#test-div-overlay-cell-cursor')
+		cy.get('#test-div-overlay-cell-cursor-border-0')
 			.should(function (elem) {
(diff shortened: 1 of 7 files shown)


kendy:
@dennisfrancis: Looks great to me in general, let's get it in :-)  Would be great to add few purpose comments for some of the methods - can you do?  Thank you!

mmeeks:
Would love to back-port the auto-fill marker patches too for 6.4.9 - that'd be great - good to have more in TS & canvas sections =)
Thanks !

dennisfrancis:
@kendy I've added purpose comments.

PR #2333, opened 2021-05-11
Add the possibiliy to include arbitrary arguments in the generated Tr…

tml1024:
…ace Events in JS, too

For now, an argument object should have all keys and values as strings
not containing any spaces because of the trivial way in which it is
parsed.

This is a bit sad, yes, but necessary until we move to some more
structured syntax for the messages.

Also fix an incorrect comment.

Change-Id: I8a1408a4a1787b66a3cf7b26b3d92c07df244c47

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -1369,25 +1369,34 @@ app.definitions.Socket = L.Class.extend({
 
 	traceEventRecordingToggle: false,
 
-	emitInstantTraceEvent: function (name) {
+	_stringifyArgs: function (args) {
+		return (args == null ? '' : (' args=' + JSON.stringify(args)));
+	},
+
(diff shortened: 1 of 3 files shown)


tml1024:
Let's continue iterating until the Cypress test succeeds.

tml1024:
> If you like it lets get it in; but ~everything in the world goes through those _onMessage functions 

Nota that this PR does not touch the _onMessage functions.

mmeeks:
My mistake - sounds perfect then =) as/when CI looks happy.

tml1024:
Wait, there are no checks for even build on Linux any longer?

tml1024:
Superseded by https://github.com/CollaboraOnline/online/pull/2372

PR #2334, opened 2021-05-11
private/pedro/user feedback

pedropintosilva:
- add initial feedback option
- loleaflet: add initial iframe dialog
- loleaflet: add feedback handler
- loleaflet: handle iframe load error
- add option --with-feedback-location
- simplify
- Feedback: Style iframe

### DIFF
--- a/loleaflet/src/map/handler/Map.Feedback.js
+++ b/loleaflet/src/map/handler/Map.Feedback.js
@@ -30,7 +30,6 @@ L.Map.Feedback = L.Handler.extend({
 				this._iframeDialog.remove();
 
 			this._iframeDialog = L.iframeDialog(window.feebackLocation);
-			this._iframeDialog.create();
 		}
 	},
 
--- a/loleaflet/src/map/handler/Map.Feedback.js
+++ b/loleaflet/src/map/handler/Map.Feedback.js
@@ -26,15 +26,26 @@ L.Map.Feedback = L.Handler.extend({
 
 	onFeedback: function () {
 		if (window.localStorage.getItem('WSDFeedbackEnabled')) {
+			if (this._iframeDialog && this._iframeDialog.hasLoaded())
+				this._iframeDialog.remove();
+
 			this._iframeDialog = L.iframeDialog(window.feebackLocation);
 			this._iframeDialog.create();
 		}
 	},
 
+	onError: function () {
+		window.localStorage.removeItem('WSDFeedbackEnabled');
+		this._iframeDialog.remove();
+	},
+
 	onMessage: function (e) {
 		var data = e.data;
 
-		if (data == 'never') {
+		if (data == 'show') {
+			this._iframeDialog.show();
+		}
+		else if (data == 'never') {
 			window.localStorage.removeItem('WSDFeedbackEnabled');
 			this._iframeDialog.remove();
 		} else if (data == 'later') {
--- a/loleaflet/src/map/handler/Map.Feedback.js
+++ b/loleaflet/src/map/handler/Map.Feedback.js
@@ -1,0 +1,52 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * L.Map.Feedback.
+ */
+
+L.Map.mergeOptions({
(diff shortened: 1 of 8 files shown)


pedropintosilva:
@eszkadev could you be the second pair of eyes that reviews this PR? Particularly the part regarding building by default without enable_feedback on (as requested by @kendy today)

loleaflet/src/map/handler/Map.Feedback.js
```diff
@@ -0,0 +1,62 @@
...
+			setTimeout(L.bind(this.onFeedback, this), this._map.options.feedbackTimeout);
+		} else if (data == 'submit') {
+			window.localStorage.removeItem('WSDFeedbackEnabled');
+			this._iframeDialog.remove();
+		}
+	}
+});
+
+if (window.isLocalStorageAllowed) {
+	L.Map.addInitHook('addHandler', 'feedback', L.Map.Feedback);
```
eszkadev, line 61: I think we should add feedback functionality only when `window.feebackLocation` is defined and not empty to not show feedback request for customer builds where it is not wanted

eszkadev: I know this file is included only when feedback is enabled but to be sure...

loleaflet/src/control/Toolbar.js
```diff
@@ -609,6 +611,9 @@ L.Map.include({
 				if (!calledFromMenu) {
 					localStorage.setItem('WSDWelcomeVersion', app.socket.WSDServer.Version);
 				}
+				map.welcomeClosed = true;
+				// call Map.Feedback.js or sent a onMessage value?
+				// app.socket._onMessage('feedback-later');
```
pedropintosilva, line 616: @hcvcastro (cc: @eszkadev ) as the commit message explains I think we need to avoid showing the feedback iframe while the welcome dialog is opened, specially because they both appear without being triggered by the user. Bottom line is I couldn't figure out how to do that... so help would be welcomed

pedropintosilva: feel free to remove this commit and add your own

eszkadev: @pedropintosilva somtehing like this should be fine:

`var checkWelcomeAndRunFeedback = function() {
   if (welcome.exists())
         setTimeout(checkWelcomeAndRunFeedback, 3000); // wait 3s more
   else
         runFeedback();
}

setTimeout(checkWelcomeAndRunFeedback, initialDelay)`

pedropintosilva: @hcvcastro : I'm getting a bit nervous with the time passing (and the probably of dealing again and again with some conflicts) of this PR so, I end up removing the WIP commit and only adjusting the timer value. And let's merge as soon as we get the green again. The additional commit (to prevent welcome and feedback dialog appearing at the some time) can be added later and, at least for now, the longer timer should already make for a better experience.

pedropintosilva: > @pedropintosilva somtehing like this should be fine:
> 
> ```
> var checkWelcomeAndRunFeedback = function() {
>    if (welcome.exists())
>          setTimeout(checkWelcomeAndRunFeedback, 3000); // wait 3s more
>    else
>          runFeedback();
> }
> 
> setTimeout(checkWelcomeAndRunFeedback, initialDelay)
> ```

@eszkadev the time should always be `30000` and what we should do is not show feedback If welcome.exists(). Meaning: not delaying; instead really not showing at all until the close of welcome is triggered via vex `_showWelcomeDialogVex`:`beforeClose`. Just upon being close we start a timer. This will ensure that a person can take as much time as desired to read the welcome dialog and check all the various links that are in it.

pedropintosilva: but as I said before let's add that in in a different PR, so this one can get merged without further delay

pedropintosilva: Thanks @eszkadev for the guidance I have added a new PR : https://github.com/CollaboraOnline/online/pull/2969

Issue #2149, opened 2021-04-26
Notebookbar Character Spacing


Andreas-Kainz:
**Describe the bug**
Character Spacing isn't availbale at cool.

![2021-04-26 11_08_36-Untitled 1 - LibreOfficeDev Writer](https://user-images.githubusercontent.com/8517736/116058154-d46e7c00-a67f-11eb-95c1-0565d45614e4.png)

Andreas-Kainz:
IF the character spacing command would be available the character section would have in writer, calc, impress, draw everywhere the same width. Cause in calc there is the border line command, so for consistency it would be a big improvement. And for sure, it's a missing command.

Issue #2085, opened 2021-04-22
Notebookbar: Home tab: positioning direct formatting icons


pedropintosilva:
We are having holes while not using space to separate different group of icons:
![image](https://user-images.githubusercontent.com/65948705/115679725-29925100-a353-11eb-9e33-e2513adfa1ed.png)

In Calc and Writer (the most flagrant cases) we can:
- add space before and after that specific group with half the size of the icon with (=16px)

In Impress there is not that much we can do without rethink other things
- but for now we could maybe just 2 times the previous space (=32px)

Andreas-Kainz:
![2021-04-22 10_03_38-example ods - Dateien - Nextcloud](https://user-images.githubusercontent.com/8517736/115680826-382d3800-a354-11eb-8841-b4550dcd7893.png)

The plan is to have somewhere an end align element. will have to have a look to the css file.

PR #2337, opened 2021-05-12
Notebookbar: Home tab: positioning direct formatting icons #2085

Andreas-Kainz:
Calc has the most icons in the direct formatting group so the width of the fontnamecombobox was enlarged to optimize the group at calc
Writer, Impress/Draw use one icon less so I added an 44px space. By add Character Spacing command #2149 there are everywhere the same number of commands, so the 44px margin can be removed.

Change-Id: I8782cc738d79174a7f78f9e9e3fbd604f64e2233

### DIFF
--- a/loleaflet/css/notebookbar.css
+++ b/loleaflet/css/notebookbar.css
@@ -400,7 +400,19 @@
 }
 
 #fontnamecombobox.notebookbar .select2 {
-	width: 150px !important;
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
### writer after
Screenshot_20210512_085405

### calc after
Screenshot_20210512_085423

PR #2341, opened 2021-05-12
Properly process URLs with existing query

mikekaganski:
URLs from host may already contain a query, like

  http://server.tld/foo/wopi/files/bar?desktop=baz

Previously, we simply appended '?' and then our parameters, resulting in

  http://server.tld/foo/wopi/files/bar?desktop=baz?access_token=...

which later was processed as if the question mark and following parameter
were part of previous parameter's value, so proper 'access_token' parameter
was missing.

This checks if the URL contains '?', and uses proper separator accordingly.

Change-Id: I1a237c0e47e1eea3704ef7d4a8a596283ea2a241

### DIFF
--- a/loleaflet/js/global.js
+++ b/loleaflet/js/global.js
@@ -752,7 +752,9 @@ window.app.definitions = {};
 		global.socket = new global.FakeWebSocket();
 		window.TheFakeWebSocket = global.socket;
 	} else {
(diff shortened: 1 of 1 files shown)


mmeeks:
A lovely fix - thanks for that Mike ! =) are you sure this is the only place this needs doing ? anyhow +1 for the back-port to co-6-4

mikekaganski:
> are you sure this is the only place this needs doing ?

Not much in this world I'm sure about - and of course I'm not sure here, but I don't see other possible places easily :-)

mmeeks:
Great =)

Issue #2357, opened 2021-05-13
Help dialogs: not selectable and flimsy 


pedropintosilva:
**To Reproduce**
1. User opens any help dialog (keyboard shortcuts or general help)
2. Dialog tens to close accidentally if the user is reading and by mistake clicks somewhere else
3. User tries to select text either to save in his own notes, to share or to simply report something
4. Text is unelectable and prone to close the dialog (user tried to select it multiple times and ends up clicking out side in the overlay closing it complexity)

**Expected behavior**
- This type of dialogs with a lot of text should only close when pressing close button
- Should be selectable

PR #2359, opened 2021-05-13
Help dialogs: avoid closing it by mistake and make it selectable

pedropintosilva:
- Only close when user expressly pressed the close button
- Make it possible to select text

This type of dialogs with a lot of text should only be close when
pressing close button so to avoid accidentally closing it. The text
should also be selectable, allowing the user to save it in his own notes
, to share or to use it to simply report something

Fixes https://github.com/CollaboraOnline/online/issues/2357

Change-Id: I88e01904bad4d0fa2504296c892e28ba6df9d112

### DIFF
--- a/loleaflet/css/override-vex.css
+++ b/loleaflet/css/override-vex.css
@@ -74,6 +74,15 @@
 	outline: none;
 }
 
+.vex-selectable {
+	-webkit-touch-callout: text;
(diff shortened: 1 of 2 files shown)


pedropintosilva:
- tested locally
  - there are no failures related to this change

PR #2362, opened 2021-05-13
avoid rendering for execution of ~every message

dennisfrancis:
Send the canvas container into a paused state just before a batch of
queued messages are going to be executed. During this state the
container dirties itself for every requestRedraw() call. After all
messages in the current slurp batch is executed, resume rendering
starting with a re-render if necessary based on dirty flag.

Change-Id: Ice5347aac78f4f19642b96edb929836b15c55599

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -269,6 +269,9 @@ L.Socket = L.Class.extend({
 	},
 
 	_emitSlurpedEvents: function() {
+		if (this._map && this._map._docLayer) {
+			this._map._docLayer.pauseDrawing();
+		}
 		// console.log2('Slurp events ' + that._slurpQueue.length);
 		for (var i = 0; i < this._slurpQueue.length; ++i) {
 			var evt = this._slurpQueue[i];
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Looks great I think we prolly need it for the tile event slurping too.
Thanks!

mmeeks:
Performs very nicely =) good stuff.

dennisfrancis:
@mmeeks I've added the same for tile event slurping. I tried to use the same timer for both but it needs a periodic cleanup timer to do the left over tile events in the tile slurp queue. Without the periodic cleanup using setInterval, I see the final few tile updates not getting rendered (tile sitting in the queue till next message arrives :) ). This is very visible in multi user editing - when the other user is editing and stops typing, the first user do not see the last (few?) character(s) they typed until some message is sent to the first user.

mmeeks:
Hmm - I'm not a fan of the second commit here, the random 30ms delay is unhelpful I think =) - can we get just the first patch in - I'll sketch / implement what I was thinking about for merging the slurping - I think we can simplify the asynchronicity here for the core.

dennisfrancis:
@mmeeks I've removed the original second commit and added one for the scroll section fix. This avoids any scrolling till the end of msg slurping and it repaints the canvas only once.

mmeeks:
Looks lovely to me - thanks Dennis =)

PR #2376, opened 2021-05-14
Private/gokay/autofillmarker co 6 4

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,3 +142,4 @@ Makecache
 autogen.input
 loleaflet/src/layer/tile/TilesSection.js
 loleaflet/src/layer/tile/ScrollSection.js
+loleaflet/src/layer/tile/AutoFillMarkerSection.js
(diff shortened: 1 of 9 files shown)


mmeeks:
Would be great - currently as you move the cursor you can see it sort of 'jump' on top / is not in the same frame: more noticeable now cursor moving is faster too =)

dennisfrancis:
@gokaysatir Are there more patches to port regarding auto-fill marker or this the final set ?

gokaysatir:
@dennisfrancis  This is final set.

dennisfrancis (commented):
This patch is empty : "LOUtil: Add rectangle class". The patch "LOUtil: Rectangle class follow-up." is also empty Can you fix ?

dennisfrancis:
I see the same marker behaviour on moving the cursor in the svg version too. I'll fix it in a follow-up.

PR #2377, opened 2021-05-14
Render grid-lines in horizontal strips for performance.

mmeeks:
Interestingly for large CPU rendered skia surfaces - as used by
many browsers, rendering well spaced vertical lines is inordinately
expensive: 10x slower than horizontal ones, presumably due to cache
effects.

Grid rendering is around ~40% of our rendering performance measured
by knocking out various elements we render and re-timing.

By banding our vertical lines to re-use pixels close to where we
are drawing horizontal lines, and by creating more smaller paths
we get at ~25% overall performance improvement - 70ms -> 53ms
avg of ~10 frames.

Change-Id: I8ad1007b74866bd3d7b1b838f226a3efadcccbed

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


dennisfrancis (approved):
Looks good. Thanks!

PR #2388, opened 2021-05-17
Distro/collabora/co 6 4

gueniewie:
* Resolves: # 
* Target version: master 

### Summary
I hope I am in the correct repository ?
I Installed on a new test system (oracle 8.3) loolwsd and core on this system it is not possible to create a correct install of certs i have always permission denied when i crate the path (letsencrypt).

I have also a test system with a older installation from loolwsd and core on this system the installation from the certs was correct, also after updating the loolwsd and core it is working with no problem??

Why it is not possible to install the correct path from certs?

### TODO

- [x] ...

### Checklist


timar:
This pull request was made by mistake, probably.

PR #2405, opened 2021-05-19
loleaflet:added dropdown button in classic mode for fontcolor,backcolor,backgroundcolor

pedropintosilva:
Change-Id: Ieac36f003c5b1258415d0c449edf669c84d4020c


pedropintosilva:
this PR failed test is a bug:
in the `top_toolbar_spec.js` I experimented changing the viewport size of cypress (default is 1000 by 600) in the `cypress_test/cypress.json` by adding:
```javascript
    "viewportWidth": 1280,
    "viewportHeight": 660
```
and now the bullet tests pass! The culprit seems to be the `w2ui-scroll-right` that in smaller screens hides the bullet icon and thus . Also checking the initial error in the interactive mode gives that clue:

```javascript
Timed out retrying after 6050ms: cy.click() failed because this element:

<td id="tb_editbar_item_defaultbullet" style="" class="" valign="middle" title="Toggle Bulleted List">...</td>

is being covered by another element:

<div class="w2ui-scroll-right" onmousedown="event.preventDefault()" onclick="if($(this).closest('#toolbar-up').length>0 || $('#toolbar-down').width() > 768){var el=w2ui['editbar']; if (el) el.scroll('right');}" style="display: block;"></div>
```

_merging this_

PR #2407, opened 2021-05-19
Notebookbar: Add document icon in the case app icon is missing

pedropintosilva:
Add mime type icon under document-header according to document type

Change-Id: I42e80886919372a842a1d0fecf40b4a9557b5376


Andreas-Kainz:
Screenshot_20210521_105115


@pedropintosilva can you add the option to make the background color not white for notebookbar only.

PR #2421, opened 2021-05-20
Mobile: applystyle: Set different string for static label

pedropintosilva:
Reuse existent font to set a different label for applystyle command.
- Avoid duplicates (icon value-string      value-string  > )
- Fixes layout breakage (no space)
- Set it as "translatable" string
- Translation needs to be updated

Change-Id: I2d10c772a15fcb36a7169d77ff72cb5e5f6f7ed1


pedropintosilva:
**Problem:**

> mobilewizard-applystyle-string

**Fix:**

> mobilewizard-applystyle-string-fix

PR #2423, opened 2021-05-20
Enable building the iOS app completely on macOS

tml1024:
The intent is that it should no longer be necessary to do the
JavaScript part on a Linux machine. For the configury and "make" part,
mostly just some minor portability issues had to be handled:

The readlink command on macOS has no -f option. But that use of
readlink is not necessary anyway in a tree for building a mobile app,
so bypass.

The find command on macOS always requires at least one directory name.
So when $(CUSTOM_ICONS_DIRECTORY) is empty the "find
$(CUSTOM_ICONS_DIRECTORY) -name '*.*'" command produced an annoying
error message. Handle that in a trivial way.

The echo command in /bin/sh on macOS does not have the -n option. Use
the more portable printf command instead.

To then configure and run the make (before you can do the actual build
in Xcode9, you need have Node.js installed. Personally I just
downloaded the macOS Node.js binary tarball and unpacked into
/opt/node so that /opt/node/bin/node and /opt/node/bin/npm are the
relevant executables.

You will also need to add the lxml and polib Python modules to your
user Python module library with pip3, as in "pip3 install --user lxml"
and "pip3 install --user polib".

Change-Id: I263a318181125fa37ac174fa36cb0e44e56a6607

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,7 +75,7 @@ for MODULE in lxml polib; do
         AC_MSG_RESULT([yes])
     else
         AC_MSG_RESULT([no])
-        AS_IF([test `uname -s` = Linux -o `uname -s` = FreeBSD],
+        AS_IF([test `uname -s` = Linux -o `uname -s` = FreeBSD -o `uname -s`= Darwin],
               [AC_MSG_ERROR([${MODULE} for python3 is needed. It might be in a package called python3-${MODULE}.])],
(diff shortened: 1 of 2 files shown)


tml1024:
Need to update the README, too.

tml1024:
Somehow it is hard for me to believe that this PR could cause a Cypress test failure...

tml1024:
> Need to update the README, too.

Except that it isn't a README any more but a web page somewhere.

PR #2424, opened 2021-05-20
loleaflet: fix make clean issue

hcvcastro:
applying several times make clean
install the node_modules

Change-Id: I9004787ebf2087f541bec0248bd873252c88a40d

### DIFF
--- a/loleaflet/Makefile.am
+++ b/loleaflet/Makefile.am
@@ -1,5 +1,6 @@
 MAKEFLAGS = --no-builtin-rules
 CTAGS = ctags
+BUILT_SOURCES = node_modules
 
 L10N_PO = $(wildcard $(srcdir)/po/*.po)
 BUNDLE ?= $(if $(filter true,$(ENABLE_DEBUG)),DEBUG,RELEASE)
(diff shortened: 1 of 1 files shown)


loleaflet/Makefile.am
```diff
@@ -519,13 +519,12 @@ node_modules: package.json archived-packages
...
 
-Makefile: $(srcdir)/Makefile.in $(top_builddir)/config.status node_modules
-	@case '$?' in \
-	 *config.status*) \
-	  cd $(top_builddir) && $(MAKE) $(AM_MAKEFLAGS) am--refresh;; \
-	 *) \
-	  cd $(top_builddir) && $(SHELL) ./config.status $(subdir)/$@;; \
-	esac;
+define target_makefile
+Makefile: $$(srcdir)/Makefile.in $$(top_builddir)/config.status $(1)
```
Ashod, line 523: I have never seen Makefile target defined in Makefile.am. This is managed by automake and autoconf. 

At least some comment explaining what's going on is necessary here. Why are we defining Makefile rule? why can't we have node_modules as a dependency for the rules that depend on it? Why is node_modules an argument? And why do we explicitly invoke *top-level* make ?! (That looks so wrong--it's recursive!)

loleaflet/Makefile.am
```diff
@@ -519,13 +519,12 @@ node_modules: package.json archived-packages
...
+	  @cd $$(top_builddir) && $$(MAKE) $$(AM_MAKEFLAGS) am--refresh
+endef
+
+$(eval $(call target_makefile,$(if $(findstring clean,$(MAKECMDGOALS)),,node_modules)))
```
Ashod, line 527: Clearly your 'make' skills are beyond what most people get by. But explaining in a comment what we're doing is important, because other people are not expected to know our intentions from simply reading advanced code.

Ashod (commented):
Thanks Henry. I left a couple of comments. 

I tested it and it works for me, but I still think this is dangerous stuff, especially that most people (myself included) don't understand what's going and how all this is supposed to work. Of course it's not clear what the purpose is either! I don't understand what we're doing and I don't expect most other devs to have any better ideas.

PR #2445, opened 2021-05-24
fix: empty tiles on zoom back to previous level (co-6-4)

dennisfrancis:
Don't confuse between tile object and the tile image element.
Assign the tile object to tile array correctly and update the tile image
using the tileCache.

Change-Id: I014cc47c7b6e4086384a66158fb8eb2534714b99

* Resolves: # 
* Target version: co-6-4

### Checklist


kendy:
@dennisfrancis: Looks great to me, love the more explicit "tileImg" vs. "tile" (object) split too, thank you!

PR #2457, opened 2021-05-25
Re-work the profile zone implementation here, and other stuff piling up waiting for CI to pass

tml1024:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -41,7 +41,9 @@ namespace Log
     {
     private:
         Poco::Logger* _logger;
+        static thread_local Poco::Logger* _threadLocalLogger;
         std::string _name;
+        std::string _logLevel;
         std::string _id;
         std::atomic<bool> _inited;
     public:
@@ -65,9 +67,28 @@ namespace Log
 
         const std::string& getName() const { return _name; }
 
+        void setLevel(const std::string& logLevel) { _logLevel = logLevel; }
+
+        const std::string& getLevel() const { return _logLevel; }
+
         void setLogger(Poco::Logger* logger) { _logger = logger; };
+
+        void setThreadLocalLogger(Poco::Logger* logger)
+        {
+            // FIXME: What to do with the previous thread-local logger, if any? Will deleting it
+            // destroy also its channel? That won't be good as we use the same channel for all
+            // loggers. Best to just leak it?
+            _threadLocalLogger = logger;
+        }
+
         Poco::Logger* getLogger() const { return _logger; }
+
+        Poco::Logger* getThreadLocalLogger() const { return _threadLocalLogger; }
+
     } Static;
+
+    thread_local Poco::Logger* StaticHelper::_threadLocalLogger = nullptr;
+
     bool IsShutdown = false;
 
     // We need a signal safe means of writing messages
@@ -312,7 +333,9 @@ namespace Log
         auto& logger = Poco::Logger::create(Static.getName(), channel, Poco::Message::PRIO_TRACE);
         Static.setLogger(&logger);
 
-        logger.setLevel(logLevel.empty() ? std::string("trace") : logLevel);
+        const std::string level = logLevel.empty() ? std::string("trace") : logLevel;
+        logger.setLevel(level);
+        Static.setLevel(level);
 
         const std::time_t t = std::time(nullptr);
         oss.str("");
@@ -333,7 +356,11 @@ namespace Log
 
     Poco::Logger& logger()
     {
-        Poco::Logger* pLogger = Static.getLogger();
+        Poco::Logger* pLogger = Static.getThreadLocalLogger();
+        if (pLogger != nullptr)
+            return *pLogger;
+
+        pLogger = Static.getLogger();
         return pLogger ? *pLogger
                        : Poco::Logger::get(Static.getInited() ? Static.getName() : std::string());
     }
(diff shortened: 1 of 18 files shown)


common/ProfileZone.hpp
```diff
@@ -91,7 +95,7 @@ class ProfileZone
 
     ~ProfileZone()
     {
-        if (m_nCreateTime.time_since_epoch() > std::chrono::time_point<std::chrono::system_clock>().time_since_epoch())
```
mmeeks, line 94: Wow - we had a system-call for every profile zone destructor?
I really want the ProfileZone stuff to be almost a no-op / zero performance impact when it is not enabled. This is better - but - I'm amazed to see that thing[!] even as it disappears =)

kit/Kit.cpp
```diff
@@ -1799,6 +1803,21 @@ class Document final : public DocumentManagerInterface
...
 
+#if !defined FUZZER && !defined BUILDING_TESTS
+
+// When building the fuzzer we link LOOLWSD.cpp into the same executable so the
+// Protected::addOneRecording() there gets used. When building the unit tests the one in
+// ProfileZone.cpp gets used.
+
+void ProfileZone::addOneRecording(const std::string &sRecording)
+{
+    if (!config::getBool("trace_event[@enable]", false))
```
mmeeks, line 1814: I'd expect this to be heinously slow too - for every line we add as we trace; can we have a static / global variable that is set at load time (?)

kit/Kit.cpp
```diff
@@ -1799,6 +1803,21 @@ class Document final : public DocumentManagerInterface
...
+        return;
+
+    singletonDocument->sendTextFrame("trace: \n" + sRecording);
```
mmeeks, line 1817: Hmm - we really want to send these one by one? or does the recording contain lots of back to back events? We really want to reduce the performance impact and overhead of performance tracing as much as possible so it doesn't distort what we're trying to measure.

mmeeks (commented):
A few queries =)

common/ProfileZone.hpp
tml1024: Not sure I see any system call there. And anyway, "we had" is an odd expression to use for code that has been there for maybe a week, and has not been used at all.

kit/Kit.cpp
tml1024: Ask @Ashod , who explicitly said I should use that instead of a global variable.

kit/Kit.cpp
tml1024: Yeah, clearly we need to optimise this before even seeing whether it has any impact. Will get right on it.

kit/Kit.cpp
tml1024: You use a global variables. The expert say no, we have an API for that. You use that API. Another expert says no, use a global variable. Repeat.

tml1024: From Mattermost: "@tml: fyi, the kit now has full access to loolwsd.xml :slightly_smiling_face: so we don't need to pass any new configs to kit explicitly and we don't need any new global or other variables in kit... see ConfigUtil.{hpp,cpp}."

common/ProfileZone.hpp
mmeeks: You're quite right; I had assumed one of these would have done a gettimeofday or equivalent; but I'm wrong =)

kit/Kit.cpp
mmeeks: heh - so a locally scoped static bool = would make both happy I think.

wsd/Storage.cpp
```diff
@@ -651,6 +652,8 @@ std::unique_ptr<WopiStorage::WOPIFileInfo> WopiStorage::getWOPIFileInfo(const Au
                                                                         const std::string& cookies,
                                                                         LockContext& lockCtx)
 {
+    ProfileZone profileZone("getWOPIFileInfo", { {"url", _fileUrl} });
```
mmeeks, line 655: Lovely =) Ash'll need to convert it to an async pair as/when we get async save in I guess but that's fine.

kit/Kit.cpp
```diff
@@ -1809,13 +1814,46 @@ class Document final : public DocumentManagerInterface
...
+        flushProfileZoneRecordings();
+    else if (profileZoneRecordings.size() == 0 && profileZoneRecordings.capacity() < profileZoneRecordingsCapacity)
+        profileZoneRecordings.reserve(profileZoneRecordingsCapacity);
+    profileZoneRecordings.emplace_back(recording);
 }
+
+#else
+
+static void flushProfileZoneRecordings()
+{
```
mmeeks, line 1854: Looks great thanks for that; do we send the frame here?

common/Log.cpp
```diff
@@ -65,9 +67,28 @@ namespace Log
 
         const std::string& getName() const { return _name; }
 
+        void setLevel(const std::string& logLevel) { _logLevel = logLevel; }
+
+        const std::string& getLevel() const { return _logLevel; }
+
         void setLogger(Poco::Logger* logger) { _logger = logger; };
```
mmeeks, line 74: How do we avoid leaking that ? Presumably at the end of the DocumentBroker::poll - we could check if this is set, and unset / free / clean it up (?) I guess it's only for debugging but still =)

Anyhow - looks lovely; thanks Tor!

tml1024: There is a FIXME about that: 
 ```
            // FIXME: What to do with the previous thread-local logger, if any? Will deleting it
            // destroy also its channel? That won't be good as we use the same channel for all
            // loggers. Best to just leak it?
```

PR #2467, opened 2021-05-26
Make sure that double fits into uint64_t

vmiklos:
Change-Id: Ic521383f71c1865ae18fc880b79493c52285e29b


vmiklos:
writer/hamburger_menu_spec.js failed, let's try again.

tml1024:
Looks OK to me. So this fixes a problem that shows up when the fuzzer intentionally feeds garbage to this, not in actual use, right? (I don't mean that it shouldn't be fixed anyway.)

vmiklos:
> So this fixes a problem that shows up when the fuzzer intentionally feeds garbage to this, not in actual use, right?

Yes, this is about "don't crash on garbage input". I'll push it once CI passes.

vmiklos:
Now writer/shape_properties_spec.js failed. (Just logging these here to see if there is a pattern or not.)

PR #2471, opened 2021-05-26
Add UI blocking functions to UIManager

eszkadev:
when UI is blocked any interactions with sidebar/dialogs,
    jsdialogs or menu/toolbar items are blocked.
    
    to use: fire 'blockUI' / 'unblockUI' events
    or call directly: map.uiManager.blockUI() / .unblockUI()

### DIFF
--- a/loleaflet/src/control/Control.LokDialog.js
+++ b/loleaflet/src/control/Control.LokDialog.js
@@ -1193,6 +1193,9 @@ L.Control.LokDialog = L.Control.extend({
 		L.DomEvent.on(canvas, 'mousedown mouseup', function(e) {
 			L.DomEvent.stop(e);
 
+			if (this._map.uiManager.isUIBlocked())
+				return;
+
 			if ((this._isSidebar(id) || this.isCalcInputBar(id)) && this.hasOpenedDialog()) {
 				this.blinkOpenDialog();
 				return;
--- a/loleaflet/src/control/Control.UIManager.js
+++ b/loleaflet/src/control/Control.UIManager.js
@@ -88,7 +88,7 @@ L.Control.UIManager = L.Control.extend({
 						type: 'container',
 						vertical: 'true',
 						children: [
-							{id: 'busyspinner', type: 'spinner'},
+							{id: 'busyspinner', type: 'spinnerimg'},
 							{id: 'busylabel', type: 'fixedtext', text: label}
 						]
 					}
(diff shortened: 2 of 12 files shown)


loleaflet/src/control/Control.LokDialog.js
```diff
@@ -1193,6 +1193,9 @@ L.Control.LokDialog = L.Control.extend({
 		L.DomEvent.on(canvas, 'mousedown mouseup', function(e) {
 			L.DomEvent.stop(e);
 
+			if (this._map.uiManager.isUIBlocked())
```
Ashod, line 1196: Do we need this for the keydown/keyup events too, or do we already block typing into the document ?

Ashod (commented):
Thanks Szymon, just one question.

loleaflet/src/control/Control.LokDialog.js
eszkadev: good catch, I blocked only the UI (dialogs, sidebar, etc), not document editing. so we need this for typing. but probably also for mouse events in the document

Ashod: Yes, sounds good.

loleaflet/src/control/Control.UIManager.js
```diff
@@ -71,16 +75,53 @@ L.Control.UIManager = L.Control.extend({
...
+
 		this.map.on('showbusy', function(e) {
-			if (w2ui['actionbar'])
-				w2utils.lock(w2ui['actionbar'].box, e.label, true);
+			if (window.mode.isMobile()) {
+				if (w2ui['actionbar'])
+					w2utils.lock(w2ui['actionbar'].box, e.label, true);
+			} else {
+				closeBusyPopup();
+				openBusyPopup(e.label);
```
Ashod, line 115: I fear this will block the UI too frequently. If I remember correctly, showbusy is used during saving, which might be frequent. Best to check how often we use showbusy and avoid blocking if it's too frequent, instead, we can issue the blockui command from WSD. But if showbusy isn't frequent, this can be helpful.

Ashod (commented):
Thanks! Looking better. One comment about blocking in showbusy.

loleaflet/src/control/Control.UIManager.js
eszkadev: no, I'm not blocking the UI on showbusy.
I'm calling showbusy when blocking the UI :)
But used my new popup in showbusy so all will look unified

PR #2489, opened 2021-05-27
connect just once to the zoomend message in a more sensible place.

mmeeks:
Change-Id: I9ac799a0c9a0f77a61a85954d270af6e7ca28dac

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Makes perfect sense to me from the code reading perspective.

No idea though why it was done the way it was done here, so if we wanted to keep it there, it might be safer to just do an map.off() and then map.on() - but as said, no idea why it was done as it is, the code looks safe after the change too, the functionality is guarded at places.

PR #2490, opened 2021-05-27
UI blocking (master)

eszkadev:


loleaflet/src/core/Socket.js
```diff
@@ -1018,6 +1021,21 @@ app.definitions.Socket = L.Class.extend({
...
+			msg = null;
+
+			if (textMsg === 'rename') {
+				msg = _('The document is being renamed and will reload shortly');
+			}
+
+			this._map.fire('blockUI', {message: msg});
+			return;
+		}
+		else if (textMsg.startsWith('unblockui: ')) {
```
Ashod, line 1035: Since the message is optional (even for blockui, but certainly for unblockui, which will almost never have a message), we shouldn't expect the trailing space.

I suggest we do the startsWith for both 'blockui:' and 'unblockui:' *without* the trailing space.

loleaflet/src/core/Socket.js
```diff
@@ -1018,6 +1021,21 @@ app.definitions.Socket = L.Class.extend({
 			var mobileEvent = textMsg.substring('mobile: '.length);
 			this._map.fire(mobileEvent);
 		}
+		else if (textMsg.startsWith('blockui: ')) {
+			textMsg = textMsg.substring('blockui: '.length);
```
Ashod, line 1025: Trim the textMsg, in case it has some extra whitespace, is a good idea.

Ashod (commented):
Thanks Szymon, great stuff! There is a minor conflict and I left a couple of comments to improve the command handling from WSD. Otherwise, will continue testing and add notes here, if any.

loleaflet/src/core/Socket.js
eszkadev: done

loleaflet/src/core/Socket.js
eszkadev: done, fixed in both PR

Ashod:
@eszkadev, @pedropintosilva, Thank you.

eszkadev:
I need this changes in my current work. I will push it afer I solve conflicts

PR #2498, opened 2021-05-28
Mobile BottomBar Align

Andreas-Kainz:
- Mobile Icon closedocument fix
- Mobile BottomBar Align dropdown


Andreas-Kainz:
@eszkadev this patch didn't work cause the Align dropdown menu wasn't shown. Do you know what I miss to have the Align commands behind a dropdown? I take the code from classic toolbar.

eszkadev:
> @eszkadev this patch didn't work cause the Align dropdown menu wasn't shown. Do you know what I miss to have the Align commands behind a dropdown? I take the code from classic toolbar.

I think there is type: "menu" and you can define list of items in the dropdown.

https://github.com/CollaboraOnline/online/blob/master/loleaflet/src/control/Control.TopToolbar.js#L181

Andreas-Kainz:
At patch set 1 I use menu as type but it didn't work at bottomBar.

Andreas-Kainz:
> I think there is type: "menu" and you can define list of items in the dropdown.
> https://github.com/CollaboraOnline/online/blob/master/loleaflet/src/control/Control.TopToolbar.js#L181

I got it, many thanks

PR #2503, opened 2021-05-29
Mobile BottomBar fix height

Andreas-Kainz:
The height of the bottom bar items is 35px the height of BottomBar was defined with 34px so that it look always as the bottomBar was cutted.

Height is changed from 34px to 37px (check the blue select frame, it now wasn't cuted)

Change-Id: If30e56c582ca652cb35e73f669f440696c6b03b4


Andreas-Kainz:
@nikslor fix for BottomBar cut issue

PR #2506, opened 2021-05-30
Mobile BottomBar Calc Align Menu

Andreas-Kainz:
Move Align settings into a drop down menu instead of buttons, cause calc has 7 align options.

Change-Id: I1bbe1ed8b8ef7ada8e56486fef44b962e8dc50d7


cypress_test/integration_tests/mobile/calc/bottom_toolbar_spec.js
```diff
@@ -126,54 +126,6 @@ describe('Interact with bottom toolbar.', function() {
 			.should('have.attr', 'colspan', '1024');
 	});
 
-	it('Apply left/right alignment', function() {
```
vmiklos, line 129: Why did you remove this test? If it's by accident, could you please restore it? Thanks. CC @pedropintosilva

Andreas-Kainz: I will get some help to write tests, so hopefully I can define that at calc, impress/draw the align stuff will be available within a submenu. But first I have to learn how to write tests.
In addition I don't know what's the benefit of this tests.

PR #2517, opened 2021-05-31
Avoid higlighting additional rows/cols

eszkadev:
When used 120% zoom and selected any row after 50
there were multiple rows highlighted.

Change-Id: I4bc5f4272bf65fb0f9d00d96527890281377c0a8


gokaysatir (approved):
Thanks for the fix :) I tested.

PR #2523, opened 2021-05-31
Avoid maximum/minimum logging level terminology for newly introduced …

tml1024:
…settings

(Also in variable names.)

Better to just use terms that say what we mean, i.e. most verbose and
least verbose.

Change-Id: I80d74fda8b80bd34d194c3df97d246a41368189b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
I think the "Unit & Cypress Tests" thing is stuck, let's ignore it. This PR doesn't affect the UI anyway.

PR #2525, opened 2021-05-31
cypress:introducing notebookbar tests

Rash419:
Change-Id: I60bcadff6d4602a51a40d02d8cbe445b0a180a64

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


cypress_test/integration_tests/desktop/writer/top_toolbar_spec.js
```diff
@@ -22,8 +23,7 @@ describe('Top toolbar tests.', function() {
 	});
 
 	it('Apply highlight color.', function() {
-		cy.get('#tb_editbar_item_backcolor')
-			.click();
+		desktopHelper.actionOnSelector('backColor', (selector) => { cy.get(selector).click(); });
```
lpranam, line 26: now need to rename this file as now it doesn't contain only top toolbar tests

lpranam (changes requested):
please also update the readme, so we know how to run the notebookbar tests,

cypress_test/integration_tests/desktop/writer/top_toolbar_spec.js
Rash419: > now need to rename this file as now it doesn't contain only top toolbar tests

it contains all top toolbar functionality only

lpranam: but it also tests corresponding notebook bar functionality and top toolbar is classic mode specific.

Rash419:
> please also update the readme, so we know how to run the notebookbar tests,

done and I think we don't need to change the name of the spec.In both case upper portion is known as toolbar

PR #2530, opened 2021-06-01
Do not include js files in TS compilation

lpranam:
Change-Id: Ie2df89ba09597a34aca9221bf082b1f2df3c7d07

* Target version: master 

### Summary
problem:
in compilation tsc removes blank lines from the file
this causes problem in debugging and mapping with IDEs

https://github.com/microsoft/TypeScript/issues/843

### Checklist


dennisfrancis (approved):
Making tsc process/copy all js files to dist/ is no longer necessary (It was necessary before when the project compilation was turned on). Thanks for the fix.

PR #2532, opened 2021-06-02
Include the thread name in the Trace Events emitted from the Online C++ code…

tml1024:
…, too

Make sure to avoid most of the code path through createArgsString() if
Trace Event generation is not turned on.

Change-Id: Id86b2c9b4ded9d402647b7a12b9290d0122ba2aa

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/TraceEvent.cpp
+++ b/common/TraceEvent.cpp
@@ -20,27 +20,22 @@ thread_local int TraceEvent::threadLocalNesting = 0; // level of overlapped zone
 
 static std::mutex mutex;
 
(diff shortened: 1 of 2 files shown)


tml1024:
Note that after this, the thread name is still missing from the Trace Events coming in from core in the LOK_CALLBACK_PROFILE_FRAME callbacks (handled in GlobalCallback() in Kit.cpp). Do we really need to parse the JSON there and add an "args" property if not already there, and then a "thread" property to its value, and then stringily it again? Or do we need to tell core the thread's name so that core can add it itself to the generated Trace Events?

PR #2534, opened 2021-06-02
fix: no painting on switching to earlier slides (co-6-4)

dennisfrancis:
Before slurping events work, on switching parts the canvas was getting
painted as side effect of other events triggered from
Parts.js:setPart(). Let us enforce an explicit and possibly immediate
canvas update on part/slide switch in case we have all the necessary
tiles in the tile array or the cache in the doclayer method
_updateOnChangePart().

Change-Id: I776c1a2476adb1db2054eecdbba78feff290981a

* Resolves: # 
* Target version: co-6-4

### Checklist


mmeeks (approved):
Thanks Dennis ! =) would be good to have in 6.4.9 I think too.

PR #2537, opened 2021-06-02
Mobile Calc Property sidebar panel fix number format and cell appearance

Andreas-Kainz:
Icon size and alignment is now well at Number Format and Cell Appearance (Borders).
Same padding/margin for the icons than the ui-header-left

Change-Id: Ifff67d06074d4d39a4148237a3c3d40255c73392


Andreas-Kainz:
### Before
Screen Shot 2021-06-03 at 22 44 52 

Screen Shot 2021-06-03 at 22 44 58

### After
after

PR #2539, opened 2021-06-03
fix tile invalidation for tiles of other zooms

dennisfrancis:
tile-size is always 256 core pixels, so twips size of a tile in
different zooms are different depending on zoom scale. But the document
coordinates of the tiles in twips are zoom invariant.

Problem it resolves: Open any writer/impress document, switch to 85%, 120% zooms and then back to 100%.
Make a very noticeable change like color change. Now switch to 85% or 120% zoom, the change is not there completely or there are blank tiles.

Change-Id: I648134aaf84d451d93dad54d62b66062fb1d9204

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Looks great to me from the code reading perspective - please merge :-)

PR #2560, opened 2021-06-07
HTTP fixes based on fuzzing

Ashod:
- configure: allow for setting --with-lo-path with --enable-fuzzers
- wsd: http: performWrites can be called after destroying the socket
- wsd: better handling of invalid port in URL
- wsd: http: sync requests accept external poller
- wsd: http: an incomplete response is an error
- wsd: http: always destroy the socket last
- wsd: http: do not set the state explicitly on parsing error

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -857,7 +857,7 @@ AS_IF([test -n "$LOKIT_PATH"],
       [CPPFLAGS="$CPPFLAGS -I${LOKIT_PATH}"])
 lokit_msg="$LOKIT_PATH"
 
-AS_IF([test "$ENABLE_IOSAPP" != "true" -a "$ENABLE_ANDROIDAPP" != "true" -a "$enable_fuzzers" != "yes"],
+AS_IF([test "$ENABLE_IOSAPP" != "true" -a "$ENABLE_ANDROIDAPP" != "true"],
       [AC_MSG_CHECKING([for LibreOffice path])
       if test -n "$with_lo_path"; then
           # strip trailing '/' from LO_PATH, 'ln -s' with such path will otherwise fail
@@ -865,7 +865,8 @@ AS_IF([test "$ENABLE_IOSAPP" != "true" -a "$ENABLE_ANDROIDAPP" != "true" -a "$en
           AC_MSG_RESULT([found])
       else
           AC_MSG_RESULT([not found])
-          AC_MSG_ERROR([LibreOffice path must be configured: --with-lo-path])
+          AS_IF([test "$enable_fuzzers" != "yes"],
+                [AC_MSG_ERROR([LibreOffice path must be configured: --with-lo-path])])
       fi
       ])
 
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -857,15 +857,16 @@ AS_IF([test -n "$LOKIT_PATH"],
       [CPPFLAGS="$CPPFLAGS -I${LOKIT_PATH}"])
 lokit_msg="$LOKIT_PATH"
 
-AS_IF([test "$ENABLE_IOSAPP" != "true" -a "$ENABLE_ANDROIDAPP" != "true" -a "$enable_fuzzers" != "yes"],
+AS_IF([test "$ENABLE_IOSAPP" != "true" -a "$ENABLE_ANDROIDAPP" != "true"],
```
vmiklos, line 860: I didn't need this as I always configure a git checkout for fuzzing or not and never switch between the two, but why not.

net/HttpRequest.hpp
```diff
@@ -831,7 +839,7 @@ class Response final
 
     StatusLine _statusLine;
     Header _header;
-    State _state; //< The state of the Response.
+    std::atomic<State> _state; //< The state of the Response.
```
vmiklos, line 842: Is this accessed from multiple threads? I had the impression that most of your HTTP work is about having an async HTTP client which uses a single thread.

vmiklos (approved):
I don't see anything obviously problematic with this, though don't trust me too much regarding the socket code, I didn't follow its development closely.

Re fuzzing: how did you discover the problems? Jenkins runs the 3 fuzzers we have in this repo, and they didn't find anything http-related recently, so I assume this is something else. My ideal is that whenever you fix something based on fuzzing, the fuzzer is added to fuzzer/ and the input that triggered the problem is added to fuzzer/*data/.

configure.ac
Ashod: Right, switching is how I hit it and thought it's not a bad idea to be permissive.

Ashod:
> Re fuzzing: how did you discover the problems? Jenkins runs the 3 fuzzers we have in this repo, and they didn't find anything http-related recently, so I assume this is something else. My ideal is that whenever you fix something based on fuzzing, the fuzzer is added to fuzzer/ and the input that triggered the problem is added to fuzzer/*data/.

I have a new fuzzer. Will submit a PR separately. The reason it's not here is to keep the size of the PR small and because I had to remove the asan library, which new versions of clang don't have/need/like.

net/HttpRequest.hpp
Ashod: It's true that async can still work on the same thread, but in practice the poll thread can be a different one, so it's best to make the state atomic. We typically handle the async result on the same poll thread, though.

vmiklos:
> I have a new fuzzer. Will submit a PR separately. The reason it's not here is to keep the size of the PR small

Ah, I see. :-)

PR #2588, opened 2021-06-10
Log also the average tile PNG size and the average compression ratio

tml1024:
Change-Id: I77fb1ae0e9ef892b23212222b44dc3dc522c7ce3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Png.hpp
+++ b/common/Png.hpp
@@ -51,6 +51,7 @@
 
 #include <cassert>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


tml1024:
Cypress, bah. I just can't believe this change would cause a Cypress failure. But I am open to convincing other arguments. Still, I am pushing the button now.

PR #2591, opened 2021-06-10
Transfer LOK callback processing to our main thread if necessary.

mmeeks:
While we are processing data in the poll handler, we can have
another helpful thread from the core sending things to us:

loolforkit(Socket::assertCorrectThread(char const*, int))
loolforkit(WebSocketHandler::sendFrame(std::shared_ptr<StreamSocket> const&, char const*, unsigned long, unsigned char, bool) const)
loolforkit(WebSocketHandler::sendMessage(char const*, unsigned long, WSOpCode, bool) const)
loolforkit(Document::postMessage(char const*, int, WSOpCode) const)
loolforkit(Document::sendFrame(char const*, int, WSOpCode))
loolforkit(ChildSession::sendTextFrame(char const*, int))
loolforkit(ChildSession::loKitCallback(int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&))
loolforkit(Document::GlobalCallback(int, char const*, void*))
/core/instdir/program/libsofficeapp.so()
/core/instdir/program/../program/libfwklo.so()
/core/instdir/program/libscfiltlo.so()
/core/instdir/program/libexpwraplo.so()
/core/instdir/program/libooxlo.so(oox::core::FastParser::parseStream(com::sun::star::xml::sax::InputSource const&, bool))
/core/instdir/program/libooxlo.so(oox::core::FastParser::parseStream(com::sun::star::uno::Reference<com::sun::star::io::XInputStream> const&, rtl::OUString const&))
/core/instdir/program/libooxlo.so(oox::core::XmlFilterBase::importFragment(rtl::Reference<oox::core::FragmentHandler> const&, oox::core::FastParser&))
/core/instdir/program/libscfiltlo.so()
/core/instdir/program/libcomphelper.so(comphelper::ThreadTask::exec())
/core/instdir/program/libcomphelper.so()
/core/instdir/program/libuno_salhelpergcc3.so.3(salhelper::Thread::run())
/core/instdir/program/libuno_salhelpergcc3.so.3()
/core/instdir/program/libuno_sal.so.3()
/lib/x86_64-linux-gnu/libpthread.so.0()
/lib/x86_64-linux-gnu/libc.so.6(clone)

This is most likely from eg. threaded parsing of various file formats,
or progress messages while loading / saving, sent while the SolarMutex
is locked.

We assume that:
  + there will only ever be one thread running inside kitPoll
    this is enforced by vcl/headless.
  + and so, we can safely mutate SocketPoll and Socket buffer
    state from this thread.
  + there are only two entry points for callbacks from threads
    in core - which are both instrumented, and if necessary
    their work is shifted to this thread.

Memory corruptions matching the above trace, suggesting that
multiple writes have collided between kitPoll and an event
callback have been seen in the wild.

Change-Id: I5b084cbfec1ea326b6e17c9e5590a8c8e35b3854

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -149,6 +149,11 @@ static void flushTraceEventRecordings();
 // socket buffer & event processing in a single, thread.
 bool pushToMainThread(LibreOfficeKitCallback cb, int type, const char *p, void *data);
 
+// Abnormally we get LOK events from another thread, which must be
+// push safely into our main poll loop to process to keep all
+// socket buffer & event processing in a single, thread.
+bool pushToMainThread(LibreOfficeKitCallback cb, int type, const char *p, void *data);
+
 #if !MOBILEAPP
 
 static LokHookFunction2* initFunction = nullptr;


mmeeks:
Tor - would love some thoughts around the iOS usage here; particularly multiple documents.

kit/Kit.cpp
```diff
@@ -1994,6 +2005,24 @@ class KitSocketPoll final : public SocketPoll
...
+    {
+        if (mainPoll && mainPoll->getThreadOwner() != std::this_thread::get_id())
+        {
+            LOG_TRC("Unusual push callback to main thread");
+            std::shared_ptr<std::string> pCopy;
+            if (p)
+                pCopy = std::make_shared<std::string>(p, strlen(p));
+            mainPoll->addCallback([=]{
+                LOG_TRC("Unusual process callback in main thread");
+                callback(type, pCopy ? pCopy->c_str() : nullptr, data);
```
Ashod, line 2019: Is `data` guaranteed to live beyond the callback? Could the thread end-up with a dangling pointer?

Ashod (commented):
Thanks! Looks nice and simple. One question about the lifetime of `data` in the callback.

kit/Kit.cpp
vmiklos: Good catch, at least core.git hands out a data pointer which is only valid for the duration of the callback, and then core frees it. I'm not sure if there is some layer between this function and core that would perhaps avoid this problem in practice.

mmeeks: Well the data is really a closure that we provided to them in the first instance =) I checked the ViewCallback which has a CallbackDescriptor pointer - which is eternal and I updated the comment there to help review ;-) For the GlobalCallback the *data is the Document and our lifecycle there is basically being killed by a remote process for fast cleanup, or calling _exit() similarly, so I think we're good here.

tml1024:
Building the iOS app now for the first time in several weeks to see I I can notice any problem when having multiple documents open.

kit/Kit.cpp
vmiklos: Ah I see, then this is not a problem.

tml1024:
Fixed a couple of other bit-rot issues in the iOS app first. After that, could not notice any new problems with this PR included, when I had two documents open at the same time.

kit/Kit.cpp
```diff
@@ -821,9 +826,12 @@ class Document final : public DocumentManagerInterface
     static void GlobalCallback(const int type, const char* p, void* data)
     {
         if (SigUtil::getTerminationFlag())
-        {
             return;
-        }
```
tml1024, line 826: Just nit-picking, but why such cosmetic changes in a functional PR?

mmeeks: to make it more consistent =) [and sensible]

mmeeks:
CI a mess it seems; passes fine on co-6-4 and co-4-2, lets just merge it.

PR #2598, opened 2021-06-11
Guard against REMOVE messages for data: URLs for tunnelled dialogs in…

tml1024:
… iOS app

Trying to remove such a "file" caused a crash.

Sure, loleaflet should ideally not send such REMOVE messages at all.

No idea why the image for a tunnelled dialog shows up as a data: URL
and not as a file: URL pointing to a BMP file, like the document
tiles.

Change-Id: I4866ae2c94c626f1947a14353803a9c3d75d5ffa

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Ignoring Cypress test failure for a PR that touches only iOS-only code.

PR #2599, opened 2021-06-11
Fix crash in iOS app

tml1024:
No separate Kit process in the mobile apps, and the iOS app even can
have several documents open simultaneously, so the singletonDocument
variable does not make sense.

If we at some stage want to generate Trace Event log files also in the
iOS or Android app (perhaps not likely), will have to handle it in a
different fashion. All the code runs in one process anyway in the
mobile apps, so it wouldn't make sense to "send" collected Trace
Events to the code standing in for the WSD process.

Change-Id: I5777b48e98121c3ed55e590d5009e2658af2ef22

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -114,9 +114,11 @@ using namespace LOOLProtocol;
 
 extern "C" { void dump_kit_state(void); /* easy for gdb */ }
 
+#if !MOBILEAPP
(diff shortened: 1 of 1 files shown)


tml1024:
No point in waiting for the Cypress test here.

PR #2605, opened 2021-06-11
IE11 compatibility changes.

gokaysatir:
Change-Id: If595e742ecef633741892c4a714f74d93e97c6a9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/marker/Cursor.ts
+++ b/loleaflet/src/layer/marker/Cursor.ts
@@ -56,7 +56,11 @@ class Cursor {
 		this.map.getCursorOverlayContainer().appendChild(this.container);
 		this.visible = true;
 		this.domAttached = true;
-		this.onFocusBlur(new FocusEvent('focus'));
+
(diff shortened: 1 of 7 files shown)


loleaflet/src/layer/marker/Cursor.ts
```diff
@@ -56,7 +56,11 @@ class Cursor {
 		this.map.getCursorOverlayContainer().appendChild(this.container);
 		this.visible = true;
 		this.domAttached = true;
-		this.onFocusBlur(new FocusEvent('focus'));
+
+		var isIE11 = !!window.MSInputMethodContext && !!(<any>document).documentMode;
+		if (!isIE11)
+			this.onFocusBlur(new FocusEvent('focus'));
```
eszkadev, line 62: so in IE11 we just don't support hiding cursor when window is inactive?

gokaysatir: Yes it seems, other than that, everything seems fine.

eszkadev (approved):

PR #2614, opened 2021-06-14
android: Implemented conversion to odf  format for plain/text types

merttumer:
In addition, refactored the restating the app bits for copy requests
Now, we dont need to reload the activity for the new document
instead LO will switch to the new.

Signed-off-by: merttumer <mert.tumer@collabora.com>
Change-Id: I0319086ece0b7dcd0bbed865ac2a9681c7a1edb5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
Looks great to me from the code point of view, please go ahead & merge!

PR #2623, opened 2021-06-15
Bail out of getPixmap gracefully, when rendering failed.

mikekaganski:
And only insert the new pixmap after the check.

Change-Id: I310d321443152967b3aed8992ece3b55287d01cc


thebearon (approved):
Works nicely, thanks for the fix, Mike!

PR #2624, opened 2021-06-15
private/pedro/busypopup overlay

pedropintosilva:
- Busypopup: use less padding and fix text capitalization
- Busypopup: Create overlay when in presence of modal dialog

### DIFF
--- a/loleaflet/src/control/Control.JSDialog.js
+++ b/loleaflet/src/control/Control.JSDialog.js
@@ -28,6 +28,7 @@ L.Control.JSDialog = L.Control.extend({
 
 	closeDialog: function(id) {
 		var builder = this.dialogs[id].builder;
+
 		L.DomUtil.remove(this.dialogs[id].container);
 		delete this.dialogs[id];
 		builder.callback('dialog', 'close', {id: '__DIALOG__'}, null, builder);
--- a/loleaflet/src/control/Control.UIManager.js
+++ b/loleaflet/src/control/Control.UIManager.js
@@ -85,6 +85,10 @@ L.Control.UIManager = L.Control.extend({
 				type: 'modalpopup',
 				children: [
 					{
+						id: 'jsdialog-overlay',
+						type: 'overlay'
+					},
+					{
 						id: 'busycontainer',
 						type: 'container',
 						vertical: 'true',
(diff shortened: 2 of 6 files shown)


loleaflet/src/control/Control.JSDialog.js
```diff
@@ -28,6 +28,10 @@ L.Control.JSDialog = L.Control.extend({
 
 	closeDialog: function(id) {
 		var builder = this.dialogs[id].builder;
+
+		if (this.dialogs[id].type === 'modalpopup')
```
pedropintosilva, line 32: This condition is never met and I wonder if this is the right place or if there is additional stuff to be add in `closeBusyPopup()` or even in `hideBusy()`

pedropintosilva (commented):
^

loleaflet/src/control/Control.JSDialog.js
```diff
@@ -28,6 +28,10 @@ L.Control.JSDialog = L.Control.extend({
...
+			L.DomUtil.remove('jsdialog-overlay');
```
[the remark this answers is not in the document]
eszkadev: no, it's ok

`('body').find` doesn't look like valid code

loleaflet/src/control/Control.JSDialog.js
eszkadev: see line 118 (https://github.com/CollaboraOnline/online/pull/2624/files#diff-49fe2ec29af28fec557a60dbea747d3af4be9aa24d352ddbec4945e57098ba11L118)

in this.dialogs[id] there is no field type

loleaflet/src/control/Control.JSDialog.js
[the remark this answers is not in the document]
eszkadev: better would be to add for `this.dialog[id]` an `overlay`  member and remove it here by direct passing the reference

loleaflet/src/control/Control.UIManager.js
```diff
@@ -84,6 +84,10 @@ L.Control.UIManager = L.Control.extend({
 				jsontype: 'dialog',
 				type: 'modalpopup',
 				children: [
+					{
+						id: 'jsdialog-overlay',
+						type: 'container'
```
eszkadev, line 89: it shouldn't be done here, in the future I plan to reuse it, it would be not good if you need to remember that you need to add this overlay

eszkadev: could you create new control in JSDialogBuilder which creates div with CLASS jsdialog-overlay?

pedropintosilva: understood, but isn't this type `modalpopup` supposed to be for those situations where there is really just one dialog and its overlay?

pedropintosilva: ok

PR #2636, opened 2021-06-17
Fix Calc tab positions, sidebar width and Status bar

pedropintosilva:
- Calc: Fixes Tab list that was out of place and not visible
- Calc: Fixes inconsistency with background colours
- Calc: Fixes selected tab status by improving styling
  - Also use CSS vars instead
- Fixes overall (for all applications) toolbar-down (status bar)
  - Border was being overlapped and thus not being displayed

Change-Id: I766ca3ee3d82c89343a318d3adcf2d2e3a7dcf46

### DIFF
--- a/loleaflet/css/loleaflet.css
+++ b/loleaflet/css/loleaflet.css
@@ -34,7 +34,6 @@
 	width: 1px;
 }
 #document-container {
-	border-top: 1px solid var(--gray-color);
 	background: #DFDFDF;
 	position: relative;
 	top: 0;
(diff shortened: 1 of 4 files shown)


pedropintosilva:
I'm merging this. The failed test is unrelated to this change and to calc. It's from writer and specifically annotations : `spec=writer/annotation_spec.js`

I have ran the same CI test (interactively with `make run-desktop spec=writer/annotation_spec.js`) multiple times sometimes and it passes sometimes it doesn't but the culprit seems to be the code itself and not the Cypress tests (since I saw with my own eyes the annotation not being there)

PR #2637, opened 2021-06-18
Mobile: menuwizard (hamburger): Fixes contents position

pedropintosilva:
- Add mobile class to #toolbar-wrapper to avoid targeting anything else
  - Sets Menuwizard so it goes above toolbar (position absolute) so to be
  properly aligned from the top
  - Sets toolbarwrapper with normal z-index (according to its parent) so
  menuwizard can be above it

Note, no need to touch #toolbar-hamburger.menuwizard-opened has it comes
already with z-index properly set

Behavior might be related with changes from
ff20ec7

Change-Id: I1a66057b7827309a6ad3871927099ac1c9339657

### DIFF
--- a/loleaflet/css/device-mobile.css
+++ b/loleaflet/css/device-mobile.css
@@ -695,7 +695,7 @@ td[id^='tb_spreadsheet-toolbar_item']:focus table.w2ui-button div.w2ui-icon, td[
 }
 
 #mobile-wizard.menuwizard {
-	position: relative;
+	position: absolute;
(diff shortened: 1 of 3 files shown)


pedropintosilva:
Fixes this: 
mobile-hamburger-bug

pedropintosilva (commented):
Did some digging there is no reason to start to use relative position now. Also made sure that `#toolbar-wrapper` is being targetted properly within `toolbar-mobile.css`

PR #2641, opened 2021-06-18
HTTP 302 Found - redirrect support for getWOPIFileInfo

eszkadev:
With some loadbalancers it may happen that HTTP 302 Found
response with redirect location will appear.

### DIFF
--- a/test/UnitWOPIHttpRedirect.cpp
+++ b/test/UnitWOPIHttpRedirect.cpp
@@ -22,6 +22,7 @@ class UnitWopiHttpRedirect : public WopiTestServer
         Load,
         Redirected,
         GetFile,
+        Redirected2,
         Loaded
     } _phase;
 
@@ -49,6 +50,8 @@ public:
         std::string redirectUri = "/wopi/files/0";
         Poco::RegularExpression regRedirected(redirectUri);
         Poco::RegularExpression regContents("/wopi/files/0/contents");
+        std::string redirectUri2 = "/wopi/files/2/contents";
+        Poco::RegularExpression regContentsRedirected(redirectUri2);
 
         LOG_INF("Fake wopi host request URI [" << uriReq.toString() << "]:\n");
 
@@ -117,7 +120,7 @@ public:
 
             return true;
         }
-        // GetFile - for redirected URI
+        // GetFile - first try
         else if (request.getMethod() == "GET" && regContents.match(uriReq.getPath()))
         {
             LOG_TST("Fake wopi host request, handling GetFile: " << uriReq.getPath());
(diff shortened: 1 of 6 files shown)


test/UnitWOPIHttpRedirect.cpp
```diff
@@ -0,0 +1,151 @@
...
+    {
+        constexpr char testName[] = "UnitWopiHttpRedirect";
+
+        switch (_phase)
+        {
+            case Phase::Load:
+            {
+                initWebsocket("/wopi/files/1?" + params);
+
+                helpers::sendTextFrame(*getWs()->getLOOLWebSocket(), "load url=" + getWopiSrc(),
```
Ashod, line 133: Consider using WSD_CMD. See UnitWOPI and other WOPI tests (though not all use it) for WSD_CMD and other improvements that make the tests more readable and more informative in case of errors (including the phase transition logs etc.)

wsd/Storage.cpp
```diff
@@ -635,6 +635,21 @@ std::unique_ptr<WopiStorage::WOPIFileInfo> WopiStorage::getWOPIFileInfo(const Au
...
+        if (response.getStatus() == Poco::Net::HTTPResponse::HTTP_FOUND ||
+            response.getStatus() == Poco::Net::HTTPResponse::HTTP_MOVED_PERMANENTLY)
+        {
+            const std::string& location = response.get("Location");
+            Log::StreamLogger logRedirect = Log::trace();
+            if (logRedirect.enabled())
+            {
+                logRedirect << "WOPI::CheckFileInfo redirect to URI [" << LOOLWSD::anonymizeUrl(location) << "]:\n";
+                LOG_END(logRedirect, true);
+            }
```
Ashod, line 647: Why not LOG_TRC ?

test/UnitWOPIHttpRedirect.cpp
```diff
@@ -0,0 +1,151 @@
...
+            const std::string mimeType = "application/json; charset=utf-8";
+
+            std::ostringstream oss;
+            oss << "HTTP/1.1 200 OK\r\n"
+                "Last-Modified: " << Util::getHttpTime(_fileLastModifiedTime) << "\r\n"
+                "User-Agent: " WOPI_AGENT_STRING "\r\n"
+                "Content-Length: " << responseString.size() << "\r\n"
+                "Content-Type: " << mimeType << "\r\n"
+                "\r\n"
+                << responseString;
```
Ashod, line 112: We will convert these to http::Response at some point. Feel free to use it here (grep for samples).

test/UnitWOPIHttpRedirect.cpp
```diff
@@ -0,0 +1,151 @@
...
+            for (const auto& pair : request)
+            {
+                logger << '\t' << pair.first << ": " << pair.second << " / ";
+            }
+
+            LOG_END(logger, true);
+        }
+
+        // CheckFileInfo - returns redirect response
+        if (request.getMethod() == "GET" && regInfo.match(uriReq.getPath()))
```
Ashod, line 62: I'd strongly recommend using a state machine to track the phase we are in. For example, this conditional should be hit once and it should be hit first. Once hit, the phase changes and we assert that it's not called in the wrong phase. Similarly, the `if` for the new URL, it must be hit after this one, and not more than once. See UnitWOPIFileURL for an example.

Ashod (commented):
Good stuff. Thanks Szymon! Left some notes.

test/UnitWOPIHttpRedirect.cpp
eszkadev: I don't see any 'http::Response', does it exist even in co-6-4-9? (before async save)

PR #2643, opened 2021-06-20
leaflet:NB:added hyperlink in copy,paste and cut button

Rash419:
Change-Id: Ia2aa29bf064fb2bb518b80a6d5fa5617a1334f2c

* Target version: master 

### Summary

As we discovered when we implemented this for the menu bar, it is really important in some browsers - particularly Safari, but others - to have the right security context for 'paste' to be able to work when you click it.

In order to do that we need to have a hyperlink to '#' for the buttons in the notebookbar - at least, any of those that want to be able to cut/copy but particularly paste.

### DIFF
--- a/loleaflet/src/control/Control.JSDialogBuilder.js
+++ b/loleaflet/src/control/Control.JSDialogBuilder.js
@@ -2449,7 +2449,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		var controls = {};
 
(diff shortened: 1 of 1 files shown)


lpranam (approved):
code wise looks good to me. but Does it solve the original problem. I do not have details about original problem

Rash419:
I actually also shared the phab ticket in mattermost about it

PR #2674, opened 2021-06-24
Separate SSL Context for Client and Server Sockets

Ashod:
- wsd: separate the SSL locks from the context
- wsd: separate client SSL context from the server

### DIFF
--- a/net/Ssl.hpp
+++ b/net/Ssl.hpp
@@ -27,33 +27,15 @@
 class SslContext
 {
 public:
-    static void initialize(const std::string& certFilePath, const std::string& keyFilePath,
-                           const std::string& caFilePath,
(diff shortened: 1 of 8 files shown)


net/Ssl.hpp
```diff
@@ -68,9 +50,61 @@ class SslContext
...
-
     SSL_CTX* _ctx;
 };
 
+namespace ssl
+{
+class Manager
+{
+public:
+    static void initializeServerContext(const std::string& certFilePath,
```
mmeeks, line 61: Would have done this with an enum, and an array rather than two sets of methods myself but ... ;-) fine for now.

mmeeks (approved):
Like the lock cleanup, lets get this in I have a few concerns I'll mail you about.

PR #2676, opened 2021-06-24
Don't hide download progress when async download has started

mikekaganski:
Also make sure to not show the "Confirm copy to clipboard" in the end
of the download which is not intended for the local clipboard, but is
immediately uploaded to the target document

Change-Id: I09f27a234cc92fd0700bf29654f44cbaa158de47


mmeeks (approved):
Looks really good Mike - thanks ! =)

PR #2677, opened 2021-06-24
private/pedro/backporting welcome dialog missing btn

pedropintosilva:
- fix:added dismiss button in welcome page dialog
- Welcome dialog: do not mix desktop and mobile plus styling


pedropintosilva:
Again unrelated tests and gain related to annotations

PR #2679, opened 2021-06-24
Backporting tests that includes notebookbar 

Rash419:
* Resolves: # 
* Target version: co-6-4


kendy:
@Rash419: Thanks for this, let's get it in :-)

PR #2689, opened 2021-06-25
Mobile: Hamburger menu: latestupdates: Fix missing icon

pedropintosilva:
Fixes the used id by removing `-` (to be the same as used by
`_onlineHelpControl` control in the desktop:Notebookbar).
In return the id is then used to call the correct svg file name.

Change-Id: I63951d20b204b61e64a6c0d7e394465e230b7377


pedropintosilva:
it's falling here within a multi-user unrelated test but it passes locally. Note: The same changes also passed via CI against master

PR #2698, opened 2021-06-26
Revert "wsd: initialize SSL if enabled or termination is enabled"

Ashod:
The original behavior must be preserved, which is not to
initialize SSL when termination is enabled but SSL is disabled.

This reverts commit 1c9541a6a636647086397e1310084cc990b6ed70.


mmeeks (approved):
Looks sensible to me; thanks Ash =)

PR #2707, opened 2021-06-28
Add argument information to the ProfileZone for encodeSubBufferToPNG

tml1024:
Change-Id: I688958fe3b1329e410a96aeda573dd23cee5784d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
I think this can be dropped, not that important, and possibly a bad idea anyway to have such a different way to use ProfileZone.

PR #2709, opened 2021-06-28
Fix async copy to clipboard

mikekaganski:
... after 1aea12bed0313e4e54f82646c8a77556a68c2407, where I blindly
assumed that the _doAsyncDownload is only called from one file. The
affected code path was in loleaflet/src/control/Control.DownloadProgress.js,
and one place in loleaflet/src/map/Clipboard.js was wrong marking
a "POST" operation as "for clipboard".

Change-Id: I5bc1e68603ab8281716bcec6f8586524b6e01c95


mmeeks (approved):
Heh - I totally agree that the coupling to what looks like a generic 'DownloadProgress' control to explicitly call back into the Clipboard code is -extremely- no-obvious. That's something we could fix on master of online if this was not so fragile =)

PR #2710, opened 2021-06-28
HTTP 302 - updates after review

eszkadev:
Change-Id: Iaac4384385e05f33472b6678422b4a444319b59c

### DIFF
--- a/test/UnitWOPIHttpRedirectLoop.cpp
+++ b/test/UnitWOPIHttpRedirectLoop.cpp
@@ -42,7 +42,7 @@ public:
         Poco::URI uriReq(request.getURI());
         Poco::RegularExpression regInfo("/wopi/files/[0-9]+");
         std::string redirectUri = "/wopi/files/";
-        static unsigned fileId = 1;
+        static unsigned redirectionCount = 0;
 
         LOG_INF("Fake wopi host request URI [" << uriReq.toString() << "]:\n");
 
@@ -54,23 +54,23 @@ public:
             assertCheckFileInfoRequest(request);
 
             std::string sExpectedMessage = "It is expected to stop requesting after " + std::to_string(RedirectionLimit) + " redirections";
-            LOK_ASSERT_MESSAGE(sExpectedMessage, fileId <= RedirectionLimit + 1);
+            LOK_ASSERT_MESSAGE(sExpectedMessage, redirectionCount <= RedirectionLimit);
 
             LOK_ASSERT_MESSAGE("Expected to be in Phase::Load or Phase::Redirected", _phase == Phase::Load || _phase == Phase::Redirected);
             _phase = Phase::Redirected;
 
-            if (fileId == RedirectionLimit + 1)
+            if (redirectionCount == RedirectionLimit)
                 _phase = Phase::Polling;
 
             std::ostringstream oss;
             oss << "HTTP/1.1 302 Found\r\n"
-                "Location: " << helpers::getTestServerURI() << redirectUri << fileId << "?" << params << "\r\n"
+                "Location: " << helpers::getTestServerURI() << redirectUri << redirectionCount << "?" << params << "\r\n"
                 "\r\n";
 
             socket->send(oss.str());
             socket->shutdown();
 
-            fileId++;
+            redirectionCount++;
 
             return true;
         }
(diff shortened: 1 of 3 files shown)


test/UnitWOPIHttpRedirectLoop.cpp
```diff
@@ -52,22 +53,25 @@ class UnitWopiHttpRedirectLoop : public WopiTestServer
 
             assertCheckFileInfoRequest(request);
 
-            LOK_ASSERT_MESSAGE("It is expected to stop requesting after 21 redirections", fileId <= 22);
+            std::string sExpectedMessage = "It is expected to stop requesting after " + std::to_string(RedirectionLimit) + " redirections";
+            LOK_ASSERT_MESSAGE(sExpectedMessage, fileId <= RedirectionLimit + 1);
```
Ashod, line 57: Shouldn't we expect that the **last** retry be RedirectionLimit, and no redirections after that? Meaning, should we check that `fileId <= RedirectionLimit` ?

wsd/Storage.cpp
```diff
@@ -697,7 +697,7 @@ std::unique_ptr<WopiStorage::WOPIFileInfo> WopiStorage::getWOPIFileInfoForUri(Po
             }
             else
             {
-                LOG_TRC("WOPI::CheckFileInfo redirected too many times\n");
+                LOG_WRN("WOPI::CheckFileInfo redirected too many times - URI [" << uriAnonym << "]\n");
```
Ashod, line 700: Each log is on its own line. No need to have `\n`.

Ashod (approved):
Thanks Szymon for the follow up! One minor question, but looks good otherwise.

test/UnitWOPIHttpRedirectLoop.cpp
eszkadev: we start with `fileId = 1;` (see line 44, above) not from 0 which is the "initial URL". So `RedirectionLimit + 1` is good.

/wopi/files/0    fileId = 1     redirectionLimit=21
/wopi/files/1    fileId = 2     redirectionLimit=20
...
/wopi/files/20  fileId = 21   redirectionLimit=1
/wopi/files/21  fileId = 22   redirectionLimit=0
/wopi/files/22 is not requested

I used fileId because it's easier than extracting X from /wopi/files/X

wsd/Storage.cpp
eszkadev: done

test/UnitWOPIHttpRedirectLoop.cpp
Ashod: Right. That's the issue (granted, it's a minor one). You see, the limit is called 'Redirection Limit', but it's not representing the number of redirections. It makes more sense to start from 0 (meaning, we didn't do any redirections). Then, we ++ and it becomes 1 *after* we get the first request and reply back with the *first* redirection.

Yes, this is the server side, and we are counting the number of times the server returned 302. We are *not* counting the number of times the client requested the same URL (of course there is a relationship between the request and the response, but we are counting the response and not the requests).

When we do the assertion, at that point, we can compare the number of requests, or the number of redirections, or any combination we want. But the numbers would represent exactly what they are counting (the number of times we reply).

The reason I'm mentioning all this is because when there is a +1 when comparing with a counter, it typically implies that it has the wrong name (in this case, it seems we are really counting the requests, but comparing it with the redirections, which is unnecessary and confusing). This means that it will be confusing to someone else (even if it works correctly). 

Since we are doing both the count and the assertion in the same component (the server, when handling the request), I expect us to count the number of times we did reply with 302 (i.e. did redirection) and refuse to do any more when we exceed the limit. That is, we start with 0 (no redirections done), compare `fileId < RedirectionLimit` (the last time this assertion passes we will send the 21st redirection and fileId will be 21), and we expect *not* to get the 22nd request. If we get another one, fileId will be 21 and that is RedirectionLimit. We fail the assertion. (I would also rename fileId to redirectionCount, then it becomes very clear: redirectionCount = 0; and we assert `redirectionCount < RedirectionLimit`, then every time we send back 302 we ++).

Ashod (approved):
Thanks for the explanation, Szymon! It's a minor point, but I took to opportunity to explain why I find the `+1` a sign that we can improve things. Hope it makes sense. This is not a stopper, but didn't merge to give you a chance to read my explanation first. If you don't update it, I'll merge tomorrow.

Ashod (approved):
Thank you!

PR #2711, opened 2021-06-28
loleaflet: show some options in context menu only when there are more then one sheets

Rash419:
Change-Id: If96461a5670ceeda035a77715cf24d419a0c2151

* Target version: master 

### Summary
If there is only one sheet following options are enabled which doesnot make sense:

Delete Sheet
Show Sheet
Hide Sheet
Move Sheet Right/left

Proposed solution:
The above options should be disabled when there is only one sheet


dennisfrancis (approved):
Looks good. Thanks!

PR #2714, opened 2021-06-28
leaflet: refactored uno commands in top toolbar

lpranam:
Change-Id: I43fb0796507987817b8eb82828a506b1f874d8ca

* Resolves: # 

### Summary
added missing uno commands
modified uno property to use full command (i.e: use prefix ".uno:")

some item had custom handling and adding uno property would interfere the correct functionality
but to make freemium work properly each item needs a command
to solve the problem a new propery to store the command introduced: freemiumUno

This unifies the button properties with other JSONs

This is also necessary for freemium, without uno property button can not be blocked

### TODO

- [ ] ...

### Checklist


kendy:
If this is needed, let's get it in.  I am not overly excited about the "freemiumUno" vs. "uno" - this will create confusion.  Let's talk about the possibilities there :-)

PR #2729, opened 2021-06-30
Show verbose feedback when toggling log level and Trace Event generation

tml1024:
Sadly I had to use a global variable to access the actual
contents of the Help>About window as being displayed in its key event
handler. If somebody can figure out the correct clean way to do it, be
my guest. (Look for the ExpertlyTrickForLOAbout.)

Change-Id: I88582e490d566d127d60226e045535e12c172e9c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
The Cypress thing will fail, let's ignore that.

PR #2730, opened 2021-06-30
Avoid assertion failure if toggling Trace Event generation back and f…

tml1024:
…orth

Change-Id: Ie6b7144ecf16faa9d22e7d87801d9dbe9b97d3fc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
(The assertion failure happens now when we are surrounding the very functionality that can toggle Trace Event recording with a ProfileZone, so it can be on when the ProfileZone is constructed and off when the ProfileZone is destructed...)

PR #2737, opened 2021-07-01
Initialize the vectors instead of reserving

mikekaganski:
The vectors are accessing the elements without pushing to the back.

Change-Id: Ia01b54f80b70173b994e776e05710e92248f0e23


thebearon (approved):
Thanks, Mike!

PR #2738, opened 2021-07-01
fix: avoid crash in non writer apps

dennisfrancis:
when calling hasTableSelection(). Do not depend on stray/duplicate
tableselection messages that can ensure a valid _currentTableData
member of TileLayer.

Change-Id: I957fdfc43c1435ca98eb204dcb3ec1e76b85ce89

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Thanks, at least the TableOverlay.js piece looks reasonable to me.

PR #2756, opened 2021-07-05
Move side bars into flex container.

gokaysatir:
So, we no longer need to set positions of HTML elements using margins or absolute positioning or attributes like "left: 30px".

Tested on mobile, desktop, tablet views with & without notebookbar.

Change-Id: Ib301dfa8c06a9cf12acd33caf1b6a852a2faab20

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/common/mobile_helper.js
+++ b/cypress_test/integration_tests/common/mobile_helper.js
@@ -4,6 +4,16 @@ var helper = require('./helper');
 
 // Enable editing if we are in read-only mode.
 function enableEditingMobile() {
+
+	//https://stackoverflow.com/a/63519375/1592055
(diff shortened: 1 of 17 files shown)


pedropintosilva (changes requested):
I have tested and everything seems ok! : ) (desktop, mobile, tablet, multiple browsers)
But could you please rename the id to something more unique? 
from `flex-row-container` to `main-content` or `document-sidebar-container`

pedropintosilva (approved):
Thanks for the new commit. Just to make the life easier when choosing what to port (and to avoid missing them) I have rebased and modified your first commit (and removed the last one)

PR #2763, opened 2021-07-05
Stop clipboard copy operation on selection change

mikekaganski:
So that when user decided to not start download or copy to clipboard,
and moved to other things, the notification gets disposed.

Change-Id: Ifebd61594a87a37e83b0ec81b62c501792913a98

### DIFF
--- a/loleaflet/src/control/Control.DownloadProgress.js
+++ b/loleaflet/src/control/Control.DownloadProgress.js
@@ -85,6 +85,17 @@ L.Control.DownloadProgress = L.Control.extend({
 		return this._started;
 	},
 
+	currentStatus: function () {
+		if (this._closed)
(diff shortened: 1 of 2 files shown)


mmeeks:
Surely we hide the cursor when we alt-tab window switch: that is a big feature - so we don't have blinking cursors when there is no focus. I would strongly recommend using the events we have (you can trivially find them by enabling debugging by tripple clicking help->about and watching the console log) - that track the selection changes we see in the document: rather than cursor changes. But - of course, perhaps this is ok - I've not tested.

Might be nice to have a chatty method name for a helper function so it's easier to grasp the intent.

mmeeks:
Looks great Mike thanks for cleaning that up =)

PR #2776, opened 2021-07-06
Enabled bulk insert for sections.

gokaysatir:
Now, one can add all sections at once.

addSection(sectionObject, bulkInsert = true) => call for them all.

addSection(sectionObject, bulkInsert = false) => call for the last section (shouldn't be a document object), so it will trigger renewAllSections function.

Change-Id: Ie83265b5a57c11bd828a980c1797d6bfa3d6e287

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CanvasSectionContainer.ts
+++ b/loleaflet/src/layer/tile/CanvasSectionContainer.ts
@@ -285,6 +285,7 @@ class CanvasSectionContainer {
 	// Above 2 properties can be used with documentBounds.
 	private drawingPaused: boolean = false;
 	private dirty: boolean = false;
+	private sectionsDirty: boolean = false;
 
 	// For window sections.
 	private windowSectionList: Array<CanvasSectionObject> = [];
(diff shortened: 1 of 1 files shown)


mmeeks:
Hmm - I'd prefer a freeze & thaw method generally - it's easier to understand, and AFAIR we already use something like this for the slurped event processing (?) But of course - if this works - that's good. Did you re-test the calc document with lots of comments' performance ? =)

gokaysatir:
In slurped event processing, we have the events one by one and we process them together i think. Here, we have all the info.
"Bulk insert" removes unnecessarily repeating of some heavy functions and i hope it will work.

On my end, document is not that slow. "importComments" function is a lot faster now. Thinking about the screen shot given in the bug report, "importComments" function was very slow, this commit should change that.

mmeeks (approved):
Beautiful - thanks Gokay - hopefully it makes all sorts of other things faster too =)

gokaysatir:
Thanks for the help:)

PR #2785, opened 2021-07-06
CanvasTileLayer: Correct the check for the tile source.

gokaysatir:
Change-Id: I4990f8c9d44c92f711fb4d973bf44ef77812c757

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir:
@dennisfrancis that change came with my previous changes and something that i forgot to remove i think. This patch restores your original code. Can you review when you have time.

dennisfrancis (approved):
Looks good. Thanks!

mmeeks:
Hmm - by the time we get a tile - we should have a valid src - since we don't allow tiles to come out of the core unless they are valid nowadays.

PR #2794, opened 2021-07-07
Guard against two uncaught JS exceptions

tml1024:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Saw these two in the loolwsd.log when running the perf-test.

tml1024:
Seems that the Cypress thing is stuck or something, and I see "FTL  Failed to fork child processes" in the console output. In other words, situation normal, CI dysfunctional.

tml1024:
How not surprising, the Cypress thing failed. Let's iterate once more.

tml1024:
Let's assume the Cypress failure is unrelated.

PR #2802, opened 2021-07-08
Don't output errno information when a syscall hasn't failed

tml1024:
Change-Id: I90a65c444d9a300f7dd446db6c3617d7f580855c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1080,10 +1080,9 @@ public:
                 len = readData(buf, sizeof(buf));
                 last_errno = errno;
 
-                LOG_TRC('#' << getFD() << ": Read incoming data " << len << " bytes in addition to "
-                            << _inBuffer.size() << " buffered bytes ("
-                            << Util::symbolicErrno(last_errno) << ": " << std::strerror(last_errno)
-                            << ')');
+                if (len > 0)
+                    LOG_TRC('#' << getFD() << ": Read incoming data " << len << " bytes in addition to "
+                            << _inBuffer.size() << " buffered bytes");
 
                 if (len < 0 && last_errno != EAGAIN && last_errno != EWOULDBLOCK)
                     LOG_SYS_ERRNO(last_errno, '#' << getFD() << ": Socket read returned " << len);
@@ -1337,10 +1336,9 @@ public:
                 len = writeData(_outBuffer.getBlock(), size);
                 last_errno = errno; // Save right after the syscall.
 
-                LOG_TRC('#' << getFD() << ": Wrote outgoing data " << len << " bytes of "
-                            << _outBuffer.size() << " buffered bytes ("
-                            << Util::symbolicErrno(last_errno) << ": " << std::strerror(last_errno)
-                            << ')');
+                if (len > 0)
+                    LOG_TRC('#' << getFD() << ": Wrote outgoing data " << len << " bytes of "
+                            << _outBuffer.size() << " buffered bytes");
 
 #ifdef LOG_SOCKET_DATA
                 if (len > 0 && !_outBuffer.empty())


tml1024:
@Ashod , you have touched that stuff, if you like to keep things as is, please just close this PR without merging.

tml1024:
Ping @Ashod .

Ashod:
@tml1024: sorry, hadn't seen this. These are purely for debugging socket issues and communication issues. If we make it conditional and remove the errno stuff, it will be far less useful, because the issue often is not understanding what is going on when no data is going around. It was added to help troubleshoot some issues in the past and it proved useful.

Is it too noisy? I suppose it could be and we could have a compile-time directive to disable them (like we do with `LOG_SOCKET_DATA`).

Or is there another reason?

tml1024:
> These are purely for debugging socket issues and communication issues

Sure, but what "issue" is there if no system call has failed? Why report an errno value set by some random failing system call in history, not at all related to what the code is doing right there?

tml1024:
As in:

```TRC  #18: Read incoming data 188 bytes in addition to 0 buffered bytes (EOPNOTSUPP: Operation not supported)```

The EOPNOTSUPP is *not* an error indicated by the recvmsg() system calls performed by the readFD() function called by the readData() function. (How could it be; the recvmsg() succeeded, it received 188 bytes.)

tml1024:
If the thinking here is something along the lines of _but what if some code is changed in the future so that there might be a system call failure deeper down in the stack that is somehow relevant but not reported_, is the solution then to report the errno here, higher up in the call stack ? Why not report system call failures exactly at the spot where they fail?

Ashod:
> As in:
> 
> `TRC #18: Read incoming data 188 bytes in addition to 0 buffered bytes (EOPNOTSUPP: Operation not supported)`
> 
> The EOPNOTSUPP is _not_ an error indicated by the read() or recvmsg() system calls (in the readFD() and readData() functions). How could it be; the read() or recvmsg() clearly succeeded, it received 188 bytes.

I see now. I think you are looking at the mobile implementation, while these logs were written for the naked read/write sys-calls. I suppose you could reset errno, if it's giving nonsense values.

I mean, when you say "The EOPNOTSUPP is _not_ an error indicated by the read()", how is the log to know this? If we check the value of errno and only log when it's value is non-zero, then we *will* print EOPNOTSUPP, because we cannot know what you are asserting (that it's not an error indicated by read).

The sys-calls should clear errno and if we have fake socket or something similar, we should clear errno. Otherwise, we can only assume there was an error that set errno.

Notice that we assume that the `readData()` and `writeData()` calls should manage errno. For example, they should preserve it from the relevant sys-call. Again, the design here is that these functions are equivalent to the sys-call, and if they aren't then the implementation should reset errno or set it as appropriate.

But I don't think removing the log is a good idea, because the whole idea is to use errno (as reported by the sys-call) to help us troubleshoot. If it's messed up in some cases, then we should fix those cases.

tml1024:
>  I think you are looking at the mobile implementation,

No. That was from a _make run_ on Linux.

> when you say "The EOPNOTSUPP is not an error indicated by the read()", how is the log to know this?

From the return value of the system call, which (I hope, to lazy to check right now) is propagated through the one level of wrapper. Only when a system call has failed is errno set to a value.

> The sys-calls should clear errno

No system call "clears" errno if they succeed. A system call sets errno if it fails, but doesn't touch it on success.

But oh well, I give up.

kendy:
Can't we just clear the errno before the readData() and before the writeData() and be done with that? ;-)

tml1024:
Nah.

PR #2807, opened 2021-07-08
Improve cell cursor/selection/autofill-marker style

dennisfrancis:
* Resolves: # 
* Target version: co-6-4

### Summary
Improved styling of cell cursor/selection autofill-marker.

1. Improve styling of cursor/selection/autofill-marker
    
    * Autofillmaker should follow the grid (not go overboard as it is now)
    * Add one white border to the selection (but very discreet, 1 pixel)
    * The autofillmarker, selected outline and cell cursor should share the
      same color. That color should come from css var (so it fits better
      with surrounding elements, cell headers etc).
    
    Caveats:
    
    * Autofill-marker in mobile devices was positioned and sized
      differently. With this change, they stay the same except for the color
      and the contrast border. Styling this needs more thought.

2. Separate the styling of cell and text selections
    
    We now have a very specific design for cell-area selections which
    involves many paths. However we don't need the same styling for text
    selections. Use a simple CPolygon based drawing for text-selections till
    we come up with a better styling for them.

3. Fix render ordering of CPaths based on groups [cell-cursor, cell-selection etc..] and viewId.

### Checklist

### DIFF
--- a/loleaflet/css/loleaflet.css
+++ b/loleaflet/css/loleaflet.css
@@ -21,6 +21,9 @@
 	--gray-light-txt--color: #696969;
 	--gray-light-bg-color: #EFEFEF;
 	--gray-color: #b6b6b6;
+
+	/* Shared color for cell and selection border */
+	--cell-cursor-selection-border-color: #0074e8;
 }
 /* clip technique: hide visually but keep it available to screen readers */
 .visuallyhidden {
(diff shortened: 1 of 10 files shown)


pedropintosilva:
Desktop
calc-cursors

Mobile
calc-mobile-cursors

pedropintosilva (approved):
Everything seems ok, thanks!
Tested on FF and Chrome and also multi-session

Desktop
calc-cursors

image


Mobile
calc-mobile-cursors

dennisfrancis:
Thanks Pedro for the review and the gif that shows the current state!

PR #2817, opened 2021-07-08
BusyPopup: show after 0.7s delay

eszkadev:
This will prevent us from flickering if popup has to be shown
for less than 0.7s.

Change-Id: I5904e5f836dae40b83ca24d19e0b2fb779eb504b

### DIFF
--- a/loleaflet/src/control/Control.UIManager.js
+++ b/loleaflet/src/control/Control.UIManager.js
@@ -8,6 +8,7 @@
 L.Control.UIManager = L.Control.extend({
 	mobileWizard: null,
 	blockedUI: false,
+	busyPopupTimer: null,
 
 	onAdd: function (map) {
 		this.map = map;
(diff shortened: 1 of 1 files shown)


mmeeks:
Looks good, but it's something of a tough one - ultimately if we take 0.8s we get a flicker ;-) I suspect we should fade that thing out after it is shown elegantly (but perhaps we already do in some other CSS magic? =)

eszkadev:
> Looks good, but it's something of a tough one - ultimately if we take 0.8s we get a flicker ;-) I suspect we should fade that thing out after it is shown elegantly (but perhaps we already do in some other CSS magic? =)

I have patch for that, but it depends on some other change, so will be backported later after we finish Busy Popup for mobile.

pedropintosilva (approved):
nice, thanks @eszkadev. I will be now checking and working on the Busy Popup for mobile so then we can test/work on your fadein/out patches :+1:

PR #2818, opened 2021-07-08
http response: check if result would fit into chunkLen

vmiklos:
Change-Id: I6e3b73461653d2ab3cedaa3f6ca7fbfd2a826edb
(cherry picked from commit 57d3d0f3919115b77761ae26af4ff40140fb7f98)


vmiklos:
@Ashod could you please take a look? Thanks.

vmiklos:
> 09:34 < kendy> @vmiklos: Given that "digit" is always less than 16, you can avoid the "- digit" in the condition (because the division turns that into 0 anyway), making it constexpr, can't you?

Done.

vmiklos:
unit-wopi-ownertermination failed, retry.

vmiklos:
unit-wopi-ownertermination again, rebase.

net/HttpRequest.cpp
```diff
@@ -553,6 +553,12 @@ int64_t Response::readData(const char* p, int64_t len)
                     if (digit < 0)
                         break;
 
+                    // Can assume that digit is always less than 16.
+                    if (chunkLen >= std::numeric_limits<int64_t>::max() / 16)
+                    {
+                        // Would not fit into chunkLen.
+                        return len - available;
```
Ashod, line 560: Wouldn't this be a fatal error? Would be be able to recover? If not, best to return -1 and fail.

Ashod (commented):
Thanks Miklos!

net/HttpRequest.cpp
vmiklos: Good point, -1 is better here, this is really a garbage input, we just should not crash on this.

vmiklos:
@Ashod OK CI is now happy with the 3rd commit, addressing your review. Please let me know if you need anything more from me. Thanks!

Ashod (approved):
Thanks Miklos, looks good to me.

Ashod:
@vmiklos, do we want to squash the 3 commits before merging?

vmiklos:
I would leave them separate, otherwise my life is harder when I try to make sure all 3 are in all relevant branches. :-)

PR #2821, opened 2021-07-08
add a status indicator label

hcvcastro:
- kit: add payload to 'statusindicatorstart'
- loleaflet: check the 'statusindicatorstart' label


timar (changes requested):
Will it work with localized UI? It seems that you check for an en-US string.

hcvcastro:
> Will it work with localized UI? It seems that you check for an en-US string.

No, Thanks,  updated.

PR #2824, opened 2021-07-09
wsd: don't re-render tiles while closing a session.

mmeeks:
During the save during shutdown we can get big invalidation
emissions, there is no point in re-rendering the view at that
stage; so skip it.

Change-Id: Ie85ef57648815a1872ca35dcb1bdfc9fa54e64c0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -2100,6 +2100,12 @@ void ClientSession::handleTileInvalidation(const std::string& message,
         return;
     }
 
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Looks good. I wonder if there are other similar cases where we process messages on a disconnected client that we can drop. Perhaps not an issue unless they spike the cpu.

mmeeks:
No idea wrt. other cases - but this was sloow =)

PR #2828, opened 2021-07-09
Log Last 4 Uno Commands During a Fatal Crash

gopi487krishna:
Records the uno commands from different instances of Child Session and
dumps the last 4 uno commands into the Crash log during a fatal crash

Signed-off-by: Gopi Krishna Menon <krishnagopi487.github@outlook.com>
Change-Id: I838f71769dc08df7076c040f3d72c15f7607e9d3

### Checklist

- [X ] Code is properly formatted

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -78,6 +78,14 @@ namespace SigUtil
 #endif
     }
 
+    UnoCommandsDumperFn dumpUnoCommandsInfoFn = nullptr;
+
+    void registerUnoCommandInfoHandler(UnoCommandsDumperFn dumpUnoCommandsInfo)
+    {
+        dumpUnoCommandsInfoFn = dumpUnoCommandsInfo;
+    }
+
+
 #if !MOBILEAPP
     /// This traps the signal-handler so we don't _Exit
     /// while dumping stack trace. It's re-entrant.
@@ -242,6 +250,11 @@ namespace SigUtil
         else
             Log::signalLog(" Fatal signal received: ");
         Log::signalLog(signalName(signal));
+
+        if (dumpUnoCommandsInfoFn != nullptr)
+        {
+            dumpUnoCommandsInfoFn();
+        }
 
         struct sigaction action;
 
--- a/common/SigUtil.hpp
+++ b/common/SigUtil.hpp
@@ -48,6 +48,12 @@ namespace SigUtil
 
     void checkDumpGlobalState(GlobalDumpStateFn dumpState);
 
+    typedef void (*UnoCommandsDumperFn)(void);
+
+    extern UnoCommandsDumperFn dumpUnoCommandsInfoFn;
+
+    void registerUnoCommandInfoHandler(UnoCommandsDumperFn dumpUnoCommandsInfo);
+
 #if !MOBILEAPP
     /// Wait for the signal handler, if any,
     /// and prevent _Exit while collecting backtrace.
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -886,6 +886,27 @@ int main(int argc, char**argv)
         return std::string(message, size);
     }
 
+    /// Eliminates the prefix from str(if present) and returns a copy of the modified string
+    inline
+    std::string eliminatePrefix(const std::string& str, const std::string& prefix)
+    {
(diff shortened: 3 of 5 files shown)


common/SigUtil.cpp
```diff
@@ -243,6 +251,11 @@ namespace SigUtil
             Log::signalLog(" Fatal signal received: ");
         Log::signalLog(signalName(signal));
 
+        if(dumpUnoCommandsInfoFn != nullptr)
```
tml1024, line 254: Our code style is to leave a space after keywords like 'if'.

common/SigUtil.hpp
```diff
@@ -48,6 +48,12 @@ namespace SigUtil
 
     void checkDumpGlobalState(GlobalDumpStateFn dumpState);
 
+    extern "C" { typedef void (*UnoCommandsDumperFn)(void); }
```
tml1024, line 51: Why the 'extern "C"' ?

common/Util.hpp
```diff
@@ -886,6 +886,24 @@ int main(int argc, char**argv)
...
+
+        std::tie(prefix_pos,str_pos)= std::mismatch(prefix.begin(),prefix.end(),str.begin());
+
+        if( prefix_pos == prefix.end() )
+        {
+            return std::string(str_pos,str.end());
+        }
+        else
+        {
+            return std::string();
```
tml1024, line 903: This function needs a comment to describe what its semantics are. Without reading the code, I would not expect it to return an empty string if str doesn't start with prefix.

tml1024: Also, compared to the startsWith() functions in the same file, this is much different in style. (Those use memcmp() directly on the string bytes.) That is probably not a problem, just a slight inconsistency.

kit/ChildSession.cpp
```diff
@@ -832,6 +859,50 @@ bool ChildSession::getStatus(const char* /*buffer*/, int /*length*/)
...
+
+void UnoCommandsRecorder::addUnoCommandInfo(const std::string& unoCommandInfo)
+{
+    // Create a new C string on heap with the contents of unoCommandInfo
+    char* command = new char[unoCommandInfo.length()+1];
+    strncpy(command,unoCommandInfo.c_str(),unoCommandInfo.length()+1);
+
+    unsigned long long currentIndex = _currentpos.fetch_add(1);
+
+    char* oldCommand = _unocommands[currentIndex%4].exchange(command);
```
tml1024, line 895: Use the NUM_UNO_COMMANDS  here.

tml1024 (commented):
See comments.

common/SigUtil.cpp
gopi487krishna: Ok Tor got it . Will update the changes

kit/ChildSession.cpp
gopi487krishna: Done

common/Util.hpp
gopi487krishna: A one line description of eliminate prefix added

kit/ChildSession.cpp
```diff
@@ -832,6 +859,50 @@ bool ChildSession::getStatus(const char* /*buffer*/, int /*length*/)
     return sendTextFrame("status: " + status);
 }
 
+void ChildSession::dumpRecordedUnoCommands()
+{
+    std::atomic<char*>* recordedCommands = unoCommandsRecorder.getRecordedCommands();
+
+    for(int i = 0; i < unoCommandsRecorder.NUM_UNO_COMMANDS; i++)
```
tml1024, line 866: At least here is a space missing. Sadly I am not sure if our git hooks try to enforce use of clang-format or not.

tml1024 (commented):
See inline comment.

kit/ChildSession.cpp
gopi487krishna: In libreoffice core they did. Now I dont know what is happening. But the hooks are installed and working because I am getting error messages of trailing whitespaces. Also code signing is done by hooks so...

tml1024:
Please do some dummy change (just git commit --amend without actually changing anything in the commit message is enough) and force-push, to re-try the checks.

PR #2882, opened 2021-07-15
Use custom local branch name as explicit remote user branch name for PR

mikekaganski:
In a scenario where onee has several parallel PRs, one needs to use
several local branches, and when updating PRs, _needs to provide PR
branch name explicitly_ each time (my-PR-name-* in the example):

  git checkout featureFoo
  [ edit session ]
  git commit
  ./g review my-PR-name-1
  git checkout featureBar
  [ edit session ]
  git commit
  ./g review my-PR-name-2

This changes the processing, so that when the user's local branch name
is different from the tracked branch name, it is treated as if user
used that name explicitly in a call:

  git checkout -b featureFoo origin/master
  git checkout -b featureBar origin/distro/collabora/co-6-4
  [ ... ]
  git checkout featureFoo
  [ edit session ]
  git commit
  ./g review # this uses "featureFoo" as implicit second argument
  git checkout featureBar
  [ edit session ]
  git commit
  ./g review # this uses "featureBar" as implicit second argument

so no need to remember which PR name was used where.
This only works for local branch names that are not equal to a remote
branch name, or to the last part (after last /) of any distro branch.

Change-Id: I8b95f92f6e205e3c93ef30f4813f3549c2315872

### DIFF
--- a/g
+++ b/g
@@ -15,6 +15,21 @@
 # 3) Uses './g review' to submit a pull request against <branch>.
 #
 
+# not_in_standard_branches $BRANCH $REMOTE
+not_in_standard_branches() {
+    # e.g., with $1 = co-4-2, and $2 = origin
+    # check for origin/co-4-2 branch
+    if ! git rev-parse --quiet --verify $2/$1 >/dev/null; then
+        return 1 # indicate failure - there's a top-level branch named that way
+    fi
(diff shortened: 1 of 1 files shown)


mikekaganski:
https://github.com/CollaboraOnline/online/pull/2880 is the correct one

PR #2885, opened 2021-07-15
Parameters for perf-test

merttumer:
now it's possible to pass parameters
typing_speed, typing_duration, single_view, node_inspect

also replaced the old perf-test with new-perf-test

Change-Id: I03f1f14db4ddd9afb9d4a9a60a87c135497db70c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/Makefile.am
+++ b/loleaflet/Makefile.am
@@ -723,6 +723,14 @@ check-local: $(MOCHA_TS_JS_FILES)
 
 SSL_FLAG = `xmllint --xpath 'string(/config/ssl/enable)' $(abs_top_builddir)/loolwsd.xml`
 
+# the delay (in miliseconds) between each emulated key stroke for each view
+typing_speed=100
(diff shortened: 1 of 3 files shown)


loleaflet/Makefile.am
```diff
@@ -723,25 +723,17 @@ check-local: $(MOCHA_TS_JS_FILES)
 
 SSL_FLAG = `xmllint --xpath 'string(/config/ssl/enable)' $(abs_top_builddir)/loolwsd.xml`
 
+typing_speed="100"
```
mmeeks, line 726: units in a comment ? =)

mmeeks (approved):
Looks good to me =) thanks Mert! do improve in a follow-up if this passes CI.

loleaflet/Makefile.am
merttumer: Nodejs treats them as string always

PR #2894, opened 2021-07-16
systemplate: ensure we have write access to directories before cleaning.

mmeeks:
Some of the folders we setup from the system are:

dr-xr-xr-x 1 michael users 2824 Jun  9 21:05 systemplate/usr/share/fonts/

before adjustment, and resist removal.

Change-Id: I8e5a96264b98d8091b205f7469bc46c401f47ab1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy (approved):
I wonder if we should use && instead of the ; ?  But approving either way...

PR #2900, opened 2021-07-16
js: reduce keyup emission to allow event de-duplication.

mmeeks:
We emit lots of 'keyup' events that don't match any keydown
event like this:

OUTGOING: textinput id=0 text=a
OUTGOING: textinput id=0 text=l
OUTGOING: key type=up char=0 key=517
OUTGOING: key type=up char=0 key=523

since we emit a textinput instead. So just emit events
when we have modifiers, or already sent an input event.

Change-Id: I0f86438d093497668bb5b749ff63ecd87d9b7636

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/map/handler/Map.Keyboard.js
+++ b/loleaflet/src/map/handler/Map.Keyboard.js
@@ -392,7 +392,13 @@ L.Map.Keyboard = L.Handler.extend({
 				keyEventFn('input', charCode, unoKeyCode);
 			}
 			else if (ev.type === 'keyup') {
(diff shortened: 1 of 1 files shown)


tml1024:
(Let's re-try the CI, maybe it works better now.)

tml1024:
Maybe that failing Cypress test is not a false positive...

tml1024:
OK, maybe this will pass CI. I modified the patch so that also for the Enter key (UNO keycode 1280) we send the "key type=up" message.

PR #2903, opened 2021-07-19
wsd: use the latest writable session

Ashod:
The latest writable session will have the longest
lifetime before expiring. We prefer the most recent
session that fits the requirements for use to upload
and take the storage lock.

Change-Id: I2cefd06612ea99380e24a47e12835bc92c929f5b

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1646,24 +1646,30 @@ std::string DocumentBroker::getWriteableSessionId() const
 {
     assertCorrectThread();
 
+    // Consider each session once, starting from the last.
     std::string savingSessionId;
-    for (auto& sessionIt : _sessions)
+    for (std::size_t index = _sessionsByTime.size() - 1; index < _sessionsByTime.size(); --index)
     {
-        // Save the document using an editable and loaded session, or first ...
-        if (savingSessionId.empty()
-            || (!sessionIt.second->isReadOnly() && sessionIt.second->isViewLoaded()
-                && !sessionIt.second->inWaitDisconnected()))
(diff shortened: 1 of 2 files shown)


mmeeks (changes requested):
Really not sure I like another list of sessions there that is sorted by time - that looks like a nightmare to maintain & keep in sync with the other list.

Can we not do a simple copy/crop readonly/sort of the existing, single _sessions list when we want to get a writeable session?

Also - when a user hits 'save' - do we use the session from that user ? =) if not we should.

Thanks !

Ashod:
> Really not sure I like another list of sessions there that is sorted by time - that looks like a nightmare to maintain & keep in sync with the other list.

There is only one point of insertion into the `_sessions` and one point of erasing. So it's not as bad as it sounds.

> Can we not do a simple copy/crop readonly/sort of the existing, single _sessions list when we want to get a writeable session?

Yes, but then we need to track the age of each and sort by that after copying. That seems more complex, but can do, sure.

> Also - when a user hits 'save' - do we use the session from that user ? =) if not we should.

We don't, because the user saving might have read-only view, but the document might still be writable.

If we use the current user's token, then this logic of using the latest-writable will only apply for auto-save. It will simplify things, but the read-only user will now result in errors broadcast to everyone, which I suppose we need to suppress, OR, replace with a private message to the user in question saying they can't save.

mmeeks:
I'd really prefer not to duplicate state - can't we come up with a 'goodness' metric and if we want to be complicated - then iterate through the sessions and update our preferred sessionId - if we get a better metric ? Then we don't need a new list or a sort.

One thing to test for - is closing the last editable session I guess and ensuring that we still have the right token to write with after it is gone / going for the final save (?)

PR #2911, opened 2021-07-19
Allow pasting into document even when DisableCopy is active

eszkadev:
From our API description:
DisableCopy
Disables copying from the document in libreoffice online backend.
Pasting into the document would still be possible.

Change-Id: I1f7e272ad3b04208b534eff990fed895951b4751


merttumer (approved):
Looks good, it does what it says.

PR #2912, opened 2021-07-19
SSL Certificate Verification

Ashod:
- wsd: move isLocalhost to a common area
- wsd: support certificate verification
- wsd: ssl certificate verification requirements
- wsd: killpoco: cert verification
- wsd: test: improve modify test
- wsd: prevent recycling the listening port during tests

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -15,10 +15,66 @@
 #include "SslSocket.hpp"
 #endif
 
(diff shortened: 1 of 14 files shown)


mmeeks:
I'm assured that we need this for cert. checking in 6.4.10 - and there is a risk it is more stringent than 6.4.9 - but we'll have to fix that as/when. Lets get it in and a new release out ASAP.

PR #2913, opened 2021-07-19
Merge textinput and removetextcontext messages when possible

tml1024:
When adding a freshly arrived textinput message to the queue, check if
there is any outstanding one queued, and if so, remove it, and prefix
its text text to the text in the fresh message.

Change-Id: I712eebfb10d410bb424157c8df46b2848a236d88

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/MessageQueue.hpp
+++ b/common/MessageQueue.hpp
@@ -181,16 +181,17 @@ protected:
             // If any messages of these types are present before the current ("textinput") message,
             // no combination is possible.
             if (queuedTokens.size() == 1 ||
-                queuedTokens[1] == "key" ||
-                queuedTokens[1] == "mouse" ||
-                queuedTokens[1] == "removetextcontext" ||
-                queuedTokens[1] == "windowkey" ||
-                (queuedTokens[0] != tokens[0] && queuedTokens[1] == "textinput"))
+                (queuedTokens[0] == tokens[0] &&
+                 (queuedTokens[1] == "key" ||
+                  queuedTokens[1] == "mouse" ||
+                  queuedTokens[1] == "removetextcontext" ||
+                  queuedTokens[1] == "windowkey")))
                 return std::string();
 
             std::string queuedId;
             std::string queuedText;
-            if (queuedTokens[1] == "textinput" &&
+            if (queuedTokens[0] == tokens[0] &&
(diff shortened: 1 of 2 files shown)


tml1024:
Using the "mash the keyboard as fast as you can" test, against a "make run" on a very lightly loaded machine, I got these combination operations to occur fairly reliably a few times per second, when inputting text in a table.

tml1024:
(Oops, hit the "close with comment" button by mistake.)

tml1024:
And yeah, this can't go in as such yet, also need to make sure we don't merge text input messages with some backspacing in-between.

tml1024:
I had to unset the "draft" state to get the checks to be run, but I know that this is *not* ready for merging. Don't touch that button please.

tml1024:
One thing still missing from the code is to make sure that textinput messages are merged only when they are from the same client. Apparently to ensure that it is enough to check that the first token in the message, the thing that looks like "child-123" is identical. Possibly we could even do so that if there is a textInput message from another client in the queue, we don't look for ones from the same client any further down, to avoid potential problems?

tml1024:
OK, now I try to take care of the situation I mentioned in my previous comment.

tml1024:
@mmeeks , I have a patch to merge removetextcontext messages from the same client similarly, should I wait until this PR is merged and then submit it as a follow-up PR, or add the commit for that to this PR? (Once I have verified manually locally that it works as expected.)

mmeeks:
Hi Tor - glad it's progressing; we can put a couple of patches in the same PR I guess (?) =)
It is a very interesting Q regarding the input streams of different clients. Clearly it is most ideal - if we can merge together and glup into one event input from the same client - regardless of other client's editing activity; otherwise I fear we get slowest when we most want to be fast (?). So I would be inclined to walk the queue back filtering on a matching client-id, and ignore other views' activity in that filter - hopefully that gives the best win (?)

tml1024:
@mmeeks OK. Now this PR contains more commits.

mmeeks:
Looks great to me and reasonably safe I think - lets merge it =) does it improve life when we have multiple concurrent editors I guess - this is the real question =)

PR #2923, opened 2021-07-20
cleanup: don't assume that the web-server poll is still alive.

mmeeks:
Since a99edb5d29 this gets a different lifecycle, and is often
joined and gone by the time we get to cleanup.

Change-Id: Ic21ca837d484a9a853c7f8f683e876406b00a943

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Pushed manually with DCO =)

PR #2924, opened 2021-07-21
Don't use the obsolete 'type' parameter in 'textinput' messages

tml1024:
Our JS code does not send 'textinput' messages with a 'type' parameter
either, so it is pointless to do it in a unit test.

(We still have code that checks for the 'type' parameter in received
'textinput' messages. That is pointless, too, and should be removed.)

Also, document what testWriterTyping() apparently is doing.

Change-Id: I4954379339f14e4a310a76788521001048d84e37

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me =) the type is obsolete; did you get around to updating protocol.txt ?

Issue #2771, opened 2021-07-06
CODE 6.4.10 - Well, this is embarrassing


manu-p:
**Describe the bug**
Since 6.4.10, the message "Well, this is embarrassing..." has appeared on an remote instance that had been working fine for a long time before.
Configuration:
NC instance "cloud.NC-**client**" running NC 20.0.6 and Collabora 3.8.3 using a CODE server named "code.NC-**server**" (below).

NC instance "cloud.NC-**server**" running NC 20.0.6 and Collabora 3.8.3 and CODE in a docker, the same code.NC-server.

Since latest 6.4.10 everything keeps working fine on NC-server, CODE is opening odt, ods, docx, ... files as it did before. 
But on NC-client, CODE is not opening files anymore as it did before. The message "Well, this is embarrassing..." is displayed instead.

I managed to revert to the previous 6.4.9 image in docker on NC-server and everything's working fine again on NC-client and NC-server.

thebearon:
Can you share what errors the CODE docker logs show when you try opening a file?

manu-p:
Hi @thebearon thanks.

Not easy, unless I can find old docker logs somewhere ?
My browser tells me yesterday I've been looking for:
`ERR Socket #36 SSL BIO error: error:14094438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error (0: Success)| ./net/SslSocket.hpp`
and
`ERR Socket #22 SSL BIO error: closed unexpectedly (-1). (0: Success)| ./net/SslSocket.hpp:294`
which were amongst the 1st ERR messages I could find in "docker logs -f"

The NC-server is running Ubuntu 18.04 with nginx 1.14. 
Docker version 20.10.7, build f0df350
The subdomain code.NC-server has its own LetsEncrypt certificate which I forced to renew, in case.

At last, I reverted all nginx and loolwds.xml files back to their previous setups which used to work fine before friday and, as said before, managed to revert to CODE 6.4.9 image in docker.

These two servers are production servers, I can't play too much with their setup.

stavros-k:
```
Ready to accept connections on port 9980.

wsd-00007-00045 2021-07-09 19:36:27.117439 [ websrv_poll ] WRN client - server version mismatch, disabling browser cache. Expected: 2c4ce99| wsd/FileServer.cpp:288

[...]

Issue #2752, opened 2021-07-04
Loolwsd 6.4.10 - Debian 9 - SNI enforced on OwnCloud/NextCloud instance get a 403 reply on CODE


Athar42:
**Describe the bug**
With the latest update to Loolwsd 6.4.10 (and other collabora updates related), the server can't work anymore on a client which used the "SSLStrictSNIVHostCheck" setting to ON.
This setup was working prior the latest updates.

**To Reproduce**
1. On the client (hosting the NextCloud application for example), just set the "SSLStrictSNIVHostCheck" to ON
2. Try to open a document which Collabora would open, it will fail with an access denied.
3. Set the previous setting to off, test again, and it'll pass.

**Expected behavior**
Should work.

**Actual behavior**
Document can't load as the loolwsd service didn't comply with this Apache directive "SSLStrictSNIVHostCheck".

**Desktop (please complete the following information)**
 - OS: Debian 9.13
 - CODE : 
```
 # loolwsd --version-hash
51b8716
# loolwsd --version
terminate called after throwing an instance of 'std::runtime_error'
Abandon
```
**As you may see, even the "--version" is broken there**
But in the log, we have this : 
`wsd-20943-20943 2021-07-04 15:30:34.621808 [ loolwsd ] INF  Loolwsd version details: 6.4.10 - 51b8716 - on "Debian GNU/Linux 9 (stretch)"| wsd/LOOLWSD.cpp:4069`

**Additional context**
On the webserver : 
Error.log
`[Sun Jul 04 17:04:41.466646 2021] [ssl:error] [pid:tid] AH02033: No hostname was provided via SNI for a name based virtual host`

Access.log
```
CODE SERVER IP - - [04/Jul/2021:17:04:39 +0200] "GET /index.php/apps/richdocuments/wopi/files/32_odfl?access_token=WXYZ&access_token_ttl=0 HTTP/1.1" 403 2255 "-" "LOOLWSD HTTP Agent 6.4.10"
CODE SERVER IP - - [04/Jul/2021:17:04:41 +0200] "GET /index.php/apps/richdocuments/wopi/files/32_odfl?access_token=WXYZ&access_token_ttl=0&permission=edit HTTP/1.1" 403 2255 "-" "LOOLWSD HTTP Agent 6.4.10"
```

On CODE server log : 
[...]

PR #2925, opened 2021-07-21
ssl: setup hostname for SNI

eszkadev:
For some servers we receive failure with HTTP 403 Forbidden in WOPI::CheckFileInfo

"Reason: The client software did not provide a hostname using Server
Name Indication (SNI), which is required to access this server"


eszkadev:
see: #2771

eszkadev:
see: https://github.com/CollaboraOnline/online/issues/2752

PR #2928, opened 2021-07-21
TopToolbar get Context related commands

Andreas-Kainz:
Change-Id: Ic23c242b9c097a90819c43088d4d6573517c9043


Andreas-Kainz:
@eszkadev Settings come from 
https://github.com/CollaboraOnline/online/commit/95440e3e9de57aae49fe3fda225b6f700dcd1794

Patch work, BUT the icon wasn't shown, cause the command wasn't visible (before) If you add uno:bold with context: ['Table'], the icon was visible.

Andreas-Kainz:
Can you make the context command property more generic, cause I'd like to use it also at MobileBottomBar also. There it is even more important, cause on mobile, you have less space.

eszkadev:
will take a look when I will have some time

eszkadev:
hi, @Andreas-Kainz  I've added context support for mobile toolbar

Andreas-Kainz:
Screenshot_20210820_075046

hi, @eszkadev I tested the PR but it's as it is before, the context related command was added but you don't see the icon.

how to reproduce:
- classic mode
- insert table
- top toolbar added commands but didn't show an icon (if the icon wasn't loaded before)

pedropintosilva:
@Andreas-Kainz I have checked the PR, it works. The icons are missing because you need add them in ``loleaflet/css/toolbar.css`` see list starting from line `:757`

Nevertheless I would really like to keep this PR clear and without any icons (100% technical)
So the commit ``TopToolbar get Context related commands`` should do just that, it should be without the added icons, meaning without:
```
{type: 'break',  id: 'breaksidebar', hidden: true, context: ['Table']},
{type: 'button',  id: 'insertcolumnsbefore',  img: 'insertcolumnsbefore', uno: '.uno:InsertColumnsBefore', context: ['Table']},
{type: 'button',  id: 'insertrowsbefore',  img: 'insertrowsbefore', uno: '.uno:InsertRowsBefore', context: ['Table']},
```

And then create a new PR with a commit per each contextual icon group.

Also, and I think it's also your idea, per context we should probably refrain from adding too many icons specially since we are talking about classic mode (add in a maximum 2 or 3 icons context icons)

Andreas-Kainz:
ah perfect now I know how it will work. thanks. will prepare an updated pr without the table stuff

Andreas-Kainz:
thanks for all the help. fix was done with #3248

PR #2930, opened 2021-07-22
Support aborting unloading

Ashod:
- leaflet: explanatory message when loading fails while unloading
- wsd: allow new views while unloading
- wsd: log discovery.xml response at trace level

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -482,7 +482,7 @@ void DocumentBroker::pollThread()
 #endif
         if (_sessions.empty() && (isLoaded() || _docState.isMarkedToDestroy()))
         {
-            if (isAsyncSaveInProgress())
+            if (_saveManager.isSaving() || isAsyncSaveInProgress())
             {
                 LOG_DBG("Don't terminate dead DocumentBroker: async saving in progress for docKey [" << getDocKey() << "].");
                 continue;
@@ -1089,7 +1089,8 @@ bool DocumentBroker::attemptLock(const ClientSession& session, std::string& fail
 DocumentBroker::NeedToUpload DocumentBroker::needToUploadToStorage() const
 {
     // When destroying, we might have to force uploading if always_save_on_exit=true.
-    if (isMarkedToDestroy())
+    // If unloadRequested is set, assume we will unload after uploading and exit.
+    if (isMarkedToDestroy() || _docState.isUnloadRequested())
     {
         static const bool always_save
             = LOOLWSD::getConfigValue<bool>("per_document.always_save_on_exit", false);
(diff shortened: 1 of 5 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -1975,6 +1980,12 @@ size_t DocumentBroker::removeSession(const std::string& id)
         // If last editable, save and don't remove until after uploading to storage.
         if (!lastEditableSession || !autoSave(isPossiblyModified(), dontSaveIfUnmodified))
```
mmeeks, line 1981: I worry about the last readonly session - how do we mark to destroy for those sessions ?

wsd/DocumentBroker.cpp
```diff
@@ -1934,10 +1941,8 @@ size_t DocumentBroker::removeSession(const std::string& id)
         }
         std::shared_ptr<ClientSession> session = it->second;
 
-        // Last view going away, can destroy.
-        if (_sessions.size() <= 1)
```
mmeeks, line 1938: Ensuring we handle _sessions.size == 0 carefully is important to not leak sessions of course.

wsd/DocumentBroker.cpp
```diff
@@ -1999,6 +2010,14 @@ void DocumentBroker::disconnectSessionInternal(const std::string& id)
             LOOLWSD::dumpEndSessionTrace(getJailId(), id, _uriOrig);
 #endif
 
+            if (_docState.isUnloadRequested())
+            {
+                // We must be the last session, flag to destroy if unload is requested.
+                assert(_sessions.size() == 1 && "Unload-requested with multiple sessions.");
+                _docState.markToDestroy();
+                LOG_TRC("Unload requested while disconnecting session [" << id << "], marking to destroy.");
```
mmeeks, line 2018: lets have _sessions.size() in that trace - just in case.

wsd/DocumentBroker.cpp
```diff
@@ -3066,6 +3088,7 @@ void DocumentBroker::dumpState(std::ostream& os)
     os << "\n  doc activity: " << DocumentState::toString(_docState.activity());
     if (_docState.activity() == DocumentState::Activity::Rename)
         os << "\n  (new name: " << _renameFilename << ')';
+    os << "\n  unload requested: " << _docState.isUnloadRequested();
```
mmeeks, line 3091: glad to see it there =)

mmeeks (changes requested):
I would like every use-case of: isMarkedToDestroy properly audited.

Here is the first one I saw that looked dodgy:

DocumentBroker::NeedToUpload DocumentBroker::needToUploadToStorage() const
{
    // When destroying, we might have to force uploading if always_save_on_exit=true.
    if (isMarkedToDestroy())

I didn't read further. I assume we have a unit test for that though (?).

In general I'd prefer a 'state' that moves from Normal -> Unloading -> ToDestroy rather than two booleans but ... perhaps we can cope with that =)

wsd/DocumentBroker.cpp
Ashod: Good point. Improved.

wsd/DocumentBroker.cpp
Ashod: Restored.

kendy: Not sure about the previous version of the patch; the current one turned <= 1 into == 1; but as we talked with @Ashod , we cannot get this far with empty _sessions (we search the id in there, and return if it was not found).

kendy:
Tested locally, and it does what it is supposed to do for me :-)  Also checked that the number of sessions is behaving as it should (eg. we temporarily have 2 sessions when the browser window is closed, but another browser window connects to the document while autosaving due to the closed window).

Tried to review the code too; I think we are good, thank you!

PR #2935, opened 2021-07-22
Print a header in front of the list of last UNO commands

tml1024:
Change-Id: I1f7194edb833d8569ca7bd87cadd903deb84ef0e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
To at least see that we hit that part of the code, even if for some reason no UNO commands have been recorded.

PR #2937, opened 2021-07-22
renamefile: Respond with reload instead

merttumer:
This way it will reload the document without getting
additional errors like "this is embaressing.."
and not timed out on large files due to retrying for a limited
time.

Change-Id: Ied341200ca21e1b7243daaf62813bc9ed69fecd3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -691,6 +691,10 @@ app.definitions.Socket = L.Class.extend({
 					showMsgAndReload = true;
 				}
 			}
+			else if (textMsg.startsWith('reloadafterrename')) {
+				msg = _('Reloading the document after rename');
+				showMsgAndReload = true;
+			}
 
 			if (showMsgAndReload) {
 				if (this._map._docLayer) {
@@ -1110,7 +1114,6 @@ app.definitions.Socket = L.Class.extend({
 	_renameOrSaveAsCallback: function(textMsg, command) {
 		this._map.hideBusy();
 		if (command !== undefined && command.url !== undefined && command.url !== '') {
-			this.close();
 			var url = command.url;
 			var accessToken = this._getParameterByName(url, 'access_token');
 			var accessTokenTtl = this._getParameterByName(url, 'access_token_ttl');
@@ -1134,17 +1137,6 @@ app.definitions.Socket = L.Class.extend({
 			var docUrl = url.split('?')[0];
 			this._map.options.doc = docUrl;
 			this._map.options.previousWopiSrc = this._map.options.wopiSrc; // After save-as op, we may connect to another server, then code will think that server has restarted. In this case, we don't want to reload the page (detect the file name is different).
-			this._map.options.wopiSrc = encodeURIComponent(docUrl);
-
-			// if this is save-as, we need to load the document with edit permission
-			// otherwise the user has to close the doc then re-open it again
-			// in order to be able to edit.
-			if (textMsg.startsWith('saveas:'))
-				this._map.options.permission = 'edit';
-			this._map.loadDocument();
-			this._map.sendInitUNOCommands();
-
-
 			if (textMsg.startsWith('renamefile:')) {
 				this._map.fire('postMessage', {
 					msgId: 'File_Rename',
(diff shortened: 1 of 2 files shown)


loleaflet/src/core/Socket.js
```diff
@@ -1130,20 +1129,25 @@ app.definitions.Socket = L.Class.extend({
...
 
+			// setup for loading the new document, and trigger the load
+			var docUrl = url.split('?')[0];
+			this._map.options.doc = docUrl;
+			this._map.options.previousWopiSrc = this._map.options.wopiSrc; // After save-as op, we may connect to another server, then code will think that server has restarted. In this case, we don't want to reload the page (detect the file name is different).
+			var encodedURL = encodeURIComponent(docUrl);
+			var reload = this._map.options.wopiSrc != encodedURL;
+			this._map.options.wopiSrc = encodedURL;
+
+			if (reload) {
```
Ashod, line 1146: Minor point: since you have previousWopiSrc, you could avoid the `reload` variable and replace these 4 lines with:
```
			this._map.options.wopiSrc = encodedURL;
			if (this._map.options.previousWopiSrc != encodedURL) {
```

wsd/DocumentBroker.cpp
```diff
@@ -1491,9 +1491,33 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
             Poco::URI::encode(filename, "", encodedName);
             const std::string filenameAnonym = LOOLWSD::anonymizeUrl(filename);
 
+            constexpr const char* needle = "?access_token";
+            std::size_t index = getPublicUri().toString().find(needle);
```
Ashod, line 1495: This is problematic because access_token doesn't have to be the first parameter, it could be later, in which case it will be "&access_token". Perhaps search for "access_token=" ? (But best to use Poco::URI and getQueryParameters).

wsd/DocumentBroker.cpp
```diff
@@ -1491,9 +1491,33 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
...
+            std::string url1, url2;
+            if (index != std::string::npos)
+            {
+
+                url1 = getPublicUri().toString().substr(0, index);
+                url2 = url.substr(0, index);
```
Ashod, line 1501: What is `url` and why are we mixing `getPublicUri()` with `url` ?

What happens when `index` is shorter or longer than `url`? 

If they are identical, then let's use the same string please.

wsd/DocumentBroker.cpp
```diff
@@ -1491,9 +1491,33 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
...
             oss << "renamefile: " << "filename=" << encodedName << " url=" << url;
             broadcastMessage(oss.str());
+            if (!strcasecmp(url1.c_str(), url2.c_str()))
+            {
+                // WOPISrc has not changed
+                // However, due to rename, timestamp in the storage has changed
+                // and any further upload requests will result to a conflict
+                // Do notify the client about the conflict
+                // This will make the client reload the document properly
+                broadcastMessage("close: documentconflict");
```
Ashod, line 1519: Broadcasting `renamefile:` command followed with `close:` is not ideal. The client might actually not process the `close:` for some time after getting `renamefile:` and cause other issues/bugs. Why can't we send only one of these? (Either `renamefile:` or `close:`, but not both.)

wsd/DocumentBroker.cpp
```diff
@@ -1491,9 +1491,33 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
...
             std::ostringstream oss;
             oss << "renamefile: " << "filename=" << encodedName << " url=" << url;
             broadcastMessage(oss.str());
+            if (!strcasecmp(url1.c_str(), url2.c_str()))
+            {
+                // WOPISrc has not changed
+                // However, due to rename, timestamp in the storage has changed
+                // and any further upload requests will result to a conflict
+                // Do notify the client about the conflict
+                // This will make the client reload the document properly
```
Ashod, line 1518: I wonder why couldn't we do this in the JS code? We do send `renamefile:` with the new URL, right? And it has the old one. So, it knows we are renaming and has the old and new url, why can't it close and reopen? Why should WSD send it an explicit `close:`?

Ashod (commented):
Thanks Mert! Good stuff. Left some notes, but the main one is that we should be able to have this logic in the UI/JS side, no need to send explicit `close:` from WSD. Let me know what you think.

Ashod (approved):
Great, thanks @merttumer!

PR #2938, opened 2021-07-22
Wait for async save result on DocumentBroker stop

eszkadev:
WIP, should be merged on top of https://github.com/CollaboraOnline/online/pull/2930

When async save was started but we want to stop DocumentBroker we shouldn't kill the socket which listens to save result.
Because when we kill socket we will destroy DocumentBroker before saving is completed, this allows for a minimal time to connect again to the document and receive old content from the WOPI storage.
When we wait for async save result we will keep DocumentBroker alive and next session can be connected and receive correct content.

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -482,6 +482,12 @@ void DocumentBroker::pollThread()
 #endif
         if (_sessions.empty() && (isLoaded() || _docState.isMarkedToDestroy()))
         {
+            if (isAsyncSaveInProgress())
+            {
+                LOG_DBG("Don't terminate dead DocumentBroker: async saving in progress for docKey [" << getDocKey() << "].");
+                continue;
+            }
+
             // If all sessions have been removed, no reason to linger.
             LOG_INF("Terminating dead DocumentBroker for docKey [" << getDocKey() << "].");
             stop("dead");
@@ -3096,6 +3102,16 @@ void DocumentBroker::dumpState(std::ostream& os)
             proxy->dumpProxyState(os);
     }
 #endif
+}
+
+bool DocumentBroker::isAsyncSaveInProgress() const
+{
+    if (!_storage)
+        return false;
+
+    StorageBase::AsyncUpload::State state = _storage->queryLocalFileToStorageAsyncUploadState().state();
+
+    return state == StorageBase::AsyncUpload::State::Running;
 }
 
 /* vim:set shiftwidth=4 softtabstop=4 expandtab: */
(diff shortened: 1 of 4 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -389,6 +389,12 @@ void DocumentBroker::pollThread()
             continue;
         }
 
+        if (isModified() && _storage->queryLocalFileToStorageAsyncUploadState().state() == StorageBase::AsyncUpload::State::Running)
+        {
+            _stop = false;
```
Ashod, line 394: Can't we add this check before setting `_stop = true;` ?

We should be able to do that when we get the socket disconnection, or whenever the trigger condition is met.

Resetting the _stop flag has many obvious issues, but one thing that might not be obvious is that some other logic might have already been triggered by the time we reset it, but we can't undo that logic (certainly not when we cannot know about it here). 

So it's always best to avoid changing the _stop flag in the first place.

Ashod (commented):
Thanks Szymon! I think I understand the idea, but it would be far better to do the check before setting the _stop flag, instead of resetting it after the fact.

I don't think there are infinite loops or anything like that in this solution, but once the _stop flag is set, it's impossible to undo it reliably (because we don't know other logic is affected).

wsd/LOOLWSD.cpp
```diff
@@ -2237,7 +2237,10 @@ class PrisonerRequestDispatcher : public WebSocketHandler
         {
             std::unique_lock<std::mutex> lock = docBroker->getLock();
             docBroker->assertCorrectThread();
-            docBroker->stop("docisdisconnected");
+            if (docBroker->isAsyncSaveInProgress())
+                LOG_DBG("Don't stop DocumentBroker on disconnect: async saving in progress.");
+            else
+                docBroker->stop("docisdisconnected");
```
Ashod, line 2243: We need UnloadRequested here (#2930).

wsd/Storage.hpp
```diff
@@ -627,6 +627,14 @@ class WopiStorage : public StorageBase
     /// Total time taken for making WOPI calls during uploading.
     std::chrono::milliseconds getWopiSaveDuration() const { return _wopiSaveDuration; }
 
+    virtual AsyncUpload queryLocalFileToStorageAsyncUploadState() override
```
Ashod, line 630: I thought I had this function! Apparently I only had it in StorageBase, but never overridden it. Thanks for adding it.

Ashod (approved):
Great, thanks Szymon!

PR #2940, opened 2021-07-23
Improve Logging Last n Uno Commands in Online

gopi487krishna:
This commit fixes and improves the logging of unocommands from every
Child Session

Signed-off-by: Gopi Krishna Menon <krishnagopi487.github@outlook.com>
Change-Id: Icd32d90e7c01dc3ac080a010f5a70078d4bc495e

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -56,7 +56,8 @@ using Poco::URI;
 using namespace LOOLProtocol;
 
 bool ChildSession::NoCapsForKit = false;
-UnoCommandsRecorder ChildSession::unoCommandsRecorder;
+int ChildSession::CurrentPos = 0;
+std::shared_ptr<UnoCommandsRecorder>ChildSession::UnoCommandsRecorders[ChildSession::NUM_UNO_RECORDERS];
 
 namespace {
 
@@ -86,7 +87,7 @@ std::string formatUnoCommandInfo(const std::string& sessionId, const std::string
     unoCommandInfo.push_back(')');
     unoCommandInfo.append(" : ");
     unoCommandInfo.append(Util::eliminatePrefix(unoCommand,".uno:"));
-    unoCommandInfo.append(" - ");
+    unoCommandInfo.append("\t-\t");
     unoCommandInfo.append(recorded_time);
     unoCommandInfo.push_back('\n');
 
@@ -109,6 +110,8 @@ ChildSession::ChildSession(
     _isDocLoaded(false),
     _copyToClipboard(false)
 {
+    unoCommandsRecorder = std::make_shared<UnoCommandsRecorder>();
+    registerUnoCommandsRecorder(unoCommandsRecorder);
     SigUtil::registerUnoCommandInfoHandler(&ChildSession::dumpRecordedUnoCommands);
     LOG_INF("ChildSession ctor [" << getName() << "]. JailRoot: [" << _jailRoot << "].");
 }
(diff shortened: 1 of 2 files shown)


gopi487krishna:
@tml1024  Kindly review it when you get time Tor

kit/ChildSession.cpp
```diff
@@ -56,7 +56,8 @@ using Poco::URI;
 using namespace LOOLProtocol;
 
 bool ChildSession::NoCapsForKit = false;
-UnoCommandsRecorder ChildSession::unoCommandsRecorder;
+int ChildSession::CurrentPos = 0;
+std::shared_ptr<UnoCommandsRecorder>ChildSession::UnoCommandsRecorders[ChildSession::NUM_UNO_RECORDERS];
```
tml1024, line 60: Why did you introduce several UNO command recorders?  Why bother with shared_ptr here? Just more complication, more things that can be wrong. Michael said that we want this to be as minimal as possible.

A Kit processes handles the editing of one document and then dies, there is no need to try to make sure memory is deallocated cleanly after a ChildSession goes away or anything like that.

kit/ChildSession.cpp
```diff
@@ -868,44 +871,56 @@ bool ChildSession::getStatus(const char* /*buffer*/, int /*length*/)
...
     }
 }
 
 void UnoCommandsRecorder::addUnoCommandInfo(const std::string& unoCommandInfo)
 {
     // Create a new C string on heap with the contents of unoCommandInfo
-    char* command = new char[unoCommandInfo.length()+1];
-    strncpy(command, unoCommandInfo.c_str(), unoCommandInfo.length()+1);
+    char* newCommand = new char[unoCommandInfo.length()+1];
+    strncpy(newCommand, unoCommandInfo.c_str(), unoCommandInfo.length()+1);
```
tml1024, line 912: This is an anti-pattern, I think. strncpy() is hard to use meaningfully and correctly, I tend to avoid it. This is not a meaningful use of strncpy(). You know the length of the source string and you have allocated enough space for it on the previous line, so you could just use strcpy().

There is absolutely no extra "security" in using strncpy() as here. (For instance, if somebody later for some reason reduces the amount of space allocated with the new operator but forgets to change the last parameter to strncpy(), using strncpy() has no benefit, it will gladly overwrite the space allocated for the destination.)

And actually you could just use strdup(), and then deallocate with free().

tml1024 (changes requested):

kit/ChildSession.cpp
gopi487krishna: Each child session has its own uno commands recorder which records the uno commands that were fired by the user Tor

gopi487krishna: This output might make it clear

Image Pasted at 2021-7-23 01-00

gopi487krishna: Tor if they are not shared pointers then how will we manage their memory when they are out of scope? 

Example Scenario:

There are 50 recorders recording from 50 different child session. Suppose a new user comes in so the 51st recorder will be replaced at the 0 th position of bookkeeping array.  So what happens to the old recorder at 0th position and how it  will be deallocated? We cannot deallocate it with childsession destructor because the bookeeping array will be holding it as well. (There is no way to detect whether an address was free or not).

gopi487krishna: Using a shared pointer when we can ensure that atleast 1 reference is alive is safe inside a signal handler. We are not allocating or freeing anything. In fact we are not even modifying the control block in any way.

gopi487krishna: In summary what I want to tell is that there is a need for memory management in case of unorecorders or else  it will result in leaking memory

kit/ChildSession.cpp
gopi487krishna: Yes I will update it with strdup.

kit/ChildSession.cpp
gopi487krishna: image

PR #2949, opened 2021-07-25
OnScrollBy: Fix bug.

gokaysatir:
Code from old implementation was forgotten to update.

Change-Id: I4d5090ebf6ac5fa0e9746e542c47c0e682160ba5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/js/global.js
+++ b/loleaflet/js/global.js
@@ -3,7 +3,10 @@
 
 window.app = { // Shouldn't have any functions defined.
 	definitions: {}, // Class instances are created using definitions under this variable.
-	dpiScale: 1, roundedDpiScale: 1,
+	dpiScale: 1,
+	roundedDpiScale: 1,
+	twipsToPixels: 0, // Twips to pixels multiplier.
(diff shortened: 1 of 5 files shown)


kendy:
@gokaysatir: Unfortunately cannot merge it in the current form - when you start selecting in cell eg. H100, and leave the mouse pointer at the top (so that it scrolls towards the first row), when it hits the 1st row, the scrolling does not stop at the 1st row; instead it continues scrolling and the 1st row grows vertically.

kendy (changes requested):
See the comment. wrt. the problematic behavior.

eszkadev (approved):
I tested, now it works correctly

kendy:
@gokaysatir: So this improves the situation for me, but still does not fix the issue completely - when you let it scroll upwards, it scrolls, but does not extend the selection; only when you move the mouse pointer back into the sheet, it finally grows the selection.

merttumer:
It still needs some but certainly improves the issue. We need these patches for another problem, so merging them.

PR #2956, opened 2021-07-26
loleaflet: Process the reconnection in _onSocketClose in a timeout.

kendy:
We need to make sure that the message slurping processes the events
first, because there could have been a message like "close: idle" from
the server.

Without this, we'd immediately reconnect, because the "close: idle" was
not processed yet by the message slurping.

Co-author: Jan Holesovsky <kendy@collabora.com>
Change-Id: Id3e27164fb00de69f661ba7580075b36100a03c1

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -1388,10 +1388,18 @@ app.definitions.Socket = L.Class.extend({
 		this._map['wopi'].resetAppLoaded();
 		this._map.fire('docloaded', {status: false});
 
-		if (!this._reconnecting) {
-			this._reconnecting = true;
-			this._map._activate();
-		}
+		// We need to make sure that the message slurping processes the
+		// events first, because there could have been a message like
+		// "close: idle" from the server.
+		// Without the timeout, we'd immediately reconnect (because the
+		// "close: idle" was not processed yet).
+		var that = this;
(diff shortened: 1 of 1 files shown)


loleaflet/src/core/Socket.js
```diff
@@ -1389,8 +1389,15 @@ app.definitions.Socket = L.Class.extend({
...
-			this._map._activate();
+			// We need to make sure that the message slurping processes the
+			// events first, because there could have been a message like
+			// "close: idle" from the server.  Without this, we'd immediately
+			// reconnect (because the "close: idle" was not processed yet).
+			var that = this;
+			setTimeout(function () {
+				that._reconnecting = true;
+				that._map._activate();
+			}, 1 /* ms */);
```
Ashod, line 1400: Maybe we can give the server a bit more time, like 50ms or 500ms? Otherwise, looks great, thanks!

Ashod (commented):
Thanks Kendy!

loleaflet/src/core/Socket.js
kendy: As we talked - I believe the problem is JS only; if the server needed more time, it would have been a problem even in the pre-message-slurping era - so I hope we are fine here.

Ashod (approved):
Looks great, thanks @kendy.

PR #2969, opened 2021-07-27
Feedback: Do now show dialog if welcome is present

pedropintosilva:
- Do not bother user with multiple dialog at the loading time
- Add additional timer after the welcome dialog is closed before
calling feedback again

Change-Id: Ib432e13205b1b3bef4dfa7ed8752a6561a4f0466


eszkadev (approved):
code looks ok, but I didin't test as I don't know how to make 2 dialogs appear :)

Issue #2812, opened 2021-07-08
Mobile: do not use span inside of a button element


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

On master we have now a bug where buttons have an additional unnecessary span, which in turn doesn't get properly styled:

![image](https://user-images.githubusercontent.com/65948705/124917511-39394180-dff4-11eb-93a0-efbda4dd0921.png)

_on distro/collabora/co-6-4 it's ok_

Arkadipta-Chatterjee:
Can i work on this please?

pedropintosilva:
unfortunately this seems a bit more complex than what I originally thought https://github.com/CollaboraOnline/online/pull/2974

Darshan-upadhyay1110:
It's fixed in latest version.

COOLWSD version: 24.04.5.2
(git hash: 
df56d78c)
LOKit version: Collabora OfficeDev 24.04.5.2
(git hash: 
b57fd97)
Served by: Ubuntu 23.10
Server ID: 0dc2679d

PR #2974, opened 2021-07-27
Mobile: do not use span inside of a button

pedropintosilva:
Change-Id: I88f1b97939c59927686bd72ea295f87abffcbe29

### DIFF
--- a/loleaflet/src/control/Control.JSDialogBuilder.js
+++ b/loleaflet/src/control/Control.JSDialogBuilder.js
@@ -1341,15 +1341,19 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		var wrapper = L.DomUtil.create('div', '', parentContainer); // need for freemium overlay
 		var pushbutton = L.DomUtil.create('button', 'ui-pushbutton ' + builder.options.cssClass, wrapper);
 		pushbutton.id = data.id;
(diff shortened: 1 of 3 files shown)


loleaflet/src/control/Control.JSDialogBuilder.js
```diff
@@ -1347,9 +1347,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			image.src = data.image;
```
eszkadev, line 1347: this image inserted into pushbutton will be overwritten by innerHTML call later I think

eszkadev (changes requested):

loleaflet/src/control/Control.JSDialogBuilder.js
eszkadev: also please check if sidebar popups which have buttons with images are correctly shown

pedropintosilva:
@Rash419 could you please take it over?

- I have fixed https://github.com/CollaboraOnline/online/issues/2812
- But now we need to make sure that we account for the cases Szymon described above

Rash419:
> @Rash419 could you please take it over?
> 
> * I have fixed [Mobile: do not use span inside of a button element #2812](https://github.com/CollaboraOnline/online/issues/2812)
> * But now we need to make sure that we account for the cases Szymon described above
> 
> feel free to add commits to this PR

Yes syzmon is right if there will a button with image and text both the image will be replaced with text but I could find any instance where both is there in button @pedropintosilva

eszkadev (approved):
looks good

PR #2985, opened 2021-07-28
jsdialog: fix buttons alignment

eszkadev:
before fix buttons in jsdialogs were positioned vertically (ok and cancel buttons)

### DIFF
--- a/loleaflet/css/jsdialogs.css
+++ b/loleaflet/css/jsdialogs.css
@@ -345,15 +345,7 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 	cursor: auto;
 }
 
-/* button box */
-
-.jsdialog.ui-button-box-right {
-	float: right;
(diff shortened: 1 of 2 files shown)


loleaflet/css/jsdialogs.css
```diff
@@ -345,15 +345,7 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 	cursor: auto;
 }
 
-/* button box */
-
-.jsdialog.ui-button-box-right {
```
pedropintosilva, line 350: Avoid repeating declaration for .ui-button-box-right in multiple files

pedropintosilva (approved):

loleaflet/css/jsdialogs.css
eszkadev: @pedropintosilva will you push it also for master (my change is there already)

Issue #2846, opened 2021-07-12
Error parsing HTTP status line when there is no reason phrase


avialluri:
If there is no reason phrase (OK) in HTTP status line the following error is thrown during the call to CheckFileInfo:
wsd-17022-17044 2021-07-12 16:58:59.290312 [ docbroker_001 ] ERR  StatusLine::parse: missing line break| net/HttpRequest.cpp:289
This worked in 6.4.9 - 0b3211b , but is failing with 6.4.10 version.

**To Reproduce**
1. Access a fileId . This will trigger the call to CheckFileInfo
2. Make sure the response does not have the reason phrase : i.e only
     HTTP/1.1 200
     (reason phrase OK is missing above..which is what tomcat 9.x + will return)
3. HttpRequest.cpp fails with the above error message. This worked in 6.4.9 - 0b3211b , but is failing with 6.4.10 version.
4. 
wsd-17022-17044 2021-07-12 16:58:59.290312 [ docbroker_001 ] ERR  StatusLine::parse: missing line break| net/HttpRequest.cpp:289
wsd-17022-17044 2021-07-12 16:58:59.290444 [ docbroker_001 ] ERR  WOPI::CheckFileInfo (31ms) failed or no valid JSON payload returned. Access denied. Original response: [].| wsd/Storage.cpp:786

**Expected behavior**
CheckFileInfo gets the json data and able to parse and continue succesfully

**Actual behavior**
CheckFileInfo fails with the above error message (it does have json payload but fails due to the HTTP status line parse issue)

**Screenshots**

**Desktop (please complete the following information)**
 - OS: amazon linux 2

---------- Headers in response from CheckFileInfo:
HTTP/1.1 200
Set-Cookie: JSESSIONID=5308B30917BCD4CF201FCB495FA59282; Path=/xyz; HttpOnly
Content-Type: application/json;charset=UTF-8
Transfer-Encoding: chunked
Date: Mon, 12 Jul 2021 17:49:42 GMT

jiandanfeng:
I am facing the same issue with 6.4.10 version.

PR #2991, opened 2021-07-29
wsd: http: reason phrase is optional

Ashod:
* Resolves: #2846 

### Summary
Not having reason phrase in the http status is not an error.

### Checklist


vmiklos (approved):
Looks reasonable to me, but co-6-4 is kind of frozen at the moment, so please do ask @kendy before actually merging this.

vmiklos (approved):

PR #2993, opened 2021-07-29
Don't try to set empty hostname for SNI

eszkadev:
to reduce logging warnings


kendy (approved):
Looks good, thank you!

PR #2994, opened 2021-07-29
Handle poll results in SocketPoll in a round-robin order

tml1024:
We don't want to always start with the most recently added socket.

Change-Id: Ic4b4bf6c19c5d119e6e6f9b398789a4c77b47a10

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -186,6 +186,7 @@ namespace {
 
 SocketPoll::SocketPoll(const std::string& threadName)
     : _name(threadName),
+      _pollStartIndex(0),
       _stop(false),
       _threadStarted(0),
       _threadFinished(false),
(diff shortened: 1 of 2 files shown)


tml1024:
(Tentative patch, do not merge just to be "helpful". But I do want the checks to be run so I can't mark it as "draft".)

tml1024:
Ah, this crashes right away even in "make run"... truly work in progress;)

tml1024:
OK, now it even seems to work as intended.

tml1024:
Changed commit message and comments to match reality better. It is of course not the "poll order" that this change modifies, but the order in which we handle the sockets that the poll result indicated.

mmeeks:
Lets get this into master but not co-6-4 =) Thanks.

Issue #2910, opened 2021-07-19
Admin console broken in loolwsd 6.4.10 (with Apache 2.4.48, but not with 2.4.38)


thkoch2001:
The bug is described here:
https://forum.collaboraonline.com/t/websocket-connection-to-admin-console-adminws-is-failing/475

Duplicate issue:
https://forum.collaboraonline.com/t/code-cannot-launch-admin-console-after-update/477

TL;DR:
- Error message: “Server has been shut down; please reload the page.”
- websocket connection failing
- log: `Admin::handleInitialRequest bad request| wsd/Admin.cpp:378`
- This if should be true but isn't:
  `if (request.find("Upgrade") != request.end() && Util::iequal(request["Upgrade"], "websocket"))`
- This commit touched the line recently: https://github.com/CollaboraOnline/online/commit/54a10729f3b5ac5913a051a0b810214e30a5b83a

thkoch2001:
This expression looks unnecessary brittle and complicated:

    request.find("Upgrade") != request.end()

Why not:

    request.has("Upgrade")
?

kendy:
@thkoch2001: Thank you for the report!  Please does it happen also with the 6.4.10-6 that has been released 2 days ago?  Thank you in advance!

thkoch2001:
Yes, I updated yesterday and confirmed the issue also with 6.4.10-6.

thebearon:
@thkoch2001 would you be able to capture the exact HTTP request that is sent when opening the admin console? I didn't see an error with a terminating reverse proxy setup, but the request could possibly be analyzed.

thkoch2001:
I confirmed with Apache module [dumpio](https://httpd.apache.org/docs/current/mod/mod_dumpio.html) that the Apache server receives this request:

```
GET /lool/adminws/ HTTP/1.1\r\n
Host: collaboraonline.REDACTED\r\n
Connection: Upgrade\r\n
Pragma: no-cache\r\n
Cache-Control: no-cache\r\n
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/REDACTED (KHTML, like Gecko) Chrome/REDACTED Safari/REDACTED\r\n
Upgrade: websocket\r\n
Origin: https://collaboraonline.REDACTED\r\n
Sec-WebSocket-Version: 13\r\n
Accept-Encoding: gzip, deflate, br\r\n
[...]

PR #3010, opened 2021-07-30
use Request.has instead of .find and .end

thkoch2001:
I noticed this while looking into issue #2910 but it is not a fix for this issue.


pedropintosilva:
@thkoch2001 thanks (maybe @mrkara can review this) maybe meanwhile could you please sign off your commit? :

Ashod (approved):
Thanks @thkoch2001, looks good.

PR #3012, opened 2021-07-30
wsd: log stopping-although-modified as warning

Ashod:
This log entry can be misleading on its own.
When shutting down, the modified flag is not reset,
and that in itself is not a cause for concern.
So this log entry is seen much more often than we
expect. If we truely fail to save/upload, there
will be other errors that will be the tell-tale
signs of a serious issue. So it's reasonable to
downgrade this error to a warning (and watch for
other errors).

Change-Id: Ic879c3fbac887ed5b64bdd1a31aa38bcfd863850


kendy (approved):
Yes, let's do that, thank you!

PR #3022, opened 2021-08-03
Fix view jump flicker on docload in Calc (co-6-4)

dennisfrancis:
* Resolves: # 
* Target version: co-6-4

### Summary

Fixes view area jump (looks like a flicker) during document load in Calc.
Note that without the fix, the view jump is apparent only if last edited cell in the sheet is not in a view containing A1. 
So try something like A250.
The root cause of the jump is that the initial view area always the default (ie. starting cell A1) unless the client gets a cell cursor message and it adjusts the view such that it contains it. In this patch set we request for cell cursor early on during CalcTileLayer initialization. We avoid any canvas painting (except for clearing of the canvas during resizes) till we get the first cell cursor.

### Checklist


gokaysatir (approved):

dennisfrancis:
Fixed a merge conflict after zoom-animation PR got merged.

PR #3031, opened 2021-08-04
kit: improve StringVector usage in ChildSession::getClipboard()

vmiklos:
It returns a temporary.

Change-Id: I5a9df90a4e7dda05cc387e5dc11842ea8e237504


mmeeks (approved):
Ah - that looks very helpful =)

PR #3046, opened 2021-08-09
co-6-4: loleaflet: avoid reload page if it is integrated with WOPI

hcvcastro:
Change-Id: I3c6ba97c04c001affbc62b039290d2a7a8c8f91a

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -497,7 +497,9 @@ app.definitions.Socket = L.Class.extend({
 						this._map.uiManager.showSnackbar(reloadMessage, _('RELOAD'), function() {window.location.reload();});
 						setTimeout(window.location.reload, 5000);
 					} else {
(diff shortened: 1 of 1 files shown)


mmeeks:
Hmm - interesting; are you sure this is the right fix? what is it about WOPI that helps here?
I would assume that we want to do the reload command but without the dialog showing up - if we are doing that already - to refresh just that iframe (?) or ... ?

hcvcastro:
No explicit dialog to reload when WOPI is used, the auto-reconnecting will refresh the document.

mmeeks:
Surely the purpose of the reload is to refresh the javascript though - since when the server version changes the JS can too and we badly don't want to leave old JS running against a new loolwsd - hence the reload.

So - we should do an automatic reload in this case, as/when we detect a different serverid I think.

hcvcastro:
I see, working on it ...

PR #3059, opened 2021-08-10
loleaflet: comment improvement

Rash419:
Change-Id: I8b50a889542193591eb2ed97924e4f6a07a4aba2

* Target version: co-6-4

### Summary
image

### TODO:
@pedropintosilva  apparently we need to add padding to the righ-side also
image


pedropintosilva:
> TODO: @pedropintosilva apparently we need to add padding to the righ-side also

Thanks I have already removed all that padding from that side

pedropintosilva:
didn't have permission to push to this PR plus it seems that your private branch already has new stuff above it, so I have create a new one : https://github.com/CollaboraOnline/online/pull/3063

PR #3064, opened 2021-08-11
log: avoid logging after shutdown on all log levels

vmiklos:
This can happen on an assertion failure, and causes noise while looking
for real memory errors.

Also disable an unstable test, it's not clear that it passes depending
on how loaded the machine is (just increasing timeouts doesn't seem to
help).

Change-Id: I049bd0a06f41e2d43702ec12a2c35944bc5200d8

### DIFF
--- a/common/Log.hpp
+++ b/common/Log.hpp
@@ -299,7 +299,7 @@ namespace Log
     do                                                                                             \
     {                                                                                              \
         auto& log_ = Log::logger();                                                                \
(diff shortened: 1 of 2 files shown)


vmiklos:
FreeBSD fails with:

> fetch: http://arrowd.name/libreoffice-include.tar.xz: Connection refused

Which looks unrelated, so ignoring that failure. The rest is green.

PR #3072, opened 2021-08-13
loleaflet: store recent and custom colors locally

Rash419:
Change-Id: I85fe51d4c7a839b9743df6c739e4d0a02daacf40

* Target version: master


lpranam (approved):
Looks good. Works fine. 

Though one suggestion for future work, when we use a colour that already exists in the recent colours, we should move it back to the first position instead of keeping the recent panel unchanged. But For now good to go.

PR #3073, opened 2021-08-13
fix calc's incorrect sheet switch

dennisfrancis:
'updateparts' event must be fired only after requesting new part to
core, else both tiles and sheet-geometry will be wrong after the user
triggers sheet switch.

The move of 'updateparts' firing code was done in
commit 6da00622d08060b45452f0e6cbbf6077b117503b
  Fix PDF view: Switching pages using buttons

but that move became unnecessary after the next PDF related commit
commit 9ca5d68ba21fbaf94049f85b1730390573b80184
  Make page buttons scrollby view height instead instead of switching the page directly

Change-Id: Ib806e31965ecada5305f4ea647f1de6236b27486

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/loleaflet/src/control/Parts.js
+++ b/loleaflet/src/control/Parts.js
@@ -42,13 +42,6 @@ L.Map.include({
 			return;
 		}
 
-		this.fire('updateparts', {
-			selectedPart: docLayer._selectedPart,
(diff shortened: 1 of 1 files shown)


merttumer (approved):
Looks good and makes sense. Thank you

PR #3078, opened 2021-08-13
Unit: add new 'Tool' type of unit test handler.

mmeeks:
Dummy, but useful for eg. fuzzing.

Change-Id: I9819cda898a7bf1c0fba3d1fccbb63b16955f01d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Pushed with DCO =)

PR #3081, opened 2021-08-15
wsd: autosave using the last editing session

Ashod:
Change-Id: I556b7b49b2d1abc0e61ea7ecaa8e5f9406c408f3

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -180,6 +180,7 @@ DocumentBroker::DocumentBroker(ChildType type,
     _stop(false),
     _closeReason("stopped"),
     _lockCtx(new LockContext()),
+    _lastEditingSessionId(),
     _tileVersion(0),
     _debugRenderedTileCount(0),
     _wopiLoadDuration(0),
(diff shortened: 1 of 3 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -2772,6 +2776,12 @@ bool DocumentBroker::forwardToChild(const std::string& viewId, const std::string
     const auto it = _sessions.find(viewId);
     if (it != _sessions.end())
     {
+        if (!it->second->isReadOnly() && it->second->isViewLoaded() &&
```
mmeeks, line 2779: Not a fan of this as a location for this work. I would expect to hook into DocumentBroker::updateLastActivityTime - and get the client-session to pass its own const &view-id into that. I would expect to compare with what we already have, so as to avoid duplicating the std::string on the heap very frequently.

With that fix - the approach looks ok to me.

Thanks !

mmeeks (changes requested):
Needs a tweak or two I think.

wsd/DocumentBroker.cpp
Ashod: > Not a fan of this as a location for this work. I would expect to hook into DocumentBroker::updateLastActivityTime - and get the client-session to pass its own const &view-id into that.

Unfortunately, the `AdminModel` also calls `DocumentBroker::updateLastActivityTime` (but it doesn't have a session to pass).

I can add another similar helper for this, so we can do it as you describe, but without changing `updateLastActivityTime`.

> I would expect to compare with what we already have, so as to avoid duplicating the std::string on the heap very frequently.

Sure.

mmeeks:
Looks great to me - thanks Ash ! =) lets get it in as/when CI is happy.

PR #3082, opened 2021-08-16
log: avoid logging after shutdown on all log levels

lpranam:
This can happen on an assertion failure, and causes noise while looking
for real memory errors.

Also disable an unstable test, it's not clear that it passes depending
on how loaded the machine is (just increasing timeouts doesn't seem to
help).

Change-Id: I049bd0a06f41e2d43702ec12a2c35944bc5200d8


mmeeks:
Thanks Pranam =)

PR #3093, opened 2021-08-18
cypress: make annotation specs compaitable with NB

Rash419:
Change-Id: I76f70262c13896ccb5bd41f68ac12b337db7f7c2

* Target version: master 

### Summary
I have removed one test from comment scrolling because the behaviour of the comment scrolling has been changed 
Before the comments were scrolling on the individual basics per slide but now they show up in group per slides
hence we don't need that test

### DIFF
--- a/cypress_test/integration_tests/common/desktop_helper.js
+++ b/cypress_test/integration_tests/common/desktop_helper.js
@@ -270,17 +270,47 @@ function deleteImage() {
 		.should('not.exist');
 }
 
-function insertMultipleComment(numberOfComments) {
-	numberOfComments = numberOfComments || 1;
+function insertMultipleComment(docType, numberOfComments = 1) {
+	var mode = Cypress.env('USER_INTERFACE');
+
+	cy.get('#toolbar-up .w2ui-scroll-right').then($button => {
+		if ($button.is(':visible'))	{
+			$button.click();
+		}
+	});
+
+	if (mode === 'notebookbar') {
+		cy.wait(500);
+
+		cy.get('#Insert-tab-label').then($button => {
+			if (!$button.hasClass('selected')) {
+				$button.click();
+			}
+		});
+	}
+
+	if	(docType === 'writer') {
+		cy.get('#toolbar-up .w2ui-scroll-right').click();
+	}
+
 	for (var n=0;n<numberOfComments;n++) {
 
-		cy.get('#menu-insert').click().get('#menu-insertcomment').click();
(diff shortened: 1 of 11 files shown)


cypress_test/integration_tests/common/desktop_helper.js
```diff
@@ -270,18 +270,41 @@ function deleteImage() {
...
+function insertMultipleComment(docType, numberOfComments = 1, isFirstCall = true) {
+	if (isFirstCall) {
+		var mode = Cypress.env('USER_INTERFACE');
+
+		cy.get('#toolbar-up .w2ui-scroll-right').click();
+
+		if (mode === 'notebookbar') {
+			cy.get('#Insert-tab-label').click();
+
+			(docType === 'writer') ?
```
lpranam, line 284: a simple if is better in this case

cypress_test/integration_tests/common/desktop_helper.js
```diff
@@ -270,18 +270,41 @@ function deleteImage() {
 		.should('not.exist');
 }
 
-function insertMultipleComment(numberOfComments) {
-	numberOfComments = numberOfComments || 1;
+// we don't need to scroll if toolbar is already scrolled
+// for the mentioned purpose isFirstCall parameter is used
+function insertMultipleComment(docType, numberOfComments = 1, isFirstCall = true) {
```
lpranam, line 275: why to have an extra parameter? instead you could just check if the 'scroll-right' is available then do it straight away!?

cypress_test/integration_tests/desktop/impress/annotation_spec.js
```diff
@@ -13,6 +13,8 @@ describe('Annotation Tests', function() {
 
 		if (Cypress.env('INTEGRATION') === 'nextcloud') {
 			desktopHelper.hideSidebarIfVisible();
+		} else if (Cypress.env('USER_INTERFACE') === 'notebookbar') {
+			cy.get('#ModifyPage').click();
```
lpranam, line 17: This seems like a bug can you confirm please. Modify page should not work as sidebar button.

lpranam (changes requested):

cypress_test/integration_tests/desktop/impress/annotation_spec.js
Rash419: I don't think it is a bug : )

lpranam (approved):

PR #3095, opened 2021-08-18
Add fourth number to LOOLWSD versioning

NickWingate:
Signed-off-by: NickWingate <nicholas.wingate03@gmail.com>
Change-Id: Ib7db15dfafa06c4ddcdef28a8831a91bb847a389

* Resolves: # 
* Target version: master 

### Summary

Allows for a fourth version number to be specified for LOOLWSD in configure.ac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
 
 AC_PREREQ([2.63])
 
-AC_INIT([loolwsd], [21.06.1], [https://github.com/CollaboraOnline/online/issues], [loolwsd], [https://collaboraonline.github.io/])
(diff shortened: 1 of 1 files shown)


mmeeks:
Nice to see your first 'm4' commit - well worth reading the health warning here: https://www.gnu.org/software/m4/manual/m4.html#Intro =)

Andras - I'll leave a final review up to you; will it break our shell - tools / tooling elsewhere ? Quite possibly we don't want it in some places, but do in some others - like help->about where we need it.

timar (approved):

PR #3098, opened 2021-08-18
leaflet: show table markers when table has no entry

lpranam:
Change-Id: Idcf62a2f787d098b1fcc78b57b4d07f9de685b31

* Target version: master 

### Summary
Problem:
First row/column are treated specially as headings,
that's why heading are not sent through the entry.

Due to that if table has only single column or row,
entry will be empty and marker will never appear

### Checklist


mikekaganski (approved):
This is fine together with https://git.libreoffice.org/core/+/549dd4b3849435b4f4331e33fe056fb9832b6c6c - thanks!

Issue #2685, opened 2021-06-25
Remove isInternetExplorer code paths


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @mmeeks 

The master branch of online will no longer support Internet Explorer, so we should remove lots of legacy code that relates to this.

### Code pointers

git grep -10 isInternetExplorer

### Additional notes, tips & tricks

Note - it may be a good idea to do this as several patches. First by having a last use of this logic:

loleaflet/js/global.js: var isInternetExplorer = (navigator.userAgent.toLowerCase().indexOf('msie') != -1 ||
loleaflet/js/global.js-                 navigator.userAgent.toLowerCase().indexOf('trident') != -1);

And warning the user that this browser is not supported with a suitable dialog.

Then as a follow-on we should dung out things like:

loleaflet/src/map/Clipboard.js- compatRemoveNode: function(node) {
loleaflet/src/map/Clipboard.js:         if (L.Browser.isInternetExplorer)
loleaflet/src/map/Clipboard.js-                 node.removeNode(true);
loleaflet/src/map/Clipboard.js-         else // standard
loleaflet/src/map/Clipboard.js-                 node.parentNode.removeChild(node);
loleaflet/src/map/Clipboard.js- },

and in-line the 'removeChild' method everywhere instead carefully.

Please separate the Clipboard.js cleanup to its own patch - since this is very involved code.

PR #3101, opened 2021-08-18
Remove isInternetExplorer code paths

NickWingate:
* Resolves:  [#2685](https://github.com/CollaboraOnline/online/issues/2685)
* Target version: master 

### Summary
 Removes any code referencing isInternetExplorer, excluding a popup in `/loleaflet/src/main.js
`
In `Clipboard.js` also inlines `compatRemoveNode` as it only exists for to handle IE.

### TODO

- [ ] Confirm if the code for the alert is in the correct place (currently in `main.js`) I don't know if this is the responsibility of `main.js` or if should be another file
- [ ] The alert could also be a confirm dialog that redirects to a page with more information about supported browsers/IE

### Checklist

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -386,16 +386,10 @@ app.definitions.Socket = L.Class.extend({
 		e.imgBytes = new Uint8Array(e.data);
 
 		// search for the first newline which marks the end of the message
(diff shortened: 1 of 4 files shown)


lpranam (approved):
Looks good. let's wait for CI to pass.

PR #3104, opened 2021-08-19
loleaflet: when user select existing color in colorPicker it comes on first position in recentRow

Rash419:
Change-Id: I200789fc5aa9cd5a6fc8a0bc311cbe71dde33e42

* Target version: master


lpranam (approved):
Looks good

Issue #2816, opened 2021-07-08
Tablet: toggleMenubar has wrong icon


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

1. Use tablet or test it with F12 -> device -> tablet
2. The icon of the farthest right (hamburger) works as a toggle to hide/show the top menubar (File, Edit, etc..)

![image](https://user-images.githubusercontent.com/65948705/124927119-8fab7d80-dffe-11eb-8111-0e17c2b8a7f1.png)

Use the same icon as in the desktop:
 `.w2ui-icon.fold` (`background: url('images/fold.svg') no-repeat center center / 19px !important`) visible in `toolbar.css`

ViAsmit:
I would like to work on this issue. May I know which folder should I look for the source code of this page.

Arkadipta-Chatterjee:
Can i work on this issue?

kendy:
@pedropintosilva: Is this actually fixed by that commit?

shrikanta8:
@pedropintosilva is this issue being assigned to someone?

pedropintosilva:
@shrikanta8 : Are you still working or intend to work on this?

Ayushbajpai19:
I would like to take this issue for hacktoberfest

callmetushar123:
@pedropintosilva Hello since no one is working on it please assign it to me. I will fix it :D

pedropintosilva:
Fixed with https://github.com/callmetushar123/online/commit/f762ee0904b0efafd693454e7f7cf8169b9f04f2, thanks! :partying_face:

callmetushar123:
@pedropintosilva i wanted to work on some medium level issue. Can you please be my mentor?

PR #3105, opened 2021-08-19
Fixes Tablet: toggleMenubar has wrong icon #2816

pedropintosilva:
Signed-off-by: ASMIT VIMAL <asmit.vml.962@gmail.com>
Change-Id: Id0b3e0ac7364792d8f3ed18ae1b02c5f032cd2bd

### DIFF
--- a/loleaflet/css/toolbar-mobile.css
+++ b/loleaflet/css/toolbar-mobile.css
@@ -12,3 +12,5 @@
 	border-top: none;
 	z-index: auto;
 }
+
+.w2ui-icon.fold{ background: url('images/fold.svg') no-repeat center center / 19px !important; opacity: 0.8; filter: alpha(opacity=80);}


pedropintosilva:
@ViAsmit I'm afraid this didn't fix it

pedropintosilva:
There is no need to repeat the CSS rule, we are already using the image , but the JS side is calling the wrong file. How to fix it?
1. find where the top toolbar is being created `loleaflet/src/control/Control.TopToolbar.js`
2. Search for one of the ids you see (when inspecting the html on a browser using responsive mode = tablet) for example `hamburger-tablet`
3. You will find it in the line `265` of `loleaflet/src/control/Control.TopToolbar.js`
4. change `{type: 'button',  id: 'hamburger-tablet',  img: 'hamburger'` to `{type: 'button',  id: 'hamburger-tablet',  img: 'hamburger'`

PR #3112, opened 2021-08-20
fix: loleaflet: cursor behaviour when mouseleaves from verticalScrollbar

Rash419:
Change-Id: I26ddd343309492d6e930b8d4b232d932804eb269

* Target version: master


gokaysatir (approved):
Thanks for the fix :)

lpranam (approved):

PR #3123, opened 2021-08-23
loleaflet: add demo for opening multiple docs and switching between them

### LINKED ISSUE #3122: Add new example in the documentation with multiple loleaflet iframes opened at the same time
**Is your feature request related to a problem? Please describe.**
It's not a problem, but more of a general use case in various cloud-like portals. A common use case in such portals is to have multiple documents open at the same time, perhaps even copy-pasting and then editing from one to another.

**Describe the solution you'd like**
A `html` demo page that showcases this common use case in my opinion. The user should be able to open several documents at the same time and navigate between them.

**Describe alternatives you've considered**
-

**Additional context**
-

alessioalex-1and1:
closes #3122

Change-Id: Ibf4823b03cb70812c84d9125f4092b87bec6613b

* Resolves: # 
* Target version: master 

### Summary

open document prompt

demo multiple docs

### TODO

- [ ] ...

### Checklist


pedropintosilva:
I think the cypress tests need to be adjusted here, and I wonder what would be the best plan of action here, maybe somehow to adjust `cypress_test/integration_tests/multiuser` ? what do you think @Rash419 ?

Rash419:
> I think the cypress tests need to be adjusted here, and I wonder what would be the best plan of action here, maybe somehow to adjust `cypress_test/integration_tests/multiuser` ? what do you think @Rash419 ?

@pedropintosilva  I don't think it is anything related to multiuser tests.

pedropintosilva:
@Rash419 oh right where did I have my head.. you are right (sorry for the noise) and thanks for checking that :+1:

Issue #2160, opened 2021-04-26
loleaflet-api.html: Copy documentation into source code as comments


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**

### Detailed description and rationale
We have documentation in `loleaflet/loleaflet-api.html` it would be great to transfer those bits directly to source code via comments in front of declarations for self documenting purposes. Please start small, picking one option at the time per PR

##### 1. Pick one
`tileWidthTwips` is documented in `loleaflet/loleaflet-api.html`:
![image](https://user-images.githubusercontent.com/65948705/116096540-abfb7780-a6a9-11eb-9963-71924e8ad599.png)

##### 2. Find where that is declared in the code
Search for `tileWidthTwips` using terminal or your favourite IDE and hunt where is firstly declared. In this case is in `loleaflet/src/map/Map.js` `:40`

##### 3. Add comment

![image](https://user-images.githubusercontent.com/65948705/116098283-3d1f1e00-a6ab-11eb-8ad4-ddc3391617f6.png)

##### 4. Push Pull request
In your commit message please follow this `Copy loleaflet-api.html into code: [option/command/etc]`. So in our example that would be `Copy loleaflet-api.html into code: tileWidthTwips`

### Code pointers

 where: `loleaflet/loleaflet-api.html`

##### Options
- [ ] doc
- [ ] server
- [ ] webserver
- [ ] permission
- [ ] timestamp
- [ ] documentContainer
- [ ] toolbarContainer
- [ ] renderingOptions
- [ ] print
- [ ] autoFitWidth
- [ ] zoom
- [x] tileWidthTwips https://github.com/CollaboraOnline/online/pull/2749
- [x] tileHeightTwips https://github.com/CollaboraOnline/online/pull/2205
- [x] defaultZoom https://github.com/CollaboraOnline/online/pull/2830
- [x] cursorURL https://github.com/CollaboraOnline/online/pull/3121

##### General Methods
- [ ] search()
- [ ] highlightAll()
- [ ] setPermission()
- [ ] getDocSize()
- [ ] getDocType()
- [ ] getPageSizes()
- [ ] scroll()
- [ ] scrollDown()
- [ ] scrollRight()
- [ ] scrollTop()
- [ ] scrollLeft()
- [ ] scrollOffset()
- [ ] getPreview()
- [ ] getCustomPreview()
- [ ] removePreviewUpdate()
- [ ] fitWidthZoom()

[...]

PR #3126, opened 2021-08-23
Copy documentation into source code as comments of remaining Options …

siddharth25pandey:
* Resolves: # [loleaflet-api.html: Copy documentation into source code as comments #2160](https://github.com/CollaboraOnline/online/issues/2160)
* Target version: master 

* Added the comments of the remaining  **Options**  from the documentation
* Removed the `toolbarContainer` from **loleaflet-api.html** as per reference of commit id [4358d2a04ec5465d55e336aa826788c17df96d71](https://github.com/CollaboraOnline/online/commit/4358d2a04ec5465d55e336aa826788c17df96d71)


siddharth25pandey:
@pedropintosilva Please review, Thanks!

loleaflet/src/control/Control.Menubar.js
```diff
@@ -1522,6 +1522,8 @@ L.Control.Menubar = L.Control.extend({
 		} else if (id === 'shareas') {
 			this._map.openShare();
 		} else if (id === 'print') {
+			// print
+			// Whether the print handler is active (for Chrome).
```
pedropintosilva, line 1526: This seems to be the wrong place to add those since we are just checking the id and then calling `this._map.print();` we need to find where that is declared in the code. 


Option | Type | Default | Description
-- | -- | -- | --
print | Boolean | true | Whether the print handler is active (for Chrome).

notice that we are talking about a boolean option

loleaflet/src/main.js
```diff
@@ -26,6 +26,8 @@ if (reuseCookies !== '') {
 
 var filePath = getParameterByName('file_path');
 var permission = getParameterByName('permission') || 'edit';
+// permission
+// The document's permission.
```
pedropintosilva, line 30: This should be placed where permission **option** is declared in the first place

loleaflet/src/map/Map.js
```diff
@@ -109,6 +109,9 @@ L.Map = L.Evented.extend({
 		Cursor.imagePath = options.cursorURL;
 
 		if (options.webserver === undefined) {
+			// webserver
+			// The webserver access to hosting loolwsd. Normally it is derived from 'server'
+			// but can be overridden with an own value in case of proxying. Example: http://localhost:9980
```
pedropintosilva, line 114: maybe in this case adding it before the if would be better (so in the case someone changes that condition the comments will not get in the way

pedropintosilva (changes requested):
@siddharth25pandey : Could you please add the missing change id ? (that's why one of the tests are failing )

CC: @lpranam could you please review this?

loleaflet/src/main.js
siddharth25pandey: @pedropintosilva It was added after line number **28**, this is the place where the **declaration** of permission was done

loleaflet/src/map/Map.js
siddharth25pandey: Okay will do it

loleaflet/src/control/Control.Menubar.js
siddharth25pandey: Let me know where it is declared first

siddharth25pandey:
> @siddharth25pandey : Could you please add the missing change id ? (that's why one of the tests are failing )
> 
> CC: @lpranam could you please review this?

@pedropintosilva I already added the missing commit id of `toolbarContainer` as a link in the PR Template. Let me know if I am thinking wrong

loleaflet/src/map/Map.js
siddharth25pandey: @pedropintosilva Done

lpranam (changes requested):
Hello, Thank you for taking the interest and time to contribute in collabora online.

I think all the comments you have added are for the variables/objects which are pretty self-explanatory by their names themselves. So adding these comments makes it redundant. So I would be happier if we don't add these particular comments. 

On the other hand nice catch for finding absolute API and removing it from the docs. 

So just remove the comments and I will approve and merge the PR :)

On the other hand for your future work, when you add comments, ideally they are supposed to be above the element they are related to. Sometimes if the comment is short you can also add it inline. 

thanks.

siddharth25pandey:
> Hello, Thank you for taking the interest and time to contribute in collabora online.
> 
> I think all the comments you have added are for the variables/objects which are pretty self-explanatory by their names themselves. So adding these comments makes it redundant. So I would be happier if we don't add these particular comments.
> 
> On the other hand nice catch for finding absolute API and removing it from the docs.
> 
> So just remove the comments and I will approve and merge the PR :)
> 
> On the other hand for your future work, when you add comments, ideally they are supposed to be above the element they are related to. Sometimes if the comment is short you can also add it inline.
> 
> thanks.

@lpranam Thanks for the review, I am adding the comments as directed by @pedropintosilva in [issue](https://github.com/CollaboraOnline/online/issues/2160). Please go through it and let me know the same.
Thanks!

PR #3132, opened 2021-08-25
Remove more leftover files taking space on the device in the iOS app

tml1024:
Remove the app's copy of the document after core has saved it and we
have stored the edited version at its real location.

Remove possible leftover tile bitmaps when the app starts.

Change-Id: Ibc41be38c2cfb689c532640d148116bc06a248ab

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -544,7 +544,7 @@ namespace RenderTiles
             const int offsetX = positionX * pixelWidth;
             const int offsetY = positionY * pixelHeight;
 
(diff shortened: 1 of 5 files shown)


mmeeks (approved):
Looks fine to me; though I don't see it solving the big, fast leak - is there some place I can try it ?

tml1024:
To test it you would have to build the app yourself.

tml1024:
But sure, this change doesn't fix tile .bmp leaks, it just removes old leftover tiles the next time the app process starts.

PR #3151, opened 2021-08-27
Remove firstline assignment from ctor

NickWingate:
* Resolves: #
* Target version: master 

### Summary

Move _firstLine assignment from ctor to firstLine() method, therefore only making the call to generate _firstLine when necessary. 

### Todo

- [x] Change return type of firstLine to make it pass by reference not value
- [x] Check where rewriteDataBody() is called and if _data is changed, to see the purpose of Message.hpp:127 and if any logic is changed

### Checklist


mmeeks:
Looks great to me - thanks Nick ! =)

PR #3155, opened 2021-08-27
private/pedro/mobile calc formula bar size style fixing

pedropintosilva:
- Calc, mobile: fixed formula bar size
- leaflet: avoid showing formula bar in read only mode


lpranam (approved):
Looks good to me.

pedropintosilva:
Thanks @Rash419 ! But now it's failing in a completely unrelated test (writer track changes)

PR #3162, opened 2021-08-31
Annotations for small windows

eszkadev:
This change introduces comments list with reduced size for small browser windows or tablets.

### DIFF
--- a/cypress_test/integration_tests/common/desktop_helper.js
+++ b/cypress_test/integration_tests/common/desktop_helper.js
@@ -270,7 +270,7 @@ function deleteImage() {
 		.should('not.exist');
 }
 
-function insertMultipleComment(docType, numberOfComments = 1) {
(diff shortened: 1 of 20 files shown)


pedropintosilva:
Thanks! I have added a couple of changes. 

I think we could also improve further:
* The way the user can close the container
  * Right now if the user clicks again the avatar bubble (that opened the container) it closes and opens again. I think it would be cool to simply act as close button (when the container is opened)
  * I would then add a different icon (close) on mouseover as seen in https://share.collabora.com/index.php/s/tzNEdrjSeqP68SH
* The way the elements get resizes when the user triggers the reply action 

The following video shows those issues:

https://user-images.githubusercontent.com/65948705/132020879-69cd5d9d-574c-424d-9c9c-71d70b394815.mp4

Andreas-Kainz:
cooll shit. great work

PR #3166, opened 2021-09-01
Remove leftover tiles in the iOS app already when the document is closed

tml1024:
Normally (ideally), tiles (.bmp files) are removed as soon as the JS
has displayed them. But occasionally something goes wrong and they are
left behind. (For instance, it seems to happen if the user closes the
document immediately when it shows up.)

Do not leave them on disk until the app starts the next time.

Change-Id: I0c764280a69a16ad3b7b67c329832fd5331c2e1e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/MobileApp.cpp
+++ b/common/MobileApp.cpp
@@ -9,35 +9,79 @@
 #include <map>
 #include <mutex>
 
+#include "Log.hpp"
 #include "MobileApp.hpp"
 
 #if MOBILEAPP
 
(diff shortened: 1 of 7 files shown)


mmeeks (approved):
Lets get it in and see =) thanks Tor !

tml1024:
Argh, I was too fast when manipulating the ifdefs for temporary logging, accidentally put the actual beef of this change inside #if 0... Will file a follow-up PR.

PR #3171, opened 2021-09-01
Cypress: wait before some checks and deselect text

pedropintosilva:
Attempt to avoid random errors by:
- Wait before checking selection
- Wait before checking blinking-cursor
- Also press leftarrow key before checking that there is no selection
  - I wonder why we have ctrl + a before in the first place

Change-Id: If196ba79c36776c3add4600a58997055489b0eed


timar:
I think this PR is not important any longer. Please re-open, if you disagree.

PR #3175, opened 2021-09-01
Calc, mobile: fixed formula bar size (hide it on readonly)

pedropintosilva:
- Remove negative margins (no need for that anymore) and instead use
the ideal height values from the get go
- Also Remove parent's height rule (better to follow children's value)
- Remove rules that are never applied (since they are always superseded
by others)
- Align addressInput to left column header
- addressInput: Use the same text alignment as used in column header
- Set the addressInput border color to the same color as in formula
input field

Change-Id: If2ed681e408fb12dadc33ed47e3c68d8237324f9

### DIFF
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -13,6 +13,10 @@ function clickFormulaBar() {
 	// canvas, which is accurately sized.
 	// N.B. Setting the width of the inputbar_container
 	// is futile because it messes the size of the canvas.
+	helper.doIfOnMobile(function() {
+		helper.waitUntilIdle('.inputbar_container');
+	});
+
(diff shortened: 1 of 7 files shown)


pedropintosilva:
Cherry picked @Rash419 7007506108320be908672b10f5f045b079dcf345

and resolve conflict : 

kept `skip` (that was in upstream) in ```it.skip('Switch oneline-multiline mode of input bar', function() {``` instead of replacing it with `only` , [cypress_test/integration_tests/mobile/calc/formulabar_spec.js](https://github.com/CollaboraOnline/online/pull/3175/commits/7397a956d574ad2e002f6d12bdebdf69ba156657#diff-97da57578392cb351203d817c29f9b2112824ec89621466b831ba8249d0bd487R57)

Issue #3084, opened 2021-08-16
Share URL creation code & drop un-necessary map.options. pieces ...


mmeeks:
A pure refactor is needed to share code that generates URLs for the backend
and to consolidate it into js/global.js

git grep -5 serviceRoot

gives some idea of where the problems are that need consolidating. I
would like to see a single use of that and to:

* bin the map.options stuff - this is global state, and the map is baggage.
* a helper method in global.js that can take a sub-path eg. '/ws' or '/insertfile'
* this method should append the WOPISrc= of course to the URLs
* It should centralize the URL building and ensure the docParamsPart= logic is included
* We should test clipboard, file insertion, websocket setup, help service, welcome dialog, download, print etc.

Thanks !

Darshan-upadhyay1110:
@Ashod is it okay to close this issue ? As I see, you have mentioned this as `resolved` in https://github.com/CollaboraOnline/online/pull/3179

Darshan-upadhyay1110:
The requested refactor has landed: backend URL construction is consolidated into helpers in browser/js/global.js (global.host + global.serviceRoot + path), and map.options.serviceRoot usage is gone from browser/src. Ashod already marked this resolved via PR #3179.

So closing this...

PR #3179, opened 2021-09-02
Hexify URLs

### LINKED ISSUE #3086: Optional hexification of URLs...
It appears that some IIS server setups - particularly in Azure infrastructure cannot stop themselves from de-mangling URL components that contain URLs so things like:

GET wss://foo.com/gateway/wopi/client/lool/https%3A%2F%2Fbaa.com%2Fget%2Fwopi%2Fhost%2Ffiles%3Faccess_token%3D1234456/ws

get mashed into:

GET wss://foo.com/gateway/wopi/client/lool/https:/baa.com/get/wopi/host/files?access_token=1234456/ws

Note broken URL, and un-escaping of escaped '/'s. We don't have access to configure such intervening mangling software.

As such - we should provide a configure option in loolwsd.xml - which allows hexification of the URL: ie. we provide:

GET wss://foo.com/gateway/wopi/client/lool/0x68747470733a2f2f6261612e636f6d2f6765742f776f70692f686f73742f66696c65733f6163636573735f746f6b656e3d313233343435362f7773/ws

We should use this instead of URL encoding, so we don't stack too much madness & bloat on top of ourselves.

Thanks.

Ashod:
This resolves #3086 and #3084.

- wsd: hexifying helpers and tests
- wsd: hexified URL support
- leaflet: move url manipulation to helper
- leaflet: support hexifying the body of the lool URI
- leaflet: replicate options.webserver in window.webserver
- leaflet: move map.options.server to window.host
- leaflet: move map.options.webserver to window.webserver
- leaflet: move map.options.serviceRoot to window.serviceRoot
- leaflet: makeURL -> makeHttpUrl
- leaflet: makeWsUrl
- wsd: configuration setting to control embedded-url hexifying
- wsd: throw when an invalid URL is used to create an http session
- leaflet: move wopiSrc into window and central URL maker
- wsd: improved dehexifying of URLs
- leaflet: centralize wopiSrc URL forming

### DIFF
--- a/loleaflet/js/global.js
+++ b/loleaflet/js/global.js
@@ -768,10 +768,9 @@ window.app = { // Shouldn't have any functions defined.
 
 	var docParams, wopiParams;
 	var filePath = global.getParameterByName('file_path');
-	var wopiSrc = global.getParameterByName('WOPISrc');
-	if (wopiSrc != '') {
-		global.docURL = decodeURIComponent(wopiSrc);
-		wopiSrc = '?WOPISrc=' + wopiSrc + '&compat=';
+	global.wopiSrc = global.getParameterByName('WOPISrc');
+	if (global.wopiSrc != '') {
+		global.docURL = decodeURIComponent(global.wopiSrc);
 		if (global.accessToken !== '') {
 			wopiParams = { 'access_token': global.accessToken, 'access_token_ttl': global.accessTokenTTL };
 		}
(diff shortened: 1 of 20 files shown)


loleaflet/src/core/Socket.js
```diff
@@ -1139,6 +1139,8 @@ app.definitions.Socket = L.Class.extend({
 			this._map.options.doc = docUrl;
 			this._map.options.previousWopiSrc = this._map.options.wopiSrc; // After save-as op, we may connect to another server, then code will think that server has restarted. In this case, we don't want to reload the page (detect the file name is different).
 			this._map.options.wopiSrc = encodeURIComponent(docUrl);
+			window.wopiSrc = this._map.options.wopiSrc;
```
mmeeks, line 1137: I thought we were killing this._map.options.wopiSrc etc. - and using these via the global settings everywhere to reduce confusion.

wsd/RequestDetails.cpp
```diff
@@ -140,17 +170,6 @@ void RequestDetails::processURI()
     _fields[Field::Type] = _uriString.substr(off, posDocUri - off); // The first is always the type.
     std::string uriRes = _uriString.substr(posDocUri + 1);
 
-    // Decode hexified component, if any.
-    if (Util::startsWith(uriRes, "0x"))
```
mmeeks: This gave me more confidence - de-hexifying just a specific piece of the URL seems much safer to me during parsing.

loleaflet/js/global.js
```diff
@@ -849,12 +869,7 @@ window.app = { // Shouldn't have any functions defined.
 	} else {
 		// The URL may already contain a query (e.g., 'http://server.tld/foo/wopi/files/bar?desktop=baz') - then just append more params
 		var docParamsPart = docParams ? ((global.docURL.indexOf('?') >= 0) ? '&' : '?') + docParams : '';
-		var websocketURI = global.makeWsUrl('/lool/');
```
mmeeks: nice cleanups & code redux here =)

mmeeks (approved):
Lets get this in - and iterate it later.

mmeeks:
Some follow-on work needed: I want to be really sure we are not searching the whole URL for lool/0x - but only the piece of the path we know can be escaped.

This is also going to need some seriously hard testing.

PR #3181, opened 2021-09-03
Implemented an option to pass PostMessageOrigin

merttumer:
External apps load loolwsd inside the iframe
and loolwsd makes postmessages to parent window.
We receive the postMessageOrigin from checkfileInfo yet
we still send some messages before we even go to the WOPI Api
in that case, if parent window runs on a different domain, we
end up with CORS blocking by the browser. To prevent that we can
allow safely passing the origin inside the first post like access_token
and sanitize it with Poco::URI::encode.

Change-Id: I5724f2d103603a599d45b7f61da81fb30834ef0e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
Looks good :)

PR #3185, opened 2021-09-03
rename loleaflet.html -> cool.html

hcvcastro:
- initial rename loleaflet -> browser
- initial replace path loleaflet -> browser


kendy:
Instead of moving all the sources - can we just rename stuff in the dist/ dir during the build of the bundle or so?

hcvcastro:
> Instead of moving all the sources - can we just rename stuff in the dist/ dir during the build of the bundle or so?

yes, we can rename during build of the bundle

PR #3187, opened 2021-09-06
kit: fix build

vmiklos:
When building against core.git master:

kit/ChildSession.cpp:2668:13: error: enumeration value 'LOK_CALLBACK_SC_FOLLOW_JUMP' not handled in switch [-Werror,-Wswitch]

It's not clear how we want to handle this in Online, so have a default
label in debug builds as well.

Change-Id: Ia0b1cef33e8a5a7fbc1518586ea1e3bb13aa5f22


vmiklos:
Only FreeBSD failed.

PR #3189, opened 2021-09-06
Make some temporary debug output use speech synthesis instead of NSLog()

tml1024:
As a debugging aid, use iOS speech synthesis to speak the amount of
disk space used by the app and its data. This is to help trace spae
wasted by leaked tile bmp files. (Sure, sure, the leaks should be
fixed.)

But it is still normally inside #if 0. Will have to figure out some
way for the user to turn it on at will. For instance by creating a
document with some very specific name. Like CO-DEBUG-FLAG.odt.

Change-Id: I594f5155ca92051f18ac59ede679951a92ba0846

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Not sure what to do with this. In a way, it's fun to keep around as an example of how relatively easy it is to use the speech API. On the other hand, there is absolutely no need for this PR as such any more, no tiles are kept in the file system any longer.

tml1024:
So yeah, best to just close.

PR #3193, opened 2021-09-07
Stop using .bmp files for tiles on iOS

tml1024:
We don't want leaking tiles to take up valuable space on the device.
Not even just while a document is open.

Change-Id: I761c2bb2ab121ee76b5535e96bd5702d0fce6856

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Thanks Tor, sad to see it go - no doubt we'll ~immediately discover the leak is actually some chart-stream nonsense instead ;-) but - lets try it like this.

tml1024:
The CI seems to be stuck.

tml1024:
Oh well, this is iOS-only, let's merge anyway.

PR #3196, opened 2021-09-07
jsdialog: prevent from document losing focus on toolbutton click

eszkadev:
not focus notebookbar on button click


merttumer (approved):
Looks good

PR #3204, opened 2021-09-08
wsd: move convert-to docs into the jail

Ashod:
We now download the convert-to files into the
child-root/tmp directory and then move it into
the jail that will convert it. This way ownership
and cleanup become contained within our child-root
and jail subsystems. This reduces the chances of
leaking convert-to files and simplifies the design.

In addition, we avoid an extra file copy and improve
the security of the convert-to API.

(cherry picked from commit fad4222a2a91785e714df7ffad1ffd4ddb8f4142)

Change-Id: I450c24d0d0dc0da447c8072b0701c3b48d07c81b

### DIFF
--- a/test/UnitConvert.cpp
+++ b/test/UnitConvert.cpp
@@ -50,6 +50,7 @@ public:
 
         config.setBool("ssl.enable", true);
         config.setInt("per_document.limit_load_secs", 30);
+        config.setBool("storage.filesystem[@allow]", false);
     }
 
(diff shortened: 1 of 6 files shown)


vmiklos:
I don't see anything obviously problematic with this. I tested that convert-to still works after this and that we indeed don't emit a warning now that hardlinking fails (on a setup where /tmp and the build direcatory are on separate partitions). Thanks Ash!

PR #3205, opened 2021-09-08
fix slow doc loading that has lots of comments [co-6-4]

dennisfrancis:
* Resolves: # 
* Target version: co-6-4

### Summary
Fix slow doc loading that has lots of comments by avoiding eager importing of each comment at document load. Instead setup each comment only when it is necessary.

### Checklist


eszkadev (approved):
Looks good

PR #3211, opened 2021-09-09
cypress: calc: added row and column operation specs

Rash419:
Change-Id: I5f6fefad5dfcfc41f2d5d6f40dc1aec182c75d7e

* Target version: master


loleaflet/src/control/Control.Menubar.js
```diff
@@ -625,7 +625,7 @@ L.Control.Menubar = L.Control.extend({
 				{uno: '.uno:FormatLine'},
 				{uno: '.uno:FormatArea'}
 			]},
-			{name: _UNO('.uno:SheetMenu', 'spreadsheet'), type: 'menu', menu: [
+			{name: _UNO('.uno:SheetMenu', 'spreadsheet'), id: 'sheet', type: 'menu', menu: [
```
lpranam, line 628: wouldn't it be better to use `sheetmenu` as id to avoid confusion?

lpranam (changes requested):

loleaflet/src/control/Control.Menubar.js
Rash419: if you see the ID in DOM it would be menu-sheet

lpranam: okay then go ahead.

PR #3212, opened 2021-09-09
android: show gridlines if background color not set

eszkadev:
when we don't receive background color we need to assume we want gridlines.
In snapshot android builds gridline was missing.


dennisfrancis (approved):
Looks good. Thanks! Probably the callback introduced in https://gerrit.libreoffice.org/c/core/+/120524 is not getting triggered in the android case for some reason.

PR #3219, opened 2021-09-13
mobile: keep cursor visible when keyboard pops up

eszkadev:
example:
1. open spreadsheet, let's say you see 20 rows
2. doubleclick on 19th row to edit cell

before: cursor was under the keyboard
after: cursor is visible in the center of the visible area


merttumer (approved):
Works great, thank you

PR #3221, opened 2021-09-13
private/pedro/readonly ruler fix

pedropintosilva:
- Revert "Readonly mode: Do not display ruler"
- Readonly mode: Hide ruler

### DIFF
--- a/loleaflet/src/control/Control.UIManager.js
+++ b/loleaflet/src/control/Control.UIManager.js
@@ -166,9 +166,10 @@ L.Control.UIManager = L.Control.extend({
 			L.DomUtil.remove(L.DomUtil.get('spreadsheet-toolbar'));
 			L.DomUtil.remove(L.DomUtil.get('presentation-controls-wrapper'));
 
-			if ((window.mode.isTablet() || window.mode.isDesktop()) && this.map.isPermissionEdit()) {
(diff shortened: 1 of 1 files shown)


loleaflet/src/control/Control.UIManager.js
```diff
@@ -166,9 +166,10 @@ L.Control.UIManager = L.Control.extend({
 			L.DomUtil.remove(L.DomUtil.get('spreadsheet-toolbar'));
 			L.DomUtil.remove(L.DomUtil.get('presentation-controls-wrapper'));
 
-			if ((window.mode.isTablet() || window.mode.isDesktop()) && this.map.isPermissionEdit()) {
-				var showRuler = this.getSavedStateOrDefault('ShowRuler');
-				L.control.ruler({position:'topleft', interactive:true, showruler: showRuler}).addTo(this.map);
+			if ((window.mode.isTablet() || window.mode.isDesktop())) {
+				var showRuler = window.docPermission === 'readonly' ? false : this.getSavedStateOrDefault('ShowRuler');
```
lpranam, line 170: I think maybe we can move this condition inside the `if` conditions. because if it is false we don't show ruler.

lpranam (changes requested):
everything else looks good

PR #3225, opened 2021-09-13
Hide/Restrict uno commands

lpranam:
control uno command functionality and disable functions as needed.

* Target version: master


mmeeks (commented):
Look fine - I'm not a big fan of the 'freemium' word though. I wonder if we can actually put all of this through a single (lets just rename it) LibreOfficeKit API - called blockCommands - or something - and then we can distinguish why we blocked it later on the kit side ?
I imagine people rightly have an allergy to the word 'freemium' (as do I ;-).
Thanks !

PR #3227, opened 2021-09-14
android: Fix creating a new document causes an exception

merttumer:
When we create a document inside a network storage
we need to do it in a seperate thread

Change-Id: I43cb5f09dfab2f6c31de4647b4a312a3c29e746b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy (approved):
Looks great to me, thank you for that!

PR #3228, opened 2021-09-14
android fix: use correct instance for Unit classes getter

eszkadev:
On android we had issue with not working release builds
(worked when optimization flag -O0 was used for Unit.cpp).

Global which stored instance of UnitBase subtype was returned
after cast to a choosen subtype. Unfortunately in case when
android didn't work UnitWSD was returned in UnitKit::get() !

Thich patch makes us sure we don't cast to not-correct type.

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -75,35 +75,36 @@ UnitBase *UnitBase::linkAndCreateUnit(UnitType type, const std::string &unitLibP
 bool UnitBase::init(UnitType type, const std::string &unitLibPath)
 {
 #if !MOBILEAPP
-    assert(!Global);
+    assert(!get(type));
 #else
     // The LOOLWSD initialization is called in a loop on mobile, allow reuse
-    if (Global)
+    if (get(type))
         return true;
 #endif
 
     if (!unitLibPath.empty())
     {
-        rememberGlobalInstance(type, linkAndCreateUnit(type, unitLibPath));
-        LOG_DBG(Global->_testname << ": Initializing");
+        UnitBase* instance = linkAndCreateUnit(type, unitLibPath);
+        rememberInstance(type, instance);
+        LOG_DBG(instance->_testname << ": Initializing");
 
-        if (Global && type == UnitType::Kit)
+        if (instance && type == UnitType::Kit)
         {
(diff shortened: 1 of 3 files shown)


common/Unit.cpp
```diff
@@ -185,6 +206,12 @@ void UnitWSD::lookupTile(int part, int width, int height, int tilePosX, int tile
     }
 }
 
+UnitWSD& UnitWSD::get()
+{
+    assert(GlobalWSD && GlobalWSD->_type == UnitType::Wsd);
```
Ashod, line 211: If we guarantee that GlobalWSD always holds an instances of UnitWSD, why do we need to store the type and compare it?

common/Unit.cpp
```diff
@@ -237,6 +280,11 @@ void UnitBase::returnValue(int &retValue)
     if (TimeoutThread.joinable())
         TimeoutThread.join();
 
+    if (Global == GlobalKit)
+        GlobalKit = nullptr;
+    if (Global == GlobalWSD)
+        GlobalWSD = nullptr;
```
Ashod, line 286: If we have two instances in the same process, surely we want to delete both. The `delete Global;` below will only delete one.

common/Unit.cpp
```diff
@@ -24,6 +24,8 @@
 #include <common/SigUtil.hpp>
 
 UnitBase *UnitBase::Global = nullptr;
+UnitKit *GlobalKit = nullptr;
+UnitWSD *GlobalWSD = nullptr;
```
Ashod, line 28: If we have two separate variables for each type, why do we need the `UnitBase::Global`? Isn't it better to replace `UnitBase::Global` with `UnitBase::GlobalKit` and `UnitBase::GlobalWSD` ?

Ashod (approved):
Thanks Szymon! I left some comments for a better approach, but if this works, then let's get it in.

common/Unit.cpp
eszkadev: we need UnitBase::Global for UnitBase::get() which we use for example in WebSocketHandler.cpp where we don't know which kind of test we run (it returns the last set UnitBase subclass)

PR #3232, opened 2021-09-15
cypress: added autofilter spec for calc desktop

Rash419:
Change-Id: I0df97b82bea545daff4184af3ff4e135bcaad846

* Target version: master


cypress_test/integration_tests/desktop/calc/autofilter_spec.js
```diff
@@ -0,0 +1,169 @@
...
+	function toggleAutofilter() {
+		//enable/disable autofilter
+		if (mode === 'notebookbar') {
+			helper.waitUntilIdle('#toolbar-up');
+
+			cy.get('#toolbar-up .w2ui-scroll-right').then($button => {
+				if ($button.is(':visible'))	{
+					$button.click();
+				}
+				if ($button.is(':visible'))	{
```
lpranam, line 32: repeated statement! if you meant to do it then maybe while loop is better option.

lpranam (changes requested):

PR #3240, opened 2021-09-16
Fix helm host binding to path in ingress.yaml

printminion:
* Target version: master 

### Summary
Without this change I get 404 since host name is not set properly

### DIFF
--- a/kubernetes/helm/collabora-online/templates/ingress.yaml
+++ b/kubernetes/helm/collabora-online/templates/ingress.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   rules:
   - host: {{ .Values.hosts.host }}
-  - http:
+    http:
       paths:
       - path: "/"
         pathType: Prefix


lpranam (approved):
Looks good. Thank you for the contribution. 

CI is failing at the moment because commit is not signed. Can you please update it? thanks.

printminion:
Looks like something not caused by this change. probably some "hiccup"

https://cpci.cbg.collabora.co.uk:8080/job/github_against_cp_online_master_debug/7330/console
```
 1) Repair Document
       Repair by user-2:
     AssertionError: Timed out retrying after 30000ms: expected '<div.leaflet-cursor.blinking-cursor.blinking-cursor-hidden>' to be 'visible'

This element `<div.leaflet-cursor.blinking-cursor.blinking-cursor-hidden>` is not visible because it has CSS property: `display: none`
```

lpranam:
Don't worry I will take care of it 😃

PR #3243, opened 2021-09-16
Action_Paste postMessage to work around the pasting mechanism.

mmeeks:
In some cases, the integration needs to paste the data directly to the
document, outside of our mechanism that detects whether we should use
the internal paste, or not.

Change-Id: I5e5de09a990ed68451af34dd208a81391a0e9cfa

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Assume CI is broken; adding this seems extremely unlikely to disrupt an un-related cypress test.

PR #3272, opened 2021-09-21
No need to use a separate 'that' variable in the _slurpMessage function

tml1024:
Storing 'this' into a separate variable, typically called 'that,' is
needed only to enable it to be captured in a (callback) function that
you create in the scope of the 'that' variable. No callback or other
function is created inside the _slurpMessage function.

Change-Id: Iae2c6844a3dfb0ff486ee6762ee129893829f272

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Not sure if I understand how and why 'that' is typically used, is the commit message correct?

mmeeks (approved):
Makes perfect sense - this thing got re-factored a few times into and out of callbacks AFAIR.

PR #3275, opened 2021-09-22
mobile: reset IgnorePanning on touch end

eszkadev:
When we move marker we set IgnorePanning to true.
Unfortunately onUp handler is not called as Map.TouchGesture
overrides that event - so release variable there too.


merttumer (approved):
Looks good to me

PR #3286, opened 2021-09-23
Mobile: actionbar items: Fix misalignment and css

pedropintosilva:
- Not all elements were uniformly spaced out
- CSS was not targeting properly all elements
- Remove the need to add manually every element by id
every time we add a new one (or change it)

Change-Id: I6de930e351822c3477c26c8e6569c41975cee142


pedropintosilva:
resolves: image

Issue #2484, opened 2021-05-27
Mobile BottomToolbar context related


Andreas-Kainz:
The BottomBar on mobile show always content for context-text, but there are other context related stuff. 

Can the BottomBar show context related commands. For Impress and Draw where you aren't by default at context-text it is very strange, but the workflow can be improved everywhere with context related commands.

Andreas-Kainz:
@eszkadev how can I add context related toolbar commands?

eszkadev:
we don't have possibility for that now. I had PR for that in the past (you can look in the history) but then you decided to abandon that because you wanted to focus on improving notebookbar.

Andreas-Kainz:
And now notebookbar is improved so I can work on the toolbars. So the old PR will give the option to have context related stuff?

eszkadev:
yes, it added "context: xxxx" property to the toolbar item

Andreas-Kainz:
@eszkadev can you please make me one example how it work?

Andreas-Kainz:
Context for Table and Graphic work. But context for default|Text|DrawText didn't work (Like it was defined for the notebookbars).

Andreas-Kainz:
There are two open questions to fix this issue:

1. initial reload toolbar cause when I make context ['Text'] in writer, they should be shown by default, cause default = Text (in writer).
2. context ['Draw|DrawLine|3DObject|MultiObject|Graphic|DrawFontwork'] should be possible. So a command was visible at more than one context entry. Like we have at Notebookbar

Andreas-Kainz:
As @pedropintosilva is in charge of UX, I add him. In addition mybe he can help ;-)

PR #3287, opened 2021-09-24
Mobile Context Toolbars for writer [WIP] #2484

Andreas-Kainz:
Change-Id: I8d7711a9002f7a49e586beda8605485df85d2ede

* Resolves: #2484
* Target version: master 

### Summary
Show Context Related Toolbar commands within Mobile Bottom Toolbar

### TODO
- [ ] You can have ONLY one context item, so `context: ['Graphic']` and `context: ['Draw']` are the same. Would be nicer if you can use `context: ['Graphic|Draw']` like in https://github.com/CollaboraOnline/online/blob/dad92dee2523f67c5d592727fbd9b0c9eb8e400a/loleaflet/src/control/Control.NotebookbarWriter.js#L57
- [ ] `context: ['default]`  wasn't defined so at the initial load you don't see commands.


Andreas-Kainz:
This is a WIP PR. as default isn't available you don't see any command at the initial load of the bottom bar only if you change from text to another context, text will work.
In addition you see that if we can't use more than one context entry, the toolbar has a lot of copy paste elements.

@eszkadev I know you have a lot on your todo list. I can't fix it myself and I need this two items fixed to go ahead with context toolbars, ...

Andreas-Kainz:
This PR isn't for push to master. It's for fix the issues.

eszkadev:
@Andreas-Kainz it is possible to define multiple contexts, [ ... ] is an array, so you can do: ['Text', 'Graphic']

eszkadev:
@Andreas-Kainz I've added support for "default" context

BTW: please rebase

Andreas-Kainz:
Thanks for the work Szymon. It work very well. As there was a lot of changes in the past and the change here is very tiny. I make the change with a new MR. Thanks for your hard work.

PR #3300, opened 2021-09-26
leaflet: allow only more restrictive permission change

Ashod:
Change-Id: Iee192f7e75f0ede12c0b3cf34bf04ff3ed50c110


gokaysatir (approved):
Thank you for the fix and explanations :)

PR #3314, opened 2021-09-28
Zoom-out animation improvements

dennisfrancis:
* Target version: co-6-4

### Summary
* Zoom frames are drawn directly using tiles in tile-array or cache. 
* Draws grid before drawing tiles every zoom frame.
* A zoom level is chosen for every zoom-frame which has tiles that has maximum coverage for the zoom-frame area based on a score indicator. Tile of this zoom level is used to paint the frame.
* draw/pdf-zoomframe: needs special treatment of y coordinates
* adds SheetGeometry methods convert b/w core-pixels between zooms needed for Calc.
* Use sheetGeometry to transform the view coordinates for the chosen zoom-level for the frame. However this is not done for indivdual tiles. As a result there could be small alignment discrepancies between tiles and other stuff we draw using sheetGeometry such as grids and overlays.

### Tested on
* Desktop/Mobile Chrome browser
* Desktop/Mobile Firefox browser

### Caveats
* In Mobile Firefox, pinch zooming in general is noticeably slow. On doing this quickly it misses render frames and in most cases it is perceived as a view jump. Rendering is done with requestAnimationFrame API and it works well in Chrome.
* There is no change in tile prefetch code yet. For the first zoom-out, if some tiles are missing for all zoom levels,  the area appears  with default background (or empty grid in Calc).
* For non-Calc apps there is no definite info about the margins surrounding the page area in the client, so it is not possible to draw the zoom-frames at the "expected" centered position when zooming out from a view where there is no left (and / or) right margin. In such cases after the pinch zoom is finished the view will jump to make the page centered.
    * This lack of information can also impact the scoring of zoom-levels and it may sometimes choose a non-optimal zoom-level to draw the frame.
* For Calc, sheetGeometry conversions are not done for each tile to calculate its position on the frame at the frame scale, so there will be small discrepancies between tile contents and the overlays/grids.

### Checklist

### DIFF
--- a/loleaflet/Makefile.am
+++ b/loleaflet/Makefile.am
@@ -31,7 +31,7 @@ DIRECTORY_TS = $(srcdir)/src/layer/tile $(srcdir)/src/control # Add ts directori
 SRC_TS_ALL := $(foreach dir, $(DIRECTORY_TS), $(wildcard $(dir)/*.ts))
 SRC_TS_JS_FILES := $(SRC_TS_ALL:.ts=.js)
 
-$(SRC_TS_ALL:.ts=.js): $(SRC_TS_ALL:.ts=.ts)
+$(SRC_TS_JS_FILES): %.js: %.ts
 	$(builddir)/node_modules/typescript/bin/tsc $(@:.js=.ts) --outfile $@ --module none --lib dom,es2016 --target ES5
 
 MOCHA_DIR = $(srcdir)/mocha_tests
--- a/loleaflet/src/layer/tile/CalcTileLayer.js
+++ b/loleaflet/src/layer/tile/CalcTileLayer.js
@@ -1136,6 +1136,30 @@ L.SheetGeometry = L.Class.extend({
 			this._rows.getTileTwipsAtZoom(point.y, zoomScale));
 	},
 
+	// accepts a point in core-pixel coordinates at current zoom
+	// and returns the equivalent point in core-pixels at the given zoomScale.
+	getCorePixelsAtZoom: function (point, zoomScale) { // (L.Point, number) -> L.Point
+		if (!(point instanceof L.Point)) {
+			console.error('Bad argument type, expected L.Point');
+			return point;
+		}
+
+		return new L.Point(this._columns.getCorePixelsAtZoom(point.x, zoomScale),
+			this._rows.getCorePixelsAtZoom(point.y, zoomScale));
+	},
+
+	// accepts a point in core-pixel coordinates at *given* zoomScale
+	// and returns the equivalent point in core-pixels at the current zoom.
+	getCorePixelsFromZoom: function (point, zoomScale) { // (L.Point, number) -> L.Point
+		if (!(point instanceof L.Point)) {
+			console.error('Bad argument type, expected L.Point');
+			return point;
+		}
+
+		return new L.Point(this._columns.getCorePixelsFromZoom(point.x, zoomScale),
+			this._rows.getCorePixelsFromZoom(point.y, zoomScale));
+	},
+
 	// accepts a point in print twips coordinates and returns the equivalent point
 	// in tile-twips.
 	getTileTwipsPointFromPrint: function (point) { // (L.Point) -> L.Point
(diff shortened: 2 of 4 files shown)


tml1024:
Looking into it. @mmeeks first asked me to try this in co-2021, but the set of commits seems to need quite some massaging to apply to that branch. Will test in cp-6.4.

tml1024:
Does indeed seem to have a beneficial effect. Not perfect, but then few things are. (When zooming in the location in the document that was the centre of your pinch gesture does not stay at the same location on screen. Earlier it was better to keep it there, but then the screen typically went completely blank for a while inbetween which was far worse of course.)

tml1024:
Yeah, clear improvement for presentation and spreadsheet documents, too.

tml1024 (approved):

mmeeks:
Nice - I wonder if we can preserve the zoom center more effectively & have something near perfect ? =) Dennis ?
Any chance of testing on PC with manual zooming etc. and if you're really happy lets get it into co-6-4 =)
Nice work Dennis!

tml1024:
I tested with Safari on macOS against a _make run_ with the commits. It looked worse than in the iOS app for some reason. I still got completely blanked-out tiles both when zooming out and back in.

tml1024:
(Sigh, I had confused zooming in and out in my comments. Corrected them now to match what I mean. I hope.)

tml1024:
Hmm, actually the behaviour in a desktop browser against _make run_ seems to depend on whether I use gestures on my touchpad, the plus and minus buttons, or the zoom level menu to zoom. Using the buttons or menu, it is much better.

dennisfrancis:
I believe there is no change in the zoom-frame location/size selection. There are two different situations:
1. The view just before pinch-zoom has no page-margin because the zoom level is high enough. Here are the screen-casts of how co-6-4 behaves for me with/without the PR. Both of them preserve the pinch center.

https://user-images.githubusercontent.com/132960/135799513-63936b5f-064b-467e-9531-86d7b1820b5b.mp4

https://user-images.githubusercontent.com/132960/135799544-6f7f23d4-3b76-4f22-9a95-5915c0b6bb45.mp4

2. The view just before pinch-zoom has page-margin(s). Here are the screen-casts of how co-6-4 behaves for me with/without the PR. Both of them do not preserve the the pinch center.

https://user-images.githubusercontent.com/132960/135799691-811f2a4e-e31e-4e11-bac8-f899fbb81630.mp4

https://user-images.githubusercontent.com/132960/135799708-90270c1e-b786-432c-a675-5cb4dfff973e.mp4


To fix the second case, we need to fix the definition of post zoom view area (from pre-canvas days) to take the margins into account, else there will be a jump after the animation finishes. I'll investigate if this can be done accurately for all apps (esp. Impress) without the margin data.

dennisfrancis:
I'll try to setup touchpad pinch-zoom in my Ubuntu machine and reproduce the completely blanked out tiles for zooming (in and out).

dennisfrancis:
I tried to setup touchpad based gestures (libinput-gestures), but unfortunately they are not getting detected in Chrome, our hammer pinch callbacks are not getting called...

dennisfrancis:
@tml1024 If possible can I get your console logs (info+error+warnings) for trackpad pinch (zoom-in/out) after applying the attached diff ? Thanks!
[debug-trackpad-pinchzoom.diff.txt](https://github.com/CollaboraOnline/online/files/7276768/debug-trackpad-pinchzoom.diff.txt)

tml1024:
@dennisfrancis Oddly enough, I don't see those added DEBUG log messages at all. I am investigating what events the browser actually gets from the trackpad gestures.

tml1024:
@dennisfrancis The trackpad pinch events apparently arrive as mouse wheel events with the Control key pressed. But the onMouseWheel function in ScrollSection.ts returns right away if ctrlKey is true? I don't understand;) Continuing to investigate. Or does this already help you?

dennisfrancis:
> @dennisfrancis The trackpad pinch events apparently arrive as mouse wheel events with the Control key pressed. But the onMouseWheel function in ScrollSection.ts returns right away if ctrlKey is true? I don't understand;) Continuing to investigate. Or does this already help you?

Yes, that helps a lot. I'll investigate ctrl+mousewheel. Sorry for the late reply!

tml1024:
The function _onWheelScroll in Map.Scroll.js seems to be involved, too. And in that the event is not ignored if the control key is down.

PR #3325, opened 2021-10-01
stress: fix disable-ssl build

vmiklos:
test/test.cpp and wsd/Storage.cpp has the same pattern.

Change-Id: Ic1dfbd3a5368eb9831bde9d77ea1bb813834d42b


vmiklos:
Code scanning failure is only in indexing/, ignore.

PR #3330, opened 2021-10-03
Shutdown improvements

Ashod:
- wsd: avoid misbehaving library cleanup
- wsd: allow shutting down during startup


vmiklos:
How does this relate to df30e0e92 from master, would it make sense to include a backport of that commit in this PR?

mmeeks (approved):
Looks like we wait a long time, but this looks fine =)

PR #3333, opened 2021-10-04
cypress: restore writer shape properties tests

eszkadev:
Make the shape path tests more readable and debuggable by asserting the coordinates in path commands of the shape path rather than on the whole string. The pixel positions are not stable across the branches (eg. co-2021 vs cp-6-4 in core) so were are doing "approximate" asserts.

Change-Id: I4ad0c9db32bf0397f1086b3039c7604385e7ca73

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/mobile/writer/shape_properties_spec.js
+++ b/cypress_test/integration_tests/mobile/writer/shape_properties_spec.js
@@ -1,11 +1,65 @@
-/* global describe it cy beforeEach require afterEach Cypress */
+/* global describe it cy beforeEach require afterEach Cypress expect */
 
 var helper = require('../../common/helper');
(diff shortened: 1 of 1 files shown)


eszkadev:
@dennisfrancis this is port of your patch from co6-4. It doesn't pass the modified test. Can you look if it is needed for that branch? (comment in the code says 'it can be a bug in core' so maybe it was fixed in co2021?)

dennisfrancis:
@eszkadev I've updated the tests to only do approximate asserts on the coordinates as some of the numbers have changed from that of cp-6.4 core. For now it seems a delta of 30 units is sufficient without changing the 'expected' values from co-6-4 patch.

PR #3341, opened 2021-10-05
First create UI then propagate permissions change

eszkadev:
Our permissions change callback modifies the UI so
first we have to create the UI when reconnecting.

This fixes the bug when on pdf renaming notebookbar appeared
above menubar.


merttumer (approved):
Looks and works great

PR #3342, opened 2021-10-05
leaflet: moved Certificate notification to statusbar

lpranam:
* Target version: master 

### Checklist


lpranam:
@pedropintosilva can you have a look if it needs any new icon?
thanks

PR #3350, opened 2021-10-06
Turn off desynchronized for non-desktop to avoid flickering

eszkadev:


dennisfrancis (approved):
Makes sense. Probably worth removing this for desktop case too if the perf numbers mentioned in the commit msg of c60987ab25f2e2a6389bef8fec65d9d7d88b3726 are maintained on a fresh profile. Thanks!

PR #3355, opened 2021-10-07
cypress: fix: writer/annotation_spec.js failing for NB

Rash419:
Change-Id: Ibf80c954036cd98c0da09f45b3ff9ce13acc9e44

* Target version: master


Rash419:
@dennisfrancis thanks for catching the failing test : ) , I wonder why error doesnot show in CI. It is possible that as we are using 2 commands for cypress test:
1) make check - for all classic desktop , mobile and multiuser tests
2) make check-notebookbar - for notebookbar desktop

Even though CI is running all the NB tests but at the end of displaying the result it is only considering the end result of first step

PR #3363, opened 2021-10-07
Fix two different background colours

merttumer:
Change-Id: Ibbfb39f9cccb69d69bcdd39983e076b49b77b8bf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
just writing here notes of our meeting @merttumer :

* Before we used to be able to change the color by setting a different background to `.leaflet-container`
  * That is now not possible because the z-index has changed on master (now if we set a background it will cover the document) that's why @gokay changed to : https://github.com/CollaboraOnline/online/commit/f33f24c16c7a22a62231381d6287205889614d23

Rash419:
seems like cypress is encountring an exception 
image

merttumer:
Closing this since it doesnt fix the root cause and the color variable there is not the actual background

PR #3389, opened 2021-10-08
Private/mmeeks/stressbits

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Miklos - any chance you can try out the fuzzer - I've no doubt broken it ;-) I put a FIXME in for that piece - I guess it would be good to understand how the filename of the test file gets passed in: is that present in the trace itself ? but then ... a bit confused about that I think. Would love to kill this use of the (grim) Poco / LOOLWebSocket that used to be needed for loolstress.

vmiklos:
Sure, checking.

mmeeks:
Oh - great, if there is nothing to break - I'll push then ;-) I would expect that there is some quite fertile ground for fuzzing some replays of our traces via loolstress in the future though; but we should start with some test-files to test against I guess.

PR #3402, opened 2021-10-12
Turn off desynchronized canvas for desktop

eszkadev:
It causes blurry/flickering on canvas while resizing window.
This was previously introduced for rendering optimization in:
78115809c7c87166e8ff9479fe503eafb1fcc69e
Canvas: not alpha, and de-synchronized.

https://github.com/CollaboraOnline/online/commit/78115809c7c87166e8ff9479fe503eafb1fcc69e


mmeeks:
I would be inclined just to remove the desynchronized attribute from the call in the 1st instance: I suspect it could cause some unexpected horrors: and it's far off the tested path for the browsers - otherwise great; thanks Szymon =)

PR #3410, opened 2021-10-13
Introduce --enable-bundle to decouple bundling from --enable-debug

tml1024:
Without --enable-bundle, --enable-debug works as before: Individual JS
and CSS files are sourced in loleaflet.html.

But with both --enable-debug and --enable-bundle, a non-empty
bundle.js and bundle.css is produced and that is used instead. It is
important for developers to be able to test also the use of bundle.js
because there can be significant differences in behaviour. Developers
typically have to use --enable-debug because otherwise there are
various stict run-time checks in the C++ code that you don't want to
bother with when developing, like having a separate login for COOL.

Change-Id: If7d0220b7910fe0f7a2f391b856cd0dedc788152

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,10 @@ AC_ARG_ENABLE([debug],
               AS_HELP_STRING([--enable-debug],
                              [Enable debugging, link with debugging version of Poco libraries]))
 
+AC_ARG_ENABLE([bundle],
+              AS_HELP_STRING([--enable-bundle],
+                             [Enable creating bundled JS and CSS even with --enable-debug]))
+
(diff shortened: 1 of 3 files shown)


tml1024:
The need for this was made apparent recently, when some minor weirdness in our Typescript resulted in a bundle.js that did not work, but as developers don't use a bundle.js, it was noticed only in the iOS app, which always uses a bundle.js. Naturally the assumption was then that it was an iOS-specific problem, but in fact it happened on Linux, too, in a "make run" scenario. if one forced it to use bundle.js.

PR #3415, opened 2021-10-13
Escape strings in JSON

mikekaganski:
This should take care e.g. for SharePoint user IDs, which have the
form of "domain\username", and the backslash must be escaped.

Change-Id: I75936070ad1661dc9b03e05a19b64159b0758018


kendy:
@mikekaganski : Looks good, please do merge.  Thank you!

Issue #2697, opened 2021-06-25
optimize startsWith ...


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @mrkara 

### Detailed description and rationale
Currently we do some degree of wasteful duplication and copying of items out of our common/StringVector.h using the array dereference operator so eg.
                else if (Util::startsWith(tokens[0], "child-"))
or
        else if (Util::startsWith(keyValue[0], "Text"))

These duplicate and create temporary std::strings which are immediately freed.

### Code pointers

We should add a 'StringVector' method startsWith(0, "child-") - that passes a token index in and returns true/false for the startsWith method - re-using the impl. of Util::startsWith on the underlying _string at the right offset for the token without duplication and/or heap allocation.

Thanks !

PR #3424, opened 2021-10-14
Optimize startsWith calls on StringVector elements

hugopeixoto:
* Resolves: #2697 
* Target version: master 

### Summary

There was a pattern of calling Util::startsWith on StringVector
elements:

- Util::startsWith(tokens[0], "boo")
- Util::startsWith(tokens.getParam(token), "boo")

These two expressions would cause a new string to be allocated and
immediately released. To optimize this, a StringVector::startsWith
method is introduced.

This method works by calling compare directly on the underlying
StringVector string, avoiding creating a temporary string.

### TODO

~manually check that things work; I don't seem to be able to `make run` (no such rule), must have missed a step. Tests pass, and I checked that they break with a wrong startsWith implementation, so I have some confidence that this works, but would appreciate confirmation.~

### Checklist

### DIFF
--- a/common/StringVector.hpp
+++ b/common/StringVector.hpp
@@ -136,6 +136,33 @@ public:
         return _string.compare(token._index, token._length, string, N) == 0;
     }
 
+    // Checks if the token text at index starts with the given string
+    template <std::size_t N>
+    bool startsWith(std::size_t index, const char (&string)[N]) const
+    {
+        if (index >= _tokens.size())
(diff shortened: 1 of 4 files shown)


vmiklos:
Looks good to me, thanks!

Could you please create a follow-up PR and add tests for the new startsWith() member functions? You could extend testStringVector in test/WhiteBoxTests.cpp.

Issue #115, opened 2020-10-02
Poco::Time* replacement with std::chrono


mrkara:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: _TBD_

We have a mix of `Poco::Timestamp` and `std::chrono` - we should move to using the latter consistently.

It would be good to move _wsd/*_ first - since it doesn't use much Poco API anymore - although it would be good to consider whether we need some convenience wrappers around `std::chrono` to make it somewhat less verbose - particularly the cast to milliseconds is very ugly.

Thanks !

yusufketen:
Should we write wrapper function to use Poco functions with std::chrono instead of Poco::Timespan or Poco::Timestamp functions?

kendy:
@yusufketen: Not sure I understand what do you mean by "function to use Poco functions with std::chrono"?  "Poco" is a library that we would like to avoid in the long term, and use only the std:: implementation of things.

As the initial description says, std::chrono is a bit too verbose in some cases, so it might be needed to create some convenience functions in Util.hpp or so.

To find the files that use Poco::Timestamp:

git grep -l 'include.*Timestamp'

PR #3425, opened 2021-10-14
Remove unused Poco/Timestamp include

hugopeixoto:
* Resolves: #115 
* Target version: master 

### Summary

There was a stray include in `test/WopiTestServer.hpp`. I removed it, and things still build / tests pass. I think this doesn't affect the runtime at all, being limited to the `test/` directory, so it should be safe to merge.

### Checklist


vmiklos:
Indeed looks unused, thanks!

PR #3433, opened 2021-10-17
Move adding the start script after running the install script.

by-erik:
This will speed up docker build if only something in start-collabora-online.sh has changed.

* Target version: master 

### Checklist


timar (approved):
Thanks!

PR #3443, opened 2021-10-19
wsd: add HttpEcho fuzzer

Ashod:
Change-Id: I75d21bd0e25221ee6621097a2605d62c4bb2ae4d


fuzzer/HttpResponse.cpp
```diff
@@ -7,7 +7,8 @@
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
 {
     http::Response response;
-    response.readData(reinterpret_cast<const char*>(data), size);
+    for (size_t i = 0; i < size; ++i)
+        response.readData(reinterpret_cast<const char*>(data), i);
```
vmiklos, line 11: Is this just something speculative? libfuzzer already watches coverage and already tries to insert \0 chars at random positions in the input, sounds like you try to duplicate that here, but without checking if it improves coverage or not.

I.e. it seems this is not helpful, but perhaps I miss something.

configure.ac
```diff
@@ -811,7 +811,7 @@ fi
 AC_MSG_CHECKING([whether to turn sanitizers on])
 if test "x$with_sanitizer" != "x"; then
     AC_MSG_RESULT([yes ($with_sanitizer)])
-    SANITIZER_FLAGS="-O1 -fno-omit-frame-pointer -fsanitize=$with_sanitizer -frtti -static-libasan"
+    SANITIZER_FLAGS="-O1 -fno-omit-frame-pointer -fsanitize=$with_sanitizer -frtti"
```
vmiklos, line 814: This is fine, my understanding is that the Jenkins fuzzer setup uses the sanitizer options from lode.git (set via a custom CC and CXX), so you can tweak this to your needs without problems.

configure.ac
```diff
@@ -821,7 +821,7 @@ fi
 AC_MSG_CHECKING([whether to build fuzzers])
 if test "$enable_fuzzers" = "yes"; then
     AC_MSG_RESULT([yes])
-    LIBFUZZER_FLAGS="-fsanitize=fuzzer-no-link"
+    LIBFUZZER_FLAGS="-fsanitize=fuzzer-no-link,address"
```
vmiklos, line 824: Same here. Jenkins enables both ASAN and UBSAN for fuzzers, btw.

fuzzer/HttpEcho.cpp
```diff
@@ -0,0 +1,149 @@
...
+        if (!(X))                                                                                  \
+        {                                                                                          \
+            assert(!(X));                                                                          \
+            __builtin_trap();                                                                      \
+        }                                                                                          \
+    } while (0)
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    static HttpRequestTests test;
```
vmiklos, line 125: It would be good to have an initial "seed" for this fuzzer, see e.g. fuzzer/httpresponse-data/load as an example to follow. Once you have that, you can update fuzzer/README to document how to to invoke the fuzzer.

Last: running such a fuzzer has the cost of 100% using one CPU core, so it would be useful if that cost brings benefits. Do you know if this fuzzer already found some problem in practice or is this more just a speculative one? Asking to find out if our Jenkins should have a 4th fuzzer running 7/24 or that would be a waste of resources.

vmiklos (commented):
You asked about the configure.ac bits, those look fine to me. For the rest, it looks good, but having initial input for the fuzzer would be important I think, otherwise it's hard to find interesting edge cases. Could you please address that? Thanks!

fuzzer/HttpResponse.cpp
Ashod: What I'm hoping to do here is to give the parser partial data. The fuzzer (as I understand, at least) generates data to trigger different paths/conditions in the code. One of the most important ones is the parser not having enough data to parse a token. Finding out all the different parser cases is really hard, so I'm trying to help the fuzzer by re-using a single data to cover all possible 'incomplete data' cases. 

I have to agree that this is probably useful for catching early issues and once we have a well-tested parser it probably wastes cycles. I'll consider reverting it.

configure.ac
Ashod: Perhaps we should enable more sanitizers here too?

fuzzer/HttpEcho.cpp
Ashod: Very good point. I've ran the fuzzer multiple times and generated several billion runs. I found a couple of issues (one or two crashes and a deadlock) and fixed them.

How likely is it to find other issues? Not very likely, but if we make a change to the parser then it'd be good to re-run the fuzzer. Should we have a core dedicated 24/7? Probably not.

configure.ac
vmiklos: It's basically your sandbox currently, so up to you. :-) If you want to get closer to the Jenkins setup, then enabling both "address" and "undefined" probably makes sense.

fuzzer/HttpEcho.cpp
vmiklos: Ah, in case it found a single real problem in the past, then I'm happy to have this in. :-)

configure.ac
Ashod: I've opted to leave this up to the user. I do remind in the help that configuring `--with-sanitizer=address,undefined` is a good idea, but don't force it. This maximizes flexibility.

vmiklos:
This looks good to me in general, but I'm not confident enough to review the net/ changes, which are also part of this PR.

I guess feel free to self-approve for that bit.

FWIW the fuzzers build for me with and without this PR, but clientsession fuzzer is now broken at runtime. It fails the same way https://bee.cbg.collabora.co.uk/job/fuzzer-clientsession/146/console fails. Can you reproduce that on your end?

```
./clientsession_fuzzer -max_len=16384 fuzzer/data/
```

is broken, while the admin fuzzer and a httpreponse fuzzer is OK.

vmiklos (approved):
The fuzzer/, kit/, test/ and wsd/ bits look OK to me, can't comment on the net/ changes.

Ashod:
> The fuzzer/, kit/, test/ and wsd/ bits look OK to me, can't comment on the net/ changes.

Thanks @vmiklos!

PR #3494, opened 2021-10-26
Keep branding theme dynamic

juliushaertl:
Should resolve the issue I've noticed back when testing https://github.com/CollaboraOnline/online/pull/3201 but when retesting it again now i think I've catched the issue there, but would appreciate some double checking if this really makes sense.

As far as I remember local static variables would be kept even when leaving the block so that seems to have caused that once the css/js links were initialized a change of the theme url parameter was not taking any effect.


kendy:
@juliushaertl: Looks good to me, let's get it in :-)

Unfortunately I think you need to rebase to get rid of the cypress failure - can you do that please?

kendy (approved):

juliushaertl:
Sure, rebased and CI seems happy now :)

PR #3505, opened 2021-10-27
Simplified DocKey processing in incoming requests

### LINKED ISSUE #3085: Simplify incoming request processing
The existing code in wsd/ does quite well to encapsulate this into RequestDetails but by the time we get to creating a DocumentBroker we start to pull things out and do weird stuff to the RequestDetails:

        const std::string url = requestDetails.getLegacyDocumentURI();

        LOG_INF("URL [" << url << "].");
        const auto uriPublic = DocumentBroker::sanitizeURI(url);
        LOG_INF("URI [" << uriPublic.getPath() << "].");
        const auto docKey = DocumentBroker::getDocKey(uriPublic);
        LOG_INF("DocKey [" << docKey << "].");
        const std::string fileId = Util::getFilenameFromURL(docKey);
        Util::mapAnonymized(fileId, fileId); // Identity mapping, since fileId is already obfuscated

This should all be brought back into the RequestDetails - which we should pass around and/or duplicate on the heap to push into the setupTransfer etc. 

The DocumentBroker::sanitizeURI() should presumably be done commonly inside the RequestDetails, as well as the breaking out of the key and so on etc.

There is altogether too much copy/paste between handleClientWsUpgrade, handleClientProxyRequest, handlePostRequest etc. - which I would like to ensure is shared and centralized.

Thanks !

Ashod:
Resolves #3085

- wsd: move sanitizeURI into RequestDetails
- wsd: move getDocKey into RequestDetails

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -87,21 +87,6 @@ void DocumentBroker::broadcastLastModificationTime(
     broadcastMessage(message);
 }
 
-std::string DocumentBroker::getDocKey(const Poco::URI& uri)
-{
-    // If multiple host-names are used to access us, then
-    // they must be aliases. Permission to access aliased hosts
-    // is checked at the point of accepting incoming connections.
-    // At this point storing the hostname artificially discriminates
-    // between aliases and forces same document (when opened from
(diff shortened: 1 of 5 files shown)


Ashod:
FreeBSD failure is unrelated.

PR #3519, opened 2021-10-29
snackbar: show when disconnected and on desktop

eszkadev:
this backports snackbar to 6.4 to avoid showing browser's alert when server was reconnected


pedropintosilva:
Tested, thanks!

PR #3522, opened 2021-10-29
android: Do copy back in another thread for network operations

merttumer:
If network provider is used, on some devices download/upload
operations can cause an exception. Creating a new file is already
covered but better to cover saving back operation too.

Change-Id: I90ffa890ed864e730c16c68bc777b1d6371fd0a8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy (approved):
Makes perfect sense, thank you!

PR #3524, opened 2021-10-29
quarantine: create hard links for quarantining when file is saved

lpranam:
number of versions to maintain per file can be specified in loolwsd.xml
on exceeding specified quarantine size oldest file(s) is deleted

Change-Id: I3ca55b9ab29a82988f19fe0acd43e0fae2c2a423

* Target version: master 

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ compile_commands.json
 systemplate
 test-driver
 jails
+quarantine
 cache
 test/data/hello-world.odt
 test/data/hello-world.ods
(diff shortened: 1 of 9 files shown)


tml1024:
Briefly looking, looks good. Didn't check yet how it works, will do.

tml1024:
@kendy and others had improvement suggestions in chat, so I assume you will be looking into this first, before this is merged?

kendy:
@tml1024: I don't think the review needs to block on my request (to have a 'quarantine_files.enable' config entry to enable / disable as a whole, and to default to false when not configured with --enable-debug), please do review & let's get it merged.

Not sure what the other feedback was, though?

@lpranam: Thanks a lot for the work!

tml1024 (approved):

PR #3525, opened 2021-10-31
kill reuse_cookies

Ashod:
This remove reuse_cookies as a setting and feature
altogether. It was originally a workaround to aid
with authentication, prior to having access_token.
However, it proved to be less useful than originally
anticipated, primarily because cookies nowadays have
security restrictions in browsers. In addition to
the fact that access_token simply deprecated it.

While the documentation has also been updated,
tests still have reuse_cookies in input data.
This is intentional to ensure the code is
backwards compatible with any deployment that might
still pass URLs with reuse_cookies around.

Change-Id: If214b299b34a910face8cabc7c1335621990c85e


kendy (approved):
Nice, let's get it in, thank you!

Ashod:
FreeBSD failure is unrelated.

Issue #3365, opened 2021-10-07
Help dialog: Use cypress to take screenshots dynamically 


pedropintosilva:
**Is your feature request related to a problem? Please describe.**
Right now it's very difficult to maintain and make sure that all screenshots present in the help dialog are up-to-date. And delegate that to a contributor might still be difficult to sustain ( https://github.com/CollaboraOnline/online/issues/2194 )

On top of that we do not have those screenshots localized 

**Describe the solution you'd like**
We already use cypress for testing can we also use it to take screenshots of specific components so they can be used by the help dialog?

PR #3526, opened 2021-11-01
cypress: update help dialog screenshot automatically

Rash419:
Change-Id: I9f3e3c9cc7838f89d5a66a25c1f6dd0ef8f8a946

* Resolves: #3365 
* Target version: master 

### Summary
I have created two spec file to update the screenshot:
1) desktop/writer/help_dialog_update_spec.js
2) desktop/calc/help_dialog_update_spec.js

You can run this file individually using:
make UPDATE_SCREENSHOT=true check-desktop spec=writer/help_dialog_update_spec.js
If UPDATE_SCREENSHOT is true then  only cypress will run the spec file and replace the original screenshot


pedropintosilva (approved):

cypress_test/integration_tests/desktop/writer/help_dialog_update_spec.js
```diff
@@ -0,0 +1,263 @@
+/* global describe it cy require afterEach beforeEach Cypress*/
+
+const { hideSidebar } = require('../../common/desktop_helper');
+var helper = require('../../common/helper');
+describe('Help dialog screenshot updation', function() {
```
kendy, line 5: Can you please add a bit of an explanation in form of a comment? :-)

cypress_test/plugins/index.js
```diff
@@ -101,6 +101,11 @@ function pickTests(filename, foundTests) {
 			testsToRun = [];
 		}
 	}
+
+	if (!process.env.UPDATE_SCREENSHOT) {
```
kendy, line 105: Can we improve the condition a bit - it is hard to read: if (! something) { remove_something(); } ; would be preferred to have if (something) { add_something(); } - if possible...

kendy (commented):
Looks good to be in general; some small changes would be good.

If possible - add also info how to trigger this to a README.

cypress_test/plugins/index.js
Rash419: In this case, It is not possible

PR #3529, opened 2021-11-02
Add avatar list at the top right corner

luka-nextcloud:
Signed-off-by: Luka Trovic <luka@nextcloud.com>
Change-Id: I39f567c67a727a23fdef3d98a147fa30a13b36a4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -22,7 +22,6 @@
 }
 #toolbar-down *{
 	font-family: var(--cool-font);
-	max-height: 35px !important;
 }
 #tb_actionbar_item_LanguageStatus table table td:first-of-type{
 	min-width: max-content !important;
@@ -306,10 +305,10 @@ td[id^='tb_editbar_item_sidebar']{
 	z-index: 1050;
 	right: 4px;
 	top: 0px;
-	width: 32px;
 	height: 32px;
 	background-color: transparent;
-	display: none;
+	display: flex !important;
+	align-items: center;
 }
 
 .main-nav:not(.hasnotebookbar) ~ #closebuttonwrapper {
@@ -323,7 +322,7 @@ td[id^='tb_editbar_item_sidebar']{
 }
 
 .main-nav:not(.hasnotebookbar) ~ #closebuttonwrapper #closebutton {
-	background: url('images/closedoc.svg') no-repeat center/24px !important;
+	background: url('images/closedoc.svg') no-repeat center/35px !important;
 }
 
 #closebutton:hover {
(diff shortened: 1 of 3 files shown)


juliushaertl:
@luka-nextcloud Avatar display should work again after https://github.com/CollaboraOnline/online/pull/3531 is merged

loleaflet/html/loleaflet.html.m4
```diff
@@ -203,6 +203,11 @@ m4_ifelse(MOBILEAPP,[true],
     <input id="selectbackground" type="file" accept="image/*" style="position: fixed; top: -100em">
 
     <div id="closebuttonwrapper">
+      <div class="userInfo">
```
juliushaertl, line 206: This should rather be moved out of the closebuttonwrapper

loleaflet/src/control/Control.UserList.js
```diff
@@ -81,6 +81,12 @@ L.Control.UserList = L.Control.extend({
 		var content = L.DomUtil.create('tr', 'useritem');
 		content.id = 'user-' + viewId;
 		$(document).on('click', '#' + content.id, this.onUseritemClicked.bind(this));
+		var ownerClass = userName == 'You' ? 'onwer' : 'guess';
+
+		$('#listActiveUser').append('<div class="user-item-wrapper" id="user-wrapper-' + viewId + '"><p class="user-item" id="user-' + viewId 
+		+ '" style="background-image:url(\'' + extraInfo.avatar + '\');border-color:' + color + '"></p><span> '+ userName +'</span></div>');
+
+		$('.userInfo').append('<p class="user-top ' + ownerClass + '" id="user-top-' + viewId + '" style="background-image:url(\'' + extraInfo.avatar + '\');border-color:' + color + ';"></p>');
```
juliushaertl, line 89: Would be great if we could move the rendering of the header avatars to a separate method, ideally also just making use of plain javascript for the dom manipulation instead of jquery, so this doesn't have the external dependency on the library.

loleaflet/css/toolbar.css
```diff
@@ -1052,3 +1052,84 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
...
+	background-color: white;
+	border-radius: 100%;
+	margin-right: -10px;
+	background-size: cover;
+	background-repeat: no-repeat;
+	border-style: solid;
+}
+
+.guess {
+	opacity: 0.5;
```
juliushaertl, line 1134: I'd say we don't need a separation between the display of the current user or guest like that. Also opacity may look a bit odd when other avatars are shining through

loleaflet/src/control/Control.UserList.js
juliushaertl: A few other points to consider:
- The avatar might not be available https://github.com/CollaboraOnline/online/blob/3ab824cdfde71c79d627bbe7c69f47610fed6019/loleaflet/src/control/Control.UserList.js#L94
- For uniquely identifying the element you could use a data-view-id="" attribute

luka-nextcloud:
Here's a screenshot:
image

loleaflet/src/control/Control.UserList.js
```diff
@@ -101,7 +102,27 @@ L.Control.UserList = L.Control.extend({
 		return content;
 	},
 
-	updateUserListCount: function() {
+	renderUserAvatars: function(viewId, userName, extraInfo, color) {
+		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
```
juliushaertl, line 106: This check is duplicate with the one in https://github.com/CollaboraOnline/online/pull/3529/files#diff-41de9cb6f19860d72e4fe0ba9bae4682ea79c3312a3ca4ad3c1219cf43f0d8efR88

loleaflet/src/control/Control.UserList.js
```diff
@@ -101,7 +102,27 @@ L.Control.UserList = L.Control.extend({
...
+			var avatarElement = '<div class="user-item-wrapper" id="user-wrapper-' + viewId + '"><p class="user-item" id="user-' + viewId 
+			+ '" style="background-image:url(\'' + extraInfo.avatar + '\');border-color:' + color + '"></p><span> '+ userName +'</span></div>';
```
juliushaertl, line 108:
```suggestion
			+ '" style="background-image:url(\'' + extraInfo.avatar + '\');border-color:' + color + '"></p><span> '+ this.escapeHtml(userName) +'</span></div>';
```

loleaflet/css/toolbar.css
```diff
@@ -1052,3 +1052,85 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
...
+	right: 25px;
+}
+
+.user-item-wrapper {
+	display: flex;
+	justify-content: flex-start;
+	align-items: center;
+}
+
+.user-item {
```
juliushaertl, line 1115: Please add somea dditional styling to make the spacing fit the existing popover entries and the font also match the default `font-family: var(--loleaflet-font);`

See screenshot for comparison:
image

loleaflet/html/loleaflet.html.m4
```diff
@@ -202,6 +202,11 @@ m4_ifelse(MOBILEAPP,[true],
     <input id="insertgraphic" type="file" accept="image/*" style="position: fixed; top: -100em">
     <input id="selectbackground" type="file" accept="image/*" style="position: fixed; top: -100em">
 
+    <div class="userInfo">
+      <div class="userInfo-content" id="listActiveUser">
+      </div>
+    </div>
```
juliushaertl, line 208: Let's just use one unique id here for those elements as the class `userInfo` is not really unambiguously. See the other comment for the naming in general as well :wink: 

It could be something like this:
```suggestion
    <div id="userListHeader">
      <div id="userListSummary"></div>
      <div id="userListPopover"></div>
    </div>
```

loleaflet/src/control/Control.UserList.js
```diff
@@ -101,7 +102,27 @@ L.Control.UserList = L.Control.extend({
...
+			if ($('#follow-editor').length == 0) {
+				$('#listActiveUser').append(avatarElement);
```
juliushaertl, line 110: The general naming is a bit hard to follow, maybe stick to something like this for the elements:

- userListSummary: For the avatars shown in the header (those should be limited to a maximum number, more users will then be shown in the popover)
- userListPopover: Shows a list of all users

That naming would ideally be reflected in the css id/class names as well as the variables that contain the HTML elements created.

juliushaertl:
@luka-nextcloud This would be the open points I can think of right now, most likely best tackled in the order mentioned :wink: 

- Use the viewId as unique identifier as there might be users with the same 
  - https://github.com/CollaboraOnline/online/pull/3529/files#diff-216513472020c43e49874b42a0ed8e16b5332900a653a31e77b9adea39e5db91R131
- Clicking the avatar should trigger the following of the user cursor (see current implementation)
- The checkbox to follow the current editor should be working (see current implementation)
- Do not show on mobile (there the current implementation works fine with just showing the number of users)
- Hide the existing avatar list in the status bar if the new one is shown
- Ideally do not remove all elements to rerender but check if an element is present already in the DOM and remove it if the view gets removed or add a new element if needed
- In the end it would be good to replace jQuery usage with plain javascript methods
  - https://youmightnotneedjquery.com/ is a good reference in that regard

pedropintosilva:
@luka-nextcloud it seems there are some problems in the following lines:

```bash
browser/src/control/Control.UserList.js
  88:1  error  Expected indentation of 2 tabs but found 4 spaces  indent
  90:1  error  Expected indentation of 2 tabs but found 4 spaces  indent
```

Could you please fix that?

Note: make sure your text editor/IDE is using the proper JS linter with the appropriated configuration within online project:
* `browser/.eslintrc`

luka-nextcloud:
Sure, let me check!

pedropintosilva:
This is still not passing the tests becuase

* `#userListSummary` is now obstructing the view from other elements
  * So when a user tries to click an icon in the toolbar it triggers instead that element
  * That's why in the tests when trying to click `#tb_editbar_item_linespacing` it fails

pedropintosilva:
Probably the parent's position should also be tweaked after that, so maybe something like this :


```diff
+ #userListSummary p {
+   margin-top: 0;
+   margin-bottom: 0;
+ }

#userListHeader {
-  top: 5px;
+ top: 2px;
}
```

juliushaertl:
@luka-nextcloud Lets continue over at https://github.com/CollaboraOnline/online/pull/3677 as discussed :wink:

pedropintosilva:
this has been moved to #3677

PR #3531, opened 2021-11-02
wsd: Only add one img-src rule to the CSP header

juliushaertl:
* Target version: master 

### Summary

Starting with 352d1d1f40c96cc76968ec657ad096ee36d7ca72 the CSP header contained two entries for img-src which caused the second to no longer take effect. This change brings back the rendering of avatars that may be loaded from external sources as provided by the WOPI integrator. 

### Checklist


pedropintosilva:
:+1: checked, works , thanks a lot

PR #3532, opened 2021-11-02
Prevent from error on LanguageStatus and use translation

eszkadev:
sometimes core send "{en-US}" as a LanguageStatus
prevent from trying to parse it as JSON and remember
translated names for language codes so we can reuse it
and show in the statusbar for such codes


eszkadev:
already fixed in core: https://gerrit.libreoffice.org/c/core/+/124448

Issue #3527, opened 2021-11-01
Impress: Wrong labelling of buttons in Layout tab


SustReal:
**Describe the bug**

I noticed some inconsistencies in the labellings of some buttons in the Layout Tab of the NotebookBar in Impress Online 21.11:

This concerns:

- The "New Page" & "Duplicate Page" buttons should be labelled "Slide"
- "Slide Layout" button
-  The "Reset zoom" button has inconsistent capitalization  

**To Reproduce**
Steps to reproduce the behaviour:
1. Open a presentation in Online 21.11
2. Click on the Layout tap in the NotebookBar

**Expected behaviour**
- The "New Page" & "Duplicate Page" buttons should be labelled "New Slide" & "Duplicate Slide"
- "Slide Layout" should be named "Sidebar" (like in the other modules)
- "Reset zoom" should be "Reset Zoom" for consistent capitalization across the buttons 

**Actual behaviour**
Currently, inconsistent/ambiguous labelling of buttons

**Screenshots**
![image](https://user-images.githubusercontent.com/51329893/139649641-31652451-8161-4245-871b-3a9bf9648ce3.png)

**Desktop (please complete the following information)**
 - Browser: [Firefox on Linux]
 - Version: [90.0.2]

**Smartphone (please complete the following information)**
 - 

**Additional context**
-

Ezinnem:
@pedropintosilva, @Andreas-Kainz

Issue #3489, opened 2021-10-26
Calc: remove sheet confirmation dialog has extra space


pedropintosilva:
**Describe the bug**
Calc: remove sheet confirmation dialog has extra space right before `?` which:
1. In terms of style (written content) it shouldn't be there
2 Plus it creates this cases where the sentence breaks in an undesirable manner

**To Reproduce**
1. Remove a sheet
2. See dialog

**Screenshots**
![image](https://user-images.githubusercontent.com/65948705/138867033-5ece2a16-3e1d-4aa9-a3e9-295ee2760685.png)

**Desktop**
LOOLWSD
* 21.11-0 (git hash: 81ea2266)
* Served by: 94c16d16

LOKit
* built on a great technology base
* Collabora Office 21.06.5.1 (git hash: 1d54cb6)

PR #3533, opened 2021-11-02
loleaflet: fix: UX inconsistency

Rash419:
Change-Id: I627103970ce34cf1612dfe1c2f072e4f58596965

* Resolves: #3527 #3489 
* Target version: master


pedropintosilva (changes requested):
* Calc: Tested, works, fixes: https://github.com/CollaboraOnline/online/issues/3489
* Impress: Tested, some more places in need of a string change: 
Insert tab: image
also and even though in the issue https://github.com/CollaboraOnline/online/issues/3527 Marc says to change from `Slide Layout` to `Sidebar` please revert back to `slide layout` as it was before (this is important because even though the icon is of a sidebar the string `Slide Layout` helps with discoverability and tells the user where the slide layout options are.

pedropintosilva (approved):

Issue #3513, opened 2021-10-28
Calc:Comment button active when it should have not


NISZOpenSourceTeam:
Describe the bug
After a Chart insertion, comment button is active. When clicking on it nothing happens.
1. Insert a chart
2. Click on chart
3. Go to Review
4. Click on the comment button

**Expected behavior**
It should not be active the button

**Actual behavior**
Comment button is active

**Desktop
 - OS: Windows
 - Browser: Chrome
 - Collabora online version:
              LOOLWSD
              21.06.1.0 (git hash: 5a0738ff)
              Served by: b7a30aab
              LOKit
              Collabora Office 21.06.4.1 (git hash: 90aa650)
              "Ubuntu 20.04.3 LTS"
              Nextcloud version 22.1.1

pedropintosilva:
Thanks @NISZOpenSourceTeam for reporting it!

PR #3534, opened 2021-11-02
loleaflet: calc: fix: comment button active when there is graphicselection

Rash419:
…ction

Change-Id: Ibd19ead342a50658eae76808f2af81f378ac17cc

* Resolves: #3513 
* Target version: master


pedropintosilva (approved):
thanks, tested in both UI modes.

Note: before insert comment looked like it was disabled but was still possible to click it, this PR fixes that

PR #3535, opened 2021-11-02
Update indicator in statusbar on permission change

eszkadev:


eszkadev:
This can be tested with *.csv file on Nextcloud, it opens in read-only mode but then it can be switched to edit mode

pedropintosilva:
Thanks

PR #3541, opened 2021-11-03
Fix closing CSV dialog returns an exception

merttumer:
It should return to the integration instead

Change-Id: I1b2ce5013365faa4f67263866764ed4bd34277d7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/loleaflet/src/control/Control.JSDialogBuilder.js
+++ b/loleaflet/src/control/Control.JSDialogBuilder.js
@@ -43,6 +43,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 	_currentDepth: 0,
 
+	_firstDialogHandled: false,
+
 	setWindowId: function (id) {
 		this.windowId = id;
 	},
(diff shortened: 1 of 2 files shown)


loleaflet/src/control/Control.JSDialogBuilder.js
```diff
@@ -261,6 +263,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var encodedCommand = data.replace(' ', '%20');
 			builder.map.sendUnoCommand(encodedCommand);
 		} else if (object) {
+			// CSV and Macro Security Warning Dialogs are shown before the document load
+			// In that state the document is not really loaded and closing or cancelling it
+			// returns docnotloaded error. Instead of this we can return to the integration
+			if (!builder.map._docLoaded && !this._firstDialogHandled && (object.id === 'cancel' || eventType === 'close')) {
```
eszkadev, line 269: Some dialogs provide buttons with standard "response code" (help, ok, cancel, etc.).
For example see Fontwork dialog, after cancel button is clicked, this is emitted:
control: 'responsebutton' id:'cancel' event: 'click' state: '0'

we can check if `type` is `responsebutton` and `data` is `0` (cancel) or `7` (close)
number values: https://opengrok.libreoffice.org/xref/core/include/vcl/vclenum.hxx?r=7ccfae54&fi=RET_CANCEL#RET_CANCEL

this can be additional alternative condition for that feature, in case `object.id === 'cancel'` will not be true for some dialog.

eszkadev (approved):
Thanks, now looks good :)

PR #3544, opened 2021-11-03
debug.html:use iframe as a target after submiting form

Rash419:
Change-Id: I4b4e59721fb1deec709bc778a8e59a52fb466a3a

* Target version: master

### DIFF
--- a/cypress_test/integration_tests/mobile/writer/annotation_spec.js
+++ b/cypress_test/integration_tests/mobile/writer/annotation_spec.js
@@ -19,6 +19,8 @@ describe('Annotation tests.', function() {
 
 
 	it('Saving comment.', function() {
+		cy.wait(1000);
+
 		mobileHelper.insertComment();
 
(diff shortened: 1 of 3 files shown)


kendy (approved):
The code looks great, thank you!  Would be good to improve the commit message though - instead of describing what the patch does (can be read from the code), focus on what issue it is solving.  Can you update that please?

Rash419:
@kendy done : )

Rash419:
> @kendy done : )



> The code looks great, thank you! Would be good to improve the commit message though - instead of describing what the patch does (can be read from the code), focus on what issue it is solving. Can you update that please?

@kendy done

kendy (approved):
Approving, thank you! :-)

thebearon:
@Rash419 could you please also check the SSL disabled case? (one option is to use the `--disable-ssl` configure parameter, but it's easier to just switch the `enable` config item in `loolwsd.xml` to `false`)

PR #3545, opened 2021-11-03
Don't show disconnected message when DisableInactiveMessages

eszkadev:
Snackbar message for 'disconnected' or 'connected again'
shouldn't be shown if DisableInactiveMessages is set.

### DIFF
--- a/loleaflet/src/core/Socket.js
+++ b/loleaflet/src/core/Socket.js
@@ -502,7 +502,10 @@ app.definitions.Socket = L.Class.extend({
 						reloadMessage = _('Server is now reachable...');
 
 					var reloadFunc = function() { window.location.reload(); };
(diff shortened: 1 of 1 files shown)


kendy (commented):
When they are not shown (due the DisableInactiveMessages), I guess we should still send some kind of postMessage so that the integration can handle those - or is that not the case for these two?

eszkadev:
@kendy there is sending of `Reloading` message here.

For socket disconnection I fear it can be send in too many cases so it would be not usefull at all: socket can be closed even as a part of a 'normal' process like on idle. Maybe also when renaming and 'save as' - when we change document?

kendy (approved):
OK, good then, thank you! Approving :-)

PR #3549, opened 2021-11-03
Don't waste space for comment section

eszkadev:
Use trick to set 1px width for comment section so it doesn't
reserve space which can be used for document itself.

Thanks to that we don't see grey area on the right side when
zoom level is big enough so document is shown in a full width.

Position correctly HTML elements for comments so they are still visible.


gokaysatir (approved):
thanks for the fix:)

PR #3551, opened 2021-11-03
Unselect comments when collapsing to bubbles

eszkadev:
It prevents us from case when bubble is "higlighted"
but we cannot open comment preview.


kendy (approved):
Cool, thank you!

Issue #3520, opened 2021-10-29
Loading/Editing a spreadsheet file full with comments is slow.


gmasei11:
**Describe the bug**
Loading/Editing a specific spreadsheet file full with comments is slow. This seems to be a regression when compared to 6.4.0.

**To Reproduce**
1. Open the [Test71.ods](https://forum.collaboraonline.com/uploads/short-url/btgFJMKR6vHBa24RixYKZNiV0at.ods)
2. Select multiple times cells, rows, columns

**Expected behavior**
The load is slow but the UI works fine.

**Actual behavior**
After the load spinner disappears the page hangs for a while with a partially loaded UI, then cells appear but without data and then, after a while, the cells are filled with the data. Selecting cells, rows, columns is slow. In general, the edit is slow.

**Desktop (please complete the following information)**
 - OS: Ubuntu 20.04.3 LTS
 - Browser: chrome, firefox
 - Version: 6.4.10 debug, 2021 debug

**Additional context**
https://github.com/CollaboraOnline/online/discussions/3461
In 6.4.0 works fine.
The problem was mentioned first here: https://forum.collaboraonline.com/t/collabora-weekly-meeting-46/687#test-fest-gabriel-1 where you can find attached the file for testing.

Ezinnem:
I did not notice any slowness with the loading of the file using LOOLWSD 6.4.13 (git hash: [4e5bac7f](https://github.com/CollaboraOnline/online/commits/4e5bac7f)).

pedropintosilva:
thanks @Ezinnem  but did you try with **Version: 6.4.10** as reported by @gmasei11 ?

gmasei11:
Thanks @Ezinnem for your feedback. I still think that there is a drop in performance in 6.4.10 in comparison with 6.4.0. Indeed it is not so dramatic like in co 2021 and it is more visible in Firefox than Chrome. This could suggest that the root of the issue comes from an earlier version than co 2021. See the videos below:

6.4.0 Firefox


https://user-images.githubusercontent.com/73231749/139674778-55941671-0210-43b4-a2db-419a6417e5a7.mp4

6.4.10 Firefox


https://user-images.githubusercontent.com/73231749/139674981-f4eb293e-a9b0-4c54-88da-78eb60b8f17f.mp4

Ezinnem:
The performance issue has been worked on recently. Please see #3205.
I think that is the reason this issue is not reproducible using the recent 6.4.13 version.

[...]

PR #3556, opened 2021-11-03
Fix slow loading/editing of sheet with ~800 comments [master]

dennisfrancis:
Profiling points to updateReplyCount() as the bottleneck where it does
O(N^3) work where N is the number of comments. The innermost loop is
used to find the index of a comment-section in an array using its 'id'.
Turns out that flattening this loop to a < log(N) lookup via a js map
from id -> index pretty much cures the perceived 'slowness'. But we need
to ensure the correctness of the map by updating it whenever
'commentList' is mutated.

Time to load the document with 800 comments (arranged in 50 rows x 16
cols table):

Before the patch: 1min 31 sec
After the patch:  3.5 sec

After-load performance (comment popup/hide/scroll) with this patch is
similar to how it feels in co-6-4 which does not have this issue.

Change-Id: I4c47f14767cac4c7e9744344d24f646dd27302a6

* Resolves: #3520
* Target version: master 

### Checklist

### DIFF
--- a/loleaflet/src/layer/tile/CommentListSection.ts
+++ b/loleaflet/src/layer/tile/CommentListSection.ts
@@ -64,6 +64,9 @@ class CommentSection {
 	setPosition: (x: number, y: number) => void;
 	map: any;
 
+	// To associate comment id with its index in commentList array.
+	private idIndexMap: Map<any, number>;
+
 	constructor () {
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
Thanks for the fix:)

PR #3566, opened 2021-11-04
private/kendy/master

kendy:
- rename loleaflet.html -> cool.html
- Rename the loleaflet.html usage in the recently introduced debug.html


kendy:
Merged via another PR.

PR #3569, opened 2021-11-04
Don't output errno information when a syscall hasn't failed

kendy:
Based on Tor's patch...

Signed-off-by: Jan Holesovsky <kendy@collabora.com>
Change-Id: I04f173816d5170a666437d6842f8657798c6c19f

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1112,10 +1112,14 @@ public:
                 len = readData(buf, sizeof(buf));
                 last_errno = errno;
 
-                LOG_TRC('#' << getFD() << ": Read incoming data " << len << " bytes in addition to "
-                            << _inBuffer.size() << " buffered bytes ("
-                            << Util::symbolicErrno(last_errno) << ": " << std::strerror(last_errno)
-                            << ')');
+                if (len >= 0)
+                    // errno meaningless
+                    LOG_TRC('#' << getFD() << ": Read incoming data " << len << " bytes in addition to "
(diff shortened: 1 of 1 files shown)


tml1024:
But why bother outputting a negative len, indicating error, in `Read incoming data " << len << " bytes` and `Wrote outgoing data " << len << " bytes` ? Isn't it enough in that case to just say "Failed to read/write incoming data" and then print the symbolic errno?

Ashod:
I feel we are overthinking this. It's a trace log in a heavily called function. If we output errno in trace (which helps debugging) and it says 'success' then no harm done (even if it is unnecessary).

But now we have one additional conditional and _twice_ the code size in that function. This will surely have a negative impact on performance and doesn't help anyone at all (even when trace is disabled).

I feel this is a nit-picking exercise to make a trace log "cleaner" at the expense of more code lines and slower performance. (I'm sorry if I don't see the benefit. But I'm happy to be convinced this change is worthwhile.)

tml1024:
You see now @kendy why I gave up?

tml1024:
Not saying anything then.

kendy:
The 'if' comparing to 0 (plus a potential jump) has a totally marginal cost compared to all the stuff that is done for the logging itself - with all the string operations, mapping of errno numbers to their string representations, and actual output to the log or screen. The size is small & fits caches.

But I said - I don't want to hear about this "problem" any more; it has already been a very expensive exercise already, let's just abandon and forget.

PR #3572, opened 2021-11-04
private/pedro/color indicators ui fixes

pedropintosilva:
- Mobile: Remove unused rules, were affecting desktop
- Color picker: UI, fix radius n add visual padding
- Color picker: Don't set indicator's border so bright


pedropintosilva:
Fixing:
image

PR #3574, opened 2021-11-05
Introduce security.enable_metrics_unauthenticated config item

timar:
When enabled, the /lool/getMetrics endpoint will not require authentication.
By default authentication is required and this setting is set to "false".

Change-Id: I801130cf552eb14c231fcc0a0bdd39d9ebb6db7f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (changes requested):
Patch looks good.
I think with security code though we want to be very verbose and easy to read:

Can we have:

bool skipAuthentication = getConfigSetting();
if (!skipAuthentication) {
...
}

so it is better self documenting - also I would add some /** WARNING **/ blocks around that code since it is an authentication / security point.

Otherwise lets get this in ASAP - thanks ! =)

PR #3587, opened 2021-11-09
Increase button click target in toolbars

juliushaertl:
* Target version: master 

### Summary

As from previous discussion with @pedropintosilva this PR increases the click/touch target of toolbar icons to at least 32px. Furthermore on mobile the fixed height on the bottom toolbar was removed to make it dynamically adjust to the button size (currently at 42px)

#### Desktop
Before:
image

After:
image

#### Mobile

Before:
image

After:
image

### Checklist

### DIFF
--- a/loleaflet/css/device-mobile.css
+++ b/loleaflet/css/device-mobile.css
@@ -357,6 +357,7 @@ button.vex-dialog-button-secondary.vex-dialog-button.vex-last {
 	padding-top: 0px;
 	padding-bottom: 0px;
 	background: transparent;
+	height: 40px;
 }
 #formulabar {
 	padding: 0px !important;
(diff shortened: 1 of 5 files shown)


pedropintosilva:
thanks @juliushaertl  I think on mobile makes sense and it looks nice : ) on desktop I'm honestly not sure:
* I have tested on laptop and it seems we loose a big chunk of the icons (that are now hidden and forced to be scrolled to access them) 
* The misalignment (difference in size) between btns and select elements (dropdown menus) are now bigger 
*  Additionally maybe the toolbar can seem now a bit blocky

juliushaertl:
>  have tested on laptop and it seems we loose a big chunk of the icons (that are now hidden and forced to be scrolled to access them)
> Additionally maybe the toolbar can seem now a bit blocky

Let me experiment a bit more with some smaller increments on desktop then, maybe also just having the 1px spacing between buttons is enough. It also seems there could be some additional space saving if reducing the left padding of the toolbar itself a bit and aligning the first icon with the filetype icon above.

> The misalignment (difference in size) between btns and select elements (dropdown menus) are now bigger

True, might be nice I guess to ensure those have the same height (no matter on what size of the icons is picked.

PR #3596, opened 2021-11-10
quarantine: automatically delete quarantined files after specified time

lpranam:
Change-Id: Idabef2aec46d6b7c194ef06f4235b60d6ad29e4a

* Target version: master 

### Checklist

### DIFF
--- a/wsd/QuarantineUtil.cpp
+++ b/wsd/QuarantineUtil.cpp
@@ -90,19 +90,26 @@ namespace Quarantine
 
         std::sort(files.begin(), files.end());
 
+        std::size_t timeLimit = LOOLWSD::getConfigValue<std::size_t>("quarantine_files.expiry_min", 30);
+        const auto timeNow = std::chrono::system_clock::now();
+        const auto ts = std::chrono::duration_cast<std::chrono::seconds>(timeNow.time_since_epoch()).count();
+
         std::size_t currentSize = quarantineSize();
-        std::size_t index = 0;
-        while (currentSize >= sizeLimit && !files.empty())
+        auto index = files.begin();
+        while (index != files.end() && !files.empty())
         {
-            FileUtil::Stat file(LOOLWSD::QuarantinePath + files[index]);
-            if(file.hardLinkCount() != 1)
+            FileUtil::Stat file(LOOLWSD::QuarantinePath + *index);
+            const auto modifyTime = std::chrono::duration_cast<std::chrono::seconds>(file.modifiedTimepoint().time_since_epoch()).count();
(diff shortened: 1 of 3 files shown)


tml1024:
Looks fine, but will check in a "make run".

wsd/QuarantineUtil.cpp
```diff
@@ -78,6 +78,48 @@ namespace Quarantine
         return size;
     }
 
+    void deleteExpiredFiles(std::vector<std::string> &files)
+    {
+        if (!isQuarantineEnabled())
+            return;
+
+        // parameter is used in this function to avoid creating and sorting file list multiple times
+        if(files.empty())
```
mmeeks, line 87: We have a single call site surely ? and by the time we get here it has a list of all the files surely (?)

wsd/QuarantineUtil.cpp
```diff
@@ -78,6 +78,48 @@ namespace Quarantine
...
+        }
+
+        std::size_t timeLimit = LOOLWSD::getConfigValue<std::size_t>("quarantine_files.expiry_min", 30);
+        const auto timeNow = std::chrono::system_clock::now();
+        const auto ts = std::chrono::duration_cast<std::chrono::seconds>(timeNow.time_since_epoch()).count();
+
+        std::size_t index = 0;
+        while (!files.empty() && index < files.size())
+        {
+            FileUtil::Stat f(LOOLWSD::QuarantinePath + files[index]);
```
mmeeks, line 100: In general it's better to do the syscall just once to stat the file to get its information; can we not factor out some 'bool deleteFile (name, timeNow, &totalSize); type function that we can scan do a single pass through the files with ? =)

wsd/QuarantineUtil.cpp
```diff
@@ -90,6 +132,8 @@ namespace Quarantine
 
         std::sort(files.begin(), files.end());
 
+        deleteExpiredFiles(files);
+
         std::size_t currentSize = quarantineSize();
         std::size_t index = 0;
         while (currentSize >= sizeLimit && !files.empty())
```
mmeeks, line 139: So doing a single state in this loop below; and having some nice intermediates to flag what we are detecting and doing some easy-to-read logic around that ?

bool isExpired = <timestamp-check>
bool isInUse = hardlink-count > 1

if (isExpired || !inUse || totalSize > whatever) ...
   erase.

mmeeks (commented):
A few thoughts; it looks like it would work - so not against getting it in as-is; but can improve maintainability here quite easily I think.

timar (approved):

PR #3597, opened 2021-11-11
docker: Make Notebookbar the default again

thebearon:
...and user interface mode parametrizable.

Change-Id: Id4538f27f5aef9190c546ec8dc1d505042837aeb

* Target version: master


timar:
With this patch we will unconditionally write the loolwsd.xml which is not possible in all use cases, so I expect problems. I'm also not fan of the old style perl regex based xml manipulation, we only keep it around for backward compatibility. Finally (or I would rather say luckily) we will not release CODE 6.4 any longer. We will have CODE 21.11 by end of November. Thanks Aron for taking care of this and you are right, but considering all above I do not recommend to merge this patch.

PR #3600, opened 2021-11-11
Remove config option loleaflet_html

timar:
This means that we will no longer read loleaflet_html from old config files, which is
a good thing in 99.9% of the cases, because it contains a wrong default: loleaflet.html
For those integrators, who used this setting, we will offer an alternative slution later.

Change-Id: Ia82f812a5d3e522a9d33e7fcafcaa45130312d60


kendy (approved):
Looks great, thank you!

PR #3616, opened 2021-11-12
Spinfield: improve input number (webkit)

pedropintosilva:
On webkit based browsers spinfield controls (up an down) are hidden by
default and only appear on hover on top of that they look a bit wonky
(too small and not well positioned)

This improves that by sill using the same web native control while
styling those pseudo (vendor) classes. Extras:
- Mouse cursor change
- Different opacity values (so we retaining that visual difference
	between hover and normal state)
- Increase size of those arrows

Change-Id: I07aacfc387cbf42efec1d302094bbfdfae4da4ed


eszkadev (approved):
nice!

PR #3627, opened 2021-11-15
rename: loolconfig -> coolconfig

merttumer:
Change-Id: Ied23cb46a668f754931fc125745b674b546a0195

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ loolconvert
 loolsocketdump
 loolstress
 loolconfig
+coolconfig
 loolforkit-nocaps
 loadtest
 unittest
--- a/man/coolconfig.1
+++ b/man/coolconfig.1
@@ -1,0 +1,27 @@
+.TH COOLCONFIG "1" "May 2018" "coolconfig" "User Commands"
+.SH NAME
+coolconfig \- Configuration tool for Collabora Online.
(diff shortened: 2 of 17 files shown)


man/coolconfig.1
```diff
@@ -0,0 +1,27 @@
+.TH LOOLCONFIG "1" "May 2018" "Coolconfig" "User Commands"
```
timar, line 1: LOOLCONFIG here

.gitignore
```diff
@@ -79,7 +79,7 @@ loolmap
 loolconvert
 loolsocketdump
 loolstress
-loolconfig
```
kendy, line 82: Would be good to still keep loolconfig, but only as a symlink to coolconfig (created in a Makefile) so that it can be packed into a -compat package.

Otherwise looks good to me, thank you!

man/coolconfig.1
```diff
@@ -0,0 +1,27 @@
...
+.SH NAME
+Coolconfig \- Configuration tool for Collabora Online.
+.SH SYNOPSIS
+Coolconfig COMMAND [OPTIONS]
```
timar, line 5: coolconfig - lower case!

PR #3631, opened 2021-11-15
Default to automatically cleaning up high resource consumers.

mmeeks:
Change-Id: I570d13e696e7e2dc01dd29751df26eb532fa1a76

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
@mmeeks the setting will change only for new installations, because package upgrade does not update config file. If you have to change these as well:
```
wsd/LOOLWSD.cpp:            { "per_document.cleanup[@enable]", "false" },
wsd/LOOLWSD.cpp:    docCleanupSettings.setEnable(getConfigValue<bool>("per_document.cleanup[@enable]", false));
```
then old users who do not have this setting in their config file, will get the new default. Unfortunately we cannot change the default for old users, who already have `<cleanup desc="Checks for resource consuming (bad) documents and kills associated kit process. A document is considered resource consuming (bad) if is in idle state for idle_time_secs period and memory usage passed limit_dirty_mem_mb or CPU usage passed limit_cpu_per" enable="false">` (the old default) in their config file. It is bad design from start.

mmeeks:
Fair enough - but I'm happy to leave the legacy users with their existing setting I think; that's fine.

PR #3632, opened 2021-11-16
docker: Make Notebookbar the default again

thebearon:
...and user interface mode parametrizable.

Change-Id: Id4538f27f5aef9190c546ec8dc1d505042837aeb

* Target version: master


timar (approved):

litvin-Millogic:
Could you please explain why you are insisting on this approach?  Why not simply change the default in loolwsd.xml and that's all?

At the very least, can I suggest that you please add a comment in loolwsd.xml that alerts the user to the possibility that the UI setting may have been overridden by a passed-in parameter?  This would potentially save many users many hours of frustration...

PR #3644, opened 2021-11-16
comments: show full comments on the right when enough space

eszkadev:
Resize comments section when enough space to show comments on the
right side. Document is not centered in that case and we can use
free space from the left side.


eszkadev (commented):
PR still has to be improved, I tested it more on other computer and seems to cause a lot of flickering after rebase.

PR #3662, opened 2021-11-17
Fix debian dockerfile: explicitly set buster as source since is now oldstable

mmaridev:
* Target version: master 

### Summary

Set buster as source since is now oldstable

### Checklist


thebearon:
Hi Marco, what would be the reason behind using the older Debian version compared to the newer one?

mmaridev:
Hi Aaron,
basically that the build is currently broken. Some of the packages included in the dockerfile are no longer available in debian 11.
Since debian 10 will still be supported for a while, setting the dockefile to buster is a good workaround.

PR #3690, opened 2021-11-19
private/pedro/about dialog fix wrapping on mobile

pedropintosilva:
- About dialog: fix text wrapping on mobile
- About dialog: Use different chars around git hash


pedropintosilva:
image

pedropintosilva:
image

amended last commit, the white space was being inserted but than right after removed by the browser

timar (approved):

PR #3691, opened 2021-11-19
private/kendy/userinterfacedefault

kendy:
- Make Notebookbar the default
- Add a value "default" for the user interface


mmeeks:
Works nicely for me - thanks Kendy ! =) great to see it pass CI.

PR #3698, opened 2021-11-22
jsdialog: use unique ids for unobuttons

eszkadev:
now all uno buttons with the same command will share class not id:
.unoXXX  not  #XXX

this is needed for a11y


eszkadev:
@pedropintosilva FYI
What I see is that we don't use too much of css rules for specific uno buttons using ids.
(one I can see is InsertGraphic in branding). Worth checking if everything looks ok.

gokaysatir (dismissed):
Thanks for the fix :)

eszkadev:
but we have to fix cypress, it uses old ids...

pedropintosilva (approved):
Thanks @eszkadev !

PR #3700, opened 2021-11-22
Fix WOPISrc generation after save as

eszkadev:
This fixes regression from:
https://github.com/CollaboraOnline/online/commit/293081a6317f551315c49055475c33bb0fc9792e
commit 293081a6317f551315c49055475c33bb0fc9792e
leaflet: centralize wopiSrc URL forming

global.wopiSrc is used to generate new URL but it is not updated
after we do save as action. It was always equal to initial value
which was set on the app load. Because of that we generated URL
with not matching file ids in the WOPISrc parameter.

WOPISrc parameter is used for load balancing so it was possible
that after save as the new session will be directed to the different
node.


kendy (approved):
Sounds great, thanks a lot for the fix!

PR #3701, opened 2021-11-22
feedback: delay 2 days and count open documents

hcvcastro:
Change-Id: I0966b257f3aae1376ff5ee06d286e03886a5e0da

### DIFF
--- a/browser/src/map/handler/Map.Feedback.js
+++ b/browser/src/map/handler/Map.Feedback.js
@@ -33,10 +33,25 @@ L.Map.Feedback = L.Handler.extend({
 
 	onDocLoaded: function () {
 		if (window.localStorage.getItem('WSDFeedbackEnabled') !== 'false') {
+			var laterDate = new Date();
 			var currentDate = new Date();
-			var laterDate = window.localStorage.getItem('WSDFeedbackLaterDate');
+			var timeValue = window.localStorage.getItem('WSDFeedbackLaterDate');
+			var docCount = window.localStorage.getItem('WSDFeedbackCount');
 
-			if (laterDate !== currentDate.toDateString())
+			docCount = parseInt(docCount);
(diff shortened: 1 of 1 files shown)


browser/src/map/handler/Map.Feedback.js
```diff
@@ -84,15 +97,17 @@ L.Map.Feedback = L.Handler.extend({
...
 			window.localStorage.setItem('WSDFeedbackEnabled', 'false');
+			window.localStorage.removeItem('WSDFeedbackCount');
 			this._iframeDialog.remove();
 		} else if (data == 'feedback-later') {
 			var currentDate = new Date();
 			this._iframeDialog.remove();
-			this._map.options.feedbackTimeout = 86400000;
-			window.localStorage.setItem('WSDFeedbackLaterDate', currentDate.toDateString());
-			setTimeout(L.bind(this.onFeedback, this), this._map.options.feedbackTimeout);
+			currentDate.setDate(currentDate.getDate() + 2);
```
kendy, line 105: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate says getDate() returns the day of the month; so once it is set at the end of the month (ie. it becomes eg. 32), the if (... && currentDate > laterDate) above will never be true.

Would be better to use days since the start of epoch; like https://stackoverflow.com/questions/12739171/javascript-epoch-time-in-days .

Also I'd prefer to store the day of pressing the "Later" button, and check like "currentDay > laterButtonPressedDay + 2" so that it is easy to update the 2 to something else should it be too early to check again.

kendy (changes requested):

hcvcastro:
Changed to 5 days

kendy (approved):
Thank you very much!

PR #3702, opened 2021-11-23
notebookbar: activate in core when default ui mode

eszkadev:
Core needs "notebookbar" ui mode string to activate
notebookbar widgets. If we pass "default" notebookbar
is not activated and style previews widget doesn't work.

followup for https://github.com/CollaboraOnline/online/commit/b4f588ea3cadad124f1c99718fb7f63e82c91953


kendy:
Thanks a lot!

PR #3705, opened 2021-11-23
NB: Fix big unotoolbutton dropdown with image

pedropintosilva:
Buttons with image and text stacked up plus arrow:
Some browsers (chrome/chromium) were not displaying the arrow correctly,
it was appearing as a gray block and also badly positioned.

This change fixes that.

Change-Id: I144aa39eb17361309683592928720986e8476f36


pedropintosilva:
on chrome before:
image

now:
image

PR #3706, opened 2021-11-23
jsdialog: run postProcess also on containers

eszkadev:
- added TODO: identify containers correctly by id from JSON not
  with additional prefixes, this needs revisiting css rules...
  so do not do it now
- run postProcess on containers to hide them if needed


merttumer (approved):
Works as described

PR #3708, opened 2021-11-24
wsd: in debug build on saving file divide request data into chunks an…

Rash419:
…d write to file

Change-Id: Ie0f09d24cb3ca25c7f28b608e283342d966df71f

* Target version: master 

@Ashod  I actually tested with small file ( < 1 MB) and large file  of 16 mb it works for me

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -458,14 +458,26 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request,
             }
 
             std::streamsize size = request.getContentLength();
-            char buffer[size];
-            message.read(buffer, size);
-            LocalFileInfo::fileInfoVec[i].fileLastModifiedTime = std::chrono::system_clock::now();
-
             std::ofstream outfile;
             outfile.open(LocalFileInfo::fileInfoVec[i].localPath, std::ofstream::binary);
-            outfile.write(buffer, size);
+            constexpr std::streamsize bufferSize = 1024 * 1024;
+            std::unique_ptr<char[]> buffer(new char[bufferSize]);
+            if (size > bufferSize)
+            {
+                int numberOfChunks = (size / bufferSize) + 1;
+                for (int j=0; j<numberOfChunks; j++) {
+                    message.read(buffer.get(), bufferSize);
+                    outfile.write(buffer.get(), size);
+                }
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


wsd/FileServer.cpp
```diff
@@ -458,15 +458,27 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request,
...
             outfile.open(LocalFileInfo::fileInfoVec[i].localPath, std::ofstream::binary);
-            outfile.write(buffer, size);
+            constexpr std::streamsize bufferSize = 1024 * 1024;
+            std::unique_ptr<char[]> buffer(new char[bufferSize]);
+            if (size > bufferSize)
+            {
+                int numberOfChunks = (size / bufferSize) + 1;
+                for (int j=0; j<numberOfChunks; j++) {
+                    message.read(buffer.get(), bufferSize);
+                    outfile.write(buffer.get(), size);
```
kendy, line 470: The outfile.write(buffer.get(), size); is strange - didn't you mean to use bufferSize there?  But with that, I fear this will enlarge the file - it won't stop on the exact amount, but a multiply of bufferSize? 

You should either use tellg() or gcount() somewhere - to also handle the case when read() reads less (not sure if that can happen - but normally when implementing a buffered read, that is something that has to be considered).

Fixing the above will also help you to get rid of the if (size > bufferSize) ... else ... I think, to make the code more compact...

kendy (changes requested):

Ashod (commented):
Thanks @Rash419.

Did you have a reason to use chunking? I don't think it's needed. For me, simply using dynamic allocation fixes the issue for files 100 Mb in size. Have you tried anything larger that failed? See my PR #3711.

Thanks!

Rash419:
@Ashod I miss understood your comment I thought it needs chunking so that there are no memory leaks. Anyways thanks for the fix I am closing this PR . Sorry for the noise @kendy @Ashod

Ashod:
> @Ashod I miss understood your comment I thought it needs chunking so that there are no memory leaks. Anyways thanks for the fix I am closing this PR . Sorry for the noise @kendy @Ashod

@Rash419 No worries! Apologies if I implied it was a leak. Hope it's clear that the issue was allocating a large array on the stack. By the way, it's still worth chunking, as Kendy says. For example, we should be able to use std::copy here, which will copy in chunks. Feel free to add that, if you want.

Thanks!

PR #3709, opened 2021-11-24
Welcome dialog: new layout

pedropintosilva:
- Add slider
	- Less content and easier to scan/read
- More clear call to actions
	- But less CAT per screen / slide

Change-Id: I5faf9ac5ef3bfa57f19d1b9f5f2bdcee4c91ecb7


timar:
For translation, you do not have to do anything special. Weblate takes welcome.html and uses its HTML filter to parse it. We do not us po files here.

kendy:
> For translation, you do not have to do anything special. Weblate takes welcome.html and uses its HTML filter to parse it. We do not use po files here.

@timar: Ouch, too bad I didn't read this earlier; could have introduced the variable only for the version and handle it some way in the JS, instead of moving all the texts to the JS itself.  But they are completely reworked anyway, so hopefully I've caused no real damage; sorry about that...

timar:
> > For translation, you do not have to do anything special. Weblate takes welcome.html and uses its HTML filter to parse it. We do not use po files here.
> 
> @timar: Ouch, too bad I didn't read this earlier; could have introduced the variable only for the version and handle it some way in the JS, instead of moving all the texts to the JS itself. But they are completely reworked anyway, so hopefully I've caused no real damage; sorry about that...

@kendy no, actually you simplified things, the fewer translation projects the better.

PR #3718, opened 2021-11-25
WopiStorage: log what is the final suggested target for save-as

vmiklos:
unit-wopi-saveas sometimes fails on me with sanitizers. The reason is
still not yet clear (the test is quite stable without them), but at
least try to find out if the suggested target is already wrong when we
put them to the HTTP header or it goes wrong later.

Change-Id: I25f65fd6e211022aa93c699cdaef49eb8978fdee


vmiklos:
Only code scanning failed in unrelated files, let's get this in.

PR #3719, opened 2021-11-25
xlsb type should be readonly and added option to saveas odf

merttumer:
Change-Id: I1c18ca74d1436f669f0415285efee78abb2c1e5a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Permission.js
+++ b/browser/src/control/Permission.js
@@ -4,25 +4,31 @@
  */
 /* global app $ _ vex */
 L.Map.include({
+	readonlyStartingFormats: {
+		'txt': { canEdit: true, odfFormat: 'odt' },
+		'csv': { canEdit: true, odfFormat: 'ods' },
+		'xlsb': { canEdit: false, odfFormat: 'ods' }
+	},
+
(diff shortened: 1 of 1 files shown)


browser/src/control/Permission.js
```diff
@@ -84,18 +90,27 @@ L.Map.include({
...
+			return false;
+		return true;
 	},
 
 	_proceedEditMode: function() {
+		var fileName = this['wopi'].BaseFileName;
+		if (fileName) {
+			var extension = this._getFileExtension(fileName);
+			var extensionInfo = this.readonlyStartingFormats[extension];
+			if (!extensionInfo.canEdit)
```
eszkadev, line 111: extensionInfo can be `undefined` I think for formats not present in `readonlyStartingFormats`

eszkadev: ok, false alarm, this is called only when `_shouldStartReadOnly` is true...

eszkadev (approved):
Looks good, and works

Issue #2538, opened 2021-06-03
How do I disable the option to save a document to a new file when the orginal file has changed?


objectivejs:
The WOPI server I'm working on deals with timestamps. If I edit a file, touch it on the server then try to save it, a box pops up with the option to overwrite the file, abandon the modifications in which case the file is reloaded, or to save the file to another file. This last option should not be possible since UserCanNotWriteRelative is set to true on the original file. If I do save it to another file, the original file is overwritten.

Ezinnem:
Fix on #3750

PR #3728, opened 2021-11-26
fix: dont give an option to saveas if UserCannotWriteRelative is true

merttumer:
On documentconflict, we give 3 options;
discard, overwrite, save-as
the third option should be impossible if the user is not allowed

Change-Id: I30571d885259a02c4170ccf7e291d4456bf1880c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -818,27 +818,32 @@ app.definitions.Socket = L.Class.extend({
 
 				vex.closeAll();
 
+				var dialogButtons = [
+					$.extend({}, vex.dialog.buttons.YES, { text: _('Discard'),
+						click: function() {
+							this.value = 'discard';
+							this.close();
+						}}),
+					$.extend({}, vex.dialog.buttons.YES, { text: _('Overwrite'),
+						click: function() {
+							this.value = 'overwrite';
+							this.close();
+						}})
+				];
+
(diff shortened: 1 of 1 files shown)


merttumer:
Forgot to mention #2538

browser/src/core/Socket.js
```diff
@@ -833,12 +833,12 @@ app.definitions.Socket = L.Class.extend({
...
 						                                      }}),
-						$.extend({}, vex.dialog.buttons.YES, { text: _('Save to new file'),
+						(that._map['wopi'].UserCanNotWriteRelative ? undefined : $.extend({}, vex.dialog.buttons.YES, { text: _('Save to new file'),
 						                                      click: function() {
 							                                      this.value = 'saveas';
 							                                      this.close();
-						                                      }})
-					],
+						                                      }}))
+					].filter(Boolean),
```
eszkadev, line 841: I think this code is a bit complicated. To make it more readable maybe we can construct that array before `vex.dialog.open` call.
eg.
```
var buttons = [ <first entry>, <second entry> ];
if (!that._map['wopi'].UserCanNotWriteRelative)
    buttons.push( <optional entry> )
```

eszkadev (changes requested):

eszkadev (approved):
Thanks, works as expected.

PR #3730, opened 2021-11-26
Use `cp -a` instead of `cp -ar`.

arrowd:
According to Linux man, `-r` and `-R` are synonims and `-a` implies `-R`.
However, FreeBSD `cp` errors out when both `-r` and `-a` are given.
This fixes `make install` on FreeBSD.

Change-Id: I5b325eb654ca52ad0f988e777c5bf5447d385ab5


timar:
Thank you!

PR #3737, opened 2021-11-29
jsdialogs: make message box fields non editable

eszkadev:
In co2021 type of labels in core were changed for message boxes.
This should be non-editable fields.

messsagebox


eszkadev:
for testing go to Review -> Spell checking  (when reaching end of document, message box appears)

pedropintosilva:
thanks!

PR #3739, opened 2021-11-29
jsdialog: don't send event if closed on core request

eszkadev:
When we received event that dialog was closed - don't
sent close message... It was causing crash in some cases.

example:
1. open empty calc spreadsheet
2. review -> spell checking
   (message box appears - don't close it)
3. click on Ignore once
Before: crash
After: no crash


merttumer (approved):
works as expected

PR #3742, opened 2021-11-29
add option to open xlsb type with android

merttumer:
Also request odf if editing is needed otherwise continue readonly.

Change-Id: I21060b24f6b4733024d8f0ddee504ee9eb554372

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy (approved):
Suggest a follow-up to clean the if (!canDocumentBeExported()) - and instead use it directly when we assign values to isDocEditable:

```
- isDocEditable = true;
+ isDocEditable = canDocumentBeExported();
```

or:

```
- isDocEditable = savedInstance....;
+ isDocEditable = canDocumentBeExported() && savedInstance....;
```

PR #3752, opened 2021-11-30
Add a space also in front of the "git hash" text on the About dialog

tml1024:
Change-Id: I7e2348f2c1cc9bb9d7b3e4814984e9e11695d88c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
I think this is not needed if https://github.com/CollaboraOnline/online/pull/3753 is merged, it seems  I forgot tablets and those rules including adding brackets were just taking effect o desktop (on mobile we do not need them as they go to a new line )

tml1024:
Ah ok, will close this then.

PR #3759, opened 2021-11-30
Make coolwsd-generate-proof-key prepared for a custom prefix.

arrowd:
The downstream packagers can patch the "PREFIX=/" line to their liking.

While there, converts tabs into spaces for consistency.

Change-Id: I424f6fe6efac8e51f74f5848586e7fd10fa40904

### DIFF
--- a/coolwsd-generate-proof-key
+++ b/coolwsd-generate-proof-key
@@ -1,7 +1,19 @@
 #!/usr/bin/env bash
 
+if [ "$1" == "-h" ]; then
+    echo "Usage: coolwsd-generate-proof-key [-h] [CONFIGDIR]"
+    echo "CONFIGDIR defaults to /etc/coolwsd"
+    exit 0
+fi
+
+if [ "$1" ]; then
+    CONFIGDIR="$1"
+else
+    CONFIGDIR="/etc/coolwsd"
+fi
+
 SUDO=''
-if [ ! -w "/etc/coolwsd" ]; then
+if [ ! -w "$CONFIGDIR" ]; then
     if (( $EUID != 0 )); then
         if hash sudo 2>/dev/null; then
             SUDO='sudo'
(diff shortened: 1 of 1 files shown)


arrowd:
Bump.

vmiklos:
@timar or @mikekaganski might be capable of reviewing this.

mikekaganski:
Mostly @timar - I am afraid I am not very familiar with *nix file storage conventions and jails ... sorry :-) - but looks perfectly reasonable from my PoV.

timar (changes requested):
I do not want to nitpick a lot, and this patch is basically indifferent to Collabora's packages, but to generalize the location, would you consider using @LOOLWSD_CONFIGDIR@ and convert coolwsd-generate-proof-key to coolwsd-generate-proof-key.in so this location can be decided build-time?

arrowd:
> I do not want to nitpick a lot, and this patch is basically indifferent to Collabora's packages, but to generalize the location, would you consider using @LOOLWSD_CONFIGDIR@ and convert coolwsd-generate-proof-key to coolwsd-generate-proof-key.in so this location can be decided build-time?

This is a nice suggestion. But I also have one - how about just passing the path to the config as the script parameter?

If there are no args the script would use `/etc/coolwsd` and otherwise use `$1`?

timar:
> > I do not want to nitpick a lot, and this patch is basically indifferent to Collabora's packages, but to generalize the location, would you consider using @LOOLWSD_CONFIGDIR@ and convert coolwsd-generate-proof-key to coolwsd-generate-proof-key.in so this location can be decided build-time?
> 
> This is a nice suggestion. But I also have one - how about just passing the path to the config as the script parameter?
> 
> If there are no args the script would use `/etc/coolwsd` and otherwise use `$1`?

OK!

arrowd:
Bump.

arrowd:
Another bump.

timar (approved):

PR #3765, opened 2021-12-01
fix saveAs for files with encoding characters

merttumer:
Appearently core expects encoded path for filenames on saveas
and decodes it at the end.

So if the file name is something like a%20b
and it becomes a b in the storage.

Change-Id: Ib200c92859b4ab3f262cfe88747563c42e2c6a5c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Looks good, thanks @merttumer. Would be great to have a unit-test. Perhaps we can modify the conflict (and/or save-as) test to use a file that reproduces the issue?

merttumer:
@Ashod thanks, will do as a follow-up

Issue #3773, opened 2021-12-01
Font size selector juming UI


Andreas-Kainz:
### Issue
![Screenshot_20211202_002613](https://user-images.githubusercontent.com/8517736/144330433-d8e5b1cc-12f7-4fe8-898d-648fcb45e843.png)
![Screenshot_20211202_002606](https://user-images.githubusercontent.com/8517736/144330443-db216194-fdab-46b5-8d48-60bc954cb8e1.png)
![Screenshot_20211202_002447](https://user-images.githubusercontent.com/8517736/144330450-15f4d11d-0f46-4fe0-9140-1e37570f957a.png)

### Fix
define a min-width
![Screenshot_20211202_002802](https://user-images.githubusercontent.com/8517736/144330639-4d9a76c0-13a6-484f-ba0a-ecc968f80dee.png)
![Screenshot_20211202_002813](https://user-images.githubusercontent.com/8517736/144330646-b9b85f59-88cf-4054-8e7a-d54ce1a873d0.png)
![Screenshot_20211202_002819](https://user-images.githubusercontent.com/8517736/144330650-901ce7b0-602f-4b13-b05b-099c424cf2e9.png)

Andreas-Kainz:
I think I will use the css class from notebookbar so that there is always a fixed width and the Font Size will get ... elements.

I don't see the benefit of an jumping UI.

PR #3774, opened 2021-12-01
Fix Font size jumping width #3773

Andreas-Kainz:
Change-Id: Ie8b470dc597f4e9e7199550fe57cb98a9e5d42f1

* Resolves: #3773 
* Target version: master 

### Summary
define a min-width solve the issue that the UI will jump if you switch between 10 and 10,5 font size width.

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -452,6 +452,9 @@ button.leaflet-control-search-next
 	line-height: 24px;
 	padding-right: 8px;
 }
+#tb_editbar_item_fontsizes .select2-container{
+	min-width: 55px !important;
+}
 .fontsizes-select {
 	width: 55px !important;
 }


Andreas-Kainz:
Screenshot_20211202_002802
Screenshot_20211202_002813
Screenshot_20211202_002819

If you select an area where you have more than one font size the dropdown switch to Font Size as this string will be translated AND use it as min-width will make the dropdown to width when select a font size number, I defined a min-width so that larger widths are possible.

pedropintosilva:
perfect, could you please also target the id (that is differently named) on notebookbar? Or if you prefer I can merge this and then you do a follow up PR with the `NB` in the name

pedropintosilva:
Also afterwards would be cool to fix line height of those dropdowns the text looks as if it is floating

Andreas-Kainz:
> Also afterwards would be cool to fix line height of those dropdowns the text looks as if it is floating

I know but I don't want to mix different stuff. Otherwise you can't review easy.

Andreas-Kainz:
> perfect, could you please also target the id (that is differently named) on notebookbar? Or if you prefer I can merge this and then you do a follow up PR with the `NB` in the name

I checked NB but there the width was defined that's the reason I didn't make a change for NB. Anyhow if it's there an issue. I will make a separate request.

pedropintosilva:
ok thanks

pedropintosilva:
> > perfect, could you please also target the id (that is differently named) on notebookbar? Or if you prefer I can merge this and then you do a follow up PR with the `NB` in the name
> 
> I checked NB but there the width was defined that's the reason I didn't make a change for NB. Anyhow if it's there an issue. I will make a separate request.

yes there is with but it's not possible to read Font size, it;s cropped

PR #3782, opened 2021-12-02
browser: show busy label if cleaning session

hcvcastro:
Also, the ReconnectCount under normal condition
it should be 0, otherwise the document is cleaning the session.

Change-Id: I07b383a7810b8bc06e207fc600ebec99eaac9c77

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1002,7 +1002,7 @@ app.definitions.Socket = L.Class.extend({
 				}, 500 * this.ReconnectCount * this.ReconnectCount); // Quadratic back-off.
 
 				if (this.ReconnectCount > 1) {
-					this._map.fire('error', {msg: errorMessages.docunloadingretry});
+					this._map.showBusy(errorMessages.docunloadingretry, false);
 				}
 			}
 
@@ -1476,6 +1476,9 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		console.debug('_onSocketClose:');
+		if (this.ReconnectCount > 0)
+			return;
+
 		var isActive = this._map._active;
 		this._map.hideBusy();
 		this._map._active = false;


browser/src/core/Socket.js
```diff
@@ -1476,6 +1476,9 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		console.debug('_onSocketClose:');
+		if (this.ReconnectCount > 0)
+			return;
```
Ashod, line 1480: I worry that we still need some of the other flags below to be set for some reconnection logic to work correct.

For example, if we don't removeAllViews() here, and we do reconnect, what/who will removeAllViews() later? Same question for resetting other objects.

I fear that the logic always assumes that reconnection will have a reset state, but here we are skipping resetting.

How was this tested? Is there a clear scenario that needs this fix? Could this cause other regressions that we haven't thought about yet?

Ashod (commented):
Thanks @hcvcastro, interesting stuff. Left some comments/questions. Let's discuss as necessary.

browser/src/core/Socket.js
hcvcastro: Interesting questions:

				// We're connected: cancel timer and dialog.
				this.ReconnectCount = 0;

do you mean Reset the counter?

did you find a corner case removeAllViews()?
can you provide me logs or data or steps to reproduce as evidence?

steps to test:

1) load the document
2) force an interrupt the service (whatever method you like)
3) the re-connection starts..

Ashod: @hcvcastro, sorry, by "reset" I mean to set the values of the state members to a known value.

For example, `this._map._active = false;`. If we skip this, the code that depends on `map._active` will see it as `true` and will assume we are connected/active.

Thanks for the testing steps. Here are some more specific scenarios that we should make sure work correctly.

What happens when reconnection doesn't work (but the document was loaded)? (This will happen when the server crashes.)
What happens when loading the document doesn't work, and we have to retry? (Do we give up now? Do we give up after retrying a few times?)
What happens when the server is shut down (cleanly) after loading a doc, and then started up, after a minute or two? (This is a supported "recycling" scenario.)

If the above scenarios work correctly, then we should be good. Please confirm they work after trying multiple times, to be sure.

Thanks.

browser/src/core/Socket.js
```diff
@@ -1476,6 +1476,9 @@ app.definitions.Socket = L.Class.extend({
...
+
 		var isActive = this._map._active;
 		this._map.hideBusy();
```
Ashod, line 1483: Btw, why aren't we only skipping the hideBusy for this fix, instead of returning early?

For example:
```
if (this.ReconnectCount <= 0) {
    this._map.hideBusy();
}
```
And leave all the other logic as-is. Wouldn't that be safer? We only want to show the busy label when reconnecting, no? If not, then what are we trying to do?

hcvcastro: Because if you have test your proposal, then you will show an undesired side effects with vex open message dialogs,
not related to reconnecting.

>> We only want to show the busy label when reconnecting, no? If not, then what are we trying to do?
Yes, it only changes the label, no more further changes. and if you analyze:

t1: re-connection starts 
t2: websocket close because no success connection
t3: _onSocketClose executes with all undesired side effects like vex open dialogs messages
t4: re-connection starts again Count+1

If you test without these changes, you will notice that 2 vex dialogs opens, in sequence

hcvcastro: > What happens when reconnection doesn't work (but the document was loaded)? (This will happen when the server crashes.)
_onSocketClose will be called

> What happens when loading the document doesn't work, and we have to retry? (Do we give up now? Do we give up after retrying a few times?)
re-connection start  and until  Max Count retries, after that it will give up with a polite vex message dialog

> What happens when the server is shut down (cleanly) after loading a doc, and then started up, after a minute or two? (This is a supported "recycling" scenario.)

on Client Side, what ever reason if the link websocket is disconnected, it will execute _onSocketClose

hcvcastro: > If the above scenarios work correctly, then we should be good. Please confirm they work after trying multiple times, to be sure.

The scenarios you describe is correct, but I just only changing instead to show a vex dialog to a label message,
there is a test file case, I will link you to test it, I can confirm it works

Ashod (approved):
Discussed offline and agreed that my suggested approaches weren't giving the results expected. @hcvcastro has tested the cases I could think of. Let's get it in.

Issue #3716, opened 2021-11-24
select command fill highlight color which has to big focus


Andreas-Kainz:
#### Now
![Screenshot_20211124_222841](https://user-images.githubusercontent.com/8517736/143316379-2691761f-fd4a-4024-90ef-47ae7ba538cc.png)

#### Past
![Screenshot_20211124_224028](https://user-images.githubusercontent.com/8517736/143316423-e8ca851e-e5ff-4ca8-a174-a33002ae050e.png)

#### LibreOffic
![Screenshot_20211124_224310-1](https://user-images.githubusercontent.com/8517736/143316584-64c10408-dec8-4bd3-8602-76180a8efa7a.png)

In LibreOffice there is gray with a darker gray border for select and blue for hover. In the past cool also use a gray element for selected, but now the filled highlight color was used which is from my point of view the wrong color, cause the command was selected and not highlighted. With make the icon white, it give a good contrast but from my point of view the selected commands has to much focus.

Other office suited also use gray for selected commands.

Ezinnem:
@pedropintosilva

Andreas-Kainz:
What's the reason to have a different select layout for toggled command's like Auto Spell Check?

I'm for a nice and simple gray selection for everything AND an highlight color hover frame. Like it is in picture **LibreOffice**

PR #3785, opened 2021-12-03
Update highlight color layout

Andreas-Kainz:
Change-Id: I0cb4a98b322dcfd9de036a618866ec8fef1a4a98

* Resolves: #3716
* Target version: master 

### Summary
synchronize the select, hover and select:hover layout between classic toolbar, notebookbar and sidebar. Each action has it's own layout. 

### TODO

**select | hover | select**
Screenshot_20211203_012403
**select and hover | unselect | select**
Screenshot_20211203_012410-1

layout is the same at toolbar, notebookbar and sidebar.

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -173,26 +173,39 @@ td.jsdialog .jsdialog.cell.sidebar {
 #table-textorientbox.sidebar .jsdialog .radiobutton {
 	border: 1px solid transparent;
 }
+
+/* non selected */
+.hasnotebookbar .ui-content.unotoolbutton.has-label,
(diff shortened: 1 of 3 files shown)


Andreas-Kainz:
If someone want to change this 3 status elements in the future, you only have to edit the 3 classes in toolbar.css and jssidebar.css the notebookbar classes are integrated into jssidebar.css.

As the toolbar and the notebookbar/sidebar elements work a bit different, it isn't possible to have one class for all implementations.

Andreas-Kainz:
I'm sorry pedro that I didn't wait for your review. As I check the complete UI it's very important to have the overall layout. 

So if you don't like the PR please update it or revert.

PR #3788, opened 2021-12-03
TileCache: improve parsing in parseInvalidateMsg()

vmiklos:
(cherry picked from commit e2515a7b15d47ad75d79048eff9fa4651e323ad0)

Change-Id: I373f85e68964610fe88cb6fa057c548b9a50fb51

### DIFF
--- a/wsd/TileCache.cpp
+++ b/wsd/TileCache.cpp
@@ -342,17 +342,21 @@ std::pair<int, Util::Rectangle> TileCache::parseInvalidateMsg(const std::string&
     }
     else
     {
-        int part;
-        int x;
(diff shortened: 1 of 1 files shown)


vmiklos:
@Ashod could you please review this? Thanks!

This is just to fail better when garbage arrives on the websocket, similar to your 2907d802a5 where you checked for the token count.

Ashod (approved):
Looks good. In master, we can add a `getNonNegTokenInteger` that returns false when the parsed value is negative. I think that's more readable and we can search/replace more easily than adding conditionals. Thanks!

PR #3791, opened 2021-12-03
unit test for save-as files with encoding characters in the name

merttumer:
this test ensures that double encoded filenames are not a problem.
file names can already be encoded like a%20b and this was causing
problems and fixed with #3786

Change-Id: Ibaf0456f26c3a2e8d3c2863c2b925c7a5b9fc79c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitWOPISaveAsWithEncodedFileName.cpp
+++ b/test/UnitWOPISaveAsWithEncodedFileName.cpp
@@ -1,0 +1,94 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <config.h>
+
+#include <WopiTestServer.hpp>
+#include <Log.hpp>
+#include <Unit.hpp>
+#include <UnitHTTP.hpp>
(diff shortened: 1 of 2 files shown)


test/UnitWOPISaveAsWithEncodedFileName.cpp
```diff
@@ -0,0 +1,94 @@
...
+    UnitWOPISaveAsWithEncodedFileName()
+        : WopiTestServer("UnitWOPISaveAsWithEncodedFileName")
+        , _phase(Phase::LoadAndSaveAs)
+    {
+    }
+
+    void assertPutRelativeFileRequest(const Poco::Net::HTTPRequest& request) override
+    {
+        // spec says UTF-7...
+        LOK_ASSERT_EQUAL(std::string("/path/to/hello+ACU-20world.pdf"), request.get("X-WOPI-SuggestedTarget"));
```
Ashod, line 36: Is "hello+ACU-20world.pdf" correct? Is that related to UTF-7 encoding, as the comment implies?

Ashod (approved):
Thanks @merttumer. Looks good. Just a couple of questions.

test/UnitWOPISaveAsWithEncodedFileName.cpp
merttumer: yes it is correct and it is related to UTF-7

merttumer: we send utf-7 encoded to integration %20 becomes +ACU-20 and later it is decoded to %20

PR #3792, opened 2021-12-03
Sidebar alignment fixes

Andreas-Kainz:
Change-Id: I825900d31c4e6ac20d5b8946859c477f4eb026ea

first elements are always left align
image

hide rotate setting cause it didn't work and the sizing isn't good
image

fix dropdown element
image


Andreas-Kainz:
**old master layout**

image

image

image

Issue #3764, opened 2021-12-01
Styles dropdown menu


Andreas-Kainz:
In the Styles drop down menu there is on top a seach (great!) and than there is a from my point of view not realy ordered list of styles.

![2021-12-01 13_29_00-writer odt - Files - Nextcloud](https://user-images.githubusercontent.com/8517736/144234887-324e49f8-418f-4496-85d1-9594606e5cbd.png)

I would prefer to offers in the drop down menu predefined styles (like we have in the LibreOffice menubar styles section) which are manual arranged, than a separator and styles which are used in the document. with search you can searh at ALL available styles.

I know md is simplier, but therefore it's very clean.
![2021-12-01 13_30_01-template md - Nextcloud](https://user-images.githubusercontent.com/8517736/144235203-5c064a6a-784e-402d-ac64-cad81cdc7b4b.png)

Andreas-Kainz:
![image](https://user-images.githubusercontent.com/8517736/144702413-1b18cc01-84f3-406d-ad53-c00b62ca9065.png)

Sidebar Styles dropdown layout

PR #3793, opened 2021-12-04
Fix styles dropdown: make all styles searchable, remove one separator

Andreas-Kainz:
Change-Id: I5f1809858c5aecdef04d9c736711dbb25c80afaa

* Resolves: #3764
* Target version: master 

### Summary
3 sections
1. section predefined styles
2. section favorite styles
3. section ALL available styles

As there is a search (compare to the sidebar) I listed all styles so the search result will find all styles that are available.
separator between predefined styles (default and clear) was removed to have less separations.

Fix is for writer (tested also for the other apps) maybe add favorites to the other apps to.

### DIFF
--- a/browser/src/control/Control.TopToolbar.js
+++ b/browser/src/control/Control.TopToolbar.js
@@ -433,8 +433,6 @@ L.Control.TopToolbar = L.Control.extend({
 				return;
 			var commands = commandValues.Commands;
 			if (commands && commands.length > 0) {
-				// Inserts a separator element
-				data = data.concat({text: '\u2500\u2500\u2500\u2500\u2500\u2500', disabled: true});
 
 				commands.forEach(function (command) {
 					var translated = command.text;
@@ -447,7 +445,7 @@ L.Control.TopToolbar = L.Control.extend({
 			}
 
 			if (this.map.getDocType() === 'text') {
-				styles = commandValues.ParagraphStyles.slice(7, 19);
+				styles = commandValues.ParagraphStyles.slice(7);
 				topStyles = commandValues.ParagraphStyles.slice(0, 7);
 			}
 			else if (this.map.getDocType() === 'spreadsheet') {


pedropintosilva (changes requested):
I think something is not right:
image

Andreas-Kainz:
Is it cool draw isn't it?

Andreas-Kainz:
> I think something is not right: image

The behavior is the same before the PR that Default Paragraph Style is last element.

Screenshot_20211206_220157

Default Style should be Default Paragraph Style from my point of view.

Andreas-Kainz:
The issue from @pedropintosilva isn't related to the PR. So please review the PR and than add an issue if the some behavior from master should be changed.

pedropintosilva:
> Is it cool draw isn't it?

no, it's also in writer.

> The issue from @pedropintosilva isn't related to the PR. So please review the PR and than add an issue if the some behavior from master should be changed.

True, sorry it's just very dificult to review stuff now and later (after it's merge and we want to rework or improved merged stuff) if there is no description in the commit message and if the commit subject is just "[component] update" : p

Could you please amend the commit  and add more info? If I understand it correctly this would do:
* In the subject (1st line): `Fix styles dropdown: make all styles searchable, remove separators`
* Empty line
* In the body:
```
The search in #styles-select didn't include all the styles in its search
results, some styles at the end of the list were being excluded.
List all the available styles by removing the end limit of the content
extraction.

As the predefined styles had unnecessary separators around them since
..explain why. Get rid of those separators  (default and clear).
```

I didn't understand this `3 sections...` as I didn't see that in the code.


-------
Notes for reference:
* The commit message should tell what the commit does, in imperative mood. If necessary, you can mention how the code used to work, too. But what actually is done by the commit should be in present tense. Not what you did. Not "Fixed hang in foo()" but "Fix hang in foo()". Your subject line should be able to fit into this `If applied, this commit will [your subject line here]`
* The first line should be clearly less than 80 characters. It is supposed to be a minimal summary. It can be called the subject of the commit. Do not end it with a period.
* If the change is very simple conceptually, that one line might be enough.
* The rest of the commit message describes the change in greater detail. This is the body of this commit message. The body is written after the subject line with one blank line in between. The blank line is used by various tools (such as 'git log', 'git show', etc.) to differentiate subject from body. Further paragraphs are also separated by blank lines.
* Explain the problem being solved by this commit. More importantly, explain why these changes are being made, as opposed to how. The 'why' part is your responsibility, the 'how' part is code's responsibility.

pedropintosilva (commented):
waiting for amended commit here

pedropintosilva (approved):

PR #3799, opened 2021-12-06
Sidebar: Fix Split Button dropdown hover state

pedropintosilva:
- Fix hover state: before, hovering into the arrow component would
	affect its parent's width
- Make it more visible that the dropdown has two action in it (it is a
	split button dropdown) by:
	- adding background color
	- adding divider, to emphasize that we are talking about two different
		btns
- Use CSS var instead of random gray

Change-Id: I841749d143bcc240a1b4d820868b6a7af53db599

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -726,10 +726,16 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	width: 15px;
 	height: 23px;
 	display: inline-block;
+	border-left: 1px solid transparent;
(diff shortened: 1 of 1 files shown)


pedropintosilva:
https://user-images.githubusercontent.com/65948705/144889416-3d77ccd8-1a56-4314-8bbd-08718018be28.mp4


https://user-images.githubusercontent.com/65948705/144889420-794385b7-1b72-4246-8ad8-8d07b534b12c.mp4

Issue #2393, opened 2021-05-18
Mobile Device UX Layout


Andreas-Kainz:
## General UX Separation
1. There is a top toolbar and a bottom toolbar (+ bottom sidebar). What I like is the separation between top and bottom. On top there is global stuff and at bottom edit related stuff.
2. Top Toolbar can be defined to be visible only when scroll down.

### Top Toolbar
1. Edit button 
2. Document Name and Header name or read only, .... 
3. Comments
4. Search
5. Menubar

### Bottom Toolbar
- In view mode nothing cause bottom is edit related stuff.
- Would be awesome if the scroolbar could have the TOC

### Mockup View Mode
![Mobile_Mode_View](https://user-images.githubusercontent.com/8517736/118647894-28353680-b7e2-11eb-937e-6b8fd248e3ed.png)
Sure it's very similar to what we already have.

### Top Toolbar
1. OK (save) button
2. Instead of the Document Name there are Undo / Redo and Insert
3. Comments
4. Search
5. Menubar

### Bottom Toolbar
Toolbar is context- related

- "property" button depend on what is selected 
- Several uno commands

Click on "property" will extend the bottom toolbar to show the sidebar stuff on bottom, so it's clear, that the almost left command will extend the bottom toolbar with additional commands.

### Mockup Edit Mode
![Mobile_Mode_Edit](https://user-images.githubusercontent.com/8517736/118650489-ce823b80-b7e4-11eb-888c-417fac66da09.png)

Andreas-Kainz:
I can do the context related bottom toolbar if wanted.

PR #3801, opened 2021-12-07
Mobile Context BottomToolbar

Andreas-Kainz:
Change-Id: I5199d45cce08933229bc68e2048f1285fd6594bf

* Resolves: #2393
* Target version: master 

### Summary
BottomToolbar for mobile (writer) with context related commands.
For now with default, Text, Table. Other context commands will follow. Also other apps (calc, impress, draw). But with a separate patch. I'm not a big fan of large patches.


Andreas-Kainz:
Ah I see @timar already add @pedropintosilva to review this patch.

There are two options:
1. review this patch
2. wait until contextual bottombar is finished (for mobile)
whatever @pedropintosilva prefer. Plan to be ready until 09.12. to the next cool meeting.

Andreas-Kainz:
@pedropintosilva you can review tha PR and after that we can talk about the description. I would add screenshots.

Andreas-Kainz:
#### Open issues calc mobile
- [ ] calc mobile insert to a cell so you can type a text or function, the context- isn't know. 

vlcsnap-2021-12-09-09h39m22s666

vlcsnap-2021-12-09-09h39m32s604

It's cool that this separate context- toolbar is available. I would add to the menu commands for functions (=, sum, average, ...) but this commands aren't available for now. So what should I add to this context element. 

In addition I have to know how to activate commands for this menu.

Other option would be to hide the toolbar cause the focus is on type something so have there more hight available, would be a benefit.

pedropintosilva:
@Andreas-Kainz I think there too many things in this one commit (that has no description ). Could you please:
1. Maybe separate this commit in 3 commits?
  -  Add Mobile context bottom toolbar
  - Mobile: Fix [x-component] height [and add in the commit's message body why it needed to be change)
  - Mobile: Replace view comments icon [and add  in the commit's message body why it needed to be change)

pedropintosilva:
image
regarding this it seems that we already had an internal discussion. I have created a separated issue https://github.com/CollaboraOnline/online/issues/3826 for that. Thanks for the heads up!

Andreas-Kainz:
this request will be separated to different requests so it's easier to review and also easier to develop.

PR #3806, opened 2021-12-07
private/ash/tests64

Ashod:
- wsd: test: visual marker for mismatched strings in failed tests
- wsd: test: safer assertions without side-effects
- wsd: test: more informative string assertion in TileQueueTests
- wsd: test: support logging passing test assertions
- wsd: test: killpoco web-sockets in testHandshake
- wsd: test: use dynamic memory for large buffers
- wsd: test: use chrono
- wsd: test: better logging
- wsd: test: improved test state machine
- wsd: test: kill poco in pasteBlank

### DIFF
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <chrono>
 #include <test/testlog.hpp>
 
 #include <atomic>
@@ -68,7 +69,7 @@ public:
 protected:
     // ---------------- Helper API ----------------
     /// After this time we invoke 'timeout' default 30 seconds
-    void setTimeout(int timeoutMilliSeconds);
+    void setTimeout(std::chrono::milliseconds timeoutMilliSeconds);
 
     enum class TestResult
     {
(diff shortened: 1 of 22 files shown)


common/Unit.hpp
```diff
@@ -433,4 +434,70 @@ class UnitKit : public UnitBase
...
+/// the states must have automatic, sequential, values.
+/// But also has some advantages, for example it can be used inside classes.
+/// Some ideas from https://stackoverflow.com/questions/28828957/enum-to-string-in-modern-c11-c14-c17-and-future-c20
+/// and from https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms
+
+#define STRINGIFY(NAME, e) #NAME "::" #e,
+#define CONCAT(X, Y) X##Y
+#define CALL(X, ...) X(__VA_ARGS__)
+
+#define APPLY1(MACRO, NAME, e) MACRO(NAME, e)
```
vmiklos, line 448: Are you sure this is impossible to do with variadic templates? Introducing new macros in C++ code is always a bit of a code smell to me. :-) Perhaps @mikekaganski has an idea.

Other than that, I don't see anything obviously problematic with this, let's get this in as-is, given that it's already in master.

vmiklos (approved):

common/Unit.hpp
Ashod: Agreed in general. The challenge here is that the arguments are enum entries. They aren't variables and they don't exist yet, so we can't reference them. The macros are used to duplicate the entries into the original `enum { ... }` clause and a `const char* [] = { ... }` declaration.

Reflections (in C++23?) might make this easier, but I don't know of any other way. Happy to hear suggestions.

PR #3807, opened 2021-12-07
Improved unload handling

Ashod:
- cypress: close the test document before reloading 
- cypress: give each test a unique test document 
- wsd: remove DocBroker from Admin last 
- wsd: save and wait until document isn't modified 
- wsd: handle the modified status for inactive views 
- wsd: test: add slow save and upload test 
- wsd: do not count disconnecting sessions as active 
- wsd: test: fail when modified docs are unsaved 
- wsd: do not create client sessions on docs marked to destroy 
- wsd: more frequent polling while unloading documents 
- stress: accept self-signed certs in the stress dev/test app 
- wsd: improved logging around session add/remove 
- wsd: stop DocBroker when loading fails 
- wsd: improve logging and doc state dumping


wsd/DocumentBroker.cpp
```diff
@@ -1802,6 +1808,38 @@ bool DocumentBroker::autoSave(const bool force, const bool dontSaveIfUnmodified)
     return sent;
 }
 
+void DocumentBroker::autoSaveAndStop(const std::string& reason)
+{
+    // Don't hammer on saving.
+    if (_saveManager.timeSinceLastSaveRequest() >= std::chrono::seconds(1) &&
```
mmeeks, line 1818: Anytime I see a hard-coded time I'm un-happy =) particularly when it's in seconds.
Does this delay document close / cleanup / re-load / shutdown ? why do we need to wait ? what essential state changes here ?
Still perhaps better than the 5 seconds it replaced - but ... que? =)

wsd/LOOLWSD.cpp
```diff
@@ -4385,11 +4385,16 @@ int LOOLWSD::innerMain()
...
 
+    int returnValue = EX_OK;
+    UnitWSD::get().returnValue(returnValue);
+
+    LOG_INF("Process [loolwsd] finished with exit status: " << returnValue);
+
     // At least on centos7, Poco deadlocks while
     // cleaning up its SSL context singleton.
-    Util::forcedExit(EX_OK);
+    Util::forcedExit(returnValue);
```
mmeeks, line 4395: good stuff =)

mmeeks (commented):
Lots of interesting stuff here; good to see after a re-base I think =) nice to see lots of new tests - thanks Ash ! - then again, I assume the new tests also tend to fail without the fixes (?)

wsd/DocumentBroker.cpp
Ashod: It's a small delay to avoid hammering Core with `.uno:Save` requests, which causes high cpu, floods the queues and doesn't allow for Core to get to idle. It's an arbitrary delay to avoid the above (which was observed in the wild) while being short enough as to not hold up unloading.

No queues here, in this 'unloading' state this function will be invoked on each poll, so it can really overheat quickly. Here we just skip issuing `.uno:Save` for those seconds since the last request/response.

Happy to move these compile-time constants somewhere central and name them, of course, but they get a bit detached if used only once, such as this case.

Ashod:
> Lots of interesting stuff here; good to see after a re-base I think =) nice to see lots of new tests - thanks Ash ! - then again, I assume the new tests also tend to fail without the fixes (?)

Yes! The new tests reproduce cases that fail. Good effort spent to find reproducers, then minimize them to narrow in on the root conditions.

Also, the check that triggers the failures in older tests were only enabled after the fix was applied. So all the patches do build and pass the tests. The ordering is done in "retrospective" fashion. Without the fix, with the new check enabled, at least 4 tests fail (including the new one). Also, the fixes are the minimal set of changes to get all tests passing (I started with a much larger set of fixes for individual failures before reducing them to their bare essentials, for obvious reasons).

Ashod:
+1 from Michael.

Ashod:
@mmeeks, all green!

mmeeks:
Kendy - can you review this ? =) We want it in for Monday's release.

tml1024 (approved):
Looks OK. (Sad to see whitespace changes in the same commit, though.)

tml1024:
Argh, that review was supposed to be just for one commit, will un-review it, if I can figure out how.

tml1024 (commented):
.

Ashod:
> Looks OK. (Sad to see whitespace changes in the same commit, though.)

Ah, hadn't noticed them until now. Wasn't intentional, my editors are all set to remove trailing whitespace and unless I notice and exclude from the commit (which is a pain, because every commit/amend has to exclude them), they end up with functional changes. Thanks for pointing out!

kendy (approved):
I went through the commits but the ec0e861ec73c5460ebeb4999d3faad2d705e7f6e and it looks good to me.  Had some small suggestions here and there, but told them to @Ashod privately, no need to block because of them.

tml1024 (approved):
What can I say? All the commit messages looks sane, and I don't see anything obviously wrong in the code. I would have preferred to have this split up into several smaller PRs, though, as the commits seem fairly independent, even if probably need to be applied in order. But I probably am just not used to the GitHub way of working, where collecting multiple commits into one PR is the norm?

mmeeks:
I looked at some before too - lets get them in =) thanks Ash & Kendy & Tor.

Ashod:
> I would have preferred to have this split up into several smaller PRs, though, as the commits seem fairly independent, even if probably need to be applied in order.

@tml1024, you are right. The reason why they are all in one PR is because most of the commits were fixes to Cypress failures that I couldn't reproduce. In the end the few original commits grew in size and in number to 14. I did offer to split them up, but I think that would have added time/effort to getting them in (yes, they must be in the correct order for the tests to pass), so we left them as-is.

PR #3809, opened 2021-12-08
fix double encoding url on convert-to

merttumer:
Problem is, all save-as requests received urlencoded
but when it is a wopi request we change the url bit of the message
with correct path jail filename parts. At that stage it becomes decoded
already. saveAs in the core also decodes the url and thats why we should
re-encode it. But unless it is a wopi request it should be already
encoded and we should not re-encode it again.

Change-Id: I08b3a1cff6c3aa9271222c255f4a35aa5b984819

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -2339,6 +2339,7 @@ bool ChildSession::saveAs(const char* /*buffer*/, int /*length*/, const StringVe
 
     // if the url is a 'wopi:///something/blah.odt', then save to a temporary
     Poco::URI wopiURL(url);
+    bool encodeURL = false;
     if (wopiURL.getScheme() == "wopi")
     {
         std::vector<std::string> pathSegments;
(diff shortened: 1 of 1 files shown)


kit/ChildSession.cpp
```diff
@@ -2360,6 +2361,9 @@ bool ChildSession::saveAs(const char* /*buffer*/, int /*length*/, const StringVe
         const std::string tmpDir = FileUtil::createRandomDir(jailDoc);
         const Poco::Path filenameParam(pathSegments[pathSegments.size() - 1]);
         url = std::string("file://") + jailDoc + tmpDir + '/' + filenameParam.getFileName();
+        // url becomes decoded at this stage
+        // on saveAs we should send encoded!
+        encodeURL = true;
```
Ashod, line 2366: Wouldn't it be simpler to decode here, and always decode below?

Ashod (commented):
Thanks for the quick fix, @merttumer. See question/suggestion please.

kit/ChildSession.cpp
merttumer: Ah, when we do that it has become double decoded. say the original file name is a%20b and it goes to the this place as a%2520b and we decode here it as you suggest and it becomes a%20b after that  we decode for the second time as explained for the wopi case it has become a b then we encode always as you suggest, it would become a%20b and finally after saveAs it will be a b but the original filename was a%20b. But this patch we only encode the second time if the file is decoded for the wopi case. For the other cases, it is already encoded and we dont do anything.

Ashod (approved):
Looks plausible. Thanks @merttumer!

PR #3810, opened 2021-12-08
common: introduce COOLProtocol::getNonNegTokenInteger()

vmiklos:
To simplify code that wants to extract a non-negative number from a
token value. See
<https://github.com/CollaboraOnline/online/pull/3788#pullrequestreview-826099897>.

Change-Id: I389dfa1a2838e501fc308dbfd7927b142a361922

### DIFF
--- a/common/Protocol.hpp
+++ b/common/Protocol.hpp
@@ -101,6 +101,13 @@ namespace COOLProtocol
         return false;
     }
 
+    /// Extracts a name and value from token. Returns true if value is a non-negative integer.
+    template <std::size_t N>
+    inline bool getNonNegTokenInteger(const std::string& token, const char (&name)[N], int& value)
+    {
+        return getTokenInteger(token, name, value) && value >= 0;
+    }
+
     inline bool getTokenString(const StringVector& tokens,
                                const std::string& name,
                                std::string& value)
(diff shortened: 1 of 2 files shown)


vmiklos:
@Ashod could you please take a look? Thanks.

common/Protocol.hpp
```diff
@@ -101,6 +101,18 @@ namespace COOLProtocol
         return false;
     }
 
+    template <std::size_t N>
+    inline bool getNonNegTokenInteger(const std::string& token, const char (&name)[N], int& value)
```
Ashod, line 105: A documentation comment would be great!

common/Protocol.hpp
```diff
@@ -101,6 +101,18 @@ namespace COOLProtocol
...
+    {
+        bool ret = getTokenInteger(token, name, value);
+        if (!ret)
+        {
+            return ret;
+        }
+
+        return value >= 0;
```
Ashod, line 113: I guess if we want to be more terse we could put it on one line:

`return getTokenInteger(token, name, value) && value >= 0;`

But this is just a personal preference. Your version is perfectly fine, of course.

Ashod (approved):
Thanks @vmiklos! Looks great.

common/Protocol.hpp
vmiklos: Added.

common/Protocol.hpp
vmiklos: Sure, why not.

vmiklos:
@Ashod updated, please take a look. :-)

Ashod (approved):
Great! Thanks @vmiklos.

PR #3812, opened 2021-12-08
add pdf version option for conversion

gmasei11:
Change-Id: I8a9a05498649dfea80ec425671326aa0c62d4917

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3292,6 +3292,21 @@ private:
                     // we want it enabled (i.e. we shouldn't set the option if we don't want it).
                     options = ",FullSheetPreview=trueFULLSHEETPREVEND";
                 }
+                const std::string pdfVer = (form.has("PDFVer") ? form.get("PDFVer") : "");
+                if (!pdfVer.empty())
+                {
+                    if (strcasecmp(pdfVer.c_str(), "PDF/A-1b") && strcasecmp(pdfVer.c_str(), "PDF/A-2b") && strcasecmp(pdfVer.c_str(), "PDF/A-3b")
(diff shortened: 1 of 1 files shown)


kendy (approved):
Looks good to me.  I'm only wondering if the user gives a wrong PDF version - if we should give the 400, or rather default to whatever the core defaults to (ie. not add the PDFVer= at all)?

Either way, up to you, no problem to merge as is, the above can be done in a follow-up if necessary.

Thank you!

gmasei11:
Thanks @kendy  for your feedback!

> Looks good to me. I'm only wondering if the user gives a wrong PDF version - if we should give the 400, or rather default to whatever the core defaults to (ie. not add the PDFVer= at all)?

Let me explain:

In core a wrong parameter is treated as an error. I think this is the best approach because a provided PDFVer parameter means that the caller wanted to enforce a PDF version on conversion and, for whatever reason, provided something wrong. In this case, in my opinion, is not wise to succeed the conversion with a default version because maybe this is not what the caller wants and it can end in misunderstanding (caller: I requested a PDF version and the result has a different PDF version), but more appropriate is to fail and give the possibility to the caller to rectify its request.

In core I just followed the existing rule of handling errors which means that the conversion fails. Unfortunately, the conversion hangs if the core fails the conversion. On the other hand, a check in Online guarantees that in case of wrong parameter the conversion doesn't hang and the appropriate HTTP error is provided.

I agree that it may look odd to check twice in two different places but I think that each component makes the check at its own level and handles in its own way. A call to core may come not necessary from online.

If you disagree with this reasoning just let me know and I can make the appropriate changes.

Thanks!

PR #3813, opened 2021-12-08
MobileWizard: avoid re-scaling ui-drawing-area

pedropintosilva:
Drawing areas such as those found in impress lists: chart type or slide
layout were being downsized and thus blurriness was evident.
-> Remove old CSS rule that does not serve any purpose now and fix that.

Top level drawing areas (where the whole panel is occupied by it:
slide layout) were being set with wrong position (not center
aligned).
-> Always aligned to the center without extra margins that, otherwise
would offset them

Change-Id: If4f4057cffb566026374d4d2e725dc663ff6f3e5

### DIFF
--- a/browser/css/mobilewizard.css
+++ b/browser/css/mobilewizard.css
@@ -1017,9 +1017,8 @@ input[type='checkbox'][disabled] {
 	width: 30% !important;
 }
 
-.mobile-wizard.ui-drawing-area {
+.mobile-wizard .ui-drawing-area-container:not(:only-of-type) .ui-drawing-area {
 	margin: 10px 5% !important;
-	width: 90%;
 }
 
 #TableEditPanel + .ui-content #misc_label.mobile-wizard {


eszkadev (dismissed):
tested with csv import and seems to be ok (not bigger than screen)
but slide layout seems to be not centered:
notcenter

pedropintosilva:
@eszkadev thanks for testing it! It seems I shouldn't really on levels (`.mobile-wizard div:not(.level-0) > .ui-drawing-area`)  Pushed --force, plus I thinks this is a more elegant solution with less convoluted selectors

eszkadev (approved):
now it is centered

pedropintosilva:
Merging this as the failing cypress test reported by CI does not happen locally. I have ran the failing spec `writer/insert_object_spec.js` both headless and interactively and everything worked:



* Insert objects via insertion wizard.
  - [x] Insert local image.passed
  - [x] Insert comment.passed
  - [x] Insert default table.passed
  - [x] Insert custom table.passed
  - [x] Insert header.passed
  - [x] Insert footer.passed
  - [x] Insert footnote.passed
  - [x] Insert endnote.passed
  - [x] Insert page break.passed
  - [x] Insert column break.passed
  - [x] Insert hyperlink.passed
  - [x] Insert shape.

PR #3814, opened 2021-12-08
browser: fix 'rulerupdate:' missing message

hcvcastro:
Otherwise, the 'rulerupdate:' message, it never processed
and the ruler show like an empty box.

Change-Id: Ibe0e76e3615c2c9460aba65dbe76fbad51e9bf42


merttumer (approved):
Works as expected

PR #3816, opened 2021-12-08
browser: initialize layout hidden for ruler control

hcvcastro:
Otherwise, if the server does not send data the ruler
will not show as an empty box.

Change-Id: I990e43fb0e950f6085f9d17d30129327993a72ac


merttumer (approved):
Works as expected

PR #3818, opened 2021-12-08
support for a custom path to NSS certdb files for sign/encrypt

quikee:
The path of the NSS certdb is defined in the coolwsd.xml config
file and if the config file is set and contains the certdb files,
then the db files are copied to the jail into /tmp/certdb folder.
Also the /tmp/certdb path is set to LO_CERTIFICATE_DATABASE_PATH
env. var, which is then used as the default certdb in LibreOffice
when LOKit starts up.

Change-Id: I72e8f28f27a0306fef9319bc6212cd99cb3f8212

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2391,6 +2391,50 @@ void wakeCallback(void* pData)
 
 #ifndef BUILDING_TESTS
 
+namespace
+{
+#if !MOBILEAPP
+void copyCertificateDatabaseToTmp(Poco::Path const& jailPath)
+{
+    std::string aCertificatePathString = config::getString("certificates.database_path", "");
+    if (!aCertificatePathString.empty())
+    {
+        auto aFileStat = FileUtil::Stat(aCertificatePathString);
+
+        if (!aFileStat.exists() || !aFileStat.isDirectory())
+        {
+            LOG_WRN("Certificate database wasn't copied into the jail as path '" << aCertificatePathString << "' doesn't exist");
+            return;
+        }
+
+        Poco::Path aCertificatePath(aCertificatePathString);
+
+        Poco::Path aJailedCertDBPath(jailPath, "/tmp/certdb");
+        Poco::File(aJailedCertDBPath).createDirectories();
+
+        bool bCopied = false;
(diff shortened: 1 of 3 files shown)


kit/Kit.cpp
```diff
@@ -2391,6 +2391,32 @@ void wakeCallback(void* pData)
 
 #ifndef BUILDING_TESTS
 
+namespace
+{
+#if !MOBILEAPP
+void copyCertificateDatabaseToTmp(Poco::Path jailPath)
```
Ashod, line 2397: The suggestion here is correct, this can be `const Poco::Path& jailPath`.

kit/Kit.cpp
```diff
@@ -2391,6 +2391,32 @@ void wakeCallback(void* pData)
...
+{
+    auto aCertificatePathString = config::getString("certificates.database_path", "");
```
Ashod, line 2399: We usually avoid `auto` for trivial types, so this would be `const std::string aCertificatiePathString`.

kit/Kit.cpp
```diff
@@ -2391,6 +2391,32 @@ void wakeCallback(void* pData)
...
+    if (!aCertificatePathString.empty())
+    {
+        Poco::Path aCertificatePath(aCertificatePathString);
```
Ashod, line 2402: Do we want to check if this path exists and is valid, and log a warning otherwise? I think that will help users troubleshoot typos in the path and things like that easily.

Ashod (approved):
Thanks @quikee! Looks good. A couple of cosmetic suggestions.

kit/Kit.cpp
```diff
@@ -2391,6 +2391,50 @@ void wakeCallback(void* pData)
...
 
+namespace
+{
+#if !MOBILEAPP
+void copyCertificateDatabaseToTmp(Poco::Path const& jailPath)
+{
+    std::string aCertificatePathString = config::getString("certificates.database_path", "");
+    if (!aCertificatePathString.empty())
+    {
+        auto aFileStat = FileUtil::Stat(aCertificatePathString);
```
Ashod, line 2402: Nice!

Ashod (approved):
Looks great. Thanks @quikee!

PR #3825, opened 2021-12-09
jsdialog: handle initial focus in dialog

eszkadev:
it requires in the core: https://gerrit.libreoffice.org/c/core/+/126565


merttumer (approved):
initial focus works as expected

PR #3827, opened 2021-12-09
Private/timar/shortcuts

timar:
* Resolves: shortcuts in menus were not localized and were wrong on Mac
* Target version: master


timar:
Failure on FreeBSD 13 is completely unrelated to this PR. We can ignore it.

thebearon (approved):
I can confirm the shortcuts are localized with eg. German language, and also shown with appropriate symbols on a Mac.

alessioalex-1and1 (commented):
Looks good to me, thanks for fixing more specific cases! I wasn't aware of those!

PR #3828, opened 2021-12-09
Backports and minor fixes

Ashod:
- wsd: Termination flag requires having ShutdownRequested
- wsd: fix 'potentially uninitialized' variables
- wsd: cache misses are not errors
- wsd: add endsWith helper with tests
- wsd: correct use the exit code from UnitWSD


kendy (approved):
Looks good, thank you!

PR #3830, opened 2021-12-09
Actually pass freebsd_flags variable to nmount() call.

arrowd:


timar:
Unit & Cypress & Notebookbar Tests — FAILURE
Surely it's false alert, because these tests run on Linux, and change was in FreeBSD-only code.

PR #3832, opened 2021-12-09
Mobile: Comments: Fix empty state's height

pedropintosilva:
Before we were not passing any additional class to mobile-wizard-content
as a result and, even thought child elements were set to 100% height,
the empty state was still cropped. Pass additional class to parent and
make sure child is set to full height (minus header bar).

Change-Id: I42013bfc9976edcf13680fe428484a086dcf1074

### DIFF
--- a/browser/css/mobilewizard.css
+++ b/browser/css/mobilewizard.css
@@ -1110,6 +1110,15 @@ input[type='checkbox'][disabled] {
 	grid-row: 1;
 }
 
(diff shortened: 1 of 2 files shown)


pedropintosilva:
the failing cypress test seems to have nothing to do with this PR

Andreas-Kainz (approved):

pedropintosilva:
> the failing cypress test seems to have nothing to do with this PR

could you take a look @Rash419 ?

PR #3838, opened 2021-12-09
Calc Sidebar: Number Format Section Decimal places alignment

Andreas-Kainz:
Check out Calc default Sidebar there is the Number Format section.
In the Decimal places group there are two div elements that should
be not displayed. With this PR the elements get property
display: none

In addition one line has a height of 44px in the sidebar
As the height is 52px the additional padding was removed

The result is that all rows in the sidebar are well aligned
and use a height of 44px.

Fix helps also NextCloud office.

Change-Id: I813d3a270096a0cc3d8c0740b6ac346bd8e2564a

**result**
image

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -329,3 +329,12 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 .sidebar .spinfieldcontainer input {
 	width: calc(100% - 28px);
 }
+
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2),
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) {
+	display: none;
+}
+
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) {
+	padding: 0 4px 0 0;
+}


browser/css/jssidebar.css
```diff
@@ -329,3 +329,12 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 .sidebar .spinfieldcontainer input {
 	width: calc(100% - 28px);
 }
+
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2),
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > div:nth-child(1) {
```
pedropintosilva, line 334: This works but to have this type of nth-child chain is not that good, in terms of how it looks/maintainability and more important because probably in short time this change will stop to work due to any small change on the sidebar structure be that from online or core

browser/css/jssidebar.css
```diff
@@ -329,3 +329,12 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
...
+	display: none;
+}
+
+#NumberFormatPropertyPanel > tr:nth-child(1) > td:nth-child(1) > div:nth-child(1) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) {
```
pedropintosilva, line 338: We could, use [:empty](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) instead..hm but in this case those element do have something (even if they appear empty) inside

pedropintosilva (changes requested):
Maybe better would be to:
Find out why these elements' childs are being generated on desktop but then getting `.hidden ` and see if we can fix the root cause Example:
```html
<div class="row jsdialog sidebar">
    <div class="cell jsdialog sidebar">
        <p class="jsdialog sidebar ui-text hidden" id="denominatorplaceslabel">Denominator places:</p>
    </div>
</div>
```
Maybe @szymon can shed some light here

eszkadev:
@Andreas-Kainz we cannot hide those fields, if you set "fraction" in the number type you will see they are used

Andreas-Kainz:
Thanks @eszkadev than I have to find an other solution.

Andreas-Kainz:
fix the issue where it was in LibreOffice see https://gerrit.libreoffice.org/c/core/+/126688

than it need some time to have it in cool, but it's the real place where the issue should be solved. 

maybe I will fix temporary cool with css.

Issue #3829, opened 2021-12-09
COOL Header Section rearrange


Andreas-Kainz:
Nextcloud office has an option to show/hide header elements depend on the window size.

## NextCloud Office
Show filename and save date cause the width of the window is big enough
![2021-12-09 12_31_51-Nextcloud](https://user-images.githubusercontent.com/8517736/145388975-68297e09-a179-4687-80ea-7c21e8937405.png)

Hide filename cause the width of the window isn't big enough.
![2021-12-09 12_32_12-Nextcloud](https://user-images.githubusercontent.com/8517736/145388989-2a83d6f0-68f4-4c93-8d83-cfecb3e6a7b5.png)

## Suggestion
- show everything if the window is width enough
- hide filename if width isn't big enough
- hide savedate if width isn't big enough

I suggest to have the behaviour for classic and notebookbar mode.

Andreas-Kainz:
I can do the work, but first @pedropintosilva I want to have your go to backport the stuff from NextCloud office. 

Best would be, if NC would backport the stuff.

As I wrote I would have the same layout for classic and notebookbar layout (notebookbar layout didn't offer now saved time).

PR #3713, opened 2021-11-24
Classic: shortened size of menus

pedropintosilva:
To occupy less space critical in some languages where the last save
string is quite long also increase its height so the click target
is not too small

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>
Change-Id: I16affef5b5ecb484a14c7f6ff03ba8a28b3eef63

PR #3839, opened 2021-12-09
Cool Header with filename and save date depend on window width

Andreas-Kainz:
**background color**
Remove background color for each single header element, so it's
possible to switch the global header color easy

**align header text**
The header elements are vertical center align

**resize window**
the filename was shown depend on the window with
the save date was shown depend on the window with
first the filename was hidden.

Change-Id: I9e275ab7105ace8147f3d07ca73a59a3818dbc29

* Resolves: #3829
* Target version: master

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -16,7 +16,7 @@
 
 #document-header {
 	position: relative;
-	background: #ffffff;
+	background: transparent;
 	margin-right: 5px;
 }
 /*avoid multiple document-headers in readonly mode*/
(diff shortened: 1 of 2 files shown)


pedropintosilva (changes requested):
Thanks @Andreas-Kainz I will check it locally, but at the first sight do you think we could convert all or any of those rules that are inside of media queries into normal rules inside of `browser/css/device-tablet.css` or `browser/css/device-mobile.css`

device-xxx.css is conditionally loaded depending on each case (as you can see in https://github.com/CollaboraOnline/online/blob/master/browser/html/cool.html.m4#L346 )

pedropintosilva:
and it seems something affected the tabs, see the cropped one:
image

Andreas-Kainz:
I want to make first classic toolbar and after that tabbed one. But however I can add tabbed to.

I understood that you want to have clear decisions communicated at the git history. I myself I'm fit simple and small patches for example.

Andreas-Kainz:
> and it seems something affected the tabs, see the cropped one: image

fixed

Andreas-Kainz:
Screenshot_20211219_233447
now resizing work at standard toolbar AND notebookbar.

Andreas-Kainz:
@pedropintosilva review and feedback would be awesome.

browser/css/menubar.css
```diff
@@ -16,7 +16,7 @@
 
 #document-header {
 	position: relative;
-	background: #ffffff;
+	background: transparent;
```
pedropintosilva, line 19: I have created a PR with a commit just for the colours, maybe we can divide further this PR into several commits, I will help you with that.

browser/css/menubar.css
```diff
@@ -63,14 +63,60 @@
 	display: none;
 }
 .main-nav {
-	height: 32px; /* on mouseover menubar items, border emerges */
+	height: 36px; /* on mouseover menubar items, border emerges */
```
pedropintosilva, line 66: I think this is not needed 5 days ago on of my PRs from Nov. got merged that modifies this: (more info: https://github.com/CollaboraOnline/online/pull/3713)

browser/css/menubar.css
```diff
@@ -63,14 +63,60 @@
...
 	width: auto;
 	background: #ffffff;
-	padding: 3px 3px 3px 0;
+	padding: 0;
 	white-space: nowrap;
 	z-index: 12;
 	display: flex;
+	align-items: center;
```
pedropintosilva, line 73: Vertical align nav bar and its child: maybe better to have a commit only fr this

browser/css/menubar.css
```diff
@@ -63,14 +63,60 @@
...
+}
+
+nav:not(.hasnotebookbar) #main-menu-state:not(:checked) ~ #main-menu {
+	display: flex !important;
+}
+
+nav:not(.hasnotebookbar) #main-menu {
+	width: 100%;
+	height: auto;
+}
```
pedropintosilva, line 83: Part of Vertical align nav bar and its child: no need to specifically target classic menu since #main-menu is only visible when using classic mode

browser/css/menubar.css
```diff
@@ -63,14 +63,60 @@
...
+		transform: translate(-50%, 0);
+		z-index: 1200;
+	}
+}
+@media screen and (max-width: 1199px) {
+	nav:not(.hasnotebookbar).main-nav #document-titlebar {
+		display: none;
+	}
+}
+
```
pedropintosilva, line 119: These media queries, some maybe can be in some of the device specific CSS files we have others probably not specially if we want it to react to window resize but then we should test these first in a real tablet (not the responsive mode that can sometimes be tricky) to see if it affects it or not

pedropintosilva (changes requested):

browser/css/menubar.css
pedropintosilva: https://github.com/CollaboraOnline/online/pull/3906

browser/css/menubar.css
pedropintosilva: _cherry-picking and creating a new commit just with your Vertical align nav bar changes_

pedropintosilva: Tested and the lack of padding made it look very cramped up (without enough white space around), pick everything else though: https://github.com/CollaboraOnline/online/pull/3952

Andreas-Kainz:
@pedropintosilva can you please fix the conflict.

Andreas-Kainz:
the change was a backport of what nextcloud office did. So I don't know what each single change did, cause the change come from nextcloud office.

PR #3842, opened 2021-12-10
jsdialog: handle default button

eszkadev:
- default button is the first button in the form
- change main container of a dialog into <form>
- create hidden default button as a first element in the form


merttumer (approved):
works great, thanks

PR #3844, opened 2021-12-10
theme: added config flag to switch between integration and cool themes

lpranam:
Change-Id: I679328e29f2e2297491c73249632627d01a1bf42

* Target version: master 

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -301,7 +301,8 @@ m4_ifelse(MOBILEAPP,[true],
       window.frameAncestors = '';
       window.socketProxy = false;
       window.tileSize = 256;
-      window.uiDefaults = {};],
+      window.uiDefaults = {};
+      window.useIntegrationTheme = 'false';],
      [window.host = '%HOST%';
       window.serviceRoot = '%SERVICE_ROOT%';
       window.hexifyUrl = %HEXIFY_URL%;
@@ -314,6 +315,7 @@ m4_ifelse(MOBILEAPP,[true],
       window.enableWelcomeMessage = %ENABLE_WELCOME_MSG%;
       window.enableWelcomeMessageButton = %ENABLE_WELCOME_MSG_BTN%;
       window.userInterfaceMode = '%USER_INTERFACE_MODE%';
+      window.useIntegrationTheme = '%USE_INTEGRATION_THEME%';
       window.enableMacrosExecution = '%ENABLE_MACROS_EXECUTION%';
       window.outOfFocusTimeoutSecs = %OUT_OF_FOCUS_TIMEOUT_SECS%;
       window.idleTimeoutSecs = %IDLE_TIMEOUT_SECS%;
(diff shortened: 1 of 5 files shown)


wsd/FileServer.cpp
```diff
@@ -958,7 +958,9 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
         COOLWSD::getConfigValue<bool>("hexify_embedded_urls", false) ? "true" : "false";
     Poco::replaceInPlace(preprocess, std::string("%HEXIFY_URL%"), hexifyEmbeddedUrls);
 
-    std::string themePreFix = (theme == "nextcloud") ? theme + "/" : "";
+
+    bool useIntegrationTheme = config.getBool("user_interface.use_integration_theme", true);
+    std::string themePreFix = (theme == "nextcloud") && useIntegrationTheme ? theme + "/" : "";
```
Ashod, line 963: Could be `const`.

wsd/FileServer.cpp
```diff
@@ -1017,6 +1019,11 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
 
     Poco::replaceInPlace(preprocess, std::string("%USER_INTERFACE_MODE%"), userInterfaceMode);
 
+    std::string useIntegrationThemeString = "true";
+    if (!useIntegrationTheme)
+        useIntegrationThemeString = "false";
```
Ashod, line 1024: Why not use tertiary operator here too?

`const std::string useIntegrationThemeString = useIntegrationTheme ? "true" : "false";`

Actually, it could even be `constexpr auto`, but simple works too.

Ashod (approved):
Looks very good. Thanks @lpranam!

Just a couple of nit-picks, but let's get it in, no blockers.

browser/html/cool.html.m4
```diff
@@ -311,6 +311,7 @@ m4_ifelse(MOBILEAPP,[true],
       window.enableWelcomeMessage = %ENABLE_WELCOME_MSG%;
       window.enableWelcomeMessageButton = %ENABLE_WELCOME_MSG_BTN%;
       window.userInterfaceMode = '%USER_INTERFACE_MODE%';
+      window.useIntegrationTheme = '%USE_INTEGRATION_THEME%';
```
tml1024, line 314: Here you should probably set window.useIntegrationTheme also in the true branch of the `m4_ifelse(MOBILEAPP,[true],` , a dozen lines earlier.

tml1024 (changes requested):

browser/html/cool.html.m4
```diff
@@ -298,7 +298,8 @@ m4_ifelse(MOBILEAPP,[true],
       window.frameAncestors = '';
       window.socketProxy = false;
       window.tileSize = 256;
-      window.uiDefaults = {};],
+      window.uiDefaults = {};
+      window.useIntegrationTheme = '%USE_INTEGRATION_THEME%';],
```
tml1024, line 302: Yes, but we can't use the %FOO% things for the mobile apps, as in their case the cool.html file does not pass through the FileServer.cpp code, but is used as such. Just hardcode whatever is the suitable value of useIntegrationTheme for the mobile apps.

tml1024 (changes requested):

tml1024 (approved):

PR #3847, opened 2021-12-11
Private/mmeeks/tracing2

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -448,7 +448,7 @@ app.definitions.Socket = L.Class.extend({
 			if (e.image.completeTraceEvent)
 				e.image.completeTraceEvent.abort();
 		};
-		e.image.completeTraceEvent = this.createCompleteTraceEvent('cool._extractTextImg');
+		e.image.completeTraceEvent = this.createAsyncTraceEvent('loadTile');
 		e.image.src = img;
 	},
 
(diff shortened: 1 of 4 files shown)


browser/src/core/Socket.js
```diff
@@ -289,7 +289,28 @@ app.definitions.Socket = L.Class.extend({
 									       {'_slurpQueue.length' : String(queueLength)});
 		if (this._map && this._map._docLayer) {
 			this._map._docLayer.pauseDrawing();
+
+			// Queue an instant timeout early to try to measure the
+			// re-rendering delay before we get back to the main-loop.
+			if (this.traceEventRecordingToggle)
+			{
```
tml1024, line 296: Why don't you use the existing createCompleteTraceEvent function (in browser/src/core/Socket.js) but duplicate code from it?

browser/src/core/Socket.js
```diff
@@ -1714,7 +1714,9 @@ app.definitions.Socket = L.Class.extend({
 		result.finish = function () {
 			if (this.active) {
 				var now = performance.now();
-				that.sendMessage('TRACEEVENT name=' + name + ' ph=X ts=' + Math.round(now * 1000) + ' dur=' + Math.round((now - this.begin) * 1000)
+				that.sendMessage('TRACEEVENT name=' + name +
+						 ' ph=X ts=' + Math.round(this.begin * 1000) +
+						 ' dur=' + Math.round((now - this.begin) * 1000)
```
tml1024, line 1751: Good catch.

tml1024 (approved):

browser/src/core/Socket.js
mmeeks: Re: duplicating it - fair cop we could re-factor that but I want to add the timeout early to get in as fast as possible after we return from idle - but track the start time from the end of the method - could be done a nicer way I guess by updating 'begin' or something in an un-finished object - feel free to clean it up.
Thanks !

Issue #3770, opened 2021-12-01
Sidebar dropdown and insert field design


Andreas-Kainz:
**Describe the bug**
In the sidebar we have ordinary left align a label and right align a dropdown menu or an insert field element.

**To Reproduce**
1. open sidebar
2. select an object

**Expected behavior**
if the dropdown menu and the insert field element have the same width the alignment of the sidebar elements would be perfect. 
1. This is easy doable with the .css file, but as the insert field element has a (language specific) unit, it would look perfect, only when the unit will be shown inside the insert field.
2. In additon have an +/- elment left and right to the insert field would make the up/down buttons as they are now easier to select by mouse AND by touch.

**Screenshots before**
![2021-12-01 15_47_01-](https://user-images.githubusercontent.com/8517736/144255870-3de3bd05-96a6-42d8-a932-e12ce4004b49.png)

**Screenshots after**
![2021-12-01 15_47_01-](https://user-images.githubusercontent.com/8517736/144258117-b573e4da-2c53-4857-a14a-160d0a5fdace.png)

Andreas-Kainz:
maybe it's not good to mix two ideas.

This issue is about to have the measurement unit inside the insert feeld.

![2021-12-06 09_12_46-Präsentation1 - PowerPoint](https://user-images.githubusercontent.com/8517736/144810710-94b03a11-eb83-468a-95ec-0e4e445d6d76.png)

have the measurement inside the insert field would make the UI more align.

pedropintosilva:
Yeah maybe for that other issue just visit : https://github.com/CollaboraOnline/online/issues/3324

pedropintosilva:
> maybe it's not good to mix two ideas.
> 
> This issue is about to have the measurement unit inside the insert feeld.
> 
> ![2021-12-06 09_12_46-Präsentation1 - PowerPoint](https://user-images.githubusercontent.com/8517736/144810710-94b03a11-eb83-468a-95ec-0e4e445d6d76.png)
> 
> have the measurement inside the insert field would make the UI more align.

I have no objection to it

Issue #3324, opened 2021-09-30
iOS App: spinbox instead of simple input field (HTML/JS sidebar)


nikslor:
**Describe the bug**
I'm not sure if it's a bug, but I think so. In the writer sidebar, there are four widgets/fields to improve the line padding (above and below the line) as well as the indentation/spacing (left and right). In the iOS app those fields are simple input fields (at least they look like this) - in other platforms those fields are spin boxes (with +/- buttons). Is this a bug in Safari or is it a big in JSdialogs or is it the intended behavior?

I add tml/szymon to the loop: maybe one of you can tell me if this is a bug? Thanks!

**To Reproduce**
1. Create/open a writer document in the iOS app
2. Write some text and open the sidebar
3. Tap into the field to increase the line padding 

**Expected behavior**
The field should be easy to use with a finger and offer +/- buttons

**Actual behavior**
There are not +/- buttons

**Screenshots**
iOS app (no +/- buttons):
![image](https://user-images.githubusercontent.com/1211300/135463639-8e7dc550-2f25-4883-98c7-0f9fae0d75dd.png)

Online (with +/- buttons):
![image](https://user-images.githubusercontent.com/1211300/135463759-e7bc9a4f-adf4-4478-b874-7fff9e5289e2.png)

**Tablet**
 - iPadOS: 14.8
 - Version: 21.6.1 (2)

nikslor:
Best would be to have something that looks similar to this:

![image](https://user-images.githubusercontent.com/1211300/135464677-5ef33a11-1181-44b1-989e-0e3cbcfcb4b3.png)

This works well with finger.

nikslor:
And this is how it looks in the GTK version:

![image](https://user-images.githubusercontent.com/1211300/135465519-704cedd6-a97a-40fa-bb6d-2d03cf71ce3e.png)

eszkadev:
it's <input type="number">  handled by the browser...
In firefox it has +/- buttons always.
In chromium it shows arrows only when focused.
so it depends on the browser.

But for webkit-based browsers it can be hacked to show arrows always: https://stackoverflow.com/questions/25194631/is-it-possible-to-always-show-up-down-arrows-for-input-number
maybe it works also for safari?

I don't have safari to test this. @tml or @pedropintosilva can you try?

tml1024:
[...]

PR #3848, opened 2021-12-12
Sidebar spinfield input field update #3770

Andreas-Kainz:
**spinfield unit**
- spinfield unit element was moved inside the spinfield
- there is space enough for 2 letters

**spinfield sync with listbox**
all items are similar to listbox elements
- add 4px border radius
- add border color
- add hover effect

**spinfield and listbox alignment**
- spinfield and listboxes should be right align
- therefore a fixed width for spinfield was added
- fixed spinfield width was designed for paragraph section
- some spinfields need movement

Change-Id: Ic90b450a423f7c9b728e39c854a02e4b02dfe7b7

* Resolves: #3770 
* Target version: master 

**screenshots**
left after, right before

Screenshot_20211212_174251
- spinfield unit is within the spinfield

Screenshot_20211212_180538
- right alignment


Andreas-Kainz:
And the spinbox get an defined height so all sidebar rows have the same height.

eszkadev:
Did you try it with different browsers? I'm wondering if all of them show numbers aligned to the left

Andreas-Kainz:
> Did you try it with different browsers? I'm wondering if all of them show numbers aligned to the left

With Firefox and Chrome. The numbers weren't shown aligned to the left. The spinfield unit was only moved 40px to the right.

In general I would prefer something like a predefined class for spinboxes at least it would also help with #3324

Andreas-Kainz:
**left Firefox, right Chrome**
2021-12-13 08_46_10-example odt - Files - Nextcloud

Andreas-Kainz:
any open questions?

pedropintosilva (approved):
The changes work to achieve what is described in the commit message

PR #3852, opened 2021-12-13
Document has been changed: set cancel btn on mobile to block

pedropintosilva:
Even though the on mobile the display set to flex seems to work
introduced with 95bd65d2f658ca8c378b50f4ff09ef1424c15c13
the fact is that we do not need flex here and it might cause different
unexpected results depending on the device

This commit is not needed on 6.4 (as it was already fixed while cherry
picking Document has been changed related commits)

Change-Id: Ib50d6fa9e1e456c3dfdceaae802c2fe1edad15a7


gokaysatir (approved):
Thanks for the fix:)

PR #3854, opened 2021-12-13
check if we save-as successfully before proceed with downloadas

merttumer:
Change-Id: I77c76f424a5767a19be812f35dfadf9e56cbfb20

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kit/ChildSession.cpp
```diff
@@ -1113,10 +1113,16 @@ bool ChildSession::downloadAs(const char* /*buffer*/, int /*length*/, const Stri
...
 
-    getLOKitDocument()->saveAs(url.c_str(),
+    bool success = getLOKitDocument()->saveAs(url.c_str(),
                                format.empty() ? nullptr : format.c_str(),
                                filterOptions.empty() ? nullptr : filterOptions.c_str());
 
+    if (!success)
+    {
+        LOG_ERR("SaveAs Failed for id=" << id << " [" << url << "].");
+        return false;
```
eszkadev, line 1123: I think we should send some error message here, client shows 'busy popup' on save as, we should hide that and show error for a user

eszkadev: maybe this getError in the error handling code will provide more info: https://github.com/CollaboraOnline/online/blob/f88a38a4e7d25f1d34350b313df42f3ae0acbe76/kit/ChildSession.hpp#L345

eszkadev: `getLOKitDocument()->getError()` should work I think?

merttumer: thanks

eszkadev (approved):
Now, looks great, thanks :)

Issue #3758, opened 2021-11-30
Editing cursors are out of place when using browser zoom


juliushaertl:
**Describe the bug**
When using the browser zoom other users editing cursors are out of place if using a different zoom level than 100%

**To Reproduce**
1. Have multiple users editing
2. Change the browser zoom level (e.g. clicking the toolbar and using ctrl+scrollwheel)
3. See that the other users cursors are out of place

Happens on both master, 21.11 as well as 6.4

As seen today in the call with @pedropintosilva

Ezinnem:
Confirmed on COOLWSD 21.11.0.3 git hash: 265c275a

Ezinnem:
Confirmed on COOLWSD 21.11.0.3 git hash: 265c275a

eszkadev:
should be fixed now

PR #3856, opened 2021-12-13
Hide cursor during zoom animation #3758

eszkadev:
Avoid showing colored user names during zoom animation,
because we didn't update their position according to the
animation and they were at the initial positiona during
whole animation.


dennisfrancis (approved):
Looks good to me. Thanks!

PR #3862, opened 2021-12-14
Allow time for images to load before processing slurped msg. queue.

mmeeks:
Finger in the air - wait for up to four at a time before emitting
events.

Also do text conversion once, and re-use the result. Share
prettification of events for tracing messages.

Change-Id: Ib7780d055ee332cd1b2718a0a6fb4b194fe2c450

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -272,18 +272,54 @@ app.definitions.Socket = L.Class.extend({
 			     'background:#ddf;color:black', color, 'color:black');
 	},
 
-	_queueSlurpEventEmission: function() {
+	_countSlurpedImages: function() {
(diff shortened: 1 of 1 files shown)


mmeeks:
Hmm - unclear that this works as well as I'd hope. The concept is not terribly silly but whether it makes a real difference is hard to measure - particularly when scrolling down a spreadsheet with a different background color which is a good test case.

llunak:
I'm not familiar with the JS stuff, so I don't feel particularly qualified to review this.

mmeeks:
Abandon this - deltas allow us to immediately de-compress and load the images =)

PR #3870, opened 2021-12-16
cypress:added retries parameter in some fragile tests

Rash419:
Change-Id: Ia99f75302a988fc2ef744807e01ba6441ff161ec

* Target version: master

### DIFF
--- a/cypress_test/integration_tests/multiuser/calc/repair_document_spec.js
+++ b/cypress_test/integration_tests/multiuser/calc/repair_document_spec.js
@@ -3,7 +3,7 @@
 var helper = require('../../common/helper');
 var calcHelper = require('../../common/calc_helper');
 
-describe('Repair Document', function() {
+describe('Repair Document',function() {
 	var testFileName = 'repair_doc.ods';
 
 	beforeEach(function() {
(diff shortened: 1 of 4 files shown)


cypress_test/integration_tests/multiuser/calc/repair_document_spec.js
```diff
@@ -3,7 +3,7 @@
 var helper = require('../../common/helper');
 var calcHelper = require('../../common/calc_helper');
 
-describe('Repair Document', function() {
+describe('Repair Document',function() {
```
Ashod, line 6: The space was actually better, and consistent with the other cases.

Ashod (commented):
Can we do this for all tests? I don't see why we can't retry any failed test at least once, instead of re-running the whole build, which we often do to make sure it wasn't a random failure. There might be an easy way to configure the default retry count of all tests.

cypress_test/integration_tests/multiuser/calc/repair_document_spec.js
eszkadev: +1, better to not change unrelated lines

eszkadev:
Can we do this also for other "repair_document_spec.js" (writer and impress) ? I noticed it failed for me also in impress case.

eszkadev:
@Ashod maybe let's check if it helps first before we apply it to all tests (these failures were mostly present on gerrit's Jenkins)

eszkadev (approved):

Issue #3805, opened 2021-12-07
Sidebar: Paragraph align vertical


Andreas-Kainz:
the arrangement of the writer sidebar section with activated vertical align elements didn't fit perfect

![image](https://user-images.githubusercontent.com/8517736/145110325-a1f0a852-6078-4dbb-aa14-e6754d1674cb.png)

Ezinnem:
The screenshot looks a lit different from what I have on my end.

Please what is your screen resolution?

Ezinnem:
@pedropintosilva @eszkadev

Andreas-Kainz:
I will fix this bug, if it's really a bug myself. It's here as issue to have it on my to-do list.

pedropintosilva:
So it took me a while to understand where this bug came from and how #3873 is supposed to solve it. I discover that it's basically a regression coming from the following commit 27be41fb787762111cd1212524e3f3c48b4d8cb1 (note please in the future add the hash of the commit from where the regression started, and maybe better to revert the commit first with `git revert [commit-hash-number]` and then add additional commits if necessary above it) .  

Please see notes in the closed PR

PR #3873, opened 2021-12-17
Sidebar/Paragraph Table selection vertical align commands new row

Andreas-Kainz:
BUG #3805

Vertical Alignment commands move to the next table-row
cause there is not enough space for 9 commands.

Change-Id: I9455ec2d5a5990be2a17c9e7eac1320858ed5824

* Resolves: #3805
* Target version: master 

**Screenshot Result**
image

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -275,9 +275,8 @@ td.jsdialog .jsdialog.cell.sidebar {
 	left: -7px;
 }
 
-#table-verticalalignment {
-	position: relative;
-	top: 6px;
+#table-box1 > #box1 > div:nth-child(3) {
+	display: table-row;
 }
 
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
**Screenshot Before**

image

pedropintosilva:
So I think it's better to close this PR and merge #3876 first and if you still want to adjust the specific element only for draw and /or impress you can try something like this in your future PR:

```css
#document-container.drawing-doctype + #sidebar-dock-wrapper #table-verticalalignment {
	position: relative;
	top: 6px;
}
```

this way we:
- keep the git history more clean (better to find and hack in the future be that via git log, grep or git blame)
- avoid things like `#table-box1 > #box1 > div:nth-child(3)` that can bite us in the back side later on when the structure changes etc

PR #3874, opened 2021-12-17
cypress:disable repair_document_spec.js temporarily

Rash419:
Change-Id: I02b42b8943ecfedb5a950d062a0fef2a7e3f3e6b


eszkadev (approved):

cypress_test/integration_tests/desktop/impress/annotation_spec.js
```diff
@@ -15,7 +15,7 @@ describe('Annotation Tests', function() {
 		if (Cypress.env('INTEGRATION') === 'nextcloud') {
 			desktopHelper.hideSidebarIfVisible();
 		} else if (Cypress.env('USER_INTERFACE') === 'notebookbar') {
-			cy.get('#ModifyPage').click();
+			cy.get('#ModifyPage1').click();
```
eszkadev, line 18: please use `.unoModifyPage` instead, we changed uno buttons identifiers to be unique, but it is not guaranted that it will always have '1' here :)

so instead of #XXX always use .unoXXX for unobuttons

eszkadev (approved):

PR #3875, opened 2021-12-17
wsd:fix: Document conflict window not shown

Rash419:
Change-Id: I699aaed92d41f76a8eb9de6bdb65608099802663

* Target version: master

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1619,7 +1619,7 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
         LOG_ERR("PutFile says that Document changed in storage");
         _documentChangedInStorage = true;
         const std::string message
-            = isModified() ? "error: cmd=storage kind=documentconflict" : "close: documentconflict";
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Thanks @Rash419. Looks reasonable. 

I'd suggest the typical convention of having log messages in the injunctive. That is, make the subject read as what you want to do. For example:

`wsd: consider possibly-modified docs when handling conflicts`

And you can give the following explanation in the body of the commit log:

```
When the storage returns modified-in-storage or conflict status
to an upload attempt, we prompt the user to handle the conflict
when the document is modified. Previously, we only prompted
when we had the modified flag set, ignoring the possibility of
having a race with the modified flag, in case the document was
just modified, or potentially modified.

We now prompt the user even when the document is potentially
modified, giving the user the opportunity to decide what to do.
```

This is just a suggestion. Feel free to have your own message, or just use the above, as you see fit.

This is a general suggestion for all future commits. It's always good to have short and descriptive log messages followed with explanation of what the issue was and how the fix should improve things.

Feel free to ask if aren't sure or would'd like some feedback.

Thanks!

Ashod (approved):
Thanks!

Issue #3878, opened 2021-12-20
Improve COOL browser logging


alessioalex-1and1:
**Is your feature request related to a problem? Please describe.**
When I started out using the project, the first thing I did was to insert `console.log` here and there to try to figure out things. I wasted between 30 minutes and an hour not knowing that all the `console` native browser functions were overwritten by COOL to empty functions. It also happened at a later time when I forgot about this issue and it took me a couple of minutes to remember this awkward functionality.

**Describe the solution you'd like**

Native JavaScript functions should never be overwritten, for multiple reasons. COOL needs to use its own logging methods and it can setup those to not log anything to the `console` if that is the desired behavior.

**Describe alternatives you've considered**
This manifests when making the 'production' build, so normally one might not encounter this problem. However until recently you couldn't separate having multiple files or a single file based on a build flag (apart from the debug vs production one), so if you wanted to mimic the production JavaScript && CSS build into a single file you needed to run the build for production. Nevertheless this shouldn't be an issue requiring workarounds to fix.

**Additional context**
-

PR #3879, opened 2021-12-20
browser: improve logging by using custom logging methods, no console overriding

alessioalex-1and1:
Change-Id: I0f6c0cbe2510c3e621cb75a8a76ef79c752e3d99

* Resolves: #3878
* Target version: master 

### Summary

Instead of using `console` methods everywhere now we will be using `window.app.console` while having the same methods as before. It is the easiest transition to a better logging solution as we will keep using the same methods as we used before with a prefix basically (like for example `window.app.console.log(<string>)`) and it allows us to set those functions to do nothing when needed.

### TODO

- [ ] Make sure I haven't omitted any `console` function somewhere, I have used the following command to track them down:

```bash
grep -r "console\." . --exclude-dir="node_modules" --exclude-dir=".git" --exclude-dir="archived-packages" --exclude-dir="dist" --exclude-dir="jails" --exclude-dir="po" --exclude-dir="test" --exclude-dir="admin" -l --include \*.js
```

Note: I haven't touched external libraries found in `browser/js` such as 'Autolinker', 'select2', 'w2ui' or 'vex'. They should stay the same imo.

### Checklist


alessioalex-1and1:
closing this PR in favor of https://github.com/CollaboraOnline/online/pull/3883

not sure what happened but after squashing into a single commit the Github interface is not showing the proper diff with 45 files changed but instead it's showing 261 files changed

PR #3883, opened 2021-12-20
browser: improve logging by using custom logging methods instead of c…

alessioalex-1and1:
…onsole ones directly

Change-Id: Id4ad5711c85998ce71ba0849fca5fe6dc580f395

* Resolves: #3878
* Target version: master 

### Summary

Instead of using `console` methods everywhere now we will be using `window.app.console` while having the same methods as before. It is the easiest transition to a better logging solution as we will keep using the same methods as we used before with a prefix basically (like for example `window.app.console.log(<string>)`) and it allows us to set those functions to do nothing when needed.

### TODO

- [ ] Make sure I haven't omitted any `console` function somewhere, I have used the following command to track them down:

```bash
grep -r "console\." . --exclude-dir="node_modules" --exclude-dir=".git" --exclude-dir="archived-packages" --exclude-dir="dist" --exclude-dir="jails" --exclude-dir="po" --exclude-dir="test" --exclude-dir="admin" -l --include \*.js
```
### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -43,6 +43,7 @@ window.app = { // Shouldn't have any functions defined.
 		}
 	},
 	socket: null,
+	console: {},
 };
 
 (function (global) {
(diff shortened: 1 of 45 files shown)


mmeeks (approved):
Thanks Alexandru - looks good; I expect it helps the casual developer too.
I wonder if we should allow this stuff to be turned on & off in our tripple-click help->about dialog popup: where we mis-use these 'layer check-boxes' to select that sort of stuff.
I wonder what our stack of libraries we use underneath does too - perhaps they use console.log left and right.
Can you re-base this and lets get it in ASAP.
Thanks!

mmeeks:
We should also make CodeQL - happy:

/home/runner/work/online/online/browser/js/global.js
  78:24  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
  89:1   error  Expected indentation of 2 tabs but found 1

alessioalex-1and1:
> Thanks Alexandru - looks good; I expect it helps the casual developer too. I wonder if we should allow this stuff to be turned on & off in our tripple-click help->about dialog popup: where we mis-use these 'layer check-boxes' to select that sort of stuff. I wonder what our stack of libraries we use underneath does too - perhaps they use console.log left and right. Can you re-base this and lets get it in ASAP. Thanks!

Since it's a custom function and not involving the native `console.log` directly this should be easily doable. I will resolve the CodeQL thing and the conflicts and push again.

mmeeks:
Thanks Alexandru =)

mmeeks:
I was about to merge this - which looks fine; and then considered the scope for conflicts as we back-port to the 21.11 branch for the next week (or worse silent exceptions causing very odd behavior).

It sucks to do it, but can we delay another week ?

alessioalex-1and1:
@mmeeks sure, comment here before you want to make the merge, there might be a few conflicts to solve by then.

mmeeks:
Urgh; github makes a separate merge commit - can you re-base properly and just push this ? =)

Thanks !

mmeeks:
Lovely - thanks ! =)

PR #3890, opened 2021-12-21
Sidebar Effect Section Color width fix

Andreas-Kainz:
### Steps to reproduce
- Draw a Shape
- open Sidebar
- open Effect Section
- Change color to Light Gray 4
- width is to small for the label / cause of the preview rectangular
- as fillcolor1 and fillcolor2 has the same layout than #LB_GLOW_COLOR
  so width width was synced

Change-Id: I2c1ea98bd462cac899d9e118d520cc51749f3ba4

**before**
Screenshot_20211221_205852

**after**
Screenshot_20211221_210012

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -280,14 +280,7 @@ td.jsdialog .jsdialog.cell.sidebar {
 	top: 6px;
 }
 
-#LB_SHADOW_COLOR, #LB_GLOW_COLOR {
-	padding: 4px 10px;
-	width: 130px;
-	position: relative;
-	left: -11px;
-}
-
-#fillgrad2 {
+#LB_SHADOW_COLOR, #LB_GLOW_COLOR, #fillgrad2 {
 	padding: 4px 10px;
 	width: 160px;
 	position: relative;


pedropintosilva:
Thanks @Andreas-Kainz but I actually wonder why we even start to have the colour names there in the first place it seems very intrusive. I would rather show only the colour (image element without span) and on hover show the colour name. Unless there was a very strong reason that forced us in the past to show the label at all times It think this is a no-brainer. I think this behaviour is coming from core (some colour dropdown have name others no)

Andreas-Kainz:
Yes it come from core. I don't have a problem show the #3b3b3b color, but on the other hand it's also not to intuitive.

Andreas-Kainz:
Please fill a bug report at LibO and discuss there what should be shown. Should be fixed at LibO and not at cool.

However in cool you have two options
1. Extend the width as we already have at  area fill gradient colors
2. Use ellipse

pedropintosilva:
Will do, for now if we want to improve quickly this (and because the way these dropdowns were made in the first place on Online we do not know how this will look even if the fix is on the core side) we could just hide the colour label for now, this would ensure that all of them have the same width

pedropintosilva:
closing this now, I think this has been fixed by hiding the label

PR #3894, opened 2021-12-22
Use var(--border-radius)

Andreas-Kainz:
add a --border-radius: 4px value
and use it for sidebar and notebookbar
other stuff will follow, but I want to test where the value was visible
selected-inline values in notebookbar use border-radius

**test change**
- use notebookbar in writer
- goto tab review
- Automatic Spell Checking use now the border-radius: 4px

Change-Id: I3083cd2ff2cc3efd784aafb0c71cb2f823b9b79d

value name is similar to https://github.com/nextcloud/server/blob/master/core/css/css-variables.scss#L57

### DIFF
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -11,4 +11,6 @@
 	--gray-lighter-bg-color: #F7F7F7;
 	--gray-color: #b6b6b6;
 	--white-bg-color: #fff;
+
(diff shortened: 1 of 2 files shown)


Rash419:
@Andreas-Kainz good to get it reviewed from pedro , thanks

pedropintosilva (approved):
thanks!

PR #3896, opened 2021-12-22
notebookbar: use more advanced optioned Conditional Formatting control

merttumer:
Home tab's contitional formatting control is old and is not the same
as Format's tabs control. Now they are the same.

Change-Id: I9a08cda3a2ff98a18e711ba4a17bd8ebb8b4027b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/w2ui-1.5.rc1.js
+++ b/browser/js/w2ui-1.5.rc1.js
@@ -2529,7 +2529,7 @@ w2utils.event = {
             if (typeof options.onRender === 'function' && typeof options.render !== 'function') options.render = options.onRender;
             // since only one overlay can exist at a time
             $.fn.w2menuClick = function (event, index) {
-                var keepOpen = false;
+                var keepOpen = options.keepOpen || false;
                 if (['radio', 'check'].indexOf(options.type) != -1) {
                     if (event.shiftKey || event.metaKey || event.ctrlKey) keepOpen = true;
                 }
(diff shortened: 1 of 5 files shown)


eszkadev (commented):
looks good, but I noticed one issue:
1. open 'conditional formatting' menu
2. click on 'icon set' and overlay is opened
3. click again on 'icon set' - menu disappears and overlay is positioned in the top left corner of an app

eszkadev:
It would be also good to unify two 'conditional formatting' widgets in the notebookbar (one is in home tab, second in the formating tab). Second one opens dialog on 'icon set' selection. Or is that expected?

merttumer:
@eszkadev I pushed a fix for both of the issues addressed. Can you re-review it? Thank you.

eszkadev (approved):
All issues are fixed now

browser/js/w2ui-1.5.rc1.js
```diff
@@ -2529,7 +2529,7 @@ w2utils.event = {
             if (typeof options.onRender === 'function' && typeof options.render !== 'function') options.render = options.onRender;
             // since only one overlay can exist at a time
             $.fn.w2menuClick = function (event, index) {
-                var keepOpen = false;
+                var keepOpen = options.keepOpen || false;
```
vmiklos, line 2532: Is this intentional? Looks like a debugging leftover.

merttumer: Yes, I needed to touch the w2ui library file as agreed with @eszkadev and changed this. Not sure why this was hardcoded to false however I did not want to break the existing code and want to pass this option inside the options object so I made it this way to be safe. Since we dont compile w2ui from npm I guess this is alright.

vmiklos: Ah, I see, thanks. Perhaps it is worth to add a comment above to line to explain why is this necessary? So a next person does not "clean it up" for you. :-)

PR #3900, opened 2021-12-23
scrolling: make swipe scrolling faster

lpranam:
Change-Id: I42e3a0aa79f9cbda1f5b4d3c1a3cd6d55a28292f

* Target version: master 

### Checklist


merttumer (commented):
I think it is too much fast for writer but it is ok for calc

lpranam:
in such a case, we can also maintain different scrolling speeds for different apps if that sounds like a good idea.

PR #3903, opened 2021-12-24
log more info about administrative document adding

yunusem:
Target version: master 

### Summary
Added new LOG_INF line

### Checklist

### DIFF
--- a/wsd/AdminModel.cpp
+++ b/wsd/AdminModel.cpp
@@ -508,13 +508,13 @@ void AdminModel::addDocument(const std::string& docKey, pid_t pid,
                              const int smapsFD, const std::string& wopiHost)
 {
     assertCorrectThread();
-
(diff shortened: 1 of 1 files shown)


mmeeks (changes requested):
Looks reasonable - but you need to use the anonymization infrastructure:

git grep anonomizeUrl
git grep anonymizeUsername

That's very important to some users.

Thanks.

mmeeks (approved):
Ok - looks good; we need to iterate this though - LOG_INF is far too low a level; there is much too much debug churning out there which will impact server performance -horribly- negatively. It must not be turned on in this state on any machine.

We need a new config setting inside the <logging> sub-section:

<docstats type="bool" default="false">false</docstats>

And to check this - and then to log un-conditionally if that is on; ie. read the common/Log.hpp and in-line the writing to the Log::logger().log() without having a check for 'trace' etc. - we should do it un-conditionally if it is turned on.

Also with a nice prefix: "docstats:" eg. so a tool can extract these easily.

Thanks !

PR #3906, opened 2021-12-26
Header area DocumentName without cutted extension

Andreas-Kainz:
there is hardcoded at the input#document-name-input a fixed width
with the defined fixed width the extension was cutted
this MR get 24px additional width so that the extension was
shown with it has 4 letters.

**before**
Screenshot_20211226_102622

**after**
Screenshot_20211226_103021

Change-Id: I8736ac08e2f4ab27703b6dcedf3e1d28d73de516

### DIFF
--- a/browser/src/control/Control.DocumentNameInput.js
+++ b/browser/src/control/Control.DocumentNameInput.js
@@ -125,7 +125,7 @@ L.Control.DocumentNameInput = L.Control.extend({
 		var documentNameInput = $('#document-name-input');
 		var content = (typeof tail === 'string') ? documentNameInput.val() + tail : documentNameInput.val();
 		var font = documentNameInput.css('font');
-		var textWidth = L.getTextWidth(content, font) + 24;
+		var textWidth = L.getTextWidth(content, font) + 48;
 		var maxWidth = this._getMaxAvailableWidth();
 		//console.log('_setNameInputWidth: textWidth: ' + textWidth + ', maxWidth: ' + maxWidth);
 		textWidth = Math.min(textWidth, maxWidth);


Rash419:
@Andreas-Kainz I could not reproduce the issue 
image
In my build , it represents the file extension correctly

Andreas-Kainz:
image

you are correct @Rash419 top is firefox, bottom chrome. Issue is available only at firefox

Andreas-Kainz:
what is really strange, the element.style { width: } is different in firefox and chrome.

pedropintosilva:
@Andreas-Kainz it seems this is suposed to target readonly document cases right?

pedropintosilva:
I wonder if it wouldn't be better to simply set the width to auto , I think we do the that for when it;s editable (#document-name-input.editable)

Andreas-Kainz:
image

true @pedropintosilva it was fixed for toolbar layout with `width: auto !important;`

pedropintosilva:
@Andreas-Kainz exactly, so as I was saying probably better to assure that we have everything set to auto. Probably even better would be to remove all those pixels from JavaScript side (provided that there is no other situation that we actually need them).

I quickly tried and it seems that nothing breaks if we apply this patch: https://share.collabora.com/index.php/s/7d6YYJSgnkj3aNd could you please play/test further?

pedropintosilva:
note: as far as I saw it seems that the exact width (via JS) was needed before but not now

Andreas-Kainz:
I'm sorry but without help I have to close this PR.

pedropintosilva:
@Andreas-Kainz try to test the patch I have sent you (https://share.collabora.com/index.php/s/7d6YYJSgnkj3aNd ) by via `git apply documentnameinput-remove-width.patch` and see if it needs more work on top or if you would like adjust something

Andreas-Kainz:
> @Andreas-Kainz try to test the patch I have sent you (https://share.collabora.com/index.php/s/7d6YYJSgnkj3aNd ) via `git apply documentnameinput-remove-width.patch` and see if it needs more work on top or if you would like adjust something

patch work, as it's your patch, please close the PR and submit your patch. Thanks

pedropintosilva:
Will do. Thanks for double checking and testing !

PR #3907, opened 2021-12-26
Calc sidebar NumberFormat section alignment

Andreas-Kainz:
spinboxes are right aligned
labels are vertical center aligned
remove padding at table element 
so all commands have the same row height

Change-Id: If5b89cac71e2918d3d870066f2b895136286b8b2

**after**
image


Rash419 (approved):
Looks good , @Andreas-Kainz thanks :)

PR #3909, opened 2021-12-29
Fix InsertFile returns 400 after an idle reload

merttumer:
Change-Id: I56d38f05d8da4fc55393203764692ca2b5f864f2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected

PR #3913, opened 2021-12-31
co-6-4: 'SaveAs' Document Templates

hcvcastro:
- kit: include document properties to "status:" message
- lolealfet: parse new document properties
- loleaflet: only 'saveAs' action is allowed if document


timar:
@hcvcastro can you rebase this PR on the top of current co-6-4?

Ashod (commented):
Thanks @hcvcastro. What happens when we edit and close such a document? There is no UI, so WSD will need to handle the template saving. But I don't see how that should work.

PR #3914, opened 2021-12-31
kit: preload: enable module 'ucpgio1'

hcvcastro:
Otherwise opening a document template crash the kit,
because the provider does not exists.

i.e. loading file Alizarin.otp

Change-Id: Icd8911d75e5e699d864a328218787ae0b29cd117

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2912,7 +2912,7 @@ bool globalPreinit(const std::string &loTemplate)
     // desktop or developer's install if env. var not set.
     ::setenv("UNODISABLELIBRARY",
              "abp avmediagst avmediavlc cmdmail losessioninstall OGLTrans PresenterScreen "
-             "syssh ucpftp1 ucpgio1 ucphier1 ucpimage updatecheckui updatefeed updchk"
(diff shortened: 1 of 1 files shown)


timar (commented):
In production we build LOKit without gio support, i.e. `--disable-gio`. Maybe the patch helps when you don't disable gio in your dev build. Was that the case?

hcvcastro:
According to my tracks, the universal content broker tries to create a default provider service "com.sun.star.ucb.GIOContentProvider" which it is defined in "libucpgio1lo.so", without the provider
I cannot load the template Alizarin.otp in my local build.
In my development build I did not use the parameter --disable-gio

mmeeks:
We really don't want to be using gio in production; we want to shrink our code & dependency footprint as much as possible - bringing all of the latest glib in there is far from ideal.

Is there some way we can fix the crash and/or warn people about builds with gio ? is there a ticket for the crash ?

Thanks !

hcvcastro:
yes, I can change the implementation the UNO service default provider to something dummy (or fake) or bypass I think.

timar:
> yes, I can change the implementation the UNO service default provider to something dummy (or fake) or bypass I think.

Thank you, that'd be the proper solution. Let's close this PR.

PR #3921, opened 2022-01-04
Create a link to the master UDS file inside the jail when we aren't using abstract sockets.

arrowd:


mmeeks (approved):
Looks fine to me - thanks Gleb ! =)

PR #3923, opened 2022-01-05
wsd: fix: remove unnecessary condition

Rash419:
Change-Id: Ibc5338b34713233c540e0de938dbb76b88398ef5

* Target version: master


Ashod (approved):
Great, thanks @Rash419.

(Minor note: don't think having "fix" in the log message is necessary, because the description implies it.)

PR #3931, opened 2022-01-06
Fix selecting all cells with top left corner btn make the focus lost

merttumer:
This makes keyboard handler does not receive key events
and functions like delete does not work.

Change-Id: If49d8ff760c9db774d85d8fcdb7eb40fd85c4329

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
Looks good, fixes the issue.

Issue #3916, opened 2022-01-02
CODE 21.11.6 Regression: Does Not Open Specific Calc Spreadsheet


gohrner:
**Describe the bug**
I have a specific spreadsheet which cannot be opened with COOL 21.11.6. Previously with 6.4.x it worked fine.

Other files (spreadsheets as well as documents) CAN be opened, so the COOL 21.11.6 installation itself basically seems to work.

From a user's point of view, either:

* Some parts of CODE like menu bar, status bar and also sidebar, if expanded, load and display fine, however the document area stays white,
* or on other occasions the screen stays in the loading-artwork, and the progress bar is stuck at 0%.

Unfortunately, in both cases there's nothing interesting in the COOL logs for this file:

```text
wsd-00001-00082 2022-01-02 21:41:32.241483 +0000 [ docbroker_002 ] WRN  Waking up dead poll thread [HttpSynReqPoll], started: false, finished: false| ./net/Socket.hpp:718
wsd-00001-00082 2022-01-02 21:41:32.304258 +0000 [ docbroker_002 ] WRN  Waking up dead poll thread [HttpSynReqPoll], started: false, finished: false| ./net/Socket.hpp:718
wsd-00001-00082 2022-01-02 21:41:36.751748 +0000 [ docbroker_002 ] ERR  Error while handling poll for socket #20 at 0 in docbroker_002: JSON Exception| net/Socket.cpp:468
wsd-00001-00082 2022-01-02 21:42:20.204081 +0000 [ docbroker_002 ] WRN  Ignoring attempted read from 26| ./net/Socket.hpp:1095
^C
```

These log messages look basically the same for files that work.

COOL 21.11.6 in self-build Ubuntu-based Docker container on x64 platform. (Host is Debian Buster.)

Browser is Chromium Version 95.0.4638.69 (Offizieller Build) Built on Ubuntu , running on neon 20.04 (64-Bit).

I cannot publicly attach the problematic file here, and I need to roll back to a working version now, so my testing possibilities are limited.

Maybe someone else experienced similar behaviour?

pedropintosilva:
Thanks @gohrner for reporting it, I see that the file is private but could you perhaps answer these questions? and/or attach a "clean" version without those private bits?:

- How heavy is the file?
- How many cells roughly ?
- Does it have anything out of ordinary? Or function?

gohrner:
The file actually is not really big or complex - in my opinion.

[...]

PR #3933, opened 2022-01-06
Handle JSON exception when parsing commands

eszkadev:


eszkadev:
just wrapping JSON parsing with try {} catch {} so we don't stop the app

eszkadev:
should fix https://github.com/CollaboraOnline/online/issues/3916

eszkadev:
@Ashod could you look if we need to return something in case of an exception in that method?
(It was hard to me to understand what means false or true when following the calls, no comment above declaration either)

btw real fix for incorrect JSON is also made, but in core ;)

Ashod (approved):
I checked why leaking the exception didn't get gracefully handled. Basically, this exception is happening while we inspect the data in DocBroker, before forwarding to clients. So this scenario doesn't benefit from the default exception belts we have around external data.

Let's get it in, this is good. Thanks @eszkadev!

Issue #3725, opened 2021-11-26
Sidebar scrollbar differs from document scrollbar and partly hides icons


sebastiaanveld:
Describe the bug
Sidebar scrollbar differs from document scrollbar and partly hides icons: If you select an object that shows a lot of items in the sidebar or when you expand items in the sidebar so the bar gets to large for the screen a scrollbar appears. This scrollbar:
- Has a different look versus the scrollbar used for like the document area. 
- It, probaly cause it is larger then intended, slightly hides the most right icons in the sidebar. 

To Reproduce
- Start a document
- Add an object line an picture or slect something that has a lot of items in the sidebar, or expand all items in the sidebar so that scrollbar in the sidebar appears.
- The scrollbar is thick compared to the scrollbar of the document area. It also partly hides the most right icons in the sidebar.

Expected behavior
I would expect we use the same scrollbars for the sidebar as used in the document area. That would solve the partly hidden icons too.

Actual behavior
Different scrollbar, visually off comparted to scrollbar in document area. Icons are slightly hidden.

Screenshots
![afbeelding](https://user-images.githubusercontent.com/37268999/143536265-184d7a78-5a78-4c32-9d1e-e6ed1ead78b8.png)

Desktop (please complete the following information)
- OS: Windows 10 (2004)
- Browser: Firefox 94.0.2 (64 bit), Chrome  96.0.4664.45 (Officiële build) (64-bits)
- Version: Nextcloud staging.eu.collaboraonline.com
- COOLWSD version: 21.11.0.3 git hash: 55ed82d7
- LOKit version: Collabora Office 21.06.8.1 git hash: ceddcf2

Ezinnem:
I do notice that the scrollbar covers an icon as stated.

![scrollbar-outlook](https://user-images.githubusercontent.com/54344301/143580246-02c64ce6-b31a-4e3f-a464-b4c1428ee30b.png)

sebastiaanveld:
@Ezinnem mind you have a small scrollbar in the sidebar just like like the scrollbar that is in the document's frame. In my screenshot the scrollbar in the sidebar is different and wider in size as in using more space.

sebastiaanveld:
Looks fixed in latest CODE, using
COOLWSD versie: 21.11.3.6 git hash: [eb73aa3]
LOKit versie: Collabora Office 21.06.26.1 git hash: [524dda8]

Just to show the difference:
[...]

PR #3939, opened 2022-01-07
Fix Sidebar scroll bar width and improve consistency

pedropintosilva:
On Chrome the scroll bar was overlapping some sidebar icons; The
scroll bar thumb was not the same color as the document scroll bar.
Additionally on Firefox (using Windows) the sidebar-panel's scroll bar
would also overlap a bit of those sidebar icons.

- Fix sidebar scroll bar width and make sure is not too big
	- Done in different ways for Chrome and FF since they do not posses the
		same CSS spec (plus FF does not allow the use of exact widths)
- Fix sidebars inconsistencies regarding color
	- Use the same #bec0c0 used in the document's scroll bar albeit only
		on mouseover. In the default state it stays with the browser's default
		opacity and thus looks lighter than the document's scroll bar. This
		is not only fine but desirable so to avoid taking to much attention
		from the document itself.

First reported:
https://github.com/CollaboraOnline/online/issues/3725

Change-Id: I7b4f2aeaf2a3b7a1af88b01e36f82d1cb7746f0d

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -200,6 +200,16 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	overflow-x: hidden;
 	overflow-y: auto;
 	z-index: 1200;
+	scrollbar-width: auto;
+	scrollbar-color: #bec0c0 transparent;
+}
+
+#sidebar-panel::-webkit-scrollbar {
+	width: 8px;
+}
+
+#sidebar-panel::-webkit-scrollbar-thumb {
+	background-color: #bec0c0;
 }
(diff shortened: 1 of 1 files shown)


pedropintosilva:
before see images from: https://github.com/CollaboraOnline/online/issues/3725

---

After with #3939 :

| win-chrome |
|:--:|
| <b>Chrome</b>|

| win-ff |
|:--:|
| <b>Firefox on Windows (disregard font rendering problems, culprit: Wine) </b>|

gokaysatir (approved):
Thanks for the improvement :)

PR #3942, opened 2022-01-09
Backport Unloading and Saving

Ashod:
- cypress: close the test document before reloading
- cypress: give each test a unique test document
- wsd: remove DocBroker from Admin last
- wsd: save and wait until document isn't modified
- wsd: test: add slow save and upload test
- wsd: do not count disconnecting sessions as active
- wsd: test: fail when modified docs are unsaved
- wsd: do not create client sessions on docs marked to destroy
- wsd: more frequent polling while unloading documents
- wsd: improved logging around session add/remove
- wsd: stop DocBroker when loading fails
- wsd: improve logging and doc state dumping
- wsd: enforce TLSv1.1 as the minimum acceptable version


kendy (approved):
Great stuff, thank you @Ashod (and @Rash419 for the multi-user fix)!  Looks good to me, let's merge.

@Ashod  - I see you had to disable a test here, I suppose you'll be investigating that still, right?

Ashod:
> I see you had to disable a test here, I suppose you'll be investigating that still, right?

@kendy I couldn't get that test to pass locally even on very old commits. So either it's unstable or broken or there is something wrong. It failed on the CI too and since I could reproduce it *without* my patches, I couldn't realistically chase and fix it when I didn't even know if there were other issues with *my* patches, so I excluded it to have a chance of fixing issues with my patches before I ran out of time.

Yes, I'll chase it.

PR #3944, opened 2022-01-10
Aboud: Ipad: Fix git hash formatting

pedropintosilva:
In the wsd version git hash was getting set right after the previous
word (completely adjacent and without space and brackets)
- Fix outdated "loolwsd-*" class name and make sure the CSS rules does
	get applied

Change-Id: I77ea5ac19a5dbaa798437ec70489a1aee1051b1f

### DIFF
--- a/browser/css/device-tablet.css
+++ b/browser/css/device-tablet.css
@@ -160,12 +160,12 @@
 	margin: 0;
 }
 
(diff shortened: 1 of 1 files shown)


tml1024 (approved):
Works as intended, approved.

PR #3948, opened 2022-01-10
Notebookbar: Fix Tabs' overlapping into other elements

pedropintosilva:
Make the whole nav bar scroll-able, specially important when still on
desktop but using small factor screen or simply when resizing window

Change-Id: Ib0011c8353a5c4d09f5f150032088bc69e0f7d90

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -77,6 +77,12 @@
 }
 .main-nav.hasnotebookbar {
 	height: 32px;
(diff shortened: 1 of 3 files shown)


pedropintosilva:
Fix this: 
image

Rash419 (approved):
Looks good , Thanks. I have enabled the test that was failing due to issue.

PR #3951, opened 2022-01-10
RTL UI

eszkadev:


eszkadev:
can be tested with `&lang=ar` parameter in the URL

gokaysatir (approved):

eszkadev:
in case this will fail or we will have time to wait a bit longer there is better PR with comments in writer fixed:
https://github.com/CollaboraOnline/online/pull/3968

eszkadev:
Merged another PR with the same patches

PR #3956, opened 2022-01-11
RTL support for Calc

dennisfrancis:
* Target version: master 

### Summary
RTL support for Calc

This needs core side patches from https://cgit.freedesktop.org/libreoffice/core/log/?h=feature/lok-calc-rtl

### Checklist


gokaysatir (approved):
Thanks for the great improvements:)

browser/src/control/Control.AutofilterDropdown.js
```diff
@@ -131,6 +131,9 @@ L.Control.AutofilterDropdown = L.Control.extend({
 				top = top - splitPos.y;
 		}
 
+		if (this._map._docLayer.isCalc() && this._map._docLayer.isLayoutRTL())
+			left = this._map._size.x - left;
```
Ashod, line 135: Shouldn't we always have braces?

browser/src/control/Control.ColumnHeader.js
```diff
@@ -203,7 +209,13 @@ L.Control.ColumnHeader = L.Control.Header.extend({
 				width = 0;
 			}
 
-			width += dragDistance[0];
+			var isRTL = this.isCalcRTL();
+
+			if (isRTL)
+				width -= dragDistance[0];
+			else
+				width += dragDistance[0];
```
Ashod, line 217: Same. I think having braces around even single-line scopes is the convention used in .js/.ts.

browser/src/layer/tile/TilesSection.ts
```diff
@@ -625,6 +649,7 @@ class TilesSection {
 			var toScaleAbs = relScale * docLayer._tileSize * 15.0 / docLayer._tileWidthTwips;
 			toScaleAbs = docLayer._tileSize * 15.0 / Math.round(15.0 * docLayer._tileSize / toScaleAbs);
```
Ashod, line 650: It would seem toScaleAbs is not used. Is this a bug?

browser/src/layer/vector/CanvasOverlay.ts
```diff
@@ -55,16 +96,16 @@ class CanvasOverlay {
...
+	private transforms: TransformationsList;
 
 	constructor(mapObject: any, canvasContext: CanvasRenderingContext2D) {
 		this.map = mapObject;
 		this.ctx = canvasContext;
 		this.tsManager = this.map.getTileSectionMgr();
 		this.overlaySection = undefined;
 		this.paths = new Map<number, CPath>();
-		this.transform = new OverlayTransform();
+		this.transforms = new TransformationsList();
```
Ashod, line 108: The single `s` at the end is a very small difference. This wouldn't be a problem if there were no `transform` (without `s`) elsewhere. But since we are using both variables with a single character difference, sooner or later someone is going to have a typo and cause a bug.

Perhaps we should make this `trasnformList` to avoid confusion and accidental bugs ?

browser/src/layer/vector/SVGGroup.js
```diff
@@ -76,6 +76,11 @@ L.SVGGroup = L.Layer.extend({
 		this._update();
 	},
 
+	isCalcRTL: function () {
```
Ashod, line 79: This exists in other file(s) too. Perhaps it's better to make this part of `docLayer` and replace it with a single `this._map._docLayer.isCalcRTL()` call? Sure it's not great to have application-specific helpers in docLayer (calc, in this case), but copy pasting these 4 lines is probably worse.

browser/src/layer/vector/SVGGroup.js
```diff
@@ -133,6 +140,8 @@ L.SVGGroup = L.Layer.extend({
 
 		if (this._map) {
 			var pos = this._map.mouseEventToLatLng(evt);
+			if (this.isCalcRTL())
+				pos = this._map.negateLatLng(pos);
```
Ashod, line 144: Braces?

browser/src/layer/vector/SplitPanesSVG.js
```diff
@@ -72,6 +72,19 @@ L.SplitPanesSVG = L.SplitPanesRenderer.extend({
 		L.SplitPanesRenderer.prototype.onRemove.call(this);
 	},
 
+	_isCalcRTL: function () {
+		var docLayer = this._map._docLayer;
+		return docLayer.isCalc() && docLayer.isLayoutRTL();
```
Ashod, line 77: Consider making this part of `docLayer`. Too many copy/pastes.

browser/src/map/Map.js
```diff
@@ -989,6 +989,13 @@ L.Map = L.Evented.extend({
 		return this.options.crs.pointToLatLng(L.point(point), zoom);
 	},
 
+	/// Get LatLng coordinates after negating the X cartesian-coordinate.
```
Ashod, line 992: This describes what the function does, which is easy to read/see, but it doesn't explain why or when/where this is useful. It would be helpful to maintainers to read a note as to when this is typically used so they can more easily understand the purpose of this function. I.e. the 'why' as opposed to the 'what'.

kit/KitHelper.hpp
```diff
@@ -89,6 +90,11 @@ namespace LOKitHelper
                         if (mposs.str().empty())
                             mposs << prop.second;
                     }
+                    else if (name == "rtllayout")
+                    {
+                        if (prop.second == "1")
+                            rtlposs << i << ",";
```
Ashod, line 96: Of course we could have a single-char instead of a literal-string here:
```
rtlposs << i << ',';
```

Ashod (commented):
Thanks @dennisfrancis! Left some notes. Hope they are helpful.

browser/src/control/Control.AutofilterDropdown.js
mmeeks: Ideally not - so the code is readable ;-)

timar:
@gokaysatir gave +1 before my patch with the braces, so we can merge. 
@dennisfrancis could you please read @Ashod 's notes, and if you agree, make those changes in a follow-up commit.

browser/src/layer/tile/TilesSection.ts
dennisfrancis: These lines are not from this PR, but from older zoom-animation related changes. Looks like they weren't removed after factoring some code to private utility functions (scalePosForZoom etc). I'll remove them in a follow up.

browser/src/layer/vector/CanvasOverlay.ts
dennisfrancis: This particular variable is a private member of the CanvasOverlay class. The other transform variable is a local variable of a different type (OverlayTransform). The typescript transpiler will complain if anyone uses `transforms` instead of `transform` or vice-versa. But I agree that renaming `transforms` to `transformList` makes it more distinctive and readable.

dennisfrancis:
https://github.com/CollaboraOnline/online/pull/3972 contains the follow-up patches.

PR #3963, opened 2022-01-11
Notebookbar: Fix smartmenus visibility and set overflow dynamically

pedropintosilva:
c9b4b619cfde85553761662f9a7d5914e246b2f4 fixed the tabs overlap issue
but for the smartmenus on notebookbar (hamburger menu) to work the CSS
overflow property needs to be set (both for x and y) to visible, any
other value and the smartmenus' visibility suffers (gets cropped).
- Make set the CSS overflow property of .main-nav elm dynamically
	depending on menu state.
		- If it's open: set to `visible`
			(this disables the .main nav scroll but that's completely fine
			for when the Hamburger menu is open since it's not needed)
		- If it's close: back to `scroll hidden`.

Change-Id: I8a5522cb4a107d0583cde252e2ee551124c06f02

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -77,7 +77,7 @@
 }
 .main-nav.hasnotebookbar {
 	height: 32px;
-	overflow: scroll hidden;
+	/* overflow is set dynamically */
 	scrollbar-width: none;
 	-ms-scrollbar: none;
 }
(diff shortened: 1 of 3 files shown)


eszkadev (approved):
Tested and works fine

pedropintosilva:
This is failing in an unrelated unit test, I think..

pedropintosilva:
_rebasing and pushing_

eszkadev (approved):

PR #3966, opened 2022-01-11
this is a temporary commit which will be reverted

lpranam:
this will help solve the deadloack for failing CI between:
https://gerrit.libreoffice.org/c/core/+/128107
https://github.com/CollaboraOnline/online/pull/3960

Change-Id: If3b13b49994ace7de48b0da7a99602508c36fcd7

### Checklist


vmiklos:
FWIW I would not revert this: this fixes a dependency cycle between online.git and core.git. The idea was always to allow newer core with unchanged online, and this works usually, except for this case. I would rather forward-port this commit to master as well, permanently. :-)

lpranam:
@vmiklos okay will take care of it then :)

PR #3968, opened 2022-01-11
RTL UI (with comments fixed)

eszkadev:


gokaysatir (dismissed):
Thanks for the improvements and the fix :)

gokaysatir (approved):
Thanks also for the second fix :)

PR #3970, opened 2022-01-11
coolconfig: updated man page

thebearon:
Follow-up to a6b258a0db07a4fa720903f1fcf2e32c4677bb47

Change-Id: I79e457bb6fd410cbb40de500d2f3ce0f33f2e9ae

* Target version: master

### DIFF
--- a/man/coolconfig.1
+++ b/man/coolconfig.1
@@ -8,7 +8,7 @@ coolconfig COMMAND [OPTIONS]
 .SS "General options:"
 \fB\-h\fR, \fB\-\-help\fR                Show this usage information.
 .PP
-\fB\-\-config\-file\fR=\fIpath\fR        Specify configuration file path manually.
+\fB\-\-config\-file\fR=\fIpath\fR        Specify configuration file path manually. The default is usually \fI/etc/coolwsd/coolwsd.xml\fR.
 .SS "anonymize"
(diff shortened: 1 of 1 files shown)


man/coolconfig.1
```diff
@@ -8,7 +8,7 @@ coolconfig COMMAND [OPTIONS]
 .SS "General options:"
 \fB\-h\fR, \fB\-\-help\fR                Show this usage information.
 .PP
-\fB\-\-config\-file\fR=\fIpath\fR        Specify configuration file path manually.
+\fB\-\-config\-file\fR=\fIpath\fR        Specify configuration file path manually. The default is \fI/etc/coolwsd/coolwsd.xml\fR.
```
timar, line 11: Not always. See https://github.com/CollaboraOnline/online/blob/0340dbd8938a0ab36f369d93cc8696842361c08c/tools/Config.cpp#L89 On FreeBSD I think it's not /etc.

timar (changes requested):

man/coolconfig.1
thebearon: How about: `The default is the path to coolwsd.xml`?

timar: The default is usually _/etc/coolwsd/coolwsd.xml_. We covered 99.9% of the cases.

PR #3975, opened 2022-01-12
Don't send unnecessary resizewindow message for the Calc input bar

tml1024:
If the size is the same as last time we sent the message, don't send
it. This fixes ugly flashing of the input bar tunnelled dialog when
the symbols displayed in front of the input field change.

Change-Id: I6532d85fddede54f416ee6686958bb650fdcb4f1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
(Updated initial comment here to match the updated commit message. Pity that GitHub doesn't do that automatically when you force push the single commit of a PR.)

mmeeks (approved):
Looks very sensible to me - thanks Tor =)

PR #3986, opened 2022-01-13
RTL: fix tunneled child windows

eszkadev:
use margin-inline-start with relative offset instead of fixed position


eszkadev:
test with these core changes:
https://gerrit.libreoffice.org/c/core/+/128371
https://gerrit.libreoffice.org/c/core/+/128370/2

eszkadev:
this fixes the dropdowns for listboxes (eg, categories listbox in the function wizard) or functions dropdown in formula bar

merttumer (approved):
works as expected

PR #3987, opened 2022-01-13
Fix vex dialogs that cannot be set to flex on mobile

pedropintosilva:
- About dialog was not centered anymore
- About dialog was missing close button
- Welcome dialog was not getting resized properly

These two dialogs need to be excluded from the changes introduced in
95bd65d2f658ca8c378b50f4ff09ef1424c15c13

About dialog on mobile is still the only case where we need to have the
the close button; The main vex cntainer of Welcome and About dialogs
cannot be set to flex since they do have a custom structure different
from the other ones.

Change-Id: Iadd9e866d33de1ae945f3383a289b7b0b7113337

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -289,6 +289,8 @@ div#w2ui-overlay-actionbar.w2ui-overlay{
 	border: none;
 	box-shadow: none;
 	border-radius: none;
+}
+.vex.vex-theme-plain:not(.vex-about-dialog, .vex-welcome-mobile) .vex-content {
 	display: flex;
 }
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
close button in about works as expected

pedropintosilva:
do not merge this yet, let me do it, I need to check something first

lpranam (approved):

pedropintosilva:
hm.. this fixes but  there are other cases and probably better to avoid targeting individual dialogs if possible _closing this_ as #3992 is more elegant at fixing these issues

PR #3988, opened 2022-01-13
private/pedro/rtl mobile n tablet arrows hamburger pencil

pedropintosilva:
- RTL: Fix Floating Action Button for mobile n tablet
- RTL: Mobile: flip or rotate main top navigation icons

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -250,7 +250,7 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 
 #mobile-edit-button-image {
 	position: relative;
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
awesome! this `html[dir='rtl']` does the job
I think we can convert my user list fix with that too (I used additional rtl class...) to simplify the code

pedropintosilva:
> awesome! this `html[dir='rtl']` does the job I think we can convert my user list fix with that too (I used additional rtl class...) to simplify the code

thanks for reviewing it. I will adjust those bits from user list.

PR #3989, opened 2022-01-13
browser: fixed broken show row/col after hiding

lpranam:
Change-Id: I141dd097e326c99021f5f85b95113e79c57cadd8

* Resolves: # 
* Target version: master 

### Summary
there can only be one isCurrent row/col
in case of multiple row/col selected,
original condition may fail as row/col maybe selected
but not the current.

added condition to also check if the row/col is selected

### Checklist


gokaysatir (approved):
Thanks for the fix :)

gokaysatir (approved):
Thanks again :)

PR #3993, opened 2022-01-14
calc: fix commentsPos field is empty string instead of empty array

merttumer:
as of co-2021 we use a new jsonwriter for json callbacks and it generates an empty array.
but in cp-6.4 we use boost for generating json and it cannot create an empty array
but instead it creates an empty string

The previous fix was taking advantage of it being an empty array

Change-Id: I0b4a4b39c3cff0b2f054765dbc47527633b56151

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix and the comment on the code :)

PR #3995, opened 2022-01-14
unit-wopi-async-slow: increase timeout

vmiklos:
When running this test against a debug core.git, then this test times
out for me. But if I increase the timeout, then it passes after 316s
just fine.

Change-Id: I1cd0d227dd9aac7279f991493b9aaa7c07468fdd


vmiklos:
@Ashod are you OK with this? The alternative would be to reduce the content in that monster 700 pages document, but I'm not sure if e.g. 350 pages is enough to not destroy test coverage. Thanks.

PR #3998, opened 2022-01-15
Unloading documents now make new clients retry after unloading is complete

Ashod:
This avoids all the corner-cases of 'unloading document recovery' by pushing back on the new connection while unloading, so that the client will have to retry again later. This seems to handle reloading cases handsomely and it also restores the older behavior anyway.

- wsd: dump the state of the sessions of DocBroker
- wsd: do not load a new view when unloading


kendy (approved):
Looks great to me - thank you!

PR #3999, opened 2022-01-15
Cleanup disconnecting sessions while recovering an unloading document

Ashod:
This still allows recovering from unloading, if a new client connects to the same document, but it does still seem to suffer corner cases. #3998 avoids all the corner-cases by pushing back on the new connection while unloading, so that the client will have to retry again later.

- wsd: dump the state of the sessions of DocBroker
- wsd: cleanup disconnecting sessions when a new session connects
- wsd: exit the Kit process when the last active session is disconnected

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1063,13 +1063,16 @@ private:
         if (viewCount == 1)
         {
 #if !MOBILEAPP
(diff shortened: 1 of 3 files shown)


Ashod:
Superseded by #3998.

PR #4000, opened 2022-01-15
Extend logging information from exceptions

stellarpower:
* Resolves: will open an issue if wanted later.
* Target version: master 

### Summary
I have some issue with our docker-based deployment of CODE, because no documents can be opened in my cloud, and I get a 500 error from the initial call to /cool.html. The output to the console is very chaotic, and the only error I can see mentions some issue parsing the config file and converting a value to a boolean - but no mention at all of what line or property is the issue. I have made minimal changes to the config; only what was required for a basic setup such that the server would be able to service incoming connections, so I don't see what the issue could be without some more error information. This PR should include more information for this specific error cause by outputting the

Overall, I would say the diagnostic information output by the server could be improved. There's lots of verbosity in general messages, and it's hard to separate real root-cause errors from the application from. Often, only a basic message and then the filename and line are output - whilst having a reference to the line is definitely helpful, as a primary means of finding out what is going on, it is not helpful. A settign to allow displaying of greater error details, possibly including a backtrace, to the browser would be helpful. My colleague told me that the browser-based reporting failed for us, and fromthe line in the sources it seems that not much more than the basic message is returned from this function. Whilst a potential security risk if left on when not desired, in a staging environment, having detailed information from a 50x page would really facilitate reslving issues with the setup.

I leave the discussion open as to if more helpful logging would be required, and can open an issue with the details from this if desired. For now, the "syntax error" output by the Poco exception linked to the XML config somewhere isn't sufficient to diagnose what the issue is, so I have had to make these changes in order to progress.

### TODO

- I've spent multiple hours waiting for gitpod to build the docker images, and it's still waiting, and the build script wanted me to change my sudoers file, which I'm not prepared to do except in a virtual machine, so I'm afraid I can't say I have built and tested this properly yet. The server itself builds in gitpod, but I can't replicate the scenario that caused the error in the first place without then deploying in our original environment, which requires the docker image that is not building for me.

### Checklist


mmeeks:
Hi there; looks fine to me! can you re-base it so it is a single commit and also git commit --amend -s # to sign off on the license. Not worried about security when we just send back data that was sent to us =)
Beyond that - Henry is working on having an 'Admin' bit sent in CheckFileInfo so we can provide more verbose diagnostics for administrators setting the server up.

Thanks for contributing !

stellarpower:
I will try, should just have been one commit but github wanted me to pull the upstream master branch. I'll try again. Feel free to LMK in the comments here about the CheckFileInfo, will happily test.

stellarpower:
I hate git, afraid I just had to open a new PR - please see #4033 :)

mmeeks:
Lets close this one we have a new duplicate - thanks!

PR #4002, opened 2022-01-17
clientsession_fuzzer: fix missing UnitWSD

vmiklos:
An alternative would be to tweak online-fuzz/wsd/DocumentBroker.cpp:534
to check for Util::isFuzzing(), but this is probably a better & more
generic way.

'./clientsession_fuzzer fuzzer/data/load' now works again.

Change-Id: I4d9fa387597695ff0802b268bc4d86be51dbabb2

### DIFF
--- a/fuzzer/ClientSession.cpp
+++ b/fuzzer/ClientSession.cpp
@@ -7,6 +7,7 @@
 bool DoInitialization()
 {
     COOLWSD::ChildRoot = "/fuzz/child-root";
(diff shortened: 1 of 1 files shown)


vmiklos:
@Ashod could you please review this? Without this, the fuzzer fails early before finding anything interesting. Thanks.

vmiklos:
Let's get this in, the fuzzer is not finding anything useful without this.

PR #4005, opened 2022-01-17
fix infinite scrolling

merttumer:
it is possible to scroll outside the document
with anything that uses onScrollBy function because
it does not respect the scroll area.

Change-Id: Ia8609311b8ead97f5866ff6300ddebb0d3f5b14d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
fixes the issue with infinite scrolling

PR #4009, opened 2022-01-17
private/pedro/rtl doc header

pedropintosilva:
- Document Header: remove unnecessary float
- Document Header: remove unnecessary CSS important
- RTL: Force Doc header border to respect direction


eszkadev (approved):
Looks good

PR #4010, opened 2022-01-17
Some additional fixes to Draw UI

timar:
* Draw and Impress UNO command names are defined in the same file, therefore
  we need to refer to 'presentation' when we want to extract a string.
  * So there is no such thing as 'drawing' in this context.
  * We need to use the _UNO() function to get the names instead of the
    raw command IDs when we build Draw's menus.
  * It is not necessary to define actions that simply send a UNO command,
    some code could be deleted.
* 'duplicatedrawpage' action id was not handled, therefore the menu did not 
  work in Draw. 'duplicatepage' is good enough.
* Impress use the term 'slide', Draw use the term 'page'. We need to make
  difference on the UI, cf. "Are you sure you want to delete page/slide?".
* Draw's Insert notebookbar tab was inherited from Impress, but it was not
  right because of 'slide' vs. 'page' difference, and also, Draw pages
  cannot have headers and footers, so the Header/footer button was removed
  in Draw.
* On Draw's notebookbar the 'Draw' tab was unlocalizable. It's not a product
  name there.

Change-Id: I54a7b14b83e48a706b15cac6bb538886ef009a49

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me - thanks Andras! nice catch.

Ashod (approved):
Thanks @timar, these have been long-standing inconsistencies in the uno commands in the menus.

PR #4012, opened 2022-01-17
browser: show busy label if cleaning session

hcvcastro:
Also, the ReconnectCount under normal condition
it should be 0, otherwise the document is cleaning the session.
..
Change-Id: I07b383a7810b8bc06e207fc600ebec99eaac9c77


gokaysatir (approved):
Thank you for the fix :)

PR #4013, opened 2022-01-17
Vex Replacement

hcvcastro:
- browser: add initial IDialog class
- browser: add initial idialog.css
- browser: insert content to the dialog
- browser: use statics methods
- browser: simplify isVisible
- browser: rename 'show' -> 'open'
- browser: create dialog buttons
- browser: fix button type
- browser: initial CSS to buttons
- browser: add 'afterOpen' callback
- browser: add 'close' function
- browser: add button 'click' event listener


pedropintosilva:
Thanks Henry! But I think we have steered away from creating a new control and instead we are using JSDialog https://github.com/CollaboraOnline/online/pull/5518 to do so

PR #4017, opened 2022-01-18
RTL: incorrect cell selection to mouse clicks...

dennisfrancis:
when we set ui language to ar or any other rtl language via url.

When language is RTL, since the commit:

RTL: Serve html with correct RTL settings
a3c5c71107a1eda32c39d11ba6ee1c3aaca29e6a

The dir attribute of body will be "rtl". The div "document-container"
itself occupies full width of the viewport but one of its children div
with id="map" has to "track" the document area of the canvas
element(canvas element occupies full width). The positioning of "map"
div is done in js assuming an "ltr" case. We could change all that based
on RTL flag, but at the moment a less riskier fix is to set the dir
attribute of "document-container" unconditionally to "ltr".

This also fixes all wrong svg based marker-selections too - for instance
shape/chart selections.

Comment cell hover popups were also affected by the same issue hence
fixed by this commit as well.

Change-Id: I5e20474d5b010976de12f7beee40a5c44df1b867

* Target version: master 

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -225,7 +225,7 @@ m4_ifelse(MOBILEAPP,[true],
         <div id="slide-sorter"></div>
         <div id="presentation-toolbar" style="display: none"></div>
       </div>
-      <div id="document-container" class="readonly">
+      <div id="document-container" class="readonly" dir="ltr">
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Looks good

PR #4023, opened 2022-01-18
Toolbars use var(--color-toolbar) and var(--color-border)

Andreas-Kainz:
You can simple modify the background colors of the toolbars
for now NB and bottom toolbar
You can edit the borders and separators with --color-border
so it's very easy to remove focus for separators

Change-Id: I661513031eb231d6ffc5036df38381d7fdcac5a7

result look the same than before, 
but the values names are easier to understood 
and changes like in https://github.com/nextcloud/richdocuments/issues/1976
are more simple.

### DIFF
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -12,5 +12,8 @@
 	--gray-color: #b6b6b6;
 	--white-bg-color: #fff;
 
+	--color-main-background: #fff;
+	--color-border: #b6b6b6;
+
 	--border-radius: 4px;
 }
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -218,7 +218,7 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	width: 100%;
 	border-top: none;
 	z-index: 11 !important;
-	border-bottom: 1px solid var(--gray-color);
+	border-bottom: 1px solid var(--color-border);
 }
 
 #toolbar-logo {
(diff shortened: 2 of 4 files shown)


browser/css/color-palette.css
```diff
@@ -12,5 +12,8 @@
 	--gray-color: #b6b6b6;
 	--white-bg-color: #fff;
 
+	--color-toolbar: #fff;
```
pedropintosilva, line 15: I think it makes total sense to have `--color-border` but I would really avoid starting to have `--color-component-x` because that, I'm afraid, will lead to multiple variables with the same value but different naming 

In terms of naming conventions, I think better would be to have anything between these categories:
- Main colors:
  -  Primary, (and if needed secondary, branding etc)
- Error | Success  | Warning
- Attribute:
  - Text
  - Background
  - Border
  - Radius
  - Shadow

pedropintosilva (changes requested):

browser/css/color-palette.css
Andreas-Kainz: That's fine.  

So in this case `--color-toolbar` will be `--color-background-primary` ? cause `--color-background` will be the color of the document and the tabbs?

Andreas-Kainz: What is with select, hover. Should they be part of the name convention? `--color-select` and `--color-hover` ?

Andreas-Kainz: I use `--color-main-background` which was used also at NC for #fff

pedropintosilva: thanks, merging. Ideally we would have something like --color-secondary or --color-secondary-primary-element

PR #4024, opened 2022-01-19
NB Writer Stylesview show only images and text as tooltip

Andreas-Kainz:
In Writer with notebookbar toolbar there is the option to show
previews of writer styles within the stylesview element.

Now there is an img with the preview AND the text name of the style
with this patch the text was removed and shown as tooltip
as every other element in the NB has an tooltip

the benefit is that you see ONLY an preview of Default Paragr
but not in addition the name which is simple a duplicate

after the patch you see more previews and it's simplier to select

Change-Id: Iaa13ecde7f6c515742b2c0b303dde4b7d97e135c

image

recommendation come from https://github.com/nextcloud/richdocuments/issues/1976


Andreas-Kainz:
after the patch was in master, I will work on use var(--colors) instead of the hard coded colors.

browser/css/notebookbar.css
```diff
@@ -453,27 +453,15 @@
...
 	width: 530px;
 	overflow-x: hidden;
-	padding: 5px;
-	padding: 0px;
+	padding: 2px;
 	border: 1px solid var(--gray-color);
+	border-radius: var(--border-radius);
 	border-collapse: separate; /*To use box-shadow in Internet Explorer 9 or later*/
-	-webkit-box-shadow: 0 0 3px var(--gray-color);
-	box-shadow: inset 0 0 3px var(--gray-color);
```
pedropintosilva, line 461: This seems unrelated to the commit and related to how the main parent looks. Could you please cut this from the current commit and add another one just with that? Let me know if you need help with that

pedropintosilva (changes requested):

browser/css/notebookbar.css
Andreas-Kainz: sure will make a separate PR with how select, hover, ... should look like.

pedropintosilva: What I said it's still true: This chunk of changes seems unrelated to the commit and related to how the main parent element looks. Could you please:
a) removing that from the current commit and add another commit just with that? 
b) Or if it's difficult for you to do that, no worries, just then please add amend the commit and add in the description that this commit also changes the parent elements' look n feel (`#stylesview`)

pedropintosilva (changes requested):
https://github.com/CollaboraOnline/online/pull/4024#discussion_r790171908

Andreas-Kainz:
I have to submit a new PR cause I think I don't know what I did.

PR #4031, opened 2022-01-20
fix selectedFont is not shown on sidebar

merttumer:
if the font is not includede in the core and already exists
in the document, we need to display that as menubar's fontnamebox
does

Change-Id: Ia725fb5d031aca9382f0247772bdc9e7bbe95729

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected

PR #4033, opened 2022-01-20
Add more detail to error logs

stellarpower:
Changes as requested in #4000.


mmeeks:
No problem creating another commit - I've queued CI goodness, which is more of a test of CI than your patch I expect =)

wsd/FileServer.cpp
```diff
@@ -722,9 +722,9 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
     }
     catch (Poco::SyntaxException& exc)
     {
-        LOG_ERR("Incorrect config value: " << exc.displayText());
+		LOG_ERR("Incorrect config value: " << exc.displayText() << " :: " << exc.extendedMessage());
```
pedropintosilva, line 725: 2 Tabs were used in this line could you please swap that for spaces?

wsd/FileServer.cpp
```diff
@@ -722,9 +722,9 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
...
         sendError(500, request, socket, "500 - Internal Server Error!",
-                  "Cannot process the request");
+				  "Cannot process the request - " + exc.displayText());
```
pedropintosilva, line 727: In this line a combination of tabs and spaces were used, could please use spaces?

pedropintosilva (changes requested):
Please make sure that your text editor/IDE is showing those invisible characters

pedropintosilva:
Also another tip: in order to see all the tests passing you need get your change id. ( we use this to identify commits and it also simplifies when we are cherry-picking commit or trying to find regressions etc)

How to obtain Change-Id for your commit message?

1. `cp .git-hooks/commit-msg .git/hooks/commit-msg` Installs (by copying that file) a commit hook
2. `git commit --amend` Amends your last commit and it will add the missing change-id
3. `git push -f `

stellarpower:
> Please make sure that your text editor/IDE is showing those invisible characters

It is, I'm using Kate so it shows tabs as chevrons, I don't know why linting continually messes up as soon as I touch VCS. It doesn't look like it's gone through even on the second time.

> How to obtain Change-Id for your commit message?

Done, please let me know if that's showing okay to you.

mmeeks:
Ah - this 'extendedMessage' is actually a setter that just appends ": " + arg - to the _msg member; so 'disaplyText' will print that already - so no need for the first line of change (which fails to compile for this reason) =)

Tweak that and we should be done - and sorry it's take so long to go around [!]

stellarpower:
Thanks for that - don't understand what was going on in gitpod then. It did build and run it and I typed in the modifications, but it was always the docker build that just hung for hours, and that's what I need to test in my setup so I never manged to check the build that way.

So can we get no more information from the Poco exception? I was hoping it might contain a line number or something in an extended message. Displaying to browser is a useful conveniece, but with the error from the log it still didn't clarify what was going on - just syntax error, no line number, no context for what XML node it was parsing. Part of the problem it seems is that on firing up the server it's not validating the entirety of the config - we were able to bring it up and connect to our storage cloud just using the default config; but then to test opening documents, we need to modify (only slightly) the configuration so that it contains meaningful values - hostname, TLS settings, etc., which is then the point it started complaining.

stellarpower:
Sorry, missed this re notifications! I think rethrowing sounds agood idea, yes. I'm hardly a great reference on managing a large codebase and how best to manage the exception framework, but, often in my code I am capturing something, adding additional info, and then unless I have a specific workaround, there's not much mnore I can do than rethrow it - in this case, if the configs can't be parsed properly (or are semantically invalid), I'd say that's justification for a fatal error and aborting, it's not like we can just ignore and retry. But I agree, the key difference between having it propogate all the way up and rethrowing from a target site is the useful information we can tag onto it. There's an art to it. I reckon it's always best to fail soon rather than late (looking at you, bash, where a single space on line 5 can only show up as a real problem an hour later in line 478 with some subtly rubbish output), and at the same time we wanna interject once we are in user code at the highest level where it makes sense to the domain why we failed. I.e., "server couldn't start because the config file does not match the schema" not "assertion failed: false != true" (now looking at you, OpenCV).

I know nothing about Poco, but, would this actually be better seen as an upstream issue? Could try opening a bug there about adding more detail to their exceptions.

PR #4034, opened 2022-01-20
Admin: force un-conditional logging when docstats enabled

yunusem:
Change-Id: Iea0a75927b495735787496468ab39d5314992a50

* Target version: master 

### Checklist

### DIFF
--- a/wsd/AdminModel.cpp
+++ b/wsd/AdminModel.cpp
@@ -553,15 +553,17 @@ void AdminModel::addDocument(const std::string& docKey, pid_t pid,
     oss << memoryAllocated << ' ' << wopiHost;
     if (COOLWSD::getConfigValue<bool>("logging.docstats", false))
     {
-        Log::StreamLogger logger = Log::trace();
-        if (logger.enabled())
-        {
-            logger << "docstats : adding a document : " << filename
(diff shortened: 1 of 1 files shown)


mmeeks (changes requested):
Oh - hmm; we don't really want a Log::warn at all - we want to un-conditionally write to the channel so:

on the Poco Logger call:

	channel = Channel::Ptr getChannel() const;
then do:
        channel->log();

cf. Log.cpp(initialize) which creates the channel.

That should bypass the priority fluff and just print thus:

	_str << msg.getText() << std::endl;

Clearly this stuff is not a warning either.

yunusem:
> Oh - hmm; we don't really want a Log::warn at all - we want to un-conditionally write to the channel so:
> 
> on the Poco Logger call:
> 
> ```
> channel = Channel::Ptr getChannel() const;
> ```
> 
> then do: channel->log();
> 
> cf. Log.cpp(initialize) which creates the channel.
> 
> That should bypass the priority fluff and just print thus:
> 
> ```
> _str << msg.getText() << std::endl;
> ```
> 
> Clearly this stuff is not a warning either.

I agree, I can do better with this awesome guide.

mmeeks (approved):
Hmm - timestamp is prolly not that useful - I imagine that there are timestamps in any journal we log to.

PR #4036, opened 2022-01-21
make sure modifiedTimeUs calculates with enough precision

grandinj:
otherwise, on some platforms, the calculation might occur with 32-bits, but we really need 64 bits for microseconds.
Also tweak the return type to make sure we use 64 bits.

Signed-off-by: Noel Grandin <noel.grandin@collabora.co.uk>
Change-Id: Ie8b8a49938c7ac2cfe7c5fc1dd6629e6ea347115

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/FileUtil.hpp
+++ b/common/FileUtil.hpp
@@ -169,9 +169,10 @@ namespace FileUtil
         }
 
         /// Returns the modified unix-time in microseconds since epoch.
-        std::size_t modifiedTimeUs() const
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Thanks @grandinj!

pedropintosilva:
Thanks @grandinj and congrats on your first Online (via GH) PR :partying_face:

PR #4037, opened 2022-01-21
Admin: alternate std::localtime

yunusem:
I saw similar implementations in Util.cpp

Change-Id: I3927ef9989de2217dee6c4b3034c9ffbd8fb184d

* Target version: master 

### Checklist

### DIFF
--- a/wsd/AdminModel.cpp
+++ b/wsd/AdminModel.cpp
@@ -555,10 +555,13 @@ void AdminModel::addDocument(const std::string& docKey, pid_t pid,
     {
         std::ostringstream osst;
         Poco::AutoPtr<Poco::Channel> channel = Log::logger().getChannel();
-        const auto now = std::chrono::system_clock::now();
-        const auto current = std::chrono::system_clock::to_time_t(now);
-        osst << std::put_time(std::localtime(&current), "%F %T")
-             << " docstats : adding a document : " << filename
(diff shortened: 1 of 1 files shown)


wsd/AdminModel.cpp
```diff
@@ -555,10 +555,12 @@ void AdminModel::addDocument(const std::string& docKey, pid_t pid,
...
         std::ostringstream osst;
         Poco::AutoPtr<Poco::Channel> channel = Log::logger().getChannel();
-        const auto now = std::chrono::system_clock::now();
-        const auto current = std::chrono::system_clock::to_time_t(now);
-        osst << std::put_time(std::localtime(&current), "%F %T")
-             << " docstats : adding a document : " << filename
+        char output[64];
+        std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
+        std::time_t current = std::chrono::system_clock::to_time_t(now);
+        std::tm tm = *std::gmtime(&current);
```
eszkadev, line 561: std::gmtime Return value:
Pointer to a static internal std::tm object on success, or null pointer otherwise. The structure may be shared between std::gmtime, std::localtime, and std::ctime and may be overwritten on each invocation.

This function may not be thread-safe. 
[https://en.cppreference.com/w/cpp/chrono/c/gmtime]

PR #4040, opened 2022-01-21
wsd: Adjust log levels

thebearon:
A non-writable path isn't an error in itself.

The reason why the service tried to write the systemplate,
and the perf implications is worth reporting even on default
log level.

Change-Id: Ie168875ceb4e501291676a5aa4ddc21b88c5761f

* Target version: master


Ashod (approved):
Looks good. Thanks @thebearon!

PR #4048, opened 2022-01-24
tools: fix -Werror,-Wfortify-source

vmiklos:
'sscanf' may overflow; destination buffer in 'smap_key' has size 20, but
the corresponding specifier in tools/map.cpp:570 may require size 21.

Change-Id: Ie7356db924193777869a39b882512a864c0e10c5


vmiklos:
'Code scanning results / CodeQL' failure looks not relevant.

PR #4053, opened 2022-01-25
NB Writer Stylesview show only preview and no label

Andreas-Kainz:
In writer with notebookbar toolbar there is an option to show
previews of writer styles within the stylesview element.

Now there is an img with the preview AND the text name of the style
with this patch the text was removed and shown as tooltip

The benefit is that you see ONLY an preview of Default Paragr
but not in addition the name which is simple a duplicate

after the patch you see more previews and it's simplier to select

Change-Id: I18605af2daf3c47aee9902164060f530e71fad60

image

recommendation come from https://github.com/nextcloud/richdocuments/issues/1976

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -462,8 +462,6 @@
 
 #stylesview .ui-iconview-entry {
 	width: 30%;
-	height: 60px;
-	padding: 2px;
 	box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
compare to #4024 this is ONLY the remove of the text and add it as tooltip.

pedropintosilva:
super!

PR #4056, opened 2022-01-25
Add more detail to error logs.

mmeeks:
We get a syntax error message in the logs - but have no indication as
 to what in the config file caused this issue. Bringing back both more
 detail to the browser and logging the extended details.

Signed-off-by: stellarpower <5004545+stellarpower@users.noreply.github.com>
Change-Id: Ib8ddb20a1968f879558e59f50579ee4b18db0f09

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -724,7 +724,7 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
     {
         LOG_ERR("Incorrect config value: " << exc.displayText());
         sendError(500, request, socket, "500 - Internal Server Error!",
-                  "Cannot process the request");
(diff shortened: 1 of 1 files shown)


stellarpower:
Actually whilst at this line, I just noticed we seem to catch by non-const reference, which I think would probably be better as a const&, if you agree(?)

mmeeks:
Yes - good point - it would be nice to have a follow-on patch to fix the (several) cases where we don't catch by const reference there.
If you do:
$ git grep catch | grep -v '\.\.\.' | grep -v const | grep '\.cpp'
can you fix those up as a follow-on commit - I'd really like to get your first fix in as-is here =)

Thanks !

PR #4057, opened 2022-01-26
Use variables for text colors

Andreas-Kainz:
there are 3 color var's added

var(--color-main-text)
this is the default text color EVERYWHERE

var(--color-text-accent)
will be used when a label was selected, or during hover
also for the filename accent text color was used

var(--color-text-lighter)
was use for saved date or when a label was disabled

PR was tested for classic and notebookbar
for the sidebar and dropdown elements.

Change-Id: I21795875e58b956081f2020422e77ccecd9bfba3

image

### DIFF
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -12,8 +12,12 @@
 	--gray-color: #b6b6b6;
 	--white-bg-color: #fff;
 
-	--color-main-background: #fff;
-	--color-border: #b6b6b6;
+	--color-main-text: #696969;
+	--color-text-accent: #333333;
+	--color-text-lighter: #636363;
 
+	--color-main-background: #fff;
+
+	--color-border: #b6b6b6;
 	--border-radius: 4px;
 }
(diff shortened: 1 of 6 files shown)


Andreas-Kainz:
to use `--color-text-accent` when hover to a section, the command has sometimes been added, cause it wasn't available before, but I didn't add any other stuff than text color stuff.

Andreas-Kainz:
If I miss some text color stuff, I will make separate PR's. I didn't change colors only cause there was `color:` defined within the css files. I want to know where the label was used, so I could check if it work well.  Check was done with exotic colors so I quickly see if the color was correct changed.

pedropintosilva:
There were color to CSS vars changes mixed with actual color/design changes. I have split all changes in two commits

Andreas-Kainz:
> There were color to CSS vars changes mixed with actual color/design changes. I have split all changes in two commits

thanks. wait for +1.

pedropintosilva (approved):

PR #4059, opened 2022-01-26
calc: Fix new comments do not check tabid on insert

merttumer:
it only checks the cellPos and this can cause
modifying the existing comment when we try to add
a new one if the same cell has a comment in another
tab already.

Change-Id: I69fd607b487fe18c9a48420b7df202b7d3db311d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix :)

PR #4064, opened 2022-01-26
mount /dev via devfs on FreeBSD

arrowd:
* Target version: master 

### Summary

COOL uses `mknod()` system call to create `/dev/random` and `/dev/urandom` devices. On FreeBSD this does create a special file, but it doesn't function as proper random device. The correct way to do this is to mount a special filesystem called `devfs`.

This patch teaches `coolmount` to perform `devfs` mounts and changes COOLWSD to use that when running on FreeBSD. It also fixes the cleanup procedure to not leave dangling `devfs` mounts after the daemon shutdown.

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -125,6 +125,9 @@ void removeJail(const std::string& root)
 
     // Unmount the tmp directory. Don't care if we fail.
     const std::string tmpPath = Poco::Path(root, "tmp").toString();
+#ifdef __FreeBSD__
+    unmount(tmpPath + "/dev");
+#endif
     FileUtil::removeFile(tmpPath, true); // Delete tmp contents with prejudice.
     unmount(tmpPath);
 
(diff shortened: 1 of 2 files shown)


common/JailUtil.cpp
```diff
@@ -168,12 +171,17 @@ void cleanupJails(const std::string& root)
         for (const auto& jail : jails)
         {
             const Poco::Path path(root, jail);
+            // Delete "tmp" directory after cleaning all the jails
+            if (jail == "tmp")
+                continue;
             // Delete tmp and link cache with prejudice.
```
timar, line 177: Please amend the comment.

timar (changes requested):

common/JailUtil.cpp
arrowd: "Amend" in a sense "explain why it is needed"?

arrowd: Is it better now?

timar (approved):

PR #4073, opened 2022-01-27
Classic Toolbar Menubar no juming up at hover menu element

Andreas-Kainz:
If you hover at the classic toolbar at the menubar,
a border was shown and the string jump at 1px size.

Change-Id: I5d03a380c2bfa34bd169bb12295d535875e54a5f

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -123,9 +123,7 @@
 .lo-menu > li > a, .lo-menu > li > a.has-submenu {
 	padding: 8px;
 	z-index: 400;
-	border-left: 1px solid transparent;
-	border-right: 1px solid transparent;
-	border-top: 1px solid transparent;
+	border: 1px solid transparent;
 }
 
 .lo-menu a, .lo-menu a:hover, .lo-menu a:focus, .lo-menu a:active, .lo-menu a.highlighted {


pedropintosilva:
@Andreas-Kainz : Thanks for spotting that! I wonder if wouldn't be better and simpler to just delete that border

browser/css/menubar.css
```diff
@@ -123,9 +123,7 @@
 .lo-menu > li > a, .lo-menu > li > a.has-submenu {
 	padding: 8px;
 	z-index: 400;
-	border-left: 1px solid transparent;
-	border-right: 1px solid transparent;
-	border-top: 1px solid transparent;
+	border: 1px solid transparent;
```
pedropintosilva, line 126: in here

pedropintosilva:
so we would have: 

https://user-images.githubusercontent.com/65948705/151519748-0bd4e33a-954f-4ad3-af95-d2cb1b8156fb.mp4

pedropintosilva:
the border would still be there on hover due to 
```css
.lo-menu > li > a:hover, .lo-menu > li > a:focus, .lo-menu > li > a:active, .lo-menu > li > a.highlighted {
	background: #fff;
	color: var(--color-text-accent);
	border-color: #bbbbbb;
	border-bottom: 1px solid #bbb;
}
```
(we should probably delete that extra `border-color: #bbbbbb;`  since the border shorthand property already has that)

Andreas-Kainz:
Yes I plan to cleanup the css, but I would make separate PR's so that it's only one thing. First bugfix, second cleanup.

However I also had in mind to remove the border stuff, BUT if we have borders and hover in the css files via var(--border-color) we can easy define the same color than the fill color and we can define transparent so it's more flexible to have the border stuff added.

Andreas-Kainz:
So if you ask me, I would make a separate cleanup PR

PR #4077, opened 2022-01-27
Detect storage conflict with no clients

Ashod:
- wsd: wait for the modified flag before giving up 
- wsd: improved removeSession
- wsd: do not force uploading when in conflict
- wsd: detect storage conflict with no clients
- wsd: reuse autoSaveAndStop for consistentcy
- wsd: merge unloading logic and centralize
- wsd: test: rework and extend UnitWOPIDocumentConflict
- wsd: test: add discarding after closing test to DocumentConflict
- wsd: test: add SaveOnExit test

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1832,7 +1832,7 @@ bool DocumentBroker::autoSave(const bool force, const bool dontSaveIfUnmodified)
 
 void DocumentBroker::autoSaveAndStop(const std::string& reason)
 {
-    LOG_TRC("autoSaveAndStop for docKey [" << getDocKey() << ']');
+    LOG_TRC("autoSaveAndStop for docKey [" << getDocKey() << "]: " << reason);
 
     if (_saveManager.isSaving() || isAsyncUploading())
     {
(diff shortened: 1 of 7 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -2148,8 +2147,11 @@ std::size_t DocumentBroker::removeSession(const std::string& id)
                 LOG_DBG("No more sessions after removing last. Setting MarkToDestroy flag.");
             }
         }
-        else
-            assert(!_docState.isMarkedToDestroy());
+        else if (activeSessionCount > 0)
+        {
+            assert(!_docState.isMarkedToDestroy() &&
```
kendy, line 2170: Would be good to ERR this into the log too, even in the non-debug case (ie. asserts compiled out), so that we have a sign of something very bad happening.

wsd/DocumentBroker.cpp
```diff
@@ -1677,9 +1677,22 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
...
             = isPossiblyModified() ? "error: cmd=storage kind=documentconflict" : "close: documentconflict";
 
-        broadcastMessage(message);
+        const std::size_t activeClients = broadcastMessage(message);
         broadcastSaveResult(false, "Conflict: Document changed in storage",
                             uploadResult.getReason());
+        if (activeClients == 0)
+        {
+            // No clients were contacted; we will never resolve this conflict.
+            stop("conflict");
```
kendy, line 1676: As we talked - let's trigger a save-as in a follow-up PR please :-)

kendy (approved):
Let's get it in - thank you @Ashod ! :-)

PR #4082, opened 2022-01-27
NB add SelectAll command to Layout tab

Andreas-Kainz:
In the layout tab there are the object related stuff like
align and front/back. So I add there SelectAll command.

image

Change-Id: I4cdd9fbb0351cede9cd78dd1edd50e0612d895f0

see https://github.com/nextcloud/richdocuments/issues/1976 moved command from the hamburger menu to the tab.


pedropintosilva:
Cool! I noticed that both on Draw and Impress the tab changes upon clicking that select all bigtoolitem

pedropintosilva (approved):
merging, we can deal with this in another PR ^

PR #4088, opened 2022-01-31
private/pedro/nb statusbar toggle

pedropintosilva:
- Notebookbar: Fix statusbar and ruler labels
- Notebookbar: Add statusbar toggle to layout tab
- Notebookbar: Set statusbar toggle to react to visibility

PR #4091, opened 2022-02-01
Notebookbar: Set statusbar toggle's status on loading

pedropintosilva:
Make sure the toggle also reacts not only to the user interaction¹ but
also at the startup time

1 Which is already done with:
- e384c561798e7616f10e763788e92acec6caff9c
 - from PR: https://github.com/CollaboraOnline/online/pull/4088/

Change-Id: I29e401eede4fc11c821da91b66430ba99d1ec1aa


joergmschulz:
I think for impress it still needs to be done.

PR #4092, opened 2022-02-01
Do not request unnecessary image on selection

eszkadev:
Avoid doubled PNG generation when graphic is selected.
We only need 'rendershapeselection' request to get preview.
'gettextselection' will only setup '_selectionContent' in
the Clipboard code but it will be never used as
`_getHtmlForClipboard` will ignore it for complex selection
with graphic.


merttumer (approved):
I just tested it and I can confirm that it is much much better now than before with the core patch

Issue #3772, opened 2021-12-01
css-variables


Andreas-Kainz:
we have an `color-palette.css` file with specific colors. As the var list isn't very long and also it's not really well distinguish which var is for which thing, I suggest to update the var names, extend them AND use them everywhere.

As cool didn't has something like that I suggest to have a look at https://github.com/nextcloud/server/blob/master/core/css/css-variables.scss and use the var names or similar ones.

Andreas-Kainz:
@pedropintosilva I would like to do the job, but first I need your go for it.

Andreas-Kainz:
the toolbar use `w2ui-1.5.rc1.css` styles which has hardcoded colors, however should I add the values inside the w2ui-.css file or in a more generic `toolbar.css` file? 

I think `toolbar.css` will be the better idea, when w2ui... has been updated.

pedropintosilva:
> @pedropintosilva I would like to do the job, but first I need your go for it.

Go for it !

pedropintosilva:
> the toolbar use `w2ui-1.5.rc1.css` styles which has hardcoded colors, however should I add the values inside the w2ui-.css file or in a more generic `toolbar.css` file?
> 
> I think `toolbar.css` will be the better idea, when w2ui... has been updated.

Anything that is 3rd party library we were kinda avoiding to touch too much so for obvious reasons. However and if it's just color change maybe that's ok

PR #4094, opened 2022-02-01
Color-Palette values get a prefix -co- for there variables

Andreas-Kainz:
Change-Id: I44e700dd526b4979983208bbfad73a2b9892e224

* Resolves: #3772


pedropintosilva:
sorry but the prefx `--co` is reserved for branding only so we and people can safely customize it without braking the main CSS

Andreas-Kainz:
Perfect good to know

PR #4097, opened 2022-02-01
Notebookbar: Set ruler toggle to react to visibility

pedropintosilva:
- Make sure a trigger is fired no matter the UI mode
	- Handle it (do anything) just from withing the NB widget
	(get proper css classes)

Note: similar to what we did with statusbar in:
e384c561798e7616f10e763788e92acec6caff9c

Change-Id: I6c8a38f2e426f20cc5b7a06241f7a92fa4d938e4


pedropintosilva:
a follow up PR would be to also ensure that the toggle reacts right at the loading time as we did with statusbar control in 4b25653860a1df4271110b64b3b02ebb2c249b3b . However it seems that the ruler control is not so similar to the statusbar control :thinking:

PR #4100, opened 2022-02-02
sidebar: Don't show sidebar on pdf regardless of the permisson

merttumer:
To be safe, we can ignore the idea of this.map.isPermissionReadOnly()
control would always return true for PDF views and add
additional check for app.file.fileBasedView which is always
true and readonly true for PDF view

Change-Id: I8f0a4401087c735508534404ee987bb35c3bb0fd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks good

PR #4102, opened 2022-02-02
fixed master slide previews not being updated

lpranam:
after switching to master view previews would remain unchanged
this patch will invalidate tiles for all the previews

Change-Id: Ie7320d625bd95fe80fc76a459b576d6148b39276

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -523,18 +523,6 @@ L.Control.PartsPreview = L.Control.extend({
 		}
 	},
 
-	_showMasterSlides: function() {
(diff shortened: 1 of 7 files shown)


mmeeks (commented):
Looks like a potentially very expensive change to me. How often do we get EMPTY invalidations?

lpranam:
WE get empty invalidation only when we add/remove pages or change to master view or vice versa. 

Though I am working on a better solution that would solve this problem from the root.

mmeeks (commented):
Looks better =) I expect though that we want to avoid duplicate updates of the state changes here. I expect we want https://gerrit.libreoffice.org/c/core/+/128578 merged.

PR #4107, opened 2022-02-02
NB View Tab group commands according to LibO

Andreas-Kainz:
First Group app specific commands
Second Group Zoom commands
Third Group visibility toggle

Change-Id: I822d1ea8ef3460bea9869235b039749dad028c17

image

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -351,7 +351,7 @@
 	margin-inline-end: 5px !important;
 	text-align: center;
 }
-.unotoolbutton.notebookbar.has-label > *{ /*so they stay side by side*/
+#zoomreset, .unotoolbutton.notebookbar.has-label > *{ /*so they stay side by side*/
 	display: table-cell;
 }
 
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -847,17 +847,10 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'command': '.uno:FullScreen'
 			},
 			{
-				'id': 'showstatusbar',
-				'type': 'menubartoolitem',
-				'text': _('Toggle Status Bar'),
-				'command': _('Show Status Bar')
-			},
-			{
 				'id': 'zoomreset',
 				'type': 'menubartoolitem',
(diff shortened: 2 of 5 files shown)


Andreas-Kainz:
As there is space at the toggle status bar, would we add there an command to collapse the notebookbar @pedropintosilva ?

pedropintosilva:
> As there is space at the toggle status bar, would we add there an command to collapse the notebookbar @pedropintosilva ?

Good idea! Yes makes sense.

pedropintosilva:
> First Group app specific commands Second Group Zoom commands Third Group visibility toggle
> 
> Signed-off-by: andreas kainz [kainz.a@gmail.com](mailto:kainz.a@gmail.com) Change-Id: I822d1ea8ef3460bea9869235b039749dad028c17
> 
> image

Sounds good but it seems there are two problems with the zoom commands:
1. the reset zoom is using the wrong icon, shouldn't be [this one](https://github.com/CollaboraOnline/online/blob/master/browser/images/lc_zoomreset.svg)
2. The order of those zoom icons. I would prefer to keep the order as in the statusbar: 
<table>
<tr><td rowspan="2">Reset Zoom</td><td colpan="2">Zoom Out</td></tr>
<tr><td >Zoom In</td></tr></table>

Andreas-Kainz:
Zoom reset didn't work as bigtoolitem that's the reason for the updated layout.

Updated icon, yes I also had this in mind. Will be a separate PR cause the icon is the same than before.

Andreas-Kainz:
I could fix all your _very useful_ comments @pedropintosilva 

image

browser/css/notebookbar.css
```diff
@@ -351,7 +351,7 @@
 	margin-inline-end: 5px !important;
 	text-align: center;
 }
-.unotoolbutton.notebookbar.has-label > *{ /*so they stay side by side*/
+#zoomreset, .unotoolbutton.notebookbar.has-label > *{ /*so they stay side by side*/
```
pedropintosilva, line 354: clever way of doing it : ) without additional JS. Note for myself: Then maybe in the feature we can somehow improve the bigtoolitem or add an additional parameter to apply that while not applying anything else . I need to think about it :thinking:

pedropintosilva (approved):
Perfect!!

PR #4114, opened 2022-02-03
Image operations fixes (co-6-4)

eszkadev:
- fixes error sometimes shown in the console
- fixes preview update - before it was leaving empty rectangle when new preview arrived during rotation

(for more info how to reproduce see the commit message)


merttumer (approved):
works as expected

PR #4116, opened 2022-02-04
color-palette use var --color-border-*

Andreas-Kainz:
add a description to color-palette to define
which --color-border-* color should be used.

replace hardcoded colors with the equivalent
--color-border-* variable

Change-Id: I7be11904eee2fe7f9330a87915cdb71e6a9af3ba

### DIFF
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -30,9 +30,9 @@
 	--color-primary-lighter: #63BBEE;
 
 	--color-border: #b6b6b6;
-	--color-border-dark: #EFEFEF;
-	--color-border-darker: #b6b6b6;
-	--color-border-lighter: #f1f1f1;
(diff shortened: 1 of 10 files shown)


Andreas-Kainz:
sure there are a few hardcoded border colors, but I change the color ONLY when I know where the color was used. So each changed value was checked in the UI.

pedropintosilva:
Hover status somehow got affected notebookbar > buttons

https://user-images.githubusercontent.com/65948705/152501311-8037803a-97ff-4a85-86a6-c6c926bbe514.mp4

Andreas-Kainz:
Thanks for thre Review Pedro, sorry you are right, will check what's going wrong.

Andreas-Kainz:
I think it's not reviewable so I make a #4119 and hope that it's easier to review.

PR #4122, opened 2022-02-07
wsd: preserve the original modified time as string

Ashod:
We no store the origina modified time of the document
as we receive from the storage server in string
form and send it back as-is. This avoids any
potential issues with the roundtrip of conversion
to and from a timestamp.

Change-Id: I524bea8f36c3ce62dcd00c4fe6a1e7e083287ed1

### DIFF
--- a/wsd/DocumentBroker.hpp
+++ b/wsd/DocumentBroker.hpp
@@ -670,12 +670,6 @@ private:
         bool lastRequestSuccessful() const { return _lastRequestSuccessful; }
 
 
-        /// Set the modified time of the document.
-        void setModifiedTime(std::chrono::system_clock::time_point time) { _modifiedTime = time; }
-
-        /// Returns the modified time of the document.
-        std::chrono::system_clock::time_point getModifiedTime() const { return _modifiedTime; }
-
 
         /// Helper to get the current time.
         static std::chrono::steady_clock::time_point now()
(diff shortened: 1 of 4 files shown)


wsd/DocumentBroker.hpp
```diff
@@ -775,16 +769,10 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
         }
 
         /// Set the last modified time of the document.
-        void setLastModifiedTime(std::chrono::system_clock::time_point time)
-        {
-            _request.setModifiedTime(time);
-        }
+        void setLastModifiedTime(std::chrono::system_clock::time_point time) { _modifiedTime = time; }
```
kendy, line 772: Now we have a setLastModifiedTime here that works with the time_point, and the other that works with a string.  I wonder - could or should we rename one of these, for better understanding?

But I guess that's not blocking the integration - so feel free to update in a follow-up PR if you agree with the above :-)

kendy (approved):
Looks great to me in general, thank you!

wsd/DocumentBroker.hpp
Ashod: Makes sense. Thanks for the review!

Ashod: Done, waiting for the build.

PR #4124, opened 2022-02-07
mobile: show go back btn before docload

merttumer:
on ios case, there is no device back button available
so it is impossible to go back before the document load

Change-Id: Ifce5e4777b8dd014cbdc527fb65f0d0be21f2cd2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.Toolbar.js
```diff
@@ -1057,8 +1057,12 @@ function onUpdatePermission(e) {
 			}
 		}
 		if (e.perm === 'edit') {
+			$('#toolbar-mobile-back').removeClass('editmode-off');
+			$('#toolbar-mobile-back').addClass('editmode-on');
 			toolbar.set('closemobile', {img: 'editmode'});
```
eszkadev, line 1062: this line will not work anymore I think

eszkadev, line 1062: this line will not work anymore I think

browser/src/control/Control.UIManager.js
```diff
@@ -42,13 +42,20 @@ L.Control.UIManager = L.Control.extend({
...
+		var that = this;
 
 		if (window.mode.isMobile() || !enableNotebookbar) {
 			var menubar = L.control.menubar();
 			this.map.menubar = menubar;
 			this.map.addControl(menubar);
 		}
 
+		if (window.mode.isMobile()) {
+			$('#toolbar-mobile-back').on('click', function() {
```
eszkadev, line 54: it would be good later move this to Control.MobileTopBar.js file to keep related things together. you can access uiManager by: `map.uiManager`

eszkadev (approved):

browser/src/control/Control.UIManager.js
merttumer: I will, thank you :)

merttumer: I just checked that Control.MobileTopBar.js is initialized after docloaded event so we cannot move the logic there. The idea is to have this before docloaded

eszkadev: ah, ok

PR #4128, opened 2022-02-07
Mobile: Fix vex dialogs that still need vex-close

merttumer:
context: with 95bd65d all vex dialogs on mobile have vex-close hidden

About dialog on mobile is still the only case where we need to have the
the close button:
	- Add generic CSS class that can be later re-used if needed for other
		cases

Change-Id: Ibf13350c892a77c87ffa2c9a67dcbfe02a3e20f6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
Thanks @merttumer , I have added additional ports here : https://github.com/CollaboraOnline/online/pull/4131

PR #4129, opened 2022-02-07
Make some shortcuts on Mac clients match expectations better

tml1024:
The expected shortcuts are documented in
https://support.apple.com/en-us/HT201236 . This patch handles a small
subset of those:

Command–Up Arrow: Move the insertion point to the beginning of the
document.

Command–Down Arrow: Move the insertion point to the end of the
document.

Command–Left Arrow: Move the insertion point to the beginning of the
current line.

Command–Right Arrow: Move the insertion point to the end of the
current line.

Option–Left Arrow: Move the insertion point to the beginning of the
previous word.

Option–Right Arrow: This should move the insertion point to the end of
the next word. But I don't think LO has such a shortcut so we move to
the beginning of the next word instead.

Change-Id: I6b0d36b0842388f0d49adf9b75fa9480df79c805

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1,4 +1,4 @@
-/* -*- js-indent-level: 8 -*- */
+/* -*- js-indent-level: 8; fill-column: 100 -*- */
 /*
  * L.CanvasTileLayer is a layer with canvas based rendering.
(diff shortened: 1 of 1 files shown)


tml1024:
(This was for T20750.)

PR #4132, opened 2022-02-07
Private/mert/mobile back button backport

merttumer:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
I tested and it works there is something going on : 
image

but it seems unrelated to this pr (see https://github.com/CollaboraOnline/online/pull/4136 and https://github.com/CollaboraOnline/online/pull/4137 )

Then we should maybe improve how it looks when we open the hamburger menu (when on edit mode) but then again it's a next step a nothing block this to be merged. Thanks @merttumer !

PR #4134, opened 2022-02-07
welcome CSS styles

hcvcastro:
- browser: makefile: the welcome folder has images and CSS files
- wsd: pre-process the welcome.html file
- welcome: add CSS tags to welcome dialog
- welcome: add welcome.css file
- welcome: update the welcome.html url


pedropintosilva:
@gokaysatir it seems this didn;t pass unit & cypress test. I have re-triggered it. I also test it and now the css are there, I cannot attest for the make files as I do not understand them. Oh and just to confirm the dialog appears if the django server is running is not nothing appears

hcvcastro:
@pedropintosilva that patches works only if the welcome server is down (django), you can test it without starting the django server

pedropintosilva:
@hcvcastro yes but it's what I did, I killed django and then make run again in online and couldn;t see notthing

pedropintosilva:
@hcvcastro yes but it's what I did, I killed django and then make run again in online and couldn;t see nothing and when pressing the Help > Latest updates 

I get error in the console
```js
 Uncaught ReferenceError: RPMGetInnerMostURI is not defined
    <anonymous> chrome://browser/content/certerror/aboutNetError.js:26
aboutNetError.js:26:19
    <anonymous> chrome://browser/content/certerror/aboutNetError.js:26
    InnerModuleEvaluation self-hosted:2411
    evaluation self-hosted:2358
```

PR #4136, opened 2022-02-07
Calc: Fix additional empty bezel appearing above formulabar

pedropintosilva:
It seems this element is not used at all on mobile devices and it seems
to be useful for in terms of: d7b834dd7c953ca27344b244942c43fb2049e0a0
but again it seems pretty safe to set to display none on mobile

Change-Id: I92e4b85147dd98d47f27a1ddd72bb222494864d8


pedropintosilva:
image

PR #4145, opened 2022-02-08
Check if section.boundsList is defined before use

eszkadev:


gokaysatir:
I cannot see the reason for these guards. This seems to me that this fix will shadow the root cause.

gokaysatir (approved):
Thanks for the improved safety :)

I have a comment but these are safe anyway.

PR #4147, opened 2022-02-09
Limit Failure to Store

Ashod:
- gitingore: .clangd and .cache
- wsd: count save and upload failures
- wsd: limit store failure when unloading


kendy (approved):
OK, let's get it in - thank you!

PR #4148, opened 2022-02-09
wsd: introduce remote config

Rash419:
Change-Id: Id0e2b4b524a80de88b2b3559dd18c4a95b8163fb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -9,6 +9,7 @@
 #include <stdexcept>
 #include "COOLWSD.hpp"
 #include "ProofKey.hpp"
+#include "CommandControl.hpp"
 
 /* Default host used in the start test URI */
 #define COOLWSD_TEST_HOST "localhost"
@@ -73,6 +74,10 @@
 #include <Poco/Net/PartHandler.h>
 #include <Poco/Net/SocketAddress.h>
 #include <net/HttpHelper.hpp>
+#include <Poco/Net/AcceptCertificateHandler.h>
+#include <Poco/Net/Context.h>
+#include <Poco/Net/KeyConsoleHandler.h>
+#include <Poco/Net/SSLManager.h>
 
 using Poco::Net::HTMLForm;
 using Poco::Net::PartHandler;
@@ -117,6 +122,7 @@ using Poco::Net::PartHandler;
 #include "DocumentBroker.hpp"
 #include "Exceptions.hpp"
 #include "FileServer.hpp"
+#include <common/JsonUtil.hpp>
 #include <common/FileUtil.hpp>
 #include <common/JailUtil.hpp>
 #if defined KIT_IN_PROCESS || MOBILEAPP
@@ -169,6 +175,7 @@ using Poco::StreamCopier;
 using Poco::URI;
 using Poco::Util::Application;
 using Poco::Util::HelpFormatter;
+using Poco::Util::LayeredConfiguration;
 using Poco::Util::MissingOptionException;
 using Poco::Util::Option;
 using Poco::Util::OptionSet;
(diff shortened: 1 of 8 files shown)


wsd/COOLWSD.cpp
```diff
@@ -1003,6 +1010,254 @@ COOLWSD::~COOLWSD()
...
+        }
+    }
+
+    void applyNewConfig(Poco::JSON::Object::Ptr remoteJson)
+    {
+        std::map<std::string, std::string> newAppConfig;
+
+        fetchWopiHostPatterns(newAppConfig, remoteJson);
+
+        #ifdef ENABLE_FEATURE_LOCK
```
tml1024, line 1117: We don't use the convention of indenting preprocessor lines anywhere else in the code, please don't start using it here. Consistency is important. (And then no extra indentation of the lines inside the ifdef block either.)

wsd/COOLWSD.cpp
```diff
@@ -1003,6 +1010,254 @@ COOLWSD::~COOLWSD()
...
+                booleanFlag.toString();
+            }
+        }
+        else
+        {
+            booleanFlag = "false";
+        }
+    }
+
+    enum ConfigPriority
```
tml1024, line 1250: Wouldn't it be better to use a more modern scoped enum (enum class) here? And then the PRIO_ prefix of the PRIO_JSON enumerator would be unnecessary as already the ConfigPriority identifier contains "Priority".

tml1024: Or wait, as the PRIO_JSON value is passed to a function that expects an int, why use an enum at all, why not just constexpr int PRIO_JSON = -200;?

wsd/Storage.cpp
```diff
@@ -872,11 +881,40 @@ WopiStorage::WOPIFileInfo::WOPIFileInfo(const FileInfo &fileInfo,
...
+        }
+
+        if (isCommandDisabled || isReadOnly)
+        {
+            isUserLocked = true;
+        }
+        CommandControl::LockManager::setHostReadOnly(isReadOnly);
+    }
+    CommandControl::LockManager::setLockedUser(isUserLocked);
+#endif
```
tml1024, line 915: Would be good to have an #else branch that only does a `(void) uriObject;` to avoid `error: unused parameter ‘uriObject’ [-Werror=unused-parameter]`.

wsd/COOLWSD.cpp
Rash419: ok

tml1024 (approved):
Let's get this in now I think, further refinements can be added separately.

Rash419:
> Let's get this in now I think, further refinements can be added separately.

@tml1024 just fixed a cypress thing in CI https://github.com/CollaboraOnline/online/pull/4274
it should pass now

wsd/COOLWSD.cpp
```diff
@@ -1004,6 +1011,277 @@ COOLWSD::~COOLWSD()
...
+
+            sslRemoteParams.verificationMode =
+                (sslRemoteParams.caLocation.empty() ? Poco::Net::Context::VERIFY_NONE
+                                                    : Poco::Net::Context::VERIFY_STRICT);
+            sslRemoteParams.loadDefaultCAs = true;
+
+            Poco::Net::Context::Ptr sslClientContext =
+                new Poco::Net::Context(Poco::Net::Context::CLIENT_USE, sslRemoteParams);
+
+            Poco::Net::SSLManager::instance().initializeClient(
```
Ashod, line 1068: We are initializing Poco SSL, but we are not using Poco sockets? 

`httpSession->syncRequest(request);` doesn't use Poco and so the `Poco::Net::SSLManager` is unnecessary here. Instead, we would want `ssl::Manager::initializeClientContext()`.

See `StorageBase::initialize()` for how both `Poco::Net::SSLManager` and our `ssl::Manager` are initialized. Of course we want to do the initialization once, and we want to have only one function that does that. So perhaps we can move `StorageBase::initialize()` to `COOLWSD::initializeSSL()` and remove the `// checks certificates` try/catch above?

Notice that we want to initialize the SSL layer when either Storage *or* Remote config need SSL.

Right now SSL in Remote config will work because we are reusing the Storage settings. But when/if they are different, it will stop working. And of course the above initialization code is completely unnecessary (I think the call that I highlighted here will check that it is already initialized and return immediately.)

Ashod (commented):
No objections to merge, but my comments on SSL initialization should be well understood (and improved later).

PR #4151, opened 2022-02-09
fix sidebar can be visible on pdf

merttumer:
it appears that sidebar callback can arrive before
docloaded message when it is already loaded in the background
Also app.file.fileBasedView parameter is set after docloaded
so it wasnt a correct fix. perm: message arrives before document
initialization and we can hide sidebar unless the arrived messsage is
not 'edit'. if the user have edit mode, it will be up to the users preference

Change-Id: Ifb220d8bb8baad0078bcd32f71ee6853d31adfb8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -10,6 +10,7 @@ window.app = { // Shouldn't have any functions defined.
 	file: {
 		editComment: false,
 		readOnly: true,
+		disableSidebar: false,
 		size: {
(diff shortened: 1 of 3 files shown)


merttumer:
> Hello :) Thanks for the fix @merttumer
> 
> Can you rename "hideSidebar" with "disableSidebar".

sure

gokaysatir (approved):
Thanks for the fix :)

Issue #4008, opened 2022-01-17
Can't compile  : npm ERR! code EINTEGRITY


HumanG33k:
**Describe the bug**

**To Reproduce**
```bash
#!/bin/bash

#### Goals
# Improve debian support (fix packaging)
# Add support of "new" init system 
# Work on the code

# neofetch 
#   ..,,;;;::;,..                   root@collabora 
#           `':ddd;:,.              --------- 
#                 `'dPPd:,.         OS: Devuan GNU/Linux 4 (chimaera) x86_64 
#                     `:b$$b`.      Host: VirtualBox 1.2 
#                        'P$$$d`    Kernel: 5.10.0-10-amd64 
#                         .$$$$$`   Uptime: 15 hours, 6 mins 
#                         ;$$$$$P   Packages: 1141 (dpkg) 
#                      .:P$$$$$$`   Shell: bash 5.1.4 
#                  .,:b$$$$$$$;'    Resolution: preferred 
#             .,:dP$$$$$$$$b:'      Terminal: /dev/pts/1 
#      .,:;db$$$$$$$$$$Pd'`         CPU: Intel i5-4590 (1) @ 3.292GHz 
# ,db$$$$$$$$$$$$$$b:'`             GPU: 00:02.0 VMware SVGA II Adapter 
#:$$$$$$$$$$$$b:'`                  Memory: 73MiB / 1982MiB 
# `$$$$$bd:''`
#   `'''`                                                   



### PATHS
#
# /usr/src/collabora
# tree -L 2
# .
# ├── build.sh  (this File)
# └── workspace
#     ├── collabora-online
#     └── libreofficecore

function require_system()
{
  apt install aptitude -y
}

function require_dependencies()
{
  aptitude install -y \
    dialog

  aptitude install -y \
    libpoco-dev python3-polib libcap-dev npm \
    libpam-dev wget git build-essential libtool \
    libcap2-bin python3-lxml libpng-dev libcppunit-dev \
    pkg-config fontconfig chromium
}

function grab_assets()
{
  mkdir -p workspace/libreofficecore
  cd ./workspace/libreofficecore
  export LOCOREPATH=$(pwd)
  wget https://github.com/CollaboraOnline/online/releases/download/for-code-assets/co>
  tar xvf core-co-2021-assets.tar.gz
  cd -
}

function setLOCOREPATH()
{
  cd ./workspace/libreofficecore
  export LOCOREPATH=$(pwd)
  cd -
}

[...]

PR #4146, opened 2022-02-08
Fixed broken freeBSD CI

lpranam:
Signed-off-by: Pranam Lashkari <lpranam@collabora.com>
Change-Id: I39630d59f432ddd0712063ded0629ab0f4953839

* Target version: master 

### Checklist

- [ ] Code is properly formatted
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #4153, opened 2022-02-09
Use package-lock.json instead of shrinkwrap

jblemee:
Signed-off-by: Jean-Baptiste Lemée <git@lemee.co>
Change-Id: I52f299b9b623e80383663111df3318881ec3d0ea

* Resolves: #4008
* Target version: master 

### TODO

- [x] Revert  #3815

### Checklist


pedropintosilva:
Hi @jblemee  and, albeit in draft state, thanks for pushing your first PR !

This seems a good idea, but somehow if I remember correctly there was something mentioned in one of the community meetings along the line of "we shouldn't pin versions by pushing shrinkwrap" but maybe I'm just confused. Could you check this @alessioalex-1and1 @timar ?

alessioalex-1and1 (approved):

pedropintosilva:
Ah so I just talked with @gokaysatir and we cannot do this if we want to keep the possibility of building without internet

pedropintosilva:
nevertheless this error should now be fixed with @lpranam 's commit: https://github.com/CollaboraOnline/online/pull/4146/files

pedropintosilva:
@jblemee I noticed that your branch https://github.com/hyperlex/online/commits/chore/remove-shrinkwrap is not up to date and so it does not include that above commit mentioned. Could you test it again?

jblemee:
> @jblemee I noticed that your branch https://github.com/hyperlex/online/commits/chore/remove-shrinkwrap is not up to date and so it does not include that above commit mentioned. Could you test it again?

I've test and it's works now with shrinkwrap. Maybe we should update shrinkwrap with a version compatible with lasts npm/node versions ?

PR #4119, opened 2022-02-04
color-palette: use color-palette vars instead of hardcoded colors

Andreas-Kainz:
add a description to color-palette to define

Signed-off-by: andreas kainz <kainz.a@gmail.com>
Change-Id: I23f78c8c6e87417095c2c0bc029fda5f43f5a488

* Resolves: #3772

PR #4164, opened 2022-02-11
Private/pedro/color schemes

pedropintosilva:
improvement on https://github.com/CollaboraOnline/online/pull/4119


Andreas-Kainz:
can we push it to master?

pedropintosilva:
@Rash419 I have merged what it seems to be needing some update after these changes , could you then double check it

PR #4174, opened 2022-02-12
Count and report on various internal exceptions.

mmeeks:
An initial set of seven of these, easy to add more as/when needed.

Change-Id: I6c65e052d00f9eaa10adee3c9464043e4c594848

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


arrowd:
You probably need `#include <string>` to fix build with libc++

mmeeks:
> You probably need `#include <string>` to fix build with libc++
Of course - I assumed I had already, but looked at the wrong branch I think =) thanks !

PR #4190, opened 2022-02-14
fix copy comments breaking the canvassection completely

merttumer:
core sends same id for copy comments and this breaks the
existing implementation because sections are named based on the id.

Change-Id: Ia410227b7a42ac8ea7682f12aa980b34ee5c940b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix :)

PR #4193, opened 2022-02-14
browser: add missing error method for logging

alessioalex-1and1:
Change-Id: Id4aa67e98779dd0d5c92172425baa753fcaca27c

* Resolves: relates to #3883 
* Target version: master 

### Summary

The `.error()` method was missing from the logging ones, so `window.app.console.error()` would not work. 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -68,7 +68,7 @@ window.app = { // Shouldn't have any functions defined.
 	// enable later toggling
 	global.setLogging = function(doLogging)
 	{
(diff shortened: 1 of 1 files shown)


alessioalex-1and1:
@mmeeks this would also have to be backported to those versions you added the methods to ensure there's no bug. Just add that `error` method to those as well in case it's missing.

mmeeks:
Thanks; will poke.

PR #4201, opened 2022-02-14
Expose a11y check online

NickWingate:
* Resolves: # 
* Target version: master 

### Summary

Exposes the accessibility checker used in LO writer and core for use in COOL Writer.
Can be accessed in Notebookbar under review and also the hamburger drop down in the top left under tools>accessibility check.
Also added in the menubar, but I don't see this anywhere in the actual UI- possibly not used not sure?

Note I just used the svg for the a11y logo from [their website](https://www.a11yproject.com/) and changed the color, don't know if we want this to be something else

### TODO

- [ ] Possibly change logo
- [x] Does mobile need accessibility checker?
- [ ] Should the checker be enabled when the document is readonly?
- [ ] Update [online help menu](https://help.collaboraoffice.com/21.06/en-US/text/shared/05/err_html.html?System=&DbPAR=WRITER&HID=svx/ui/accessibilitycheckdialog/dialogButtons) with content from [libre office help menu](https://help.libreoffice.org/7.3/en-GB/text/swriter/01/accessibility_check.html?&DbPAR=WRITER&System=WIN)

### Checklist

### DIFF
--- a/browser/src/unocommands.js
+++ b/browser/src/unocommands.js
@@ -5,6 +5,7 @@ var unoCommandsArray = {
 	'AcceptTrackedChange':{text:{context:_('Accept Change'),menu:_('Accept'),},},
 	'AcceptTrackedChangeToNext':{text:{context:_('Accept Track Change and select the next one'),menu:_('Accept and Move to Next'),},},
 	'AcceptTrackedChanges':{text:{context:_('Manage Track Changes'),menu:_('~Manage...'),},},
+	'AccessibilityCheck':{text:{menu:_('~Accessibility Check...'),},},
 	'AddName':{spreadsheet:{menu:_('~Define...'),},},
 	'AlignBlock':{spreadsheet:{menu:_('Justified'),},},
 	'AlignBottom':{spreadsheet:{menu:_('Align Bottom'),},text:{menu:_('Align Bottom to Anchor'),},},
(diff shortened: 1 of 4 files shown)


NickWingate:
@pedropintosilva it seems no icon exists for the accessibility checker anywhere in core. Would you mind making one to replace the svg i took from https://github.com/a11yproject/a11yproject.com/blob/main/src/img/logo.svg when you have the time. Thanks!

mmeeks:
Looks good; is the a11y checker already non-modal in the core ? =)

NickWingate:
In online the dialog is non-modal but in the LO writer it is modal.  I'll work on getting the core dialog to be non-modal too when I improve other UX stuff.

browser/src/unocommands.js
```diff
@@ -24,7 +25,7 @@ var unoCommandsArray = {
 	'AutoFormatMenu':{text:{menu:_('AutoCorr~ect'),},},
 	'BackColor':{text:{menu:_('Highlight Color'),},},
 	'BackgroundColor':{global:{menu:_('Background Color'),},},
-	'BasicShapes':{global:{context:_('Shapes'),menu:_('Shapes'),},},
+	'BasicShapes':{global:{context:_('Basic Shapes'),menu:_('Basic Shapes'),},},
```
eszkadev, line 28: this is unintended I think, please revert this line :)

mmeeks:
Great - so I think we need to fix the core modality issue before enabling this =) otherwise collaborative editing will really suffer with big tangles on the stack =)

eszkadev (approved):

merttumer (approved):

PR #4220, opened 2022-02-16
Send error on failed load and/or exception during document load

mmeeks:
This avoids an endless spinner for the loading mobile app in this case.

Change-Id: I047797cbdb066a236afc40337d2eba08f9c0c84b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks!

Issue #4215, opened 2022-02-16
replace floppy disk icon for save


Andreas-Kainz:
The Save icon is a floppy disk symbol, which was discussed quite often at LibO cause the time when floppy disks was used, is long long ago.

Compare to LibO COOL is an online based office suite. So I suggest to change the save icon from a floppy disk symbol to something else.

![2022-02-16 11_28_22-mobileOfficeApps docx - Dateien - ARSENAL RACE](https://user-images.githubusercontent.com/8517736/154245928-80248368-0bcf-46d5-9436-174b0e863fe4.png)

Andreas-Kainz:
On mobile we use an ✓ symbol.

![2022-02-16 11_29_36-mobileOfficeApps docx - Dateien - ARSENAL RACE](https://user-images.githubusercontent.com/8517736/154246097-60564f4f-db0e-4d9e-9b84-eb0a6e3362b3.png)

Which is from my point of view a bit to generic for desktop, but could be used also than we are more similar between desktop and mobile.

Andreas-Kainz:
As cool data storage is ordinary cloud based, we can use an cloud symbol for save.

![image](https://user-images.githubusercontent.com/8517736/153958180-674ba166-c393-40a3-8947-5caa7d72de1e.png)

pedropintosilva:
ref: https://github.com/CollaboraOnline/online/pull/4200#pullrequestreview-883080023 ; https://github.com/CollaboraOnline/online/pull/4209

Even if after the fact, thanks for creating this issue : )

I'm not opposed to change the save icon but if we do so we probably should think a bit long and hard of what it would be instead.

Classic floppy disc has many advantages to name a few:
* If initially used as a representation of the physical device where you store stuff it then evolved in something else. Not a representation of a floppy disk but as the classic symbol representing the action of saving
* Our user base is not so young that would dramatically benefit them (quite the opposite I would say)
  * But even when asking to kids and pre-teenagers to spot what's the icon for the saving action they will always point to what they call "classic"/"saving" "icon/square"
 
With that been said I think we can do a change and then wait for the feedback (as there are still many users not using branding packages, so just CODE bare-bones and sometimes with their own CSS customizations above) 

## Object VS Action

* I really don't like how the [cloud](https://github.com/CollaboraOnline/online/issues/4215#issuecomment-1041342494) is portrait here, it does not point to the act but instead to the (false) object
  * There is no cloud only computers : p nah just joking
* So I would really focus on the action it self
[...]

PR #4224, opened 2022-02-17
document-logo use 16px app icons

Andreas-Kainz:
update the menubar.css file to make the icon size 16 x 16px
add 16px x-office-* icons so that no scaling is needed.

Signed-off-by: andreas kainz <kainz.a@gmail.com>
Change-Id: If6d84218920e8200e644fa8ded5c950f61c02197

* Resolves: #4214
* Target version: master 

#### Classic Toolbar
Screenshot_20220217_233606

#### Notebookbar

The icon size is now the same than the notebookbar-shortcuts-bar

Screenshot_20220217_233039
Screenshot_20220217_233113
Screenshot_20220217_233147
Screenshot_20220217_233207

PR #4225, opened 2022-02-17
Save icon use text color and primary-color for feedback

Andreas-Kainz:
All save icons are now gray,
like the print icon or the hamburger menu icon.

The gray color is the color-main-text as the icon
is next to text, it's not so visible.

Indicators like saveas or modified use color-primary
in the icons and in the .css file

Change-Id: I7311b71dff356d2c9aaf9d03d97f33a8ae9fb830

* Resolves: #4215

#### Classic Toolbar
Screenshot_20220217_234942

Screenshot_20220217_235028

#### Notebookbar
Screenshot_20220218_000506


Andreas-Kainz:
fyi the document-icon will have the same size than the hamburger and save icon in the shortcuts list. See #4224

rizmut:
As additional info, Colibre in LibreOffice has been updated. It's up to you whether to sync up or not.

Andreas-Kainz:
@pedropintosilva your prefernce for gray or colourful?

pedropintosilva:
I think a neutral colour like gray would be better, specially because we then use colourful indicator on top (dot)

pedropintosilva:
I would prefer to avoid that purple, specially because we use a different colour on top via the dot indicator. So either app colour or a neutral colour would be fine by me. Thanks _merging_

rizmut:
> > As additional info, Colibre in LibreOffice has been updated. It's up to you whether to sync up or not.
> 
> good point, will update to the save icons to the new ones https://github.com/LibreOffice/core/blob/master/icon-themes/colibre_svg/cmd/lc_save.svg
So you did not update it, did you?

PR #4241, opened 2022-02-19
CSS Var: fix for toolbar dropdown menu border radius

Andreas-Kainz:
the border-radius was needed at .select2.container--default and
.select2.selection--single and __rendered to have nice borders

Change-Id: I6ae386d65f45c49f221a3c987b101132eb4d56c7

image


Andreas-Kainz:
Toolbar is done with css var's

Issue #3955, opened 2022-01-10
Add user toggle for the ui mode to switch while editing


juliushaertl:
**Is your feature request related to a problem? Please describe.**
Users have different preferences on the ui mode to either use the notebookbar or the classic toolbar.

**Describe the solution you'd like**
An option to switch while editing a document between the two modes. E.g. through the Menu

View > Switch to classic toolbar

or
 
View > Switch to notebook bar

**Describe alternatives you've considered**
A user setting on the integration side could be used to store the default, but immediate switching would be more intuitive.

**Additional context**
This issue is the outcome of previous discussion with @jancborchardt and @pedropintosilva

Ezinnem:
@eszkadev @pedropintosilva, please can you take a look sometime?

eszkadev:
`this._map.fire('changeuimode', {mode: msg.Values.Mode, force: false});`   from browser/src/map/handler/Map.WOPI.js
switches the mode.
for more info see: browser/src/control/Control.UIManager.js   `onChangeUIMode`

eszkadev:
Switching the mode should also notify the integrator by post message. In UIManager we should send in  `onChangeUIMode`:
eg. `map.fire('postMessage', {msgId: 'UI_ChangedMode', args: {Mode: 'notebookbar'}});`

Issue #4115, opened 2022-02-03
add command to collapse notebookbar tabs


Andreas-Kainz:
As there is space at the toggle status bar, would we add there an command to collapse the notebookbar.

![image](https://user-images.githubusercontent.com/8517736/152250462-7befa285-22d7-4f64-8588-735f72247350.png)

Andreas-Kainz:
feedback from Pedro see https://github.com/CollaboraOnline/online/pull/4107#issuecomment-1028792062

pedropintosilva:
Little update on this: We already have the collapse functionally implemented with 1fb8e2475464d7adc6fb7d3590ad0f72a1e9a0e9 , bb9ee333bdb0f16ef8f414a37c5294607e4afee9 and 30e65384535f384889a9a07d5c666393c493ddc4 and we also have added a tooltip with 657b193eb94c8738f57b7a267d047e6b7685b8c0

so all the code is already available in the browser/src/control/Control.UIManager.js namely isNotebookbarCollapsed extendNotebookbar and collapseNotebookbar

PR #4254, opened 2022-02-21
Add toggle for UI mode to menubar and notebookbar

juliushaertl:
* Resolves: #3955 
* Target version: master 

### Summary

This implements a toggleuimode action that allows switching between classic and notebookbar UI mode which is then exposed in the menubar (View > Use Notebookbar UI) and notebookbar (View > Classic UI).

@eszkadev I'd be interested about feedback on the implementation as I could not get a clear understanding on how the actions are shared between the menubar and notebookbar and if the approach to trigger them in the way done is the best.

@pedropintosilva Do you have a suggestion for a good icon from the existing ones maybe? For now I added a temporary commit with a material icon, but should be something matching the upstream icon theme there of course.

https://user-images.githubusercontent.com/3404133/154939267-0169e562-9cb8-4d58-9062-5f1512b9fe95.mov

### TODO

- [x] Switching the mode should also notify the integrator by post message. In UIManager we should send in onChangeUIMode: eg. `map.fire('postMessage', {msgId: 'Action_ChangeUIMode_Resp', args: {Mode: 'notebookbar'}});`
- [x] Find a proper icon for the notebookbar to switch to classic UI
- [x] Figure out why the menu bar is not expanded after switching from notebookbar to classic
- [x] Move **before** or after the regular view toggle options in the menu/notebookbar tab

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1159,9 +1159,12 @@ L.Control.Menubar = L.Control.extend({
 		map.on('addmenu', this._addMenu, this);
 		map.on('commandvalues', this._onInitLanguagesMenu, this);
 		map.on('updatetoolbarcommandvalues', this._onStyleMenu, this);
+
+		this._resetOverflow();
 	},
 
 	onRemove: function() {
+
 		this._map.off('doclayerinit', this._onDocLayerInit, this);
 		this._map.off('updatepermission', this._onRefresh, this);
 		this._map.off('addmenu', this._addMenu, this);
@@ -1170,6 +1173,8 @@ L.Control.Menubar = L.Control.extend({
 
 		this._menubarCont.remove();
 		this._menubarCont = null;
+
+		this._resetOverflow();
 	},
 
 	_addMenu: function (e) {
@@ -1432,10 +1437,22 @@ L.Control.Menubar = L.Control.extend({
 		}
 	},
 
-	_onMouseOut: function() {
-		$('.main-nav.hasnotebookbar').css('overflow', 'scroll hidden');
+	// needed for smartmenus to work inside notebookbar
+	_setupOverflow: function() {
+		$('.main-nav.hasnotebookbar').css('overflow', 'visible');
+		$('.notebookbar-scroll-wrapper').css('overflow', 'visible');
+	},
+
+	_resetOverflow: function() {
+		$('.main-nav').css('overflow', '');
 		$('.notebookbar-scroll-wrapper').css('overflow', '');
 	},
+
+	_onMouseOut: function(e) {
+		var self = e.data.self;
+		self._resetOverflow();
+	},
+
 	_checkedMenu: function(uno, item) {
 		var constChecked = 'lo-menu-item-checked';
 		var state = this._map['stateChangeHandler'].getItemValue(uno);
@@ -1449,9 +1466,8 @@ L.Control.Menubar = L.Control.extend({
 	},
 
 	_beforeShow: function(e, menu) {
-		$('.main-nav').css('overflow', 'visible');
-		$('.notebookbar-scroll-wrapper').css('overflow', 'visible');
 		var self = e.data.self;
+		self._setupOverflow();
 		var items = $(menu).children().children('a').not('.has-submenu');
 		$(items).each(function() {
 			var aItem = this;
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -111,8 +111,8 @@ L.Control.Menubar = L.Control.extend({
 					{name: _UNO('.uno:ZoomMinus', 'text'), id: 'zoomout', type: 'action',},
 					{name: _('Reset zoom'), id: 'zoomreset', type: 'action'},
 					{type: 'separator'},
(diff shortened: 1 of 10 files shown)


Andreas-Kainz:
great. as there is only one command in the view tab, I would add as second command #4115 

for the icon stuff, I can do the icon work if wanted.

juliushaertl:
> great. as there is only one command in the view tab, I would add as second command https://github.com/CollaboraOnline/online/issues/4115

Sure, makes sense to me. 

> for the icon stuff, I can do the icon work if wanted.

That would be much appreciated.

pedropintosilva:
Thanks @juliushaertl !

pedropintosilva:
@juliushaertl : oh I see @Andreas-Kainz is on it already, great thanks : )

eszkadev:
Hi @juliushaertl   it looks great. I've only noticed that style previews were not initialized (it regressed from 6.4). I pushed one commit here to fix that. It also requires core change I made here: https://gerrit.libreoffice.org/c/core/+/130341

browser/src/control/Control.Menubar.js
```diff
@@ -1491,6 +1492,13 @@ L.Control.Menubar = L.Control.extend({
 							$(aItem).removeClass(constChecked);
 						}
 
+					} else if (id === 'toggleuimode') {
+						if (window.userInterfaceMode === 'notebookbar') {
+							$(aItem).text(_('Use Classic UI'));
+						} else {
+							$(aItem).text(_('Use Notebookbar UI'));
```
eszkadev, line 1499: I think this piece of code is not needed?  NB will use own label defined somewhere else and in classic mode we alwyas need "Use Notebookbar UI" text

eszkadev:
> @eszkadev I'd be interested about feedback on the implementation as I could not get a clear understanding on how the actions are shared between the menubar and notebookbar and if the approach to trigger them in the way done is the best.

This is done correctly. Code sharing between menubar and notebookbar is a bit complicated due to old requirements (which are no longer valid I think), but we are progressing with cleanups from time to time so I hope it will be better in the future :)

browser/src/control/Control.Menubar.js
```diff
@@ -1411,7 +1412,7 @@ L.Control.Menubar = L.Control.extend({
 	},
 
 	_beforeShow: function(e, menu) {
-		$('.main-nav.hasnotebookbar').css('overflow', 'visible');
+		$('.main-nav').css('overflow', 'visible');
```
eszkadev, line 1415: When switching classic -> nb -> classig I've got:
switch

this line can be a cause

jancborchardt:
@juliushaertl as mentioned in the meeting, just for the record, we should probably have the "Use NotebookBar view" as first entry in that list of 4 toggles.

juliushaertl:
Good for review from my side. One thing I'd like to get adjusted would be to switch the menubartoolitem to a bigtoolitem on the notebookbar but it seems that I miss something there with the automatic conversion to ids for the icons. @pedropintosilva Do you have any hint there how that could be achieved properly maybe?

eszkadev:
Only `menubartoolitem` calls the action implemented in Control.Menubar.js file, if we want to convert that to `bigtoolitem` we need to create `bigmenubartoolitem` or add custom implementation for toggle ui `bigtoolitem`.

eszkadev:
retrigger checks

browser/src/control/Control.Menubar.js
eszkadev: I can still reproduce this issue in Chrome:
https://user-images.githubusercontent.com/5307369/159482695-f349ffc1-3372-4d45-be05-4420ca936198.mp4

jancborchardt:
Wording-wise, I know it's hard to drop "Notebookbar", but we can rename "Classic" to "Compact", right @pedropintosilva? :)

browser/src/control/Control.UIManager.js
```diff
@@ -301,6 +301,8 @@ L.Control.UIManager = L.Control.extend({
 
 		document.body.setAttribute('data-userInterfaceMode', uiMode.mode);
 
+		this.map.fire('postMessage', {msgId: 'Action_ChangeUIMode_Resp', args: {Mode: uiMode}});
```
eszkadev, line 304: AI: We will need to update SDK with new event: `Action_ChangeUIMode_Resp`

juliushaertl:
- [ ] Fix style loading when starting from classic and switching to notebook bar
- [ ] Fix overflow setting on chrome
- [ ] Keep naming for now, naming to be discussed further in a follow up
  - Change naming of classic to compact mode
  - Change naming of notebookbar naming to tabbed mode

gmasei11:
We discussed this topic internally and, in general, we have a positive feedback on this proposal. This means that is OK for us to use this implementation of UI mode switching feature instead of our: exposing the feature only in the View menu/tab and removing it from toolbar. But if, in the meantime, you've changed your mind and have other proposals, especially after the last COOL weekly meeting discussion, please share them.

eszkadev (approved):

browser/src/control/Control.UIManager.js
eszkadev: done

juliushaertl:
@eszkadev Thanks a tons for taking over the finishing :)

PR #4255, opened 2022-02-21
Notebookbar: Remove ellipsis from generated uno button labels

juliushaertl:
* Target version: master 

### Summary

Contributes to the temporary solution of https://github.com/CollaboraOnline/online/issues/1285#issuecomment-824547942 by applying the removal of potentially existing ellipsis on a wider range by performing this cleanup in the JSDialogBuilder.

I could imagine that this also has an effect on other jsdialog based components, however I could not see any noticeable impact on the sidebar for example.

What do you think about that approach @pedropintosilva @Andreas-Kainz 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -362,6 +362,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_cleanText: function(text) {
 		if (!text)
 			return '';
+		if (text.endsWith('...'))
+			text = text.slice(0, -3);
 		return text.replace('~', '');
 	},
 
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -362,6 +362,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_cleanText: function(text) {
 		if (!text)
 			return '';
+		if (text.endsWith('...'))
```
juliushaertl, line 365: Depending if IE11 support would still be required, this would need to be adjusted to be backward compatible.

Andreas-Kainz: IE11 support was dropped. So no need to care.

pedropintosilva:
Tested, thanks @juliushaertl !

Issue #4260, opened 2022-02-21
Sidebar Layout update


Andreas-Kainz:
Sidebar elements should be align.

Therefore there is 
- left-align the label
- right-align the widget

Andreas-Kainz:
make all widgets the same width, what do you think @pedropintosilva @juliushaertl ? 

#### Result / After
![image](https://user-images.githubusercontent.com/8517736/155043859-03cf5181-11b0-4e53-8de9-9729397eaf87.png)

#### Before
![Screenshot_20220222_012545](https://user-images.githubusercontent.com/8517736/155043889-4d8f8ee3-b438-43c5-b339-1d79bca43cd0.png)

PR #4261, opened 2022-02-21
Sidebar: hide separator and margin for frame-label

Andreas-Kainz:
Issue #4260

separators are nice, but didn't bring that much  for visual separation.
it's more visual noise than help for separation

compare to separators margins are a good idea for separation
so the frame-label get an margin which make also the label
layout better.

Change-Id: I402591327e6cdd5328bd95e5e2c342a56b66471d

#### writer default sidebar layout
Screenshot_20220221_203224

#### writer table sidebar layout
Screenshot_20220221_203055

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -44,7 +44,7 @@
 }
 
 .jsdialog.vertical p, .ui-frame-label {
-	margin: 0px;
+	margin: 10px 0 2px;
 	color: var(--color-main-text);
 	font-size: 10px;
 	font-weight: bold;
@@ -70,6 +70,7 @@
 }
 
 .jsdialog.ui-separator.horizontal {
+	display: none;
 	width: 100%;
 	padding-top: 6px;
 	/* border-left: 1px dashed #888; */
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -341,12 +341,12 @@ td.jsdialog .jsdialog.cell.sidebar {
 }
 
 #spacinglabel {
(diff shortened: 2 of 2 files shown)


Andreas-Kainz:
#### writer table sidebar layout in master (before)

image

browser/css/jsdialogs.css
```diff
@@ -44,7 +44,7 @@
 }
 
 .jsdialog.vertical p, .ui-frame-label {
-	margin: 0px;
+	margin: 10px 0 2px;
```
pedropintosilva, line 47: maybe we can avoid touching this

pedropintosilva (changes requested):
I agree but I would prefer to keep using the separator even if it's invisible (so any additional spacing would be added/adjust in that element)

Andreas-Kainz:
thanks for the feedback
- the horizontal separator is displayed, but use the background color, so there is no visual noise
- margins wasn't touched

pedropintosilva:
no problem : ) @Andreas-Kainz , 
I'm ok with this right now but if we don't want those lines wouldn't better and **cleaner** to just :
```diff
 .jsdialog.ui-separator.horizontal {
-       width: 100%;
        padding-top: 6px;
-       /* border-left: 1px dashed #888; */
-       background-image: linear-gradient(#888 22%, rgba(255,255,255,0) 0%);
-       background-position: center top;
-       background-size: 14px 1px;
-       background-repeat: repeat-x;
-       box-shadow: inset 0 1px 0 0 var(--color-border-lighter);
 }

```

PR #4269, opened 2022-02-22
Sidebar Impress widget right align as everywhere else

Andreas-Kainz:
In all apps the sidebar widgets are justified: flex-end
only in impress the rule wasn't selected. But than the
alignment didn't work in impress.

Change-Id: I5f44496bb570f64a0b21d14fb4493665dbd31d07

#### before
Screenshot_20220222_231540

#### after
Screenshot_20220222_231757

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -180,7 +180,7 @@ div#thousandseparator.checkbutton.jsdialog.sidebar {
 }
 
 #settransparency, #fontheight,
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
as I didn't know what's the reason for `:not(.presentation-doctype)` @pedropintosilva maybe you know why.

pedropintosilva:
Thanks! Good catch : ) 
I think I needed to add that because the first look was not really good : 
image

so that was a temporary solution so we could have: 
image

but I totally agree with you, do you think you could add one more commit here that addresses that situation only within `SlideBackgroundPanel` ?

Andreas-Kainz:
`SlideBackgroundPanel` 

sure for SlideBackgroundPanel (and also in Draw) I will make an fix but not for everything else.

Issue #4070, opened 2022-01-27
Add last-mod string to notebookbar


Andreas-Kainz:
#### Classic toolbar layout

![2022-01-27 12_59_51-](https://user-images.githubusercontent.com/8517736/151354866-b6eef67d-1a7e-4405-a564-536a23379a94.png)

#### notebookbar layout

should show also the `Last saved:` (`last-mod`) information in the header area.

Ezinnem:
Confirmed in the Notebookbar UI COOLWSD version: 21.11.1.4 git hash: [739da71](https://github.com/CollaboraOnline/online/commits/739da71)

Andreas-Kainz:
If someone can give me am codepointer, it would be cool. @pedropintosilva ?

Andreas-Kainz:
add .uno:Save next to the last-mod string

![save](https://user-images.githubusercontent.com/8517736/155394797-aa084144-497d-4961-9739-df8f8f3c01c8.png)

Andreas-Kainz:
@eszkadev do you have a codepointer or something like this. It's one of the last todo for https://github.com/nextcloud/richdocuments/issues/1976

PR #4279, opened 2022-02-23
Shortcuts-bar remove menubar

Andreas-Kainz:
all commands from menubar was added to the notebookbar
so no need for the hamburger menubar button

Change-Id: Icb283b40c269d2f92a29ad16dba8dda9d0f860d2

I didn't remove the code from getShortcutsBarData cause there are 3rd party apps that use this Shortcuts-bar to add there commands there. 

@pedropintosilva I checked all commands. All commands are available somewhere else. So it's possible to hide the hamburger menubar button.


Andreas-Kainz:
image

Andreas-Kainz:
The `.uno:Save` command I would suggest to move to the last-mod string #4070

eszkadev:
nice, we can also remove widget implementation for that menu:
from `browser/src/control/Control.NotebookbarBuilder.js`
line: `this._toolitemHandlers['.uno:Menubar'] = this._menubarControl;`
and whole: `_menubarControl` function

pedropintosilva (approved):
it seems good, thansk @Andreas-Kainz

Issue #3882, opened 2021-12-20
harmonize font size


Andreas-Kainz:
**before**
![2021-12-20 09_10_09-mobileOfficeApps docx - Files - Nextcloud](https://user-images.githubusercontent.com/8517736/146734281-76ee9e32-078f-4fab-9cd0-7105df5ac0d4.png)

**after**
![2021-12-20 09_08_53-Koralmbahn Überblick xlsx - Excel](https://user-images.githubusercontent.com/8517736/146733785-2542fa3c-d7ce-4502-b934-ca8a7b5a0099.png)

### issue
Elements like Style selector (Title) have different font sizes at the toolbar and sidebar. In notebokbar toolbar layout the labels use 1em (16px) font height. In classic toolbar it's 12px.

### goal
- harmonize the font size and define it at the root css file.
- easier to change the font size, cause you have to change only a view values in best case one value.

Andreas-Kainz:
**default font size: 14px** (instead of 12px as it's now for most areas)

PR #4280, opened 2022-02-23
Color-Palette Vars added

Andreas-Kainz:
the following vars were added
- header-height
- default-font-size
- color-box-shadow

value names come from
- https://github.com/nextcloud/server/blob/master/core/css/css-variables.scss

Change-Id: I05a081d5b27f6e9d8102002805c9194062f8ea0f

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -28,6 +28,7 @@
 	height: 32px;
 	line-height: 0em;
 	color: var(--color-main-text);
+	font-size: var(--default-font-size);
 	min-width: 62px;
 	background-color: var(--color-background-dark);
 	border: 1px solid var(--color-border);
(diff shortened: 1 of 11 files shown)


Andreas-Kainz:
Harmonize font size see #3882

browser/css/btns.css
```diff
@@ -1,4 +1,4 @@
-.jsdialog-container .ui-button-box button,
+ffont.jsdialog-container .ui-button-box button,
```
juliushaertl, line 1:
```suggestion
.jsdialog-container .ui-button-box button,
```

That is a typo I guess?

Andreas-Kainz: thanks a lot

Issue #4282, opened 2022-02-24
Group download as icons into a dropdown menu for the Notebookbar UI


alessioalex-1and1:
**Is your feature request related to a problem? Please describe.**
There are a lot of 'download as' icons when clicking on the File tab, Notebookbar UI mode enabled. 

**Describe the solution you'd like**
There needs to be a setting that you can check to enable those into a dropdown so that it's more manageble.

**Describe alternatives you've considered**
none

**Additional context**

This is how the icons look like at the moment:

![How it is at the moment](https://cldup.com/Ir_bPCTKzP.png)

This is how the icons look could like at (via enabling a setting):

![How it should be via a setting](https://cldup.com/nptZ0r-O-R.png)

Andreas-Kainz:
This is how the LibreOffice (writer) File tab look like

![2022-02-24 12_28_40-Untitled 1 - LibreOffice Writer](https://user-images.githubusercontent.com/8517736/155515893-5124c510-7ef0-4c2c-9fb8-7805ae7ad8bb.png)

Andreas-Kainz:
And here is the file (writer) menubar section
![2022-02-24 12_28_09-Untitled 1 - LibreOffice Writer](https://user-images.githubusercontent.com/8517736/155516014-ae086f9f-9848-415e-a378-15ee7cbea636.png)

Andreas-Kainz:
if you click on Export you get a save dialog with the following options
![2022-02-24 12_27_57-Untitled 1 - LibreOffice Writer](https://user-images.githubusercontent.com/8517736/155516206-8261063f-f1ed-4e9a-bb2e-a93503d9f97f.png)

Andreas-Kainz:
what I want to say with the screenshots from LibO
- In LibO there are two groups `Save/SaveAs/SaveRemote` and `Export/PDF/EBUP`
- In Cool there are less commands (available now), but as the Export commands are also available within cool, we could offer the commands from LibO

#### suggestions for cool
![File](https://user-images.githubusercontent.com/8517736/155520843-025bb66a-49e6-470e-a6af-3d0ed298f84b.png)

So we have the benefit for one click download the most often used formats and two clicks for all other options that are available within LibO.

Andreas-Kainz:
is it possible to add Export as PDF (which open first the dialog where you can define settings), export png, jpg, html, ... ? 

Cause for this more exotic downloads an dropdown menu could be very useful.

alessioalex-1and1:
[...]

PR #4283, opened 2022-02-24
browser: add option for grouping the download as icons

alessioalex-1and1:
Change-Id: I3c8c1964879bec8c4e31a85baefb7b442ceae29f

* Resolves: #4282
* Target version: master 

### Summary

This PR groups the download as icons into a dropdown in the toolbar (select `File` tab), Notebookbar UI. It adds a boolean property to the xml file and reads from that (by adding it into the html file). If that is enabled then there's a different logic when displaying the menu. I have also added a clean function that strips the menu options of empty objects (because those would modify the HTML structure of the menu needlessly).

### TODO

- Change the setting for `group_download_as` from `false` to `true` in `coolwsd.in.xml` and see the download as icons being grouped together in the toolbar

### Checklist


pedropintosilva (approved):
Thanks it works, I wonder if it wouldn't be better to make it a bigtoolitem but then again if cusers or integrators etc want to add more items they will probably appreciate if it's small as it is, merging it , thanks!

PR #4295, opened 2022-02-25
NB: Fix Save savemodified indicators' contrast

pedropintosilva:
Context:
With the f4cde4a18c8a402f5028971987193a2b28aa3d69 there is no visual
delimiter between the icon and the indicator which results in lack of
contrast.

- Add back shadow and tweak it so we do have better contrast between
them
- Also tweak indicators' position so it's aligned to the top and so it
feels more like an indicator (notifying the user)

Change-Id: I2ff6e00383a7ad7305680c92b587fc1b6890a886

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -100,9 +100,10 @@
 	border-radius: 2px;
 	filter: none !important;
 	background: var(--color-primary-lighter);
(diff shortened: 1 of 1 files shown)


pedropintosilva:
_no need to wait for tests, as this is basically a partly revert of something_

pedropintosilva:
save-savemodified
Fixed: 
save-savemodified-fixed

PR #4296, opened 2022-02-25
autofilter: use gridwindow instead of screen offset

eszkadev:
core is adjusted to send offset for autofilter popup
related to gridwindow (spreadsheet area) instead of
absolute screen coordinates - what helps to simplify
code.

requires: https://gerrit.libreoffice.org/c/core/+/130528


pedropintosilva:
I tested, it seems good : )

merttumer (approved):
Looks good

PR #4298, opened 2022-02-25
Annotations: fix collapsed comment container

pedropintosilva:
with some recent colors changes (css vars) the container of collapsed
annotations (comment and reply) and its background was superseded with
another css rule:
 - Add to the rule the id so it does not get overruled
 - Also add background-color instead of just bg

Change-Id: If8c847a80d761e177ad1d12b6c4f7ea9f7690ad2


pedropintosilva:
This seems to not be needed anymore, new changes were add later on and already merged before this PR

PR #4299, opened 2022-02-25
Dialog colors use var's

Andreas-Kainz:
colorpicker use var's

Change-Id: I621005329766b529e1c031da4c3716f9f587c462

image


Andreas-Kainz:
Its not that easy to describe each single change. All the changes are in general about use color-scheme variables and normalization of stuff.

However it's better to describe in tiny step's and without typo. So I start again.

PR #4301, opened 2022-02-25
Toolbar/Notebookbar scroll-left/right use var's

Andreas-Kainz:
the buttons use the buttons look and hover effect.
Also an image was used, that can be themed.

image

See #4302

Change-Id: Ida8cc19cefa0fc8ff17e5f9d0b14869a313a4d25

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -194,6 +194,23 @@ w2ui-toolbar {
 .w2ui-scroll-left,
 .w2ui-scroll-right {
 	z-index: 1001;
+	background-color: var(--color-background-dark) !important;
+	box-shadow: var(--color-box-shadow);
+	border: 1px solid var(--color-border-dark);
+	border-radius: 100%;
+	width: 32px;
+}
+.w2ui-scroll-left:hover,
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
This PR isn't finished. In addition to the new design, I suggest to arrange both buttons next to each other.

The existing solution has the disadvantage that when you want to scroll left/right you have to navigate with the mouse a lot.

I suggest that both buttons are right aligned so you can easier scroll left/right. At notebookbar you have space in the height, so you can have the buttons in two rows.

Andreas-Kainz:
@pedropintosilva any feedback here?

pedropintosilva:
> This PR isn't finished. In addition to the new design, I suggest to arrange both buttons next to each other.

I think it makes sense : ) and it would be a nice improvement, let me know once is finished or if you need any help here

> At notebookbar you have space in the height, so you can have the buttons in two rows.

Agreed :)

Andreas-Kainz:
> and maybe we can make it more "embed" in the interface so it does still attention from the buttons that trigger actions image

good idea!

Andreas-Kainz:
I add a new PR #4615

Issue #2720, opened 2021-06-29
Use `make`'s `-C` argument instead of `cd cypress_test` in the "a test failed" message


mikekaganski:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @mikekaganski

When a cypress test fails, a message is shown, like this:

>     CypressError: a test failed, please do one of the following:
> 
>     Run the failing test in headless mode:
>     	cd cypress_test && make check-mobile spec=impress/searchbar_spec.js
> 
>     Run the failing test with video recording:
>     	cd cypress_test && ENABLE_VIDEO_REC=1 make check-mobile spec=impress/searchbar_spec.js
> 
>     Open the failing test in the interactive test runner:
>     	cd cypress_test && make run-mobile spec=impress/searchbar_spec.js

The options suggect to `cd` to `cypress_test` first, then run the `make` command. This needs one also to `cd` back after the debugging session.

To simplify the task, one may use `make`'s `-C` command line argument [1], and have the current directory unchanged.
The task is, therefore, to simply update the script suggesting the debug options to use the `-C` instead of `cd`.

### Code pointers
`cypress_test/run_parallel.sh`

### Links to related documentation
[1] https://linux.die.net/man/1/make

PR #4304, opened 2022-02-26
Use make's -C argument instead of cd in cypress_test/run_parallel.sh

bayramcicek:
Signed-off-by: Bayram Çiçek <bayram.cicek@libreoffice.org>
Change-Id: I781070c32d89c329effbc233cf356fe983b85c1e

* Resolves: #2720
* Target version: master 

### Summary

- `cd cypress_test && make` is equal to `make -C cypress_test`
-  current directory will remain unchanged with `make -C`. 

### Checklist

### DIFF
--- a/cypress_test/run_parallel.sh
+++ b/cypress_test/run_parallel.sh
@@ -89,23 +89,23 @@ print_error() {
         echo -e "\n\
         CypressError: a test failed, please do one of the following:\n\n\
         Run the failing test in headless mode:\n\
-        \tcd cypress_test && make USER_INTERFACE=notebookbar check-${COMMAND} spec=${SPEC}\n" >> ${ERROR_LOG}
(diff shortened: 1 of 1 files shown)


pedropintosilva:
Thanks @bayramcicek ! that's a very good idea, I just wonder if it's ok for others using other systems since make -C* doesn't seem to be mentioned in the POSIX Programmer's Manual


----

`*`
```bash
-C dir, --directory=dir
            Change  to  directory dir before reading the makefiles or doing anything else.  If multiple -C options are specified, each is interpreted relative to the previous one: -C / -C etc is equivalent
            to -C /etc.  This is typically used with recursive invocations of make.
```

pedropintosilva:
Thanks @bayramcicek ! maybe we could update the `cypress_test/README` also (`cd cypress_test && make UPDATE_SCREENSHOT=true check-desktop spec=writer/help_dialog_update_spec.js` or add more info around `So you can
just execute it from the root folder or under the
cypress_test folder to run cypress tests only.`

bayramcicek:
@pedropintosilva: sure! . I'll work on it :)

Issue #4305, opened 2022-02-26
Sidebar Layout Harmonization


Andreas-Kainz:
The goal is to harmonize the sidebar layout

#### Global Sidebar Layout:
- Sections Headers use `--font-size-header`
- Separation between Sections have padding and a separator line

#### Section Layout:
- `align-left` are the Labels
- `align-right` are the widgets (commands)
- Labels use `-default-font-size`

PR #4312, opened 2022-02-26
Sidebar Layout Harmonization #4305

Andreas-Kainz:
The goal is to harmonize the sidebar layout

#### Global Sidebar Layout:

- Sections Headers use --font-size-header
- Separation between Sections have padding and a separator line

#### Section Layout:

- align-left are the Labels
- align-right are the widgets (commands)
- Labels use -default-font-size

Change-Id: Ib08ca0340d99dac7e7479bc43a04d069d550d85c

* Resolves: #4305 
* Target version: master 

#### Writer Default + Table
Table

#### Shape
shape


Andreas-Kainz:
@pedropintosilva @juliushaertl this patch isn't finished. I only checked for now writer and I know that there has to be some fixes in the default view of calc, I will do the fixes. What I need is feedback about the direction of this patch from you, cause it change the design of the sidebar. 

- the labels are the same than any other string. 
- the Section Headers use default font and height, ...
- the labels are vertical align with the header label
- ...

The goal is to have an clean sidebar. I also think to drop the horizontal line at each section.

pedropintosilva (changes requested):
understood, regarding the icon, it would be better to refrain from using hamburger menu icon because:
- it does not really trigger menus, it opens dialogs with more options
- on top of that in some cases we might already have hamburger menus in the screen (ipad) and it will look strange to have the same type of icon for very different things

juliushaertl:
Looks nice I think. Regarding to the icon I agree with @pedropintosilva but also find the different icons currently un use not that ideal in terms of giving the user a quick idea on what a click would trigger. Maybe something like a preferences icon could be more suitable there as the type of preferences is already indicated by the heading text? Something like https://thenounproject.com/search/icons/?iconspage=1&q=preferences

Andreas-Kainz:
About the icon. I prefer the different icons like it was done by nextcloud office. It's great and the hamburger menu was only an backup when there is no specific icon.

Andreas-Kainz:
if you are ok with this PR I have to fix the conflict.

Issue #4316, opened 2022-02-27
Many inconsistent and problematic icons in sidebar


rizmut:
![gambar](https://user-images.githubusercontent.com/5697991/155879510-cdb42046-1243-42e6-aee9-8418d3314907.png)

As you can see above some icons are not well balanced. Here I mention one by one
1. Italic icon has thinner line than other font based icon
2. Shadowed, increase and Decrease icons are larger than the rest of icons
3. Font color is way too hidden by the color area below
4. The arrow in Right-To-Left and Left-To-Right a little bit hard to see

PR #4317, opened 2022-02-27
Update icons in sidebar

rizmut:
Signed-off-by: Rizal Muttaqin <riz_17_oke@yahoo.co.id>
Change-Id: I4d44746709049ced93fba5e163b5446546f56fd2

* Resolves: #4316 
* Target version: master 

### Summary
This PR provides updated icons for sidebar

**Before:**
gambar

**After:**
gambar

### Checklist


Andreas-Kainz:
The icons are all fine and come from core.

However the strikethrough icon I prefer the older one where I think it's better to see the strikethrough. 

About the Shadow icon I also liked the older one better. But you are correct it should have a blue shadow. However do you have an idea of an shadow icon which look more balanced and easier to understood?

pedropintosilva:
I agree with @Andreas-Kainz maybe it would help if in the case strikethrough we  would have white space surrounding the blue line (as it happens in the previous version); and in the shadow case I'm a bit neutral (I see advantages either way).

pedropintosilva:
but I would merge this and then wait for a follow up from @rizmut : )

pedropintosilva:
it seems there was a problem while getting online dependencies within GH actions, re-running them

PR #4328, opened 2022-03-01
fix copy annotations on shared-view doc causes uncaught exception

merttumer:
Change-Id: I281fcb3a613166bc9be4b3bc09dbe0b28fdca6bc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix :)

PR #4342, opened 2022-03-03
Add a remote font configuration thing

tml1024:
The coolwsd.xml file can contain a URI of a JSON file on some server
that contains URIs of fonts. These fonts are downloaded. Just like the
remote configuration thing, the URIs are checked once a minute and the
JSON or the fonts are re-downloaded if their contents change.

If a font has been removed from the JSON then the corresponding
downloaded file is removed, too.

It is still unclear how exactly handle this with LibreOffice core and
Fontconfig.

Change-Id: If78058ddff5ed05c7a82d7ea465a7a414fd0d861

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -120,6 +120,8 @@ class Document;
 static Document *singletonDocument = nullptr;
 #endif
 
+_LibreOfficeKit* loKitPtr = nullptr;
+
 /// Used for test code to accelerating waiting until idle and to
 /// flush sockets with a 'processtoidle' -> 'idle' reply.
 static std::chrono::steady_clock::time_point ProcessToIdleDeadline;
@@ -2446,8 +2448,6 @@ void copyCertificateDatabaseToTmp(Poco::Path const& jailPath)
 #endif
 }
 
-
-
 void lokit_main(
 #if !MOBILEAPP
                 const std::string& childRoot,
@@ -2967,13 +2967,12 @@ bool globalPreinit(const std::string &loTemplate)
         }
     }
 
-    LokHookPreInit* preInit = reinterpret_cast<LokHookPreInit *>(dlsym(handle, "lok_preinit"));
+    LokHookPreInit2* preInit = reinterpret_cast<LokHookPreInit2 *>(dlsym(handle, "lok_preinit_2"));
     if (!preInit)
     {
-        LOG_FTL("No lok_preinit symbol in " << loadedLibrary << ": " << dlerror());
+        LOG_FTL("No lok_preinit_2 symbol in " << loadedLibrary << ": " << dlerror());
         return false;
     }
-
     initFunction = reinterpret_cast<LokHookFunction2 *>(dlsym(handle, "libreofficekit_hook_2"));
     if (!initFunction)
     {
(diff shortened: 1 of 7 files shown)


kit/Kit.cpp
```diff
@@ -2576,6 +2577,17 @@ void lokit_main(
                     return false;
                 }
 
+                // Mount the directory for fonts downloaded from servers, if configured
+                const std::string jailTmpFontDir = Poco::Path(jailPath, tmpFontDir.substr(1)).toString();
+                LOG_INF("Mounting temp font dir " << tmpFontDir << " -> " << jailTmpFontDir);
+                if (!JailUtil::bind(tmpFontDir, jailTmpFontDir)
+                    || !JailUtil::remountReadonly(tmpFontDir, jailTmpFontDir))
```
mmeeks, line 2584: We worked quite hard to get the number of bind mounts per jail down to a small number - and need to do more work to get it lower still really. Is it really unavoidable that we have to have a new bind mount for fonts ?

wsd/COOLWSD.cpp
```diff
@@ -1278,6 +1285,113 @@ class RemoteConfigPoll : public RemoteJSONPoll
...
+
+                const std::string fontFile = COOLWSD::TmpFontDir + "/" + Util::encodeId(Util::rng::getNext()) + ".ttf";
+                std::ofstream fontStream(fontFile);
+                fontStream.write(body.data(), body.size());
+                if (fontStream.good())
+                {
+                    LOG_DBG("Got " << body.size() << " bytes for " << it.first << " and wrote to " << fontFile);
+
+                    it.second.pathName = fontFile;
+                    COOLWSD::sendMessageToForKit("addfont " + fontFile);
```
mmeeks, line 1374: I like the addfont message =)

kit/Kit.cpp
```diff
@@ -2962,12 +2974,13 @@ bool globalPreinit(const std::string &loTemplate)
         }
     }
 
-    LokHookPreInit* preInit = reinterpret_cast<LokHookPreInit *>(dlsym(handle, "lok_preinit"));
+    LokHookPreInit2* preInit = reinterpret_cast<LokHookPreInit2 *>(dlsym(handle, "lok_preinit_2"));
     if (!preInit)
     {
-        LOG_FTL("No lok_preinit symbol in " << loadedLibrary << ": " << dlerror());
+        LOG_FTL("No lok_preinit_2 symbol in " << loadedLibrary << ": " << dlerror());
```
mmeeks, line 2980: Hmm - seems like the preinit2 thing looks sensible. I would suggest though that we add a new C wrapper to the LibreOfficeKitInit.h - or somesuch that uses it (?) bit of a pain changing these methods. I might also be inclined if we are fiddling with the signature there to add some const char *options parameter or something for future use - to avoid the next inevitable lok_preinit_3 biting us too soon ;-)

mmeeks (commented):
Looks good to me - thanks Tor - a few comments in-line.

kit/Kit.cpp
tml1024: Well, the point is that I need to font files to be accessible using the same pathname both in the non-jailed ForKit process (which passes the font pathname on to the OutputDevice::AddTempDevFont() function) and in a jailed Kit process (which opens and mmaps the font file using the same pathname in the FreetypeFontFile::Map() function).

But sure, I can attempt to modify the core code so that I can pass a separate pathname to be used when the font file is opened later.

mmeeks: Makes perfect sense - I can see how that would be easier but hoarding the bind-mounts makes sense. Then again - bind mounts are not really the right thing to do we should really be using: https://man7.org/linux/man-pages/man7/namespaces.7.html and (I think) creating a new namespace per kit process - particularly becase:

 Namespace lifetime
       Absent any other factors, a namespace is automatically torn down
       when the last process in the namespace terminates or leaves the
       namespace. 

Cleanup is much nicer, and it doesn't clutter your 'mount' command something horribly ;-) and possibly we could do these things hierarchically or something (or something ;-) really we need a container whizz to advise us - I've been meaning to talk to Lennart or someone at a FOSDEM in-person but not got to one recently.

tml1024:
Using separate pathnames for the font files in ForKit when calling AddTempDevFont() and then in Kit, inside the jail, when/if the font is actually used, does seem to be way too complicated, and possibly depends on FontConfig internals.

I will instead attempt to do it so that the downloaded fonts are linked (or copied) into each jail and each Kit process calls AddTempDevFont() on their local copy of each font.

tml1024:
Nah, I did not do what I said earlier after all. Just some more trickery with symlinks in the systemplate was needed. Now no new bind mount is needed, and the downloaded fonts appear as the same pathnames both in ForKit and KIt. Thanks to @mmeeks for pointing me in the right direction.

tml1024:
Now I can't figure out how to turn off the "Draft" setting.

tml1024:
(Ah, had to click the "Ready for review" button to get out of the Draft state.)

mmeeks (approved):
Looks great to me - thanks Tor - lets get it in when CI likes it =)

tml1024:
Now it should compile, after I merged the required core bits.

tml1024:
Hmm, apparently those checks build against tarballs of core, so the compilations can't succeed until new ones have been made. And I didn't cherry-pick to the 22.05 branch yet. Will do.

PR #4349, opened 2022-03-04
Check if needed anchor exist before use

eszkadev:
in one document I have it has "21" instead of "16"


merttumer (approved):
RTL anchor to be added as a follow-up. This fixes the problem right now.

PR #4355, opened 2022-03-08
calcRTL: dark-overlay incorrectly covers the chart

dennisfrancis:
Dark overlays like any other overlay need regular document coordinates.
But in calc-rtl mode, charts (like shapes) have negative x document
coordinate internal representation.

Before this fix, the CDarkOverlay used to get negative x coordinates
hence the positions of the four dark rectangles were wrong.

Change-Id: Iea3202e2a0c0594cd8f894ee96123a0281cf2f38

* Resolves: # 
* Target version: master 

### Checklist


merttumer (approved):
Works as expected, thanks

PR #4365, opened 2022-03-09
Remove extra white space & update gear element

rizmut:
Signed-off-by: Rizal Muttaqin <riz_17_oke@yahoo.co.id>
Change-Id: I6e2293ca096ba45ce0dc1740ba2fc80c900e92ca

* Target version: master


Andreas-Kainz:
Thanks for removing the white background. I don't know if you can submit both PR's or you get an merge conflict.

Could you close both PR's and make a new PR with the updated icons? Or please ask someone how knows better how git work for a review.

rizmut:
> Thanks for removing the white background. I don't know if you can submit both PR's or you get an merge conflict.
> 

Currently the icons has different approach: ones have  empty space gap, others have white space gap. Which one we should use?
> Could you close both PR's and make a new PR with the updated icons? Or please ask someone how knows better how git work for a review.

The previous PR has been merged

Andreas-Kainz:
the white background was added cause **if** we could style icons, it's better. As long as the icons can't be styled (with .css), I would use the icons from LibO otherwise we have more work with no benefit. 

So lets take the icons from LibO core.

PR #4368, opened 2022-03-09
calcRTL: chart selection's svg does not match the tile

dennisfrancis:
The svg sent from core does not exactly match the corresponding contents
in tile of the shape - especially the legend texts. This commit does not
attempt to fix that.

Before the commit e64ff372fc54e42fb1337ac3b76200c021018c96, the svg
embedded image was always hidden when a shape/chart/ole is merely
selected and it is only shown while the shape is being resized or moved.
Because of the strict inequality used with getAttribute('opacity') this
condition is always true at least in Chrome as getAttribute() returns a
string.  Using a non-strict inequality operator results in the intended
behaviour and also hides the svg preview when chart/shape is not dragged
or resized.

Change-Id: I0aef873e392b0959e93057e6399e8f77cf70042c

* Resolves: # 
* Target version: master 

### Checklist


browser/src/layer/vector/SVGGroup.js
```diff
@@ -258,7 +258,7 @@ L.SVGGroup = L.Layer.extend({
 	_hasVisibleEmbeddedSVG: function () {
 		var result = false;
 		this._forEachSVGNode(function (svgNode) {
-			if (svgNode.getAttribute('opacity') !== 0)
+			if (svgNode.getAttribute('opacity') != 0)
```
eszkadev, line 261: hmm, maybe better idea would be to use strict inequality after cast - parseInt()  ?

eszkadev (changes requested):

browser/src/layer/vector/SVGGroup.js
eszkadev: `try ... catch` for parsing error inside loop will be then needed probably.

dennisfrancis: parseInt does not throw exceptions (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt). On erroneous cases it returns NaN which would fail strict comparison checks.

eszkadev (approved):

PR #4370, opened 2022-03-09
Editing of document templates does not work well currently

timar:
Change-Id: I67c57be272f3b4818f9bcd3a9b3b6df61846cbdb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
The cypress tests probably fail because of the readonly-ness of the template file types. @Rash419 could you please suggest how to fix that?

Rash419:
I will take a look at it tomorrow

Rash419:
> The cypress tests probably fail because of the readonly-ness of the template file types. @Rash419 could you please suggest how to fix that?

I will take a look tomorrow thanks

PR #4378, opened 2022-03-10
cypress: removed repeated tests and enabled notebookbar tests

Rash419:
Change-Id: I2c31b7433559d776dbde179c2659e23d8b104d59

* Target version: master


cypress_test/integration_tests/desktop/writer/top_toolbar_spec.js
```diff
@@ -587,4 +603,77 @@ describe('Top toolbar tests.', function() {
 		}).then(() => expect(data).to.deep.eq(expectedData));
 
 	});
+
+	it('Apply superscript.', function() {
```
lpranam, line 607: if we are testing the top toolbar shouldn't we use buttons from the top toolbar for this and all the tests below?

cypress_test/plugins/index.js
```diff
@@ -97,7 +96,7 @@ function pickTests(filename, foundTests) {
 	}
 
 	if (process.env.USER_INTERFACE === 'notebookbar') {
-		if (!isNotebookbarTest(filename,whitelists.notebookbarOnlyList)) {
+		if (!isNotebookbarTest(filename,blacklists.notebookbarOnlyList)) {
```
lpranam, line 99: it is preferred to not use words like blacklist/whitelist: https://tools.ietf.org/id/draft-knodel-terminology-00.html

lpranam (changes requested):

cypress_test/integration_tests/desktop/writer/top_toolbar_spec.js
Rash419: superscript and subscripts buttons are not there for classic menu

PR #4379, opened 2022-03-10
fix empty selection handles appear on calc

merttumer:
Change-Id: I56a263711bd034ff1a56357eb5efe713fe076db3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -6993,6 +6993,8 @@ L.MessageStore = L.Class.extend({
 			return;
 		}
 
+		this._cleanUpSelectionMessages(this._ownMessages);
+
 		var ownMessages = this._ownMessages;
 		Object.keys(this._ownMessages).forEach(function (msgType) {
 			callback(ownMessages[msgType]);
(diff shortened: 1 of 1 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -7002,5 +7004,17 @@ L.MessageStore = L.Class.extend({
...
+	_cleanUpSelectionMessages: function(messages) {
+		// must be called only from _replayPrintTwipsMsg !!
+		// check if textselection is empty
+		// if it is, we need to handle textselectionstart and textselectionend
+		// otherwise we get handles without selection and they also may appear in the wrong cell
+		// but it is also reproducible on the same cell too. e.g. selection handles without selection
+		if (!messages && !messages['textselection'] && messages['textselection'] !== 'textselection: ')
+			return;
+		messages['textselectionstart'] = 'textselectionstart: ';
+		messages['textselectionend'] = 'textselectionend: ';
```
eszkadev, line 7018: I'm wondering if we cannot do something in code which handles these messages?
(so we will keep message store clean from message modifications)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -7002,5 +7004,17 @@ L.MessageStore = L.Class.extend({
 		Object.keys(othersMessages).forEach(function (msgType) {
 			othersMessages[msgType].forEach(callback);
 		});
+	},
+
+	_cleanUpSelectionMessages: function(messages) {
+		// must be called only from _replayPrintTwipsMsg !!
```
eszkadev, line 7010: is that comment valid?
I see it's called just few lines above not in any `_replayPrintTwipsMsg`
(or do you mean that code above is called from `_replayPrintTwipsMsg` - I didn't check the stack...)

eszkadev (approved):
Works correctly, fixing empty selections.
Just added 2 questions, if it's not needed to be changed please push :)

browser/src/layer/tile/CanvasTileLayer.js
merttumer: the messages is bind to _onMessage through forEach method and it is handled there only. Unfortunately we cannot check the state there because same function is used when we actually get the message from the API as well and we may not be sure about it.

browser/src/layer/tile/CanvasTileLayer.js
merttumer: Yes, forEach is part of _replayPrintTwipsMsg :)

PR #4381, opened 2022-03-11
Fixes to save/upload corner-cases

Ashod:
- wsd: wait longer when restarting forkit
- wsd: stop loading when low on disk space
- wsd: gracefully close the document when disconnected from Kit
- wsd: last modified timestamp of the upload file is not now


kendy (approved):
OK, let's get it in, thank you!

PR #4388, opened 2022-03-11
Can't open password-protected files

dennisfrancis:
* Resolves: # 
* Target version: master 

### Summary
Reproduction:

    Opening a password protected document gives an error corresponding to
    'faileddocloading' and then a blank file is displayed.
    
Fix:

    Since 0b76383346887fa6fbf732e680a79a8d5e38411b, when Document::load()
    fails without any exception, onLoad() sends
    
    "error: cmd=load kind=faileddocloading"
    
    to the client. This is a problem when load() fails just because the
    client did not provide password for a password protected document.
    Besides for all "exception free cases" load() already sends the correct
    error message(s) to the client. So this patch also avoids sending
    duplicate error messages.

This PR also adds a check for this case in the relevant unit test.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitPasswordProtected.cpp
+++ b/test/UnitPasswordProtected.cpp
@@ -54,7 +54,7 @@ UnitBase::TestResult UnitPasswordProtected::testPasswordProtectedDocumentWithout
         // Send a load request without password first
         helpers::sendTextFrame(socket, "load url=" + documentURL);
 
-        const auto response = helpers::getResponseString(socket, "error:", testname);
+        auto response = helpers::getResponseString(socket, "error:", testname);
         StringVector tokens(Util::tokenize(response, ' '));
         LOK_ASSERT_EQUAL(static_cast<size_t>(3), tokens.size());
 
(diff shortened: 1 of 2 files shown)


test/UnitPasswordProtected.cpp
```diff
@@ -64,6 +64,9 @@ UnitBase::TestResult UnitPasswordProtected::testPasswordProtectedDocumentWithout
         COOLProtocol::getTokenString(tokens[2], "kind", errorKind);
         LOK_ASSERT_EQUAL(std::string("load"), errorCommand);
         LOK_ASSERT_EQUAL(std::string("passwordrequired:to-view"), errorKind);
+
+        response = helpers::getResponseString(socket, "error:", testname);
+        LOK_ASSERT_MESSAGE("Unexpected second error message", !response.size());
```
Ashod, line 69: Perhaps we could also compare the full error message we expect? Because in some cases there could be a different failure, which should break this test.

Ashod (approved):
Thanks @dennisfrancis, looks good. (I wouldn't have removed the braces around `return false;`, since they were there already, but that's just a stylistic preference.)

PR #4404, opened 2022-03-14
JS btn: fix disabled status

pedropintosilva:
With changes from 76121778cac405508839484bae120dfa78a70a61
- We started to overwrite default user agent styles and with it
decreasing the difference between disabled and enabled buttons making it
very difficult to distinguish between the two.
- Also border width was set to different values (disabled btn getting it
from our stylesheet and enabled from browsers styles) causing the btn to
"jump" in size and position when changing between states

Fix those and add new css vars for btns so to avoid mistakes
in the future

Change-Id: I3ec865ecbbff44c90357d1d6de44a2e6f633bc3f

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -4,10 +4,18 @@
 	text-transform: full-size-kana;
 }
 
+button.jsdialog {
+	color: var(--color-text);
+	background-color: var(--color-background);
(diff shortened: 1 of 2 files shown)


pedropintosilva:
before : disabled state: 
image

PR #4415, opened 2022-03-15
calc: fix cells become empty after closing sidebar

merttumer:
this patch not only fixes the issue in the title
also fixes similar problems such as lets say the
browser page is not maximized and then it is decided
to be maximized. This operation normally does trigger redrawing
but it only draws tiles that are in the cache already and does not
trigger tile invalidation and cells can stay empty until something
retriggers tile invalidation again

Change-Id: I5981aa36c77a87626809c92e72e1cd6263826953

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix :)

PR #4419, opened 2022-03-15
Menubar add DrawText and VerticalText commands

Andreas-Kainz:
Change-Id: I6d07df1a3a97b9e896c883e87803d2f2b986ee56


Andreas-Kainz:
In Notebookbar mode this two commands are available, so I suggest to have them in classic mode also?

@pedropintosilva @eszkadev

PR #4420, opened 2022-03-16
mobile: fix fullscreen-presentation does not work on readonly mode

merttumer:
we send the slideshow buttons reference to onClick function
in that item.disabled is false but later we reassign it from
w2ui get function and then it becomes disabled true on readonlymode
The logic is incorrect because onClick function does not respect
wheter the item is passed or not and tries to reassign it anyway.

Change-Id: I025e26c455a7673a002f894f49452b360550b178

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -53,7 +53,8 @@ function onClose() {
 	}
 }
 
-function onClick(e, id, item) {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Fixes the issue

PR #4422, opened 2022-03-16
wsd: reduce polling when blocked on ssl handshake

Ashod:
While SSL is handshaking, there can be no general
application data communication. During that early
stage of connecting we have data to send (the
request, headers, etc.) and so we poll on POLLOUT.
Naturally, we also always want to poll on POLLIN,
because we can never know when there is data to
read (especially true for web-sockets).

The problem is when SSL will not send data just
yet because it is handshaking. It is typically
waiting for handshake negotiation data to read,
so when we POLLOUT, poll immediately returns, but
writing (via SSL_write) fails with WANTS_READ
error. This goes on in a busy-loop until the
negotiation data is available for read and the
handshake is completed. Very inefficient.

The solution is to poll on whatever SSL needs
during the handshake, exclusively. Once the
handshake is complete, we poll on whatever we
need. However, SSL can renegotiate at any time,
so we also merge with what it needs.

In addition, we avoid the unnecessary read when
poll doesn't give us POLLIN in revents, since the
read will more likely than not fail (except in
the rare case when data becomes available in the
interim). Notice that SSL_read will return
SSL_WANTS_READ when there is no data, which
is misleading (since SSL isn't in need of data to
read at all, nor are we, for that matter).
Best not to do noisy reads unnecessarily.

Change-Id: I6a7ed7d871ed257b30062cc720a8b8c7acbab3b7

### DIFF
--- a/net/SslSocket.hpp
+++ b/net/SslSocket.hpp
@@ -11,6 +11,7 @@
 #include <sstream>
 #include <string>
 
+#include "Common.hpp"
 #include "Ssl.hpp"
 #include "Socket.hpp"
 
@@ -147,6 +148,28 @@ public:
                       int64_t & timeoutMaxMicroS) override
     {
         ASSERT_CORRECT_SOCKET_THREAD(this);
+
+        // We cannot communicate with a client when SSL wants
+        // to negotiate the handshake. Here, we give priority to
+        // SSL's needs. Only when SSL is done negotiating
+        // (i.e. wants neither to read, nor to write) do we see
+        // what data activity we have.
+        if (EnableExperimental && _doHandshake)
+        {
+            if (_sslWantsTo == SslWantsTo::Write)
+            {
+                return POLLOUT;
+            }
+
+            if (_sslWantsTo == SslWantsTo::Read)
+            {
+                return POLLIN;
+            }
+
+            // The first call will have SslWantsTo::Neither and
+            // we will fall-through and call the base.
+        }
+
         int events = StreamSocket::getPollEvents(now, timeoutMaxMicroS); // Default to base.
         if (_sslWantsTo == SslWantsTo::Write) // If OpenSSL wants to write (and we don't).
             events |= POLLOUT;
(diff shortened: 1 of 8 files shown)


net/SslSocket.hpp
```diff
@@ -188,11 +207,13 @@ class SslStreamSocket final : public StreamSocket
                     return rc != 0;
             }
 
-            _doHandshake = false;
-
             if (rc == 1)
             {
                 // Successful handshake; TLS/SSL connection established.
+                _doHandshake = false;
+                _sslWantsTo = SslWantsTo::Neither; // No special needs until we are told otherwise.
```
Ashod, line 214: If we don't reset, we may do an extra `POLLOUT` when not needed, if we ended the handshake with WANTS_WRITE. This will get cleared out after the first `SSL_read` or `SSL_write`. But best to reset and have a clean start (though almost always we would have data to send, and we would therefore `POLLOUT`, first thing after the handshake completes, anyway).

net/SslSocket.hpp
```diff
@@ -147,6 +147,28 @@ class SslStreamSocket final : public StreamSocket
                       int64_t & timeoutMaxMicroS) override
     {
         ASSERT_CORRECT_SOCKET_THREAD(this);
+
+        // We cannot communicate with a client when SSL wants
+        // to negotiate the handshake. Here, we give priority to
+        // SSL's needs. Only when SSL is done negotiating
+        // (i.e. wants neither to read, nor to write) do we see
+        // what data activity we have.
+        if (_doHandshake)
```
mmeeks, line 156: can we put a #if TESTING type thing around this - so that we isolate the functional change only for the 22.05 branch ? - we should have a configure switch that sets that and ask Andras to have that on for CODE (I guess) ? =)

net/SslSocket.hpp
mmeeks: Looks sensible enough - proof is in the pudding; lets add that conditional & get it in =)

mmeeks (commented):
Lets have a testing configure option we use for 22.05 here =)

PR #4423, opened 2022-03-16
infobar: show message every five days

hcvcastro:
Change-Id: Iedf6aff2f026f4530bfed06cd4bf5e7acae15a50

### DIFF
--- a/browser/src/map/handler/Map.Infobar.js
+++ b/browser/src/map/handler/Map.Infobar.js
@@ -16,7 +16,19 @@ L.Map.Infobar = L.Handler.extend({
 
 		if (viewInfo && !this.enabled() && viewInfo.userextrainfo &&
 		    viewInfo.userextrainfo.is_admin) {
-			this.enable();
+			var laterDate = new Date();
+			var currentDate = new Date();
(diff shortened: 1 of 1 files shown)


mmeeks:
Looks wonderful to me - +1 etc. - like the solution; every 5 days seems reasonable. Thanks !
Would love someone to test it =)

browser/src/map/handler/Map.Infobar.js
```diff
@@ -16,7 +16,19 @@ L.Map.Infobar = L.Handler.extend({
 
 		if (viewInfo && !this.enabled() && viewInfo.userextrainfo &&
 		    viewInfo.userextrainfo.is_admin) {
-			this.enable();
+			var laterDate = new Date();
+			var currentDate = new Date();
+			var timeValue = window.localStorage.getItem('InfoBarLaterDate');
+			if (!timeValue || isNaN(timeValue)) {
+				/* + 5 seconds */
+				laterDate.setTime(currentDate.getTime() + 5000);
```
gokaysatir, line 24: Maybe we need to make this -5000 not +5000 so
"if (currentDate > laterDate)" will work :))
Thanks for the fix :) we just need to change that "+" to "-" i think.

Issue #3184, opened 2021-09-03
Classic mode: Add back removed space between selectors and icons


pedropintosilva:
**Describe the bug**
In the classic mode there is a new regression that effectively removed the essential space between dropdowns (selectors) and the following icons: 
* Clear direct formatting icon
* bold  icon

**To Reproduce**
1. Run it locally with ```make run```
2. Make sure you set the respective interface mode (only needed if you have changed before since it is the default): 
  * in your ```loolwsd.xml``` change the ```<user_interface>``` like so:
```xml
    <user_interface>
      <mode type="string" desc="Controls the user interface style (classic|notebookbar)" default="classic">classic</mode>
    </user_interface>
```
3. Notice the lack of space
4. Get acquainted with git commands and find where the regression comes from (from which commit)
5. Add back the removed spaces 

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/65948705/132022537-5bed5847-f0a1-44cc-8bf1-3c9aaed747c2.png)

**Desktop (please complete the following information)**
 - branch: master

pedropintosilva:
pointer:
```bash
git log --grep="toolbar.*.breaks" -i
```

codewithvk:
@pedropintosilva can I take this issue?

pedropintosilva:
@codewithvk yes ! :)

pedropintosilva:
@codewithvk  : are you still working on this? If so please let me know and I will reassigned that to you :)

Andreas-Kainz:
> pointer:
> 
> ```shell
> git log --grep="toolbar.*.breaks" -i
> ```

I see two PR's but both aren't related to .css changes.

Andreas-Kainz:
I can add margin-left and margin-right of 2px but than the toolbar commands need way more space see

#### before
![image](https://user-images.githubusercontent.com/8517736/154795047-3634d574-ff7a-48cb-904f-23f22cf63ccb.png)

#### after
![image](https://user-images.githubusercontent.com/8517736/154795037-6aa8197e-5de6-4fa2-9d32-008e8bb216fb.png)

In the sidebar there is a `margin-right: 5px` setting

pedropintosilva:
https://github.com/CollaboraOnline/online/pull/4438

PR #4425, opened 2022-03-17
Add back removed breaks from TopToolbar, see #3184

kcmaxwell:
Breaks were removed in commit df925222a2c4ce94024b8df7eeb0d014e2d82271

Signed-off-by: Kristopher Maxwell <kristophermaxwellc@gmail.com>
Change-Id: I376207d654ccbe98fc1d1b4b84363748190ab646

* Resolves: # 3184
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
Thanks @kcmaxwell ! I have re-triggered the cypress test since it failed the build midway

pedropintosilva:
@kcmaxwell I have tested your patch but I think this was removed in the past because the drop-down menus already count as visual dividers making the additional bezels unecessary:

image

kcmaxwell:
Ok, would you still like to have the space between the dropdowns and the other icons, just no bezels?
I could do that with the CSS margin property on the dropdowns.
Let me know if that sounds good or if you want something different.

pedropintosilva:
Good idea : ) I think that would be the best, retaining white space but without the visual clutter of those breaks (since the drop-down menus being already different visually from the surrounding elements already do that role)

pedropintosilva:
maybe even better would be to add something like `visible: false` on the JS side that would then set `visibility: hidden` instead of having loosely breaks in the CSS set to hidden and then people forget about them : p

---

Note: The `hidden: true` set `display` to `none` in the CSS side

pedropintosilva:
or if you prefer or easier to add the .hidden class which we already have on the css side : )

pedropintosilva (approved):
You what I will merge this : ) and then we can make them hidden since you already waited some time

PR #4431, opened 2022-03-18
co-22.05: mobile writer/table_properties_spec.js

eszkadev:
- use .unoXXX class instead of id
- check only if height was changed on Optimal Row Height command


Rash419 (approved):
Cypress part looks good, thanks : )

PR #4433, opened 2022-03-18
image operation: make sure embedded svgs are shown in writer only

lpranam:
Change-Id: I1b3cfefdd6abb7f81ff741eee4edab7e349c932d

* Target version: master 

### Summary
fixed regression caused by https://github.com/CollaboraOnline/online/commit/1c86fd2ff869d36156bb97b3f44fc0cf7fec037c
regression: while we move textbox in impress previews would not go away

### Checklist


eszkadev (approved):
Writer case works

PR #4436, opened 2022-03-18
jsdialog: find also controls with modified id

eszkadev:
In some cases we add 'table-' prefix for toolboxes due to
some old requirements in the mobile-wizard. We need to find these
modified ids in a fallback on widget update.

In the future we should get rid of 'table-' prefix.

This fixes missing numbering/bullets items in the impress sidebar


pedropintosilva (approved):
Tested on desktop and mobile, it works:
- icons related to the lists are now properly displayed in the sidebar ( including when selecting a text box on impress)

pedropintosilva:
Thanks!

PR #4438, opened 2022-03-18
Allow items to be shown invisible on the TopToolbar

pedropintosilva:
This is important because we in many places might want to still have the
respective element's space but without actually showing it.
Example: divider's space without that being represented by a bezel

Change-Id: Ib58cc9564144c4ad48fa240878a3654add5d4d03

### DIFF
--- a/browser/src/control/Control.TopToolbar.js
+++ b/browser/src/control/Control.TopToolbar.js
@@ -84,6 +84,9 @@ L.Control.TopToolbar = L.Control.extend({
 	// tablet:true means show it in normal Online from a tablet browser, and in a mobile app on a tablet
 	// tablet:false means hide it in normal Online used from a tablet browser, and in a mobile app on a tablet
 
+	// hidden means display:none
+	// invisible means visibility:hidden
+
 	getToolItems: function() {
 		var that = this;
 		return [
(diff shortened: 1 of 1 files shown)


pedropintosilva:
@kcmaxwell what do you think?

pedropintosilva:
(also @eszkadev are you ok with this change? )

kcmaxwell:
That looks good to me. Don't forget to close issue #3184 after you've merged this pull request.

browser/src/control/Control.TopToolbar.js
```diff
@@ -275,6 +278,11 @@ L.Control.TopToolbar = L.Control.extend({
 				window.hideTooltip(this, e.target);
 			},
 			onRefresh: function(event) {
+				if (event.item && event.item.invisible == true) {
+					var toolItem = $(this.box).find('#tb_'+ this.name +'_item_'+ w2utils.escapeId(event.item.id));
+					toolItem.css('visibility', 'hidden');
+				}
+
 				if ((event.target === 'styles' || event.target === 'fonts' || event.target === 'fontsizes') && event.item) {
 					var toolItem = $(this.box).find('#tb_'+ this.name +'_item_'+ w2utils.escapeId(event.item.id));
```
eszkadev, line 287: Does it run without errors? I think there is `var toolItem` redeclared. `var` should be used only in the first occurence in the function according to our checkers.

eszkadev: also we could have only one place with getting this item (now we duplicated this line of code).
Like:
```
if (event.item)
    var toolItem = $(this.box).find('#tb_'+ this.name +'_item_'+ w2utils.escapeId(event.item.id));
 if (toolItem && event.item.invisible == true)
    ....
 if (toolItem && (event.target === 'styles' || event.target === 'fonts' || event.target === 'fontsizes'))
    ....
 ```

File: browser/src/control/Control.TopToolbar.js [hunk 2]
```diff
@@ -275,10 +278,15 @@ L.Control.TopToolbar = L.Control.extend({
 				window.hideTooltip(this, e.target);
 			},
 			onRefresh: function(event) {
-				if ((event.target === 'styles' || event.target === 'fonts' || event.target === 'fontsizes') && event.item) {
+				if ((event.item && (event.target === 'styles' || event.target === 'fonts' || event.target === 'fontsizes'))
+					|| (event.item && event.item.invisible == true)) {
```
eszkadev, line 282: it can be simplified to:
`if (event.item && (event.target === 'styles' || event.target === 'fonts' || event.target === 'fontsizes' || event.item.invisible)`

eszkadev (dismissed):
Just a nitpick: we can simplify conditions, but I approve

browser/src/control/Control.TopToolbar.js
eszkadev: also better to use '===' (strict equal - with type checking) so it will not convert other type values like '0' (string) to true

pedropintosilva: Good pointer and much appreciated!

pedropintosilva:
> Just a nitpick: we can simplify conditions, but I approve

Thanks for reviewing, I have simplified it and made sure to use strict equal as well.

eszkadev (approved):
Now looks great :)

PR #4440, opened 2022-03-20
util: removed duplicate string trimming function

lpranam:
Change-Id: Ia150975467d1df8a9d05c8543bcb5d7369d29099

* Target version: master 

### Checklist

### DIFF
--- a/common/CommandControl.cpp
+++ b/common/CommandControl.cpp
@@ -35,9 +35,7 @@ void LockManager::generateLockedCommandList()
     std::string command;
     for (std::size_t i = 0; i < commandList.size(); i++)
     {
-        // just an extra check to make sure any whitespace does not sniff in command
-        // or else command will not be recognized
-        command = Util::trim_whitespace(commandList[i]);
+        command = commandList[i];
         if (!command.empty())
         {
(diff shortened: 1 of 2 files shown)


common/CommandControl.cpp
```diff
@@ -35,9 +35,7 @@ void LockManager::generateLockedCommandList()
     std::string command;
     for (std::size_t i = 0; i < commandList.size(); i++)
     {
-        // just an extra check to make sure any whitespace does not sniff in command
-        // or else command will not be recognized
-        command = Util::trim_whitespace(commandList[i]);
```
Ashod, line 40: Should trimming still happen? Why not use `Util::trimmed()` ?

Ashod (commented):
Thanks @lpranam. Why not keep the trimming?

common/CommandControl.cpp
lpranam: I checked and `tokenize` does not include any space so its already trimmed no need to check again.

eszkadev:
I will retrigger build as ssl cert is now ok for Jenkins and gtihub check will be able to work :)

Ashod (approved):

PR #4442, opened 2022-03-21
SSL logging and other improvements

Ashod:
- browser: protect against undefined keyEventFn
- wsd: always include the BIO errors in SSL logs
- wsd: trace poll events and revents flags arround ppoll
- wsd: test: correctly show the failing code in the assertion message
- wsd: minor improvements to StringVector::equals overloads
- wsd: exclude deprecated openssl API

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -735,8 +735,8 @@ public:
     {
         if (newSocket)
         {
-            LOG_DBG("Inserting socket #" << newSocket->getFD() << ", address "
-                                         << newSocket->clientAddress() << ", into " << _name);
+            LOG_DBG("Inserting socket #" << newSocket->getFD() << ", address ["
+                                         << newSocket->clientAddress() << "], into " << _name);
             // sockets in transit are un-owned.
             newSocket->resetThreadOwner();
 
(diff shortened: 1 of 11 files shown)


net/Socket.hpp
```diff
@@ -1117,15 +1117,18 @@ class StreamSocket : public Socket,
...
-                LOG_TRC('#' << getFD() << ": Read incoming data " << len << " bytes in addition to "
-                            << _inBuffer.size() << " buffered bytes ("
-                            << Util::symbolicErrno(last_errno) << ": " << std::strerror(last_errno)
-                            << ')');
-
-                if (len < 0 && last_errno != EAGAIN && last_errno != EWOULDBLOCK)
-                    LOG_SYS_ERRNO(last_errno, '#' << getFD() << ": Socket read returned " << len);
-            }
-            while (len < 0 && last_errno == EINTR);
+                if (len < 0)
```
mmeeks, line 1122: I really think EAGAIN and EWOULDBLOCK are too noisy to log but ... if you need them for a bit - why not.

net/SslSocket.hpp
```diff
@@ -19,7 +20,7 @@
 class SslStreamSocket final : public StreamSocket
 {
 public:
-    SslStreamSocket(const std::string hostname, const int fd, bool isClient,
+    SslStreamSocket(std::string hostname, const int fd, bool isClient,
```
mmeeks, line 23: de-const-ipating hostname ?

mmeeks (approved):
Looks fine to me - minor comments =) thanks Ash.

net/Socket.hpp
Ashod: We did log them previously. This reduces the total log entries, actually.

net/SslSocket.hpp
Ashod: Haha. Actually, the argument is moved, so the const doesn't help here. Clang-tidy fix.

PR #4457, opened 2022-03-22
wsd: cleanup: moved all parsing related static method and containers …

Rash419:
Change-Id: I1821b19eca2132e5b03ba6ab41364ea58865a6ba

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -130,6 +130,7 @@ coolwsd_sources = common/Crypto.cpp \
                   wsd/FileServerUtil.cpp \
                   wsd/RequestDetails.cpp \
                   wsd/Storage.cpp \
+                  wsd/HostUtil.cpp \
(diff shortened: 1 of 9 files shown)


kendy (approved):
Looks good in general. Please remove the HostUtis contructor, change the includes a bit and check the breakage on iOS; other that than I'm happy - let's merge :-)

Rash419:
@kendy How it works: 
For example: host = nextcloud-abc , aliases = nextcloud-pqr, nextcloud-xyz
if nextcloud-abc locked_host settings is defined that nextcloud-pqr and xyz will use that settings
if it is not defined it will use fallback settings  :)

kendy (approved):
Thank you @Rash419 , let's do it! :-)

PR #4469, opened 2022-03-23
comments: we need id with a number type

eszkadev:


eszkadev:
This only covers the issue, I prepared fix in the core: https://gerrit.libreoffice.org/c/core/+/132054

PR #4470, opened 2022-03-23
private/pedro/secondary n primary buttons plus hover statsu

pedropintosilva:
- Remove duplicated n inactive margin from annotation btns
- Autofilter buttons do not need special css targeting
- Refactor commentSection buttons to avoid using multiple css classes
- Add missing hover state for primary buttons
- Switch hover color to a lighter color


gokaysatir (approved):

PR #4472, opened 2022-03-24
annotation: fix layouting problem on selected comments

merttumer:
modify and reply causes relayouting due to expansion of the
commentsections but it is based on whether they are selected or not
currently. If they are already selected we dont relayout them again
and it causes overlapping. To avoid that we can simply unselect
comments just before reply or modify.

Change-Id: Ifc2e546504b36b46f98a9b22ce791b4279239457

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Tested, it works

PR #4478, opened 2022-03-25
fixed edit button not appearing in ipads

lpranam:
Change-Id: I949b3bc6d72541cba18867847cd43f0c6e57a32f

* Target version: master 

### Checklist

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -916,6 +916,6 @@ div#fontstyletoolbox + div#style.mobile-wizard {
 	height: 100%;
 }
 
-.div-infobar-wrap ~ #mobile-edit-button {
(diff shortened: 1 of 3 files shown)


pedropintosilva (approved):
it works thanks @lpranam ! Nevertheless I think we might have another problem here:

pedropintosilva:
on top of that and after checking the code it seems we are not removing elements from DOM when we get an error we are simply setting display none on the wrapper so... I think I cannot rely on display not to hide or show the edit button on mobile (to avoid this over lay 
image
 (

pedropintosilva:
so for the time being better to just remove that CSS altogether  L /

pedropintosilva (approved):

PR #4481, opened 2022-03-25
Notify server about visible slide

eszkadev:
Fixes problem with:
- showing wrong slide number in statusbar
- inserting comment at correct page
in pdfs where we see slides in continuous mode.

Sending slide id from online to server is much easier
than calculating based on visible area on the server
where we have to put that data into TabControl widget
which is used in different places.

It updates visible slide on scroll and also when selected
from the previews bar.


dennisfrancis (approved):
Looks good. Thanks!

PR #4486, opened 2022-03-28
Notebookbar: set overflow visible before removing notebookbar when ch…

gmasei11:
…anging layout

Change-Id: Ifde1603e33399a50173c78b37f20d1c30db9f536

* Resolves: # 
* Target version: master 

### Summary
When changing the layout from notebookbar to classic the popup menus associated with main menu options are not visible.

### TODO

- [ ] ...

### Checklist


pedropintosilva (dismissed):
thanks

pedropintosilva (approved):

PR #4487, opened 2022-03-28
Cypress: Do not hard target buttons based on their position

pedropintosilva:
No need to specify if it's first or second button as there is always
only one primary button. On top of that this avoids maintenance work
whenever for some reason the button is not the first in line.

Change-Id: Idd3ef60023e88d1a9a8fd6f465dc4c63c068debd


pedropintosilva:
I have spotted this as a possible problem with https://github.com/CollaboraOnline/online/pull/4446 (in there I already fixed for hyperlink) in this PR I'm spreading out the changes to the other instances

Rash419 (approved):
Thanks : )

PR #4491, opened 2022-03-29
clipboard: introduce paste special menu entries

eszkadev:
It makes paste special feature more visible to the users
Also move paste special popup code to the clipboard.js so it is easiet to share the code.

What to do with notebookbar mode: to be discussed...

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -347,7 +347,7 @@ L.Control.JSDialog = L.Control.extend({
 
 		var dialogInfo = this.dialogs[data.id];
 		if (!dialogInfo.invalidated && dialogInfo.setupPosFunc) {
-			setTimeout(dialogInfo.setupPosFunc(true), 100);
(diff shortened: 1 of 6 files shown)


mmeeks:
Looks lovely to me - ctrl-alt-shift-v is un-findable for anyone who doesn't look at help - which is ~95% of users =)

pedropintosilva (dismissed):
Thanks 
image

pedropintosilva:
I think we should probably do the same for notebookbar mode

pedropintosilva:
have the paste bigtoolitem to be a menu (icon plus arrow down) that would show that option (so, reusing the menu contro we are already using in other places like formatting Mark etc

eszkadev:
- hints: done
- Paste Special dialog now doesn't trigger error in the console when enabled for jsdialogs (see https://forum.collaboraonline.com/t/how-to-help-with-jsdialogs-development-porting-dialogs-to-native-html/964/2)

pedropintosilva (approved):

PR #4493, opened 2022-03-29
fix trackchange comments are not displayed on docload

merttumer:
fix typo on clearList loop that prevents skipping trackchange
comments from being deleted

Also I removed the line in WriterTileLayer.js containing
clearList because clearList is already called on importComments
function so it was unnecessary

Change-Id: Ic4086a80b126aec62e382078f53d68bc17da8a1c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected

PR #4494, opened 2022-03-29
l10n: if english variant is not specified then choose US as default

lpranam:
Change-Id: Iba4f96c75d69e52044f6460c4e26d0ab1d187d3c

* Target version: master 

### Checklist

### DIFF
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -157,7 +157,10 @@ void Session::parseDocOptions(const StringVector& tokens, int& part, std::string
         }
         else if (name == "lang")
         {
-            _lang = value;
+            if (value == "en")
+                _lang = "en-US";
+            else
+                _lang = value;
             ++offset;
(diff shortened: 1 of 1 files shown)


common/Session.cpp
```diff
@@ -157,7 +157,10 @@ void Session::parseDocOptions(const StringVector& tokens, int& part, std::string
         }
         else if (name == "lang")
         {
-            _lang = value;
+            if (value == "en")
+                _lang == "en-US";
```
merttumer, line 161: LGTM, except this typo :=)

lpranam: done. thanks

timar (approved):
You are right, it's easier to fix it here, than finding out what is wrong with the language fallbacks in core. :)

timar:
We need this in co-6-4, too.

PR #4499, opened 2022-03-30
Close iframeDialogs when pressing escape key

pedropintosilva:
Change-Id: Ief0919048bce1a27bee1c6c263a693d487f9c7f2


gokaysatir (approved):
Thanks for the improvement :)

PR #4502, opened 2022-03-31
Gracefully unload when there is no Kit and nothing to upload

Ashod:
- wsd: remove unsed function in DocumentBroker
- wsd: remove dup entry in dumpState
- wsd: unload when no Kit and nothing to upload


mmeeks (approved):
Looks sensible to me - thanks Ash =)

PR #4503, opened 2022-03-31
wsd: test SSL hostname

Ashod:
Change-Id: I0895d90d2c9be8a76afd209446ae39cd33fa5d2e

### DIFF
--- a/net/SslSocket.hpp
+++ b/net/SslSocket.hpp
@@ -20,10 +20,10 @@
 class SslStreamSocket final : public StreamSocket
 {
 public:
-    SslStreamSocket(const std::string& hostname, const int fd, bool isClient,
+    SslStreamSocket(const std::string& host, const int fd, bool isClient,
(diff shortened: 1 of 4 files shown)


net/SslSocket.hpp
```diff
@@ -48,12 +48,12 @@ class SslStreamSocket final : public StreamSocket
             throw std::runtime_error("Failed to create SSL.");
         }
 
-        if (!hostname.empty())
+        if (!hostname().empty())
```
mmeeks, line 51: Do we want this functional change instead of using host ?

Ashod: `hostname()` returns the member that is the `host` argument. Not a functional change (as the test should attest).

mmeeks: If you're happy with that; I worry about the system hostname() function but as you like.

mmeeks (commented):
Looks much more robust - thanks Ash ! =)

mmeeks (approved):

PR #4519, opened 2022-04-02
browser: Disable browser reload on F5

thebearon:
In Impress, launch slideshow.

Change-Id: Ib4b72f2f1278eb30b0177da4e3c3ef9bfd6d11a8

* Target version: master

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -334,6 +334,16 @@ L.Map.Keyboard = L.Handler.extend({
 			return;
 		}
 
+		// don't trigger browser reload on F5, launch slideshow in Impress
+		if (ev.type === 'keydown' && keyCode === 116) {
+			ev.preventDefault();
+			if (this._map._docLayer._docType === 'presentation')
+			{
+				this._map.fire('fullscreen');
+			}
+			return;
+		}
+
 		var unoKeyCode = this._toUNOKeyCode(keyCode);
 
 		if (this.modifier) {


eszkadev (approved):
Works as described. I've noticed that f5 in the slideshow does the reload. Maybe we should insert similar code in presentation engine in core to prevent reloading?

thebearon:
Thanks for the review, Szymon!

Good idea, I'll open a GH issue on it for now.

gitgt:
I tested this. 

For some reason, after the new login, when presentation is opened and F5 pressed for the 1st time, it just reloads. 
(Not sure why "Explore The New 22.05.10.1" also repeats even if I closed it in the previous login and then logged out).
Only on 2nd press of F5 it starts prez in Chrome. Chrome 110 profile was empty. 
Also, if prez is started and ESC stopped, another F5 doesn't start it again. 

Another issue was with my Firefox browsers. F5 appeared as not working at all, it would just create a small image of the slide. 
Until I figured out that I need to press in the document after loading. 
This is a good idea, but not fully intuitive.

thebearon:
@gitgt you seem to be encountering a bug with focus, ie. there is different content from multiple sources in a browser tab, and the focus can be in either part, which then reacts to actions. I'm not sure how that's coded, and what the default behavior is, there's definitely a security aspect, as I do see this error message in the browser console sometimes:
`Blocked autofocusing on a <textarea> element in a cross-origin subframe.`
(it probably matters if the WOPI host and COOL are on the same or on different servers)

I'd say it's worth reporting those separately, as it's a separate can of worms (probably as a single issue for now).

Also, good point about the keyboard shortcut, the shortcuts help should be amended with that sometimes.

PR #4524, opened 2022-04-04
clipboard: add Paste Special also in context menu

eszkadev:


pedropintosilva (approved):
Tested, it works, thanks!

PR #4525, opened 2022-04-04
paste special shortcuts in manus and context 6.4

eszkadev:
added Paste Special to menubar, notebookbar and context menu


pedropintosilva (approved):
Tested both on classic and Notebookbar mode, it works, thanks!

PR #4528, opened 2022-04-04
mobile: hide send feedback button

merttumer:
we dont have welcome dialog for mobile

Change-Id: I7b0196418e5c3b00f87630de42b131b3deb3ae15

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -2051,6 +2051,9 @@ L.Control.Menubar = L.Control.extend({
 			}
 		}
 
+		if (item.id === 'feedback' && !this._map.feedback)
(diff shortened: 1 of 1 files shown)


hcvcastro (approved):

timar (commented):
Hmm, not bad, it does not look a root cause fix to me. Why do we see the menu item, when we do not configure the feedback URL for mobile app? 
We do not have this menu item in Collabora Online, only in CODE.

merttumer:
FWIW, for mobile theres another function generateFullMenuStructure is called to build the hamburger menu. That's why it was displayed.

PR #4541, opened 2022-04-04
jsdialog: autofilter: fix fox co-22.05

eszkadev:
jsdialog: autofilter: unify with regular popup (from co-22.05)
- in 22.05 core autofilter was welded and converted into regular popup
- for now we have to support both versions for 2021 and 22.05:
    - for 22.05 handled in Control.JSDialog.js
    - for 2021 handled in Control.AutofilterDropdown.js
- leter we need to remove Control.AutofilterDropdown.js
- special handling in Control.JSDialog.js should be separated later...

requires core changes:
https://gerrit.libreoffice.org/c/core/+/132538
https://gerrit.libreoffice.org/c/core/+/132539
https://gerrit.libreoffice.org/c/core/+/132619

### DIFF
--- a/browser/src/control/Control.AutofilterDropdown.js
+++ b/browser/src/control/Control.AutofilterDropdown.js
@@ -1,6 +1,9 @@
 /* -*- js-indent-level: 8 -*- */
 /*
- * L.Control.AutofilterDropdown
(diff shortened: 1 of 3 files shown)


merttumer (approved):
Works as expected

PR #4544, opened 2022-04-05
Tests, logs, and non-functional improvements

Ashod:
Misc test, logging, assertions, and similar improvements.

All of these changes are non-functional and any functional change should be flagged in the review.

- wsd: move tokenizer helpers into StringVector
- wsd: log the duration of save and upload when done
- wsd: move STATE_ENUM to common/StateEnum.hpp
- wsd: use STATE_ENUM for DocBroker states
- wsd: minor improvements to logging and macros
- wsd: modernize UnitWOPI
- wsd: logging and minor cosmetics
- wsd: add LOG_ASSERT to replace assert with extra logging
- wsd: include the ClientSession name in logs
- wsd: use STATE_ENUM for SessionState
- wsd: support serializing STATE_ENUM to ostream 
- wsd: use STATE_ENUM for http

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -275,6 +275,7 @@ shared_headers = common/Common.hpp \
                  common/JailUtil.hpp \
                  common/Log.hpp \
                  common/Protocol.hpp \
+                 common/StateEnum.hpp \
                  common/StringVector.hpp \
                  common/Seccomp.hpp \
                  common/Session.hpp \
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -81,7 +81,7 @@ namespace FileUtil
 {
     std::string createRandomDir(const std::string& path)
     {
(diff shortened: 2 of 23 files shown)


vmiklos:
Would it make sense to sort out the gcc-6.3 failure before reviewing the code manually?

Ashod:
> Would it make sense to sort out the gcc-6.3 failure before reviewing the code manually?

That was because I forgot to add the new header to Makefile.am. Unfortunately, it seems the Android build is broken.

At any rate, I broke up this PR into two, since it was very large. The other one is #4551. They are both ready for review (ignoring the broken Android build). They are both non-functinional.

Thanks in advance!

vmiklos (approved):
I don't see anything obviously problematic with these changes. The APPLY* macros you move around are ugly, but I guess at the moment variadic templates are not good enough to do the same.

PR #4547, opened 2022-04-05
common: isValidUtf8: ensure chunk len is 4 at max

vmiklos:
UTF-8 is expected to encode a code point using 1 to 4 bytes.

Change-Id: I9b4f346738865fc18242e51a33c97e5b38ee90d8

### DIFF
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -303,7 +303,14 @@ namespace Util
             if (data[i] >> 6 != 0x3)
                 return i;
             int chunkLen = 1;
-            for (; data[i] & (1 << (7-chunkLen)); chunkLen++);
+            for (; data[i] & (1 << (7-chunkLen)); chunkLen++)
+                if (chunkLen > 4)
+                    return i;
+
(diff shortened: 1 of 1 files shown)


common/Util.hpp
```diff
@@ -303,7 +303,9 @@ namespace Util
             if (data[i] >> 6 != 0x3)
                 return i;
             int chunkLen = 1;
-            for (; data[i] & (1 << (7-chunkLen)); chunkLen++);
+            for (; data[i] & (1 << (7-chunkLen)); chunkLen++)
+                if (chunkLen > 4)
+                    return i;
             for (; chunkLen > 1; --chunkLen)
                 if (data[++i] >> 6 != 0x2)
```
mmeeks, line 310: heh - we should also check that i + chunkLen < len after we've worked out the chunkLen.
my bad ! thanks for catching; can you tweak ? =)

Ashod (approved):
Good stuff, @vmiklos. Looks like the CI failure is unrelated.

PR #4548, opened 2022-04-05
Classic: Add missing split cell dialog trigger to writer

pedropintosilva:
When working with tables it is now possible to to slip cells
where: Top menu: Table -> Split Cell...

Change-Id: Ia242358245cebb6bded55aa995bd8c80900b9aaa

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -274,6 +274,7 @@ L.Control.Menubar = L.Control.extend({
 					{uno: '.uno:EntireRow'},
 					{uno: '.uno:EntireColumn'},
 					{uno: '.uno:EntireCell'}]},
+				{uno: '.uno:SplitCell'},
 				{uno: '.uno:MergeCells'},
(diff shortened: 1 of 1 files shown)


eszkadev:
> it works but regarding the disable status:
> 
> @eszkadev I wonder how I can disable this dialog trigger `Split Cells...` id the user is not inside of the table. I was expecting to automatically get the right status set by the `_beforeShow` `if (itemState === 'disabled') {` condition but it seems that the `itemState` does not get `disabled` state. When debugging it seems it gets `.uno:SplitTable=disabled` whih seems odd

SplitCell != SplitTable :)
I see uno command you added is .uno:SplitCell and status we receive .uno:SplitTable

We have to enable status updates in core for .uno:SplitCell.
Done here: https://gerrit.libreoffice.org/c/core/+/132585

eszkadev:
I've added some documentation in forum, how to do it. In case it will be needed in the future: https://forum.collaboraonline.com/t/faq-how-to-work-with-uno-commands/1009

PR #4551, opened 2022-04-06
Break-up WhiteBoxTests

Ashod:
This breaks up `WhiteBoxTests.cpp` to reduce its build time. This translation-unit was (still is) the single largest and longest-compiling one, taking multiple minutes on my box. This reduces the build time and makes the tests more readable as they are now in more relevant test files.

- wsd: test: split out RequestDetails tests
- wsd: test: move Authorization tests to RequestDetailsTests
- wsd: move tokenizer helpers into StringVector
- wsd: test: move StringVector tests to StringVectorTests


Ashod:
> This looks OK to me. Nitpick: "wsd: move tokenizer helpers into StringVector" doesn't say why you move this: ideally you should provide some argument in the body of the commit message, so the next person doesn't move tokenizer helpers back to where they were before. :-)

Good point. Updated the commit log and also moved the tests (some 300 lines) out of `WhiteBoxTests.cpp`, which further reduces it. Thanks!

vmiklos:
> Updated the commit log

Perfect, thanks. :-)

PR #4553, opened 2022-04-06
Correctly hide keyboard on mobile wizard

merttumer:
current approach is lacking the acceptInput state
because when we close the keyboard we need to set it
to false when we hide the keyboard otherwise canAcceptInput is not going
to change and controls for that will fail such as panStart and panEnd in
TouchGesture. Pannig while the mobile wizard is on causes keyboard going
on top of the mobile wizard and after that it causes endless screen
flickering.

Change-Id: I3248ef7f344fe53a57b7c77f4222d1fed5740896

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.MobileWizard.js
+++ b/browser/src/control/Control.MobileWizard.js
@@ -191,7 +191,7 @@ L.Control.MobileWizard = L.Control.extend({
 	},
 
 	_hideKeyboard: function() {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
works as expected

PR #4554, opened 2022-04-06
Fix endless keyboard show/hide flicker

merttumer:
when we decide the keyboard is closed by checking the screen size
we should not force keyboard to close again instead we just need to
set the state

Change-Id: I486cda0598ab02e426aefd726475f0e70e0eee52

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
I was not able to reproduce the issue on my devices, so I add +1 to not block the fix

PR #4558, opened 2022-04-07
wsd: cleanup firsthost code

Rash419:
we added AllHosts to give admin the err log that host is not in alias_groups but now as we removed the host list entries from configuration we don't need that log

Change-Id: I8b5e9e6b7df7df59befb496c12966c7ddc60c707

* Target version: master


kendy (approved):
Looks like a very nice cleanup to me, thank you!

PR #4559, opened 2022-04-07
Fix clone formatting button does not have state on notebookbar

merttumer:
Change-Id: Ib5c2ff5629a8334d9c2866642e1d6d1c794aff82

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2339,10 +2338,16 @@ L.Control.JSDialogBuilder = L.Control.extend({
 					if (state && state === 'true') {
 						$(button).addClass('selected');
 						$(div).addClass('selected');
+						if (data.command === '.uno:FormatPaintbrush') {
+							$('.leaflet-pane.leaflet-map-pane').addClass('bucket-cursor');
```
eszkadev, line 2342: it would be nice to move this to some global handler, so we will get correct cursor also when no button is created

merttumer: done, thank you for the suggestion

eszkadev (approved):
this looks good :)

PR #4560, opened 2022-04-07
test: increase buffer size in SocketProcessor()

vmiklos:
The webframe buffer size used to be READ_BUFFER_SIZE (64k), but that's
no longer enough in unit-uno-command since
<https://github.com/libreoffice/core/commit/a20e761cb1112357bfae35e24b49fae872bce3f6>
(jsdialog: fix toolboxes with children, 2022-03-17) if you have enough
fonts installed locally, because one toolbox in Writer is the font
dropdown, and that lists all the fonts you have.

getResponseMessage() already uses this increased buffer size, so this
also improves consistency.

Change-Id: Ic2ad0e8e7443365395b16618d6c82465252ba455


vmiklos:
CC @eszkadev

PR #4565, opened 2022-04-07
Save and Upload Improvements: Part 1

Ashod:
**Part 1**

A number of fixes and improvements to Save and Upload cases. Each commit explains the need for the change.
Where appropriate, `ExperimentalEnabled` is used.

Most were found through tests (a few through reviewing code). Improved test coverage included (which fail without these fixes) in part-3 #4567. Please review part-2 #4566 after this.

_Best to review each commit separately._

- wsd: detect ECONNRESET and EPIPE after reads and writes
- wsd: check the stop flag immediately after poll
- wsd: correct active-session counting
- wsd: add and use isWritable helper in session
- wsd: do not close the socket while have data to read

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1294,10 +1294,19 @@ protected:
 
         if (!EnableExperimental || events & POLLIN)
         {
-            closed = !readIncomingData() || closed;
+            // readIncomingData returns false only if the read len is 0 (closed).
+            // Oddly enough, we don't necessarily get POLLHUP after read(2) returns 0.
+            const bool reading = readIncomingData();
+            closed = !reading || closed;
             LOG_TRC('#' << getFD() << " Incoming data buffer " << _inBuffer.size()
                         << " bytes, closeSocket? " << closed << ", events: " << std::hex << events
                         << std::dec);
+            if (EnableExperimental && closed && reading)
+            {
+                // We might have outstanding data to read, wait until readIncomingData returns false.
+                LOG_DBG('#' << getFD() << ": Closed but will drain incoming data per POLLIN.");
+                closed = false;
+            }
         }
 
 #ifdef LOG_SOCKET_DATA
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1355,9 +1355,12 @@ protected:
             if ((events & POLLOUT) && !_outBuffer.empty())
             {
                 writeOutgoingData();
-                if (errno == EPIPE)
+                const int last_errno = errno;
+                if (last_errno == EPIPE || (EnableExperimental && last_errno == ECONNRESET))
                 {
-                    LOG_DBG('#' << getFD() << ": Disconnected while writing (EPIPE).");
+                    LOG_DBG('#' << getFD() << ": Disconnected while writing ("
+                                << Util::symbolicErrno(last_errno)
+                                << "): " << std::strerror(last_errno) << ')');
                     closed = true;
                     break;
                 }
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -274,6 +274,12 @@ void DocumentBroker::pollThread()
         _poll->poll(unloading ? SocketPoll::DefaultPollTimeoutMicroS / 16
                               : SocketPoll::DefaultPollTimeoutMicroS);
 
+        if (EnableExperimental && _stop)
+        {
+            LOG_DBG("Doc [" << _docKey << "] is flagged to stop after returning from poll.");
+            break;
+        }
+
 #if !MOBILEAPP
         const auto now = std::chrono::steady_clock::now();
 
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1785,23 +1785,26 @@ std::string DocumentBroker::getWriteableSessionId() const
     assertCorrectThread();
 
     std::string savingSessionId;
-    for (auto& sessionIt : _sessions)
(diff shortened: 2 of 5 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -1849,7 +1852,7 @@ bool DocumentBroker::autoSave(const bool force, const bool dontSaveIfUnmodified)
     // Note: a loaded view cannot be disconnecting.
     const auto itLastEditingSession = _sessions.find(_lastEditingSessionId);
     const std::string savingSessionId =
-        (itLastEditingSession != _sessions.end() && itLastEditingSession->second->isReadOnly() &&
+        (itLastEditingSession != _sessions.end() && itLastEditingSession->second->isWritable() &&
```
mmeeks, line 1855: from isReadOnly -> isWritable - looks like a logic error - do we miss a not ?

wsd/DocumentBroker.cpp
```diff
@@ -2981,10 +2983,7 @@ std::size_t DocumentBroker::countActiveSessions() const
     std::size_t count = 0;
     for (const auto& it : _sessions)
     {
-        if (it.second->isViewLoaded() && !it.second->inWaitDisconnected())
-        {
-            ++count;
-        }
+        count += it.second->isLive();
```
mmeeks, line 2986: Not thrilled by adding a bool to a size_t - of course it may work - but it looks highly un-readable and paranoid programmers who know there is no boolean register type may freak out (like me)

Can we have it in the conditional as before ?

wsd/DocumentBroker.cpp
```diff
@@ -274,6 +274,12 @@ void DocumentBroker::pollThread()
         _poll->poll(unloading ? SocketPoll::DefaultPollTimeoutMicroS / 16
                               : SocketPoll::DefaultPollTimeoutMicroS);
 
+        if (EnableExperimental && _stop)
```
mmeeks, line 277: how is 'stop()' called by ->poll directly (or someone else?) and why does the normal processing take significant time? - surely it should not or everything will crawl ? =)

net/Socket.hpp
```diff
@@ -1294,7 +1294,20 @@ class StreamSocket : public Socket,
 
         if (!EnableExperimental || events & POLLIN)
         {
+            // readIncomingData returns false only if the read len is 0.
             closed = !readIncomingData() || closed;
+            if (EnableExperimental)
+            {
+                const int last_errno = errno;
+                if (last_errno == EPIPE || last_errno == ECONNRESET)
```
mmeeks, line 1302: Hmm - why bother? there is great merit in simplicity and having fewer code-paths to test.
Is there a genuine problem going around the loop again to detect a HUP in the next poll ?
Of course - if we treat 0 length reads as an error - then perhaps it makes sense to use these error values too but - lets not add complexity and expand our test-matrix just for fun.
Is there a good, hard reason to have a duplicate closing code-path here ? are there good numbers that show a win ?
If not - I'd not do it myself.

mmeeks (changes requested):

wsd/DocumentBroker.cpp
Ashod: It was a bug, evidently, since a read-only session shouldn't be used for saving (and indeed it couldn't have been the last-editing-session in the first place, so `isReadOnly()` would be returning false anyway, and we'd fall back to choose a writable session).

wsd/DocumentBroker.cpp
Ashod: `bool` is implicitly promoted to integral types. The language rules guarantee that `true` is 1 and `false` is 0. (The opposite is also true, when coercing an integral value into a bool.) The language guarantees that no other value can be stored in a bool, so we can't end up adding arbitrary numbers here.

We can have the conditional, of course, it just looks rather verbose and unnecessary.

wsd/DocumentBroker.cpp
Ashod: `stop()` gets called through the Socket handler. What happens is that when the async uploading finishes, the handler (which is in DocumentBroker) is invoked and it figure that was the last upload before exiting, calls `stop()`.

Without checking for `_stop` after `poll()` we end up uploading yet again (when `always_save_on_exit` is set), which besides making testing harder (since depending on timing differences we might do an extra upload), it also uploads redundantly (this can't be helped on the storage side because `always_save_on_exit` forces the upload).

net/Socket.hpp
Ashod: As discussed, we will check HUP only when we get no POLLIN. So this is obsolete and I'll remove it.

However, for the write case it's worth noting that at least on my system I see ECONNRESET, which we don't check for, rather than EPIPE. So checking for either of these after the write seems sensible. What happens when we don't detect disconnection is that we make testing exit conditions much harder and unreliable (because depending on timing you get extra activity--or not). For example, in the third PR in this series we count the number of uploads done, and when we don't detect disconnections or `_stop` (see above), we do extra uploads, and the test fails.

I'd love to have all externally-observable activity tested and get to a point where we don't do arbitrary activity.

Ashod:
@mmeeks, all concerns/issues resolved. I've also updated the commit logs to be more informative (in addition to the explanations above). And while I increased the number of commits by 1, I reduced the diff by a handful of lines!

I left the socket commits separate to better document the change and explanation in the commit log. All socket changes are protected with EnableExperimental, and they are both very small and highly readable (in addition to the comments and commit-logs).

mmeeks:
iOS build failures look un-related, and this is a really nice cleanup - thanks Ash ! =)

PR #4566, opened 2022-04-07
Save and Upload Improvements: Part 2

Ashod:
**Part 2**

A number of fixes and improvements to Save and Upload cases. Each commit explains the need for the change.

Most were found through tests (a few through reviewing code). Improved test coverage included (which fail without these fixes) in part-3 #4567.

- wsd: do not prevent saving when storage is missing
- wsd: add canUploadToStorage helper
- wsd: add canSaveToDisk helper
- wsd: needToSaveToDisk helper and improved autoSaveAndStop
- wsd: delay uploading if the last attempt had failed

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1347,14 +1347,24 @@ void DocumentBroker::uploadToStorage(const std::string& sessionId, bool force)
         _storage->forceSave();
     }
 
-    constexpr bool isRename = false;
-    uploadToStorageInternal(sessionId, /*saveAsPath*/ std::string(),
-                            /*saveAsFilename*/ std::string(), isRename, force);
+    if (force || _storageManager.lastUploadSuccessful() ||
+        (_storageManager.timeSinceLastUploadRequest() > std::chrono::seconds(5) &&
+         _storageManager.timeSinceLastUploadResponse() > std::chrono::seconds(5)))
+    {
+        constexpr bool isRename = false;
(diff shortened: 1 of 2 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -1823,16 +1859,18 @@ bool DocumentBroker::autoSave(const bool force, const bool dontSaveIfUnmodified)
...
+    const CanSave canSave = canSaveToDisk();
+    if (canSave != CanSave::Yes)
     {
-        // Nothing to do.
-        LOG_TRC("Nothing to autosave [" << _docKey << "].");
+        LOG_DBG("Cannot save to disk: " << name(canSave));
         return false;
     }
 
-    if (_docState.isDisconnected())
```
Ashod, line 1833: Being disconnected from the Kit is handled by `canSaveToDisk()` above. We only need to check we are modified, or forced autosave, to proceed.

mmeeks (approved):
Had a good read, I see you like descriptive states these days: why not =) it prolly reads better.
Lets get it into master but hold-fire for 21.11 =)

Ashod:
> Had a good read, I see you like descriptive states these days: why not =) it prolly reads better. Lets get it into master but hold-fire for 21.11 =)

Yes, readability and also to simplify the implicit state dependencies. Also, we increasingly have more combinations to handle, so simple implicit either-or cases don't cut it anymore. This should allow for better growth as well.

PR #4568, opened 2022-04-08
Update gerrit and git shields

pedropintosilva:
Change-Id: I95029916bda70206654eeadfc69180490f8ee532


pedropintosilva:
this is just a readme we should probably do something to automatically skip the tests for these type of files :thinking:

PR #4571, opened 2022-04-08
sidebar: fix alignment in chart deck

eszkadev:
changes, mainly in Chart Editing Panel
(select chart -> press "edit chart")

- simplified sidebar panel label with arrow
  arrow has simpler rules without left/right (for RTL)
- turned off margins for .jsdialog.root-container
  what caused elements to overflow in chart panel
  (eg drawingarea with chart types)
- changed panel width definition from fixed pixels to
  "100% - scrollbar"
- RTL: converted left/right -> margin-inline-* for some
  spinfields in chart panel

also includes :not(autofilter) in one rule which is changed in not merged PR...:
https://github.com/CollaboraOnline/online/pull/4541

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -21,7 +21,7 @@
 	z-index: 2001;
 }
 
-.jsdialog.root-container {
+.jsdialog.root-container:not(.autofilter):not(.sidebar) {
 	margin: .5em 1em; /*same as in .ui-dialog */
 }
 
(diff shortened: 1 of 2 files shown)


pedropintosilva (approved):
Tested and it works.

solves this bug:
https://archive.org/download/sidebar-fix-alignment-chart-deck/edit-chart-color.mp4



After this is merged we can start to hunt any CSS rules that were specifically targeting sub-elements in the sidebar and setting margins or padding. We now can simple rely on expanders width and if we want a marging to applied everywhere we can then do a combination of:
- CSS calc width 
- margin

PR #4573, opened 2022-04-08
Calc: Add missing row/columns height/width menus to the notebookbar

pedropintosilva:
Layout tab does not have the:

- optimal height/width
- width/height

User expects to see those options in the Layout tab (as in LO).
However it seems in LO those menus have all the row/column related
options and as this might be worst for the tablet/touch screen laptop
(to have all those option hidden in the dropdown) and since we already
have the insert/delete options as icons:
- Simply add height and width specific menus
https://archive.org/download/notebookbar-layout-add-row-column-menus/notebookbar-layout-add-row-column-menus.png

Change-Id: Iebdc6df833bb3cb9d125e23be01e5197ec0d80c3


eszkadev (approved):

eszkadev:
@pedropintosilva we have to revert "Optimal ..." entries which trigger dialog.
It doesn't work in multi user case (blocks other users which were present at that time in document)

Issue #4564, opened 2022-04-07
Ambiguous description of "Export as..." functions in the NotebookBar


SustReal:
**Is your feature request related to a problem? Please describe.**
In the File Tab of the NotebookBar the "Export as..." functions are not clear enough. It is easily confused with the "Save as..." function. 

 
**Describe the solution you'd like**
An "Export as..." note after the print symbol could make it clearer that the following buttons are intended for export. (However, a separate "Export as..." button would also be conceivable). 

**Describe alternatives you've considered**
I repeatedly click on the "Save as..." button instead of the export functions when I want to export files. For me, the current solution is not intuitive, although I am trying to learn it...

**Additional context**
![grafik](https://user-images.githubusercontent.com/51329893/162272915-490753b0-e8fb-47ce-8c7e-10fb2fd94345.png)

thebearon:
@pedropintosilva, what do you think?

pedropintosilva:
I agree that the current situation is not ideal. We should communicate without shadow of a doubt that the user can:
- Save the current document as another format or same format but different name (**save as**)
- "Export as" AKA **Download** to his local machine the current document as any of the available formats

So that we are doing right now is skipping one step (in conveying the "download" part) and jumping right into the "as" part resulting in some users getting confused (and spending more time in scanning all the options)

pedropintosilva:
I have discussed this with at the time @gmasei11 and @alessioalex-1and1 and we agree that we should allow for another way for the integrator to display those options (specially because they might want to add other btns and the white space is needed). And so @alessioalex-1and1 wen ahead and created a patch exactly for that (already merged): https://github.com/CollaboraOnline/online/pull/4283

![image](https://user-images.githubusercontent.com/65948705/162464508-2876efbe-56b2-4952-85e6-9e0239032fdb.png)

pedropintosilva:
So I think what we can do right now and that would require the less amount of effort is:
- Improve that download as icon so it does not feel out of place
- Make it a bigtoolitem (bigger)
- Have the `<group_download_as` option set to true by default in the next CODE release
- Wait/request for any user feedback on the topic

pedropintosilva:
then depending on the feedback we can regroup and think if instead what it makes sense is to introduce a complete new component: `"tab group label"`

pedropintosilva:
fixed

PR #4575, opened 2022-04-08
Improve group_download_as: replace icon

pedropintosilva:
As discussed in https://github.com/CollaboraOnline/online/issues/4564
the current download as 2 is quite different form the surrounding icons
- Remove and use colibre's icon

Change-Id: I5808af8a50c1f8025b38502ba5fcd2db3f98c9d5


pedropintosilva:
@gmasei11 I agree this icon comes from Colibre theme (which is used in the core as export) and here we are using it as download. I will adjust the icon : )

pedropintosilva:
I also start a draft PR to improve the implementation of this and to make more small and consistent with the other menus we already use : https://github.com/CollaboraOnline/online/pull/4576

gmasei11:
> @gmasei11 Also I think we should remove the `as` from `Download as` contrary to the "Save as" which has point (to convey to the user that the action will open a dialog/more options) in here we have a dropdown, it is already clear that the component is not an action in itself but a dropdown to more actions

It's up to you :)

pedropintosilva:
only iOS fails and it is unrelated

PR #4576, opened 2022-04-08
Private/pedro/group download as improvements

pedropintosilva:
This is a draft PR related to https://github.com/CollaboraOnline/online/issues/4564 and it's an attempt into simplifying the implementation with 80ce6c16319d551192b0808fa5744e988106363f 

Context: while working on https://github.com/CollaboraOnline/online/pull/4575 I noticed that maybe we are creating a very specific implementation instead of replicating what we do with other menus

### DIFF
--- a/browser/src/control/Control.NotebookbarImpress.js
+++ b/browser/src/control/Control.NotebookbarImpress.js
@@ -217,18 +217,10 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 
 		if (hasGroupedDownloadAs) {
 			content.push({
-				'id': 'downloadas-container',
-				'type': 'container',
-				'text': '',
-				'enabled': 'true',
-				'children': [
-					{
-						'id': 'downloadas',
-						'type': 'menubartoolitem',
-						'text': _('Download'),
-						'command': '.uno:InsertGraphic'
-					}
-				]
+				'id': 'DownloadAs:DownloadAs',
+				'type': 'menubutton',
+				'text': _('Download'),
+				'enabled': 'true'
 			});
 
 			content.push({
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -239,6 +239,22 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			{text: _UNO('.uno:JumpUpThisLevel', 'text'), uno: 'JumpUpThisLevel'},
 			{text: _UNO('.uno:ContinueNumbering', 'text'), uno: 'ContinueNumbering'}
 		];
+		this._menus['DownloadAs'] = [
+			{name: _('PDF Document (.pdf)'), id: 'downloadas-pdf', type: 'action', docType: []},
+			{name: _('ODF text document (.odt)'), id: 'downloadas-odt', type: 'action', docType: ['text']},
+			{name: _('Word 2003 Document (.doc)'), id: 'downloadas-doc', type: 'action', docType: ['text']},
+			{name: _('Word Document (.docx)'), id: 'downloadas-docx', type: 'action', docType: ['text']},
+			{name: _('Rich Text (.rtf)'), id: 'downloadas-rtf', type: 'action', docType: ['text']},
+			{name: _('EPUB (.epub)'), id: 'downloadas-epub', type: 'action', docType: ['text']},
+			{name: _('ODF presentation (.odp)'), id: 'downloadas-odp', type: 'action', docType: ['presentation']},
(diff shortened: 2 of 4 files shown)


pedropintosilva:
@eszkadev : I wonder if this is the right direction to make it bigtoolitem menu
* If it is then what would be the best way to do it for all the apps while avoid redundancy?
  * What about `Control.Menubar.js` : Ideally I would like to reuse the code (menu entries) there for the Notebookbar
  * I was thinking if there was a way to simply have all the menu entries for all apps within `this._menus['DownloadAs']` : `Control.JSDialogBuilder.js` and then add some propriety to each one that would just show them in the correct app something like `{name: _('ODF Drawing (.odg)'), id: 'downloadas-odg', type: 'action', text: false, spreadsheet: false, presentation: false},` or the opposite just sating which app should be `drawing: true`
* If this is not the way to go please feel free to:
  * close this PR with just a big of context
  * Maybe we should be using the upcoming [bigMenubarToolItem](https://github.com/CollaboraOnline/online/pull/4254/commits/8e2babd4471551a27b18d3caa9f45ed109209635#diff-3e0f2254afef982a82c7a71206ed2db22c9119b82da912a52a9234046711208dR20) for this case (?)

browser/src/control/Control.NotebookbarImpress.js
```diff
@@ -199,7 +199,7 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 				'enabled': 'true',
 				'children': [
 					{
-						'id': 'downloadas2',
+						'id': 'downloadas',
 						'type': 'menubartoolitem',
 						'text': _('Download as'),
 						'command': '.uno:InsertGraphic'
```
eszkadev, line 205: just noticed: what it does here? InsertGraphic command in download as?

eszkadev:
> * something like `{name: _('ODF Drawing (.odg)'), id: 'downloadas-odg', type: 'action', text: false, spreadsheet: false, presentation: false}`

yes, we can do this. We just need to add something in `menubutton` implementation in `JSDialogBuilder.js` :
```
$(control.container).w2menu({
	items: builder._menus[menuId],
	onSelect: function (event) {
		builder.map.sendUnoCommand('.uno:' + event.item.uno);
	}
});
```

like:
`items: filterApp(builder._menus[menuId], builder.map.getDocType()),`

and `var filterApp = function(menu, docType) { <remove bad entries>; return menu; }` which removes entries which are not matching the current app (but it should leave entries where apps are not defined at all)

example menu entries (avoid defining just `text:` - it can be confusing):
`{id: 'abc', docType: ['text', 'presentation']}`     <- visible only when current document is text or presentation
`{id: 'abc', docType: []}`     <- always visible
`{id: 'abc'}`     <-  always visible

browser/src/control/Control.NotebookbarImpress.js
pedropintosilva: exactly , that's why I really would like to re-implement this part, I think Alexandru just use a random command as a placeholder

pedropintosilva:
> > * something like `{name: _('ODF Drawing (.odg)'), id: 'downloadas-odg', type: 'action', text: false, spreadsheet: false, presentation: false}`
> 
> yes, we can do this. We just need to add something in `menubutton` implementation in `JSDialogBuilder.js` :
> 
> ```
> $(control.container).w2menu({
> 	items: builder._menus[menuId],
> 	onSelect: function (event) {
> 		builder.map.sendUnoCommand('.uno:' + event.item.uno);
> 	}
> });
> ```
> 
> like: `items: filterApp(builder._menus[menuId], builder.map.getDocType()),`
> 
> and `var filterApp = function(menu, docType) { <remove bad entries>; return menu; }` which removes entries which are not matching the current app (but it should leave entries where apps are not defined at all)
> 
> example menu entries (avoid defining just `text:` - it can be confusing): `{id: 'abc', docType: ['text', 'presentation']}` <- visible only when current document is text or presentation `{id: 'abc', docType: []}` <- always visible `{id: 'abc'}` <- always visible

Thanks @szymon I have followed your code pointers and returned the `filteredMenu` and it looks as expected (e.g.: items from calc or impress only do not appear in writer)

I have removed the generic InserGraphic uno command but, and as you have hinted, it seems it really needs a uno command placeholder in our current implementation

pedropintosilva:
rebased against master, no need to cherry-pick those jsdialogs commits they are already merged on master

gokaysatir:
@pedropintosilva is this PR ready to be reviewed?

pedropintosilva:
https://github.com/CollaboraOnline/online/pull/5153

PR #4577, opened 2022-04-08
Private/gokay/debugging improvements

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1,6 +1,10 @@
 /* -*- js-indent-level: 8 -*- */
 /* global Uint8Array */
 
+/*
+	For extending window.app object, please see "docstate.js" file.
+	Below definition is only for the properties that this (global.js) file needs at initialization.
+*/
 window.app = {
 	socket: null,
 	console: {}
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1,50 +1,9 @@
 /* -*- js-indent-level: 8 -*- */
 /* global Uint8Array */
 
-window.app = { // Shouldn't have any functions defined.
-	definitions: {}, // Class instances are created using definitions under this variable.
-	dpiScale: window.devicePixelRatio,
-	roundedDpiScale: Math.round(window.devicePixelRatio),
-	twipsToPixels: 0, // Twips to pixels multiplier.
-	pixelsToTwips: 0, // Pixels to twips multiplier.
-	file: {
-		editComment: false,
-		readOnly: true,
-		disableSidebar: false,
-		size: {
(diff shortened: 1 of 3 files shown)


browser/js/global.js
```diff
@@ -1,50 +1,9 @@
...
-			pixels: [0, 0] // This can be larger than the document's size.
-		}
-	},
-	tile: {
-		size: {
-			pixels: [0, 0],
-			twips: [0, 0]
-		}
-	},
+window.app = {
```
eszkadev, line 4: window.app is duplicated here and in docstate.jsm is that intended?

browser/src/docstate.js
```diff
@@ -0,0 +1,81 @@
...
+			pixels: [0, 0] // This can be larger than the document's size.
+		}
+	},
+	tile: {
+		size: {
+			pixels: [0, 0],
+			twips: [0, 0]
+		}
+	},
+	socket: window.app.socket,
```
eszkadev, line 45: self reference? window.app.socket = window.app.socket ?

browser/js/global.js
gokaysatir: :) yep, hmm it's not duplicated indeed. Recreated, i needed the same variable. Writing it like that looks much more familiar while reading the code than adding new properties.

eszkadev: ok, so can't we have only one place with window.app? I fear someone will be adding properties not in the correct place...
or these files are not always present both at the same time?

gokaysatir: Oh, that's right. We should add some comments there. Unfortunately "global.js" file needs those 2 variables at initialization. So i had to leave those 2 variables there. And the rest of the object is defined in "docstate.js" file.

They will be both present at the same time. I first tried to move "global.js" to outside of html but it's not a good solution for various reasons. The second option was this current solution.

Or we can move those 2 variables outside of "window.app" object.

My take would be adding a comment there to ensure that noone expands the object inside that file.

browser/src/docstate.js
gokaysatir: Thanks for the valuable comments :)

browser/js/global.js
eszkadev: yes, good idea, with comment nobody will be confused :)

eszkadev (approved):

browser/src/docstate.js
merttumer: cant we do something like window.app = {...window.app, ...{rest_here}} just to say we are expanding window.app? but of course it does not really matter in terms of functionality.

PR #4579, opened 2022-04-09
jsdialog: formulabar conversion

eszkadev:


eszkadev:
requires core changes:
https://gerrit.libreoffice.org/c/core/+/132757
https://gerrit.libreoffice.org/c/core/+/133010

cypress is not ready - checks canvas, some tests needs to be adjusted, eg: https://github.com/CollaboraOnline/online/pull/4611

merttumer (approved):
works as expected

PR #4580, opened 2022-04-10
Do not display update infobar when the current version is higher (e.g…

timar:
…. development version)

Change-Id: Id17c28535fe0d47518189da25582d540bd597b81


gokaysatir (approved):
Reviewed & tested. Thanks for the improvements :)

PR #4583, opened 2022-04-11
fix Notebookbar menus flickering

merttumer:
regression started with: 23156808083ebbf31aa42f87cab74999258ecbba

added accidentally removed ","

Change-Id: I18bf3a44615085d9cf3066b696ac316dd9196254

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -283,7 +283,7 @@ td.jsdialog .jsdialog.cell.sidebar {
 }
 
 /* non selected */
-#clearFormatting
+#clearFormatting,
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Thank you!

pedropintosilva:
merging right away , this is a fix of a missing `,` on CSS. Unfortunately It passed in all tests (we should probably have stylelint that we already use as part of the tests) this change will also not change nothing

pedropintosilva:
also this reminds me that we should probably add CSS analysis (with stylelint) as an extra test. After all we already use it in the building steps

this would flag these situations

PR #4585, opened 2022-04-11
Add Print Area (ranges) commands to classic mode in calc

pedropintosilva:
commit 5545808fe94740bd0e97d138af6f55a402e4f127 (HEAD -> private/pedro/T34684, origin/private/pedro/T34684)
Date:   Mon Apr 11 15:14:36 2022 +0200

Add Print Area (ranges) commands to **notebookbar** mode in Calc

Add new svg icon
Add menuitem to Layout tab with all the missing commands

Note: EditPrintArea was not be added because:
It triggers a dialog that is not async at the moment. When tested with
multiple users within the same server it does not work. If the one user
opens it the other one is not able to open it at the same time.
AddPrintArea was not added since it can also be problematic when multiple
users edit the same document.
    
------

commit cbaa8c04eba19afa0523975e3a19a0bb4a6e98eb
Date:   Mon Apr 11 13:54:09 2022 +0200

Add Print Area (ranges) commands to **classic** mode in calc

There was only one way to successfully set a print rage: by defining
name ranges first and then within the dialog the tick the checkbox
'print range' under `Range options` expander.

All the other expected top menu entries within format main menu were
missing, Add except EditPrintArea and AddPrintArea:
- DefinePrintArea
- DeletePrintArea

EditPrintArea couldn't be added:
It triggers a dialog that is not async at the moment. When tested with
multiple users within the same server it does not work. If the one user
opens it the other one is not able to open it at the same time.

AddPrintArea wans't added because it might also be problematic when
multiple users edit the same document
    
-----


eszkadev (changes requested):
It doesn't work in multi user case:
- A opens "Edit..."
- B cannot open the same dialog at that time

eszkadev (approved):

PR #4586, opened 2022-04-11
Add Print Area (ranges) commands to notebookbar mode in Calc

pedropintosilva:
Add new svg icon
Add menuitem to Layout tab with all the missing commands

  Note: EditPrintArea wan not be added because:
  It triggers a dialog that is not async at the moment. When tested with
  multiple users within the same server it does not work. If the one user
  opens it the other one is not able to open it at the same time.

Change-Id: I276b6ab08b0f541aba9211d2595786a835d873b8


pedropintosilva:
#4585 needs to get merge first (since it reuses the newly added command in unocommands.js)

pedropintosilva:
#4585 has this commit as well, so we can effectively use CI without waiting for one PR to get merged first

PR #4589, opened 2022-04-11
Fix fallback welcome dialog on phone + other backports from online welcome dialog

timar:
Change-Id: Ifdf373c2c0b5a0962610d5fcb35f5845603a7a03


merttumer (approved):
Works as expected

PR #4593, opened 2022-04-12
add sparkline action to the menubar and notebookbar

quikee:
Just a to make the sparkline actions accessible from the menu
and notebookbar.

Change-Id: I80a483397184a8a9c0eb28048b3636ec91b95081

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -1001,6 +1001,11 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'command': '.uno:InsertObjectChart'
 			},
 			{
+				'type': 'bigtoolitem',
+				'text': _('Sparkline'),
+				'command': '.uno:InsertSparkline'
+			},
+			{
 				'id': 'Insert-Section-PivotTable-Ext',
 				'type': 'container',
 				'children': [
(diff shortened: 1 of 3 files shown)


browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -994,6 +994,11 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'text': _UNO('.uno:InsertObjectChart'),
 				'command': '.uno:InsertObjectChart'
 			},
+			{
+				'type': 'bigtoolitem',
+				'text': _UNO('.uno:InsertSparkline'),
```
pedropintosilva, line 999: I think it would be better to avoid having the text from the uno command in the Insert tab because otherwise it would be "Insert Sparkline".

So probably best to replace this with ` _('Sparkline'),` so in does not have the "Insert" and it stays consistent with the other items on that specific tab

browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -1473,7 +1478,74 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
...
+								'type': 'toolitem',
+								'text': _UNO('.uno:UngroupSparklines'),
+								'command': '.uno:UngroupSparklines'
+							}
+						]
+					}
+				],				
+				'vertical': 'true'
+			},
 		];
```
pedropintosilva, line 1549: Nice, yes to have it in the format tab seems for now the best specially because 1. we do not have better place and 2. it keep the 1 - 1 relation with top menu bar :+1: I would maybe opt to have a menu item here called "Sparklines" so it would occupy less space in the format tab and it would work similar to the classic mode counterpart.

I think this can be done similar to how we did with [Statistics](https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L137) or [Format Paragraph](https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L160) etc

pedropintosilva (changes requested):

browser/src/control/Control.NotebookbarCalc.js
pedropintosilva: I think this is ok right @eszkadev ?

Andreas-Kainz:
Will add an icon for colibre. Rizal make a Ramadan break.

pedropintosilva (approved):

pedropintosilva:
Thank you!

PR #4595, opened 2022-04-12
jsdialog: added button in about dialog to open WidgetTestDialog

Rash419:
Change-Id: Ib525515f27640114f5d8bc83314b63f006893d37

* Target version: master

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -283,6 +283,7 @@ m4_ifelse(MOBILEAPP,[true],
             <div style="margin-inline-end: auto;"><div id="lokit-version" dir="ltr"></div></div>
             m4_ifelse(MOBILEAPP,[],[<div id="served-by"><span id="served-by-label"></span>&nbsp;<span id="os-info"></span>&nbsp;<wbr><span id="coolwsd-id"></span></div>],[<p></p>])
             <div id="slow-proxy"></div>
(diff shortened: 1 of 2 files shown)


Rash419:
core patch : https://gerrit.libreoffice.org/c/core/+/132891

browser/html/cool.html.m4
```diff
@@ -283,6 +283,7 @@ m4_ifelse(MOBILEAPP,[true],
             <div style="margin-inline-end: auto;"><div id="lokit-version" dir="ltr"></div></div>
             m4_ifelse(MOBILEAPP,[],[<div id="served-by"><span id="served-by-label"></span>&nbsp;<span id="os-info"></span>&nbsp;<wbr><span id="coolwsd-id"></span></div>],[<p></p>])
             <div id="slow-proxy"></div>
+            <div id="js-dialog">JSDialogs: <a href="javascript:void(app.socket.sendMessage('uno .uno:WidgetTestDialog'))">View widgets</a></div>
```
eszkadev, line 286: I think we want it only in DEBUG build, could you check if there is some macro for it?
Like above for mobile app: `m4_ifelse(MOBILEAPP`

eszkadev (approved):

eszkadev:
It would be nice if that link would close the "About dialog", can we access vex from window object?  vex has some method like closeAll if I remember...

eszkadev (approved):

eszkadev:
before push I'm waiting for core patch for co-22.05

eszkadev:
"iOS against Core co-22.05" passed before and now is broken also in other PRs

PR #4596, opened 2022-04-12
Rename the variable CHROOT to SYSTEMPLATE to match what it means

tml1024:
This fixes my own misguided choice of variable name in the initial
commit of this script.

Also, use the term "symbolic link" or "symlink" instead of "soft-link"
in some comments. That is the correct terminology. (OK, so formally,
there is nothing called "hard-link", but I kept that anyway.)

Change-Id: Ic9bbc01a8a355d370464f2915c4f2db087b4038a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Sigh, I give up, I have more important changes upcoming that will be confused by this or the other way around.

PR #4598, opened 2022-04-13
Fix repair document item on notebookbar writer

pedropintosilva:
Context:
  Repair document on the Notebookbar is currently always a small item
  
  https://archive.org/download/repairdoc-writer-smallitem/repairdoc-writer-smallitem.png
  
  ,it should be a bigtooltiem when the extra additional/optional item
  Sign document is not available:
  - Set it to small item and vertically stacked when hasSigning is true
  - Set it to bigtoolitem when hasSigning is false

Change-Id: I26a8e1b721bf2550127eff1ccfdc44ab4876db52

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -216,25 +216,38 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 			]);
 		}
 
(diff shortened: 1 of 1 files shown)


Rash419 (approved):
I don't know what happend but it is working now , Error on my side sorry for that : )
Looks Good thanks :+1:

PR #4602, opened 2022-04-13
Calc: Remove nonexistent row/columns menu unocommands

pedropintosilva:
Introduced in 71ef338dc543f59b6ed53b0605cb3f0389209996 and in
8e815bdff49154debcfb8a69816c89d71d01603d

this creative invention of new UNO commands is not the way to go.
We cannot invent uno command names and add them to unocommands.js, that
is a generated file, and generated from core. If we cannot pick the
desired string from unocommands.js, as last resort we can write out the
text, e.g. _('Column Width').

Affected:
.uno:ColumnWidthMenu,
.uno:RowHeightMenu,
.uno:ColumnFormatMenu,
.uno:RowFormatMenu

Change-Id: I93adca8b416ae986171af2771adf4007d76d6422


timar (approved):

pedropintosilva:
This change has no impact on cypress tests whatsoever

PR #4611, opened 2022-04-15
cypress: prepare for jsdialog formulabar

eszkadev:
check if canvas exists


eszkadev:
cypress should be fixed in main PR

Issue #4302, opened 2022-02-25
Scroll buttons for Classic / Notebookbar Toolbars right aligned for faster mouse navigation


Andreas-Kainz:
This PR isn't finished. In addition to the new design, I suggest to arrange both buttons next to each other.

The existing solution has the disadvantage that when you want to scroll left/right you have to navigate with the mouse a lot.

I suggest that both buttons are right aligned so you can easier scroll left/right. 

#### Classic Toolbar
![image](https://user-images.githubusercontent.com/8517736/155773338-d841f13e-fef7-4c8a-a1a7-147a05276273.png)

#### notebookbar layout
At notebookbar you have space in the height, so you can have the buttons in two rows.

![image](https://user-images.githubusercontent.com/8517736/155772688-2ae2ee76-f644-4bb7-b042-854d914cce59.png)

Andreas-Kainz:
@pedropintosilva ^ feedback is welcome here or in the PR. 

The movement of the buttons wasn't (yet) added to the PR. But if you give the go (or someones else), I will finish the PR.

PR #4615, opened 2022-04-15
Scroll buttons are right aligned for faster nav

Andreas-Kainz:
As the scroll buttons are left for move left
and right for move right, the mouse movement
is very long to move the toolbars left/right
this PR move all commands to the right

In addition the layout change to some
32px large buttons (bigger in width)
and borders are rounded for faster
visual feedback.

Change-Id: I9cfddfc682fe634985087f550eada32f7ec32478

* Resolves: #4302
* Target version: master 

#### classic toolbar
Screenshot_20220415_231554

#### notebookbar
Screenshot_20220415_234210

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -58,6 +58,13 @@
 	border-spacing: 0px;
 }
 
+/* scroll buttons */
+.hasnotebookbar .w2ui-scroll-left,
+.hasnotebookbar .w2ui-scroll-right {
+	top: 9px;
+	margin-right: 5px;
+}
+
 /* shortcuts bar */
 
 .notebookbar-shortcuts-bar {
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -194,6 +194,30 @@ w2ui-toolbar {
 .w2ui-scroll-left,
 .w2ui-scroll-right {
 	z-index: 1001;
+	background-color: var(--color-background-lighter) !important;
+	box-shadow: var(--color-box-shadow);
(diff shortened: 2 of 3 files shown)


jancborchardt:
So I actually think the current way it works is great and I’m quite hesitant of this change, reasons being:
Right now if you are looking for something you remember is on the left side of the bar and you see it’s overflowing, it is not immediately obvious how you can access it.
You have to know that there is a control for this on the right, and that it controls the left.

By contrast, currently you just see the arrow and you click it. And if it doesn’t overflow it’s not there.

I get the intention of putting controls together and this works well in cases when the controls are the main use case, like for muting/video etc, or forward/back in a browser. For a row with overflow it doesn’t necessarily work as well, as there it’s not the scrolling you mainly want to do, but you just want to reach another thing. :)

Here are 2 examples of how the current way is done elsewhere too:

Firefox does it in the tab bar on overflow:
image

We use a similar pattern in Nextcloud Talk for the participants ribbon (or if there are multiple pages in grid view):
image

Andreas-Kainz:
I thought that the way's between left and right are to long, but I also don't have issues when we have them left and right it's easier to recognize what the buttons good for.

Andreas-Kainz:
Thanks for the feedback. I will close the PR and when I come back with an update about that toppic invite you to the review.

jancborchardt:
> Maybe we can have the option to use the mouse wheel for left and right scrolling when the mouse is located at the toolbar area?

That sounds like a possibly good enhancement, yep!

PR #4616, opened 2022-04-15
writer notebookbar stylesview previews no padding

Andreas-Kainz:
remove padding-top: 5px so there is no scrolling
needed for two rows.

image

Change-Id: I9a0450074cc55486d0f7de8a39b7edff8e66c752

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -400,6 +400,7 @@
 #stylesview .ui-iconview-entry {
 	width: 30%;
 	box-sizing: border-box;
+	padding-top: 0px;
 }
 
 #stylesview .ui-iconview-entry.selected {


pedropintosilva (changes requested):
Thanks! Removing the top padding decreases the need for scroll (even thought we will always have scroll if the document has more styles than the hello-world.odt) but this is also a good improvement in terms of alignment :+1: 

Just one note: Wouldn't be better to instead of adding a padding-top 0 have this removed https://github.com/CollaboraOnline/online/blob/master/browser/css/jsdialogs.css#L521 ? I think it affects other places where we use the icon view (fontwork)

Andreas-Kainz:
Thanks for the feedback. Remove padding-top from jsdialogs.css didn't work well. There is a padding around all fontwork elements. 

So no only .stylesview need `padding-top: 0px;` as defined in the PR.

Andreas-Kainz:
@pedropintosilva do I have to rebase the PR. I don't think that my PR is the issue of the Test.

pedropintosilva:
something happened because your PR has 28 commits

pedropintosilva:
we need only your commits

pedropintosilva:
let me fix that for you

Andreas-Kainz:
I think it can go to master @pedropintosilva

pedropintosilva (approved):
approved, Thanks for checking it out the possibilities here.

PR #4617, opened 2022-04-15
Notebookbar dropdown arrow is visible at hover

Andreas-Kainz:
Change-Id: Idc294d215d7191f28ed6e4090ab5734d44e6bfe7

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -236,10 +236,6 @@
 	padding: 0px 5px !important;
 }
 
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
image

pedropintosilva (approved):
I though we just needed to revert the color change done in https://github.com/CollaboraOnline/online/commit/19715471c7f05507294156a806daf3d96f452b41 to keep the hover effect on the arrow. But now that I think about it that is not so important and since we already have the main button highlight I think it's always good to remove stuff if we can. Thanks

Issue #3790, opened 2021-12-03
Improve Welcome screen


pedropintosilva:
- [Intro](#intro)
- [What we have](#what-we-have)
  - [Desktop](#desktop)
  - [Mobile](#mobile)
- [Room for improvement](#room-for-improvement)
  - [on Desktop](#room-for-improvement-on-desktop)
  - [on Mobile](#room-for-improvement-on-mobile)
- [CO logo](#co-logo)

<h3 id="intro">Intro</h3>
We now have better revamped welcome screen both on desktop and on mobile. The main direction of any improvement here is to make sure we keep it clean and simple, any over the top elements or extra wording should be carefully thought out with these CTA in mind:

1. Communicate that this is a new release
* Make more discoverable that we do have Apps (Android, ChromeOS, iOS)
  * To get more downloads and users to test those 
2. Communicate the release notes
* In a succinct way (a short paragraph)
* Make more discoverable the web page where all those release notes are at
  * To get more people involved in the insides of CO
    * So to avoid outdated bug reports or questions in the forum
    * So they can share them with everyone else
    * So they can start to test those new improvements/features 
5. Get involved
* Make more discoverable our community website
* Attract more people to contribute and hack in all the different fronts
6. Show that there are an option for enterprise-ready versions
* To decrease the number of unaware CODE users that end up choosing other tools. Basically to about "Oh that's cool but I really need some enterprise/professional version with official support for my company/organization"

Note: very important to use active verbs and to avoid to avoid multiple colors per screen with different elements fighting for attention

<h3 id="what-we-have">What we have</h3>

<h4 id="desktop">Desktop</h4>
https://user-images.githubusercontent.com/65948705/144588044-5e9c0601-9646-414a-ae31-67f2fc327d56.mp4

<h4 id="mobile">Mobile</h4>
https://user-images.githubusercontent.com/65948705/144588040-f48e0fde-2d4f-41aa-b5d4-85ce9a1990de.mp4

https://user-images.githubusercontent.com/65948705/144588043-c9bb34dc-d58e-44e8-8706-cfcb5fdab9b9.mp4

<h3 id="room-for-improvement">Room for improvement</h3>
As seen above there are still room for improvement:

<h4 id="room-for-improvement-on-desktop">Room for improvement on Desktop</h4>
- [ ] Desktop: Next btn: Maybe re-using the buttons from the JSdialogs are not as good as I thought, maybe we can use arrows instead
[...]

PR #4619, opened 2022-04-16
Update welcome screen for smaler sizes and mobile

Andreas-Kainz:
* Resolves: #3790
* Target version: master 

devices-mobiles.css
- rules for mobile wasn't needed
  everywthing is in welcome.css

welcome.css
- screen behind welcome dialog use --color-overlay
  as at every dialog
- no fixed width and height
  max-width and max-height was the old width/height
  width and height are now 95%
  dialog window was center justify
- content width is 96%
  when smaler than 500px padding was added to <p>
- fig get top an bottom margin
  so now all fig's have the same height
- hyperlinks use --color-primary
  and not underline

Change-Id: If921bc2abc3977930f7d60256d8011d8986a22c9

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -904,14 +904,3 @@ div#fontstyletoolbox + div#style.mobile-wizard {
 	margin: 2px;
 	font-size: 29px;
 }
-
-/* Welcome Dialog */
-.iframe-welcome-modal {
-	display: block;
-	border-radius: 0;
-}
-.iframe-welcome-modal,
-.iframe-welcome-content {
-	width: 100%;
-	height: 100%;
-}
--- a/browser/css/welcome.css
+++ b/browser/css/welcome.css
@@ -1,4 +1,5 @@
-.iframe-welcome-wrap {
+.iframe-welcome-wrap,
+.iframe-welcome-content {
 	position: absolute;
 	z-index: 1106;
 	display: flex;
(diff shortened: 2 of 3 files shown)


Andreas-Kainz:
#### mobile
Screen Shot 2022-04-17 at 00 19 47

Andreas-Kainz:
#### desktop
image

Andreas-Kainz:
What I would like to have support for color-palette.css (and therefor branding.css would also be usefull, but maybe this is already available) to use var's instead of hard-coded stuff.

Andreas-Kainz:
My proposal is to have the navigation within one flex box with left and right navigation

new

But this would need first an +1 and than a separate PR.

pedropintosilva:
> My proposal is to have the navigation within one flex box with left and right navigation

I think if on one hand it can look tidier on the other hand it would require an extra effort from the user side to interact with those because:
- They would be even more down (and distant from the other main action: close)
- They would be more discreetly placed following a bottom banner and thus harder to spot
- They would be placed more near to edges increasing yet again the mouse travelling

Andreas-Kainz:
The movement for the buttons, I would make a separate PR. So this PR is ONLY for make the dialog mobile friendly.

pedropintosilva (approved):
Thanks, ideally we would have just on PR (good :) ) but with more than 1 commit  for all these changes but nevermind for now :) 

Looks nice!


Screen Shot 2022-04-19 at 11 37 16

I was not sure in using this kinda of modal look and feel for mobile (afraid of wasting space specially on small devices) but I think if you prefer I'm not opposed to it. I also noticed that the fig is now cropped  and the close overlaps it (so we might want to look into that)

PR #4622, opened 2022-04-18
private/ash/misc

Ashod:
- wsd: remove argument default-value
- wsd: fallback only when COOLStatusCode is missing
- wsd: minor cleanup of send-frame members
- wsd: remove identical simulateSocketError override


dennisfrancis (approved):
Looks good. Thanks!

PR #4632, opened 2022-04-20
wsd: use the default DH parameters in SSL

Ashod:
OpenSSL 3 deprecated the manual DH parameter
functions. Instead, it encourages the use
of the built-in parameters. Since this
API also works on the 1.1 version, there is
no need to support multiple implementations.

Change-Id: I900cc11c3ca09f1d85b7d88cfbf537d802f69846

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (changes requested):
Unfortunately on old distros that we support (e.g. CentOS 7) there's still OpenSSL 1.0.2.

timar (approved):

PR #4634, opened 2022-04-21
wsd: do not set default close reason

lpranam:
we preserve original reason,
so setting default reason means always using default reason
and ignoring any other reason

Change-Id: I837a13a129aa759a319406affc6e89425bf31361

* Target version: master 

### Checklist


Ashod (approved):
Thanks @lpranam, looks good. Let's wait for the CI to be sure it doesn't break anything.

PR #4635, opened 2022-04-21
We do not use tunneled sidebar any more

timar:
Change-Id: If3dc1dae35250c54d62e69396aea7df51873d4cb


eszkadev (commented):
@timar but what about dialogs? we still have many of them tunneled

timar:
> @timar but what about dialogs? we still have many of them tunneled

If they have icons on them (it's rare), colibre_svg will be used. We do not ship collabora_svg icon set in brand packages any more (in 21.11 and 22.05). I don't remember who removed collabora_svg icon set from brand packages but as matter of fact they are not there, so it has no effect, if we configure collabora_svg.

eszkadev (approved):

PR #4639, opened 2022-04-21
calc header: avoid reselecting selected row/col range

lpranam:
if right clicked on already selected row/column range,
don't reselect the right clicked row/column

Change-Id: I718a0a727913c38bc506fd2b808fba8feaebb0c8

* Target version: master 

### Checklist


eszkadev (approved):
nice, works as expected

Issue #4620, opened 2022-04-17
Calc: Misaligned Function wizard button next to formula bar.


thebearon:
See screenshot:
![calc_misaligned_formula_bar](https://user-images.githubusercontent.com/19438782/163702307-2cfb0964-80bb-4536-a00b-f518611e2924.png)
(this is with branch `distro/collabora/co-21-11`, but it's the same with `master` branch)

This used to be fine, the alignment issue started with 8d350b28453eee6d84a8bd2b9285d5664706f342.
```
Author:     Andreas Kainz <kainz.a@gmail.com>
AuthorDate: Mon Mar 7 00:21:52 2022 +0100
Commit:     Andreas Kainz <andreas_k@abwesend.de>
CommitDate: Tue Mar 8 08:33:48 2022 +0100

    Mobile Search use var colors and is vertical aligned
```
CC @Andreas-Kainz @pedropintosilva.

Andreas-Kainz:
Thanks will check.

Andreas-Kainz:
And fix during the next day's.

pedropintosilva:
Thanks

pedropintosilva:
the merged PR is for 21.11 since Master now is pared with core 22.05 (and more changes from formula bar are coming from @eszkadev making it rendered in the client side )

PR #4642, opened 2022-04-22
private/pedro/formulabarfixes

pedropintosilva:
commit 7825802342fdb5f20a752cd0765589022423b3e2 (HEAD -> distro/collabora/co-21-11)
Date:   Fri Apr 22 12:12:23 2022 +0200

Formulabar: Remove FF-only rules

It's not needed anymore plus it was also affecting desktop.

----

commit e7b577e7566812c443f060c3c7c31275ce97ca67
Date:   Fri Apr 22 12:10:16 2022 +0200

formulabar InputAddress: Use the same border style as used in core

This might need to be removed once we change the way we render formulabar
- So if it is render natively we can just use a more nuanced color
across the board

----

commit 79ea15df320095c54a965bd1e1d8d17609799e7b
Date:   Fri Apr 22 12:06:37 2022 +0200

Formulabar inputAdd: Add missing margin left

Before it seems that only FF was getting that margin
- Yes even though that rule is supposed to only affect mobile

Fix it. And use the same margin everywhere.

----

commit bcf4234cd422494f29b6b1cb31ba58d348f5e0e8
Date:   Fri Apr 22 11:50:01 2022 +0200

Fix formulabar and formulabar items' misalignment

It seems Formulabar and surrounding items were being aligned with
multiline mode (when input is expanded) in mind but:
- This affecting the standard view (when is collapsed) and causing
misalignment
- On top of that, when the input is expanded the alignment was not 100%
effective

https://archive.org/download/formulabar-misalignment/formulabar-misalignment.png

Fix it:
- Align everything to the top but only when the input field is expanded
- Remove padding for alignments that can get automatically aligned, and
use padding just for the function icon and just when needed (multiline)
- Fix AdressInput's size padding and margin

fixes issue #4620

----

commit 3998ac1e4f318cb4ed12ac2425eb6fdc10caeb77
Date:   Fri Apr 22 10:09:27 2022 +0200

Fix formulabar background discrepancy

For now and because formula bar comes from core side and thus it's
via canvas we have no way to change its background (it comes in white).
Currently we have a color discrepancy:
https://archive.org/download/formulabar-bg-discrepancy/formulabar-bg-discrepancy.png

Fix it. Remove background of its container element which we were
forcing another color other than white and let it inherit the #fff from
wui toolbar.

### DIFF
--- a/browser/css/toolbar-mobile.css
+++ b/browser/css/toolbar-mobile.css
@@ -1,6 +1,5 @@
 @-moz-document url-prefix() {
 	#formulabar #formulaInput, #formulabar #addressInput{
-		margin-left: 5px;
 		color: var(--color-main-text);
 		background-color: var(--color-main-background);
 		border: 1px solid var(--color-border) !important;
--- a/browser/css/toolbar-mobile.css
+++ b/browser/css/toolbar-mobile.css
@@ -1,14 +1,3 @@
-@-moz-document url-prefix() {
-	#formulabar #formulaInput, #formulabar #addressInput{
-		color: var(--color-main-text);
-		background-color: var(--color-main-background);
-		border: 1px solid var(--color-border) !important;
-		border-radius: 0px;
-	}
-	#toolbar-up{
-		top: -1px;
-	}
-}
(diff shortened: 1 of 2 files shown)


pedropintosilva:
better to try to fix this for 21.11 since Master now is pared with core 22.05 (and more changes from formula bar are coming from @eszkadev )

PR #4644, opened 2022-04-22
Avoid using only "real" paths for the tmpfont thing

tml1024:
If I have used a path with symlinks in it when changing directory to
my build directory, what gets put in config.status as ac_pwd is that
path, not a realpath version. That then propagates to
ac_abs_to_builddir and to SYSTEMPLATE_PATH, which is what Makefile.am
passes for the --o:sys_template_path option to coolwsd.

Change-Id: I7575120090986e6207497c5ce740aedd6075e48f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eneiluj:
Nice! Tested on my side. It works fine.

tml1024:
Let's consider Julien's comment as a review;)

PR #4647, opened 2022-04-22
Private/eszkadev/port nb switch

eszkadev:
port to 21.11 of ui mode switch


eszkadev:
build bots were wrong - I will recreate PR

PR #4648, opened 2022-04-22
21.11 port nb switch

eszkadev:
this ports ui mode switch


eszkadev:
rash will do that later

PR #4650, opened 2022-04-23
calc: remove unused code after jsdialog conversions

eszkadev:
* old jsdialog autofilter
* tunneled formulabar


merttumer (approved):
Works as expected

Issue #4382, opened 2022-03-11
Add a dark mode


florealcab:
Application is so beautiful, but is missing a dark mode.

An option to choose theme between light and dark should be a good thing for users eyes :)

danir-de:
Doing this automatically via reading the current system dark-mode would be even better: If the user is in dark mode, automatically use dark mode, if not use light mode.

florealcab:
Yes, an automatic behavior is a good practice. Some times, this setting could not be detected (on some linux systems for example) so the common way to achieve that is an option with 3 choices :
- Automatic (follow system setting)
- Dark
- Light

If we use it on nextcloud (my case), could also follow the nextcloud parameter for dark theme

zak39:
Please, if you want to make a dark mode, you should not change the background only.

The background color, the color of the paper and the color of the icons should be studied.

I see many apps that only change the background color.

A good example, the "grafikart website" : https://grafikart.fr/

When we switch to dark mode, the fonts, buttons and many more change too.

It's very pleasant to read

What do you think ?

danir-de:
@zak39 I understand where that idea is coming from, but that is a whole other topic.

Also applying a dark theme to the document would make it unusable for most people, since documents cannot be shared or printed in dark mode and would have to constantly be converted between a light and dark color scheme. 

This might work for "simple" proprietary apps, but for full document suits where a user can easily change the text color to black, that color has to stay black and cannot be "converted" into another color just for a dark mode, since that would completely destroy a document. 
Even if you would take that into account... How should that black text be displayed on a dark mode page? Should it be converted to white or should it stay black? There is another element that's darkgray. Should that be displayed as lightgray? What happens with the dark transparent image, one user has inserted in the document? You can't simply invert that!

I think that this issue should firstly focus on an UI-only dark mode, if you want, you can tackle the problem in another issue later - first things first!

zak39:
Yes, I agree with you :wink: 

This is precisely one of the mistake I saw on some applications.

I'm sorry, I misspoke ^^'

I meant to say, if you set the same black color as the GitHub color for the background. I think the contrast will be high and it's very hard to focus on the document

Do you know what I mean ?

Andreas-Kainz:
The JS UI stuff use now var-colors from https://github.com/CollaboraOnline/online/blob/master/browser/css/color-palette.css 

[...]

PR #4654, opened 2022-04-23
Add color-scheme-dark.css support

Andreas-Kainz:
@media (prefers-color-scheme: dark) setting
is available by use color-scheme-dark.css
instead of color-scheme.css

sure the scheme need some update
but there are other open points like
use dark color for document
use dark colibre icons

this is only the .css change
so no full dark mode update

* Resolves: #4382 
* only use dark color-sheme-dark.css  
* Target version: master 

Screenshot_20220423_195656

Change-Id: I5c2adcbb75583a952f419af507d7cfe5c1343e3c

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -122,6 +122,7 @@ COOL_CSS_LST =\
 	$(srcdir)/css/leaflet-spinner.css \
 	$(srcdir)/css/selectionMarkers.css \
 	$(srcdir)/css/color-palette.css \
+	$(srcdir)/css/color-palette-dark.css \
 	$(srcdir)/css/cool.css \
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
@pedropintosilva if there is enough time before the next release, let's push it. why? because it's way easier to find issues with css var's.

pedropintosilva (approved):
Thanks, there are some things that need fix within the dark pallet. Let me merge this and then I will add those as new issues or as part of the existent issue

Andreas-Kainz:
> Thanks, there are some things that need fix within the dark pallet. Let me merge this and then I will add those as new issues or as part of the existent issue

yes please separate issues for the theming. but now it's way easier to switch and see the issues.

PR #4655, opened 2022-04-24
scroll buttons use var colors and update design

Andreas-Kainz:
scroll buttons when the window size is to small,
use var colors

the design was updated to be more like a regular
button. width was larger, 32px like other commands
vertical line like we have in toolbars

Change-Id: I9448c596f473dff29e322b040dd95755bf5aefb6

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -194,6 +194,19 @@ w2ui-toolbar {
 .w2ui-scroll-left,
 .w2ui-scroll-right {
 	z-index: 1001;
+	width: 32px;
+	background-color: var(--color-background-lighter) !important;
+	box-shadow: none;
+}
+.w2ui-scroll-left {
+	border-right: 1px solid var(--color-border-dark);
+}
+.w2ui-scroll-right {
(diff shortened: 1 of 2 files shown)


pedropintosilva:
I have rebased and force-pushed

pedropintosilva:
However I'm not sure about these changes:
* on the NB (tabbed view)  the additional left lateral border feels unease since we do not have top border in the nb tab
image
* On the classic (compact view) : the indicator looks like it's floating (with gaps on the bottom and top). If the floating look is intended we should have such a huge bezel (e.g: left border ).
image

Andreas-Kainz:
The scroll button should look like a regular button not something alien.

So I made the button the same width than regular ones and a separator. Sure I'm not 100% happy with the design, but I wand to have something that fit's better into the overall design and a button didn't have a border or something.

pedropintosilva:
> * we should have such a huge bezel (e.g: left border ).

we shoul**dn't**

Andreas-Kainz:
As I didn't have a good proposal I will close the PR and make a new one ONLY for the use of var colors. The layout want be changed as long as we don't have a good idea.

PR #4656, opened 2022-04-24
.w2ui-button use var color for :hover

Andreas-Kainz:
image

Change-Id: I3b77ba1e99dee24e74f9f839de5590d0a0def939

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -593,6 +593,9 @@ button.leaflet-control-search-next
 	border: 1px solid transparent;
 	color: var(--color-main-text);
 }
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
I think better to don't duplicate CSS selectors that target the same element and state. Why not instead, add the new declaration to : https://github.com/CollaboraOnline/online/blob/master/browser/css/toolbar.css#L596

Andreas-Kainz:
this can be pushed. Suggestion from @pedropintosilva was done. @Rash419

Rash419 (approved):
Thanks

PR #4663, opened 2022-04-25
framed.doc.html: fix: cannot change UI view (master)

Rash419:
Change-Id: Ifa842f64ed70137f06ebdfc053fa5a3afd2f4185

* Target version: master

### DIFF
--- a/browser/src/map/handler/Map.WOPI.js
+++ b/browser/src/map/handler/Map.WOPI.js
@@ -451,9 +451,7 @@ L.Map.WOPI = L.Handler.extend({
 			}
 		}
 		else if (msg.MessageId === 'Action_ChangeUIMode') {
-			if (msg.Values && msg.Values.Mode !== null && msg.Values.Mode !== undefined) {
-				this._map.fire('changeuimode', {mode: msg.Values.Mode, force: false});
-			}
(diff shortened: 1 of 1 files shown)


browser/src/map/handler/Map.WOPI.js
```diff
@@ -449,7 +449,11 @@ L.Map.WOPI = L.Handler.extend({
 		}
 		else if (msg.MessageId === 'Action_ChangeUIMode') {
 			if (msg.Values && msg.Values.Mode !== null && msg.Values.Mode !== undefined) {
-				this._map.fire('changeuimode', {mode: msg.Values.Mode, force: false});
+				if (msg.Values.Mode === 'notebookbar') {
```
eszkadev, line 452: shouldn't be `msg.Values.Mode === 'notebookbar' || msg.Values.Mode === 'tabbed'` ?

Rash419: why would we need tabbed ? https://github.com/CollaboraOnline/online/blob/bef3556a1fbdba334abc2e6f67ec8d05f7c67785/browser/html/framed.doc.html#L99
mode value will we either classic and notebookbar

eszkadev (approved):
looks good

PR #4665, opened 2022-04-25
Configurable Delayed Save and Upload

Ashod:
Adds support to configure the minimum time between save and upload (in case of retrying failed attempts).

- wsd: add min_time_between_save_ms config
- wsd: add min_time_between_upload_ms config
- wsd: config: merge auto- and idle-save comment and make uint
- wsd: move the autosave interval into SaveManager
- wsd: cosmetic and logging 
- wsd: move minTimeBetweenSaves into SaveManager 
- wsd: move minTimeBetweenUploads into StorageManager 
- wsd: remove lastUploadTime as it is redundant

### DIFF
--- a/coolwsd.xml.in
+++ b/coolwsd.xml.in
@@ -52,6 +52,7 @@
         <limit_load_secs desc="Maximum number of seconds to wait for a document load to succeed. 0 for unlimited." type="uint" default="100">100</limit_load_secs>
         <limit_store_failures desc="Maximum number of consecutive save-and-upload to storage failures when unloading the document. 0 for unlimited (not recommended)." type="uint" default="5">5</limit_store_failures>
         <limit_convert_secs desc="Maximum number of seconds to wait for a document conversion to succeed. 0 for unlimited." type="uint" default="100">100</limit_convert_secs>
+        <min_time_between_saves_ms desc="Minimum number of milliseconds between saving the document on disk." type="uint" default="500">500</min_time_between_saves_ms>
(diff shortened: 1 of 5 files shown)


mmeeks (approved):
Some thoughts. Not so impressed by the cosmetic changes.
Otherwise - it looks ok. What happens when we hit save manually - and yet we did that sub <timeout> earlier ? looks like that would be dropped, and not re-done after a short delay. Similarly if we save before closing a doc (and why not) - then make a small tweak - looks like we would not-quite busy-loop there waiting for the chance for autoSaveAndStop to do something.

Nevertheless - I guess all this code gives a Nextcloud time to get its locking done and its thumbnailing actioned between us saving and saving again - and not getting confused.

Ashod:
Thanks @mmeeks, great feedback!

> What happens when we hit save manually - and yet we did that sub earlier ? looks like that would be dropped, and not re-done after a short delay. 

Manual saving should be "forced" so it wouldn't be subject to this logic. Unfortunately, we currently have a single `force` flag which is used both to clobber the version in storage and to force saving. Plan is to separate them and then it should be trivial to make manual saves forced, without conflating it with clobbering in storage (which obviously is dangerous).

Until then, it shouldn't be common to have a modify + save issued within the 500ms delay (modify is necessary because we don't save when nothing is modified by default).

> Similarly if we save before closing a doc (and why not) - then make a small tweak - looks like we would not-quite busy-loop there waiting for the chance for autoSaveAndStop to do something.

For closing, we are generally more aggressive. I'll double check that we don't become a hostage to this delay.

PR #4667, opened 2022-04-25
jsdialog: fontwork dialog fix

eszkadev:
fixes selection in fontwork dialog - regression from last jsdialog updates


lpranam (approved):
works as expected :)

PR #4668, opened 2022-04-25
Remember last UI mode

eszkadev:
This also makes us sure we are in sync due to
single point of knowledge about UI Mode


merttumer (approved):
works as expected

PR #4670, opened 2022-04-25
Remember last UI mode (21.11)

eszkadev:
This also makes us sure we are in sync due to
single point of knowledge about UI Mode


eszkadev:
was approved for master already

PR #4678, opened 2022-04-25
Sidebar scrollbar-width: thin

Andreas-Kainz:
The sidebar isn't a main navigation area,
so the width of the scrollbar should be thin.

In addition the width of the sidebar is "limited"
so an large scrollbar didn't help that much

Change-Id: I8a8ab9ca7866004c8e581c5bbe88125830e2ba7b

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -199,7 +199,7 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	overflow-x: hidden;
 	overflow-y: auto;
 	z-index: 1200;
-	scrollbar-width: auto;
+	scrollbar-width: thin;
 	scrollbar-color: var(--color-border) transparent;
 }
 


pedropintosilva (commented):
Thanks, did you test these changes in multiple browsers on windows? We had a bug on he way sidebar scrollbar was being displayed depending on the browser on windows. This https://github.com/CollaboraOnline/online/commit/9f4d9c3f7e1d3e381dc53763bed6b3f143823247 has fixed #3725 . If this only improves, great but it would be good to test it thoroughly

Andreas-Kainz:
ok I can test on windows, linux and with different browsers. In general the right sidebar in impress for example has also a thin scrollbar.

pedropintosilva:
sorry for the delay on this.

>The sidebar isn't a main navigation area,
> so the width of the scrollbar should be thin.

Agreed, but the scrollbar on the sidebar is already narrow when compared with the document's scrollbar 

> In addition the width of the sidebar is "limited"
> so an large scrollbar didn't help that much

I understand that the scrollbar occupies space but testing this PR makes it extremely difficult to manipulate the scrollbar with a mouse via grabbing it
image

PR #4679, opened 2022-04-25
Sidebar RTL fixes

Andreas-Kainz:
border on left and right so it work also for RTL languages

image

Change-Id: Ic973cfee4c8fd563008477ab38a9e4e3e69c8300


pedropintosilva:
so in this case would be `border-inline-start`

pedropintosilva (approved):
:heart:

PR #4683, opened 2022-04-26
private/pedro/backporting from master

pedropintosilva:
- Remove toggle prefix from status bar and ruler
- Fix repair document item on notebookbar writer
- Fix leaflet-control alignment
- Notebookbar: Fix SaveAs Id, Capitalize Each Word
- Notebookbar: Fix rev-history Id, Capitalize Each Word

### DIFF
--- a/browser/css/leaflet.css
+++ b/browser/css/leaflet.css
@@ -139,7 +139,7 @@
 .leaflet-right .leaflet-control {
 	float: right;
 	margin-right: 10px;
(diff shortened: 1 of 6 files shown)


pedropintosilva:
all commits tested locally :+1:

pedropintosilva:
interesting cypress test fails due to 

```bash
The response we received from your web server was:

  > 404: Not Found
```

PR #4684, opened 2022-04-26
notebookbar: fix ui mode switch

eszkadev:
previously without data in localstorage it started with
compact mode always


Rash419 (approved):
Works as expected thanks : )

PR #4691, opened 2022-04-28
Socket Logs with Context

Ashod:
- wsd: make socket FD log prefix implicit and cleanup logs
- wsd: improved socket logging
- wsd: simplify socket read logging
- wsd: better logging of unix socket binding
- wsd: cleanup log initialization

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -375,20 +375,10 @@ namespace Log
         Static.setLevel(level);
 
         const std::time_t t = std::time(nullptr);
-        oss.str("");
-        oss.clear();
-
-        oss << "Initializing " << name << '.';
-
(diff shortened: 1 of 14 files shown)


mmeeks (commented):
Well - it looks reasonable; the logPrefix stuff is clever - but I really don't like optimizing for the verbose logging case. Lets not allocate _logPrefix members and take another set of allocations at construction and hold that all around constantly if we can avoid it and just append eg. '#' << GetFD() or whatever to the stream in the virtual function instead. Safer too in case members are mutated during use.
Thanks.

mmeeks (approved):
Thanks Ash - looks lovely =)

PR #4693, opened 2022-04-28
Prevent errors on document load

eszkadev:
Prevent errors on load when section may be missing
    
    In the past I've seen very often that for example
    commentList section was missing but used without any
    check.
    
    This commit checks if section exists before use and
    warns if it doesn't so we can see that in the console
    logs.

Second commit:
    removed unnecessary console.error with notebookbar state used for debugging and commited by mistake


lpranam (approved):

gokaysatir:
Thanks @eszkadev :)

PR #4697, opened 2022-04-29
backports 21.11: jsdialog improvements

eszkadev:
- build drawingarea even if is empty (so we can later apply update events)
- remove old hack which added "table-" prefix to some widgets
- implemented textarea with raw keyboard handling (by the core, user cannot type directly into the field, only see the result from server) `not used in 21.11`
- implemented buttons with "symbol" instead of image `not used in 21.11`
- implemented `toolitem` which can easily share code with eg. menubar `customtoolitem` and `map.dispatch(command)`

This PR is for unifying jsdialog code between 21.11 and master so cherry-picking will be easier

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -18,6 +18,11 @@ button.jsdialog {
 	opacity: 0.7;
 }
 
+/*limit icon to button height*/
+button.jsdialog img {
+	max-height: 100%;
+}
+
 .annotation-btns-container,
 .vex-dialog-buttons {
 	display: flex;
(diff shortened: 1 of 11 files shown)


eszkadev:
@pedropintosilva it removed "table-" prefix from containers. I didn't noticed change (I think we no longer use that in CSS)  but would be good to quickly look if there is no regressions in sidebar/mobilewizard.
Nothing should change after this PR, this is just a code change.

eszkadev:
not needed

eszkadev:
closed by mistake

pedropintosilva:
Thanks  for the heads up @eszkadev I will test it. I also noticed with your changes (on master) there are a couple of things we can remove and/or simplify. #4700 aims to do just that.

pedropintosilva (approved):

PR #4698, opened 2022-04-29
formulabar: fix focus also on mobile

eszkadev:
- also hide autosum button - it was hidden before conversion


merttumer (approved):
Looks good

PR #4700, opened 2022-04-29
private/pedro/jsdialogs rm css n fix bg n radiobtn

pedropintosilva:
- JSdialog, p element: Remove #table- from CSS it was removed from JS
- Macro dialog: remove unecessary CSS rules
- JSdialog: Don't add margins to multiple inner containers
- Do not add padding too all td jsdialog cells
- JSdialog: fix disabled radio buttons border
- Dialogs should have white background

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -733,35 +733,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	font-size: var(--default-font-size);
 }
 
-#MacroSelectorDialog #table-box2 {
-	width: auto;
-}
-
-#MacroSelectorDialog #table-box2 > .row > .cell.jsdialog {
-	width: auto;
-}
-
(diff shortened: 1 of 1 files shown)


eszkadev:
Macro dialog should be checked, can we just remote these rules?

eszkadev:
BTW Macro dialog is missing in notebookbar, it was in the hamburger menu

pedropintosilva:
> BTW Macro dialog is missing in notebookbar, it was in the hamburger menu

Yes I noticed when I tried to test both macro dialogs (warning and the selection dialog)

pedropintosilva:
> Macro dialog should be checked, can we just remote these rules?

what is remote? (maybe a typo = remove?) . I have tested everything seems fine (a part from other color regression from css vars)

gokaysatir (approved):
Thanks for the improvements @pedropintosilva :)

PR #4708, opened 2022-04-30
wsd: return status from writeOutgoingData

Ashod:
Without knowing whether the write succeeded
or failed, we cannot trust errno has our
error or some earlier and unrelated error.

This was caught when there were two sockets,
one of which disconnected. The write to the
disconnected one returned -1 and set errno
to ECONNRESET. We subsequently wrote to the
second socket, which succeeded. However,
because errno wasn't reset, and since
writeOutgoingData didn't return anything
to indicate the success, errno's ECONNRESET
value meant the second socket was also
disconnected, which was incorrect.

writeOutgoingData now returns the last return
value from writeData so we can make informed
decision as to whether to check errno or not.

Also, to avoid incorrecly propagating errno,
we now capture errno only when readData and
writeData return -1. This has the nice
side-effect that we reset errno to 0 when
no errors occur during our call.

Change-Id: I911b31390f37cc71938bc4a6ae75393dbf24bb9d

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1106,7 +1106,7 @@ public:
 #if !MOBILEAPP
         // SSL decodes blocks of 16Kb, so for efficiency we use the same.
         char buf[16 * 1024];
-        ssize_t len;
+        ssize_t len = 0;
         int last_errno = 0;
         do
         {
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Not a huge fan of the last_errno thing from a readability perspective - but I forget what the alternative was: an errno push/pop inside our streaming I guess ;-) anyhow - looks good - thanks ! =)

mmeeks:
Lets get it into the stable branches too if we can =) thanks !

PR #4712, opened 2022-04-30
Don't try to hide comment section if mobile wizard not enabled.

mmeeks:
Change-Id: I37069fd83cf421b762c82a5584964a09961afaef

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.MobileWizardPopup.js
+++ b/browser/src/control/Control.MobileWizardPopup.js
@@ -55,7 +55,7 @@ L.Control.MobileWizardPopup = L.Control.extend({
 	_hideWizard: function () {
 		this.removeContainer();
 
-		if (app.sectionContainer) {
+		if (window.commentWizard === true && app.sectionContainer) {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks good

eszkadev:
But is that fix for desktop bug, right? I think this code shoudn't be executed then...
Here we add `mobilewizard` component without any condition:
https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L91
So we should check if it is possible to add `mobilewizard` only for mobile.

On desktop we use only `Control.MobileWizardPopup.js` which is separate component - used for comment bubbles, to show list of comments.

eszkadev:
in master probably fixed by this: https://github.com/CollaboraOnline/online/commit/b713340f245ecb0322b4d826b9db71e62eb61d09

timar:
b713340 was cherry picked to co-21-11, this PR is obsolete.

PR #4720, opened 2022-05-02
Allow specifying the remote_font_config URI in the remote_config

kendy:
This way, there is only one entry point for all the remote
configuration, and the remote_font_config does not have to be specified
in coolwsd.xml, in case the remote_config JSON specifies that.

Signed-off-by: Jan Holesovsky <kendy@collabora.com>
Change-Id: I857bab3d300ddb5c0c3df0a9f62eda20f43c3ff7


mmeeks (approved):
Looks good to me; thanks Kendy ! =)

PR #4723, opened 2022-05-02
docker: map domain env variable to alias_groups

Rash419:
show a verbose warning about using domain

Change-Id: I41fdea3419eb2719fa665a62de33e3a237cf3df8

This is what the message look like in docker logs:
image

### DIFF
--- a/docker/from-packages/scripts/start-collabora-online.pl
+++ b/docker/from-packages/scripts/start-collabora-online.pl
@@ -13,11 +13,14 @@ use strict;
 # create the part of the config that contains aliases based on the aliasgroupN envvars
 sub generate_aliases() {
     my $output = '';
+    if ($ENV{'domain'} && $ENV{'aliasgroup1'}) {
+        print "WARNING: Both aliasgroupX and domain are provided, aliasgroupX takes the precedence where X=1,2,3...\n";
+    }
+
     foreach (sort keys(%ENV)) {
         if (/^aliasgroup/) {
             my $value = $ENV{$_};
             my @aliases = split(',', $value);
-
             if (@aliases) {
                 $output .= "                <group>\n";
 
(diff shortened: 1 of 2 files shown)


docker/from-packages/scripts/start-collabora-online.pl
```diff
@@ -34,9 +35,42 @@ ()
 
                 $output .= "                </group>\n";
             }
+        } elsif (/^domain/) {
+            my $value = $ENV{$_};
+            $message .= "WARNING: The 'domain=$value' is deprecated.\n For your convenience, we interpret it as the following:\n";
+            my @hosts = split('\|', $value);
+            if (@hosts) {
```
kendy, line 42: I suppose the WARNING should read a bit differently when the domain= is completely empty (ie. if (@hosts) is false)?

docker/from-packages/scripts/start-collabora-online.pl
```diff
@@ -34,9 +35,42 @@ ()
...
+                    $message .= "This means that people from $hosts[0] will not be able to access documents from ";
+                    for ($b = 1; $b < @hosts; $b = $b + 1)
+                    {
+                        $message .= "$hosts[$b], ";
+                    }
+                    $message .= " and vice versa";
+                }
+                $message .= ". If you want to allow the access instead, use this configuration instead:\n     aliasgroup1=";
+                $i = 0;
+                foreach(@hosts) {
```
kendy, line 61: You can use join() instead of this foreach()

docker/from-packages/scripts/start-collabora-online.pl
```diff
@@ -34,9 +35,42 @@ ()
...
                 $output .= "                </group>\n";
             }
+        } elsif (/^domain/) {
+            my $value = $ENV{$_};
+            $message .= "WARNING: The 'domain=$value' is deprecated.\n For your convenience, we interpret it as the following:\n";
+            my @hosts = split('\|', $value);
+            if (@hosts) {
+                my $i = 0;
+                foreach (@hosts) {
+                    $i++;
```
kendy, line 45: Should we also add a check here for the case there is some form of regexp in the domain= ?  Or will it work in the alias_groups?

docker/from-packages/scripts/start-collabora-online.sh
```diff
@@ -36,8 +36,7 @@ fi
 
 # Replace trusted host and set admin username and password - only if they are set
 if test -n "${domain}"; then
```
kendy, line 38: I guess we should actually turn this into a 'if test -n "${domain}" -o -n "${aliasgroup1}"; then' ... so that it is not executed twice when there is both domain and aliasgroupX in there?  Or better - ignore the domain when there is an aliasgroupX, but warn that both are provided, and the aliasgroupX takes the precedence?

kendy (commented):
Looks good in general, commented about few corner cases :-)

docker/from-packages/scripts/start-collabora-online.pl
Rash419: In this case join() does not work

docker/from-packages/scripts/start-collabora-online.pl
Rash419: it will work

PR #4741, opened 2022-05-05
A more robust resizeobserver polyfill detection.

mmeeks:
Instead of messing with parsing useragent - lets just check whether
the type exists.

Change-Id: Ib654dc71ec12fd643c906bc50fa9dff1791dc14e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thank you :)

PR #4744, opened 2022-05-05
issue 4730 update: check for new versions

hcvcastro:
- infobar: remove unused functions
- infobar: show a snackbar message
- infobar: clean up unused code
- infobar: fire 'infobar' message
- wsd: infobar: add config value "fetch_update_check"
- wsd: infobar: fetch and save the latest version data
- wsd: infobar: send latest version to the client side
- wsd: infobar: clean up unused code

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4954,6 +4954,9 @@ int COOLWSD::innerMain()
 
     LOG_DBG("Initializing DelaySocket with " << SimulatedLatencyMs << "ms.");
     Delay delay(SimulatedLatencyMs);
+
+    const auto fetchUpdateCheck = std::chrono::duration_cast<std::chrono::milliseconds>(
+        std::chrono::hours(std::max(getConfigValue<int>("fetch_update_check", 10), 0)));
 #endif
 
     ClientRequestDispatcher::InitStaticFileContentCache();
@@ -5056,6 +5059,9 @@ int COOLWSD::innerMain()
 #endif
 
     const auto startStamp = std::chrono::steady_clock::now();
+#if !MOBILEAPP
+    auto stampFetch = startStamp - (fetchUpdateCheck - std::chrono::milliseconds(60000));
+#endif
 
     while (!SigUtil::getTerminationFlag() && !SigUtil::getShutdownRequestFlag())
     {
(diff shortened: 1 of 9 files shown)


wsd/COOLWSD.cpp
```diff
@@ -4879,6 +4879,51 @@ class COOLWSDServer
...
+        http::Request request(uriFetch.getPathAndQuery());
+
+        if (!sessionFetch || !socketPoll)
+            return;
+
+        request.add("Accept", "application/json");
+        http::Session::FinishedCallback fetchCallback =
+            [sessionFetch, socketPoll](const std::shared_ptr<http::Session>& httpSession)
+        {
+            const std::shared_ptr<const http::Response> httpResponse = httpSession->response();
```
mmeeks, line 4898: This seems to assume that the POCO config is thread-safe - lets have some locking around this - and ideally not use the poco config to store the information.

wsd/COOLWSD.cpp
```diff
@@ -5094,6 +5139,7 @@ int COOLWSD::innerMain()
             = std::chrono::duration_cast<std::chrono::milliseconds>(timeNow - stampFetch);
         if (fetchUpdateCheck > std::chrono::milliseconds::zero() && fetchUpdateCheck >= durationFetch)
         {
+            processFetchUpdate();
```
mmeeks, line 5142: Do we do this on-start ? if fetchUpateCheck is non-zero?

wsd/COOLWSD.cpp
```diff
@@ -5056,6 +5059,7 @@ int COOLWSD::innerMain()
 #endif
 
     const auto startStamp = std::chrono::steady_clock::now();
+    auto stampFetch = startStamp;
```
mmeeks, line 5107: Lets subtract the durationFetch from that and add say 1 minute - so we fetch early after startup the first time.

mmeeks (changes requested):
Looks like we're going in a good direction - some work on thread safety needed first though.

wsd/COOLWSD.cpp
hcvcastro: Yes, it is thread safe

wsd/COOLWSD.cpp
hcvcastro: validation in case user disable

wsd/COOLWSD.cpp
hcvcastro: done

mmeeks (approved):
Changes are fine - thanks Henry.

Having said that we don't want to break mobile - so we need some #if MOBILE or whatever conditionals around the pieces that cause link errors I think.

Editing the project files there is tough, and of course we really don't need to call-out on mobile to check for updates with the app-store in-place.

Thanks !

PR #4750, opened 2022-05-06
jquery-ui-lightness/ui-icons_ for dark mode

Andreas-Kainz:
switch to ffffff instead of 222222 when the prefers-color-scheme is dark

image

Change-Id: I3e3c68879c6b6a831d5fbf095597ddae6eb51f77

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -485,11 +485,15 @@ input[type='number']:hover::-webkit-outer-spin-button {
 .ui-listbox::-ms-expand {
 	display: none;
 }
-
+@media (prefers-color-scheme: dark) {
+	.ui-listbox-arrow {
+		background: url('images/jquery-ui-lightness/ui-icons_ffffff_256x240.png') no-repeat transparent !important;
+	}
+}
 .ui-listbox-arrow {
 	content: '';
 	background: url('images/jquery-ui-lightness/ui-icons_222222_256x240.png') no-repeat transparent;
-	background-position: -67px -18px;
+	background-position: -67px -18px !important;
 	display: inline-block;
 	position: absolute;
 	width: 11px;


Rash419:
@Andreas-Kainz rebased the PR now it should pass the tests

Andreas-Kainz:
@Rash419 thanks can you review the PR?

Rash419:
@Andreas-Kainz how to enable dark mode ?

Andreas-Kainz:
Dark mode was enabled when the user use dark mode at the browser.

For the moment there is no setting it's automatic if the user use dark mode in general.

Andreas-Kainz:
At Firefox with dev tools. 
dark

Rash419 (commented):
Thanks : )

Rash419 (approved):
Thanks : )

pedropintosilva:
Thank @Andreas-Kainz and @Rash419 for reviewing but I though the conclusion of our discussion in one of the community calls was to avoid the media query altogether and instead do the change vis JS (since anyhow we will need to do JS changes for the generated icons). I think the conclusion was:
1. Tabbed view and sidebar (both views) is currently using generated img elements 
   * To do: make part of those strings (paths) change depending on the dark/light mode for every SVG
2. Compact view is still using some static hard-coded SVG names in the CSS files
   * To do: make those icons also automatically generated from JS by re-using already our code from JSBuilder
   * To do: make sure part of the path also changes depending on the dark/light mode
3. So we should probably hunt for all the in the CSS files, remove them and use only one condition in the JS side like @kendy  suggested:
```js
if (window.matchMedia) {
        if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
            colorScheme = 'Dark';
        } else if (window.matchMedia('(prefers-color-scheme: light)').matches) {
            colorScheme = 'Light';
        }
    }
```

Issue #4736, opened 2022-05-05
branding icon folder


Andreas-Kainz:
Cool offers the simple option to make `*branding.css` files which replace the origin.css file.

I would suggest the same for icons/images.

If there is in the browse/images folder a branding folder cool wan't take the image from the main folder, the image would choosen from the branding subfolder.

![2022-05-05 08_09_46-images - Dolphin](https://user-images.githubusercontent.com/8517736/166870814-b2e06a33-5ed3-4526-a66e-84078d205b7f.png)

Andreas-Kainz:
This is the `branding.css code` from nextcloud where the `lc_formatpaintbrush.svg` icon was replaced by a brading icon. And this code have to be done for EVERY icon. If we would have within the js part of cool the option if branding subfolder has the lc_formatpaintbrush.svg icon, use it, than you don't need any addition thing, if you want to use an different icon theme, or a partly different icon theme.

```
.w2ui-icon.formatpaintbrush,
.w2ui-button .w2ui-icon.formatpaintbrush,
#tb_editbar_item_formatpaintbrush .w2ui-icon,
#tb_actionbar_item_formatpaintbrush .w2ui-icon {
  background-image: url("images/lc_formatpaintbrush_branding.svg") !important;
  background-size: 24px !important;
}

#NotebookBar .unoformatpaintbrush > img,
#NotebookBar .unoFormatPaintbrush > img,
#NotebookBar #FormatPaintbrush #FormatPaintbrushimg {
  -moz-box-sizing: border-box;
  box-sizing: border-box;
  width: 24px !important;
  height: 24px !important;
  padding-left: 24px;
  /* Equal to width of new image */
  background: url("images/lc_formatpaintbrush_branding.svg") no-repeat center !important;
  background-size: 24px !important;
}

.unoFormatPaintbrush:not(.ui-expander-icon-right) > img.unobutton,
.unospan-FormatPaintbrush > img.unobutton,
#FormatPaintbrush #FormatPaintbrushimg,
#FormatPaintbrush img.menu-entry-icon {
  -moz-box-sizing: border-box;
  box-sizing: border-box;
  width: 24px !important;
  height: 24px !important;
  padding-left: 24px;
  /* Equal to width of new image */
  background: url("images/lc_formatpaintbrush_branding.svg") no-repeat center;
  background-size: 24px !important;
}
```

Andreas-Kainz:
I would like to use this feature for the option to have dark mode icons.

[...]

PR #4751, opened 2022-05-06
colibre-dark icons

Andreas-Kainz:
as subfolder at images/dark

Change-Id: I720cb82715760661faa74b7a9acb96ad97808ab5

* Resolves: #4736
* Target version: master


Andreas-Kainz:
This are only the dark icons. 

The coding part when user use dark theme, the icons from the dark subfolder should be used, isn't part of this PR. Therefore a separate PR has to be done. But with this PR the "design" part was done.

Rash419:
@Andreas-Kainz  lets wait to @pedropintosilva 's review : )

pedropintosilva (approved):
Thanks @Andreas-Kainz !

pedropintosilva:
@Andreas-Kainz just note that might be useful, in some places we already have a icon alias so to avoid to have duplicated SVG files just because they have different ids/commands [see Control.JSDialogBuilder.js#L2324](https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2324)

PR #4765, opened 2022-05-12
Expose object name dialog

NickWingate:
Requires core patch: https://gerrit.libreoffice.org/c/core/+/167296
(to enable JSDialog)

Format -> Name (allows to rename object to have unique name in the navigator)


mmeeks:
@eszkadev  can we review this and/or find someone to take over the draft and merge it ? =)

eszkadev (changes requested):
It exposes tunneled dialogs, have to be enabled in the core first.

"Name" dialogs allow to name objects so we can identify them in the navigator - very useful.
"Description" - setups invisible description property? I don't see why it is needed.

eszkadev:
if 2 icons are the same but have different name in the code - please use alias to not use additional space. It can be done in Control.JSDialogBuilder.js - find long list of strings mapped to other strings.

eszkadev (approved):

PR #4778, opened 2022-05-17
make cypress tests work also with 16384 columns

llunak:
Some places tested for AMJ being the last column, but with 16k
columns it's XFD. Without this those tests run in an endless
loop when run with Calc set to 16k columns.

Signed-off-by: Luboš Luňák <l.lunak@centrum.cz>
Change-Id: I6249e0ec975634b60aaf3291e021b74b4aa3cc5d


Rash419 (approved):
:+1:

PR #4780, opened 2022-05-17
fix: typing after idle

eszkadev:
Fixes exception introduced in:
commit 04187501c16b7122d693cc133d87f040284d63a5
Unify ui switcher naming


lpranam (approved):
This does not fix the issue with typing after idle for me but anyways a good fix to get in. maybe also rename the commit to avoid confusion in future :)

PR #4782, opened 2022-05-17
Compact View isn't part of Sidebar, Statusbar

Andreas-Kainz:
Compact View is separate in the menubar from Sidebar and Statusbar
So it shouldn't be between them in the View Tab.

image

Change-Id: Ib46e9c4a4df24432eae5a7fef335e7cde68b1766

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -944,11 +944,6 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'vertical': 'true'
 			},
 			{
-				'type': 'bigtoolitem',
-				'text': _UNO('.uno:Sidebar'),
-				'command': '.uno:Sidebar'
-			},
-			{
 				'id': 'toggleuimode',
 				'type': 'bigmenubartoolitem',
 				'text': _('Compact view'),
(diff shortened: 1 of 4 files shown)


Andreas-Kainz:
If you think it's only a Change cause of a change, I can close the PR.

From my point of view Compact view shouldn't be between Sidebar and Statusbar. It should be the first item or the last one in this "group".

pedropintosilva (changes requested):
Thanks @Andreas-Kainz for looking at this. I think we can improve here. Even thought the top menus on the compact view don't need to be 1 to 1 with tabbed view tabs (sometimes they really shouldn't) in this case I would say you are right. Nevertheless I would prefer to have the view button to the the first and the sidebar the last because:
1. Then the Sidebar icon is the rightmost icon which is nice since the sidebar comes from that sode
2. The view button would be in the leftmost side from those icons. That's ideal since it's an action that is not only hierarchy above the other actions in that group of buttons but also probably more important to the user to spot straight away 

TLDR: 
image

Andreas-Kainz:
Thank's for the feedback I also didn't know what first, so I make one variant and get now your feedback. 

Will update the PR.

pedropintosilva:
no sweat : ) and thanks for improving this up ;)

Andreas-Kainz:
I will make the showstatusbar from menubartoolitem to a bigmenubartoolitem I think you already did it @pedropintosilva do you have a PR example?

pedropintosilva (changes requested):
Something odd in impress:
image

Andreas-Kainz:
> Something odd in impress: image

you mean cause Status Bar isn't a bigtoolitem? That's the same within master. Open an issue. I don't want to fix it with this PR cause I don't want to mix different stuff.

pedropintosilva:
> you mean cause Status Bar isn't a bigtoolitem?

No, because of this you have added other icons to the view tab, in Impress

PR #4784, opened 2022-05-18
Show visual indicator for print ranges in Calc (master)

dennisfrancis:
* Target version: master 

### Summary

* Introduce new callback LOK_CALLBACK_PRINT_RANGES
* Read the print ranges for the current sheet from LOK_CALLBACK_PRINT_RANGES message and draw dashed grid lines for the rows/columns bordering the range.

Caveats:

* Desktop calc does not draw print ranges on document load if there are
print ranges set previously on save. Cool also has the same behaviour.

* If there are multiple print ranges in the same sheet desktop calc does
not show any dashed lines indicating the ranges. Cool also adopts this
behaviour for now.

* Styling of the dashed lines was not investigated much for writing this
patch hence it uses the same drawing style as normal grids.

### Checklist


gokaysatir (approved):

PR #4785, opened 2022-05-18
cypress: added new function selectCellsInRange

Rash419:
after the jumbo sheet patch selecting entire row and merging all the cells becomes costly and cypress timeouts
therefore select first 100 cells and test the merging functionality

Change-Id: I8568bbdbabdccc55c2938f886619a1d178ae7d22

* Target version: master


llunak (approved):
Yes, this makes these tests now pass with 16k columns, thank you.

PR #4791, opened 2022-05-19
use the more efficient getSelectionTypeAndText()

llunak:
Otherwise calling getSelectionType() first and then getTextSelection()
creates the XTransferable2 twice, and then possibly converts it
to the text format twice, which may be expensive.

Also some cypress tests use selectEntireSheet() and then check
the clipboard contents, which with the switch to 16k columns in Calc
may be slow enough in debug builds when done twice to time out
the test, and this helps a bit there.

Signed-off-by: Luboš Luňák <l.lunak@centrum.cz>
Change-Id: If166f67c216281d32dbb1d3e10b51177b42a9668

This requires https://gerrit.libreoffice.org/c/core/+/134615 .

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1137,11 +1137,12 @@ bool ChildSession::getTextSelection(const StringVector& tokens)
         return true;
     }
 
-    std::string selection;
     getLOKitDocument()->setView(_viewId);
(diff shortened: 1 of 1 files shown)


llunak:
Ping?

Ashod (approved):
Thanks @llunak, looks good. I suppose Core has to go in first for the CI to pass.

PR #4793, opened 2022-05-19
Proxy: send explicit charset=utf-8

mmeeks:
https: //github.com/nextcloud/richdocuments/issues/2196
Change-Id: I4fce74290e8bf8a6681922a2165772933443703e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (approved):
I think we can merge this, because it won't make things worse. Testing of richdocumentscode that has this patch will show if the fix was good. It would be difficult and time consuming to test the patch in a development environment.

PR #4794, opened 2022-05-19
fix: typing after idle

lpranam:
Reset document info on idle

resetting document info will force to use new status of the document

problem:
reconnecting after idle, user could not type anything in document,
this was due to some properties were to set assuming they existed from last session (i.e: clientzoom)
resetting document info will force to use new status message and set all the properties again correctly

Change-Id: Ibcf395eee3a4e0b99413de0178331d42a1354253

* Target version: distro/collabora/co-21-11 

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -306,7 +306,7 @@ L.Control.UIManager = L.Control.extend({
 
 		document.body.setAttribute('data-userInterfaceMode', uiMode.mode);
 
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Looks great to me - thanks Pranam ! =) hopefully it fixes a number of other sillies after resume from idl.e

eszkadev (approved):
works as expected

eszkadev:
it has to be rebased so it will know about missing item I think

lpranam:
closed in favour of #4797

PR #4797, opened 2022-05-20
fix: typing after idle

lpranam:
Reset document info on idle

resetting document info will force to use new status of the document

problem:
reconnecting after idle, user could not type anything in document,
this was due to some properties were to set assuming they existed from last session (i.e: clientzoom)
resetting document info will force to use new status message and set all the properties again correctly

* Target version: distrp/collabora/co-21-11 

### Checklist


eszkadev (approved):

lpranam:
no idea why FreeBSD is not using the latest core. even tried opening new PR :(

Issue #4781, opened 2022-05-17
btns.css use wrong var name


Andreas-Kainz:
Check the var names at btns.css

https://github.com/CollaboraOnline/online/blob/a4885396bb595e358d580091b7eb474cf3c69db0/browser/css/btns.css#L8

PR #4799, opened 2022-05-21
btns var color name fixes

Andreas-Kainz:
btns use var names that didn't exist

* Resolves: #4781
* Target version: master 

Change-Id: I13f2acbdd91cf05391da81adc721156d125f3de7

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -5,8 +5,8 @@
 }
 
 button.jsdialog {
-	color: var(--color-text);
-	background-color: var(--color-background);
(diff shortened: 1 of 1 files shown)


pedropintosilva (commented):
I have tested and yes these are not defined anywhere but we do have still other files using that background variable namely: https://github.com/CollaboraOnline/online/blob/master/browser/css/mobilewizard.css#L678 and in https://github.com/CollaboraOnline/online/blob/master/browser/css/override-vex.css#L351 so I wonder when did we removed the definition of that var

PR #4800, opened 2022-05-21
wsd: browser: handle 413 Entity Too Large from storage

Ashod:
It seems at least some storage hosts report disk-full
errors via 413. This logic of handling 413 error as
disk-full was introduced in
https://github.com/CollaboraOnline/online/commit/f8e0b8c11e94810fdde73f8b4d27b80a989670cb.

Here we handle 413 as it is defined, as Entity Too Large.

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -825,6 +825,9 @@ app.definitions.Socket = L.Class.extend({
 			if (command.errorKind === 'savediskfull') {
 				storageError = errorMessages.storage.savediskfull;
 			}
+			else if (command.errorKind === 'savetoolarge') {
+				storageError = errorMessages.storage.savetoolarge;
+			}
 			else if (command.errorKind === 'savefailed') {
 				storageError = errorMessages.storage.savefailed;
 			}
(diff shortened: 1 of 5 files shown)


thebearon:
Thanks for improving the message, Ash! What I think could be further indicated is that this is often a message from the reverse proxy, and not from the WOPI host itself.

Ashod:
> What I think could be further indicated is that this is often a message from the reverse proxy, and not from the WOPI host itself.

Thanks for the feedback Aron. Considering that the end-user would have no idea what a proxy is (let alone a reverse proxy), and the fact that they couldn't do anything about it anyway, I'm not so sure that speculating to the end-user that the issue might be the reverse proxy (which we don't even know for sure) is productive. Telling them to contact the admins is reasonable.

We can document somewhere else for the admins (including we can add a warning log with the reverse proxy hint that the admins could check (but then we wouldn't want to flood the logs for something they might have configured on purpose. After all, an admin might want to set reasonable limits to maintain a good service.)

Ashod:
@thebearon: added descriptive message in the warning log that could benefit the admins.

thebearon:
Thanks, Ash!

lpranam (approved):

PR #4803, opened 2022-05-23
browser: contentcontrol: add dropdown type

Rash419:
Change-Id: I178561b081f9c1cb3c7db081fcdab606c6426cc5

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -132,6 +132,7 @@ COOL_CSS_LST =\
 	$(srcdir)/css/impress-mobile.css \
 	$(srcdir)/css/spreadsheet.css \
 	$(srcdir)/css/writer-mobile.css \
+	$(srcdir)/css/writer.css \
 	$(srcdir)/css/editor.css \
 	$(srcdir)/css/iframedialog.css \
 	$(builddir)/node_modules/jquery-contextmenu/dist/jquery.contextMenu.css \
@@ -223,7 +224,6 @@ COOL_JS_LST =\
 	src/layer/tile/CanvasSectionProps.js \
 	src/layer/tile/CanvasSectionContainer.ts \
 	src/layer/tile/CommentSection.ts \
-	src/layer/tile/ContentControlSection.ts \
 	src/layer/tile/CommentListSection.ts \
 	src/layer/tile/TilesSection.ts \
 	src/layer/tile/AutoFillMarkerSection.ts \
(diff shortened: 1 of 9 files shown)


vmiklos (approved):
Thanks, works for me as a user. I can insert a dropdown, I can add list items, I can select a list item. I can't comment on the JS style or design, leaving that to @gokaysatir or @pedropintosilva .

pedropintosilva:
_testing it_

pedropintosilva (approved):
@Rash419 I see that now we:
- Moved any insert action related to the forms from `Insert` tab to `Form` tab
- We have added `insertdropdowncontrol` and `contentcontrolproperties`

I have nothing against with these ^. Even though in the core side the tab label is different (those actions live within `Tools` tab). Here, in the context of Collabora Online, and since we do not have so many options I think it makes sense to have `Form` tab

pedropintosilva (changes requested):
But within the document this doesn't seem right:
image

pedropintosilva:
but I think we should merge this as it is and then in a separated PR tackle these styles

pedropintosilva (approved):

Rash419:
> @Rash419 I see that now we:
> 
> * Moved any insert action related to the forms from `Insert` tab to `Form` tab
> * We have added `insertdropdowncontrol` and `contentcontrolproperties`
> 
> I have nothing against with these ^. Even though in the core side the tab label is different (those actions live within `Tools` tab). Here, in the context of Collabora Online, and since we do not have so many options I think it makes sense to have `Form` tab. Also moving insert actions (that are not used so often) away from the insert tab and into their own specific tab seems also reasonable

@pedropintosilva In core form controls are there in both tools and form tab

pedropintosilva:
missing icons added in https://github.com/CollaboraOnline/online/pull/4832

browser/src/layer/ContentControlLayer.js
```diff
@@ -0,0 +1,228 @@
...
+					'enabled': true,
+					'children': [
+						{
+							'id': 'dropdownlist',
+							'type': 'treelistbox',
+							'text': '',
+							'enabled': true,
+							'children': [
+								{
+									'id': '',
```
eszkadev, line 143: I think we can remove these scrollbar/scrollbarbox items

Rash419: thanks  : )

PR #4804, opened 2022-05-23
Fix Insertcontentcontrols strings on tabbed view

pedropintosilva:
Don't use strings coming from uno commands as they include
"Insert":
1. There is no need to say that these actions will "Insert", they
are already in the Insert tab
2. All other actions do not have Insert so it would be not consistent
3. On top of that with Insert in each icon the labels look quite long

Signed-off-by: Pedro Pinto da Silva <pedro.silva@collabora.com>
Change-Id: I286e95187d5df08cdf2d71bd99f3b77a17ef12ad


Rash419 (approved):
Thanks : )

PR #4814, opened 2022-05-24
jsdialog: handle tab key in autofilter popup fixes #4716

### LINKED ISSUE #4716: AutoInput list in spreadsheet causes inoperability
1. Open a spreadsheet file.
2. Press the Alt and Down arrow keys to open the drop-down list menu.
3. Calc operation becomes unavailable.
4. The user has no choice but to press the reload button on the browser.

However, users will not be completely inoperable. With the drop-down list menu displayed.
1. Mouse hover state within the drop-down list menu is identified but not clickable.
2. It is only possible to enter text in the designated cells.
2.1. I haven't found the law yet. Random arrow keys or other keyboard operations will allow you to move cell selections.
2.2. Once the user has made the Delete Details panel visible, they can no longer hide it. Mouse clicks are still not recognized.

Is there a problem with the mouse judgment when the drop-down list menu is displayed?
When the cell is ready to be moved, it seems to accept keyboard operations such as copy and paste just fine.

---
COOLWSD version: 21.11.4.1 
LOKit version:
 Collabora Office 21.06.27.1
Served by: CentOS Linux 7 (Core)

eszkadev:
This added begin and end mark to the popup so we can detect
we are leaving the popup and we should close.

This prevents us from blocking the whole UI by focusing
map and typing before popup was closed.

Also close popup when server confirmed it should be closed.

Fixes #4716

### DIFF
--- a/browser/src/control/Control.AutofilterDropdown.js
+++ b/browser/src/control/Control.AutofilterDropdown.js
@@ -78,6 +78,8 @@ L.Control.AutofilterDropdown = L.Control.extend({
 
 			this.builder.setWindowId(null);
 			this.subMenuBuilder.setWindowId(null);
+
+			this.map.focus();
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


merttumer (approved):
Works as expected

PR #4818, opened 2022-05-25
Try to make the trace event emission more thread-safe

tml1024:
Change-Id: I26ef3ca58c1fa6c237a49a42bb92ecf1a1227db3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -114,6 +114,9 @@ class Document;
 static Document *singletonDocument = nullptr;
 #endif
 
+static const std::thread::id nullThreadId;
+static std::thread::id mainKitSocketPollThread;
+
 _LibreOfficeKit* loKitPtr = nullptr;
 
 /// Used for test code to accelerating waiting until idle and to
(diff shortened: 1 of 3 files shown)


kit/Kit.cpp
```diff
@@ -1866,12 +1869,15 @@ class Document final : public DocumentManagerInterface
...
 static constexpr int traceEventRecordingsCapacity = 100;
 static std::vector<std::string> traceEventRecordings;
+static std::mutex traceEventRecordingsMutex;
 
 static void flushTraceEventRecordings()
 {
     if (traceEventRecordings.size() == 0)
         return;
 
+    assert(mainKitSocketPollThread == std::this_thread::get_id());
```
mmeeks, line 1879: Surely this also needs to take the mutex ?

kit/Kit.cpp
```diff
@@ -1907,7 +1913,8 @@ void TraceEvent::emitOneRecordingIfEnabled(const std::string &recording)
     if (singletonDocument == nullptr)
         return;
 
-    singletonDocument->sendTextFrame("forcedtraceevent: \n" + recording);
```
mmeeks, line 1910: Another thing we can do is use the thread-safe "callback in a poll thread" thing - so:

whateverPoll->addCallback( { singletonDocument->sendTextFrame(); });

would presumably work fine (?) that should be sent next time that poll spins in the right thread.

mmeeks (changes requested):
Thanks for that - would be good to get this nailed too.

kit/Kit.cpp
tml1024: Yes.

PR #4820, opened 2022-05-26
Private/pedro/jsformulabar input height n width

pedropintosilva:
commit bf9754eb3d5b7bedcc3a5a15cb0ebc14d94b7186

Formulabar: fix sc_input_window height (when collapsed)
    
Formula input window (textarea) had different height from address 
input
  - Set it at the same height as seen in toolbar.css:211

----

commit df1dafe3f69d5030f30ab53feeb19126f05f08d6

Fix formulabar width, use flex property
    
Avoid using percentages when we can already use the grow property
(it avoids different results across browsers)

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1169,7 +1169,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 /* formulabar */
 #sc_input_window.formulabar {
 	border: 1px solid var(--color-border);
-	width: 95% !important;
+	flex-grow: 1;
 	resize: none !important;
 	max-width: unset !important;
 	font-size: var(--default-font-size) !important;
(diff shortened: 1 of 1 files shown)


pedropintosilva:
**not related to this PR** but probably Next would be good to, note to myself: fix jumping icons and margins, probably better to tweak :
jsdialogs.css:1183
```js
.formulabar.unotoolbutton {
	margin-inline-end: 4px !important;
}

.formulabar.unotoolbutton:not(.has-dropdown), .formulabar.unotoolbutton img, .formulabar.ui-pushbutton {
	height: 24px !important;
	width: 24px !important;
	min-width: unset !important;
}
```

probably by removing `.formulabar.unotoolbutton:not(.has-dropdown)` and setting elements with `.formulabar.unotoolbutton` with two `margin-block-start`. One for one the formulabar is collapsed and another one for when the formulabar is expanded
- We probably need extra class in the parent that signifies that

eszkadev (approved):
looks good

Issue #4821, opened 2022-05-26
Remove unused CSS related to old formulabar


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

### Detailed description and rationale
Formulabar implementation has suffered great improvements thanks to @eszkadev . It is not rendered anymore via html canvas. It is now rendered as normal html elements.
![image](https://user-images.githubusercontent.com/65948705/170461549-659f98f1-c491-4e83-aa93-7d108a93318c.png)

Hunt down CSS bits that are targeting elements that do not exist anymore, e.g.: `#formulaInput`

### Code pointers
`toolbar.css` and maybe also in other CSS files

#### How to test it
* without development environment: https://github.com/CollaboraOnline/online#gitpod
* locally with development environment: https://collaboraonline.github.io/post/build-code/

DekotaNelson:
I can take care of this issue and will get started on it.

pedropintosilva:
Thanks @DekotaNelson ! BTW it will be cool if you could attend our weekly dev meeting :) It's every Thursday at 12:00 Central Europe : https://collaboraonline.github.io/post/communicate/

PR #4826, opened 2022-05-27
Removed unused CSS related to old formulabar.

DekotaNelson:
Resolves: Issue #4821
Target version: master

Summary
Removed CSS bits that are targeting elements that do not exist anymore and were left over from the formulabar in the browser calc editor being converted from html canvas to normal. Expected there to be more unused bits, but everything else was in use.

TODO:
N/A

Checklist
[x] Code is properly formatted
[x] All commits have Change-Id
[x] I have run tests with make check
[x] I have issued make run and manually verified that everything looks okay
[x] Documentation (manuals or wiki) has been updated or is not required
Signed-off-by: Dekota Nelson [nelson.dekota@gmail.com](mailto:nelson.dekota@gmail.com)

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -212,24 +212,6 @@ w2ui-toolbar {
 	width: 100px;
 }
 
-#formulaInput {
-	height: 29px;
-	width: 100%;
-}
-
-#calc-inputbar-wrapper {
-	display: block;
-	border: none;
-}
-
-#calc-inputbar {
-	display: block;
-	width: 100%;
-	position: relative;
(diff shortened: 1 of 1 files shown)


pedropintosilva (commented):
Thanks @DekotaNelson for hunting this easy hack down :)

I have noticed the deletion of `#calc-inputbar` and `#calc-inputbar-wrapper`. Those elements still exist so, I guess you test those outs and ended up discovering that those rules are not needed now?

pedropintosilva (approved):
I tested these out and everything seems fine :+1:

PR #4839, opened 2022-06-01
jsdialog: use correct codes for shortcuts CTRL + char

eszkadev:
Handles for example: CTRL + "i" in the formulabar what triggers italic font

we should do the same for other modifiers: shift, alt...


gokaysatir (approved):

PR #4846, opened 2022-06-02
wsd: handle socket closed on read better

Ashod:
A corner case where read doesn't return 0
on a closed socket can result in a stuck
state where we attempt to read when we
do get ECONNRESET (which we didn't check
on reads).

This makes the interface of readIncomingData
the same as writeOutgoingData by returning
the actual return value of the read syscall.

So now we handle both return 0 as well as
error codes returned on failed calls.

The logic hasn't changed, just that now
we handle errors better and similar to the
write case.

Change-Id: I0b38a63da4e6c92a482948478d5d8d446e0b8b58

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1094,8 +1094,8 @@ public:
     }
 
     /// Reads data by invoking readData() and buffering.
-    /// Return false iff the socket is closed.
-    virtual bool readIncomingData()
+    /// Returns the last return from writeData. 0 implies socket is closed.
+    virtual int readIncomingData()
     {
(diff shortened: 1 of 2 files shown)


mmeeks (commented):
if (ignoringInput())
        {
            LOG_WRN("Ignoring attempted read from " << getFD());
            return false; // error - close it.
        }

Needs thought in the read method surely ?

mmeeks (approved):
Apart from the lingering boolean there this looks sensible enough - thanks Ash =) should prolly backport it I guess (?)

PR #4849, opened 2022-06-02
ignore if comment is selected while pasting

lpranam:
if comment is selected but is not in editing,
proceed to normal pasting in document

problem:
while pasting content multiple times with comment,
previously pasted comment stays selected and can't further normally

Change-Id: Ia594c725f4fe7108d6aea1591d88b2c306202be3

* Target version: master 

### Checklist


gokaysatir (approved):
Thanks for the improvement :)

PR #4850, opened 2022-06-02
Impress add context: MasterPage

### LINKED ISSUE #4808: uno:SlideMasterPage not available at tabbed toolbar
`uno:SlideMasterPage` is available in the menubar (classic view) but the command isn't available in the notebookbar layout.

Andreas-Kainz:
* Resolves: #4808 
* Target version: master 

Fix #4808 by add SlideMasterPage to impress view tab
and add a context related MasterPage tab.

Update the icons to use the master*.svg colibre icons from core

image

Change-Id: Ibbd24d1c20a48e26591d217311c5989794aa841e


Andreas-Kainz:
would be cool to have `.uno:InsertMasterPage` in the master page tab, but the command isn't available. Any idea @eszkadev?

Andreas-Kainz:
`.uno:InsertMasterPage` is now available (including an additional Master Tab) at tabbed view.

pedropintosilva (approved):
Tested

pedropintosilva (changes requested):
* Insert master slides does not work collaboratively: Once one user changes to that view every user gets their view changed as well 
* The `View` -> `Master slide` should be clickable also when the Master view is active so the user can exit by pressing that (following the same logic present in the `status bar` icon)
  * Currently the user opens the Master slide view , the only way to close the view is by going to `Master` tab -> `Close master view`

pedropintosilva:
> * Insert master slides does not work collaboratively: Once one user changes to that view every user gets their view changed as well but without the context tab

This bug already existed before this PR (just tested and confirmed that). It should not be a blocker to merge this
> 
>     * The `View` -> `Master slide` should be clickable also when the Master view is active so the user can exit by pressing that (following the same logic present in the `status bar` icon)
>       
>       * Currently the user opens the Master slide view , the only way to close the view is by going to `Master` tab -> `Close master view`
^ this is still an issue, but let's merge this and do a follow up PR to improve that toggle

pedropintosilva (approved):

pedropintosilva (approved):

Issue #4672, opened 2022-04-25
Basic Shapes should be Shapes only


Andreas-Kainz:
As you can see in the screenshot, the insert basic shapes didn't contain only basic shapes. The insert command contain ALL shapes which are available within cool / LibO.

So I suggest to rename the string to Shapes (in addition is't shorter).

![grafik](https://user-images.githubusercontent.com/8517736/165109496-2276c41e-b34a-4d81-bda4-9d581fbee4a3.png)

pedropintosilva:
hm I think this was fixed already I think I did 2 times, but probably the unocommands.js got re-updated and reverted back to the original string

pedropintosilva:
maybe better to fix it in another way

PR #4851, opened 2022-06-02
Insert Shapes instead of Basic Shapes

Andreas-Kainz:
* Resolves: #4672
* Target version: master 

In the insert tab there is the command uno:BasicShapes
by select BasicShapes the user can insert shapes
at least Basic Shapes, Symbol Shapes, Block Arrows,
Stars and Banners, Callouts, Flowchart, ...

So uno:BasicShapes will let you insert all available type of shapes
therefore the text label should be Shapes instead of Basic Shapes.

Change-Id: I692b7a5b03b905599648c7af4a016684b12b335a


Andreas-Kainz:
If you have time @pedropintosilva please review. Thanks.

PR #4856, opened 2022-06-03
cypress: re-enable impress/table_operation_spec.js

Rash419:
modify the test to use 2x2 table when testing 'insert column before/after' tests
no idea why we don't get svg when there is 3x2 table

Change-Id: Id927fede691c4483206fa2b65c450f11cc980581
* Target version: master


gokaysatir (approved):
Thanks for the improvements:)

PR #4864, opened 2022-06-06
jsdialog: avoid focus in special popups like busypopup

eszkadev:
so user can continue typing....


merttumer (approved):
works as expected

PR #4872, opened 2022-06-08
jsdialog: focus map after popup is closed

eszkadev:
So we can type after we closed popup in the sidebar.


merttumer (approved):
works as expected

PR #4879, opened 2022-06-09
wsd: alias: fix a case when regex with '*' is added in host tag of al…

Rash419:
…ias_group

if added regex in host has '*' in it and group has no aliases, when we receive first request from host pattern we considered it as original host and all the host following first request which matches the pattern are considered as aliases
for example:
<group>
	<host>https://.*:80</host>
</group>
if we receives first request from asustuf then behaviour will similar to the following config:
<group>
	<host>https://asustuf:80</host>
	<alias>https://.*:80</alias>
</group>

Change-Id: I70fb91a4bb7bf38ed79db9efd9fe4bc46db325e1

### DIFF
--- a/wsd/HostUtil.cpp
+++ b/wsd/HostUtil.cpp
@@ -11,6 +11,7 @@
 
 Util::RegexListMatcher HostUtil::WopiHosts;
 std::map<std::string, std::string> HostUtil::AliasHosts;
+std::set<std::string> HostUtil::hostList;
 std::string HostUtil::FirstHost;
 bool HostUtil::WopiEnabled;
 
@@ -94,6 +95,7 @@ void HostUtil::parseAliases(Poco::Util::LayeredConfiguration& conf)
         try
         {
             const Poco::URI realUri(uri);
+            HostUtil::hostList.insert(realUri.getHost());
             HostUtil::addWopiHost(realUri.getHost(), allow);
         }
         catch (const Poco::Exception& exc)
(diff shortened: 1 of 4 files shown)


wsd/HostUtil.cpp
```diff
@@ -148,6 +150,14 @@ std::string HostUtil::getNewUri(const Poco::URI& uri)
     {
         newUri.setAuthority(value);
     }
+    else
+    {
+        const std::string val = Util::getValue(hostList, newUri.getHost());
```
kendy, line 155: Would be really good to add a comment here of the form - "It is allowed for the host to be a regex. In that case, the first who connects is treated as the 'host', and stored to the AliasHosts here"

wsd/HostUtil.cpp
```diff
@@ -148,6 +150,14 @@ std::string HostUtil::getNewUri(const Poco::URI& uri)
...
+        if (!val.empty() && val.find('*') != std::string::npos)
```
kendy, line 156: I wonder if we can use something better to check it was a regex match - what about to compare 'val' to the newUri.getHost()? If they are not equal, it was a regex match, and the pair should be added to AliasHosts, I think - can you please check?

kendy (approved):
Looks great in general, so approving; but if you can check the 2 comments I'v added please?

kendy (approved):
Looks great to me, thanks a lot! :-)

PR #4889, opened 2022-06-10
jsdialog: correct popup close in 21.11

eszkadev:
jsdialog: be sure popup overlay will not block user
    
    Don't add multiple overlays with the same id.
    Remove overlay always when it exists and we close.
    Close dropdown entires by dropdown click - specific to 21.11
    where popups in the core are handled differently.

------------------------------------------------------------------------------------------------------------

    jsdialog: use correct element to close popup
    
    For toolitems with dropdowns we should toggle the button.
    Use correct parent so we don't use button from other component.
    Later we send toggle event to properly close the popup on the server.
    
    example: Calc sidebar, Underline popup
    Before this patch when we clicked in outside the popup it dissapeared
    but app was completly blocked.

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -68,6 +68,13 @@ L.Control.JSDialog = L.Control.extend({
 		var builder = this.dialogs[id].builder;
 
 		if (sendCloseEvent) {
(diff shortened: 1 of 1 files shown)


merttumer (approved):
works as expected

PR #4891, opened 2022-06-10
jsdialog: combobox is not a container type

eszkadev:
add combobox to list of non-containers and cleanup


merttumer (approved):
Works as expected

PR #4896, opened 2022-06-13
fix hidden cursor 21.11

eszkadev:
when other view moved cursor outside out view - all of
them were hidden because we used selector for all the
elements matching cursor class - instead of current
instance


merttumer (approved):
works as expected

PR #4897, opened 2022-06-13
Fix blinking dialog & warnings about notebookbar in Impress (21.11)

eszkadev:
notebookbar: retry to activate when needed
    
    Don't generate spam in the console log with warnings about not
    activated notebookbar in apps where we don't use any welded widgets.

Allow to send view mode command when dialog is opened
    
    We automatically send notebookbar activation command in some cases.
    Allow this command when dialog is opened and don't blink.


gokaysatir (approved):

PR #4844, opened 2022-06-02
browser: contentcontrol: add date type

Rash419:
add uno command to insert date in form tab
renamed 'Content Control Properties' -> 'Properties'

Signed-off-by: Rash419 <rashesh.padia@collabora.com>
Change-Id: Ib4a439cd349bb79430d590e0b80c825372c629d3

* Target version: master

PR #4902, opened 2022-06-14
Promote content control properties to bigtoolitem in tabbed view

pedropintosilva:
Content control properties is not stacked on top of any other button
and it will probably never be. Promote it to bigtoolitem so it can use
more vertical space

Change-Id: I39b624b3930611f91f98d413646062a773183eec


pedropintosilva:
added in https://github.com/CollaboraOnline/online/pull/4844/commits/9d3eaf33b266c1e30e29a1241b6be7d6689ebb97

PR #4904, opened 2022-06-14
jsdialog: formulabar: send key modifierwith all key events

eszkadev:
This will allow to create multiline cell content using Shift+Enter


pedropintosilva:
One note though: 
:x:  when pressing <kbd>Shift</kbd>+<kbd>Ctrl</kbd>+<kbd>Enter</kbd> the cursor enters the cell (moves the cursor to inside of the cell) and it should instead move the cursor to the input line
:heavy_check_mark: <kbd>F2</kbd> correctly enters the cell

PR #4909, opened 2022-06-15
mobile: contentcontrol improvements

Rash419:
add: form icon in insert menu
fix: datecontentcontrol icon in form menu, jsdialog: options for all content control types were shown

Change-Id: Id3f425f3a80cb09f58b3f83224bb19ebb365e8f3

* Target version: master 
* Core patch - https://gerrit.libreoffice.org/c/core/+/135902
* Todo:
mobile: fix insertion of new items in dropdown type


pedropintosilva (approved):
Tested it works.
Maybe a follow up would be to move `properties` from insert panel to the properties panel

PR #4924, opened 2022-06-17
css: improve contentcontrol datepicker

Rash419:
Change-Id: Ie03578e47e123b6100fa361005a148a50bffa6d6

* Target version: master 

### Summary
Before:
image
After:
image

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -900,3 +900,41 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 #about-dialog-info {
 	flex-direction: column;
 }
+
+/* ovveride some jquery css for contentcontrol datepicker */
+#datepicker .ui-state-highlight {
+	background: var(--color-background-dark) !important;
+	font-weight: bold !important;
+}
+
+#datepicker .ui-state-default {
+	background: var(--color-main-background);
+	font-weight: normal;
+	color: var(--color-main-text);
+	border: 1px solid var(--color-border-lighter);
+}
+
+#datepicker .ui-state-active {
+	border: 1px solid var(--color-border-dark);
+}
+
+#datepicker .ui-state-hover {
+	border: 1px solid var(--color-border-darker);
+	background: var(--color-background-lighter);
+}
+
+#datepicker .ui-datepicker-next span {
+	background-image: url('images/lc_nextrecord.svg');
+}
+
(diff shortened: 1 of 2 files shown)


Rash419:
@pedropintosilva decided to override the css rules because it gives me more control on how date picker looks

browser/css/cool.css
```diff
@@ -900,3 +900,54 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
...
+#datepicker .ui-state-default {
+	background: #f7f7f7 ;
+	font-weight: normal;
+	color: black;
+	border: 1px solid #fff;
+}
+
+#datepicker .ui-widget-content {
+	background: #fff;
+}
```
pedropintosilva, line 913: This is not needed we can safely change the colour directly in the jquery-ui-lightness.css`:896`, ideally to `var(--color-background-lighter)`

browser/css/cool.css
```diff
@@ -900,3 +900,54 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
...
+#datepicker .ui-state-highlight {
+	background: lightgray;
+	font-weight: bold;
+}
+#datepicker .ui-state-active {
+	border: 1px solid #a9a9a9;
+}
+#datepicker .ui-widget.ui-widget-content {
+	border: 1px solid #cccace;
+}
```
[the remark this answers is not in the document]
pedropintosilva: probably there are more cases like these two I have pointed out

Rash419:
@pedropintosilva I think we should only edit `.ui-datepicker` rules in jquery-ligthness.css . If we edit for example `ui-state-active`
directly from jquery-lightness.css it will affect all the widgets , ui-state-active's child elements have some specific rules for example
`.ui-state-active a` we also need to change that then

browser/css/cool.css
```diff
@@ -900,3 +900,51 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
...
+
+#datepicker .ui-widget-content {
+	background: #fff;
+}
```
pedropintosilva, line 925: This is not needed we can safely change the colour directly in the jquery-ui-lightness.css`:896`, ideally to `var(--color-background-lighter)`

pedropintosilva (changes requested):

PR #4927, opened 2022-06-18
wsd: reuse parsed message

Ashod:
Change-Id: I9d67f92ba99037954400e07a7d1101e5de3716fd


mmeeks (approved):
Looks sensible in the end we should share parsed messages in the write queue I guess =)

PR #4932, opened 2022-06-20
browser: add fzstd lib

hcvcastro:
usage

	var decompressed = fzstd.decompress(compressed);
	// Second argument is optional: custom output buffer
	var outBuf = new Uint8Array(100000);

	// IMPORTANT: fzstd will assume the buffer is sufficiently sized, so it
	// will yield corrupt data if the buffer is too small. It is highly
	// recommended to only specify this if you know the maximum output size.
	window.fzstd.decompress(compressed, outBuf);

Change-Id: I0b378f9395b2442bac9f0790f7633e6a6cb0402e


timar:
This has been merged already.

PR #4933, opened 2022-06-21
Fix heap-use-after-free in HttpRequestTests

Ashod:
- wsd: test: HttpRequestTests timeout cleanup
- wsd: test: fix heap-use-after-free


vmiklos (approved):
Looks good, thanks. Please wait till CI gives a +1, then this could go in.

PR #4935, opened 2022-06-21
jsdialog: formulabar: handle tab key

eszkadev:
Open spreadsheet, click on formulabar, type, press tab -> it should move to the next cell in a row


merttumer (approved):
Works as expected

PR #4938, opened 2022-06-21
notebookbar: unify color control with core

eszkadev:
in desktop case color widget should show the latest selected
color not current one. split the button into arrow with dropdown
and "last color" button (icon)


pedropintosilva (approved):
tested

PR #4940, opened 2022-06-22
proxy request handler

hcvcastro:
- Revert "serve lokit-extra-img.svg with integrator theme, too"
- Revert "wsd: fetch external resource in debug build, but not for unit tests"
- Revert "wsd: add indirect request logo image"
- Revert "wsd: add fetch external method"
- wsd: proxy: add ProxyRequestHandler files
- wsd: proxy: add handler request function
- wsd: proxy: add a callback response handler
- wsd: proxy: add cache response
- wsd: proxy: handle cache max age
- wsd: proxy: modify basic headers to forward
- wsd: proxy: handle the proxy request

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -970,6 +970,7 @@ public:
     /// Get the timeout, in microseconds.
     std::chrono::microseconds getTimeout() const { return _timeout; }
 
+    std::shared_ptr<Response> response() { return _response; }
     std::shared_ptr<const Response> response() const { return _response; }
     const std::string& getUrl() const { return _request.getUrl(); }
 
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -971,6 +971,7 @@ public:
     std::chrono::microseconds getTimeout() const { return _timeout; }
 
     std::shared_ptr<const Response> response() const { return _response; }
+    const std::string& getUrl() const { return _request.getUrl(); }
 
     /// The onFinished callback handler signature.
     using FinishedCallback = std::function<void(const std::shared_ptr<Session>& session)>;
--- a/wsd/ProxyRequestHandler.cpp
+++ b/wsd/ProxyRequestHandler.cpp
@@ -14,4 +14,22 @@
 #include <net/HttpRequest.hpp>
 #include <net/HttpHelper.hpp>
 
+void ProxyRequestHandler::handleRequest(const std::string& relPath,
+                                        const std::shared_ptr<StreamSocket>& socket)
+{
+    Poco::URI uriProxy(ProxyServer);
+
+    uriProxy.setPath(relPath);
+    auto sessionProxy = http::Session::create(uriProxy.getHost(),
+                                              http::Session::Protocol::HttpSsl,
+                                              uriProxy.getPort());
+    sessionProxy->setTimeout(std::chrono::seconds(10));
+    http::Request requestProxy(uriProxy.getPathAndQuery());
+
+    if (!sessionProxy->asyncRequest(requestProxy, *COOLWSD::getWebServerPoll()))
+    {
+        HttpHelper::sendErrorAndShutdown(400, socket);
+    }
+}
+
 /* vim:set shiftwidth=4 softtabstop=4 expandtab: */
--- a/wsd/ProxyRequestHandler.cpp
+++ b/wsd/ProxyRequestHandler.cpp
@@ -68,9 +68,14 @@ void ProxyRequestHandler::handleRequest(const std::string& relPath,
                         HttpHelper::sendErrorAndShutdown(400, socket);
                     }
                 }
+                catch(std::exception& exc)
+                {
+                    LOG_ERR("ProxyCallback: " << exc.what());
+                    HttpHelper::sendErrorAndShutdown(400, socket);
+                }
                 catch(...)
                 {
-                    LOG_DBG("ProxyCallback: Unknown exception");
+                    LOG_ERR("ProxyCallback: Unknown exception");
                     HttpHelper::sendErrorAndShutdown(400, socket);
                 }
             };
--- a/wsd/ProxyRequestHandler.cpp
+++ b/wsd/ProxyRequestHandler.cpp
@@ -14,10 +14,19 @@
 #include <net/HttpRequest.hpp>
 #include <net/HttpHelper.hpp>
 
+std::map<std::string, std::shared_ptr<http::Response>> ProxyRequestHandler::CacheFileHash;
+
 void ProxyRequestHandler::handleRequest(const std::string& relPath,
                                         const std::shared_ptr<StreamSocket>& socket)
 {
     Poco::URI uriProxy(ProxyServer);
+
+    const auto cacheEntry = CacheFileHash.find(relPath);
+    if (cacheEntry != CacheFileHash.end())
+    {
+        socket->sendAndShutdown(*cacheEntry->second);
+        return;
+    }
 
     uriProxy.setPath(relPath);
     auto sessionProxy = http::Session::create(uriProxy.getHost(),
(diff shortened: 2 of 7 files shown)


net/HttpRequest.hpp
```diff
@@ -678,6 +678,13 @@ class Response final
         saveBodyToMemory();
     }
 
+    Response(const Response& response)
+        : _statusLine(std::move(response._statusLine)),
+          _header(std::move(response._header)),
+          _body(std::move(response._body))
```
Ashod, line 684: Unfortunately, this won't work.

This is a copy constructor, but, instead of copying, it is moving the members of its argument. A copy constructor should only copy and should not leave the argument in "unspecified state" (as the standard calls moved-from objects). If we want a move constructor, then we should add it (we would need `Response&& response` parameter).

net/HttpRequest.hpp
```diff
@@ -971,6 +978,7 @@ class Session final : public ProtocolHandlerInterface
     std::chrono::microseconds getTimeout() const { return _timeout; }
 
     std::shared_ptr<const Response> response() const { return _response; }
+    const std::string getUrl() const { return _request.getUrl(); }
```
Ashod, line 981: Why not return a reference? After all, that's what `_request.getUrl()` returns (so returning a copy always costs an allocation, even when not needed).

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto sessionProxy = http::Session::create(uriProxy.getHost(),
+                                              http::Session::Protocol::HttpSsl,
+                                              uriProxy.getPort());
+    sessionProxy->setTimeout(std::chrono::seconds(10));
+    http::Request requestProxy(uriProxy.getPathAndQuery());
+    http::Session::FinishedCallback proxyCallback =
+        [](const std::shared_ptr<http::Session>& httpSession)
+            {
+                const auto callbackNow = std::chrono::system_clock::now();
+                auto sessionLink = LinkSession.find(reinterpret_cast<unsigned long>(httpSession.get()));
```
Ashod, line 50: We could use `void*` and avoid this cast. But, even better (and good news!) we don't need this map _at all_.

We simply need to capture the `std::shared_ptr<StreamSocket>` in this lambda! Simply change line 47 with this:
```
       [socket](const std::shared_ptr<http::Session>& httpSession)
```
And rename `sessionLink` to `socket` in the code below.

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                }
+
+                try
+                {
+                    const std::shared_ptr<const http::Response> httpResponse = httpSession->response();
+                    if (httpResponse->statusLine().statusCode() == 200)
+                    {
+                        if (MaxAge == std::chrono::system_clock::time_point())
+                        {
+                            MaxAge = callbackNow + std::chrono::hours(10);
```
Ashod, line 63: Perhaps we can have the lifetime as a named `constexpr` somewhere more visible?

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (MaxAge > std::chrono::system_clock::time_point() && timeNow > MaxAge)
+    {
+        CacheFileHash.clear();
+        MaxAge = std::chrono::system_clock::time_point();
+    }
+
+    const auto cacheEntry = CacheFileHash.find(relPath);
+    if (cacheEntry != CacheFileHash.end())
+    {
+        socket->sendAndShutdown(*cacheEntry->second.get());
```
Ashod, line 36: `shared_ptr` has a very friendly `operator*` exactly for this case. Use `*cacheEntry->second` (no need to call `.get()`). Same pattern applies elsewhere in this PR.

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    if (httpResponse->statusLine().statusCode() == 200)
+                    {
+                        if (MaxAge == std::chrono::system_clock::time_point())
+                        {
+                            MaxAge = callbackNow + std::chrono::hours(10);
+                        }
+
+                        std::shared_ptr<http::Response> entry =
+                            std::make_shared<http::Response>(*httpResponse.get());
+                        CacheFileHash[httpSession->getUrl()] = entry;
```
Ashod, line 68: Since we aren't really using this `http::Response` as it was intended, and since we really don't want to share it at all, why not store the instance directly (no pointers). (Alternatively, we could use `std::unique_ptr`, again, since don't need the overhead of `shared_ptr`.)

I'd use `std::unordered_map`, because it has a much better locality than `std::map` (i.e. CPU cache and memory friendly), and should also be faster, on average. Once the cache is built, it's unlikely to change much. Then, we can simply do this: `CacheFileHash.emplace_back(httpSession->getUrl(), *httpResponse);` 
This will store a copy of the `Response` object directly in the map. Then when we have a cache-hit, we simply do: `socket->sendAndShutdown(cacheEntry->second);`

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                        entry->add("Server", HTTP_SERVER_STRING);
+                        entry->add("Date", Util::getHttpTimeNow());
+                        sessionLink->second->sendAndShutdown(*entry.get());
+                    }
+                    else
+                    {
+                        HttpHelper::sendErrorAndShutdown(400, sessionLink->second);
+                    }
+                }
+                catch(...)
```
Ashod, line 79: Seems useful to catch `std::exception`, which more often than not will have valuable information for us. Also, reasonable to log as error here, or possibly warning, since these should be rare, but useful to understand edge-cases better.

wsd/ProxyRequestHandler.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void ProxyRequestHandler::handleRequest(const std::string& relPath,
+                                        const std::shared_ptr<StreamSocket>& socket)
+{
+    Poco::URI uriProxy(ProxyServer);
+    const auto timeNow = std::chrono::system_clock::now();
+
+    if (MaxAge > std::chrono::system_clock::time_point() && timeNow > MaxAge)
+    {
+        CacheFileHash.clear();
```
Ashod, line 29: This is race-free _only_ if this is called from `COOLWSD::getWebServerPoll()` exclusively. Otherwise, there is a race condition between the callback below where we do the insertion and this code.

Ashod (commented):
Thanks @hcvcastro. Good stuff. Looks promising. I left some notes and suggestions. Let me know if we need to discuss.

wsd/ProxyRequestHandler.cpp
hcvcastro: Yes, I did before capturing the socket, I wanted to try some alternatives. It seems reasonable to capture the socket
to simplify the code

wsd/ProxyRequestHandler.cpp
hcvcastro: Yes, simplify a constexpr time_point() by default is zero, I could not find any constant to apply

wsd/ProxyRequestHandler.cpp
hcvcastro: The idea is moving only pointers rather than moving large data blob, I have some other ideas to simplify to moving pointers

wsd/ProxyRequestHandler.cpp
hcvcastro: fair enough, I did it, but it seems forget to stage the chunk

wsd/ProxyRequestHandler.cpp
hcvcastro: The data logs says that the poller server thread only calls request handler and callback functions
unless you have data to refute them.

mmeeks:
Lets get this in - Henry has a follow-on fix soon.

PR #4943, opened 2022-06-23
jsdialog: listbox and scrollwindow are not containers which need additional parent

eszkadev:
This helps to keep id unique and process events correctly.


pedropintosilva (approved):
I have tested.

Before this PR containers were getting focus when going through elements via `TAB`:
image

With this PR this ^ is fixed

PR #4944, opened 2022-06-23
mobilewizard.css #fontsizecombobox.mobile-wizard only

Andreas-Kainz:
the css rule for #fontsizecombobox at file `mobilewizard.css` change also the behaviour at desktop sidebar so added `.mobile-wizard` and than the rule was for mobile only.

#### Mobile
Screenshot_20220624_004131

#### Desktop before
Screenshot_20220624_004035

#### Desktop after
Screenshot_20220624_004046

After the change the right alignment work as it should.

Change-Id: I704857bca49556abebca018ef7a284c7b2159530

### DIFF
--- a/browser/css/mobilewizard.css
+++ b/browser/css/mobilewizard.css
@@ -39,7 +39,7 @@ span#main-menu-btn-icon {
 	display: flex;
 }
 
-#fontsizecombobox {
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Thanks!

PR #4945, opened 2022-06-23
Sidebar fixed with and no jumping frame in frame no additional padding

Andreas-Kainz:
`.sidebar.ui-expander-content` get fixed `with: 309px;`
frame inside frame remove `padding-inline: 10px;`

**fixed with** can be shown best at sidebar table context
**frame inside frame** issue was at effects sidebar section and chart section

#### Before
https://user-images.githubusercontent.com/8517736/175432924-c7167ff5-6857-48f3-8e47-ca94386c410d.mp4

#### After
https://user-images.githubusercontent.com/8517736/175432956-89f621c0-2554-4b60-b039-58c1e5101c99.mp4

Change-Id: I263bb6cf33337e9dd1471e68142ba909d777d13f

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -93,7 +93,16 @@
 }
 
 .sidebar.ui-expander-content {
+	width: 309px;
 	padding-inline: 10px;
+}
+
+/* frame inside frame */
+#ChartElementsPanel .sidebar.ui-expander-content,
+#softedgeframe .sidebar.ui-expander-content,
+#glowframe .sidebar.ui-expander-content {
+	width: 100% !important;
+	padding-inline: 0px;
 }
 
 #selectcolor {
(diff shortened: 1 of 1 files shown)


browser/css/jssidebar.css
```diff
@@ -93,9 +93,18 @@
 }
 
 .sidebar.ui-expander-content {
+	width: 309px;
```
eszkadev, line 96: is that needed?

Andreas-Kainz: yes the 309px is needed, cause otherwise the sidebar elements jump. Maybe we don't need the 330px for the complete sidebar width. The idea was to fix the existing bug and than check if we can reduce stuff.

Andreas-Kainz: | **Issue at Master** | **Fix** |
| ------------- | ------------- |
|2022-06-24 11_45_38-Sidebar fixed with and no jumping frame in frame no additional padding by Andrea|2022-06-24 11_49_33-Sidebar fixed with and no jumping frame in frame no additional padding by Andrea|
| This is the existing sidebar layout if you are at writer when tab selected | with the defined with of 309px |

pedropintosilva:
This is now not needed since the sidebar structure revamp

PR #4946, opened 2022-06-23
Sidebar .toolbox items get flex-end aligned

Andreas-Kainz:
If there are more than one `.toolbox` element at an `.sidebar.cell` the `div.toolbox` get `flex-end` aligned.

It's a very generic approach and the goal is, that with additional PR's some css rules can be removed.

#### Before
Screenshot_20220624_015527

#### After
Screenshot_20220624_015533

Change-Id: Ic2f1268a6bc55f5f500a35e224fa69faa8a60652

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -165,7 +165,9 @@ div#thousandseparator.checkbutton.jsdialog.sidebar {
 #document-container ~ #sidebar-dock-wrapper .sidebar.jsdialog.vertical > .sidebar.jsdialog.row > table.sidebar.ui-grid > tr.sidebar > td.sidebar:last-child:not(:first-child),
 .sidebar.jsdialog.cell > #table-box3 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:last-child,
 .sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:last-child,
-.sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:nth-child(4) {
+.sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:nth-child(4),
+.sidebar.cell:not(:nth-child(1)) > .vertical > .row > .toolbox,
+.sidebar.cell:not(:nth-child(1)) > .toolbox {
 	display: flex;
 	justify-content: flex-end;
 }


browser/css/jssidebar.css
```diff
@@ -165,7 +165,9 @@ div#thousandseparator.checkbutton.jsdialog.sidebar {
 #document-container ~ #sidebar-dock-wrapper .sidebar.jsdialog.vertical > .sidebar.jsdialog.row > table.sidebar.ui-grid > tr.sidebar > td.sidebar:last-child:not(:first-child),
 .sidebar.jsdialog.cell > #table-box3 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:last-child,
 .sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:last-child,
-.sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:nth-child(4) {
+.sidebar.jsdialog.cell > #table-box4 > .sidebar.jsdialog.row > .sidebar.jsdialog.cell:nth-child(4),
+.sidebar.cell:not(:nth-child(1)) > .vertical > .row > .toolbox,
+.sidebar.cell:not(:nth-child(1)) > .toolbox {
```
pedropintosilva, line 170: Thanks but I'm afraid these will affect other cases. Ideally we would avoid those multiple :nth-child, I also wonder if those #table-box ids can be redone (in core) :thinking: 

TLDR: Do you think it will be possible to target those via other classes or maybe better straight from JS?

pedropintosilva (changes requested):

browser/css/jssidebar.css
Andreas-Kainz: yes you are right it's complex but we have in the sidebar sometimes table and sometimes div's so it's very difficult to catch dem.

Here the idea is to catch the right element which mean the last element within a table, div, ... excl. if there is only one element.

I also want to remove all the other complex stuff, but reduction (delte) stuff mean a lot of review and checks. So first add one css rule to fix the existing "visual bug" and than reduce stuff.

Andreas-Kainz: I'm way further by reduce rules, but one step after another.

Andreas-Kainz:
One year later, the sidebar didn't look better than 2022. I will close the request and maybe start from scratch, but please tell me the direction I should go.

pedropintosilva:
Well is not that nothing has changed. On the contrary, with a11y many controls now have different structure and are different HTML elements. Bottom line is: master branch is far from being in a releasable state (closing the gap but still..) so we need to me mindful of changes that fix something while possibly adding instability somewhere else. We need to aim for clear commits and steer away from big commit that not only is difficult to parse but also difficult to revert if needed

pedropintosilva:
Also ids have changes in some.places , could you please rebase this PR? I will then can try to check and test it left and right, thanks!

Andreas-Kainz:
I know that a lot of stuff had changed and get better an better. what I mean with nothing has changed was, that the placement of the sidebar elements in master didn't look perfect see screenshot.

image

If it's ok, I'll start a new PR where I will offer a better alignemnt of the sidebar elements. And I also know, with NO mixed stuff.

PR #4949, opened 2022-06-25
notebookbar: unify color control in all components

eszkadev:
Keep in sync colors in sidebar and notebookbar when
we change one of them.

Followup for
    commit 13ef21aa1ecff764ed7c5b24bee83e25297ca21e
    notebookbar: unify color control with core


pedropintosilva (approved):
Good catch @eszkadev sorry I didn;t see that when testing before. I can confirm, now swithcing the color in the sidebar correctly propagates and shows in the notebookbar

PR #4962, opened 2022-06-29
comments: don't lose focus if is edited

eszkadev:
When spreadsheet was opened after few seconds we received CellCursor
message.
If comment was opened end edited at that time - it dissapeared


Rash419 (approved):
Thanks : )

PR #4965, opened 2022-06-29
Private/pedro/improvements n fixes user list popover

pedropintosilva:
commit da9f10df79ec63b86af625d820c0cc8c0866d6f1 (HEAD -> private/pedro/improvements-n-fixes-userListPopover, origin/private/pedro/improvements-n-fixes-userListPopover)

    Fix padding in follow editor's entry (userListPopover)
    
    Use the same padding as the user names and thus fixing the
    alignment (vertical: checkbox vs avatars)
    
    Change-Id: I1acb8cf5e3e5dc8aad083151c9c54b38aa552d6c

commit 58a1ec616fe5c9787751aaa1fd5aa8eca2ca9412

    Add following label to userListPopover (avatar list)
    
    Until now user was reporting not fully understanding all the
    features that user userListPopover offers. Namely, not knowing
    that not only is possible to follow the editor (checkbox) but
    that it is also possible to follow a specific user from the list
    
    Make it easier to understand what's the current following status
    by adding "Following" under the selected user name in the list
    
    Change-Id: I0beba5df06f2cc7a9a349ef8f93db6b403befb9b

commit 36ea2ca662410d968928a0481cc4f554e53e566e

    Hide avatars list when follow editor changes status
    
    Get rid of the popover once the user completes the action
    Probably better to avoid having hanging popover (that have no
    close btn) and instead close them, specially because this
    one can always be re-opened at anytime from omnipresent
    toolbar and without risk of triggering any additional change
    
    Change-Id: Id7946512e6c54622e3ed20b28d56178c05b18b5c

commit 47aa7fc950a04d0ca52506fd225a9fb61a5f8c56

    Hide avatars list when a user entry is selected
    
    Make sure the avatar list popover automatically gets out of the
    way once the user selects an entry (follow user)
    
    Change-Id: Ie83a587e4f9900ffb0d70cadbfb13cfe81d8b849

commit 9a45988e82e0c3d0280d8361960de2ec7be6ac79

    Avatars list: Add missing hover styles
    
    Change-Id: I6a5b215316a988ad0649982415c2d8ce6fab1212

commit 62fe226aaf62c3e811a5fec027ad4dc7dbbe710c

    Use the same bg for #userListPopover and its pointing triangle
    
    Before the triangle was getting a different bg when compared
    to #userListPopover
    
    Change-Id: If110d9e1178bedb7e858804e169689ee97a5a385

commit 58afa7b12d58ced2dffba145ea13ff7a5b126778

    Fix position of avatar list
    
    Instead of trying to re-position #userListPopover for every case
    (compact view, tabbed view and rtl), use flex to align the elements
    - This also fixes the positioning discrepancy of both the popover
    but also the arrow (triangle)

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1234,6 +1234,8 @@ html[dir='rtl'] #userListPopover::after {
 
 #userListPopover .user-list-item.selected-user {
 	background-color: var(--color-background-dark);
+	display: grid;
+	grid-template-columns: 0fr 1fr;
 }
 
 #userListPopover #follow-editor {
@@ -1248,6 +1250,22 @@ html[dir='rtl'] #userListPopover::after {
 	width: 16px;
 	box-shadow: none !important;
 	background-position: center;
+}
+
+#userListPopover .user-list-item:not(.selected-user) .user-list-item--following-label {
+	display: none;
+}
+
+#userListPopover .user-list-item.selected-user .user-list-item--following-label {
+	display: flex;
+	padding-inline-start: 8px;
+	font-size: var(--default-font-size);
+	padding-block-end: 4px;
+	color: var(--color-text-lighter);
+}
+
+#userListPopover .user-list-item.selected-user .user-list-item--name {
+	padding: 4px 0 0 8px;
 }
 
 @media only screen and (max-width: 600px) {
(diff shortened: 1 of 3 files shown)


pedropintosilva:
Screenshot_20220629_144330

eszkadev (approved):
thanks, for me fixes issue with half-visible popover after selection

PR #4969, opened 2022-07-01
Fix TypeError when using formulabar variable in Writer

eszkadev:
This is for "0 JS errors" task


eszkadev:
to test: open any tunelled window in Writer and do some actions, mouse movements, clicks etc

Rash419 (approved):
Thanks  : )

PR #4970, opened 2022-07-01
WIP: introduce new options to localize upsell dialog

Rash419:
Change-Id: Ib9b89672c395eebab4085c7682c5383d4b5c460e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


kendy:
> looks good but probably the configuration setup can be improved and put all the feature locking config in same place?

@lpranam: We have considered various pro's and con's there, and in the end this turned out to be the most convenient way how to make it possible to have various unlock_link's for various alias groups.  The alternative would be to have a map of alias_group's host -> unlock_link in the feature_lock, but that could easily lead to missing this or that host in the separate list...

Thank you for the review - and hope this is bearable for you! :-)

common/CommandControl.hpp
```diff
@@ -49,31 +53,51 @@ class LockManager
...
+    }
+    static std::string getUnlockLink()
+    {
+        return config::getString("feature_lock.unlock_link", "");
     }
-    static std::string getUnlockLink() { return config::getString("feature_lock.unlock_link", ""); }
     static std::string getUnlockDescription()
     {
-        return config::getString("feature_lock.unlock_description", "");
+        return config::getString(translationPath + ".unlock_description", "");
```
kendy, line 71: @Rash419: I think instead of config::getString(translationPath + ".unlock_description", "");, I guess you want a function that checks if the particular "unlock_description" etc. exist in the given language.  I suspect it might happen there is an incomplete translation, and only eg. unlock_description is provided, but not the other - which I think will fail, and provide an empty string, instead of the default one (?)

wsd/COOLWSD.cpp
```diff
@@ -3475,6 +3565,22 @@ class ClientRequestDispatcher final : public SimpleSocketHandler
                 const auto pos = uri.find(ProxyRemoteStatic);
                 if (pos != std::string::npos)
                 {
+                    if (Util::endsWith(uri, "lokit-extra-img.svg"))
```
kendy, line 3572: This looks dangerous to me - from what I can see, we'll soon end up in a situation where we try to load the rating from an image URI or the other way around ;-)

Let's instead add serverURI as the 1st parameter of the ProxyRequestHandler::handleRequest(), and for the rating server, let's still have a ratingServerProxy constant that we can use there, and let's get rid of the setProxyServer() call.

kendy (commented):
@Rash419: Lots of great stuff, thank you! Can you have a look the comments before we integrate it, please?

Rash419:
@kendy thanks for the review I have updated the PR :)

kendy (approved):
Looks good to me, thank you, let's get it in!

PR #4973, opened 2022-07-03
USR2 to Backtrace and Signal-Safe Logging

Ashod:
These patches implement SIGUSR2 to backtrace the stack of the receiving process. For the moment, we backtrace the main thread only. In addition, we implement a long-desired, signal-safe method for capturing stack-traces in log files. Here we dump the backtrace in question into a special signalLog file and we then extract it (outside of the signal handler) and dump into the log file.

The changes are large mostly to accommodate these changes. That is, there is quite a bit of refactoring that isn't functional.

- wsd: jail: rename jails to childroot to distinguish them
- wsd: jail: move LO_JAIL_SUBPATH to JailUtil
- wsd: jail: no need to pass loSubPath between processes
- wsd: jail: add Poco namespace where missing in LinkOrCopy
- wsd: minor cleanup of addNewChild
- wsd: sig: prime libgcc and backtrace
- wsd: sig: merge setTerminationSignals with setFatalSignals
- wsd: sig: move signalLog helpers to SigUtil
- wsd: sig: write signalLog to SignalLogFD
- wsd: sig: support writing signalLog to files
- wsd: sig: dump signal logs into files and extract later

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -196,11 +196,11 @@ void cleanupJails(const std::string& root)
         LOG_WRN("Jails root directory [" << root << "] is not empty. Will not remove it.");
 }
 
-void setupJails(bool bindMount, const std::string& jailRoot, const std::string& sysTemplate)
+void setupChildRoot(bool bindMount, const std::string& childRoot, const std::string& sysTemplate)
 {
     // Start with a clean slate.
-    cleanupJails(jailRoot);
-    Poco::File(jailRoot + JAIL_TMP_INCOMING_PATH).createDirectories();
+    cleanupJails(childRoot);
+    Poco::File(childRoot + CHILDROOT_TMP_INCOMING_PATH).createDirectories();
 
     disableBindMounting(); // Clear to avoid surprises.
 
@@ -209,7 +209,7 @@ void setupJails(bool bindMount, const std::string& jailRoot, const std::string& 
     {
         // Test mounting to verify it actually works,
         // as it might not function in some systems.
-        const std::string target = Poco::Path(jailRoot, "cool_test_mount").toString();
+        const std::string target = Poco::Path(childRoot, "cool_test_mount").toString();
         if (bind(sysTemplate, target))
         {
             enableBindMounting();
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -156,8 +156,7 @@ void cleanupJails(const std::string& root)
         return;
     }
 
-    // FIXME: technically, the loTemplate directory may have any name.
-    if (FileUtil::Stat(root + "/lo").exists())
+    if (FileUtil::Stat(root + '/' + LO_JAIL_SUBPATH).exists())
     {
         // This is a jail.
         removeJail(root);
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -23,6 +23,7 @@
 #include <string>
 
 #include "Log.hpp"
+#include <SigUtil.hpp>
 
 namespace JailUtil
 {
@@ -122,7 +123,7 @@ static bool safeRemoveDir(const std::string& path)
 
 void removeJail(const std::string& root)
 {
-    LOG_INF("Removing jail [" << root << "].");
+    LOG_INF("Removing jail [" << root << ']');
 
     // Unmount the tmp directory. Don't care if we fail.
     const std::string tmpPath = Poco::Path(root, "tmp").toString();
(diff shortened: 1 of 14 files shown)


common/SigUtil.cpp
```diff
@@ -418,6 +418,13 @@ namespace SigUtil
         action.sa_handler = handleUserSignal;
 
         sigaction(SIGUSR1, &action, nullptr);
+
+#if !defined(__ANDROID__)
+        // Prime backtrace to make sure libgcc is loaded.
+        constexpr int maxSlots = 1;
```
mmeeks, line 586: Like this one =)

wsd/COOLWSD.cpp
```diff
@@ -511,8 +511,11 @@ static bool prespawnChildren()
 
 #endif
 
-static size_t addNewChild(const std::shared_ptr<ChildProcess>& child)
```
mmeeks, line 515: This is all fine of course, though I've no problem with const & generally.

common/JailUtil.cpp
```diff
@@ -156,7 +156,7 @@ void cleanupJails(const std::string& root)
         return;
     }
 
-    if (FileUtil::Stat(root + '/' + LO_JAIL_SUBPATH).exists())
+    if (FileUtil::Stat(JailUtil::getLoPath(root)).exists())
```
mmeeks, line 163: This one seems marginal - is it all tested ? I worry about our test coverage of this generally.

common/JailUtil.cpp
```diff
@@ -156,8 +156,7 @@ void cleanupJails(const std::string& root)
         return;
     }
 
-    // FIXME: technically, the loTemplate directory may have any name.
-    if (FileUtil::Stat(root + "/lo").exists())
+    if (FileUtil::Stat(root + '/' + LO_JAIL_SUBPATH).exists())
```
mmeeks: Not sure I like this one. What do we do about the cases where we can't chroot ? do these continue working ? do we have good tests for that ? That mode is really important for eg. OpenShift - and we should really have tests for it.

mmeeks (changes requested):
So - some cosmetic changes here that bring quite some validation burden I think.
Beyond that - this signal logging stuff is far from ideal I think.
* 98%+ of our users use stdout/stderr logging that streams to the journal - so Poco logging is a waste of their time I think.
* when we get a USR2 - to get a backtrace the process is normally hung - so we want to get data out of it fast; we can't count on it ever getting back to the mainloop - and training a random admin to do a dozen other unusual things to get data out of intermediate temporary files will not fly.
* the code is far more complex.

I'd prefer that we just log to stderr/stdout and avoid the Poco logging completely - and perhaps re-direct stdout/stderr to a file with something rather simple in the future to kill it completely ;-)

So - can you split this and get the non-controversial bits in ? =)

wsd/COOLWSD.cpp
Ashod: The reason for the change isn't `const&` it's when the `ChildProcess` instance is copied. Specifically, with the copy taken at the call-site, the copying is done outside the lock. With the `const&`, copying is done while the lock is taken, extending the duration of the lock and making contention more likely. I also reduced logging while the lock is taken for the same reason.

common/JailUtil.cpp
Ashod: Removed. Although it's a thin wrapper to avoid these hard-to-read concatenations and magic constants, I agree that testing wouldn't be cheap. Removed those changes altogether.

common/JailUtil.cpp
Ashod: I merely made "lo" a constexpr called LO_JAIL_SUBPATH (`constexpr const char LO_JAIL_SUBPATH[] = "lo";`). 

Really no idea about OpenShift and test coverage. There is no functional change here.

common/SigUtil.cpp
```diff
@@ -93,10 +94,23 @@ namespace SigUtil
     void checkDumpGlobalState(GlobalDumpStateFn dumpState)
     {
 #if !MOBILEAPP
+        assert(dumpState && "Invalid callback for checkDumpGlobalState");
         if (DumpGlobalState)
         {
-            dumpState();
             DumpGlobalState = false;
```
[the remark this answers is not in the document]
mmeeks: Hah =) ok - but given the confusion lets rename that flag and make the logic clearer there. Thanks =)

mmeeks (approved):
Fine - but of course there is quite a lot of risk for rather little feature-function benefit in this set - so lets not get it in until CODE<next> is branched ~Wed - but check with Andras.

PR #4974, opened 2022-07-03
GCC12 Build Fixes

Ashod:
### Summary

GCC 12 build fixes.

Please see commit logs for details.

### DIFF
--- a/wsd/ProofKey.cpp
+++ b/wsd/ProofKey.cpp
@@ -184,7 +184,11 @@ std::vector<unsigned char> Proof::RSA2CapiBlob(const std::vector<unsigned char>&
     if (exponent.size() > 4)
         throw ParseError("Proof key public exponent is longer than 4 bytes.");
     // make sure exponent length is correct; assume we are passed big-endian vectors
-    std::vector<unsigned char> exponent32LE(4);
+    // GCC 12 doesn't like it when the capacity of exponent is > ours.
+    // std::copy gives the following compile-time error:
+    // error: writing 16 bytes into a region of size 4
+    std::vector<unsigned char> exponent32LE(exponent.capacity());
+    exponent32LE.resize(4);
     std::copy(exponent.rbegin(), exponent.rend(), exponent32LE.begin());
 
     std::vector<unsigned char> capiBlob = {
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1762,7 +1762,7 @@ bool ChildSession::selectText(const StringVector& tokens,
 {
     std::string swap;
     unsigned winId = 0;
-    int type, x, y;
+    int type = 0, x = 0, y = 0;
     if (target == LokEventTargetEnum::Window)
     {
         if (tokens.size() != 5 ||


Ashod:
@mikekaganski for 756c1babf573a6cee7 I tried the simplest solutions (creating `exponent32LE` with the same size as `exponent`, and to change the range in the `std::copy` call to be 4, for example), but couldn't make the error go away. The only fix was to simply loop exactly 4 times.

It seems that GCC thinks the `exponent` coming from Poco is 16 bytes. My version is 1.10.1, which hasn't changed (except, it's also built with gcc12).

Interesting what you make of the error and if you have a better fix. Thanks!

mikekaganski:
It is very strange; it ignores the preceding length check. What are the compiler command line for that CPP? Maybe there's some environment flags active that make the compiler to optimize by reorder the execution, and some compiler bug appears then? It definitely looks to me that the code was correct with that regard, or I miss something.

wsd/ProofKey.cpp
```diff
@@ -185,7 +185,10 @@ std::vector<unsigned char> Proof::RSA2CapiBlob(const std::vector<unsigned char>&
         throw ParseError("Proof key public exponent is longer than 4 bytes.");
     // make sure exponent length is correct; assume we are passed big-endian vectors
     std::vector<unsigned char> exponent32LE(4);
-    std::copy(exponent.rbegin(), exponent.rend(), exponent32LE.begin());
```
mikekaganski, line 188: Could you please also check if

    std::reverse_copy(exponent.begin(), exponent.end(), exponent32LE.begin());

makes a difference?

wsd/ProofKey.cpp
```diff
@@ -185,7 +185,10 @@ std::vector<unsigned char> Proof::RSA2CapiBlob(const std::vector<unsigned char>&
         throw ParseError("Proof key public exponent is longer than 4 bytes.");
     // make sure exponent length is correct; assume we are passed big-endian vectors
     std::vector<unsigned char> exponent32LE(4);
```
mikekaganski, line 187: Additional idea: maybe it tries to "optimize" by some memcpy-like call, taking into account not the vector size, but allocation size - and std::vector's default ctor allocates 16 elements by default (IIRC). So maybe replace the

    std::vector<unsigned char> exponent32LE(4);

with

    std::vector<unsigned char> exponent32LE;
    exponent32LE.resize(4);

so that the ctor allocates (reserves) 16 bytes, and resize makes the first 4 bytes "used"?

wsd/ProofKey.cpp
Ashod: Can try, but I doubt it will make a difference (because I agree with your next comment).

wsd/ProofKey.cpp
Ashod: Right. This is exactly what I also thought is happening. The fact that it is missing the early return means this is a bug on their side.

Another fix would be to construct `expoent32LE` with the default ctor (so we get the default allocation) and then we resize to 4 (so we make 4 bytes usable). This should work, but it's easy for someone to "optimize" this and break it again. I'll try it anyway as it's cleaner and add a comment if it works.

wsd/ProofKey.cpp
Ashod: This didn't work. I got this error:
```
/usr/include/c++/12/bits/stl_algo.h:1169:21: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 1169 |           *__result = *__last;
      |           ~~~~~~~~~~^~~~~~~~~
```

wsd/ProofKey.cpp
Ashod: The default constructor doesn't work. Instead, I use `exponent.capacity()` followed by `.resize(4)` and that works.

mikekaganski (approved):

PR #4976, opened 2022-07-04
Allow buttons in vex dialogs to wrap

pedropintosilva:
In cases when the button text is too long
or when using CO in, for example, DE language and when
the length of the parent dialog was not enough the btns
would not stack up, instead that would overlap. This would
make reading those btns impossible.

vex-btns-overlap

Change-Id: I801ea330db21add35d0a81466a1fba9501a61899


Rash419 (approved):
Looks great thanks : )

PR #4981, opened 2022-07-05
private/pedro/formula bar improvmeents n fixes 05 jul

pedropintosilva:
- Formula bar: improve autosum menu button
- Formula bar buttons add hover effect
- Remove Toolbar-mobile.css, avoid bleeding out rules to desktop
- Formula bar: Set text area and expander with the same style
- Formula bar: improve expander img
- Formula bar: Increase font size of the text area
- Formula bar: Make sure the outline doesn't bleed out
- Formula bar: Fix formula and sum btns shifting when expanding


eszkadev (approved):
looks good

PR #4971, opened 2022-07-01
Keep custom buttons when switching the view

eszkadev:
When we use API call InsertButton and we add custom buttons
we want to keep them when we switch between compact/tabbed view.

PR #4984, opened 2022-07-06
Keep custom buttons when switching the view

eszkadev:
When we use API call InsertButton and we add custom buttons
we want to keep them when we switch between compact/tabbed view.


pedropintosilva:
port of https://github.com/CollaboraOnline/online/pull/4971 and it works as expected on master. Thanks!

pedropintosilva (approved):

PR #4985, opened 2022-07-06
private/pedro/formula bar padding fixes

pedropintosilva:
- Fix horizontal padding around formulabar
- Fix formulabar vertical padding and odd numbers
- Fix formulabar's item address input field vertical alignment
- Formulabar: no need to use calc for line-height
- Fix formulabar's address input vertical pos on mobile
- Formulabar: remove bottom padding from toolbar-wrapper


eszkadev (approved):
Thanks, now it looks well aligned

PR #4997, opened 2022-07-08
cypress: allow correct plain text clipboard formatting (master)

eszkadev:
Our plain text clipboard output generated additional newlines
at the beginning of every paragraph. To fix that we need
fix in core side which now fails due to expected bad format in cypress.
This is interim state where we accept both old and new format.

see for reference: https://gerrit.libreoffice.org/c/core/+/136893
lok: don't pretty print html for online
Change-Id: I2b17d62398d947fcf1d3fb1ed6005c3063d114f2


Rash419 (approved):
Looks great : )
Thanks

PR #5002, opened 2022-07-08
If integration theme does not exits, fall back to default theme

timar:
Change-Id: I9939ee243aa81b50969168df2e8bb5c8cbf2a34c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Tested, it works. If the integrator theme is not present and even if the coolwsd value is true it falls-back correctly tot he default theme.

PR #5008, opened 2022-07-12
Added support for setting the timezone in Docker

tobia:
### Summary

Several functions in Collabora Online depend on the server timezone setting.
The most common is possibly the rendering of dates and times using the `<text:date>`
tag in order to include the current date and time in the header or footer of documents.

Unfortunately current docker images do not provide any means to change the timezone
from the default UTC. Linux docker users should be able to mount the system's `/etc/localtime`
and similar system files, but this is a hack, does not support users in other platforms, and
does not seem to work in recent versions of the docker images.

This commit adds the the `tzdata` package to Ubuntu-based docker builds and the
`timezone` package to openSUSE builds. This allows docker users to run the containers
with a custom timezone by simply passing a TZ env variable. For example:

```
docker run -e TZ=Australia/Sydney collabora/code
```

Docker images based on Debian already include the `tzdata` package, therefore no change is needed.

### Checklist

### DIFF
--- a/docker/README
+++ b/docker/README
@@ -5,6 +5,9 @@ working docker container with Collabora Online.
 
 Docker image can be built from packages or from source code.
 
+All docker images built from this directory can be run with a TZ environment variable
+to set the server timezone to something other than UTC.
+
 Examples:
 
 1. Build latest CODE based on Ubuntu 18.04 LTS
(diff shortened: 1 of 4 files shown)


pedropintosilva:
Hello @tobia and thanks for taking the time to setup Collabora Online and attempt to contribute! I think this one is best to be reviewed by @timar :)

timar (approved):

PR #5009, opened 2022-07-13
wsd: fix warnings on GCC 12

Ashod:
### Checklist


mmeeks (approved):
Fair enough calming down (erroneous) warnings like this, there is no significant perf. impact =)

PR #5012, opened 2022-07-13
HAProxy: better load balancing

lpranam:
Instead of using URL parameter hash to direct all the users on same document to same server,
use stick-tables for it

original problem:
Problem was load balancer provided us with required persistence and balance both.
we directed traffic based on the url_param which means,
all the traffic with same url_param will go to same pod/server.
All the new values of url_param will be directed to different servers in round-robin way
this can be less efficient in a high traffic environment

solution:
this new method allows us to balance and manage persistence separately
so we can specify our desired algorithm for load balancing
and persistence is managed using stick-table

Change-Id: I80d63a2fc33f796c4534d94fb8de1e44c62ee9e2

* Target version: master 

### Checklist

### DIFF
--- a/kubernetes/helm/collabora-online/templates/ingress.yaml
+++ b/kubernetes/helm/collabora-online/templates/ingress.yaml
@@ -5,6 +5,12 @@ metadata:
   namespace: {{ .Values.namespace.collabora }}
   annotations:
     nginx.ingress.kubernetes.io/upstream-hash-by: "$arg_WOPISrc"
+    haproxy.org/backend-config-snippet: |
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
I like this part of the docs:
"  - stick-tables are replicated between all nodes in a multi-master fashion;"
https://cbonte.github.io/haproxy-dconv/1.7/intro.html
which is vital for us - and I expect the deployment we're looking at.

I'd like us to test this on share asap before updating the on-line docs.

Thanks Pranam ! =)

PR #5021, opened 2022-07-14
formulabar: mobile input fix

eszkadev:
formulabar: mobile: reuse TextInput for mobile IME
    
    Fixes typing on Android using GBoard in Chrome
    before this patch we didn't get any input due to
    lack of keyCode, we need to use 'input' events
    which are already handled in TextInput.js
    
    This patch does that only for mobile case, later
    we can try to unify it with desktop so we remove
    our custom formulabar input even there.

mobile: fix TypeError, missing Intl API
    
    this error was shown on socket disconnect on mobile Chrome in older
    phone

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1452,18 +1452,25 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		edit.addEventListener('change', function() {
 			if (callback)
 				callback(this.value);
-			if (data.rawKeyEvents) {
+			if (data.rawKeyEvents || data.useTextInput) {
 				// here event.keyCode has some non-ascii code
 			} else
 				builder.callback('edit', 'change', edit, this.value, builder);
 		});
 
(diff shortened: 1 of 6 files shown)


eszkadev:
This change requires in core:
https://gerrit.libreoffice.org/c/core/+/137086
https://gerrit.libreoffice.org/c/core/+/137085

It doesn't fix multi-line cell editing issues with linebreaks, which will be handled later in the followup.

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1461,107 +1461,131 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.rawKeyEvents) {
 			var modifier = 0;
 
-			edit.addEventListener('keydown', function(event) {
-				if (edit.disabled)
-					return;
+			if (window.mode.isMobile() || window.mode.isTablet()) {
```
eszkadev, line 1464: here added different handlers for mobile, other changes are just indentation level change

merttumer (approved):

PR #5030, opened 2022-07-16
wsd: slightly improve logging in case of version mismatch

thebearon:
Change-Id: I586da3390487727f0d4b743d4d285cce59f7d55b

* Target version: master

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -507,7 +507,9 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
 
         if (requestUri.getPath().find("browser/" COOLWSD_VERSION_HASH "/") == std::string::npos)
         {
-            LOG_WRN("client - server version mismatch, disabling browser cache. Expected: " COOLWSD_VERSION_HASH);
+            LOG_WRN("Client - server version mismatch, disabling browser cache. "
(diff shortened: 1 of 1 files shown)


wsd/FileServer.cpp
```diff
@@ -507,7 +507,7 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
 
         if (requestUri.getPath().find("browser/" COOLWSD_VERSION_HASH "/") == std::string::npos)
         {
-            LOG_WRN("client - server version mismatch, disabling browser cache. Expected: " COOLWSD_VERSION_HASH);
+            LOG_WRN("client - server version mismatch, disabling browser cache. Expected: " COOLWSD_VERSION_HASH "; Actual URI path with version hash: " + requestUri.getPath());
```
Ashod, line 510: Best to use clang-format, which would break this long line, to make the code consistent and easier for other without scrolling side-ways (when diffing side-by-side, or 3-way merging, we have less than half or 1/3rd the screen width).

Ashod (approved):
Thanks @thebearon!

wsd/FileServer.cpp
Ashod: Also, please use `<<` in logs instead of `+`.

thebearon: Thanks a bunch for the hint, hopefully it's better now.

PR #5035, opened 2022-07-18
tracing: re-work code, improve thread safety.

mmeeks:
We shouldn't write to the socket outside the main thread ever or
we'll get memory corruption; cf. a number of suspicious traces
around this code.

Change-Id: Ibeb36fffbecce960012a55b3ab4fd40c9996ae68

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024 (approved):
Looks OK, once I passes CI.

PR #5038, opened 2022-07-19
private/pedro/T36365

pedropintosilva:
- Add UI toggles to view menu (compact view) on tablets (type: text)
- Remove unnecessary separator from view menu entries (type: text)
- Make view menu entries behave similar across types


Rash419 (approved):
Thanks : )

PR #5046, opened 2022-07-20
Tablet: Fix Save icon floating when in presence of changes

pedropintosilva:
The dot above the save icon was not scrolling together
with its parent:
     - Make sure the little dot is always under a parent with
       position set to relative
     - Adjust top position due to that ^ change
     - It also fixes for narrow window on desktop.

Change-Id: I24278d09ace6e54769771fafb6f113ce9d4f1542

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -107,7 +107,7 @@
 	margin-left: 11px;
 }
 
-#Save1.savemodified:after {
+#File .unoSave.savemodified:after {
 	content: '';
 	display: block;
 	width: 6px;
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -107,6 +107,9 @@
 	margin-left: 11px;
 }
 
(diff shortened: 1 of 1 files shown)


browser/css/notebookbar.css
```diff
@@ -107,6 +107,9 @@
 	margin-left: 11px;
 }
 
+#Save1.savemodified {
```
eszkadev, line 110: `.unoSave` class could be used instead of `Save1`, `1` is generated to make ids unique but it's not guaranted to be in the same place, also might be changed in the future by some side effects of other changes...

eszkadev (approved):
fixes the issue :)

browser/css/notebookbar.css
pedropintosilva: Yes we have this problem still in more than on place, but thanks for the reminder, I think better to don;t include yet another bad practice here

pedropintosilva:
I have added another commit before the fix just to clean that up before introducing a new rule

PR #5050, opened 2022-07-22
calc: handle DoubleClick on AutoFillMarker

eszkadev:
This will allow to fill column with formula when data
are present in neighborhood


gokaysatir (approved):
Thanks for the fix :)

PR #5057, opened 2022-07-25
calc: scroll when we reached end of view on AutoFill

eszkadev:


eszkadev:
WIP:
- doesn't work with zoom !== 100
- desktop only

eszkadev:
Fixed: Desktop
Mobile: still uses old approach without scrolling (it's different case because it doesn't use Map.Mouse which handled auto scrolling for desktop)

merttumer (commented):
works great, thanks

merttumer (approved):

PR #5062, opened 2022-07-26
Fix the close button on windows without a title

Minion3665:
* Target version: master 

### Summary

- Previously going into Help > About > View widgets would show a window
  where the close button was rendered slightly out of the top of the
  window
- This was because without the title taking up a line, the title bar
  shrunk to be too small to contain the close button
- This PR makes sure that the title can never take up less than a
  line in height

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -757,6 +757,10 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	display: none;
 }
 
+.ui-dialog-title {
+	min-height: 1em;
+}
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Thanks.
Minimum-height does make sense and I have tested in other JSDialogs(e.g.: Fontwork Gallery from insert > Fontwork) and everything seems fine

As a follow up PR:
It would be good to also set the missing title for the WidgetTestDialog (About > View Widgets)
* Something like `JSDialogs Widgets`

PR #5067, opened 2022-07-27
Alias and remove duplicate SVG files

### LINKED ISSUE #5052: Remove duplicated SVG files and use iconURLAliases instead
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva @eszkadev 

### Detailed description and rationale
We want to use the existent possibility to specify name alias for icons that are  generated via jsbuilder (sidebar, notebookbar) so we can:
-  Avoid having multiple duplicated icons with different names:
  - Example: 142097b3402e0d9aa3221a6e326c6db20137e34a
- Avoid solely removing icons files
  - Even though we want to use the same icon, in some cases the uno commands and thus the generated id are actually different from ones used in writer and elsewhere (uno.AlignLeft <> uno.LeftPara)

### Code pointers
[iconURLALiases](https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2444)

### Links to related documentation
https://collaboraonline.github.io/post/build-code/

### Additional notes, tips & tricks
...

Minion3665:
* Resolves: #5052
* Target version: master 

### Summary

- This PR fixes #5052
- Files which were directly in `/browser/images` were the only ones I
  considered in-scope for this (not files in `/browser/images/dark` as I
  did not want any aliasing there to influence light theme)
- Files which weren't named `lc_*.svg` were not considered for this, as
  the function with the alias list transformed the names into that
  format
- Dupes were found with `find . ! -empty -type f -exec md5sum {} + | sort | uniq -w32 -d --all-repeated=separate`
  It is possible (although unlikely) that there was a hash collision and
  I incorrectly aliased an image. The names I was aliasing seemed sane
  though, so I doubt this was the case.
- I changed the behavior of the aliases object such that it is *after*
  the lowercasing of the name and not before; this was so that I didn't
  have to worry about if capitals were used (as there were some SVGs
  with different combinations of underscores etc. and I didn't want to
  assume they all always needed to be in CamelCase)

Signed-off-by: Skyler Grey <skyler3665@gmail.com>
Change-Id: I4d7972900195edeaa8f5f8941d18ab556319a7f7

### Checklist


pedropintosilva (approved):
I have tested on the various doc types and everything seems fine. I have also noticed that you are lowercasing before the list, so we can forgo of the camelCase when manually adding those to the list, I see no problem with that:) Maybe it's even better since It makes it equal to the actual used file names.

eszkadev (approved):
code looks good, thanks

PR #5070, opened 2022-07-27
android: fix closing the document on interactive dialogs

merttumer:
We get stuck on exitting the app progressbar on mobile
with the interactive dialogs such as csv import or macro security
dialog, we have a different use case for killing the document
as in the normal case. We are not running a separate process
but a thread and we dont trigger regular killing use case.

This piece is not relevant for android because we always have
one connection through fakesocket until the main thread is killed
and it is waiting on the mutex to finish before we finally exit the
document activity.

Change-Id: I99bd333152d40a04f95d4747705a721112317bb2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks @merttumer.

PR #5075, opened 2022-07-28
calc: scroll when we reached end of view on AutoFill

eszkadev:
- allow to use auto scrolling feature by event propagation
- it required to create new modified event which will
  contain coordinates of the autofill marker


merttumer (approved):
works great

PR #5077, opened 2022-07-28
fix selecting redline comment jumps to scroll to beginning

merttumer:
redline comments dont have position properties set
and some calculation needed in order to have it
and scrolling to comment relies on position property.

We have now instead getter for it and if it is a redline
comment it is calculated otherwise the default one is returned
because it is already set.

Change-Id: I6a3cefdd1c881783832de3de8fc1df7c849aad09

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works great, no jumping :)

PR #5080, opened 2022-07-28
impress: unittest: adjust to additional view

eszkadev:
To allow rendering slide previews without any fields under edit
there was a change introducing additional view in the core.

see: https://gerrit.libreoffice.org/c/core/+/137565


eszkadev:
this workaround is not needed, havereal fix

PR #5081, opened 2022-07-28
private/pedro/mwizard fix spans

pedropintosilva:
- Fix non-used automatically generated spans on mobileWizard
- Mobile: Make sure all unotoolbutton share padding and margin


eszkadev (approved):
looks ok

PR #5091, opened 2022-08-02
IWYU: Clean up includes in common/*

thebearon:
By first running
 bear make -j8
...then on individual files:
 iwyu_tool.py -v -p . <filename>
(some suggestions were disregarded)

Also:
- reordered includes, first from this project,
  then from Poco, finally from std,
- removed unnecessary 'config.h' include from:
   Authorization.cpp, Protocol.cpp, Session.cpp, SpookyV2.cpp
- 'config.h' include was missing from these, haven't added them
   as they haven't been causing a problem:
   Log.hpp, SigUtile.hpp, Util.hpp
- removed unnecessary LibreOfficeKitEnums.h include from
   Protocol.hpp/cpp

Change-Id: Iaf3e095b66622a4f3483ee478a7c531521b3bc91

* Target version: master 

### Checklist

### DIFF
--- a/common/Authorization.cpp
+++ b/common/Authorization.cpp
@@ -5,14 +5,14 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-#include <config.h>
-
 #include "Authorization.hpp"
 #include "Log.hpp"
 #include "StringVector.hpp"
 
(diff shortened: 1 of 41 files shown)


thebearon:
Will get back to it later.

Issue #5054, opened 2022-07-22
Add dark mode toggle


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @eszkadev 

### Detailed description and rationale
Currently dark mode kicks in depending on user's system due to
```css
@media (prefers-color-scheme: dark) {
```
https://github.com/CollaboraOnline/online/blob/master/browser/css/color-palette-dark.css#L1

It would be good as a next step to add:
- new icon within View tab (Tabbed view)
- new menu entries to view menu (Compact view)
- Ideally and at least for now it would be good to always load the normal palette and load the dark one only upon user's action (clicking the toggle)
  -  At least for now that the dark mode is not complete
  - 
### Code pointers
https://web.dev/terra-dark-mode/ has a nice JS snippet (using window.matchMedia) that we could use for explicitly loading a piece of CSS in one or the other case.

pedropintosilva:
this seems to be a problem already report on https://forum.collaboraonline.com/t/switch-between-light-dark-theme-code-22-05/1259/3 that people are getting dark mode by default depending on their system and as this is not ready yet would be good to get this easy hack in

Pransav290302:
I would like to add that dark mode toggle button with nice design @pedropintosilva ...

pedropintosilva:
Hi @Pransav290302 and welcome! I think Christopher has already a WIP fix for this but I think we have a couple of nice easy hacks waiting for someone to pick them up ;) like this one: https://github.com/CollaboraOnline/online/issues/4920

Pransav290302:
Yess,i wll help @ChrisGlad by providing him an icon .

pedropintosilva:
Thanks @Pransav290302 I have pinged you in the #5084 feel free to answer so I can assign you (Github doesn't allow to assign people unless they comment first)

pedropintosilva:
fixed with https://github.com/CollaboraOnline/online/pull/5102 and added to feature branch

PR #5092, opened 2022-08-02
Adding a dark mode toggle

ChrisGlad:
Signed-off-by: Christopher Gladman <chris.gladman2004@gmail.com>
Change-Id: I0e0c282cec24d9bb7fbfc4131ba3875976cea5fd

* Resolves: # <!https://github.com/CollaboraOnline/online/issues/5054>
* Target version: master 

### Summary
This makes changes to the Menu and Notebook bars, adding a new 
"Dark mode" option. 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
Thanks @ChrisGlad  I will review it but maybe better to do a bit of rewording, here are some tips:
- Start the commit subject (AKA title) with a capital letter
- Add commit description
- This is not LibreOffice
- The commit message should tell what the commit does, in imperative mood. If necessary, you can mention how the code used to work, too. But what actually is done by the commit should be in present tense. Not what you did. Not "Fixed hang in foo()" but "Fix hang in foo()"
- The first line is supposed to be a minimal summary. Do not end it with a period.
- If the change is very simple conceptually, that one line might be enough
- The rest of the commit message describes the change in greater detail. This is the body of this commit message. The body is written after the subject line with one blank line in between. The blank line is used by various tools (such as 'git log', 'git show', etc.) to differentiate subject from body. Further paragraphs are also separated by blank lines.
- Explain the problem being solved by this commit. More importantly, explain why these changes are being made, as opposed to how. The 'why' part is your responsibility, the 'how' part is code's responsibility.

pedropintosilva:
@ChrisGlad thanks but that doesn't change the commit message as you see when you do `git log` locally or when you click in the commits tab here in the  web interface

pedropintosilva:
- `git commit --amend`
- `git push origin name-of-your-branch  --force` in this case I guess `Request`

pedropintosilva (changes requested):
Also it would be good if the toggle would get selected upon switching to dark mode

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -1085,6 +1085,12 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 				],
 				'vertical': 'true'
 			},
+			{
+				'id':'toggledarktheme',
+				'type': 'bigmenubartoolitem',
```
pedropintosilva, line 1090: Cool, nice to use bigmenubartoolitem so we can share stugg between Menubar and notebookbar :+1:

ChrisGlad:
> Also it would be good if the toggle would get selected upon switching to dark mod

The code should add a tick in the Menubar item when dark mode is active. Do you want me to try something similar for the Notebookbar?

pedropintosilva:
> > Also it would be good if the toggle would get selected upon switching to dark mod
> 
> The code should add a tick in the Menubar item when dark mode is active. Do you want me to try something similar for the Notebookbar?

WRT menubar it looks nice! For the Notebookbar I meant that the would be good to have the button being set to `selected` when the dark mode is active (We have other buttons doing something similar such as ruler)

browser/src/control/Control.UIManager.js
```diff
@@ -51,6 +51,38 @@ L.Control.UIManager = L.Control.extend({
 			|| forceCompact === false;
 	},
 
+	toggleDarkMode: function() {
+		// the initial theme is 'light' unless it is already set to 'dark'
+		var InitialScheme = 'light';
```
eszkadev, line 56: we use camelCaseStartingFromSmallCharacter :)

lpranam (changes requested):
3 problems I noticed.

1. in Tabbed view button does not show if selected
2. If i reopen the doc or refresh the page, theme is again set to light theme. Interesting thing is, if it was supposed to be dark mode, in compact view dark mode option is checked even if it is currently showing light mode. in short in compact view, menu option is correctly checked even if theme is displayed incorrectly.
3. dark mode icon seems a little pixelated and also moon has a light gray area on the bottom edge but I think this can be fixed as follow-up.

pedropintosilva:
@lpranam 

> 3 problems I noticed.
> 
>     1. in Tabbed view button does not show if selected
I think I have pointed out this in https://github.com/CollaboraOnline/online/pull/5092#issuecomment-1202313115, but let's let that for a follow-up PR. Right now we really need to take out dark mode from being loaded automatically as it is without this PR.
> 
>     2. If i reopen the doc or refresh the page, theme is again set to light theme. Interesting thing is, if it was supposed to be dark mode, in compact view dark mode option is checked even if it is currently showing light mode. in short in compact view, menu option is correctly checked even if theme is displayed incorrectly.
Makes sense. But let's first merge this PR and then the follow-up can fix that :)

>     3. dark mode icon seems a little pixelated
 and also moon has a light gray area on the bottom edge but I think this can be fixed as follow-up.

Yes, this is just a placeholder, there is already a person interesting in contribute with an icon. See https://github.com/CollaboraOnline/online/issues/5054#issuecomment-1199931471

lpranam:
...and also you can put all those changes in the same commit you don't need to have 2 commits here. 
multiple ways you can do it
1. you can reset both your commits and recommit everything in one commit (to reset top 2 commits `git reset HEAD~2`)
2. you can just reset the last commit and amend the previous commit
3. you can squash them. https://gist.github.com/lpranam/4ae996b0a4bc37448dc80356efbca7fa

pedropintosilva:
Merged to feature branch with PR #5102

PR #5093, opened 2022-08-02
writer: introducing new mention feature

Rash419:
Change-Id: I326cd663737a2e8e04bedeccd8f130fec4014c26

* Core patch: https://gerrit.libreoffice.org/c/core/+/140015
* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1920,9 +1920,34 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			};
 
 			text.addEventListener('click', clickFunction);
-			text.addEventListener('keypress', function onEvent(event) {
-				if (event.key === 'Enter')
+			text.addEventListener('keydown', function onEvent(event) {
+				var preventDef = false;
+				var listElements = $('#' + treeViewData.id + ' li');
+				var currIndex = parseInt(entry.row);
+				var treeLength = treeViewData.entries.length;
+				var spanElement = 'span.ui-treeview-cell';
+				if (event.key === 'Enter') {
 					clickFunction();
+					preventDef = true;
+				} else if (event.key === 'ArrowDown') {
+					if (currIndex === treeLength - 1)
+						listElements.eq(0).find(spanElement).focus();
+					else
+						listElements.eq(currIndex + 1).find(spanElement).focus();
+					preventDef = true;
+				} else if (event.key === 'ArrowUp') {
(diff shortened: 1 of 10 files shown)


Rash419:
@eszkadev @pedropintosilva @kendy  to test this just open framed.doc.html start typing mention
review, feedback, suggestion appreciated

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1920,9 +1920,34 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
+					else
+						listElements.eq(currIndex + 1).find(spanElement).focus();
+					preventDef = true;
+				} else if (event.key === 'ArrowUp') {
+					if (currIndex === 0)
+						listElements.eq(treeLength - 1).find(spanElement).focus();
+					else
+						listElements.eq(currIndex - 1).find(spanElement).focus();
+					preventDef = true;
+				} else if (treeViewData.id === 'mentionList' && event.key !== 'Tab' && event.key !== 'Shift') {
```
eszkadev, line 1944: hmm, I would prefer to keep widgets generic, not having any logic for particular widget instances for custom dialogs.
Can we do it different way?

For example pass custom keydown callback to the widget which will be called before original one and will return boolean which will indicate if event was consumed and doesn't have to be handled anymore.

pedropintosilva (commented):
Thanks @Rash419 I have encountered some interesting behaviour. If zoom in (brwoser zoom) the list shows up in the wrong place:

image

Rash419:
> Thanks @Rash419 I have encountered some interesting behaviour. If zoom in (brwoser zoom) the list shows up in the wrong place:
> 
> image

@pedropintosilva I guess with every jsdialog popup we have 
@eszkadev thougths ?

browser/src/control/Control.JSDialogBuilder.js
Rash419: yep makes sense , thanks : )

eszkadev (approved):
From code reading perspective looks ok now :)

eszkadev (changes requested):
tested again rebased:
- nice, border is fixed now
but what I noticed:
1. type "@a"
2. select something from the list (here Alexandra)
3. you got in the document: "@a@Alexandra"

eszkadev (approved):
ok, needed core patch :)

pedropintosilva (approved):

PR #5096, opened 2022-08-03
DeepL translation integration

merttumer:
this patch relies on a core change:
https://gerrit.libreoffice.org/c/core/+/137199

Change-Id: Ie5f5dada18539586e7c5ea41b0a296b33ae8ee2e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


merttumer:
@pedropintosilva we need a new icon for uno:Translate

pedropintosilva:
I tried to create an icon reminiscent from the existent colibre theme language (used in review tab)

I also wonder why we are using the new translate btn in format instead of placing it in the Review tab

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -741,7 +741,12 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 				'type': 'bigtoolitem',
 				'text': _UNO('.uno:TransformDialog'),
 				'command': '.uno:TransformDialog'
-			}
+			},
+			{
```
pedropintosilva, line 745: Could you please move this into `getReviewTab`, after `LanguageMenu`

pedropintosilva (changes requested):

pedropintosilva (approved):

PR #5097, opened 2022-08-03
browser: Add proper id to StyleApply buttons on notebookbar

juliushaertl:
Change-Id: Ic654c9b3316a73a09033f58414c2e98043ccf0de

* Resolves: # 
* Target version: master 

### Summary

Use proper ids instead of the generated ones from the uno command which are not valid otherwise:

Before: `StyleApply%3FStyle%3Astring%3DDefault%26FamilyName%3Astring%3DCellStyles`
After: `StyleApplyDefault`

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -637,16 +637,19 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 						'type': 'toolbox',
 						'children': [
 							{
+								'id': 'StyleApplyDefault',
 								'type': 'toolitem',
 								'text': _('Default'),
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
It's ok but we should do something so we will use that id int the future. Currently in the HTML I still see command as an ID. (But it has to be done carefully because probably we have some old-style id based css rules)

juliushaertl:
> It's ok but we should do something so we will use that id int the future. Currently in the HTML I still see command as an ID. (But it has to be done carefully because probably we have some old-style id based css rules)

CSS rules with the odd format `` wouldn't work anyways. It might be good to apply the button id rather as a class to avoid duplicate id attribute cases. This is actually the case already in parts for the .unospan-{id} which is what we use in branding:

Screenshot 2022-08-04 at 13 21 32

PR #5098, opened 2022-08-03
browser: Give formulabar a fixed button width to avoid jumping

juliushaertl:
Change-Id: Id38951ddc79e99573008c63d4e6c7ed36fcc52a2

* Resolves: # 
* Target version: master 

### Summary

The unoAutoSumMenu button has a different size than unoacceptformula button, which leads to a minor jump when clicking into the input field. Setting a fixed width solves this.

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1198,6 +1198,10 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	opacity: 0.8;
 }
 
+#acceptformula {
+	margin-inline-start: 3px;
+}
+
 .formulabar.unotoolbutton:hover {
 	opacity: 1;
 }


pedropintosilva:
@juliushaertl I noticed one test is inexistent in your PRs maybe because you have old branch, could you rebase it, please?

juliushaertl:
Done, sorry used the wrong origin 🙈

pedropintosilva:
no sweat, I thought something was amiss

browser/css/jsdialogs.css
```diff
@@ -1196,6 +1196,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 .formulabar.unotoolbutton {
 	margin-inline-end: 4px !important;
 	opacity: 0.8;
+	width: 37px;
```
pedropintosilva, line 1199: Thanks and thanks for explaining to me where the jump occurs:


https://user-images.githubusercontent.com/65948705/183068080-7237ca42-e01d-4532-a55f-8c957316e272.mov



This fixes it! But having that extra space always after the `#FunctionDialog` looks a bit odd hm... maybe it would be better to refrain from changing every unotoolbutton in the formulabar and instead make sure `#acceptformula` occupies the exact same space as the former icon AutoSumMenu (that occupies more right now due to the additional arrow):

```css
#acceptformula {
    margin-inline-start: 3px;
```

What do you think?

pedropintosilva (changes requested):

browser/css/jsdialogs.css
juliushaertl: Makes sense, pushed as you suggested :)

pedropintosilva (approved):
Thanks!

PR #5100, opened 2022-08-03
Unique ids and proper label for row height / column width

juliushaertl:
Change-Id: I486c2709fd2dd5d133c4c46416dcea5ef0479fc1

* Resolves: # 
* Target version: master 

### Summary

- Use unique ids for the row height and column width menus
- Use proper label "Row height" instead of "Row width"

### Checklist


pedropintosilva (approved):
Nice catch, much appreciated

Issue #5053, opened 2022-07-22
Remove any trace of ID with digit as suffix within CSS files


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

### Detailed description and rationale
Elements with a digit at the end. Example #Save1 is generated to make ids unique, but it's not guaranteed to be always in the same place, meaning: sometimes the suffix might be added to the other element instead that shares the same base ID. Moreover, it can also get changed in the future as a result of other changes.

### Code pointers
Example: 160467ac964fbda60da7f72f7ea1e76576a3af3d

### Links to related documentation
...

### Additional notes, tips & tricks
...

ArshErgon:
Hey @pedropintosilva quick question, do we have to change the classes name also? and can you assign this to me?

pedropintosilva:
@ArshErgon Cool, assigned :) hm.. not sure if I understand. No need to change classes, I think. You can use other existing classes already or parent classes. So in the example of `#Save1` it was already fixed with https://github.com/CollaboraOnline/online/commit/160467ac964fbda60da7f72f7ea1e76576a3af3d, but we probably have more of these lying around

ArshErgon:
@pedropintosilva I am asking about this example: `#style1`  like this has a suffix with an integer, but `#style1` is already mentioned as an `<element id="#style1">` do I have to remove this also? I think removing this will cause trouble, they would be many cases like this, Should rename it as `#styleOne`

ArshErgon:
@pedropintosilva please close this issue it has already been solved.

pedropintosilva:
ok @ArshErgon thanks for taking the time to check and test it left and right, I highly appreciated !

PR #5101, opened 2022-08-04
remove suffix digit

ArshErgon:
* Resolves: # 
* #5053
* Target version: master 

### Summary
I went through all the CSS files, I only find suffix in digit in `browser/css/notebook.css`
the `#style1`, `#style2`,` #style3` they aren't defined yet inside the HTML or JavaScript files. So I rename the `#style1' to 
 `#style`, I have commented the duplicates out because it's not defined, if I done this right, I will delete those function or make them in one line.
Please tell me if I did right.

### TODO
* I see two files  `browser/css/color-palette-dark.css` and `browser/css/color-palette.css` they are actually the same just 2 lines of codes are different, I was thinking to make them one, inside `browser/css/color-palette-dark.css` there's this `@media (prefers-color-scheme: dark)` if I make another function and change the `dark` to `light` and put these two function inside one file lets say `color-palette.css` I'm pretty sure, It would work.

* We can speed up our CSS files if we minify them.

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -533,11 +533,12 @@
 	text-align: center;
 }
 
-#style1 img ~ .ui-drawing-area-placeholder-container .ui-drawing-area-placeholder {
-	font-family: Liberation Serif;
-	font-size: var(--default-font-size);
+#style img ~ .ui-drawing-area-placeholder-container .ui-drawing-area-placeholder {
+	font-family: Liberation Sans;
+	font-size: calc(var(--default-font-size)*2);
+	font-weight: bold;
 }
-
+/* 
 #style2 img ~ .ui-drawing-area-placeholder-container .ui-drawing-area-placeholder {
 	font-family: Liberation Serif;
 	font-size: var(--default-font-size);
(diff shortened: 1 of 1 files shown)


browser/css/notebookbar.css
```diff
@@ -533,11 +533,12 @@
 	text-align: center;
 }
 
-#style1 img ~ .ui-drawing-area-placeholder-container .ui-drawing-area-placeholder {
```
eszkadev, line 536: hi, thanks for contributing to the project :)
Unfortunately this is not what issue #5053 is about. This particular case is ok. We want to fix only "UNO toolbuttons".
Because currently when we have two buttons with the same command - id will be the same. So we add the number, but it's not guaranted that they will appear in the same order so better to use: `#Command1` -> `.unoCommand`.
So you should check if the element with id `XXXX1` is the toolbutton and has class `unoXXXX`, then it should be fixed.

eszkadev (changes requested):

browser/css/notebookbar.css
ArshErgon: okay @eszkadev , so what if I hit something like `#slide-2 fig` it doesn't have an `uno` class. I have to skip right? if yes, I went through all the CSS files, and I could not find any other.

eszkadev: yes `slide-2` is also ok :), then maybe the issue is solved already

pedropintosilva:
@ArshErgon so it seems there is no more suffixes WRT UNO toolbuttons, let me close this up. Let me close this up. Thanks agains @ArshErgon for testing and inspecting the code. Now we can be sure this can be closed !

PR #5105, opened 2022-08-04
Backport of Stuck-Saving

Ashod:
- wsd: test: handle save notifications in tests
- wsd: test: filterSendMessage -> filterSendWebSocketMessage
- wsd: test: cleanup and cosmetics
- wsd: Message supports new find and contains operations
- wsd: test: support filtering Kit messages
- wsd: timeout on saving stuck with test

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -442,6 +442,14 @@ void DocumentBroker::pollThread()
                 if (_docState.isDisconnected())
                 {
                     // We will never save. No need to wait for timeout.
+                    LOG_DBG("Doc disconnected while saving. Ending save activity.");
+                    _saveManager.setLastSaveResult(false);
+                    endActivity();
+                }
+                else
+                if (_saveManager.hasSavingTimedOut())
+                {
+                    LOG_DBG("Saving timedout. Ending save activity.");
                     _saveManager.setLastSaveResult(false);
                     endActivity();
                 }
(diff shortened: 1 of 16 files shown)


thebearon (approved):
Thanks Ash!

wsd/DocumentBroker.cpp
```diff
@@ -1787,6 +1795,9 @@ void DocumentBroker::setLoaded()
         _docState.setLive();
         _loadDuration = std::chrono::duration_cast<std::chrono::milliseconds>(
                                 std::chrono::steady_clock::now() - _threadStart);
+        _saveManager.setSavingTimeout(
```
mmeeks, line 1798: Not convinced by this here.

mmeeks (commented):
Not sure that the arithmetic here is good around saving and the time it should take.
I can imagine a document loading very fast say 50ms and taking 10 seconds to save on a busy system. No data on that - but not sure we cope with that well.

What's the story there - why 5 seconds ? and so on =)

wsd/DocumentBroker.cpp
Ashod: It's a way to cap it, with the intent to improve it and tune it. Agreed that it's a bit finger-in-the-air formula, but being able to cap the wait prevents getting stuck. Of course adding tracking of save duration and using that would be better. (We could also make it configurable, but we have too many of those already.)

PR #5106, opened 2022-08-04
Disable testing splitting cells

Minion3665:
- The split cells test currently uses tunneled dialogs to make sure that
  we can split table cells
- There is an open gerrit review to switch the split cells dialog to
  JSDialogs (https://gerrit.libreoffice.org/c/core/+/137791) which is
  blocked because the tunneled dialog needs to be removed in it, and
  that fails this test
- This test needs to be temporarily disabled to give coreside a chance
  to merge that review and build a new version of the libreoffice
  archive. Once that has happened, this test should be replaced with a
  rewritten version

Signed-off-by: Skyler Grey <skyler3665@gmail.com>
Change-Id: I0bb51e3cdbc04020e03c8d87696d014a0f693554

* Target version: master 

### Checklist

CC: @eszkadev @pedropintosilva

### DIFF
--- a/cypress_test/integration_tests/desktop/writer/table_operation_spec.js
+++ b/cypress_test/integration_tests/desktop/writer/table_operation_spec.js
@@ -312,7 +312,7 @@ describe('Table operations', function() {
 			.should('have.attr', 'width', '50%');
 	});
 
-	it('Split Cells', function() {
+	/*it('Split Cells', function() {
 		helper.typeIntoDocument('{downarrow}');
 
 		helper.typeIntoDocument('{ctrl}{a}');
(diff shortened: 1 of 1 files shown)


cypress_test/integration_tests/desktop/writer/table_operation_spec.js
```diff
@@ -312,7 +312,7 @@ describe('Table operations', function() {
 			.should('have.attr', 'width', '50%');
 	});
 
-	it('Split Cells', function() {
+	/*it('Split Cells', function() {
```
eszkadev, line 315: it can be done by `it.skip(` but as we will fix it soon it's ok :)

eszkadev (approved):

PR #5107, opened 2022-08-05
private/ash/storageflags

Ashod:
- wsd: remove unused uploadLocalFileToStorage
- wsd: mark last-upload as failed in early-failure
- wsd: track modifying user commands
- wsd: cosmetics and comments
- wsd: move storage attributes to DocBroker
- wsd: test: modernize UnitWOPIAsyncUpload_ModifyClose


mmeeks (approved):
Seems reasonable - though far from excited about the unknown / not-understood problem with the stuck-save unit test. Can we investigate that much further ?

PR #5109, opened 2022-08-05
private/pedro/backporting pr 5038

pedropintosilva:
- Add UI toggles to view menu (compact view) on tablets (type: text)
- Remove unnecessary separator from view menu entries (type: text)
- Make view menu entries behave similar across types


gokaysatir (approved):

Issue #4880, opened 2022-06-09
Add dark icons for content controls (writer > form)


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @rizmut  

### Detailed description and rationale
Add missing content controls icons for dark mode 

### Code pointers
Path `browser/images/dark/`
Before creating new icon from scratch please check if [LO Colibre theme](https://github.com/LibreOffice/core/tree/master/icon-themes/colibre_dark/cmd) has it already

### Additional notes, tips & tricks
Some icons can already exist in that path but need to be replaced for dark mode. Example [browser/images/dark/lc_insertpicturecontentcontrol.svg](https://github.com/CollaboraOnline/online/pull/4871/files#diff-363b72ee007c505e873439c8fe886881e9cf4e6d9c1d2587473fe461d185cde4)

PR #5029, opened 2022-07-15
Revise dark variants

rizmut:
Signed-off-by: Rizal Muttaqin <riz_17_oke@yahoo.co.id>

* Target version: master 

### TODO

- [ ] ...

### Checklist

- [ ] Code is properly formatted
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #5117, opened 2022-08-07
Update dark icons and add content control icons

Andreas-Kainz:
Update dark icons to use the correct colors
add content control icons for writer -> form

Change-Id: I4023a434e9fad27e58fadb843ff10e69f4dcf77e

* Resolves: #4880
* Target version: master


pedropintosilva:
Let me reference here the previous PR from @rizmut https://github.com/CollaboraOnline/online/pull/5029

pedropintosilva (changes requested):
It would prefer if this per was against `feature/dark-mode` so it enters super fast and so the development around dark-mode can be made easier

PR #5130, opened 2022-08-10
jsdialog: hanlde popup updates correctly

eszkadev:
Since 22.05 popups (eg autofilter) are handled by the same
code as dialog (Control.JSDialog.js). Allow processing their
messages.


Rash419 (approved):
Works as expected , thanks : )

Issue #2809, opened 2021-07-08
Add outline to color indicator in the Classic mode


pedropintosilva:
1. build it locally : https://collaboraonline.github.io/post/build-code/ or use [gitpod](https://github.com/CollaboraOnline/online#gitpod) 
2. Switch the ui mode in the `coolwsd.xml`
  - in `<user_interface>` change the value to `classic`  like so:
```xml
    <user_interface>
      <mode type="string" desc="Controls the user interface style. The 'default' means: Take the value from ui_defaults, or decide for one of classic or notebookbar (default|classic|notebookbar)" default="default">classic</mode>
    </user_interface>
```
3. Do a `make run` and observe the problem
4. Try to fix it :)
 - Make the colour rectangle be visible even when there is no colour (via some sort of outline etc)

_______

Right now our toolbar colour dropdown components ( `tb_editbar_item_fontcolor` `tb_editbar_item_backcolor`) look like this if:

- no text is selected
- the text that is selected has is black and has no backcolor (https://github.com/CollaboraOnline/online/issues/2810)

![image](https://user-images.githubusercontent.com/65948705/124913889-faa18800-dfef-11eb-88e6-b1dbc0c11ed2.png)

This is how it looks when we select a text with back color:
![image](https://user-images.githubusercontent.com/65948705/124913935-0725e080-dff0-11eb-8469-1455ee0ab125.png)

georg480:
![image](https://user-images.githubusercontent.com/79664365/131212665-77b640a1-5c67-4bd0-a696-a570f63834ba.png)

pedropintosilva:
@georg480 sorry for the delay.

The problem only when using classic mode. In the Notebookbar mode we already have that border : ) (that's why it's a easy hack )

-----

Switch to classic mode

1. Via `loolwsd.xml` change `notebookbar`
 to `classic` 
 
 ```
    <user_interface>
      <mode type="string" desc="Controls the user interface style (classic|notebookbar)" default="classic">notebookbar</mode>
    </user_interface>
```

pedropintosilva:
Hi @georg480 , are you still working or intend to work on this?

PR #5135, opened 2022-08-10
Add outline to color indicator for classic mode

bayramcicek:
- In classic mode, toolbar color indicators weren't visible when no
  text is selected. The reason is that color indicators didn't have
  border(outline) color.
- In fact, color indicators were just a box-shadow property. So,
  it was not possible(?) to add properties like border and
  border-radius.
- No need to use box-shadow property. A `<div>` tag can represent
  color indicators.
- Removed box-shadow property.
- Added `<div>` tag that has `class="selected-color-classic"`.

Signed-off-by: Bayram Çiçek <bayram.cicek@libreoffice.org>
Change-Id: I1b1b68fdc015fd3719d1c59ef7f782036a5934a5

* Resolves: #2809
* Target version: master 

### Summary

- Outline/border added to color indicator by creating `<div>` element.
- This commit only resolves outline/border issue in classic mode.

### Other issues

In classic mode: 
- [ ] Changes in toolbar's font&highlighting color, doesn't change the sidebar's font&highlighting color.
- [ ] In toolbar, selecting the same color, resets the color indicator to white.
- [x] In Calc's toolbar, there is no icon/image for Background Color (`class="backgroundcolor"`).
- [ ] In Draw's toolbar, font&highlighting color components doesn't work at all.

* Issues listed above still exist regardless of this patch. 

### Checklist

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -517,6 +517,22 @@ nav:not(.spreadsheet-color-indicator) ~ #toolbar-wrapper #toolbar-up.w2ui-toolba
 #toolbar-down table.w2ui-button .w2ui-tb-image {
 	margin: 5px 9px !important;
 }
+#toolbar-down table.w2ui-button .textcolor {
+	margin-bottom: 0px !important;
+}
+#toolbar-down table.w2ui-button .backcolor {
+	margin-bottom: 0px !important;
+}
+#toolbar-down table.w2ui-button .selected-color-classic {
+	height: 5px;
+	width: 26px;
+	margin: auto;
+	position: relative;
+	top: -6px;
+	border: 1px solid var(--color-border-dark);
+	border-radius: 7px;
+	outline: 1px solid var(--color-background-lighter);
+}
 .w2ui-toolbar .w2ui-break{
 	background-image: none;
 	background-color: #dae6f3;
--- a/browser/css/w2ui-1.5.rc1.css
+++ b/browser/css/w2ui-1.5.rc1.css
@@ -2943,6 +2943,15 @@ button.w2ui-btn-small:focus:before {
   border: 1px solid transparent;
   background-color: transparent;
 }
+.w2ui-toolbar table.w2ui-button .selected-color-classic {
+  height: 7px;
+  width: 26px;
+  position: relative;
+  top: -6px;
+  border: 1px solid var(--color-border-dark);
+  border-radius: 7px;
+  outline: 1px solid var(--color-background-lighter);
+}
 .w2ui-toolbar table.w2ui-button .w2ui-tb-image {
   width: 22px;
   height: 16px;
--- a/browser/js/w2ui-1.5.rc1.js
+++ b/browser/js/w2ui-1.5.rc1.js
@@ -6231,7 +6231,36 @@ w2utils.event = {
             var img  = '<td>&#160;</td>';
             var text = item.text;
             if (typeof text == 'function') text = text.call(this, item);
-            if (item.img)  img = '<td><div class="w2ui-tb-image w2ui-icon '+ item.img +'"></div></td>';
+
+            if (item.img) {
+                // color indicator container for classic mode
+                var colorContainer = '<div class="selected-color-classic"></div>';
+
+                /**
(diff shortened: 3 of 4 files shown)


eszkadev (commented):
It looks nice. @pedropintosilva if you are ok with it we can merge :)

eszkadev:
btw I've noticed in classic mode color picker still works like in the past (doesn't remember last used color but shows color of the currently selected text)

browser/css/w2ui-1.5.rc1.css
```diff
@@ -2943,6 +2943,15 @@ button.w2ui-btn-small:focus:before {
   border: 1px solid transparent;
   background-color: transparent;
 }
+.w2ui-toolbar table.w2ui-button .selected-color-classic {
+  height: 7px;
+  width: 24px;
```
pedropintosilva, line 2948: Probably this should be 26px, so it is aligned to the top element (since the sibling has 24px width + 1 margin left +1 margin right = 26px)

pedropintosilva (changes requested):
Thanks @bayramcicek for this! It works as expected and if @eszkadev is ok with the w2ui file changes then I'm happy to merge this. The only thing I could point out is:

pedropintosilva (approved):
Awesome, thanks!

bayramcicek:
Thanks to both of you for review! 
(BTW, I have no idea why one of the Cypress tests gives failure.)
(`make -C cypress_test check-mobile spec=writer/shape_properties_spec.js` <-- fails.)

pedropintosilva:
hm cypress is not passing on CI, I also tested locally and I confirm the test didn;t pass

pedropintosilva:
I'm now rebasing this branch locally with master to see if there is nay difference in the cypress outcomes

pedropintosilva (changes requested):
There is one more thing needed before we can merge this. Dimensions on mobile are being affected by this. You can see the cypress running by ``


This happens because we now have two elements that are occupying more height than expected: image Plus there misalignment going now there.

Could you please:
- Make sure we use `margin:auto` for `.selected-color-classic` when on mobile #toolbar-down
- Set `.selected-color-classic` with smaller height on mobile e.g.: height: 5px;
- Also set `#toolbar-down table.w2ui-button .w2ui-tb-image` element (inside  `#tb_editbar_item_backcolor` and `#tb_editbar_item_fontcolor`) with `0` bottom margin

browser/css/device-mobile.css
```diff
@@ -517,6 +517,25 @@ nav:not(.spreadsheet-color-indicator) ~ #toolbar-wrapper #toolbar-up.w2ui-toolba
...
 }
+#toolbar-down table.w2ui-button .textcolor {
+	margin: 5px 9px 0px !important;
+}
+#toolbar-down table.w2ui-button .backcolor {
+	margin: 5px 9px 0px !important;
+}
+#toolbar-down table.w2ui-button .backgroundcolor {
+	margin: 5px 9px 0px !important;
+}
```
pedropintosilva, line 528: I have tested and as far as I can see we don't use `.backgroundcolor` on mobile #toolbar-down so this can be removed

browser/css/device-mobile.css
```diff
@@ -517,6 +517,25 @@ nav:not(.spreadsheet-color-indicator) ~ #toolbar-wrapper #toolbar-up.w2ui-toolba
 #toolbar-down table.w2ui-button .w2ui-tb-image {
 	margin: 5px 9px !important;
 }
+#toolbar-down table.w2ui-button .textcolor {
+	margin: 5px 9px 0px !important;
+}
+#toolbar-down table.w2ui-button .backcolor {
+	margin: 5px 9px 0px !important;
```
pedropintosilva, line 524: maybe better to use `margin-bottom` instead, so to avoid duplications since we can inherit the rest

pedropintosilva (changes requested):
Thanks! It now passes all tests, and it works both on mobile and desktop :partying_face: . There are just 2 details that could be improved, sorry for nitpicking here.

pedropintosilva (approved):
Thanks!

Issue #1951, opened 2021-04-09
Can't edit file properties


manu-p:
**Describe the bug**
While the Collabora Office 6.4 Help page mention [here](https://help.collaboraoffice.com/6.4/en-US/text/swriter/01/04090004.html) and [here](https://help.collaboraoffice.com/6.4/en-US/text/shared/01/01100300.html?&DbPAR=WRITER&System=UNIX) some capabilities to edit "file properties", there is no actual way to edit the properties, as it is done in LO-Writer.

**To Reproduce**
1. Edit/create file properties in LO-Writer, for instance Title, Subject and occasionally some custom properties. Edit the document content using these fields.
2. Upload the .odt file in Nextcloud.
3. Edit the file using CODE.
4. Double-Click on a field in the document content to edit it.

**Expected behavior**
The ability to modify the field content.

**Actual behavior**
No option to edit the field: Edit is greyed

**Screenshots**
![Capture d’écran de 2021-04-09 11-14-36](https://user-images.githubusercontent.com/14906222/114159386-082e6f80-9926-11eb-9e6e-5cb5bdb52066.png)

**Desktop (please complete the following information)**
 - OS: Ubuntu 20.04
 - Browser: FF 87.0 and Brave 1.22.71 Chromium: 89.0.4389.114

**Additional context**
Workaround: download the document to edit fields with LO-Writer and re-upload it back in the cloud...

manu-p:
Hi, 
At this point, I have a choice: either nobody has read this ticket and that's why it is not even qualified as an anomaly or an issue, or the function does not exist in CODE and nobody says so explicitly... I'm leaning towards the second option. Anyone here to confirm?

manu-p:
OK, thanks @thebearon, and is there any plan to implement this standard feature in CODE, and by when?

PeterHagen:
I'm specially missing out the custom properties:

<img width="620" alt="Screenshot 2022-01-01 at 23 23 57" src="https://user-images.githubusercontent.com/12300899/147861371-57fd5138-0ff6-4e21-911f-010849c39ba7.png">

Daenou:
Just wanted to open a new issue but fortunately I saw that this already exists.
We're facing this issue everyday and it is a reason why CollaboraOnline isn't as popular within our organization. We use custom properties a lot and having to download the document and edit them in LibreOffice, only to reupload them defeats the whole purpose of LibreOffice. I would very much like this to be implemented.

PR #5137, opened 2022-08-11
Enable file properties online

NickWingate:
* Resolves: #1951
* Target version: master 
* Core patch: https://gerrit.libreoffice.org/c/core/+/138010

### Summary
Expose file properties dialog online
Currently this is a scaled back version with fewer features

Fixed various related bugs:
- Tabs not correctly selecting
    -  Use local index of correct div in `contentDivs` instead of depending on internal ID which causes problems when IDs are non-sequential
    - `Control.JSDialogBuilder.js:1046`
- Errors when an empty grid is sent
    - `Control.JSDialogBuilder.js:572`

### To-be implemented:
- [ ] Digital signatures dialog
- [ ] Security tab page (protect document)
- [ ] Several custom property features:
    - [ ] DateTime, Date, Number types
    - [ ] Custom property name

### Checklist


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -114,6 +114,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		this._controlHandlers['spinnerimg'] = this._spinnerImgControl;
 		this._controlHandlers['image'] = this._imageHandler;
 		this._controlHandlers['scrollwindow'] = this._scrollWindowControl;
+		this._controlHandlers['scrollbarbox'] = this._ignoreHandler;
```
eszkadev, line 117: I see no effect of this, what does it fix?

NickWingate: Accidentally left in, ignore.

cypress_test/integration_tests/common/helper.js
```diff
@@ -921,14 +921,14 @@ function waitUntilIdle(selector, content, waitingTime = mobileWizardIdleTime) {
...
-function clickOnIdle(selector, content, waitingTime = mobileWizardIdleTime) {
+function clickOnIdle(selector, content, waitingTime = mobileWizardIdleTime, forceClick = false) {
 	cy.log('Clicking on item when idle - start.');
 
 	waitUntilIdle(selector, content, waitingTime);
 
 	if (content) {
 		cy.contains(selector, content)
-			.click();
+			.click({force: forceClick});
```
eszkadev, line 931: hmm, this one looks like it will never fail - but in some cases should, for example object is covered by other element.
It should not be changed. Why it was needed?

eszkadev: if button was not visible and it failed (in notebookbar) we just do a scroll right in other tests - this can help

eszkadev:
looks good, please fix the cypress thing and it's done :)

eszkadev (approved):

PR #5143, opened 2022-08-11
Improve keyboard focus

juliushaertl:
Change-Id: I712d5457b65c0c8a84efdfb3d031160c8455ea3c

* Resolves: # 
* Target version: master 

### TODO

- [x] Convert notebook bar tabs to buttons
- [x] Convert close button to an actual button
- [ ] Fix styling of close button
- [ ] Figure out why the focus is brought back to the leaflet container after some time
- [ ] Find a way to register a global keyboard shortcut to move the focus out of leaflet to the menu bar
- [ ] Extend further cases of the JSDialogBuilder to use proper elements

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -198,7 +198,7 @@ m4_ifelse(MOBILEAPP,[true],
       </div>
 
       <div id="closebuttonwrapper">
-        <div class="closebuttonimage" id="closebutton"></div>
+        <button class="closebuttonimage" id="closebutton"></button>
       </div>
      </nav>
 
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -998,9 +998,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 				var title = builder._cleanText(item.text);
 
-				var tab = L.DomUtil.create('div', 'ui-tab ' + builder.options.cssClass, tabsContainer);
+				var tab = L.DomUtil.create('button', 'ui-tab ' + builder.options.cssClass, tabsContainer);
 				tab.id = item.name + '-tab-label';
 				tab.number = item.id - 1;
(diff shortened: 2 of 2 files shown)


juliushaertl:
>  Figure out why the focus is brought back to the leaflet container after some time
>  Find a way to register a global keyboard shortcut to move the focus out of leaflet to the menu bar

@pedropintosilva @eszkadev I would need some pointers for those two items. I've noticed that the keyboard events are catched in Map.Keyboard.js and mostly forwarded to core as UNO commands but I'm unsure what a proper way of achieving the above mentioned goals would be.

I think for the shortcuts it might make most sense to stick to what is used on Collabora Office on the desktop:
- F6 | Repeatedly pressing F6 switches the focus and circles through the following objects:•Menu bar •Every toolbar from top to bottom and from left to right •Every free window from left to right •Document
- F10 or Alt | Switches to the Menu bar and back.

eszkadev:
> Figure out why the focus is brought back to the leaflet container after some time

In Calc there is `CellCursor` message from core which often moves focus to the map. We don't do that if input is focused. Worth checking that place: https://github.com/CollaboraOnline/online/blob/master/browser/src/layer/tile/CanvasTileLayer.js#L4300

eszkadev:
> Find a way to register a global keyboard shortcut to move the focus out of leaflet to the menu bar

Yes, `Map.Keyboard.js` seems to be good point to start.

We have one place where we pass keyboard event to JSDialogs (only dialogs and popups, not sidebar or notebookbar) to check if it should be handled there (ESC key):
https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.Keyboard.js#L252

But most of global actions are in `_handleKeyEvent`, like F1 for Help.

If you need to access UI components: `map.uiManager` -> `Control.UIManager.js` has code related to the toolbar, notebookbar or sidebar.

juliushaertl:
Closing as this is beeing addressed differently already in the meantime

PR #5144, opened 2022-08-12
Scrollbar: step by step quick scroll

eszkadev:
Instead of jump to the position on a scrollbar, the quick scroll
action (click above/below scrollbar marker) should move us one
page up/down.


eszkadev:
this is demo, I would like to add one thing:
- continue scrolling when user holds button pressed

pedropintosilva (approved):

pedropintosilva:
Thanks Szymon! We can then later discuss  Gokay if it makes sense to do something to scrollbar dynamic size etc but this is good to be merged :+1:

Issue #5084, opened 2022-07-29
Add dark mode toggle icon


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

### Detailed description and rationale
This icon is to be used in the new dark mode toggle #5054 

### Code pointers
* Where: browser/images
* Branch: Please push PR against [feature/dark-mode](https://github.com/CollaboraOnline/online/tree/feature/dark-mode) (not master)

pedropintosilva:
@Pransav290302 ^

hanswang123456:
Can I attempt this issue?

pedropintosilva:
Hi @hanswang123456 yes!

pedropintosilva:
I have just assigned this to you

hanswang123456:
Feel free to work on the issue too @Pransav290302

hanswang123456:
I have made made a prototype for the darkmode icon. let me know if I need to re-title the pull request or change up the design.

hanswang123456:
@pedropintosilva, I've replaced the previous pull request and made sure to only have 1 commit.

PR #5152, opened 2022-08-14
fix(images):New Prototype Darkmode Icon

hanswang123456:
* Resolves: #5084
* Target version: feature

### TODO

- [] ...

### Checklist


pedropintosilva (changes requested):
Hi @hanswang123456 and thanks for your first attempt at contributing to Collabora Online :partying_face: 

* Thanks for the icon! Maybe it would be good if you could
  * Use the same shade of black as use already in other filled icons `#3A3A38` instead of pitch black (`#000000`)
  * Also, maybe it would be better to avoid using double "stroke" (border) and use only one border, so it is consistent with other circular icons (`lc_about.svg`, `lc_feedback.svg`,  `lc_accessibilitycheck.svg`)
* Then please squash all your commits into one, :) better to avoid commits that change previous unmerged commits.

hanswang123456:
Ill reopen a pull request

PR #5156, opened 2022-08-16
jsdialog: unify clear formatting icon

eszkadev:
There is no reason for custom handler.
Thanks to that we don't need special css.


pedropintosilva (approved):
Tested and it works , really cool to see all those lines being eliminated, thanks!  I have pushed 1 additional fix for sidebar's #resetattr and its hover state + another one that removes related

PR #5160, opened 2022-08-17
wsd: log unmount errors only when mounting is enabled

Ashod:
When bind-mounting is disabled, we do a best-effort
to unmount any lingering mount-points. This is a
cleanup that helps with leftovers. Here, we
do a debug log level in case the unmounting fails,
and error, otherwise, when bind-mounting is enabled.

Change-Id: I199d6234aebfd84e6be812e5b7d3758273086815


thebearon (approved):
Thanks, Ash!

PR #5162, opened 2022-08-17
private/pedro/jsdialog tabs improvements

pedropintosilva:
- Remove gray block bellow JSDialog tabs
- Increase jsdialog tab btn contrast
- Increase jsdialog tab btn text contrast
- Fix jsdialog title background and re-use CSS var


eszkadev (approved):
looks good, added fix to avoid "black focus border" on tab switch

PR #5163, opened 2022-08-17
comment bubbles: don't use undefined container on comment remove

eszkadev:
- Show error instead of throwing an TypeError exception.
- Prevent us from showing empty popup when removed comment.

noticed when testing: https://gerrit.libreoffice.org/c/core/+/138398


lpranam (approved):
Looks good, works as expected.

PR #5176, opened 2022-08-22
jsdialog: find real input on setText action

eszkadev:
- it is needed for mobile wizard spin fields

Removes annoying bug in Chrome. When using spinfields in sidebar - when you click on "arrow up" it focuses "arrow down" what causes revert of the last change on next click.

requires code change: https://gerrit.libreoffice.org/c/core/+/138700


Rash419 (approved):
Thanks I tested again with core patch and it works great  : )

PR #5180, opened 2022-08-23
jsdialog: allow boolean as treeview entry state

eszkadev:
after core commit:
https://cgit.freedesktop.org/libreoffice/core/commit/?h=distro/collabora/co-22.05&id=62d5622540c7251bb870a837b5ca6a836fac6f01
DumpAsPropertyTree: use more efficient overloads of JsonWriter::put

we use boolean not string

this also enables test broken by that core commit


Rash419 (approved):
Thanks :+1:

PR #5187, opened 2022-08-24
jsdialog: correctly setup & test value for spinfields

eszkadev:
Adding attribute by jQuery and testing in cypress was
not working: visible value was correct after change but
in DOM we had original value which was tested by cypress.

Use val() to setup value and have.value for cypress to test.

Unify inputs so we don't use attributes but always props.


Rash419 (approved):
Looks great! thanks : )

PR #5189, opened 2022-08-26
Make the repair document dialog into a JSDialog

Minion3665:
- The repair document dialog is a leaflet dialog
- As this dialog is online-side only, I'll use the JSDialog builder
  directly rather than writing a UI file
- Most of the code to build the dialogs is already very adaptable from
  looking at the protocol of existing JSDialogs that come from the
  server & where the code goes, however a small change had to be made to
  dialog modifications to allow us a callback function, as otherwise we
  would send off to the server whenever we pressed on a list item
- Finally, if there's no list to show, we'll show a bit of text saying
  that there's nothing, as otherwise it looks like an empty/broken dialog

Change-Id: Iab9e943a428e66b05e28819c2ee1001a2deffd2c

* Target version: master 

### Checklist

a gif of it working
a screenshot of multiple clients using it

### DIFF
--- a/browser/src/control/Control.DocumentRepair.js
+++ b/browser/src/control/Control.DocumentRepair.js
@@ -2,89 +2,108 @@
 /*
  * L.Control.DocumentRepair.
  */
-/* global _ _UNO */
+/* global _ */
 L.Control.DocumentRepair = L.Control.extend({
-	options: {
-		position: 'topright'
-	},
+	builder: null,
+	actions: null,
+	selected: null,
 
-	initialize: function (options) {
-		L.setOptions(this, options);
-	},
+	addTo: function (map) {
+		this.remove();
 
-	onAdd: function () {
-		this._initLayout();
+		var data = {
+			id: 'DocumentRepairDialog',
+			dialogid: 'DocumentRepairDialog',
+			type: 'dialog',
+			text: 'Repair Document',
+			title: 'Repair Document',
+			jsontype: 'dialog',
+			responses: [
+				{
+					id: 'ok',
+					response: 1
+				},
+				{
+					id: 'cancel',
+					response: 0
+				},
+			],
+			enabled: true,
+			children: [
+				{
+					id: 'dialog-vbox1',
+					type: 'container',
+					text: '',
+					enabled: true,
+					vertical: true,
+					children: [
+						{
+							type: 'treelistbox',
+							id: 'versions',
+							enabled: false,
+						},
+						{
+							id: 'dialog-action_area1',
+							type: 'container',
+							text: '',
+							enabled: true,
+							vertical: true,
+							children: [
+								{
+									id: '',
+									type: 'buttonbox',
+									text: '',
+									enabled: true,
+									children: [
+										{
+											id: 'cancel',
+											type: 'pushbutton',
+											text: '~Cancel',
+											enabled: true
+										},
+										{
+											id: 'ok',
+											type: 'pushbutton',
+											text: '~Ok',
+											enabled: true,
+											'has_default': true,
+										}
+									],
+									vertical: false,
+									layoutstyle: 'end'
+								},
+							],
+						},
+					]
+				},
+			],
+		};
 
-		return this._container;
-	},
+		this._map = map;
+		this.actions = [];
 
-	_initLayout: function () {
-		this._container = L.DomUtil.create('div', 'leaflet-control-layers');
-		this._container.style.visibility = 'hidden';
+		var dialogBuildEvent = {
+			data: data,
+			callback: this._onAction.bind(this),
+		};
 
-		var closeButton = L.DomUtil.create('a', 'leaflet-popup-close-button', this._container);
-		closeButton.href = '#close';
-		closeButton.innerHTML = '&#215;';
-		L.DomEvent.on(closeButton, 'click', this._onCloseClick, this);
 
-		var wrapper = L.DomUtil.create('div', 'leaflet-popup-content-wrapper', this._container);
-		var content = L.DomUtil.create('div', 'leaflet-popup-content', wrapper);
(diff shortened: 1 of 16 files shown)


browser/src/control/Control.DocumentRepair.js
```diff
@@ -2,89 +2,116 @@
...
-		inputButton.type = 'button';
-		inputButton.value = _('Jump to state');
-		L.DomEvent.on(inputButton, 'click', this._onJumpClick, this);
-		content.appendChild(document.createElement('br'));
-		content.appendChild(document.createElement('br'));
-		content.appendChild(inputButton);
+	addTo: function (map) {
+		this.remove();
+		/*
+		this.builder = L.control.jsDialogBuilder({
```
eszkadev, line 15: some leftover?

Minion3665: > some leftover?

yep

I also still need to fix cypress tests, but thanks for reminding me

pedropintosilva:
@Rash419 could you please check if this is ok or if there is a better way to do this using this or other helper WRT Cypress tests. 
Note: it seems we were handling multi user repair document in different ways depending on the document (but I would except if not the whole repair document at least part of it should be handled in centralized place)

Rash419:
Everything looks great  : ) . I just noticed we don't have repair document test for mobile/writer and mobile/impress. If you can add them then It would be great.

Minion3665:
> Everything looks great : ) . I just noticed we don't have repair document test for mobile/writer and mobile/impress. If you can add them then It would be great.

Sure thing; while looking @ the other mobile test it seems that mobile is just not working at all, so I'll fix up mobile and then make those tests

pedropintosilva:
@Minion3665 thanks for hunting these cypress tests down :) Here are some things it might help:

- To run the test in multiuser mode: `make run-multi spec=writer/repair_document_spec.js`
- It seems the typing ends up to be interrupted ('H') and cypress is already opening the repair dialog and then trying to check if the whole word is there

image

pedropintosilva:
or you can also do it headless `make check-multi spec=writer/repair_document_spec.js`

Minion3665:
> * `make run-multi spec=writer/repair_document_spec.js`

Thank you for the tip
i'm sort of 50% doing that, as it seems that whenever I try to `check-multi` the make script crashes, however I strongly suspect this just to be my setup. Running the commands that make outputs works successfully (at some point I might make a script to automate that), however it obviously doesn't spot stuff like eslint failures

image

Minion3665:
> @Minion3665 thanks for hunting these cypress tests down :) Here are some things it might help:
> 
> * To run the test in multiuser mode: `make run-multi spec=writer/repair_document_spec.js`
> * It seems the typing ends up to be interrupted ('H') and cypress is already opening the repair dialog and then trying to check if the whole word is there
> 
> image

Also, due to running in docker I can't run in interactive mode which hampers testing a bit. My screenshots appear to come out fine though (and the test has passed on my local machine during the last commit) so I'm hoping that I've fixed that anyway

Minion3665:
I've just uploaded another patchset with the frame variable removed, and it seemed to pass eslint so hopefully it should pass testing here too; I guess we'll see in an hour or so

pedropintosilva:
It still fails for me, I got same result as shown in https://user-images.githubusercontent.com/65948705/188681254-d94fd6fa-f462-45da-b242-c5708b067173.png
 ```javascript
15  iframe  #iframe2
16  find    #DocumentRepairDialog
16  assert  expected <form#DocumentRepairDialog.jsdialog-container.ui-dialog.ui-widget-content.lokdialog_container> to exist in the DOM 18
18  iframe  #iframe2    19
19  find    #versions   contains
.ui-listview-entry, Typing: “World”
```

**Comment by pedropintosilva** [comment]
Waiting after typing resulted in passed test on my side (I have run it twice, and it passes):

```diff
diff --git a/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js b/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
index 45d4d18ac..01a00a032 100644
--- a/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
+++ b/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
@@ -20,6 +20,7 @@ describe('Repair Document', function() {
                cy.customGet('.leaflet-layer', frameId1).click();
 
                helper.typeIntoDocument('Hello World', frameId1);
+               cy.wait(2000);
 
                repairHelper.rollbackPastChange('Typing: “World”', frameId2);
```

Minion3665:
> Waiting after typing resulted in passed test on my side (I have run it twice, and it passes):
> 
> ```diff
> diff --git a/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js b/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
> index 45d4d18ac..01a00a032 100644
> --- a/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
> +++ b/cypress_test/integration_tests/multiuser/writer/repair_document_spec.js
> @@ -20,6 +20,7 @@ describe('Repair Document', function() {
>                 cy.customGet('.leaflet-layer', frameId1).click();
>  
>                 helper.typeIntoDocument('Hello World', frameId1);
> +               cy.wait(2000);
>  
>                 repairHelper.rollbackPastChange('Typing: “World”', frameId2);
> ```

If we're going to need this everywhere we rollback (to avoid issues with interrupting other stuff) would it make sense to put in the helper, or do you think that putting it in each test is better?

Minion3665:
@pedropintosilva cypress is passing (although iOS and freebsd still fail I don't think that's my fault)

pedropintosilva:
I think we can merge it, but let's wait for @Rash419 to review it :)

Rash419 (approved):
Thanks
Looks good : )

PR #5191, opened 2022-08-26
Private/mmeeks/setclip

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Looks good. If we could, best to avoid default-value args. Thanks!

Issue #4920, opened 2022-06-16
Improvements for following another person's cursor.


gokaysatir:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva @eszkadev 

### Detailed description and rationale
- [ ] The "follow cursor" feature should stop following other user's cursor when the current user types something.
- [x] There should be an indicator to show which user's cursor is being followed. 

### Code pointers
Here is the function which setups the "following" of a user:
https://github.com/CollaboraOnline/online/blob/master/browser/src/map/Map.js#L1803

### Links to related documentation
...

### Additional notes, tips & tricks
...

Rimuruweeb:
sir, can please provide code pointers ? and can you guide me ?.

eszkadev:
@Rimuruweeb hi, here is function which setups the "following" of a user:
https://github.com/CollaboraOnline/online/blob/master/browser/src/map/Map.js#L1803

pedropintosilva:
> There should be an indicator to show which user's cursor is being followed. 
this was done by me and already pushed

pedropintosilva:
I think @Rimuruweeb is not work on this so, I'm assigning this to you @Minion3665 if that's ok :)

pedropintosilva:
@GulsahKose is on this now, thanks !

PR #5195, opened 2022-08-30
Stop follow user feature if a key has been pressed by the original user.

GulsahKose:
Change-Id: Ibfd160fd8fb80a9dfd4c71dfb9a4a2725be48465

* Resolves: #4920
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
From code perspective looks ok

pedropintosilva (approved):
Awesome! I have tested with multiple users, with and without integrator. It works, thanks :)

PR #5196, opened 2022-08-30
private/pedro/cleanup buttons

pedropintosilva:
- Reduce CSS selectors targeting buttons
- Don't use vex btn default btn CSS classes (rename) 
- Remove master slides sidebar's btn CSS declarations
- Add mobile-wizard-titlebar-btn and remove hard coded rules
- Fix button order on newAnnotationVex
- Don't load btn hover status CSS rules when on mobile

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -24,8 +24,8 @@ button.jsdialog img {
 	flex-wrap: wrap;
 }
 
-[class*='button-secondary'],
-.vex.vex-theme-plain .vex-dialog-button.vex-dialog-button-secondary {
+.button-secondary,
+button:not(.ui-corner-all):not(.button-primary) {
 	box-sizing: border-box;
 	height: 32px;
 	line-height: 0em;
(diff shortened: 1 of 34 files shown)


eszkadev (commented):
great! reduction of complex css rules is waht we need :)

I've only noticed one issue on mobile:
act_find_repl_mob

after I click active button (it is selected then?) in mobile find and replace dialog - it becomes white with white text

eszkadev:
and cypress is looking for old classes: `AssertionError: Timed out retrying after 6000ms: Expected to find element: `.vex-dialog-form .vex-dialog-button-primary`, but never found it. Queried from element: <body.vex-open>`

pedropintosilva:
> great! reduction of complex css rules is waht we need :)
> 
> I've only noticed one issue on mobile:

Thanks, and good catch I could only see that in chrome which led me to do a proper more generic fix for that (don;t use hover styles when not on desktop)

pedropintosilva:
> and cypress is looking for old classes: `AssertionError: Timed out retrying after 6000ms: Expected to find element: `.vex-dialog-form .vex-dialog-button-primary`, but never found it. Queried from element: <body.vex-open>`

Yup I have rebase it and add the needed adjustments.

pedropintosilva:
**NOT READY TO BE MERGED:** I need to test:
* Cypress locally
* Welcome Dialog, feature dialog, conflict dialog on desktop and mobile

pedropintosilva:
Merging as this was approved already

PR #5201, opened 2022-09-01
wsd: Cache result of disk space check, too

thebearon:
...when it's requested.

Previously always an empty string was returned on subsequent
calls, and the error message was only shown once.

Change-Id: I7645c7770e3e1712c8ef1b3ad145daec14530fb0

* Target version: master 

### Checklist


Ashod (approved):
Looks reasonable. Thanks @thebearon.

thebearon:
The failure in the iOS chain is unrelated.

PR #5205, opened 2022-09-05
Split config.h into separate headers to avoid extensive rebuild

thebearon:
Eg. if only the version was changed, config.h was updated, and
everything was recompiled.

New structure, maintain these manually:
config.h.in - for configured variables that are stable if
              configuration parameters are unchanged
config_version.h.in - for version/hash related variables

----
config_unused.h.in - still generated by autoheader with all
                     variables, don't use it

Change-Id: Id9a50a9f1e798a3b3814778d8683b7d7cb57bb29

* Target version: master 

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,9 @@ coolwsd.xml
 aclocal.m4
 autom4te.cache
 config.h
-config.h.in
+config_unused.h
+config_unused.h.in
+config_version.h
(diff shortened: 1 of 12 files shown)


mmeeks:
Oh - bother =) sorry - got too used to CI giving broken answers. Let me look.

PR #5208, opened 2022-09-06
Calc: limit scrollbar to data area

eszkadev:
When we moved our view to the end of spreadsheet - scrollbar works with large range.
This PR will reduce the range again to small value (only area containing data) when we enter again data area.

Requires core changes:
https://gerrit.libreoffice.org/c/core/+/139472
https://gerrit.libreoffice.org/c/core/+/139473


eszkadev:
Still WIP

eszkadev:
To test:
1 open spreadsheet
2 go to last row: ctrl + arrow down
3 move scrollbar to lets say 1/3 of its range and type some data

result: we have 2 scrolbar modes, above data we limit scrollbar to the last used row/column. when we reach the end of data - we unlock it and allow to scroll far away.

pedropintosilva:
Tested, it feels great. Thanks!

dennisfrancis (approved):
Looks good! Thank you.

PR #5217, opened 2022-09-07
formulabar: fix selection in multi-line cell

eszkadev:
indexOf returns position - do not sum the positions This fixes not working cursor position change in
formulabar when we had multiple lines.


Rash419 (approved):
Looks good, Thanks !

PR #5218, opened 2022-09-07
docker: option to add your own CA certificate

timar:
Change-Id: Iace0b0f2c99aeaecc50214a1559a1f51410c803e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/docker/from-packages/scripts/start-collabora-online.sh
+++ b/docker/from-packages/scripts/start-collabora-online.sh
@@ -61,5 +61,9 @@ fi
 # Generate WOPI proof key
 coolwsd-generate-proof-key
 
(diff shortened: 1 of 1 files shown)


genofire:
Yes i thought that this works.
But sadly this task need root permission, which is not given:
https://github.com/CollaboraOnline/online/blob/d2f06e078c29cf492f6625ca6a5fae1531438d45/docker/from-source/Debian#L42

It should not work, and that is okay, out of security look.

I do not know an alternative solution ATM (beside to build a new layer, like we do ATM).

timar:
Oh yes, this was an oversight on my side. Startup script runs unprivileged. So, as you said, it's easier to have the workaround with +1 layer, when necessary.

PR #5221, opened 2022-09-07
Attempt to solve FreeBSD issue.

gokaysatir:
Change-Id: I8d3e85ab0b1d590a130569dd74d3c45cade708f5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -45,7 +45,6 @@
   ],
   "license": "BSD-2-Clause",
   "dependencies": {
-    "canvas": "^2.6.1",
     "jsdom": "^16.4.0",
     "tmp": "^0.2.1"
   },


timar:
Don't we use canvas module in our code? Do you remember why it was introduced? If we don't use it, then it's not a FreeBSD fix, is it?

gokaysatir:
> Don't we use canvas module in our code? Do you remember why it was introduced? If we don't use it, then it's not a FreeBSD fix, is it?

Yes the change was made by @mmeeks for performance tests. I checked it with him. I couldn't find any places that the canvas package is used.

If i was wrong and it was being used for perf tests, then we will install the package onto test machines, globally with "npm install -g canvas".

My guess is: That package has some dependencies and one or some of those dependencies is/are the root cause of the issue.

I don't know if it is technically not a fix but i am suggesting to remove a package that is causing an issue possibly because of a dependency. Furthermore, we don't seem to need it :)

Thanks for the comment.

timar:
OK, it's even better, if we remove an unneeded dependency, that caused issues on various platforms.

timar (approved):

PR #5225, opened 2022-09-08
add 'Connection: close' header to http responses.

gmasei11:
If the connection is closed right after the response was sent then it's wise to add 'Connection: close' header to the response so that the client optimizes its behaviour: e.g. does not reuse the socket for further http requests. Normally  a client should retry a request if the reuse of an old socket fails and that should solve the problem but still this is an overhead.

Change-Id: I29f1498610c567024def3beb1ad7014f2c15a232

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (commented):
Thanks for the patch!

As discussed offline, let's split out the `headerOnly` bit, as it's unrelated (and happens to be false anyway).

Ashod (approved):
Let's get it in and improve the `headerOnly` as a follow up.

PR #5233, opened 2022-09-09
private/pedro/tabbedview improve toolbar tabs

pedropintosilva:
- Remove notebookbar tab indicators on tablet
- Add margin to tabbed view: tabs
- Tabbed view: center document name in the toolbar
- Tabbed view: Add "Tap to collapse" tooltip

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -974,7 +974,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		};
 	},
 
-	_tabsControlHandler: function(parentContainer, data, builder) {
+	_tabsControlHandler: function(parentContainer, data, builder, tabTooltip) {
+		if (tabTooltip === undefined) {
+			tabTooltip = '';
+		}
 		if (data.tabs) {
 			var tabs = 0;
 			for (var tabIdx = 0; data.children && tabIdx < data.children.length; tabIdx++) {
@@ -1003,6 +1006,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				var isSelectedTab = data.selected == item.id;
 				if (isSelectedTab) {
 					$(tab).addClass('selected');
+					tab.title = tabTooltip;
 					singleTabId = tabIdx;
 				}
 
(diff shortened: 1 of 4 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1003,6 +1003,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				var isSelectedTab = data.selected == item.id;
 				if (isSelectedTab) {
 					$(tab).addClass('selected');
+					tab.title = _('Tap to collapse');
```
eszkadev, line 1006: this will be added not only in notebookbar but in all the dialogs...

it should be added to the JSON in https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.NotebookbarBuilder.js#L377
I would add that to every item as 'tooltip'
{
    ...
    tabs: [ ... { id: ..., number:... ,   tooltip: _('Tap to collapse'), ... }, ... ],
    ...
}
and here we just use that value

eszkadev (changes requested):

browser/src/control/Control.JSDialogBuilder.js
eszkadev: also we need to change `_createTabClick` function above so we will get the tooltip not only for the tab which initially was selected, but after we activate other tab also

eszkadev (changes requested):

browser/src/control/Control.JSDialogBuilder.js
pedropintosilva: Will do

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -376,7 +382,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 
 	_overriddenTabsControlHandler: function(parentContainer, data, builder) {
 		data.tabs = builder.wizard.getTabs();
-		return builder._tabsControlHandler(parentContainer, data, builder);
+		return builder._tabsControlHandler(parentContainer, data, builder, _('Tab to collapse'));
```
eszkadev, line 385: Here is "Tab" in other places "Tap" :)

pedropintosilva:
Szymon already reviewed it before reasing it, thanks

PR #5239, opened 2022-09-12
wsd: simplify infobar fetching

Ashod:
Since the request is really synchronous,
there is no benefit in firing up a thread
and doing an asynchronous request when
we will block until the result is in.

Luckily, we have a much simpler sync-
request API that avoids threads and
callbacks.

Also, avoid failure logs when no
infobar URL is provided.

Change-Id: I06a2f7fb042c8e8d5599bb06ecee5b1afa51911a

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -5105,33 +5105,34 @@ void COOLWSD::processFetchUpdate()
 {
     try
     {
-        Poco::URI uriFetch(std::string(INFOBAR_URL));
+        const std::string url(INFOBAR_URL);
+        if (url.empty())
+            return; // No url, nothing to do.
+
+        Poco::URI uriFetch(url);
         uriFetch.addQueryParameter("product", APP_NAME);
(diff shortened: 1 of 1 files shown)


wsd/COOLWSD.cpp
```diff
@@ -5105,33 +5105,34 @@ void COOLWSD::processFetchUpdate()
 {
     try
     {
-        Poco::URI uriFetch(std::string(INFOBAR_URL));
+        const std::string url(INFOBAR_URL);
```
hcvcastro, line 5108: I think Poco exception is thrown if uriFetch is invalid

hcvcastro:
It looks good,  in fact, it was my first attempt to do it, but it will block timeout milliseconds
mainWait.poll() (housekeeping,prisoner wakeup, brokers checks) and I did not want any risk to alter (side effects)
those stable functionalities, anyway the thread will span every 8 hours. if tested well, then let's do it :)

hcvcastro (approved):

gokaysatir:
Thanks for the improvements & review :)

PR #5240, opened 2022-09-12
wsd: improve SSL error reporting

Ashod:
ERR_get_error cleared the returned error,
which meant the subsequent call to print
the error in string form, using
ERR_print_errors_cb, didn't find anything
to report.

Change-Id: If131a8cc0d2c1d8bbf705ed38f144b38abf6c8c6


mmeeks (approved):
Seems plausible; we do either clear the error or check return values so we don't need to elsewhere I hope =)

Ashod:
> Seems plausible; we do either clear the error or check return values so we don't need to elsewhere I hope =)

Correct. In the path touched we always call `ERR_print_errors_cb()` which empties the error queue.

PR #5244, opened 2022-09-12
 Save different sidebar states for Impress. (Szymon's version)

eszkadev:
works with core fix: https://gerrit.libreoffice.org/c/core/+/139729

 sidebar: show on demand

Previously sidebar was always active on launch.
This patch together with core fix changes that so
initially it is off and we don't waste resources
for rendering sidebar which is hidden just after short
time.

We don't know the state of sidebar (which deck is activated)
on the core side so in some cases we need to switch decks
in 2 steps: send SidebarShow, and when we received result
we change deck to other one or we do nothing. This allows us
to switch correctly the decks.

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1822,12 +1822,6 @@ L.Control.Menubar = L.Control.extend({
 			return;
 		}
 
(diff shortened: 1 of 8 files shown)


gokaysatir (approved):
Great improvements, thanks :)

PR #5246, opened 2022-09-13
cypress: use class instead of id in impress/apply_paragraph_props_tex…

eszkadev:
followup so core fix will pass the CI
don't use generated ids


Rash419 (approved):
Looks good : )

PR #5249, opened 2022-09-13
cypress: fix mobile repair test

eszkadev:
rollbackPastChange function already opens the menu so don't do it twice

regression from: https://github.com/CollaboraOnline/online/commit/fd0c49a0be757c047674164a941914d15c18fd17


gokaysatir (approved):
Thanks for the fix @eszkadev :)

PR #5256, opened 2022-09-15
jsdialog: simplify structure

eszkadev:
This is PR which introduces CSS grid to the jsdialogs instead of old table design.
It should be polished before merge.

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -33,7 +33,6 @@
 	height: 28px;
 	padding-inline-start: 4px;
 	position: relative;
-	margin-inline-start: 8px;
 	color: var(--color-text-dark);
 	background-color: var(--color-background-dark);
 }
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -10,6 +10,10 @@
 	vertical-align: middle;
 }
 
+.ui-expander-content.sidebar.expanded > .sidebar .ui-grid-cell {
+	justify-content: space-between;
+}
+
 #document-container:not(.mobile) + #sidebar-dock-wrapper {
 	width: 330px;
 	padding: 0;
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -68,6 +68,10 @@
 	margin-inline-start: -14px;
 }
 
+#ParaPropertyPanel.sidebar #box3 {
+	margin-inline-end: 64px;
+}
+
 #fontnamecombobox > .sidebar.jsdialog.ui-listbox,
 #applystyle > .sidebar.jsdialog.ui-listbox {
 	min-width: 186px;
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -13,6 +13,7 @@
 .ui-expander-content.sidebar.expanded > .sidebar .ui-grid-cell {
 	justify-content: space-between;
 	padding: 2px 0;
+	align-items: center;
 }
 
 #document-container:not(.mobile) + #sidebar-dock-wrapper {
@@ -137,6 +138,11 @@
 	position: relative;
 	display: flex;
 	justify-content: space-between;
+}
+
+#textorientbox.sidebar button.ui-pushbutton {
+	min-width: 32px;
+	height: 32px;
 }
 
 #table-box4.sidebar.jsdialog.vertical {
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -275,10 +275,13 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	width: 100%;
 }
 
-.sidebar#ChartTypePanel #grid6 {
-	display: block;
+.sidebar#ChartTypePanel #stack:not(.hidden) + #grid6 {
+	display: block !important;
 }
 
+.sidebar#ChartTypePanel #stack.hidden + #grid6 {
+	display: none;
+}
 
 /* not available commands*/
 
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -64,16 +64,16 @@
 	width: 100% !important; /* frame inside frame */
 }
 
-.sidebar.ui-expander-content {
-	padding-inline: 10px;
+#AlignmentPropertyPanel.sidebar.ui-grid #box3,
+#ParaPropertyPanel.sidebar.ui-grid #box1.sidebar.ui-grid-cell,
+#NumberFormatPropertyPanel.sidebar.ui-grid #grid1 {
+	justify-content: space-between !important;
 }
 
 /* section content */
 
-#AlignmentPropertyPanel.sidebar.ui-grid #box3,
-#ParaPropertyPanel.sidebar.ui-grid #box1.sidebar.ui-grid-cell,
-#NumberFormatPropertyPanel.sidebar.ui-grid #grid1,
-.sidebar.ui-grid {
+.sidebar.ui-expander-content {
+	padding-inline: 10px;
 	width: 300px;
 	row-gap: 4px;
 	column-gap: 4px;
@@ -124,7 +124,9 @@ div.sidebar.ui-grid .checkbutton.sidebar,
 
 .sidebar.ui-grid.ui-grid-cell .menubutton {
 	justify-content: space-between;
+	justify-self: end;
 	padding: 4px;
+	margin-right: 1px;
 }
 
 /* widget width */
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -12,6 +12,7 @@
 
 .ui-expander-content.sidebar.expanded > .sidebar .ui-grid-cell {
 	justify-content: space-between;
+	padding: 2px 0;
 }
 
 #document-container:not(.mobile) + #sidebar-dock-wrapper {
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -202,6 +202,11 @@ div#thousandseparator.checkbutton.jsdialog.sidebar {
 	justify-content: space-between;
 }
 
+#CellAppearancePropertyPanel.sidebar div {
+	grid-template-columns: repeat(4, auto) !important;
+	justify-content: space-between;
+}
+
 .sidebar-container > .root-container.jsdialog.sidebar {
 	margin: 0;
 }
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -306,11 +306,6 @@ td.jsdialog .jsdialog.cell.sidebar {
 	bottom: 15px;
 }
 
-.sidebar #ratio {
-	position: relative;
-	left: 163px;
-}
-
 button#button2.ui-pushbutton.jsdialog.sidebar {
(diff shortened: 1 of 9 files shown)


Andreas-Kainz:
### Align widgets left and right with space between

| before | after |
| -------- | ------ |
| Screenshot_20220927_232818 | Screenshot_20220927_232739 |

Andreas-Kainz:
### Add padding to grid-cell

`padding: 2px;` was added at notebookbar implementation to `cell` so I also add this 2px at the sidebar grid-cell class.

| before | after |
| -------- | ------ |
| Screenshot_20220927_233802 | Screenshot_20220927_233747 |

Andreas-Kainz:
### Paragraph section alignment

The spinfields ans margins need some alignment value.

| before | after |
| -------- | ------ |
| Screenshot_20220927_235242 | Screenshot_20220927_235227 |

Andreas-Kainz:
### Cell Appearance at calc sidebar appearance

| before | after |
| -------- | ------ |
| Screenshot_20220928_000447 | Screenshot_20220928_000454

Andreas-Kainz:
### Alignment smaller buttons at calc sidebar

| before | after |
| -------- | ------ |
| Screenshot_20220928_003258 | Screenshot_20220928_003316 |

@pedropintosilva do you have an idea how to align the spin boxes and wrap text to the right?

browser/css/jssidebar.css
```diff
@@ -10,6 +10,12 @@
 	vertical-align: middle;
 }
 
+.ui-expander-content.sidebar.expanded > .sidebar .ui-grid-cell {
```
eszkadev, line 13: won't be that rule needed not only in expander in sidebar but also in any other grid-cell ?

eszkadev: and in all jsdialogs not only in sidebar?

browser/css/jssidebar.css
```diff
@@ -143,6 +153,11 @@
 	justify-content: space-between;
 }
 
+#textorientbox.sidebar button.ui-pushbutton {
```
eszkadev, line 156: I have proposal to add these "custom" rules for specific objects at the end of file in one section like:
/* CUSTOMIZED WIDGETS */
so it will be easy to find non-generic rules.

eszkadev:
@Andreas-Kainz looks awesome :)

browser/css/jssidebar.css
Andreas-Kainz: as we are editing jssidebar.css from my point of view there should be ONLY sidebar related css rules. If we want them to have also for jsdialogs, than I would move them to jsdialogs. Which would mean more testing, cause it has to work also at dialgoues.

Andreas-Kainz:
As I understood at the end of the day EVERYTHING in the sidebar (content area) should be within a `grid-cell `. So that we have to define rules only for that element. This `grid-cell` class is missing for the context related sidebar stuff. I will have a look at that.

However `grid-cell` is fine cause also the notebookbar has also the cell class. My plan is to simplify the css classes maybe to harmonize them with the notebookbar (table) implementation so the structure is similar. Than I will also add some var's for `margin` and `padding` spaces which mean, that we can easy switch between finger/mouse layout.

However first sidebar stuff than harmonization with dialogues and than var's. I want to leave you @eszkadev the message, that I'm interested in support you at this PR.

Andreas-Kainz:
@eszkadev as everything has `grid-cell` class. Can we make make a `grid-row` class for each line?

css classes

Now everythins is a `grid-cell` like the grid-row from the image is also a `grid-cell` so it's difficult to define specific rules.

eszkadev:
> @eszkadev as everything has `grid-cell` class. Can we make make a `grid-row` class for each line?
> Now everythins is a `grid-cell` like the grid-row from the image is also a `grid-cell` so it's difficult to define specific rules.

This is what we wanted to do: use CSS grid which automatically does the layouting. We don't want very deep structure. We just removed rows in this PR :)

eszkadev:
> However `grid-cell` is fine cause also the notebookbar has also the cell class. My plan is to simplify the css classes maybe to harmonize them with the notebookbar (table) implementation so the structure is similar. Than I will also add some var's for `margin` and `padding` spaces which mean, that we can easy switch between finger/mouse layout.

please do not touch notebookbar in this PR. this change is applied only to sidebar and dialogs now. later we reuse that for notebookbar also

eszkadev:
It would be nice to have unified structure based on one grid in every panel:
Here example, Line panel has 3 objects inside, box and 2 grids, alignment in that case is different without setting fixed width...

alignment

Andreas-Kainz:
that's the reason for https://bugs.documentfoundation.org/show_bug.cgi?id=148017 to fix this. On the other hand it take ordinary some time to bring the changes from LibO (core) to COOL. However I will start to harmonize the sidebar .ui files. 

Maybe I can add you for the reviews at gerrit and I would like to fix also https://bugs.documentfoundation.org/show_bug.cgi?id=127771 but there I need some help. 

So that we have an standardizised layout at LibO and COOL.

Andreas-Kainz:
However do you want to wait for the changes at LibO sidebar or make the fix at LibO and Cool and hope that when the change at LibO was available at Cool it will also work there.

eszkadev:
Port of sidebar simplification for core from @Andreas-Kainz to 22.05: https://gerrit.libreoffice.org/c/core/+/141380

eszkadev:
> Port of sidebar simplification for core from @Andreas-Kainz to 22.05: https://gerrit.libreoffice.org/c/core/+/141380

Each element goes into the next cell in the grid, I've noticed panels where label should have span=2 because our code assigns only one slot:
effects

Andreas-Kainz:
So can I use a prebuild LibO from wget https://github.com/CollaboraOnline/online/releases/download/for-code-assets/core-co-22.05-assets.tar.gz and check this?

eszkadev:
> So can I use a prebuild LibO from https://github.com/CollaboraOnline/online/releases/download/for-code-assets/core-co-22.05-assets.tar.gz and check this?

No, I just used this PR and mentioned gerrit patch - locally built. I have some idea how we can automatically generate missing placeholder elements for above issues - have to check. Maybe we don't have to review all the .ui with that :)

eszkadev:
> effects2

Above issues are fixed, apart of writer's position and size panel when selected the shape. It's some grid in a grid problem which should be probably fixed in the .ui. TODO: after this rework is accepted.

eszkadev:
> > effects2
> 
> Above issues are fixed, apart of writer's position and size panel when selected the shape. It's some grid in a grid problem which should be probably fixed in the .ui. TODO: after this rework is accepted.

fixed in last commit here

Andreas-Kainz:
As the 1st column use `justify-content: start;` and the 2nd (or last) column use `justify-content: end;` the simplest solution is use the col number as css class.

rect10828

I also had a very simple jssidebar.css file, no direct formatting is needed, but I want to discuss first an issue.

#### Issue
Some sidebar widgets didn't has a ui-grid-cell css class and so the widgets also didn't get a ui-grid-col- class. The missing widgets are:
- menubutton
- ui-pushbutton
- ui-listbox-container
- checkbutton
- toolbox
- spinfieldcontainer

can this widgets get an `ui-grid-col-n` class?

eszkadev:
> 

@Andreas-Kainz you can do that just using CSS (what is better approach because take into account also cell visibility state):
see: https://stackoverflow.com/a/56315469

eszkadev:
I don't like approach from  patch:
[Instead of grid-cell each cell item use grid-col-n so it's easy to se…](https://github.com/CollaboraOnline/online/pull/5256/commits/774f08cfc3faa62ed88fb777fc7c7cc1ccdc6d3f)

What if we will dynamically hide one of the items? It will be broken. Please delegate this work to the browser and CSS language :)

Andreas-Kainz:
I know that you can do this with css, but at the .js part you define that the items will be at collumn-n. So if a (hidden) widget is not at the correct column, it has to be fixed at the .js part and the .css part define the widgets width, alignment, ...

I can do the column/row location of widgets, but if you change something at the .js part like a widgets wan't be hidden any more, you have to change the .js part AND the .css part. If we have the column number at the .js part, the .css part is way simplier AND update stuff is super easy.

eszkadev:
Problem is that we can at some point dynamically remove one of the elements, or do other modifications -> then numbers will NOT be correct... And browser & CSS always have good results. Also it means - less code. Code is only in CSS. No need to change both JS and CSS.

Also one note here: we should always use "visibility: hidden" in grids instead of "display: none". First one keeps element in the correct place. Second one removes node completely from the calculations.

eszkadev:
I've tested without last patch but with merged core patches (2 x simplified .ui files) and it looks great. I don't see any not working widgets. Everything seems to be properly aligned. First column to the left, second to the end. But @Andreas-Kainz  do you want to have content of a second column be aligned to the right? So this screenshot is wrong in your opinion:
aligned after patches

Asking to be sure I know what do you want to get here :)

Andreas-Kainz:
No change at .js file from me. only changes at jssidebar.css file.

Andreas-Kainz:
| writer | calc | impress |
| -------- | ------ | ------------ |
|writer| calc|impress |

Andreas-Kainz:
all .css classes are for .sidebar so no change anywhere else.

eszkadev:
@Andreas-Kainz thank you, now I see what you aligned to the right :)
PR is quite old so I rebased. Unfortunately everything was broken after that so I added some fixes.

eszkadev:
impress
noticed one more bug in .ui probably

Andreas-Kainz:
Should it be fixed at the .css part or in the .js part?

There is one issue if you are in sidebar charts and select XY chart type.

eszkadev:
> 

I think we need to fix .ui file for Impress

Andreas-Kainz:
Ok will check on weekend. In general I like the simple structure from now. Also the .css file is way simplier.

Andreas-Kainz:
The layout isn't related to the .ui file. it's related to the bullet group which is sometimes visible, sometimes not. The solution would be to add span = 2 when the bullet items are visible.

| first select | after select a text item |
| --------------- | --------------------------------- |
| Screenshot_20221126_083027 | Screenshot_20221126_083054 |

Andreas-Kainz:
Here is the second issue

| charts line type | charts XY (Scatter) |
| ---------------------- | --------------------------- |
| Screenshot_20221126_085242 | Screenshot_20221126_085216 |

If I set
```
.sidebar#ChartTypePanel #grid6 {
   display: none;
}
```
I have issues at some other chart type selections.

eszkadev:
> > 
> 
> I think we need to fix .ui file for Impress
> 
> > Here is the second issue
> > charts line type 	charts XY (Scatter)
> > Screenshot_20221126_085242 	Screenshot_20221126_085216
> > If I set
> > ```
> > .sidebar#ChartTypePanel #grid6 {
> >    display: none;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I have issues at some other chart type selections.
> 
> Good catch. CSS grid requires `visibility: hidden` instead of `display: none` - this should help

I fixed that issue.

PR #5262, opened 2022-09-16
Mobile: hide non-working context menus, and fix missing icon

thebearon:
* Target version: master 

### Checklist

- [] I have run tests with `make check`


pedropintosilva (approved):
Great catch! Thanks!

PR #5274, opened 2022-09-20
keyboard: block shortcuts in readonly mode

lpranam:
Change-Id: Ife814304e397620cf55b57b2025143eaab9a3280

* Target version: master 

### Checklist


browser/src/map/handler/Map.Keyboard.js
```diff
@@ -646,5 +646,14 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 		return false;
+	},
+
+	readOnlyAllowedShortcuts: function(e) {
+		if (this._isCtrlKey(e) && e.shiftKey && e.key === '?')
```
eszkadev, line 652: we should add comment what it is: "Open keyboard shortcuts help page"

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -646,5 +646,14 @@ L.Map.Keyboard = L.Handler.extend({
...
+			return true;
+		else if (e.type === 'keydown' && !e.shiftKey && !e.ctrlKey && !e.altKey && !e.metaKey && e.keyCode === 112)
```
eszkadev, line 654: we have constants for keyCodes now: https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.Keyboard.js#L220

Also it would be good to add a comment what this condition detects :)
Is this for help? (F1 without special keys like ctrl)

eszkadev:
But in general works as expected

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -332,7 +332,7 @@ L.Map.Keyboard = L.Handler.extend({
 	// any 'beforeinput', 'keypress' and 'input' events that would add
 	// printable characters. Those are handled by TextInput.js.
 	_onKeyDown: function (ev) {
-		if (this._map.uiManager.isUIBlocked())
+		if (this._map.uiManager.isUIBlocked() || (this._map.isPermissionReadOnly() && !this.readOnlyAllowedShortcuts(ev)))
```
eszkadev, line 335: ctrl + alt + c (insert comment) doesn't work in pdf (readonly with commenting)

gokaysatir (approved):
Thanks for the improvements and the review :)

PR #5279, opened 2022-09-20
masterpage: followup

eszkadev:
helps with:
- activation of a closed sidebar on master page activation
- parses correctly message with `mode` param, should be separated with `,` not a space

this is required for next set of patches which will activate the `mode` param


gokaysatir (approved):

PR #5280, opened 2022-09-20
fix: calc: comments in frozen row headers aren't shown after scrolling

Rash419:
Change-Id: I6eeb716c118778c1a386437c751bda3e9e3a5a24

* Target version: master


gokaysatir (approved):

PR #5284, opened 2022-09-21
Make the wrapper's display "fixed" so it doesn't interfere with the w…

gokaysatir:
…indow.

Change-Id: I7176a0dbc235f8202cbbd4028ac187293d648d8a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/welcome.css
+++ b/browser/css/welcome.css
@@ -1,6 +1,6 @@
 .iframe-welcome-wrap,
 .iframe-welcome-content {
-	position: absolute;
+	position: fixed;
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Thanks @gokaysatir ! I have tested. Approved.

Still I didn't quite understand what was the problem by just reading the commit message. Nevertheless, I see no problem with this PR :) and I see how the containing block could be different when using absolute and if we would have scrolling.

Issue #5164, opened 2022-08-17
ERR  Poco Exception: File not found: /opt/cool/child-roots/kKwSDrf3cR8mcwCG//delete.me (Starting coolwsd in unprivileged lxc container fails)


brknkfr:
**Describe the bug**
I installed the most recent `coolwsd` (version 22.05.5.3-1) on an unprivileged lxc container (debian 11 bullseye on amd64). It seems that coolwsd isn't starting up properly. It is falling into a strange `prisoner_poll` loop and a connection to port 9980 is not possible (curl -k http://127.0.0.1:9980 times out). The bug is probably related to https://github.com/CollaboraOnline/online/issues/5155.

**To Reproduce**
1. Install code-brand as [indicated](https://www.collaboraoffice.com/code/linux-packages/) into an unprivileged container
2. Disable SSL in config file.
3. Start coolwsd with systemctl start coolwsd.
4. See error. I guess the problems lies somewhere here in following log entry: 

```
kit-03320-03318 2022-08-17 17:17:37.561993 +0200 [ kit_spare_001 ] ERR  Poco Exception: File not found: /opt/cool/child-roots/kKwSDrf3cR8mcwCG//delete.me| kit/Kit.cpp:2968
kit-03320-03318 2022-08-17 17:17:37.562037 +0200 [ kit_spare_001 ] INF  Kit process for Jail [kKwSDrf3cR8mcwCG] finished.| kit/Kit.cpp:2978
Forced Exit with code: 0
kit-03320-03318 2022-08-17 17:17:37.562065 +0200 [ kit_spare_001 ] FTL  Forced Exit with code: 0| common/Util.cpp:1096
frk-03318-03318 2022-08-17 17:17:38.555792 +0200 [ forkit ] WRN  No live Kits exist, and we are not terminating yet.| kit/ForKit.cpp:304
```
After that, it repeats following message:

```
wsd-03594-03602 2022-08-17 17:25:48.415089 +0200 [ prisoner_poll ] TRC  Poll completed with 0 live polls max (5000000us)(timedout)| net/Socket.cpp:355
wsd-03594-03602 2022-08-17 17:25:48.415147 +0200 [ prisoner_poll ] TRC  #18: Starting handling poll results of prisoner_poll at index 1 (of 2): 0| net/Socket.cpp:433
[...]

Issue #5155, opened 2022-08-16
upgraded coolwsd on ubuntu 18.04 lts from 21.06 to 22.05 - doesn't start anymore (no docker)


redglobuli:
**Describe the bug**
Everything worked with 21.06, now i upgraded to 22.05 and coolwsd isn't responding on 9980 anymore.

The instance is running on OpenVZ VPS.

curl -k http://127.0.0.1:9980 times out

journalctl shows (i set logging to trace) a loop like this:

```
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521830 +0200 [ forkit ] TRC  ppoll start, timeoutMicroS: 1000000 size 1| net/Socket.cpp:337
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521816 +0200 [ forkit ] TRC  #13: setupPollFds getPollEvents: 0x1| net/Socket.hpp:859
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521800 +0200 [ forkit ] TRC  cleanupChildren reaped 0 children to have 0 left: | kit/ForKit.cpp:325
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521785 +0200 [ forkit ] TRC  cleanupChildren with 0 children| kit/ForKit.cpp:290
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521769 +0200 [ forkit ] TRC  forkLibreOfficeKit limit: 0| kit/ForKit.cpp:436
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521751 +0200 [ forkit ] TRC  #13: Revents: 0x0| net/Socket.hpp:1296
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521733 +0200 [ forkit ] TRC  #13: Starting handling poll results of forkit at index 0 (of 1): 0| net/Socket.cpp:435
Aug 16 09:25:39  coolwsd[6975]: frk-06983-06983 2022-08-16 09:25:39.521647 +0200 [ forkit ] TRC  Poll completed with 0 live polls max (1000000us)(timedout)| net/Socket.cpp:356
[...]

PR #5289, opened 2022-09-23
wsd: disable mounting when unmount fails

Ashod:
It seems that on some systems it is possible
to succeed in mounting, but fail to unmount.

This is proving very problematic and it is
best to detect unmount failures and disable
mount-binding altogether.

Should avoid the issues seen in #5155 and #5164.


mmeeks (approved):
Very strange. I'd like to also fix the fall-through case that we know does not work from this log too.
We can't be copying files into a folder that doesn't exist through the path I identified.

Belt+braces.

mmeeks:
We don't need code-22.05 I think - it's ab obsolete branch =) but thanks Pranam! =)

PR #5290, opened 2022-09-23
annotation: comments: allow comment operation in read only mobile view

lpranam:
Change-Id: I91e631549545696c573ed873126e225947a1ad18

* Target version: master 

### Checklist


eszkadev (changes requested):
I think it allows to add comment in read-only file

eszkadev (commented):
ok, it is only for calling the command... but there should be protection in ui to not show these commands in read-only file

eszkadev (approved):

eszkadev (changes requested):
tested again, it allows to add comment (open the dialog) in read-only file but after modal is closed no comment appears. it should be blocked

lpranam:
closed in favour of #5410

PR #5295, opened 2022-09-23
wsd: re-create the jail directory when mounting fails

Ashod:
This guarantees that the jail directory is always
    created, especially in case mounting fails and
    we cleanup and fallback.
    
    Also, move the directory creation and setting
    the perms into a helper, and reuse.


mmeeks (approved):
Lovely - thanks Ash ! =)

PR #5303, opened 2022-09-27
embedded video: show inside selection marker

eszkadev:
Draft showing example video preview when activating any shape...


eszkadev:
continued in: https://github.com/CollaboraOnline/online/pull/5349

PR #5305, opened 2022-09-28
slidesorter: use correct styles for desktop & selection more visible

eszkadev:
Mark as important so will not be overriden by the mobile-wizard rules.

Use primary color for selection so it will be visible...

Previously we had primary color for current slide (from mobile-wizard) and grey for other selected slides.

slidesorter


pedropintosilva (approved):
Thanks!

PR #5313, opened 2022-10-03
Cleanup remote config

Rash419:
* Resolves: # 
* Target version: master

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -1220,13 +1220,23 @@ public:
 
         try
         {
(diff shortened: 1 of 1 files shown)


genofire:
Could we put it outside of `ENABLE_FEATURE_LOCK` ?

Rash419:
> Could we put it outside of `ENABLE_FEATURE_LOCK` ?

@genofire not really, to use feature locking COOL should be built with --enable-feature-lock flag ENABLE_FEATURE_LOCK is set based on the flag has been passed or not  : )

kendy (approved):
All looks good to me, thank you! :-)

Issue #4454, opened 2022-03-21
Speed-up variable substitution


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @Ashod 

### Detailed description and rationale

When we serve our initial HTML files - we have a reasonably large file loleaflet.html - and before we serve it we do a lot of repeated substitutions in it:

### Code pointers

checkout:

wsd/FileServer.cpp

We do things like:

    Poco::replaceInPlace(preprocess, std::string("%ACCESS_TOKEN%"), escapedAccessToken);
    Poco::replaceInPlace(preprocess, std::string("%ACCESS_TOKEN_TTL%"), std::to_string(tokenTtl));
    Poco::replaceInPlace(preprocess, std::string("%ACCESS_HEADER%"), escapedAccessHeader);

Which continually re-parse and re-allocate a std::string which is far from optimal.

We should instead build an unordered_set of all the variables we want to replace; and then search through the file for a first '%' and then if it has a reasonably short, ALL_CAPS type variable after it - look that variable up and substitute it and if not just continue appending it to a string stream or somesuch to save lots of extra allocations.

### Links to related documentation

We prolly want something like:

std::string s;
s.reserve(existing file-size);

and just s.append(chunk);

as we go along.

Would be nice to have a util method that does this cleanly for us that takes an unordered_set I guess.

Thanks !

Ashod:
I've been mulling over this for a week now (while looking at implementing a feature), not knowing this easy hack existed! (Note to self: check your missed notifications more frequently).

Anyway, my idea is very similar to the description (with a small tweak). Here it goes:

1. While caching files, store templates with `%` variables in a special container that has the contents string pre-tokenized (ideally, this container is StringVector or uses it directly). This is done only once at startup.
2. Construct an `unordered_map` with <variable-name, value> pairs before `preprocessFile` and pass it as input. This is done for each request.
3. In `preprocessFile` simply run over the pre-tokenized file contents and for every non-variable token just append it to the output (which should have reserved size > original file-size, since many substitutions are larger than the variable length), and for each variable token substitute with the value from the `unordered_map`, if it exists, otherwise do what is done right now (for backwards compatibility, unless we know it's save to do something different and more efficient).

mmeeks:
Great =) of course this is not urgent, but would be a great easy hack for a newbie.

ArshErgon:
[...]

PR #5314, opened 2022-10-04
wsd: fileserver.cpp changed variable replacemenct method

NoahbJohnson15:
Change-Id: Ie1a6c00f23a898c4244025f83d78cdd19ff73b9b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/mocha.log
+++ b/browser/mocha.log
@@ -1,0 +1,173 @@
+
+
+  Bounds parse() tests
+    Bounds.parse() call with an empty string argument
+incomplete rectangle
+      ✓ should return undefined
+    Bounds.parse() call with an string argument with 3 numbers
+incomplete rectangle
+      ✓ should return undefined
+    Bounds.parse() call with an string argument with 4 numbers
+      ✓ should return a valid Bounds
+      ✓ and the Bounds should be correct in position and size
+    Bounds constructor call
+      ✓ correctness check with PointConstructable[] argument
+
+  CPointSet empty() tests
+    new CPointSet()
+      ✓ should be empty
+    CPointSet constructed from Points
+      ✓ should be not be empty
+    CPointSet constructed from array of CPointSets
+      ✓ should be not be empty
+    After setting an empty Point array to a non-empty CPointSet
+      ✓ the CPointSet should be empty
+
+  SheetGeometry public methods tests
+    Document zoom level = 10
+      Tests for part#0 - Varying row/column sizes
+        ✓ correctness of getPart()
+        ✓ correctness of getViewColumnRange()
+        ✓ correctness of getViewRowRange()
+        ✓ correctness of getRowData()
+        ✓ correctness of getColumnGroupLevels()
+        ✓ correctness of getRowGroupLevels()
+        ✓ correctness of getColumnGroupsDataInView()
+        ✓ correctness of getRowGroupsDataInView()
+        ✓ correctness of getTileTwipsAtZoom()
+        ✓ correctness of getTileTwipsPointFromPrint()
+        ✓ correctness of getPrintTwipsPointFromTile()
+        ✓ correctness of getTileTwipsSheetAreaFromPrint()
+        ✓ correctness of getSize()
+        ✓ correctness of getCellRect()
+        ✓ correctness of getCellFromPos()
+        ✓ correctness of getSnapDocPosX() and getSnapDocPosY()
+      Tests for part#1 - Hidden rows
+        ✓ correctness of getPart()
+        ✓ correctness of getViewColumnRange()
+        ✓ correctness of getViewRowRange()
+        ✓ correctness of getRowData()
+        ✓ correctness of getColumnGroupLevels()
+        ✓ correctness of getRowGroupLevels()
+        ✓ correctness of getColumnGroupsDataInView()
+        ✓ correctness of getRowGroupsDataInView()
+        ✓ correctness of getTileTwipsAtZoom()
+        ✓ correctness of getTileTwipsPointFromPrint()
+        ✓ correctness of getPrintTwipsPointFromTile()
+        ✓ correctness of getTileTwipsSheetAreaFromPrint()
+        ✓ correctness of getSize()
+        ✓ correctness of getCellRect()
+        ✓ correctness of getCellFromPos()
+        ✓ correctness of getSnapDocPosX() and getSnapDocPosY()
+      Tests for part#2 - Filtered rows
+        ✓ correctness of getPart()
+        ✓ correctness of getViewColumnRange()
+        ✓ correctness of getViewRowRange()
+        ✓ correctness of getRowData()
+        ✓ correctness of getColumnGroupLevels()
+        ✓ correctness of getRowGroupLevels()
+        ✓ correctness of getColumnGroupsDataInView()
+        ✓ correctness of getRowGroupsDataInView()
+        ✓ correctness of getTileTwipsAtZoom()
+        ✓ correctness of getTileTwipsPointFromPrint()
+        ✓ correctness of getPrintTwipsPointFromTile()
+        ✓ correctness of getTileTwipsSheetAreaFromPrint()
+        ✓ correctness of getSize()
+        ✓ correctness of getCellRect()
+        ✓ correctness of getCellFromPos()
+        ✓ correctness of getSnapDocPosX() and getSnapDocPosY()
+      Tests for part#3 - Row groups
+        ✓ correctness of getPart()
+        ✓ correctness of getViewColumnRange()
+        ✓ correctness of getViewRowRange()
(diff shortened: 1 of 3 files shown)


pedropintosilva:
Hello @NoahbJohnson15 and welcome!

If I'm not mistaken this is related to https://github.com/CollaboraOnline/online/issues/4454 right? Let me assign @Ashod  as reviewer

pedropintosilva (changes requested):
Could you please amend the commit and insert a commit subject (title) and commit message (description)?

* `git commit --ammend`
* And add 3 lines on top of `Signed-off-by: NoahJohnson <noahbjohnson18@gmail.com>`
  * `wsd: fileserver.cpp changed variable replacement method`
  * Empty line
  * `This speeds-up variable substitution #4454`

note: I also see a new added log file, I wonder if that's ok

browser/mocha.log
```diff
@@ -0,0 +1,173 @@
+
+
+  Bounds parse() tests
+    Bounds.parse() call with an empty string argument
+incomplete rectangle
+      ✓ should return undefined
+    Bounds.parse() call with an string argument with 3 numbers
+incomplete rectangle
+      ✓ should return undefined
```
Ashod, line 9: Since this is unrelated to the fileserver, best to leave it out. If need help with removing this file from your commit, just ask. Happy to help with git commands.

wsd/FileServer.cpp
```diff
@@ -562,7 +562,8 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
             endPoint == "uno-localizations.json" ||
             endPoint == "uno-localizations-override.json")
         {
-            preprocessFile(request, requestDetails, message, socket);
+            std::map<std::string, std::string> replaceVar = {{"ACCESS_TOKEN",""}, {"SOCKET_PROXY", ""}, {"ACCESS_TOKEN_TTL", ""}, {"ACCESS_HEADER", ""}, {"HOST", ""},  {"VERSION", "COOLWSD_VERSION_HASH"}, {"COOLWSD_VERSION", "COOLWSD_VERSION"}, {"SERVICE_ROOT", ""}, {"UI_DEFAULTS", ""}, {"POSTMESSAGE_ORIGIN", ""}, {"PROTOCOL_DEBUG", ""}, {"HEXIFY_URL", ""}, {"BRANDING_CSS", ""}, {"BRANDING_JS", ""},  {"CSS_VARIABLES", ""}, {"DOCUMENT_SIGNING_DIV", ""} , {"DOCUMENT_SIGNING_URL", ""}, {"BROWSER_LOGGING", ""}, {"GROUP_DOWNLOAD_AS", ""},  {"OUT_OF_TIMEOUT_SECS", ""}, {"IDLE_TIMEOUT_SECS", ""}, {"ENABLE_WELCOME_MSG", ""}, {"AUTO_SHOW_WELCOME", ""}, {"UI_RTL_SETTINGS", ""},   {"USE_INTEGRATION_THEME", ""}, {"ENABLE_MACROS_EXECUTION", ""},  {"AUTO_SHOW_FEEDBACK", ""}, {"FEEDBACK_URL", "FEEDBACK_URL"}, {"WELCOME_URL", "WELCOME_URL"}, {"BUYPRODUCT_URL", ""}, {"FRAME_ANCESTORS", ""}, {"USER_INTERFACE_MODE",""}};
```
Ashod, line 565: Since we are setting the map keys and values below, I wonder what advantage is there in having default values that we later override. It seems that we only need default values for the cases that we don't have a value for. So we can simplify this greatly.

This line is excessively long. We use clang-format, which you can also use to automatically format all your code, but you can also manually break this on multiple lines, roughly making each line equal in length (about 80-100 characters long).

This helps readability and maintainability.

wsd/FileServer.cpp
```diff
@@ -872,7 +873,8 @@ namespace
 void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
                                               const RequestDetails &requestDetails,
                                               Poco::MemoryInputStream& message,
-                                              const std::shared_ptr<StreamSocket>& socket)
+                                              const std::shared_ptr<StreamSocket>& socket,
+                                              std::map<std::string, std::string> replaceVar)
```
Ashod, line 877: Passing the argument as reference (`std::map<std::string, std::string>& replaceVar`) is better.

wsd/FileServer.cpp
```diff
@@ -940,29 +942,27 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
     std::string socketProxy = "false";
     if (requestDetails.isProxy())
         socketProxy = "true";
-    Poco::replaceInPlace(preprocess, std::string("%SOCKET_PROXY%"), socketProxy);
+    replaceVar["SOCKET_PROXY"] = socketProxy;
```
Ashod, line 945: Shouldn't we include the `%` too? Why do we drop the percentage marks from all the variables?

wsd/FileServer.cpp
```diff
@@ -984,9 +984,9 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
     }
 #endif
 
-    Poco::replaceInPlace(preprocess, std::string("<!--%BRANDING_CSS%-->"), brandCSS);
-    Poco::replaceInPlace(preprocess, std::string("<!--%BRANDING_JS%-->"), brandJS);
-    Poco::replaceInPlace(preprocess, std::string("<!--%CSS_VARIABLES%-->"), cssVarsToStyle(cssVars));
+    replaceVar["BRANDING_CSS"] = brandCSS;
+    replaceVar["BRANDING_JS"] = brandJS;
+    replaceVar["CSS_VARIABLES"] = cssVarsToStyle(cssVars);
```
Ashod, line 989: Shouldn't we retain the "" as well? We then wouldn't need to special case in the replace loop below.

wsd/FileServer.cpp
```diff
@@ -1116,14 +1114,30 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
...
         LOG_TRC("Denied all frame ancestors");
         cspOss << imgSrc << "; ";
     }
 
+    std::string newpreprocess;
+    newpreprocess.reserve(preprocess.length() + 50);
+    std::stringstream checkVar(preprocess);
+    std::string medium;
+
+     while(getline(checkVar, medium, '$'))
```
Ashod, line 1130: Isn't the variable marker `%` ? Perhaps a comment to explain what this loop is doing is in order.

wsd/FileServer.cpp
```diff
@@ -1116,14 +1114,30 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
...
+     {
+        if(medium.substr(medium.length()-4, medium.length()) == "<!--")
+            medium = medium.substr(0, medium.length()-4);
+        if(medium.substr(0, 3) == "-->")
+            medium = medium.substr(3, medium.length());
```
Ashod, line 1135: Why do we delete the comment markers? Was that done in the old algorithm?

wsd/FileServer.cpp
```diff
@@ -1116,14 +1114,30 @@ void FileServerRequestHandler::preprocessFile(const HTTPRequest& request,
...
     }
 
+    std::string newpreprocess;
+    newpreprocess.reserve(preprocess.length() + 50);
+    std::stringstream checkVar(preprocess);
+    std::string medium;
+
+     while(getline(checkVar, medium, '$'))
+     {
+        if(medium.substr(medium.length()-4, medium.length()) == "<!--")
```
Ashod, line 1132: `substr` is rather expensive, because it creates a copy. We can improve this (perhaps later).

Also, best to follow the formatting of the file, for consistency. Leave space after the `while` and `if`, and add a line break after `else`. You can also use clang-format on the lines you've changed/added.

Ashod (changes requested):
Thanks @NoahbJohnson15 for the patch! 

I left some comments and questions. The formatting and performance issues are minor. The ones related to the functionality are more critical, of course.

In addition, I'm curious how you tested this. It would be best to diff the output of this function before and after your change. They should be identical. Also, it's a good idea to add new test to verify your algorithm.

Other than that, there are test failures. But we'll get to that after you update the patch with the requested changes and we'll see if it still fails or not. So don't worry about that just yet.

Finally, you can use the title from the issue you are fixing in the commit message, which makes things easier for everyone. For example:
```
wsd: speed-up variable substitution in file-server
```

Btw, even though your approach is not what is proposed in #4454, it's fine, as long as it works as expected. We can improve it further later. (There is a detailed description of the algorithm in one of my comments).

Thanks, and feel free to ask any questions you may have!

pedropintosilva:
Hello @NoahbJohnson15 are you still working on this ? Or is there anything I can help with?

PR #5320, opened 2022-10-07
Add percentage sign to cell automatically.

GulsahKose:
Change-Id: I422f63c6d3963bdf125fc00e7dafc4428883c7ce

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/marker/TextInput.js
+++ b/browser/src/layer/marker/TextInput.js
@@ -79,6 +79,21 @@ L.TextInput = L.Layer.extend({
 			'?':  [ 63, 0,      0, 4287 ],
 			'_':  [ 95, 0,      0, 5384 ]
 		};
+
+		// unoKeyCode values of the digits.
+		this._unoKeyMap = {
+			'48': 96,   // 0
+			'49': 97,   // 1
+			'50': 98,   // 2
+			'51': 99,   // 3
+			'52': 100,  // 4
+			'53': 101,  // 5
+			'54': 102,  // 6
+			'55': 103,  // 7
+			'56': 104,  // 8
+			'57': 105   // 9
+		};
+
 	},
 
 	onAdd: function() {
@@ -563,6 +578,12 @@ L.TextInput = L.Layer.extend({
 		this._lastContent = oldContent;
 	},
 
+	_isDigit: function(asciiChar) {
+		if (asciiChar >= 48 && asciiChar <= 57)
+			return true;
+		return false;
+	},
+
 	// Fired when text has been inputed, *during* and after composing/spellchecking
 	_onInput: function(ev) {
 		if (this._map.uiManager.isUIBlocked())
@@ -680,7 +701,15 @@ L.TextInput = L.Layer.extend({
 		this._lastContent = content;
 
 		if (newText.length > 0) {
-			this._sendText(this.codePointsToString(newText));
+			// When the cell formatted as percent, to trig percentage sign addition
+			// automatically we send the first digit character as KeyEvent.
+			if (this._map.getDocType() === 'spreadsheet' &&
+			    content.length === 1 && ev.inputType === 'insertText' &&
+				this._isDigit(newText)) {
+				this._sendKeyEvent(newText, this._unoKeyMap[newText], 'input');
+			}
+			else
+				this._sendText(this.codePointsToString(newText));
 		}
 
 		// was a 'delete' and we need to reset world.


pedropintosilva:
thanks @GulsahKose so, and please correct me if I'm wrong,

* The goal is to fix the missing percentage sign `%` that doesn't appear automatically when the user starts to type into a cell (that was previously formatted as percent `.uno:NumberFormatPercent `)
* Right now, with this commit, we have a new `this._hasAutoPercent` that we are attempting to use to know if `%` was programmatically added (not typed by the user).
  * But  `this._hasAutoPercent`, as of now, never gets back to false. We need to find a place that is called when the cell cursor moves (when the user exits that cell) something like the opposite of the existent `_onBeforeInput`
    * As we saw during our call `_finishFormulabarEditing` seems to get called but only one hitting enter (doesn't get call if the user changes the position of cell cursor via mouse)

eszkadev:
I feel that it's still a bit hacky solution...  it keeps some state we need to be sure we set correctly because it can break content later.
I really think it should be done in the core, not online. Core knows what is happening. Online just sends key codes, doesn't know current cell state etc., also I think we don't want to duplicate formatting logic in both core and online (when core can do it better).

Issue probably is caused because we "type" in online using different method, we use text composition (IME) here which is not used in the desktop application in this case - what triggers different code path.

This is done using events: [CommandEventId](https://opengrok.libreoffice.org/s?refs=CommandEventId&project=core)
Especially: StartExtTextInput, ExtTextInput, EndExtTextInput

@GulsahKose have you checked maybe the difference between desktop app / online typing? do we have some additional formatting in desktop case?

eszkadev:
Here is handled typing in calc's cells for online:
https://opengrok.libreoffice.org/xref/core/sc/source/ui/view/gridwin.cxx?r=467085bf#3095

And here fro desktop:
https://opengrok.libreoffice.org/xref/core/sc/source/ui/view/gridwin.cxx?r=467085bf#3557

Found just by code reading so I can be wrong :) I hope it has some differences.

GulsahKose:
Hi Szymon, Actually firstly I attempt to solve this problem on core side. In core side we are appending that percent sign manually. You can see here: https://opengrok.libreoffice.org/xref/core/sc/source/ui/app/inputhdl.cxx?r=ce6babf7#3863

The problem is when we send this cell context via online, we send without percent sign and problem occurs.

Maybe we can manipulate the text on core side again but user won't never see the percent sign during edit. (This behaviour will be different from the core side). 

So I'm trying to manipulate string during edit. So trying to use content, cursor etc... I can see my way is not good. _onInput is very generic function but do you have any different idea how can i add this percent sign? Is there any special cell handling on online side?

eszkadev:
> Hi Szymon, Actually firstly I attempt to solve this problem on core side. In core side we are appending that percent sign manually. You can see here: https://opengrok.libreoffice.org/xref/core/sc/source/ui/app/inputhdl.cxx?r=ce6babf7#3863

Great, so can we add something like that in the [ScGridWindow](https://opengrok.libreoffice.org/s?defs=ScGridWindow&project=core)::[Command](https://opengrok.libreoffice.org/s?refs=Command&project=core) code path? I belive that is used when typing in Online

GulsahKose:
Let me try. Thank Szymon o/

GulsahKose:
Hello Szymon, I tried to apply same code path https://gerrit.libreoffice.org/c/core/+/141160

But I have a problem now. Here I need to add more control like the KeyEvent https://opengrok.libreoffice.org/xref/core/sc/source/ui/app/inputhdl.cxx?r=ce6babf7#3861

In my patch CommandEvent don't know the char info.  I tried to use GetExtTextInputData() and other data infos but all return nullptr. As a result it adds % to every character e% C%. Do you have any idea? Should I modify CommadEvent itself maybe what do you think about?

eszkadev:
I guess this function doesn't work well in online: https://opengrok.libreoffice.org/xref/core/sc/source/ui/app/inputhdl.cxx?r=ce6babf7#2674
Probably we are in eMode==SC_INPUT_NONE after every typed character due to handling the multiple views switching?

So maybe we can just check the content of an EditEngine - so it should have only numbers inside when we are adding %.
Hope it will help.

browser/src/layer/marker/TextInput.js
```diff
@@ -680,7 +681,20 @@ L.TextInput = L.Layer.extend({
...
+				this._lastContent = newText;
+				var percentText = this._textArea.value;
+				this._textArea.value = percentText.slice(0,2) + '%' + percentText.slice(2);
+				this._hasAutoPercent = true;
+			}
+
 			this._sendText(this.codePointsToString(newText));
+
+			if (content.length === 2 && this._hasAutoPercent) {
+				this._setCursorPosition(2);
```
mikekaganski, line 696: IMO, this only has to be done in the same event when the percent was appended.

Imagine this:
1. I type "1"
2. It appends "%", sets _hasAutoPercent, and moves the cursor to position 2 (OK)
3. I type "2" (the current text is 12%)
4. I press "Del"

At this point, the text is "12", and its length is 2, and _hasAutoPercent is true. It will put the cursor between 1 and 2, while it should stay after the 2.

So you don't need _hasAutoPercent.

eszkadev: Agree, not using any "state" variable would be perfect :)

GulsahKose:
I've updated commit as Mike suggested. Problematic behaviors still there Let me describe them:
1-) When i finish the typing into cell and press enter cursor still behind of the percentage sign and it moves sign to next cell
2-) Using _setCursorPosition function doesn't move the cursor visually. But in practice it adds the characters true position.

pedropintosilva:
testing this now, maybe then could you please amend the commit so `WIP: ` is removed from it?

pedropintosilva:
@GulsahKose Seems much better, there is just one thing. When pressing `Esc` key it results in something unexpected:
- typing 1
- The percentage is added automatically
- But pressing `Esc` and then typing again 1 the percentage is not added anymore
- Furthermore by then pressing "Enter" key it gives different result when compared to the other cases

Maybe a video is easier to understand:

https://user-images.githubusercontent.com/65948705/195557986-67a78368-ab61-431f-8993-5792222d266d.mp4

pedropintosilva:
Thanks, this is not a ^ regression and it is already a huge improvement!

pedropintosilva (approved):

PR #5321, opened 2022-10-07
browser: fix: using IME input candidates cannot be selected correctly

Rash419:
Change-Id: I7fac8ddfcdf5eeb3f03fe79bdf538fb52d311cce

* Target version: master 

### Summary
This fixes both the chrome and firefox bugs  :)

### DIFF
--- a/browser/src/layer/marker/TextInput.js
+++ b/browser/src/layer/marker/TextInput.js
@@ -879,13 +879,11 @@ L.TextInput = L.Layer.extend({
 			return;
 
 		this._map.notifyActive();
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
seems to be reasonable, we shouldn't clear buffer when composing
@Rash419 Unfortunately I cannot test as I have trouble to setup japanese IME - can we ask someone else to verify?

Rash419:
> seems to be reasonable, we shouldn't clear buffer when composing
> @Rash419 Unfortunately I cannot test as I have trouble to setup japanese IME - can we ask someone else to verify?

@Ezinnem can you try testing it ? 
@eszkadev or I can share the clip

PR #5331, opened 2022-10-11
browser: set the document view only if it is offline

hcvcastro:
The cursor will hide and the permission will be updated
after the document is re-connected.

Change-Id: Ib0a584371317adabe9898dad8bb0cd96717cefff


dennisfrancis (approved):
Tested, it works as expected. Thanks!

PR #5351, opened 2022-10-18
Hide insert menu item when in readonly mode

pedropintosilva:
except if we are in presence of PDF
- Insert is needed there (insert > comment)

Change-Id: I20276e5690fa515a6fb497d1bb3447d0808f02d0

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -2019,6 +2019,11 @@ L.Control.Menubar = L.Control.extend({
 	_createMenu: function(menu) {
 		var itemList = [];
 		var docType = this._map.getDocType();
+		var isReadOnly = this._map.isReadOnlyMode();
+
+		if (isReadOnly && !app.file.editComment) {
+			this._hiddenItems.push('insert');
+		}
 		for (var i in menu) {
 			if (this._checkItemVisibility(menu[i]) === false)
 				continue;


browser/src/control/Control.Menubar.js
```diff
@@ -2002,6 +2002,10 @@ L.Control.Menubar = L.Control.extend({
 	_createMenu: function(menu) {
 		var itemList = [];
 		var docType = this._map.getDocType();
+		var isReadOnly = this._map.isPermissionReadOnly();
+		if (docType !== 'drawing' && isReadOnly) {
```
eszkadev, line 2006: We should check probably if it is .pdf, so it will not appear in readonly .odg, but probably it is not possible here directly (on client side).

We have 'app.file.fileBasedView' variable which indicates that we don't watch separate slides but have continuous scrolling - which is used in PDFs. Maybe that one we should use? I think @gokaysatir knows if that is enough condition.

pedropintosilva: @eszkadev makes sense. What do you think @gokaysatir ? And what about `BaseFileName` https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.WOPI.js#L102 ?

pedropintosilva: I have tried with `app.file.fileBasedView` but it ends up being true on pdf, impress and drawing when readonly-mode:
```
false [Control.Menubar.js:2006:10](localhost:9980/browser/src/control/Control.Menubar.js)
true [Control.Menubar.js:2006:10](localhost:9980/browser/src/control/Control.Menubar.js)
true [Control.Menubar.js:2006:10](localhost:9980/browser/src/control/Control.Menubar.js)
```

pedropintosilva: I know we avoid depend on file extension so I wonder if it's not better to check if all children are disabled and if it is add it to the hidden array:
The problem is that the direct children do not have disabled status: 
```javascript
Array(8) [ li#menu-saveas, li#menu-savecomments, li#menu-shareas, li#menu-print, li#menu-rev-history, li#menu-downloadas, li, li#menu-closedocument.readonly ]
​
0: <li id="menu-saveas" class="" role="menuitem">
​
1: <li id="menu-savecomments" class="" role="menuitem">
​
2: <li id="menu-shareas" class="" role="menuitem">
​
3: <li id="menu-print" class="" role="menuitem">
​
4: <li id="menu-rev-history" class="" role="menuitem">
​
5: <li id="menu-downloadas" class="" role="menuitem">
​
6: <li class="" role="menuitem">
​
7: <li id="menu-closedocument" class="readonly" role="menuitem">
​
length: 8
​
<prototype>: Array []
Control.Menubar.js:2076:10
```

just one level under we have that info in the `<a>`

pedropintosilva: @gokaysatir 
> filebasedview shouldn't be set to true on read only mode. It seems to be a bug to me.

eszkadev: I think so

GulsahKose: I have an idea to detect file type according to mimetype. We are detecting docType here for Impress
ImpressTileLayer.js: _onStatusMsg: function (textMsg)
  var command = app.socket.parseServerCmd(textMsg)
  this._docType = command.type;

If I'm correct, server sends that textMsg that icludes "type" info. If we send the mime type too from server side to online, we can define a getDocMimeType() Is there any obstacle for us to do this?

eszkadev: yes, we can use that if it is already sent.
Only we have to check if ALWAYS `createMenu` occurs after we receive status message. In other words: if we create menubar after we received status message (eg. in `initializeSpecializedUI` in `UIManager`) or before/concurrently (in `initializeBasicUI`). So we will be sure there is no race here.

GulsahKose: As I learn the max info about the document as following:
    void dumpState(std::ostream& oss)
    {
        oss << "Kit Document:\n"
            << std::boolalpha
            << "\n\tpid: " << getpid()
            << "\n\tstop: " << _stop
            << "\n\tjailId: " << _jailId
            << "\n\tdocKey: " << _docKey
            << "\n\tdocId: " << _docId
            << "\n\turl: " << _url
            << "\n\tobfuscatedFileId: " << _obfuscatedFileId
            << "\n\tjailedUrl: " << _jailedUrl
            << "\n\trenderOpts: " << _renderOpts
            << "\n\thaveDocPassword: " << _haveDocPassword // not the pwd itself
            << "\n\tisDocPasswordProtected: " << _isDocPasswordProtected
            << "\n\tdocPasswordType: " << (int)_docPasswordType
            << "\n\teditorId: " << _editorId
            << "\n\teditorChangeWarning: " << _editorChangeWarning
            << "\n\tmobileAppDocId: " << _mobileAppDocId
            << "\n\tinputProcessingEnabled: " << _inputProcessingEnabled
            << "\n";

Extra we know WOPISrc the local path of the document. For xml based filetypes we can reach the "mimetype" if we unzip the files and read the xml files. But for PDF is not xml based and couldn't find if it includes any mimetype info.

It seems to me we have to rely file extension.

pedropintosilva:
I wonder if we could avoid adding the object directly to the DOM (at least until we now each ones should be added)

- Create arrays of objects (top menu entries) without adding it to the DOM
- Then create the sub arrays (the subitems) again without adding it to the DOM
- Check if all subitems are disabled
- Add to hidden list the top menus that have all their subitems disabled
- Finally Add objects to the DOM

hm.... or all top menu items start inside of hidden until we reach on of the subitems that are not disabled -> then the top menu gets removed from hidden list

GulsahKose:
Detecting if all submenus disabled seems the very hard way. So I returned to detecting mimetype somehow. Here is my detections
In ImpressTileLayer.js we are detecting docType . So here is the right place to detect mimetype too.

ın initiliaze() function we can reach the local file path via  "window.wopiSrc"

If we read that files first 4 byte is equeal to "%PDF" we can sure the file is a pdf file. we can define isPdf function to check.
        var fs = require('fs');
        var Buffer = require('buffer').Buffer;                                                                               
        fs.open(window.wopiSrc, 'r', function(status, fd) {
            if (status) {
                console.log(status.message);
                return;
            }
            var buffer = Buffer.alloc(4);
            fs.read(fd, buffer, 0, 4, 0, function(err, num) {
                console.log(buffer.toString('utf8', 0, num));
            });
        });
To try that way I'm getting 41:12  error  'require' is not defined  no-undef
Do you know why I'm getting this. Is it proper way to read a file's bytes?

thebearon:
There's something I don't understand. When COOL is passed a PDF, it already knows that the file doesn't open in a regular draw view, but in the semi-read-only PDF view (which is advertised as `view_comment` in the discovery XML). Why does it have to figure out that piece of information again?

mmeeks:
I share Aron's question really =) we should re-use some property on the document that is propagated nicely from the core - IIRC we have something like this already for the continuous scrolling feature (?) we use for PDFs. so ...

pedropintosilva:
right but if I understood correctly checking for that is not working (we seems to have another bug there) https://github.com/CollaboraOnline/online/pull/5351#discussion_r999438858

eszkadev:
there is a `isAllowChangeComments()` function which can help: https://github.com/CollaboraOnline/online/blob/7043365475ef7d997e2e553eeb524bc5b6aff65c/wsd/ClientSession.hpp#L69

eszkadev (approved):
looks good
ah didn't found `file.editComment` earlier, that was that easy :)

GulsahKose (approved):

PR #5354, opened 2022-10-18
notebookbar: repair icon should be big when not in 2 rows

eszkadev:


pedropintosilva (approved):
Thanks, perfect.
Also tested with docsign option enabled, it works.

PR #5359, opened 2022-10-19
wsd: add lang param to conversion

gmasei11:
This is needed, for example, for date type cells
for which the format language is set to default.
In this case the load language is used and it
determines the display/output format.

Change-Id: I49ec4940377b261971224d37ea5fbe8ed7006157

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Change seems good to me =) thanks for that.

PR #5364, opened 2022-10-21
jsdialog: apply edit field type

eszkadev:
works with: https://gerrit.libreoffice.org/c/core/+/141619


dennisfrancis (approved):
Looks good. Thanks!

PR #5365, opened 2022-10-21
masterpage: followup (part 2)

eszkadev:
this one requires core change: https://gerrit.libreoffice.org/c/core/+/137869

### DIFF
--- a/browser/src/control/Control.Sidebar.js
+++ b/browser/src/control/Control.Sidebar.js
@@ -146,7 +146,8 @@ L.Control.Sidebar = L.Control.extend({
 	},
 
 	changeDeck: function(unoCommand) {
-		app.socket.sendMessage('uno ' + unoCommand);
+		if (unoCommand !== null)
+			app.socket.sendMessage('uno ' + unoCommand);
 		this.setupTargetDeck(unoCommand);
 	},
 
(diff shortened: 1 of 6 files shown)


browser/src/control/Control.Sidebar.js
```diff
@@ -146,7 +146,8 @@ L.Control.Sidebar = L.Control.extend({
 	},
 
 	changeDeck: function(unoCommand) {
-		app.socket.sendMessage('uno ' + unoCommand);
+		if (unoCommand !== null)
+			app.socket.sendMessage('uno ' + unoCommand);
 		this.setupTargetDeck(unoCommand);
```
Ashod, line 151: If `unoCommand == null`, should we still call `this.setupTargetDeck` ?

Ashod (approved):
Thanks @eszkadev. See questions please.

browser/src/control/Control.Sidebar.js
eszkadev: yes, we want to reset it

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1623,6 +1623,12 @@ L.CanvasTileLayer = L.Layer.extend({
 		}
 		else if (textMsg.startsWith('status:') || textMsg.startsWith('statusupdate:')) {
 			this._onStatusMsg(textMsg);
+
+			// update tiles and selection because mode could be changed
```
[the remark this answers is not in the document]
eszkadev: yes, we send status because it contains slide list and mode so we can refresh slide previews

PR #5368, opened 2022-10-24
browser: Make sure that the header avatar list is limited

juliushaertl:
Change-Id: I0897e2ef3ca74ae20fdc056b863d9940805cbabe

* Resolves: https://github.com/nextcloud/richdocuments/issues/2574
* Target version: master 

### Summary

The header avatar list was only limited on the initial attempt to create it. This PR makes sure that the list will also be limited when users join afterwards.

### Checklist


browser/src/control/Control.UserList.js
```diff
@@ -139,8 +140,21 @@ L.Control.UserList = L.Control.extend({
 
 		var that = this;
 
+		var headerUserList = this.options.listUser.slice(-this.options.userLimitHeader);
+
+		// Remove users that should no longer be in the header
+		Array.from(document.querySelectorAll('#userListSummary [data-view-id]')).map(function(element) {
```
juliushaertl, line 146: @eszkadev I wasn't fully sure if I could already use arrow functions within this code parts. Those should be widely adopted if IE11 is not supported anymore: https://caniuse.com/arrow-functions

Maybe you can quickly tell if switching to those would be ok, that would make the code a bit less tangled :)

pedropintosilva: As far as I can remember we have dropped IE11 support since co-2021

juliushaertl: Decided to stick with the current one as no other .js file uses arrow functions

juliushaertl:
@pedropintosilva I somehow am unable to find why exactly this is failing in CI, can you maybe check and give me a pointer?

eszkadev (approved):
looks good

eszkadev:
CI should be fixed now, please rebase

PR #5373, opened 2022-10-25
wsd: support embedded video 

eszkadev:


Ashod (approved):
Thanks @eszkadev. This would work in edit-mode, and we'd need to fix the fullscreen case.

Let's get @mmeeks's +1 before we merge please.

mmeeks (approved):
Lets get this in then =)

PR #5381, opened 2022-10-26
Enable Show Status Bar menu entry in readonly mode

pedropintosilva:
Without this commit if the status bar is hidden, it's impossible
to bring it back and so the user is unable to:
- Search within readonly mode
- View the number of pages
- Zoom in our out via selecting the specific values

Change-Id: If30a0729862788f7475ce800b643db2b386722f3


gokaysatir (approved):
Thanks for the fix :)

PR #5382, opened 2022-10-26
Add read-only mode tooltip when hovering disabled menu items

pedropintosilva:
So user can have yet another visual cue that the current mode
is read-only

Change-Id: Ie7862e963603541c04d1b6662fbc7350f6e675a0


gokaysatir (approved):
Thanks :) I think we will need to translate this sentence.

And PR needs a re-base.

PR #5385, opened 2022-10-27
Writer: Use page twip coordiantes for switching the pages on read-onl…

gokaysatir:
…y mode.

Change-Id: Ie7ed6eb6d4f0178ff2f1f514e6e9f9321caf965d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Tested with read-only share, also tested via open PDF. It works. The current page number doesn't get updated when on PDF

PR #5386, opened 2022-10-27
Hide insert menu item when in readonly mode

GulsahKose:
To do some PDF specific operations we need to sure document is really PDF instead of relying only extension.

Change-Id: I02d710293c174e642e66e27f8ca545859521434c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -2019,6 +2019,13 @@ L.Control.Menubar = L.Control.extend({
 	_createMenu: function(menu) {
 		var itemList = [];
 		var docType = this._map.getDocType();
+		var isReadOnly = this._map.isPermissionReadOnly();
+		var isAllowChangeComments = this._map.isAllowChangeComments();
+
+		if (!isAllowChangeComments && isReadOnly) { // Checks if PDF or not indirectly. If It is not a PDF(view_comments)
+			//this._hiddenItems.push('insert');     // and a readonly doc, all subitems of insert menu are disabled so not
+		}                                           // have a reason to show insert menu anymore.
+
 		for (var i in menu) {
 			if (this._checkItemVisibility(menu[i]) === false)
 				continue;
--- a/browser/src/control/Permission.js
+++ b/browser/src/control/Permission.js
@@ -258,5 +258,9 @@ L.Map.include({
 
 	isPermissionEdit: function() {
 		return this._permission === 'edit';
+	},
+
+	isAllowChangeComments: function() {
+		return this.options.allowChangeComments;
 	}
 });
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -620,6 +620,11 @@ app.definitions.Socket = L.Class.extend({
 				this._map.options.permission = 'readonly';
 			}
 
+			if (json.allowComments) {
+				console.debug('xxx: we allowed comments');
+				this._map.options.allowChangeComments = true;
+			}
+
 			if (this._map._docLayer) {
 				this._map.setPermission(this._map.options.permission);
 			}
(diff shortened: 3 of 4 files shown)


pedropintosilva:
I'm still puzzled how to detect PDF (edit with Collabora Online) because none of the permission functions including the supposedly true only when editing PDF (isPermissionEditForComments) doesn't help at all in detecting PDF so either we are doing something wrong or permissions are getting reset when creating the menu. I would probably also try to add those changes within _checkItemVisibility (I think as @eszkadev has mentioned somewhere else)
```javascript
_checkItemVisibility: function(menuItem) {
+		// if (menuItem.id === 'insert' && isPDF) {
+		console.debug('isPermissionEditForComments: ' + this._map.isPermissionEditForComments());
+		console.debug('isPermissionReadOnly: ' + this._map.isPermissionReadOnly());
+		console.debug('isPermissionEdit: ' + this._map.isPermissionEdit());
+		console.debug('disableSelection: ' + this._map.disableSelection());
+		console.debug('shouldStartReadOnly: ' + this._map._shouldStartReadOnly());
+		//    return false
+		// }
```

none of those seems to help, here is the console output:

ODT readn only
```
isPermissionEditForComments: true
isPermissionReadOnly: true
isPermissionEdit: false
disableSelection: undefined
shouldStartReadOnly: false
```

PDF (Edit with Collabora Online)
```
isPermissionEditForComments: true
isPermissionReadOnly: true
isPermissionEdit: false
disableSelection: undefined
shouldStartReadOnly: false
```

ODP read only
```
isPermissionEditForComments: true
isPermissionReadOnly: true
isPermissionEdit: false
disableSelection: undefined
shouldStartReadOnly: false
```

ODG read only mode
```
isPermissionEditForComments: true
isPermissionReadOnly: true
isPermissionEdit: false
disableSelection: undefined
shouldStartReadOnly: false
```

browser/src/control/Control.Menubar.js
```diff
@@ -2019,6 +2019,13 @@ L.Control.Menubar = L.Control.extend({
 	_createMenu: function(menu) {
 		var itemList = [];
 		var docType = this._map.getDocType();
+		var isReadOnly = this._map.isPermissionReadOnly();
+		var isPermissionEditForComments = this._map.isPermissionEditForComments(); // Checks if PDF or not indirectly.
```
eszkadev, line 2023: current implementation of `isPermissionEditForComments`
is used to detect that we can add comments on mobile when not in "Edit mode" - so it means that actually file is writable:
https://github.com/CollaboraOnline/online/blob/7e22a01d4d0a2fe026a4b89f31678706d6148b6f/browser/src/control/Permission.js#L248

eszkadev:
> @eszkadev mentioned `isAllowChangeComments()` - > [#5351 (comment)](https://github.com/CollaboraOnline/online/pull/5351#issuecomment-1293549652) but @GulsahKose I have no idea how to reach that from the client side

yes, we need to pass value of that variable from the server to the client - so we can use that

GulsahKose:
I've update my commit and transferred to isAllowChangeComments() to client side but when we chek it alone it is true for all drawing doc types. Maybe we need to combine readonly mode again eg: (isReadOnly() &&  isAllowChangeComments()) Then it is true for only PDF.  What do you think about?

Now I'll setup a NC and try with more case.

eszkadev:
> 

If it works - great. In edit mode we also can edit comments so probably ` isAllowChangeComments` it is correct.

pedropintosilva:
> (isReadOnly() && isAllowChangeComments()) Then it is true for only PDF.

exactly

pedropintosilva:
solved with #5351

PR #5391, opened 2022-10-28
Increase the timeout in case copy-paste whole column.

GulsahKose:
To copy paste whole column into another document needs more time to get a reply from server.

Change-Id: I5321b42e321179d541464f8cbdc49fe01502c353

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks @GulsahKose :)

PR #5392, opened 2022-10-30
Dont show error for wrong shape content

eszkadev:
This happens when second time embedded video was selected.

(Error in the browser console about wrong URL)


Ashod (approved):
Thanks @eszkadev!

PR #5403, opened 2022-11-01
wsd: improved UnitWOPILock to avoid race-condition

Ashod:
Change-Id: Ib91f53f4ebbfef8d93cebe24d48fea26df479dfa

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -1863,6 +1863,13 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
             docBroker->setInteractive(false);
             docBroker->setLoaded();
 
+            if (UnitWSD::isUnitTesting())
+            {
+                UnitWSD::get().onDocBrokerViewLoaded(
+                    docBroker->getDocKey(),
+                    std::dynamic_pointer_cast<ClientSession>(shared_from_this()));
+            }
+
 #if !MOBILEAPP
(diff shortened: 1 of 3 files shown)


wsd/ClientSession.cpp
```diff
@@ -1863,6 +1863,13 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
             docBroker->setInteractive(false);
             docBroker->setLoaded();
 
+            if (UnitWSD::isUnitTesting())
+            {
+                UnitWSD::get().onDocBrokerViewLoaded(
+                    docBroker->getDocKey(),
```
eszkadev, line 1869: we send `status:` message very often in calc (when we scroll down for example)
so this function will be called multiple times but should be only one time, I think?

Ashod: Right. It's possible for this to get called multiple times in that case. It won't affect this test (it doesn't load calc or do anything to trigger `status:` more than once (during loading).

Perhaps we can rename this to `onDocBrokerStatusMessage`, that's one solution to signal that this is not strictly loading only (although, that's what `status:` meant, originally).

For now, I don't see this as an issue, since the reuse of `status:` in calc is a special case. A test that needs to handle this event will need to know that. Renaming is reasonable, and I don't mind, but I rather do it when we have a case that will benefit from it. Right now, it will make the current use less readable, I think.

eszkadev (approved):
looks good

PR #5405, opened 2022-11-01
Limit the automatic addition of percentage sign.

GulsahKose:
In mobile view there is no way to select a cell and type something into. We had to switch edit mode or use formula bar. So automatic percentage sign addition is useless on mobile view.

Change-Id: Ie0fca8121534a53304414d98e6fbdef82a349d89

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks good

PR #5406, opened 2022-11-01
private/pedro/statusbar never goes readonly

pedropintosilva:
- Code refactoring: avoid long statusbar css rules
- Fix statusbar display discrepancies when in readonly mode

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -35,10 +35,6 @@
 #tb_actionbar_item_LanguageStatus .w2ui-tb-caption{
 	float: none;
 }
-#toolbar-down.readonly #tb_actionbar_item_InsertMode,
-#toolbar-down.readonly #tb_actionbar_item_InsertMode + td[id^='tb_actionbar_item_break'],
-#toolbar-down.readonly #tb_actionbar_item_StatusSelectionMode,
-#toolbar-down.readonly #tb_actionbar_item_StatusSelectionMode + td[id^='tb_actionbar_item_break'],
 #toolbar-down.readonly #tb_actionbar_item_LanguageStatus .w2ui-button .w2ui-tb-down {
 	display: none;
 }
--- a/browser/src/control/Control.StatusBar.js
+++ b/browser/src/control/Control.StatusBar.js
@@ -456,6 +456,11 @@ L.Control.StatusBar = L.Control.extend({
 
 	onPermissionChanged: function(event) {
 		var isReadOnly = event.perm === 'readonly';
+		if (isReadOnly) {
+			$('#toolbar-down').addClass('readonly');
+		} else {
+			$('#toolbar-down').removeClass('readonly');
+		}
 		$('#PermissionMode').parent().html(this._getPermissionModeHtml(isReadOnly));
 	},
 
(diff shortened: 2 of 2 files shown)


browser/src/control/Control.Notebookbar.js
```diff
@@ -165,8 +165,10 @@ L.Control.Notebookbar = L.Control.extend({
 			this._showNotebookbar = true;
 			this.showTabs();
 			$('.main-nav').removeClass('readonly');
+			$('#toolbar-down').removeClass('readonly');
```
eszkadev, line 168: Please try to avoid modifying other components in unrelated files. Then we have clear separation between them.
This should be done in the `Status bar` code or in the `UIManager` which shows/hides components (if it depends on the used view mode).

Statusbar:  https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.StatusBar.js#L445

UIManager: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L629

browser/css/toolbar.css
```diff
@@ -35,10 +35,6 @@
 #tb_actionbar_item_LanguageStatus .w2ui-tb-caption{
 	float: none;
 }
-#toolbar-down.readonly #tb_actionbar_item_InsertMode,
-#toolbar-down.readonly #tb_actionbar_item_InsertMode + td[id^='tb_actionbar_item_break'],
-#toolbar-down.readonly #tb_actionbar_item_StatusSelectionMode,
```
eszkadev, line 40: this is nice! :)

browser/src/control/Control.StatusBar.js
```diff
@@ -405,8 +405,13 @@ L.Control.StatusBar = L.Control.extend({
 
 		this._updateToolbarsVisibility();
 
-		if (isReadOnly)
+		if (isReadOnly) {
 			statusbar.disable('LanguageStatus');
+			statusbar.hide('InsertMode');
+			statusbar.hide('break6');
+			statusbar.hide('StatusSelectionMode');
+			statusbar.hide('break7');
```
eszkadev, line 413: This could go into `_updateToolbarsVisibility` function -> will be applied on any status change, not only when created.

eszkadev: (with else block for showing elements)

browser/src/map/Map.js
```diff
@@ -186,6 +186,7 @@ L.Map = L.Evented.extend({
 				L.DomUtil.addClass(this._container.parentElement, 'readonly');
 				if (window.mode.isDesktop() || window.mode.isTablet()) {
 					L.DomUtil.addClass(L.DomUtil.get('toolbar-wrapper'), 'readonly');
+					L.DomUtil.addClass(L.DomUtil.get('toolbar-down'), 'readonly');
```
eszkadev, line 189: This should go into statusbar component, update permission handler. "Similar things together".

pedropintosilva: I don't think 
```javascript
					L.DomUtil.addClass(L.DomUtil.get('toolbar-down'), 'readonly');
```
is needed anymore :)

browser/src/control/Control.StatusBar.js
pedropintosilva: done, thanks

browser/css/toolbar.css
pedropintosilva: :)

browser/src/control/Control.Notebookbar.js
pedropintosilva: done, added to statusbar thanks

eszkadev (approved):
looks good

PR #5414, opened 2022-11-02
Improve zstd ios build script

tml1024:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
All looks great to me - thanks Tor =)
We should check the build instructions on the web page too I think.

PR #5416, opened 2022-11-02
Add logout button to admin console

### LINKED ISSUE #5360: Add logout button to admin console
**Is your feature request related to a problem? Please describe.**
I'd like to be able to logout from the admin console, browser/dist/admin/admin.html

**Describe the solution you'd like**
A logout button in the top-right corner

**Describe alternatives you've considered**
Close and reopen the browser

**Additional context**
Using coolwsd 22.05

pedropintosilva:
Adds possibility to logout from admin console
fixes  #5360

Change-Id: I30f2300ac90252fc08d24bd0aeb949454ac5e044


pedropintosilva:
He @pmarini and thanks for reporting this #5360. I'm not sure there is a way to do this without clearing the browser cookies (since we use http auth for admin console) but then again I'm not versed in C++ or C for that matter. I have added this draft PR but it's missing the clearing browser cookies (saw some references to `HTTPBasicCredentials credentials(request);` in FileServer.cpp)

@pmarini if you want to contribute feel free to push additional commits to this PR or create a new one. (we can then close this up)

Note: I don't know if is ok to clear the browser cookies better to hear what @Ashod has to say about it and follow

Ashod:
> Note: I don't know if is ok to clear the browser cookies better to hear what @Ashod has to say about it and follow

While I'm not the author of the admin login logic, that's my understanding. A token is stored in as a cookie, which is verified upon requesting the admin page. If the cookie is missing, a new login is required. 

But to be clear, we only need to clear our JWT cookie, _not_ all cookies.

browser/admin/src/AdminSocketSettings.js
```diff
@@ -41,6 +41,21 @@ var AdminSocketSettings = AdminSocketBase.extend({
...
+			document.getElementById('btnLogout').onclick = function() {
+				var dialog = (new DlgYesNo())
+					.title(_('Confirmation'))
+					.text(_('Are you sure you want to logout from admin console?'))
+					.yesButtonText(_('OK'))
+					.noButtonText(_('Cancel'))
+					.type('warning')
+					.yesFunction(function() {
+						console.debug(document.cookie);
+						document.cookie = 'JWT= ;expires=Thu, 18 Dec 2013 12:00:00 UTC;max-age=0;path=/;';
```
pedropintosilva, line 54: Doesn't work and I have no idea why. Locally in a test file it works: https://jsfiddle.net/ppsilva/o4Lgzf5p/

pedropintosilva: also tried with brand new clean ff profile and still doesn't work.. hm

pedropintosilva: so I think removing this cookie from client side is impossible because it seems to be set with HttpOnly option. so probably only possible to do so in the server side

PR #5426, opened 2022-11-04
segregate edit/read only permissions better

lpranam:
Change-Id: I85861352929d150aaf93da5570bda960b1ad46f0

* Target version: master 

### Summary
functional changes are contained in files:
browser/src/control/Permission.js
browser/src/map/handler/Map.WOPI.js
wsd/DocumentBroker.cpp

rest of the changes are refactoring and renaming of functions

### Checklist

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -123,7 +123,7 @@ L.Control.ContextMenu = L.Control.extend({
 
 	_onContextMenu: function(obj) {
 		var map = this._map;
-		if (!map.isPermissionEdit()) {
+		if (!map.isEditMode()) {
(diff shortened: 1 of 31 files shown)


browser/src/main.js
```diff
@@ -78,7 +78,6 @@ window.addEventListener('beforeunload', function () {
 	}
 });
 
-window.docPermission = permission;
```
eszkadev, line 81: so we will ignore completly `&permission=xxxxx` from the URL and use only WOPI prop?
what about existing integrators? They can use that still

lpranam: thanks for the catch, I forgot about URL param

gokaysatir (approved):

PR #5431, opened 2022-11-05
systemplate: include /lib/*/nss

Ashod:
It seems the location of the NSS dynamic
libraries have changed recently. This is
true on Ubuntu 22.04 and it seems on
Clear Linux.

We expected to find them in /usr/lib but
they are now in /lib. We support both
now, so the systemplate should work on
both new and old systems.

The symptom of not having the NSS libs
is failure to open password-protected
documents. The tests failed, which
helped to reproduce and debug.

Change-Id: Ifb4cbc4e2c852464ffcdcc19801689ea60355042

### DIFF
--- a/coolwsd-systemplate-setup
+++ b/coolwsd-systemplate-setup
@@ -44,6 +44,7 @@ cd / || exit 1
 find etc/ld.so.* \
      lib/ld-* lib64/ld-* \
      lib/libnss_* lib64/libnss_* lib/*/libnss_* \
+     lib/*/nss/*.so \
      lib/libresolv* lib64/libresolv* lib/*/libresolv* \
      var/cache/fontconfig \
      etc/fonts \


mmeeks (approved):
Looks great to me - thanks Ash =)

thebearon:
What's strange is that looking at the `libnss3` package, it seems the libraries are still put under `/usr/lib` in Ubuntu 22.04:
https://packages.ubuntu.com/jammy/amd64/libnss3/filelist

But indeed, the library seems to be copied to `/lib/x86_64-linux-gnu` as well. Still, I wouldn't call that "changed".

timar:
We use `--without-system-nss` in CPLinux-LOKit.conf. So nss libs are in /opt/collaboraoffice/program. In official bulids at least.

Ashod:
> What's strange is that looking at the `libnss3` package, it seems the libraries are still put under `/usr/lib` in Ubuntu 22.04: https://packages.ubuntu.com/jammy/amd64/libnss3/filelist
> 
> But indeed, the library seems to be copied to `/lib/x86_64-linux-gnu` as well. Still, I wouldn't call that "changed".

@thebearon, yes, I should have been more careful with what I meant was changed. You are right, I can see the `.so` files in both `/usr/lib` and `/lib`, but `dlopen` is not looking inside `/usr/lib`, for some reason that I didn't try to find out (by that point I had spent far more time than I had planned, so didn't chase it further). So the end result is that fresh build of Core simply can't load NSS and the crypto stuff fail, even though it has the `.so` files it needs inside the jail, in `/usr/lib`.

Ashod:
> We use `--without-system-nss` in CPLinux-LOKit.conf. So nss libs are in /opt/collaboraoffice/program. In official bulids at least.

Thanks, @timar. Good to know!

PR #5432, opened 2022-11-07
Unlock only after uploading

Ashod:
- wsd: test: log multiple exitTest cases with different results
- wsd: log the socket FD used to listen to client connections
- wsd: test: reduce the duration of UnitWOPIStuckSave
- wsd: test: move assertLockRequest to WopiTestServer
- wsd: unlock the document only after uploading

### DIFF
--- a/test/UnitWOPIUnlock.cpp
+++ b/test/UnitWOPIUnlock.cpp
@@ -1,0 +1,167 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <config.h>
+
+#include "lokassert.hpp"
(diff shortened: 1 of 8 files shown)


test/UnitWOPIUnlock.cpp
```diff
@@ -0,0 +1,167 @@
...
+            LOK_ASSERT_MESSAGE("Lock String cannot be empty", !lock.empty());
+            _lockState = newLockState;
+            _lockString = lock;
+            TRANSITION_STATE(_phase, Phase::Modify);
+        }
+        else if (_phase == Phase::Unlock)
+        {
+            LOK_ASSERT_EQUAL_MESSAGE("Expected X-WOPI-Override:UNLOCK", std::string("UNLOCK"),
+                                     newLockState);
+            LOK_ASSERT_MESSAGE("Document is not unlocked", _lockState != "UNLOCK");
```
eszkadev, line 72: shouldn't be   `_lockState == "LOCK"`   to be sure it's not empty?

eszkadev (approved):

test/UnitWOPIUnlock.cpp
Ashod: On line 61 we assert that `newLockState` is "LOCK" and then we assign it to `_lockState`. I suppose we could make it more explicit.

PR #5443, opened 2022-11-09
Don't mix mouse move and auto-scroll-to-point handling

eszkadev:
If user moves mouse - reset current auto scroll settings. So we don't get scheduled events changing current state and causing movement in the other direction.


eszkadev:
merged in: https://github.com/CollaboraOnline/online/pull/5208

PR #5447, opened 2022-11-09
Send column selection text when Function Wizard is open.

gokaysatir:
Change-Id: I09ca97f2f7ec19cd4836f0dbfb8cbf863f3c5f3e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
Do we need the same for rows?

eszkadev (approved):
works fine

gokaysatir:
> Do we need the same for rows?

Done, thank you @eszkadev :)

PR #5449, opened 2022-11-09
private/pedro/remove macro dialog css

pedropintosilva:
- Remove unnecessary CSS for MacroSelector dialog


pedropintosilva:
How to test it:
* set to true: `cat coolwsd.xml | grep macro`
* File tab > Run macro
* Test that the Macro selector dialog looks ok

eszkadev (approved):
dialog seems to look good

Issue #5436, opened 2022-11-07
Memory leak with big remote config


klauserber:
**Describe the bug**
We are running collabora online inside of Kubernetes at a larger scale. Since we have activated the full count of allowed hosts and aliases we observe a constant increase of memory consumption. The configuration is loaded via remoteconfig poll once a minute and it is about 1 MB large containing about 5000 hosts each with one alias.

**To Reproduce**
1. Configure collabora online for remote configuration
2. Deliver a big json configuration of min 1 MB size
3. Do nothing else with the instance.
4. Monitor the memory consumption
5. See that the memory consumption is increasing continuously

**Expected behavior**
The memory utilization should be not increasing under the above conditions.

**Actual behavior**
See section 'To Reproduce'.

**Screenshots**
<img width="1380" alt="Bildschirmfoto 2022-11-04 um 17 18 36" src="https://user-images.githubusercontent.com/9998995/200307250-1d8829ec-bc59-48f6-9ed2-69699924c2c6.png">

**Desktop (please complete the following information)**
not relevant

**Smartphone (please complete the following information)**
not relevant

**Additional context**
Container Image: registry.gitlab.collabora.com:22.05.6.3.1
Kubernetes: 1.23.6

PR #5453, opened 2022-11-11
fix: memory leak with remote config

Rash419:
when eTag header is not passed from remote server it keeps on parsing the new settings without clearing old ones

Change-Id: Iaad9fb6c0e7fd45be3e121b0c0bebaba17d90aab

* Resolves: #5436 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks @Rash419 :)

mmeeks (approved):
Seems very sensible to me =)

PR #5456, opened 2022-11-11
enable link to non-system libpng and openssl

timar:
Change-Id: I109242bde268e63e9918af0f44fcb456e2b8f6c5

* Resolves: # 
* Target version: master 

### Summary
The goal is to create binaries that can run on any Linux distro.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,7 +65,7 @@ nginxconfig_DATA = etc/nginx/coolwsd.conf
 ACLOCAL_AMFLAGS = -I m4
 
 # quick and easy for now.
-include_paths = -I${top_srcdir}/common -I${top_srcdir}/net -I${top_srcdir}/wsd -I${top_srcdir}/kit ${ZLIB_CFLAGS} ${ZSTD_CFLAGS}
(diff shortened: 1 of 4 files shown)


Makefile.am
```diff
@@ -65,7 +65,7 @@ nginxconfig_DATA = etc/nginx/coolwsd.conf
 ACLOCAL_AMFLAGS = -I m4
 
 # quick and easy for now.
-include_paths = -I${top_srcdir}/common -I${top_srcdir}/net -I${top_srcdir}/wsd -I${top_srcdir}/kit ${ZLIB_CFLAGS} ${ZSTD_CFLAGS}
+include_paths = -I${top_srcdir}/common -I${top_srcdir}/net -I${top_srcdir}/wsd -I${top_srcdir}/kit ${ZLIB_CFLAGS} ${ZSTD_CFLAGS} ${PNG_CFLAGS}
```
Ashod, line 68: Line is getting a bit long. Maybe we should break on the next update.

Ashod (approved):
Very good, I needed this for openssl (when working on supporting 3.x with backwards compatibility with 1.x). Thanks @timar!

PR #5460, opened 2022-11-12
Add "Formula Object" button to Insert tab of notebookbar

mikekaganski:
Change-Id: Icd06d49fe18b37fd664219a951ba9ae822659150


pedropintosilva (approved):
Thanks @mikekaganski ! I have tested the commit works as described: the buttons is present and inserts the formula object in text documents. Things I have noticed:

the canvas are outside of document page changes colour and not fully (the left side doesn't get dark): 

https://user-images.githubusercontent.com/65948705/202243174-1efad53c-208b-4f81-8793-81cef7d1a17f.mp4

image 


The sidebar, if it's closed, should open automatically otherwise it user has very few clues of what is happening (but probably could be a follow-up [?]): 

image

I also couldn't find out how to edit the formula:
| Online: where to edit the formula?                                                                               | Core: Edit panel (bellow) appears on double click                                                               |
|-----------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| image | image |

mikekaganski:
@pedropintosilva thank you for testing!

> the canvas are outside of document page changes colour and not fully (the left side doesn't get dark):

This is the same with charts, so if this looks undesirable, likely we need to change some common code that dims things outside of current in-place edited object. I suggest we make it a separate task if needed?

> The sidebar, if it's closed, should open automatically otherwise the user has very few clues on what is happening (but probably could be a follow-up [?]):

Yes, I haven't been able to figure how to make that happen automatically. A follow-up would be reasonable approach.

> I also couldn't find out how to edit the formula:

The formula is edited in-place, not in a separate code window. That was the original idea behind the respective GSoC "visual formula editor" project by Jonas Finnemann Jensen (https://jonasfj.dk/downloads/GSoC2010/slide-osc10.pdf). This functionality is exposed to the Online here; you may click inside the window to put cursor; or you may use keyboard to position the cursor; you use Shift+arrows to select; and you type things like "a^2<down_arrow>+b=c" in the box, or you use the sidebar. The code window (in the sidebar) is a topic for a follow up.

Hope this clarifies this :)

PR #5461, opened 2022-11-12
Re-Enable TileTests and Minor Misc Improvements

Ashod:
- wsd: test: reduce the duration of UnitWOPIStuckSave
- wsd: log the socket FD used to listen to client connections
- wsd: test: log multiple exitTest cases with different results
- wsd: test: enable tile-tests
- wsd: test: re-order tests to minimize execution time


thebearon (approved):
Thanks, Ash!

PR #5462, opened 2022-11-13
wsd: test: fix UnitWOPILock once and for all

Ashod:
This fixes a race in UnitWOPILock and the
fact that the second view wasn't really
read-only due to empty extensions being
commentable, by default.

Also, improve UnitWOPIUnlock.


eszkadev (approved):
nice, it's more readable now :)

PR #5463, opened 2022-11-13
Log Improvements

Ashod:
- wsd: remove default parameters in Session
- wsd: log the session in question
- wsd: log edit and comment ext in discovery.xml


eszkadev (approved):
looks ok

PR #5464, opened 2022-11-13
Test Improvements and Refactoring

Ashod:
- wsd: test: header and forward declaration cleanup
- wsd: test: simplify UnitTimeout
- wsd: test: refactor client connection management

The client connection management had nothing to
do with the WopiTestServer and it was there
for convenience. The issue was that it was
not available to tests that do not need
the custom WopiTestServer and depending
on it would make them unnecessarily
complicated.

This adds a new intermediary class to manage
client connections in tests. The reason
this logic didn't move to UnitWSD or even
UnitBase is simply because the client
connection logic depends on the helpers
namespace and other test-specific code
that isn't available in COOLWSD. In short,
UnitBase and UnitWSD are primarily an
interface that COOLWSD and co depend on
and cannot contain test-specific logic
that isn't linked to coolwsd.

We also move the UnitWebSocket helper class.

### DIFF
--- a/test/UnitWSDClient.hpp
+++ b/test/UnitWSDClient.hpp
@@ -1,0 +1,118 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "helpers.hpp"
+#include "testlog.hpp"
+#include "Unit.hpp"
+#include "Util.hpp"
+
+#include <Poco/URI.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+class UnitWebSocket final
+{
+    std::shared_ptr<http::WebSocketSession> _httpSocket;
+
+public:
+    /// Get a websocket connected for a given URL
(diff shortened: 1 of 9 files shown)


test/UnitWSDClient.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include "Unit.hpp"
+#include "Util.hpp"
+
+#include <Poco/URI.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+class UnitWebSocket final
```
vmiklos, line 19: Nit: please document this new class.

test/UnitWSDClient.hpp
```diff
@@ -0,0 +1,118 @@
...
+private:
+    /// The WOPISrc URL.
+    std::string _wopiSrc;
+
+    /// Websockets to communicate.
+    std::vector<std::unique_ptr<UnitWebSocket>> _wsList;
+};
+
+/// Send a command message to WSD from a WopiTestServer on the given connection.
+#define WSD_CMD_BY_CONNECTION_INDEX(INDEX, MSG)                                                    \
```
vmiklos, line 107: Nit: does this have to be a macro? I don't immediately see something in the body that would be impossible to do with an inline function.

vmiklos (approved):
I don't see anything obviously wrong with this, see inline comments for some nits.

test/UnitWSDClient.hpp
Ashod: Will do in a follow up.

test/UnitWSDClient.hpp
Ashod: Good point. This macro is used several times and the idea was to hide the internal details. My hope is to replace it with something more sensible at some point.

PR #5469, opened 2022-11-15
Multitest Support

Ashod:
This adds support for multiple tests
within a single unit-test .so.

We did have many, many tests within
the converted "old style" tests already.
Except, they couldn't support the proper
event-driven tests, nor were they running
as fast as possible. They also blocked the
main thread in COOLWSD.

With this support, we can now avoid
creating a new .so for each test case.
This will reduce the effort to add new
tests and the build time. In addition,
the execution of each unit-test is
dominated by the initialization, which
takes in the order of 5 seconds. Running
multiple tests after this high load time
reduces the per-test average time.

Later we will support listing tests,
running particular tests (i.e. filtering)
and other debugging and troubleshooting
facilities.

Each commit is self-contained, builds and passes tests. Best reviewed by commit.

- wsd: test: improvements to UnitWSDClient
- wsd: test: use a global array to support multiple tests
- wsd: test: make dlhandle static and properly cleanup
- wsd: test: support multitests
- wsd: test: rename setHandle to initialize
- wsd: test: merge the results of all tests

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -27,10 +27,11 @@
 #include <common/SigUtil.hpp>
 #include <common/Message.hpp>
 
-UnitBase *UnitBase::Global = nullptr;
 UnitKit *GlobalKit = nullptr;
 UnitWSD *GlobalWSD = nullptr;
 UnitTool *GlobalTool = nullptr;
+UnitBase** UnitBase::GlobalArray = nullptr;
+int UnitBase::GlobalIndex = -1;
 char * UnitBase::UnitLibPath;
 static std::thread TimeoutThread;
 static std::atomic<bool> TimeoutThreadRunning(false);
@@ -39,7 +40,7 @@ std::timed_mutex TimeoutThreadMutex;
 /// Controls whether experimental features/behavior is enabled or not.
 bool EnableExperimental = false;
 
-UnitBase *UnitBase::linkAndCreateUnit(UnitType type, const std::string &unitLibPath)
+UnitBase** UnitBase::linkAndCreateUnit(UnitType type, const std::string& unitLibPath)
 {
 #if !MOBILEAPP
     void *dlHandle = dlopen(unitLibPath.c_str(), RTLD_GLOBAL|RTLD_NOW);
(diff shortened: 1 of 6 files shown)


common/Unit.cpp
```diff
@@ -187,9 +223,57 @@ void UnitBase::rememberInstance(UnitType type, UnitBase* instance)
...
+    if (GlobalArray)
+    {
+        // By default, this will check _setRetValue and copy _retValue to the arg.
+        // But we call it to trigger overrides and to perform cleanups.
+        int retValue = GlobalResult == TestResult::Ok ? EX_OK : EX_SOFTWARE;
+        GlobalArray[GlobalIndex]->returnValue(retValue);
+        if (retValue)
+            GlobalResult = TestResult::Failed;
+
+        for (int i = 0; GlobalArray[i] != nullptr; ++i)
```
vmiklos, line 240: If GlobalArray is a list of owning pointers, would it be better to go with std::vector<std::unique_ptr<>> instead of this manual memory management? Especially with the C-style, nullptr-terminated array.

test/UnitWSDClient.hpp
```diff
@@ -16,16 +16,30 @@
 #include <string>
 #include <vector>
 
+/// Send a command message to WSD from a UnitWSDClient instance on the given connection.
+#define WSD_CMD_BY_CONNECTION_INDEX(INDEX, MSG)                                                    \
```
vmiklos, line 20: I think I already commented on this elsewhere, I wonder if this could be a function instead of a macro.

vmiklos (approved):
Looks good, I like the idea to have less boilerplate for a new test. Some nits inline, but you can address those in follow-up commits.

common/Unit.cpp
Ashod: Yes, I suppose that's what we should use. I left it as simple as possible until I got something working, but there is more room to improve, and this is one of them.

test/UnitWSDClient.hpp
Ashod: Yes. This patch should've been included in the previous PR you reviewed, but for some reason I hadn't fixed up the relevant commit with this one. I'll revisit this macro at some point and clean it up (convert to function).

PR #5486, opened 2022-11-17
wsd: fix: wopi.host not working when alias_groups are not defined

Rash419:
regression from commit 3aa3334a7b2b0abad7a07beefa881be62a83020b

Change-Id: I601ea3bde0419129e39c1a1454c22c676d4870bd

* Resolves: # 
* Target version: master


thebearon (approved):
Thanks for fixing!

PR #5492, opened 2022-11-18
Enable read-only status label for tablets

pedropintosilva:
So tablet user is able to check what is the current edit mode
by looking at the status bar

Change-Id: I4990ca748762eb100b0ca9e791b4ae2a0843f2f6


eszkadev (approved):
looks good

PR #5494, opened 2022-11-18
calc: refresh cursor on reconnect

eszkadev:
When doing save as we reconnect the app to different spreadsheet. We need to invalidate cursor position and geometry to avoid showing incorrect selection.


dennisfrancis (approved):
Thanks!

PR #5497, opened 2022-11-21
Test Improvements

Ashod:
- wsd: test: return the correct type from hooks multi
- wsd: test: simplify WSD_CMD_BY_CONNECTION_INDEX
- wsd: test: log test start and finish visibly
- wsd: test: non-numeric filenames in WopiTestServer
- wsd: test: stop DocBroker when unit-test finishes
- wsd: test: fail -> onDataLoss
- wsd: test: capture the UnitWSD instance per DocBroker
- wsd: test: filter tests using COOL_TEST_OPTIONS
- wsd: test: log all test names before running
- wsd: test: capture the reason in exitTest
- wsd: test: improved assert logging on failure
- wsd: test: invoke configure for each test
- wsd: test: capture the UnitBase instance in WebSocketHandler
- wsd: test: filter subsequent tests
- wsd: test: do not invoke WSD test when finished

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -102,6 +102,24 @@ UnitBase** UnitBase::linkAndCreateUnit(UnitType type, const std::string& unitLib
     return nullptr;
 }
 
+void UnitBase::filter()
+{
+    // For now, support only filtering.
+    static const char* TestOptions = getenv("COOL_TEST_OPTIONS");
+    if (TestOptions == nullptr)
+        return;
+
+    const std::string filter = Util::toLower(TestOptions);
+    for (GlobalIndex = 0; GlobalArray[GlobalIndex] != nullptr; ++GlobalIndex)
+    {
+        const std::string& name = GlobalArray[GlobalIndex]->getTestname();
+        if (strstr(Util::toLower(name).c_str(), filter.c_str()))
+            break;
+
+        LOG_INF("Skipping test [" << name << "] per filter [" << TestOptions << "]");
+    }
+}
+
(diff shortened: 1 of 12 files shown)


common/Unit.cpp
```diff
@@ -75,31 +96,43 @@ UnitBase** UnitBase::linkAndCreateUnit(UnitType type, const std::string& unitLib
...
+    // For now, support only filtering.
+    static const char* TestOptions = getenv("COOL_TEST_OPTIONS");
+    if (TestOptions == nullptr)
+        return;
+
+    const std::string filter = Util::toLower(TestOptions);
+    for (; GlobalArray[GlobalIndex] != nullptr; ++GlobalIndex)
+    {
+        const std::string& name = GlobalArray[GlobalIndex]->getTestname();
+        if (strstr(Util::toLower(name).c_str(), filter.c_str()))
```
vmiklos, line 129: Would it make sense to have a way to support exact matches here? test/test.cpp:46 is the old code, that one works with regular expressions.

vmiklos (approved):
Looks OK in general, but please see inline comment for a nit. And perhaps it would make sense to document how to use the new COOL_TEST_OPTIONS.

common/Unit.cpp
Ashod: I certainly should improve this. It was a quick way to add support for filters and other options, and to improve later. 

Yes, agreed that supporting regex is a good idea, but for the simplest case, partial match is better because it's more user friendly (there is no harm in over-matching, but annoying the user to type the exact name of a test is a bit silly, IMNSHO).

PR #5498, opened 2022-11-21
Log Improvements

Ashod:
- wsd: clang-tidy: suppress bogus assert diagnostics
- wsd: http: correct written data-size reporting
- wsd: ssl: log the operation context in errors
- wsd: correct logged upload duration
- wsd: improve various logs

### DIFF
--- a/net/WebSocketSession.hpp
+++ b/net/WebSocketSession.hpp
@@ -305,7 +305,7 @@ private:
     {
         const auto header = COOLProtocol::getFirstLine(message);
         const bool match = COOLProtocol::matchPrefix(prefix, header);
-        LOG_DBG(context << (match ? "Matched" : "Skipped") << " message [" << prefix
+        LOG_DBG(context << (match ? " Matched" : " Skipped") << " message [" << prefix
                         << "]: " << header);
         return match;
     }
@@ -343,8 +343,6 @@ private:
 
             if (_shutdown && _outQueue.isEmpty())
             {
-                LOG_DBG(
-                    "WebSocketSession: closing gracefully after sending all the data, as flagged.");
                 sendCloseFrame();
             }
         }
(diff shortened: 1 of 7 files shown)


net/WebSocketSession.hpp
```diff
@@ -343,8 +343,6 @@ class WebSocketSession final : public WebSocketHandler
 
             if (_shutdown && _outQueue.isEmpty())
             {
-                LOG_DBG(
-                    "WebSocketSession: closing gracefully after sending all the data, as flagged.");
```
eszkadev, line 347: just to confirm: we don't want this log?

wsd/DocumentBroker.cpp
```diff
@@ -1266,13 +1269,19 @@ void DocumentBroker::handleSaveResponse(const std::string& sessionId, bool succe
...
+    // Rename even if no new save, in case we have an older version.
     if (rename(oldName.c_str(), newName.c_str()) < 0)
     {
         // It's not an error if there was no file to rename, when the document isn't modified.
         const auto onrre = errno;
-        LOG_DBG("Failed to renamed [" << oldName << "] to [" << newName << "] ("
-                                      << Util::symbolicErrno(onrre) << ": " << std::strerror(onrre)
-                                      << ')');
+        if (success || onrre != ENOENT)
+            LOG_ERR("Failed to rename [" << oldName << "] to [" << newName << "] ("
```
eszkadev, line 1278: I guess you wanted to print something different here than the else block will :)

eszkadev (approved):
added 2 comments, but looks good

wsd/DocumentBroker.cpp
Ashod: The log-level is different. I could make it more complex to reduce duplication, but the number of lines and readability will probably not improve.

I thought of adding more info, or to differentiate, but for some reason couldn't find anything to change. So I left them identical.

PR #5500, opened 2022-11-21
Tag polls correctly as running synchronously in the client thread.

mmeeks:
This avoids wakeup warning us that the thread is not started to wakeup.

Change-Id: Ifd5c483717024cb1c30521433d0d10acead01e2f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Ash - looks like we forgot to tag these as not involving a separate thread back in the day ... =)

Ashod (approved):
Makes sense. I meant to chase these warnings, but never gotten around to them. Thanks for chasing!

PR #5503, opened 2022-11-22
Fix PDF detection when deciding if edit button should be visible

pedropintosilva:
Before this commit both mobile and is_lock_readonly (upsell)
were being affected. Namely, the edit buttons was not being
shown for presentations and drawings

Change-Id: I903ea43ca5ea1c94eec6b8f63b8b5626b8ee2203


Rash419 (approved):
Thanks : ), works as expected

Rash419 (approved):
Tested again : ), looks good

PR #5505, opened 2022-11-22
config: add REST API Protocol to LanguageTool

hcvcastro:
Change-Id: I0e05ed89d02a9b467e1707db1638fc9ee554586a

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -1974,6 +1974,7 @@ void COOLWSD::innerInitialize(Application& self)
         { "languagetool.user_name", ""},
         { "languagetool.enabled", "false"},
         { "languagetool.ssl_verification", "true"},
+        { "languagetool.rest_protocol", ""},
         { "deepl.api_url", ""},
         { "deepl.auth_key", ""},
         { "deepl.enabled", "false"},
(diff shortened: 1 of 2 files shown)


wsd/COOLWSD.cpp
```diff
@@ -2418,6 +2419,8 @@ void COOLWSD::innerInitialize(Application& self)
     setenv("LANGUAGETOOL_APIKEY", apiKey.c_str(), 1);
     bool sslVerification = getConfigValue<bool>(conf, "languagetool.ssl_verification", "");
     setenv("LANGUAGETOOL_SSL_VERIFICATION", sslVerification ? "true" : "false", 1);
+    const std::string restProtocol = getConfigValue<std::string>(conf, "languagetool.rest_protocol", "");
+    setenv("LANGUAGETOOL_RESTPROTOCOL", restProtocol.c_str(), 1);
```
Ashod, line 2423: We shouldn't need envars (same for previous ones). The kit does get the contents of the coolwsd.xml file. See ConfigUtil and how it's used in Kit.cpp and ForKit.cpp by searching for `config::`. It's as easy as doing `config::getString("languagetool.rest_protocol", "");`

Ashod (commented):
Thanks @hcvcastro. One note.

wsd/COOLWSD.cpp
Ashod: I can see now that this envar is used in Core. Still, if we must use the envar, best to set it in ForKit and not here in WSD. I think that's cleaner. Same for other LANGUAGETOOL envars.

hcvcastro: I agree, but it is not my requirements to touch those LanguageTools settings. Maybe someone can do the followup clean patch :)

Ashod: Surely we can move these `setenv` calls to ForKit. That's all I'm suggesting. But we can do that in a follow up.

gokaysatir (approved):

PR #5507, opened 2022-11-23
wsd:fix: memory leak properly

Rash419:
introduced in e0752419575f09fef3d770f8cccf78c92a2d8612 commit. clearing new remote config  after addWriteable resulted into removing data top of top layer from Poco::LayeredConfiguration. This patch makes sure to remove old layer of remote config before applying the new layer

Change-Id: If11bd6e4b3a0a9f4ee32ae9bb43b076479e9bd84

* Resolves: # 
* Target version: master

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -990,6 +990,15 @@ public:
             setRaw(pair.first, pair.second);
         }
     }
+
+    void reset(const std::map<std::string, std::string>& map)
+    {
+        clear();
+        for (const auto& pair : map)
+        {
+            setRaw(pair.first, pair.second);
+        }
+    }
 };
 
 #if !MOBILEAPP
@@ -1178,14 +1187,15 @@ public:
     RemoteConfigPoll(LayeredConfiguration& config) :
         RemoteJSONPoll(config, "remote_config.remote_url", "remoteconfig_poll", "configuration")
     {
+        constexpr int PRIO_JSON = -200; // highest priority
+        _persistConfig = new AppConfigMap(std::map<std::string, std::string>{});
+        _conf.addWriteable(_persistConfig, PRIO_JSON);
     }
 
     virtual ~RemoteConfigPoll() { }
 
     void handleJSON(Poco::JSON::Object::Ptr remoteJson) override
     {
-        constexpr int PRIO_JSON = -200; // highest priority
-
         std::map<std::string, std::string> newAppConfig;
 
         fetchAliasGroups(newAppConfig, remoteJson);
@@ -1197,10 +1207,7 @@ public:
 #endif
 
         fetchRemoteFontConfig(newAppConfig, remoteJson);
-
-        AutoPtr<AppConfigMap> newConfig(new AppConfigMap(newAppConfig));
-        _conf.addWriteable(newConfig, PRIO_JSON);
-        newConfig->clear();
+        _persistConfig->reset(newAppConfig);
 
 #if ENABLE_FEATURE_LOCK
         CommandControl::LockManager::parseLockedHost(_conf);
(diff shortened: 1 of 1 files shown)


wsd/COOLWSD.cpp
```diff
@@ -1182,6 +1182,9 @@ class RemoteConfigPoll : public RemoteJSONPoll
 
     virtual ~RemoteConfigPoll() { }
 
+    // keeps track of remote config layer
+    AutoPtr<AppConfigMap> persistConfig = nullptr;
```
Ashod, line 1186: Is there a particular reason to have this as a public member *and* not to use the naming convention for member variables?

wsd/COOLWSD.cpp
```diff
@@ -1198,9 +1201,12 @@ class RemoteConfigPoll : public RemoteJSONPoll
...
 
-        AutoPtr<AppConfigMap> newConfig(new AppConfigMap(newAppConfig));
-        _conf.addWriteable(newConfig, PRIO_JSON);
-        newConfig->clear();
+        // delete old remote config layer before applying the new layer
+        if (!persistConfig.isNull())
+            persistConfig->clear();
+
+        persistConfig = new AppConfigMap(newAppConfig);
+        _conf.addWriteable(persistConfig, PRIO_JSON);
```
Ashod, line 1209: Are we sure this fixes the leak? It's not obvious to me what happens to the `AppConfigMap` instances we are adding in `_conf` by calling `addWriteable`.

It seems that `_conf` is taking ownership of the `AutoPtr`, so how is this different from the old solution, other than using a member variable instead of a local one?

Shouldn't we instead simply modify the `persistConfig` instance, instead of creating a new one? That is, we create it and add to `_conf` and after that we simply update it. So we never create more than one `AppConfigMap`. Wouldn't that fix the leak and make things simpler?

Ashod (commented):
Thanks @Rash419. Please see comments/questions.

wsd/COOLWSD.cpp
```diff
@@ -1533,6 +1558,10 @@ class RemoteConfigPoll : public RemoteJSONPoll
         }
         return booleanFlag.toString();
     }
+
+private:
+    // keeps track of remote config layer
+    AutoPtr<AppConfigMap> PersistConfig = nullptr;
```
Ashod, line 1564: Better, but why the new naming convention? We use `_camelCase` for class members.

Also, it's good to prefix the namespace (`Poco::` in this case). We use namespaces everywhere for a reason (it makes maintenance easier, avoid clashes, helps easily identify the context, etc.).

wsd/COOLWSD.cpp
```diff
@@ -1198,9 +1207,15 @@ class RemoteConfigPoll : public RemoteJSONPoll
 
         fetchRemoteFontConfig(newAppConfig, remoteJson);
 
-        AutoPtr<AppConfigMap> newConfig(new AppConfigMap(newAppConfig));
-        _conf.addWriteable(newConfig, PRIO_JSON);
-        newConfig->clear();
+        if (PersistConfig.isNull())
+        {
+            PersistConfig = new AppConfigMap(newAppConfig);
+            _conf.addWriteable(PersistConfig, PRIO_JSON);
```
Ashod, line 1213: Why not do this in the constructor, and avoid this check here?

wsd/COOLWSD.cpp
```diff
@@ -990,6 +990,15 @@ class AppConfigMap final : public Poco::Util::MapConfiguration
             setRaw(pair.first, pair.second);
         }
     }
+
+    void updateAppConfigMap(const std::map<std::string, std::string>& map)
+    {
+        clear();
```
Ashod, line 996: Nice!

But `update` doesn't imply "clear everything and replace with the new values". Do we really have to clear? Why can't we just add new entries and replace existing ones, without removing anything?

Notice that if you really need to call `clear()` for your case, then it's best to rename this helper, since this can be used in other places. Perhaps `reset`, or just `set`, is a good name.

Also, no need to repeat the class name in the member functions. `update()`, `reset()`, `set()`, are all good names, no need for `AppConfigMap` everywhere.

Ashod (commented):
Thank you @Rash419. It's better. Left some questions/notes.

Ashod (approved):
Great, thanks @Rash419!

PR #5511, opened 2022-11-23
Fix jsdialog spinfield jumping on hover

pedropintosilva:
Visible when using the following core change:
https://gerrit.libreoffice.org/c/core/+/142974

Note: no need to have generic CSS rules for spinfields
only for the sidebar

Change-Id: Id2c9df00d1067537adb9ec5e3db0eab8bd1b2290


eszkadev:
rowheight
for me doesn't work in chrome, no arrows and unit in the center

eszkadev (approved):
works in the end, I had to restart browser after using mobile mode :)

PR #5512, opened 2022-11-23
cypress tests: svg export filter: text fields

mcecchetti:
Some tests for checking that text fields in masterpages are replaced
with the expected content.
In the sample presentations are included the default masterpage text
fields, copy of default text fields created through copy and paste,
finally there are text fields created through the Impress insert text
field menu entry.
There are also small textfields for testing when the placeholder text
(that is <number>, <footer, ...) spans several lines.

Change-Id: I85fb98a723b3d7c446e166a6aebbb99558de7046


thebearon:
@mcecchetti is it clear why the new test is failing on CI?

pedropintosilva:
@mcecchetti should this be closed? Or are we waiting for a review?

mcecchetti:
@pedropintosilva 
Oh thanks for pointing this PR out. It was completely slipped out of my mind.
Here, the point is that fullscreen presentation cypress tests are completely disabled now.
So the first step would be to enable them again and get them passing.
Anyway there're tasks with a higher priority, actually. 
Maybe on next free hacking week I'll make a try.

PR #5514, opened 2022-11-24
private/mike/MathInSd

mikekaganski:
- Make sure that correct view is active during tile painting
- Add "Formula Object" button to Insert tab of notebookbar in Draw/Impress


mikekaganski:
Michael, could you please review specifically bc8bf064f2196463389905a48a3633b355b35e0e? I don't know if this is a correct approach at all - neither in terms of performance impact, nor in any other possible effects (is it even OK to set the view at a random place?).

timar:
This PR is almost 2 years old, and there is no progress. I don't think that it should be in the review queue that we check daily. Please re-open, when something happens.

PR #5517, opened 2022-11-24
Hyperlink warning: Use JSDialog instead of Vex

pedropintosilva:
Change-Id: I3339792d9a712efe28519fcb63dbe14002f8976c

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -711,6 +711,85 @@ L.Control.UIManager = L.Control.extend({
 		this.map.fire('hidebusy');
 	},
 
+	showHyperlinkWarn: function(url, action) {
+		console.debug(action);
+		var json = {
+			id: 'hyperlinkWarnDialog',
+			dialogid: 'hyperlinkWarnDialog',
+			type: 'dialog',
+			text: 'Are you sure you want to visit the following URL?',
+			title: 'Hyperlink',
+			jsontype: 'dialog',
+			enabled: true,
+			children: [
+				{
+					id: 'dialog-vbox1',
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -711,6 +711,85 @@ L.Control.UIManager = L.Control.extend({
 		this.map.fire('hidebusy');
 	},
 
+	showHyperlinkWarn: function(url, action) {
```
pedropintosilva, line 714: Better to have a generic showWarnDialog function where we start to use slowly to convert simple dialogs of that type, starting with Hyperlink warning dialog

browser/src/control/Control.UIManager.js
```diff
@@ -711,6 +711,85 @@ L.Control.UIManager = L.Control.extend({
...
+		console.debug(action);
+		var json = {
+			id: 'hyperlinkWarnDialog',
+			dialogid: 'hyperlinkWarnDialog',
+			type: 'dialog',
+			text: 'Are you sure you want to visit the following URL?',
+			title: 'Hyperlink',
+			jsontype: 'dialog',
+			enabled: true,
+			children: [
```
pedropintosilva, line 724: `responses` group was removed because we only use that when core expects an answer. Example in the repair document dialog we do need that see `Control.DocumentRepair.js` and the respective 
```javascript
			responses: [
				{
					id: 'ok',
					response: 1
				},
				{
					id: 'cancel',
					response: 0
				},
			],
```

pedropintosilva:
This PR should not be merged. I have added notes from my meeting with @eszkadev . He is taking this from here, thanks!

PR #5518, opened 2022-11-24
jsdialog: introduce showModal

eszkadev:
showModal is going to be replacement for vex

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -753,6 +753,92 @@ L.Control.UIManager = L.Control.extend({
 		app.socket._onMessage({textMsg: 'jsdialog: ' + JSON.stringify(json), callback: builderCallback});
 	},
 
+	// Modals
+
+	/// shows modal dialog
+	/// json - JSON for building the dialog
+	/// callbacks - array of { id: widgetId, type: eventType, func: functionToCall }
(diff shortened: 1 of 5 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -753,6 +753,92 @@ L.Control.UIManager = L.Control.extend({
...
+			'init_focus_id': 'response',
+			children: [
+				{
+					id: 'info-modal-container',
+					type: 'container',
+					vertical: true,
+					children: [
+						{
+							id: 'label1',
+							type: 'fixedtext',
```
eszkadev, line 802: these label1 and label2 could be named differently so we can distinguish them and style properly on mobile...

pedropintosilva (approved):

pedropintosilva:
Thanks @eszkadev I also wonder if we could , either set the maxHeight for what it will come to be the main parent mobilewizard panel or if we should instead have an additional condition as we have for `if (data.type === 'snackbar')`:`Control.MobileWizardWindow.js`

I basically would like to set the maxHeight within `L.Control.MobileWizardWindow` to `auto` when we are in presence of `modalpopup` type otherwise it's always `maxHeight: '45vh',` which mean it's quite bigger than our hyperlink

PR #5523, opened 2022-11-26
Reduce Duplication in WebSocketHandler

Ashod:
- wsd: simplify WebSocketHandler construction
- wsd: avoid locking weak_ptr when possible

### DIFF
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -93,7 +93,7 @@ public:
 
         // As a server, respond with 101 protocol-upgrade.
         assert(!_isClient);
-        upgradeToWebSocket(*socket, request);
+        upgradeToWebSocket(socket, request);
     }
 
     /// Status codes sent to peer on shutdown.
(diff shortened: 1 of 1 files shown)


net/WebSocketHandler.hpp
```diff
@@ -60,17 +60,21 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
-    WebSocketHandler(bool isClient, bool isMasking) :
+    WebSocketHandler(bool isClient, bool isMasking)
+        :
 #if !MOBILEAPP
-        _lastPingSentTime(std::chrono::steady_clock::now()),
-        _pingTimeUs(0),
-        _isMasking(isClient && isMasking),
-        _inFragmentBlock(false),
-        _key(isClient ? PublicComputeAccept::generateKey() : std::string()),
+        _lastPingSentTime(std::chrono::steady_clock::now() -
```
Ashod, line 66: This isn't relevant for `isClient=true` case, since only the server (i.e. `isClient=false`) sends pings.

PR #5525, opened 2022-11-27
Improved Save, Load, and State-Dump Processing

Ashod:
Each commit describes what it does. These are unrelated, but small enough to get reviewed together.

- wsd: more flexible 'save' command parsing
- wsd: improve DocBroker state dumping logic
- wsd: faster forwardToChild


vmiklos (approved):
I don't see anything obviously problematic with this.

PR #5540, opened 2022-11-29
browser: adjust print/PDF export settings

thebearon:
- don't print to PDF forms, it's unnecessary, and buggy at the moment,
- don't print/export comments as notes, they aren't relevant.

Change-Id: I511940f1a97e1c4a6ed6afb87165c9854df56b78

* Resolves: ...
* Target version: master 
### Checklist


vmiklos (approved):
Looks good till we figure out why form fields cause a problem.

PR #5541, opened 2022-11-30
Avoid Repeated ClientSession Lookup

Ashod:
Reduce the number of lookups in `_sessions` when we already have the ClientSession instance.

- wsd: pass ClientSession to finalRemoveSession
- wsd: pass ClientSession to disconnectSessionInternal
- wsd: pass ClientSession to checkAndUploadToStorage


vmiklos (approved):
Looks OK to me. Finally a PR which is mechanical enough that I can confidently review, even if I don't have too deep Online domain knowledge. ;-)

PR #5542, opened 2022-11-30
More Avoid Repeated ClientSession Lookup

Ashod:
- wsd: pass ClientSession to uploadAsToStorage
- wsd: pass ClientSession to sendUnoSave and simplify
- wsd: pass ClientSession to forwardToChild

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1365,10 +1365,21 @@ void DocumentBroker::checkAndUploadToStorage(const std::shared_ptr<ClientSession
             // If we have nothing to upload, do the rename now.
             if (needToUploadState == NeedToUpload::No)
             {
-                LOG_DBG("Renaming in storage as there is no new version to upload first.");
-                std::string uploadAsPath;
-                constexpr bool isRename = true;
-                uploadAsToStorage(_renameSessionId, uploadAsPath, _renameFilename, isRename);
+                const auto it = _sessions.find(_renameSessionId);
+                if (it == _sessions.end())
+                {
+                    LOG_ERR("Session with sessionId ["
+                            << _renameSessionId << "] not found to rename docKey [" << _docKey
+                            << "]. The document will not be renamed.");
+                    broadcastSaveResult(false, "Renaming session not found");
+                }
+                else
+                {
+                    LOG_DBG("Renaming in storage as there is no new version to upload first.");
+                    std::string uploadAsPath;
+                    constexpr bool isRename = true;
+                    uploadAsToStorage(it->second, uploadAsPath, _renameFilename, isRename);
+                }
 
                 endRenameFileCommand();
                 return;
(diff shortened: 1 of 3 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -1146,11 +1146,21 @@ void DocumentBroker::startRenameFileCommand()
 
     blockUI("rename"); // Prevent user interaction while we start renaming.
 
+    const auto it = _sessions.find(_renameSessionId);
```
Ashod, line 1149: This is needed because this is the session that started renaming. We could hold its session, but that seems problematic (we'd be extending its lifetime artificially). Instead, we just look up if that session still exists when we are ready to upload.

wsd/DocumentBroker.cpp
```diff
@@ -1344,10 +1354,21 @@ void DocumentBroker::checkAndUploadToStorage(const std::shared_ptr<ClientSession
             // If we have nothing to upload, do the rename now.
             if (needToUploadState == NeedToUpload::No)
             {
-                LOG_DBG("Renaming in storage as there is no new version to upload first.");
-                std::string uploadAsPath;
-                constexpr bool isRename = true;
-                uploadAsToStorage(_renameSessionId, uploadAsPath, _renameFilename, isRename);
+                const auto it = _sessions.find(_renameSessionId);
```
Ashod, line 1357: Same as the previous rename session lookup.

wsd/DocumentBroker.cpp
```diff
@@ -1642,10 +1652,21 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
             {
                 case DocumentState::Activity::Rename:
                 {
-                    LOG_DBG("Renaming in storage after uploading the last saved version.");
-                    std::string uploadAsPath;
-                    constexpr bool isRename = true;
-                    uploadAsToStorage(_renameSessionId, uploadAsPath, _renameFilename, isRename);
+                    const auto it = _sessions.find(_renameSessionId);
```
Ashod, line 1655: Same.

vmiklos (approved):
Hm, it's a bit tricky to know if we don't extend the lifetime of a session with this, but I don't see any obviously wrong conversion. Let's hope we have good enough test coverage for this.

Ashod:
> Hm, it's a bit tricky to know if we don't extend the lifetime of a session with this, but I don't see any obviously wrong conversion. Let's hope we have good enough test coverage for this.

Good question. The only storage is `_sessions` in DocBroker, which should have the ClientSession instance that is invoking DocBroker in it (we don't have temp sessions here). So if nothing is getting stored, the lifetime shouldn't be extended beyond the calls themselves, which either start in the DocBroker poll (via `_sessions` lookup) or through a network input (these are called from the ClientSession into DocBroker passing `client_from_this()`).

And, yes, the tests cover lifetime management quite well. We have a lot of tests around that, so I expect something will break if we mess up (though it might be timing sensitive, of course).

PR #5544, opened 2022-11-30
introduce indirection url endpoint to fetch routeToken 

Rash419:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -672,6 +672,69 @@ window.app = {
 		}, false);
 	}
 
+	// indirect socket to wrap the asyncness around fetching the routetoken from indirection url endpoint
+	global.IndirectSocket = function(uri) {
+		var that = this;
+		this.uri = uri;
+		this.binaryType = '';
+		this.unloading = false;
+		this.readyState = 0; // connecting
+		this.innerSocket = undefined;
+
+		this.onclose = function() {};
+		this.onerror = function () {};
+		this.onmessage = function () {};
+		this.onopen = function () {};
+
+		this.close = function() {
+			this.innerSocket.onerror = function () {};
+			this.innerSocket.onclose = function () {};
+			this.innerSocket.onmessage = function () {};
+			this.innerSocket.close();
+		};
+
+		this.send = function(msg) {
+			this.innerSocket.send(msg);
+		};
+
+		this.setUnloading = function() {
+			this.unloading = true;
+		};
+
+		var http = new XMLHttpRequest();
+		http.open('GET', global.indirectionUrl + '?WOPISrc=' + global.wopiSrc, true);
+		http.responseType = 'json';
+		http.addEventListener('load', function() {
+			if (this.status === 200) {
+				global.routeToken = http.response.route_token;
+				//add routeToken in uri
+				var startIndex = that.uri.indexOf('&compat');
+				var uri = that.uri.substr(0, startIndex) + '&RouteToken=' + global.routeToken + that.uri.substr(startIndex);
+				that.innerSocket = new WebSocket(uri);
+				that.innerSocket.binaryType = that.binaryType;
+				that.innerSocket.onerror = function() {
+					that.readyState = that.innerSocket.readyState;
+					that.onerror();
+				};
+				that.innerSocket.onclose = function() {
+					that.readyState = 3;
+					that.onclose();
+				};
+				that.innerSocket.onopen = function() {
+					that.readyState = 1;
+					that.onopen();
+				};
+				that.innerSocket.onmessage = function(e) {
+					that.readyState = that.innerSocket.readyState;
+					that.onmessage(e);
+				};
+			} else {
+				window.app.console.debug('Indirection url: error on incoming response ' + this.status);
+			}
+		});
+		http.send();
+	};
+
 	global.createWebSocket = function(uri) {
 		if ('processCoolUrl' in window) {
 			uri = window.processCoolUrl({ url: uri, type: 'ws' });
@@ -680,7 +743,10 @@ window.app = {
 		if (global.socketProxy) {
 			window.socketProxy = true;
 			return new global.ProxySocket(uri);
-		} else {
+		} else if (global.indirectionUrl != '') {
+			window.indirectSocket = true;
+			return new global.IndirectSocket(uri);
+		} else {			
 			return new WebSocket(uri);
 		}
 	};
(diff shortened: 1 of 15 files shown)


browser/js/global.js
```diff
@@ -675,6 +675,68 @@ window.app = {
...
+				that.innerSocket.onclose = function() {
+					that.readyState = 3;
+					that.onclose();
+				};
+				that.innerSocket.onopen = function() {
+					that.readyState = 1;
+					that.onopen();
+				};
+				that.innerSocket.onmessage = function(e) {
+					that.onmessage(e);
```
mmeeks, line 731: Looks great  I  might be inclined to clone the 'readyState' across both here (each time before the onmessage)  and also in the 'onerror()' callback just before we emit it =)

browser/js/global.js
```diff
@@ -675,6 +675,68 @@ window.app = {
...
+		};
+
+		var http = new XMLHttpRequest();
+		http.open('GET', global.indirectionUrl, true);
+		http.responseType = 'json';
+		http.addEventListener('load', function() {
+			if (this.status === 200) {
+				global.routeToken = http.response.route_token;
+				//add routeToken in uri
+				var startIndex = that.uri.indexOf('&compat');
```
mmeeks, line 715: I would really prefer to get the full URL back from the server including the routing token - rather than adding bits into the URI like this. ie. lets do this on the server side. Can we do that ? =)

browser/src/core/Socket.js
```diff
@@ -1187,6 +1187,19 @@ app.definitions.Socket = L.Class.extend({
 				this._map.zotero._onMessage(subTextMsg);
 			}
 		}
+		else if (textMsg.startsWith('updateroutetoken') && window.indirectionUrl != '') {
+			var http = new XMLHttpRequest();
```
mmeeks, line 1191: This looks fine - I wonder though if we can have a method on our Indirect websocket'y thing - that allows the refresh there ? so we can handle all that in one place (?) and again same comment on fetching the whole URL - in the future I'd hope we can switch from one server to a completely different one around the world with a different domain name =)

wsd/Admin.cpp
```diff
@@ -294,6 +295,10 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
         // Let's send back the current log levels in return. So the user can be sure of the values.
         sendTextFrame("channel_list " + _admin->getChannelLogLevels());
     }
+    else if (tokens.equals(0, "clusterresize"))
+    {
+        COOLWSD::alertAllUsersInternal("updateroutetoken");
```
mmeeks, line 300: So - looks like this might work -but- it will create a huge volume of requests - eg. 100k of them for a big server.
In almost all of these cases - we want to do the same thing for all of these users: to update all the client's routing token to a different string.

Can we not let this server know what its new routing token/URL is -and- then just tell all the clients what it is - without needing to migrate anything and/or send a server request to ask for a new routing token ?

ie. we just update our internal state / URL in the JS client without needing to ask the server again - and without re-connecting, and/or doing anything expensive or disruptive to the websocket - we shouldn't even need a refresh.

Thanks

wsd/Admin.cpp
```diff
@@ -294,6 +295,10 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
         // Let's send back the current log levels in return. So the user can be sure of the values.
         sendTextFrame("channel_list " + _admin->getChannelLogLevels());
     }
+    else if (tokens.equals(0, "clusterresize"))
```
mmeeks, line 298: I'd expect to have something more specific here: "updateroutetoken <URL>" or whatever message - such that we can tell all of our clients about their new URL - without them needing to re-connect to the server.

mmeeks (changes requested):
Looks a lot better in the JS ! - we're really getting there.

I think the 'clusterresize' concept is not what we want.

IIRC we had some notes on this from our last chat ? =)

wsd/Admin.cpp
Rash419: Thanks, I have updated the PR now COOL recieves a JSON with mapping of new serverId and routeToken , then based on that map COOL can update the routeToken/URL

wsd/COOLWSD.cpp
```diff
@@ -1551,6 +1555,33 @@ class RemoteConfigPoll : public RemoteJSONPoll
...
+        }
+
+        if (monitors.isNull() || monitors->size() == 0)
+        {
+            LOG_INF("Overriding monitors failed because array is empty or "
+                    "null");
+            return;
+        }
+
+        for (size_t i = 0; i < monitors->size(); i++)
```
mmeeks, line 1580: Does this suffer from an ever growing number of duplicate monitors ? if the list shrinks does this work nicely or leave older things away ? (or perhaps newAppConfig is just the new data we're over-writing with?)

wsd/Admin.cpp
```diff
@@ -294,6 +298,26 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
...
+        if (JsonUtil::parseJSON(tokens[1], object))
+        {
+            const std::string routeToken =
+                JsonUtil::getJSONValue<std::string>(object, Util::getProcessIdentifier());
+            if (routeToken.empty())
+            {
+                LOG_ERR("Failed to update the route token, invalid serverId to routeToken json : "
+                        << tokens[1]);
+            }
+            COOLWSD::alertAllUsersInternal("updateroutetoken " + routeToken);
```
mmeeks, line 314: Looks fine - I wonder if we need to store the routeToken internally for our server (?) might be nice to have it around so we can dump it with a dumpState() method (?)

browser/js/global.js
mmeeks: I'd use the that.readyState; rather than '1' in case there is some dis-connecting / other state we want to clone here I think; in general rather than hard-coding it it would be nice to just copy it from the underlying websocket I think (?) then any odd behavior there will be mirrored in our wrapper (?)

browser/src/core/Socket.js
```diff
@@ -1187,6 +1187,9 @@ app.definitions.Socket = L.Class.extend({
 				this._map.zotero._onMessage(subTextMsg);
 			}
 		}
+		else if (textMsg.startsWith('updateroutetoken') && window.indirectionUrl != '') {
```
mmeeks, line 1190: Looks good - do we have a wsd/protocol.txt update to reflect this ? would be nice to update the docs there for both the Admin and the client websocket.

mmeeks (commented):
Looks better =) a few more questions, but getting there !

Rash419:
closing this in favour of : https://github.com/CollaboraOnline/online/pull/5649

PR #5546, opened 2022-11-30
Delete row/column break from menu

eszkadev:
This feature didn't have any visual indicator


mmeeks (approved):
Looks good to me - I think we had a couple of these buttons in the old-style desktop menus as well as for mobile =)

Thanks !

timar:
CodeQL scanning - action / Analyze (cpp) failure has nothing to do with this patch to Javascript files.

PR #5548, opened 2022-11-30
Initial WASM support for Online

thorstenb:
* Target version: master 

### Summary

Get initial cross-building of Online + Core as a WebAssembly binary going (see https://wiki.documentfoundation.org/Development/WASM for the core LibreOffice effort).

Have a small sample app, heavily borrowing from Android and Gtk viewer, at `wasm/wasmapp.cpp`

### TODO

- [ ] still wasmapp.cpp needs to start doing something useful

### Checklist


tml1024:
Ah, this is obsolete now, I been doing cherry-picking in much smaller batches. It is far from trivial to get the cherry-picks to pass CI. Some of the commits to feature/wasm for instance use C++ features that some of the older toolchains used in CI don't recognise.

Issue #5420, opened 2022-11-02
icon size use css var's


Andreas-Kainz:
I would suggest var's for icon sizes similar to how LibO did it

Icon size
- Toolbar
- Notebookbar (`standard` + `bigtoolitem`)
- Sidebar

And in addition I would like to add some var's for padding stuff around icon sizes. So we can switch between touch and mouse devices.

Andreas-Kainz:
@pedropintosilva any suggestion?

Ezinnem:
This can be closed now

Issue #5395, opened 2022-10-31
Tabbed view: bigtoolitem has decreased in size


pedropintosilva:
**Describe the bug**
The `bigtoolitem` went from `32px` to `24px`. 

**To Reproduce**
1. View > tabbed view
2. Look at the icons that have label underneath 
3. Observe their size (they are not bigger than the icons with inline labels)

**Expected behavior**
`bigtoolitem` should be bigger than the icons with inlined labels

**Actual behavior**
Has exactly the same size as the rest (24px)

**Screenshots**

https://user-images.githubusercontent.com/65948705/198964693-5a97cd26-bb81-465a-ae0b-f756f06c8844.mp4

**Desktop (please complete the following information)**
Master

**Additional context**
`.hasnotebookbar .ui-content .unobutton` is not being applied. Plus, probably worth to change the way we are targetting those buttons from CSS side

```html
<div class="unotoolbutton notebookbar ui-content unospan-Insert unoInsertGraphic has-label has-dropdown" id="InsertGraphic1" tabindex="0"><img class="ui-content unobutton" src="lc_insertgraphic.svg" id="InsertGraphic1img" alt="InsertGraphic1"><span class="ui-content unolabel">Image</span><i class="unoarrow"></i></div>
```

sharmadhiraj86:
Hii @pedropintosilva, I would like to solve this issue. Please assign me this issue.

2022H1030042G:
Hi @pedropintosilva, I would like to work on this issue. Can you assign it to me?

Andreas-Kainz:
Nobody has to assign to an issue. Go for it and submit a PR.

pedropintosilva:
@sharmadhiraj86 are you still interested in working on this? Otherwise I can assign to @2022H1030042G

2022H1030042G:
Can you share the details where do I have to make the changes (under which directory/file)?

pedropintosilva:
@2022H1030042G the first step would be to have Collabora Online working locally on  your computer so you can work and test your changes : https://collaboraonline.github.io/post/build-code/ do you have it? Alternatively you can use GitPod are you familiar with that?

2022H1030042G:
@pedropintosilva, I'm using Gitpod for now.

pedropintosilva:
ok, so the file is notebookbar.css and it seems  `.hasnotebookbar .ui-content .unobutton` rules are being overruled

2022H1030042G:
[...]

PR #5561, opened 2022-12-04
icon size use css variables #5420

Andreas-Kainz:
add variables to color-scheme.css file
replace hardcoded sizes with var's at
- notebookbar
- toolbar
- sidebar

Change-Id: I8e6a0e76ea7f32bda3a8bcd781e75b06d800ec88

* Resolves: #5420 and #5395 
* Target version: master

### DIFF
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -45,6 +45,9 @@
 	--default-height: 24px;
 	--header-height: 38px;
 
+	--icon-size: 24px;
+	--icon-size-big: 32px;
+
 	/* Annotations */
 	--annotation-input-size: 240px;
 }
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -785,7 +785,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 /* color selector */
 .unotoolbutton .selected-color {
 	height: 5px;
-	width: 24px;
+	width: var(--icon-size);
 	position: relative;
 	top: -6px;
 	border: 1px solid var(--color-border-dark);
@@ -854,6 +854,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 /* toolbox */
+.no-label.has-dropdown,
 .jsdialog.toolbox {
 	display: flex;
 	align-items: center;
@@ -1137,8 +1138,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 /* toolbuttons */
 
 .jsdialog .ui-content.unobutton {
-	width: 24px;
-	height: 24px;
+	width: var(--icon-size);
+	height: var(--icon-size);
 	margin-inline-end: 5px;
 	vertical-align: middle;
 }
@@ -1208,8 +1209,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 .formulabar.unotoolbutton img, .formulabar.ui-pushbutton {
-	height: 24px !important;
-	width: 24px !important;
+	height: var(--icon-size) !important;
+	width: var(--icon-size) !important;
 	min-width: unset !important;
 }
 
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -81,8 +81,8 @@
 }
 
 .notebookbar-shortcuts-bar .notebookbar.unotoolbutton {
-	height: 27px;
-	width: 24px;
+	height: var(--icon-size);
+	width: var(--icon-size);
 	display: flex;
 	margin: 0px !important;
 	align-items: center;
@@ -202,9 +202,6 @@
 /* unobuttons */
 
 .hasnotebookbar .ui-content .unobutton {
-	width: 32px;
-	height: 32px;
-	margin-inline-end: 5px;
 	vertical-align: middle;
 }
 
@@ -223,13 +220,13 @@
 }
 
 .unotoolbutton.notebookbar .unobutton {
-	width: 24px !important;
-	height: 24px !important;
+	width: var(--icon-size);
+	height: var(--icon-size);
 	margin-inline-end: 0px !important;
 }
 
 .has-dropdown--color.notebookbar {
-	height: 24px !important;
+	height: var(--icon-size);
 }
 
 .unotoolbutton.notebookbar .unolabel {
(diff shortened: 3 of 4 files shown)


Andreas-Kainz:
In general there are more hardcoded icon sizes, but in general it would be only usefull if everything else toolbar, notebookbar, ... size (height) would also change on desktop, mobile, ... so a huge amount of work where you can break a lot of stuff and less output. So from my point of view, the PR is finished.

browser/css/notebookbar.css
```diff
@@ -81,8 +81,8 @@
 }
 
 .notebookbar-shortcuts-bar .notebookbar.unotoolbutton {
-	height: 27px;
```
pedropintosilva, line 84: but in here it seems we are changing from 27 to 24 . Does that affect or cause regressions? I have tested and it seems that is save

browser/css/notebookbar.css
```diff
@@ -202,9 +202,6 @@
 /* unobuttons */
 
 .hasnotebookbar .ui-content .unobutton {
-	width: 32px;
-	height: 32px;
-	margin-inline-end: 5px;
```
pedropintosilva, line 207: Better to add this to another commit. It is unrelated to this commit. Let's please add another with this in and why it is not needed (outdated rules,etc, now being overruled by more recent changes in notebookbar.css:77 etc)

browser/css/color-palette.css
```diff
@@ -45,6 +45,9 @@
 	--default-height: 24px;
 	--header-height: 38px;
 
+	--icon-size: 24px;
+	--icon-size-big: 32px;
+
 	/* Annotations */
```
pedropintosilva, line 51: Thanks!

browser/css/jsdialogs.css
```diff
@@ -785,7 +785,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 /* color selector */
 .unotoolbutton .selected-color {
 	height: 5px;
-	width: 24px;
+	width: var(--icon-size);
```
pedropintosilva, line 788: Thanks

browser/css/jsdialogs.css
```diff
@@ -854,6 +854,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 /* toolbox */
+.no-label.has-dropdown,
 .jsdialog.toolbox {
```
pedropintosilva, line 858: This seems to be unrelated to this commit. Could you please add a different one with that change in and with the respective commit message?

browser/css/jsdialogs.css
```diff
@@ -1137,8 +1138,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 /* toolbuttons */
 
 .jsdialog .ui-content.unobutton {
-	width: 24px;
-	height: 24px;
+	width: var(--icon-size);
+	height: var(--icon-size);
```
pedropintosilva, line 1142: Thanks

browser/css/jsdialogs.css
```diff
@@ -1208,8 +1209,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 .formulabar.unotoolbutton img, .formulabar.ui-pushbutton {
-	height: 24px !important;
-	width: 24px !important;
+	height: var(--icon-size) !important;
+	width: var(--icon-size) !important;
```
pedropintosilva, line 1213: Thanks

browser/css/notebookbar.css
```diff
@@ -81,8 +81,8 @@
 }
 
 .notebookbar-shortcuts-bar .notebookbar.unotoolbutton {
-	height: 27px;
-	width: 24px;
+	height: var(--icon-size);
+	width: var(--icon-size);
```
pedropintosilva, line 85: Thanks

browser/css/notebookbar.css
```diff
@@ -223,13 +220,13 @@
...
-	width: 24px !important;
-	height: 24px !important;
+	width: var(--icon-size);
+	height: var(--icon-size);
 	margin-inline-end: 0px !important;
 }
 
 .has-dropdown--color.notebookbar {
-	height: 24px !important;
+	height: var(--icon-size);
```
pedropintosilva, line 229: Thanks

browser/css/notebookbar.css
```diff
@@ -298,7 +295,8 @@
 }
 
 .unotoolbutton.notebookbar.has-label:not(.inline) img {
-	width: 24px !important;
+	width: var(--icon-size-big);
+	height: var(--icon-size-big);
```
pedropintosilva, line 299: This seems to be unrelated to this commit since we are changing value from 24 to 32px

browser/css/notebookbar.css
```diff
@@ -337,15 +335,6 @@
...
 
-#clearFormatting.notebookbar div img, #FormatPaintbrush.notebookbar div img {
-	width: 24px !important;
-	height: 24px !important;
-}
-
-#clearFormatting.notebookbar, #FormatPaintbrush.notebookbar{
-	height: 24px !important;
-}
-
```
pedropintosilva, line 348: Seems to be unrelated to this commit. Could we please have this in a different commit?

browser/css/toolbar.css
```diff
@@ -380,9 +380,9 @@ td[id^='tb_editbar_item_sidebar']{
 	margin: 0 !important;
 }
 
-.w2ui-tb-image {
-	width: 24px !important;
-	height: 24px !important;
+.w2ui-toolbar table.w2ui-button .w2ui-icon {
+	width: var(--icon-size);
+	height: var(--icon-size);
```
pedropintosilva, line 385: Thanks, but this seems to no only change to css vars but also going from affecting only related children from w2ui-toolbar to every single w2ui-tb-image no matter the parent and even if that are not a w2ui-icon. Was this intended?

browser/css/toolbar.css
```diff
@@ -1162,8 +1162,8 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
 
 #userListHeader .user-info,
 #userListHeader .avatar-img {
-	width: 24px;
-	height: 24px;
+	width: var(--icon-size);
+	height: var(--icon-size);
```
pedropintosilva, line 1166: Thanks!

pedropintosilva (changes requested):

pedropintosilva:
polite ping

Andreas-Kainz:
There is already btn size value in use. This request isn't needed any more

PR #5566, opened 2022-12-05
Don't scroll when cursor still visible on the screen

eszkadev:
When we have 2 users:
A is typing on the page 0
B is at the bottom of the document

We scrolled view B when page overflow happened what caused "flickering". This patch prevents us from scrolling when cursor is still on the screen and visible.


pedropintosilva (approved):
I have tested this against staging.eu and I can confirm:

With this PR operating the same document with multiple users (as described in the commit message) doesn't cause cursor glitch (jumping cursor)

Thank you Szymon!

PR #5567, opened 2022-12-05
Remove the 'Insert Bibliography Entry' feature

timar:
LibreOffice has its own solution to handle bibliography. It is outdated, and reportedly everyone prefers Zotero or similar solutions. Moreover in Online this feature does not work, because the bibliography database is a file in the jail that is not retained, and because tunneled dialogs have serious UX problems.

Change-Id: Ibf6195c0265c767b3acf94012ee3f82339cf34d3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Thanks Andras ! =)

mmeeks:
I guess it's a translation horror - but the "Table of contents, index or bibliography" header in that dialog may need a tweak (?) =)

PR #5571, opened 2022-12-07
Update dependencies.

gokaysatir:
Change-Id: I1ce4c803dd29f65a12f3ed8de323f131353c77cd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,7 @@ task:
   name: "FreeBSD 13"
   freebsd_instance:
     matrix:
-      - image_family: freebsd-13-0-snap
+      - image_family: freebsd-13-1-snap
   deps_script:
     - sed -i.bak -e 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
     - env ASSUME_ALWAYS_YES=yes pkg update -f


timar:
@arrowd do you know what's wrong the FreeBSD CI?

arrowd:
`image_family` should be changed to `freebsd-13-1-snap`, because 13.0 isn't supported anymore. Let's see if this fix the issue.

gokaysatir:
Okay @arrowd thanks for checking :)

arrowd:
This is a bug on our side, which is already fixed in Ports: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268098

It'd take several days for this fix to reach binary packages, though, so the CI will stay broken for some time.

gokaysatir:
> This is a bug on our side, which is already fixed in Ports: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268098
> 
> It'd take several days for this fix to reach binary packages, though, so the CI will stay broken for some time.

Okay we will ignore this error until then. Thank you.

arrowd:
This should be fixed now. Try re-running FreeBSD CI.

timar:
My PR https://github.com/CollaboraOnline/online/pull/5582 passed the FreeBSD 13 check.

arrowd:
Ah, this is an old CI run that is shown there. It should also pass now.

gokaysatir:
I rebased the commit, yes then this should also pass. Thank you @arrowd.

arrowd:
In its current form this PR only bumps FreeBSD version from 13.0 to 13.1. Your first version was also adding some dependencies.

gokaysatir:
That's right, after you told that the bug is on FreeBSD side, i removed them. Those dependencies were for checking to see if there is a missing dependency.

PR #5572, opened 2022-12-07
zotero: fieldmark should be read-only

timar:
Change-Id: If9d816932ff074c40ee159a8c36885c5bdda9fac

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks good, this matches <https://github.com/libreoffice/core/blob/eb5879f190a773a28148de8eaa6e4afa2f375eda/sw/qa/uibase/shells/shells.cxx#L282-L288>.

PR #5573, opened 2022-12-07
No need to cd back before exiting a shell script that is run as a com…

tml1024:
…mand

The autogen.sh script is not supposed to be sourced ("source autogen.sh"), but run as a command ("./autogen.sh"). After all, it has the executable bit set. Also, the function called "failed" in it does an "exit" at the end. That would be a rather rude thing to do as soon as something goes wrong if the script is sourced.

Change-Id: Ia0e4bbb2b9bed93fb4dba5c0f46a1760ec6e50d7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -3,7 +3,6 @@
 srcdir=`dirname $0`
 test -n "$srcdir" || srcdir=.
 
-olddir=`pwd`
 cd "$srcdir"
 
 function failed {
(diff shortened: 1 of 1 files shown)


tml1024:
FreeBSD failure is surely unrelated.

PR #5574, opened 2022-12-07
Remove 'Current Index' button from References tab

timar:
The label is misleading, it should read 'Update Index'. On the other hand this is a context sensitive command, it is active only when the cursor is in the Index field. So probably we do not need it on the tab bar. We have the context sensitive local menu entry 'Update Index'.

Change-Id: I9d77ed5079cb9dbbcd04f780620a7dbcae9804f7

image


pedropintosilva (changes requested):
I would prefer to keep that button in the tab specially because:
1. It already gets disabled when not in presence of an index
2. It gets enabled when within an index
3. Users that depend on always visible buttons (as opposed to contextual actions) my end up thinking that we just removed completely that feature 

Maybe we could just update the label? https://github.com/CollaboraOnline/online/pull/5575

PR #5584, opened 2022-12-12
Fix the insert hyperlink behaviour on Calc.

GulsahKose:
Change-Id: Ib8d5bedba3a6d9c05a26a73edd81b52cd94e3358

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
I have tested and it works:
- now when enter in cell edit mode and no test is selected by the user -> inserting an hyperlink does what is expected: triggers hyperlink dialog with empty inputfields
- also now it fixes the weird copy /past behaviour (before when in edit mode, copying (ctrl+c)would select and copy without user interaction (this is not expected)

kendy (commented):
A bit hard for me to see all the corner cases here, so if it fixes the problem, let's merge it ;-)  There is one thing though - all the other places that call 'this._map._clip.clearSelection();' guard that by 'if (this._map._clip)'.  I think we should do the same here.

pedropintosilva (approved):

PR #5585, opened 2022-12-12
Test Improvements

Ashod:
- wsd: test: support silent assertions on success
- wsd: test: add decodeURIComponent
- wsd: test: better assertion of test target
- wsd: test: minor cleanup in UnitBase
- wsd: test: proper test options
- wsd: test: support keepgoing option
- wsd: test: correctly calculate test timeout
- wsd: test: do not assert after test failure
- wsd: test: cleanup exitTest logs
- wsd: test: logging fixup

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -138,6 +138,12 @@ void UnitBase::initTestSuiteOptions()
             LOG_INF("Setting the 'filter' test option to [" << filter << ']');
             GlobalTestOptions.setFilter(filter);
         }
+        else if (pair.first == "keepgoing")
+        {
+            const bool keepgoing = pair.second == "1" || pair.second == "true";
+            LOG_INF("Setting the 'keepgoing' test option to " << keepgoing);
+            GlobalTestOptions.setKeepgoing(keepgoing);
+        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


common/Unit.cpp
```diff
@@ -116,21 +119,45 @@ UnitBase** UnitBase::linkAndCreateUnit(UnitType type, const std::string& unitLib
     return nullptr;
 }
 
-void UnitBase::filter()
+void UnitBase::initTestSuiteOptions()
 {
-    // For now, support only filtering.
     static const char* TestOptions = getenv("COOL_TEST_OPTIONS");
```
Ashod, line 124: This needs documenting, somewhere (but where?).

vmiklos (approved):
I don't see anything obviously problematic with this. Though if you use the "keep going" option, then probably you'll have to do something like googletest, which differentiates between an assert that keeps going on failure and one that stops on failure. Otherwise if you have this pattern:

```
Foo* foo = getFoo();
assert(foo);
foo->bar();
```

Then the old (non-keep-going) case gives a proper error message and no crash, while the new (keep-going) case results in a crash, which can be viewed as a regression in developer experience.

Ashod:
Good point. For now, I'm ignoring assertions if the test has failed. This way at least it's not too noisy, since it's very likely that subsequent assertions will also fire after a test fails.

But I like the idea. I think they use CHECK for the "log, but don't stop, if it fails." We can do something similar.

PR #5586, opened 2022-12-12
Misc Improvements

Ashod:
- wsd: use ClientSession in refreshLock
- wsd: flush sockets only when we have any
- wsd: support short-form names in STATE_ENUM
- wsd: shorter wait between cleanup checks on exit
- wsd: DocBroker dumps isStorageOutdated
- wsd: clang-tidy fix of initializing a string member


gokaysatir (approved):

PR #5587, opened 2022-12-12
Do Not Upload to Read-Only Storage

Ashod:
This fixes a corner-case where if we had a document that was not editable, but commenting were allowed (e.g. a PDF), we uploaded even when the Storage had UserCanWrite=false. Includes tests.

- wsd: set _isViewFileExtension for LocalStorage
- wsd: correct setting read-only and allow-comments
- wsd: move isWritable to Session and rename
- wsd: support isWritable flag in Session
- wsd: disable editing when Storage UserCanWrite=false
- wsd: test: add read-only storage tests


quikee (approved):
Looks fine from what I can see

PR #5590, opened 2022-12-12
wsd: improve rename action handling

Ashod:
This corrects the ending of renaming by
moving that logic only after we get the
async upload results (either success,
failure, or timeout).

It also cleans up a few cases, such as
not needing to special case renaming with
identical names. This ensures that
subsequent attempts are not ignored.

Change-Id: Ib7a432d8a3b6ef5448291763cd31cc844c997e7e


JaumePujantell (approved):
Looks good to me.

hcvcastro (commented):
Thanks. Looks good

hcvcastro (approved):

PR #5593, opened 2022-12-13
zotero: add zotero citation data json with field mark

lpranam:
Change-Id: I52710acb3a56e8930438d07cfe28be491e0161b7

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -196,6 +196,32 @@ L.Control.Zotero = L.Control.extend({
 		};
 	},
 
+	getCitationJSONString: function(item) {
+		var resultJSON = {};
+		resultJSON['citationID'] = Math.random().toString(36).substring(2,12);
+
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.Zotero.js
```diff
@@ -196,6 +196,32 @@ L.Control.Zotero = L.Control.extend({
 		};
 	},
 
+	getCitationJSONString: function(item) {
+		var resultJSON = {};
+		resultJSON['citationID'] = Math.round(Date.now() / 1000).toString();
```
timar, line 201: This does not look like a random string. We need a random string, 10 characters long, with characters from the [a-zA-Z0-9] range.

timar (changes requested):

browser/src/control/Control.Zotero.js
lpranam: updated

timar (approved):

PR #5599, opened 2022-12-15
jsdialog: align vertically grid content

eszkadev:


pedropintosilva (approved):
Thanks

pedropintosilva:
approved but a test is failing, also tested locally and it also fails but I think it's completely unrelated to this change:
`make -C cypress_test run-desktop spec=writer/image_operation_spec.js`

```
find
.leaflet-pane.leaflet-overlay-pane svg svg
90
assert
expected <svg> to exist in the DOM
91
assert
expected 394 to be a number
92
assert
expected 394 to be close to 139 +/- 10
```

eszkadev:
> 

I've got the same error on other PR - so seems to be not related

PR #5600, opened 2022-12-15
Zotero: setup minimal size for dialog

eszkadev:
to avoid large position change when loaded


browser/css/jsdialogs.css
```diff
@@ -1251,7 +1251,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
```
pedropintosilva, line 1251: Maybe we could add a default that is the same as the max for the list so the whole dialog doesn't flick/changes:

```css
#zoterolist {
	width: 800px;
}
```

pedropintosilva (approved):
Tested, it works. Maybe we could also add default width for the list?

browser/css/jsdialogs.css
eszkadev: done

PR #5605, opened 2022-12-17
Make export EPUB/PDF always with options dialog (Calc, Impress, Draw)

timar:
Change-Id: I86880ab3c54994554ab7f91921d028f962318e5a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2658,7 +2658,15 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (!data.command)
 			data.command = data.id;
 
+		if (data.inlineLabel !== undefined) {
+			var backupInlineText = builder.options.useInLineLabelsForUnoButtons;
+			builder.options.useInLineLabelsForUnoButtons = data.inlineLabel;
+		}
+
 		var control = builder._unoToolButton(parentContainer, data, builder);
+
+		if (data.inlineLabel !== undefined)
+			builder.options.useInLineLabelsForUnoButtons = backupInlineText;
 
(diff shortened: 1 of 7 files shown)


timar:
@eszkadev I found an issue with the non grouped download icons in case of Writer, therefore I did not change Control.NotebookbarCalc.js, Control.NotebookbarDraw.js, and Control.NotebookbarImpress.js.
There are no icons. Shoud we duplicate icons with the new names or is there a better workaround? I hope there is, because otherwise we have to duplicate icons in themes as well.
image

eszkadev:
> @eszkadev I found an issue with the non grouped download icons in case of Writer, therefore I did not change Control.NotebookbarCalc.js, Control.NotebookbarDraw.js, and Control.NotebookbarImpress.js. There are no icons. Shoud we duplicate icons with the new names or is there a better workaround? I hope there is, because otherwise we have to duplicate icons in themes as well. image

@timar yes, we have aliasing for icons:
https://github.com/CollaboraOnline/online/blob/6d1b392b7ef9ea529708b653240125ab2871af9c/browser/src/control/Control.JSDialogBuilder.js#L2331

timar:
I added the icon aliases, but the buttons do not have labels, only tooltips. Probably it's acceptable.
image
I will add them to Calc, Draw and Impress, too.

timar:
On the other hand, it looks a bit ugly in Calc:
image

eszkadev:
> I added the icon aliases, but the buttons do not have labels, only tooltips. Probably it's acceptable

Fixed missing text. inlineLabel property will do the job

eszkadev (approved):

PR #5612, opened 2022-12-20
Reduce Warning Logs

Ashod:
These patches reduce the number of superfluous WRN log entries significantly. Using the unit-tests as one reference, the number of warning log entries go down from around 2800 to around 170. The goal is to remove noisy logs without sacrificing the usefulness of the warnings, as they were designed. The commit logs may have more information.

- wsd: test: better post-close check in UnitCopyPaste
- wsd: preserve close reason when disconnected from Kit
- wsd: better log levels
- wsd: wait InitialPingDelayMicroS before first ping
- wsd: send WS close frame only when we have a socket
- wsd: do not warn of expected DocBroker disconnection
- wsd: support flushing control in WSProcess::sendFrame
- wsd: request Kit termination early when unloading
- wsd: do not send duplicate 'disconnect' to the Kit
- wsd: smarter warnings and errors on kit disconnection
- wsd: do not warn when waking up dead poll-thread at shutdown
- wsd: asyncRequest needs only wakeup its own poll thread
- wsd: test: cleanup redundant wakupWorld after sending a command

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -1114,7 +1114,10 @@ namespace Util
 
     void forcedExit(int code)
     {
-        LOG_FTL("Forced Exit with code: " << code);
+        if (code)
+            LOG_FTL("Forced Exit with code: " << code);
+        else
+            LOG_INF("Forced Exit with code: " << code);
         Log::shutdown();
 
 #if CODE_COVERAGE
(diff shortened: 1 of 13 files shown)


gokaysatir (approved):

PR #5367, opened 2022-10-24
browser: fix indent in package.json

vmiklos:
Signed-off-by: Miklos Vajna <vmiklos@collabora.com>
Change-Id: Icdba3a955783d34dc851876ee47cf15c949269ab

PR #5618, opened 2022-12-22
Misc Environment

Ashod:
On every build, nodejs (v12.22.9 for me) re-sorts
the packages in package.json. I expect this is
the case for everyone, so best to commit it sorted.

- gitignore: differentiate files from dirs
- browser: package.json always get resorted anyway


vmiklos (approved):
Makes sense to me. Similar to #5367

gokaysatir (approved):
Thanks @Ashod :)

PR #5619, opened 2022-12-22
wsd: make sessions readonly on disk errors

Ashod:
Change-Id: I0c59c4fb3c99dee0f6e18850ca795467a9421b33


mmeeks (approved):
Hmm - I guess we can do this; why not.

PR #5625, opened 2022-12-26
browser: menu: ignore menu item with empty text

hcvcastro:
if command "None" and empty item text, there is
an unhandled exception, and no grammar suggestions are shown.

Change-Id: Ie2c832a2a3ff7fe9944024fa003370819b16a432


gokaysatir (approved):
Thanks @hcvcastro :)

PR #5213, opened 2022-09-07
improve HelmChart

genofire:
Make this HelmChart to state of the Art ;)

with nice optional Features:
- Monitoring with Prometheus:
  - ServiceMonitor
  - Alerting (with default PrometheusRules)
  - Grafana Dashboards

- Dynamic Config reloading:
   - extra nginx-pod (+service, + ingress)
   - json will be generated from values.yaml into a ConfigMap

---

After Merge, please Release it to https://artifacthub.io/

PR #5627, opened 2022-12-27
updated helm chart

Rash419:
- now works with default configuration
- updated readme
- fixed unknown field "spec.terminationGracePeriodSeconds" in deployment.yaml
  error. This field should be defined in pod spec not in deployment spec
- improved how aliasgroups are defined in helm chart
- change default to kubernetes "Deployment" instead of "StatefulSet"
- added support for kubernetes hostAliases

Change-Id: Ief1e2b3f0ea130425853babbd28462d48e50280f

* Resolves: # 
* Target version: master

### DIFF
--- a/kubernetes/helm/collabora-online/templates/configmap.yaml
+++ b/kubernetes/helm/collabora-online/templates/configmap.yaml
@@ -12,9 +12,9 @@ data:
   {{- end }}
   server_name: {{ .Values.collabora.server_name }}
   {{- range $k,$v := .Values.collabora.aliasgroups }}
-  {{- $alias := $v.domain }}
+  {{- $alias := $v.host }}
   {{- if $v.aliases }}
-  {{- $alias := printf "%s|https://%s:443" $alias (join ":443|https://" $v.aliases ) }}
+  {{- $alias := printf "%s,%s" $alias (join "," $v.aliases) }}
   aliasgroup{{ add $k 1 }}: {{ $alias }}
   {{- else }}
   aliasgroup{{ add $k 1 }}: {{ $alias }}
--- a/kubernetes/helm/collabora-online/templates/deployment.yaml
+++ b/kubernetes/helm/collabora-online/templates/deployment.yaml
@@ -6,7 +6,6 @@ metadata:
   labels:
     {{- include "collabora-online.labels" . | nindent 4 }}
 spec:
-  terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
   minReadySeconds: {{ .Values.deployment.minReadySeconds }}
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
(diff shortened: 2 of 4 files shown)


lpranam (approved):

kubernetes/helm/collabora-online/templates/configmap.yaml
```diff
@@ -12,9 +12,9 @@ data:
   {{- end }}
   server_name: {{ .Values.collabora.server_name }}
   {{- range $k,$v := .Values.collabora.aliasgroups }}
-  {{- $alias := $v.domain }}
+  {{- $alias := $v.host }}
```
genofire, line 15: This is a breaking change - was it necessary?

kubernetes/helm/collabora-online/templates/deployment.yaml
```diff
@@ -34,6 +33,17 @@ spec:
         {{- include "collabora-online.selectorLabels" . | nindent 8 }}
         type: main
     spec:
+      {{- if .Values.deployment.hostAliases }}
+      hostAliases:
```
genofire, line 37: Why so complianced, following should be the some with one line
```yaml
hostAliases: {{- toYaml .Values.deployment.hostAliases | nindent 6 }}
```

genofire:
There is / was #5213 why solves a some of that issues ....

kubernetes/helm/collabora-online/templates/configmap.yaml
Rash419: Not a necessary but. Wanted to be consistent with how alias_groups are defined everywhere . Sure if it breaks let me change it

kubernetes/helm/collabora-online/templates/deployment.yaml
Rash419: Thanks

kubernetes/helm/collabora-online/templates/configmap.yaml
Rash419: https://github.com/CollaboraOnline/online/pull/5632

kubernetes/helm/collabora-online/templates/deployment.yaml
genofire: Yaml does not need quoting -> and if golang / helm could parse it - golang / kubernetes will it parse it on the same way.

kubernetes/helm/collabora-online/templates/configmap.yaml
genofire: it is okay - i was just wondering ...

Rash419: @genofire should I revert the change or no ?

kubernetes/helm/collabora-online/templates/deployment.yaml
Rash419: I see , you mean quotes are not necessary ? what about ip then ? It should be list of ip and inside each element of ip list, there is list of hostnames. sorry If I didnot understand your point 
https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/

Rash419: ok thanks. I will update it after testing it

PR #5630, opened 2022-12-29
zotero: in style dialog show selected style

lpranam:
Change-Id: I420c5f22f8b4aa880f202fe8b0d981f091c1dfa7

* Target version: master 

### Checklist


lpranam:
maybe in future, we can automatically scroll to the selected entry when dialogue is opened

timar:
> maybe in future, we can automatically scroll to the selected entry when dialogue is opened

Yes, it would be definitely necessary, especially with the current implementation with the 10000 styles.

timar (approved):

PR #5631, opened 2022-12-29
zotero: fix locale drop down showing incorrect language

lpranam:
user was allowed to select language without selecting the style, if style is not selected language data will also be not written in the doc.

force user to select style too if want to set the language.

Change-Id: Ie323b19189858e601f3fa979d747d7c688e4baf1

* Target version: master 

### Checklist


timar:
But when you have the style already selected, and you just want to change the language, then it's not possible now, you have to re-select the style. This is counter-intuitive. 
There's a typo in commit message 'fixe'.

lpranam:
Fixed the mentioned problem :)

timar (approved):

PR #5633, opened 2023-01-02
Fix error in conversion from microseconds to whole milliseconds

tml1024:
I had added some extra debugging output to the !MOBILEAPP branch to log values of timeoutMaxMicroS and timeoutMaxMS and was wondering why I saw things like:

    timeoutMaxMicroS=5000000 timeoutMaxMS=5009

and

    timeoutMaxMicroS=0 timeoutMaxMS=9

Change-Id: Iac8599ce5b00ef90d62eabc29c5d92858e276bb6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -342,13 +342,13 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         timeout.tv_nsec = (timeoutMaxMicroS % (1000 * 1000)) * 1000;
         rc = ::ppoll(&_pollFds[0], size + 1, &timeout, nullptr);
 #  else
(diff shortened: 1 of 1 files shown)


tml1024:
Note that as such I don't have any indication that this change would make something work better (not in a mobile app, not in WASM), but still, this change makes the code actually work as was intended, right?

kendy (approved):
Looks like the right thing to do, thank you!

tml1024:
The failing check seems random, but let me rebase and try again.

mmeeks:
Would love it on the stable branch too =) my bad - one too many 9's as it were.

PR #5635, opened 2023-01-02
Private/eszkadev/zotero note

eszkadev:
- notes from groups
- show whole note
- insert only one note, not duplicated

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -383,8 +383,8 @@ L.Control.Zotero = L.Control.extend({
 		}
 	},
 
-	fillNotes: function (items) {
-		var index = 0;
+	fillNotes: function (items, notesCount) {
+		var index = notesCount;
 		for (var iterator = 0; iterator < items.length; ++iterator) {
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
I have tested, all notes now appear without being cropped , including group notes:
image

But (probably unrelated to this PR) : 
* When I press refresh the dialog changes to citation style dialog (it happen with or without citation style set)
* The second time I click it takes quite a bit of time to show the notes


https://user-images.githubusercontent.com/65948705/210254673-62994368-fec8-493e-a835-f266b0feee40.mp4

timar:
@pedropintosilva The style list on pressing Refresh is not a regression of this PR. The behaviour of Refresh depends on the  dialogType, and 'insertnote' is not handled. 
cf. https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.Zotero.js#L660
Let's get this PR in, and @eszkadev will fix the issue next.

eszkadev:
refresh is fixed here: https://github.com/CollaboraOnline/online/pull/5637
after refresh we need to download all the data again - it can take more time

PR #5636, opened 2023-01-03
Zotero citation update

lpranam:
* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -477,6 +477,54 @@ L.Control.Zotero = L.Control.extend({
 		app.socket.sendMessage('commandvalues command=.uno:SetDocumentProperties?namePrefix=ZOTERO_PREF_');
 	},
 
+	setFetchedCitationFormat: function(style) {
+		if (!style)
+			style = this.settings.style;
+
+		var that = this;
+		fetch('https://www.zotero.org/styles/' + style)
+			.then(function (response) { return response.text(); })
+			.then(function (html) {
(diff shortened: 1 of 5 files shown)


lpranam:
known bug: inserting numeric citations in the middle will not renumber all citations, instead numbers are in order of insertion for new citations instead of appearance in the document.

lpranam:
The aforementioned problem should be solved now with the latest commit.

browser/src/control/Control.Zotero.js
```diff
@@ -314,30 +316,92 @@ L.Control.Zotero = L.Control.extend({
...
+
+	handleCitationText: function(citationItems) {
+		//Some citations have style so trying to change just inner most text without affecting other html
+		var citationString = '';
+		var citationNode = $(new DOMParser().parseFromString(citationItems[0].citation, 'text/html').body);
+
+		var that = this;
+		citationItems.forEach(function(item) {
+			var itemHTML = $(new DOMParser().parseFromString(item.citation, 'text/html').body);
+			var innerMostNode = itemHTML.find('*:not(:has("*"))')[0];
```
eszkadev, line 359: can we try to not use `jQuery` and instead use `HTMLElement.querySelector` or similar API (one dependency less)

lpranam: done

pedropintosilva:
this now is failing on cypress macro test on CI side.. re-triggering cypress

eszkadev (approved):

PR #5638, opened 2023-01-03
jsdialog: use unified naming for treeview icons

eszkadev:
unified header and no header treeviews
icon shouldn't contain old extension


Rash419 (approved):
LGTM!

PR #5639, opened 2023-01-03
private/pedro/references zotero ui jsdialogs fixes

pedropintosilva:
- JSDialog: add margin offsets to buttonbox L and R
- JSdialog: Fix container's padding
- JSDialog: edit controls (input fields): do not force their width
- Citations: Fix extra h margins affecting search and locale controls
- **Updated:** `So it uses new json property instead of checking for the specific id` Citations: Fix refresh button alignment
- **Removed:** `It seems it is better to do in Zotero dialog implementation in the callback as` @eszkadev `mentioned bellow` ~~Allow edit control to have placeholder + remove zotero search label~~

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -610,12 +610,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 												+ (data.layoutstyle ? data.layoutstyle : ''), parentContainer);
 		container.id = data.id;
 
+		var isButtonBoxLeft = data.leftaligned === 'true';
 		var leftAlignButtons = [];
 		var rightAlignButton = [];
 
 		for (var i in data.children) {
 			var child = data.children[i];
-			if (child.id === 'help')
+			if (child.id === 'help' || isButtonBoxLeft === true)
 				leftAlignButtons.push(child);
 			else
 				rightAlignButton.push(child);
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -128,9 +128,18 @@ L.Control.Zotero = L.Control.extend({
 							layoutstyle: 'end',
 							children: [
 								{
-									type: 'pushbutton',
-									id: 'zoterorefresh',
-									text: _('Refresh')
+									id: 'zoterorefresh-buttonbox',
+									type: 'buttonbox',
+									leftaligned: 'true',
+									children: [
+										{
+											type: 'pushbutton',
+											id: 'zoterorefresh',
+											text: _('Refresh')
+										}
+									],
+									vertical: false,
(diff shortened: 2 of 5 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -615,7 +615,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		for (var i in data.children) {
 			var child = data.children[i];
-			if (child.id === 'help')
+			if (child.id === 'help' || child.id === 'zoterorefresh')
```
eszkadev, line 618: I don't like this. We should put additional property in the JSON like 'leftaligned: true' and check here - so it will be more generic

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1455,6 +1455,18 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
+			if (edit.value === data.text) {
+				edit.value = '';
+			}
+		});
+
+		edit.addEventListener('focusout', function() {
+			if (edit.value === '') {
+				edit.value = data.text;
+			}
+		});
```
eszkadev, line 1468: This will be applied to any edit field. I can imagine there are places we don't want that behaviour.
Also it can break getting the real information because it will provide us old value if content was deleted when we close dialog. I think this should be done in Zotero dialog implementation in the callback... only for this specific case.

pedropintosilva: understood, so I will remove this but it would be great to have this (the possibility to have a placeholder that goes away when you start type, actually how it happens on Zotero app)

browser/src/control/Control.JSDialogBuilder.js
pedropintosilva: Makes sense, I have add that json property and have proceed accordingly. The 'help' case is still needed, I have checked but, and since it comes directly from core, we do not have any other place where we detect if we are in presence of a container with a help button

browser/src/control/Control.JSDialogBuilder.js
pedropintosilva: removed that last commit

browser/src/control/Control.Zotero.js
```diff
@@ -128,9 +128,18 @@ L.Control.Zotero = L.Control.extend({
 							layoutstyle: 'end',
 							children: [
 								{
-									type: 'pushbutton',
-									id: 'zoterorefresh',
-									text: _('Refresh')
+									id: 'zoterorefresh-buttonbox',
```
eszkadev, line 131: why this change in JSON is needed? what it does, I don't see difference.
also noticed that maybe removing  `layoutstyle` will do the same what `leftaligned` ?

PR #5646, opened 2023-01-04
Don't pointlessly assign global.socket = app.socket

tml1024:
Note that the code changed is inside an anonymous function that is called with one argument: window, for the function parameter called global. Thus the assignment that this change removes sets window.socket.

The window.socket variable is not used by code later, so this change can be this minimal.

The window.socket variable has earlier been set to the WebSocket object used to communicate with the COOL server. Let's keep it for that purpose. app.socket is the app.definitions.Socket object. Even though JavaScript is not type-safe, it is very confusing to use the same variable for completely different object types at different stages in the code execution. I am working on some new code that will want to access the WebSocket object also later.

Change-Id: I6273a6d6acee3fa46ab9fedc5742ae44c8bfa567

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/main.js
+++ b/browser/src/main.js
@@ -70,7 +70,6 @@ L.Map.THIS = map;
 
 map.loadDocument(global.socket);
 
-global.socket = app.socket;
 window.addEventListener('beforeunload', function () {
 	if (map && app.socket) {
 		if (app.socket.setUnloading)


gokaysatir (approved):
Thanks for the catch :)

tml1024:
As the "Unit & Cypress & Notebookbar Tests against Core co-22.05 " fails even with my change plus its revert, that it fails with just my change is probably not alarming. But will try again tomorrow.

PR #5647, opened 2023-01-04
browser: hide fontworks ui for non-odf files

dennisfrancis:
Change-Id: I8497a6b34186ce4167e4f864579a49a7d9910fb6

* Resolves: # 
* Target version: master 

### Summary

Hide fontworks ui for non-odf files. Revert this when actual fix lands in core master.

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -759,6 +759,7 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 	},
 
 	getInsertTab: function() {
+		var isODF = L.LOUtil.isFileODF(this._map);
 		var content = [
 			{
 				'type': 'bigtoolitem',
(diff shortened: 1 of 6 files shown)


eszkadev (approved):

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -2313,7 +2317,9 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 							{
 								'type': 'toolitem',
 								'text': _UNO('.uno:FontworkGalleryFloater'),
-								'command': '.uno:FontworkGalleryFloater'
+								'command': '.uno:FontworkGalleryFloater',
+								// Fontwork export/import not supported in other formats.
+								'visible': isODF ? 'true' : 'false',
 							}
 						]
 					},
```
eszkadev, line 2325: to perfect result, it would be good to set VerticalText to bigtoolitem when there is no Fontwork so it will be consistent with other icons (single big button)

dennisfrancis: Thanks, will add that in a followup.

dennisfrancis: Vertical text as 'toolitem' with Fontwork item visible.
image

Vertical text as 'bigtoolitem' with Fontwork item hidden. The label is shown twice for some reason.
image

PR #5649, opened 2023-01-04
Indirection/endpoint/support

Rash419:
* Resolves: # 
* Target version: master

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -363,7 +363,8 @@ m4_ifelse(MOBILEAPP,[true],
       window.useIntegrationTheme = 'false';
       window.checkFileInfoOverride = {};
       window.deeplEnabled = false;
-      window.zoteroEnabled = false;],
+      window.zoteroEnabled = false;
+      window.indirectionUrl='';],
      [window.host = '%HOST%';
       window.serviceRoot = '%SERVICE_ROOT%';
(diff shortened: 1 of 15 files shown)


kendy (changes requested):
Looks good to me in general, thank you a lot!

Unfortunately, there are few places that add whitespace at the end of the lines, I suspect you should configure your editor to strip those when saving?

I feel bad to dis-approve this PR just based on this - but from what I can see, so far we have been able to maintain the COOL codebase free from this problem :-)  So can you please go through these commits, make sure they are not introducing spaces on empty lines or before the line ends? Then I'll approve & let's merge it!

Rash419:
> Looks good to me in general, thank you a lot!
> 
> Unfortunately, there are few places that add whitespace at the end of the lines, I suspect you should configure your editor to strip those when saving?
> 
> I feel bad to dis-approve this PR just based on this - but from what I can see, so far we have been able to maintain the COOL codebase free from this problem :-) So can you please go through these commits, make sure they are not introducing spaces on empty lines or before the line ends? Then I'll approve & let's merge it!

Thanks, I have updated the PR and Editors' settings. I want you check specifically this commit https://github.com/CollaboraOnline/online/pull/5649/commits/1639c0bc6be7782b83e9bb0846922908a101fe2b
If we can do better there regarding updating the RouteToken in server side : )

kendy (approved):
Looks good to me, thank you!

PR #5651, opened 2023-01-04
Add UserPrivateInfo

eszkadev:
it will be used to store user data which shouldn't be shared between views


lpranam (approved):
looks good to me.

PR #5654, opened 2023-01-05
writer: vex-desktop: clicking away drops comment

dennisfrancis:
In case of desktop browser window whose size is only a little bigger than the page, our native comments get collapsed and we use vex. In this case don't close the comment dialog on clicking outside the dialog or when switching browser tabs.

Change-Id: I92c9d800d58a6910d6df01371e6748c864641cff

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -412,7 +412,11 @@ class CommentSection {
 				else {
 					this.cancel(comment);
 				}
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
I have tested on master and it works as expected:

* ESC key correctly cancels and closes the popup
* clicking anywhere else will not accidentally kill the comment
* SAve and cancel works
Thanks!

PR #5657, opened 2023-01-05
Add missing icons for macro dialog

eszkadev:


pedropintosilva (approved):
Tested, it works. Thanks for uploading the icons also, we will adjust or replace some of those but better to just get this is ASAP

PR #5658, opened 2023-01-05
Add float logical values fallback for buttonboxes

pedropintosilva:
Before this commit, browsers such as chrome (without enabling
experimental features) were not supporting those float values

Change-Id: I2530e7bf46d161ac980559c181a4e03947edbeee

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -76,6 +76,18 @@ button:not(.ui-corner-all):not(.button-primary) {
 
 /* button box */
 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
This works if the <html> has dir attribute **but** currently that <html> (inside of iframe) never gets that attribute. It seems `%UI_RTL_SETTINGS%` from poco is not working properly https://github.com/CollaboraOnline/online/blob/master/browser/html/cool.html.m4#L10

PR #5663, opened 2023-01-05
jsdialog: scroll treelist to selected entry

eszkadev:


pedropintosilva (changes requested):
Thanks @eszkadev ! It scrolls successfully but it seems not enough. The selected option is still not visible

https://user-images.githubusercontent.com/65948705/211023689-90571595-6cf0-4dec-b82f-85a7b72b596f.mp4

maybe something needs to be offset or maybe better yet would be if we could relly on what the user sees it -> https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -367,13 +367,25 @@ function _treelistboxControl(parentContainer, data, builder) {
 			tr.tabIndex = 0;
 			_headerlistboxEntry(tr, data, data.entries[i], builder);
 		}
+
+		setTimeout(function () {
+			var firstSelected = tbody.querySelector('.ui-listview-entry.selected');
+			if (firstSelected)
+				table.scrollTop = firstSelected.offsetTop;
```
pedropintosilva, line 374: for me it worked better with the scrollintoview
```diff
@@ -371,7 +371,7 @@ function _treelistboxControl(parentContainer, data, builder) {
 		setTimeout(function () {
 			var firstSelected = tbody.querySelector('.ui-listview-entry.selected');
 			if (firstSelected)
-				table.scrollTop = firstSelected.offsetTop;
+				firstSelected.scrollIntoView();
```

pedropintosilva (changes requested):

browser/src/control/jsdialog/Widget.TreeView.js
eszkadev: I remember that we had issue with scrollIntoView moving whole iframe in some cases (leaving blank space above notebookbar) so I wanted to avoid it. But we can try.

eszkadev: Unfortunately it does scroll whole app with scrollIntoView, I guess it will be even more visible on mobile :(

eszkadev: I think it was timing issue on your end, I updated the patch with different approach.

pedropintosilva (changes requested):
Looks good on Chrome/ium but on FF is still not visible. Maybe we can just change a bit the offset so it is a bit down and not just right as the first line (so we have a bit of error margin)

pedropintosilva (approved):
Now it works

PR #5671, opened 2023-01-06
Zotero refmark

lpranam:
now odf documents will by default use refmarks,
MS docs will use field marks by default

Change-Id: I57ef5de9bef8d1e5f6b5029e478d9a43e38dbc4d

* Target version: master 

### TODO
With reference marks, citation renumbering will not work as of now due to the core not sending remarks in sequence of them appearing in doc. All else should work fine. core patch work is in queue. MS doc formats will work as expected

### Checklist


lpranam:
required core patches:
https://gerrit.libreoffice.org/c/core/+/145149
https://gerrit.libreoffice.org/c/core/+/145167

timar (approved):

PR #5673, opened 2023-01-06
Start cherry-picking commits from the feature/wasm branch

tml1024:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
(I can't reproduce the problem locally, trying "make dist". (After also doing "make distclean" and/or "git clean".))

tml1024:
Sigh, so let's get at least the two first commits in then. This hopefully won't run into the mysterious make dist problem.

PR #5682, opened 2023-01-10
Show epub settings dailog on mobile view.

GulsahKose:
On mobide view Download as -> EPUB should show EPUB download settings dialog before downloading the document same like desktop view.

Change-Id: Iebe05e65f331de6737be88288fac34244cfcde18

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.Menubar.js
```diff
@@ -1723,7 +1723,7 @@ L.Control.Menubar = L.Control.extend({
 			this._map.openSaveAs();
 		} else if (id === 'exportpdf') {
 			this._map.dispatch('exportpdf');
-		} else if (id === 'exportepub') {
+		} else if (id === 'exportepub' || id === 'downloadas-epub') {
```
eszkadev, line 1726: I would recommend to rather change 'downloadas-epub' id to 'exportepub' in this file above so we will be unified with desktop variant. Then it will be handled only in one place here and not below in the "downloadas-"

eszkadev: and the same for pdf (it also should show the dialog)

GulsahKose: Yes make sense, I'll try

pedropintosilva:
From the user perspective, I have tested on responsive mode and physical decide. It works, thanks.

timar:
I fixed the typos in commit messages, but Cypress failed again, and I checked the output. It seems the failure is related to the patch, we have mobile UI test for PDF export, and we changed the behaviour. I think the tests should be updated.

GulsahKose:
Now I understand why tests failing. Actually commit is ok but after the change we have to do additional steps to download a PDF or EPUB. Before it was downloading direcly but now we have a setting dialog to download. I think we should change the test scenario. I'll try to change test.

eszkadev (approved):
works as expected

PR #5695, opened 2023-01-12
Add min and max width to jsdialog modalPopup

pedropintosilva:
with the introduction of showModal:
https://github.com/CollaboraOnline/online/pull/5518 we now have
a type of a dialog that, without this commit, can grow without
limits and can be as small as its content which is not ideal.

Change-Id: I8768925377606924fd70fab8be40b0b755d802ac

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -32,6 +32,8 @@
 }
 
 .jsdialog-container.modalpopup {
+	min-width: 198px;
+	max-width: 498px;
 	border: 1px solid #a4a4a4 !important;
 	border-radius: var(--border-radius);
 	box-shadow: 0 -1px 2px 0 rgba(0, 0, 0, 0.15), 0 2px 2px 0 rgba(0, 0, 0, 0.1);


browser/src/control/Control.JSDialog.js
```diff
@@ -312,6 +312,8 @@ L.Control.JSDialog = L.Control.extend({
 		};
 
 		if (isModalPopup && data.hasClose) {
+			container.style.minWidth = '198px';
```
eszkadev, line 315: maybe better to keep visual properties in CSS? what do you think?
it already has class: .modalpopup.jsdialog-container so can be targetted
if it's important that it has close button then we can put hasClose class also

pedropintosilva: Yup, I never too sure were to put it. On one hand if it stays in the JS then anyone editing JS files will see it and it will avoid people add in style properties via JS (that override the stuff that is on the CSS side). But I agree I prefer to keep it in the css

eszkadev (approved):

PR #5697, opened 2023-01-12
Zotero: split JSON and add field type in settings

eszkadev:
dialog was rearranged:
fieldtype


pedropintosilva:
Please **do not** merge until it's tested against latest @lpranam commits

pedropintosilva (changes requested):
Thanks, maybe using a dropdown would look better while reducing the decision anxiety that the radio button can cause specially when the user might not know what's best. With the dropdown the user will assume that the default is ok. On top of that if  we then need to add more options it will still look ok and we also have the freedom to had addiitonal `(info)` that can help the user decide. Example: `Fields (info hint)` `Bookmarks (info hint)` where `info hint` helps the user to understand the implication between choosing one or the other.

 Screenshot_20230112_114008

lpranam (approved):
code looks good to me but maybe we can consider Pedro's suggestion

eszkadev:
Updated PR:

fieldtype

pedropintosilva (approved):

PR #5698, opened 2023-01-12
Zotero: enable ok only if selected style

eszkadev:
bug in citation preferences... when opened ok is already active and causes error when clicked


pedropintosilva:
Thanks @eszkadev now the error is fixed. But The end result, I feel, can unexpected for the user:

1. Open Citation preferences
2. Dialogs opens with the last style used (awesome)
3. Click ok
  * not possible to click ok it's disabled even thought a item from the list is selected
4. User needs to selected what is already selected (clicking again in that row item) and then press ok

Would be better to already activate the OK button with specific style ready to be send

eszkadev:
then I close this PR, I fixed that with field type selector PR

PR #5702, opened 2023-01-13
Private/gokay/scroll perf

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1060,7 +1060,7 @@ L.CanvasTileLayer = L.Layer.extend({
 				setTimeout(this.update.bind(this), 200);
 			}, this._painter);
 		}
-		else {
+		else if (this._docType !== 'spreadsheet') { // See scroll section. panBy is used for spreadsheets while scrolling.
 			this._map.on('movestart', this._painter.startUpdates, this._painter);
 			this._map.on('moveend', this._painter.stopUpdates, this._painter);
 		}
--- a/browser/src/layer/tile/ScrollSection.ts
+++ b/browser/src/layer/tile/ScrollSection.ts
@@ -579,7 +579,8 @@ class ScrollSection {
 			this.sectionProperties.drawVerticalScrollBar = true;
 			this.sectionProperties.mouseIsOnVerticalScrollBar = true;
 			this.sectionProperties.mapPane.style.cursor = 'pointer';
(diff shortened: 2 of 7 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1060,7 +1060,7 @@ L.CanvasTileLayer = L.Layer.extend({
 				setTimeout(this.update.bind(this), 200);
 			}, this._painter);
 		}
-		else {
+		else if (this._docType !== 'spreadsheet') { // See scroll section. panBy is used for spreadsheets while scrolling.
 			this._map.on('movestart', this._painter.startUpdates, this._painter);
```
mmeeks, line 1064: The profile suggested that both startUpdates and stopUpdates did a synchronous full re-draw during their execution ;-) - which I guess is not what is/was intended. Is that still so for spreadsheets ? =)

gokaysatir: It was working for spread sheets and updates were starting and stopping immediately without another function in between. So i cancelled it. I will also check the behaviour for Writer and others. It can be removed or improved i guess.

Also for scrolling, another functionality is used for Calc, so we don't seem to need this.

eszkadev:
With this PR when document is loaded - I see flickering (black area on canvas appears for a short time)

browser/src/layer/tile/ScrollSection.ts
```diff
@@ -608,7 +609,8 @@ class ScrollSection {
 			this.sectionProperties.drawHorizontalScrollBar = true;
 			this.sectionProperties.mouseIsOnHorizontalScrollBar = true;
 			this.sectionProperties.mapPane.style.cursor = 'pointer';
-			this.containerObject.requestReDraw();
+			if (!this.containerObject.draggingSomething && !(<any>window).mode.isDesktop())
+				this.containerObject.requestReDraw();
```
eszkadev, line 613: on mobile when dragging "scroll indicator handler (circle with arrows up/down) and view goes outside currently rendered view - we get blank screen. It is rendered only when we finish dragging. Is that expected?

eszkadev: no, it works when tried again :)

eszkadev (approved):
apart of flickering on load, looks good

gokaysatir:
Closed in favor of another PR.

PR #5703, opened 2023-01-13
Zotero: ask integrator if key is missing

eszkadev:
- It sends post message to the integrator: UI_ZoteroKeyMissing if API key was not provided in CheckFileInfo and user tried to perform any action on bibliography
- It shows snackbar with message that key is missing
- It hides bibliography functionality if zotero.enable = false is set in coolwsd.xml


juliushaertl (commented):
Didn't test yet, but the post message looks reasonable for richdocuments and thanks for adding a config option for that 👍

pedropintosilva (approved):
Works as described 

When the feature is enabled but API key is missing:

image

and when the feature set is disabled via coolwsd.xml

image

PR #5708, opened 2023-01-14
kit: default timeout to ~5 seconds when we get a negative wait.

mmeeks:
I can only assume that recent LOK's often don't have any lingering timeouts for cursor blinking or whatever, such that we get an infinite wait passed as a negative microseconds-count.

Best not to interpret that as a zero wait to avoid a busy loop.

Change-Id: If79228af969f4598f07681deb355c72f72602d19

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2479,6 +2479,8 @@ void documentViewCallback(const int type, const char* payload, void* data)
 /// Called by LOK main-loop the central location for data processing.
 int pollCallback(void* pData, int timeoutUs)
 {
+    if (timeoutUs < 0)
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Looks good. Should we try and avoid the negative value case in Core, so as to avoid blindly overriding here and use a sensible value that works for Core too?

mmeeks:
Best practice is not to have a timeout at all - it saves power to go completely silent (and logging lines ;-) - so we should slowly increase our timeout there from 5 seconds to infinity over time I guess. Prolly something to have enabled in CODE FWIW - we could add a conditional there for the CODE version to bump it to 60 seconds or so I guess - I'm optimistic it shouldn't cause problems.

PR #5718, opened 2023-01-17
Switch focus to editor after JSDialog close event

GulsahKose:
Similar to  LokDialog _onDialogClose, after the closing a JSDialog we should change the focus editor again.

Change-Id: I348ead91838703590d5b9753d6028ff7f70f6111

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -93,7 +93,6 @@ L.Control.JSDialog = L.Control.extend({
 		}
 		else {
 			this.clearDialog(id);
-			this.map.focus();
 		}
 	},
 
@@ -156,6 +155,21 @@ L.Control.JSDialog = L.Control.extend({
 		else if (data.action === 'close')
 		{
 			this.close(data.id, false);
+
+			// Manage focus
+			var dialogs = Object.keys(this.dialogs);
+			if (dialogs.length) {
+				var lastKey = dialogs[dialogs.length - 1];
+				var container = this.dialogs[lastKey].container;
+				container.focus();
+				var initialFocusElement =
+					container.querySelector('[tabIndex="0"]:not(.jsdialog-begin-marker)');
+				initialFocusElement.focus();
+			}
+			else if (!this.hasDialogOpened()) {
+				this._map.fire('editorgotfocus');
+			}
+
 			return;
 		}
 


eszkadev (approved):
looks good, I see one possible improvement here:
- when we open dialog with subdialog, after closing child dialog we will get focus on map
it would be good to focus previous dialog, to get that we can use code similar to: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialog.js#L603

GulsahKose:
@eszkadev 
>     * when we open dialog with subdialog, after closing child dialog we will get focus on map
>       it would be good to focus previous dialog
Here  i tried to do focus that dialog but somehow focus still on editor. You can test with "Find and Replace" then its "Help" dialog.

Any chance to test?

eszkadev:
> Here i tried to do focus that dialog but somehow focus still on editor. You can test with "Find and Replace" then its "Help" dialog.
> 
> Any chance to test?

This is caused by this line: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialog.js#L96

It seems I've tried to fix that issue in the past but did that incompletly. Feel free to remove it and replace with your fix.

eszkadev:
diff --git a/browser/src/control/Control.JSDialog.js b/browser/src/control/Control.JSDialog.js
index 8c2606f06..69db38a9c 100644
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -93,7 +93,6 @@ L.Control.JSDialog = L.Control.extend({
                }
                else {
                        this.clearDialog(id);
-                       this.map.focus();
                }
        },
 
@@ -162,12 +161,13 @@ L.Control.JSDialog = L.Control.extend({
                        if (dialogs.length) {
                                var lastKey = dialogs[dialogs.length - 1];
                                var container = this.dialogs[lastKey].container;
-                               container.focus();
+                               var initialFocusElement =
+                                       container.querySelector('[tabIndex="0"]:not(.jsdialog-begin-marker)');
+                               initialFocusElement.focus();
                        }
                        else if (!this.hasDialogOpened()) {
                                this._map.fire('editorgotfocus');
                        }
-
                        return;
                }
 
Here is working solution. Example with find&Replace > Help doesn't work due to some jsdialog bug (I will try to fix later). But for example File -> Properties -> set password works

eszkadev:
here is additional fix needed for Help & Find&Replace to work: https://github.com/CollaboraOnline/online/pull/5725

GulsahKose:
Thanks Szymon, I've updated patch as you suggest. Then we can merge following PR to fix Find and Replace dialog issue.
> here is additional fix needed for Help & Find&Replace to work: #5725

eszkadev (approved):

PR #5723, opened 2023-01-18
add ThemeDialog to the notebook bar and main menu

quikee:
Change-Id: Ia50656414c0db2173d2233c4864b67e2c5dbd2e3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (changes requested):
The button should be visible only when we are in presence of  Microsoft OOXML documents right now the option appears to all file types

pedropintosilva (changes requested):
Theme dialog appears
Changing to other theme
Then the new palette is available via dialogs such as paragraph etc (provided that the user chooses `Theme colors`  
image

The button is available in the tabbed view

but the option is not available when I change to compact view 
image

PR #5738, opened 2023-01-20
Export PDF to storage

eszkadev:
It allows to export PDF directly to the storage, without need to download it.

This is copy of the save as functionality flow (reuses interaction with integrator)

`Save as` scenario looks like:
1. we sent UI_SaveAs to integrator
2. Integrator opens dialog, user selects target file name, integrator sends response to us
3. client sends to server `saveas` with target file name and format
4. we export file locally on the server
5. server uploads local file to the integrator's storage
6. server sends message to the client
7. client reloads app, by opening new file

Differences in export:
* In point 4. we use uno command to open additional dialog with options. After that we receive callback: LOK_CALLBACK_EXPORT_FILE with: PENDING or ABORT status. Then again after file was exported we got LOK_CALLBACK_EXPORT_FILE with exported local file path.
* In point 7. we don't reopen newly exported file. We only show the message to the user about successful operation with file name.

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1122,6 +1122,9 @@ app.definitions.Socket = L.Class.extend({
 		else if (textMsg.startsWith('saveas:') || textMsg.startsWith('renamefile:')) {
 			this._renameOrSaveAsCallback(textMsg, command);
 		}
+		else if (textMsg.startsWith('exportas:')) {
+			this._exportAsCallback(command);
+		}
 		else if (textMsg.startsWith('warn:')) {
 			var len = 'warn: '.length;
 			textMsg = textMsg.substring(len);
@@ -1250,6 +1253,11 @@ app.definitions.Socket = L.Class.extend({
 		if (this._map._docLayer && !msgDelayed) {
 			this._map._docLayer._onMessage(textMsg, e.image);
 		}
+	},
+
+	_exportAsCallback: function(command) {
+		this._map.hideBusy();
+		this._map.uiManager.showInfoModal('exported-success', _('Exported to storage'), _('Successfully exported: ') + decodeURIComponent(command.filename), '', _('OK'));
 	},
 
 	_renameOrSaveAsCallback: function(textMsg, command) {
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -329,6 +329,7 @@ bool ChildSession::_handleInput(const char *buffer, int length)
                tokens.equals(0, "selectgraphic") ||
                tokens.equals(0, "resetselection") ||
                tokens.equals(0, "saveas") ||
+               tokens.equals(0, "exportas") ||
                tokens.equals(0, "useractive") ||
                tokens.equals(0, "userinactive") ||
                tokens.equals(0, "windowcommand") ||
@@ -451,6 +452,10 @@ bool ChildSession::_handleInput(const char *buffer, int length)
         else if (tokens.equals(0, "saveas"))
         {
             return saveAs(tokens);
+        }
+        else if (tokens.equals(0, "exportas"))
+        {
+            return exportAs(tokens);
         }
         else if (tokens.equals(0, "useractive"))
         {
@@ -2497,6 +2502,76 @@ bool ChildSession::saveAs(const StringVector& tokens)
     return true;
 }
 
+bool ChildSession::exportAs(const StringVector& tokens)
+{
+    std::string wopiFilename, url;
+
+    if (tokens.size() <= 1 ||
+        !getTokenString(tokens[1], "url", url))
+    {
+        sendTextFrameAndLogError("error: cmd=exportas kind=syntax");
+        return false;
+    }
+
+    Poco::URI wopiURL(url);
+    if (wopiURL.getScheme() == "wopi")
+    {
+        std::vector<std::string> pathSegments;
+        wopiURL.getPathSegments(pathSegments);
+
+        if (pathSegments.size() == 0)
+        {
+            sendTextFrameAndLogError("error: cmd=exportas kind=syntax");
+            return false;
+        }
+
+        wopiFilename = wopiURL.getPath();
+    }
+    else
+    {
+        sendTextFrameAndLogError("error: cmd=exportas kind=syntax");
+        return false;
+    }
+
+    // for PDF and EPUB show dialog with export options first
+    // when options will be chosen and file exported we will
+    // receive LOK_CALLBACK_EXPORT_FILE message
+    std::string extension = FileUtil::extractFileExtension(wopiFilename);
+
+    const bool isPDF = extension == "pdf";
+    const bool isEPUB = extension == "epub";
+    if (isPDF || isEPUB)
+    {
+        // We don't have the FileId at this point, just a new filename to save-as.
+        // So here the filename will be obfuscated with some hashing, which later will
+        // get a proper FileId that we will use going forward.
+        LOG_DBG("Calling LOK's exportAs with: '" << anonymizeUrl(wopiFilename) << "'.");
+
+        getLOKitDocument()->setView(_viewId);
+
+        std::string encodedWopiFilename;
+        Poco::URI::encode(wopiFilename, "", encodedWopiFilename);
+
+        _exportAsWopiUrl = encodedWopiFilename;
+
+        const std::string arguments = "{"
+            "\"SynchronMode\":{"
+                "\"type\":\"boolean\","
+                "\"value\": false"
+            "}}";
+
+        if (isPDF)
+            getLOKitDocument()->postUnoCommand(".uno:ExportToPDF", arguments.c_str(), false);
+        else if (isEPUB)
+            getLOKitDocument()->postUnoCommand(".uno:ExportToEPUB", arguments.c_str(), false);
+
+        return true;
+    }
+
+    sendTextFrameAndLogError("error: cmd=exportas kind=unsupported");
+    return false;
+}
+
 bool ChildSession::setClientPart(const StringVector& tokens)
 {
     int part = 0;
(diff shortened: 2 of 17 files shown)


eszkadev (commented):
Still WIP

mmeeks (changes requested):
Really prefer this as a new 'Export' top-level I think, cf. ticket =)

wsd/DocumentBroker.hpp
```diff
@@ -611,7 +611,7 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
     /// Upload the doc to the storage.
     void uploadToStorageInternal(const std::shared_ptr<ClientSession>& session,
                                  const std::string& saveAsPath, const std::string& saveAsFilename,
-                                 const bool isRename, const bool force);
+                                 const bool isRename, const bool isExport, const bool force);
```
eszkadev, line 614: can be done via enum? (  SaveAs | Rename | Export )

eszkadev:
> Really prefer this as a new 'Export' top-level I think, cf. ticket =)

done.

For now PDF is enabled in menu but EPUB not (requires some changes in richdocuments for upload, will be added later)

kit/ChildSession.cpp
```diff
@@ -2381,6 +2386,18 @@ bool ChildSession::resetSelection(const StringVector& tokens)
     return true;
 }
 
+namespace
+{
+static std::string extractFormat(const std::string& path)
```
Ashod, line 2391: Surely we have this already. At least we have `Util::split()` and `Util::splitLast()` that make the implementation trivial, and in that case, consider having this in FileUtil, please, with a more meaningful name like `extractFileExtension` ("format" and "extension" aren't the same, of course, since a file can have the wrong extension).

See `Util::splitUrl()` for a very similar function.

wsd/DocumentBroker.cpp
```diff
@@ -1752,7 +1753,7 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
                                                    << filenameAnonym << "] successfully.");
 
                 std::ostringstream oss;
-                oss << "saveas: url=" << url << " filename=" << encodedName
+                oss << (_uploadRequest->isExport() ? "exportas:" : "saveas:") << " url=" << url << " filename=" << encodedName
```
Ashod, line 1756: Why is it important to differentiate between the exportas URL and the saveas URL in the client? 

Do we even need to tell the client the URL of the export?

wsd/DocumentBroker.hpp
Ashod: It seems we are passing this flag around only to generate the notification message to the client to be "exportas: url=..." instead of "saveas: url=...", but it's not clear to me why we need this distinction at all. Let's discuss please.

kit/ChildSession.cpp
```diff
@@ -3146,6 +3233,33 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
...
+        {
+            sendTextFrame("blockui: ");
+            return;
+        }
+
+        sendTextFrame("unblockui: ");
+
+        if (isAbort)
+        {
+            _exportAsWopiUrl = "";
```
Ashod, line 3249: `std::string` has `clear` member for this purpose.

kit/ChildSession.cpp
```diff
@@ -3146,6 +3233,33 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
         break;
     case LOK_CALLBACK_EXPORT_FILE:
     {
+        bool isAbort = payload == "ABORT";
+        bool isPending = payload == "PENDING";
```
Ashod, line 3237: Having enum for these would be best, but I'm not sure I understand when we expect "pending" and when we expect "abort".

kit/ChildSession.cpp
```diff
@@ -3146,6 +3233,33 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
...
+        {
+            _exportAsWopiUrl = "";
+            return;
+        }
+        else if (!_exportAsWopiUrl.empty())
+        {
+            std::string encodedURL;
+            Poco::URI::encode(payload, "", encodedURL);
+
+            sendTextFrame("exportas: url=" + encodedURL + " filename=" + _exportAsWopiUrl);
```
Ashod, line 3257: What's the purpose of sending this? We also send it when the upload is complete, which looks like we are sending this message twice, once on exporting the document on the server (locally) and once when we upload to storage.

This would seem wrong, since the client can't do anything with the exported file on the server (locally) and can only possibly do something (like reloading) when the file is uploaded.

Maybe I'm missing the purpose of these "exportas" messages.

Ashod (commented):
Thanks @eszkadev. Good stuff, but I think there are a few things we can discuss and improve.

wsd/DocumentBroker.hpp
eszkadev: `save as` closes old document and opens the newly exported file, `export` only exports the file and informs user about the result. Difference is visible in the client side where message handling happens

browser/src/core/Socket.js
```diff
@@ -1252,6 +1255,11 @@ app.definitions.Socket = L.Class.extend({
 		}
 	},
 
+	_exportAsCallback: function(command) {
+		this._map.hideBusy();
+		this._map.uiManager.showInfoModal('exported-success', _('Exported to storage'), _('Successfully exported: ') + decodeURIComponent(command.filename), '', _('OK'));
```
eszkadev, line 1260: here is `export` result handling, while below we see standard `save as` - closing the document and opening the new one

wsd/DocumentBroker.cpp
eszkadev: we need to return the file name only as it can differ from what user typed (when file already exists in the storage)

kit/ChildSession.cpp
eszkadev: at this point file was exported locally, now we need to send the message so we will start the uploading part

kit/ChildSession.cpp
eszkadev: done

kit/ChildSession.cpp
eszkadev: This comes from the dialog execution result:
- ok/export -> PENDING
- cancel -> ABORT
- error during export -> ERROR
- export finished with success -> URL in the payload

mmeeks:
Silly question - but there is some property we use to determine if 'save-as' should be shown (or not) as in - will the integration support our post-message handshake to get the dialog shown and the save-URL (?)

Just to check - we are checking that for 'save-as' and 'export' in the  notebookbar mode ? =) Thanks!

kit/ChildSession.cpp
```diff
@@ -2497,6 +2502,76 @@ bool ChildSession::saveAs(const StringVector& tokens)
...
+
+    getLOKitDocument()->setView(_viewId);
+
+    std::string encodedWopiFilename;
+    Poco::URI::encode(wopiFilename, "", encodedWopiFilename);
+
+    // for PDF and EPUB show dialog with export options first
+    // when options will be chosen and file exported we will
+    // receive LOK_CALLBACK_EXPORT_FILE message
+    std::string detectedFormat = FileUtil::extractFileExtension(wopiFilename);
```
Ashod, line 2549: `const std::string extension`

kit/ChildSession.cpp
```diff
@@ -2497,6 +2502,76 @@ bool ChildSession::saveAs(const StringVector& tokens)
...
+
+    // We don't have the FileId at this point, just a new filename to save-as.
+    // So here the filename will be obfuscated with some hashing, which later will
+    // get a proper FileId that we will use going forward.
+    LOG_DBG("Calling LOK's exportAs with: '" << anonymizeUrl(wopiFilename) << "'.");
+
+    getLOKitDocument()->setView(_viewId);
+
+    std::string encodedWopiFilename;
+    Poco::URI::encode(wopiFilename, "", encodedWopiFilename);
```
Ashod, line 2544: This could be done under the `if (isPDF || isEPUB)`  conditional.

kit/ChildSession.cpp
```diff
@@ -3146,6 +3221,46 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
...
+        bool isPending = payload == "PENDING";
+        bool exportWasRequested = !_exportAsWopiUrl.empty();
+
+        if (isPending) // local save has been started
+        {
+            sendTextFrame("blockui: ");
+            return;
+        }
+
+        sendTextFrame("unblockui: ");
```
Ashod, line 3235: Doing this unconditionally may be problematic. We should match "blockui:" with "unblockui:" and not issue more or less of either one. Here we are issuing move "unblockui:" than we should (for example, or regular export we will be sending "unblockui:").

wsd/DocumentBroker.cpp
Ashod: My question was whether the client code differentiates between an exportas and saveas. I assume it does, otherwise we could just send "saveas" at this line.

wsd/DocumentBroker.hpp
Ashod: Thank you. This answers my question above. 

Regarding using an enum, yes, this function was already messy and now it's more complicated. I had plans on cleaning it up. Now there is even more reason.

Ashod (commented):
Thanks @eszkadev. Looks good to me. Left a couple of suggestions, but nothing major.

eszkadev:
> Silly question - but there is some property we use to determine if 'save-as' should be shown (or not) as in - will the integration support our post-message handshake to get the dialog shown and the save-URL (?)
> 
> Just to check - we are checking that for 'save-as' and 'export' in the notebookbar mode ? =) Thanks!

@mmeeks done, I forgot to handle that in menubar

mmeeks (approved):
Thanks Szymon =)

PR #5743, opened 2023-01-23
Fix w2ui menus appearing with unnecessary scrollbars

pedropintosilva:
Before this commit, on chrome/mium, menus such as
image or save as in the tabbed view (and with the browser zoom
set to bellow 80%) were being displayed with scrollbars.

Change-Id: Id32e86229fe5d7aad293d24a20c9bfa555fbe78b


pedropintosilva:
This needs deep testing, until then we cannot merge it. I did some testing and found no issues with switching the positon but I did not test it as much as I would like

eszkadev (approved):

PR #5744, opened 2023-01-24
browser: Make navigation shortcuts work in read-only mode

thebearon:
F5 for slideshow, too.
Regression from 47c52b90f3a295e29106a61596645e774949a0aa

Change-Id: I056703f575460076693b245f790c7284854162be

* Target version: master 


lpranam (approved):
code looks good

PR #5745, opened 2023-01-24
DO NOT MERGE formulabar: allow arrow key navigation

eszkadev:


eszkadev:
better fix in https://github.com/CollaboraOnline/online/pull/5784

PR #5750, opened 2023-01-25
Don't show pdf export for readonly templates

eszkadev:


eszkadev:
included in https://github.com/CollaboraOnline/online/pull/5738

PR #5751, opened 2023-01-25
Improve cypress tests

NickWingate:
Signed-off-by: NickWingate <nick.wingate@collabora.com>
Change-Id: I99c06fdde0bcf442952d5bbfc12002e56e520f47

* Resolves: # 
* Target version: master 
* Phab ticket: https://phabricator.collabora.com/T39833

### Summary
Improve cypress test efficiency with two main optimisations:
- Remove `cy.wait()` calls where possible and use HTML checks instead
- See if possible to run multiple tests on a document inbetween setup and teardown. e.g. Test if **bold** and _italic_ work in the same document to avoid the exact same calls before and after.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/desktop/writer/image_operation_spec.js
+++ b/cypress_test/integration_tests/desktop/writer/image_operation_spec.js
@@ -40,8 +40,6 @@ describe('Image Operation Tests', function() {
 		cy.get('#selectheight input').clear({force:true})
 			.type('2{enter}', {force:true});
 
-		cy.wait(1000);
-
 		assertImageSize(139, 93);
 
 		//Keep ratio checked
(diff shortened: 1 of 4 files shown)


pedropintosilva:
Thanks @NickWingate it would be great if we could do maybe do smaller PRs that can be reviewed and merged easily and faster instead of a on-going big draft PR ;)

mmeeks:
@neilguertincollabora - possibly you've fixed this already; any chance you can opine on Nick's somewhat older patch here - and (ideally) close it by merging anything that remains to be done - we're trying to get older things out of the review queue; Thanks Nick & Neil ! =)

neilguertincollabora:
A lot of waits were removed in https://github.com/CollaboraOnline/online/pull/7954. So the few remaining here would have to be re-implemented and re-tested.

I assume that "See if possible to run multiple tests on a document inbetween setup and teardown." got moved to another PR. That would be nice to have but I don't see it here.

PR #5760, opened 2023-01-26
More WASM hacking

tml1024:
Apparently handle_cool_message() gets called from a web worker thread and then using emscripten_run_script() in wasmapp.cpp to run a JS snippet that refers to the 'window' variable will not work. That variable exists only in the main thread. So use MAIN_THREAD_EM_ASM() instead.

Hardcode the document URL for now also in wasmapp.cpp.

Try to send the HULLO message from COOLWSD::innerMain().

Change-Id: Ic48042c6d0c6239a3b82e74f0565056a15f3d98d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wasm/wasmapp.cpp
+++ b/wasm/wasmapp.cpp
@@ -12,7 +12,7 @@ int coolwsd_server_socket_fd = -1;
 const char* user_name;
 const int SHOW_JS_MAXLEN = 70;
 
-static std::string fileURL;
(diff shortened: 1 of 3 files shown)


tml1024:
Sigh. A change to wasm/wasmapp.cpp can *not* have any impact on the Cypress tests. Oh well, it was fun while CI worked without false negatives for some hours.

PR #5761, opened 2023-01-26
Keep focus on sheet if empty part of the notebookbar clicked

GulsahKose:
When we click empty part of notebook bar, focus should be on sheet. Without this patch, focus is on browser itself and CTRL+S saves the HTML instead of current document.

Change-Id: Ia0a89ae761eeeb3bb21a655c5181425df19e1c94

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -13,6 +13,7 @@ L.Control.UIManager = L.Control.extend({
 
 	onAdd: function (map) {
 		this.map = map;
+		var that = this;
 		this.notebookbar = null;
 		// Every time the UI mode changes from 'classic' to 'notebookbar'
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
code looks good

GulsahKose:
Hi @eszkadev I add a seperate commit that fix the same focus problem on navigation bar. When you have time would be nice to have your review. Thanks.

eszkadev:
@GulsahKose does the same code work for classic/compact mode? If yes - we should move it to the UI manager and use in both modes.
If not and it breaks something in menubar - we should deregister the callback in onRemove (happens when we switch between modes)

GulsahKose:
Now it works on tabbed view and compact view.

PR #5764, opened 2023-01-26
zotero: restric user from changing field type

lpranam:
if document contains existing zotero citation,
restrict user from changing the field storage type.

currently online does not have any ways to change the field type maybe in future we can add some core APIs to support this

Change-Id: I58caf403744431d34344f5a1cb635781409d7e68

* Target version: master 

### Checklist


lpranam:
@pedropintosilva maybe we can add some note in the dialog about this, if document has existing citations we can not change it's type or something like that?

eszkadev (approved):
code looks good

PR #5765, opened 2023-01-26
jsdialog: convert delete slide modal from vex

eszkadev:
can be found in menubar -> slide -> delete

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -3,7 +3,7 @@
 * Control.Menubar
 */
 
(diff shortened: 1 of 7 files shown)


pedropintosilva (changes requested):
Thanks @eszkadev that was very good pick for a dialog to convert! It seems to work but I'm afraid Control.PresentationBar.js should also suffer changes otherwise the left pane -> 'deletepage' button will trigger another non-jsdialog pop up

eszkadev:
> 

good catch, updated

pedropintosilva (approved):
Thanks!

PR #5766, opened 2023-01-26
zotero: added warning for unreachable citation items

lpranam:
In cases where zotero item is added by a user and another user opens the doc, if the new user does not have that item in the database, he may fail to do activity related to that item, so we warn the user when we are setting up the zotero

Change-Id: I9056f38334cb8d9b127ddb594adb083d25bd4472

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -56,6 +56,7 @@ L.Control.Zotero = L.Control.extend({
 			this.citationCluster[values.citationID] = [];
 			var citationString = L.Util.trim(values.properties.plainCitation, this.settings.layout.prefix, this.settings.layout.suffix);
 			var citations = citationString.split(this.settings.layout.delimiter);
+			var itemUriList = [];
 			values.citationItems.forEach(function(item, i) {
 				//zotero desktop versions do not store keys in cslJSON
 				//extract key from the item url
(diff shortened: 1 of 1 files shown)


pedropintosilva (commented):
I didn't manage to have this working.

1. Created a new document
2.  Inserted a citation
3.  Verified that the citation was inserted
4. save and close
5. Head over to Zotero web library and remove that reference
6. Head back to Nextcloud and re-open the document 
7. No warning message appeared

browser/src/control/Control.Zotero.js
```diff
@@ -110,9 +113,42 @@ L.Control.Zotero = L.Control.extend({
...
+
+			var catagory = uriObject.pathname.startsWith('/groups/') ? '/groups/' : '/users/';
+
+			var id = uriObject.pathname.substring(catagory.length);
+			id = id.substring(0, id.indexOf('/'));
+			if (!this.groupIdList.has(id) && id !== this.userID.toString()) {
+				this.showUnsupportedWarning = false;
+				this.map.uiManager.showInfoModal('zoterounreachablewarn', _('Zotero Warning'),
+					_('The document contains some citations which may be unreachable through web API.'),
+					_('It may cause some problems while editing citations or bibliography.'), _('Close'), function() {});
```
pedropintosilva, line 146: Would be good to do not have the line break in between

pedropintosilva (changes requested):

lpranam:
@pedropintosilva good catch, I think it's unlikely that people delete items from their database, so this scenario has low chances and shouldn't be a blocker for this PR.

We rely on userId and groupId to check whether an item is reachable or not, even deleted items have a valid and reachable groupId or userID which is causing the problem here.

timar (approved):

PR #5770, opened 2023-01-27
Sidebar: Fix to allow ratio to span across columns

pedropintosilva:
There is an element displayed as a column before ratio
and we then hide it which is fine but, when ratio label
becomes to lengthly (Seitenverhältnis beibehalten) it then
gets cropped due to lack of space.
- Make it span across columns
- Do not align to the end
- Remove left margin from checkbox

Change-Id: I02ec7a608c0b330a52fa70db3b453fe04fefa805

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -220,9 +220,16 @@ div.sidebar.ui-grid .checkbutton.sidebar,
 
 /* writer */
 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
normally nothing gets cropped but if the sidebar has scroll then it gets cropped 
image
 this means #5749 doesn't fixed for all cases so here is another PR

eszkadev (approved):

PR #5778, opened 2023-01-29
private/ash/tests

Ashod:
- wsd: test: wrap DocBroker destroy callback
- wsd: test: start next test only after destroying the current


vmiklos (approved):
I don't see anything obviously problematic with these changes.

PR #5782, opened 2023-01-30
I don't see the reason to use the intermediate libonline.a for WASM

tml1024:
Change-Id: I520aa232f6ac51a6a57f36d8ac775150aa095a4d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
iOS build failure is caused by ENOSPC.

PR #5783, opened 2023-01-30
WASM: Copy all the necessary files from builddir\wasm

tml1024:
to browser\dist, otherwise it has to be done by hand after every wasm build.

Signed-off-by: Balázs Varga (allotropia) <balazs.varga.extern@allotropia.de>
Change-Id: I46c432116d0d42016dfc1573009d284f82c52114

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Still "No space left on device" for the iOS build, so let's ignore that.

PR #5794, opened 2023-02-01
Add access keys to tabbed view

pedropintosilva:
Change-Id: Ia3c751b27e5b3c6ae9a409c144479006d1c33d53

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -12,64 +12,76 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 				'text': _('~File'),
 				'id': '-1',
 				'name': 'File',
+				'shortcut': 'f'
 			},
 			{
 				'text': _('~Home'),
 				'id': this.HOME_TAB_ID,
 				'name': 'Home',
-				'context': 'default|Text|DrawText'
+				'context': 'default|Text|DrawText',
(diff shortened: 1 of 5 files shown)


browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -12,64 +12,75 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
...
 				'name': 'File',
+				'shortcut': '1'
 			},
 			{
 				'text': _('~Home'),
 				'id': this.HOME_TAB_ID,
 				'name': 'Home',
-				'context': 'default|Text|DrawText'
+				'context': 'default|Text|DrawText',
+				'shortcut': '2'
```
pedropintosilva, line 22: I have tried with 'f' and 'h' -> it worked (`alt` + `shift` + `letter`) for the cases like `h` but not `f` :
* it seems t conflict with browser's `alt` + `f` in Collabora Online (but when I attempted in a fresh file html/js file it worked)
* There are more than on item starting with `F` which is not a problem in itself but coming with good keys seem problematic (for example on the core side `Format` and `Form` have the same access key which makes it impossible to open `Format` with access key

So maybe number would be a good option since it fixes all the problems above mentioned but we would then need to a Key listener (or reuse one) for  when `alt` + `shift` are pressed and then show a visual hint (circle with a number) inside

PR #5795, opened 2023-02-01
private/pedro/fix hyperlink ux plus required

pedropintosilva:
- Do not error out when inserting empty hyperlinks
- Fix the insertion of ghost hyperlinks


eszkadev (approved):
looks good

PR #5799, opened 2023-02-01
Simplify iOS application test builds in Xcode

plubius:
This commit makes the following changes to the Xcode project:

- Use the correct application bundle ID.
- Link zstd to a local build like is done for poco to avoid dynamically linking libzstd in /opt/local/lib or /usr/local/lib.
- Eliminate the need to manually edit paths in the Xcode project file by removing the core group of files from Xcode's dependency list and adding the lobuilddir-symlink softlink in the applicable Xcode include and link search paths. Note: this change may cause Xcode to fail to detect when a LibreOffice file has changed but cleaning and rebuilding the iOS application in Xcode takes only a minute or so on a Silicon Mac so this seems like a reasonable tradeoff to avoid needing to manually edit paths in the Xcode project file.

Change-Id: I84ca8e567a0936a9ea106c5dfbeb1b15f930edd2
Signed-off-by: Patrick Luby <patrick.luby@collabora.com>

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


plubius:
More details about why I remove the core group on files:

The previous Xcode project file tried to list a large number of LibreOffice files as build dependencies and the list has grown out of date. Some files no longer exist in LibreOffice and some new ones weren't in the list.

Since we aren't actually building the LibreOffice files in Xcode (they should already have been built before Online is built, my thinking was that trying to resync the LibreOffice dependency list would be a large maintenance task for a build that can be cleaned and rebuilt in a minute.

I did verify that any Online files built in Xcode will rebuild if a LibreOffice header file is touched so Xcode is doing its own dependency checking for Online files. But interestingly, Xcode never complained about the LibreOffice files that no longer exist. I would think that Xcode would complain about a missing dependency for such files so it is uncertain if Xcode was actually doing anything with the core group files.

tml1024:
The reason why a random assortment of core files have been added over time is not because they would be ”build dependencies” (there is no attempt to compile them by the project) but so that one can set a breakpoint in them without problems. Otherwise the only way to set a breakpoint in core code is to hope that the call stack from some code in Online where you have a breakpoint contains that source file. Or by single-stepping from an Online source file until you reach the source file where you want to set a breakpoint. At least it used to be like this, but possiby it has changed in more recent Xcode versions?

plubius:
> The reason why a random assortment of core files have been added over time is not because they would be ”build dependencies” (there is no attempt to compile them by the project) but so that one can set a breakpoint in them without problems. Otherwise the only way to set a breakpoint in core code is to hope that the call stack from some code in Online where you have a breakpoint contains that source file. Or by single-stepping from an Online source file until you reach the source file where you want to set a breakpoint. At least it used to be like this, but possiby it has changed in more recent Xcode versions?

That is in interesting feature. So if there is value in keeping them, my next question is how did the list get created originally? I can set all the core group to reference paths through the lobuilddir-symlink softlink, but I don't know how to select the files that we should include in the group.

I don't know if this is a good idea, but one approach is to start clean and add back files, one by one, when we find a need to debug a file. Of course, that is a bad idea if we need to export the files' symbol map at release time.

I am open to ideas. What do you think? I would assume that uploading symbols to Apple makes debuggin easier, but my experience is only on macOS building outside of Xcode with stripped binaries so I've never had the benefit of having a symbol map generated by Xcode

tml1024:
I have added source files whenever I have needed to set a breakpoint in them;) Files that no longer exist can be removed of course, but keeping a small set of core source files listed (without being part of the build from Xcode’s POV) is not harmful.

The commit messages from my adding of core source files now and then typically say something like ”add some more core source files for debugging convenience”, IIRC.

plubius:
OK. So I still need to update the steps for building on macOS at https://github.com/CollaboraOnline/CollaboraOnline.github.io/blob/master/content/post/build-code-ios.md (I have that on my todo list) so I can add some text about the need to add files to the core group for debugging.

If I add a few core files back in the Xcode and resubmit a new PR, would that be enough? Or would you like to see other changes?

plubius:
After thinking about @tml1024's point about debugging, I now think the safest change would be for me to add back the core group of files and just update their paths to begin with ../../../lobuilddir-symlink/. So, I am closing this PR and I will submit a new PR soon.

PR #5802, opened 2023-02-02
Handle CTRL+A selection event of formulabar

GulsahKose:
Current event listeners can listen CTRL+A event on sheet but we don't have a listener for CTRL+A to update clipboard on selection changed on formula bar. As a result copy paste was not working normally.

With that patch we can update selection of clipboard object even in we are on formula bar's text input areas. And copy paste works.

Change-Id: I07221197dc9ecacc47abb5324be0f8c09505b5cb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
what exactly doesn't work in formulabar?
I've recently fixed copying from it using ctrl+a and ctrl+c in : https://github.com/CollaboraOnline/online/pull/5786

GulsahKose:
Duplicated fix.

PR #5811, opened 2023-02-05
Simplify iOS application test builds in Xcode

plubius:
Replace leading "../" in absolute paths with "$(SOURCE_ROOT)/../" to avoid the need to manually edit paths in the Xcode project file.

Also, all group files that have their group set to "path relative to group" so replace all absolute paths in group files with the appropriate number of leading "../" and replace "ios-device" in paths with "lobuilddir-symlink".

Lastly, verify that every file listed in the Xcode sidebar exists. A couple dozen files in the core group have been moved or deleted in LibreOffice so those were removed.

Signed-off-by: Patrick Luby <patrick.luby@collabora.com>
Change-Id: I6fbcca5820b02590fb80d9170b85814e14222015

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


plubius:
Added second patch.

Modify the Info.plist to ensure that CFBundleVersion is always incremented

AppStoreConnect requires each upload, whether it is released or not, to be
higher than the previous successful upload's CFBundleVersion. So this
script sets the CFBundleVersion to the first and second components of
CFBundleShortVersionString and the UTC timestamp of when this script was
run is appended as the third component (or the second component if there
isn't a second component in CFBundleShortVersionString).

plubius:
Argh! I don't know how to rebase locally without pushing other people's commits in GitHub. So, I am going to close this pull request and do a clean pull request when I have a batch of related patches to push.

PR #5822, opened 2023-02-09
Prevent double feature unlock dialog

GulsahKose:
If we have already a feature unlock dialog open, don't let open the second one.

Change-Id: I7c7235bc1a64c48ba69f11d0599bbb9839b7fe71

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Command.js
+++ b/browser/src/control/Control.Command.js
@@ -3,7 +3,8 @@
  * Feature blocking handler
  */
 
-/* global $ vex _ */
+/* global $ vex _ isAnyVexDialogActive*/
+
 L.Map.include({
 
 	Locking: {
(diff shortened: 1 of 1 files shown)


eszkadev:
It is ok I think. But I fear that when we will convert that dialog from vex to jsdialogs then we will forget about this condition and we will have the same bug again :) 
What about moving that directly to the 'open-dialog' function? 
 https://github.com/CollaboraOnline/online/blob/a8cd0ef8aba03f97227629becd66105b3f882fc1/browser/src/control/Control.Command.js#L59

GulsahKose:
Good idea. I'll update the patch thanks o/

eszkadev (approved):

PR #5825, opened 2023-02-10
admin: fix: monitor doesnot try connecting again after disconnecting

Rash419:
regression from 49bac0da174e31fcdbe5a4458f2f2100d606b6dc

Change-Id: Id006db718d49da2b1f9142c4d6d5c6985b18ecc9


kendy (approved):
Maybe we could use erase_if() instead of if (find...) {erase() } ; but looks good to me in general.  If the erase_if() is indeed better here, please do it in a follow-up PR :-)

PR #5833, opened 2023-02-12
Simplify notebookbar buttons code

eszkadev:
- this reduces code we need to show buttons (uses customtoolitem to invoke actions instead of new widget implementation per button)
- shares code between notebookbar and menubar better (without constructing menubar classes)
- makes us sure we keep focus on map after click on such buttons
- more conversions can be done later, to finally remove menubartoolitems

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -138,9 +138,10 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 					hasShare ?
 						{
 							'id': 'ShareAs',
-							'type': 'menubartoolitem',
+							'type': 'customtoolitem',
 							'text': _('Share'),
-							'command': '.uno:shareas'
+							'command': 'shareas',
+							'inlineLabel': true
 						} : {},
 					hasRevisionHistory ?
 						{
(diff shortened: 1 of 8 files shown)


browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -1092,9 +1094,9 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'vertical': 'true'
 			},
 			{
-				'type': 'bigtoolitem',
+				'type': 'bigcustomtoolitem',
 				'text': _UNO('.uno:HyperlinkDialog'),
-				'command': '.uno:HyperlinkDialog'
+				'command': 'hyperlinkdialog'
```
pedropintosilva, line 1099: With this change the button id has changed to something not ideal `Hyperlink...` can we add id here and set to `HyperlinkDialog` (as before)?

pedropintosilva (changes requested):
Everything works, thanks !

browser/src/control/Control.NotebookbarCalc.js
pedropintosilva: HTML id

eszkadev: fixed

pedropintosilva (approved):
Testes, awesome. Thanks!

PR #5836, opened 2023-02-13
Improved Media Playback

Ashod:
- wsd: log improvements around media handling
- wsd: play media when selected by other views


eszkadev (changes requested):
For me it doesn't play the video at all with this PR. I tried also with two views.
Selection of the video object doesn't show the player. (But I see I receive mediashape event)
Wasn't here intention to change the event name or merge into graphic selection but somehow it is missing in this PR?

eszkadev (approved):
works as expected, retriggered CI

PR #5849, opened 2023-02-14
wsd: .uno:Copy is allowed for internal use

Ashod:
When copying is disabled, we still allow
copy/pasting *within* the document. What
is disabled is exporting the copied data
to the browser.

Luckily, this code was broken, and it
wasn't really blocking .uno:Copy at all.

Change-Id: I50a01eefbe1de3758b4451385bbc51fbde5878a8


thebearon (approved):
Thanks, Ash!

PR #5850, opened 2023-02-14
Hyperlink dialog: Use input field instead of textarea for text field

pedropintosilva:
This change also fixes the weird cursor (since the textarea looked
like a input field with bad heigh and alignment). On top of that
there is no great reason to use a short textarea for text field
instead of a input field

Change-Id: I44ac38bce565b67ed9e0c4d16f4bd78cd50765d4


eszkadev (approved):

eszkadev:
we discussed: it's better to avoid breaking multi-paragraph hyperlinks - @pedropintosilva will create another solution

PR #5852, opened 2023-02-14
Insert link

eszkadev:
1) Insert > "Pick Link" button is shown when `CheckFileInfo` contains `EnableRemoteLinkPicker`.
    After press `UI_PickLink` is send via postMessage to integrator.
    Action should be finished with sending back `Action_InsertLink` to editor with JSON `{ url: .... }`
    Links then is inserted into document.
    
2) When user clicks on link, preview popup appears with "Loading..." text
    At that time editor sends via postMessage `Action_GetLinkPreview` with JSON `{ url: ... }`
    Integrator should send back   `Action_GetLinkPreview_Resp` with JSON ` { url: ..., title: 'some text', image: 'base64 encoded image'}`

### DIFF
--- a/browser/src/map/handler/Map.WOPI.js
+++ b/browser/src/map/handler/Map.WOPI.js
@@ -24,6 +24,7 @@ L.Map.WOPI = L.Handler.extend({
 	DownloadAsPostMessage: false,
 	UserCanNotWriteRelative: true,
 	EnableInsertRemoteImage: false,
+	EnableInsertRemoteLink: false,
 	EnableShare: false,
 	HideUserList: null,
 	CallPythonScriptSource: null,
@@ -118,6 +119,7 @@ L.Map.WOPI = L.Handler.extend({
 			overridenFileInfo.DownloadAsPostMessage : !!wopiInfo['DownloadAsPostMessage'];
 		this.UserCanNotWriteRelative = !!wopiInfo['UserCanNotWriteRelative'];
 		this.EnableInsertRemoteImage = !!wopiInfo['EnableInsertRemoteImage'];
+		this.EnableRemoteLinkPicker = !!wopiInfo['EnableRemoteLinkPicker'];
 		this.SupportsRename = !!wopiInfo['SupportsRename'];
 		this.UserCanRename = !!wopiInfo['UserCanRename'];
 		this.EnableShare = !!wopiInfo['EnableShare'];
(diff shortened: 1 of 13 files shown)


juliushaertl:
Looks sane from a first look. Note that there would also be a "description" available that we could pass for the link preview.

Filed https://github.com/nextcloud/richdocuments/issues/2813 on our side and @julien-nc will look into this for richdocuments.

eszkadev:
How it looks now with example image:
link-popup

pedropintosilva:
https://github.com/nextcloud/richdocuments/pull/2814

juliushaertl (approved):
Tested and works like a charm 👍

pedropintosilva (approved):
Overall I think we can merge this.

- Pick link feature is detected and the button is shown  or in the case of compact view the menu entry is shown
- Insert (normal) hyperlink works: preview is added to the hyperlink's tooltip

However the tooltip is positioned without taking into account the height of its preview
- Before the tooltip would appear on top or bellow depending on the space that is left (within the document)

Plus when inserting Link preview
- it needs the protocol first
- it also replaces the text (the text selection)

browser/src/map/handler/Map.WOPI.js
```diff
@@ -398,6 +400,36 @@ L.Map.WOPI = L.Handler.extend({
 				this._map.insertURL(msg.Values.url);
 			}
 		}
+		else if (msg.MessageId == 'Action_InsertLink') {
+			if (msg.Values) {
+				var command = {
+					'Hyperlink.Text': {
+						type: 'string',
+						value: 'Link'
```
pedropintosilva, line 408: 'Link' hardcoded as discussed

pedropintosilva (changes requested):

browser/src/map/handler/Map.WOPI.js
eszkadev: fixed

pedropintosilva (changes requested):
Thanks, now the text is not being replaced!

but I have found out that now the richdocuments dialog doesn't close at first enter:

https://user-images.githubusercontent.com/65948705/222406609-11f1c9ae-44ae-4107-928b-0a496045dd5a.mp4

pedropintosilva (approved):
My previous review (need to insert a link again and press enter to make the dialog close) seems to be related to richdocuments only cc: @juliushaertl  (thanks Szymon) However upon refreshing the page the bug dissapears (dialog closes as expected)

PR #5853, opened 2023-02-14
browser: Separate comment reply sections from other comments

OFFTKP:
Separate comment reply sections from other comment threads if there's multiple comment threads in the same location in a document

Signed-off-by: Paris Oplopoios <paris.oplopoios@collabora.com>
Change-Id: If817b3769eb35d730860ea2817e024efe2b684c6

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -66,7 +66,8 @@ export class CommentSection extends CanvasSectionObject {
 		this.sectionProperties.arrow = null;
 		this.sectionProperties.initialLayoutData = null;
 		this.sectionProperties.showResolved = null;
-		this.sectionProperties.marginY = 10 * app.dpiScale;
+		this.sectionProperties.marginY = 20 * app.dpiScale;
+		this.sectionProperties.marginYThread = 10 * app.dpiScale;
 		this.sectionProperties.offset = 5 * app.dpiScale;
 		this.sectionProperties.layoutTimer = null;
 		this.sectionProperties.width = Math.round(1 * app.dpiScale); // Configurable variable.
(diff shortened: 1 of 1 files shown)


OFFTKP:
Let's wait for https://gerrit.libreoffice.org/c/core/+/147021 to be merged first and backported

pedropintosilva:
> Let's wait for https://gerrit.libreoffice.org/c/core/+/147021 to be merged first and backported

I see it was merged on master, was it already backported?

pedropintosilva (changes requested):
| Before | After| 
| --------- | --------- |
| Screenshot_20230220_150828 | image |

thanks @OFFTKP it definitely looks better in terms of being able to now see clearly when there are more then a comment (not a reply) in the same place. However now the replies look like they are top level comments. Ideally we would just add that small(est) space when between top comments inserted in the same place

| Ideally |
| ------ |
| image |

pedropintosilva (changes requested):
I have tested again with make clean. It seems I was getting some weird results.  Apologies.
but somehow not all top level comments in the same stop are rendered equally:


https://user-images.githubusercontent.com/65948705/220144529-0f9a5c26-7c54-48a3-9b63-fc67e679a8ff.mp4

pedropintosilva:
I have also noticed that the css class `annotation-active` that before was being removed correctly (ise the user clicks away from the comment or if the user clicks in another comment) but now it seems that the class is somehow retained

OFFTKP:
I pushed a change, could you try it once you get time?

pedropintosilva (changes requested):
Thanks @OFFTKP :

* :heavy_check_mark: CSS class is now applied and then removed (no more multiple actibe threads, thanks!)
* :heavy_check_mark: Clicking in each top comment (attached to the same word) raises successfully that comment and its replies
* :heavy_check_mark: Clicking in a reply correctly raises that whole tread
* :heavy_check_mark: Threads have space between them
* :black_square_button: Threads that were initialized (top comments were inserted) in the same spot should have smaller space in between them. So it's visible the difference between threads attached to the same text/location and threads attached to different parts of the text

OFFTKP:
It should work with this latest change (my test process was add comment to a word, add another comment to the word, and reply to the 1st comment).

However a known bug here is that there's no differentiation (AFAIK) in odt between comments having a "parent" (as in they are in the same thread) and parent like .docx has "paraIdParent" which refers to the comment it's a reply to.
Which means that .odt with replies will not roundtrip correctly but .docx should.

pedropintosilva:
> It should work with this latest change (my test process was add comment to a word, add another comment to the word, and reply to the 1st comment).

Yes it looks good (see ODT's "Editing" screenshot)  but after reloading we get a different result (see ODT's "After reloading")


> However a known bug here is that there's no differentiation (AFAIK) in odt between comments having a "parent" (as in they are in the same thread) and parent

I see but maybe we could infer this by looking at the timestamp. Each comment has a timestamp (I can see that when opening with core) so, if we have 3 comments as part of the same thread the oldest will be always the Parent, right?

> like .docx has "paraIdParent" which refers to the comment it's a reply to. Which means that .odt with replies will not roundtrip correctly but .docx should.

Confirmed, docx remains the same (see screenshot)


## ODT

| Editing | After reloading |
| ---- | ---- |
| Screenshot 2023-02-24 at 11-19-21 Online | Screenshot 2023-02-24 at 11-43-16 Online| 

## DOCX
| Editing and after reloading, no change |
| --- |
| Screenshot 2023-02-24 at 11-38-54 Online |

PR #5859, opened 2023-02-16
zotero hide if not supported

eszkadev:


pedropintosilva (changes requested):
I have tested
| oc | NC with zotero option empty |
| --- | --- |
| UI is correctly hidden | UI is hidden (not expected)|
| | 11:37:59.040 Cookie “nc_sameSiteCookielax” with the “SameSite” attribute value “Lax” or “Strict” was omitted because of a cross-site redirect. [Control.Zotero.js] |

pedropintosilva (approved):
On a second though let's just merge this for now.

Context:
NC richdocuments still doesn't send empty string (API Key) it will @juliushaertl 

but then we will not show the snackbar also...

if api key is empty it meant to be "integrator will show popup"

so it will be non functional button without feedback when user doesn't configured API key
pedro.silva
[11:48 AM](https://chat.collabora.com/collabora/pl/adskpbwpn3bw984psofuxz8aje)

yeah it stuck they basically need to implement their dialog and at the same time provide empty string (API key) when is not set

PR #5869, opened 2023-02-17
formulabar: allow pasting from external source

eszkadev:
this fixes regression introduced in:
commit 164b148d6ee6e27acdc1f25aa9c06ef9c39b193a
formulabar: allow copying using keyboard

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -530,6 +530,12 @@ L.Clipboard = L.Class.extend({
 		return false;
 	},
 
+	_isFormulabarSelected: function() {
+		if ($('#sc_input_window').is(':focus'))
+			return true;
+		return false;
+	},
+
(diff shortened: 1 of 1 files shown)


GulsahKose:
Hi Szymon, This patch  fixes he regression. But now it causes one more problem. To reproduce new problem you can follow that steps:
1) Copy (CTRL+C) a string like "xxx" (doesn't matter internal or external)
2) Select a non empty cell from sheet.
3) Ctrl+A to formula bar
4) Ctrl+V

Instead replace all text with "xxx", it appends at the end.
If I remove that patch problem disapperars.

eszkadev:
@GulsahKose good catch! fixed that

GulsahKose:
@eszkadev Looks good to me. o/

GulsahKose (approved):

Issue #5868, opened 2023-02-17
Mobile: hide View > Formatting Marks when on read only mode 


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @eszkadev 

### Detailed description and rationale
The Formatting marks should not be visible in the hamburger menu when on read only mode.

#### Read only mode on mobile
To trigger the mobile view simply open Collabora Online on browser with the browser's responsive view on (F12 and then ctrl + shift + m )

* The read only mode is what you will see right from the start (noticed the pencil icon that when clicked enters the edit mode)
* Menu entry should not be visible on read only mode. `View` > `Formatting Marks` should be hidden as it does nothing in this mode

| Read-only mode | Hide Formatting Marks |
| --------- | --------- |
| ![Screen Shot 2023-02-17 at 11 11 06](https://user-images.githubusercontent.com/65948705/219617610-fb51b99e-f25c-4166-8351-13b2d0636f02.png) | ![Screen Shot 2023-02-17 at 11 11 16](https://user-images.githubusercontent.com/65948705/219617701-d0b7a9f6-2b24-4061-8ac3-f85234827650.png) |

### Code pointers
The menu is created in `browser>src>control>Control.Menubar.js` . That menu option is called `ControlCodes`
In that file ^ items' visibility is being controled by `_checkItemVisibility`. Check `	if (this._map.isReadOnlyMode()) {`

pedropintosilva:
Let me assign this to you since you already have a PR @bayramcicek :)

bayramcicek:
> Let me assign this to you since you already have a PR @bayramcicek :)

Sure :)

Cypress tests seem failed again.

PR #5874, opened 2023-02-18
hide Formatting Marks when on read-only mode

bayramcicek:
On mobile, the Formatting marks should not be visible in the hamburger menu when on read-only mode.

- new menuItem.id added for Formatting Marks as "formattingmarks"
- the id is added to the switch case so that it can be hidden on read-only mode

Change-Id: I17098e0cd0b59556ab85d93b51745ff376f6156e

* Resolves: #5868
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Awesome, thanks!

pedropintosilva:
now retriggering the cypress tests to see if it passes

pedropintosilva:
I'm merging this since:
* insert formatting marks triggers always firs mobileHelper.enableEditingMobile(); and so this change doesn't affect that test
* And because I have ran the hamburger test `make -C cypress_test run-mobile spec=writer/hamburger_menu_spec.js` and everything passed

PR #5878, opened 2023-02-20
Drop POCO_HAVE_FD_EPOLL for Emscripten

tml1024:
Not sure why I now suddenly needed this, too, when I recompiled Poco for WASM (with different options than before).

Change-Id: I7221947e0cd2264535b6d34e0640b95d7cfc7cf4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


tml1024:
Seems that we have false Cypress failures again, yay.

PR #5879, opened 2023-02-21
Fixes for bugs in TestFlight2 release

plubius:
Swap the ID names of the font name combobox elements and don't add the "select2-selection--single" class.

Unfortunately, this change does not set the height of the font name combobox to the font size combobox. I made a few tries at overriding the CSS that sets the height of buttons to 36px but I was unsuccessful.


pedropintosilva (approved):
Tested on ipad. It works. Thanks!

PR #5883, opened 2023-02-21
Fix writer/help_dialog_update_spec.js spec issues

NickWingate:
* Resolves: # 
* Target version: master 

### Summary

Has several problems stopping tests from passing:
~~- [ ] Running entire spec causes all 11 tests to hang on 'pending' with no progress (may just be an issue on my end)~~
    - Just needed the `UPDATE_SCREENSHOT=true` tag
- [x] `Document Repair` test asserts for a tunneled dialog which is now a jsdialog
- [x] `Comment` test tries to click based on a selector which could be three different elements
- [ ] 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
LGTM!

Issue #5838, opened 2023-02-13
iOS App: Font selevtion in "Home" tab not working


nikslor:
**Describe the bug**
The font selection in the "Home" tab is not working in the current TestFlight version (22.05.202302092239).

IIRC it brought up the native iOS font selection (Tor should know more about this) - but I'm not sure anymore and currently it just doesn't do anything.

**To Reproduce**
1. Open a writer document using the TestFlight iOS app
2. Tap on the "Home" tab
3. Tap on the font selection widget

**Expected behavior**
A dialog with a font selection should be shown.

**Actual behavior**
Nothing happens.

**Screenshots**
![image](https://user-images.githubusercontent.com/1211300/218396008-21edd89b-63dc-4b06-8457-93dd5695fc70.png)

**Tablet**
 - OS: iPadOS 15.6.1
 - Version: 22.05.202302092239

plubius:
I see this too. The Font Size toolbar button renders as a combobox and displays a drop down list when you tap it so I would think that the Font toolbar button would look and behave the same as the Font Size toolbar button.

plubius:
By accident, I found that if you switch to Compact view (View > Compact view) and then switch back (View > Use Tabbed View), the Font toolbar button works. So my current theory is that the Font toolbar button is disabled and is waiting for an event that never arrives to enable it.

nikslor:
Works now, tested with 22.05.10 (22.2.202302182232).
Should I close it, or do you want @plubius?

plubius:
@nikslor Thank you for testing my fix. I think I will leave the bug open until the fix is released in the regular App Store. I added a milestone as a hacky way to note that the bug is fixed and tested but is still in TestFlight.

thebearon:
@plubius @nikslor I'm seeing two issues with the latest Testflight build:
- the font name box is considerably taller than the font size one,
- after picking a font from the font picker popup, while the font is updated in the document, the popup itself with the font list isn't dismissed, and I have to separately tap on Cancel.

Are these only affecting me?

nikslor:
@thebearon I also see the two issues you mentioned in 22.05.10 (22.2.202302180032)

plubius:
I see both issues as well.

The first issue is the generated HTML is probably missing some style that limits the height of the button. I'll look and see if there are any obvious differences between the styles used for the font size button.

The second issue is likely a native code issue. I will try debugging that.

plubius:
[...]

PR #5884, opened 2023-02-21
Fix font name combobox style bug in issue #5838

plubius:
Swap the ID names of the font name combobox elements and don't add the "select2-selection--single" class.

Unfortunately, this change does not set the height of the font name combobox to the font size combobox. I made a few tries at overriding the CSS that sets the height of buttons to 36px but I was unsuccessful.

Change-Id: Iee8b51a114f5a89627c7b83a4c937c71cad4b213

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -234,9 +234,9 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 		if (commandName === '.uno:CharFontName') {
 			if (window.ThisIsTheiOSApp) {
 				if (state === '')
-					$('#fontnamecombobox').html(_('Font Name'));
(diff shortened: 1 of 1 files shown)


pedropintosilva:
> Unfortunately, this change does not set the height of the font name combobox to the font size combobox. I made a few tries at overriding the CSS that sets the height of buttons to 36px but I was unsuccessful

I'm not sure I understood. Are you referring to `#fontnamecombobox`?

pedropintosilva (approved):
Tested on an ipad. it works, thanks!

plubius:
> I'm not sure I understood. Are you referring to `#fontnamecombobox`? but to I didn't see any difference in size between that and `#fontsize`: 

Interesting. In your screen snapshot, #fontnamecombobox is an HTML select whereas on my iPad it is a button: Screen Shot 2023-02-21 at 10 20 22 AM

PR #5890, opened 2023-02-23
Remove template list URL from application settings

plubius:
Change-Id: I72cfbde2a0a062d49e89ce81999785b2bffc0d1f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/ios/Mobile/AppDelegate.mm
+++ b/ios/Mobile/AppDelegate.mm
@@ -254,6 +254,7 @@ static void updateTemplates(NSData *data, NSURLResponse *response)
             templateListURL = nil;
     }
 
+#if 0 // Remove template list URL from application settings
     if (templateListURL == nil)
         templateListURL = [[NSUserDefaults standardUserDefaults] stringForKey:@"templateListURL"];
 
@@ -278,6 +279,7 @@ static void updateTemplates(NSData *data, NSURLResponse *response)
                     }] resume];
         }
     }
+#endif
 
     // Also look for managed configuration setting of the user name.
 
--- a/ios/Mobile/Resources/Settings.bundle/Root.plist.in
+++ b/ios/Mobile/Resources/Settings.bundle/Root.plist.in
@@ -38,6 +38,7 @@
 			<key>Title</key>
 			<string>Settings</string>
 		</dict>
+		<!-- Remove template list URL from application settings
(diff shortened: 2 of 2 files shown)


plubius:
@tml1024 @timar Is this what you were expecting to see after removal of the template list URL setting?:

Screenshot 2023-02-22 at 9 03 50 PM

tml1024:
Yes, and then also all the newly dead code related to that can be removed. The download(), downloadTemplate(), updateTemplates() functions. But maybe we first should ponder how to figure out if any user actually is using that feature... Oh well, if we remove it, and somebody complains, we just revert the removal;)

plubius:
> Oh well, if we remove it, and somebody complains, we just revert the removal;)

I would vote leaving commented out code in for at least a few more TestFlight releases so that it can quickly be reverted if someone is still using that feature. If someone is still using it, I wouldn't be surprised if we won't hear about it until after we release it through the regular App Store.

tml1024:
Well, experience has shown that code just left in #if 0 or commented out tends to stay that way for long. That is one reason why we use version control, so that we don't need to comment out code, but can simply remove it. It is trivial to find. Code that is commented out or ifdeffed out is not compiled anyway, so it might bit-rot because of changing APIs, and restoring it is not any more trivial than restoring it from the commit that removed it. But if you *promise* to remove it for real later this year, OK... ;)

plubius:
> Well, experience has shown that code just left in #if 0 or commented out tends to stay that way for long. That is one reason why we use version control, so that we don't need to comment out code, but can simply remove it. It is trivial to find.

I get your point. My concern in this case is that this requirement to purge this feature is very sudden and I don't see any compelling reason to remove a feature that doesn't seem to be causing anyone any harm.

Maybe there are good reasons to remove this feature but I do not know what they are. I am trusting your judgement that this is a desirable change to make.  But since I will likely be on the sharp point of any user pushback, I would like to hedge my bets.

nikslor:
I'm in favor of removing it as well.

plubius:
@tml1024 @nikslor OK. How about we do this:

1. Commit this pull request and include it in the next TestFlight build
2. I submit a separate pull request to the master branch that completely removes the disabled code

Sound good?

Issue #5851, opened 2023-02-14
use getrandom function


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @Ashod 

### Detailed description and rationale
Currently we need the capability CAP_MKNOD in order to create our jails - but we only use this to create /dev/random and /dev/urandom. There is however no need for these nodes on modern systems:
   https://lwn.net/Articles/711013/ 
Suggests this is un-necessary - and we can detect and use 'getrandom' on Linux, and getentropy on BSD which should use direct sys-calls. That should let us drop this un-necessary capability.

### Code pointers
git grep /dev/urandom # in online - and also in core.
git grep CAP_MKNOD

Hopefully not a horribly hard one. Quite probably we want to test hard vs. eg. inserting https:// URLs to ensure our required openssl / nss  etc. libraries can live without /dev/urandom and /dev/random.

Ashod:
Worth mentioning that we also need to test password-protected files and document signing, which both depend on /dev/[u]random to be available.

Ashod:
While `getrandom()` can replace `/dev/[u]random` on Linux, we still need to support non-Linux systems. So, the correct way to support this is to check in configure.ac for `getrandom()` and only then enable using it, via `#ifdef` compiler directives.

We cannot remove the old code, just yet.

Ashod:
The Linux side is now supported, thanks to #5897 and @bayramcicek. However, BSD is still missing, so will keep this open until we support BSD as well.

arrowd:
Thanks for not forgetting FreeBSD when adding new features!

PR #5897, opened 2023-02-23
use getrandom() instead of /dev/[u]random

bayramcicek:
Change-Id: Id64b35a02718f59014b18f5c667a7465bcf8cb43

* Resolves: #5851
* Target version: master 

### Summary
- instead of using /dev/[u]random devices, use
getrandom() and/or getentropy() to make direct
system calls if sys/random.h available.

- if getrandom() fails, we need to fall back to 
"/dev/[u]random" approach.

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -16,6 +16,11 @@
 
 #include <csignal>
 #include <poll.h>
+
+#ifdef HAVE_SYS_RANDOM_H
+#  include <sys/random.h>
+#endif
+
 #ifdef __linux__
 #  include <sys/prctl.h>
 #  include <sys/syscall.h>
@@ -125,15 +130,23 @@ namespace Util
 
             // a poor fallback but something.
             std::vector<char> random = getBytes(length);
-            int fd = open("/dev/urandom", O_RDONLY);
             int len = 0;
-            if (fd < 0 ||
-                (len = read(fd, random.data(), length)) < 0 ||
-                std::size_t(len) < length)
+#ifdef HAVE_SYS_RANDOM_H
+            len = getrandom(random.data(), length, GRND_NONBLOCK);
+
+            // if getrandom() fails, we fall back to "/dev/[u]random" approach.
+            if (len != length)
+#endif
             {
-                LOG_ERR("failed to read " << length << " hard random bytes, got " << len << " for hash: " << errno);
+                const int fd = open("/dev/urandom", O_RDONLY);
+                if (fd < 0 ||
+                    (len = read(fd, random.data(), length)) < 0 ||
+                    std::size_t(len) < length)
+                {
+                    LOG_ERR("failed to read " << length << " hard random bytes, got " << len << " for hash: " << errno);
+                }
+                close(fd);
             }
-            close(fd);
 
             hex.rdbuf()->setLineLength(0); // Don't insert line breaks.
             hex.write(random.data(), length);
--- a/configure.ac
+++ b/configure.ac
@@ -1080,6 +1080,26 @@ else
     AC_DEFINE([HAVE_STD_FILESYSTEM_EXPERIMENTAL],0,[Whether the std::filesystem is in the experimental header.])
 fi
 
+SYS_RANDOM=
+AC_LANG_PUSH([C++])
+AC_LINK_IFELSE(
+    [AC_LANG_SOURCE([[
+    #include <sys/random.h>
+    int main()
+    {
+        char buffer[8];
+        getrandom(buffer, sizeof(buffer), GRND_NONBLOCK);
+    }
+    ]])],
+    [SYS_RANDOM=TRUE])
+AC_LANG_POP([C++])
+
+if test "$SYS_RANDOM" == "TRUE" ; then
+    AC_DEFINE([HAVE_SYS_RANDOM_H],1,[Define to 1 if you have the <sys/random.h> header file.])
+else
+    AC_DEFINE([HAVE_SYS_RANDOM_H],0,[Define to 1 if you have the <sys/random.h> header file.])
+fi
+
 AS_IF([test -n "$LOKIT_PATH"],
       [CPPFLAGS="$CPPFLAGS -I${LOKIT_PATH}"])
 lokit_msg="$LOKIT_PATH"


bayramcicek:
I commented-out the code pieces to reveal some code pointers. ^

common/Util.cpp
```diff
@@ -21,9 +21,11 @@
 #  include <sys/syscall.h>
 #  include <sys/vfs.h>
 #  include <sys/resource.h>
+#  include <sys/random.h>
```
bayramcicek, line 24: according to [https://man7.org/linux/man-pages/man2/getrandom.2.html](https://man7.org/linux/man-pages/man2/getrandom.2.html) I added `#include <sys/random.h>` but some tests fail as: `compiler couldn't find the file »sys/random.h« referenced from common/Util.cpp:24:26`

common/Util.cpp
```diff
@@ -125,15 +127,14 @@ namespace Util
 
             // a poor fallback but something.
             std::vector<char> random = getBytes(length);
-            int fd = open("/dev/urandom", O_RDONLY);
+            int fd = getrandom(random.data(), length, GRND_NONBLOCK);
```
bayramcicek, line 130: Is this proper usage of `getrandom()`? I made a lot of research on the internet but couldn't find any proper migrations from `/dev/[u]random` to `getrandom()/getentropy()` on other projects (in order to have an idea).

common/Util.cpp
Ashod: This would mean that your system doesn't have this header. What distro are you using, and what version is it? (Notice that this function is only available on Linux.)

common/Util.cpp
Ashod: The function is straightforward. I assume you spent some time reading the documentation at https://man7.org/linux/man-pages/man2/getrandom.2.html. If not, please read it a couple of times.

The return value is the number of bytes that was written. This is equivalent to `len` below. Furthermore, since `fd` was the file descriptor, which you have deleted, we also need to delete the `read` below, since there is no longer a file to read.

Ashod (commented):
Thanks @bayramcicek! Good start. I left replies to your questions. Feel free to ask anything further, as necessary.

common/Util.cpp
bayramcicek: The header file works fine on my system(ManjaroLinux 22.0.4). Some tests here -on Jenkins- fails and don't recognize `#include <sys/random.h>` . for example: (GCC 6.3.0 check — FAILURE)  [https://cpci.cbg.collabora.co.uk:8080/job/github_online_master-gcc-6.3.0/5611/](https://cpci.cbg.collabora.co.uk:8080/job/github_online_master-gcc-6.3.0/5611/). I just wonder why GCC couldn't find the header.

Thanks!

Ashod: Right, I meant to add this in the issue report but forgot. We need to support non-linux systems. So, we can't remove the old approach completely. We need to wrap them in `#ifdef` statement (you'll find examples in the code). Also, I think we need to check for `sys/random.h` in configure.ac so we fallback.

configure.ac
```diff
@@ -103,6 +103,10 @@ for MODULE in lxml polib; do
     fi
 done
 
+AC_CHECK_HEADERS([sys/random.h],
+                [],
+                [AC_MSG_NOTICE([cannot find sys/random.h])])
+
```
bayramcicek, line 109: I added a check for sys/random.h header. we can use `#ifdef HAVE_SYS_RANDOM_H` directive.

common/Util.cpp
```diff
@@ -16,6 +16,11 @@
 
 #include <csignal>
 #include <poll.h>
+
+#ifdef HAVE_SYS_RANDOM_H
+#  include <sys/random.h>
+#endif
+
```
bayramcicek, line 23: But `HAVE_SYS_RANDOM_H` is always disabled (even on my linux system). When I run `make` I can see that my system finds `sys/random.h` header but here `sys/random.h` never includes because `HAVE_SYS_RANDOM_H` is defined in:
- `config_unused.h`       as `#define HAVE_SYS_RANDOM_H 1`
- `config_unused.h.in`   as `#undef HAVE_SYS_RANDOM_H`
and those two files are ignored in .gitignore so you can't see HAVE_SYS_RANDOM_H anywhere.

I don't know why `AC_CHECK_HEADERS([sys/random.h]` define `HAVE_SYS_RANDOM_H` in files that ignored, instead of defining them in e.g. `config.h.in` (or some other files that visible and not ignored by .gitignore).
Hi @Ashod; is there any solution for this or am I doing something wrong here? Thanks.

Ashod: I imagine the better approach is not just checking for the header, but to actual check that the function exists and compiles. The way to do that is to have a minimal code that must compile for the feature to be enabled. We do this for other features. For example, see how it's done to check for C++17. Something similar should work in this case, with the result of having some compile-time directive set. The code can then have `#ifdef` based on that directive to enable this code vs the existing one.

Hope this helps.

common/Util.cpp
```diff
@@ -125,16 +130,23 @@ namespace Util
 
             // a poor fallback but something.
             std::vector<char> random = getBytes(length);
-            int fd = open("/dev/urandom", O_RDONLY);
             int len = 0;
+#ifdef HAVE_SYS_RANDOM_H
+            len = getrandom(random.data(), length, GRND_NONBLOCK);
+            if(len < 0)
```
Ashod, line 136: FYI, we use clang-format to format code (there .clang-format file in the project). (There is always a space after `if`, `for`, and `while`.)

common/Util.cpp
```diff
@@ -125,16 +130,23 @@ namespace Util
...
+            {
+                LOG_ERR("The requested entropy was not available: " << errno);
+            }
+#else
+            int fd = open("/dev/urandom", O_RDONLY);
```
Ashod, line 141: It would be great if we fall back to the "/dev/[u]random" approach, instead of failing altogether.

configure.ac
```diff
@@ -1080,6 +1080,28 @@ else
...
 
+SYS_RANDOM=
+AC_LANG_PUSH([C++])
+AC_LINK_IFELSE(
+    [AC_LANG_SOURCE([[
+    #include <sys/random.h>
+    #include <vector>
+    int main()
+    {
+        std::vector<char> buffer(1);
```
Ashod, line 1091: No need for `std::vector`, just use a `char buffer[8];`

Notice that this code is actually wrong. It won't matter since we aren't running it, it's just for testing the compile stage. Still, it's invalid code (there is a buffer overflow).

Here is a simpler version:
```
int main()
{
    char buffer[8];
    getrandom(buffer, sizeof(buffer), GRND_NOBLOCK);
}
```

Ashod (commented):
Thanks @bayramcicek for the update. I left a couple of suggestions, but it's getting close.

Most important one is that I think we should use the old method when/if getrandom fails to give us the data we need. If it works, great, if it doesn't we still have a fallback.

bayramcicek:
Updated.
One last question: Should we need to use getentropy() for BSD systems?

Thank you for your time and reviews!

Ashod (approved):
Thank you @bayramcicek. Looks very good.

We have two follow up tasks to this one:

1) Support BSD.
2) After this is well tested for a certain period of time, we can remove `/dev/[u]random` in Linux and BSD. (Probably after a few months, at least.)

Feel free to do the BSD, if you want. But feel no obligation otherwise.

Thanks again.

PR #5903, opened 2023-02-24
Link target

eszkadev:
- endpoint extract-link-targets
- endpoint get-thumbnail
- jump to target after document is opened

### DIFF
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -1029,6 +1029,11 @@ function onCommandResult(e) {
 	} else if (map.zotero && commandName === '.uno:DeleteBookmark' && e.result.DeleteBookmark.startsWith('ZOTERO_BREF_')) {
 		map.zotero.setCustomProperty(e.result.DeleteBookmark, '');
 		map.zotero.handleRefreshCitationsAndBib(false);
+	} else if (commandName === '.uno:OpenHyperlink') {
+		// allow to process other incoming messages first
+		setTimeout(function () {
+			map._docLayer.scrollToPos(map._docLayer._visibleCursor.getNorthWest());
+		}, 0);
 	}
 }
 
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -283,7 +283,7 @@ bool ChildSession::_handleInput(const char *buffer, int length)
     }
     else if (tokens.equals(0, "getthumbnail"))
     {
-        if (tokens.size() < 2)
+        if (tokens.size() < 3)
         {
             sendTextFrameAndLogError("error: cmd=getthumbnail kind=syntax");
             return false;
@@ -295,16 +295,14 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             return false;
         }
 
-        int part = -1;
-        std::string timestamp, doctemplate;
-        parseDocOptions(tokens, part, timestamp, doctemplate);
+        int x, y;
+        if (!getTokenInteger(tokens[1], "x", x))
+            x = 0;
 
-        assert(!getDocURL().empty());
-        assert(!getJailedFilePath().empty());
+        if (!getTokenInteger(tokens[2], "y", y))
+            y = 0;
 
         bool success = false;
-        const int x = 0;
-        const int y = 0;
         const int width = 120;
         const int height = 120;
         const auto mode = static_cast<LibreOfficeKitTileMode>(getLOKitDocument()->getTileMode());
@@ -319,13 +317,23 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             oss << "sendthumbnail:\n";
 
             // encode PNG to base64
-            Poco::Base64Encoder encoder(oss);
-            encoder.rdbuf()->setLineLength(0);
-            encoder.write(pngThumbnail.data(), pngThumbnail.size());
-            encoder.close();
+            try
+            {
+                Poco::Base64Encoder encoder(oss);
+                encoder.rdbuf()->setLineLength(0);
+                encoder.write(pngThumbnail.data(), pngThumbnail.size());
+                encoder.close();
 
-            std::string sendThumbnailCommand = oss.str();
-            success = sendTextFrame(sendThumbnailCommand.data(), sendThumbnailCommand.size());
+                std::string sendThumbnailCommand = oss.str();
+                success = sendTextFrame(sendThumbnailCommand.data(), sendThumbnailCommand.size());
+            }
+            catch (const std::exception& e)
+            {
+                LOG_ERR("Encoding thumbnail failed: " << e.what());
+                std::string error = "sendthumbnail: error";
+                sendTextFrame(error.data(), error.size());
+                success = false;
+            }
         }
 
         return success;
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -304,10 +304,18 @@ bool ChildSession::_handleInput(const char *buffer, int length)
         bool success = false;
         const int width = 120;
         const int height = 120;
+
+        // These arbitrary magic numbers are from the original interrupted WIP, surely this will need more work.
+        constexpr float zoom = 0.5f;
+        constexpr int pixelWidthTwips = width * 15 / zoom;
+        constexpr int pixelHeightTwips = height * 15 / zoom;
+        constexpr int offsetXTwips = 15 * 15; // start 15 px/twips before the target to get a clearer thumbnail
+        constexpr int offsetYTwips = 15 * 15;
+
         const auto mode = static_cast<LibreOfficeKitTileMode>(getLOKitDocument()->getTileMode());
 
         std::vector<unsigned char> thumbnail(width * height * 4);
-        getLOKitDocument()->paintThumbnail(thumbnail.data(), x, y);
+        getLOKitDocument()->paintTile(thumbnail.data(), width, height, x - offsetXTwips, y - offsetYTwips, pixelWidthTwips, pixelHeightTwips);
 
         std::vector<char> pngThumbnail;
         if (Png::encodeBufferToPNG(thumbnail.data(), width, height, pngThumbnail, mode))
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -24,6 +24,7 @@
 #include <Poco/URI.h>
 #include <Poco/BinaryReader.h>
 #include <Poco/Base64Decoder.h>
+#include <Poco/Base64Encoder.h>
 #if !MOBILEAPP
 #include <Poco/Net/HTTPResponse.h>
 #include <Poco/Net/HTTPSClientSession.h>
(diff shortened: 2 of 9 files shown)


browser/src/control/Control.Toolbar.js
```diff
@@ -1029,6 +1029,11 @@ function onCommandResult(e) {
 	} else if (map.zotero && commandName === '.uno:DeleteBookmark' && e.result.DeleteBookmark.startsWith('ZOTERO_BREF_')) {
 		map.zotero.setCustomProperty(e.result.DeleteBookmark, '');
 		map.zotero.handleRefreshCitationsAndBib(false);
+	} else if (commandName === '.uno:OpenHyperlink') {
+		// allow to process other incoming messages first
+		setTimeout(function () {
+			map._docLayer.scrollToPos(map._docLayer._visibleCursor.getNorthWest());
```
Ashod, line 1035: I think we should avoid the timeouts and instead have a queue, as we already do with messages.

kit/ChildSession.cpp
```diff
@@ -249,6 +250,84 @@ bool ChildSession::_handleInput(const char *buffer, int length)
...
+            sendTextFrameAndLogError("error: cmd=extractlinktargets kind=docnotloaded");
+            return false;
+        }
+
+        assert(!getDocURL().empty());
+        assert(!getJailedFilePath().empty());
+
+        const char *data = _docManager->getLOKit()->extractRequest(getJailedFilePath().c_str());
+        LOG_TRC("Extracted link targets: " << data);
+        bool success = sendTextFrame("extractedlinktargets: " + std::string(data));
```
Ashod, line 272: This will blow up when `data` is null. Also, `data` needs to be freed. This is a memory leak.

kit/ChildSession.cpp
```diff
@@ -249,6 +250,84 @@ bool ChildSession::_handleInput(const char *buffer, int length)
...
+        if (!getTokenInteger(tokens[2], "y", y))
+            y = 0;
+
+        bool success = false;
+        const int width = 120;
+        const int height = 120;
+        const auto mode = static_cast<LibreOfficeKitTileMode>(getLOKitDocument()->getTileMode());
+
+        std::vector<unsigned char> thumbnail(width * height * 4);
+        getLOKitDocument()->paintThumbnail(thumbnail.data(), x, y);
```
Ashod, line 302: It would be far better to pass the size of the thumbnail to the API. Having it implicit makes it fragile.

kit/ChildSession.cpp
```diff
@@ -249,6 +250,84 @@ bool ChildSession::_handleInput(const char *buffer, int length)
...
+            {
+                Poco::Base64Encoder encoder(oss);
+                encoder.rdbuf()->setLineLength(0);
+                encoder.write(pngThumbnail.data(), pngThumbnail.size());
+                encoder.close();
+            }
+            catch (const std::exception& e)
+            {
+                LOG_ERR("Encoding thumbnail failed: " << e.what());
+                // trigger bad data exception in decoder so we will send HTTP 500
```
Ashod, line 321: Let's please not trigger corner cases in other parts of the code to do control flow. This is a bad idea for many reasons.

Instead, we should send back an error message, for example: `sendthumbnail: failed to generate thumbnail`.

The decoder can then properly parse this, since the API for messages is that binary data is started at a new line, but the first line is always part of the command/message and its parameters. We already do this for other commands/messages.

wsd/COOLWSD.cpp
```diff
@@ -4490,13 +4516,14 @@ class ClientRequestDispatcher final : public SimpleSocketHandler
                 }
 
                 std::string lang = (form.has("lang") ? form.get("lang") : "");
+                std::string target = (form.has("target") ? form.get("target") : "");
```
Ashod, line 4519: Minor: best to use `std::string()` instead of `""`.

wsd/ClientSession.cpp
```diff
@@ -2049,6 +2091,64 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+            httpResponse.setBody(stringJSON, "application/json");
+            _saveAsSocket->sendAndShutdown(httpResponse);
+
+            LOG_TRC("Removing export-link-targets ClientSession.");
+
+            // Remove us.
+            docBroker->removeSession(client_from_this());
+
+            // Now terminate.
+            docBroker->stop("Finished extracting link targets handler.");
```
Ashod, line 2112: Neither of these methods is the proper way to do this. If we want to end the docBroker instance, we should handle the message there and properly call `closeDocument()`.

wsd/ClientSession.cpp
```diff
@@ -2049,6 +2091,64 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+            return true;
+        }
+        else if (tokens.equals(0, "sendthumbnail:")) {
```
Ashod, line 2115: We should use clang-format to have consistent formatting. This is not the style that the project uses.

wsd/ClientSession.cpp
```diff
@@ -2049,6 +2091,64 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+
+            int firstLineSize = firstLine.size() + 1;
+            std::string base64thumbnail(payload->data().data() + firstLineSize, payload->data().size() - firstLineSize);
+
+            // decode back to PNG
+            try
+            {
+                std::istringstream istr(base64thumbnail);
+                std::ostringstream ostr;
+                Poco::Base64Decoder b64in(istr);
```
Ashod, line 2126: Do we really need to encode and then decode? Why can't we send the binary data as-is?

wsd/ClientSession.cpp
```diff
@@ -2049,6 +2091,64 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+                http::Response httpResponse(http::StatusCode::InternalServerError);
+                httpResponse.set("Content-Length", "0");
+                _saveAsSocket->sendAndShutdown(httpResponse);
+            }
+
+
+            LOG_TRC("Removing get-thumbnail ClientSession.");
+
+            docBroker->removeSession(client_from_this());
+            docBroker->stop("Finished get-thumbnail handler.");
```
Ashod, line 2150: Same.

wsd/DocumentBroker.cpp
```diff
@@ -3603,18 +3603,46 @@ bool ConvertToBroker::startConversion(SocketDisposition &disposition, const std:
...
+    std::vector<char> loadRequest(_load.begin(), _load.end());
+    clientSession->handleMessage(loadRequest);
+}
+
+void ExtractLinkTargetsBroker::sendStartMessage(std::shared_ptr<ClientSession> clientSession, const std::string& encodedFrom)
+{
+    // load first
+    ConvertToBroker::sendStartMessage(clientSession, encodedFrom);
+
+    auto docBroker = this;
```
Ashod, line 3630: What is going on here?

wsd/DocumentBroker.cpp
```diff
@@ -3603,18 +3603,46 @@ bool ConvertToBroker::startConversion(SocketDisposition &disposition, const std:
...
+
+void GetThumbnailBroker::sendStartMessage(std::shared_ptr<ClientSession> clientSession, const std::string& encodedFrom)
+{
+    clientSession->setThumbnailSession(true);
+    clientSession->setThumbnailTarget(_target);
+
+    // load first
+    ConvertToBroker::sendStartMessage(clientSession, encodedFrom);
+
+    // then we setup cursor in correct place and trigger rendering
```
Ashod, line 3643: Is this incomplete code, or is this comment unnecessary?

wsd/DocumentBroker.hpp
```diff
@@ -1510,8 +1510,44 @@ class ConvertToBroker final : public StatelessBatchBroker
...
+    std::string _target;
+
+public:
+    /// Construct DocumentBroker with URI and docKey
+    GetThumbnailBroker(const std::string& uri,
+                    const Poco::URI& uriPublic,
+                    const std::string& docKey,
+                    const std::string& lang,
+                    const std::string& target)
+                    : ConvertToBroker(uri, uriPublic, docKey, "", "", lang)
```
Ashod, line 1545: We shouldn't initialize `std::string` with `""`. Same elsewhere.

wsd/ClientSession.cpp
```diff
@@ -2021,6 +2045,24 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+                            // cursor already in correct place
+                            std::ostringstream renderThumbnailCmd;
+                            renderThumbnailCmd << "getthumbnail x=" << x << " y=" << y;
+                            docBroker->forwardToChild(client_from_this(), renderThumbnailCmd.str());
+                        }
+                        else
+                        {
+                            // this is initial cursor position message
+                            // we set new position wait for second message
+                            setThumbnailTarget("");
```
Ashod, line 2063: It's not clear what the comment is trying to say, but `std::string()` is what is needed here. (But should the initial state of the target be an empty string anyway? The comment is confusing.)

Ashod (commented):
Thanks @eszkadev for reviving this. There are some very good bits, but there are a couple of serious issues too. Please see the notes and let's discuss as necessary.

browser/src/control/Control.Toolbar.js
eszkadev: setTimeout -> adds task to the task queue in JS...

kit/ChildSession.cpp
eszkadev: unfortunately it is hardcoded on the core side - we can improve that that in the next iteration

kit/ChildSession.cpp
eszkadev: when trying to se4nd pure binary data:
```coolforkit: ./net/WebSocketHandler.hpp:781: int WebSocketHandler::sendFrame(const std::shared_ptr<StreamSocket>&, const char*, uint64_t, unsigned char, bool) const: Assertion `"invalid utf-8 - check Message::detectType()" && false' failed.```

Comment (Ashod) [reply kit/ChildSession.cpp]: Could it be because we are calling `sendTextFrame` instead of `sendBinaryFrame` ?

File: wsd/ClientSession.cpp [hunk 14]
```diff
@@ -2049,6 +2092,76 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+            http::Response httpResponse(http::StatusCode::OK);
+            httpResponse.set("Last-Modified", Util::getHttpTimeNow());
+            httpResponse.set("X-Content-Type-Options", "nosniff");
+            httpResponse.setBody(stringJSON, "application/json");
+            _saveAsSocket->sendAndShutdown(httpResponse);
+
+            LOG_TRC("Removing export-link-targets ClientSession.");
+
+            // Remove us.
+            docBroker->removeSession(client_from_this());
```
Ashod, line 2110: There is no good reason to do this. Why do we need to remove ourselves?

wsd/ClientSession.cpp
```diff
@@ -2049,6 +2092,76 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
...
+
+            // Now terminate.
+            docBroker->closeDocument("extractedlinktargets");
+            return true;
+        }
+        else if (tokens.equals(0, "sendthumbnail:"))
+        {
+            LOG_TRC("Sending get-thumbnail response.");
+
+            int firstLineSize = firstLine.size() + 1;
```
Ashod, line 2120: This is only used when `!error`, so it can move under the `if (!error)`.

wsd/DocumentBroker.cpp
```diff
@@ -3603,18 +3603,40 @@ bool ConvertToBroker::startConversion(SocketDisposition &disposition, const std:
...
+    ConvertToBroker::sendStartMessage(clientSession, encodedFrom);
+
+    const auto command = "extractlinktargets url=" + encodedFrom;
+    forwardToChild(clientSession, command);
+}
+
+void GetThumbnailBroker::sendStartMessage(std::shared_ptr<ClientSession> clientSession, const std::string& encodedFrom)
+{
+    clientSession->setThumbnailSession(true);
+    clientSession->setThumbnailTarget(_target);
```
Ashod, line 3635: Why not set the '_thumbnailSession' flag from `setThumbnailTarget` when `_target` is not empty ? Why do we need to call a separate function explicitly?

kit/ChildSession.cpp
eszkadev: yes, it fixed that

eszkadev:
Do not merge, API change on core side will be done

kit/ChildSession.cpp
Ashod: Yes, noted. Thanks.

wsd/DocumentBroker.cpp
Ashod: I see. Thanks.

Ashod (approved):
Thanks @eszkadev. Good stuff. I haven't merged myself, in case you want to wait for the Core API to be updated and update this too. Either way, looks good to me.

tml1024:
Pushed a commit to the branch that adapts it to current core.

tml1024:
I think we could merge this branch now and continue working on the problems in master. As long as you don't invoke the new endpoints, it can't break anything, right? (Famous last words?)

mmeeks:
Lets get a first cut merged and get this into snapshot builds that are easy to play with to accelerate integration. As long as the public facing API - and -particularly- the link / URL syntax that we then go an embed into millions of documents is good and sustainable I'm happy =)

mmeeks:
So lets get it in and polish later ... =)

PR #5912, opened 2023-02-27
jsdialog: show overlay for all modals

eszkadev:
All modals shown with uiManager.showInfoModal should have overlay which helps to focus on modal dialog.


pedropintosilva (approved):
Thanks!

PR #5915, opened 2023-02-27
Hide sidebar button on readonly mode.

gokaysatir:
Change-Id: I29a899a1113c6188558e8b85bf9adde5bc9049fe

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -58,7 +58,7 @@
 	top: -9px;
 }
 
-.main-nav.readonly #table-optionstoolboxdown {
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
So this change seems reasonable and it works for compact view. However I have noticed that this doesn't work for tabbed view. It seems on tabbed view we are not add the CSS class .readonly to the nav (which leads to menu breakage)

pedropintosilva (approved):

PR #5919, opened 2023-02-28
js: added localization for jquery-ui datepicker

Rash419:
replaced jquery-ui-dist node_module with jquery-ui for localization

Change-Id: I6dd95cb8859cd2e0626195a0b87b496fe44afc49

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/layer/tile/ContentControlSection.ts
+++ b/browser/src/layer/tile/ContentControlSection.ts
@@ -68,6 +68,7 @@ export class ContentControlSection extends CanvasSectionObject {
 
 		if (json.date) {
 			this.sectionProperties.datePicker = true;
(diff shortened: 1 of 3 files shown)


browser/src/layer/tile/ContentControlSection.ts
```diff
@@ -68,6 +68,12 @@ export class ContentControlSection extends CanvasSectionObject {
 
 		if (json.date) {
 			this.sectionProperties.datePicker = true;
+			var languageAndCode = this.map['stateChangeHandler'].getItemValue('.uno:LanguageStatus');
```
timar, line 71: It's the document language not the UI language, right?
I can accept this, if desktop LibreOffice and/or MS Word work the same.

timar (approved):

PR #5927, opened 2023-03-01
implement document migration

Rash419:
Change-Id: Iefb9ed77f78b13e1558e995b3e3a2cf3022b53db

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -759,7 +759,7 @@ window.app = {
 				}.bind(this);
 				setTimeout(timeoutFn, 10000, global.indirectionUrl, that.uri);
 			} else {
-				window.app.console.debug('Indirection url: error on incoming response ' + this.status);
+				window.app.console.error('Indirection url: error on incoming response ' + this.status);
 			}
 		});
 		http.send();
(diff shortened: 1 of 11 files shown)


mmeeks (commented):
Looks plausible to me, I'd like @quikee to review in more detail. We should reach out to the JECI team who (IIRC) used this monitor API to implement locking in Alfresco to check this protocol change works for them.

Rash419:
@quikee  @mmeeks  I have not changed the order of the arguments or added new argurments in the middle of `addoc` command in the code.  I just added them to documentation`<sessionid> <username> <userid>` because they were missing from the documentation. The `<username> <userid>`  was added in this [patch](https://github.com/CollaboraOnline/online/commit/bc7fa9a8120a5fb94ff032785186f4453416ab3a#diff-0df5e0a933d8cc635eb43daf28da4ae2b7d722f245c0654d644e50f95cbe8a91R489) six years ago. They have not updated the documentation as you can see from the patch. So I have just added them to documentation. The new argument I added are at the end of the command only [see this](https://github.com/CollaboraOnline/online/pull/5927/files#diff-0df5e0a933d8cc635eb43daf28da4ae2b7d722f245c0654d644e50f95cbe8a91R555). I should have seprated the commits to relay the message clearly. Sorry for the confusion

quikee (approved):
Looks good to me

PR #5928, opened 2023-03-01
browser: add author data to calc comment

hcvcastro:
The recent changes in LO Core, when edit a comment
in Calc the author data is not preserved.

Change-Id: I4f26ef2044d28bf55ff6a00a2e585ece79c6eb9b


pedropintosilva (approved):
Tested, it seems to work. No comments in spreadsheet do not loos author name nor author image.

Context:
Previously Insert a comment in a spreadsheet document and then editing it would result in lost of that data

PR #5931, opened 2023-03-02
Private/gokay/dont vex 3 clipboard

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -10,6 +10,7 @@ L.Control.UIManager = L.Control.extend({
 	blockedUI: false,
 	busyPopupTimer: null,
 	customButtons: [], // added by WOPI InsertButton
+	modalIdPretext: 'modal-dialog-',
 
 	onAdd: function (map) {
 		this.map = map;
@@ -821,8 +822,13 @@ L.Control.UIManager = L.Control.extend({
 		app.socket._onMessage({ textMsg: 'jsdialog: ' + JSON.stringify(closeMessage) });
 	},
 
+	/// Returns generated (or to be generated) id for the modal container.
+	generateModalId: function(givenId) {
+		return this.modalIdPretext + givenId;
+	},
+
 	_modalDialogJSON: function(id, title, cancellable, widgets, focusId) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.generateModalId(id);
 		focusId = focusId ? focusId : 'response';
 		return {
 			id: dialogId,
@@ -854,7 +860,10 @@ L.Control.UIManager = L.Control.extend({
 	/// callback - callback on button press
 	/// withCancel - specifies if needs cancal button also
 	showInfoModal: function(id, title, message1, message2, buttonText, callback, withCancel) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.generateModalId(id);
+		var responseButtonId = id + '-response';
+		var cancelButtonId = id + '-cancel';
+
 		var json = this._modalDialogJSON(id, title, true, [
 			{
 				id: 'info-modal-tile-m',
@@ -879,12 +888,12 @@ L.Control.UIManager = L.Control.extend({
 				enabled: true,
 				children: [
 					withCancel ? {
-						id: 'cancel',
+						id: cancelButtonId,
 						type: 'pushbutton',
 						text: _('Cancel')
 					} : { type: 'container' },
 					{
-						id: id + '-response', // Try to keep the id unique.
+						id: responseButtonId,
 						type: 'pushbutton',
 						text: buttonText,
 						'has_default': true,
@@ -898,7 +907,7 @@ L.Control.UIManager = L.Control.extend({
 
 		var that = this;
 		this.showModal(json, [
-			{id: 'response', func: function() {
+			{id: responseButtonId, func: function() {
 				if (typeof callback === 'function')
 					callback();
 				that.closeModal(dialogId);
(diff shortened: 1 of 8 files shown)


browser/src/map/Clipboard.js
```diff
@@ -862,20 +859,16 @@ L.Clipboard = L.Class.extend({
...
 		}
-		vex.dialog.alert({
-			unsafeMessage: msg,
-			callback: function () {
-				self._map.focus();
-			}
-		});
+
+		this._map.uiManager.showInfoModal('copy paste warning');
+		document.getElementById('copy paste warning').innerHTML = msg;
```
eszkadev, line 871: element id with spaces? is that allowed?

gokaysatir: yep, that's allowed.

browser/src/control/Control.UIManager.js
```diff
@@ -884,10 +884,11 @@ L.Control.UIManager = L.Control.extend({
 						text: _('Cancel')
 					} : { type: 'container' },
 					{
-						id: 'response',
+						id: id + '-response', // Try to keep the id unique.
```
eszkadev, line 887: this ID is used below to bind callback (line 901), please update that also

browser/src/map/Clipboard.js
eszkadev: how to write CSS rule for that for example? :)

browser/src/map/Clipboard.js
```diff
@@ -887,45 +880,37 @@ L.Clipboard = L.Class.extend({
...
-			afterOpen: function() {
-				self._map.focus();
-			}
-		});
+
+		this._map.uiManager.showInfoModal('paste special dialog');
+
+		// showInfoModal creates the id in following format: 'modal-dialog-paste special dialog'.
+		// We will use this for closing the dialog.
+		this.pasteSpecialDialogId = 'modal-dialog-paste special dialog';
```
eszkadev, line 911: It is fragile in case someone will change the modal dialog code generating id.
Maybe create helper function in UIManager: `generateModalId(id)` which will be reused in all the places where we typed: `var dialogId = 'modal-dialog-' + id;`. So there will be only one place to change in the future.

then this variable is not needed anymore. you will do `closeDialog(generateModalId('paste special dialog'));`

cypress_test/integration_tests/mobile/calc/hamburger_menu_spec.js
```diff
@@ -182,19 +182,7 @@ describe('Trigger hamburger menu options.', function() {
...
-
-		cy.get('.vex-dialog-message')
-			.should('have.text', 'Please use the copy/paste buttons on your on-screen keyboard.');
-
-		cy.get('.vex-dialog-buttons .button-primary')
-			.click();
-
-		cy.get('.vex-dialog-form')
-			.should('not.exist');
+		cy.get('#mobile-wizard-content').should('exist');
```
eszkadev, line 185: maybe check if some unique id exists? (this only tells mobile wizard is opened. I'm not even sure if that container doesn't exist when not opened).
I think there is generated `modal-dialog-something`   container?

browser/src/map/Clipboard.js
gokaysatir: Let me change that in case we need special cases, thanks :)

browser/src/control/Control.UIManager.js
```diff
@@ -821,8 +822,13 @@ L.Control.UIManager = L.Control.extend({
...
 	},
 
+	/// Returns generated (or to be generated) id for the modal container.
+	generateModalId: function(givenId) {
+		return this.modalIdPretext + givenId;
+	},
+
 	_modalDialogJSON: function(id, title, cancellable, widgets, focusId) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.modalIdPretext + id;
```
eszkadev, line 831: better to use `generateModalId()` everywhere (single place to change)

browser/src/control/Control.UIManager.js
```diff
@@ -854,7 +860,10 @@ L.Control.UIManager = L.Control.extend({
 	/// callback - callback on button press
 	/// withCancel - specifies if needs cancal button also
 	showInfoModal: function(id, title, message1, message2, buttonText, callback, withCancel) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.modalIdPretext + id;
```
eszkadev, line 863: generateModalId()

browser/src/control/Control.UIManager.js
```diff
@@ -913,7 +923,7 @@ L.Control.UIManager = L.Control.extend({
 	/// buttonText - text inside OK button
 	/// callback - callback on button press
 	showInputModal: function(id, title, message, defaultValue, buttonText, callback) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.modalIdPretext + id;
```
eszkadev, line 926: generateModalId

browser/src/control/Control.UIManager.js
```diff
@@ -967,7 +977,7 @@ L.Control.UIManager = L.Control.extend({
 	/// buttonText - text inside OK button
 	/// callback - callback on button press
 	showConfirmModal: function(id, title, message, buttonText, callback) {
-		var dialogId = 'modal-dialog-' + id;
+		var dialogId = this.modalIdPretext + id;
```
eszkadev, line 980: generateModalId

eszkadev (approved):
now looks good :)

gokaysatir:
> now looks good :)

Thanks for checking the changes :)

PR #5933, opened 2023-03-02
browser: add author data to calc comment

pedropintosilva:
The recent changes in LO Core, when edit a comment
in Calc the author data is not preserved.

Change-Id: I4f26ef2044d28bf55ff6a00a2e585ece79c6eb9b


pedropintosilva:
tested it , it works as the master counter part commit. Now comments do not loose the user's name

PR #5934, opened 2023-03-02
Improving message on TTL expiry handling

pedropintosilva:
when session expires some users might need to login again
in their integrator. Do not exclude those users. Otherwise
they might try to refresh and no session renewal happens.

Change-Id: I305d185fc8e9072e52b668e12e7ea80c55e9b93d


Ashod (approved):
Looks good, thanks @pedropintosilva.

PR #5938, opened 2023-03-03
fix(SaveAs): Fix format extraction on SaveAs with filenames that contain multiple dots

juliushaertl:
* Target version: master 

### Summary

This fixes an issue where using save as to store a file with a name containing multiple dots like "example.1.odt" by adapting the logic to extract the file extension from the provided path.

### Checklist


eszkadev (approved):
code looks good

pedropintosilva (approved):
Tested. It works as described. It is now possible to save as into a document with dots in the name

PR #5939, opened 2023-03-03
jsdialog: chart wizard conversion

eszkadev:
Calc Insert -> Chart


pedropintosilva (commented):
[x] function wizard
[x] char wizard now the missing icons are not missing, the dialog is possible to close, and it is not tunnelled again
| [Before this PR (e82f58e5)](https://github.com/CollaboraOnline/online/commits/e82f58e5) | After this PR |
| --- | --- |
| image | image |

is this expected ^ or am I missing something that in return is giving me the tunnelled window

pedropintosilva (approved):

PR #5943, opened 2023-03-04
For .eslintrc change, see: https://github.com/typescript-eslint/types…

gokaysatir:
…cript-eslint/issues/1824

idle handler is moved into typescript and outside of map variable.

vex dialog of handler is replaced with JDSdialog.

Change-Id: I341c28e9e3f1659987b871a235a50757f42d4d2e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -56,6 +56,12 @@ L.Control.JSDialog = L.Control.extend({
 		}
 	},
 
+	closeAll: function() {
+		var dialogs = Object.keys(this.dialogs);
+		for (var i = 0; i < dialogs.length; i++)
+			this.close(dialogs[i]);
+	},
+
 	closeDialog: function(id, sendCloseEvent) {
 		if (!id || !this.dialogs[id]) {
 			console.warn('missing dialog data');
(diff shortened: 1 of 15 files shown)


eszkadev:
how to trigger that idle dialog? I changed pre view idle/out of focus values in coolwsd.xml to 30s but nothing happens, when I have or not have focus on window and doing nothing...

eszkadev (approved):

browser/src/control/Control.JSDialog.js
```diff
@@ -56,6 +56,12 @@ L.Control.JSDialog = L.Control.extend({
 		}
 	},
 
+	closeAll: function() {
+		var dialogs = Object.keys(this.dialogs);
+		for (var i = 0; i < dialogs.length; i++)
+			this.close(dialogs[i]);
```
eszkadev, line 62: if there will be any pure JS dialog present - we will got 'syntax/parsing error' from the server because we will send message with incorrect window ID which is not expected by the server

eszkadev: But I think you can ignore that. Because our pure JS dialogs have custom callback which doesn't send anything to server.

PR #5947, opened 2023-03-05
wsd: simplify passing mobileAppDocId to getNewChild

Ashod:
It seems more complex to have #ifdef guarded logic
around a function call to pass or not pass an argument
when we that same argument is already defaulted in
that function.

If the difference between the two paths is primarily
the argument, then we can always pass it, making sure
that in the case that defaulting would have been used
that we pass the same default value. This is now done
with assertions.

Change-Id: I333eb3f42cb51dfab8584ae8c5aa52ee45c5a9a1

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -2510,9 +2510,13 @@ std::size_t DocumentBroker::removeSession(const std::shared_ptr<ClientSession>& 
 
         // So just go down the same code path as for normal Online:
 
-        // If last editable, save and don't remove until after uploading to storage.
-        if (!lastEditableSession || !autoSave(isPossiblyModified(), dontSaveIfUnmodified))
+        // If last editable, save (if not saving already) and
+        // don't remove until after uploading to storage.
+        if (!lastEditableSession ||
+            (!_saveManager.isSaving() && !autoSave(isPossiblyModified(), dontSaveIfUnmodified)))
+        {
             disconnectSessionInternal(session);
+        }
 
         // Last view going away; can destroy?
         if (activeSessionCount <= 1)


tml1024:
In fact, the whole reason why that mobileAppDocId thing was introduced in the first place might well have evaporated in the meantime. I have no idea whether any end-user of the app uses the feature that was the reason for it: Being able to have multiple documents open in the single process of the iOS app, in separate "windows". See tdf#128502 . Commit 7f25109f72738706359a63c9062764699f00f568 in online introduced it. It was hard to implement that as nothing in Online or core was prepared to actually have several documents open in a LibreOfficeKit-using process at the same time. Even having a plain LibreOfficeKit client on Linux that would have several documents at the same time turned out to be very fundamentally broken. See for instance a22cba07a470385dcb930c56c90cba81b317773e in core.

If we in fact aren't interested in keeping this "multi-tasking" feature in the iOS app, and if Apple's guidelines don't require that a document-centric app supports it, I would recommend we simply revert the unnecessary complexity from online and core. Then we should *document* that LibreOfficeKit does not support having multiple documents open simultaneously.

Adding @plubius .

tml1024:
Oh, and the feature in question might well have bit-rotted badly since then, of course.

tml1024:
An article about the feature in general: https://appletoolbox.com/how-to-open-two-instances-of-the-same-app-on-your-ipad-with-ipados/

tml1024:
But as such, this simplification is fine of course, even if I would prefer to eventually remove the mobileAppDocId thing completely, assuming we have no interest in the technical debt of having that presumably rarely used "multi-tasking" feature. Will approve once CI passes without compilation errors.

Ashod:
Thanks @tml1024 for the thoughts. If there is little or no value in maintaining the mobileAppDocId and the multi-doc feature, we can remove it. I wasn't looking at doing that, but am interested in simplifying things, as it's costly to maintain these multiple compile-time versions and code-paths in the common functions. Hence this PR.

plubius:
Let me know when this gets committed to master and I can do a quick build of Online master on the iPad.

tml1024 (approved):

plubius:
I tested out the patch with master Online and master LibreOffice builds and I don't seen any obvious change in behavior with the patch.

I did some light testing in ["split screen" mode](https://user-images.githubusercontent.com/104370511/223732314-619235f0-1546-46e1-a6a3-96806185e795.png) and open the same document in both for editing.

Note for the future about running multiple instances: there is some difficulty in getting keyboard and tap focus when two instances are running, but the same focus behavior is there without the patch so that shouldn't prevent merging this pull request.

PR #5961, opened 2023-03-07
Remove vex dialog from command.js.

gokaysatir:
Change-Id: Ib5df6c75c5f837cb7ccc0060f6b762d2a9fea1bc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Command.js
+++ b/browser/src/control/Control.Command.js
@@ -3,7 +3,7 @@
  * Feature blocking handler
  */
 
-/* global $ vex _ isAnyVexDialogActive*/
+/* global $ _ isAnyVexDialogActive*/
 
 L.Map.include({
 
@@ -60,12 +60,6 @@ L.Map.include({
 	openUnlockPopup: function(cmd) {
 		if ((this.isRestrictedUser() && this.isRestrictedItem(cmd)) || isAnyVexDialogActive())
 			return;
-		var lockingOnMobile = '';
-
-		if (window.mode.isMobile()) {
-			lockingOnMobile = 'mobile';
-		}
-		var that = this;
 		var message = [
 			'<div class="container">',
 			'<div id="unlock-image" class="item illustration"></div>',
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.Command.js
```diff
@@ -81,19 +75,23 @@ L.Map.include({
...
+				this.jsdialog.closeAll();
+			else
+				this.fire('closemobilewizard');
+		}.bind(this), true);
+		document.getElementById('unlock-features-popup').querySelectorAll('p')[0].outerHTML = message;
+
+		document.getElementById('unlock-features-popup-cancel').onclick = function() {
+			if (this.jsdialog)
+				this.jsdialog.closeAll();
+			else
```
eszkadev, line 92: instead of this conditional code you can simply use `closeModal` method: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L822

It sends jsdialog message so on mobile it is handled by mobile-wizard, on desktop by jsdialog control.

browser/src/control/Control.Command.js
```diff
@@ -81,19 +75,13 @@ L.Map.include({
...
-					window.open(that.Locking.unlockLink, '_blank');
-			},
-			buttons: [
-				$.extend({}, vex.dialog.buttons.NO, { text: _('Cancel') }),
-				$.extend({}, vex.dialog.buttons.YES, { text: _('Unlock') })
-			]
-		});
+		message = message.join();
+
+		this.uiManager.showInfoModal('unlock-features-popup', this.Locking.unlockTitle, ' ', ' ', _('Unblock'), function() {
```
eszkadev, line 80: typo: "Unblock" -> "Unlock"

eszkadev (changes requested):

browser/src/control/Control.Command.js
gokaysatir: Aaargh :D Thanks for the catch :)

browser/src/control/Control.Command.js
```diff
@@ -81,19 +75,13 @@ L.Map.include({
...
-					window.open(that.Locking.unlockLink, '_blank');
-			},
-			buttons: [
-				$.extend({}, vex.dialog.buttons.NO, { text: _('Cancel') }),
-				$.extend({}, vex.dialog.buttons.YES, { text: _('Unlock') })
-			]
-		});
+		message = message.join();
+
+		this.uiManager.showInfoModal('unlock-features-popup', this.Locking.unlockTitle, ' ', ' ', _('Unlock'), function() {
```
eszkadev, line 80: what about cancel button? maybe we should use showConfirmModal ?

gokaysatir: I guess there wasn't a cancel button in the initial implementation. Maybe we can ask @pedropintosilva ?

eszkadev: there was, see 5 lines above. there is "Cancel" and "Unlock" button

PR #5962, opened 2023-03-07
Use a JSDialog for infobar.

gokaysatir:
Change-Id: Ia253f7dcfac28310accbc71267cfded15c9cc2cd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.Infobar.js
```diff
@@ -3,40 +3,20 @@
...
-		}
-
-		vex.dialog.open({
-			message: e.msg,
-			className: 'vex-theme-bottom-right-corner',
-			showCloseButton: true,
-			buttons: buttons,
-			callback: callback
-		});
+		this.map.uiManager.showInfoBar('info-bar', e.msg, e.action, e.actionLabel);
```
eszkadev, line 17: what about removing whole Infobar file and replacing with that one line? what do you think?

gokaysatir: I also thought that we could indeed :) Let me remove that, thanks :)

gokaysatir:
infobar was used for both version check and info purposes. I renamed the version check to versionBar. There is no conflict as of now.

eszkadev (approved):

Issue #5843, opened 2023-02-13
iOS App: some labels are broken in the sidebar


nikslor:
**Describe the bug**
Some buttons in the sidebar have broken labels in the current TestFlight version.

**To Reproduce**
1. Open a writer document in the TestFlight iOS app
2. Open the sidebar

**Expected behavior**
The labels of each button should be readable.

**Actual behavior**
Some labels are broken.

**Screenshots**
![image](https://user-images.githubusercontent.com/1211300/218404082-62605795-e557-47e4-a02a-a8ecd11bce82.png)

**Tablet**
 - OS: iPadOS 15.6.1
 - Version: 22.05.202302092239

plubius:
Update: I did a little debugging this afternoon and was able to trace the incoming misencoded strings in the sidebar to the POLLIN handler in [DocumentViewController userContentController:didReceiveScriptMessage:] which packages the incoming data and pushes it to JavaScript.

So, the next place I'll be looking at is in LOKit. The iOS app assumes the incoming data from LOKit are in UTF-8 format so it seems odd for the reencoding of strings to happen in LOKit but if I find recoding code, maybe I can link it back to some JavaScript global setting and then I can look for something we can control in WKWebView.

plubius:
After many tedious hours of debugging, I have found the cause of the mangled non-ASCII characters: [CODocument send2JS:length:] receives valid UTF-8. For example, é (\u00e9) is encoded as two bytes: 0xc3 0xa9).

Then, the code replaces each byte that is >= 0x80 with literal '\x##' characters so the 2 bytes for an 'é' in the UTF-8 string are replaced with '\xc3\xa9' characters (8 characters).

@tml1024 added the >= 0x80 in commit 2b84a35fe803b14e15e0df954ebe1f18387078db but removing that conditional causes rendering in JavaScript to fail. So, obviously the characters need to be escaped for the JavaScript code to be able to parse it before sending it to [WebKit evaluateJavaScript:completionHandler:].

In this super [hacky debug patch](https://github.com/CollaboraOnline/online/files/10892287/iosappoutput.txt), and force the 0xc3 0xa9 combination of byes to be replaced with '\xe9' characters (4 characters), an 'é' is properly rendered by the JavaScript.

[...]

PR #5965, opened 2023-03-07
Another fix for issue #5843

plubius:
Limit escaping of non-ASCII text and splitting on the first newline to only certain message types on iOS.


mmeeks (approved):
Looks great to me - thanks Patrick ! =)

PR #5971, opened 2023-03-09
Revert "wsd: break cyclic setModified call"

Ashod:
The setModified is on an internal, AdminModel
specific, Document object. It is not the
DocumentBroker.

This reverts commit e33d6aa2ca3231befd6634ddf2316712d681ded1.

Change-Id: Ib706927051678cfc19dc1246d679ef530eb8d586


Rash419 (approved):
Thanks : )

PR #5972, opened 2023-03-09
Dark mode integration and view separation

OFFTKP:
Needs patch https://gerrit.libreoffice.org/c/core/+/148615 to work

Should separate views based on formatting marks between multiple browsers

Tests are currently failing because the patch above hasn't been merged


OFFTKP:
Set DCO to passing, it was failing pointing that commit "Fix UI selection bug" was not signed off correctly even though it was signed. If this is a mistake let me know.

eszkadev (commented):
JS part looks good

mmeeks:
Lets get this in then =)

thebearon:
This needs the 22.05 backport of the following core gerrit patch to build, no?
https://gerrit.libreoffice.org/c/core/+/148776

mmeeks:
We don't want dark mode in 22.05 - too risky.

thebearon:
Oh, ok so online master doesn't build on core 22.05 anymore, got it.

PR #5976, opened 2023-03-09
jsdialog: hide pushbuttons which shouldn't be visible

lpranam:
Change-Id: I3e9f70b57360c4d44e5a359904c7b865b3a044f5

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1529,6 +1529,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === 'false' || data.enabled === false)
 			$(pushbutton).prop('disabled', true);
 
+		if (data.visible === 'false' || data.visible === false)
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1529,6 +1529,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === 'false' || data.enabled === false)
 			$(pushbutton).prop('disabled', true);
 
+		if (data.visible === 'false' || data.visible === false)
+			L.DomUtil.addClass(pushbutton, 'hidden');
```
eszkadev, line 1533: this also will work, but we do that in `postProcess` for all controls for code sharing for common properties.
I think root cause is that we don't execute `postProcess` after running pushbutton handler inside `_buttonBox`.

eszkadev: I added fix https://github.com/CollaboraOnline/online/pull/5984

PR #5979, opened 2023-03-09
jsdialog: use display grid for vertical button box

lpranam:
Change-Id: I288b1c1b013dd0a1395f6f18890b09582f309fbc

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -668,6 +668,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				builder._controlHandlers[child.type](right, child, builder);
 		}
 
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -668,6 +668,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				builder._controlHandlers[child.type](right, child, builder);
 		}
 
+		if (data.vertical === 'true' || data.vertical === true) {
```
eszkadev, line 671: we are done with the conversion of types, so currently we should get always boolean in JSON, no need to handle also string :)

eszkadev (approved):

PR #5981, opened 2023-03-09
jsdialog: hide fixed text which shouldn't be visible

lpranam:
Change-Id: I2af0b60f4545b2349dfbfd98fd24e4bd02f4fe95

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1840,7 +1840,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			L.DomUtil.addClass(fixedtext, 'ui-text');
 		}
 
-		if (data.hidden)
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1840,7 +1840,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			L.DomUtil.addClass(fixedtext, 'ui-text');
 		}
 
-		if (data.hidden)
+		if (data.hidden || data.visible === 'false' || data.visible === false)
```
eszkadev, line 1843: this should already happen in `postProcess`, is that label inside buttonbox maybe?

eszkadev: I added fix: https://github.com/CollaboraOnline/online/pull/5984

Issue #5887, opened 2023-02-21
iOS App: Saving error "The server encountered a savefailed error while parsing save command"


Ezinnem:
**Describe the bug**
Saving a file in the iOS App build gives the error `The server encountered a savefailed error while parsing save command`

1. Create a new document
2. Type some text and Press the Save button

See the Parsing error dialog

`The server encountered a savefailed error while parsing save command`

See recording:
**To Reproduce**
1. Create a new document
2. Type some text and Press the Save button

**Expected behavior**
The File saves without any error dialog

**Actual behavior**

There is a dialog with th text: `The server encountered a savefailed error while parsing save command`

**Screenshots**

**IOS(please complete the following information)**
 - iOS 22.05.10 build

**Additional context**
See recording:

https://user-images.githubusercontent.com/54344301/220436534-6ae242f7-3e1c-442a-a280-abf455cd33a2.MP4

Ezinnem:
@plubius

plubius:
I can reproduce this bug. The error dialog looks like it is a JavaScript alert so my current theory is that an error is bubbling up from either the iOS app native code or the LibreOffice C++ code. I will need to do some debugging to determine what is happening in the code.

thebearon:
It's strange the app is giving this message instead of the actual error message this should result in: `Document cannot be saved.`
Unrelated to the actual save issue, just strange.

plubius:
I found the code that sends a "savefailed" message. Looks like the native code can't save the file. Maybe this is a false negative since closing the document successfully saves any changes:

WRN  ToClient-001: Session [001] on document [/private/var/mobile/Containers/Data/Application/9C068D3A-B274-4142-8720-89F1A8B26A24/tmp/1/Blank.odt] has no write permissions in Storage and cannot save.| ClientSession.cpp:655

sebastiaanveld:
Just updated #3760, this seems to be the same issue.

plubius:
The fix for this bug is now in the latest TestFlight build. Please close if the bug is fixed for you.

sebastiaanveld:
Tested and works for me in Collabora Office 22.05.10 (22.5.202303132008) in Testflight

plubius:
My fix is in the latest test patch. Can you close the bug if the fix works for you?

nikslor:
Works for me in 22.05.11 (22.5.202303212054); but I'll leave it to @Ezinnem to close as she reported it.

Issue #3760, opened 2021-11-30
ios saving unreliable


kr4z33:
Trying out Collabora Office via testflight on ios, ran into some kind of issue with saving an ods file on my nextcloud

Version is reported in help/about as  21.06.1.0 and 21.06.4.1

Using the spreadsheet editor on a new ods file: saving to Nextcloud was unreliable, with the following behavior:

I was able to get an early revision saved, good times. After working a while, I noticed the disk icon in the upper left would get a green dot overlaid, consistent with indicating there was unsaved work. After tapping the disk icon, the dot would go away, consistent with an indication that work done to the file was saved. Sounds good. After a few hours working this way. I put the device away for other tasks.

The following day I went to reopen the file to make further edits, and found that only a very early version of the file existed in the Nextcloud storage. apparently tapping the disk icon in Collabora was not actually saving the work, as visual indications in Collabora would have suggested.

I would prefer Collabora to be very noisy about unsuccessful saving and unsaved work and to keep a local cache of recent work to avoid this kind of loss.

sebastiaanveld:
Was testing this also and found is works reliably for me on an iPhone but not using an iPad. On iPad i get an save error when saving the modified file. After the error message appears, the Save button in the Collabora Office App still shows there are changes the need to be saved (blue circle still appears). 

Steps to duplicate:
- Start the Nextcloud App on the iPad
- On an .odt document select from the menu Open In > Collabora Office
- File opens in the local Collabora Office App
- Click the Edit button, make a modification to the document

![image](https://user-images.githubusercontent.com/37268999/224551488-053aeadb-c1c8-4e6a-8624-e6cbf3b4f63c.png)

- From the top menu bar, select the Save button
- Error appears Server ran into a savefailed-error while performing saving (Dutch localized in the screenshot)

![image](https://user-images.githubusercontent.com/37268999/224551499-2646eeeb-c8dd-4933-aa00-d03631e4ca2a.png)

-After a few second the error message goes away automatically, if not clicking the OK button. Mind the Save button still the shows modifications are made that need to be saved (blue circle still shows). The file in Nextcloud does not shows it's updated, but could be this is down after the file in the Collabora App is closed?
- Close the file using the close X-button
- Go back to the Nextcloud App, file shows is modified
- Reopen the file using the Collabora Online/Nextcloud Office just by cicking the file, this makes sure the "updated" file is used that in the App's silo.
- The modification shows up so the file was, despite the shown error, saved correctly.

[...]

Issue #5841, opened 2023-02-13
iOS App: "Help > SendFeedback" not working


nikslor:
**Describe the bug**
The button "Send > Feedback" is not working in the current TestFlight Version (22.05.202302092239)

If sending feedback is not possible from the iOS app, the button should not be visible (IMHO).

**To Reproduce**
1. Open a writer document in the TestFlight iOS app
2. Change to the "Help" tab
3. Tap on "Send Feedback"

**Expected behavior**
A dialog should show up (I guess)

**Actual behavior**
Nothing happens

**Tablet**
 - OS: iPadOS 15.6.1
 - Version: 22.05.202302092239

plubius:
I have seen this too in several toolbar icons. Tapping toolbars like these leaves it drawn like it is still pressed so my current theory is that an unexpected exception is being thrown by the underlying LibreOffice code.

plubius:
I think I know why this toolbar button is failing: the JS code is trying to load the URL in window.feedbackUrl defined in cool.html:

window.feedbackUrl = '%FEEDBACK_URL%';

What the Online server code does is replace sever %...% strings when it serves cool.html and other files to the web client. But this code is never executed in the iOS app because the "server" part of the app directly loads local files directly into the WKWebView.

It makes sense to me that Online dynamically replaces these strings with each request because the values can be update on the server and pushed to the web client. But in the iOS app, the "server" part is bundled locally in the app so I would like to avoid duplicating the dynamic replacement of these strings.

So, my current solution is to edit the makefile to do the string replacement during the build (iOS builds only) instead of at runtime.

Below are the 4 strings that need to be replaced on iOS. If anyone knows of others, let me know:

%UI_RTL_SETTINGS%
%WELCOME_URL%
%FEEDBACK_URL%
%BUYPRODUCT_URL%

plubius:
I have this bug and now the Help > Send Feedback button is hidden by default. The button was displayed because window.feedbackUrl was set to '%FEEDBACK_URL%' which is an invalid URL. In the Online releases, %FEEDBACK_URL% is replaced dynamically with an empty string or a client's custom ticketing or feedback website.

I found out that there isn't any such custom website for the iOS or Android apps so, with my fix, %FEEDBACK_URL% is now set to an empty string. That, in turn, causes the JS code to hide the Send Feedback button.

[...]

PR #5988, opened 2023-03-10
Fix issues #5887 and #5841 on iOS

plubius:
The iOS app saves directly to local disk so, other than for "view file extension" document types or other cases that I am missing, we can assume the document is writable until a write failure occurs.

Change-Id: I72e80e74f1132d41c11015a80bcc0d79c77acb26


plubius:
If there are other cases on iOS where a document should be read-only, let me know.

AFAICT, the non-Mobile code also checks user privileges and document locking to determine if a document is read-only. I don't think there are any user privilege issues since iOS gives the app dedicated folders to write to and it appears to me that LocalStorage::updateLockState() doesn't do anything.

sebastiaanveld:
Guess this also fixes #3760.

PR #5993, opened 2023-03-10
trim memory when user is inactive.

mmeeks:
Builds on a new LOK API to do the same in the core.

Change-Id: I4443cd2ffbb6c7af0726162aec2ba78fc354d901

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Needs the API  merged to core first really ...

PR #5997, opened 2023-03-11
Fix issues #5887 and #5841 on iOS

plubius:
The iOS app saves directly to local disk so, other than for "view file extension" document types or other cases that I am missing, we can assume the document is writable until a write failure occurs.


plubius:
Note: this is a copy of pull request #5988 for the master branch.

timar (approved):

PR #5998, opened 2023-03-12
browser: Separate comment reply sections from other comments

OFFTKP:
Separate comment reply sections from other comment threads if there's multiple comment threads in the same location in a document

Change-Id: If817b3769eb35d730860ea2817e024efe2b684c6

old PR: https://github.com/CollaboraOnline/online/pull/5853


OFFTKP:
Sorry Pedro, the old PR got deleted because I deleted the forked repo and switched to private branches, did you need any changes before this is merged? Here's the old PR https://github.com/CollaboraOnline/online/pull/5853

PR #6001, opened 2023-03-12
a11y: TextInput: switch from <textarea> to content editable <div>

mcecchetti:
Refactored TextInput implementation. A content editable <div> is used in place of <textarea>.
 elements are used as pre/post spaces.
Minimal changes to FormulaBarJSDialog, since some code was dependendent on
TextInput details.
Refactored some cypress helper funcions, too, since cypress tests expected a <textarea> element.

Change-Id: I2141d7769588bb5bf8a0fbb62f036cb78f914966


browser/src/layer/marker/TextInput.js
```diff
@@ -289,16 +290,32 @@ L.TextInput = L.Layer.extend({
...
-		this._textArea.select();
+		this._setSelectionRange(0, this.getPlainTextContent().length);
 	},
 
 	getValue: function() {
-		var value = this._textArea.value;
+		var value = this.getPlainTextContent();
 		if (this._map.formulabar && this._map.formulabar.hasFocus())
-			value = this._map.formulabar.getValue();
+			value =  ' ' + this._map.formulabar.getValue() + ' ';
```
mmeeks, line 299: Curious about the extra spaces here really =)

browser/src/layer/marker/TextInput.js
```diff
@@ -289,16 +290,32 @@ L.TextInput = L.Layer.extend({
...
+		return 	this._textArea.textContent;
+	},
+
+	getHTML: function() {
+		return 	this._textArea.innerHTML;
+	},
+
+	resetContent: function() {
+		this._textArea.innerHTML = this._initialContent;
+		this._textArea.firstChild.textContent = '\xa0';
```
mmeeks, line 313: Do we really need this ? surely we have the <img - now which doesn't need a textContent ? =)

browser/src/layer/marker/TextInput.js
```diff
@@ -769,11 +789,7 @@ L.TextInput = L.Layer.extend({
 		// MSIE/Edge cannot compare a string to "\n" for whatever reason,
 		// so compare charcode as well
 		if (text === '\n' || (text.length === 1 && text.charCodeAt(0) === 13)) {
-			// The composition messages doesn't play well with just a line break,
```
mmeeks, line 772: I guess we should test this with an input method - like SCIM on Linux - any chance of setting that up ? you can type eg. 'nippon' and get the character(s) for Japan in the Japanse locale

browser/src/layer/marker/TextInput.js
```diff
@@ -874,8 +890,9 @@ L.TextInput = L.Layer.extend({
 			this._deleteHint = 'delete';
 		else {
 			this._deleteHint = '';
-			this._linebreakHint = ev.keyCode === 13 && ev.shiftKey;
 		}
+		this._newlineHint = ev.keyCode === 13;
```
mmeeks, line 894: nice =)

browser/src/layer/marker/TextInput.js
```diff
@@ -990,6 +1008,14 @@ L.TextInput = L.Layer.extend({
 		}
 	},
 
+	_sendNewlineEvent: function() {
+		// The composition messages doesn't play well with just a line break,
+		// therefore send a keystroke.
+		var unoKeyCode = this._linebreakHint ? 5376 : 1280;
```
mmeeks, line 1014: looks good.

browser/src/layer/marker/TextInput.js
```diff
@@ -1014,6 +1041,129 @@ L.TextInput = L.Layer.extend({
 			this._textArea.setAttribute('data-accept-input', accept);
 		}
 		this._acceptInput = accept;
+	},
+
+	_hasPreSpace: function() {
+		var preSpace = document.getElementById('pre-space');
```
mmeeks, line 1047: Hmm - can we avoid searching the whole document, and just search in this._textArea; I'm not a huge fan of this "lets search the huge global variable for a string called 'fish'" or whatever =)

mmeeks (commented):
Lots of great stuff here =) lets do a few tweaks, get CI working & merge it !

browser/src/layer/marker/TextInput.js
mcecchetti: If you look at changes in `Control.FormulaBarJSDialog.js` you can see that extra spaces were already there. I just moved the code in TextInput since refactoring is more difficult when encapsulation is broken. As far as the reason extra spaces are needed is that `_onInput` implementation expects them. More on next comment.

browser/src/layer/marker/TextInput.js
mcecchetti: As I told above `_onInput` expects extra spaces. I'd very like to refactor `_onInput` and get rid of extra spaces, anyway I'd need some enlightenment on part of `_onInput` where we check for content length < 1 and < 2: by reomoving extra spaces it is no more possible to perform these checks:

		// We use a different leading and terminal space character
		// to differentiate backspace from delete, then replace the character.
		if (content.length < 1 || !this._hasPreSpace()) { // missing initial space
			window.app.console.log('Sending backspace');
			if (!ignoreBackspace)
				this._removeTextContent(1, 0);
			this._emptyArea();
			return;
		}
		if (!this._hasPostSpace()) { // missing trailing space.
			window.app.console.log('Sending delete');
			this._removeTextContent(0, 1);
			this._emptyArea();
			return;
		}
		if (content.length < 2) {
			window.app.console.log('Missing terminal nodes: ' + this._deleteHint);
			if (this._deleteHint == 'backspace' ||
				this._getSelectionStart() === 0)
			{
				if (!ignoreBackspace)
					this._removeTextContent(1, 0);
			}
			else if (this._deleteHint == 'delete' ||
				this._getSelectionStart() === 1)
				this._removeTextContent(0, 1);
			else
				window.app.console.log('Cant detect delete or backspace');
			this._emptyArea();
			return;
		}

For the former check the original if clause was:

    if (content.length < 1 || content[0] !== preSpaceChar)

    
Maybe they are no more needed now but on first pass I preferred to be conservative, since I didn't want to miss some use cases.

browser/src/layer/marker/TextInput.js
mcecchetti: Oh well, I have no experience with SCIM or alike input method. I'll add it to my todo list. Anyway the code here is the same as before just moved into `_sendNewLineEvent()`

browser/src/layer/marker/TextInput.js
mcecchetti: Thanks!

browser/src/layer/marker/TextInput.js
mcecchetti: No problem I will rewrite it

browser/src/layer/marker/TextInput.js
mmeeks: Well - lets clean this up to make it maintainable - if we don't believe that is used anymore - lets get rid of it. One of the problems in the past was that auto-correction and/or random Android input methods would not only move your cursor to 'help' but also change the content of the text area. I think that's less likely with our content-editable approach and HTML - it'd be too risky / complex for them =)

So lets drop this content.length < 1 complexity - I'm sure we'll get new complexity back for something else =)

browser/src/layer/marker/TextInput.js
mmeeks: Quite possibly we could use a local member variable this.preSpace - which holds a pointer to the element and managing that nicely (?)

browser/src/layer/marker/TextInput.js
mcecchetti: Extra spaces removed in new commit version

browser/src/layer/marker/TextInput.js
mcecchetti: TextInput simplified; extra spaces and text nodes inside  delimiters removed.

browser/src/layer/marker/TextInput.js
mcecchetti: SCIM tested: it works as with <textarea>

browser/src/layer/marker/TextInput.js
mcecchetti: New rewrite just check textArea first/last child

mcecchetti:
Locally I haven't these cypress test failures

mcecchetti:
As discussed on the dataport status call I'm going to merge without waiting for all cypress tests to pass (since they are passing locally)

PR #6003, opened 2023-03-13
wsd: do not disconnect the last session if saving

Ashod:
Previously we only waited if a new auto-save has
been issued at the point of disconnecting the
last session. However, we could have a save in
progress, particularly when it's a slow one,
and we should equally wait for it before
disconnecting the last session.

Failing to do so will exit the kit right after
finishing the save and will rob us of the
ability to upload the just-saved version.

Change-Id: I678689a95211fa8b80bcd3e8d4537de7de5d6632

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -2510,9 +2510,13 @@ std::size_t DocumentBroker::removeSession(const std::shared_ptr<ClientSession>& 
 
         // So just go down the same code path as for normal Online:
 
-        // If last editable, save and don't remove until after uploading to storage.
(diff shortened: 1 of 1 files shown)


mmeeks:
Apparently unrelated Android problem - I assume that the disconnectSessionInternal will in fact happen when the save completes and we won't leak these - but lets merge it.

Ashod:
> I assume that the disconnectSessionInternal will in fact happen when the save completes and we won't leak these

@mmeeks, indeed. Technically, we can even remove `disconnectSessionInternal` (with some modification), but the reason it's there is to minimize the unload time when truly we don't have anything more to do (and so we don't go over sessions on each poll to look for disconnected ones to clean them up). But the cleanup will happen when we do upload, and we correctly unload the doc.

PR #6004, opened 2023-03-13
We aparently want thumbnails to be of size 1200x630 pixels

tml1024:
Tweak the "zoom" variabe to be 1 so that with this thumbnail size, for a typical document we get more of it into the thumbnail.

Change-Id: I419afccbf57eae23062ab4c849dd41293f068f00

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -302,11 +302,11 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             y = 0;
 
         bool success = false;
-        const int width = 120;
-        const int height = 120;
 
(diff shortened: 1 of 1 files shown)


tml1024:
No idea why that "Fast Build against Core co-22.05" fails with `class lok::Office {aka class lok::Office}’ has no member named ‘extractRequest’` ; the extractRequest thing was added to co-22.5 (and co-23.05, that the check actually seems to be using?) already last week.

tml1024:
It's hard to take our CI seriously.

PR #6005, opened 2023-03-13
Remove vex from documentconflict message.

gokaysatir:
Change-Id: I7b57fd57ce4809e4d829e6e4224f4486c7c88259

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -917,6 +917,55 @@ L.Control.UIManager = L.Control.extend({
 		], cancelButtonId);
 	},
 
+	/// buttonObjectList: [{id: button's id, text: button's text, ..other properties if needed}, ...]
+	/// callbackList: [{id: button's id, func_: function}, ...]
+	showModalWithCustomButtons: function(id, title, message, cancellable, buttonObjectList, callbackList) {
+		var dialogId = this.generateModalId(id);
+
+		for (var i = 0; i < buttonObjectList.length; i++)
+			buttonObjectList[i].type = 'pushbutton';
+
+		var json = this._modalDialogJSON(id, title, !!cancellable, [
+			{
+				id: 'info-modal-tile-m',
+				type: 'fixedtext',
+				text: title,
+				hidden: !window.mode.isMobile()
+			},
+			{
+				id: 'info-modal-label1',
+				type: 'fixedtext',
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -917,6 +917,55 @@ L.Control.UIManager = L.Control.extend({
 		], cancelButtonId);
 	},
 
+	/// buttonObjectList: [{id: button's id, text: button's text, ..other properties if needed}, ...]
+	/// callbackList: [{id: button's id, func_: function}, ...]
+	showModalWithCustomButtons: function(id, title, message, cancellable, buttonObjectList, callbackList) {
```
eszkadev, line 922: maybe "yes/no" modal can now use this function internally

gokaysatir: I guess no because yesno modal only has 2 buttons. This one can have more than 2 buttons.

browser/src/core/Socket.js
```diff
@@ -1248,6 +1170,41 @@ app.definitions.Socket = L.Class.extend({
...
+				var filename = this._map['wopi'].BaseFileName;
+				if (filename) {
+					filename = L.LOUtil.generateNewFileName(filename, '_new');
+					this._map.saveAs(filename);
+				}
+			}.bind(this)});
+		}
+
+		var title = vex._escapeHtml(_('Document has been changed'));
+		var message = vex._escapeHtml(_('Document has been changed in storage. What would you like to do with your unsaved changes?'));
```
eszkadev, line 1203: _escapeHtml is not needed now, we don't build html  from strings

eszkadev (approved):

browser/src/core/Socket.js
gokaysatir: Thanks for the comment :) Let me remove that forgotten part while i solve the conflict.

browser/src/control/Control.UIManager.js
eszkadev: hmm I don't see any limitation here, it just uses array of button objects

gokaysatir: Right, i mis-read your comment at first :) we can shorten that part indeed.

PR #6016, opened 2023-03-14
Remove vex from comments.

gokaysatir:
Change-Id: Ie548bc5576bc7a7afb426f9cf9ef4a7c1cafbb88

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -30,7 +30,6 @@ L.Map.include({
 
 declare var L: any;
 declare var app: any;
-declare var vex: any;
 declare var $: any;
 declare var _: any;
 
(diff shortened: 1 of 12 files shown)


eszkadev:
comment
width is now too low

gokaysatir:
> comment width is now too low

Thanks for the check :) Done.

browser/src/layer/tile/CommentListSection.ts
```diff
@@ -375,39 +374,44 @@ export class CommentSection extends CanvasSectionObject {
...
+		document.getElementById('input-modal-input').outerHTML = '<textarea name="comment" id="input-modal-input" class="cool-annotation-textarea" required>' + (commentData.text && isMod ? commentData.text: '') + '</textarea>';
+
+		// Allow close on click away only in desktop.
+		if (document.getElementsByClassName('mobile-wizard jsdialog-overlay cancellable').length > 0) {
+			document.getElementsByClassName('mobile-wizard jsdialog-overlay cancellable')[0].addEventListener('click', function() {
+				this.cancel(comment);
+			}.bind(this));
+		}
+		else {
+			document.getElementById('modal-dialog-new-annotation-dialog-overlay').addEventListener('click', function() {
```
eszkadev, line 407: It looks a bit like a hack when we create dialog and them manually override callbacks and replacing nodes.
What about using `_modalDialogJSON` function already with textarea within it. Also it is possible to setup callbacks when calling `showModal` function. (Something like internals of `showInfoModal`, adjusted to needs). For me it will look much more clean and would take similar number of lines of code.

eszkadev (approved):

browser/src/layer/tile/CommentListSection.ts
eszkadev: I implemented that

Issue #5839, opened 2023-02-13
iOS App: "Copy" button in "Home" tab doesn't work


nikslor:
**Describe the bug**
"Copy" button in the "Home" tab of the current TestFligh verson (22.05.202302092239) is not working.

**To Reproduce**
1. Open a writer document in the TestFlight iOS app
2. Write some text and mark it
3. Tap the "Copy" button in the "Home" tab

**Expected behavior**
The "Paste" button should change its state and become "active" so the user can paste the copied content.

**Actual behavior**
The "Paste" button doesn't change its state and also the keyboard shortcut "cmd + v" will not work - so presumably the "Copy" function didn't work.

**Screenshots**
![image](https://user-images.githubusercontent.com/1211300/218397634-ee1ae1b7-7a9c-4bc6-9441-0e18f63e88c4.png)

**Tablet**
 - OS: iPadOS 15.6.1
 - Version: 22.05.202302092239

Ezinnem:
I tested this and I could not reproduce on my end.

@nikslor , please can you test again?

nikslor:
@Ezinnem it's indeed better now; but still not perfectly good. I sometimes still encounter the situation that the "Paste" button in the menu bar is no clickable (after using the "Copy" button) - exactly how you describe it in #5908. But I could not yet find the pattern that triggers this. FYI: I also encountered it in "writer". I'll test this a bit more and think we should not yet close the issue.

plubius:
Although this is related to issue #5908, this bug is definitely its own separate bug.

From my testing, it appears that the enabled/disabled state of the Paste button is set only once at when the document is opened. So if the system clipboard is empty, the Paste button is disabled.

I will investigate if the something on either the iOS side is failing to notify the JS code that the state of the system clipboard has changed.

mmeeks:
Hmm - so - I'm rather surprised that paste is ever inactive. Ultimately we can't easily tell if there is any content on the system clipboard from the browser - so I'd expect Paste to be constantly enabled ~everywhere.

plubius:
> Hmm - so - I'm rather surprised that paste is ever inactive. Ultimately we can't easily tell if there is any content on the system clipboard from the browser - so I'd expect Paste to be constantly enabled ~everywhere.

I haven't looked that closely at the JS code but it does seem that the JS code is polling the clipboard state when a document is opened.

[...]

PR #6029, opened 2023-03-15
Fix GitHub issue #5839 and never disable Paste toolbar button

plubius:
Behave the same as Contol.Menubar and never disable Paste toolbar button. Native clients that run LibreOffice locally may send a "statechanged: .uno:Paste=disabled" message when opening a document if the system clipboard is empty, So, we ignore such messages or else the current document's Paste toolbar button will never be enabled.


plubius:
Note: this is a near duplicate of pull request #6028 for the master branch.

timar (approved):

PR #6033, opened 2023-03-16
zotero: show buttons when integrator supports it

eszkadev:
Empty api key string indicates that integrator supports Zotero. When any functionality is accessed by the user we show notification to integration so it will show a popup for user with instructions.


pedropintosilva (approved):
beautiful

pedropintosilva:
PR for master branch : #6033

PR #6035, opened 2023-03-16
export as pdf: we need to decode path

eszkadev:
When using Nextcloud with source file with spaces in the name eg. "file name.docx" and we tried to export as PDF then error appeared.

In the logs we had "SaveAs produced no output in '<...>file%20name.docx', producing blank url.

We need to decode file name similar to convert-to case.


Ashod (approved):
Looks good to me. Thanks @eszkadev.

PR #6042, opened 2023-03-17
TEST BUILD

eszkadev:
at place where builds were passing... to test if any merged change caused fails or it's about CI itself


timar:
Thanks for this PR @eszkadev 
I'll try to fix 'Unit & Cypress & Notebookbar Tests against Core co-23.05'.

PR #6044, opened 2023-03-17
mobile: don't steal focus from save-as dialog

pedropintosilva:
When integrator shows dialog eg. for save as with the
input fields, keyboard appears resizing our frame.
We shouldn't then steal focus from input.

Change-Id: Ifda6e7049b1b329fb051966e38fa854c93c05065


pedropintosilva:
backporting #[544423656](https://github.com/CollaboraOnline/online/pull/6041/commits/54442365635f87299adb37a0a6cc57867dc0baee) from PR #6044 from @eszkadev

PR #6063, opened 2023-03-21
fix: date picker always in norwegian

Rash419:
Change-Id: Id03cccad6964e7555de6300b0eac619bf521802b

* Resolves: # 
* Target version: co-22-05


timar (approved):
I approve this on the basis that it looks good. We need to verify in production environment. And it's interesting that it worked without this fix in dev environment.

PR #6064, opened 2023-03-21
Zotero: execute commmands with opened dialog

eszkadev:


timar (approved):
Pedro approved the co-22-05 version, it's good surely.

PR #6072, opened 2023-03-22
Hide download in tabbed view when HideExportOption is true

eszkadev:


thebearon (changes requested):
This change seems to hide Repair in Calc when `HideExportOption` is given.

eszkadev:
> This change seems to hide Repair in Calc when `HideExportOption` is given.

good point, fixed

thebearon (approved):

PR #6073, opened 2023-03-22
jsdialog: Improve cosmetics of Find and replace

pedropintosilva:
- Make sure the grid columns of 'Find' and 'Replace'
share the same minimum
  - the maximum parameter in minmax function (so min-content) is
    ignored when that value is lower than the min value but it is
    there for when the language has quite long 'Find' or 'Replace'
    making that max width increase according to its contents
- Remove extra margin from checkbox so it gets aligned

Change-Id: Ica38b407f2a5936e3657dc067ac0658ef5b8aeed

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -142,6 +142,14 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	display: grid;
 }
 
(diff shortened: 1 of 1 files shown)


eszkadev (approved):

pedropintosilva:
this does not change ids, nor classes, cypress should pass

PR #6081, opened 2023-03-22
jsdialog: fix switching the tabs

eszkadev:
This fixes regression from:
commit e5d86f0b3c4114fa8374fdc83373f0cc9a6f1a20
We can use foreach when there is indexed assinments.

there is no index field but id

it was not sending commang to the backend to switch the tab because of that


gokaysatir (approved):
Thank you for the fix :)

PR #6082, opened 2023-03-22
jsdialog: fix dialog jumping on a screen

eszkadev:
This is fix for regression from:
commit a2d666d53adcb23e60ed61328e8e3fe44fdc1af5
JSDialog overlay: Use CSS rules for positioning.

When switching a tab in Format -> Characters dialog, update message was received and dialog got request to update position. Even when we had nothing to update we did modification to dialog position causing jump.

This patch makes updates less agressive and does update only when calculated value is really different.
[jump-2023-03-22_14.01.13.webm](https://user-images.githubusercontent.com/5307369/226917756-f1b03286-d43b-4daa-b187-a3fcf2be08bb.webm)

NOTE: for proper testing it requires working tab control, there is other regression related to that: https://github.com/CollaboraOnline/online/pull/6081


gokaysatir (approved):
Thank you for the improvements :)

PR #6088, opened 2023-03-23
mobile: use attribute 'visibility' to hide toolbar back button

hcvcastro:
Otherwise, the toolbar re-layout half the size.

Change-Id: I492a98f822a829d814b4c6e6b4765ec14a65cc98


pedropintosilva (approved):
Awesome, thanks! No only this is works but it also improves cypress tests outcomes. :+1:

PR #6091, opened 2023-03-23
Fix main tabs text styles on tabbed view

pedropintosilva:
Tabs were changed to buttons and thus we need to readjust the CSS
so text color and others are correctly applied to the text

Also add :not exception so the tabs are not styles as regular buttons

Change-Id: I1bc1681e86f17ee7afb435f7c3fc099b80abac2e


gokaysatir (approved):
Thank you for the improvements :)

PR #6112, opened 2023-03-29
jsdialog: use labelFor property for labels

eszkadev:
requires: https://gerrit.libreoffice.org/c/core/+/149717


gokaysatir (approved):
Thanks @eszkadev :)

PR #6114, opened 2023-03-30
Add remotefontconfigurl as env for Docker entrypoint

sistason:
* Target version: master 

### Summary

This commit adds the environment variable `remotefontconfigurl` to docker images and templates it's value into the coolwsd.xml in the same way as the other environment variables currently handled this way like `remoteconfigurl`.

This allows the `remote_font_config` URL to be set in the config via ENV.

### Checklist

- [x] I have successfully build the container
- [x] I have successfully build the container and tested that with a remotefontconfigurl as ENV I can successfully use the remote_font_config feature
- [x] Documentation (manuals or wiki) is not required, as the other env options (`remoteconfigurl`) is also undocumented afaik...

### DIFF
--- a/docker/from-packages/scripts/start-collabora-online.pl
+++ b/docker/from-packages/scripts/start-collabora-online.pl
@@ -97,6 +97,7 @@ sub rewrite_config($) {
     open(CONFIG, '<', $config) or die $!;
 
     my $in_aliases = 0;
+    my $in_remote_font_config = 0;
     while (<CONFIG>) {
         if (/<remote_url (.*)>.*<\/remote_url>/) {
             my $remoteurl = $ENV{'remoteconfigurl'};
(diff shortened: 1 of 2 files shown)


pedropintosilva:
I have discussed this with @timar . He is of the opinion that we kept the existing env variables for backward compatibility, in fact we can pass any config option as a command line switch to coolwsd. So this is already possible, the same way we do for ssl : ```-e extra_params=--o:ssl.enable=false``` so in this case ```-e extra_params=--o:remoteconfigurl=https://yourdomain/config/config.json```

sistason:
oh, I did not know this option for actually setting any config option, but yeah it makes sense, I already set a ton of them that way :)

Anyways, then it was a exercise in collaborating with github projects for me. Thanks for the fast response and it's a solution for me either way :)

PR #6119, opened 2023-03-31
jsdialog: make tabs accessible using keybaord

eszkadev:
This changes tabs into button elements so they can be switched using keyboard Enter key. It's also better from a11y point of view

Styles has to be !important to override button styling


gokaysatir (approved):
Thanks for the improvements:)

Issue #6103, opened 2023-03-27
iOS App: Read only mode -> File > Export as > PDF Document (PDF) leads to "crash"


nikslor:
**Describe the bug**
In the read only mode (e.g. after opening a document) the function under File > Export as > PDF Document (PDF) leads to "crash". Well the app doesn't crash, but it is stuck and a user can't exit from there.

**To Reproduce**
1. Open a writer document, don't change to edit mode
2. Tap on File > Export as > PDF Document (PDF)

**Expected behavior**
The file dialog should show up, where a user can select the target folder.

**Actual behavior**
The app is stuck in an operation (probably converting or saving).

**Tablet**
 - OS: iPadOS 16.3.1
 - Version: TestFlight 22.05.11 (22.5.202303212054)

Ezinnem:
I can confirm this experience.

Ezinnem:
I can confirm this experience.

plubius:
I can reproduce this as well. It appears that the iOS C++ code or the LibreOffice code is sending the following message so I will see if I can find which code generates this message:

2023-03-30 18:59:58.164631 -0400 [ app2js ] TRC  To JS: blockui: | CODocument.mm:99

plubius:
Note to myself: the LibreOffice PDF Export Options dialog is appearing in the current TestFlight build but not in the 21.x release so I need to see if suppressing display of that dialog changes anything.

plubius:
The fix for this bug is in the current TestFlight. Does the fix work for you?

nikslor:
Tested again in 22.05.12 (22.5.202304051456); works as expected now :)
Thanks a lot @plubius I think you can close this issue as fixed.

Issue #6104, opened 2023-03-27
iOS App: Read only mode -> File > Export as > EPUB (.epub) doesn't work


nikslor:
**Describe the bug**
In the read only mode (e.g. after opening a document) the function under File > Export as > EPUB (.epub) doesn't work. Nothing seems to happen (from a user pov).

**To Reproduce**
1. Open a writer document, don't change to edit mode
2. Tap on File > Export as > EPUB (.epub)

**Expected behavior**
The file dialog should show up, where a user can select the target folder.

**Actual behavior**
Nothing seems to happen (from a user pov).

**Tablet**
 - OS: iPadOS 16.3.1
 - Version: TestFlight 22.05.11 (22.5.202303212054)

plubius:
The fix for this bug is in the current TestFlight. Does the fix work for you?

nikslor:
Tested again in 22.05.12 (22.5.202304051456); works as expected now :)
Thanks a lot @plubius I think you can close this issue as fixed.

PR #6122, opened 2023-03-31
Fix issues #6103 and #6104 by setting the command to "downloadas-" on iOS

plubius:


pedropintosilva (approved):
Tested, it works, Thanks!

PR #6126, opened 2023-04-01
Dark mode improvements

OFFTKP:


OFFTKP:
Since we're in a bit of a rush to merge this and it seems that only cypress tests are failing, let's merge this

PR #6130, opened 2023-04-03
handle Uncaught TypeError: Cannot read properties of null

timar:
sheets[i].href can be null. Seen with proxy.php.

Change-Id: I3d774ace869d141973a92abab8ef9ac83e1e0018

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
During checks this code path is not exercised. The code is for the proxy (richdocumentscode) only. So it's not responsible for any failure.

PR #6137, opened 2023-04-04
access keys support is being prepared.

gokaysatir:
Change-Id: I3ba2249d8e2441d013c09bf8b552498661fd069f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -14,6 +14,14 @@
 	--cell-cursor-selection-border-color: var(--color-primary);
 }
 
+.access-key {
+	text-decoration: none;
+}
+
+.activate-underlines .access-key {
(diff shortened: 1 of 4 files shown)


pedropintosilva:
@gokaysatir @mmeeks as I have in my old pr: https://github.com/CollaboraOnline/online/pull/5794 we should toggle the mark only when press the modifiers. Ideally it would be an element that can be styled and hide/show independently of the text so in the case of my PR the semantic html element `mark`

pedropintosilva:
this then allows us to sure use underline but then we can also use something like floating hints

mmeeks:
Yep - for notebookbar - using floating hints would be just great =)

PR #6138, opened 2023-04-05
cypress: dont assert for fullwhite canvas

Rash419:
- for some reason in cypress env tiles still have formatting marks even if we hide it
- cannot reproduce outside of cypress env

Change-Id: If5224a5438576f81b3e1d1fa7f5d21293783516d

image

* Target version: master


gokaysatir (approved):
Thanks @Rash419 :)

PR #6140, opened 2023-04-05
jsdialogs: fix the initial focus if an editable element has it

hcvcastro:
The focusId element already has the focus, no need to
set initial element focus.

Change-Id: I25e3b5ba962aad9ce5bb1df26398297971b6a9e8


eszkadev (approved):
looks good

PR #6143, opened 2023-04-06
Disable Change Password button for user that don't own the document

JaumePujantell:
Change-Id: I2e1fc1e5b6e50cd7ec5ad9d78db36e76ce6426c9

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -846,7 +846,7 @@ bool ClientSession::_handleInput(const char *buffer, int length)
     }
     else if (tokens.equals(0, "removesession"))
     {
-        if (tokens.size() > 1 && (_isDocumentOwner || !isReadOnly()))
+        if (tokens.size() > 1 && (isDocumentOwner() || !isReadOnly()))
         {
             std::string sessionId = Util::encodeId(std::stoi(tokens[1]), 4);
             docBroker->broadcastMessage(firstLine);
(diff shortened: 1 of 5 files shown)


mmeeks (changes requested):
So - this looks fine - but one of the big confusions in the code base is the attempt to duplicate state between the supposedly shared 'Session' that has stuff in the kit and the coolwsd. I'm trying to kill that slowly.

I see no benefit to propagating this state to the Kit process - we don't seem to use it there (?); so can we back out the complexity of pushing the setting down from ClientSession -> Session, and leave it in ClientSession ? That should shrink the change substantially =)

Beyond that - we do a nice job of sending the setting to the JS and disabling the button: good - but all policy should also be enforced on the server-side - we should not allow pwd changing by non-owners there. Is that there ? (was quite a lot of change to read ;-).

Thanks !

wsd/ClientSession.cpp
```diff
@@ -877,8 +877,37 @@ bool ClientSession::_handleInput(const char *buffer, int length)
...
+        if (tokens.size() > 2)
+        {
+            std::string jsonString = tokens.cat("", 2);
+            try
+            {
+                Poco::JSON::Parser parser;
+                const Poco::Dynamic::Var result = parser.parse(jsonString);
+                const auto& object = result.extract<Poco::JSON::Object::Ptr>();
+                const std::string id = object->has("id") ? object->get("id").toString() : "";
+                if (id == "changepass" && _wopiFileInfo && !isDocumentOwner())
```
Ashod, line 891: Doesn't seem like we need `_wopiFileInfo` here.

wsd/ClientSession.cpp
```diff
@@ -877,8 +877,37 @@ bool ClientSession::_handleInput(const char *buffer, int length)
...
+                }
+            }
+            catch (const std::exception& exception)
+            {
+                return forwardToChild(firstLine, docBroker);
+            }
+        }
+        else
+        {
+            return forwardToChild(firstLine, docBroker);
```
Ashod, line 907: Since we always default to `forwardToChild`, probably better to have this as the unconditional fallback at the end, and return early where we don't want to default. The code should be simpler and more readable.

mmeeks:
Looks fine - can we squash it into one commit so we don't have these adds and removes in adjacent commits ? =)
Just do - git rebase -i HEAD~3 # and follow the instructions - replace 1st word with 'f' and 'f' for the two follow on commits (or perhaps you knew =).

Otherwise great- lets get it in - thanks ! =)

mmeeks (approved):
Looks great - thanks Jaume =)

PR #6146, opened 2023-04-06
Set focus to last element after a popup window is closed.

gokaysatir:
Change-Id: I99251c1f35b23a5b1a41bbe1820efcf090353db4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -68,6 +68,13 @@ L.Control.JSDialog = L.Control.extend({
 			return;
 		}
 
+		try {
+			this.dialogs[id].lastFocusedElement.focus();
+		}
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Code looks correct for general use. But probably we have to do more for dialogs which receive full update at some point (so onJSDialog is called again but focused element is inside old instance, then we will lost information about real lastFocusedElement)

gokaysatir:
> Code looks correct for general use. But probably we have to do more for dialogs which receive full update at some point (so onJSDialog is called again but focused element is inside old instance, then we will lost information about real lastFocusedElement)

Thank you for the comment, let me keep this in mind. We can pass the variable to new one when need arises.

PR #6147, opened 2023-04-06
Provide a way to focus on the menu from map.

gokaysatir:
Change-Id: I43bbb0d6a795ef11d43eb7ab2845f6f4674ca722

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -456,8 +456,27 @@ L.Map.Keyboard = L.Handler.extend({
 		}
 
 		// handle help - F1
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
when I press alt + shift + F1 correctly focus in the home tab

but I wonder if that's expected. Wouldn't be better to focus on the `selected` tab instead. This way user is able to quickly enter in that tabs' content or collapse

gokaysatir:
> when I press alt + shift + F1 correctly focus in the home tab
> 
> but I wonder if that's expected. Wouldn't be better to focus on the `selected` tab instead. This way user is able to quickly enter in that tabs' content or collapse

I knew you would come with a better idea :) Thanks, i will try it that way.

PR #6156, opened 2023-04-10
Private/gokay/loop focus uno dialogs

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

vmiklos:
Would it make sense to rather support alt-down here? core.git d65e85178abdd4f1bf068f161e49204e068bb5da and Word does with alt-down, not space/enter.

gokaysatir:
Thank you for the comment. It was showing an error upon pressing space and i wanted to turn that into feature :)

Let me note this and i will also add alt-down shortcuts while keeping the current shortcuts. Would it work?

PR #6163, opened 2023-04-10
Avoid duplicate ids.

gokaysatir:
Change-Id: Ie2ff0d212e0dc55eadf32e9fbbfb5abd0ee8bf68

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -1070,6 +1070,7 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 				$(table).addClass(this.options.cssClass);
 				$(table).addClass('vertical');
 				var childObject = L.DomUtil.create('div', 'row ' + this.options.cssClass, table);
+				childObject.id = tableId ? tableId + '-row' : '';
 			} else {
 				childObject = td;
 			}
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -2453,13 +2453,13 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 			'enabled': 'true',
 			'children': [
 				{
-					'id': tabName + ' Tab',
+					'id': tabName + '-Tab',
 					'type': 'container',
 					'text': '',
 					'enabled': 'true',
 					'children': [
 						{
-							'id': tabName,
+							'id': tabName + '-container',
 							'type': 'container',
 							'text': '',
 							'enabled': 'true',


browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -1065,11 +1065,12 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 
 			var hasManyChildren = childData.children && childData.children.length > 1;
 			if (hasManyChildren) {
-				var tableId = childData.id ? childData.id.replace(' ', '') : '';
+				var tableId = childData.id ? childData.id.replace(' ', '') + ' container': String(parseInt(Math.random() * 1000));
 				var table = L.DomUtil.createWithId('div', tableId, td);
 				$(table).addClass(this.options.cssClass);
 				$(table).addClass('vertical');
 				var childObject = L.DomUtil.create('div', 'row ' + this.options.cssClass, table);
+				childObject.id = tableId + ' row';
```
eszkadev, line 1073: please do not use spaces in 'id', so we can target it with css

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -1065,11 +1065,12 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 
 			var hasManyChildren = childData.children && childData.children.length > 1;
 			if (hasManyChildren) {
-				var tableId = childData.id ? childData.id.replace(' ', '') : '';
+				var tableId = childData.id ? childData.id.replace(' ', '') + ' container': String(parseInt(Math.random() * 1000));
```
eszkadev, line 1068: this can be problematic... we have to use exactly the same id as we receive with JSON so later when we receive actions like 'hide' we will find that element

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -1065,11 +1065,12 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 
 			var hasManyChildren = childData.children && childData.children.length > 1;
 			if (hasManyChildren) {
-				var tableId = childData.id ? childData.id.replace(' ', '') : '';
+				var tableId = childData.id ? childData.id.replace(' ', '') + '-container': String(parseInt(Math.random() * 1000));
```
eszkadev, line 1068: there is function `_makeIdUnique` in parent class (JSDialogBuilder) for that

gokaysatir: Thanks for checking. I left this part as is for now. It doesn't assign an id if table doesn't have one.

eszkadev (approved):

PR #6166, opened 2023-04-11
private/pedro/css refactor vars

pedropintosilva:
- CSS Refactoring: one btn-size to rule them all
- CSS refactor: Fix duplicated & outdate px across light/dark modes


pedropintosilva:
this is just CSS refactoring, no need to wait for CodeQL cpp to finish and the same is true of the rest

PR #6167, opened 2023-04-11
Better Modified Tracking and Upload Headers

Ashod:
The primary purpose of this PR is to make sure the IsModifiedByUser header passed to storage is accurate. We now get this flag from Core in the uno:Save response, so there are no race conditions.

- wsd: do not allow multiple saves in flight
- wsd: process the modified state from uno:Save
- wsd: track the storage attributes more accurately
- wsd: test: new test for superfluous save commands
- wsd: test: rulerupdate is an expected message

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1372,10 +1372,30 @@ bool DocumentBroker::isStorageOutdated() const
     return currentModifiedTime != lastModifiedTime;
 }
 
-void DocumentBroker::handleSaveResponse(const std::shared_ptr<ClientSession>& session, bool success,
-                                        const std::string& result)
+void DocumentBroker::handleSaveResponse(const std::shared_ptr<ClientSession>& session,
+                                        const Poco::JSON::Object::Ptr& json)
 {
     assertCorrectThread();
+
+    const bool success = json->get("success").toString() == "true";
+    std::string result;
+    if (json->has("result"))
+    {
+        const Poco::Dynamic::Var parsedResultJSON = json->get("result");
+        const auto& resultObj = parsedResultJSON.extract<Poco::JSON::Object::Ptr>();
+        if (resultObj->get("type").toString() == "string")
+            result = resultObj->get("value").toString();
+    }
+
+    if (json->has("wasModified"))
+    {
+        // If Core reports the modified state before saving,
+        // use it to report to the Storage with more confidence.
+        const bool wasModified = (json->get("wasModified").toString() == "true");
+        LOG_DBG("Core reported that the file was " << (wasModified ? "" : "not ")
+                                                   << "modified before saving");
+        _nextStorageAttrs.setUserModified(wasModified);
+    }
 
     // Record that we got a response to avoid timing out on saving.
     _saveManager.setLastSaveResult(success || result == "unmodified");
@@ -2400,6 +2420,7 @@ bool DocumentBroker::sendUnoSave(const std::shared_ptr<ClientSession>& session,
     oss << '}';
 
     // At this point, if we have any potential modifications, we need to capture the fact.
+    // If Core does report something different after saving, we'll update this flag.
     _nextStorageAttrs.setUserModified(isModified() || haveModifyActivityAfterSaveRequest());
 
     // Note: It's odd to capture these here, but this function is used from ClientSession too.
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1397,6 +1397,13 @@ void DocumentBroker::handleSaveResponse(const std::shared_ptr<ClientSession>& se
         _nextStorageAttrs.setUserModified(wasModified);
     }
 
+    // Update the storage attributes to capture what's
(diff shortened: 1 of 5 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -2118,10 +2118,35 @@ DocumentBroker::NeedToSave DocumentBroker::needToSaveToDisk() const
     return NeedToSave::No;
 }
 
+bool DocumentBroker::manualSave(const std::shared_ptr<ClientSession>& session,
+                                bool dontTerminateEdit, bool dontSaveIfUnmodified,
+                                const std::string& extendedData)
+{
+    // If we aren't saving already.
+    if (_docState.activity() != DocumentState::Activity::Save)
```
mmeeks, line 2151: silly question - how does this 'manual' save get re-dispatched later when we are not saving ? =)

wsd/DocumentBroker.cpp
```diff
@@ -2118,10 +2118,35 @@ DocumentBroker::NeedToSave DocumentBroker::needToSaveToDisk() const
...
+
+    LOG_DBG("Document [" << _docKey << "] is currently saving and cannot issue another save");
+    return false;
+}
+
 bool DocumentBroker::autoSave(const bool force, const bool dontSaveIfUnmodified)
 {
     assertCorrectThread();
 
+    // If we aren't saving already.
```
mmeeks, line 2167: Fair enough to cancel an autosave if we are saving I guess.

wsd/DocumentBroker.cpp
```diff
@@ -1397,6 +1397,13 @@ void DocumentBroker::handleSaveResponse(const std::shared_ptr<ClientSession>& se
         _nextStorageAttrs.setUserModified(wasModified);
     }
 
+    // Update the storage attributes to capture what's
+    // new and applies to this new version and reset the next.
+    // These are the attributes of the version to be uploaded.
+    // Note: these are owned by us and this is thread-safe.
+    _currentStorageAttrs.merge(_nextStorageAttrs);
```
mmeeks, line 1404: It was not -that- clear to me how these attributes are associated with an ongoing save operation; is it implied that there is only one?
I would assume that we have an existing asynchronous object to represent an ongoing save which this shadows - but on the DocumentBroker - is it in the right place ?

mmeeks (commented):
Didn't wrestle hard with this or understand it deeply.
Added some random questions to perhaps help =)

wsd/DocumentBroker.cpp
Ashod: If the user wants to save again in the future, they can use any of the shortcuts to do so. This only prevents multiple saves in flight. (Unless I misunderstood the question.)

wsd/DocumentBroker.cpp
Ashod: Good questions. There are 3 sets of attributes now, to capture the attributes we are uploading (to allow for re-trying when we fail), to track the attributes of an in-flight save (to be merged with the uploading attributes once a new version is saved), and the "current" attributes for modifications that the user is making (while a save and an earlier upload might be happening in parallel).

The idea is to track the attributes for each of these 3 parallel activities separately, and only merge them when one activity fails and needs to retry with the new state of the doc/file.

DocumentBroker is the right place, because it's the only place where we track saving and uploading as well as the modification state of the document.

Ashod:
@gmasei11, would appreciate if you or Cosmin could take a look at these changes that address some of the issues you reported. Of course testing the fix would be even better, if possible.

Thanks in advance!

wsd/DocumentBroker.cpp
mmeeks: Hmm - but if I have a large document, hit save, while it is saving (or at least up-loading) - I type "Sell everything right now!" and hit save - the save will effectively not happen; is that so ? - hopefully it won't loose my last modification in the long run, but ... is that expected? and/or will we get the right modification status in the UI - ultimately, as long as the UI says it is not saved - I suppose that's ok for now.

Ashod: Correct. The document will still say modified. I see your concern, though, in that we don't want to lose the last modification. That shouldn't happen even before this change, as we are more aggressive with saves, even when modified=false, if we had input from the user, we still save. This PR makes the modified flag more reliable and the reduces in-flight save commands to avoid unnecessary duplicates.

Notice that uploading doesn't affect saving, so that could still happen in parallel.

mmeeks: Ah - it is fine; it is a "de-bounce" feature; and caveat user etc. =) just wanted to understand; lets not over-complicate things by handling it.

cosminh11:
I ran some local tests on a local installation and I can confirm that these changes fix the repeated uploads with the same content (when doing multiple saves in flight).

Ashod:
> I ran some local tests on a local installation and I can confirm that these changes fix the repeated uploads with the same content (when doing multiple saves in flight).

Thanks @cosminh11 for the confirmation!

Ashod:
Unrelated test failures. Merging as reviewed and vetted.

PR #6169, opened 2023-04-11
Make menu dropdowns accessible.

gokaysatir:
Change-Id: Ie48d5c8f43d54ba998faf862a51367346d7e1257

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

eszkadev:
it would be also good to add  handler so we can navigate using arrow keys + add aria role for menu and items

PR #6170, opened 2023-04-11
Rename Link Picker to Smart Picker

pedropintosilva:
So it's consistent with the latest NC strings

Change-Id: Ic1d49c6d871aee86f80aeb5c4e6936b33f3fc93c


pedropintosilva:
@eszkadev and @juliushaertl as discussed I have updated the strings, but I still see the popup title "Link Picker" from NC side

juliushaertl (approved):
Thanks a lot

pedropintosilva:
this is label text only, the id remains the same

PR #6172, opened 2023-04-12
Show correct launch URI

eszkadev:
this fixes regression from:
commit 158e1a125f9a839893f5289329ba37584077cf5b
wsd: avoid using the default port for unit-tests

where 0 was shown as port in example launch URI on make run


eszkadev:
we will prepare different fix

PR #6173, opened 2023-04-12
jsdialog: conditional formatting dialog

eszkadev:
requires:
https://gerrit.libreoffice.org/c/core/+/150250
https://gerrit.libreoffice.org/c/core/+/150251

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -52,6 +52,10 @@ button:not(.ui-tab):not(.ui-corner-all):not(.button-primary):not(.unobutton) {
 	justify-content: space-between;
 }
 
+.jsdialog.ui-button-box {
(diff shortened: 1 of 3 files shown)


pedropintosilva (changes requested):
Thanks @eszkadev but it doesn't look right  the whole layout

|  Online | Core |
| ---------- | ---------- |
| Screenshot_20230414_122625 | Screenshot_20230414_122457 |

I think we need to introduce a grid modifier that can be used in cases like that where we need the grid to have all equal column width (even if no content available) with `grid-template-columns: repeat(' + colT  + ', 1fr);` : 

Screenshot_20230414_123057

eszkadev:
we have simplified dialog now

PR #6175, opened 2023-04-12
Fix button ration

pedropintosilva:
Before this commit we were setting buttons in places such as
the sidebar with weird ratio (not 1:1) this might lead to problems
when aligning, adjusting gaps. Better to adjust gap or margin if we
need bigger spaces in between

Change-Id: Id60cd6353de2ace2a066c0d1624000a09f0ca98e


pedropintosilva:
this change doesn't affect tests

PR #6176, opened 2023-04-12
private/ash/misc

Ashod:
- wsd: announce the debug URLs to the user after server runs
- wsd: better socket logging


mmeeks (approved):
All looks good to me. CI failure looks typical - Lets get it in.

PR #6180, opened 2023-04-13
Do not assume who the user needs to contact for storage space

pedropintosilva:
- I some situations %storageserver domain is not the right place
to redirect users (integrator might have another support system in
place)
- Do not assume the document will be read-only, integrator might have
something different in place

Change-Id: I0008819665a96677cca06d9654e6629cfcac98b1

### DIFF
--- a/browser/src/errormessages.js
+++ b/browser/src/errormessages.js
@@ -39,7 +39,7 @@ if (window.ThisIsAMobileApp) {
 } else {
 	errorMessages.storage = {
 		loadfailed: _('Failed to read document from storage. Please contact your storage server (%storageserver) administrator.'),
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Thanks @pedropintosilva. Looks good, but should we keep "Document will now be read-only." as it's useful information ?

pedropintosilva:
@Ashod thanks for the review. I removed the read-only information because I thought that is not true for all cases. Meaning, isn't it possible to  1.run out of space 2. when attempting to load a document the only thing that loads is part  of the UI with that respective message ? It has happened to me. So, I would assume that having "Document will now be read-only." is false in those cases and thus a bit confusing. What do you think?  (eager to know your thoughts as you probably are more versed in this area than I)

Ashod (approved):
Thanks @pedropintosilva, looks good.

PR #6185, opened 2023-04-13
cookie fixes

hcvcastro:
- browser: admin: fix vex dialog on socket close
- wsd: fix secure cookie Set-Cookie

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -259,8 +259,7 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request,
     Poco::Net::HTTPCookie cookie("jwt", jwtToken);
     // bundlify appears to add an extra /dist -> dist/dist/admin
     cookie.setPath(COOLWSD::ServiceRoot + "/browser/dist/");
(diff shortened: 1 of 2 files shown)


wsd/FileServer.cpp
```diff
@@ -259,8 +259,7 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request,
     Poco::Net::HTTPCookie cookie("jwt", jwtToken);
     // bundlify appears to add an extra /dist -> dist/dist/admin
     cookie.setPath(COOLWSD::ServiceRoot + "/browser/dist/");
-    cookie.setSecure(COOLWSD::isSSLEnabled() ||
```
timar, line 262: I wonder if it's good. What happens if coolwsd server is behind a reverse proxy, we reach the reverse proxy by https, and SSL termination is enabled?

hcvcastro: Can you please test with a reverse proxy to collect data, what happens?. My expectation the cookies will be sent, otherwise if it is not a reverse proxy the cookie is blocked by the browser.

Ashod (approved):
Thank @hcvcastro. Per offline note about test passing in the scenario highlighted by @timar, let's get it in.

PR #6188, opened 2023-04-14
Socket Improvements

Ashod:
- wsd: drain the wakeup pipe on each read
- wsd: update the poll round-robin index right before usage
- wsd: better merge new sockets into poll sockets
- wsd: make removeSockets private
- wsd: always handle socket polls without skipping

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -426,8 +426,9 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
         // We use the _pollStartIndex to start the polling at a different index each time. Do some
         // sanity check first to handle the case where we removed one or several sockets last time.
+        ++_pollStartIndex;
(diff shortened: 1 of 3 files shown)


net/Socket.cpp
```diff
@@ -426,6 +426,7 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
         // We use the _pollStartIndex to start the polling at a different index each time. Do some
         // sanity check first to handle the case where we removed one or several sockets last time.
+        ++_pollStartIndex;
```
mmeeks, line 433: I'm missing a modulus here. This looks like it increments until it hits size-1 and stays there. Is that expected ?

mmeeks (approved):
All good - but needs one change in the middle; please do push with that fixed.

mmeeks:
Cypress seems unrelated - lets get it in.

PR #6190, opened 2023-04-14
jsdialog: a11y: keyboard navigation for treeviews/listboxes

eszkadev:
Can be tested with:

treeview:
    - calc run macro dialog
    - Zotero / Writer -> reference -> Add citation dialog (left one)

listbox:
    - writer -> fromat -> character (font listbox)
    - mention popup

header listbox:
    - Zotero / Writer -> reference -> Add citation dialog (right one)

In commit messages I added more details, based on https://www.w3.org/WAI/ARIA/apg/patterns/ recommendations


gokaysatir (approved):

eszkadev:
all 3 failed tests passed locally

Issue #5876, opened 2023-02-19
Large PDF renders really slow leads in IOS App to blank screen


sebastiaanveld:
**Describe the bug**
Large PDF renders really slow (attached document) in (Staging) CODE. 
Using the IOS App opening this file after rendering (slowly) few slides leads into a blank screen (white page) after which I have to force close the App.
Using multiple other PDF's render sometimes partly leaving few pages blank (this was an issue a was about trying to duplicate).
This issue may be related #5215 ?

**To Reproduce**
1. Go to 'Collabora Nextcloud Staging' https://staging.eu.collaboraonline.com
2. Upload the document "Cloudervaringsdocument+NCSC" 
3. Use option "Open with Nextcloud Office" for Editing.
4. Document opens, but renders pages really show. After all previews have rendered click a page, which builds the visually as one looks at it.
5. Open the same document with the Collabora Office IOS App on iPhone
6. Few previews render and while scolling down page numbers are show and filled while being rendered. 
7. Suddenly the page turns white and I have to force close the App to get it back running.

**Expected behavior**
CODE: PDF should render more quickly in general (if possible<G>)
IOS APP: In IOS App the page should not turn white, turning the page/app unusable.

**Actual behavior**
CODE: slow rendering.
IOS APP: page turns while while scrolling, turning the app unusable.

**Screenshots**
Just a white page in the IOS App.

**Desktop (please complete the following information)**
 - OS: Windows 11, latest patches
 - Browser: Firefox 110 64 bit
 - Version: Collabora Staging:
COOLWSD version: 23.05.0.1 git hash: [https://github.com/CollaboraOnline/online/commit/d3518f7fe000984a72a9ef4b1cbaf188b973d23f]
LOKit version: built on Collabora Office 23.05.0.1 git hash: [ff9a1ba]
Served by: openSUSE Leap 15.3 [6e930f1a]

**Smartphone (please complete the following information)**
 - Device: iPhone 14
 - OS: IOS 16.3.1
 - Collabora Office Testflight 22.05.10 (22.5.202302180032)

**Additional context**
[Cloudervaringsdocument+NCSC.pdf](https://github.com/CollaboraOnline/online/files/10777062/Cloudervaringsdocument%2BNCSC.pdf)

plubius:
[...]

PR #6191, opened 2023-04-14
Related to issue #5876: reduce excessive copying of message data

plubius:
Many .png tile messages are very large so avoid using -[NSString stringByAppendingString:] calls to create the JavaScript string.


mmeeks (approved):
Looks great =) I wonder if we could pre-allocate std::vector<char> data - to avoid lots of reallocs as we push lots of stuff to it; since it's not long-lived a larger alloc might be good - length * 4 or whatever ? =)

plubius:
> Looks great =) I wonder if we could pre-allocate std::vector data - to avoid lots of reallocs as we push lots of stuff to it; since it's not long-lived a larger alloc might be good - length * 4 or whatever ? =)

Good idea. I've added a new commit to this pull request. I did some printfs and it appears that binary messages consistently use up at least 75% of (length * 4) + 1.

common/MobileApp.cpp
```diff
@@ -38,7 +38,10 @@ DocumentData & DocumentData::get(unsigned docId)
 void DocumentData::deallocate(unsigned docId)
 {
     assert(idToDocDataMap.find(docId) != idToDocDataMap.end());
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused"
     DocumentData &d = get(docId);
```
mmeeks, line 43: Urk, all these pragmas make for grim reading & maintainability. Can't we just delete this redundant line ? =) Ditto for other places where things are not needed (?) What are all the:

#pragma clang diagnostic ignored "-Wconditional-uninitialized"

warnings in fact ? - can we do something more elegant there; we really want fewer conditionals & pre-processing in the code if we can.

mmeeks (commented):
Reserving looks awesome =) thanks !

common/MobileApp.cpp
plubius: OK. I was able to remove all the #pragma clang's as I found the build setting in Xcode that was causing this: I changed "Uninitialized Variables" from "Yes (Aggressive)" to "Yes".

I also added comments to the two cases that have odd code that generated warnings: the destructor case and the "comma in an if statement" case. Maybe they are cruft but maybe these specific cases were intentional so better to be safe and leave the code as is but without warnings.

Overall my goal was to cut down the number of warnings to a very small number. I found that several deprecation warnings were being missed because they scrolled out of view in Xcode and since Apple routinely deprecates things, I wanted to get those pushed to the top of the build warnings list.

plubius:
OK. I amended the last patch to revert all of the #pragma clang's. I found the "Uninitialized Variable" build setting in Xcode and changed it from "Yes (Aggressive)" to "Yes" and that removes the false positive uninitialized warnings.

The other two warning cases I solved by slight modifications to the code but without any logic changes and added a comment to each.

I didn't want to clean up the last two warning cases because, although they appear odd, my goal was only to eliminate the very large number of Xcode warnings. Because of all the noise, I found that I was not seeing several deprecated warnings. Apple deprecates things constantly and many deprecations turn into crashes on newer OS versions so I wanted to increase the visibility of deprecated warnings without changing any code.

mmeeks (approved):
Looks great - thanks Patrick ! =)

PR #6195, opened 2023-04-15
More Socket Improvements

Ashod:
Please see commit log for detailed explanation of these changes.

- wsd: correct log argument order
- wsd: more tracing of socket and poll activity
- wsd: reduce SocketPoll lock contention
- wsd: handle read failure on closed socket better
- wsd: return SSL_do_handshake error-codes unmodified

### DIFF
--- a/net/SslSocket.hpp
+++ b/net/SslSocket.hpp
@@ -60,13 +60,17 @@ public:
 
         if (isClient)
         {
+            LOG_TRC("Setting SSL into connect state");
             SSL_set_connect_state(_ssl);
             if (SSL_connect(_ssl) == 0)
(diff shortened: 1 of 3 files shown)


net/SslSocket.hpp
```diff
@@ -226,7 +234,10 @@ class SslStreamSocket final : public StreamSocket
             {
                 rc = handleSslState(rc, "handshake");
                 if (rc <= 0)
-                    return rc != 0;
```
Ashod, line 229: Offensive line here.

mmeeks (approved):
Lets try to get this lot in - hopefully it will improve things. Thanks Ash !

PR #6196, opened 2023-04-15
Cypress: wait for cursor to move before checking where it moved.

mmeeks:
Change-Id: Ieb6685a29dbb5dbe7e8479b7d604332bb6657dc1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/common/helper.js
+++ b/cypress_test/integration_tests/common/helper.js
@@ -629,7 +629,8 @@ function closeDocument(fileName, testState) {
 			return;
 		}
 
-		cy.get('#uptime')
+		// ensure the page is fully loaded & active.
+		cy.get('#uptime').should('exist')
 			.should('not.have.text', '0');
 
 		// We have all lines of document infos as one long string.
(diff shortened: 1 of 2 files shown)


mmeeks:
Not had time to test this - and no doubt it doesn't work well - but lets see ... =)

Ashod (approved):
Looks plausible to me. Thanks!

cypress_test/integration_tests/common/helper.js
```diff
@@ -629,7 +629,8 @@ function closeDocument(fileName, testState) {
 			return;
 		}
 
-		cy.get('#uptime')
+		// ensure the page is fully loaded & active.
+		cy.get('#uptime').should('exist')
```
JaumePujantell, line 633: I fixed this in https://github.com/CollaboraOnline/online/pull/6209, the proper way is to use .its(text) witch waits for the text property to exist.

mmeeks: I saw that - thanks ! =) what frightens me most is that I don't really see how it can not have a value - we serve the page with that - and if it's a partial page load - our cy.get() should continue dispatching until the element exists - and it should have the value '0' as it is served, and then get atomically set to something valid by the JS - without any other states - at least that's what I read; which scared me.

mmeeks:
Abandoning - I suspect I was not that confident in how Cypress worked =)

PR #6197, opened 2023-04-16
Fix debug build breakage due to undefined DEBUG_ABSSRCDIR macro on iOS

plubius:
On iOS and Android, a custom build process and the Online server code runs locally in the same process as the app so defer modifying those separate build processes until they are needed on mobile platforms.


plubius:
Cancelling this pull request as the commits are in pull request #6218

PR #6201, opened 2023-04-17
Tabbed view: Tabs: Improve html structure

pedropintosilva:
Best to avoid span elements around button text
The innerHTML here is harmless since we are using a string created
by us within the same function.

Change-Id: I72af17cd2788e7ba4b3ec1c255058e18884e0be1


gokaysatir (approved):

pedropintosilva:
cypress passed locally

PR #6202, opened 2023-04-17
Fix #6192: onMessage error: e.data.startsWith is not a function

### LINKED ISSUE #6192: onMessage error: e.data.startsWith is not a function
Bug was originally written for Nextcloud Rich Documents because I didn't realize Collabora was a separate UI program from Nextcloud: https://github.com/nextcloud/richdocuments/issues/2914

**Describe the bug**
When the Chrome extension [Ruffle](https://ruffle.rs) is installed (which injects JS into the browser), loading a Collabora document results in multiple errors with the following message:

```
bundle.js:22418 Uncaught TypeError: e.data.startsWith is not a function
    at NewClass.onMessage (bundle.js:22418:33)
    at handler (bundle.js:21815:35)
```

It occurs around this bit of code:

```js
        onMessage: function(e) {
            if (e.data.startsWith("updatecheck-show"))
                return;
            var data = JSON.parse(e.data);
```

I believe it can be fixed by prepending this line to the top of onMessage:
```js
if (typeof e.data !== 'string') return; // Some extensions may inject scripts resulting in load events that are not strings
```

Here's an example `e.data` which is not a string, causing the error:
```json
{
    "to": "ruffle_content13813631390",
    "index": 0,
    "data": {}
}
```

**To Reproduce**
1. Install Ruffle chrome extension
2. Load a Collabora document

**Expected behavior**
N/A

**Screenshots**
N/A

**Client details:**
Chromium, Desktop

## Server details

**Operating system**: Windows

**Web server:** nextcloud-aio

**Database:** N/A

**PHP version:** N/A

**Nextcloud version:** 25 (beta)

**Version of the richdocuments app** beta channel

**Version of Collabora Online** beta channel

timar:
Change-Id: I6e548370c9569117b401f5db3f750fb3633cb2cd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Feedback.js
+++ b/browser/src/map/handler/Map.Feedback.js
@@ -111,6 +111,9 @@ L.Map.Feedback = L.Handler.extend({
 	},
 
 	onMessage: function (e) {
+		if (typeof e.data !== 'string')
+			return; // Some extensions may inject scripts resulting in load events that are not strings
+
 		if (e.data.startsWith('updatecheck-show'))
 			return;
 
--- a/browser/src/map/handler/Map.Welcome.js
+++ b/browser/src/map/handler/Map.Welcome.js
@@ -90,6 +90,9 @@ L.Map.Welcome = L.Handler.extend({
 	},
 
 	onMessage: function (e) {
+		if (typeof e.data !== 'string')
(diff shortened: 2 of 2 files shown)


mmeeks (approved):
Looks great to me =)

PR #6203, opened 2023-04-17
Fix assert failure when running "My Mac (Designed for iPad)" in Xcode

plubius:
LANG values such as en_US.UTF-8 trigger an assert in the LibreOffice code so replace all "_" characters with "-" characters.


mmeeks (approved):
=)

PR #6208, opened 2023-04-18
Tabbed view: Fix context tabs visibility

pedropintosilva:
With the a11y improvements from
9d3d0cdee73e84b8446324cf55fb1feefe720cd8
we now add !important to visibility: hidden that in turn makes the
browser ignore the inline display block. The result: we get a new tab
content without the tab itself (it's hidden)
- Fix this by removing the 'hidden' class when the tab button needs to
be shown

Change-Id: I276d3fa62d89b32935de0357d0b24f558fcf3fd6


pedropintosilva:
cd cypress_test
make run-multi spec=writer/track_changes_spec.js 

is unrelated to this change. Plus that sometimes fails sometimes doesn't with or without this change

PR #6209, opened 2023-04-18
Various cypress fixes

JaumePujantell:
This changes make cypress tests more reliable, specially in slow or loaded machines.

Change-Id: I574b0db3e81e57a2dc682fbe2ed3f0301ad8e52d


mmeeks:
Looks good to me - the infamous failing calc/sheet_operations_spec.js seems to pass in your build - quirk or not - who knows =) but looks reasonable; lets get it in and iterate ! =) tinderboxes are pure red for a while here  ...

PR #6211, opened 2023-04-19
use getentropy() for BSD systems

bayramcicek:
Change-Id: Id6a2629c06d641eb4e7cf3991de4036d2f7b346e

* Resolves: #5851 
* Target version: master 

### Summary

- instead of using /dev/[u]random devices, 
use getentropy() to make direct system calls if the system is a BSD.

- if getentropy() fails, we need to fall back to "/dev/[u]random" approach.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -20,6 +20,10 @@
 
 #include <poll.h>
 
+#ifdef HAVE_GETENTROPY
+#  include <unistd.h>
+#endif
+
 #ifdef HAVE_SYS_RANDOM_H
 #  include <sys/random.h>
 #endif
@@ -114,12 +118,24 @@ namespace Util
             std::vector<char> v(length);
 
             int len = 0;
-#ifdef HAVE_SYS_RANDOM_H
+            /* use "/dev/[u]random" by default in case of
+            getentropy() or getrandom() fails,
+            or they don't available */
+            bool useDevRandom = true;
+#ifdef HAVE_GETENTROPY
+            len = getentropy(v.data(), length);
+
+            // getentropy() works, no need to use "/dev/[u]random"
+            if (len != -1)
+                useDevRandom = false;
+#elif defined HAVE_SYS_RANDOM_H
             len = getrandom(v.data(), length, GRND_NONBLOCK);
 
-            // if getrandom() fails, we fall back to "/dev/[u]random" approach.
-            if (len != length)
+            // getrandom() works, no need to use "/dev/[u]random"
+            if (len != -1)
+                useDevRandom = false;
 #endif
+            if (useDevRandom)
             {
                 const int fd = open("/dev/urandom", O_RDONLY);
                 if (fd < 0 ||
--- a/configure.ac
+++ b/configure.ac
@@ -1070,6 +1070,18 @@ else
     AC_DEFINE([HAVE_SYS_RANDOM_H],0,[Define to 1 if you have the <sys/random.h> header file.])
 fi
 
+AC_LINK_IFELSE(
+    [AC_LANG_SOURCE([[
+    #include <unistd.h>
+    int main()
+    {
+        char buffer[8];
+        getentropy(buffer, sizeof(buffer));
+    }
+    ]])],
+    [AC_DEFINE([HAVE_GETENTROPY],1,[Define to 1 if you have getentropy() function in <unistd.h> header file.])],
+    [AC_DEFINE([HAVE_GETENTROPY],0,[Define to 1 if you have getentropy() function in <unistd.h> header file.])])
+
 AS_IF([test -n "$LOKIT_PATH"],
       [CPPFLAGS="$CPPFLAGS -I${LOKIT_PATH}"])
 lokit_msg="$LOKIT_PATH"


bayramcicek:
Hi @Ashod. I'd appreciate it if you could review this patch.

common/Util.cpp
```diff
@@ -131,7 +138,13 @@ namespace Util
             // a poor fallback but something.
             std::vector<char> random = getBytes(length);
             int len = 0;
-#ifdef HAVE_SYS_RANDOM_H
+#ifdef BSD
+            len = getentropy(random.data(), length);
+
+            // if getentropy() fails, it will return -1.
+            // then, we should fall back to "/dev/[u]random" approach.
```
Ashod, line 145: I guess we could say `we fall back to getrandom(), if available, otherwise we try "/dev/[u]random".` (or something similar).

Ashod (approved):
Thanks @bayramcicek! Sorry for the delayed review. Looks good to me, but added @arrowd to get more feedback. Let's wait for Gleb's input, otherwise, happy to merge this soon.

arrowd:
Looks good, although I'd introduce the `HAVE_GETENTROPY` check and use that, rather than using platform checks.

common/Util.cpp
bayramcicek: > I guess we could say `we fall back to getrandom(), if available, otherwise we try "/dev/[u]random".` (or something similar).

But if we are in `#ifdef BSD` code block and  `getentropy()` fails in here, then we can't(?) execute `#elif defined HAVE_SYS_RANDOM_H` code clock - because it is `#elif` directive? Then we can't fall back to `getrandom()`, but we can only use `"/dev/[u]random"` - if `getentropy()` fails. Therefore, imho, we can(?) only use something like `// if getentropy() fails, try "/dev/[u]random"` - if we stick with the current approach.

but:

If we want to fall back to `getrandom()` if `getentropy()` fails in `#ifdef BSD` code block(as you stated), then I can update the code accordingly.

(to sum up) Which one we should use? :
1) if `BSD` -> use `getentropy()` -> if `getentropy()` fails -> (skip `getrandom()`) -> use `"/dev/urandom"`. (current approach)
or
2) if `BSD` -> use `getentropy()` -> if `getentropy()` fails -> use `getrandom()` -> if `getrandom()` fails -> use `"/dev/urandom"`.

Thanks for the review.

arrowd: According to man, `getentropy` fails either with `EFAULT` due to invalid buffer pointer or `EIO` in case of "some other fatal error", which probably means that there's something really wrong with the system.

And at least of FreeBSD `getrandom` does the same as `getentropy`, so I think there is no point in trying both functions.

bayramcicek:
> Looks good, although I'd introduce the `HAVE_GETENTROPY` check and use that, rather than using platform checks.

Imho;
On my system(Manjaro/Linux), I can use `getentropy` from both `<unistd.h>` and `<sys/random.h>`. So, as far as I understand(but I may be wrong), linux-based systems can also use `getentropy` in `<unistd.h>`. But we should use `getentropy` only for BSD systems. Therefore, I didn't remove the platform check, but moved it to `configure.ac` file.

Thanks for the reviews!

configure.ac
```diff
@@ -1100,6 +1100,31 @@ else
     AC_DEFINE([HAVE_SYS_RANDOM_H],0,[Define to 1 if you have the <sys/random.h> header file.])
 fi
 
+HAVE_GETENTROPY=
+AC_LANG_PUSH([C++])
```
arrowd, line 1104: Why is this needed?

configure.ac
```diff
@@ -1100,6 +1100,31 @@ else
...
+AC_LINK_IFELSE(
+    [AC_LANG_SOURCE([[
+    #if defined(__unix__) || !defined(__APPLE__) && defined(__MACH__)
+        #include <sys/param.h>
+        #if defined(BSD) // FreeBSD, NetBSD, OpenBSD, DragonFly BSD
```
arrowd, line 1109: Why check for BSD? Ultimately we're onle interested if we have `getentropy` in the `unistd.h` header. I'd just check for that without looking back at underlying OS.

arrowd (approved):

configure.ac
bayramcicek: I realize that there is no need to check the script for C++ compiler. Man page of autoconf (1) says `configure scripts check for the C compiler and its features by default.`. Then, imho, `AC_LANG_PUSH([C++])` is not necessary so I removed it.

(1): [https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Language-Choice.html](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Language-Choice.html)

bayramcicek:
Hi @Ashod. Should we merge this patch? 
Thanks.

mmeeks:
@bayramcicek looks like a great patch - can you re-base and get this merged ? =)

common/Util.cpp
```diff
@@ -114,7 +118,12 @@ namespace Util
             std::vector<char> v(length);
 
             int len = 0;
-#ifdef HAVE_SYS_RANDOM_H
+#ifdef HAVE_GETENTROPY
+            len = getentropy(v.data(), length);
+
+            // if getentropy() fails, try "/dev/[u]random"
+            if (len == -1)
```
thebearon, line 125: I find the if condition and body being separated by macro checks (and thus possibly not even being a conditional block) a bit dangerous. Not sure what the best way to handle it would be, I would perhaps add a bool variable for the if condition that has a default value true, and the code inside the #ifdef/#elif sets it according to the condition.

bayramcicek: Done.
Thanks.

mmeeks (approved):
Great - lets get it in =)

PR #6212, opened 2023-04-19
Fixes for issue #5876: manage canvas contexts and content process crashes

plubius:


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -6765,9 +6765,23 @@ L.CanvasTileLayer = L.Layer.extend({
...
 
+		// Partial fix for issue #5876 discard canvas contexts immediately
+		// WKWebView has a hardcoded memory limit for all canvas contexts
+		// so immediately convert the canvas to an image and release the
+		// canvas context's backing store.
+		// This bug only appears in the iOS app because most .png tiles are
+		// passed as raw data, not as "data:" strings, in the iOS app.
+		if (window.ThisIsTheiOSApp) {
+			tile.el = document.createElement('img');
+			tile.el.src = canvas.toDataURL();
```
mmeeks, line 6779: Ok - so 'toDataURL' does a compression as PNG - which uses deflate - which is hyper-slow and (I guess) single threaded. I would expect this to introduce a major negative performance impact =) not least because to get it back and apply a delta eg. typing 't' we need to de-compress, render to canvas, apply the 't' delta as pixels, and then compress back to png =)

Probably we should keep a cache of the last <N> tiles which we could manange in this method - and keep at least the last screen-full of them around in canvas form, and compress the rest to 'img' =)

Possibly we could also listen to some OS 'trim' event and compress them if the OS is desparate for space a s bonus =)

Thanks !

plubius:
Very good points about performance. My patch really isn't production-ready, but at least we now know how to avoid Apple's canvas context limit.

So I think that I can replace my patch with code that turns "var ctx = canvas.getContext('2d')" into a loop like the following:

var ctx = null;
while ((ctx = canvas.getContext('2d')) == null) {
    // Iterate through this._tiles, find the first where tile.el is an HTMLCanvasElement,
    // and convert that HTMLCanvasElement to an image. In theory, only one tile needs
    // be converted for canvas.getContext('2d') to succeed since all of the tiles are the
    // same size.
}

After I implement the above, I will have to do something similar for the large non-tile canvases elsewhere in the code that exceed the maximum width*height. I am not sure if it is better to clip those window-sized canvases if getContext('2d') returns null or if we should scale to the smaller canvas. Neither option is likely ideal.

plubius:
> src/layer/tile/WriterTileLayer.js: delete this._tileCache[key];
> 
> IIRC you had a magic to improve real canvas deletion (?) and also (quite probably) we should do a better job about size / managing our _tileCache there so it doesn't grow indefinitely - that would be awesome =)

Good point. I'll add the following magic commands before the delete. For Apple's browsers, setting the size of the canvas to zero appears to reduce the limit count immediately:

if (this._tileCache[key].el  && this._tileCache[key].el instance of HTMLCanvasElement) {
    this._tileCache[key].el.width = 0;
    this._tileCache[key].el.height = 0;
}
delete this._tileCache[key];

plubius:
After using my updated patch for a bit, the "Total canvas memory use exceeds the maximum limit (224 MB)." message only appears a few times during the course of loading every slide.

This is good news as it means that only a few tiles had to be converted from canvases to images and nearly all canvases remain canvases.

Now I just need to add code to handle the non-title canvases.

plubius:
OK. I think this pull request is ready for merging.

PR #6214, opened 2023-04-21
Consistency checking at various points: load, save.

mmeeks:
Check that everything is where it should be; catch and warn about jails that are corrupted beyond our control, or perhaps manually un-mounted etc.

Change-Id: Ib5f7769ded119d7c6126984b040d0e5565ca8055

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -146,6 +146,8 @@ static std::string JailRoot;
 static void flushTraceEventRecordings();
 #endif
 
+
+
 // Abnormally we get LOK events from another thread, which must be
 // push safely into our main poll loop to process to keep all
 // socket buffer & event processing in a single, thread.
@@ -157,6 +159,36 @@ static LokHookFunction2* initFunction = nullptr;
 
 namespace
 {
+    // for later consistency checking.
+    static std::string UserDirPath;
+    static std::string InstDirPath;
+
+    std::string pathFromFileURL(const std::string &uri)
+    {
+        std::string decoded;
+        Poco::URI::decode(uri, decoded);
+        if (decoded.rfind("file://", 0) != 0)
+        {
+            LOG_ERR("Asked to load a very unusual file path: '" << uri << "' -> '" << decoded << "'");
+            return std::string();
+        }
+        return decoded.substr(7);
+    }
+
+    void consistencyCheckFileExists(const std::string &uri)
+    {
+        std::string path = pathFromFileURL(uri);
+        if (path.empty())
+            return;
+        FileUtil::Stat stat(path);
+        if (!stat.good() && stat.isFile())
+            LOG_ERR("Fatal system error: created file passed into document doesn't exist: '" << path << "'");
+        else
+            LOG_TRC("File path '" << path << "' exists of length " << stat.size());
+
+        consistencyCheckJail();
+    }
+
 #ifndef BUILDING_TESTS
     enum class LinkOrCopyType
     {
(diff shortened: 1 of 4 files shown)


kit/Kit.cpp
```diff
@@ -157,6 +159,36 @@ static LokHookFunction2* initFunction = nullptr;
...
+    std::string pathFromFileURL(const std::string &uri)
+    {
+        std::string decoded;
+        Poco::URI::decode(uri, decoded);
+        if (decoded.rfind("file://", 0) != 0)
+        {
+            LOG_ERR("Asked to load a very unusual file path: '" << uri << "' -> '" << decoded << "'");
+            return std::string();
+        }
+        return decoded.substr(6);
```
Ashod, line 175: Not 7? "file://" has 7 characters.

kit/Kit.cpp
```diff
@@ -157,6 +159,36 @@ static LokHookFunction2* initFunction = nullptr;
...
+        std::string path = pathFromFileURL(uri);
+        if (path.empty())
+            return;
+        FileUtil::Stat stat(path);
+        if (!stat.good())
+        {
+            LOG_ERR("Fatal system error: created file passed into document doesn't exist: '" << path << "'");
+        }
+        consistencyCheckJail();
+        LOG_TRC("File path '" << path << "' exists of length " << stat.size());
```
Ashod, line 189: Perhaps this needs to be under `else`, otherwise in case of `!stat.good()` we'd report that it both does and doesn't exists.

kit/Kit.cpp
```diff
@@ -3126,6 +3163,33 @@ void runKitLoopInAThread()
...
+
+        FileUtil::Stat lo(InstDirPath + "/unorc");
+        if ((failedLo = (!lo.good() || !lo.isFile())))
+            LOG_ERR("Fatal system error: Kit jail is missing its LibreOfficeKit directory at '" << InstDirPath << "'");
+
+        FileUtil::Stat user(UserDirPath);
+        if ((failedUser = (!user.good() || !user.isDirectory())))
+            LOG_ERR("Fatal system error: Kit jail is missing its user directory at '" << UserDirPath << "'");
+
+        if (failedTmp || failedLo || failedUser)
```
Ashod, line 3184: Also return, otherwise we'd report an error and then say "Passed system consistency check".

Before returning, this is where we'd want to have `warned = true;`.

kit/Kit.cpp
```diff
@@ -3126,6 +3163,33 @@ void runKitLoopInAThread()
 
 #endif // !BUILDING_TESTS
 
+void consistencyCheckJail()
+{
+    static size_t warned = 0;
+    if (!(warned++))
```
Ashod, line 3169: Don't we want to set this to true (or ++) only when we do actually detect inconsistency (and report) ?

Meaning, we do want to check as long as we don't see any issues, but then disable the check from that point on.

Ashod (commented):
Thanks @mmeeks. Good stuff. Left a couple of notes.

mmeeks:
Thanks for the feedback; all resolved.

kit/Kit.cpp
```diff
@@ -157,6 +159,36 @@ static LokHookFunction2* initFunction = nullptr;
...
 {
+    // for later consistency checking.
+    static std::string UserDirPath;
+    static std::string InstDirPath;
+
+    std::string pathFromFileURL(const std::string &uri)
+    {
+        std::string decoded;
+        Poco::URI::decode(uri, decoded);
+        if (decoded.rfind("file://", 0) != 0)
```
Ashod, line 170: I think this should be compared to `std::string::npos`, otherwise, this will not work for uri that _starts_ with "file://".

Ashod (approved):
Thanks @mmeeks, looks good to me.

PR #6225, opened 2023-04-24
Jail cleanup - include coolwsd pid + hash into a jails sub-directory.

mmeeks:
This avoids removing the jails of other running coolwsd that share the same jails/ directory, such as unit-tests, cypress tests etc.

Assert on fatal system error to help catch this during test builds.

Remove old unit-test specific approach, and generalize it, since apparently we did a recursive cleanup of other people's jails anyway.

Remove over-complicated recursive approach, for a rather simpler one-level scheme. Avoid following symlinks too for safety.

Remove obsolete lo_jail_subpath setting.

Change-Id: I371f8c0e22f64fb2befb4b58f72cefa39567f3b5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -149,9 +149,18 @@ static bool safeRemoveDir(const std::string& path)
     return true;
 }
 
-void removeJail(const std::string& root)
+void removeAuxFolders(const std::string &root)
 {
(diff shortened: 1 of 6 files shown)


mmeeks:
CI bits are unrelated Cypress foo ...

PR #6242, opened 2023-04-27
admin: added a seprate state for the uploaded document

Rash419:
- previously we set the Modified state in admin panel only when wsd successfully upload the document to storage
- modified state updation depends on core, it might happen that uploading is already done and successfull but we didnot get the modified state from the core which will show wrong modified status in admin panel

Change-Id: I014a8f92753fc6a93b37921d0f3cdce390bef35e

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/admin/src/AdminSocketOverview.js
+++ b/browser/admin/src/AdminSocketOverview.js
@@ -155,6 +155,16 @@ function upsertDocsTable(doc, sName, socket, wopiHost) {
 	if (add === true) { row.appendChild(isModifiedCell); } else { row.cells[0] = isModifiedCell; }
 	isModifiedCell.className = 'has-text-centered';
 
+	var isUploadedCell = document.createElement('td');
+	isUploadedCell.id = 'up' + doc['pid'];
+	isUploadedCell.innerText = doc['uploaded'];
+	if (add === true) {
+		row.appendChild(isUploadedCell);
+	} else {
+		row.cells[0] = isUploadedCell;
+	}
+	isUploadedCell.className = 'has-text-centered';
+
 	// TODO: Is activeViews always the same with viewer count? We will hide this for now. If they are not same, this will be added to Users column like: 1/2 active/user(s).
 	if (add === true) {
 		var viewsCell = document.createElement('td');
@@ -221,7 +231,7 @@ var AdminSocketOverview = AdminSocketBase.extend({
 		this.base.call(this);
 
 		this.socket.send('documents');
-		this.socket.send('subscribe adddoc rmdoc resetidle propchange modifications');
+		this.socket.send('subscribe adddoc rmdoc resetidle propchange modifications uploaded');
 
 		this._getBasicStats();
 		var socketOverview = this;
@@ -287,6 +297,7 @@ var AdminSocketOverview = AdminSocketBase.extend({
 				'elapsedTime': '0',
 				'idleTime': '0',
 				'modified': 'No',
+				'uploaded': 'Yes',
 				'views': [{ 'sessionid': docProps[2], 'userName': decodeURI(docProps[3]) }]
 			};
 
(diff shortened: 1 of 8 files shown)


browser/admin/src/AdminSocketOverview.js
```diff
@@ -379,6 +386,15 @@ var AdminSocketOverview = AdminSocketBase.extend({
 			var $mod = $(document.getElementById('mod' + sPid));
 			$mod.text(value);
 		}
+		else if (textMsg.startsWith('uploaded')) {
+			textMsg = textMsg.substring('uploaded'.length);
+			docProps = textMsg.trim().split(' ');
+			sPid = docProps[0];
```
Ashod, line 392: Should we check that `docProps` actually has length 2?

browser/admin/src/AdminSocketOverview.js
```diff
@@ -379,6 +386,15 @@ var AdminSocketOverview = AdminSocketBase.extend({
...
+			var value = docProps[1];
+
+			var $up = $(document.getElementById('up' + sPid));
+			$up.text(value);
```
Ashod, line 396: Surely we should check that `$up !== undefined`. Also, the `$` seems a typo, as we don't use '$' in variable names.

wsd/DocumentBroker.cpp
```diff
@@ -1726,13 +1726,7 @@ void DocumentBroker::handleUploadToStorageResponse(const StorageBase::UploadResu
...
 
 #if !MOBILEAPP
-    if (lastUploadSuccessful && !isModified())
-    {
-        // Flag the document as un-modified in the admin console.
-        // But only when we have uploaded successfully and the document
-        // is current not flagged as modified by Core.
-        Admin::instance().modificationAlert(_docKey, getPid(), false);
-    }
+    Admin::instance().uploadedAlert(_docKey, getPid(), lastUploadSuccessful);
```
Ashod, line 1729: I think this should be `lastUploadSuccessful && !isModified()`.

Because if by the time we finish uploading there is further modification to the document (i.e. `isModified() == true`), we will set uploaded here to `true`, which is wrong (since there is a newer version to be uploaded). And since the `DocumentBroker::setModified` is already called with `true` argument, we will have the uploaded state wrong.

Instead, if here we use `lastUploadSuccessful && !isModified()`, then in that case the uploaded state will be correct and when the user does save and upload, it will get updated here when and if by the time we upload the document is not marked as modified.

However, there is a race condition when the `modifiedstatus=false` arrives late from Core. For that case, see my notes in `DocumentBroker::setModified()`.

wsd/DocumentBroker.cpp
```diff
@@ -3373,11 +3367,13 @@ std::size_t DocumentBroker::countActiveSessions() const
...
+    // Flag the document as modified in the admin console.
+    Admin::instance().modificationAlert(_docKey, getPid(), value);
     if (value)
     {
-        // Flag the document as modified in the admin console.
-        // But only flag it as unmodified when we do upload it.
-        Admin::instance().modificationAlert(_docKey, getPid(), value);
+        // Flag the document uploaded status to false in admin panel,
+        // as the document got modified
+        Admin::instance().uploadedAlert(_docKey, getPid(), !value);
```
Ashod, line 3376: We are setting the uploaded state _only_ when the document is modified. But we should consider the race condition that exists between uploading and getting the `modifiedstatus=false` from Core.

Since we can't control when we will get the `modifiedstatus`, we need to simply update the uploaded state when we do, in a generic way.

That is, we shouldn't call `uploadedAlert` only when modifiedstatus=true, instead, we should call this and pass the actual state of the upload, which is given by `needToUploadToStorage()`. But then there might be an upload in flight already, so we need to take that into account too.

Calling this _unconditionally_ here should work:
```
    Admin::instance().uploadedAlert(
        _docKey, getPid(), !isAsyncUploading() && needToUploadToStorage() == NeedToUpload::No);
```

Ashod (commented):
Thanks @Rash419. Good stuff. I added some comments and suggestions.

This is inherently a tricky part, so please try to challenge my suggestions and find cases that we might have missed.

wsd/DocumentBroker.cpp
Rash419: Thanks, now the whole picture got cleared to me why my approach was wrong

wsd/DocumentBroker.cpp
Rash419: thanks

wsd/DocumentBroker.cpp
Rash419: I have one question wouldn't it be confusing when admin sees Modified=No but Uploaded=Yes
as Uploaded is true because it updates the status based on `needToUploadToStorage() -> isStorageOutdated()`

Ashod (approved):
Thanks @Rash419. Looks good to me.

PR #6245, opened 2023-04-27
dark-mode use correct jquery-ui icons with data-theme value

Andreas-Kainz:
dark theme was switched with [data-theme='dark'] instead of prefer-color-scheme so the new setting respect [data-theme='dark'].

#### dropdown icon color is fixed

Screenshot_20230427_230607

Screenshot_20230427_230549

Change-Id: I9214fda63f6caea0a36ab8be7c26baf4e1c0f46e

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -593,11 +593,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 .ui-listbox::-ms-expand {
 	display: none;
 }
-@media (prefers-color-scheme: dark) {
-	.ui-listbox-arrow {
-		background: url('images/jquery-ui-lightness/ui-icons_ffffff_256x240.png') no-repeat transparent !important;
-	}
-}
+
 .ui-listbox-arrow {
 	content: '';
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
- There are too many curly brackets and spaces (not valid css). Simply ```[data-theme='dark'] .ui-listbox-arrow {```

Nice to have if you can in an additional commit `css: refactor ui-list-box-arrow on dark theme`
- This works as the additional data-theme value is one extra level of specificity thus overriding the generic '''.ui-listbox-arrow {```
  - but please place it after that  so it's easier to understand the priorities when reading the code

Andreas-Kainz:
I'm not 100% sure what you mean, but it work and the css was cleaned up.

pedropintosilva:
your first commit was:
```css
[data-theme='dark'] {
	.ui-listbox-arrow {
		background: url('images/jquery-ui-lightness/ui-icons_ffffff_256x240.png') no-repeat transparent;
   }
}
```

but now after your force-pushed it's ok:
```css
[data-theme='dark'] .ui-listbox-arrow {
	background: url('images/jquery-ui-lightness/ui-icons_ffffff_256x240.png') no-repeat transparent;
}
```

pedropintosilva (approved):

Issue #6236, opened 2023-04-26
Tabbed view : Dropdown arrows are misplaced


pedropintosilva:
**Describe the bug**

![image](https://user-images.githubusercontent.com/65948705/234613772-648fbf8f-f9cf-415d-b61a-a5b4bb983482.png)

They should be placed in paralleled with the icons as seen in 22.05

**Versions**
| Where | branches |
| -- | -- |
| Online | master |
| Core | distro/collabora/co-23.05 |

GulsahKose:
This commit  be04111173dfa85f7f20cc0eb0e5babbe220a665 caused the regression.

thebearon:
@pedropintosilva This seems fine to me, is anything left to do here?

timur-g:
Seems good to me, not clear if there was something else here to do with icons. I close.

PR #6250, opened 2023-04-28
Fix down arrow/triangle position in sidebar dropdown icons

pedropintosilva:
before this commit the triangle was misplaced, under the icon instead
of side by side

(might be related to https://github.com/CollaboraOnline/online/issues/6236)

Change-Id: Id8d118471dcb02b5bc79399a4319ee514af03c5b


GulsahKose (approved):
Sidebar seems nice with that patch. Thanks Pedro.

PR #6256, opened 2023-04-28
Dark Mode: box-shadow tooltip fix

Andreas-Kainz:
tooltip use var colors.

**Before**
Screenshot_20230428_224357

**After**
Screenshot_20230428_230408

**After light mode**
Screenshot_20230428_230424

Change-Id: Ic8addda2167be038d7e435d424448824be73127a

### DIFF
--- a/browser/css/jquery-ui-lightness.css
+++ b/browser/css/jquery-ui-lightness.css
@@ -896,10 +896,10 @@ button.ui-button::-moz-focus-inner {
 	font-size: 1em;
 }
 .ui-widget.ui-widget-content {
-	border: 1px solid var(--color-border-dark);
+	border: 1px solid var(--color-border-lighter);
 }
 .ui-widget-content {
-	border: 1px solid #dddddd;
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
Thanks but it would be better if a shadow is not light then the background bellow otherwise it wouldn't be a shadow ; )

Andreas-Kainz:
good point, but in dark-mode the background color is #000 so darken than black isn't possible.

pedropintosilva:
right so the way I see it we have two options:

1. make sure the background is not pure black and reserve that for shadows 
2. Or remove shadow altogether from those
  * The tooltip's background is lighter so maybe we don't need nothing else. But if you find out that the contrast is still a bit poor (difficult to distinguish where the tooltip ends and where the bellow elements start) then we can add a light border to the tooltips

pedropintosilva:
I'm merging this as it is already an improvement. Thanks! I will then push a follow up (dark mode only)

pedropintosilva (approved):

PR #6257, opened 2023-04-28
Dark Mode Dialog css fixes

Andreas-Kainz:
input elements use css var's

| before | after |
| --------- | -------- |
| Screenshot_20230428_231244 | Screenshot_20230428_234537 |
| Screenshot_20230428_234608 | Screenshot_20230428_231314

Change-Id: Ie2ebe22bcf7fc377894f4b43c1bb081a3d5f5c16

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -2,9 +2,8 @@
 
 .jsdialog *[disabled] {
 	color: var(--color-text-lighter);
-	border: 1px solid var(--color-btn-border-dis);
-	filter: grayscale(1);
-	opacity: 0.7;
+	border: 1px solid var(--color-border-lighter);
+	background: var(--color-background-lighter);
 }
 
 /*limit icon to button height*/
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -64,12 +64,13 @@ button:not(.ui-tab):not(.ui-corner-all):not(.button-primary):not(.unobutton) {
 
 .ui-pushbutton[disabled] {
 	color: var(--color-text-lighter) !important;
-	background-color: white !important;
+	background-color: var(--color-background-lighter) !important;
+	border: 1px solid var(--color-border-lighter) !important;
 }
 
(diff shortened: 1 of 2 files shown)


Andreas-Kainz:
| before | after |
| ---------- | ------- |
| Screenshot_20230429_000957 | Screenshot_20230429_002221 |
| Screenshot_20230429_001603 | Screenshot_20230429_002209 |

Andreas-Kainz:
| before | after |
| --------- | -------- |
| Screenshot_20230429_004131 | Screenshot_20230429_004010 |

browser/css/btns.css
```diff
@@ -64,12 +64,13 @@ button:not(.ui-tab):not(.ui-corner-all):not(.button-primary):not(.unobutton) {
...
 	color: var(--color-text-lighter) !important;
-	background-color: white !important;
+	background-color: var(--color-background-lighter) !important;
+	border: 1px solid var(--color-border-lighter) !important;
 }
 
 .ui-pushbutton[disabled]:hover {
 	cursor: not-allowed !important;
-	background: white !important;
+	background: var(--color-background-darker) !important;
```
pedropintosilva, line 73: Follow up (not a blocker for this PR) the background here( in .ui-pushbutton[disabled]:hover) is not bieng applied  due to rules from `.button-secondary:hover, button:not(.ui-corner-all):not(.button-primary):not(.unobutton):hover`

pedropintosilva (changes requested):
just please add the missing `color` to the var name and we can merge it : )

browser/css/btns.css
Andreas-Kainz: To be honest, I don't know that much about the css rules at device-desktop.css so I can't change there that much (without destroy something).

What I don't understood is, that there are so much rules with `!important` elements.

pedropintosilva (approved):
Thanks! (note:next time better to rebase and amend the existing initial commit instead)

PR #6258, opened 2023-04-28
Dark Mode add color-scheme: dark setting

Andreas-Kainz:
see https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

| before | after |
| -------- | ------- |
| Screenshot_20230428_235549 | Screenshot_20230428_235601 |

Change-Id: Ie563059174471d50b5f9a8e24d1f08b2540799c5


pedropintosilva (approved):
beautiful, thanks!

PR #6261, opened 2023-04-30
jsdialog ui-widget-header border-radius and color fix

Andreas-Kainz:
ui-widget-header has no border-bottom-radius
ui-widget-content has no border-top-radius
cause both widgets are next to each other

ui-icon use the blue hightlight colors
and the blue icons for more consistancy

Screenshot_20230430_223758

Change-Id: Id891548c56b540f8d0b27f65b805c1823fe28906

### DIFF
--- a/browser/css/jquery-ui-lightness.css
+++ b/browser/css/jquery-ui-lightness.css
@@ -972,10 +972,9 @@ a:visited.ui-button,
 .ui-widget-header .ui-state-focus,
 .ui-button:hover,
 .ui-button:focus {
-	border: 1px solid #fbcb09;
-	background: #fdf5ce url("images/ui-bg_glass_100_fdf5ce_1x400.png") 50% 50% repeat-x;
+	border: 1px solid var(--color-primary-dark);
+	background: var(--color-primary) url("images/ui-bg_glass_100_fdf5ce_1x400.png") 50% 50% repeat-x;
 	font-weight: bold;
-	color: #c77405;
 }
 .ui-state-hover a,
 .ui-state-hover a:hover,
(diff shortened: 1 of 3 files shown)


Andreas-Kainz:
@pedropintosilva if you have time for a review, would be a pleasure and not to complicated.

browser/css/jquery-ui-lightness.css
```diff
@@ -1108,9 +1104,6 @@ a.ui-button:active,
 .ui-state-error-text .ui-icon {
 	background-image: url("images/ui-icons_ffd27a_256x240.png");
 }
-.ui-button .ui-icon {
```
Andreas-Kainz, line 1111: No need for different background-images. One background-image (black/white) and the other color stuff was done with  hover effect.

pedropintosilva (approved):

rizmut:
Somehow I see those tabs as individual button spread out.

Andreas-Kainz:
For the tabs I have to make a separate PR. I don't like them they may look like Firefox tabs, but not like real tabs.

Andreas-Kainz:
Thanks for the feedback

rizmut:
Personally in the start I don't like Firefox tab as iy mimick to much a button which breaks the logical connection between them. Now I can accept that, but these one make it worst. I see latest macOS approach feel more natural as all the buttonly tab are wrapped together by a beveled rounded rectangle. So it keeps the relation.

Andreas-Kainz:
At notebookbar there are tabs and the dialogue tabs should look the same. If someone will change the tab layout at notebookbar, the dialogue tabs layout should change too.

Issue #6260, opened 2023-04-30
jsdialog-tabs


Andreas-Kainz:
What's the reason that `jsdialog-tabs` didn't located the `ui-tabs` `tablist`?

![Screenshot_20230430_223254](https://user-images.githubusercontent.com/8517736/235375130-0b380561-8c7d-4707-8a0b-0e28400bebd5.png)

Andreas-Kainz:
@pedropintosilva or someone else?

pedropintosilva:
I'm also intrigued, I have checked notebookbar tabs and multiple dialogs with tabs and `jsdialog-tabs` is always empty and set to display none. Do we really need it if we don't otherwise use it? cc: @gokaysatir

Andreas-Kainz:
I would like to use it for the dialogue tabs, so we can have the same layout for the dialogue tabs and the notebookbar tabs.

eszkadev:
This is leftover not deleted after recent rework (tabs inside tabs). In notebookbar we need it (tabs placed in some dedicated place). But in dialogs we put tab list where tab control is placed, not in that special container anymore.

PR #6262, opened 2023-05-01
jsdialog: don't create empty tabs container #6260

eszkadev:
In dialogs we put tab list directly where tab control is placed. We don't need special container anymore (setTabs is not used).


eszkadev:
3 failed cypress tests are passed locally

pedropintosilva (approved):

PR #6272, opened 2023-05-03
browser: Missing fonts dialog only worked if welcome dlg was enabled

thebearon:
Change-Id: Idd0d7a5d74e5f4b0ef82d12877a46e55c172a992

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -980,7 +980,7 @@ app.definitions.Socket = L.Class.extend({
 				msg += fontsMissingObj.fontsmissing[i];
 			}
 
-			if (!this._map.welcome.isGuest() && this._map.welcome.shouldWelcome() && window.autoShowWelcome)
+			if (this._map.welcome && !this._map.welcome.isGuest() && this._map.welcome.shouldWelcome() && window.autoShowWelcome)
 			{
 				setTimeout(function() {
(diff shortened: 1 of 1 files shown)


browser/src/core/Socket.js
```diff
@@ -980,7 +980,7 @@ app.definitions.Socket = L.Class.extend({
 				msg += fontsMissingObj.fontsmissing[i];
 			}
 
-			if (!this._map.welcome.isGuest() && this._map.welcome.shouldWelcome() && window.autoShowWelcome)
+			if (this._map.welcome && !this._map.welcome.isGuest() && this._map.welcome.shouldWelcome() && window.autoShowWelcome)
```
Ashod, line 983: I'm confused. This only checked that `welcome` is defined. But that wouldn't change the behavior that the "fontsmissing" dialog will *still* only work if welcome dialog is enabled, or I missed something?

Was this supposed to be  `if (!this._map.welcome || (!this._map.welcome.isGuest() && this._map.welcome.shouldWelcome() && window.autoShowWelcome))` ?

thebearon: If you check the surrounding code, you will see that both paths show the missing fonts dialog, just in a different way, see below the following line:
https://github.com/CollaboraOnline/online/blob/7d2f7360843b82908dd6c405b6815d6ea38d5563/browser/src/core/Socket.js#L983

timar (approved):

PR #6280, opened 2023-05-04
[Backports] Impress: Fix imperceptible active (current) slide border

pedropintosilva:
- Make it bigger
- Also size is not enough, specially for slides that have backgrounds
  - Add white inner margin so there is a clear distinction between
- what is the slide preview (image) and its contour

Change-Id: Ifee386b98d74c937efa35c19551334482ec18632


pedropintosilva:
note: make -C cypress_test run-desktop spec=impress/slide_operations_spec.js passes

Issue #6275, opened 2023-05-03
Tabbed view: tabs lose format and style when modalpopup is opened


pedropintosilva:
**Describe the bug**

**To Reproduce**
1. Help tab
2. Click on Online Help, or Keyboard shortcuts, 
3. See error in the tabs spacing and position

**Screenshots**
![image](https://user-images.githubusercontent.com/65948705/235932723-cfa30637-b73f-421a-a7e9-5ac22c8a51a4.png)

PR #6282, opened 2023-05-04
Make dialog button style to specific that dialog #6275

GulsahKose:
In tabbed view when we open help or keyboard shortcuts dialog "button" style of cool-help.html overrides the button.ui-tab.notebookbar. So all buttons (even in notebook bar) uses dialog buttons' text-decoration: underline and padding:0

Here we make the dialog's button style specific to related dialogs.

Change-Id: Iacaab42efa527fc5c6ada21fd007e4a352912b60

* Resolves: #6275 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the fix @GulsahKose :)

PR #6294, opened 2023-05-06
wsd: http: killpoco in prisoner http handler

Ashod:
We use our own status codes, which
are type-safe and use streaming
operators to serialize and log.


Ashod:
Duplicate.

PR #6295, opened 2023-05-07
Misc and Logging

Ashod:
- wsd: move streaming operators to the global space
- wsd: simplify log streaming
- wsd: http: native support for logging http::Header
- wsd: move streaming operator overloads
- wsd: minor logging cleanup
- wsd: make unreusable classes final
- wsd: use helper where available


JaumePujantell (approved):
Looks good to me!

PR #6302, opened 2023-05-08
Fix white screen over document on dark mode

OFFTKP:
Upon switching to dark mode the document would disappear and replaced with a white screen. The above change fixes that. My assumption is that for some reason it would silently crash and display nothing.

Change-Id: Ia59f1a4c10b4198243a017ba9cc1cbeff50619d0

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -1,5 +1,5 @@
 [data-theme='dark'] {
-	color-scheme: dark;
+	--color-scheme: dark;
 	/*LibreOffice Colors: https://wiki.documentfoundation.org/Marketing/Branding#Colors
 	----------------------------------[to do]*/
 	--blue1-txt-primary-color: 3, 105, 163;


OFFTKP:
@Andreas-Kainz , this change was by you and it would make the document (leaflet) turn white when switching to dark mode. I am going to push this to fix the upstream and if it's not the correct fix for the reason you wanted it, you can change it in a follow up commit.

Andreas-Kainz:
see #6258

according to https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme I'm not sure, this change is related to your issue, cause color-scheme: dark is for scralbars, ...

OFFTKP:
image

This is what it looks like on firefox on my linux machine. Perhaps my firefox installation is broken? This doesn't happen on chromium for me. Could you test on firefox?

Andreas-Kainz:
You are right the dark mode on firefox didn't work well. Let's revert the origin PR #6258

OFFTKP:
Closing this PR, opened #6309

PR #6304, opened 2023-05-09
Remove scrollbar button in mobile view

Darshan-upadhyay1110:
Change-Id: Icec596bc4691b5bdd04ab2c687cc8c0c54608412

* Resolves: # 
* Target version: master 

### Summary

Issue: 

- Occasionally a button appears on top right, which does not belong to the default control elements. It looks like a rewind button in a player.On tap nothing will happen and the button disappears.

Fixed this issue which now not showing any scroll button when document loads in android app. I have tested.

@pedropintosilva  can you  please review the changes.

### DIFF
--- a/browser/js/w2ui-1.5.rc1.js
+++ b/browser/js/w2ui-1.5.rc1.js
@@ -6197,6 +6197,10 @@ w2utils.event = {
                 }
             }
 
+            if (window.mode.isMobile()) {
+                $('#toolbar-wrapper.mobile .w2ui-scroll-right, #toolbar-wrapper.mobile .w2ui-scroll-left').hide();
+            }
+
             // event after
             this.trigger($.extend(edata, { phase: 'after' }));
             return (new Date()).getTime() - time;


pedropintosilva (approved):
Thanks @Darshan-upadhyay1110 !
Unfortunately I think we can not do this for tablets other wise the following blue element (that has that class) would be hidden on tablets:
image
 **Plus** we also use these elements in the bottom toolbar on mobile. So, this would also hide always the right scrollbar in top and **bottom** bar. Better To do something different and easier.

Example:
We are already setting display rules in such elements within device-mobile.css so the easiest would be to hide there the scroll indicators (both left and right) and just for for toolbar `#toolbar-wrapper`

pedropintosilva (changes requested):
See my comment ^

pedropintosilva (approved):
thanks @Darshan-upadhyay1110 !

pedropintosilva:
I see 2 commits one changing things from the first one.

Please rebase /squash it so we have only one commit here and then we can merge it

pedropintosilva (approved):
Thanks

browser/js/w2ui-1.5.rc1.js
```diff
@@ -6197,6 +6197,10 @@ w2utils.event = {
                 }
             }
 
+            if (window.mode.isMobile()) {
```
pedropintosilva, line 6200: Please remove the space after `{` it's appearing as a linting problem

pedropintosilva (changes requested):

browser/js/w2ui-1.5.rc1.js
Darshan-upadhyay1110: okay i will do that

Darshan-upadhyay1110: Done and pushed to the existing commit.

pedropintosilva (approved):

PR #6305, opened 2023-05-09
jsdialog: css: wrap tabs in dialogs

eszkadev:
This commit allows to wrap tabs in dialogs so
if we have 10 of them we don't force large width
for the dialog. Result is visible in dialogs like
format -> character or format -> paragraph


eszkadev:
closed in favor https://github.com/CollaboraOnline/online/pull/6322

PR #6311, opened 2023-05-09
jsdialog: dialog adjusted to size of biggest tab

eszkadev:
this uses css grid to position all the tabs stacked on each other at row 1 col 1. then when tab is hidden we use 'visibility: hidden' instead of 'display: none' so we hide content but leave white space needed for it. this way we don't change dialog's size on tab switch when tab has different size requirement, but we match dialog to the size of biggest content


eszkadev:
closed in favor https://github.com/CollaboraOnline/online/pull/6322

Issue #6293, opened 2023-05-05
Overlay with color name of the gradient in chart sidebar looks misplaced


nikslor:
**Describe the bug**
The color name of a gradient in the chart sidebar looks misplaced in 23.05.x

![image](https://user-images.githubusercontent.com/1211300/236550133-040ca7cb-8067-484b-bab6-51de7f3632eb.png)

**To Reproduce**
1. Open a writer document
2. Add a chart
3. Edit the chart
4. Scroll to the "Area" part in the sidebar
5. Choose "Gradien" as "Fill:"
6. Hover over the color
7. See error

**Expected behavior**
The color name should not overlay the actual color.

**Actual behavior**
The color name overlays the actual color.

thebearon:
Confirmed, and it's the same in 22.05, the oldest I tested with is 22.05.3-1.
Looks like a bug specific to JS sidebar.

CC @eszkadev @pedropintosilva.

PR #6314, opened 2023-05-09
fixed overlay with color name of the gradient in chart sidebar 

Darshan-upadhyay1110:
fixed overlay with color name of the gradient in chart sidebar 
Change-Id: Ie3c98a908efedfd666cbafa22f21f907f947eb7a

* Resolves: #6293 
* Target version: master 

@pedropintosilva , @GulsahKose

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -130,12 +130,12 @@ div.sidebar.ui-grid .checkbutton.sidebar,
 	position: absolute;
 	display: block;
 	width: auto;
-	background-color: var(--color-background-darker);
+	background-color: transparent;
 	color: var(--color-text-darker);
-	padding: 2px 4px;
+	padding: 10px 4px;
 	border-radius: var(--border-radius);
 	white-space: nowrap;
-	margin-inline-start: -14px;
+	margin-inline-start: 2px;
 }
 
 .sidebar.ui-grid.ui-grid-cell .menubutton {


browser/css/jssidebar.css
```diff
@@ -130,12 +130,12 @@ div.sidebar.ui-grid .checkbutton.sidebar,
 	position: absolute;
 	display: block;
 	width: auto;
-	background-color: var(--color-background-darker);
+	background-color: transparent;
```
pedropintosilva, line 133: I think we should make sure the hover status for the menubutton has the same border color and background color as other components such as.jsdialog.ui-listbox:hover 

also `.menubutton.sidebar:hover span` should be improved  ideally to look like any other regular `.ui-tooltip`

pedropintosilva (changes requested):
let me push one commit before you do any further changes

pedropintosilva:
Thanks @Darshan-upadhyay1110 but there were other fixes that were needed before attempting to do this one, and some other PRs that changes some CSS values. So I ended up pushing https://github.com/CollaboraOnline/online/pull/6413/commits 

Could you please test that ^ ? I see there are now a regression WRT width of those components (but it is something unrelated and new introduced by a new width rule that is applied in the overall btn styles so it is not a blocker for my PR)

Darshan-upadhyay1110:
okay sure i will test that

Issue #6285, opened 2023-05-05
Texts in Page Number dialog needs some padding 


Ezinnem:
**Describe the bug**

**To Reproduce**
1. Open a new .odt file
2. Go to Insert > Page Number

Notice that the `Preview` text needs some padding

**Expected behavior**
![Screenshot from 2023-05-04 20-51-03](https://user-images.githubusercontent.com/54344301/236400749-0f04ba7f-2120-48d7-9935-483e2a6c6ce5.png)

**Actual behavior**
![Screenshot from 2023-05-04 20-50-01](https://user-images.githubusercontent.com/54344301/236402295-a63ed038-eb1e-4f2b-aa86-4169a3d446c5.png)

**Desktop (please complete the following information)**
 COOLWSD version:
23.05.0.1 git hash: [9de01e11](javascript:void(window.open('https://github.com/CollaboraOnline/online/commits/9de01e11'));)

Ezinnem:
@pedropintosilva @eszkadev

pedropintosilva:
I'm afraid the problem here is not the padding, but rather the `Preview` label is now misplaced (and then yes it both columns should have a gap)

GulsahKose:
Current window looks like this. Is is OK? 

![Screenshot from 2023-05-09 17-59-26](https://github.com/CollaboraOnline/online/assets/2954981/2c3369d2-f1b6-472b-98a0-4460236070b4)

PR #6316, opened 2023-05-09
Fix the layout of page number diaog

GulsahKose:
Change-Id: I00bd8d75a2e0ae509f78eda3ae1af687da088419

* Resolves: #6285
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -171,6 +171,14 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	margin: auto 0px;
 }
 
+/* Page number dialog placed under Insert menu */
+#PageNumberDialog #rightVBox,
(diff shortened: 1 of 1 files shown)


browser/css/jsdialogs.css
```diff
@@ -171,6 +171,12 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	margin: auto 0px;
 }
 
+#rightVBox.ui-grid-cell, #leftVBox.ui-grid-cell {
```
eszkadev, line 174: can we use some unique prefix with the id of the dialog so we don't apply that rule unintended?

#PageNumberDialog i think it is

eszkadev: and maybe a comment above for which dialog it is

eszkadev (approved):

PR #6322, opened 2023-05-10
jsdialog: fix jumping Page Style dialog

eszkadev:
Fixes position and dragging in Writer -> Format -> Page Style dialog

It was caused by `<button>` with id = style inside dialog's `<form>`. In that case style property was a reference to the dialog, not CSS properties map. It cannot be changed... so Introduced new parent container which is div for setting position.

This PR also makes dialogs more compact, using minimal width thanks to using 'position: absolute' inside parent container. What helps to not cover whole screen with dialog.


eszkadev:
dialog-position

gokaysatir:
Keyboard shortcuts and Online help dialogs start at the bottom of the screen. Other than that, looks beautiful.

PR #6324, opened 2023-05-10
Add date time fields under the insert menu

GulsahKose:
In Calc compact view we have insert current date and time actions but in tabbed view we don't have them. For equality we need to add in tabbed view too.

Change-Id: I5498f47a8dc96a5403cdf17be52ef3a7ad66862c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
Unrelated Fails, merging

PR #6325, opened 2023-05-10
Handle LOK_CALLBACK_DOCUMENT_PASSWORD_RESET

JaumePujantell:
Change-Id: Ibc2be5bc7676b1b67b7494c200011e4622a5b294


JaumePujantell:
This pull request requires https://gerrit.libreoffice.org/c/core/+/151651

kit/Kit.cpp
```diff
@@ -1152,6 +1152,17 @@ class Document final : public DocumentManagerInterface
             }
             return;
         }
+        else if (type == LOK_CALLBACK_DOCUMENT_PASSWORD_RESET) {
```
Ashod, line 1155: Please format using clang-format (easy to set vscode to apply it automatically for the modified lines when you save).

kit/Kit.cpp
```diff
@@ -1152,6 +1152,17 @@ class Document final : public DocumentManagerInterface
...
+            Document* document = dynamic_cast<Document*>(descriptor->getDoc());
+            if (document)
+            {
+                document->_isDocPasswordProtected = p ? !payload.empty() : false;
```
Ashod, line 1159: Isn't it simpler like this?
`document->_isDocPasswordProtected = p && !payload.empty();`

kit/Kit.cpp
```diff
@@ -1152,6 +1152,17 @@ class Document final : public DocumentManagerInterface
...
+                document->_haveDocPassword = document->_isDocPasswordProtected;
+                document->_docPassword = payload;
+                document->_docPasswordType = PasswordType::ToView;
```
Ashod, line 1162: Does this imply that we can't reset the password to modify a document?

(Perhaps another reason why reusing `LOK_CALLBACK_DOCUMENT_PASSWORD` and `LOK_CALLBACK_DOCUMENT_PASSWORD_TO_MODIFY` is a good idea.)

Ashod (commented):
Thanks @JaumePujantell. I left a couple of notes, and we will need to wait for the Core to pass the build.

My only major concern is that it seems we don't support resetting the password to modify docs, which is a serious limitation.

Ashod (commented):
Thanks @JaumePujantell. I left a couple of notes, and we will need to wait for the Core to pass the build.

My only major concern is that it seems we don't support resetting the password to modify docs, which is a serious limitation.

Ashod (approved):
Thanks @JaumePujantell. Looks good to me, but we should add a unit-test to defend the new API. We already have password tests, so we can extend them or add a new one to reset. Let me know if you need help, I can give you pointers.

Also, the CI needs a retrigger.

JaumePujantell:
> 

To test this, since the password is changed through jsdialogs, should I write in the test all the dialogevents that happen, or is a cleaner way to write the test?

thebearon:
This got merged as #6711. Following up on the test separately.

PR #6352, opened 2023-05-14
Add possibility to define help root URL from brand package

timar:
Previously the help root URL was defined in coolkitconfig.xcu, but it was inconvenient to change. Now we have this setting in coolwsd.xml, therefore it can be changed more easily. For example a brand package can disable the Help buttons on dialogs by executing the following command in postinstall script:
  coolconfig set help_url ""

Change-Id: I6d0bdd71ca908df3d2dd20bd321aff9e93896f52

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -2081,7 +2081,8 @@ void COOLWSD::innerInitialize(Application& self)
         { "deepl.auth_key", "" },
         { "deepl.enabled", "false" },
         { "zotero.enable", "true" },
-        { "indirection_endpoint.url", "" }
+        { "indirection_endpoint.url", "" },
+        { "help_url", HELP_URL }
     };
 
     // Set default values, in case they are missing from the config file.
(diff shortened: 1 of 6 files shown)


timar:
Corresponding core patch: https://gerrit.libreoffice.org/c/core/+/151739

wsd/COOLWSD.cpp
```diff
@@ -2547,6 +2548,15 @@ void COOLWSD::innerInitialize(Application& self)
     setenv("DEEPL_API_URL", apiURL.c_str(), 1);
     setenv("DEEPL_AUTH_KEY", authKey.c_str(), 1);
 
+#if !MOBILEAPP
+    const std::string helpUrl = getConfigValue<std::string>(conf, "help_url", HELP_URL);
+    setenv("LOK_HELP_URL", helpUrl.c_str(), 1);
```
Ashod, line 2553: Is there already logic that read LOK_HELP_URL envar and uses it? Is that in a separate PR?

Ashod (approved):
Thanks @timar. Looks good to me.

wsd/COOLWSD.cpp
timar: yes, see https://gerrit.libreoffice.org/c/core/+/151739

PR #6357, opened 2023-05-15
wsd: quarantine: thread-safe quarantining

Ashod:
The QuarantineMap is shared between DocBroker
instances, which makes it subject to concurrent
access and modification. As such, it must be
protected via a lock.

Change-Id: I2e884ec54a843b17b077c84ecf07656ad8af752a


mmeeks (approved):
Thanks Ash =)

PR #6362, opened 2023-05-16
Private/gokay/document conflict

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/core/Socket.js
```diff
@@ -953,6 +954,7 @@ app.definitions.Socket = L.Class.extend({
 							if (value === 'discard') {
 								// They want to refresh the page and load document again for all
 								that.sendMessage('closedocument');
+								window.location.reload();
```
Ashod, line 957: This should be fine, although I think it worked previously simply because we always reloaded when we got disconnected, and the 'closedocument' will disconnect us from WSD. At any rate, the `reload()` call should be fine, I expect.

browser/src/core/Socket.js
```diff
@@ -797,21 +797,22 @@ app.definitions.Socket = L.Class.extend({
 				}, 3000);
 			}
 
-			// Close any open dialogs first.
-			vex.closeAll();
+			if (textMsg === 'idle' || textMsg === 'oom') {
```
Ashod, line 800: I don't understand why we need this change, and how it's related to the document conflict case.

Ashod (approved):
Thanks @gokaysatir. Looks good, but not sure I understand the 'idle' and 'oom' changes.

browser/src/core/Socket.js
gokaysatir: It causes to close open vex window(s) even when a vex window will not open. I modified it while testing the "close: documentconflict" case.

browser/src/core/Socket.js
gokaysatir: Thanks for checking :)

PR #6367, opened 2023-05-16
wsd: better detection of modification during a conflict

Ashod:
When we detect document change in storage, we have to
either prompt the user(s) to either reload from storage,
discarding their own changes, or force saving their
changes and clobber the version in storage.

However this implies that the loaded document was
modified. Because otherwise, we do reload the
document from storage, automatically.
This change improves the detection logic to take
into account not just the saved state of the
document, but also the upload state. That is,
if a previous modification was not uploaded yet,
then the user does have changes that they need to
consider to discard or save.

Change-Id: I5f03593bebc5b565fc19e78562896bcdcb6112e2


mmeeks:
Seems fine to me =)

PR #6374, opened 2023-05-16
fix wrong cell background

hcvcastro:
- browser: reset last selected color
- browser: fix color not a number
- browser: check a valid color


eszkadev:
it didn't fix the issue, background color of a cell is not visible (it's always white)

PR #6358, opened 2023-05-15
Fix debug build breakage due to undefined function on iOS

plubius:

PR #6375, opened 2023-05-16
Fix iOS build breakage due to Poco replacement changes

plubius:
iOS (and maybe Android) do not compile or use the HttpRequest class. The socket in these apps is between two threads in the same process, not a web browser and a web server so no HTTP headers are passed over the socket.

From what I can see, on iOS this call only gets called once when a document is opened and the socket is empty. Which makes sense to me since on iOS we use WKWebView's special JavaScript streams to send messages to the server code.

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3463,6 +3463,7 @@ private:
                 << Util::dumpHex(std::string(data.data(), std::min(data.size(), 256UL))));
 #endif
 
+#if !MOBILEAPP
         // Consume the incoming data by parsing and processing the body.
         http::Request request;
         const int64_t read = request.readData(data.data(), data.size());
@@ -3482,6 +3483,9 @@ private:
 
         // Remove consumed data.
         data.eraseFirst(read);
+#else
+        Poco::Net::HTTPRequest request;
+#endif
 
         try
         {
@@ -3551,7 +3555,6 @@ private:
 #else
             pid_t pid = 100;
             std::string jailId = "jail";
-            LOG_ASSERT_MSG(socket->getInBuffer().empty(), "Unexpected data in prisoner socket");
             socket->getInBuffer().clear();
 #endif
             LOG_TRC("Calling make_shared<ChildProcess>, for NewChildren?");
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -1595,6 +1595,22 @@ class CanvasSectionContainer {
 		this.canvas.width = newWidth;
 		this.canvas.height = newHeight;
 
+		// Partial fix for #5876 reduce size if canvas context is null
+		// WKWebView has a hard limit on the number of bytes of canvas
+		// context memory that can be allocated. Canvas with sizes that
+		// are too large will return a null context so reduce the size
+		// of the canvas until a non-null context is obtained.
+		if ((window as any).ThisIsTheiOSApp) {
+			this.context = this.canvas.getContext('2d', { alpha: false });
+			while (!this.context && this.canvas.width >= 2 && this.canvas.height >= 2) {
+				this.canvas.width /= 2;
+				this.canvas.height /= 2;
+				this.context = this.canvas.getContext('2d', { alpha: false });
+			}
+			for (var i: number = 0; i < this.sections.length; i++)
+				this.sections[i].context = this.context;
+		}
+
 		// CSS pixels can be fractional, but need to round to the same real pixels
 		var cssWidth: number = newWidth / app.dpiScale; // NB. beware
 		var cssHeight: number = newHeight / app.dpiScale;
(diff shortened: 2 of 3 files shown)


plubius:
This is now ready for review. The first patch fixes building and running on iOS without undoing the Poco replacement changes for Collabora Online.

The second patch is a port of the second patch in pull request #6358 from the distro/collabora/co-22-05 branch.

wsd/COOLWSD.cpp
```diff
@@ -3482,6 +3483,9 @@ class PrisonerRequestDispatcher final : public WebSocketHandler
 
         // Remove consumed data.
         data.eraseFirst(read);
+#else
+        Poco::Net::HTTPRequest request;
```
Ashod, line 3487: Why not have the `http::Request request;` on line 3468 outside the `#if !MOBILEAPP` directive, two lines above, and avoid this? Why not have `request` of type `http::Request`? (We are reducing Poco dependency.)

wsd/COOLWSD.cpp
```diff
@@ -3551,7 +3555,6 @@ class PrisonerRequestDispatcher final : public WebSocketHandler
 #else
             pid_t pid = 100;
             std::string jailId = "jail";
-            LOG_ASSERT_MSG(socket->getInBuffer().empty(), "Unexpected data in prisoner socket");
```
Ashod, line 3554: What was wrong with this assertion? We don't expect residue data in the socket, regardless of whether there was data that we consumed, or it was always empty. What am I missing?

Ashod (commented):
Thanks @plubius. Left a couple of comments.

plubius:
This pull request is way out of date at this point so I am closing it. It is pretty clear to me that the iOS and Android code needs to just bypass all this header parsing and jail management code since no HTTP headers are passed over the socket on iOS and Android. After all, the "socket" isn't really a socket on iOS or Android,, it is just a buffer shared among threads since there is no separate server process.

Realistically, we are going to keep running into this every time the Online server code is changed because apparently very few people understand that the message processing architecture for iOS and Android is different than Online.

plubius:
Note: when you are finished removing Poco from the Online master branch, let me know. When I get time I'll see if I can come up with a new pull request that bypasses all of the server handling code. As I noted above, iOS (and maybe Android) do not compile or use the HttpRequest class since there are no headers passed across the socket and that would likely confuse and/or break any header handling logic in the HttpRequest class.

I suspect that iOS and Android need a small number of snippets of the code so an empty Poco request object plus all the #if !MOBILE has been used to get to those few snippets. Most likely, I'll just copy whatever snippets are needed elsewhere and bypass compiling this server file completely.

mmeeks:
Ooh - that's a really good idea =) I've been trying to write out Poco for a long time - but using the mobile apps as a Poco-lite version that doesn't require the HTTP parsing stuff is a great idea as a 1st step, and killing the dep. on those platforms would be quite helpful too =)

plubius:
> Ooh - that's a really good idea =) I've been trying to write out Poco for a long time - but using the mobile apps as a Poco-lite version that doesn't require the HTTP parsing stuff is a great idea as a 1st step, and killing the dep. on those platforms would be quite helpful too =)

So I did a "grep -rl Poco:: *" in master and it looks like there are dozens of source files that iOS builds that still use Poco so does that mean you guys are still in an early phase of this conversion? If so, it might end up being easier to add net/HttpRequest.cpp to the iOS build and put #if !MOBILE around stuff in HttpRequest's methods that don't apply to or will fail on mobile. Right now, the number of build failures on iOS is small so maybe next week I can try that out and see if I can get iOS building and running. That will still require some work here and there as the conversion proceeds, but at least we'll have a way remove Poco from iOS hopefully without much hassle. Sound reasonable?

mmeeks:
Yes - we're in an early stage of de-poco-ization - people peck at it on an ongoing basis when we have time; of course, meanwhile other people use more Poco methods left and right ;-) We cut out a fair number of headers in the past though by re-working things. Poco logging is one of our more grim (and not co-incidentally under-performing) pieces that gets ~everywhere unfortunately. ... adding more #if !MOBILEs is an option - I just really hate the degree to which we have all this conditionality.

I'd prefer we loose more of those by adding a pair of iOS and Android stubs files that we can easily maintain to stub out methods that are not useful on mobile in some central place - have often needed that but not had the place to put them.

plubius:
I have gotten my 23.05 build up-to-date and rebuilt. Interestingly, there is only one compile error due to, as expected, an undefined http::Request::readData() symbol. Tomorrow I'll add the applicable net/* header and source files to the Xcode project and see how far I get.

Ashod (approved):
Thanks @plubius!

plubius:
OK. I've got iOS to build and run with Xcode 14.2. It's not pretty, but hopefully it has no effect on Online:

https://github.com/CollaboraOnline/online/pull/6589

Note: it's just a quick and dirty disabling of the Online server logging classes when building on mobile. Maybe it's not the ideal solution, but at least iOS builds and runs.

PR #6389, opened 2023-05-19
Sidebar calc Alignment fix button size of align icons

Andreas-Kainz:
#### Sidebar calc

| before | after |
| ---------- | ------- |
| Screenshot_20230519_113538 | Screenshot_20230519_113503 |

The Alignment icons (Align left, right, center, top, bottom, ... ) in every other cool app use `--btn-size` icon width/height, only at calc the size increase to 32px which cause the issue, that the 9 buttons are to large for the sidebar.

Change request is to use the correct size for the Alignment icons and ONLY `textorientbox` buttons use the 32px size.

Change-Id: I954c7885946dfd4e13261622673d1b97ef59001c

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -289,8 +289,8 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	height: auto;
 }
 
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Looks good

PR #6390, opened 2023-05-19
Sidebar writer table justify-start everything

Andreas-Kainz:
| before | after |
| --------- | -------- |
| Screenshot_20230519_130827 | Screenshot_20230519_131207 |

Change-Id: I94b047d206b4019ca15ec4b0343402643250658c


pedropintosilva:
@Andreas-Kainz I think we might not need these changes. The following PR https://github.com/CollaboraOnline/online/pull/6425 already fixes this, I think.

PR #6395, opened 2023-05-22
Fix: Speed-up callbacks issue# 4837

shehr7:
Change-Id: I05a14226589ab7ed23d273574dd175ca9ef2469d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -624,7 +624,7 @@ public:
             }
         }
 #endif
-        wakeup();
+        speedwakeup();
     }
 
     bool isAlive() const { return (_threadStarted && !_threadFinished) || _runOnClientThread; }
@@ -719,6 +719,16 @@ public:
         wakeup(_wakeup[1]);
     }
 
+void speedwakeup()
+{
+    if (_newSockets.empty() && _newCallbacks.empty())
+    {
+        _newSockets.push_back(Socket);
+        _newCallbacks.push_back(CallbackFn);
+        wakeup();
+    }
+}
+
     /// Global wakeup - signal safe: wakeup all socket polls.
     static void wakeupWorld();
 
@@ -737,7 +747,7 @@ public:
 
             std::lock_guard<std::mutex> lock(_mutex);
             _newSockets.emplace_back(std::move(newSocket));
-            wakeup();
+            speedwakeup();
         }
     }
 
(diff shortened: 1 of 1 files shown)


net/Socket.hpp
```diff
@@ -719,6 +719,16 @@ class SocketPoll
         wakeup(_wakeup[1]);
     }
 
+void speedwakeup()
+{
+    if (_newSockets.empty() && _newCallbacks.empty())
+    {
+        _newSockets.push_back(Socket);
+        _newCallbacks.push_back(CallbackFn);
```
mmeeks, line 727: Hmm - does this even compile ? where is CallbackFn from ?

net/Socket.hpp
```diff
@@ -719,6 +719,16 @@ class SocketPoll
         wakeup(_wakeup[1]);
     }
 
+void speedwakeup()
+{
+    if (_newSockets.empty() && _newCallbacks.empty())
+    {
+        _newSockets.push_back(Socket);
```
mmeeks, line 726: ditto for socket - what's up there?

net/Socket.hpp
```diff
@@ -624,7 +624,7 @@ class SocketPoll
             }
         }
 #endif
-        wakeup();
+        speedwakeup();
```
mmeeks, line 627: Is there a need to rename the method ? and if so, why - are there usefully two variants that do different things ?

net/Socket.hpp
```diff
@@ -719,6 +719,16 @@ class SocketPoll
         wakeup(_wakeup[1]);
     }
 
+void speedwakeup()
```
mmeeks, line 722: one-line comment to document why this one should be used would be helpful.
Also - a check that the lock is held that we expect to be locked here.

mmeeks (changes requested):
Please check the code at least compiles before asking for review.

pedropintosilva:
Thanks @shehr7 but better to spend sometime reading and re-reading the easy hack, investigating and ultimately testing it locally (so we can review runable patches)

shehr7:
It's weird that my compiler doesn't show these errors. Anyways I will check it again.

mmeeks:
> It's weird that my compiler doesn't show these errors. Anyways I will check it again.

But surely reading the code you can see that this can't compile; Quite possibly you are running 'make' in core and not in online (or somesuch).

PR #6396, opened 2023-05-22
jsdialog: position autofilter popup correctly in RTL mode

eszkadev:


gokaysatir (approved):
Thanks for the improvements.

PR #6398, opened 2023-05-22
browser: impress: introduce slide sorter context menu

lpranam:
added context menu entries which works out of the box

Change-Id: Ic0b598a575c4f0fb337ae79bc51cac2330d1115f

* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -31,6 +31,7 @@ L.Control.PartsPreview = L.Control.extend({
 		this._container = container;
 		this._partsPreviewCont = preview;
 		this._partsPreviewCont.onscroll = this._onScroll.bind(this);
+		this._idNum = 0;
 	},
 
 	onAdd: function (map) {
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.PartsPreview.js
```diff
@@ -218,6 +218,37 @@ L.Control.PartsPreview = L.Control.extend({
...
+			$.contextMenu({
+				selector: '.' + imgClassName,
+				className: 'slide-preview-contextmenu',
+				items: {
+					newslide: {
+						name: _('New Slide'),
+						callback: function() { this._map.insertPage(); }
+					},
+					duplicateslide: {
+						name: _('Duplicate Slide'),
```
timar, line 231: Please use existing translations everywhere, if possible, e.g. here `name: _UNO('.uno:DuplicateSlide', 'presentation')`.

timar (changes requested):

timar (changes requested):
* Right click does not select slide/page. So the menu command will have effect on the selected slide/page, and not on the slide/page where the context menu appears.
* The style (css) of this context menu is different to other context menus. E.g. the font is different.

timar (approved):

browser/src/control/Control.PartsPreview.js
```diff
@@ -171,13 +172,14 @@ L.Control.PartsPreview = L.Control.extend({
 	_createPreview: function (i, hashCode, bottomBound) {
 		var frameClass = 'preview-frame ' + this.options.frameClass;
 		var frame = L.DomUtil.create('div', frameClass, this._partsPreviewCont);
-		frame.id = 'preview-frame-part-' + i;
+		frame.id = 'preview-frame-part-' + this._idNum;
```
eszkadev, line 175: why do we need this additional `_idNum` variable?  wasn't  `i` already doing the same?
what when we switch to master page and go back? or we add/delete slide?
I think this function will be called again but index will not be reset

lpranam:
closed in favor of #6529

PR #6400, opened 2023-05-22
browser: fix formula bar 'useTextInput' property

hcvcastro:
The desktop application can use IME composed
characters.

Change-Id: I0db3b23c8859fe25cc3ea19985d4da19e779626f


eszkadev:
it doesnt' work:
1. type into formulabar wods separated with space: "Hello world"
Result: you will see characters mixing positions after space

PR #6425, opened 2023-05-24
Sidebar: Align everything to start and remove unnecessary css

pedropintosilva:
Better to align everything to start so if the components differ in
width they will still be aligned (as opposed to be laid out in a staircase)

This is a follow up from 1bfbdf807208b98be56aa558f749e32c7a67acdf coming with
the following PR https://github.com/CollaboraOnline/online/pull/6389

Change-Id: Ic607567d30656cc122f9cbfd71bd3d910ebfe9bf

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -94,10 +94,8 @@ img.sidebar.ui-image {
 	line-height: var(--default-height);
 }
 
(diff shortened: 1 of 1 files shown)


Andreas-Kainz (approved):
Thanks

Issue #4837, opened 2022-05-31
Speed-up callbacks ....


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @Ashod 

### Detailed description and rationale
When we send messages across threads (one thread per SocketPoll) we have to use 'addCallback' or 'insertSocket' these queue up the work to be done in the main thread and they also send a 'wakeup'.

Since we take a lock anyway here, we should check whether we actually need to send a wakeup and if so, not send another one - otherwise we just fill up the internal wakeup socket with data for no good reason, spin the main-loop pointlessly and waste CPU, scheduler cycles and goodwill from developers tracing the code =)

We should not add locking to wakeup() itself. Quite probably though we should do a larger read from the wakeup pipe to empty it inside our ::poll function though - since there is no benefit in spinning and waking up again if there have been multiple wakeups.

### Code pointers

```
net/Socket.hpp see:
    /// Protects _newSockets and _newCallbacks
    std::mutex _mutex;
    std::vector<std::shared_ptr<Socket>> _newSockets;
    std::vector<CallbackFn> _newCallbacks;

```
And uses of _newSockets and _newCallbacks. We should have a function that checks if _newSockets and _newCallbacks is empty before doing a 'wakeup()'.

Thanks ! =)

callmetushar123:
Hello. Can you point out where this part of the code is present in the file?

thebearon:
For the record, #6467 causes a slight regression in coolwsd shutdown times (5-10 s in a debug build).

While that won't affect users, it could be frustrating for devs, who start up and shut down builds often.

PR #6426, opened 2023-05-24
 Speed-up callbacks, issue #4837 

shehr7:
* Resolves: #4837
* Target version: master 

### Summary
Changes:

    condition on _newsockets and _newcallbacks --> Inside wakeup function
    Use loop to read Wakeup pipe inside poll function --> in socket.cpp 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -367,11 +367,11 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         // Clear the data.
 #if !MOBILEAPP
         int dump[32];
-        dump[0] = ::read(_wakeup[0], &dump, sizeof(dump));
+        dump[0] = ::read(_wakeup[i], &dump, sizeof(dump));
         LOG_TRC("Wakup pipe read " << dump[0] << " bytes");
 #else
         LOG_TRC("Wakeup pipe read");
-        int dump = fakeSocketRead(_wakeup[0], &dump, sizeof(dump));
+        int dump = fakeSocketRead(_wakeup[i], &dump, sizeof(dump));
 #endif
 
         std::vector<CallbackFn> invoke;
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -614,6 +614,7 @@ public:
             }
         }
 #endif
+    if (!_newSockets.empty() || !_newCallbacks.empty())
         wakeup();
     }
 
@@ -697,7 +698,6 @@ public:
             LOG_WRN("Waking up dead poll thread ["
                     << _name << "], started: " << (_threadStarted ? "true" : "false")
                     << ", finished: " << _threadFinished);
-
         wakeup(_wakeup[1]);
     }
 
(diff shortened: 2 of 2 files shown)


net/Socket.cpp
```diff
@@ -365,13 +365,20 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
...
     {
         // Clear the data.
+        //Size of wakeup array
+        size_t arraySize = sizeof(_wakeup) / sizeof(_wakeup[0]);
 #if !MOBILEAPP
         int dump[32];
-        dump[0] = ::read(_wakeup[0], &dump, sizeof(dump));
-        LOG_TRC("Wakup pipe read " << dump[0] << " bytes");
+        //Empty wakeup pipe inside poll function.
+        for (size_t ii = 0; ii < arraySize; ii++)
```
GulsahKose, line 373: Here you can use "i" instead of "ii". Scope will be in loop so don't need to concern about it.

net/Socket.hpp
```diff
@@ -715,8 +715,11 @@ class SocketPoll
             LOG_WRN("Waking up dead poll thread ["
                     << _name << "], started: " << (_threadStarted ? "true" : "false")
                     << ", finished: " << _threadFinished);
-
-        wakeup(_wakeup[1]);
+        // Check if _newSockets and _newCallbacks is empty before doing a 'wakeup()'
+        if (!_newSockets.empty() || !_newCallbacks.empty())
+        {
+            wakeup(_wakeup[1]);
```
GulsahKose, line 721: Here you can remove the curly braces too.

GulsahKose (commented):
Hi Shehr,

Thanks for the fix and comments. Logic seems good to me. Just I can say something about naming etc. You can see my comments in lines.

net/Socket.cpp
```diff
@@ -365,13 +365,20 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
...
+        {
+            dump[0] = ::read(_wakeup[ii], &dump, sizeof(dump));
+            LOG_TRC("Wakup pipe read " << dump[0] << " bytes");
+        }
 #else
         LOG_TRC("Wakeup pipe read");
-        int dump = fakeSocketRead(_wakeup[0], &dump, sizeof(dump));
+        for (size_t ii = 0; ii < arraySize; ii++) {
+            int dump = fakeSocketRead(_wakeup[ii], &dump, sizeof(dump)); }
```
GulsahKose, line 381: Same. please use i instead of ii

GulsahKose:
Unfortuanately i detected performance decline. You can test it in Writer. Just try to select one word. You'll see, there is some delay. I'll review more detailed and let you know.

shehr7:
@GulsahKose I'm sorry, for I ran into some errors and had to update the file. I will let you know once the commit is up to date

shehr7:
@GulsahKose I am not sure about this issue as I tested in calc and writer, and experienced a better performance. Might be because of some other reason.

shehr7:
@Ashod @GulsahKose  I have made some amendments. Kindly check to see if it works, for me, this version works faster than the previous ones.

Issue #6416, opened 2023-05-23
[Linking API] Endless preview for sheets and presentation slides


juliushaertl:
After further testing for #6229 i noticed that getting a preview for slides or sheets of a spreasheet is also running endless requests.

Two example files are below:

Requests used:

```
curl --request PUT \
  --url http://192.168.178.65:9980/cool/get-thumbnail \
  --header 'Content-Type: multipart/form-data' \
  --form 'data=@/Users/julius/Demo files/Market analysis.ods' \
  --form target=market-segments
```

```
curl --request PUT \
  --url http://192.168.178.65:9980/cool/get-thumbnail \
  --header 'Content-Type: multipart/form-data' \
  --form 'data=@/Users/julius/Demo files/Market analysis.ods' \
  --form 'target=Slide 1'
```

[Market analysis.ods](https://github.com/CollaboraOnline/online/files/11547034/Market.analysis.ods)
[presentation.odp](https://github.com/CollaboraOnline/online/files/11547036/presentation.odp)

COOLWSD version:
23.05.0.1 git hash: 
[cfd94eaa](https://github.com/CollaboraOnline/online/commits/cfd94eaa)
LOKit version:
Collabora Office 23.05.0.1 git hash: 
[805c4cd](https://hub.libreoffice.org/git-core/805c4cd)

cc @eszkadev @pedropintosilva

Issue #6229, opened 2023-04-25
[Linking API] Image previews with get-thumbnail endpoint doesn’t work for some elements


juliushaertl:
**Describe the bug**

Request a preview of a target through the linking API using the following example document [example.odt](https://github.com/CollaboraOnline/online/files/11323201/example.odt) only works for some types.

Not working, request is running infinitely:
- Tables (e.g. `Revisionsheettable|table`)
- Headings (e.g. `1.GENERAL INFORMATION|outline`)
- Sections: (e.g `Intro|region`)

Working:
- Images
- OLE objects
- Drawing objects

```
curl --request PUT \
  --url http://collabora.dev.local/cool/get-thumbnail \
  --header 'Content-Type: multipart/form-data' \
  --form data=@/Users/julius/example.odt \
  --form 'target=1.GENERAL INFORMATION|outline'
```

Tested with CODE 22.05.13.1

PR #6428, opened 2023-05-24
linking-api: always response for thumbnail render requests #6416

eszkadev:
followup for:
commit 012b00d
linking-api: always response for thumbnail render requests #6229

in calc and impress we don't get cursor message when switching to target (slide, range), so don't check if target was reached, statechanged messages are good enough signal we are loaded already


timar (approved):

GulsahKose (commented):
Works as expected. Thanks

PR #6433, opened 2023-05-25
Fixes: Sidebar icons are cut off

pedropintosilva:
This hard coded width is not needed and was imposing that width value
to #indentfieldbox even if there is other elements present resulting
in the whole container being bigger than the sidebar itself

This issue became more evident with
https://github.com/CollaboraOnline/online/commit/27fddae9c6b3fe0dc0c948bb132034f9ffa38012 (even
though it's not related to the bug but rather exposed it even more)

Change-Id: I90e684b395b950e4cc25a96a9875d4702016411d

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -261,10 +261,6 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	left: -10px;
 }
 
-#ParaPropertyPanel.sidebar.ui-grid #indentfieldbox {
-	width: 158px;
-}
-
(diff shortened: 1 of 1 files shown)


pedropintosilva:
adding you @Andreas-Kainz since the initial width was added by you (probably needed in the past ?) but hopefully we can merge this now

pedropintosilva:
This does not need to wait for cypress tests. It was passing until now even thought the icons were cropped. And now they aren't. On top of that there is no additional change apart for the removal of the hard coded px on that specific id.

PR #6437, opened 2023-05-25
fix: Avoid important property on all classic toolbar w2ui icons

juliushaertl:
Change-Id: I2a2ace9ffc2a31d500b6819f790a07e5762a5f6a

* Target version: master 

### Summary

This is basically a regex replace for any usage where the background image was enforced with `!important` in toolbar.css for the classic toolbar. From my perspective those were not needed at all and also make life harder during icon adjustments in other css parts like branding. 

I went through the classic toolbar on all apps and couldn't find a problem, but probably needs some careful checking from @pedropintosilva 

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -602,703 +602,703 @@ button.leaflet-control-search-next
 	color: var(--color-text-dark);
 }
 
-.w2ui-icon.basicshapes_rectangle { background: url('images/lc_rect.svg') no-repeat center !important; }
-.w2ui-icon.basicshapes_round-rectangle { background: url('images/lc_rect_rounded.svg') no-repeat center !important; }
-.w2ui-icon.basicshapes_quadrat { background: url('images/lc_square.svg') no-repeat center !important; }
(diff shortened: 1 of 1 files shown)


vmiklos:
Hi Julius! Just a nit: if you push to a new branch in the future, so that you can create a PR from it, please name such branches as private/juliushaertl/xyz. This helps to have a better overview of actual branches which are not just sources for PRs. Thanks. :-)

pedropintosilva (approved):
Nice! This was one of the low hanging fruits waiting to be done, thanks :)

note to myself: Maybe there are still other legacy places where !important rules could be removed

juliushaertl:
> Hi Julius! Just a nit: if you push to a new branch in the future, so that you can create a PR from it, please name such branches as private/juliushaertl/xyz. This helps to have a better overview of actual branches which are not just sources for PRs. Thanks. :-)

Sure, will stick to that naming scheme then next time :)

juliushaertl:
Not sure about cypress, test seems to pass fine locally

PR #6444, opened 2023-05-26
zotero: initialize only one time

eszkadev:
we didn't check if zotero was already initialized what caused reload of the notebookbar when any new user joined to collaborative editing


gokaysatir (approved):
Thanks :)

gokaysatir (approved):
Thanks :)

mmeeks:
Would love that in 22.05 =) thanks !

PR #6445, opened 2023-05-26
Fix debug build breakage due to undefined function on iOS

plubius:


plubius:
This pull request is now ready for review and this pull request replaces pull request #6358.

Issue #6443, opened 2023-05-26
Make canvas background and page background adjustable through CSS


juliusknorr:
If I remember correctly at some point it was possible to adjust the canvas background (e.g. behind the actual document) through CSS where the style property value was picked up by javascript and used to render a dynamic background color. However it seems that currently the value is either hardcoded or taken from the core messages itself:

https://github.com/CollaboraOnline/online/blob/master/browser/src/layer/tile/CanvasTileLayer.js#L1750-L1759
https://github.com/CollaboraOnline/online/blob/master/browser/src/layer/tile/CanvasSectionContainer.ts#L494

@pedropintosilva What do you think about getting the css backgroundColor of the #document-container and use that instead so it can be dynamically adjusted

Sample code could be `window.getComputedStyle(document.getElementById('document-container')).backgroundColor`

pedropintosilva:
It would be equally good to have the page color coming from css var. @OFFTKP has been helpful in the IRC and pointed to:

> the color itself is defined in coolkitconfig.xcu.in:74 for light and coolkitconfig.xcu.in:391 for dark. The theme is switched through an uno command (uno:ChangeTheme)

Andreas-Kainz:
when this is fixed we can have a lot at #6449

juliushaertl:
I've briefly chatted with @pedropintosilva and @eszkadev about this and we though it would be nice to have a separate css variables for this to be more flexible.

In CSS it could still default to --color-main-background but it can be separately be adapted trough the iframe [css_variables](https://sdk.collaboraonline.com/docs/theming.html#content-of-hidden-field-css-variables).

Darshan-upadhyay1110:
The canvas background now comes from the --color-canvas CSS variable and the document background from --color-background-document, both defined in the colour-palette CSS and adjustable by integrators via the theming css_variables mechanism 

[f09f7436366c](https://github.com/CollaboraOnline/online/commit/f09f7436366c9a9345c3d8097c6ab8da6bd41216)
Dark mode: Canvas, get directly the css var property value

Pedro Pinto Silva, 2023-06-05
GitHub
[738de114d71a](https://github.com/CollaboraOnline/online/commit/738de114d71a573dcce74c03916dd2407be1b781)
Make writer page background match document color
Paris Oplopoios, 2023-06-28

PR #6451, opened 2023-05-26
--color-main-background CSS variable will be used for canvas backgrou…

gokaysatir:
…nd as default.

To change the behaviour: call app.setBackgroundColorMode(false); ^ This will take the variables sent from core side into account. Otherwise they are ignored.

Change-Id: Ib7321bf7c65ff9bcb1937957cea3a070b0eecd32

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


juliushaertl (approved):
Tested and works

juliushaertl:
@pedropintosilva Would be great to get this in to make @jancborchardt happy with adjusted dark mode backgrounds :wink: 

I think https://github.com/CollaboraOnline/online/issues/6443#issuecomment-1568293061 could also be addressed in a follow up.

Issue #6406, opened 2023-05-23
23.05 - Draw - In Compact View, the Page Layout icon highlights Slide Layout


Ezinnem:
**Describe the bug**

**To Reproduce**

1. Open .odg file
2. Switch to compact View

Hover on the Page Layout Icon, see that it highlights Slide Layout

**Expected behavior**
Hover text should be "Page Layout"

**Actual behavior**
Hover text reads "Slide Layout"

**Screenshots**

https://github.com/CollaboraOnline/online/assets/54344301/31956e1e-5160-4919-88a5-f11f77eea8f4

**Desktop (please complete the following information)**

COOLWSD version:
23.05.0.1snapshot git hash: https://github.com/CollaboraOnline/online/commit/012b00d6d4530fb7e74fac5c32696c82de028036(E)

Phabricator link: [T42483](https://phabricator.collabora.com/T42483)

pedropintosilva:
@codewithvk was interested in fixing this , I think

codewithvk:
I can work on it.

codewithvk:
![image](https://github.com/CollaboraOnline/online/assets/61119120/ca3c25d0-a992-4500-ace0-c626820c1e7f)
It seems like, We did intentionally use the text "Slide Layout" as in `View` we have text "Slidebar".

@pedropintosilva @Ezinnem

pedropintosilva:
@codewithvk true. But now that I think about it's probably best to change it (specially because the sidebar's contents can change depending on the object selected)

I have assign this to you :)

codewithvk:
One more doubt, 
![image](https://github.com/CollaboraOnline/online/assets/61119120/6115c0b9-7c3d-432d-8e3b-7908a719b7e2)
Why do we have Page Layout and Sidebar both?

PR #6452, opened 2023-05-26
Draw - Change tool tip text to Page Layout

codewithvk:
Change-Id: Id44046569e5904f39ed8b57f2094f0744a5daaa8

* Resolves: #6406 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Darshan-upadhyay1110:
@codewithvk thanks:. Need to advice you about test this change in both impress as well as in Draw. 

i think if you change here Slide layout to page layout it will effect the presentation slide layout tool tip. 

and that we can not do because we want slide layout in presentation as tool-tip and page layout in  draw as tool tip

Darshan-upadhyay1110 (changes requested):
This might effect to presentation so please verfiy in both impress as well as in draw for sidebar tooltip. 

i am seeing it will change value of tool tip on impress as well if you change this here 

image

codewithvk:
Hey @Darshan-upadhyay1110 , Thanks for feedback.

I was confused about one thing. Although I posted in the Issue and included a screenshot there, I wanted to share the screenshot here as well. In the Impress Compact view, we can see both the Sidebar and Page Layout options. These options seem to have the same functionality.
image

Could you please review the discussion in the Issue and provide your valuable feedback?

Issue Link [Here](https://github.com/CollaboraOnline/online/issues/6406#issuecomment-1564843937)

Darshan-upadhyay1110:
hii @codewithvk i got your point , this issue we might need to discuss but the fix here you suggested is not based on sidebar and page layout option.
assume you do not have side bar in compact view but still this fix will be changing the tool tip value from slide layout to page layout ( which should not in impress ).

Tip: try to do fix which only have effect to draw tab ( not change values for others ) :).

hope this clears the thing for current issue.

Darshan-upadhyay1110 (commented):
Great looks good. @eszkadev will you take a look into it once?

timar:
I'm sorry that I did not notice this PR, because unfortunately it's not good. 
- It is good to remember, that `'drawing'` makes no sense as parameter of the `_UNO()` function. In core Impress and Draw are the same module.
-  The first line of browser/src/unocommands.js is `// Don't modify, generated using unocommands.py` and you modified it by hand. Your change disappeared the next time we ran unocommands.py and nobody noticed.
You could have added simply `_('Page layout')` to top toolbar.

PR #6454, opened 2023-05-26
Sidebar writer Paragraph Section BackgroundColor justify-content: end

Andreas-Kainz:
Background Color use justify-content: end
image

Change-Id: I23a5690b0b33dae426914ebef30fae6d97d598c7

PR #6392, opened 2023-05-20
Sidebar Paragraph and Table Section has 1fr 1fr (50/50) width

Andreas-Kainz:
The Spacing and Indent section has the same width
The Table section has now the same width

image

Change-Id: I0d1680b8fd0e97e109d86203fba7fe2ca0c82f51

PR #6455, opened 2023-05-26
Sidebar Paragraph and Table Section has 1fr 1fr (50/50) width

Andreas-Kainz:
see #6392 
This is needed to have the colums same width and
make everything nicely align

| before | after |
| --------- | -------- |
| Screenshot_20230527_010109 | Screenshot_20230527_010927 |

The Background Color command will be correct aligned with #6454 

Change-Id: I21887ba06606d6ea4bc10eaafb3964f2659c90e2

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -262,6 +262,16 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	width: 118px;
 }
 
+#TableEditPanel.sidebar.ui-grid #grid1,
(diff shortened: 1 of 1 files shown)


Andreas-Kainz:
ping @pedropintosilva

pedropintosilva (commented):
Thanks @Andreas-Kainz but with this change any other scenario (where table properties are not display) it looks worst:

image

So maybe it would be better to have the table properties full width (?)

PR #6457, opened 2023-05-27
Test Improvements

Ashod:
- wsd: test: fix uninitialized member
- wsd: test: initialize logs before unit-tests
- wsd: test: improved UnitLoad
- wsd: test: log resetting wopi event counters
- wsd: test: better upload conflict test setup
- wsd: test: simplify UnitWOPIHttpRedirectLoop
- wsd: test: use http::Response
- wsd: test: simplify UnitWopiHttpRedirect
- wsd: test: use http::Response
- wsd: test: break connectLOKit when shutdown is signaled
- wsd: test: better polling in tests
- wsd: test: merge UnitWopiHttpRedirectLoop with Redirect
- wsd: test: reduce test run time
- wsd: test: better Ownertermination test


test/UnitLoad.cpp
```diff
@@ -180,7 +181,7 @@ UnitBase::TestResult UnitLoad::testExcelLoad()
 
         // Expected format is something like 'status: type=text parts=2 current=0 width=12808 height=1142 viewid=0\n...'.
         StringVector tokens(StringVector::tokenize(status, ' '));
-        LOK_ASSERT_EQUAL(static_cast<size_t>(7), tokens.size());
+        LOK_ASSERT_EQUAL(static_cast<size_t>(9), tokens.size());
```
eszkadev, line 184: above comment with example content has 7 tokens, would be good to update

eszkadev: how it become 9 now? this check wasn't executed in the past?

Ashod: I updated it somewhere else and missed this one. Will update in a follow up patch.

Yes, this test was disabled. I enabled it and 4 or 5 others.

Thanks!

Issue #6405, opened 2023-05-23
23.05 - Draw - Page Layout cannot be open unless from the Sidebar Menu Item


Ezinnem:
**Describe the bug**

**To Reproduce**

1. Create a new .odg file
2. Using the icon in the top-right corner, try to close / reopen the Page Layout (sidebar)

Notice that it appears and disappears within seconds.

It can only be opened using the Sidebar item in the Menu bar.

**Expected behavior**
Page Layout should appear when the user clicks on the icon.

**Actual behavior**
Page Layout appears and disappears when the user clicks on the icon

**Screenshots**

https://github.com/CollaboraOnline/online/assets/54344301/828250f5-cc71-4237-8ff3-f3906318540b

**Desktop (please complete the following information)**
COOLWSD version:
23.05.0.1snapshot git hash: https://github.com/CollaboraOnline/online/commit/012b00d6d4530fb7e74fac5c32696c82de028036(E)

Phabricator link: [T42482](https://phabricator.collabora.com/T42482)

PR #6465, opened 2023-05-29
Fix Draw - Page Layout cannot be open from top tool sidebar icon

Darshan-upadhyay1110:
Change-Id: If9a57484a8335034e3007f47513cd4827c8decbc

* Resolves: #6405 
* Target version: master 

- Fixed sidebar toogle functionality in odg (draw) file

### DIFF
--- a/browser/src/control/Control.NotebookbarDraw.js
+++ b/browser/src/control/Control.NotebookbarDraw.js
@@ -778,7 +778,7 @@ L.Control.NotebookbarDraw = L.Control.NotebookbarImpress.extend({
 			{
 				'type': 'bigtoolitem',
 				'text': _UNO('.uno:ModifyPage', 'drawing', true),
-				'command': '.uno:ModifyPage'
+				'command': '.uno:SidebarDeck.PropertyDeck'
 			},
 			{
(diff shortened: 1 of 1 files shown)


eszkadev (changes requested):
this cannot be done because then it will not be possible to switch back from navigator I think

browser/src/control/Control.Sidebar.js
```diff
@@ -152,7 +152,7 @@ L.Control.Sidebar = L.Control.extend({
 	},
 
 	changeDeck: function(unoCommand) {
-		if (unoCommand !== null)
+		if (unoCommand !== null&& unoCommand !== '.uno:ModifyPage')
```
codewithvk, line 155: I don't think we should add a condition here. Instead for the top right button, we can add fix here: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.TopToolbar.js#L242

Darshan-upadhyay1110: yes I forget to remove this . thanks :)

eszkadev (approved):
looks good :)

PR #6467, opened 2023-05-29
Fix: Speed-up callbacks

shehr7:
Change-Id: I796d71e48ae7110fb8d88eda781e362faba75aa2

* Resolves: #4837
* Target version: master 

### Summary

- Lock in callback function was already placed. 
- Poll already reads the whole value of wakeup[0] and it does not need to read wakeup[1].
- Check the condition to look for new sockets and call-backs before calling wakeup() is placed here. 

The third one was to check for callbacks and sockets before calling wakeup, which I implemented, and also studied more ways but checked that most of it was implemented.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -614,6 +614,7 @@ public:
             }
         }
 #endif
+    if (!_newSockets.empty() || !_newCallbacks.empty())
         wakeup();
     }
 
@@ -718,8 +719,10 @@ public:
             newSocket->resetThreadOwner();
 
             std::lock_guard<std::mutex> lock(_mutex);
+            const bool wasEmpty = _newSockets.empty() && _newCallbacks.empty();
             _newSockets.emplace_back(std::move(newSocket));
-            wakeup();
+            if (wasEmpty)
+                wakeup();
         }
     }
 
@@ -746,7 +749,7 @@ public:
     void addCallback(const CallbackFn& fn)
     {
         std::lock_guard<std::mutex> lock(_mutex);
-        bool wasEmpty = _newCallbacks.empty();
+        const bool wasEmpty = _newSockets.empty() && _newCallbacks.empty();
         _newCallbacks.emplace_back(fn);
         if (wasEmpty)
             wakeup();


GulsahKose:
I checked the commit according to Michael's solution description. It seems logical to me. But I don't know exactly how to observe the performance increase. It would be nice to have one more review. Thanks @shehr7

net/Socket.hpp
```diff
@@ -697,7 +698,6 @@ class SocketPoll
             LOG_WRN("Waking up dead poll thread ["
                     << _name << "], started: " << (_threadStarted ? "true" : "false")
                     << ", finished: " << _threadFinished);
-
```
Ashod, line 700: This blank line helps readability, since the previous log line is multi-lined.

net/Socket.hpp
```diff
@@ -719,7 +719,8 @@ class SocketPoll
 
             std::lock_guard<std::mutex> lock(_mutex);
             _newSockets.emplace_back(std::move(newSocket));
-            wakeup();
+            if (!_newSockets.empty() || !_newCallbacks.empty())
```
Ashod, line 722: There is no need to check for `!_newSockets.empty()` here, since we just inserted an item in it, so this can never be false.

This means that this check is unnecessary, since the condition will always be true, and we indeed must always call `wakeup()` after inserting a new socket.

However, there is something better to do here! Which was done in `addCallback()` (before you removed it). So let me comment there, and you can apply the same idea here too.

net/Socket.hpp
```diff
@@ -746,9 +747,8 @@ class SocketPoll
     void addCallback(const CallbackFn& fn)
     {
         std::lock_guard<std::mutex> lock(_mutex);
-        bool wasEmpty = _newCallbacks.empty();
         _newCallbacks.emplace_back(fn);
-        if (wasEmpty)
+        if (!_newSockets.empty() || !_newCallbacks.empty())
```
Ashod, line 751: This condition is not helpful because it must always be true (we just inserted an item into `_newCallbacks`). However, the old solution was smart, so we should bring it back (and apply it in the `insertNewSocket()` case too).

You see, normally when we insert into either `_newCallbacks` or `_newSockets`, we want to `wakeup()`. But this `wakeup()` is expensive, because it needs to write into the pipe. The good news is that we can optimize this when we know `wakeup()` is already called. The previous code here did that: it checked if `_newCallbacks.empty()` *before* inserting and called `wakeup()` only when it was empty. Meaning, it only calls `wakeup()` the first time we insert something, but after that, any further insertions don't need separate calls to `wakeup()` since that needs to be done only once (util we read and empty `_newCallbacks` and `_newSockets`.

So, the solution here and in the `insertNewSocket()` function is this:
```
const bool wasEmpty = _newSockets.empty() && _newCallbacks.empty();
_newCallbacks.emplace_back(fn);
if (wasEmpty)
    wakeup();
```
(Same solution for the `insertNewSocket()` case, but change the second line accordingly.)

Ashod (commented):
Thanks @shehr7. I left some notes.

Ashod (approved):
Thanks @shehr7. Looks good to me.

pedropintosilva:
hm @Ashod I think I'm noticing some performance hit after this got in, could you please re-check this? If so, please revert it or ping me and I will be happy to do so

PR #6468, opened 2023-05-29
private/marco/a11y/contenteditabl

mcecchetti:
- a11y: TextInput: switch from <textarea> to content editable <div>
- a11y: populating editable area with focused paragraph

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -249,6 +249,7 @@ COOL_JS_LST =\
 	src/layer/ImageOverlay.js \
 	src/layer/marker/ProgressOverlay.js \
 	src/layer/marker/TextInput.js \
+	src/layer/marker/A11yTextInput.js \
 	src/layer/marker/Icon.js \
 	src/layer/marker/Icon.Default.js \
 	src/layer/marker/Marker.js \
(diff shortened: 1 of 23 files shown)


mmeeks:
I assume this requires: https://gerrit.libreoffice.org/c/core/+/152363 first ?

mmeeks:
Nice - I forget what we need to do to re-run the checks; but re-basing and force pushing to the branch is prolly a good plan to re-trigger here (?) =)

mcecchetti:
Usually I close and re-open the PR. ;-)
Anyway, as far as I can remember new merged commits are not immediately available here

eszkadev (approved):
looks good in general

one question: do we want to enable that unconditionally? probably worth adding some switch in coolwsd for that?

mcecchetti:
> looks good in general

Thanks for your review.

> 
> one question: do we want to enable that unconditionally? probably worth adding some switch in coolwsd for that?

So, I plan to add such a switch, too. I hope to include this option in my next patch.
Later we should also allow a user  to enable/disable this feature through the UI.
I point out that at present it's enabled for Writer only.

mcecchetti:
So, all tests checking the editable area content are passing now.
Also desktop and mobile tests are passing except for some `beforeEach` that's randomly failing.
Since failures are few, random and unrelated I think that's safe to merge the patch without waiting for having an all green run.

PR #6469, opened 2023-05-29
prevent non numerical input

hcvcastro:
- browser: add 'getLang' function to get the current ...
- browser: use Intl built-in object to get number format
- browser: improve _preventNonNumericalInput function

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -111,7 +111,11 @@ window.app = {
 			(window.DocumentTouch && document instanceof window.DocumentTouch)) && !chromebook;
 
 	var isInternetExplorer = (navigator.userAgent.toLowerCase().indexOf('msie') != -1 ||
-			navigator.userAgent.toLowerCase().indexOf('trident') != -1);
+				  navigator.userAgent.toLowerCase().indexOf('trident') != -1);
+
+	var navigatorLang = navigator.languages && navigator.languages.length ? navigator.languages[0] :
+	    (navigator.language || navigator.userLanguage || navigator.browserLanguage || navigator.systemLanguage);
+
 
 	global.L = {};
 
@@ -228,7 +232,11 @@ window.app = {
 
 		// @property retina: Boolean
 		// `true` for browsers on a high-resolution "retina" screen.
-		retina: (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1
+		retina: (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1,
+
+		// @property lang: String
+		// browser language locale
+		lang: navigatorLang
 	};
 
 	global.mode = {
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -43,6 +43,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_menus: null,
 	_colorPickers: null,
 	_colorLastSelection: {},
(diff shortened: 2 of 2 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1518,6 +1542,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_metricfieldControl: function(parentContainer, data, builder, customCallback) {
 		var value;
 		var controls = builder._controlHandlers['basespinfield'](parentContainer, data, builder, customCallback);
+		if (!L.Browser.chrome) {
+			controls.spinfield.onkeypress = L.bind(builder._preventNonNumericalInput, builder);
```
eszkadev, line 1546: can't we add this only one time in `basespinfield` which is shared in metric and formatted field?

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -285,6 +292,20 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
 	},
 
+	_preventNonNumericalInput: function(e) {
+		e = e || window.event;
+		var charCode = (typeof e.which == 'undefined') ? e.keyCode : e.which;
+		var charStr = String.fromCharCode(charCode);
+		if (charStr.match(/^[0-9]+$/))
+			return;
+
+		for (var part in this._numberParts) {
```
eszkadev, line 302: this is not enough, we can still type:
'1.1,,,,,,' or '-13-----'

Best would be to use single regexp where we can use minus sign only on the beginning and separators at correct positions. This regex should check input field value, not only current key code

browser/src/control/Control.JSDialogBuilder.js
hcvcastro: are we sure that basespinfield class is numeric?
what about integers, currency?

browser/src/control/Control.JSDialogBuilder.js
hcvcastro: oh, sound good, thanks

eszkadev:
now it looks better :)
I think we can add also `onchange` event listener which will remove '-' signs which are not at the first position ? what do you think?

browser/js/global.js
```diff
@@ -228,7 +228,13 @@ window.app = {
 
 		// @property retina: Boolean
 		// `true` for browsers on a high-resolution "retina" screen.
-		retina: (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1
+		retina: (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1,
+
+		getLang: function () {
+			return (navigator.languages && navigator.languages.length ? navigator.languages[0] :
+				navigator.language || navigator.userLanguage ||
+				navigator.browserLanguage || navigator.systemLanguage);
```
Ashod, line 236: Is this supposed to be 
`navigator.languages && navigator.languages.length ? navigator.languages[0] : (navigator.language || navigator.userLanguage ||
				navigator.browserLanguage || navigator.systemLanguage)` 
or 
`(navigator.languages && navigator.languages.length ? navigator.languages[0] :
				navigator.language) || navigator.userLanguage ||
				navigator.browserLanguage || navigator.systemLanguage` ?

Perhaps adding `()` would make it more readable.

Ashod (approved):
Thanks @hcvcastro. One comment, but otherwise looks good.

hcvcastro:
Thanks, I hope it is clear now

PR #6471, opened 2023-05-30
jsdialog: don't vertically center by default grids

eszkadev:
by default every grid had vertically centered content, so in case of small boxes inside it looks bad:
vertcent


eszkadev:
passed tests locally

PR #6479, opened 2023-05-30
Restore precache next slide

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

- [X ] I have run tests with `make check`


dennisfrancis (approved):
Thanks for the fix!

PR #6481, opened 2023-05-30
Dark mode improvements

OFFTKP:
* Replaced documentbackgroundcolor command with applicationbackgroundcolor command. This command is responsible for changing the background color on the Online side in Calc and Draw/Impress
* Added tile/delta dumping option in the About debug menu
* Fixed "graphic dragging" delta related bug


kit/Delta.hpp
```diff
@@ -347,9 +348,12 @@ class DeltaGenerator {
...
                     {
-                        char cnt = output[lastCopy];
-                        if (output[lastCopy + 1] + cnt == (char)(match) &&
-                            output[lastCopy + 2] + cnt == (char)(y))
+                        // check if we can extend the last copy
+                        uint8_t cnt = output[lastCopy];
+                        if (output[lastCopy + 1] + cnt == match &&
+                            output[lastCopy + 2] + cnt == y &&
+                            // make sure we're not copying from out of bounds of the previous tile
+                            output[lastCopy + 1] + cnt + 1 < prev.getHeight())
```
mmeeks, line 356: Seems reasonable - I wonder if we have a unit test for this use-case =) would be good to have one, and also one to check that we do generate multi-row copy outputs still.

test/DeltaTests.cpp
```diff
@@ -112,17 +116,24 @@ std::vector<char> DeltaTests::applyDelta(
...
 //            std::cout << "new " << length << " at " << destCol << ", " << destRow << '\n';
             LOK_ASSERT(length <= width - destCol);
 
             char *dest = &output[width * destRow * 4 + destCol * 4];
-            for (size_t j = 0; j < length * 4 && i < delta.size(); ++j)
-                *dest++ = delta[i++];
+            for (size_t j = 0; j < length * 4 && i < delta.size(); j += 4)
+            {
+                // Swap BGR to RGB
+                *dest++ = delta[i + 2];
```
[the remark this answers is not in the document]
mmeeks: My expectation is that we treat a pixel as a pixel independent of its' BGRA/RGBA-ness - it's certainly far faster to handle 32bit integers in one bite.

mmeeks (commented):
Looks great - a couple of comments =)

kit/Delta.hpp
OFFTKP: we do - the one I added to DeltaTests tests exactly this - should fail without this change in Delta.hpp and we do generate multi-row copies still, there's further details in the task report.

test/DeltaTests.cpp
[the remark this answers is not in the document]
OFFTKP: I think what is happening is as the png is being loaded into a byte array is loaded as BGRA. Without that change the test fails because the graphic is the wrong color (instead of light blue it becomes orange, the R and B values are swapped)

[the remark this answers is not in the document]
OFFTKP: I'll look into a better way to do this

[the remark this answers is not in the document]
mmeeks: Ah - fair cop - if the PNG is not what we expect then we should swap that first I guess before applying a delta; but I'm surprised; possibly we're used to BGRA or somethign internally and we didn't notice yet; I'd expect PNGs to be RGB(A) though. Thanks for chasing =)

mmeeks:
Cypress test failures seem unrelated =) Would be great to get this pushed before the release next week =)

kit/ChildSession.cpp
```diff
@@ -635,6 +636,9 @@ bool ChildSession::_handleInput(const char *buffer, int length)
         else if (tokens.equals(0, "geta11ycaretposition"))
         {
             return getA11yCaretPosition();
+        else if (tokens.equals(0, "toggletiledumping"))
+        {
+            setDumpTiles(tokens[1] == "true");
```
mmeeks, line 642: Again - it cannot be that a client can swamp a server with this request in a deployed system. We need to guard this with #if ENABLE_DEBUG  =) thanks !

kit/Delta.hpp
mmeeks: fair enough.

mmeeks (changes requested):
Seems quite similar to the last commit ? =)

kit/ChildSession.cpp
OFFTKP: Now setting dumpTiles to false during the tile dumping function

OFFTKP:
Since we're in a hurry to push this, I say we do. There's still the BGR to RGB swapping we do in the tests that I need to figure out but otherwise the bug is fixed now in dark mode with this patch.

mmeeks (approved):
Thanks Paris ! =) we should also test with a calc sheet that has grid-lines turned off; it's possible some of the default color stuff is associated with that.

Issue #3216, opened 2021-09-11
Dunging out 


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @mmeeks 

### Detailed description and rationale
We have a lot of duplication in Makefile.am around the run: rules to take just two for example:

```make
run: setup-wsd
	./coolwsd --o:sys_template_path="@SYSTEMPLATE_PATH@" \
			  --o:security.capabilities="$(CAPABILITIES)" \
			  --o:child_root_path="@JAILS_PATH@" --o:storage.filesystem[@allow]=true \
			  --o:ssl.cert_file_path="$(abs_top_srcdir)/etc/cert.pem" \
			  --o:ssl.key_file_path="$(abs_top_srcdir)/etc/key.pem" \
			  --o:ssl.ca_file_path="$(abs_top_srcdir)/etc/ca-chain.cert.pem" \
			  --o:admin_console.username=admin --o:admin_console.password=admin \
			  --o:logging.file[@enable]=${OUTPUT_TO_FILE} --o:logging.level=trace \
			  --o:trace_event[@enable]=true

if ENABLE_DEBUG
run-one: setup-wsd
	./coolwsd --o:sys_template_path="@SYSTEMPLATE_PATH@" \
			  --o:security.capabilities="$(CAPABILITIES)" \
			  --o:child_root_path="@JAILS_PATH@" --o:storage.filesystem[@allow]=true \
			  --o:ssl.cert_file_path="$(abs_top_srcdir)/etc/cert.pem" \
			  --o:ssl.key_file_path="$(abs_top_srcdir)/etc/key.pem" \
			  --o:ssl.ca_file_path="$(abs_top_srcdir)/etc/ca-chain.cert.pem" \
			  --o:admin_console.username=admin --o:admin_console.password=admin \
			  --o:logging.file[@enable]=true --o:logging.level=trace \
			  --singlekit
endif
```

All of the common parameters from these coolwsd invocations should be condensed into a single shared variable for all of these parameters to ease maintenance and simplify the script.

Thanks ! =)

pedropintosilva:
WRT https://github.com/CollaboraOnline/online/blob/master/Makefile.am

h20220025:
Hi @mmeeks, I'm interested in contributing to this issue, please assign me this.
Thank you

pedropintosilva:
[...]

PR #6482, opened 2023-05-31
Condense common parameters in Makefile.am for run and run-one rules

codewithvk:
Change-Id: Ie3027c63cc31eea2dea2cb0a997d296a24476581

* Resolves: #3216 
* Target version: master


mmeeks (approved):
Looks good to me - I think there is a lot more duplication we can remove here though =)

PR #6488, opened 2023-05-31
jsdialog: style Formula Wizard

eszkadev:
funcwiz


pedropintosilva (commented):
Thanks @eszkadev unfortunately this doesn't seem to fix it
- We still have labels being concatenated as if it was a whole paragraph
- We still get some content cropped horizontally 
- Some input fields are also not vertical aligned 

(tested on firefox and chrome) 

I have pushed a different PR for this, I hope that's ok
https://github.com/CollaboraOnline/online/pull/6498

eszkadev:
ok, so closing this

PR #6496, opened 2023-06-01
fix worng image path name passing to setImage url

Darshan-upadhyay1110:
Change-Id: I58dab34e1dbe0811844a01aaa54bd889a2e6d033

* Resolves: # 
* Target version: master 

### Summary
In Mobile view we are getting error related to image icon path not found. this is because section id is empty and we are tring to set value of based on that.
image

here i am posting  the fix for that :)


eszkadev (approved):
good catch!

PR #6515, opened 2023-06-02
private/pedro/improve and fix search inputfield

pedropintosilva:
- Status bar: increase search width
- Status bar: add missing focus state for search input field


pedropintosilva:
these changes were already tested in the past and does not affect any build process

PR #6516, opened 2023-06-02
private/pedro/navigator fixalignment

pedropintosilva:
- Navigator: do not resort to absolute position to set deck
- Navigator: Fix alignment to coincide w/sidebar properties expanders

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -358,6 +358,8 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	color: var(--color-text-lighter) !important;
 }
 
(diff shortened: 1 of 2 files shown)


pedropintosilva:
@eszkadev I have tested and this seems to fix the additional top white space. Nevertheless I would love if you could review/look at it (because you might have had another reason to set it to absolute position that I somehow didn't realized

eszkadev (approved):

eszkadev:
I tried to extend it using `pos: absolute` and `left: 0, right: 0, top: 0, bottom:0`, that was why I used it. But with this change it looks fine too

PR #6517, opened 2023-06-02
Fit the images into the sidebar.

GulsahKose:
Without patch image sizes are 336px and seems cropped in sidebar. We should use same width with sidebar width.

Change-Id: I617a4e79e15d1d49ca95aec8f5a20f95734160a2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -11,6 +11,11 @@ img.sidebar.ui-image {
 	padding: 6px;
 }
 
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):

pedropintosilva:
no need to wait for cypress, specially when it was passing when it was cropped

PR #6523, opened 2023-06-04
docker: 23.05 image based on DebianSlim

timar:
This new DebianSlim dockerfile eliminates the use of installer script which was an anti-pattern.
The inotifywait hook for watching changing coolwsd.xml was removed. Was it useful? I don't know. Please complain, if you miss it. DebianSlim is still not "distroless". Probably first we should solve the name resolution problem in jails, in order to eliminate the need of copying /etc/hosts and /etc/resolv.conf to systemplate at startup.

Change-Id: I26948eb21a1bc3092b93d02d93e03269e9a70198

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
**Note:** Currently only CODE repos are present for 23.05. 
FYI @ralfbecker

ralfbecker:
Hi Timar,
the inotifywait, which was added by me, was useful: it allows EGroupware to update the Collabora configuration, eg. add the support key, and CO would restart without EGroupware needing (the rights) to talk to docker, or the user to manually restart CO after saving the config.
Did you removed it, because you did not know, or was inotifywait a problem somewhere?
Ralf

timar:
@ralfbecker I'll add it back. The problem is that there are customers who are pushing us towards a "distroless" concept, and inotifywait does not fit into this concept. Maybe we will need different dockerfiles for different customers. Thanks for your comment.

ralfbecker:
I only need that in the image with support-key, which we use for on premise installations. 
Thought a couple of days ago, a use complained the CODE did not restart automatically, after updating the config via EGroupware.
Ralf

PR #6528, opened 2023-06-05
Dark mode: Canvas, get directly the css var property value

pedropintosilva:
- Get directly the particular css var that we want instead of on relying on a container
  - This makes things easier to maintain in the future
- Use a new --color-canvas instead of relying on --color-main-background
  - color-main-background is used in many other places which means
  that if we want to have some particular UI brighter or darker it
  would affect the canvas. Better to use a different/new css var for that

Change-Id: I3d7d3ba188e96080d9f6f814a5c46f85ea378142


pedropintosilva:
no need to wait for any test other than codeQL for this specific commit

Issue #6404, opened 2023-05-23
23.05 - Pressing ENTER key does not confirm dialog entries


Ezinnem:
**Describe the bug**
Pressing ENTER key does not confirm dialog entries.

**To Reproduce**

1. Open a .odt file in Online 23.05 build
2. Add some text. select the text
3. Go to Insert > Hyperlink. Add the link to the hyperlink dialog
4. Press Enter

Notice that the dialog dismisses, but the link is not added to the text.

**Expected behavior**
ENTER key should confirm the dialog entries

**Actual behavior**
ENTER key does not confirm dialog entries

**Screenshots**

https://github.com/CollaboraOnline/online/assets/54344301/e8ef1f15-a4c2-4430-967c-78ecafe2d153

**Desktop (please complete the following information)**
COOLWSD version:
23.05.0.1snapshot git hash: [012b00d](javascript:void(window.open('https://github.com/CollaboraOnline/online/commits/012b00d'));)(E)

codewithvk:
Can i work on it?

gitgt:
@codewithvk you are welcome.

codewithvk:
@gitgt Thanks :) I want to give it try by my own, If i will get stuck at any point, I'll ask you about code pointer/help.

You can assign this issue to me :)

codewithvk:
Hey @gitgt, I figured out that when we press enter after opening the HyperLink dialog, the default target element is the clipboard.
![image](https://github.com/CollaboraOnline/online/assets/61119120/7416c1fb-3a40-4eb2-830c-3ac9b7f36bf8)

I tried to modify the behavior by working with Control.JSDialog.js -> onJSDialog. Specifically, I attempted to bind a function using the following syntax: `$('#hyperlink-link-box').off('keypress', this.onDocumentNameKeyPress);`. However, the function did not get triggered. I'm currently stuck and seeking guidance on how to approach this problem. I would appreciate your assistance.

thebearon:
This is a regression from fd71e664b6ffdef2eeb4948fa2b208b517d1f1d8:
```
Author:     Szymon Kłos <szymon.klos@collabora.com>
AuthorDate: Wed May 10 10:23:48 2023 +0200
Commit:     Szymon Kłos <eszkadev@gmail.com>
CommitDate: Mon May 15 07:55:22 2023 +0200

    jsdialog: fix jumping ans dragging Page Style dialog
```

CC @eszkadev 

[...]

PR #6533, opened 2023-06-05
jsdialog: default button inside form #6404

eszkadev:
default button should be inside form so we execute correct action on ENTER

fixes: https://github.com/CollaboraOnline/online/issues/6404


eszkadev:
all tests passes apart of navigator ones which fail on CI in other PR:
draw/navigator_spec.js
impress/navigator_spec.js
writer/navigator_spec.js

gokaysatir (approved):

Issue #6527, opened 2023-06-05
Canvas tiles color do no change when clicking the dark mode toggle


pedropintosilva:
**Describe the bug**
Background surrounding the documents doesn't change the color when clicking the dark mode toggle

This regression started with 
[--color-main-background CSS variable will be used for canvas backgrou… #6451](https://github.com/CollaboraOnline/online/pull/6451)

**To Reproduce**
1. Open a file
2. change to dark mode by pressing the toggle button under view tab
3. notice the background didn't change to the `-color-main-background` value
4. also noticed that when refreshing the page it this change it correctly

**Mentors**
@gokaysatir 

**CC**
@OFFTKP

OFFTKP:
Note: If you set `useCSSForBackgroundColor` to false, it seems to work as normal last time I tried.

Darshan-upadhyay1110:
Yes i also noticed

Darshan-upadhyay1110:
but if we pass false to every case than that will not work for branding css variables so i am finding more genric way to handle this

PR #6535, opened 2023-06-05
Fix Canvas tiles color do no change when clicking the dark mode toggle

Darshan-upadhyay1110:
Change-Id: Ifdb5bbf06cb2ed5b8aa1e03ddb6796a0bc80b1c8

* Resolves: #6527 
* Target version: master 

### Summary

Added css variable condition in setBackgroundColorMode which was not used in any part of online code. this will work for both the case the normal one and for css variables if we pass as url parms

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -74,16 +74,27 @@ L.Control.UIManager = L.Control.extend({
 
 	loadLightMode: function() {
 		document.documentElement.setAttribute('data-theme','light');
+		this.setCanvasColorAfterModeChange();
 		this.map.fire('darkmodechanged');
 	},
 
 	loadDarkMode: function() {
 		document.documentElement.setAttribute('data-theme','dark');
+		this.setCanvasColorAfterModeChange();
 		this.map.fire('darkmodechanged');
 	},
 
 	getDarkModeState: function() {
 		return this.getSavedStateOrDefault('darkTheme', window.uiDefaults['darkTheme'] ?  window.uiDefaults['darkTheme'] : false);
+	},
+
+	setCanvasColorAfterModeChange: function() {
+		if (app.sectionContainer) {
+			app.sectionContainer.setBackgroundColorMode(false);
+			app.sectionContainer.setClearColor(window.getComputedStyle(document.documentElement).getPropertyValue('--color-canvas'));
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
Thanks :) but I don't understand this PR. 

The initial problem was: the css var was only getting picked up by CanvasSectionContainer.ts on loading (when initially opening a file and when reloading). So, it was working for those cases1. But, it was not working when clicking the toggle button.
 - So the fix should be not to disable the reading of the css var but instead make sure we also re-read it and re-paint canvas when the user clicks tha toggle

1. To test this you can set NC settings to dark mode and see it in action.

Darshan-upadhyay1110:
@pedropintosilva i am understading what your are trying to say but if you notice in setClearColor there is condition for useCssBackgroundColor which is always remains true and that will not allow to set color for canvas when we toggle to dark 

image


here the method on top `setBackgroundColorMode` is never been called from any file so the var useCSSForBackgroundColor remains true forever.

Darshan-upadhyay1110:
One other way we can do is to just set 

Current 
`this.context.fillStyle =  this.clearColor ( here clear color is not being set properly because of situation i mentioned above )

so we can do this by just read value every time we clean canvas 

My suggestion 
this.context.fillStyle =  window.getComputedStyle(document.documentElement).getPropertyValue('--color-canvas');

Darshan-upadhyay1110:
The last change i pushed id working in both case plus i tested wuth NC settings works with that as well. Maybe it is okay if we can re read clear color value after re paint the canvas. 

image
image

@pedropintosilva ???

gokaysatir (changes requested):
Thank you @Darshan-upadhyay1110 for the PR. It's a nice catch.

I think we can use setClearColor for changing the background color of the canvas. clearCanvas is called for every draw call and i don't know if it would be good to set the variable there for every call.

Maybe we can add 2-3 lines into "Control.UIManager.js" to the function "loadDarkMode" to set the canvas background via setClearColor function?

You'll need to call setBackgroundColorMode function twice there, but it may be worth to do.

Thank you.

pedropintosilva (approved):
Thanks, the toggle works and uses the css var!
It also changes to dark mode automatically set (tested with integrator: NC settings set to dark, no cache)

PR #6538, opened 2023-06-06
Mobile: About dialog: remove unnecessary elements

pedropintosilva:
Change-Id: Id929475af828f6bddf4abaadea443dd3f707af4c


pedropintosilva:
no need to wait for any tests other than the regular CodeQL. This is CSS only changes plus it removes elements that used to be already removed in the past

Issue #6493, opened 2023-06-01
Dialogs are narrow


thebearon:
Open dialogs, eg. Format -> Character or Page Style in Writer.
Note how the dialogs are rather narrow, they could be much wider so they aren't crowded.

On a sidenote, it isn't obvious those light button-like items are tab headers.

![character](https://github.com/CollaboraOnline/online/assets/19438782/4e235da5-3acf-407a-a204-6addc4e069ac)
![page style](https://github.com/CollaboraOnline/online/assets/19438782/a256dab3-fcdd-4d1f-8aef-9f42cb0273ff)

thebearon:
CC @eszkadev @pedropintosilva

PR #6541, opened 2023-06-06
Expand the Page style and Character dialogs #6493

GulsahKose:
Change-Id: I5442d6a99bb202ad2a850ac42baf2140adc4d1a2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1448,3 +1448,11 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #FormulaDialog #box3 {
 	display: grid;
 }
+
+div [id^='TemplateDialog'] {
+	min-width: 590px;
+}
+
+#CharacterPropertiesDialog {
+	min-width: 656px;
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Thanks I was thinking we could target just the beginning of that id e.g.: [id^='TemplateDialog'] 

... but maybe better not since it seems quite generic (it even doesn't have page or page layout in the name) and then it can affect something else in the future.

browser/css/jsdialogs.css
```diff
@@ -1448,3 +1448,11 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #FormulaDialog #box3 {
 	display: grid;
 }
+
+#TemplateDialog8 {
```
eszkadev, line 1452: are you sure this '8' isn't something automatically added? (we add numbers to avoid duplication, maybe will be better to use "starts with TemplateDialog" ?

PR #6544, opened 2023-06-06
jsdialog: fix initial focus for combobox

eszkadev:
combobox has inner < select > html element which should be focused instead of main container

this can be tested in Insert -> Page Number in Writer


gokaysatir (approved):

eszkadev:
locally tests are ok

Issue #6381, opened 2023-05-18
"Help > About" Copy Button


Tex2002ans:
**Is your feature request related to a problem? Please describe.**

Right now, it's hard to copy/paste out of Collabora Online's:

- **Help > About** page.

Making this easier would:

- Get more users to include/post their info

for:

- bug reports
  - Github, Forum posts
- when they need help
  - reporting bugs, configuration issues, missing features, etc.

I think it indirectly would make QA/debugging much easier too, because many users:

- Currently take screenshots instead.
- Don't include the info at all.
   - (And this would help have consistent formatting of About text too! With exact git commits, etc.)

**Describe the solution you'd like**

Add a:

- Copy button

similar to LibreOffice's **Help > About LibreOffice**:

![LibreOffice 7 5 3 - Help About Info - Copy Button](https://github.com/CollaboraOnline/online/assets/5774324/b5d4ba75-c5e9-46db-b130-ee8eb06de5c2)

**Describe alternatives you've considered**

Making the formatting better when you:

- Highlight the text in the "About" window.
- `Ctrl+C` + `Ctrl+V`

Right now, it outputs in an ugly text too (with odd linebreaks, etc.).

Hopefully the button will do much better + make this much easier though. :)

pedropintosilva:
additional iteration https://github.com/CollaboraOnline/online/issues/6764

PR #6549, opened 2023-06-06
Added copy button in "Help > About" to copy version information to clipboard

rparth07:
Change-Id: I2ce4d5fe03ac3c38ff179cbd2466ba25961db92d

* Resolves: #6381
* Target version: master 

### Summary

- Added a copy button in "Help > About" to copy version related information to clipboard

### Checklist

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -656,10 +656,46 @@ L.Map.include({
 		box.insertBefore(innerDiv, box.firstChild);
 		innerDiv.innerHTML = content.outerHTML;
 
-		var form = document.getElementById('modal-dialog-about-dialog-box');
+		var form = document.getElementById('about-dialog-box');
+
 		form.addEventListener('click', this.aboutDialogClickHandler.bind(this));
 		form.addEventListener('keyup', this.aboutDialogKeyHandler.bind(this));
 		form.querySelector('#coolwsd-version').querySelector('a').focus();
+		var copyversion = L.DomUtil.create('button', 'ui-pushbutton jsdialog', null);
+		copyversion.setAttribute('id', 'modal-dialog-about-dialog-box-copybutton');
+		copyversion.innerHTML = 'Copy version';
+		copyversion.addEventListener('click', this.copyVersionInfoToClipboard.bind(this));
+		var aboutok = document.getElementById('modal-dialog-about-dialog-box-yesbutton');
+		if (aboutok) {
+			aboutok.before(copyversion);
+		}
+	},
+
+	getVersionInfoFromClass: function(className) {
+		var versionElement = document.getElementById(className);
+		var versionInfo = versionElement.innerText;
+
+		var gitHashIndex = versionInfo.indexOf('git hash');
+		if (gitHashIndex > -1) {
+		  versionInfo = versionInfo.slice(0, gitHashIndex) + '(' + versionInfo.slice(gitHashIndex) + ')';
+		}
+
+		return versionInfo;
+	},
+
+	copyVersionInfoToClipboard: function() {
+		var text = 'COOLWSD version: ' + this.getVersionInfoFromClass('coolwsd-version') + '\n';
+		text += 'LOkit version: ' + this.getVersionInfoFromClass('lokit-version') + '\n';
+		text += 'Served by: ' + document.getElementById('os-info').innerText + '\n';
+		text += 'coolwsd id: ' + document.getElementById('coolwsd-id').innerText + '\n';
+
+		navigator.clipboard.writeText(text)
+		.then(function() {
+		  window.console.log('Text copied to clipboard');
+		})
+		.catch(function(error) {
+		  window.console.error('Error copying text to clipboard:', error);
+		});
 	},
 
 	extractContent: function(html) {


rparth07:
> Oh we really needed this, Thanks!
> 
> Could you please change a bit the add html? I would prefer to avoid adding as much elements as possible. Specially because it can be hard on small screens and on mobile devices:
> 
> * So no additional heading please
> 
> Additionally the icon button doesn't look good on mobile and can be also misleading on desktop (people might think it will copy just the COOLWSD version
> 
> So I would propose to just add an additional push button before the ok button. This way, it will look good on desktop and mobile while not adding any additional vertical space:
> 
> ```diff
> diff --git a/browser/src/control/Toolbar.js b/browser/src/control/Toolbar.js
> index 995cc35f40..8cd8040814 100644
> --- a/browser/src/control/Toolbar.js
> +++ b/browser/src/control/Toolbar.js
> @@ -660,6 +660,13 @@ L.Map.include({
>  		form.addEventListener('click', this.aboutDialogClickHandler.bind(this));
>  		form.addEventListener('keyup', this.aboutDialogKeyHandler.bind(this));
>  		form.querySelector('#coolwsd-version').querySelector('a').focus();
> +		var copyversion = L.DomUtil.create('button', 'ui-pushbutton jsdialog', null);
> +		copyversion.setAttribute('id', 'modal-dialog-about-dialog-box-copybutton');
> +			copyversion.innerHTML = 'Copy version';
> +		var aboutok = document.getElementById('modal-dialog-about-dialog-box-yesbutton');
> +		if (aboutok) {
> +			aboutok.before(copyversion);
> +		}
>  	},
>  
>  	extractContent: function(html) {
> ```

Thanks, @pedropintosilva for reviewing, I updated the code, can you please review the changes?

pedropintosilva (changes requested):
I'm getting (on mobile) Uncaught TypeError: `form` is null  it seems we should be checking if that exists before adding listeners.

browser/src/control/Toolbar.js
```diff
@@ -660,6 +660,41 @@ L.Map.include({
...
+		var aboutok = document.getElementById('modal-dialog-about-dialog-box-yesbutton');
+		if (aboutok) {
+			aboutok.before(copyversion);
+		}
+	},
+
+	getVersionInfoFromClass: function(className) {
+		var versionElement = document.getElementById(className);
+		var versionInfo = versionElement.innerText;
+
```
pedropintosilva, line 676: Please remove the empty tab here

browser/src/control/Toolbar.js
```diff
@@ -660,6 +660,41 @@ L.Map.include({
...
+	getVersionInfoFromClass: function(className) {
+		var versionElement = document.getElementById(className);
+		var versionInfo = versionElement.innerText;
+	  
+		var gitHashIndex = versionInfo.indexOf('git hash');
+		if (gitHashIndex > -1) {
+		  versionInfo = versionInfo.slice(0, gitHashIndex) + '(' + versionInfo.slice(gitHashIndex) + ')';
+		}
+
```
pedropintosilva, line 681: Please remove the empty tab here

browser/src/control/Toolbar.js
```diff
@@ -660,6 +660,41 @@ L.Map.include({
...
+	  
+		return versionInfo;
+	},
+
+	copyVersionInfoToClipboard: function() {
+		var text = 'COOLWSD version: ' + this.getVersionInfoFromClass('coolwsd-version') + '\n';
+		text += 'LOkit version: ' + this.getVersionInfoFromClass('lokit-version') + '\n';
+		text += 'Served by: ' + document.getElementById('os-info').innerText + '\n';
+		text += 'coolwsd id: ' + document.getElementById('coolwsd-id').innerText + '\n';
+
```
pedropintosilva, line 690: Please remove the empty tab here

Rash419 (changes requested):
update the id on line number 659 to `var form = document.getElementById('about-dialog-box');` it will fix form null error on mobile

rparth07:
Thanks, @Rash419. Fixed form null error on mobile and removed the tab.

Rash419:
@rparth07 please squash the commits : https://www.freecodecamp.org/news/git-squash-commits/

Rash419 (approved):
Thanks, Looks good to me!

Rash419:
@rparth07 can you also rebase your PR please ? Lets wait for CI to pass

browser/src/control/Toolbar.js
```diff
@@ -656,10 +656,46 @@ L.Map.include({
...
 
-		var form = document.getElementById('modal-dialog-about-dialog-box');
+		var form = document.getElementById('about-dialog-box');
+
 		form.addEventListener('click', this.aboutDialogClickHandler.bind(this));
 		form.addEventListener('keyup', this.aboutDialogKeyHandler.bind(this));
 		form.querySelector('#coolwsd-version').querySelector('a').focus();
+		var copyversion = L.DomUtil.create('button', 'ui-pushbutton jsdialog', null);
+		copyversion.setAttribute('id', 'modal-dialog-about-dialog-box-copybutton');
+		copyversion.innerHTML = 'Copy version';
```
pedropintosilva, line 666: 'Copy version' should be _('Copy version') so it can be picked up by translations but other than that this looks great so, let's merge it + if the tests passed) and you can create a follow up PR for that string ;)

pedropintosilva (approved):

pedropintosilva:
Cypress and unit tests passed, another thing we can do in that follow up @rparth07 would be to add cypress test for that :) thanks !

timar:
See also PR#6754
I have a question though:
I noticed that navigator.clipboard can be undefined, c.f. https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined
On the other hand we can always copy text from a document. How is that possible? Can we do the same here?
If not, what would be the best? Hide the button, when navigator.clipboard is undefined?

Tex2002ans:
Hey @rparth07,

Thanks so much for adding the copy button! I'm the one who first requested it!

I noticed there is an NBSP ("non-breaking space"):

- AFTER `hash:`
   - Right after the colon.
- BEFORE the actual hash number

- - -

Example of...

Current copy WITH NBSP:

> COOLWSD version: 23.05.1.2 (git ***hash: 6f34565a***)
> LOKit version: Collabora Office 23.05.1.2 (git ***hash: 9a20ef2***)
> Served by: Linux Mint 21.1
> Server ID: c150413d

Same text WITHOUT NBSP:

> COOLWSD version: 23.05.1.2 (git hash: 6f34565a)
> LOKit version: Collabora Office 23.05.1.2 (git hash: 9a20ef2)
> Served by: Linux Mint 21.1
> Server ID: c150413d


- - -

I know that NBSP can sometimes make trouble in some programs when finding/searching (since it could be treated different from a SPACE), so...:

- Could that copy text be tweaked to just use a normal SPACE?

rparth07:
> Hey @rparth07,
> 
> Thanks so much for adding the copy button! I'm the one who first requested it!

@Tex2002ans, Thanks for the idea 🎉

> Could that copy text be tweaked to just use a normal SPACE?

Sure, we can identify and remove the NBSP, currently I'm working on writing cypress test for it, once finished, I will create an PR containing cypress test and that fix changes.

PR #6550, opened 2023-06-06
impress: introduce slide hiding/showing in presentation

lpranam:
Change-Id: If91726a1a66e1679b58603e496a5b90cde5f0b5b

* Target version: master 

### Checklist


eszkadev (approved):
code looks good

browser/src/control/Control.Menubar.js
```diff
@@ -1681,6 +1684,16 @@ L.Control.Menubar = L.Control.extend({
...
+					} else if (id === 'showslide') {
+						if (!self._map._docLayer.isHiddenSlide(self._map.getCurrentPartNumber()))
+							$(aItem).addClass('disabled');
+						else
+							$(aItem).removeClass('disabled');
+					} else if (id === 'hideslide') {
+						if (self._map._docLayer.isHiddenSlide(self._map.getCurrentPartNumber()))
+							$(aItem).addClass('disabled');
+						else
+							$(aItem).removeClass('disabled');
```
pedropintosilva, line 1696: Can we please instead of adding/removing disable css class just show/hide the specific menu item? So instead of seeing 2 options (hide slide and show slide) just one of those:

image

```diff
-                                                       $(aItem).addClass('disabled');
+                                                       $(aItem).hide();
-                                                       $(aItem).removeClass('disabled');
+                                                       $(aItem).show();
```

browser/src/control/Control.NotebookbarImpress.js
```diff
@@ -428,6 +428,34 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
...
+								'id': 'hideslide',
+								'type': 'menubartoolitem',
+								'text': _UNO('.uno:HideSlide', 'presentation'),
+								'command': '.uno:HideSlide'
+							}
+						]
+					}
+				],
+				'vertical': 'true'
+			}
```
pedropintosilva, line 458: Could we please use `'type': 'bigtoolitem'` instead for each one and show just one each time (depending on if isHiddenSlide)

pedropintosilva (changes requested):
I will then add two icons

pedropintosilva (changes requested):
Thanks @lpranam I have also added the 2 missing icons.

Still, when using tabbed view (notebookbar)
* Both buttons are visible
* Both send the uno command (even if the slide is already hidden/shown)

This is possible to reproduce in 2 case scenarios:
* Open a presentation that was saved beforehand with the one of the slides hidden (e.g. the first slide)
  * 1st slide is selected -> both buttons are visible
* Open a presentation without any hidden slide
  * Both icons are visible 

Upon pressing (calling the toggle) it hides on of them. Could you please not show both by default when using tabbed view? (we already have that `visible` option that might come in handy, e.g.: `'visible': isODF ? 'true' : 'false',` but up to you)

pedropintosilva (approved):
Thanks!

pedropintosilva:
Note to myself: core patch https://gerrit.libreoffice.org/c/core/+/152688

PR #6553, opened 2023-06-07
a11y: add config for enable/disable accessibility in general

mcecchetti:
A new section about accessibility has been appended to coolwsd.xml
config file

Change-Id: I086abdf73646639283eb655ae60f200fb64e495a


.gitpod-files/coolwsd-gitpod.xml
```diff
@@ -3,6 +3,10 @@
     <!-- Note: 'default' attributes are used to document a setting's default value as well as to use as fallback. -->
     <!-- Note: When adding a new entry, a default must be set in WSD in case the entry is missing upon deployment. -->
 
+    <accessibility desc="Accessibility settings">
+        <enable type="bool" desc="Controls whether accessibility support should be enabled or not." default="false">true</enable>
```
timar, line 7: We do not use this file in CI or in production. You can leave the setting `false`.

cypress_test/Makefile.am
```diff
@@ -154,6 +154,7 @@ define start_coolwsd
 			--o:logging.file[@enable]=true --o:logging.level=trace \
 			--o:welcome.enable=false \
 			--o:user_interface.mode=$(USER_INTERFACE) \
+	        --o:accessibility.enable=true \
```
timar, line 157: Wrong indentation.

timar (changes requested):
I think you need to add the default value to `wsd/COOLWSD.cpp` to `DefAppConfig`.

.gitpod-files/coolwsd-gitpod.xml
mcecchetti: Okay, anyway I think it doesn't make any difference since it's pretty ignored

cypress_test/Makefile.am
mcecchetti: Ah! In the file are used tabs in place of spaces. That's unusual. Anyhow, I fixed it.

mcecchetti:
> I think you need to add the default value to `wsd/COOLWSD.cpp` to `DefAppConfig`.

Okay, thanks. Added!

timar (approved):

mcecchetti:
Well Cypress test are passed, I think we can skip to wait for Fast Build against Core co-23.05

PR #6554, opened 2023-06-07
WIP: color palette with document theme

eszkadev:


eszkadev:
passed, only broken test failed: calc/formulabar_spec.js

PR #6555, opened 2023-06-07
DeltaTests::testDeltaCopyOutOfBounds fails in a build from make dist …

caolanm:
…tarball

because data/delta-graphic.png and data/delta-graphic2.png are missing

Change-Id: Ie850d9ae76946e891d2928de80c5ded4e970f4cf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (approved):
Thank you! We used to have checks for `make dist` and make from tarball, I have to investigate if we dropped that check for CI or someone ignored the fail and merged.

PR #6558, opened 2023-06-07
A11TextInput: clipboard area stolen focus from text search field

mcecchetti:
When a key was typed the focus was grabbed from the clipboard area
because of caret/text selection events received from core.

Change-Id: I60b945d7d4742faf4a946c836628bdf70b266eb7


gokaysatir (approved):
Thanks for the fix :)

PR #6561, opened 2023-06-08
Disable HTML dragging on navigator until the feature is activated.

gokaysatir:
Change-Id: I110f4efd3ff24b3f38a5756432b5847f0a852832

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -364,9 +364,6 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	width: var(--btn-size-s);
 	height: var(--btn-size-s);
 }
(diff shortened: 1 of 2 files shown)


pedropintosilva (approved):

pedropintosilva (changes requested):
Hm I'm afraid this won't do as we do need them to be draggable in other places .e.g Pivot tablet layout dialog (when you create a pivot table) . So this should only be applied to treeview with navigator id

pedropintosilva (approved):

pedropintosilva:
let's let the JS codeQL scanning finish

PR #6564, opened 2023-06-08
kit: fix missing init of _isDumpingTiles in ChildSession

vmiklos:
This was forgotten in commit 21966e1a9c2379214d3bef2dfb93c7e7ee32ae82
(Fix copy command going out of bounds during delta creation,
2023-05-31).

Change-Id: I91ce26f46cc6cb7a2dc3ab0665dc9aeea8a5c00f

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -108,7 +108,8 @@ ChildSession::ChildSession(
     _viewId(-1),
     _isDocLoaded(false),
     _copyToClipboard(false),
-    _canonicalViewId(-1)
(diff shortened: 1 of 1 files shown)


vmiklos:
@caolanm I guess long-term it would be interesting to try Coverity on Online, I found this with a local sanitizer-enabled 'make check' in online.git.

caolanm (approved):
yeah, make sense. I think there is some github + coverity integration option too

OFFTKP:
Yup good catch

Issue #6507, opened 2023-06-02
Missing icon "Open in local editor"


ia-bg:
**Describe the bug**
The icon "Open in local editor" is missing.

**To Reproduce**
1. Nextcloud 26.0.2
2. Nextcloud Office App  8.0.2
3. collabora/code 22.05.14.3
4. Open file in Nextcloud
5. See error

**Expected behavior**
There is a icon

**Actual behavior**
There is a wrong background image URL `url("https%3A%2F%2FREDACTED_DOMAIN%2Fcustom_apps%2Frichdocuments%2Fimg%2Flaunch.svg") no-repeat center !important;`

**Screenshots**
![2023-06-02-09:35:32](https://github.com/CollaboraOnline/online/assets/102674963/bd5c8d13-33a0-4be0-b8b3-4c3d9fea227f)

**Desktop (please complete the following information)**
 - OS: Windows 10 and Linux
 - Browser: Firefox, Chromium, Edge

pedropintosilva:
found in https://github.com/nextcloud/richdocuments/blob/b77f855ded8ad94990d116e9a84229e9c999a8bf/src/document.js#L330

Interestingly enough when using tabbed view it works and I see `http://localhost/nc-master/apps/richdocuments/img/launch.svg` but when switching to compact view I see what is reported. Looks like an encoding problem

gitgt:
Fixed with https://github.com/CollaboraOnline/online/commit/902fedf65545666a1e84ca5476cc9c9b0aa6cb51 for 23.05.
Hopefully it will be also in 22.05.
I close now with thanks for reporting.

PR #6567, opened 2023-06-08
Fix Missing icon "Open in local editor" #6507

pedropintosilva:
It seems we were encoding even when setting the url via CSS.

Tested using richdocuments
Tested using browser/html/framed.doc.html

Context:
Open in local editor is insert via
https://github.com/nextcloud/richdocuments/blob/b77f855ded8ad94990d116e9a84229e9c999a8bf/src/document.js#L330
using https://sdk.collaboraonline.com/docs/postmessage_api.html?highlight=insert_button#id5

Change-Id: I526a5f476e90dbbc52640bd6d76f416a654a20a5

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -121,7 +121,7 @@ L.Control.UIManager = L.Control.extend({
 			this.refreshMenubar();
 			this.refreshToolbar();
 		}
-		else { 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
no problem :) and thanks @Rash419 for force-pushing, better now to use `encodeURI` instead of nothing at all

PR #6585, opened 2023-06-09
perf: avoid creating strings by using string_view

caolanm:
Change-Id: Ibffa35d40ee9839635cbd37be8cdc9c69b834504

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks good, thanks.

Nit: I assume you did some measuring that this actually helps, next time it would be a good idea to include some before/after numbers in the commit message.

PR #6589, opened 2023-06-09
Fix iOS build errors

plubius:
Online needs a lot of detailed logging but mobile doesn't run a server so just bypass those calls on mobile.

### DIFF
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -255,12 +255,14 @@ void Session::handleMessage(const std::vector<char> &data)
     try
     {
         std::unique_ptr< std::vector<char> > replace;
+#if defined ENABLE_DEBUG && !MOBILEAPP
         if (!Util::isFuzzing() && UnitBase::get().filterSessionInput(this, &data[0], data.size(), replace))
         {
             if (!replace || replace->empty())
                 _handleInput(replace->data(), replace->size());
             return;
         }
+#endif
 
         if (!data.empty())
             _handleInput(&data[0], data.size());
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -45,7 +45,7 @@ private:
     const bool _isClient;
 
     // Last member.
-#ifdef ENABLE_DEBUG
+#if defined ENABLE_DEBUG && !MOBILEAPP
     /// The UnitBase instance. We capture it here since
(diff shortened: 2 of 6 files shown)


plubius:
iOS (and maybe Android) pull in a subset of the much more complex Online server code so this is just a periodic update for mobile.

mmeeks (changes requested):
Heh ;-) my goal is to reduce this metric:

$ git grep 'MOBILEAPP' | wc -l
339

This goes in the wrong direciton I think by adding hundreds more - a reasonable number is in the low 10s I think =)

If we don't need logging on iOS - then we should have perhaps one more ifdef in the logging code. Or we should stub that whole chunk of code with no-ops on mobile: in one place, and not everywhere.

Then as we add more logging and/or unit testing left and right - we will not have to do anything to keep iOS working. I'd really like us to get to having a much easier to maintain app there =)

Does that make sense ?

plubius:
All your ideas make sense. However, I am not terribly interested in such code cleanup. I just wanted to get something out there so that I could build and run on iOS and try out your canvas memory fix. No need to commit this pull request if you don't like it. It's good enough for me to work with.

Note: the vast majority of these #if !MOBILEAPP are due to a bunch of assertion failures in the unit testing classes. They never  get initialized on iOS or Android so testing classes crash when you open a document.

plubius:
One other important thing to note: AFAICT the code as is won't build on master if ENABLE_DEBUG is not defined. In its unpatched state, _unit in net/WebSocketHandler.hpp and _unitWsd in wsd/DocumentBroker.hpp are in #ifdef ENABLE_DEBUG blocks but then there are no such #ifdef ENABLE_DEBUG blocks around where they are used.

TL;DR even if you remove all my additional !MOBILEAPPs, you will still need to leave the #ifdef ENABLE_DEBUGs that I added or include _unit and _unitWsd data members in non-debug builds.

plubius:
> $ git grep 'MOBILEAPP' | wc -l 339

So I call BS on this statistic. Your statistic applies to the code *before* this patch so you are somehow I can't get iOS build fixes committed unless I cleanup the mess that @mmeeks and his staff created. Talk about moving the goalposts. :D

Closing this pull request as I have replaced this pull request with https://github.com/CollaboraOnline/online/pull/6625.

Issue #4792, opened 2022-05-19
Annotations, comments on hover


pedropintosilva:
**Describe the bug**
the comments on mouseover seem to have wrong styling:
![image](https://user-images.githubusercontent.com/65948705/169259074-a36b215f-02f0-453a-b155-5cf2ac991192.png)

Frédéric

**To Reproduce**
Steps to reproduce the behaviour:
1. make clean , make run
2. insert a comment
3. move the cursor above the comment when editing it

Andreas-Kainz:
What would be the correct styling? no blue border or a hover effect on blue buttons or another color for the input field?

thebearon:
Not only that, but when hovering over the Save button:
- the emphasized color is a washed blue, which doesn't really emphasize the focus,
- the pointer stays an arrow (when hovering over Cancel, it turns into a pointing hand).

ZainIftikhar7vals:
I want to contribute to this Easy Hack.
After building and running Writer in browser I added a Comment but  moving cursor above the comment when editing it doesn't change the Styling.
And Yes Hovering over the Save button the pointer stays an arrow. Should I fix Only this issue then?
![Screenshot from 2022-10-01 16-49-29](https://user-images.githubusercontent.com/88649907/193408179-849f12c0-12e7-4df3-acdb-fdddcf0ddaa0.png)

Ayushbajpai19:
I would like to take this issue for hacktoberfest

pedropintosilva:
Let's please ignore the save button for now

pedropintosilva:
Hi @ZainIftikhar7vals and sorry for the delayed response. I just tested and I can confirm that the card's background does change (when editing or otherwise). Could you please try again?

PR #6594, opened 2023-06-10
Fix: Annotations, comments on hover

codewithvk:
Change-Id: I3a1e9473cabc9df2b33e3b99168f9698d1beb86c

* Resolves: #4792 
* Target version: master 

### Summary
image


codewithvk:
Hey @pedropintosilva, thank you for your cool review. I think I understand from the issue that the hover effect is annoying, but currently, it looks awesome. When the comment is active, there is no hover effect, but otherwise, we do have hover colors. Thanks for the suggestion!

pedropintosilva (approved):
Thanks : )

PR #6595, opened 2023-06-10
Thread Sanitization

Ashod:
Some thread-safety fixes with the help of the thread sanitizer.

- wsd: signal handler cannot log
- wsd: test: better test timeout thread management
- wsd: thread-safe socket shutdown flag
- wsd: thread-safe kit pid
- wsd: signal handlers must preserve errno

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -431,17 +431,14 @@ namespace SigUtil
                 static constexpr auto msg =
                     "Crashed in unattended run and won't wait for debugger. Re-run without "
                     "--unattended to attach a debugger.";
-                LOG_ERR(msg);
-                std::cerr << msg << '\n';
+                std::cerr << msg << std::endl;
             }
             else
             {
                 signalLog(FatalGdbString);
-                LOG_ERR("Sleeping 60s to allow debugging: attach " << getpid());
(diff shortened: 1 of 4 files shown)


common/Unit.cpp
```diff
@@ -535,7 +541,7 @@ void UnitBase::endTest(const std::string& reason)
     _socketPoll->joinThread();
 
     // tell the timeout thread that the work has finished
-    TimeoutThreadMutex.unlock();
```
Ashod, line 538: Manually unlocking here was wrong because 1) we didn't take the lock and 2) the Mutex might already be unlocked.

Both cases are invalid.

mmeeks (commented):
All looks fine to me - another review needed before merge to master =)

common/SigUtil.cpp
```diff
@@ -295,6 +295,8 @@ namespace SigUtil
     static
     void handleTerminationSignal(const int signal)
     {
+        const auto onrre = errno; // Save.
```
vmiklos, line 298: Nice, joke, but perhaps go with something like oldErrno or origErrno to be more readable?

vmiklos (approved):
Looks OK to me, please wait for CI and then I'm fine to merge this. See inline comment for a nit. Thanks.

Issue #5473, opened 2022-11-15
Add Scrollbar onmouseover effect


pedropintosilva:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @eszkadev 

### Detailed description and rationale
Add onmouseover effect to scrollbar (vertical [and horizontal in the case of spreadsheets]):
* Increase scrollbar thickness more than what we have currently when onmouseover 
* decrease scrollbar thickness when **not** onmouseover

### Code pointers
* browser/src/layer/tile/ScrollSection.ts
* browser/src/layer/tile/CalcTileLayer.js

pedropintosilva:
I'm not able to assign this to you @SakshiT25 until you comment here, I think.

SakshiT25:
Hi, @pedropintosilva I would like to work on this issue.

pedropintosilva:
Hi @SakshiT25 , how is this coming along?

SakshiT25:
Hi @pedropintosilva, I'm working on this issue. I was occupied with some other task. I will try to update you by Thursday.

pedropintosilva:
awesome @SakshiT25 ! feel free to ping me or anyone else in the IRC/MAtrix/Telegram room or in the weekly community meeting https://collaboraonline.github.io/post/communicate/ :)

drugwho:
@pedropintosilva , if this issue isn't picked up, can I pick it up now?

rimuruweebb:
can you please assign this to me ? @pedropintosilva

rimuruweebb:
hello @pedropintosilva ,there's no onMouseFunction callback defined in the canvassectioncontainer.ts and do i have to make a function in canvas section and then implement it in scrollbarsection.ts ?

pedropintosilva:
I would say so (I have checked the existing `isMouseOnScrollBar` but that doesn't seem to help) but @eszkadev would probably know best

eszkadev:
maybe `onMouseEnter` and `onMouseLeave` are enough for this task?

pedropintosilva:
ah right it seems so :

```diff
@@ -632,7 +633,12 @@ export class ScrollSection extends CanvasSectionObject {
 		}
 	}
 
+	public onMouseEnter (): void {
+		console.debug('onMouseOver');
+	}
+
 	public onMouseLeave (): void {
+		console.debug('onMouseOut');
```

[...]

PR #6596, opened 2023-06-11
Add Scrollbar onmouseover effect

codewithvk:
Change-Id: If52d3b4b3afa8ad629adeeda6fd65c7a80776b06

* Resolves: #5473 
* Target version: master

### DIFF
--- a/browser/src/layer/tile/ScrollSection.ts
+++ b/browser/src/layer/tile/ScrollSection.ts
@@ -539,10 +539,22 @@ export class ScrollSection extends CanvasSectionObject {
 		}
 	}
 
+	private increaseScrollBarThickness () : void {
+		this.sectionProperties.scrollBarThickness = 8 * app.roundedDpiScale;
+		this.containerObject.requestReDraw();
+	}
+
+	private decreaseScrollBarThickness () : void {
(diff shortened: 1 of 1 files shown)


codewithvk:
@pedropintosilva Can you please review this PR or add appropriate reviewers?

pedropintosilva:
Hi @codewithvk and sorry for the delay, I missed notification for this PR. 

I will check it but @eszkadev could you please review it?

pedropintosilva (approved):
Thanks ! it works, The scrollbar gets thicker on hover. I wonder if we could do a follow up where we animate that increase

eszkadev (approved):

codewithvk:
> Thanks ! it works, The scrollbar gets thicker on hover. I wonder if we could do a follow up where we animate that increase

I think we can give try to JavaScript Animation. [Ref : JavaScript animations](https://javascript.info/js-animation) If it sound good, I will create PR for this fix.

Minion3665:
@codewithvk re:
> I think we can give try to JavaScript Animation using setTimeOut and timing functions. [Ref : JavaScript animations](https://javascript.info/js-animation) If it sound good, I will create PR for this fix.

yep, that sounds like it would work! go ahead

codewithvk:
Created Follow up PR for animation : #7218

PR #6600, opened 2023-06-12
cid#318846 Array compared against 0

caolanm:
array_null: Comparing an array to null is not useful: "false", since the test will always evaluate as true.

Change-Id: Ib34694022484e9041827980c0b67fe614446f4b9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
This looks like it was always wrong, but hopefully we don't rely on it being wrong

vmiklos (approved):
Looks good, we pass a boolean as a parameter at other places.

PR #6608, opened 2023-06-13
Don't let the open dialog if it is open already

GulsahKose:
Eg: Writer -> Format -> Theme dialog's "Add" button lets the open multiple same dialogs. We should prevent that case.

Change-Id: I4297c133eee52d04f0aa67f62efece87af6b023c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -547,6 +547,13 @@ L.Control.JSDialog = L.Control.extend({
 			}
 		}
 		else {
+			// Is the target dialog already open. So don't let the open same dialog again
+			var dialogs = Object.keys(this.dialogs);
+			for (var i = 0; i < dialogs.length; i++) {
+				if (this.dialogs[dialogs[i]].dialogid === instance.dialogid)
+					return;
+			}
+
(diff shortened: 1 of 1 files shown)


eszkadev (changes requested):
This will block "Full update" which is needed in some cases.
When JSON with the same dialog id comes we remove "old" dialog and then we recreate it using new "updated" data.

GulsahKose:
I couldn't understand "Full update" case. Is there a way that i can test this case.

And I have one more idea to prevent multiple dialogs. This dialog is the child dialog of the Theme. Can we prevent multiple same level child dialogs? Is there any case that we have several same level child dialogs?

eszkadev:
We cannot do it in the online. It has to be done in core. Because when we received JSON it means that dialog was already opened on the server and consumes resources.

eszkadev:
Example of not working update:
Writer -> Format -> Character -> Higlighting -> Color  (it doesn't switch tab with this patch)

Full update happens if we need to change structure of the dialog (add tab page, collapse dialog, etc.)

Issue #6613, opened 2023-06-13
Impress: Wrong context menu in Slides Pane in master view


lpranam:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @lpranam 

### Detailed description and rationale
There should be no context menu option for master slide previews as of now those options are not yet supported

### Code pointers
https://github.com/CollaboraOnline/online/pull/6529
...

### Links to related documentation
...

### Additional notes, tips & tricks
...

codewithvk:
I'm working on it.

PR #6618, opened 2023-06-14
Fix : Wrong context menu in Slides Pane in master view for Impress

codewithvk:
Change-Id: Ia9f1a33d2b698e64660391866ed357b4346339b0

* Resolves: #6613 
* Target version: master

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -255,6 +255,13 @@ L.Control.PartsPreview = L.Control.extend({
 		}, this);
 
 		L.DomEvent.on(img, 'contextmenu', function(e) {
+			var isMasterView = this._map['stateChangeHandler'].getItemValue('.uno:SlideMasterPage');
+			var $trigger = $('#' + img.id);
+			if (isMasterView === 'true') {
+				$trigger.contextMenu(false);
+				return;
+			}
+			$trigger.contextMenu(true);
 			that._setPart(e);
 			$.contextMenu({
 				selector: '#' + img.id,


browser/src/control/Control.PartsPreview.js
```diff
@@ -224,90 +225,98 @@ L.Control.PartsPreview = L.Control.extend({
...
+		L.DomEvent.on(img, 'contextmenu', function(e) { this._contextMenuHandler(e,img); } , this);
+
+		this._layoutPreview(i, img, bottomBound);
+		this._idNum++;
+
+		return img;
+	},
+
+	_contextMenuHandler : function(e, img) {
+		var isMasterView = $('#document-container')[0].classList.contains('slide-master-mode');
```
lpranam, line 237: you can use this method instead

```
map['stateChangeHandler'].getItemValue('.uno:SlideMasterPage');
```

lpranam (changes requested):
Thanks for the PR, I suggest please avoid white space changes so it doesn't show up in git when the actual code is not changed unless it was done on purpose for readability. 

and also make sure to test your code with draw too...

codewithvk:
@lpranam Thanks for your review! 

> Thanks for the PR, I suggest please avoid white space changes so it doesn't show up in git when the actual code is not changed unless it was done on purpose for readability.

That was done intentionally for the sake of readability. However, the pushed code now seems to be fine in terms of formatting.

> and also make sure to test your code with draw too...

Tested!

browser/src/control/Control.PartsPreview.js
```diff
@@ -169,6 +171,7 @@ L.Control.PartsPreview = L.Control.extend({
 
 		var bottomBound = this._getBottomBound();
 		for (var prev = 0; prev < this._previewTiles.length; prev++) {
+			L.DomEvent.on(this._previewTiles[prev], 'contextmenu', function(e) { this._contextMenuHandler(e, this._previewTiles[prev]); } , this);
```
lpranam, line 174: this adds a new handler every time to every slide.

lpranam (changes requested):
just thought:
how about keeping everything simple and just adding a condition to check if it is a master view, and if yes then just not opening the context menu?

browser/src/control/Control.PartsPreview.js
codewithvk: We have to add intentionally else `_contextMenuHandler` event not triggered when we switch to master view mode and Since In master view and Normal View Image ID is same so by default It will open context menu.

codewithvk:
> just thought: how about keeping everything simple and just adding a condition to check if it is a master view, and if yes then just not opening the context menu?

I considered that approach, but it could potentially become problematic. It would block all context menus or pop-ups in the master view, which doesn't seem like the right design pattern. Since I have limited knowledge of the code, it's ultimately your call, and your decision might be more beneficial.

lpranam:
> > just thought: how about keeping everything simple and just adding a condition to check if it is a master view, and if yes then just not opening the context menu?
> 
> @lpranam I considered that approach, but it could potentially become problematic. It would block all context menus or pop-ups in the master view, which doesn't seem like the right design pattern. Since I have limited knowledge of the code, it's ultimately your call, and your decision might be more beneficial.

how about adding simple condition [here](https://github.com/CollaboraOnline/online/blob/7a634d0790576c30c009f195d39647549b290691/browser/src/control/Control.PartsPreview.js#L229)? will it still be problematic?

codewithvk:
> > > just thought: how about keeping everything simple and just adding a condition to check if it is a master view, and if yes then just not opening the context menu?
> > 
> > 
> > @lpranam I considered that approach, but it could potentially become problematic. It would block all context menus or pop-ups in the master view, which doesn't seem like the right design pattern. Since I have limited knowledge of the code, it's ultimately your call, and your decision might be more beneficial.
> 
> how about adding simple condition [here](https://github.com/CollaboraOnline/online/blob/7a634d0790576c30c009f195d39647549b290691/browser/src/control/Control.PartsPreview.js#L229)? will it still be problematic?

Ohh! I was considering adding a condition at the JsDialog based on your previous message. That's why I replied that way.

Anyways, initially, I attempted to add a condition at the code pointer you mentioned. There are two edge cases that you should look into.

- First, let me tell you where it actually works. If you refresh Impress and switch to the master view without creating a first slide preview context menu (i.e., without right-clicking on the preview of the slide in normal mode), then it works like a charm.

Now, here are the two cases where it won't work:

- If you create a context menu for the first slide in normal mode (right-click on the first slide in normal mode) and then switch to the master view, if you try to click on the slide preview in the master view, you will still be able to see the context menu!

- If you create multiple context menus for the first slide in normal mode (multiple right-clicks on the first slide in normal mode) and then go to the master view and try to open the context menu, it will still open the context menu.

The reason for both cases is that we are using the same image ID for the slide previews. So, even if we prevent the creation of the context menu in your code pointer for the master view, the context menu will still exist for that image ID, which was actually created in normal mode.

browser/src/control/Control.PartsPreview.js
codewithvk: Done this one.

lpranam:
you can probably just disable the menu like this in master view

https://swisnl.github.io/jQuery-contextMenu/demo/disabled-menu.html

browser/src/control/Control.PartsPreview.js
```diff
@@ -255,6 +255,13 @@ L.Control.PartsPreview = L.Control.extend({
 		}, this);
 
 		L.DomEvent.on(img, 'contextmenu', function(e) {
+			var isMasterView = this._map['stateChangeHandler'].getItemValue('.uno:SlideMasterPage');
+			var $trigger = $('#' + img.id);
+			if (isMasterView && isMasterView === 'true') {
```
lpranam, line 260: when it's just a direct variable and not accessing a member you don't need to check if its valid

lpranam (approved):

lpranam:
can you please try to rebase this PR?

codewithvk:
> can you please try to rebase this PR?
I'm able to see one commit so far, Should I Pull from master and rebase with it?

codewithvk:
> can you please try to rebase this PR?

 I'm able to see one commit so far, Should I Pull from master and rebase with it?

PR #6619, opened 2023-06-14
Mobile: Add "Read-only" label to toptoolbar

pedropintosilva:
Affected user flows:
- User opens read-only file
  - Before this commit, there was no indication that the current mode
  is read-only.
- User opens file
  - Before this commit, the only indication about the current read
  mode was the pencil Floating Action Button.

Change-Id: I7cc3982b0c0f827a96e144ee339d8f0e601bf03e


pedropintosilva:
This a backport of #6509

Issue #5837, opened 2023-02-13
iOS App: Inserting emoji is not working as intended


nikslor:
**Describe the bug**

In the TestFlight version (22.05.202302092239) inserting an emoji sometimes work and sometimes doesn't. What I often see, is, that inserting emojis in fact act as delete/backspace if there is no text added yet.

I recorded a video and add two different emojis in a otherwise empty document. Please see attached.

https://user-images.githubusercontent.com/1211300/218394607-09c587a4-45fc-4d06-aca8-9baacff050c8.mp4

**To Reproduce**
1. Create a new writer document using the TestFligh iOS app
2. Insert the :+1: emoji
3. Insert the :open_hands: emoji

**Expected behavior**
I should see both emojis

**Actual behavior**
Nothing is shown

**Tablet**
 - OS: IPadOS 15,6,1
 - Version: 22.05.202302092239

plubius:
I can reproduce this. I can insert the first emoji and any after that are treated as delete keys. I will investigate what these key events look like when they are sent from the JS code to the iOS code. My first guess is that higher Unicode characters are getting mangled somewhere in the pipeline.

plubius:
So I did a little debugging and it looks like the JS code handles text input and then sends it to the iOS code to forward to the LibreOffice code.

What I am seeing is that the JS code sends a "textinput" message for the first emoji character entered and then sends a "removetextcontext" for every character (not just emoji characters) entered thereafter.

Seems to me that UTF-16 strings in the "surrogate pair" range (i.e. UTF-16 characters that span 4 bytes instead of 2) are getting misinterpreted by the JS code.

I'll see if I can get any closer to the source of the bug. Note for myself: it looks like the "" message originates in browser/dist/src/layer/marker/TextInput.js.

plubius:
The bug is definitely in the JS code as it also appears when using Collabora Online in the Safari web browser.

Some debugging notes for the JS experts: when a text input event generates an emoji codepoint like U+1F600, L.TextInput.codePointsToString() in browser/dist/src/layer/marker/TextInput.js properly converts the codepoint to 2 16-bit UTF-8 characters and the first surrogate pair entered works fine.  However, any subsequent text input events (even regular ASCII characters) lead to L.TextInput._removeTextContent() being called in L.TextInput._onInput(). There is no escape until you close the document. 

I have narrowed the bug down the following code snippet near the end of L.TextInput._onInput():

    var matchTo = 0;
[...]

PR #6625, opened 2023-06-14
Fix iOS build breakages by adding http::Request to iOS build

plubius:
Still needed to add a bunch of #if !MOBILAPP to disable all of the Online server's unit testing classes so that the iOS app won't crash when opening a document.

TODO: the iOS app will get stuck in an infinite loop when opening a document on iOS in http::Request::readData().

### DIFF
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -254,6 +254,7 @@ void Session::handleMessage(const std::vector<char> &data)
 {
     try
     {
+#if !MOBILEAPP
         std::unique_ptr< std::vector<char> > replace;
         if (!Util::isFuzzing() && UnitBase::get().filterSessionInput(this, &data[0], data.size(), replace))
         {
@@ -261,6 +262,7 @@ void Session::handleMessage(const std::vector<char> &data)
                 _handleInput(replace->data(), replace->size());
             return;
         }
+#endif
 
         if (!data.empty())
             _handleInput(&data[0], data.size());
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -254,15 +254,13 @@ void Session::handleMessage(const std::vector<char> &data)
 {
     try
     {
-#if !MOBILEAPP
         std::unique_ptr< std::vector<char> > replace;
-        if (!Util::isFuzzing() && UnitBase::get().filterSessionInput(this, &data[0], data.size(), replace))
+        if (UnitBase::isUnitTesting() && !Util::isFuzzing() && UnitBase::get().filterSessionInput(this, &data[0], data.size(), replace))
         {
             if (!replace || replace->empty())
                 _handleInput(replace->data(), replace->size());
             return;
         }
-#endif
 
         if (!data.empty())
             _handleInput(&data[0], data.size());
--- a/net/HttpRequest.cpp
+++ b/net/HttpRequest.cpp
@@ -116,7 +116,7 @@ static inline int64_t findEndOfToken(const char* p, int64_t off, int64_t len)
 
 int64_t Header::parse(const char* p, int64_t len)
 {
-    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min(len, 80L)));
+    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min(len, (int64_t)80L)));
     if (len < 4)
     {
         // Incomplete; we need at least \r\n\r\n.
@@ -307,7 +307,13 @@ int64_t Request::readData(const char* p, const int64_t len)
     if (_stage == Stage::Header)
     {
         // First line is the status line.
+#if !MOBILEAPP
         if (p == nullptr || len < MinRequestHeaderLen)
+#else
+        // Fix infinite loop on mobile by skipping the minimum request header
+        // length check
+        if (p == nullptr)
+#endif
         {
             LOG_TRC("Request::readData: len < MinRequestHeaderLen");
             return 0;
--- a/net/HttpRequest.cpp
+++ b/net/HttpRequest.cpp
@@ -116,7 +116,7 @@ static inline int64_t findEndOfToken(const char* p, int64_t off, int64_t len)
 
 int64_t Header::parse(const char* p, int64_t len)
 {
-    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min(len, (int64_t)80L)));
+    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min<int64_t>(len, 80L)));
     if (len < 4)
     {
         // Incomplete; we need at least \r\n\r\n.
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -48,7 +48,7 @@ private:
     /// The UnitBase instance. We capture it here since
     /// this is our instance, but the test framework
     /// has a single global instance via UnitWSD::get().
-    UnitBase* _unit;
+    UnitBase* const _unit;
 
 protected:
     struct WSFrameMask
@@ -81,14 +81,11 @@ public:
 #endif
         _shuttingDown(false)
         , _isClient(isClient)
-        , _unit(nullptr)
+        , _unit(UnitBase::isUnitTesting() ? &UnitBase::get() : nullptr)
     {
 #if MOBILEAPP
         (void) isMasking;
 #endif
-
-        if (UnitBase::isUnitTesting())
-            _unit = &UnitBase::get();
     }
 
     /// Upgrades itself to a websocket directly.
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -45,12 +45,10 @@ private:
     const bool _isClient;
 
     // Last member.
-#if !MOBILEAPP
     /// The UnitBase instance. We capture it here since
     /// this is our instance, but the test framework
     /// has a single global instance via UnitWSD::get().
-    UnitBase& _unit;
-#endif
+    UnitBase* _unit;
 
 protected:
     struct WSFrameMask
@@ -83,13 +81,14 @@ public:
 #endif
         _shuttingDown(false)
         , _isClient(isClient)
-#if !MOBILEAPP
-        , _unit(UnitBase::get())
-#endif
+        , _unit(nullptr)
     {
 #if MOBILEAPP
         (void) isMasking;
 #endif
+
+        if (UnitBase::isUnitTesting())
+            _unit = &UnitBase::get();
     }
 
     /// Upgrades itself to a websocket directly.
@@ -669,14 +668,12 @@ public:
     /// 0 for closed socket, and -1 for other errors.
     int sendMessage(const char* data, const size_t len, const WSOpCode code, const bool flush) const
     {
-#if !MOBILEAPP
-        if (!Util::isFuzzing())
+        if (UnitBase::isUnitTesting() && !Util::isFuzzing())
         {
             int unitReturn = -1;
-            if (_unit.filterSendWebSocketMessage(data, len, code, flush, unitReturn))
+            if (_unit->filterSendWebSocketMessage(data, len, code, flush, unitReturn))
                 return unitReturn;
         }
-#endif
 
         //TODO: Support fragmented messages.
 
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -45,7 +45,7 @@ private:
     const bool _isClient;
 
     // Last member.
-#ifdef ENABLE_DEBUG
+#if !MOBILEAPP
     /// The UnitBase instance. We capture it here since
     /// this is our instance, but the test framework
     /// has a single global instance via UnitWSD::get().
@@ -83,7 +83,7 @@ public:
 #endif
         _shuttingDown(false)
         , _isClient(isClient)
-#ifdef ENABLE_DEBUG
+#if !MOBILEAPP
         , _unit(UnitBase::get())
 #endif
     {
@@ -669,12 +669,14 @@ public:
     /// 0 for closed socket, and -1 for other errors.
     int sendMessage(const char* data, const size_t len, const WSOpCode code, const bool flush) const
     {
+#if !MOBILEAPP
         if (!Util::isFuzzing())
         {
             int unitReturn = -1;
             if (_unit.filterSendWebSocketMessage(data, len, code, flush, unitReturn))
                 return unitReturn;
         }
+#endif
 
         //TODO: Support fragmented messages.
 
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3613,10 +3613,8 @@ private:
     /// Prisoner websocket fun ... (for now)
     virtual void handleMessage(const std::vector<char> &data) override
     {
-#if !MOBILEAPP
-        if (UnitWSD::get().filterChildMessage(data))
+        if (UnitWSD::isUnitTesting() && UnitWSD::get().filterChildMessage(data))
             return;
-#endif
 
         auto message = std::make_shared<Message>(data.data(), data.size(), Message::Dir::Out);
         std::shared_ptr<StreamSocket> socket = getSocket().lock();
@@ -3868,7 +3866,7 @@ private:
             }
 
             // Routing
-            if (UnitWSD::get().handleHttpRequest(request, message, socket))
+            if (UnitWSD::isUnitTesting() && UnitWSD::get().handleHttpRequest(request, message, socket))
             {
                 // Unit testing, nothing to do here
             }
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3489,6 +3489,7 @@ private:
 
         // Consume the incoming data by parsing and processing the body.
         http::Request request;
+#if !MOBILEAPP
         const int64_t read = request.readData(data.data(), data.size());
         if (read < 0)
         {
@@ -3506,6 +3507,7 @@ private:
 
         // Remove consumed data.
         data.eraseFirst(read);
+#endif
 
         try
         {
@@ -3575,7 +3577,6 @@ private:
 #else
             pid_t pid = 100;
             std::string jailId = "jail";
-            LOG_ASSERT_MSG(socket->getInBuffer().empty(), "Unexpected data in prisoner socket");
             socket->getInBuffer().clear();
 #endif
             LOG_TRC("Calling make_shared<ChildProcess>, for NewChildren?");
@@ -3612,8 +3613,10 @@ private:
     /// Prisoner websocket fun ... (for now)
     virtual void handleMessage(const std::vector<char> &data) override
     {
+#if !MOBILEAPP
         if (UnitWSD::get().filterChildMessage(data))
             return;
+#endif
 
         auto message = std::make_shared<Message>(data.data(), data.size(), Message::Dir::Out);
         std::shared_ptr<StreamSocket> socket = getSocket().lock();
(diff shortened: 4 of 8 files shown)


wsd/COOLWSD.cpp
```diff
@@ -3575,7 +3577,6 @@ class PrisonerRequestDispatcher final : public WebSocketHandler
 #else
             pid_t pid = 100;
             std::string jailId = "jail";
-            LOG_ASSERT_MSG(socket->getInBuffer().empty(), "Unexpected data in prisoner socket");
```
plubius, line 3578: Note: on mobile, the socket is always empty as JavaScript messages come over a separate "fake socket" class.

wsd/DocumentBroker.hpp
```diff
@@ -1460,7 +1460,7 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
     const bool _alwaysSaveOnExit;
 
     // Last member.
-#ifdef ENABLE_DEBUG
```
plubius, line 1463: Note: #ifdef ENABLE_DEBUG is always true as config.in will always define it. In a production build config.in contains the following:

#define ENABLE_DEBUG 0

Since _unitWsd was always instantiated in Online builds, production or not, I removed the #ifdef ENABLE_DEBUG.

net/WebSocketHandler.hpp
```diff
@@ -45,7 +45,7 @@ class WebSocketHandler : public ProtocolHandlerInterface
     const bool _isClient;
 
     // Last member.
-#ifdef ENABLE_DEBUG
```
plubius, line 48: Note: #ifdef ENABLE_DEBUG is always true as config.in will always define it. In a production build config.in contains the following:

#define ENABLE_DEBUG 0

Since _unit was always instantiated in Online builds, production or not, I removed the #ifdef ENABLE_DEBUG.

mmeeks: great catch =)

common/Session.cpp
```diff
@@ -254,13 +254,15 @@ void Session::handleMessage(const std::vector<char> &data)
 {
     try
     {
+#if !MOBILEAPP
```
mmeeks, line 257: Soo ... I suggest that we guard these UnitBase.get() calls with:

UnitBase::isUnitTesting() && ...

And then I think we don't get any crashes; and we don't have any extra compile conditionals =)

net/WebSocketHandler.hpp
```diff
@@ -83,7 +83,7 @@ class WebSocketHandler : public ProtocolHandlerInterface
 #endif
         _shuttingDown(false)
         , _isClient(isClient)
-#ifdef ENABLE_DEBUG
+#if !MOBILEAPP
         , _unit(UnitBase::get())
```
mmeeks, line 87: This boggles the mind - why do we have a member in each instance that stores the result of a static function - that seems unbelievable =)

Ash - what is going on here:

commit a89ed3b5350a55d265d5ea0131f3a342e28384ff
Author: Ashod Nakashian <ashod.nakashian@collabora.co.uk>

And why can't we use the global variable; that's really under-explained; also see Patrick's point on the #ifdef FOO.

Ideally we would have a git hook to check for #ifdefs and stop them creeping in in error =)

mmeeks (commented):
I know it's not your fault that there are lots of conditionals [!]; but I'm really trying to not make the maintainability problem worse here =)

I think we can instead use the method UnitBase::isUnitTesting.

If you must get it in as-is fine; but we badly need to be converging not diverging code here for mobile I think.

Possibly (or probably) one of the problems here is the UnitWSD vs. UnitKit when they're linked into the same process, which makes life less elegant, but the UnitBase::isUnitTesting() should be fine either way to guard those accessess.

Thanks!

common/Session.cpp
plubius: OK. That founds like a reasonable approach. I'll see if I can implement that over the next day or two.

Is it safe to assume that any UnitWSD calls can also be guarded with UnitBase::isUnitTesting()?

plubius:
An update on the "the iOS app will not accept text input at all" bug that I am seeing: it seems to only be limited to Writer and Impress documents but not Calc. Also, arrow keys and formatting toolbar buttons both work, I just can't enter any new text or delete text via key events.

net/HttpRequest.cpp
```diff
@@ -307,7 +307,13 @@ int64_t Request::readData(const char* p, const int64_t len)
     if (_stage == Stage::Header)
     {
         // First line is the status line.
+#if !MOBILEAPP
         if (p == nullptr || len < MinRequestHeaderLen)
+#else
+        // Fix infinite loop on mobile by skipping the minimum request header
```
Ashod, line 313: Wouldn't it make more sense to define `MinRequestHeaderLen` differently for mobile, instead of explicitly short-circuiting here?

But even that seems odd. I mean, this is generic HTTP parser. Why would we blanket disable it on Mobile? Doesn't make sense. Surely we use HTTP on Mobile! Perhaps we don't use sockets for certain cases, but disabling such independent code would make it harder to use this code anywhere in Mobile at all.

net/WebSocketHandler.hpp
```diff
@@ -669,10 +668,10 @@ class WebSocketHandler : public ProtocolHandlerInterface
     /// 0 for closed socket, and -1 for other errors.
     int sendMessage(const char* data, const size_t len, const WSOpCode code, const bool flush) const
     {
-        if (!Util::isFuzzing())
+        if (UnitBase::isUnitTesting() && !Util::isFuzzing())
```
Ashod, line 671: We shouldn't need to check `isUnitTesting()`, because the default UnitBase object has no-op implementation. But perhaps we don't want to instantiate them on Mobile? Is that the idea here?

wsd/COOLWSD.cpp
Ashod: Great! So why is asserting that it *is* empty wrong?

wsd/DocumentBroker.cpp
```diff
@@ -149,13 +149,14 @@ DocumentBroker::DocumentBroker(ChildType type, const std::string& uri, const Poc
...
     , _alwaysSaveOnExit(COOLWSD::getConfigValue<bool>("per_document.always_save_on_exit", false))
-#ifdef ENABLE_DEBUG
-    , _unitWsd(UnitWSD::get())
-#endif
+    , _unitWsd(nullptr)
 {
     assert(!_docKey.empty());
     assert(!COOLWSD::ChildRoot.empty());
 
+    if (UnitWSD::isUnitTesting())
```
Ashod, line 157: Can we please do this in the initialization list, and have a const pointer?
```
, _unitWsd(UnitWSD::isUnitTesting() ? nullptr : &UnitWSD::get())
```

(Same elsewhere.)

wsd/DocumentBroker.hpp
```diff
@@ -1460,12 +1460,10 @@ class DocumentBroker : public std::enable_shared_from_this<DocumentBroker>
...
     /// The UnitWSD instance. We capture it here since
     /// this is our instance, but the test framework
     /// has a single global instance via UnitWSD::get().
-    UnitWSD& _unitWsd;
-#endif
+    UnitWSD* _unitWsd;
```
Ashod, line 1466: `UnitWSD* const _unitWsd;`

(Same elsewhere.)

net/WebSocketHandler.hpp
Ashod: Good question! Not obvious, but we now support running multiple tests in the same process. To do that, we have to create a new instance of the UnitX. The problem is that we have all sorts of threads and async logic and re-using the static version sends events/data to the wrong test (when we "switch" from one test to the next).

Multi-test support makes it far easier to add new variations on a given test-case (in the same unit-test files, no Makefile changes), faster build (no new translation units), and faster runtime (same process, same Kit fork). It's wins all around!

(The fact that `ENABLE_DEBUG` was always true is neither here nor there; tests are _explicitly_ disabled outside of debug builds.)

wsd/DocumentBroker.hpp
Ashod: Shouldn't we fix `ENABLE_DEBUG` instead? (Test cannot run in non-debug builds.)

I'd rather we keep this and fix configure.ac. Technically, it's correct that we need test code in debug builds only.

Ashod (commented):
Thanks @plubius. The approach in this PR looks more promising than the previous attempts. Thanks for persevering!

In addition to @mmeeks's notes, I've left some further suggestions. I hope they make sense, but feel free to ask if you think otherwise.

Thanks again!

net/WebSocketHandler.hpp
mmeeks: @Ashod ah - quite possibly we are just failing to do the no-op instantiation on mobile - *but* bear in mind that we have both Kit and WSD in the same process - so probably this one global that is both of these is a problem (or not). If you could dig out the code to do the no-op initialization for Patrick that would be much appreciated - Thanks ! =)

net/HttpRequest.cpp
mmeeks: We shouldn't use HTTP on Mobile - we should just push the messages through without wrapping in a websocket container for performance & slickness.

mmeeks:
> An update on the "the iOS app will not accept text input at all" bug that I am seeing: it seems to only be limited to Writer and Impress documents but not Calc. Also, arrow keys and formatting toolbar buttons both work, I just can't enter any new text or delete text via key events.

@mcecchetti - any chance you can have a quick look at this. Patrick - Marco has been re-working the input methods - I guess you building for iOS is not possible - but if you could read around the code and offer some suggestions that would be great =) thanks !

net/WebSocketHandler.hpp
Ashod: `UnitBase::init()` has a fallback case (when no unit-lib path is given) that creates a `UnitWSD` or `UnitKit` instance.

E.g.:
```
            rememberInstance(UnitType::Wsd, new UnitWSD("UnitWSD"));
```

plubius:
> @mcecchetti - any chance you can have a quick look at this. Patrick - Marco has been re-working the input methods - I guess you building for iOS is not possible - but if you could read around the code and offer some suggestions that would be great =) thanks !

So my first guess is that this might be a Safari/WKWebKit thing. I can confirm that the JS code is not sending some key events as messages to the fake socket JS object. For example, in Calc if I type in the cell's input field in the toolbar, I see events like the following sent to the fake socket which is expected:

2023-06-15 09:10:22.540651 -0400 [ main ] TRC  To Online: textinput id=0 text=j| DocumentViewController.mm:288

But in Writer, I see such events for arrow keys, but no events are written to the fake socket for regular ASCII character text input.

wsd/COOLWSD.cpp
plubius: > Great! So why is asserting that it _is_ empty wrong?

Because that socket is *always* empty. This code is called only once when a document is first opened and I doubt that an actual C socket is ever created so the socket buffer will always be empty.

I would guess that the only reason that iOS invokes this code is that is a side-effect of bootstrapping LibreOffice and starting the LOKit queue. After this code is called, iOS immediately tells the JS code to set app.Socket to a fake socket. The Online server components are never sent messages directly on mobile.

Instead, on iOS the fake socket routes messages to -[DocumentViewController userContentController:didReceiveScriptMessage:] in ios/Mobile/DocumentViewController.mm. This code forwards the message directly to LOKit, bypassing the Online server completely.

wsd/DocumentBroker.cpp
plubius: Done

wsd/DocumentBroker.hpp
plubius: Done

common/Session.cpp
plubius: Done

net/HttpRequest.cpp
```diff
@@ -116,7 +116,7 @@ static inline int64_t findEndOfToken(const char* p, int64_t off, int64_t len)
 
 int64_t Header::parse(const char* p, int64_t len)
 {
-    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min(len, 80L)));
+    LOG_TRC("Parsing header given " << len << " bytes: " << std::string(p, std::min(len, (int64_t)80L)));
```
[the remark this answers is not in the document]
plubius: Done

net/WebSocketHandler.hpp
plubius: I would vote against this for a couple of reasons:

1. We would probably need to pull more Online server code into the Xcode build which will likely bring in more code that doesn't build or run on iOS. AFAICT, there no tests for iOS at this time so sounds like a lot of work for no immediate gain.
2. iOS is a very resource constrained environment. Adding a bunch of code that just do nothing on iOS sounds like a very bad idea given how much effort was spent in March to make the iOS app not crash due to out of memory errors and performance bottlenecks.

By all means, if someone implements unit tests that can on iOS, go ahead and add them but can you do that in a separate pull request? Feel free to add me as a reviewer and I can see if it has any problems building or running.

mmeeks (approved):
Looks beautiful to me - thanks Patrick ! =)

mmeeks:
CI runs builds and tests on Linux - that's fine =)

plubius:
Thanks so much!

mcecchetti:
> So my first guess is that this might be a Safari/WKWebKit thing. I can confirm that the JS code is not sending some key events as messages to the fake socket JS object. For example, in Calc if I type in the cell's input field in the toolbar, I see events like the following sent to the fake socket which is expected:
> 
> 2023-06-15 09:10:22.540651 -0400 [ main ] TRC To Online: textinput id=0 text=j| DocumentViewController.mm:288
> 
> But in Writer, I see such events for arrow keys, but no events are written to the fake socket for regular ASCII character text input.

I performed some test: when Online is loaded in Safari iOS I'm able to see `textinput` events sent from the client. Typing is working fine. I checked the `TextInput` class: there are only a few code fragments specific for the iOS app, and they didn't seem to have been modified recently.  Could it be a focus issue ? I mean maybe there is something preventing the editable area to get focused. Are you using a virtual or hw-keyboard ?
A last question: does text input in Calc work both by using the formula bar and by typing directly into the cell ?

plubius:
> I performed some test: when Online is loaded in Safari iOS I'm able to see `textinput` events sent from the client. Typing is working fine. I checked the `TextInput` class: there are only a few code fragments specific for the iOS app, and they didn't seem to have been modified recently. Could it be a focus issue ? I mean maybe there is something preventing the editable area to get focused. Are you using a virtual or hw-keyboard ? A last question: does text input in Calc work both by using the formula bar and by typing directly into the cell ?

Thank you for taking a look at this. I am seeing this with both hardware and virtual keyboards. Typing directly into a cell does not work, but in both Writer and Calc, I can input text into JS dialogs. So, it looks like only the document area is affected.

I'm not sure if focus is part of the problem. I am seeing text messages come through from the JS code for "action "keys (e.g. arrows, tab, key shortcuts like Command-C/V/Z, etc.), but not for actual text.

Since this works on Safari, when I get some time I will probably need to step through the code and see if the JS code is hitting any special "is mobile" or "is iOS" code.

mcecchetti:
> Since this works on Safari, when I get some time I will probably need to step through the code and see if the JS code is hitting any special "is mobile" or "is iOS" code.

I would suggest to give a glance at this piece of code: https://github.com/CollaboraOnline/online/blob/11767353b3e15063da17eba66b0ec3b14a5e08fc/browser/src/layer/marker/TextInput.js#LL218C32-L218C32

Please note the `ThisIsTheiOSApp` flag. Though, there is obvious reason, it shouldn't work correctly for the iOS app.

plubius:
> Please note the `ThisIsTheiOSApp` flag. Though, there is no obvious reason, it shouldn't work correctly for the iOS app.

Thank you for the code pointer. I put some console.warn() calls in that function and the good news is that every pass through that function appears to be invoking this._setAcceptInput(true).

I am seeing the JS code log all incoming key events so I will start there and walk through the code until I find the branch point.

plubius:
OK. I think I know what is happening: the main text area is not receiving "input" events in the _onInput handler in browser/src/map/handler/Map.Keyboard.js.

I suspect that it is a focus issue. If the text area doesn't have focus
as when I use an external keyboard, it works fine entering text in the "Search" field in Writer's status bar but not in the main document area.

The good news is that without an external keyboard, tapping in the main document area displays the virtual keyboard and text input works perfectly. Even issue #5837 is fixed.

I'll debug this more when I get some time.

mcecchetti:
> The good news is that without an external keyboard, tapping in the main document area displays the virtual keyboard and text input works perfectly. Even issue #5837 is fixed.

Well, that's a great news that with the virtual keyboard typing is working :-)
I would point out the following code fragments:
https://github.com/CollaboraOnline/online/blob/1cf561d0ef2b9b8130bf077355700c93e0898f7d/browser/src/layer/marker/TextInput.js#L118-L130

and:
https://github.com/CollaboraOnline/online/blob/1cf561d0ef2b9b8130bf077355700c93e0898f7d/browser/html/cool.html.m4#L109-L124

Since you have problem with hw keyboard only, it could be worth checking that the above workaround is still working.
Maybe giving a glance to this commit https://github.com/CollaboraOnline/online/commit/3df718aac72138849ecafadc13f0207c3ec52283 could be helpful too.

plubius:
Thank you for the code pointers. I can confirm that the JS window event listeners are being added via the native keyboard detection code. But after much debugging, I found that by using the debug styles for the "clipboard-area" div in _setupStyles in TextInput.js, I could see that the div acts as if it is not editable.

With the debug styles, you can see the div's cursor blink and text is added to the div when running on an iPad with only a virtual keyboard. However, when I run on macOS (from Xcode) with only the macOS keyboard, there is no cursor and no text ever added.

I suspect (hope) that this is just bugs in Apple's "run iOS app on macOS Silicon" feature. So I will pickup a bluetooth keyboard for my iPad later today and see if I see the same text input problem on an actual iPad.

Nevertheless, I finally was able to get the "clipboard-area" div  to accept input when running the iOS app on macOS with the patch below. Strangely, the "-webkit-user-select: none" style for the parent pane of the div blocks input to the div even though the div has focus and is the activeElement: [iOS_running_on_macOS.txt](https://github.com/CollaboraOnline/online/files/11801163/iOS_running_on_macOS.txt)

TL;DR The switch from a textarea to a div for text input acts oddly when running an iOS app on macOS. Still need to determine if this odd behavior occurs on a real iOS device with a keyboard.

plubius:
Good news! The weird key entry bug I have been posting about does *not* occur on my iPad with a bluetooth keyboard so this is just a but in Apple's "run iOS app on macOS" feature. The iOS app is not shipped as a "Mac Catalyst" compatible app. I only use it to speed up my build and test cycle so I don't think we need to worry about this bug.

I can also confirm that issue #5837 is fixed for me in 23.05 with both a virtual keyboard and a bluetooth keyboard. Also, the native input handler popups (e.g. the accented character list when you press and hold a vowel) works properly and the popup is placed in the proper position near the cursor when using an external keyboard.

Overall, the switch to a <div> for key input in 23.05 appears to work much smoother than the <textarea> used in 22.05.

mcecchetti:
> Good news! The weird key entry bug I have been posting about does _not_ occur on my iPad with a bluetooth keyboard so this is just a but in Apple's "run iOS app on macOS" feature. The iOS app is not shipped as a "Mac Catalyst" compatible app. I only use it to speed up my build and test cycle so I don't think we need to worry about this bug.

Really a good news!
And a really weird bug. It reminded me when I hit a mouse event issue that occurred only when Chrome Development Tools panel was open. At least there is nothing to fix. :-)

Nice to hear that the switch to content editable `div` has solved the emoji issue.

PR #6626, opened 2023-06-14
impress: arrange slide/hide show menu entries better

lpranam:
added slide hide/show option in presentation bar
rearranged icons to align them with core UI

Change-Id: Ic4b121ab5d9ab21d8f48e68bb07e51a31442586e

* Target version: master 

### Checklist


pedropintosilva (changes requested):
Thanks for all these improvements :)

The hide/show button in the slide sorter doesn't seem to work as one would expect. It always hides/shows one slide only, doesn't act upon more than one slide (selection) as it happens on core side in the slide sorter

lpranam:
@pedropintosilva that bug is not related to the button, even notebookbar or menu acts similarly, I'll work on it as follow-up. This PR is just cosmetic

lpranam:
follow up: #6635

pedropintosilva (approved):

PR #6628, opened 2023-06-15
use more string_view

caolanm:
Change-Id: Iad56a97f9bef1f6f905f08002b1958072a209551

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Makes sense, thanks.

PR #6632, opened 2023-06-15
private/marco/a11y/cont edit/lists

mcecchetti:
- refactored cypress tests for editable area
- A11yTextInput: empty selection at the beginning of a list entry
- a11y: editable area: cypress tests for list editing


hcvcastro (approved):

hcvcastro:
Thanks. It looks OK

PR #6636, opened 2023-06-15
"throttled" move events can be processed after moveEnd

caolanm:
We throttle the "move" event, but in moveEnd we always call a _move anyway, so if there are throttled moves still pending by the time moveEnd is called then there is no point processing them after _moveEnd because we are up to date already when they arrive and to do would just duplicate tile requests

Change-Id: I15c6fe5f0f1b6171a299a4519138912000dd9b68

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks sensible to me - thanks Caolan =)

PR #6637, opened 2023-06-15
configure option to disable building unit tests

timar:
rational: building for release takes a lot of time and we neither package, nor run (most of) these unit tests when we make packages.
CI runs all of these tests before merging PRs, so risk is low, if we do not run the few build-time unit tests when we package.

make -j12
--enable-tests (default): 3m 18s
--disable-test: 1m 9s
3x faster...

Change-Id: I4edd98af55748432e30c65c4c910e72c0221272a

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -373,6 +373,10 @@ AC_ARG_ENABLE([cypress],
             AS_HELP_STRING([--enable-cypress],
                            [Enable cypress tests.]))
 
+AC_ARG_ENABLE([tests],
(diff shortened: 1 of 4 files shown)


vmiklos (approved):
Looks good. Idea for a follow-up PR: we still have the `PKG_CHECK_MODULES([CPPUNIT], [cppunit])` check unconditionally, it would be nice to not require cppunit at all if --disable-tests is used.

PR #6643, opened 2023-06-17
Bind-Mount Fallback to mknod

Ashod:
- killpoco: use own file stat helper
- mount: support mounting from character device
- wsd: fallback from mknode to bind-mount for random devs
- wsd: test: testInactiveClient may get jsdsialog message

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -330,7 +330,7 @@ void setupJailDevNodes(const std::string& root)
 
 #ifndef __FreeBSD__
     // Create the urandom and random devices.
-    if (!Poco::File(root + "/dev/random").exists())
+    if (!FileUtil::Stat(root + "/dev/random").exists())
     {
         LOG_DBG("Making /dev/random node in [" << root << "/dev].");
         if (mknod((root + "/dev/random").c_str(),
@@ -342,7 +342,7 @@ void setupJailDevNodes(const std::string& root)
         }
     }
 
-    if (!Poco::File(root + "/dev/urandom").exists())
+    if (!FileUtil::Stat(root + "/dev/urandom").exists())
     {
         LOG_DBG("Making /dev/urandom node in [" << root << "/dev].");
         if (mknod((root + "/dev/urandom").c_str(),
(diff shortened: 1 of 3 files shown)


common/JailUtil.cpp
```diff
@@ -304,6 +304,63 @@ void setupChildRoot(bool bindMount, const std::string& childRoot, const std::str
...
+    if (isBindMountingEnabled())
+    {
+        LOG_DBG("Failed to create random device via mknod("
+                << absPath << "). Mount must not use nodev flag. Will try bind-mounting instead: "
+                << strerror(mknodErrno));
+
+        Poco::File(absPath).createFile();
+        if (coolmount("-b", devicePath, absPath))
+        {
+            LOG_DBG("Bind mounted [" << devicePath << "] -> [" << absPath << ']');
```
mmeeks, line 338: I would like a once-per-forkit or once-per-coolwsd LOG_WRN message in this case if we have to start doing this - lets have a static bool warned = false; if (!warned) LOG_WRN("Performance issue: nodev mount permission - have to bind mount random devices.") or something in there.

tools/mount.cpp
```diff
@@ -226,19 +226,39 @@ int main(int argc, char** argv)
...
         {
-            fprintf(stderr, "%s: cannot mount from invalid source directory [%s].\n", program,
-                    source);
+            fprintf(stderr, "%s: cannot mount from invalid source [%s]. stat failed with %s\n",
+                    program, source, strerror(errno));
+            return EX_USAGE;
+        }
+
+        const bool isDir = S_ISDIR(sb.st_mode);
+        const bool isFile = S_ISCHR(sb.st_mode); // We don't support regular files.
```
mmeeks, line 237: So - this is security related; lets special-case this for bind mounting files only from /dev/[u]random to wherever - we want to have the narrowest window possible in this CAP_SYSADMIN code.

common/JailUtil.cpp
```diff
@@ -330,7 +330,7 @@ void setupJailDevNodes(const std::string& root)
 
 #ifndef __FreeBSD__
     // Create the urandom and random devices.
-    if (!Poco::File(root + "/dev/random").exists())
+    if (!FileUtil::Stat(root + "/dev/random").exists())
```
mmeeks: Nice de-poco-ization =)

mmeeks (approved):
Fine to get in, but we should have a couple of follow on cleanups there I think =)
Thanks !

tools/mount.cpp
Ashod: Yes, noted.

PR #6687, opened 2023-06-21
Notebookbar accessibility keys

gokaysatir:
Change-Id: Iae4e8d5cc6044c5aa089c1b0d7c077ee350da3b0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/dom/NotebookbarAccessibility.js
+++ b/browser/src/dom/NotebookbarAccessibility.js
@@ -1,0 +1,310 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+	This class is used for managing the accessibility keys of notebookbar control.
+*/
+
+/* global app NotebookbarAccessibilityDefinitions */
+
+/*
+	This class relies on following id convention (example for "Home" tab):
+		* Tab button id: "Home-etc..."
+		* Tab content container id: "Home-container"
+*/
+var NotebookbarAccessibility = function() {
+
+	this.initialized = false;
+
+	this.activeTabPointers = {
+		id: '',
+		contentList: [],
+		infoBoxList: []
+	},
+
+	this.definitions = new NotebookbarAccessibilityDefinitions();
+	this.tabInfoList = null; // This is to be fetched from NotebookbarAccessibilityDefinitions class at initialization.
+	this.combination = null;
+	this.filteredItem = null;
+
+	this.addInfoBox = function(anchorElement) {
+		var infoBox = document.createElement('div');
+		infoBox.classList.add('accessibility-info-box');
+		infoBox.textContent = anchorElement.accessKey;
+		var rectangle = anchorElement.getBoundingClientRect();
+		infoBox.style.top = (rectangle.top + 20) + 'px';
+		infoBox.style.left = rectangle.left + 'px';
+		document.body.appendChild(infoBox);
+		return infoBox;
+	};
+
+	this.removeCurrentTabInfoBoxesAndAcceleratorKeys = function() {
+		for (var i = 0; i < this.activeTabPointers.contentList.length; i++) {
+			document.getElementById(this.activeTabPointers.contentList[i].id).accessKey = null;
+		}
+
+		for (i = this.activeTabPointers.infoBoxList.length - 1; i > -1; i--) {
+			if (this.activeTabPointers.infoBoxList[i].parentNode !== null)
+				document.body.removeChild(this.activeTabPointers.infoBoxList[i]);
+		}
+	};
+
+	// Checks the current tab's content.
+	this.getElementOfWhichIdStartsWith = function(id) {
+		var tab = document.getElementById(this.activeTabPointers.id.split('-')[0] + '-container');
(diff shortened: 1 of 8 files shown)


browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -0,0 +1,288 @@
...
+		contentList: [],
+		infoBoxList: []
+	},
+
+	this.definitions = new NotebookbarAccessibilityDefinitions();
+	this.tabInfoList = null; // This is to be fetched from NotebookbarAccessibilityDefinitions class at initialization.
+	this.combination = null;
+	this.filteredItem = null;
+
+	this.addInfoBox = function(acnhorElement) {
```
eszkadev, line 23: typo: acnhor

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -0,0 +1,288 @@
...
+		var rectangle = acnhorElement.getBoundingClientRect();
+		infoBox.style.top = (rectangle.top + 20) + 'px';
+		infoBox.style.left = rectangle.left + 'px';
+		document.body.appendChild(infoBox);
+		return infoBox;
+	};
+
+	this.removeCurrentTabInfoBoxesAndAcceleratorKeys = function() {
+		for (var i = 0; i < this.activeTabPointers.contentList.length; i++) {
+			document.getElementById(this.activeTabPointers.contentList[i].id).accessKey = null;
```
eszkadev, line 36: what if we will modify id's in the future in the notebookbar but not here in definitions?
I would recommend to first try to get element by id, then check if it exists. Also when it doesn't exist: console.error - so we will have info in the console that something was changed

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -0,0 +1,288 @@
...
+		this.activeTabPointers.id = id;
+		this.activeTabPointers.contentList = this.tabInfoList[id].contentList;
+		this.activeTabPointers.infoBoxList = [];
+
+		for (var i = 0; i < this.activeTabPointers.contentList.length; i++) {
+			var element = document.getElementById(this.activeTabPointers.contentList[i].id);
+			if (element) {
+				element.accessKey = this.activeTabPointers.contentList[i].combination;
+				this.activeTabPointers.infoBoxList.push(this.addInfoBox(element));
+			}
```
eszkadev, line 55: warn here in console so we will know that something was changed (id) and we will be able to fix that

eszkadev:
I would recommed to not use `getElementById` for uno buttons because in our code we avoid id duplication by adding numbers at the end of id. Some buttons are duplicated so best would be to apply accelerator info box for all the occurences. we can use querySelector and do a selector "starting with id" ...

eszkadev:
this would allow us to avoid duplicated entries for the same command like eg. Insert comment
Also what is important: code adding numbers at the end of id is not deterministic! it can happen that sidebar will be loaded before notebookbar and some items will have different id for example.

So best to select based on class "unoBold" which is the same always, or id "starting with ..."

eszkadev:
But works nicely in general :)

eszkadev (approved):

gokaysatir:
Thank you for the catches @eszkadev, i made some changes. I guess we are ready now :)

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -0,0 +1,310 @@
...
+		var rectangle = anchorElement.getBoundingClientRect();
+		infoBox.style.top = (rectangle.top + 20) + 'px';
+		infoBox.style.left = rectangle.left + 'px';
+		document.body.appendChild(infoBox);
+		return infoBox;
+	};
+
+	this.removeCurrentTabInfoBoxesAndAcceleratorKeys = function() {
+		for (var i = 0; i < this.activeTabPointers.contentList.length; i++) {
+			document.getElementById(this.activeTabPointers.contentList[i].id).accessKey = null;
```
eszkadev, line 41: here getElementOfWhichIdStartsWith should be also reused I think

browser/src/dom/NotebookbarAccessibilityDefinitions.js
```diff
@@ -0,0 +1,220 @@
...
+			'File-tab-label': {
+				focusBack: true,
+				combination: '0',
+				contentList: []
+			},
+			'Home-tab-label': {
+				focusBack: true,
+				combination: '1',
+				contentList: [
+					{ id: 'ResetAttributesimg',				focusBack: true, 	combination: 'AA' },
```
eszkadev, line 22: one more improvement I can see is to avoid adding "img" here everywhere but instead use some smart selector in querySelector, but that's for the future cleanup I think

eszkadev:
now should be fine :)

mmeeks (commented):
Looks great; I guess we want to review the accelerators themselves going forward; but nice work =)

gokaysatir:
Thanks :) yes we will be updating the configuration for each document type.

PR #6690, opened 2023-06-21
test: fix -Werror,-Winconsistent-missing-override

vmiklos:
Change-Id: Ib74b8b8066300e2b3398484d45ff0a8dfa8bd432


vmiklos:
Ah, this is not needed, we already have a #6689 for this.

PR #6691, opened 2023-06-21
Navigator: fix scroll position change

aszucs3:
Save the scroll position of contenttree (whenever its scroll changed), and use its last value as an initial scroll position when the tree is reconstructed.

There is still a glitch, for a moment the tree displayed with a default scroll position (0 = scrolled to top) when the tree reconstructed before its scroll resetted, but after that, the scroll position will be good.

Change-Id: Ib790d50ee13f6fc697ae1481dcb695416c2a2c3f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (changes requested):
patch not needed when we reduce full updates in core: https://gerrit.libreoffice.org/c/core/+/153401

eszkadev:
closing in favor of core fix

PR #6692, opened 2023-06-21
private/pedro/welcomedialog darkmode

pedropintosilva:
- Welcome: Refactoring: fix indentations and quotes
- WIP Pass ui-theme to welcome dialog

### DIFF
--- a/browser/welcome/welcome.html
+++ b/browser/welcome/welcome.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html data-theme="%UI_THEME%">
 	<head>
 		<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">
 		<meta http-equiv="Pragma" content="no-cache">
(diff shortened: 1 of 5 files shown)


pedropintosilva (commented):
@eszkadev can you help me here?

eszkadev:
> @eszkadev can you help me here?

what exactly should I do?

pedropintosilva:
Ah sorry if It wasn't clear. I was looking for code pointers based on the comments I have inserted. Or...aren't those github comments visible to you?

browser/welcome/welcome.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html data-theme="%UI_THEME%">
```
eszkadev, line 3: this will not work as welcome.html isn't processed like  the main app, so variable will not be replaced

eszkadev: change would be nedded in: `FileServerRequestHandler::preprocessWelcomeFile`

eszkadev:
I think easiest is to add:     `&ui_theme=xxx` to the url and then replace %UI_THEME in preprocessing c++ code based on that param

pedropintosilva:
@eszkadev well passing things to the url was exactly how I did it in the past https://github.com/CollaboraOnline/online/commit/90cc39061859ab7229b3b42d34305e0f0080d460 (see the before diff) but now all the code is different and I have not clue on what's the best to achieve the fix that is needed

pedropintosilva:
tested, and added CSS bits, should be merged

PR #6693, opened 2023-06-21
the tile is assumed to be 256 in width

caolanm:
but as seen with unit-tiletest that is not always the case

==24170==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x63400001e808 at pc 0x558a6f505973 bp 0x7ffc0670e7d0 sp 0x7ffc0670e7c8 READ of size 8 at 0x63400001e808 thread T0 (kitbroker_002)
    #0 0x558a6f505972 in DeltaGenerator::DeltaBitmapRow::initRow(unsigned int const*, unsigned int) Delta.hpp:129:36
    #1 0x558a6f4fc318 in DeltaGenerator::DeltaData::DeltaData(unsigned int, unsigned char*, unsigned long, unsigned long, int, int, TileLocation const&, int, int) Delta.hpp:233:21
    #2 0x558a6f4f8a22 in DeltaGenerator::createDelta(unsigned char*, unsigned long, unsigned long, int, int, int, int, TileLocation const&, std::vector<char, std::allocator<char> >&, unsigned int, bool) Delta.hpp:574:17
    #3 0x558a6f4f2a35 in DeltaGenerator::compressOrDelta(unsigned char*, unsigned long, unsigned long, int, int, int, int, TileLocation const&, std::vector<char, std::allocator<char> >&, unsigned int, bool, bool, LibreOfficeKitTileMode) Delta.hpp:669:14
    #4 0x558a6f41f300 in RenderTiles::doRender(std::shared_ptr<lok::Document>, DeltaGenerator&, TileCombined&, ThreadPool&, bool, std::function<void (unsigned char*, int, int, unsigned long, unsigned long, int, int, LibreOfficeKitTileMode)> const&, std::function<void (char const*, unsigned long)> const&, unsigned int, int, bool)::$_0::operator()() const /home/vmiklos/git/libreoffice/online-san/./common/RenderTiles.hpp:304:38

so just use the rle cache for the first 256 pixels

Change-Id: I8d34ea53bd20b69184e100b56017dfc0a904eaab

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Delta.hpp
+++ b/kit/Delta.hpp
@@ -95,8 +95,8 @@ class DeltaGenerator {
             void next()
             {
                 _x++;
-                size_t index = _x >> 6;
-                if (index < _rleMaskUnits && !(_row._rleMask[index] & (uint64_t(1) << (_x & 63))))
+                size_t rleMaskIndex = _x >> 6;
+                if (rleMaskIndex >= _rleMaskUnits || !(_row._rleMask[rleMaskIndex] & (uint64_t(1) << (_x & 63))))
                     _rlePtr++;
             }
         };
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Will make that inner loop significantly slower; and I'm horrified that a non 256x256 tile exists - that will break a lot of stuff; how is that possibly generated ?

caolanm:
found the 400x400, it was an apparently arbitrary tile width used in
TileCacheTests::testSimultaneousTilesRenderedJustOnce

pr to change those 400s and so drop the guard with: https://github.com/CollaboraOnline/online/pull/6709

PR #6694, opened 2023-06-21
jsdialog: fixed dialog content being cut off

lpranam:
i.e: theme dialog(format->theme)

Change-Id: I9a09d52cd2c0d95cfae18ae12995761985d8111d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -41,6 +41,11 @@
 	min-width: 400px;
 }
 
+img.ui-drawing-area {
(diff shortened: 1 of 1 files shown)


eszkadev (changes requested):
no, this change makes all the dialogs huge (eg. format Page Style)
it was intentionally changed so dialogs uses minimum space.
I think we should do something with widgets:
 - scrolling window
 - image
 so we will avoid cutting them (they are only widgets with this problem)

PR #6696, opened 2023-06-22
Limit memory usage of internal log storage.

mmeeks:
Otherwise it could grow without bound; now limited to 100 entries, each of <128 characters.

Change-Id: I6a1b96a8cb6a67f991c3870f5b724989f65e0e74

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Log.js
+++ b/browser/src/core/Log.js
@@ -11,6 +11,14 @@ L.Log = {
 		if (!this._logs) {
 			this._logs = [];
 		}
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
Awesome catch =) would want it on all older branches too - slowly filling memory with junk log messages is not a great plan. Need a sign-off from you though Patrick if you can re-push ?

plubius (approved):
I tested this in the iOS build and it looks good to me.

PR #6701, opened 2023-06-22
Avoid crash when pasting image in debug mode

luigiiucci:
If COOL is running in debug mode ther is a
check that incoming messages are UTF-8.
We skip this tests if the message refers
to an image pasted.

Change-Id: I4f7e4a647deb22ec2a032d4d6347d910184c0701

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks ok

PR #6703, opened 2023-06-22
Add new "download as" PDF option

luigiiucci:
PDF export was avalaible only with a modal "pdf options" window. It seems a little complicated for some users who'd like to simply press the button and receive pdf.
Added a new menu item to download pdf directly.

Change-Id: Id824179adeac3b516e58c85a9e2d01838f81c892

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
Master PR: https://github.com/CollaboraOnline/online/pull/6704

PR #6704, opened 2023-06-22
Add new "download as" PDF option

pedropintosilva:
PDF export was avaliable only with a modal "pdf options" window.
It seems a little complicated for some users who'd like to
simply press the button and receive pdf.
Added a new menu item to download pdf directly.

Signed-off-by: luigiiucci <luigi.iucci@collabora.com>
Change-Id: Id824179adeac3b516e58c85a9e2d01838f81c892

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2384,6 +2384,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			// The fix for issues #6103 and #6104 changes the name of these
 			// icons so map the new names to the old names.
 			'downloadas-pdf': 'exportpdf',
+			'downloadas-direct-pdf': 'exportdirectpdf',
 			'downloadas-epub': 'exportepub',
 		};
 		if (iconURLAliases[cleanName]) {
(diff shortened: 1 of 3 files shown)


pedropintosilva:
22-05 PR: https://github.com/CollaboraOnline/online/pull/6703

pedropintosilva:
This cypress fail is unrelated to this PR and it also fails on plain master branch.
What it fails:
* make -C cypress_test run-desktop spec=writer/editable_area_spec.js
  * Unable to load undo redo document in `Editable area - Basic typing and caret moving`
    * Sub task `Typing <delete>`

pedropintosilva:
I have tested and it works. Just:
* @luigiiucci is there any reason why we didn't do the adjustments for the other apps other than writer?
* @luigiiucci the new options is not visible if `group_download_as` is set to false in `coolwsd.xml`
*

eszkadev (approved):

pedropintosilva:
> I have tested and it works. Just:
> 
>     * @luigiiucci is there any reason why we didn't do the adjustments for the other apps other than writer?
> 
>     * @luigiiucci the new options is not visible if `group_download_as` is set to false in `coolwsd.xml`
> 
>     * @luigiiucci Compact view lacks this new pdf option

@luigiiucci you can add a new commit for each of those topics ^

PR #6706, opened 2023-06-22
avoid calling getImageData if we can

caolanm:
Change-Id: I02c27170b2aa96703fb6cb153cb2f7958fcdee21

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (changes requested):
Looks workable; problem is - we conflate two things: managing local texture / JS memory, and managing remote latency (by caching in the imgDataCache).

I'd really prefer to account for these separately.

I'd suggest that if (!tile.current) - we should drop the imgDataCache - so that should be used only for things that are on-screen: since it's quite a big chunk of RAM to chew up with lots of those.

Hopefully that's easier to do too - just zip though that in the GC pass and drop anything that has an imgDataCache when it's not .current

mmeeks (approved):
Thanks Caolan =)

PR #6707, opened 2023-06-22
Push unpremul and reorder to client

caolanm:
optimized build server flamegraph before:
perf-before

optimized build server flamegraph after:
perf-after

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -6788,14 +6788,14 @@ L.CanvasTileLayer = L.Layer.extend({
 		return ctx;
 	},
 
-	_rgba: function(rawDelta, start, len) {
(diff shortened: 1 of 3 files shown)


mmeeks (approved):
Looks great to me =)
Thanks Caolan.

mmeeks (approved):
Reads beautifully =) thanks !

mmeeks:
One more minor optimization here; since we do the ==255 case and the ==0 case - we should prolly also optimize the: "is it the same as the last un-pre-multiply" case - and just copy the answer from the previous pixel - I imagine lots of times we then end up with just a slowish pixel copying operation instead of the math to un-pre-multiply which is noticably slower in the JS - would be easy as a follow-up. =)

caolanm:
my effort of https://github.com/CollaboraOnline/online/pull/6724 seemed to make things worse rather than better

PR #6709, opened 2023-06-22
drop non-standard width=400 height=400 tile request in unit test

caolanm:
there since:

commit 232499f542341168a8f5b3efb62610ffc1ec8223
Date:   Thu Apr 21 12:34:30 2016 +0300

Change-Id: I04745be0581a697bbc3b15ba1d8bb94207320a7a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Ahh - perfect =) of course, the benefit of passing this pixel size all around the place left & right in our protocol is far from clear to me ;-) quite possibly we should excise that from being even an option.

PR #6712, opened 2023-06-23
Dialog: don't ignore label's padding (ui-text)

pedropintosilva:
Before this 8a0c8c44a710b1043b678f2049e480becd18f0a5 was not being
applied furthermore better to specifically set this to only affect
labels within dialog content (since jsdialog class is also in many
other places such as sidebar and widgets)

Change-Id: Ic1cb825d6c70858a4c49123243866eaf0a6a8e50


pedropintosilva:
this is just css adjustment and it even makes it safer what it was already merged. enough to wait for codeQL to pass

PR #6724, opened 2023-06-23
attempt to optimize unpremultiply

caolanm:
Change-Id: Id3560cd1f40b5235d0115f82479b9091fcf798ad

before: simple
after: Screenshot from 2023-06-23 15-34-12

so that seems to be worse


mmeeks:
Heh; interesting https://stackoverflow.com/questions/21553528/how-to-test-for-equality-in-arraybuffer-dataview-and-typedarray seems to suggest that we might be able to wrap with a UInt32Array cheaply (possibly sharing the underlying data) and then do something more fun there.

mmeeks (approved):
Looks great to me - the subarray and TypedArray overlay stuff was always a bit unexpected to me =) but if it works & performs - that's great. Thanks Caolan !

PR #6725, opened 2023-06-23
Insert new slide to specific position

GulsahKose:
When we insert a new slide via context menu of the slidesorter, new slide should be inserted on pointer position instead of end of the last selected slide.

Change-Id: Ie5efb78d6b6ee468cda150cc416cbd2ab536d6dd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -164,6 +164,25 @@ L.Control.PartsPreview = L.Control.extend({
 		}
 	},
 
+	isPaddingClick: function (element, e, part) {
+		var style = window.getComputedStyle(element, null);
+		var nTop = parseInt(style.getPropertyValue('padding-top'));
+		var nRight = parseFloat(style.getPropertyValue('padding-right'));
+		var nLeft = parseFloat(style.getPropertyValue('padding-left'));
+		var nBottom = parseFloat(style.getPropertyValue('padding-bottom'));
+		var width = element.offsetWidth;
+		var height = element.offsetHeight;
+		var x = parseFloat(e.offsetX);
+		var y = parseFloat(e.offsetY);
+
+		if (part === 'top')         // Clicked on top padding?
+			return !(y > nTop);
+		else if (part === 'bottom') // Clicked on bottom padding?
+			return !(y < height - nBottom);
+		else                        // Clicked on any padding?
+			return !((x > nLeft && x < width - nRight) && (y > nTop && y < height - nBottom));
+	},
+
 	_createPreview: function (i, hashCode) {
 		var frameClass = 'preview-frame ' + this.options.frameClass;
 		var frame = L.DomUtil.create('div', frameClass, this._partsPreviewCont);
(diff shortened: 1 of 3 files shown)


GulsahKose:
Here is the followup core patch https://gerrit.libreoffice.org/c/core/+/153539

browser/src/control/Control.PartsPreview.js
```diff
@@ -225,12 +244,20 @@ L.Control.PartsPreview = L.Control.extend({
 		}, this);
 
 		var that = this;
-		var pcw = document.getElementById('presentation-controls-wrapper');
+		L.DomEvent.on(frame, 'contextmenu', function(e) {
+			var nPos = undefined;
+			if (this.isPaddingClick(frame, e, 'top')) {
+				nPos = frame.id.replace(/^\D+/g, '');
```
GulsahKose, line 250: Here I tried to use frame id but i saw the id's are not secure to use. Because order can change easly. Then i trid to use i variable as ordered index number but didn't work. Any idea about how can understand where the mouspointer is clicked?

pedropintosilva:
@GulsahKose didn't this got already merged?

eszkadev:
@GulsahKose can we rebase this PR, seems to be useful

GulsahKose:
Just rebased the patch, I'm working on...

GulsahKose:
This patch is ready to review. Related core patch is here: https://gerrit.libreoffice.org/c/core/+/172113

pedropintosilva (approved):
This needs one more review to be merged ideally from @eszkadev or @gokaysatir 

My opinion is that, if there is nothing staggering bad, we should just merge it. This adds an important feature and fixes the current faulty behavior and even though the sometimes right click in between the slides can give different results depending on the place the click happens. It is still better than the behavior without these patches

| Current master | With this PR and Gerrit change |
| ---------------------- | --------------------------------------------- |
| https://github.com/user-attachments/assets/747fc64b-8319-4c78-84a3-3f0564cc399c | https://github.com/user-attachments/assets/6ddedbbd-e622-4a05-85ac-7bb76e9d0df3 |

browser/src/control/Control.PartsPreview.js
eszkadev: we use "hash" to identify slides

GulsahKose: That was the old comment szymon and problem solved in time, Here slide hash's are not useful. I have to use slide index, Before slide indexes can be not in order when we move a slide. But currently problem is not present. Frame indexes give true position info us.

eszkadev:
works good in general, but I found corner case:
[impress-2024-08-22_08.55.56.webm](https://github.com/user-attachments/assets/07a65b8d-639d-41ef-afdc-fbec9ebfc97c)

eszkadev (changes requested):
1. right click on some place and insert slide (between 2 other slides, this works correctly)
2. don't click anywhere but try to again repeat step 1

result:
second time it adds slide at the end not in the correct position

GulsahKose:
@eszkadev It should be fixed now.

eszkadev (approved):
thanks it works now, just a nitpick:
- when try to right-click before first slide, it cannot be done far from the first slide, there is a drop area which doesn't handle right click to open context menu - but it's something unrelated to this change:
firstslide

PR #6735, opened 2023-06-26
Mobile: Add "Read-only" label ONLY when it's not possible to edit

pedropintosilva:
The new label added in a13f33b777b0ecc85752f832cd475736206f241e might
be a bit confusing and not necessary if the edit button (pencil) is
already present.

Better to:
* Have only the edit button (that should already be enough to signify
that the user is currently in a read-only transient mode)
* Add Read-only label for the cases were the user is opening a
read-only file (without permission to edit and thus the edit button is
absent)

Change-Id: I1fc5ca821bfd9feb75f411d2c3260a43d54e16cd


Rash419 (approved):
Works as expected, Thanks : )

PR #6736, opened 2023-06-26
private/marco/a11y/cont edit/cypress tests a11y disabled

mcecchetti:
- a cypress env var for enable/disable a11y on testing + a a11y tags
- cypress: expectTextForClipboard helper function never fails
- cypress tests when editing with a11y disabled


eszkadev (commented):
expectTextForClipboard looks good, yes we had "/n" case  only in the past, thanks for fixing

PR #6742, opened 2023-06-26
fix android ci

caolanm:
github_android_online_co-23.05/net/HttpRequest.hpp:1391:63: error: no matching function for call to 'min'
                      << Util::dumpHex(std::string(data.data(), std::min(data.size(), 256UL))));
                                                                ^~~~~~~~

Change-Id: I5b81096c696565b9e4eb4f3ca39c16bc1e5b9182

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks very sensible =)

PR #6744, opened 2023-06-26
jsdialog: avoid closing parent dialog if childe dialog is open

lpranam:
this patch assumes that dialog ids are incremental, and no two dialogs are open at the same time unless one dialog(parent dialog) has triggered another dialog

i.e: calc advance filter dialog triggers warning dialogs, if entered an invalid range and closing the advance filter dialog before the warning dialog causes a crash of LOK

Change-Id: Ic4ba3b4553fb0ee089b1ab9b50b1c023801b674a

* Target version: master 

### Checklist


eszkadev (approved):

lpranam:
reopened new PR: https://github.com/CollaboraOnline/online/pull/6869

PR #6747, opened 2023-06-27
Tablet: Add "Read-only" label only for users with no writing access

pedropintosilva:
Before this the read-only label was being added to the status bar even
if the user has writing access but is currently in that transient
read-only mode state (pencil icon is visible)

Don't show read-only label for those cases. We already have the pencil
icon/btn that signifies that.

Change-Id: I5295564244977388034c7a7930840cb45dfdd629


pedropintosilva:
Szymon said: code looks good but someone else needs to test it

pedropintosilva:
Porting this to master: https://github.com/CollaboraOnline/online/pull/6748

Darshan-upadhyay1110 (approved):
@pedropintosilva  i have tested in tablet , all works good :).

PR #6749, opened 2023-06-27
transport in rgba order

caolanm:
so if core is compiled with a cairo using rgba the pixels can be sent without need to reorder in server or client

Change-Id: Iaf0410f1eaa605b9ce2716625f6c968bca523ccb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
in conjunction with https://gerrit.libreoffice.org/c/core/+/153633

caolanm:
This pull will work in either mode, and https://gerrit.libreoffice.org/c/core/+/153633 would set server side to rgba order

mmeeks (approved):
Looks great =)

PR #6750, opened 2023-06-27
jsdialog: get precision for validation also from value

eszkadev:
when we open Writer -> Format -> Footnote field: Thickness shows an alert about incorrect value. It's caused because step is set to 1 but value is 0.5 as core uses step only for increase/decrease but not for valudation. Fix this annoying error by allowing to use 0.1 step.


eszkadev:
all tests passed apart of broken test: impress/apply_paragraph_props_text_

gokaysatir (approved):

PR #6754, opened 2023-06-27
Amend 'Copy all version information in English' on About Box

timar:
* reuse string from LibreOffice (and make it translatable)
* use an icon instead of button text, like in LibreOffice
* small amendments to text copied to clipboard

Change-Id: Ic16d786fe467c1f8ff718b1f46364541bba8e6ab

image

**Questions:** 
* How to set the style of the button popup text (title attribute)? Other toolbar/sidebar buttons have different style.
* The popup does not appear on mobile UI with log tap, unlike other button popups. Why?

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -663,7 +663,10 @@ L.Map.include({
 		form.querySelector('#coolwsd-version').querySelector('a').focus();
 		var copyversion = L.DomUtil.create('button', 'ui-pushbutton jsdialog', null);
 		copyversion.setAttribute('id', 'modal-dialog-about-dialog-box-copybutton');
-		copyversion.innerHTML = 'Copy version';
+		copyversion.setAttribute('title', _('Copy all version information in English'));
+		var img = L.DomUtil.create('img', null, null);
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
@timar so,
> reuse string from LibreOffice (and make it translatable)
* Yes the button is not translatable as noted in my [comment](https://github.com/CollaboraOnline/online/pull/6549#pullrequestreview-1501527977) from #6549 but I thought @rparth07 could just create a follow up PR for that

> use an icon instead of button text, like in LibreOffice

It was I who requested to avoid using floating icon as a button see my https://github.com/CollaboraOnline/online/pull/6549#pullrequestreview-1471474448 
  * It doesn't look good on mobile
  * It's hard to use it on tablet
  * It's misleading using a small icon (will it copy only wsd version or core version, user tries to guess from the position)
  * Better to use plain text and standard button

pedropintosilva:
> small amendments to text copied to clipboard

thanks

pedropintosilva:
failing impress/apply_paragraph_props_text_      13:27       12       11        1        -        - │
  │    spec.js is unrelated to this change. thanks @timar ! I will be creating additional issues to address the tooltip  and also would be good to either show text or change the icon once the info goes to the clipboard

PR #6759, opened 2023-06-28
rawDelta may not be sufficiently aligned for Uint32Array

caolanm:
in which case it fails, just use the Uint8ClampedArray view to look back at previous input pixel after all.

Change-Id: I1ee5d7480bf9a11145ca6514ef7efb7882c183cb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Ah - good catch - makes sense =) no unit tests for this.

thebearon (approved):
I can confirm it fixes the issue, thanks Caolán!

PR #6761, opened 2023-06-28
Tilewins2

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -5737,22 +5737,11 @@ L.CanvasTileLayer = L.Layer.extend({
 	},
 
 	_pruneTiles: function () {
+		// update tile.current for the view
 		if (app.file.fileBasedView)
 			this._updateFileBasedView(true);
 
-		var key, tile;
-
-		// keep everything current visible
-		for (key in this._tiles) {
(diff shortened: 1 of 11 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1527,26 +1480,37 @@ L.CanvasTileLayer = L.Layer.extend({
...
+		}
+
 		return tile;
 	},
 
-	_tileHasContent: function(key) {
+	_tileNeedsFetch: function(key) {
 		var tile = this._tiles[key];
-		return tile && tile.hasContent();
+		return tile && tile.needsFetch();
```
caolanm, line 1513: tiles: track invalidation by monotonic timestamp does seem to genuinely cause sheet_operation_spec.js 'Switching sheet sets the view that contains cell-cursor' to fail and it does seem to be the case that the cell cursor might be gone missing on switching sheets.

Problem seems to be here:  !tile || tile.needsFetch()

mmeeks: Ah - so should be hasContent -> !needsFetch() I guess.

caolanm (approved):
yeah, I made that change in the force-push above and we're green now

Issue #6767, opened 2023-06-28
Formulabar editing on mobile doesn't work correctly


eszkadev:
**Describe the bug**
When we open on a smartphone in web browser Calc and try to type there is weird behaviour. It is impossible to move caret to the middle of the content and type something there. It always jumps to the end and appends new characters. Sometimes text disappears completly.

**To Reproduce**
1. Go to Cal on mobile in web browser
2. Click into formulabar and type 'hello world'
3. Try to move cursor after 'hello'
4. Try to type something

**Expected behavior**
Typed text appears after hello

**Actual behavior**
Content is appended at the end, after world.

Tested on iOS with Safari (iPhone)

eszkadev:
Fix merged

Issue #6737, opened 2023-06-26
Calc: Formula bar doesnt accept euro-sign


relikd:
**Describe the bug**
I can type a euro-sign (€) in a **cell** but not in the **formulae bar**.
Inserting the character with copy & paste works. Could be a macOS only issue.

**To Reproduce**
1. Click on the formula bar
2. Type alt+E to insert euro-sign

**Expected behavior**
"€"

**Actual behavior**
""

**Screenshots**
N/A

**Desktop (please complete the following information)**
 - OS: macOS
 - Browser: Firefox
 - Version: v22.05.14.3 

**Smartphone (please complete the following information)**
N/A

**Additional context**
N/A

eszkadev:
Fix merged, should be fine now :)

thebearon:
@eszkadev Can you please note the 23.05 fix as well?

relikd:
does not seem to be fixed in 23.05.1.3

PR #6768, opened 2023-06-28
formulabar: handle Japanese input and fix mobile

eszkadev:
It converts formulabar's textarea into contenteditable div.
Input is handled by TextInput.js used for document
and already in the past for formulabar. Replaces
usage of keyevents which are not triggered with
eg. Japanese.
    
Fixes also #6737
Fixes on mobile #6767 requires: https://gerrit.libreoffice.org/c/core/+/153717
    
New widget consists of text layer with visible content
and cursor layer with selections and cursor. When clicked
it focuses TextInput.js field for typing.

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -250,7 +250,7 @@ textarea.cool-annotation-textarea {
 
 /* Related to toolbar.css */
 @-moz-document url-prefix() {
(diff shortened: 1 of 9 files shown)


hcvcastro (approved):

hcvcastro:
It looks Ok.

PR #6771, opened 2023-06-28
Make writer page background match document color

OFFTKP:
It used to be that scrolling in dark mode would show a white color while the tiles were loading - now shows the document background color.

Change-Id: Id4c3881e728ae18496bf276e8c565a7abc1de7b5


caolanm:
The ci failure looks to be the same as the one fixed by:
https://github.com/CollaboraOnline/online/commit/8803bfa9a41f3b27b8e17ce6f1f6ba5bf54c8b5c

caolanm (approved):
looks reasonable to me

pedropintosilva:
Tested it works, thanks!

PR #6772, opened 2023-06-28
scroll: avoid selecting elements while dragging scroll bar

lpranam:
Change-Id: I222d246148d5d84a5c0b3cfdb1890350f428e696

problem:
In safari and gnome web when you scroll using scroll bar,
DOM elements on the slidebar and documents would get selected

* Target version: master 

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -1132,3 +1132,9 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	text-overflow: ellipsis;
 	overflow: hidden;
 }
+
+.prevent-select {
+	-webkit-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
+}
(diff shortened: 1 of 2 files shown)


thebearon:
I see no info in the PR why this was needed, would be good to amend the commit message with reproducer details, and the fix with describing what it does, and what it affects (ie. why it's needed for Safari/WebKit), otherwise people will be clueless about it in the future.

thebearon (commented):
Can't comment on the actual code, but I can confirm it works. Thanks!

pedropintosilva (approved):
It works, thanks! (I couldn't test safari browser)

I wonder, we probably just need the standard `use-select: none` according to the browser compatibility https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility and if that's true wouldn't be better to add it directly inline (to make sure nothing overwrites it) with `element.style.userSelect = 'none';` ? (honest question, at the end of the day up to you @eszkadev @lpranam )

lpranam:
@pedropintosilva According to the link you provided, it suggests we have to use the prefix to use those properties.

On the other hand, I prefer to have things in CSS files instead of inline js, it becomes easier to debug

Issue #6755, opened 2023-06-27
Wrong label 'BackColor' on mobile UI


timar:
![image](https://github.com/CollaboraOnline/online/assets/290614/2844d848-6738-4227-9562-d8b33199a09c)

BackColor is the UNO command name. The label should read: "Character Highlighting Color" (or the localized equivalent).

COOLWSD version: 23.05.1.2

PR #6776, opened 2023-06-29
Fix : Wrong label 'BackColor' on mobile UI

Darshan-upadhyay1110:
Change-Id: I264d02b57a5e2c969ac9245cee2fa4be028e1c04

* Resolves: #6755
* Target version: master 

### Summary

Fix wrong backcolor tooltip for mobile UI
image


pedropintosilva (approved):

timar (commented):
This change will disappear when we run `scripts/unocommands.py` next time (to update translations from core). The question is: why wasn't it extracted properly?

timar:
Actually, I'd rather like this:
UNO('.uno:BackColor', 'text') instead of _UNO('.uno:BackColor') in browser/src/control/Control.MobileBottomBar.js

timar (changes requested):
See my comments above.

Darshan-upadhyay1110:
@timar i see thanks i will check this

timar (approved):

PR #6778, opened 2023-06-29
New PDF download option, porting from2205 plus without legacy hasSigning

pedropintosilva:
- Add new "download as" PDF option - part 2
- Use consistent download PDF as... labeling


JaumePujantell (approved):
Looks good to me.

PR #6779, opened 2023-06-29
restore using Uint32Array for previous pixel compare

caolanm:
removed in:

commit d859da11b098f0e7495e19adc211264e78330710
Author: Caolán McNamara <caolan.mcnamara@collabora.com>
Date:   Wed Jun 28 11:12:28 2023 +0100

    rawDelta may not be sufficiently aligned for Uint32Array

because it would failed with nonaligned data from the delta case.

Those tend to be very short, so for that case copy into a suitable aligned buffer and the lengthy keyframe case which is suitable aligned already can get the win.

before: before2

after: after

Change-Id: I8acfcb18c69380c55128c07f1749eb301c1145a1

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -6796,23 +6796,18 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	_brgatorgba: function(rawDelta) {
 		var len = rawDelta.byteLength / 4;
+		var delta32 = new Uint32Array(rawDelta.buffer, rawDelta.byteOffset, len);
 		var resultu32 = new Uint32Array(len);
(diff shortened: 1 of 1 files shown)


caolanm:
The last version of this had the chart redraw glitch, so a heads up for attempt 2 which shouldn't have that

mmeeks (approved):
Works for me - I don't believe that 'subarray' does a copy - I think it's just a pointer into the parent - so should be fast & efficient.

Alternatively we could of course ensure that all 'commands' are 32bits  in kit/Delta.hxx - I guess concatenating deltas with their 1 byte termination code is perhaps an issue (?).

From a testing perspective - we should prolly try to test the JS - which might be extract-able to something like:

browser/mocha_tests/CanvasSectionContainer.test.ts

with some stubbing =)

PR #6782, opened 2023-06-29
configure.ac: Fix missing space in test command

Minion3665:
Omitting this space leads to the error "test: too many arguments"

Change-Id: Id8a83d77adfb484dcc949a10b4b25f6ad81130f8

* Target version: master 

### Summary
When building Collabora online I noticed this error:
```
./configure: line 18286: test: too many arguments
configure: WARNING: lxml for python3 is needed. It might be in a package called python3-lxml. But
               if you are building the JS bits on another (Linux) machine, that doesn't matter
checking for polib for python3... no
./configure: line 18286: test: too many arguments
configure: WARNING: polib for python3 is needed. It might be in a package called python3-polib. But
               if you are building the JS bits on another (Linux) machine, that doesn't matter
```
It appears that a typo was made in this test command as a space was missing between a uname invocation and the = sign

### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -96,7 +96,7 @@ for MODULE in lxml polib; do
         AC_MSG_RESULT([yes])
     else
         AC_MSG_RESULT([no])
-        AS_IF([test `uname -s` = Linux -o `uname -s` = FreeBSD -o `uname -s`= Darwin],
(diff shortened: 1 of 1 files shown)


Minion3665:
(I suspect this was missed for so long because most times someone building will have polib and lxml so will never run into this issue. (Luckily? Unluckily?) I have access to a system that doesn't seem to be recognizing the presence of these modules)

timar (approved):
Thanks!

PR #6783, opened 2023-06-29
Tabbed view: Writer: Insert tab: Reorder elements

pedropintosilva:
Better to reorder these so we have higher probably to meet user's
expectations and fix the priority in which the elements are placed

Change-Id: I9ae3bcd087f82e2cfa02da780638424f37a3589c


gokaysatir (approved):

PR #6785, opened 2023-06-29
Private/aszucs3/navigator expanded

aszucs3:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -8,8 +8,8 @@
  *     id: 'id',
  *     type: 'treelistbox',
  *     entries: [
- *         { row: 0, text: 'first entry', children: [ { row: 2, text: 'first subentry' } ] },
- *         { row: 1, text: 'second entry', selected: true, state: false, ondemand: true }
+ *         { row: 0, text: 'first entry', collapsed: true, children: [ { row: 1, text: 'first subentry' } ] },
+ *         { row: 2, text: 'second entry', selected: true, state: false, ondemand: true }
  *     ]
  * }
  *
@@ -27,6 +27,8 @@
  * 'row' property is used in the callback to differentiate entries
  * 'state' property defines if entry has the checkbox (false/true), when is missing - no checkbox
  * 'ondemand' property can be set to provide nodes lazy loading
+ * 'collapsed' property means, this entry have childrens, but they are not visible, because
+ *             this branch is collapsed.
  *
  * Copyright the Collabora Online contributors.
  *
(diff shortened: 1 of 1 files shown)


browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -426,15 +434,24 @@ function _handleKeyEvent(event, listElements, builder, data) {
...
+		}
 		preventDef = true;
 	} else if (event.key === 'ArrowUp') {
 		if (currIndex < 0)
 			_changeFocusedRow(listElements, currIndex, treeLength - 1);
-		else if (currIndex > 0)
-			_changeFocusedRow(listElements, currIndex, currIndex - 1);
+		else {
+			var nextIndex = currIndex - 1;
+			while (currIndex >= 0 && listElements[nextIndex].clientHeight <= 0)
```
eszkadev, line 450: is that condition correct?
what it checks?

I would expect we want to check if `(nextIndex >= 0)` because we use it to access array listElements

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -426,15 +434,24 @@ function _handleKeyEvent(event, listElements, builder, data) {
 	if (event.key === 'ArrowDown') {
 		if (currIndex < 0)
 			_changeFocusedRow(listElements, currIndex, 0);
-		else if (currIndex < treeLength - 1)
-			_changeFocusedRow(listElements, currIndex, currIndex + 1);
-
+		else {
+			var nextIndex = currIndex + 1;
+			while (currIndex < treeLength - 1 && listElements[nextIndex].clientHeight <= 0)
```
eszkadev, line 439: currIndex or nextIndex?  (in the first condition)

aszucs3: thx, you are right :)

browser/src/control/jsdialog/Widget.TreeView.js
aszucs3: thx, you are right :)

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -162,19 +164,25 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
...
 			builder.callback('treeview', 'expand', treeViewData, entry.row, builder);
-		toggleFunction();
+		$(span).toggleClass('collapsed');
 	};
 
 	if (entry.children) {
 		var ul = L.DomUtil.create('ul', builder.options.cssClass, li);
-		for (var i in entry.children) {
-			_treelistboxEntry(ul, treeViewData, entry.children[i], builder, isTreeView, treeRoot);
+		if (!entry.collapsed) {
```
eszkadev, line 182: we should create even collapsed children. we use treeview also without dialog in the backend (in Zotero plugin) where we just show/hide nodes on collapse/expand, there is no message from the core after change

eszkadev (approved):

Issue #6764, opened 2023-06-28
About: Copy version button lacks visual feedback


pedropintosilva:
# Description
About dialog's new copy version button works but doesn't inform the user that it worked. Visually nothing happens.

# Enhancement
It would be good to display any kind of visual feedback to inform the user that information is already copied:
* make use of `showSnackbar` to display a temporary bottom notification "Version information has been copied"

## Optional enhancement 
Change the image to a check-mark temporarily

PR #6796, opened 2023-06-30
Added visual feedback after copy information into clipboard

Darshan-upadhyay1110:
Change-Id: I41debc2421841e4f2235f2d31c49d3687a096339

* Resolves: #6764 
* Target version: master 

### Summary
   * Added visual feedback to inform the user that information is already copied
   * Change the image to a check-mark temporarily
  
[Copy-to-clipboard.webm](https://github.com/CollaboraOnline/online/assets/61383886/da6e7694-db23-4a65-a084-c0df29edff47)


browser/src/control/Toolbar.js
```diff
@@ -720,6 +722,16 @@ L.Map.include({
...
+	showSnackbar: function() {
+		var timeout = 1000;
+		this.uiManager.showSnackbar('Version information has been copied', null, null, timeout);
+		var copybutton = document.querySelector('#modal-dialog-about-dialog-box-copybutton > img');
+		L.LOUtil.setImage(copybutton, 'lc_clipboard-check.svg', this._docLayer._docType);
+		setTimeout(function () {
+			L.LOUtil.setImage(copybutton, 'lc_copy.svg', this._docLayer._docType);
+		}.bind(this), timeout);
+	},
+
```
pedropintosilva, line 734: Do we need this?:

1. maybe we could avoid having another showSnackbar function (it can be quite confusing)
2. avoid changing the showSnackbar function in UIManager and instead just close it directly from Toolbar.js . I think we are doing something like ```setTimeout(function () { instance.that.closePopover(instance.id, false); }, this.options.snackbarTimeout);``` elsewhere

pedropintosilva:
Thanks, I think it's nice to reuse the snackbar for that :) maybe Szymon can take a look?

browser/src/control/Toolbar.js
Darshan-upadhyay1110: 1. showSnacbar function is for code which is kind of duplicating so i have added that in showSnacbar function. **You are right i need to rename because it is bit confusing.** 

2. So the main resone i have added timout in UImanager is because i need to pass custome timeout to toolbar.js. ( Default is 10 sec in toolbar.js )

i have made some change in toolbar.js  i think we need to add a parameter timeout because for diffrent showSnackbar messages we may want diffrent timeouts to show message.

browser/src/control/Control.JSDialog.js
```diff
@@ -474,7 +474,8 @@ L.Control.JSDialog = L.Control.extend({
 
 		instance.callback = e.callback;
 		instance.isSnackbar = e.data.type === 'snackbar';
-		instance.isModalPopUp = e.data.type === 'modalpopup';
+		instance.isModalPopUp = e.data.type === 'modalpopup' || instance.isSnackbar;
```
eszkadev, line 477: thy this line is changed?

browser/src/control/Toolbar.js
```diff
@@ -720,6 +722,16 @@ L.Map.include({
 		}
 	},
 
+	showSnackbar: function() {
```
eszkadev, line 725: please rename to something describing more what it does :)
like showHasBeenCopiedSnackbar

browser/src/control/Control.JSDialog.js
eszkadev: yes, snackbar shouldn't be modal

eszkadev (approved):

pedropintosilva (approved):
Thanks

PR #6801, opened 2023-07-03
Provide unique ids for notebookbar buttons.

gokaysatir:
Change-Id: I5a4cedf5a714722379c95dff2cd67e050fbf6294

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
hmm , I would say better to have keyboard accelerator definitions based on UNO command not some specific id - less to maintain, because when we decide to move some button between tabs, or add second occurence of it in other tab we don't have to do anything. with this solution you will need to change the list of ids.

this is why we use unique id with some number as id, but we keep   the same class for every instance of button with given uno command: .unoXXX    example all Bold buttons will have class .unoBold

isn't better idea to define in accelerators list:      Bold  uno command has always   accelerator 'B',    instead of       button bold in tab X and button Bold in tab Y has accelerator 'B'

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2491,14 +2474,11 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
 
-			if (isRealUnoCommand)
-				id = builder._makeIdUnique(id);
-
 			div.id = id;
 			data.id = id; // change in input data for postprocess
 
 			var icon = data.icon ? data.icon : builder._createIconURL(data.command);
-			var buttonId = id + 'img';
+			var buttonId = id + '-button';
```
eszkadev, line 2481: isn't that 'img' used in some branding CSS or in CSS in general?

gokaysatir: I will also update those parts i think. This is draft for now.

gokaysatir:
> hmm , I would say better to have keyboard accelerator definitions based on UNO command not some specific id - less to maintain, because when we decide to move some button between tabs, or add second occurence of it in other tab we don't have to do anything. with this solution you will need to change the list of ids.
> 
> this is why we use unique id with some number as id, but we keep the same class for every instance of button with given uno command: .unoXXX example all Bold buttons will have class .unoBold
> 
> isn't better idea to define in accelerators list: Bold uno command has always accelerator 'B', instead of button bold in tab X and button Bold in tab Y has accelerator 'B'

Maybe but it feels like it would limit the flexibility in the future. But i think i could add the definitions directly into Control.NotebookbarWriter.js into the "getFooTab" functions. That would solve both of our concerns. Let me think about this. I will turn back to this draft after multilanguage support PR.

PR #6802, opened 2023-07-03
Add action for detecting code duplicates

eszkadev:
Detector for code duplication in the PR. For now set minimal length of 26 lines, as we have some dups of 25. In the future we can decrease after refactor.


gokaysatir (approved):
Thanks for the improvement :)

Issue #6795, opened 2023-06-30
hyperlink dialog on chrome


pedropintosilva:
**Describe the bug**
Dialog looks horrible on chrome

![image](https://github.com/CollaboraOnline/online/assets/65948705/324e1bb7-9eba-4371-8ca4-bcbd53730060)

Expected: should look like the same as in FF

Darshan-upadhyay1110:
@pedropintosilva  This seems fixed i am not seeing this design error in local so we can close this issue as fixed 
![image](https://github.com/CollaboraOnline/online/assets/61383886/34de0d81-57a8-4fc7-8af3-8306e751d21b)

pedropintosilva:
ah so it's still weird

pedropintosilva:
the text input field should be full width

Darshan-upadhyay1110:
oh okay i know this one. i think i will add my soultion here for this

PR #6804, opened 2023-07-03
Fix odd looking textarea width.

Darshan-upadhyay1110:
Change-Id: Ica7977c8d334bbb8bcb3bfd3c8aca8d4876c0571

* Resolves: #6795
* Target version: master 

### Summary
Fixed hyper link textarea dialog box's width

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -676,7 +676,6 @@ input[type='number']:hover::-webkit-outer-spin-button {
 
 /* ui-textarea */
 .jsdialog.ui-textarea {
-	max-width: 400px;
 	color: var(--color-text-dark);
 	background-color: var(--color-background-dark);
 	border: 1px solid var(--color-border-dark);


pedropintosilva (changes requested):
the 400px max-width seems to have been introduced quite long time ago: https://github.com/CollaboraOnline/online/pull/1609 and possibly we don't need that hard-coded width anymore but we need to:
* Test how .jsdialog.ui-textarea in other places are affected (including mobile)
* Check if we really need a max-width (probably we don't?)
  * But if we do, can't we just us width auto ? ( I think we are already inheriting this, so possible we can just remove that max-width altogether? If we find other places where having a hard coded max width is useful than we should tweak the `#hyperlink-text-box:not(.mobile-wizard) {` in cool.css instead )

Darshan-upadhyay1110:
@pedropintosilva i have tested for many places where  js-dialog textare present. 

 * actually you are right we don't need max-width property here. 
 * i have also tested on mobile view , with branding on nextcloud and without branding css on nextcloud  all cases are good with out max-width.
 
 
image
image
image

 
  * we can remove this property and then all dialog box works fine.

Issue #6808, opened 2023-07-03
Language selection in More... dialog is not applied


federicorispo:
**Describe the bug**
When a user tries to select a different language for the document using the More... dialog, the system does not apply the selection.

**To Reproduce**
1. Go to the menu ''Tools/Language/For All Text/"
2. Click on "More..."
3. Selects a language
4. Click on "OK"
5. Nothing changes and in the console appears this error:
```
Cannot execute: .uno:LanguageStatus?Language:string=Default_Portuguese (Brazil) when dialog is opened.
```

**Expected behavior**
The system should apply the selected language.

**Actual behavior**
The system does not apply the selected language. This happens also if the user wants to change a language for:
- Selection
- Paragraph

**Screencast**
https://github.com/CollaboraOnline/online/assets/44871614/041a0dfa-63cf-4a7c-9f47-74053d1d9a9b

**Desktop (please complete the following information)**
 - OS: Ubuntu 22.04
 - Browser: tested on:
   - Firefox (113.0)
   -  chrome (114.0.5735.199)

PR #6809, opened 2023-07-03
fix: language selection in More... dialog is not applied

federicorispo:
### Info
* Resolves: #6808
* Target version: master 

### Summary
When a user selects a different language for the document using the More... dialog, the system does not apply the selection. This patch allows the system to send the LanguageStatus command even when the jsdialog is opened.

Change-Id: I7a1a39b31db59f1bc5f980d63f3040ffa9b311e5

### Checklist

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -395,6 +395,16 @@ L.Map.include({
 
 		if (this.uiManager.isUIBlocked())
 			return;
+
+		// This check is necessary to allow the user to change a language in the More... modal
+		if (this.jsdialog
+			&& this.jsdialog.hasDialogOpened()
+			&& command.startsWith('.uno:LanguageStatus?Language')
+		) {
+			app.socket.sendMessage('uno ' + command);
+			return;
+		}
+
 		if ((this.dialog.hasOpenedDialog() || (this.jsdialog && this.jsdialog.hasDialogOpened()))
 			&& !command.startsWith('.uno:ToolbarMode') && !force) {
 			console.debug('Cannot execute: ' + command + ' when dialog is opened.');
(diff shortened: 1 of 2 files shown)


browser/src/control/Toolbar.js
```diff
@@ -395,6 +395,16 @@ L.Map.include({
...
+		if (this.jsdialog
+			&& this.jsdialog.hasDialogOpened()
+			&& command.startsWith('.uno:LanguageStatus?Language')
+		) {
+			app.socket.sendMessage('uno ' + command);
+			return;
+		}
+
 		if ((this.dialog.hasOpenedDialog() || (this.jsdialog && this.jsdialog.hasDialogOpened()))
 			&& !command.startsWith('.uno:ToolbarMode') && !force) {
```
eszkadev, line 409: Here we added special parameter "force" for situation like this. Could you try use "force = true" in the source of uno command?

eszkadev: its probably here: https://github.com/CollaboraOnline/online/blob/61aa932d3088ce698dab86304f8acaf86c78116d/browser/src/control/Control.LanguageDialog.js#L19

federicorispo: @eszkadev Thanks for the quick reply. I noticed the `force` boolean but to pass it as `true` I need to pass also a null/empty json body because, as far I understand, the LanguageStatus command can be used only with query parameters:

https://github.com/CollaboraOnline/online/blob/32218cf54b862e2f422d7f17ea935628c2afbed6/browser/src/control/Toolbar.js#L343

https://github.com/CollaboraOnline/online/blob/61aa932d3088ce698dab86304f8acaf86c78116d/browser/src/control/Control.LanguageDialog.js#L20

Let me know if you prefer my approach or updating the PR passing an empty json and `true` for the force parameter

eszkadev: null json is fine :)

federicorispo: PR updated

eszkadev (approved):

PR #6812, opened 2023-07-04
Coverity5

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1158,6 +1158,7 @@ bool ChildSession::getTextSelection(const StringVector& tokens)
 
 bool ChildSession::getClipboard(const StringVector& tokens)
 {
+    std::string token;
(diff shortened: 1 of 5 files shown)


kit/ChildSession.cpp
```diff
@@ -1166,7 +1167,6 @@ bool ChildSession::getClipboard(const StringVector& tokens)
     char       **pOutStreams = nullptr;
 
     bool hasMimeRequest = tokens.size() > 1;
-    std::string token;
```
vmiklos, line 1169: Hmm, how can it happen that moving this up a little helps?

vmiklos (commented):
Looks good in general, but I don't understand the last commit. What's happening there? Thanks.

vmiklos (approved):
Ah, token is referenced by pMimeTypes, got it.

PR #6813, opened 2023-07-04
draw a checkerboard for missing tiles when debugging

caolanm:
Change-Id: I27cef6d9903e93fe4f87a8627c4fab84257129ef (cherry picked from commit 308f367ffa6daec90f24603cca07bf7f26d37825)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me - I love the makeCheckPattern =)

A couple of thoughts - there is a simple analog to this with this code:

/*		if (this._map._canvasDevicePixelGrid)
		{
			// FIXME: render this onto the canvas with hairlines (?)
			// browser/test/pixel-test.png - debugging pixel alignment.

That should be easy to do the same for.

Also - that we should prolly count how many tiles we have failed to render (and or perhaps the failed-to-render * duration metric) so we can see and try to minimize that.

But this is great =)

PR #6815, opened 2023-07-04
Notebookbar accelerators, 2 step approach: Select tab first, then pic…

gokaysatir:
…k command.

Change-Id: Idde0cd3401beb89c999b698af5ecb5a9348b5d8c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
works but needs set focus

PR #6816, opened 2023-07-04
On a cache hit for a tile the result typically had a wireId of 0

caolanm:
And tile::needsFetch on that result will always return true given the default of invalidFrom of 0 and the comparison of
  return this.invalidFrom >= this.wireId || ...
so the tile return from the cache is unusable.

It is typical for a request not to have a wireId, if we don't use the cache then doRender will send a timecombine result with wireIds set. But if we use the cache here we send a response using the wireId of the request.

Change-Id: I586406af86b0f503dfc204e20120cc8acebf5b8e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Lovely =)

PR #6817, opened 2023-07-04
Fetch enough for next vertical scroll

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Looks great to me - thanks Caolan =) I'm still worried that we have lots of amber tiles in my calc view though (with the new debug).

PR #6826, opened 2023-07-05
keyboard: show keyboard only when explicitly requested

lpranam:
problem:
on mobile phone keyboard used to flicker when you tap on cell, even if you were not editing inside cell

Change-Id: I966267817e054694dabd30c2aa8db09a1f189ce5

* Target version: master 

### Checklist


gokaysatir:
Failures may not be false-positive.

lpranam:
closed in favor of https://github.com/CollaboraOnline/online/pull/6871

PR #6827, opened 2023-07-05
asan error seen in testRleComplex

caolanm:
=============== START DeltaTests::testRleComplex
../kit/Delta.hpp:98:23: runtime error: index 4 out of bounds for type 'const uint64_t[4]' (aka 'const unsigned long[4]')
    #0 0x55d29dacd21b in DeltaGenerator::DeltaBitmapRow::PixIterator::next() libreoffice/online-san/test/../kit/Delta.hpp:98:23
    #1 0x55d29da8d514 in DeltaTests::testRleComplex() libreoffice/online-san/test/DeltaTests.cpp:316:16
    #2 0x55d29daf7a87 in void std::__invoke_impl<void, void (DeltaTests::*&)(), DeltaTests*&>(std::__invoke_memfun_deref, void (DeltaTests::*&)(), DeltaTests*&) /usr/bin/../lib64/gcc/x86_64-suse-linux/9/../../../../include/c++/9/bits/invoke.h:73:14

Change-Id: Ib6e13eaf0fb89ad086a99251d5b8edcdcedd6800

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Delta.hpp
+++ b/kit/Delta.hpp
@@ -95,7 +95,8 @@ class DeltaGenerator {
             void next()
             {
                 _x++;
-                if (!(_row._rleMask[_x >> 6] & (uint64_t(1) << (_x & 63))))
+                size_t rleMaskIndex = _x >> 6;
+                if (rleMaskIndex >= _rleMaskUnits || !(_row._rleMask[rleMaskIndex] & (uint64_t(1) << (_x & 63))))
                     _rlePtr++;
             }
         };


mmeeks (commented):
Would really prefer to ensure that we don't accept anything >256x256 pixels I think; we should just fail in some helpful way I think.

vmiklos:
Thanks, this fixes the problem for me locally.

caolanm:
> Would really prefer to ensure that we don't accept anything >256x256 pixels I think; we should just fail in some helpful way I think.

While I can add something to not end up here if a > 256x256 arises, this asan problem arises on 256x256 where the typical pattern is:

PixIterator it(row);
for (int col = 0; col < 256; ++col)
{
  // at col of 255, PixIterator internal __x is 255, next bumps that to 256 which is an index 4 into the 4 element _rleMask and does an out of bounds read
  it.next();
}

mmeeks (approved):
Thanks =)

caolanm:
https://github.com/CollaboraOnline/online/pull/6842 in case there is some > 256x256 case

PR #6829, opened 2023-07-05
private/pedro/unoToolButton alt n img alt

pedropintosilva:
- Fix unoToolButton unlocalized alt attribute and avoid digits
- Add missing img alt attribute to unoToolButton


eszkadev (approved):
looks ok

PR #6830, opened 2023-07-05
Gray out empty treeview and signal it as disabled component

pedropintosilva:
So user understands the component is disabled

Change-Id: Ide4fb7589b5be611975de4a704b458bb626be0e0

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -386,6 +386,10 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	width: 100%;
 }
 
+.ui-treeview.empty .ui-treeview-body {
+	color: var(--color-text-lighter);
+}
+
 .ui-treeview-cell.with-sub-dialog {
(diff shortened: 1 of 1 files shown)


browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -517,6 +517,7 @@ function _treelistboxControl(parentContainer, data, builder) {
 			tr.innerText = _('Headings and objects that you add to the document will appear here');
 		} else {
 			L.DomUtil.addClass(table, 'empty');
+			table.setAttribute('disabled', 'true');
```
eszkadev, line 520: empty != disabled

this can confuse people later (developers)

eg. pivot table drag and drop: when one treeview is empty but should accept drop, it is empty but not disabled. while debugging someone will spend time to understand why it is disabled suddenly

could you instead apply proper styling to `.empty` treeview in CSS?

eszkadev (approved):

PR #6832, opened 2023-07-05
impress slide previews are not 256x256 to triggers assert

caolanm:
but they don't go through the delta mechanism, so move the assert to the delta path, the other path is ok wrt gt 256x256

Change-Id: I2ba4d8affb7645349540f3a5de31d1802f04c53d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks good =)

PR #6840, opened 2023-07-06
don't append wid=X to payload

caolanm:
commit 36818e915316f12ed9988adf04bb03888f5003ac
Date:   Thu Jun 22 22:26:59 2023 +0100

    tiles: track invalidation by monotonic timestamp.

added wid=X to invalidatetiles (etc?)

but TileCache::parseInvalidateMsg is unaware of the extra arg and ends up doing nothing.

Change-Id: I6da20253807daa60064683d13e7e9471ccab4b99

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1174,8 +1174,7 @@ public:
         {
             // no point in handling invalidations or page resizes per-view,
             // all views have to be in sync
-            tileQueue->put("callback all " + std::to_string(type) + ' ' + payload +
-                           " wid=" + std::to_string(RenderTiles::getCurrentWireId()));
+            tileQueue->put("callback all " + std::to_string(type) + ' ' + payload);
         }
         else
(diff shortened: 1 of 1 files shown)


mmeeks:
We need the wid= - to be able to work out what is obsolete in our tile-cache in the JS side I believe; it stops various races.

caolanm:
hmm, one thing I got hung up on earlier is that in. CanvasTileLayer.js _onMessage we  have:

else if (textMsg.startsWith('invalidatetiles:')) {
 ... var msg = 'invalidatetiles: '; ... 
}

which processes an incoming invalidatetiles and builds a new invalidatetiles message, and that one also doesn't take the wid in its input into account when building its output

caolanm:
lets drop this one then as we want to keep the wid. https://github.com/CollaboraOnline/online/pull/6846 for that js case where I saw the wid getting lost in an intermediate parse.

PR #6842, opened 2023-07-06
return failure early in case of oversized tiles

caolanm:
it shouldn't happen, but return early if it does

Change-Id: I1733d346ff370827fd882caacebca111f790bbe1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Thanks =)

PR #6845, opened 2023-07-06
jsdialog: calendar control

eszkadev:
Can be tested in Writer -> Review -> Manage -> Filter (date dropdown)

Requires:
https://gerrit.libreoffice.org/c/core/+/154133
https://gerrit.libreoffice.org/c/core/+/154135
https://gerrit.libreoffice.org/c/core/+/154136
https://gerrit.libreoffice.org/c/core/+/154192
https://gerrit.libreoffice.org/c/core/+/154193

Fixes date picker and time field


lpranam (approved):
works fine

my suggestion for improvement as follow-up:
+ Style date drop-down needs styling if it is disabled
+ Should be able to pick year and month too

PR #6849, opened 2023-07-06
docker: remove obsolete fix

timar:
The domain resolution problem from jails was properly solved by https://github.com/CollaboraOnline/online/commit/bc8da0cb339cc685f9a24c78e8a92aadcd690479 This hack is not necessary in docker.

Change-Id: I59664b895d187b2a8930a34b096305e0b7c384e3


Ashod (approved):
Thanks @timar. Agreed that there is no reason to treat resolv.conf any differently from the other /etc files. Since we check and update them from coolwsd (as the linked commit shows), the removed code doesn't seem to be doing anything useful, as far as I can tell.

PR #6851, opened 2023-07-07
Focus on the color input popup when keyboard is used.

gokaysatir:
Change-Id: Ibf189aaa6029b3de1e1b4233d583f42e07be21d6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
I tested and in Writer Font color works, but Higlight color cannot be applied using keyboard (it closes popup and it is not applied).
Also when we focus arrow -> popup appears, but when we select icon we should apply last selected color - this is not possible using keyboard, while black outliner appears on full icon.

eszkadev (approved):

gokaysatir:
> * I tested and in Writer Font color works, but Higlight color cannot be applied using keyboard (it closes popup and it is not applied).
> * Also when we focus arrow -> popup appears, but when we select icon we should apply last selected color - this is not possible using keyboard, while black outliner appears on full icon.
> * When I open popup, some white rectangle on top of it appears.

Thanks for the valuable feedback :)

I checked the rectangle above the select box - i think you meant that one. I couldn't remove it and it also exists without this PR. That space is not margin nor padding, what else could it be :D Anyway, we can deal with it later i think.

Other issues should be solved with the latest commits by now. Thank you.

eszkadev (approved):
now is perfect

PR #6852, opened 2023-07-07
Osc ctxs and offscreen canvases are write only

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -606,11 +606,6 @@ L.TileSectionManager = L.Class.extend({
 	},
 
 	_viewReset: function () {
-		var ctx = this._paintContext();
-		for (var i = 0; i < ctx.paneBoundsList.length; ++i) {
-			this._tilesSection.oscCtxs[i].fillStyle = this._sectionContainer.getClearColor();
-			this._tilesSection.oscCtxs[i].fillRect(0, 0, this._tilesSection.offscreenCanvases[i].width, this._tilesSection.offscreenCanvases[i].height);
-		}
 	},
 
 	_getZoomDocPos: function (pinchCenter, paneBounds, splitPos, scale, findFreePaneCenter) {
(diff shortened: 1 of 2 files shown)


caolanm:
AFAICS we only write/draw *to* those offscreenCanvases/oscCtxs things and nothing reads/draws from them.

Looks to me the last read from them was removed with:

commit 1f3d5e0a3e6c90d8ba8bc8471cbcbfa94917dda9
Date:   Fri Sep 10 13:42:33 2021 +0530

    zoomframe: directly use tiles to draw zoom-frames
    
    Move the zoom frame painting to tile-section.

mmeeks:
Sorry tilewins conflicted - but looks good otherwise - lets dung stuff out !

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -605,9 +605,6 @@ L.TileSectionManager = L.Class.extend({
 		this._sectionContainer.requestReDraw();
 	},
 
-	_viewReset: function () {
```
mmeeks, line 608: There are quite a few _viewResets in sub-classes (?)

mmeeks (commented):
Generally looks great - just a tad nervous about the _viewReset s- but the other bits look fun =) nice to dung that out [!]

caolanm:
as far as I can see, before that last commit there are two _viewReset both in browser/src/layer/tile/CanvasTileLayer.js
One is TileSectionManager::_viewReset and the other is CanvasTileLayer::_viewReset

L.TileSectionManager = L.Class.extend({
...
        _viewReset: function () { // L.TileSectionManager = L.Class.extend({
        },
...
});

and 

L.CanvasTileLayer = L.Layer.extend({

        onAdd: function (map) {
                ...
                this._viewReset(); // CanvasTileLayer::_viewReset: function (e) with no e is called here

                ...
        },


        getEvents: function () {
                var events = {
                        ...
                        viewreset: this._viewReset, // CanvasTileLayer::_viewReset
                        ...
                };

                return events;
        },

        _viewReset: function (e) {
                this._reset(this._map.getCenter(), this._map.getZoom(), e && e.hard);
                if (this._docType === 'spreadsheet' && this._annotations !== 'undefined') {
                        app.socket.sendMessage('commandvalues command=.uno:ViewAnnotationsPosition');
                }

                this._painter._viewReset(); // and calls TileSectionManager::_viewReset because this._painter = new L.TileSectionManager(this))
        },
});

PR #6854, opened 2023-07-07
language: introduce option to change language of only para or selection

lpranam:
Change-Id: If9086e72f53c3620c9d3e605b9e0415bdebab730

* Target version: master 

### Checklist


browser/src/control/Control.StatusBar.js
```diff
@@ -585,6 +591,11 @@ L.Control.StatusBar = L.Control.extend({
 
 		toolbaritems.push({ id: 'morelanguages', text: _('More...') });
```
eszkadev, line 592: I'm wondering if `More...` should be then replaced with: `Set language for All` to be unified with other entries

eszkadev: cc @pedropintosilva

lpranam: to me, it makes "more..." sense 😂 . agreed to @eszkadev

pedropintosilva:
Before:
the dropdown has just some languages. For example in the following screenshto the dropdown doesn't show Russian but after pressing `More...` more languages appear in the dialog
`More...`
image

-----
Now: compiling ....

browser/src/control/Control.StatusBar.js
pedropintosilva: Sorry, I missed this notification.

Thanks @lpranam yup let's please replace `More...` with `Set Language for All text`

pedropintosilva:
all tests passed, so that string change I don;t think we need to wait again for everything to pass

pedropintosilva (approved):

PR #6860, opened 2023-07-09
Private/timar/checkci

timar:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
@mmeeks this test was primarily for parallelizing unit tests and cypress tests. The result is negative. With `make -j4` cypress failed with many errors, without parallelism it passed.

PR #6862, opened 2023-07-09
private/marco/a11y/cont edit/remove duplicated code

mcecchetti:
- Adding support for enabling/disabling a11y support from the ui
- remove dup code: A11yTextInput: removed unused code
- remove dup code: TI: keep _hasWorkingSelectionStart for logging only
- remove dup code: TI: _onCommandResult: clear content in desktop case too
- remove dup code: TextInput: uniformed code fragments to A11yTextInput
- remove dup code: A11yTI: _onKeyDown uniformed to TI version
- remove dup code: A11yTI: _onInput mentions part uniformed to TI version
- remove dup code: A11yTI: _sendKeyEvent uniformed to TI version
- TextInput: gecko: fixed not working delete key
- remove dup code: A11yTextInput inherits from TextInput
- cypress tests for accessibility toggle ui


Rash419 (approved):
Thanks, Looks good to me!

just one I found:
it throws errors when accessbility is disabled from UI. I think as a11IT layer is getting removed  we get this error. Therefore adding a simple condition to all the a11 messages in CanvasTileLayer.js  might fix the error
```
Exception TypeError: this._map._textInput.onAccessibilityCaretChanged is not a function emitting event a11ycaretchanged: {
    "position": "43"
}
```

mcecchetti:
> Thanks, Looks good to me
> 
> just one thing I found: it throws errors when accessbility is disabled from UI. I think as a11IT layer is getting removed we get this error. Therefore adding a simple condition to all the a11 messages in CanvasTileLayer.js might fix the error
> 
> ```
> Exception TypeError: this._map._textInput.onAccessibilityCaretChanged is not a function emitting event a11ycaretchanged: {
>     "position": "43"
> }
> ```
> 
> maybe can be added in the next patch : )

Thanks for pointing out this issue as we agreed on mm I will provide a core patch for fixing this problem.

PR #6863, opened 2023-07-10
Share code for findTailsGroup

eszkadev:
Introduce getTailsGroupRect method which will be different in every implementation but share findTailsGroup logic using above function.

This fixes code duplication detected by: https://github.com/CollaboraOnline/online/pull/6802


gokaysatir (approved):
Thanks Szymon :)

PR #6866, opened 2023-07-10
snackbar: do not block users

eszkadev:
Make snackbar non-modal
    
    Snackbar shoudn't block user interactions with other
    UI components. This fixes regression introduced with
    vex rework to JSDialogs where snackbar become modal.
    
    This fixes problem with "send feedback" and "Zotero"
    snackbars which block users for ~10s from editing.

------------

Center IFrameDialog using CSS
    
    instead of using hardcoded pixel distance...

### DIFF
--- a/browser/css/iframedialog.css
+++ b/browser/css/iframedialog.css
@@ -1,8 +1,13 @@
 .iframe-dialog-wrap {
 	position: absolute;
 	z-index: 1106;
-	left: calc(50% - 225px);
-	top: calc(50% - 315px);
+	left: 50%;
+	top: 50%;
+}
+
+.iframe-dialog-content {
+	margin-left: -50%;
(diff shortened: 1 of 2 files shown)


browser/css/iframedialog.css
```diff
@@ -1,8 +1,13 @@
 .iframe-dialog-wrap {
 	position: absolute;
 	z-index: 1106;
-	left: calc(50% - 225px);
-	top: calc(50% - 315px);
+	left: calc(50%);
+	top: calc(50%);
```
relikd, line 5: no need for calc() ;)

eszkadev: yes, done :)

pedropintosilva (approved):
Beautiful !

PR #6868, opened 2023-07-10
Update Dockerfile

rodinux:
The sim-stable image do not found curl and broke the build

* Resolves: build image from-package 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva:
@rodinux :Welcome and thanks for your 1st PR! 

I have added @timar as a reviewer.

From my side I was just wondering if curl is missing or some other particular tool wouldn't be better to just add that on top of slim instead of changing it to bullseye?

Issue #6861, opened 2023-07-09
Mobile sidebar header layout


Andreas-Kainz:
I suggest to make the sidebar header layout more similar to the regular sidebar header layout. Maybe we can share css vars.

| Mobile | Desktop |
| ----------- | ------------- |
| ![Screenshot_20230709-150912_1.png](https://github.com/CollaboraOnline/online/assets/8517736/e0d8c411-81c8-4d9d-affb-bdaed2f3125e) | ![CollaboraOnline_220510_Writer_Zotero_CitationPreference_FirstCitation_edit_CP-1320x743_1.png](https://github.com/CollaboraOnline/online/assets/8517736/aa486146-9f28-43a2-a3fc-d13df6854c97) |

Andreas-Kainz:
@pedropintosilva any feedback about mobile sidebar layout for the headers at the example for spacing:

Andreas-Kainz:
@pedropintosilva any feedback about mobile sidebar layout for the headers at the example for spacing:

pedropintosilva:
Thanks for reporting @Andreas-Kainz 

# Context
## Bottom Line
In the past we set the headings to include a line under them to help users spot where the sections end and start (specially useful when scrolling quickly up and down within mobile wizard

## Problems
We were getting some headings that were not needed (we already have that info repeated in the label of the commands etc) and so we ended up hiding  some of them (with display: none)

The thing is now we have more and probably we should double check if there is any other in need to be hidden

# Layout
> I suggest to make the sidebar header layout more similar to the regular sidebar header layout. 

I agree we should improve those. But I don't know what you have in mind ^. 
I was thinking that we should simplify:
* Use similar gray header as in the sidebar (sized adapted to mobile). After all this is just a non-interactive element
* Remove the bottom line
* Add divider in between sections (so, above the header) 
 basically something similar to what people expect (see android's navigation drawer headers with dividers https://m3.material.io/components/navigation-drawer/overview )

Andreas-Kainz:
I will make additional changes at the mobile "sidebar" layout, but step by step and with issues where we can discuss.

PR #6872, opened 2023-07-10
Mobile sidebar header layout #6861

Andreas-Kainz:
Change-Id: I8c10857b0ca7e166b5f9364db9ea2a01e5acfffa

* Resolves: #6861 
* Target version: master 

### Summary

Follow Androids navigation drawer divider
https://m3.material.io/components/navigation-drawer/specs#25f67916-87fa-4cbb-b390-175096ae559b

Screenshot_20230711_001250


pedropintosilva (changes requested):
Thanks! 

| This will fail in any web browser that doesn't support webkit vendor prefix | width |
| ---------------------------------------------------------- | ----------------------------------------------------------  |
| Screen Shot 2023-07-12 at 09 32 40 | So the way I see it we could add all the other vendor prefix -moz-available but if we are already setting the margin in % I don't see any reason to don't just use one width set to percentage also  |

Andreas-Kainz:
thanks for the feedback.

pedropintosilva (approved):

PR #6874, opened 2023-07-10
DarkMode checkboxes will be themed

Andreas-Kainz:
Icon's are already available, the css rule has to be themed as for the other toolbar icons.

image

Change-Id: I0ffd594ba3e897a36c2caa8a5fe9aace9e04a63d


pedropintosilva (approved):
Thanks!

PR #6885, opened 2023-07-12
comments: fix hidden comment after lost focus in modification

eszkadev:
after commit 51d19dd60e268ddf0a087c38dc3ad8290f4ac4c2 annotation: avoid saving comments on losing focus

We had regression causing collapsed comment to be completly hidden if user first opened 'modify mode' and clicked outside comment area. Then comment was collapsed but when selected again - content wasn't shown.

This commits makes us sure that when we select comment it has visible content.


lpranam (approved):
works as expected :)

Issue #6505, opened 2023-06-02
Missing icon list in darkmode & mobile view 


Darshan-upadhyay1110:
**Describe the bug**

There are some icons are missing in dark mode that are list down below 

1. below svg are mentioned on toolbar.css as dark theme but not available in dark images folder.

- plus.svg
- minus.svg
- lc_view100.svg

[ to see this icon not visible change to dark mode and see bottom toolbar righ corner ]

![image](https://github.com/CollaboraOnline/online/assets/61383886/8203321a-2897-4002-acd8-210ec0bd88bf)

it should be like this 

![image](https://github.com/CollaboraOnline/online/assets/61383886/7a5dac09-352e-482d-a0c5-911a5eef7473)

- [x] #6661
- [x] #6662
- [x] #6663
- [x] #6664
- [x] #6665
- [x] #6666
- [x] #6667
- [x] #6668
- [x] #6669
- [x] #6670
- [x] #6671
- [x] #6672
- [x] #6673
- [x] #6674
- [x] #6675
- [x] #6676
- [x] #6677
- [x] #6678
- [x] #6679
- [x] #6680
- [x] #6681
- [x] #6682
- [x] #6683
- [x] #6684

2. In mobile view 

Steps:

Open a .odt file
Click on the Add icon > Go to Formattinng Mark.
Notice that Word Joiner has no icon

![image](https://github.com/CollaboraOnline/online/assets/61383886/bc177b11-28a3-4a59-b30e-f03d8c9d28b5)
![image](https://github.com/CollaboraOnline/online/assets/61383886/7fc4cc5c-9c8d-4990-9d63-296cdb184b5b)

- [ ] #6714
- [ ] #6715
- [ ] #6716
- [ ] #6717

4. here are the list of icons which is not present in dark theme but for that we use the normal images present in images folder. ( not sure that this we need to add in dark but if any better icons are there to give rich look in dark mode for following icons then it will be great to add it )

Darshan-upadhyay1110:
Please feel free to add icons description and name which are missing.

rpear:
* plus.svg, minus.svg, lc_view100.svg
- These do appear in dark view in Online desktop 24.04.14.2, so good.
- These do not appear on the status bar in the Android app in desktop mode on a Chromebook, as zooming this way is not an option, but 2 finger trackpad zooming works well, I don't miss it, it would be a new bug if it matters.
[...]

PR #6889, opened 2023-07-12
Dark Icon Theme missing icons #6505

Andreas-Kainz:
notebookbar download icons use dark variant

Change-Id: Ic0efbe2862dd28d60defcc45361d6d2b6ff61775

* Resolves: #6505 
* Target version: master 

still icons are missing


browser/css/notebookbar.css
```diff
@@ -633,5 +633,62 @@ button.ui-tab.notebookbar {
 }
 
 [class*='as-pptx-submenu-icon'] {
-	background: url('images/lc_downloadas-pptx.svg') center;
```
pedropintosilva, line 636: Was this intentional? It changes the light icon to dark when using light theme.

pedropintosilva (changes requested):

browser/css/notebookbar.css
Andreas-Kainz: no, thank's for the review. will be fixed.

PR #6895, opened 2023-07-13
Fix pasting image assertion

eszkadev:
Revert "Avoid crash when pasting image in debug mode" This reverts commit 78558fe9af361dd0238bd3021c67e9d431dcf981.

Instead of detecting paste command and not trigerring assertion: fix frame type so it will be binary in case of paste.

### DIFF
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -778,16 +778,11 @@ protected:
             size_t offset = Util::isValidUtf8((unsigned char*)data, len);
             if (offset < len)
             {
-                static const char *imgModel = "child-000 paste mimetype=image/";
-                static std::regex imgRegex("^child-[0-9a-f]{3} paste mimetype=image/$");
-                bool isImagePaste = std::regex_match(data, data + strlen(imgModel), imgRegex);
-                if (!isImagePaste) {
-                    std::string raw(data, len);
-                    std::cerr << "attempting to send invalid UTF-8 message '" << raw << "' "
-                            << " error at offset " << std::hex << "0x" << offset << std::dec
-                            << " bytes, string: " << Util::dumpHex(raw) << "\n";
-                    assert("invalid utf-8 - check Message::detectType()" && false);
-                }
+                std::string raw(data, len);
+                std::cerr << "attempting to send invalid UTF-8 message '" << raw << "' "
+                          << " error at offset " << std::hex << "0x" << offset << std::dec
+                          << " bytes, string: " << Util::dumpHex(raw) << "\n";
+                assert("invalid utf-8 - check Message::detectType()" && false);
             }
         }
 #endif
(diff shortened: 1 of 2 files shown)


wsd/ClientSession.cpp
```diff
@@ -1343,7 +1343,11 @@ bool ClientSession::sendCombinedTiles(const char* /*buffer*/, int /*length*/, co
 bool ClientSession::forwardToChild(const std::string& message,
                                    const std::shared_ptr<DocumentBroker>& docBroker)
 {
-    return docBroker->forwardToChild(client_from_this(), message);
+    bool binary = false;
+    if (Util::startsWith(message, "paste"))
+        binary = true;
```
Ashod, line 1348: `const bool binary = Util::startsWith(message, "paste");`

net/WebSocketHandler.hpp
```diff
@@ -778,16 +778,11 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
-                bool isImagePaste = std::regex_match(data, data + strlen(imgModel), imgRegex);
-                if (!isImagePaste) {
-                    std::string raw(data, len);
-                    std::cerr << "attempting to send invalid UTF-8 message '" << raw << "' "
-                            << " error at offset " << std::hex << "0x" << offset << std::dec
-                            << " bytes, string: " << Util::dumpHex(raw) << "\n";
-                    assert("invalid utf-8 - check Message::detectType()" && false);
-                }
+                std::string raw(data, len);
+                std::cerr << "attempting to send invalid UTF-8 message '" << raw << "' "
```
Ashod, line 782: Logging should be very useful, when one is not debugging. But more important, in an optimized build, the assertion won't fire, and we will be writing to stderr, which may get lost and we wouldn't know that we are hitting this case.

Suggestion: serialize to `std::ostringstream` and then write the contents to `LOG_ERR` (`std::cerr` can we written to in debug build, to help debugging, but I don't see how it could be useful for production).

Ashod (approved):
Thanks @eszkadev. Indeed this seems a better fix.

I made two small suggestions, but no blockers from me.

net/WebSocketHandler.hpp
```diff
@@ -778,16 +778,11 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
+                          << " error at offset " << std::hex << "0x" << offset << std::dec
+                          << " bytes, string: " << Util::dumpHex(raw) << "\n";
+                assert("invalid utf-8 - check Message::detectType()" && false);
```
Ashod, line 785: `assert(!"invalid utf-8 - check Message::detectType()");`

Seems more understandable to me.

wsd/ClientSession.cpp
eszkadev: done

PR #6900, opened 2023-07-13
wsd: avoid cursor invalidation if cursor becomes not visible

lpranam:
this also helps in avoiding flickering of keyboard in mobile in some cases

Change-Id: I59312ca93136cd6bd13ad958dab0e72242f839e1

* Target version: master 

### Checklist


eszkadev (changes requested):
filtering not needed messages happens in the core init.cxx

PR #6905, opened 2023-07-14
fix: media cannot play with indirection server

Rash419:
Change-Id: I83d39fa4829938e8f217361fec3fbdab79adbac9

* Target version: master 

show error message:
image

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -745,10 +745,11 @@ window.app = {
 				};
 			} else if (this.status === 202) {
 				var msg = {
-					'MessageId': 'App_LoadingStatus',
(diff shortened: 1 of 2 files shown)


mmeeks:
Looks good to me - we should almost certainly add similar error popups for the copy/paste case - and send the ServerId that we are expecting to the other end and allow it to respond with: "wrong server" and show a similar popup error - to catch mis-configured clusters =) Thanks.

Rash419:
@mmeeks sure , Lets merge this. I will create a seprate PR for that : )

Rash419:
@mmeeks follow up PR: https://github.com/CollaboraOnline/online/pull/6915

PR #6906, opened 2023-07-14
Enter shouldn't be a requirement to trigger a tab or a button

Darshan-upadhyay1110:
Change-Id: I9ab9e004e9dc883d18b8f750a9a400e1f318240e

* Resolves: 
* Target version: master 

### Summary

Enter key will not be necessary to change tab by accessibility key

### DIFF
--- a/browser/src/dom/NotebookbarAccessibility.js
+++ b/browser/src/dom/NotebookbarAccessibility.js
@@ -225,12 +225,6 @@ var NotebookbarAccessibility = function() {
 				this.resetState();
 			}
 		}
-		else if (key === 'ENTER' || key === ' ') {
-			if (this.filteredItem !== null)
-				this.clickOnFilteredItem();
-			else
-				app.map.focus();
-		}
 		else if (event.keyCode === 16) // ShiftLeft.
 			return; // Ignore shift key.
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):

pedropintosilva (changes requested):
Thanks @darshan now it works without the need for pressing the `enter` key but we are not setting the focus correctly. When pressing `H` key to switch to `Home` tab it not only should switch to that tab but it also should set the focus there (an  outline should appear). Currently I see no focus nor focus-visible being set in the browser). Do you think you could add an extra commit for that?

Darshan-upadhyay1110:
@pedropintosilva sure

pedropintosilva:
> @pedropintosilva i confused by your comment as i am seeing focus changed to that tab after pressing key 
> image
> 
> Can you explain extact beahviour for this ? ( not getting point for outline ) 

Hm , let's merge this and work on the focus in another issue/PR (basically triggering for example the Insert tab with keyboard doesn't set 'focus-visible' = meaning pressing then Tab key doesn't work to change component)

Darshan-upadhyay1110:
oh i see got your point. i will start to work on that in other PR. Thanks :)

PR #6912, opened 2023-07-15
Fix android bugs

Rash419:
* Resolves: # 
* Target version: master 

regression from: https://github.com/CollaboraOnline/online/commit/17f739ca7710428765c40d5daac5fdaecfe708e2
https://developer.android.com/about/versions/11/privacy/storage


timar (approved):
This looks very plausible. Let's get it in and test it with snapshots. Thanks for investigating, @Rash419 !

Issue #6877, opened 2023-07-11
Menubutton should have disabled state styling different


eszkadev:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva 

### Detailed description and rationale
"Menu button" widget (button with dropdown) should have disabled state handled in CSS so user will know that it is disabled.

### Code pointers
Example dialog to test: Writer -> Review -> Manage Changes
[disabled-menubutton-2023-07-11_13.40.41.webm](https://github.com/CollaboraOnline/online/assets/5307369/cf906acc-4339-41db-9571-f21609dc5ee4)

codewithvk:
Can i work on it?

eszkadev:
@codewithvk sure :)

Harsh8055:
@codewithvk you working on it?

codewithvk:
@Harsh8055 Yes.

PR #6914, opened 2023-07-16
Fix : Menubutton should have disabled state styling different

codewithvk:
Change-Id: I1cd97503e0d3ca4333b11552ca3571093cf10bfe

* Resolves: #6877 
* Target version: master 

image

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1030,6 +1030,15 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	margin: 5px 0px 5px 5px;
 }
 
+.jsdialog.menubutton[disabled] {
+	background-color: var(--color-background-lighter) !important;
+	color: var(--color-text-lighter) !important;
+}
+
+.jsdialog.menubutton[disabled] .arrow {
+	border-top-color: var(--color-text-lighter);
+}
+
 /* sum button in formulabar */
 .formulabar.menubutton {
 	padding: 0px !important;


eszkadev:
Nice! it fixed enabled/disabled state and also text alignment inside.
One probably unwanted thing is that now menubutton become higher when compared to other listboxes / edit fields:
alignment

@pedropintosilva what do you think?

pedropintosilva (changes requested):
Thanks @codewithvk !

I think with this change we still get the misalignment , now a bit more visible. And it seems we are not really targeting the disabled menubutton and addressing the lack of styling for those.

Here is what I would prefer:
- Avoid adding more exceptions to btns.csss
- Properly set all the missing styles:
```css
.jsdialog.menubutton[disabled] span {
	color: var(--color-text-lighter);
}

.jsdialog.menubutton[disabled] .arrow {
	border-top-color: var(--color-text-lighter);
}
```
to https://github.com/CollaboraOnline/online/blob/master/browser/css/jsdialogs.css#L1032

And maybe in additional commit make sure that regular menubutton has the same size as the other elements

codewithvk:
Hi @pedropintosilva,

Thank you for your valuable suggestions and feedback on my pull request. I have incorporated several changes based on your suggestions, and I appreciate your help in improving the code.

I wanted to share a small suggestion for future interactions. While I value the code you provided, I believe it would be more beneficial to my learning process if you could offer hints or suggestions rather than providing the code directly. This way, I can have the opportunity to explore and implement the solutions on my own, which will greatly contribute to my growth as a developer.

Regarding your additional comment on making sure that the regular menubutton has the same size as the other elements, I agree that it's a good idea. However, to keep the changes focused and organized, I will address this separately in a new commit and create a separate pull request once the current one is merged.

Once again Thanks :)

pedropintosilva:
Cool, I'll check it. 

Yup I agree and I appreciate the feedback :) 

Granted the ins and outs of what should be where can be a bit all over the place and so I wanted to offer some initial aid (I just shouldn't over do it :)

pedropintosilva:
@codewithvk thanks. Can you also make sure it doesn't reach to hover state when it's disabled?

codewithvk:
> @codewithvk thanks. Can you also make sure it doesn't reach to hover state when it's disabled?

Do i have to tack care of all the inputs or just for the date picker? I can see all inputs are hover even they are disabled.
image

pedropintosilva:
Let me merge this and please create a follow up pr that fixes all the hover states on disabled items :)

codewithvk:
> Let me merge this and please create a follow up pr that fixes all the hover states on disabled items :)

Sure, Thanks!

PR #6922, opened 2023-07-18
Calc: fix document background. It should be white.

pedropintosilva:
With c04c065edcfe9e00e4f7e50834e6317caaceb9ea we start to dynamically
set the canvas background. The problem is that for Calc cells do not
have intrinsically a background and thus what we end up to see as the
document background color is actually the canvas color.
	 - Make sure that when we are in presence of calc we set the
	 canvas color (clearColor) with the value coming from
	 documentBackgroundColor
	 - To make it clear: create a new CSS variable to be used as
	 background-document

Change-Id: I4c456e4aeea233609ee58684d967984dd8afd63a

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -12,6 +12,7 @@
 	--color-text-lighter: #fff; /* secondard text, disabled */
 
 	--color-canvas: #121212;
+	--color-background-document: #121212;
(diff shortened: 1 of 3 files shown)


OFFTKP:
looks good to me

OFFTKP (approved):

PR #6924, opened 2023-07-18
Don't add outline to ShowInfoModal that have no buttons

pedropintosilva:
Large Clipboard  and document idle modals were being set with tabindex
0 allowing for manual user focus plus were being focused dynamically
for each instance:
    - Transfer all that (just once) to showinfomodal instead for the
    cases where we have buttons
        - This in turn sets the things to the parent (dialog itself)
	and not the subcontianer
        - Set tabIndex with negative value so, it only allows focus via
    code
    - This fixes the weird outline (focus) that sometimes would appear
    on sub containers of that dialog instances

Change-Id: If40192e05c4b0e25e2ce71bc612ab64fc54a29ec

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1057,6 +1057,13 @@ L.Control.UIManager = L.Control.extend({
 				that.closeModal(dialogId);
 			}}
 		], cancelButtonId);
+		if (!buttonText && !withCancel) {
+			// if no buttons better to set tabIndex to negative so the element is not reachable via sequential keyboard navigation but can be focused programatically
+			document.getElementById(dialogId).tabIndex = -1;
+			// We hid the OK button, we need to set focus manually on the popup.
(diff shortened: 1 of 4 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -1057,6 +1057,13 @@ L.Control.UIManager = L.Control.extend({
 				that.closeModal(dialogId);
 			}}
 		], cancelButtonId);
+			if (!buttonText && !withCancel) {
```
eszkadev, line 1060: indent is wrong?

eszkadev: fixed

eszkadev (approved):

browser/src/control/Control.UIManager.js
pedropintosilva: :facepalm: thanks

PR #6926, opened 2023-07-18
Fix Spell checker js Dialog

Darshan-upadhyay1110:
Change-Id: Idc0800eaa4ab751601bd34bba6c2e9cfef38de4d

* Resolves: #
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -41,11 +41,6 @@
 	min-width: 400px;
 }
 
-img.ui-drawing-area {
-	width: -webkit-fill-available;
-	width: -moz-available;
-}
-
 .jsdialog.one-child-popup {
 	border: 0;
 	margin: 0 !important;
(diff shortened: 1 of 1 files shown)


browser/css/jsdialogs.css
```diff
@@ -42,8 +42,7 @@
 }
 
 img.ui-drawing-area {
-	width: -webkit-fill-available;
-	width: -moz-available;
```
eszkadev, line 46: this breaks writer -> format -> theme (it is cut off)

eszkadev: fill available makes things blurry...  yes, please remove it

I think we can fix theme dialog case by adding:
`width: max-content;` to `.jsdialog.ui-scrollwindow`

and that is all we need :)
you can remove #suggestionslb rule and this img.ui-frawing-area rule

Darshan-upadhyay1110: ahh right !

eszkadev (approved):

PR #6932, opened 2023-07-19
wsd: trim memory on idling

Ashod:
When all users idle their activities
for at least 10 seconds, we take the
opportunity to trim the memory footprint.

We are careful not to issue a subsequent
trimming too soon. A minimum of 30 seconds
needs to pass to trim again. This should
give a decent balance between performance
and reduced memory footprint.

Change-Id: I1c8b2c4a265c9aa0ac5033762695706ad970f11e

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -13,6 +13,7 @@
 #include <config_version.h>
 
 #include <dlfcn.h>
+#include <limits>
 #ifdef __linux__
 #include <ftw.h>
 #include <sys/vfs.h>
@@ -708,6 +709,7 @@ public:
         _stop(false),
         _editorId(-1),
         _editorChangeWarning(false),
+        _lastMemTrimTime(std::chrono::steady_clock::now()),
         _mobileAppDocId(mobileAppDocId),
         _inputProcessingEnabled(true)
     {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):

kit/Kit.cpp
```diff
@@ -982,6 +984,34 @@ class Document final : public DocumentManagerInterface
...
+        double minInactivityMs = std::numeric_limits<double>::max();
+        for (const auto& it : _sessions)
+        {
+            minInactivityMs = std::min(it.second->getInactivityMS(), minInactivityMs);
+        }
+
+        if (minInactivityMs >= 9999)
+        {
+            LOG_DBG("Trimming Core caches");
+            _loKit->trimMemory(4096);
```
mmeeks, line 1006: So lets just pass a smaller value here - say 1024 - so the core can differentiate between a deep-clean of everything and an interactive try to cleanup ? =)

kit/Kit.cpp
```diff
@@ -982,6 +984,34 @@ class Document final : public DocumentManagerInterface
...
+
+            LOG_DBG("Dropping delta caches");
+            _deltaGen.dropCache();
```
mmeeks, line 1009: Our delta cache saves a ton of CPU and bandwidth, and improves interactivity - we should not drop this easily - please remove this. Possibly we could save some memory there - but the delta cache should be only ~10Mb or so per session I think.

kit/Kit.cpp
```diff
@@ -2343,11 +2376,16 @@ class KitSocketPoll final : public SocketPoll
...
             else
-                LOG_TRC("Poll of woudl not close gap - continuing");
+                LOG_TRC("Poll of would not close gap - continuing");
         }
 
         drainQueue();
 
+        if (_document)
+        {
+            _document->trimIfExcessive();
```
mmeeks, line 2386: Naming-wise is there anything Excessive here ? =) we have trimIfInactive already trimOnShortIdle ? ;-) up to you but - what was Excessive ?

mmeeks (commented):
Outline looks good, some detail tweaks =)

PR #6935, opened 2023-07-19
Adjacent tiles prefetching with delay

eszkadev:
While scrolling using scrollbar we tried to prefetch lots of tiles adjacent to the view.

Introduce timer which will protect us from prefetching tiles we will not need when scrolling is fast.
We will avoid doing it synchronously on every little step.

_update is called a lot in onMouseMove handler:

_update (CanvasTileLayer.js:6220)
...
panBy (Map.PanAnimation.js:51)
scroll (Scroll.js:11)
ScrollSection.scrollVerticalWithOffset (ScrollSection.ts:657) ScrollSection.onMouseMove (ScrollSection.ts:776)
CanvasSectionContainer.propagateOnMouseMove (CanvasSectionContainer.ts:1097) CanvasSectionContainer.onMouseMove (CanvasSectionContainer.ts:1403)

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -6211,7 +6211,15 @@ L.CanvasTileLayer = L.Layer.extend({
 		if (queue.length !== 0)
 			this._addTiles(queue);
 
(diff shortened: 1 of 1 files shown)


Rash419 (approved):
Looks good to me

PR #6937, opened 2023-07-19
rle: optimize inner loop bit-bashing.

mmeeks:
Simplified code:

+ re-use the bit mask as an inner loop counter.
+ turn alpha only rows into zero length with agreed lastPix.
+ keep lastPix around on the stack.
+ handle odd widths in a duplicate slow-path

Change-Id: Ibc7630f7187ea5f4904c6fed14dda28cdfbf7523

rle: further optimize the pixel reading & de-compression.

Change-Id: Ifcaaa45d1de6532b3fd43015c47a37daf56c2ea5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Delta.hpp
+++ b/kit/Delta.hpp
@@ -74,31 +74,53 @@ class DeltaGenerator {
         class PixIterator final
         {
             const DeltaBitmapRow &_row;
-            unsigned int _x;
-            const uint32_t *_rlePtr;
+            unsigned int _nMask; // which mask to operate on
+            uint32_t _lastPix; // last pixel (or possibly plain alpha)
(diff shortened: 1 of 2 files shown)


caolanm:
==1626772==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x607000011020 at pc 0x000000e75798 bp 0x7ffff9dc4c30 sp 0x7ffff9dc4c28
READ of size 4 at 0x607000011020 thread T0
    #0 0xe75797 in DeltaGenerator::DeltaBitmapRow::PixIterator::next() collabora-online/test/../kit/Delta.hpp:113:32

caolanm (approved):

PR #6938, opened 2023-07-19
Add Left or right arrow on a tab label should switch tabs

Darshan-upadhyay1110:
Change-Id: Id8c8e2a8bcb57c90e0762dd3916b851fff8e0a6f

* Resolves: #6934 
* Target version: master 

### Summary
- accessibility key features 
   - focus on tab when press `Alt`.
   - move back and forth tab focus by ArrowLeft and ArrowRight

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/dom/NotebookbarAccessibility.js
+++ b/browser/src/dom/NotebookbarAccessibility.js
@@ -119,12 +119,14 @@ var NotebookbarAccessibility = function() {
 	};
 
 	this.onInputFocus = function() {
+		this.addTabFocus();
 		document.body.classList.add('activate-info-boxes');
 	};
 
 	this.onInputBlur = function() {
 		document.body.classList.remove('activate-info-boxes');
 		app.map._textInput._abortComposition({ type: 'Notebookbar Accessibility' });
+		this.removeFocusFromTab();
 	};
 
 	this.isAllFilteredOut = function() {
@@ -185,8 +187,9 @@ var NotebookbarAccessibility = function() {
 		if (this.filteredItem !== null) {
 			var element = document.getElementById(this.filteredItem.id);
 			if (element) {
+				this.removeFocusFromTab();
 				element.click();
-
+				this.addTabFocus();
 				if (this.state === 0) {
 					this.setupAcceleratorsForCurrentTab(element.id);
 					this.combination = null;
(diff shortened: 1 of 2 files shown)


browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -195,13 +198,37 @@ var NotebookbarAccessibility = function() {
...
 		}
 		else
-			app.map.focus();
+			this.focusToMap();
+	};
+
+	this.addTabFocus = function() {
+		var element = this.getCurrentSelectedTab();
+		if (element) {
+			element.classList.add('add-focus-to-tab');
```
pedropintosilva, line 213: can't we just make the browser using the `:focus-visible` pseudo class to the element by, when the arrow keys are pressed we
1. `element.focus();`
2. Then switch tab

?

Plus I'm also getting sometimes that border even when the focus is already in the document and the active tab is a different one

pedropintosilva (changes requested):

browser/src/dom/NotebookbarAccessibility.js
Darshan-upadhyay1110: yes we can make direct focus to tab elment by foucs but if we do that then  accessibility info keys will be focus out and in simple case we lost all info tags like `N` for insert because focus shifted to tab.

Darshan-upadhyay1110: actully if you see this addTabFocus is only used when we have focus to accessibilty keys. 

we need 2 things at a time focus should be on accessibility keys and
that active tab should highlighted 

so to add explicit focus thing on element at same time i have introduced this class

pedropintosilva (approved):
Awesome, Thanks!

pedropintosilva:
:heavy_check_mark: arrow right and left cycles through tabs
:heavy_check_mark: the tab get correctly the browser's focus
   * It gets CSS class with border (fallback) but it also gets the focus and then browser styles it accordingly to its focus-visible default CSS rules

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -229,21 +256,59 @@ var NotebookbarAccessibility = function() {
...
+				break;
+			}
+			previousTab = previousTab.nextElementSibling;
+			}
+			this.removeFocusFromTab();
+			previousTab.click();
+			previousTab.focus();
+		}
+		else if (key === 'ARROWLEFT') {
+			var currentSelectedTab = this.getCurrentSelectedTab();
```
eszkadev, line 295: it would be good to create method getNextTab() to not duplicate code in ARROWLEFT / RIGHT

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -229,21 +256,59 @@ var NotebookbarAccessibility = function() {
...
+			var previousTab = currentSelectedTab.previousElementSibling;
+			if (!previousTab) {
+				var elements = document.querySelectorAll('.ui-tab.notebookbar:not(.hidden)');
+				previousTab = elements[elements.length - 1];
+			}
+			while (previousTab) {
+			if (!previousTab.classList.contains('hidden')) {
```
eszkadev, line 302: indent

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -229,21 +256,59 @@ var NotebookbarAccessibility = function() {
...
 			}
 		}
+		else if (key === 'ARROWRIGHT') {
+			var currentSelectedTab = this.getCurrentSelectedTab();
+			var previousTab = currentSelectedTab.nextElementSibling;
+			if (!previousTab) {
+				previousTab = document.querySelector('.ui-tab.notebookbar:not(.hidden)');
+			}
+			while (previousTab) {
+			if (!previousTab.classList.contains('hidden')) {
```
eszkadev, line 284: indent

eszkadev (commented):
looks ok, just 2 minor style comments

browser/src/dom/NotebookbarAccessibility.js
eszkadev: thanks now looks good :)

PR #6940, opened 2023-07-19
refactor: Toolbar down icons: remove unnecessary !important rule

pedropintosilva:
If we set the initial backgrounds without `!important` then there is
no need to use `!important` for dark-mode (since we are already using
a more specific targe via data-them attribute)

Change-Id: I600629dfc0485edac6ff725f6278de1501e1c9a7


pedropintosilva:
passed code scanning and codeQL no need to wait for cypress, this doesn;t affect tests

PR #6943, opened 2023-07-20
Revert "Fix wrong device detection logic."

thebearon:
On actual iPad (2048x1536 native, 1024x768 browser), in landscape view in Safari, mobile phone view is shown in online instead of tablet.

This reverts commit f1e9ee72fe48d21ac7a78b7702055b91ba8f3ba9.

Change-Id: I2cd4ba6cdcd616760c06c7db479e6d13be3d83b4

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -253,7 +253,7 @@ window.app = {
 				return true;
 			}
 
-			return L.Browser.mobile && (window.innerWidth < 768 || window.innerHeight < 768);
+			return L.Browser.mobile && (screen.width < 768 || screen.height < 768);
 		},
 		// Mobile device with big screen size.
 		isTablet: function() {


pedropintosilva:
I tested master branch without this revert and I can confirm the mobile view appears on ipad (my case mini ipad)
I still didn't test it with this PR but here is some context:

The first commit was to fix the top menu that was appearing on mobile:
image

pedropintosilva (approved):
I double check it and I tracked the values outputted  by

```
if (window.innerWidth > 0)
+                alert(window.innerWidth + ' | ' + window.innerHeight)
```

 on a phone and on a tablet and it's very interesting.. I think the weird menu that appears is something to do with the transitional readonly mode. Thus, I think this is safe to revert

I also tested with the revert and the initial top menu that appeared before and that initial commit was fixing , doesn't appear anymore on mobile both with and without the revert. Just on tablet

PR #6955, opened 2023-07-21
theme: use integration theme value from the interation message

lpranam:
problem:
till now we used values from DOM elements which were unreliable, sometimes they are not discoverable due to nested iframs. That threw some 404 due to incorrect theming path.

Change-Id: Ibc291ce9f64db799095e1edcb14c598bdd085de7

* Target version: master 

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -407,7 +407,7 @@ var brandingLink = document.createElement('link');
 brandingLink.setAttribute("rel", "stylesheet");
 brandingLink.setAttribute("type", "text/css");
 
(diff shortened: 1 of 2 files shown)


pedropintosilva (approved):
Thanks!

I'm now getting always the value.. This seems to be much more reliable (getting the value directly from FileServer.cpp instead of getting it from DOM).

Also Thanks for adding the additional comment for UI_THEME , this definitely helps as a reminder of what is what!

pedropintosilva:
It terms of safety. I think we should just merge this:
* The if statement didn't change, meaning if theme is not defined nothing will broke
* It's a very small change and fixes the 404 error

PR #6971, opened 2023-07-26
Fix Direct change by accesskey from Home to Insert tab triggers bullet command

Darshan-upadhyay1110:
Change-Id: Iaace6bfc446e91b638494a4dc69dce611b1e3bf8
Signed-off-by: Darshan-upadhyay1110 [darshan.upadhyay@collabora.com](mailto:darshan.upadhyay@collabora.com)

* Resolves: #
* Target version: master 

### Summary
 press alt (shift + alt) at all the tab's content shortcuts shouldn't be accessible and their hints shouldn't be visible.

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -512,6 +512,11 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				element.accessKey = key;
 	},
 
+	_resetAccessKeyForCurrentTab: function() {
+		if (app.UI.notebookbarAccessibility)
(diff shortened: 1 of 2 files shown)


pedropintosilva (approved):
It works but someone needs to code review it :)

Darshan-upadhyay1110:
Note:
•    Alt + N : should trigger the * notebookbar tab
•    N : Tab’s contents
o    Yes but the alt will always trigger things like insert bullet because it’s effectively using accessKey html attribute (and that’s probably how it should be since screen-readers might depend on that for aiding the user)

PR #6973, opened 2023-07-26
accessibilty: added accessibilty fields to the mobile edit button

lpranam:
Change-Id: Ifdb59ff3463a0354549316994f8e80b341e28133

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Permission.js
+++ b/browser/src/control/Permission.js
@@ -13,6 +13,10 @@ L.Map.include({
 	setPermission: function (perm) {
 		var button = $('#mobile-edit-button');
 		button.off('click');
+		button.attr('tabindex', 0);
+		button.attr('role', 'button');
+		button.attr('title', _('Enter editing'));
+		button.attr('aria-label', _('Enter editing'));
 		// app.file.fileBasedView is new view that has continuous scrolling
 		// used for PDF and we dont permit editing for PDFs
 		// this._shouldStartReadOnly() is a check for files that should start in readonly mode and even on desktop browser


browser/src/control/Permission.js
```diff
@@ -13,6 +13,9 @@ L.Map.include({
 	setPermission: function (perm) {
 		var button = $('#mobile-edit-button');
 		button.off('click');
+		button.attr('tabindex', 0);
+		button.attr('role', 'button');
+		button.attr('title', 'Enter editing');
```
pedropintosilva, line 18: title is for the tooltip. Which yup is nice to have but we should also add alt attribute

pedropintosilva (changes requested):

browser/src/control/Permission.js
lpranam: alt attribute does not exist for div technically. but should I add it anyway? image of the pencil is actually a div with the background.

pedropintosilva: aha so if it's not a button we should either:
1. Turn that thing into a button (and fix all the styles and things that will brake with that change)
2. Or for now use aria attributes https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA

lpranam: its already ARIA compatible for for button role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

lpranam: done

pedropintosilva (approved):

PR #6975, opened 2023-07-27
Tablet: Non-functional sidebar button shown in read-only mode

pedropintosilva:
Do not wrongly switch part of the UI to edit mode when on read-only
mode. Best to be only remove readonly css class when it's not needed as
opposed to removed it whenever refreshNotebookbar() is called.

Before this commit:
- We were showing tabs in wrong mode (readonly mode) -> every-time
refreshNotebookbar() was called. Thus, initial tablet mode was
appearing with icons and actions from edit mode.

Change-Id: I6ad9c8c14e60d4c3513f63b75717db0a124b73ea

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -432,11 +432,13 @@ L.Control.UIManager = L.Control.extend({
 		var selectedTab = $('.ui-tab.notebookbar[aria-selected="true"]').attr('id') || 'Home-tab-label';
 		this.removeNotebookbarUI();
 		this.createNotebookbarControl(this.map.getDocType());
+		if (this._map._permission === 'edit') {
+			$('.main-nav').removeClass('readonly');
+		}
 		$('#' + selectedTab).click();
 		this.makeSpaceForNotebookbar();
 		this.notebookbar._showNotebookbar = true;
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks good

browser/src/control/Control.UIManager.js
```diff
@@ -432,11 +432,13 @@ L.Control.UIManager = L.Control.extend({
 		var selectedTab = $('.ui-tab.notebookbar[aria-selected="true"]').attr('id') || 'Home-tab-label';
 		this.removeNotebookbarUI();
 		this.createNotebookbarControl(this.map.getDocType());
+		if (this._map._permission === 'edit') {
+			$('.main-nav').removeClass('readonly');
+		}
```
eszkadev, line 437: should we also add class in opposite case?

pedropintosilva: I thought about that. But I ended not doing because it seems control.Notebookbar.js:hideTabs is already doing that and control.Toolbar.js:onUpdatePermission as well (plus afraid to fix a inexistent problem and add regression :p )

eszkadev: ok, then better to not touch this. At some point we should unify that so it will be done in one place.

PR #6979, opened 2023-07-28
Add reset accessibilityKey

Darshan-upadhyay1110:
Change-Id: I04310f7ee4ab599a0c53c7aceaba459dd7641681

On focus away or if NotebookbarAccessibility is disabled then remove accesskey

* Resolves: #
* Target version: master


browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -349,6 +354,9 @@ var NotebookbarAccessibility = function() {
 					if (this.activeTabPointers.id !== element.id) {
 						this.setupAcceleratorsForCurrentTab(element.id);
 					}
+					if (!document.body.classList.contains('activate-info-boxes')) {
+						this.resetState();
+					}
 				}.bind(this));
 				element.addEventListener('keydown', function(event) {
 					if (event.key === 'Alt') {
```
gokaysatir, line 362: ESC key is (should) also working. Why did we need this? It's no harm though :)

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -349,6 +354,9 @@ var NotebookbarAccessibility = function() {
 					if (this.activeTabPointers.id !== element.id) {
 						this.setupAcceleratorsForCurrentTab(element.id);
 					}
+					if (!document.body.classList.contains('activate-info-boxes')) {
+						this.resetState();
+					}
```
gokaysatir, line 359: We shouldn't need this 'click' function entirely, including the "setupAcceleratorsForCurrentTab" part. Info boxes should be cleaned on blur and state should be reset at the same time.

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -238,6 +239,10 @@ var NotebookbarAccessibility = function() {
 		this.combination = null;
 		this.mayShowAcceleratorInfoBoxes = false;
 		this.filteredItem = null;
+		var accessKeyElements = document.querySelectorAll('.ui-content.unobutton');
+		for (var i = 0; i < accessKeyElements.length; i++) {
+			accessKeyElements[i].removeAttribute('accesskey');
+		}
 	};
```
gokaysatir, line 246: I would use a loop like:

loop (activeTabPointers.contentList.length)
    document.getElementById(activeTabPointers.contentList[i].id).accessKey = null;

And we can check the behaviour from scratch, please correct me if i am wrong. We need something like:
* At start, only tabs' access keys are assigned. Indeed they are never removed.
* When user presses Alt or Alt+Shift (needed on Firefox for example), info boxes are shown (state = 0).
* When user picks a tab's letter, tab's content keys are assigned (state = 1).
* When user picks a command or cancelles the process (possible on 'onblur'), content access keys are removed.

I guess we can check above behaviour and have a clean API hopefully. If you can check and improve my suggestions whereever you see possibilities, that'd be great.

Thanks :)

gokaysatir (changes requested):
Thanks for the PR :)

gokaysatir (approved):
Thanks for the improvements @Darshan-upadhyay1110 :)

PR #6992, opened 2023-07-31
Add support for URP messages in COOLWSD

Minion3665:
Change-Id: Idadfe288a78cfd72b01253dfdade150d506e3f05

* Depends on: https://gerrit.libreoffice.org/c/core/+/154680/
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -98,20 +98,6 @@ std::string formatUnoCommandInfo(const std::string& sessionId, const std::string
 
 }
 
-int sendURPToClient(void* pContext /* std::function* of sendBinaryFrame */,
-                    const signed char* pBuffer, int nLen)
-{
-    static const std::string header = "urp:\n";
-    size_t responseSize = header.size() + nLen;
-    char* response = new char[responseSize];
-    std::memcpy(response, header.data(), header.size());
-    std::memcpy(response + header.size(), pBuffer, nLen);
-    pushToMainThread([pContext, response, responseSize]() {
-        static_cast<ChildSession*>(pContext)->sendBinaryFrame(response, responseSize);
-    });
-    return 0;
-}
-
 ChildSession::ChildSession(const std::shared_ptr<ProtocolHandlerInterface>& protocol,
                            const std::string& id, const std::string& jailId,
                            const std::string& jailRoot, DocumentManagerInterface& docManager)
@@ -132,8 +118,7 @@ ChildSession::ChildSession(const std::shared_ptr<ProtocolHandlerInterface>& prot
         LOG_WRN("URP is enabled in the config: Starting a URP tunnel for this session ["
                 << getName() << "]");
 
-        m_hasURP = startURP(docManager.getLOKit(), this, &m_sendURPToLOContext, sendURPToClient,
-                            &m_sendURPToLO);
+        m_hasURP = startURP(docManager.getLOKit(), &m_URPContext);
 
         if (!m_hasURP)
             LOG_INF("Failed to start a URP bridge for this session [" << getName()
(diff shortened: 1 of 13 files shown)


kit/ChildSession.cpp
```diff
@@ -345,6 +350,17 @@ bool ChildSession::_handleInput(const char *buffer, int length)
 
         return success;
     }
+    else if (tokens.equals(0, "urp")) {
+        if (length < 4) return false;
```
eszkadev, line 354: Later when you will prepare final PR, please make sure formatting is ok. Here it uses different style than we use in c++ code.
if () {
}
vs
if ()
{
}

kit/ChildSession.cpp
```diff
@@ -106,37 +106,36 @@ int sendURPToClient(void* pContext /* std::function* of sendBinaryFrame */,
...
+    , _jailRoot(jailRoot)
+    , _docManager(&docManager)
+    , _viewId(-1)
+    , _isDocLoaded(false)
+    , _copyToClipboard(false)
+    , _canonicalViewId(-1)
+    , _isDumpingTiles(false)
+    , m_hasURP(false)
 {
     if (std::string(std::getenv("ENABLE_WEBSOCKET_URP")) == "true")
```
mmeeks, line 128: It is prolly better to get a StdlyCapsGlobal here setup on start-up of the kit; just how 'AnonymizationSalt' is setup - copy how that's done through the code - this global env. var is not really a per-instance member concept =)

net/Socket.hpp
```diff
@@ -1063,15 +1086,17 @@ class StreamSocket : public Socket,
         msg.msg_iov = &iov[0];
         msg.msg_iovlen = 1;
 
-        char adata[CMSG_SPACE(sizeof(int))];
```
mmeeks, line 1066: Fun =)

wsd/COOLWSD.hpp
```diff
@@ -174,6 +174,30 @@ class WSProcess
...
+        const char* buffer = message.data() + 4;
+        for (ssize_t i = 0; i < len;)
+        {
+            int written;
+            while ((written = ::write(urpFD, buffer + i, len - i)) < 0 && errno == EINTR)
+                LOG_TRC("EINTR writing to URP pipe: " << urpFD);
+            if (written < 0)
+            {
+                LOG_ERR("failed to write to URP pipe: " << urpFD << " : " << std::strerror(errno));
+                return false;
```
mmeeks, line 194: This seems odd - can't we use a Stream to write from the coolwsd side and shuffle the data across ?

net/Socket.cpp
```diff
@@ -333,26 +334,29 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
     setupPollFds(now, timeoutMaxMicroS);
     const size_t size = _pollSockets.size();
 
+    bool hasURP = _urpFD != -1;
```
mmeeks, line 337: I'm a tad surprised we can't use a normal StreamSocket for the URP  thing and add that to the SocketPoll in the normal way - did you try without this ? =)

mmeeks (changes requested):
This looks a lot more sensible and has waaay fewer scary Mutexen in it :-) so awesome!
Does it work ?
Ultimately I'd really like to touch the net/Socket code as little as possible, if possible - I'd hope that shouldn't be necessary. We have hard-coded wakeup socket, and then everything should be uniform: Unix Domain Sockets, IPv6 / IPv4 / listening server sockets / inotify pipes etc. - I don't think we want to change that.
But great to see it so nearly there ! =)

caolanm:
Oh yeah, it certainly works. I'll move stuff into some sort of StreamSocket based solution.

mmeeks (approved):
Looks really nice =) thanks Caolan !

caolanm:
I'll just stash here a standalone demo modified to use collabora-online and this feature, based on the example of https://bugs.documentfoundation.org/show_bug.cgi?id=152261 

[officepb-collabora-online.zip](https://github.com/CollaboraOnline/online/files/12935390/officepb-collabora-online.zip)

PR #6994, opened 2023-07-31
admin: use access token

hcvcastro:
- admin: add access token
- browser: admin: use access token

### DIFF
--- a/browser/admin/admintemplate.html
+++ b/browser/admin/admintemplate.html
@@ -61,6 +61,7 @@
 <body>
     <!--%BRANDING_JS%-->
     <script>
+        window.accessToken = '%ACCESS_TOKEN%';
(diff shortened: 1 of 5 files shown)


caolanm:
iOS app co-23.05 is a temp failure affecting a lot of merge requests from this time. Otherwise I can't really meaningfully review, but it looks sane at least.

browser/admin/admintemplate.html
```diff
@@ -61,6 +61,7 @@
 <body>
     <!--%BRANDING_JS%-->
     <script>
+        window.accessToken = '%ACCESS_TOKEN%';
```
Ashod, line 64: Wouldn't it be better to name this `jwtAccessToken = '%JWT_ACCESS_TOKEN%`, to avoid confusing with the wopi access token?

PR #7000, opened 2023-08-01
Improve a11y support for tabbed UI

mcecchetti:
Provided support for screen reader to report key accellerators
description.

When <alt> is pressed screen reader says something like: "tab control
Home tab selected".
When an accellerator is typed such as "N" screen reader says: "N
Insert tab selected".
When an accellerator for a tab item is typed the screen reader says
the typed letters and the item name as set in the 'alt' attrinute for
the button element.
For instance by typing accellerator for bold style ("1"), the screen
reader says: "1 Bold"

Change-Id: I64b198f79cb9f7f13dcd1a4fe0890e4317e512dd

### DIFF
--- a/browser/src/dom/NotebookbarAccessibility.js
+++ b/browser/src/dom/NotebookbarAccessibility.js
@@ -3,7 +3,7 @@
 	This class is used for managing the accessibility keys of notebookbar control.
 */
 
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
Thanks for the improvements.

mcecchetti:
Thanks for your review going to merge

PR #7004, opened 2023-08-02
browser: remove hyperlink pop-up

hcvcastro:
Unfortunately, we cannot show the hyperlink pop-up,
in client side based on cell formula, because the formula
had not been evaluated.

=HYPERLINK(A1,A1);

The server should fire event "hyperlinkclicked".

Change-Id: Ie2b743812493e1790b9e54f0050974c585e61e00


caolanm (approved):
looks reasonable to me

PR #7005, opened 2023-08-03
Add check for Insert_buttons to display in tablet view

Darshan-upadhyay1110:
Change-Id: I103461b02405bc03d63671b885a31735668d4111

* Resolves: #
* Target version: master 

### Summary
Added tablet check for Insert buttons to display on dom or not based on passed condition .
 - By default custom buttons will show up in tablet.

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -583,6 +583,9 @@ L.Control.UIManager = L.Control.extend({
 		}
 
 		this.customButtons.push(button);
+		if (button.tablet === false && window.mode.isTablet()) {
+			return;
+		}
 		this.insertCustomButton(button);
 	},
 


browser/src/control/Control.UIManager.js
```diff
@@ -581,9 +581,13 @@ L.Control.UIManager = L.Control.extend({
 			if (item.id === button.id)
 				return;
 		}
-
 		this.customButtons.push(button);
-		this.insertCustomButton(button);
+		if (button.tablet === undefined) {
+			button.tablet = true; // setting default value as true so in future we can use tablet prop of insert button
+		}
+		if (button.tablet || !window.mode.isTablet()) {
```
pedropintosilva, line 588: Is this a typo? if button.tablet is true or if isTablet is not true

Darshan-upadhyay1110: no it's not a typo. 
basicallly is button.tablet is set to true then we will insertCustom button 
And if button.tablet is set to false then we will check if current resolution is desktop then in this case we also need to saw the button 

only if button.tab = false and resolution is also for tablet then and then we should remove the insert button

pedropintosilva (approved):
Thanks much better

PR #7009, opened 2023-08-03
a11y: provide info about current table and cell to screen reader

mcecchetti:
When we get in one or more tables screen reader reports row and column
count.
When we get out one or more table screen report reports "out of table"
for each table.
When the fosused cell changes screen reader reports new row/col index.
Screen reader reports cell paragraph content, too.
Added also cypress tests for checking that the table/cell description
string is correct.

Change-Id: I078b29a2f76f91df479f75db76113c80405f7169


mcecchetti:
it needs https://gerrit.libreoffice.org/c/core/+/155282 to be merged for passing tests

eszkadev (approved):
Looks good

mcecchetti:
Looked at Cypress failures, locally `calc/annotation_spec.js` is passing. So I feel enough confident to merge this patch.

PR #6989, opened 2023-07-31
private/pedro/tablet show userlist hide docname

pedropintosilva:
- Tablet: Hide document file name input field
- Show userlist on tablets

PR #7013, opened 2023-08-03
Add rename document to notebookbar and menubar

NickWingate:
Change-Id: Iec25de88aa3e7089af102dd9f4bda6507308c215

Related: [#6989](https://github.com/CollaboraOnline/online/pull/6989)

Add button to notebookbar and menubar to rename document.

* Resolves: # 
* Target version: master 

### TODO

- [x] Test with integrator (nextcloud)
- [ ] Get icon
- [x] Test menubar

### Checklist

### DIFF
--- a/browser/src/control/Control.DocumentNameInput.js
+++ b/browser/src/control/Control.DocumentNameInput.js
@@ -29,6 +29,10 @@ L.Control.DocumentNameInput = L.Control.extend({
 						this.map.renameFile(value);
 					}
 				}
+				else {
+					// when user doesn't specify any extension
+					this.map.renameFile(value);
+				}
 			} else {
 				// saveAs for rename
 				this.map.saveAs(value);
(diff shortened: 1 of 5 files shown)


pedropintosilva:
@NickWingate the following is not produced by this PR but it's emphasise with it:

Right now ``documentNameConfirm`` function requires the extension to be passed. If the extension is not passed (user doesn't type it in the dialog or removes it in the input field (toolbar) before pressing enter : **Nothing happens**, nothing is renamed, no error is triggered. Could you please adjust that function?  If no extension is passed, we should just use the existent document's extension and rename it. Right now it never enters ``if (value.lastIndexOf('.') > 0) {``

NickWingate:
@pedropintosilva 
Done, if the user doesn't specify extension we use the `map.rename` function (`Control.DocumentNameInput.js:34`) which just takes filename without an extension and preserves the original extension

pedropintosilva (approved):

pedropintosilva:
waiting for another +1

eszkadev (approved):
code looks good

PR #7019, opened 2023-08-04
occasionally testTileSubscription fails

caolanm:
sometimes the deltas arrive in a different order, later in the same test we have "ordering is undefined tiles arrive in so swap if needed" for a pair of later deltas, so do the same reordering for this intermittently failing earlier test

Change-Id: If51edb9a4e22d469d9e029c38da7348c34b17832

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks @caolanm. Seems reasonable to allow for one-step out of order arrival.

PR #7026, opened 2023-08-05
Removed NBSP while coping info to clipboard

rparth07:
Change-Id: I2566a4902ed36e53ad2acfbe1c32efccb04bddda

* Resolves: # 
* Target version: master 

### Summary
This is follow up PR regarding #6549. This PR contains latest code that will remove NBSP and cypress test regarding copying version info to clipboard.

### TODO

- [ ] ...

### Checklist


Tex2002ans:
Hey @rparth07 . Do you think the root cause might actually be these 2 lines instead?

https://github.com/CollaboraOnline/online/blob/0700a5a314f770f35617299ccadba8f2a5483e60/browser/src/core/Socket.js#L629

and:

https://github.com/CollaboraOnline/online/blob/0700a5a314f770f35617299ccadba8f2a5483e60/browser/src/core/Socket.js#L653C1-L655C70

I think those 2 `&nbsp;` are the ones that introduce them inside the About dialog.

I think if we get rid of those and replace with a simple space, then it won't be copied in the first place. :)

rparth07:
> Hey @rparth07 . Do you think the root cause might actually be these 2 lines instead?
> 
> https://github.com/CollaboraOnline/online/blob/0700a5a314f770f35617299ccadba8f2a5483e60/browser/src/core/Socket.js#L629
> 
> and:
> 
> https://github.com/CollaboraOnline/online/blob/0700a5a314f770f35617299ccadba8f2a5483e60/browser/src/core/Socket.js#L653C1-L655C70
> 
> I think those 2 `&nbsp;` are the ones that introduce them inside the About dialog.
> 
> I think if we get rid of those and replace with a simple space, then it won't be copied in the first place. :)

Thanks @Tex2002ans , I understand the concern but I think `&nbsp;` was initially introduced to make sure that the space between two words stays consistent, ensuring the HTML renders correctly without unexpected line breaks, especially on smaller screens. While I haven't tested it extensively, I worry it might cause spacing issues on certain devices.

Moreover, I really think we should stick with replace method. It guarantees that our content copying works smoothly, regardless of whether `&nbsp;` is present or not. This way, we maintain a reliable and consistent experience.

Would love to hear your thoughts on this!

Tex2002ans:
> While I haven't tested it extensively, I worry it might cause spacing issues on certain devices.

Looks like `git blame` points to this back in 2021:

https://github.com/CollaboraOnline/online/commit/e44737af7266d13ff6b54dd8a4196414d4b81610

> "About dialog: **space after git hash: was being ignored**"

So I suspect the hash number might have been smushing into the string:

- `git hash:123456`

Perhaps that original spacing issue was squashed since then?

- - -

> [...] I think `&nbsp;` was initially introduced to make sure that the space between two words stays consistent, ensuring the HTML renders correctly without unexpected line breaks, especially on smaller screens.

Yeah, `&nbsp;` would typically be used to keep the 2 words together:

Original:

```
blah blah blah (git hash:
123456)
```

vs. nbsp:

```
blah blah blah (git
hash: 123456)
```

so "hash:" + "123456" would always stay stuck together, even on skinny cellphones.

> Moreover, I really think we should stick with replace method. It guarantees that our content copying works smoothly, regardless of whether `&nbsp;` is present or not.

No idea, but something you did back in August failed CI, and this PR has been staying in limbo ever since!

(It bothers me every single day when I press the dang "Copy" button! This has been haunting me for an entire year ever since we first fixed the previous issue! :P)

> Would love to hear your thoughts on this!

Well, something—anything—would be better! As long as we get rid of that dang non-breaking space in the copied text!

I pinged the devs about this PR, and they said:

> First, it needs to pass CI checks. Then it can be reviewed/merged.

rparth07:
> Well, something—anything—would be better! As long as we get rid of that dang non-breaking space in the copied text!

@Tex2002ans all CI checks are now passing. Additionally, with this change I'm not facing any non-breaking space issue. Everything appears to be working as expected. 

Can you please let me know if you are encountering any issues with these updates?

vmiklos:
Quoting <https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/908/consoleFull>:

> Test unit-crash FAILED. Finished in 15s.

retry.

vmiklos:
Quoting <https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/254/console>:

> ✖  writer/invalidations_spec.js

retry.

vmiklos (approved):
Looks good to me, thanks! :-)

PR #7028, opened 2023-08-07
ios: fix doubleclick in navi, with single clicks

aszucs3:
dblclick seems not to work on ios, but click does. We can simulate dblclick with single clicks.

Change-Id: Ic9ae58a39122025c9947294ebfd2292ffe04ae5c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -123,6 +123,8 @@ function _getCellIconId(cellData) {
 	return iconId;
 }
 
+var lastClickHelperRow = -1;
+var lastClickHelperId = '';
 function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTreeView, treeRoot) {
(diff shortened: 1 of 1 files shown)


eszkadev (commented):
I found information that `dblclick` event is supported on iOS:
https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event
https://caniuse.com/mdn-api_element_dblclick_event

eszkadev:
can you experiment with replacing jQUery's `$(text).dblclick(doubleClickFunction);`  with `addEventListener` ?

eszkadev:
on which device it didn't work? in the 'mobile app' only? or also 'in the browser'?

aszucs3:
> please instead add new click handler which will check if it is click after click case and directly will call: `doubleClickFunction();`

Thank you, i did it.

eszkadev (approved):

PR #7029, opened 2023-08-07
admin: imprv ui sizes of adminClusterOverview.html

pedropintosilva:
- Graph axis text is too small (10px) better to set it to 1em based on
parent div (16px). Good to also set any group within svg to 1em to
avoid any surprises
- Currently tablets from one card is spilling over to the next card
when there is not enough space. Better to allow scrolling.
- The left pane should be hidden even before reaching the mobile
size. This will fix the lack of space for cards and avoid everything
being so small

Change-Id: I58c88385875469b56542703663e570ab4d63fbdc

### DIFF
--- a/browser/admin/adminClusterOverview.html
+++ b/browser/admin/adminClusterOverview.html
@@ -5,6 +5,29 @@
     Admin.ClusterOverview(host);
 </script>
 
+<style>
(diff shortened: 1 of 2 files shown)


Rash419 (approved):
Thanks

PR #7035, opened 2023-08-09
Add option for print sheet

Darshan-upadhyay1110:
Change-Id: Ic4514648bd998575990bce877eda5d45884a8514

* Resolves: # 
* Target version: master 

### Summary
Added 2 options to print
    - Active sheet
    - All sheets

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -671,7 +671,10 @@ L.Control.Menubar = L.Control.extend({
 					{name: _('CSV file (.csv)'), id: 'downloadas-csv', type: 'action'}]},
 				{name: _UNO('.uno:SetDocumentProperties', 'spreadsheet'), uno: '.uno:SetDocumentProperties', id: 'properties'},
 				{type: 'separator'},
-				{name: L.Control.MenubarShortcuts.addShortcut(_UNO('.uno:Print', 'spreadsheet'), L.Control.MenubarShortcuts.shortcuts.PRINT), id: 'print', type: 'action'},
+				{name: L.Control.MenubarShortcuts.addShortcut(_UNO('.uno:Print', 'spreadsheet'), L.Control.MenubarShortcuts.shortcuts.PRINT), id: 'print', type: 'menu', menu: [
+					{name: _('Active sheet'), id: 'print-active-sheet', type: 'action'},
+					{name: _('All Sheets'), id: 'print-all-sheets', type: 'action'},
+				]},
 				{name: _('Close document'), id: 'closedocument', type: 'action'}
 			]},
 			{name: _UNO('.uno:EditMenu', 'spreadsheet'), id: 'editmenu', type: 'menu', menu: [
@@ -1857,7 +1860,9 @@ L.Control.Menubar = L.Control.extend({
 			|| id === 'deletepage'
 			|| id === 'remotelink'
 			|| id === 'toggledarktheme'
-			|| id === 'home-search') {
+			|| id === 'home-search'
+			|| id === 'print-active-sheet'
+			|| id == 'print-all-sheets') {
 			this._map.dispatch(id);
 		} else if (id === 'insertcomment') {
 			this._map.insertComment();
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -154,8 +154,8 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 			},
 			!this._map['wopi'].HidePrintOption ?
 			{
-				'id': 'print',
-				'type': 'bigtoolitem',
+				'id': 'Data-Print:Print',
+				'type': 'menubutton',
 				'text': _UNO('.uno:Print', 'spreadsheet'),
 				'command': '.uno:Print',
 				'accessibility': { focusBack: true,	combination: 'PT', de: null }
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -100,6 +100,9 @@ function onClick(e, id, item) {
 			map.toggleCommandState(getUNOCommand(item.uno));
 		}
 	}
+	else if (item.id === 'print-active-sheet' || item.id === 'print-all-sheets') {
+		map.dispatch(item.id);
+	}
 	else if (id === 'print') {
 		map.print();
 	}
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -246,12 +246,12 @@ L.Map.include({
 			'options=' + options);
 	},
 
-	print: function () {
+	print: function (options) {
 		if (window.ThisIsTheiOSApp || window.ThisIsTheAndroidApp) {
 			window.postMobileMessage('PRINT');
 		} else {
 			this.showBusy(_('Downloading...'), false);
-			this.downloadAs('print.pdf', 'pdf', null, 'print');
+			this.downloadAs('print.pdf', 'pdf', options, 'print');
 		}
 	},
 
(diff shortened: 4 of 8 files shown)


Darshan-upadhyay1110:
Not able to print only the current sheet data by Print active Sheet option. 

I tried by setting the PrintRange ( only to sheet 1 for testing purpose https://github.com/CollaboraOnline/online/pull/7035/files#diff-460c498103a44955159750887652ddfb90cbe3a2daf389c2cfb5332a824a9a24 ) but it has no effect in result.

we are getting all sheets data same as normal print.


Maybe need to introduce new UNO command in core and get PrintRange accordingly.

@eszkadev can you help here

browser/src/control/Control.NotebookbarCalc.js
pedropintosilva, line 1: Please watch the indentation. See the removed formatting

Darshan-upadhyay1110: oh i see i'll fix with final commit for this feature

Darshan-upadhyay1110:
Note for me: need to add print option in menubar of comact mode. options are added in toolbar but also need to provide option in menubar 

menubar : 
image


top-toolbar: 
image


Notebookar : Done :)

image

Darshan-upadhyay1110:
Query:

in current implementation i am facing one problem.
- if we have 4 sheets in calc and for an example sheet 1,3,4 are having some content in it but sheet 2 is empty.
- at next step if we try to print sheet 4 it gives error like this.
image
- emprty Sheet 2 :  image
- the reson here  is we use Export as for printing and because to print any current sheet we are passing this information ( `CurrentSheet` )
image
- Export as we only consider pages with content in it and igrone the pages which are empty
- so Sheet 2 will ignore and page number of all sheets after that will be change for expirt purpose.
- and so if we now passing `PageRange` as 4 ( sheet 4)  but in real only 3 pages presnt and though this error occured

@pedropintosilva do you know is there any options are available to also consider empty pages to print or just not to discard while printing other active sheet.

pedropintosilva:
So, it seems that on core that option (don't print empty sheets) is enabled by default

But I think we don't have additional option that an be set on online side https://sdk.collaboraonline.com/docs/postmessage_api.html?highlight=action_export used to then call downloadas:

```js
Map.WOPI.js
 			if (msg.Values) {
				var format = msg.Values.Format;
				var fileName = this._map['wopi'].BaseFileName;
				fileName = fileName.substr(0, fileName.lastIndexOf('.'));
				fileName = fileName === '' ? 'document' : fileName;
				this._map.downloadAs(fileName + '.' + format, format);
			}
```
So I wonder what we should do here. Any thoughts @eszkadev ?

eszkadev:
> So, it seems that on core that option (don't print empty sheets) is enabled by default
> 
> But I think we don't have additional option that an be set on online side https://sdk.collaboraonline.com/docs/postmessage_api.html?highlight=action_export used to then call downloadas:
> 
> ```js
> Map.WOPI.js
>  			if (msg.Values) {
> 				var format = msg.Values.Format;
> 				var fileName = this._map['wopi'].BaseFileName;
> 				fileName = fileName.substr(0, fileName.lastIndexOf('.'));
> 				fileName = fileName === '' ? 'document' : fileName;
> 				this._map.downloadAs(fileName + '.' + format, format);
> 			}
> ```
> 
> So I wonder what we should do here. Any thoughts @eszkadev ?

Above thing is unrelated, it's postmessage API while @Darshan-upadhyay1110  is working on online browser <-> online server API

We have to pass some new additional flag and apply it on the core side.

eszkadev:
> @eszkadev oh right, but then again there is no current option/parameter . We would need to add one and then handle it on core. Which leads to ask: couldn't we simply use our recent and already async download dialog ?:

I think there will be requests: do not show me that complex dialog I don't know what to do, I just want to print :)
And I think it is also not easy to attach it to the print workflow.

We are already passing some parameters to configure pdf generation, one more is not a big issue I think.

pedropintosilva:
ok, fair enough just wanted to be sure we didn't forget that dialog.

Darshan-upadhyay1110:
@pedropintosilva and @eszkadev
- there are some options to allow empty pages like  `IsSuppressEmptyPages`, `IsSkipEmptyPages` but that options are not beahving like they should be :(. 

-as we can see in code `PageRange` is working so this variables also should work.

- maybe we are using Exportas functinality for printing sheets and that is why some parts in core already descarding empty pages without considering this options 

 Any thoughts @pedropintosilva  , @eszkadev ?

Darshan-upadhyay1110:
@eszkadev i tried `selection` parameter option but i am confused here what shopuld be the value for that i tried with expression like this '`$Tabelle1.$A$1:$XFD$1048576` but not working. find some  example values like we passing in page range  ( 1-2 or 2-3)  but no luck

browser/src/control/Control.Toolbar.js
```diff
@@ -100,6 +100,30 @@ function onClick(e, id, item) {
 			map.toggleCommandState(getUNOCommand(item.uno));
 		}
 	}
+	else if (item.id === 'print-active-sheet') {
+		map.sendUnoCommand('.uno:.uno:DeletePrintArea');
```
eszkadev, line 104: duplicated code

browser/src/control/Control.Menubar.js
```diff
@@ -1848,6 +1851,28 @@ L.Control.Menubar = L.Control.extend({
 			this._map.dispatch('shareas');
 		} else if (id === 'print') {
 			this._map.print();
+		} else if (id === 'print-active-sheet') {
+			this._map.sendUnoCommand('.uno:.uno:DeletePrintArea');
```
eszkadev, line 1855: duplicated code

browser/src/control/Control.Toolbar.js
eszkadev: please use `this._map.dispatch(command)` and put action implementation in: browser/src/control/Toolbar.js  in `dispatch` method

browser/src/control/jsdialog/Widget.MenuButton.js
```diff
@@ -147,6 +149,32 @@ function _menubuttonControl (parentContainer, data, builder) {
 	return false;
 }
 
+function printSheets(item, builder) {
+	builder.map.sendUnoCommand('.uno:.uno:DeletePrintArea');
```
eszkadev, line 153: duplicated code

browser/src/control/Control.Toolbar.js
Darshan-upadhyay1110: okay got it

browser/src/control/jsdialog/Widget.MenuButton.js
```diff
@@ -114,6 +114,8 @@ function _menubuttonControl (parentContainer, data, builder) {
 				onSelect: function (event) {
 					if (event.item.uno)
 						builder.map.sendUnoCommand('.uno:' + event.item.uno);
+					else if (event.item.id === 'print-active-sheet' || event.item.id === 'print-all-sheets')
```
eszkadev, line 117: please rpepare generic solution, use `map.dispatch(event.item.xxx)`

eszkadev: not if (myspecial id, but detect something form JSON which will mean "use dispatch"

eszkadev: eg. `if (item.type == 'action') map.dispatch(item.id)`

browser/src/control/Toolbar.js
```diff
@@ -1073,6 +1073,29 @@ L.Map.include({
 		case 'renamedocument':
 			this.uiManager.renameDocument();
 			break;
+		case 'print-active-sheet':
+			this.sendUnoCommand('.uno:.uno:DeletePrintArea');
```
eszkadev, line 1077: dupliacted prefix: `.uno:.uno`

eszkadev (approved):
code looks good

PR #7040, opened 2023-08-09
Private/caolanm/coverity9

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/tools/Replay.hpp
+++ b/tools/Replay.hpp
@@ -423,19 +423,6 @@ public:
         if (optStats)
             optStats->addConnection();
     }
-
-    /// Attach to @server, load @filePath and replace @tracePath
-    static void replaySync(const std::string &server,
-                           const std::string &filePath,
-                           const std::string &tracePath)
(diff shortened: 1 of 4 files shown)


tools/Replay.hpp
```diff
@@ -409,8 +408,10 @@ class StressSocketHandler : public WebSocketHandler
 
     static void addPollFor(SocketPoll &poll, const std::string &server,
                            const std::string &filePath, const std::string &tracePath,
-                           const std::shared_ptr<Stats> &optStats = nullptr)
+                           const std::shared_ptr<Stats> &optStats)
```
vmiklos, line 411: Perhaps rename this optStats to just stats in a follow-up PR, given that it's no longer optional? Or even perhaps try to turn it into a reference if it's never nullptr.

vmiklos (approved):
Looks good in general, see inline comment for a nit.

PR #7050, opened 2023-08-10
fix: cursor jump when hyperlink popup clicked in mobileview

Rash419:
Change-Id: I4c5f1f2923f894156f5f05382b0f54476dea8a56

* Resolves: # 
* Target version: master


eszkadev (approved):
code looks good

PR #7064, opened 2023-08-11
cypress: fix svg text-decoration detection

eszkadev:
this was changed in the core:
commit 0a549b2c9595cd076e3b503b67df93bff380885a
svg: export text decoration for text run not paragraph


eszkadev:
Fixed tests passed

PR #7072, opened 2023-08-14
Termination Flag

Ashod:
- wsd: move statics into anonymous namespace
- wsd: simplify shutdown and termination flagging

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -41,8 +41,18 @@
 namespace
 {
 #ifndef IOS
-static std::atomic<bool> TerminationFlag(false);
-static std::atomic<bool> ShutdownRequestFlag(false);
+
+/// The valid states of the process.
+enum class RunState : char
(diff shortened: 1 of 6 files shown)


common/SigUtil.cpp
```diff
@@ -38,57 +38,63 @@
...
+/// The valid states of the process.
+enum class RunState : char
+{
+    Run = 0, //< Normal up-and-running state.
+    ShutDown, //< Request to shut down gracefully.
+    Terminate //< Immediate termination.
+};
+
+/// Single flag to control the current run state.
+static std::atomic<RunState> RunStateFlag(RunState::Run);
```
vmiklos, line 54: Note that you're inside an anonymous namespace here, so a 'static' is redundant. core.git CI even errors out in such cases.

vmiklos (approved):
Makes sense, see inline comment for a nit.

PR #7073, opened 2023-08-14
When document layer is not ready and socket message is not delayed, m…

gokaysatir:
…essage is lost.

Wait for doclayer to be ready in this case.

Change-Id: Ib3e9427a89e8be7d97f7baf92bdcb6295c695ca2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
fixes the issue

PR #7078, opened 2023-08-14
android: fix app crashes after closing the document

Rash419:
- It exits LOActivity when lokit thread finishes using lokit_main_mutex. But if we edit document and exit without saving lokit thread gets finished but COOLWSD is still running to clean up the docBrokers. It takes some time to clean up this docBrokers. We need to exit the LOActivity only after COOLWSD is finished
- for that I have added coolwsdRunningMutex again - it was removed in ad32888d7c2ec960255d826c245f7df2e93c3f1b

Change-Id: Ic9785f5aa55deafbac98efc4013bd3376d0ba62d


eszkadev:
please rebase, I fixed cypress tests some time ago

quikee (approved):
Looks fine

PR #7080, opened 2023-08-15
Don't show presentation on iOS in small frame

eszkadev:
Full screen feature is not available on iPhone,
only on iPad with iOS 12+ (https://caniuse.com/fullscreen).

Instead open svg directly in the browser as a fallback.


mcecchetti (approved):
So I checked the problem on iPad with iPadOS 16 and for me locally it works fine both with and without this patch.
Unfortunately I'm not able to perform any test on my old iPhone 6 (iOS 12.5.7): Online doesn't work at all: the page is blank (included UI). No idea when this issue started.
Anyway the code looks good to me.

PR #7086, opened 2023-08-15
jsdialog: add minimal size for treeview

eszkadev:
Without it in calc -> insert -> chart treeview
for chart type selection was invisible / 0px width


pedropintosilva:
WRT

pedropintosilva (approved):
Thanks, I have tested.

Issue #7094, opened 2023-08-16
Mobile: Missing dark mode toggle


pedropintosilva:
### Describe the Bug

We are missing dark mode toggle on mobile view. This becomes even more dramatic when the same user changed his preference (via desktop view) and now is not able to get back to light/dark mode when using the phone.

### Steps to Reproduce

1. Open Mobile View (either using web-browser dev tools or on smartphone)
2. Click on the hamburger menu
3. View
4. There is no Dark mode toggle (as we have on desktop view)

### Expected Behavior

Add menu item at the bottom with checkmark as we are already doing for other menu items (spelling):
![image](https://github.com/CollaboraOnline/online/assets/65948705/808969b9-c974-4768-ab5e-fd3e2404f607)

Note, the above screenshot is altered. Last item doesn't exist, that's what the bug is about : )

PR #7095, opened 2023-08-16
Add missing dark mode toggle in mobile view

Darshan-upadhyay1110:
Change-Id: I89100d703444e8bfe2a8960d00c62a1e1ea88a84

* Resolves: #7094 
* Target version: master 

### Summary
- Added missing dark mode icon  with checkmark.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -118,14 +118,8 @@ L.Control.UIManager = L.Control.extend({
 			};
 			app.socket.sendMessage('uno .uno:ChangeTheme ' + JSON.stringify(cmd));
 		}
-		if (this.getCurrentMode() === 'classic' || this.map.isReadOnlyMode()) {
-			this.refreshMenubar();
-			this.refreshToolbar();
-		}
-		else {
-			this.refreshNotebookbar();
-		}
-		this.refreshSidebar();
+		if (!window.mode.isMobile())
+			this.refreshAfterThemeChange();
 	},
 
 	initDarkModeFromSettings: function() {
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -441,21 +441,23 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	refreshNotebookbar: function() {
-		var selectedTab = $('.ui-tab.notebookbar[aria-selected="true"]').attr('id') || 'Home-tab-label';
-		this.removeNotebookbarUI();
-		this.createNotebookbarControl(this.map.getDocType());
-		if (this._map._permission === 'edit') {
-			$('.main-nav').removeClass('readonly');
+		if (!window.mode.isMobile()) {
```
pedropintosilva, line 444: thanks.
I was wondering, if we now need this then it means that when we toggle the theme we are calling this. Wouldn't it be better to just change the `toggleDarkMode` function? After all we already have darkmode related conditions there:

```javascript
	toggleDarkMode: function() {
		// get the initial mode
		var selectedMode = this.getDarkModeState();
		// swap them by invoking the appropriate load function and saving the state
```
[...]
```javascript
		[...]
		else {
			this.refreshNotebookbar();
		}
		this.refreshSidebar();
	},
```

Also this could then be used in the case we would need to (now or in the future) refresh any other mobile-only UI

Darshan-upadhyay1110: totally agreed. @pedropintosilva how about we seperate this refresh logic from `toggleDarkMode` function so if need refresh thing we can use in other paces as well. 

i have added this condition to `toggleDarkMode` and seprated the refresh logic in other function for more readability and also use cases of that function in future will be more helpful

pedropintosilva (commented):
Ah now seems more sensible, thanks! (I still didn't test it. )

pedropintosilva (approved):
Tested in:
* responsive view on web browser
* smartphone using integrator's web view

Rash419 (approved):
Looks good to me!

Rash419:
Tested with android but doesnot work as expected. it only changes dark mode from the toggle button only one time then toggling doesnot have any effect I will create a patch to fix this behaviour. Thanks

Rash419:
fix darkmode toggle on android: https://github.com/CollaboraOnline/online/pull/7108

PR #7110, opened 2023-08-18
clusteroverview: send routeToken with auth message

Rash419:
Change-Id: I6412d68ef0c86d7a618eaf0ad252c16dee565d77

* Target version: master


quikee (approved):
Looks OK to me

PR #7111, opened 2023-08-18
Switch tab using keyboard shortcut

Darshan-upadhyay1110:
Change-Id: I4e76a0cc382fbe97f16037c8f9d78174e8663491

* Target version: master 

### Summary

LO is offering this using `CTRL`+`PageUp` or 'PageDown`.
- but there are some browser restriction related to security purpose some of the event we can not prevent.
- for an example CTRL + PageUp will not be prevented using JS in browser ( as per browser security rules )
- so i have tried to add this shortcut using Ctrl + Alt
- we can cycle through all sheets.

Adding this small enhancement  as part of this `HackWeek` :)

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -717,6 +717,24 @@ L.Map.Keyboard = L.Handler.extend({
 				case this.keyCodes.D: // d
 					app.socket.sendMessage('uno .uno:InsertEndnote');
 					return true;
+				case this.keyCodes.pageUp:
+				case this.keyCodes.pageDown :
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
code looks good

pedropintosilva (approved):
Even if it was possible using CTRL + PageUp would be not good for accessibility (users make use of that to cycle through their web browser tabs)

Thanks, please do a follow up that adds this short cut info to the Help > Keyboard shortcuts ;)

Darshan-upadhyay1110:
Thanks pedro. i will add this information in keyborad shortcuts :)

PR #7011, opened 2023-08-03
Fix to issue #6977

shehr7:
Fixes the ui-treeview's limited occupied vertical space in writer. 
Change-Id: I8551242fc2b987f00c67088caa0ed07ee8526a72

* Resolves: #6977 
* Target version: master 

### Summary

As there are 3 hidden grids, changing their individual heights did not change anything so I forced change the height of ui-treeview.

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id
- [x] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [x] Documentation (manuals or wiki) has been updated or is not required

Issue #6977, opened 2023-07-27
ui-treeview doesn't extend the avaialble vertical space


pedropintosilva:
![image](https://github.com/CollaboraOnline/online/assets/65948705/76ed81f5-cea4-4949-b22d-4b010958b8da)

PR #7112, opened 2023-08-18
Fix 'manage changes' dialog expanding infinitely

Minion3665:
A regression in #7011 caused the manage changes dialog to continue expanding when new elements were added, no matter how big it got. This would cause you to be unable to see all the changes when there were a lot of them because you could not scroll through the list. This patch fixes the scrolling, without making the list area very small again.

Change-Id: Ib29fb161522773ea9893b7bc21d06aab19b74c52

* Target version: master 

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -14,11 +14,6 @@
 	display: flex;
 }
 
-.jsdialog #writerchanges .ui-treeview-body {
-	min-height: 300px;
-	max-height: none;
-}
-
 .jsdialog-overlay.cancellable {
 	-webkit-animation: none;
 	animation: none;
@@ -1570,4 +1565,10 @@ kbd,
 
 .warn-copy-paste {
 	width: 100%;
+}
+
+/* manage changes dialog */
+
+.jsdialog #writerchanges .ui-treeview-body {
+	min-height: 300px;
 }


browser/css/jsdialogs.css
```diff
@@ -14,9 +14,31 @@
 	display: flex;
 }
 
-.jsdialog #writerchanges .ui-treeview-body {
-	min-height: 300px;
-	max-height: none;
+#AcceptRejectChangesDialog .ui-grid {
```
eszkadev, line 17: could you move all these rules to the end of file
and before add separation comment with name of the dialog
that would make things grouped, here we have general rules for all dialogs

pedropintosilva (changes requested):
Thanks @Minion3665 

I just don't understand one thing. Why the max-height: none was set in the first place.. Meaning wouldn't be enough of a fix to remove `max-height: none;` and this inherit the general max-height rule coming from the ui-treeview-body?

Otherwise, and if we need to set all these in the css afterwards in a case by case then we should maybe improve the widget itself instead.

Minion3665:
> Thanks @Minion3665
> 
> I just don't understand one thing. Why the max-height: none was set in the first place.. Meaning wouldn't be enough of a fix to remove `max-height: none;` and this inherit the general max-height rule coming from the ui-treeview-body?
> 
> Otherwise, and if we need to set all these in the css afterwards in a case by case then we should maybe improve the widget itself instead.

@pedropintosilva So, you'd think so, but that was what was causing https://github.com/CollaboraOnline/online/issues/6977... in this case you wanted to have the treeview filling the space so we can't limit it to the normal max height...

Improving the element so it always fills the available space is an option, would you like me to look into doing that instead (possibly in a separate issue so we can get this merged for Richard Brock)?

pedropintosilva:
I would prefer to also fix this as initially I wished (via CSS grid) but I would like to steer away from adding it in a case by case basis.

This PR works better but if this will not stay in the long run wouldn't it be better to just remove the max-height for now?

```diff
 .jsdialog #writerchanges .ui-treeview-body {
        min-height: 300px;
-       max-height: none;
 }
```
| With This PR| Removing max-height|
| --- | --- |
| with-PR7112 | just-removing-max-height |

> So, you'd think so, but that was what was causing https://github.com/CollaboraOnline/online/issues/6977...

And I think we wouldn't see this problem again because we keep the min-height.

Thus I'm not against merging this but up to @eszkadev

Minion3665:
I've replaced this with the code as Pedro suggested, as well as moving it down so it conforms with the earlier review comment from Szymon... let's look into using something closer to the css that I had before for doing this to treeviews not on a case-by-case basis sometime in the future

eszkadev (approved):

PR #7113, opened 2023-08-18
jsdialog: handle vertical tabcontrol

eszkadev:
vertical tabcontrol has "vertical: true" property

for now just handle it as normal tab control,
it doesn't use tabpages but regular grids for
tab contents


aszucs3 (approved):
Looks good to me

Issue #7109, opened 2023-08-18
Rename document button works only on tabbed view


pedropintosilva:
### Describe the Bug

The newly added "Rename Document" triggers the rename dialog only when on tabbed view. It does nothing when on the compact view. IIRC this was working before.

### Steps to Reproduce

1. Switch to compact view
2. File menu
3. Rename Document
4. Nothing happens

Ezinnem:
Confimed in:
COOLWSD version:
23.05.2.2git hash: [c3aaae07](https://github.com/CollaboraOnline/online/commits/c3aaae07)(E)

PR #7114, opened 2023-08-18
Fix "renamedocument" not working in compact view

Darshan-upadhyay1110:
Change-Id: I2e9c322cc31e333d78e7b1576ed061a9091b8a54

* Resolves: #7109 
* Target version: master


pedropintosilva:
Thanks but I think we just forgot to dispatch it, better to relay on the id since are anyhow checking for that . Could you please review https://github.com/CollaboraOnline/online/pull/7116 ?

PR #7117, opened 2023-08-18
fix: empty mobile wizard when opened for the first time

Rash419:
- it was only affecting wizard with tabs

Change-Id: I6172a3a1a498982927e63676c4743be8f2ff9d80

* Target version: master


eszkadev (approved):
works :)

PR #7120, opened 2023-08-21
admin: add new command to verify auth token

Rash419:
Change-Id: I2e640a746c839ea6f4494ddbdfcb8f924150906f


quikee (approved):
Maybe we would need brute-force protection for such... allowing only a limited amount of queries per second. Otherwise looks fine.

PR #7121, opened 2023-08-21
Open user list box using keyboard ( CTRL + 'O')

Darshan-upadhyay1110:
Change-Id: I8ac5043382ffe49fe2644c76c6b77af85e1e5937

* Resolves: #
* Target version: master 

### Summary
Added keyboard shortcut to see userlist in document.

as part of my hackweek contributions :)


Darshan-upadhyay1110:
@pedropintosilva , @gokaysatir need your views on this. i am not sure about which keys is good for this but for now i have gone with => `CTRL + O`

Darshan-upadhyay1110:
@pedropintosilva i have doubt about using accessibility keys for this. if that feature is disabled then how user will directly open this using keyboard ?

pedropintosilva:
@Darshan-upadhyay1110 the Alt pressing  (hints shows up) is always enabled. What you are thinking is the document related accessibility part, but that's different

pedropintosilva:
Actually on a second though. I would like to have both:
* The direct shortcut so, user can very quickly open it even if on compact mode
* And for the accessKey we might need to rethink as we don't have that enabled outside of tabs

@Darshan-upadhyay1110 please make sure that ctrl+O is not used for anything else on any app (writer, calc, impress, draw). And check how other do.

Darshan-upadhyay1110:
@pedropintosilva

-  i have done some research to open userlist. in other platforms, There is no keyboard shortcut to open the user icons in the top right corner of Microsoft Docs, whether in the web or desktop version. The only way to open them is to click on them.
 
 -  also `Ctrl + o` is used to open a dilog box to open new files from system in other platforms ( MS online, GDocs etc....)
 
 -  in Current condition i have checked that we do not have ctrl+O used for anything else on any app (writer, calc, impress, draw)
 
 -  So we can use that :)

pedropintosilva (changes requested):
So I would avoid to use shortcuts that are popular for other things and that might trigger a certain response in some web browser.

Good to avoid Ctrl + O (known for open file); Ctrl + P (known for printing) etc

I think it is safer to have 2 modifier instead:
* CTRL + ALT + U , as in users
* CTRL + ALT + P , as in participants or as in people

Maybe I prefer P because might be less confusing and more integrator with what integrators use in their lexico (Participants)

Darshan-upadhyay1110:
@pedropintosilva  i think need to go with `Shift` and nor `alt` because when i am testing somtimes accesskeys got activated and it open the `Layout` tab insted of UserList

gokaysatir (approved):

PR #7128, opened 2023-08-22
Android dependencies update

Rash419:
* Target version: master


Rash419:
I could not udpate gradle to latest version due to this bug: https://issuetracker.google.com/issues/293294923
for some reason it could not compile resources of `ca-valencia` locale

timar (approved):

PR #7131, opened 2023-08-22
jsdialog: combobox with custom entries (font previews)

eszkadev:
To work requires chain of core changes:
https://gerrit.libreoffice.org/c/core/+/155853


eszkadev:
tests require core changes to pass (only font picking from combobox fails)

PR #7132, opened 2023-08-22
Notebookbar: introduce simple search button

### LINKED ISSUE #7098: Status bar: Increase discoverability
### Is your feature request related to a problem?

I have seen people asking around:
* Where is the status bar? How to enable it
* Where is some feature they used to access with one click at the bottom (status bar)

#### Current situation
* Status bar can be shown or hidden by default from integrators
* Some integrators (.e.g.: Nextcloud) have chosen to hide it for a variety of reasons
   * This has been experienced as a life improvement to some users (less clutter, more focus on content, more space to be used to read/edit)
   * This has been experienced as an additional barrier to other users that are either more experienced or make use of things that need to be always visible (language, word & character counting, selected rows zoom controls, Average/Sum)

##### Things we have done to improve this
* Once the user toggles `View` > `Status bar` that element will be visible from then on
* Almost all indicators and actions from status bar were also added to the tabbed view. This way, user doesn't depend on status bar. However there are still a couple of things that are Status bar only.

### Proposal / Discussion

Not only some indicators are difficult to extract from status bar but, I believe, that when the user needs them they actually make the most sense discretely at the bottom where they are. Furthermore, and since it's likely that user starts off wanting more space but ends up middle-editing wanting status bar – **Show always the Status bar. But allow it to be collapsed** 

- It would have only the zoom controls available at the right
- Plus an expand button on the left
- When the window is too small, and to avoid the collapsed status bar to be above the document, we could show less buttons there
- Expanding that element would need to change the canvas height (similar as it happens when we toggle status bar currently)

image

pedropintosilva:
This way we increase status bar discoverability ( fixes #7098 ) and
group both search functionality in one place in the home tab

Change-Id: Ia1fa8fe58903088aad37a560edd033a184d35e0e

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2250,6 +2250,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		}
 
 		var iconURLAliases = {
+			'home-search': 'recsearch',
 			'addmb-menu': 'ok',
 			'closetablet': 'view',
 			'defineprintarea': 'menuprintranges',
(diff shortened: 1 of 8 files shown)


gokaysatir (approved):
Thanks:)

PR #7133, opened 2023-08-22
Private/caolanm/coverity14

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -85,7 +85,7 @@ ClientSession::ClientSession(
                                    << "], current number of connections: " << curConnections);
 
     // populate with random values.
-    for (auto it : _clipboardKeys)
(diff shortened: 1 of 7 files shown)


wsd/ClientSession.cpp
```diff
@@ -85,7 +85,7 @@ ClientSession::ClientSession(
                                    << "], current number of connections: " << curConnections);
 
     // populate with random values.
-    for (auto it : _clipboardKeys)
+    for (size_t i = 0; i < N_ELEMENTS(_clipboardKeys); ++i)
```
vmiklos, line 88: I assume that here `const auto& it` would not be enough, would introduce some unused variable warning or so.

vmiklos (approved):

wsd/ClientSession.cpp
caolanm: yeah, exactly, and everything else range-based is just a miserable hackaround

PR #7145, opened 2023-08-23
Stop coolwsd when /etc/coolwsd/coolwsd.xml changes

Minion3665:
- This allows you to, say, have a docker container and --restart=always
  to restart when you update the config
- This patch only listens for "/etc/coolwsd/coolwsd.xml", so if you
  specify a config file that isn't there then you're out of luck... An
  improvement for a followup patch will be make it listen to wherever
  your config files actually are
- This patch doesn't work when editing /etc/coolwsd/coolwsd.xml with vim
  as vim moves the file before writing a new file in the right place.
  Again, this may be a fix for a followup patch. The current docker
  scripts only listen for modifications, so this matches that behavior

- This patch adds an option to enable this feature, it is off by default

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -162,6 +162,12 @@ using Poco::Net::PartHandler;
 #endif
 #endif
 
+#ifdef __linux__
+#if !MOBILEAPP
+#include <sys/inotify.h>
+#endif
+#endif
+
 using namespace COOLProtocol;
 
 using Poco::DirectoryIterator;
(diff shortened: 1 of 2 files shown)


mmeeks (changes requested):
Looks interesting. I'm not sure we need a new TerminatingPoll for this - I'd expect to add the inotify socket into:
while (!SigUtil::getShutdownRequestFlag()) the         mainWait.poll(waitMicroS);

which doesn't have much else to do - we don't need a new thread.

I'd expect the socket to be added to the poll though:

    void insertNewSocket(std::shared_ptr<Socket> newSocket)

With a sub-class of Socket that overrides the getPollEvents - mostly read I guess, and handlePoll: handling the incoming inotify event - which would get triggered when the mainWait.poll() wakes up and gives control there.

It'd repay you to read the strace of coolwsd too I think =)

Minion3665:
@mmeeks re 88476a80ddc6de3ff1cd6711aad40d6741167db4
> Oooh - looks much nicer :-) please do squash down

done

> There were some other comments around replicating how the previous inotify approach worked - to ensure we match and handle the vim fixme in the same way (?) - do we watch the parent folder or ? different flags? - worth reading the code there and matching what we had.

Looking at the code, this is a match to what we had-- and indeed when backupcopy is set to yes (`:set backupcopy=yes`, the default is auto which may sometimes do it... I'm not quite sure what the conditions are there) and the file is edited in the docker container the same issue occurs...

mmeeks (approved):
Looks fine; lets get this in - and follow up with getting the paths right, and also watching coolkitconfig.xcu too.

Thanks.

wsd/COOLWSD.cpp
```diff
@@ -1001,6 +1007,111 @@ class PrisonPoll : public TerminatingPoll
...
+    {
+        if (getFD() == -1)
+        {
+            LOG_WRN("Inotify - Failed to start a watcher for the configuration, disabling "
+                    "stop_on_config_change");
+            m_stopOnConfigChange = false;
+            return;
+        }
+
+        watch("/etc/coolwsd/coolwsd.xml");
```
timar, line 1027: COOLWSD_CONFIGDIR is the path of coolwsd.xml. 
What about watching COOLWSD_CONFIGDIR itself? That way we could cover coolkitconfig.xcu as well, and at the same time we would not have to worry about strange behaviour of vim. Would there be any side-effects?

mmeeks: Agreed - would prefer watching the parent directory and then matching filenames with changed size / attributes etc.

timar (approved):

wsd/COOLWSD.cpp
```diff
@@ -1001,6 +1007,109 @@ class PrisonPoll : public TerminatingPoll
...
+        }
+        else if (len == -1)
+        {
+            LOG_TRC("InotifyPoll - Got to end of data when reading inotify");
+        }
+
+        if (len <= 0)
+            break;
+
+        for (char* ptr = buf; ptr < buf + len; ptr += sizeof(struct inotify_event) + event->len)
```
caolanm, line 1100: inotify_event is one of those structs which uses a "struct hack" for a variable length trailing member.

struct inotify_event {
...
uint32_t len;      /* Size of name field */
char     name[];   /* Optional null-terminated name */
};

so sizeof(struct inotify_event) doesn't actually skip over a full struct, you would need to factor in whatever "len" claims to do that.

Either way, I don't think you can be sure you even have a complete entry even for the the first iteration to be sure "name" is null terminated if the record happens to be too long to fit in that buf buffer?

caolanm: I've read man inotify now and withdraw my comment :-)

PR #7146, opened 2023-08-23
Fix regression in tab cycle in dialogs

eszkadev:
this is regression from commit:
f8e53bd851a267fd50f369a1beb3c398d46f620a
Assign tab ids while defining tabs.

When opened Format -> Page Style dialog in Writer. There was "123" label next to the tabs.
Also when switching between widgets using tab key
it was moving focus outside window.

Core is sending id as plain number what is easy to duplicate. Let's add some unique name.


gokaysatir (approved):
Thank you for the catch @eszkadev :)

Issue #7062, opened 2023-08-11
Impress: Can't use keyboard keys (arrow, enter...) in the Insert Hyperlink popup


maks-w:
### Describe the Bug

Can't use keyboard keys (arrow, enter...) in the Insert Hyperlink popup for a Presentation document. 
![image](https://github.com/CollaboraOnline/online/assets/1763638/8954da57-95e7-4e2a-b91e-f2e5f40a831f)

The focus switch to the slide panel
![image](https://github.com/CollaboraOnline/online/assets/1763638/29f70735-0896-43be-b04b-f7861c48a9b2)

### Steps to Reproduce

1. Open Impress
2. Click on Insert > Hyperlink
3. Write something in one of the inputs
4. Use arrow keys or delete key

### Expected Behavior

Moving the keyboard cursor in the input box

### Actual Behavior

Move slide focus if you use arrow keys or delete it if you use the delete key.

### Desktop

 - **OS:** Debian GNU/Linux 11 (bullseye)
 - **Browser:** Brave
 - **Version:** Version 1.56.20 Chromium: 115.0.5790.171 (Build officiel) (64 bits)
 
 ### Collabora version
 
- COOLWSD version: 23.05.2.2
- git hash: [912dfcd0](javascript:void(window.open('https://github.com/CollaboraOnline/online/commits/912dfcd0'));)
- LOKit version: Collabora Office 23.05.2.2
- git hash: [ec17398](javascript:void(window.open('https://hub.libreoffice.org/git-core/ec17398'));)
- Served by: Debian GNU/Linux 11 (bullseye)

Ezinnem:
I can confirm this issue.

@eszkadev, please would you like to take a look?

thebearon:
There was a similar issue with comments, which was a regression from this commit:
https://github.com/CollaboraOnline/online/commit/13387b8df6a35db78f4088bc0d6b945913791bc6
```
Author:     Gökay Şatır <gokaysatir@collabora.com>
AuthorDate: Mon Jan 23 12:36:14 2023 +0300
Commit:     Andras Timar <andras.timar@collabora.com>
CommitDate: Tue Jan 31 09:40:54 2023 +0100

    Enable deleting page / part with keyboard after a preview is clicked on.
```

This change was to enable cursor / delete / backspace key in the slides pane.

The fix in 22.05 was the following PR: #6306 (this particular issue didn't need to be fixed in 23.05, because the comment view changed)

Steps for this bug were:
[...]

PR #7151, opened 2023-08-24
Make jsdialogs capture keydown events in Impress

CosmicOli:
* Resolves: #7062
* Target version: master 

### Summary
JSdialog popups on Impress would have the keydown events not captured. The events would be handled by the main document.
Now they will be captured by the correct element.

Change-Id: I543e25ddbcb1766cc677f7fbc6629e8d04a5b394

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks reasonable :)

PR #7153, opened 2023-08-24
feature-locking: send lock/restriction status from client session

lpranam:
problem:
earlier restriction/lock status was sent to kit via browser, now client session send this status directily to the kit. this design will require less communication between server and browser

Change-Id: I6b830f30fb326a5e6637e345250893cbba101de6

* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
@Ashod could you review?

Ashod (approved):
Thanks @lpranam. Looks reasonable. I would use clang-format to get a consistent formatting, though.

Issue #7156, opened 2023-08-24
Add unix socket support to coolwsd


FedericoHeichou:
### Is your feature request related to a problem?

I want to use coolwsd with unix socket.

### Describe the solution you'd like

Add a optional "Unix" like IPv4 and IPv6 listeners.

### Describe alternatives you've considered

N/A

### Additional context

N/A

PR #7158, opened 2023-08-24
Add unix socket support

FedericoHeichou:
Change-Id: I451dbee460a8a6907e438cd15f8084b8badcb40a

* Resolves: #7156 
* Target version: master 

### Summary
With this commit you can listen at a unix socket with coolwsd

### TODO

- [ ] ...

### Checklist


net/Socket.cpp
```diff
@@ -824,8 +825,33 @@ bool ServerSocket::bind(Type type, int port)
...
-    if (_type == Socket::Type::IPv4)
+    if (_type == Socket::Type::Unix)
+    {
+        struct sockaddr_un addrunix;
+
+        _socketPath = std::string(Socket::SocketPath);
+        int last_errno = 0;
+        std::memset(&addrunix, 0, sizeof(addrunix));
+        addrunix.sun_family = AF_UNIX;
+        std::memcpy(addrunix.sun_path, _socketPath.c_str(), _socketPath.length());
```
mmeeks, line 836: The paranoid in me wants to check that length against sizeof sun_path and fail elegantly if too long =)

net/Socket.cpp
```diff
@@ -877,8 +903,6 @@ std::shared_ptr<Socket> ServerSocket::accept()
     // Accept a connection (if any) and set it to non-blocking.
     // There still need the client's address to filter request from POST(call from REST) here.
 #if !MOBILEAPP
-    assert(_type != Socket::Type::Unix);
-
```
mmeeks, line 881: How does this work for client addresses ? I assume that sockaddr_in6 is big enough for whatever a unix socket whacks in clientInfo but is it ? =)

net/Socket.cpp
```diff
@@ -894,22 +918,29 @@ std::shared_ptr<Socket> ServerSocket::accept()
...
 #if !MOBILEAPP
-            char addrstr[INET6_ADDRSTRLEN];
-
-            const void *inAddr;
-            if (clientInfo.sin6_family == AF_INET)
+            if (clientInfo.sin6_family == AF_UNIX)
             {
-                auto ipv4 = (struct sockaddr_in *)&clientInfo;
-                inAddr = &(ipv4->sin_addr);
+                _socket->setClientAddress("/");
```
mmeeks, line 923: I wonder if we could have something more useful for the filtering & access control logic here? which is IIRC string based; could we use SO_PEERCRED to get a uid at least out if it?

net/Socket.cpp
```diff
@@ -50,6 +50,7 @@ constexpr std::chrono::microseconds WebSocketHandler::PingFrequencyMicroS;
 
 std::atomic<bool> SocketPoll::InhibitThreadChecks(false);
 std::atomic<bool> Socket::InhibitThreadChecks(false);
+std::string Socket::SocketPath("/tmp/coolwsd.sock");
```
mmeeks, line 53: Do we want this in a global variable really? surely we can store it on a UnixSocket on an instance or somesuch?  We already have randomized Unix sockets for the Prisoner connection:

    std::shared_ptr<ServerSocket> findPrisonerServerPort()

Could we clean this up for both uses somehow and have a UnixServerSocket sub-class that handles this with a virtual destructor in ServerSocket ? =)

mmeeks (changes requested):
This looks like a nice feature - though I'm intrigued by the use-case :-) can you expand on that. Nevertheless - lets get it merged with a few cleanups (as outlined) if possible. Grab me on libera.chat (mmeeks) if you need interactive help.

Thanks for contributing !

FedericoHeichou:
> This looks like a nice feature - though I'm intrigued by the use-case :-) can you expand on that. Nevertheless - lets get it merged with a few cleanups (as outlined) if possible. Grab me on libera.chat (mmeeks) if you need interactive help.
> 
> Thanks for contributing !

I have a container in network mode host (not bridge) but I don't want to expose coolwsd in the machine's network. So I patched in this way to expose a socket in a path inside the container

mmeeks:
Ah - nice :-) sounds interesting - would love to have some documentation fragments we can merge on setting that up too, or perhaps a forum post on your setup as/when it's tweaked and merged :-)
I assume you're ok with doing those changes ?

net/Socket.cpp
FedericoHeichou: Did you wrongly quoted the code? With this assert the code will stop

net/Socket.cpp
FedericoHeichou: I set the client address to "/" because a function already exists and handles the Unix socket in this way

https://github.com/CollaboraOnline/online/blob/52e3d4cb7ca39008fd272b6e9d26d74640dbcc33/net/Socket.cpp#L998-L1008

net/Socket.cpp
FedericoHeichou: I did this because it was the fastest solution and I don't have so much time (I am not a C++ developer but after some tries I can write some nice code). I think someone more skilled could code this in few minutes/hours but I would probably loose days

FedericoHeichou:
> Ah - nice :-) sounds interesting - would love to have some documentation fragments we can merge on setting that up too, or perhaps a forum post on your setup as/when it's tweaked and merged :-) I assume you're ok with doing those changes ?

Sure, I can write some examples (tell me where and I'll do).  
For the changes I replied/liked in each one, I'll starting work there next week

Darshan-upadhyay1110:
hii @FedericoHeichou can you please rebase and solve the conflicts ?

Thanks, Darshan :)

Darshan-upadhyay1110:
Closing this PR as it hasn’t been updated for a long time. We can always revisit it in the future or refer back to the original reported GitHub issue if needed.
 https://github.com/CollaboraOnline/online/issues/7156

PR #7160, opened 2023-08-24
Add cli option to use settings from env variables

Minion3665:
Change-Id: I75762ad620132037523fa82167a3ff17075c7027

* Target version: master 

### Summary
Currently [in docker it is possible to do configuration through
environment variables](https://col.la/dockercodeconfigviaenv), which
works using the start-collabora-online.sh start-collabora-online.pl
scripts. This PR lets COOLWSD listen to the same environment
variables directly

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/man/coolwsd.1
+++ b/man/coolwsd.1
@@ -30,5 +30,7 @@ coolwsd OPTIONS
 .PP
 \fB\-\-nocaps\fR                       Use a non\-privileged forkit, with increase in security problems.
 .PP
+\fB\-\-use\-env\-vars\fR                Override coolwsd.xml config with options from the environment variables described in https://col.la/dockercodeconfigviaenv
+.PP
 .SH "SEE ALSO"
 coolforkit(1), coolconvert(1), coolconfig(1), coolwsd-systemplate-setup(1), coolmount(1)
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -895,6 +895,7 @@ bool COOLWSD::NoSeccomp = false;
 bool COOLWSD::AdminEnabled = true;
 bool COOLWSD::UnattendedRun = false;
 bool COOLWSD::SignalParent = false;
+bool COOLWSD::UseEnvVarOptions = false;
 std::string COOLWSD::RouteToken;
 #if ENABLE_DEBUG
 bool COOLWSD::SingleKit = false;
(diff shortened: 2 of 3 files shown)


timur-g:
Can you please explain how this will work?

Minion3665:
@timur-g 
> Can you please explain how this will work?

Sure thing! The idea is to be able to replace the work done by the docker scripts (see https://github.com/CollaboraOnline/online/blob/master/docker/from-packages/scripts/start-collabora-online.sh) directly in wsd... 
```sh
if test -n "${username}"; then
    perl -pi -e "s/<username (.*)>.*<\/username>/<username \1>${username}<\/username>/" /etc/coolwsd/coolwsd.xml
fi
if test -n "${password}"; then
    perl -pi -e "s/<password (.*)>.*<\/password>/<password \1>${password}<\/password>/" /etc/coolwsd/coolwsd.xml
fi
if test -n "${server_name}"; then
    perl -pi -e "s/<server_name (.*)>.*<\/server_name>/<server_name \1>${server_name}<\/server_name>/" /etc/coolwsd/coolwsd.xml
fi
if test -n "${dictionaries}"; then
    perl -pi -e "s/<allowed_languages (.*)>.*<\/allowed_languages>/<allowed_languages \1>${dictionaries:-de_DE en_GB en_US es_ES fr_FR it nl pt_BR pt_PT ru}<\/allowed_languages>/" /etc/coolwsd/coolwsd.xml
fi
```

This will let us remove those scripts entirely from docker, but we still want to keep the ability to add settings from environment variables because it makes configuring docker easier.

The idea here is to add an extra option (`--use-env-vars`) which will get COOLWSD to look at the same environment variables that the scripts did directly. Does that answer the question?

timur-g:
Thanks, just one more. Will you update documentation?

Minion3665:
@timur-g 
> Thanks, just one more. Will you update documentation?

thanks for the reminder! I've updated the manpage to add the new option now. I'd like to edit https://sdk.collaboraonline.com too but I don't know where the source to that is... do you?

timur-g:
No, but I guess @thebearon knows.

timar (approved):

timar (changes requested):
* remoteconfigurl is missing
* domain is missing (although it was deprecated by alias_groupX so probably we can skip it)

timar:
> * remoteconfigurl is missing
> * domain is missing (although it was deprecated by alias_groupX so probably we can skip it)

I added remoteconfigurl.

timar (approved):

PR #7161, opened 2023-08-24
reduce cost of TileDesc::parse

caolanm:
perf reported 6.58% of time in collaborative multi user test on 2023-08-24 was spent in TileDesc::parse and much of that in std::unordered_map.

There are only 12 arguments in the map we care about here so we can just used a sorted array, look by name on write, and read by index.

Change-Id: Iadfc2001e298d8f4d46200c8488f0eb4cd8734c2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks nice =) I imagine there is another problem from higher-up though; I expect that when we get a queue of tiles we repeatedly parse the descriptions for no really good reason - but this looks nice.

PR #7162, opened 2023-08-24
Private/caolanm/better abbrev

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
The original thing here was to not split a utf-8 sequence when abbreviating a message so we didn't end  up with invalid utf-8. But then there was inconsistencies between the (const char*, len) and (const std::string&) which were annoying.

vmiklos (approved):
Looks sensible to me, thanks.

Issue #7164, opened 2023-08-25
DocumentBroker::sendRequestedTiles !newTileCombined.hasDuplicates() assert


caolanm:
### Describe the Bug
When a second user joins that can cause an assert of !newTileCombined.hasDuplicates(),. 

### Steps to Reproduce

1. I triggered this originally using "Open multiple urls" to open hello-world.odt in 20 tabs locally

### Actual Behavior

What I think is happening here is that the client is *sent* two canonicalidchange messages, the first message sends 1001 as viewid and the second message sends 1002 as viewid. But the first message appears before this._map._docLayer and so it put into a timer to happen later, then the second message appears but it is applied immediately it arrives, then the timer fires and the outdated obsolete canonicalidchange gets applied out of sequence which goes on to have unwanted sideeffects.

caolanm:
I think this problem has become common since around

commit a003126a652474906bb8a2ecc67d52ec4f39fab5
    Give new sockets a canonical view id in tests

and
commit d0c5321aaf8bfd3cbda1adb7830b569e971aba5c
    Fix getting canonical view id of 1000

while the delaying thing is also relatively new since

commit e660bb8e66e60cbff64fff1eb6eb39968797ee32
Date:   Mon Aug 14 14:21:05 2023 +0300

When document layer is not ready and socket message is not delayed, message is lost

caolanm:
@gokaysatir Putting a breakpoint into browser/src/layer/tile/CanvasTileLayer.js at
this._canonicalViewId = canonicalId;
I frequently see that (in a 2nd browser window) that _canonicalViewId gets set to 1002 first and then later 1001 from the delayed message so we end up with an obsolete _canonicalViewId

caolanm:
The code added in e660bb8e66e60cbff64fff1eb6eb39968797ee32 is very much like the existing code of _tryToDelayMessage so maybe what case is desired to be caught by the new thing should be added to _tryToDelayMessage instead?

mmeeks:
I suspect this could also be the cause of a new regression - whereby I lost a load of tiles on document load (I guess if there was a race) - it was easy to loose the top row of tiles in a document - until the doc was touched. And/or it was easy to loose the 2nd half of it at times =) a recent regression I've never seen before - on share as of now - share is c9e7591 currently which includes that patch.

caolanm:
[...]

PR #7170, opened 2023-08-25
cool#7164 use existing delay queue for 'canonicalidchange'

caolanm:
https://github.com/CollaboraOnline/online/issues/7164

might be best to add everything to this queue, and not drop any incoming message, and then process the queue when both this._map._docLayer and _isReady become available/true.

But perhaps this serves as a reasonable safe checkpoint to return to if it becomes necessary.

Change-Id: Ib77a4126437689709e2ca0915667b5b9b2f2f207

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1187,20 +1187,6 @@ app.definitions.Socket = L.Class.extend({
 		if (this._map._docLayer && !msgDelayed) {
 			this._map._docLayer._onMessage(textMsg, e.image);
 		}
-		else if (!this._map._docLayer && !msgDelayed) {
-			// If message is delayed and document layer is not ready at the time, message gets lost.
-			// To prevent this, we wait a little for document layer.
-			var waitForDocLayer = function(message, image) {
-				setTimeout(function() {
-					if (this._map._docLayer)
-						this._map._docLayer._onMessage(message, image);
-					else
-						waitForDocLayer(message, image);
-				}.bind(this), 300);
-			}.bind(this);
-
-			waitForDocLayer(textMsg, e.image);
-		}
 	},
 
 	_exportAsCallback: function(command) {
@@ -1316,6 +1302,7 @@ app.definitions.Socket = L.Class.extend({
 	_tryToDelayMessage: function(textMsg) {
 		var delayed = false;
 		if (textMsg.startsWith('window:') ||
+			textMsg.startsWith('canonicalidchange:') ||
 			textMsg.startsWith('celladdress:') ||
 			textMsg.startsWith('cellviewcursor:') ||
 			textMsg.startsWith('statechanged:') ||


mmeeks:
Hmm - not thrilled by the change even if it somehow works.
We should properly understand what was intended by the delaying here - do an analysis of the effect of that based on reading the protocol log from the initial startup there, work out what (if anything) really needs to get through, and do some allow-list instead of block-list I think - for the very few things we can tolerate re-ordering.

Of course, as a hack to get rid of the bug this is ok - as long as we stop and fix this properly. There are really several problems here, and Paris might have some ideas: if we need the canonicalviewid it is not to send it back to the core (that should never happen, and if it does it shouldn't be trusted) - it is perhaps useful for annotating into the tile-cache, or clearing that in the case that it changes I guess: but then it should be in-order with the other tile: and delta: messages coming in (I guess) rather than out of order somehow (!?).

Anyhow - we should pause, properly understand the problem, what this delaying concept is for and why it is there, and then "do it right" (TM) - I would not assume that code is correct in its current state.

caolanm:
Tag @OFFTKP to put in loop here.

A) so if I go back to commit 0ed983d7a3cca81c08c4078c20330002d68d5e70 and have a look at what used to happen, then...

1693480389538 INCOMING: canonicalidchange: viewid=0 canonicalid=1000
^^^ at this point this._map._docLayer is undefined so _onMessage is not called for canonicalidchange
...
1693480389642 OUTGOING: tilecombine nviewid=0 ...
...
1693481710656 INCOMING: canonicalidchange: viewid=0 canonicalid=1001 
^^^ at this point this._map._docLayer is defined so _onMessage is called for canonicalidchange, but it doesn't really matter because back then the handler for "canonicalidchange" didn't do anything with its payload except stash it in
this._debugData. The client had hardcoded nviewid=0 for tilecombine etc.
...
1693481779059 OUTGOING: tilecombine nviewid=0 ...
...
1693480389746 INCOMING: tile: nviewid=1001 ...

B) after commit d0c5321aaf8bfd3cbda1adb7830b569e971aba5c which drops the first "too early" invalidateCanonicalId  mission in Kit.cpp, and also sets this._canonicalViewId from canonicalidchange message. And adds early returns at various places if this. _canonicalViewId is unset, we then have...

1693482057296 INCOMING: canonicalidchange: viewid=0 canonicalid=1001 
^^^ at this point this._map._docLayer is defined so _onMessage is called for canonicalidchange and the handler is changed, so it sets this._canonicalViewId and that is used for later tilecombine etc
...
1693482160018 OUTGOING: tilecombine nviewid=1001 ...

c) after commit a003126a652474906bb8a2ecc67d52ec4f39fab5 first "too early" invalidateCanonicalId emission in Kit.cpp is restored so we get:

1693482580242 INCOMING: canonicalidchange: viewid=0 canonicalid=1001 
^^^ at this point this._map._docLayer is undefined so _onMessage is not called for canonicalidchange, browser _canonicalViewId remains at -1
...
<presumably problematic window of time where client will not request tiles because _canonicalViewId is -1>
...
1693482580383 INCOMING: canonicalidchange: viewid=0 canonicalid=1002 
^^^ at this point this._map._docLayer is defined so _onMessage is called for canonicalidchange and the handler is changed, so it sets this._canonicalViewId and that is used for later tilecombine etc
...
1693482671037 OUTGOING: tilecombine nviewid=1002 

d) after commit e660bb8e66e60cbff64fff1eb6eb39968797ee32

1693483549489 INCOMING: canonicalidchange: viewid=0 canonicalid=1001 
^^^ at this point this._map._docLayer is undefined so _onMessage is not called for canonicalidchange, browser _canonicalViewId remains at -1, but is gets queued to get played back later
...
1693483549717 INCOMING: canonicalidchange: viewid=0 canonicalid=1002 
...
^^^ at this point this._map._docLayer is defined so _onMessage is called for
canonicalidchange, so it sets this._canonicalViewId to 1002
...
1693483658825 OUTGOING: tilecombine nviewid=1002 ...
...
delayed canonicalidchange: viewid=0 canonicalid=1001 get processed, this._canonicalViewId set to 1001
...
1693483713231 OUTGOING: tilecombine nviewid=1001 ...
...
1693483713236 INCOMING: tile: nviewid=1002...

I'll merge this commit now which at least ensures the canonicalidchange messages get applied in the order they are received and I don't the obsolete viewid in messages triggering the newTileCombined.hasDuplicates() assert

PR #7182, opened 2023-08-29
a11y: don't focus map if focusBack is false

eszkadev:
This fixes regression introduced in commit 2ac7183e44e16a2a6956665d2d461c0873d62d95
Add Left or right arrow on a tab label should switch tabs

When we are in writer: Alt + H, K - opens spacing dropdown
Before this patch focus was stolen by the map even if the spacing item has focusBack property set to false


Darshan-upadhyay1110 (approved):
Tetsed and working fine. thanks @eszkadev :)

image

PR #7188, opened 2023-08-30
wsd: exclude readonly sessions from modification indication

Ashod:
Read only sessions cannot modify the document, so
we shouldn't consider input from such a session
as potentially modifying the document. This
becomes important during an abrupt disconnection
where we look at isPossiblyModified() as an
indicator for data-loss.

New unit-test that reproduces the corner-case
with disconnection. With graceful unloading
a previous unit-test verified that it worked
as expected.

Change-Id: Id84cda4f4599c559018247c32ea1205e154e4984

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -799,7 +799,8 @@ bool ClientSession::_handleInput(const char *buffer, int length)
             }
             else
             {
-                docBroker->updateLastModifyingActivityTime();
+                if (!isReadOnly())
+                    docBroker->updateLastModifyingActivityTime();
                 return forwardToChild(std::string(buffer, length), docBroker);
             }
         }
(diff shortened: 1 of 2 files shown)


wsd/ClientSession.cpp
```diff
@@ -799,7 +799,8 @@ bool ClientSession::_handleInput(const char *buffer, int length)
             }
             else
             {
-                docBroker->updateLastModifyingActivityTime();
+                if (!isReadOnly())
+                    docBroker->updateLastModifyingActivityTime();
```
eszkadev, line 803: "moveselectedclientparts: Moves the selected parts to a new position."

should be that possible in readonly mode?

eszkadev (approved):
Looks ok

wsd/ClientSession.cpp
Ashod: I expect that would modify the document and there not allowed.

timar:
@Ashod you did not think about the PDF case, apparently. PDF is read-only but comments can be added and saved.

PR #7190, opened 2023-08-30
cid#318945 Unchecked return value from library

caolanm:
Change-Id: I37cb5ddd7a1ce2dd1357738130f62950e76bbc6a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
@Ashod are we ok to continue here if chmod fails, is logging sufficient or should we return an error and refuse to continue?

timar (approved):

PR #7192, opened 2023-08-31
fix: comment annotation table not aligned

Rash419:
- regression from 8379ae65dd713c30ac729ae10a2d78cfc98e9a6c

Change-Id: I28edf882ce9efd586730a48a95f3bde555a63690

* Resolves: # 
* Target version: master


Darshan-upadhyay1110 (approved):
Thanks for the fix :)

Issue #6960, opened 2023-07-24
Arrow keys on notebook bar tab content should move around


Minion3665:
**Describe the solution you'd like**
As a followup to #6933 it might be nice to have the arrow keys move around inside the notebook bar, so that instead of using the tab key you could use up-down-left-right to navigate between different items

**Describe alternatives you've considered**
You could use tab and shift-tab to navigate, which is already supported

**Additional context**
[Screencast from 24-07-23 09 43 08.webm](https://github.com/CollaboraOnline/online/assets/34243578/f5172651-af23-47c8-b53a-3d7841947d07)

Ezinnem:
@Darshan-upadhyay1110 @pedropintosilva Please can you take a look?

PR #7194, opened 2023-08-31
Arrow keys traverse focus around notebookbar

CosmicOli:
* Resolves: #6960
* Target version: master 

### Summary
When focus is in the notebookbar, arrow keys allow focus to be moved.

Change-Id: Iccc9122f77c224e0190f25732ff9e4a24e4eab3e

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1220,10 +1220,114 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		}
 
 		function addKeydownEvents(element) {
+			var handleKeyDown = function(sourceElement, direction, currentTarget) {
+				var findFocusableParent = function(currentElement) {
+					if (currentElement.tagName === 'NAV') {
+						return currentElement;
+					}
+					else if (currentElement.tabIndex === -1 || currentElement.tagName == 'A') {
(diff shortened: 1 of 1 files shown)


gokaysatir (changes requested):
This is a really beautiful solution :) A fresh idea that uses ray casting. Thank you for the PR.

* I think 1px of difference is a little to small. For example it takes 25 tries to find the next one. Maybe we can make it configurable, just to be safe. Then we can set it to at least 5 pixels - to be changed if causes some unwanted selections / behaviours.
* Also there are JS errors in the console - some null things etc.
* And if user presses down arrow when the bottom control is focused, map gets the focus. This may or may not be related to your changes but it'd be good to ask @pedropintosilva if we want it to behave like this.

pedropintosilva:
yes I was aware of the initial report https://github.com/CollaboraOnline/online/issues/6960 and I agree

pedropintosilva:
> And if user presses down arrow when the bottom control is focused, map gets the focus. This may or may not be related to your changes but it'd be good to ask @pedropintosilva if we want it to behave like this.

will check

gokaysatir:
@CosmicOli thank you very much for the improvements :) I sent a small PR on top of this, in order to make this aligned with the latest changes.

gokaysatir (approved):

PR #7195, opened 2023-08-31
if we can't call _onMessage yet, delay all messages until we can

caolanm:
and not just some of them discarding the others, so we always apply all messages received in order the arrived without omissions.

TODO: apply them when the conditions application requires become true rather than depending on a timer poll.

Change-Id: I265b6ccb45e211c15c5f4daf9d6572fa051b68eb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
What seems unknown is why we used to delay some, and discard others?

eszkadev (approved):

PR #7199, opened 2023-09-01
mobile: fix images not visible with branding

Rash419:
Change-Id: I7a3d430ddd96ccb3ea0605eb0a0f2523d66225ce

* Target version: master


eszkadev:
something happened and this is no longer needed on master...

eszkadev (approved):

PR #7200, opened 2023-09-01
Writer: Set parent / child relationship in UI.

gokaysatir:
Change-Id: I7476aa13ad5d77cbc888339fb0ccd2d485b8f274

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -218,7 +218,7 @@ export class CommentSection extends CanvasSectionObject {
 	}
 
 	private createCommentStructureWriter (menuStructure: any, threadOnly: any): void {
-		var rootComment, lastChild, comment;
+		var rootComment, comment;
 		var commentList = this.sectionProperties.commentList;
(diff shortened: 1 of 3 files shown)


browser/src/layer/tile/CommentListSection.ts
```diff
@@ -1681,6 +1698,63 @@ export class CommentSection extends CanvasSectionObject {
...
+		}.bind(this));
+	}
+
+	private orderTextComments() {
+		var newOrder = [];
+
+		for (var i = 0; i < this.sectionProperties.commentList.length; i++) {
+			var comment = this.sectionProperties.commentList[i];
+
+			if (comment.sectionProperties.data.parent === '0') {
```
eszkadev, line 1747: this is repeated many times (check if parent is '0'), this has some special meaning I guess.
Would be good to create separate function for that which will name that check - will be easier to understand the code.

like `isRootComment(comment)` - or what does that mean...

eszkadev (approved):
I confirm, comments at the same position are separated now.

PR #7204, opened 2023-09-04
Detect code duplicates in make-check run

eszkadev:
It uses jscpd project by invoking it using npm script "duplication". For now set minimum lines to report as 26, because we have some 25 line duplicates.
Later we can reduce that number.


eszkadev:
works, https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/7269/consoleFull
contains:
```Checking code duplication...
Code duplication tests finished successfully.```

eszkadev:
better to use annotations in files, I will try to look how many of them we need

gokaysatir (approved):
If we want to run this command on CI machine, i guess we'll need to add the package into the archive?

PR #7212, opened 2023-09-06
Avoid logging messages when anonymization is on

cosminh11:
Change-Id: I5bcf615521d13758b2b70a0c9466a3bbc1a0a918

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Protocol.cpp
+++ b/common/Protocol.cpp
@@ -19,8 +19,6 @@
 
 namespace COOLProtocol
 {
-    bool anonymizationEnabled = false;
-
     std::tuple<int, int, std::string> ParseVersion(const std::string& version)
     {
         int major = -1;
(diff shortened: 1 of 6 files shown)


caolanm (approved):
Alright, lets merge that into my one and take it from there onwards.

I wonder if we need to hit all those places, or if only one or some of them needs it. Any idea if it's the case they all need to be protected? Or just being sure?

cosminh11:
> I wonder if we need to hit all those places, or if only one or some of them needs it. Any idea if it's the case they all need to be protected? Or just being sure?

I am sure about the one with the client session in DocumentBroker, because we had a bug report about it.
I included the other 2 because they seemed similar. Can't say for sure if we need them.
Also, there are other places where we log abbreviated messages, like the one below, but I don't know what kind of content they would have so I didn't touch them:
https://github.com/CollaboraOnline/online/blob/8ba14ab5e2eb449af0e6469cd64aaa5a7549281c/wsd/TileCache.cpp#L243-L244

PR #7217, opened 2023-09-07
Kill Poco Logger (Partially)

Ashod:
These patches replace the Console Channel of Poco with our faster and simpler implementations. The new loggers are behind EnableExperimental, so this change is safe to go in.

- wsd: test: enable experimental features in tests
- wsd: set up signal handlers early
- killpoco: our implementation of ConsoleChannel
- wsd: log: retry on write failure and flush
- wsd: support EnableExperimental in Log
- killpoco: own implementation of BufferedConsoleChannel
- killpoco: own implementation of ColorConsoleChannel
- wsd: stop the rendering thread pool before exiting
- wsd: log: BufferedConsoleChannel with thread-local buffer
- wsd: log: verify that all threads are stopped
- killpoco: replace LocalDateTime

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -29,6 +29,7 @@
 #include <Poco/PatternFormatter.h>
 #include <Poco/SplitterChannel.h>
 
+#include "Common.hpp"
 #include "Log.hpp"
 #include "Util.hpp"
 
@@ -364,7 +365,12 @@ namespace Log
             channel->setProperty("warningColor", "magenta");
         }
         else
-            channel = static_cast<Poco::Channel*>(new Log::ConsoleChannel());
+        {
+            if (EnableExperimental)
+                channel = static_cast<Poco::Channel*>(new Log::ConsoleChannel());
+            else
+                channel = static_cast<Poco::Channel*>(new Poco::ConsoleChannel());
+        }
 
         /**
          * Open the channel explicitly, instead of waiting for first log message
@@ -394,7 +400,7 @@ namespace Log
         struct tm tm;
         LOG_INF("Initializing " << name << ". Local time: "
                                 << std::put_time(localtime_r(&t, &tm), "%a %F %T %z")
-                                << ". Log level is [" << logger->getLevel() << "].");
+                                << ". Log level is [" << logger->getLevel() << ']');
     }
 
     Poco::Logger& logger()
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -35,6 +35,13 @@
 #include "Log.hpp"
 #include "Util.hpp"
 
+namespace
+{
+/// Tracks the number of thread-local buffers (for debugging purposes).
+std::atomic_int32_t ThreadLocalBufferCount(0);
+
+} // namespace
+
 namespace Log
 {
     using namespace Poco;
@@ -119,9 +126,14 @@ namespace Log
                 : _size(0)
                 , _oldest_time_us(0)
             {
+                ++ThreadLocalBufferCount;
             }
 
-            ~ThreadLocalBuffer() { flush(); }
+            ~ThreadLocalBuffer()
+            {
+                flush();
+                --ThreadLocalBufferCount;
+            }
 
             std::size_t size() const { return _size; }
             std::size_t available() const { return BufferSize - _size; }
@@ -595,6 +607,9 @@ namespace Log
     void shutdown()
     {
 #if !MOBILEAPP
+        assert(ThreadLocalBufferCount <= 1 &&
+               "Unstopped threads may have unflushed buffered log entries");
+
         IsShutdown = true;
 
         Poco::Logger::shutdown();
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -42,10 +42,28 @@ namespace Log
         static constexpr std::size_t BufferSize = 64 * 1024;
 
     public:
+        void close() override { flush(); }
+
         /// Write the given buffer to stderr directly.
-        static inline bool writeRaw(const char* data, std::size_t size)
+        static inline int writeRaw(const char* data, std::size_t size)
         {
-            return ::write(STDERR_FILENO, data, size) == static_cast<ssize_t>(size);
+            std::size_t i = 0;
+            for (; i < size;)
+            {
+                int wrote;
+                while ((wrote = ::write(STDERR_FILENO, data + i, size - i)) < 0 && errno == EINTR)
+                {
+                }
+
+                if (wrote < 0)
+                {
+                    return i;
+                }
+
+                i += wrote;
+            }
+
+            return i;
         }
 
         template <std::size_t N> inline void writeRaw(const char (&data)[N])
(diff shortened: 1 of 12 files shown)


common/SigUtil.cpp
```diff
@@ -150,6 +150,15 @@ void requestShutdown()
     /// Close the signalLog file.
     void signalLogClose()
     {
+        // We cannot shutdown the logging subsystem
+        // because freeing memory is not signal safe.
+
+        // Flush the IO buffers.
+        std::flush(std::cout);
```
mmeeks, line 157: Hmm - surely std::cout is yet another layer of C++ buffered-ness on top of what we have, and will no doubt allocate memory, and is not on the list of signal safe methods for good reason. Why are we using them? we should write to stdout or stderr - using safe, low-level system calls - not the FILE* abstraction, and not the C++ wrappers :-) cf. https://stackoverflow.com/questions/21460620/is-fflush-safe-to-be-called-from-a-signal-handler

kit/Kit.cpp
```diff
@@ -825,6 +825,7 @@ class Document final : public DocumentManagerInterface
             {
                 LOG_FTL("Document [" << anonymizeUrl(_url) << "] has no more views, exiting bluntly.");
                 flushTraceEventRecordings();
+                _pngPool.stop();
```
mmeeks, line 828: Hokay - all of the call sites you touch look like a real opportunity for code sharing & cleanup here. I expect we need to flushTraceEventRecordings() on an EX_OK and not otherwise, we need to stop the pngPool and do this Util::forcedExit etc.   - having said that I expect the Kit to be hard killed mostly by coolwsd: are we changing that ?  it is a -far- faster to go down than trying to touch all the memory we used and was paged out to fiddle with it and so on. We could of course send a USR2 signal or something to get it to flush its logs and hard exit or somesuch but ... hmm :-)

common/Log.cpp
```diff
@@ -439,9 +500,10 @@ namespace Log
         }
         else if (withColor)
         {
-            channel = static_cast<Poco::Channel*>(new Poco::ColorConsoleChannel());
-            channel->setProperty("traceColor", "green");
-            channel->setProperty("warningColor", "magenta");
+            if (EnableExperimental)
```
mmeeks, line 554: Love the conditionality here :-)

common/Log.cpp
```diff
@@ -104,6 +104,85 @@ namespace Log
...
+    {
+    protected:
+        std::size_t size() const { return _size; }
+        std::size_t available() const { return BufferSize - _size; }
+
+        inline void flush()
+        {
+            if (_size)
+            {
+                write(_buffer, _size);
```
mmeeks: Short writes ? =) EINTR ?

common/Log.cpp
```diff
@@ -42,10 +42,28 @@ namespace Log
...
         /// Write the given buffer to stderr directly.
-        static inline bool write(const char* data, std::size_t size)
+        static inline int write(const char* data, std::size_t size)
         {
-            return ::write(STDERR_FILENO, data, size) == static_cast<ssize_t>(size);
+            std::size_t i = 0;
+            for (; i < size;)
+            {
+                int wrote;
+                while ((wrote = ::write(STDERR_FILENO, data + i, size - i)) < 0 && errno == EINTR)
```
mmeeks, line 65: Ah - perhaps this is the 'write' we use that handles the problems  ? =)

wsd/COOLWSD.cpp
```diff
@@ -6204,7 +6199,22 @@ void forwardSigUsr2()
 // Avoid this in the Util::isFuzzing() case because libfuzzer defines its own main().
 #if !MOBILEAPP && !LIBFUZZER
 
-POCO_SERVER_MAIN(COOLWSD)
```
mmeeks, line 6207: Love the removal =)

mmeeks (commented):
Overall some really good stuff here :-) A few things to check and/or cleanup - near to approval but for the branch being locked down. I love the faster logging re: time fields, we need to re-think the use of FILE* or std::cout streams for signal safety and use STDOUT_FILENO, and/or STDERR_FILENO which are defined with these values in <unistd.h> I think =) [ also really no point in two layers of buffered writing which can only loose us data I think ].

Thanks ! =)

common/Log.cpp
Ashod: Yes, internal. You found the syscall with EINTR retry.

kit/Kit.cpp
Ashod: We do stop/exit the Kits from WSD. The thread stopping is to make sure the thread_local buffer is destroyed and flushed. It's not really necessary, except that with the check I added to track the number of thread_local buffers, it's cleaner to stop the threads (at least in the graceful case).

A misbehaving Kit will get 'kill -9' from WSD. The above is for the 'nice' case.

common/SigUtil.cpp
Ashod: Logging does _not_ use std::cout/cerr. This is just flushing. Technically, since we do flush stdout/stderr, and since cout and cerr are really mapped to them, this shouldn't be needed. I just put them for completeness. I'll remove them.

common/Log.cpp
```diff
@@ -119,9 +128,18 @@ namespace Log
                 : _size(0)
                 , _oldest_time_us(0)
             {
+#if ENABLE_DEBUG
```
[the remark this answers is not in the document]
Ashod: I added this #if because without the assert there is no value in doing atomic operations. It's to minimize performance noise (at runtime), rather than to silence the compiler.

Ashod:
Thanks @mmeeks for the quick review!

> also really no point in two layers of buffered writing which can only loose us data I think

We have but one buffer. The two class split is to manage the lifetime of the thread_local buffer independently of the Channel object that is created once per process. But we certainly don't need multiple buffers.

Also, removed flushing cout/cerr. Agreed that was an oversight to do from signal handlers.

common/Log.cpp
mmeeks: An old C programmer would appreciate a different name to the std C write function ;-)

common/SigUtil.cpp
mmeeks: Great; please do cut those out. We shouldn't write to stdout/stderr either but the relevant file-descriptions.

common/Log.cpp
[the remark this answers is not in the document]
Ashod: Removed.

common/SigUtil.cpp
Ashod: Removed.

kit/Kit.cpp
Ashod: Refactored.

common/Log.cpp
Ashod: Renamed.

Ashod:
@mmeeks, addressed all concerns. Hope it looks better now. Thanks!

mmeeks (approved):
Thanks for fixing all the comments Ash =) would love to get this in as/when there is a good time.

PR #7218, opened 2023-09-07
Calc: Implement JavaScript animation on mouse event

codewithvk:
Calc: Implement JavaScript animation on mouse events for enhanced user interaction

Change-Id: Ib3ed491c8e233013b2f547ed083c4b3d5b9f1a73

* Resolves: #5473 
* Target version: master

### DIFF
--- a/browser/src/layer/tile/ScrollSection.ts
+++ b/browser/src/layer/tile/ScrollSection.ts
@@ -539,14 +539,38 @@ export class ScrollSection extends CanvasSectionObject {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
Works as expected +1 
- Please change the commit subject(title) : remove "Calc:" since it can be misleading (it sounds that it only applies to Calc0
- Needs another +1

gokaysatir:
Why didn't we use "public startAnimating" functionality inside CanvasSectionContainer.ts file?

codewithvk:
This is probably an easy fix, so I encourage newcomers to take it on.  

Created an easy hack: #10313

PR #7220, opened 2023-09-07
getTokenString is another loop over the tokens that we already loop over

caolanm:
getTokenString(const StringVector& tokens...) loops over each token and calls
getTokenString(const std::string&...) with token.getParam() on each token. We already have a loop over each token, so we can merge those loops here.

Though looking further, broadcast seems unused by anything now.

To be followed up on that topic.

Change-Id: I244637043c58d90562001ec58fac494da54e55a8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me - nice catch Caolan - thanks :-)

PR #7221, opened 2023-09-07
Use dialog from core in Calc Insert Hyperlink

eszkadev:
introduce it in Calc only as it has better knowledge about cell content and what should be edited


eszkadev:
hyperlinkdialog

How new dialog looks now. Other possible improvements will be in the next PR.

aszucs3 (approved):
Looks good to me.
Thank you!

PR #7222, opened 2023-09-07
browser: notify user for unsupported video

lpranam:
detect if video is being played by browser using fps, if fps is near zero that means video is not able to play

Change-Id: I22bfc836a98ec0eb2caa58b89332b2531d69d18f

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/vector/SVGGroup.js
+++ b/browser/src/layer/vector/SVGGroup.js
@@ -3,6 +3,8 @@
  * L.SVGGroup
  */
 
+/* global _ */
+
 L.SVGGroup = L.Layer.extend({
 
 	options: {
@@ -78,6 +80,46 @@ L.SVGGroup = L.Layer.extend({
 		var point = this._map.latLngToLayerPoint(this._bounds.getNorthWest());
 		svgLastChild.setAttribute('x', point.x);
 		svgLastChild.setAttribute('y', point.y);
+
+		var videos = svgLastChild.getElementsByTagName('video');
+		this.addVideoSupportHandlers(videos);
+	},
+
+	addVideoSupportHandlers: function(videos) {
+		if (!videos)
+			return;
+
+		var that = this;
+
+		// slide show may have more than one video and it does not require any selection
+		for (var i = 0; i < videos.length; i++) {
+			var video = videos[i];
+			var sources = video.getElementsByTagName('source');
+
+			video.addEventListener('playing', function() {
+				window.setTimeout(function() {
+					if (video.webkitDecodedFrameCount === 0) {
+						that.showUnsupportedVideoWarning();
+					}
+				}, 1000);
+			});
+
+			video.addEventListener('error', function() {
+				that.showUnsupportedVideoWarning();
+			});
+
+			if (sources.length) {
+				sources[0].addEventListener('error', function() {
+					that.showUnsupportedVideoWarning();
+				});
+			}
+		}
+
+	},
+
+	showUnsupportedVideoWarning: function() {
+		var videoWarning = _('Document contains unsupported video');
+		L.Map.THIS.uiManager.showSnackbar(videoWarning);
 	},
 
 	addEmbeddedSVG: function (svgString) {


browser/src/layer/vector/SVGGroup.js
```diff
@@ -78,6 +80,62 @@ L.SVGGroup = L.Layer.extend({
...
+			this.video_.paused || this.video_.ended)
+		  return;
+		var currentTime = window.performance.now();
+		var deltaTime = (currentTime - this.startTime_) / 1000;
+		this.startTime_ = currentTime;
+
+		// Calculate decoded frames per sec.
+		var fps = (this.video_.webkitDecodedFrameCount - this.decodedFrames_) /
+				   deltaTime;
+		this.decodedFrames_ = this.video_.webkitDecodedFrameCount;
```
Ashod, line 133: Isn't it enough to check that `webkitDecodedFrameCount` is positive? Why do we care about the frame rate at all. If after a reasonable time (500ms or 1s) we don't have any frames, there probably is no video being played back.

This should work for videos, but I don't expect it to work for audio-only. Perhaps that's ok (but we shouldn't complain that an audio isn't supported when it is playing loud and clear, so best to ignore audio-only media).

lpranam: good point, indeed we don't need fps :) I'll take care of it

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -115,6 +115,22 @@ L.Map.SlideShow = L.Handler.extend({
 		var separator = (this._slideURL.indexOf('?') === -1) ? '?' : '&';
 		this._slideShow.src = this._slideURL + separator + 'StartSlideNumber=' + this._startSlideNumber;
 		this._slideShow.contentWindow.focus();
+
+		var that = this;
+		this._slideShow.addEventListener('load', function() {
+			var videos = that._slideShow.contentWindow.document.getElementsByTagName('video');
+			var videoWarning = _('Document contains uncompatible video');
```
Ashod, line 122: 'uncompatible' -> 'incompatible' (or perhaps better 'unsupported' ?)

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -115,6 +115,22 @@ L.Map.SlideShow = L.Handler.extend({
...
+			for (var i = 0; i < videos.length; i++) {
+				if (videos[i].networkState === 3) {
```
Ashod, line 124: A comment as to what 3 means is welcome.

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -115,6 +115,22 @@ L.Map.SlideShow = L.Handler.extend({
 		var separator = (this._slideURL.indexOf('?') === -1) ? '?' : '&';
 		this._slideShow.src = this._slideURL + separator + 'StartSlideNumber=' + this._startSlideNumber;
 		this._slideShow.contentWindow.focus();
+
+		var that = this;
+		this._slideShow.addEventListener('load', function() {
```
Ashod, line 120: Why aren't we calling `checkVideoSupport` from here?

browser/src/layer/vector/SVGGroup.js
```diff
@@ -78,6 +80,31 @@ L.SVGGroup = L.Layer.extend({
...
 		svgLastChild.setAttribute('y', point.y);
+
+		var video = svgLastChild.getElementsByTagName('video')[0];
+		var source = svgLastChild.getElementsByTagName('source')[0];
+		this.checkVideoSupport(video, source);
+	},
+
+	checkVideoSupport: function(video,source) {
+		var that = this;
+		var videoWarning = _('Document contains uncompatible video');
```
Ashod, line 91: Shouldn't we have this in one place only? It's duplicated here and in Map.SlideShow.js.

Ashod (commented):
Thanks @lpranam. Looks promising. Left some comments.

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -115,6 +115,12 @@ L.Map.SlideShow = L.Handler.extend({
...
+			var videos = that._slideShow.contentWindow.document.getElementsByTagName('video');
+			L.SVGGroup.prototype.checkVideoSupport(undefined, undefined, videos);
```
Ashod, line 122: Sorry, I didn't mean `checkVideoSupport` should handle the `videos` case.

I mean that for `videos` we do exactly the same thing as for `source`, so we should call the `checkVideoSupport` for each video entry.

We want to reduce code duplication, not put everything in one function.

Since the `source` tag is a child of the `video` tag, we should be able to have `checkVideoSupport` that takes a `video` node and add the listener for the `video`, query the `source` node, add the listener there too, if it exists.

_Another_ function can then be used to iterate over all videos and call `checkVideoSupport` on each `video` node.

(By the way, 'check' in 'checkVideoSupport' implies the function returns the results of the check. This is misleading, since we don't do that. Instead, we should call it 'addVideoSupportHandlers' or something more accurate.)

Ashod (commented):
Thanks @lpranam. Looks much better. Left one suggestion regarding code duplication. Best to have it in a cleaner state before we merge.

browser/src/layer/vector/SVGGroup.js
```diff
@@ -78,6 +80,48 @@ L.SVGGroup = L.Layer.extend({
...
+	addVideoSupportHandlers: function(videos) {
+		if (!videos)
+			return;
+
+		var that = this;
+
+		// slide show may have more than one video and it does not require any selection
+		for (var i = 0; i < videos.length; i++) {
+			var video = videos[i];
+			var source = video.getElementsByTagName('source')[0];
```
eszkadev, line 97: it can fail if no source tag is inside

eszkadev:
doesn't work: when I select video it shows snackbar with information that it is unsupported  (but video can be played in my browser):
[unsupvid-2023-10-11_13.49.00.webm](https://github.com/CollaboraOnline/online/assets/5307369/4c06dcc0-85fc-43af-9a95-36e35ff06b3e)

lpranam:
NOTE: This patch does not show a warning in the slideshow as the snake bar can not be displayed over the fullscreen element. will talk to @pedropintosilva about it and fix it in the follow-up.

eszkadev (approved):

eszkadev:
I think we can remove slideshow handler here? as it will not work, we need to code "snackbar" in presentation_engine.js generated by LibreOffice

timar:
I cherry-picked, resolved the conflict, and pushed to master directly. I could not figure out how to push here.

PR #7225, opened 2023-09-08
Turn canonical id assertion into warning for now

OFFTKP:


OFFTKP:
This is a trivial change and a bit urgent to not block people, so I'll merge after CI passes.

eszkadev (approved):
fixes cypress, already passed all without 1 tests

PR #7228, opened 2023-09-08
readonly

hcvcastro:
- kit: add 'readonly' property to status message
- browser: fire 'readonly' if status change


Ashod (approved):
Thanks @hcvcastro.

hcvcastro:
Merging patches due to unrelated error:

Assertion failure: prefix != res.substr(0, prefix.length()) 
Expected (prefix): [status:]
Actual (res.substr(0, prefix.length())):   []
                                           [^^^^^^^]
| helpers.hpp:432
coolwsd: ./helpers.hpp:432: std::__cxx11::string helpers::assertResponseString(const std::shared_ptr<http::WebSocketSession>&, const string&, const string&, std::chrono::milliseconds): Assertion `(exp__LINE__) == (act__LINE__)' failed.

PR #7230, opened 2023-09-09
Fix minor iOS build and runtime problems

plubius:
This patch contains changes to iOS and/or Android code only.

### DIFF
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -247,41 +247,45 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 		}
 	},
 
-	_createiOsFontButton: function(parentContainer, data, builder) {
-		// Fix issue #5838 Use unique IDs for font name combobox elements
-		var table = L.DomUtil.createWithId('div', data.id, parentContainer);
-		var row = L.DomUtil.create('div', 'notebookbar row', table);
-		var button = L.DomUtil.createWithId('button', data.id + 'ios', row);
-
-		$(table).addClass('select2 select2-container select2-container--default');
-		// Fix issue #5838 Don't add the "select2-selection--single" class
-		$(row).addClass('select2-selection');
-		$(button).addClass('select2-selection__rendered');
-
-		if (data.selectedEntries.length && data.entries[data.selectedEntries[0]])
-			button.innerText = data.entries[data.selectedEntries[0]];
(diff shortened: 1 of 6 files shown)


plubius:
@OFFTKP While testing, I may have found a bug as a side effect of commit https://github.com/CollaboraOnline/online/commit/04ac02b0339008382dcf640899ea18513de5a4a5.

When I open the following PDF document in the iOS app, the thumbnails in the left panel no longer get drawn. For some reason, when opening PDF documents, browser/src/layer/tile/CanvasTileLayer.js:1013 is never called:

https://github.com/CollaboraOnline/online/files/10777062/Cloudervaringsdocument%2BNCSC.pdf

plubius:
> @OFFTKP While testing, I may have found a bug as a side effect of commit [04ac02b](https://github.com/CollaboraOnline/online/commit/04ac02b0339008382dcf640899ea18513de5a4a5).
> 
> When I open the following PDF document in the iOS app, the thumbnails in the left panel no longer get drawn. For some reason, when opening PDF documents, browser/src/layer/tile/CanvasTileLayer.js:1813 is never called but is called for when opening .odp files:
> 
> https://github.com/CollaboraOnline/online/files/10777062/Cloudervaringsdocument%2BNCSC.pdf

Committed the following fix for this and added it to this pull request:

https://github.com/CollaboraOnline/online/pull/7230/commits/89a3df24ee6e8d5725706f6893d183550616548a

OFFTKP:
Please test if you can reproduce after my commit, I don't have an iOS devenv setup to check.

plubius:
> Please test if you can reproduce after my commit, I don't have an iOS devenv setup to check.

I just did a clean build with your commit. I can no longer reproduce the bug so your commit looks good to me.

OFFTKP:
If your PR is otherwise ready it would be great if we could merge this after CI passes because it might fix another blocker, and if not, at the very least a cypress test regression.

plubius:
> If your PR is otherwise ready it would be great if we could merge this after CI passes because it might fix another blocker, and if not, at the very least a cypress test regression.

My PR is ready. The other changes in this pull request only changes code for iOS only. I am not sure what privileges I have in this repo so if you can mark this as reviewed, I will see if I can merge this PR when CI passes.

OFFTKP:
In that case I'll merge. Cypress tests pass now.

OFFTKP (approved):

plubius:
> In that case I'll merge. Cypress tests pass now.

Thank you for your help.

PR #7239, opened 2023-09-11
a11y: improved screen reader caret tracking on text navigation

mcecchetti:
In order to allow screen reader to track caret navigation properly
even if there is some connection delay default behaviour for
Left/Right arrow key press is no more prevented in
Map.Keyboard._handleKeyEvent. In TextInput._onKeyDown, it has been
needed to handle some special caret position, such as when cursor is
at begin or end of a list item prefix.
When left/right arrow is pressed and text is selected, selection is
cleared and caret needs to be moved by one char left/right.
However, for an editable div the behaviour is different:
- when left arrow is pressed caret moves at start of previously
selected text
- when right arrow is pressed caret moves at end of previously
selected text
So we needed to prevent default behaviour and simulate the same
behaviour that occurs in LibreOffice.

Change-Id: Ifa5e9847bbbf77b934a3bcbd04545f4a55afd19d

### DIFF
--- a/browser/src/layer/marker/A11yTextInput.js
+++ b/browser/src/layer/marker/A11yTextInput.js
@@ -25,6 +25,8 @@ L.A11yTextInput = L.TextInput.extend({
 		this._lastCursorPosition = 0;
 		this._lastSelectionStart = 0;
 		this._lastSelectionEnd = 0;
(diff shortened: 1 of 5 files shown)


gokaysatir (approved):
Thanks for the improvements: :)

PR #7243, opened 2023-09-11
jsdialog: create also combobox without entries

eszkadev:
this fixes missing font name and font size comboboxes when opened spreadsheet protected from editing


gokaysatir (approved):
Thanks :)

PR #7246, opened 2023-09-12
Remove ~ character from translatable strings.

gokaysatir:
Reason: Some of the translatables are not read from the JSON file for some reason.

This is a workaround until we remove all ~ symbols from tab contents.

Change-Id: Ic3ffc9e3410099fc0d87ffbf46d208b9fc07c788

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir:
Not needed any more.

PR #7257, opened 2023-09-14
reduce cost of TileCombined::parse

caolanm:
perf reported 1% of time in collaborative multi user test on 2023-09-14 was spent in TileCombined::parse.

generic-ize and reuse the TileDesc::parse approach for TileCombined::parse to avoid need for std::unordered_map

Change-Id: Iadfc2001e298d8f4d46200c8488f0eb4cd8734c2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me - thanks =)

PR #7264, opened 2023-09-15
scripts: add a simple script to invoke a single unit test

vmiklos:
And also allow automatic re-trying to decide if this is a reproducible
failure or an unstable test.

Change-Id: I254d83cc1acc6b80ec3978e26e5fd64e676ff2a9

### DIFF
--- a/scripts/run-wsdunit
+++ b/scripts/run-wsdunit
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+#
+# 'make check' runs all wsdunit tests, this script allows running just one of them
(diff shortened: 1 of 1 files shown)


vmiklos:
> 15:29 <ash> @vmiklos: good idea! (I run individual test directly by invoking coolwsd, so this helps.) Minor suggestion is for the run script to return an error code, so we don't grep the log. +1 from me.

vmiklos:
@Ashod please take a look. Thanks!

Ashod (approved):
Thanks @vmiklos!

PR #7268, opened 2023-09-18
New copy paste dialogs / snackbars

gokaysatir:
Improvements in copy-paste:
- all dialogs/popups converted to jsdialogs
- when user does complex copy first time - dialogs explaining the process are present
- when user does complex copy after on fully successful process - snackbar is used for messages
- to quickly copy, download, confirm copy to clipboard - shortcut 3x CTRL + C can be used
- when snackbar with download button appears but user changes selection -> we hide snackbar

### DIFF
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -112,8 +112,10 @@ L.Control.DownloadProgress = L.Control.extend({
 		var buttonText = _('Copy') + ' (Ctrl + C)'; // TODO: on Mac Ctrl == Command?
 
 		if (inSnackbar) {
-			this._map.uiManager.showSnackbar(snackbarMsg, buttonText,
+			this._map.uiManager.showProgressBar(snackbarMsg, buttonText,
 				this._onConfirmCopyAction.bind(this), this.options.snackbarTimeout);
+
+			this._map.uiManager.setSnackbarProgress(100);
 
 			this.setupKeyboardShortcutForSnackbar();
 		} else {
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -107,16 +107,17 @@ L.Control.DownloadProgress = L.Control.extend({
 
 	_showDownloadComplete: function (inSnackbar) {
 		var modalId = this._getDownloadProgressDialogId();
-		var msg = _('Download completed and ready to be copied to clipboard.');
+		var snackbarMsg = _('Download completed and ready to be copied to clipboard.');
+		var dialogMsg = snackbarMsg + ' ' + _('From now on clipboard notifications will discreetly appear at the bottom.');
 		var buttonText = _('Copy') + ' (Ctrl + C)'; // TODO: on Mac Ctrl == Command?
 
 		if (inSnackbar) {
-			this._map.uiManager.showSnackbar(msg, buttonText,
+			this._map.uiManager.showSnackbar(snackbarMsg, buttonText,
 				this._onConfirmCopyAction.bind(this), this.options.snackbarTimeout);
 
 			this.setupKeyboardShortcutForSnackbar();
 		} else {
-			JSDialog.setMessageInModal(modalId, msg, '');
+			JSDialog.setMessageInModal(modalId, dialogMsg, '');
 			JSDialog.enableButtonInModal(modalId, modalId + '-response', true);
 			this.setupKeyboardShortcutForDialog(modalId);
 		}
@@ -250,6 +251,9 @@ L.Control.DownloadProgress = L.Control.extend({
 		this._map._clip.filterExecCopyPaste('.uno:Copy');
 		this._onClose();
 		this._setUserAlreadyWarned();
+
+		var msg = _('Content copied to clipboard');
+		this._map.uiManager.showSnackbar(msg);
 	},
 
 	_onClose: function () {
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -2,7 +2,7 @@
 /*
  * L.Control.DownloadProgress.
  */
-/* global _ $ */
+/* global _ $ JSDialog */
 L.Control.DownloadProgress = L.Control.extend({
 	options: {
 		snackbarTimeout: 20000,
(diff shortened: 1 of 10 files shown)


pedropintosilva (changes requested):
It would be nice if we could use again `CTRL` + `C`. To avoid the problem/situation you have mentioned maybe we could instead listen for selection changes -> once the selection changes the dialog should disappear and then no problem with using ctrl + C.

The  bottom right dialog:
* Shouldn't be on the right and shouldn't be leaflet dialog (nice if we could use snackbar)
* It shouldn't appear for the first time. For that we will rely on the JSDialog (showInfoModal)
 
image

And then after the first time we can finally use the snackbar 


image

And before that ^ we would need to add a progress bar

pedropintosilva:
Also I could trigger the primary button on the infoModal dialog when the focus is lost:
1. click anywhere in that dialog (except of the buttons)
2. Then try to trigger it with the shortcut

gokaysatir:
Thanks @eszkadev :)

This works beautifully. Only one thing: When the snackbar shows up and if the user doesn't want to download the content for external use, user has to either click on the document or download the content.

Maybe we could allow user to stay in the document and if the selection changes then process would begin from scratch - or user may not download the content at all.

Or at least we can enable ESC key to close the snackbar.

@pedropintosilva knows better, thanks for the improvements :)

pedropintosilva (changes requested):
Thanks @eszkadev ! Also much appreciated those additional changes accounting for selection change.

# 1. Multiple dialogs at same time
Everything seems to work out of the box but after
1. restarting the server
3. reloading the page
4. Move the cursor to the end of "Hello World"
5. Enter
6. Paste 40 paragraphs in one go (lorem ipsum) to the document
7. These appears:

multiple-dialogs-showing-up

Web browser logs: 
[multiple-dialogs-showing-up.txt](https://github.com/CollaboraOnline/online/files/12843807/multiple-dialogs-showing-up.txt)

# 2. Download progress dialog is too fast
This dialog is almost impossible to see, Even with a 100 pages long odt. It is important that the user sees that progress has reached the 100%. Otherwise it's perceived as a glitch or ignoring the "Download" part.

## 2.1 Proposal
The progress bar doesn't need to be tempered/ slowed down artificially. Instead,
* we would just let the filled progress bar there, visible even when switching to showDownloadComplete dialog.
* And the "Copy (Ctrl + C) " would be already visible in the Download progress dialog **but** disabled until the progress bar reaches 100%.
* Once the progress bar reaches 100% the Paragraph that is located above the progress bar ( "Downloading clipboard content") would be replaced with "Download completed and ready to be copied to clipboard."

# 3. Dialogs, each one looks different 
These 3 dialogs should be or look the same dialog. Meaning, all 3 should have the same width to be perceived as one single flow regarding the same problem/situation.

## 3.1 The Downloading dialog is too small
The title doesn't fit, but this would be fixed with `2. Download progress dialog is too fast`

# 4. When downloading the progress bar has weird visible focus
DownloadProgressDialog-focus-n-width-problem

pedropintosilva (changes requested):
Amazing @eszkadev it looks nice. 

# Ctrl + C doesn't work in the last dialog

However it seems the short cut doesn't work for me in the last dialog:

https://github.com/CollaboraOnline/online/assets/65948705/60c45500-b560-4153-a214-a4c33f4e3bb3

I have tested on FF and Chrome and I get the same result ^

# Improvements

Can we please change the following text
`Download completed and ready to be copied to clipboard.`
to
`Download completed and ready to be copied to clipboard. From now clipboard notifications will discreetly appear at the bottom.`

and trigger after that dialog a snackbar saying `Content copied to clipboard`?

browser/src/control/Control.DownloadProgress.js
```diff
@@ -2,79 +2,162 @@
...
+		var modalId = this._getDownloadProgressDialogId();
+		if (!this._userAlreadyWarned())
+			this._map.uiManager.closeModal(this._map.uiManager.generateModalId(modalId));
+	},
+
+	// Step 3. Download complete
+
+	_showDownloadComplete: function (inSnackbar) {
+		var modalId = this._getDownloadProgressDialogId();
+		var msg = _(' Download completed and ready to be copied to clipboard. From now clipboard notifications will discreetly appear at the bottom.');
```
pedropintosilva, line 110: `on` missing in `From now clipboard`. It should be `From now on`

pedropintosilva (changes requested):

browser/src/control/Control.DownloadProgress.js
pedropintosilva: Also there is a space before Download,

pedropintosilva:
Oh and can you please amend `c9943726a2b7c567086cad29e0bcfadd4064b647` : `clipboard: move all dialogs to Control.DownloadProgress` ?

```diff
diff --git a/browser/src/control/Control.DownloadProgress.js b/browser/src/control/Control.DownloadProgress.js
index c26ef25744..edd47897b6 100644
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -45,7 +45,7 @@ L.Control.DownloadProgress = L.Control.extend({
 
 	_getLargeCopyPasteMessage: function () {
 		return this._map._clip._substProductName(
-			_('If you want to share large elements with other applications (outside of %productName) it\'s necessary to first download them.'));
+			_('If you want to share large elements outside of %productName it\'s necessary to first download them.'));
 	},
 
 	_getDownloadProgressDialogId: function () {
```

pedropintosilva (changes requested):
# Text in the snackbar is circular

Could we please have a different string (from the dialog) for the snackbar? Otherwise we end up mentioning things such as "appear at the bottom", which when displayed in the snackbar sounds that we are talking about something else (like status bar)

image

# Snackbar with sticky progress bar
An additional note that would also solve the previous issue. Could we (following the same practice used in the dialogs) make the progress bar stay in the snackbar?

so it would go from :
image

to:

image

or better yet:

image

by adding to `#snackbar-container`
```css
grid-template-rows: repeat(2, auto);
grid-template-columns: repeat(2, auto);
```
and adding to `#progress`
```css
display: grid;
grid-column: 1 / 3;
```

eszkadev:
I fixed above problems, apart of snackbar layout in 2 rows. This is hard to do now without breaking other sidebars.

pedropintosilva (approved):

eszkadev (approved):

PR #7278, opened 2023-09-20
Fix "modified state notifier" visibility.

gokaysatir:
Change-Id: If4feb0806ce92fadeb90f6676809bcbefc51eb24

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -236,13 +236,11 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 			$('#applystyle').val(state).trigger('change');
 		}
 		else if (commandName === '.uno:ModifiedStatus') {
-			if (e.state === 'true') {
-				$('#Save1').addClass('savemodified');
-				$('#Save').addClass('savemodified');
-			}
-			else {
-				$('#Save1').removeClass('savemodified');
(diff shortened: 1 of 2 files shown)


pedropintosilva (commented):
Checked , some changes missing, please @gokaysatir review it

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -236,13 +236,15 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
...
-				$('#Save1').removeClass('savemodified');
-				$('#Save').removeClass('savemodified');
+			if (document.getElementById('save')) {
+				if (state === 'true') {
+					document.getElementById('save').classList.add('savemodified');
+					document.getElementById('file-save').classList.add('savemodified');
+				}
+				else {
+					document.getElementById('save').classList.remove('savemodified');
+					document.getElementById('file-save').classList.remove('savemodified');
```
pedropintosilva, line 246: I thought there is no need to check if file-save element exists since if we are in presence of notebookbar then the File tab should always exist, right? Or should we anyhow check it exists .
In the pass we didn't check at all.

pedropintosilva (approved):

pedropintosilva:
@gokaysatir please re-review it

PR #7281, opened 2023-09-20
indirection: use errorcode to map errormsg and translate it

Rash419:
Change-Id: I779bfc658c54255a675e3a047ed371e7e0586327

* Target version: master

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1,5 +1,5 @@
 /* -*- js-indent-level: 8 -*- */
-/* global Uint8Array */
+/* global Uint8Array _ */
 
 /*
 	For extending window.app object, please see "docstate.js" file.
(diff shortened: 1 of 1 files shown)


browser/js/global.js
```diff
@@ -761,12 +761,18 @@ window.app = {
 					that.onmessage(e);
 				};
 			} else if (this.status === 202) {
+				var errorMsg;
+				if (http.response.errorCode == 0) {
+					errorMsg = _('Cluster is scaling, please retry after few seconds');
```
quikee, line 766: Would be probably good to have this in a separate function.

quikee (approved):
Looks good

Rash419:
don't merge until I say

PR #7287, opened 2023-09-23
Fix mobile autofilter and add cypress test

eszkadev:


gokaysatir (approved):
Thanks for the fix :)

PR #7290, opened 2023-09-24
Fix UnitBadDocLoad test after repair dialog is enabled

eszkadev:
commit	413e7a2e6f0c363938a0a5ee5bd6eb5d56d5a8d8
lok: add broken package interaction handler
https://cgit.freedesktop.org/libreoffice/core/commit/?h=distro/collabora/co-23.05&id=413e7a2e6f0c363938a0a5ee5bd6eb5d56d5a8d8

in core added dialog which asks user if file should be repaired. This requires tests adjustment.

### DIFF
--- a/test/UnitBadDocLoad.cpp
+++ b/test/UnitBadDocLoad.cpp
@@ -62,6 +62,20 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
         // Send a load request with incorrect password
         helpers::sendTextFrame(socket, "load url=" + documentURL, testname);
 
+        // We receive jsdialog with question about repair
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Great! Thanks @eszkadev.

I also had something similar, but... https://github.com/CollaboraOnline/online/pull/7232/commits/4f8b3a6ed9840b94c368569d0ae5acf1c4d9e6f4 (I'll remove it.)

One suggestion: can we verify that the dialog we get is the corruption one?
```
LOK_ASSERT(dialog1.find("is corrupt and therefore cannot be opened.") !=
                   std::string::npos);
```

PR #7291, opened 2023-09-25
remove signing-identities.html

timar:
it belonged to the removed unused document signing code

Change-Id: Idf57b30a508381e02c8d62196de38c828cd77f2f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
This can be removed without any risk.

* `$.get('signing-identities.html', function(data) {` doesn't exist anymore
* Signing.js control has been redone and it is now relying solely on core side, no libraries, logins etc

Issue #7141, opened 2023-08-23
Mobile: buttons with text color discrepancies


pedropintosilva:
### Describe the Bug

Open mobilewizard and notice that button are being styled rather oddly :
![image](https://github.com/CollaboraOnline/online/assets/65948705/509ade73-f61d-4d2f-85b7-82d410816af8)

This seems to do with the accessKeys spilling to mobile which might not be useful here:
```html
<button class="ui-pushbutton mobile-wizard" id="ok" accesskey="O" tabindex="0">
    <u class="access-key">O</u>K
</button>
```

### Expected Behavior

 - [ ] Make sure that `#mobile-wizard button u` is set with the same color as we are setting `#mobile-wizard button`
 - [ ] Maybe we could rethink if we should allow these sub-elements to be generated for mobile

Darshan-upadhyay1110:
there are two solutions i am seeing here 
 - Simple solution which would be to add a css for `<u>` inside #mobile-wizard-button. but i am seeing this as a temporary solution.
 - proper fix would be to check if we need this <u> tag for acces key ( because some of the structrure changed here for acces keys. not sure of this need @gokaysatir's view here 
  - and if this element not needed any more then just remove that element generation thing and all willl be fixed.
  
  
  CC: @gokaysatir

PR #7292, opened 2023-09-25
Mobile: buttons with text color discrepancies

Darshan-upadhyay1110:
Will not create `<u>` element for mobile wizard

Change-Id: Iaca88ac4be2c95dc4781180019ba8fede9cdc81a

* Resolves: #7141 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -502,7 +502,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	},
 
 	_stressAccessKey: function(element, accessKey) {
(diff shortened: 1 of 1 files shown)


Darshan-upadhyay1110:
@gokaysatir i think it will okay for mobile to not create access-key element. need your comment here for this.

i have also another solution. we can fix this by just adding new css rule and that is it. but if we do not need access-key on mobile view current solution will be more usefull i guess 
:)

gokaysatir (approved):
Thanks @Darshan-upadhyay1110 :) i also think we don't need them on mobile view.

PR #7293, opened 2023-09-25
Private/mmeeks/simd

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -180,7 +180,6 @@ lokitclient_SOURCES = common/Log.cpp \
 
 noinst_LIBRARIES = libsimd.a
 libsimd_a_SOURCES = kit/DeltaSimd.c
-# FIXME: x86_only etc.
 libsimd_a_CFLAGS = @SIMD_CFLAGS@
 
 coolforkit_sources = kit/ChildSession.cpp \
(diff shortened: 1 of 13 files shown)


mmeeks:
@caolanm Measured - and it appears to (only) double the speed of the RLE code - I was hoping for more, but 2x is at least reasonable... can you review and give it some testing ?
Interestingly - if you compile with -O0 we are slower with AVX2, but at -O2 we are 2x faster - perhaps that comes from pre-optimizing the C version ;-) Thanks !

caolanm:
we're missing DeltaSimd.h, presumably needs a git add for that

Makefile.am
```diff
@@ -115,6 +115,7 @@ shared_sources = common/FileUtil.cpp \
                  common/ConfigUtil.cpp \
                  common/Authorization.cpp \
                  common/CommandControl.cpp \
+		 common/Simd.cpp \
```
hfiguiere, line 118: 'nit: indentation. My OCD screams here. Also line 319.

caolanm (approved):
passes ci now

PR #7295, opened 2023-09-25
Switch hover to JS to stop touch-hover emulation

### LINKED ISSUE #6200: iOS App: Highlight color persists when a menu option is "unselected" until another selection is made
**Describe the bug**
The highlight persists when a menu option is "unselected" until another selection is made

**To Reproduce**
In iOS app Version: 22.05.12.4 (22.5.202304132141)

1. Open a .ods file in Tabbed View
2. Go to View tab > Status bar  to reveal the status bar
3. Click on the Statusbar again to hide the status bar
4. Notice the highlight color still persists

Select Sidebar, the highlight colour disappears from the Statusbar

**Expected behavior**
The highlight color should not persist after the menu is unselected.  

**Actual behavior**
The highlight persists when a menu option is "unselected" until another selection is made

**Screenshots**
Here is a screen-recording

https://user-images.githubusercontent.com/54344301/232453083-4019ece9-f50d-4e33-9f9c-3ed8cc94135d.MP4

**Desktop (please complete the following information)**
 
 - Version: 22.05.12.4 (22.5.202304132141)

**Smartphone (please complete the following information)**
 - Device: iPad
 - OS: iOS 16

Minion3665:
This PR changes the :hover selector to be a .hover class to stop devices without a pointer "hovering" after you press on an element. This fixes #6200

These other alternatives were considered and rejected
- Using media queries at the point where we use :hover
  - There's a lot of places where we use :hover. This would be an annoying task
  - This would require us to modify files such as bulma.min.css in nontrivial ways. Bulma is a CSS framework (https://bulma.io). If we ever wanted to update this, we would have to make the changes again. With this commit, a regex find and replace with `/:hover([^_a-zA-Z0-9-]|$)/g` is adequate
- Moving all hover classes into another file, and using a single media query
  - Pretty much the same downsides as the above, except if we ever wanted to update a stylesheet we would also need to figure out which were the old styles

The following potential downsides were considered
- The event order could be such that hover is not applied properly
  - According to <https://www.w3.org/TR/DOM-Level-3-Events/#events-mouseevent-event-order>, compliant browsers will never do this
- A device could have both a pointer and a touchscreen
  - As far as I know, there's no way I can detect which is used. My current code decides to avoid breaking mouse compatibility at all by enabling hovering *even if there is a touchscreen as the main input device* so long as there is also a pointer

Change-Id: Ida56dca2e4d611cb923b3d16b7bbea959ad7d2ea

* Resolves: #6200 
* Target version: master 

### Checklist


eszkadev:
Please rebase - cypress should pass

eszkadev:
I'm wondering what is the real reason of that and what exactly happens.

I can imagine we did some event.preventDefault() which doesn't allow to unhover things...
Because if they are hovered in the first place probably that means is should work. Or is it really broken on mobile? :)

I suspect it goes like - when user clicks on a button:
- we do hover event (item starts to match :hover)
- we do click handler where we do our code and preventDefault()
here browser doesn't process it's own click handler which turns off :hover 

Could you experiment with that and check if it could happen?

Minion3665:
@eszkadev 

> I'm wondering what is the real reason of that and what exactly happens.
> 
> I can imagine we did some event.preventDefault() which doesn't allow to unhover things... Because if they are hovered in the first place probably that means is should work. Or is it really broken on mobile? :)
> 
> I suspect it goes like - when user clicks on a button:
> 
>     * we do hover event (item starts to match :hover)
> 
>     * we do click handler where we do our code and preventDefault()
>       here browser doesn't process it's own click handler which turns off :hover
> 
> 
> Could you experiment with that and check if it could happen?

Unfortunately mobile browsers do not have a click handler to turn off hover... please see this jsfiddle https://jsfiddle.net/uw0x78zk/6/ which has no javascript at all and still exhibits the same behavior

eszkadev:
I see, checked that example and it really has this issue...

eszkadev:
I would like to avoid adding own `.hover` class. Maybe we can do this:
- detect case when hover is supported by media query as you do
- in that case add to the body class `supports-hover`
- in all cases where we use `:hover` do      `.supports-hover  someselector:hover`

In my opinion this looks better, @pedropintosilva what do you think?

pedropintosilva:
> I would like to avoid adding own `.hover` class. Maybe we can do this:
> 
>     * detect case when hover is supported by media query as you do
> 
>     * in that case add to the body class `supports-hover`
> 
>     * in all cases where we use `:hover` do      `.supports-hover  someselector:hover`
> 
> 
> In my opinion this looks better, @pedropintosilva what do you think?

Ideally it would be nice to avoid additional JS checks or even CSS media queries  if it's not needed.
We already generate from src a couple of .css files (device-desktop.css, device-tablet.css, device-mobile.css) It seems to me that the most logical thing to do would be to leverage that check and just put all the hover related rules in the desktop.css at the end 


-----

Another way would be to add another data attribute to html something like `data-device` and leverage the check that we already have to add the respective value (desktop || tablet || mobile) and then just add `html:[data-device='desktop'] before any `element:hover`

Another way, but I don't think it makes sense,  to do it would be to replace all the values of all class selector with pseudo class (hover) css rules with css vars. Those css vars would be then reset on mobile and tablet to have the same value as the non-hover css vars.
 * Problems
   * At the end we will end up with a bigger change when doing this
   * We would need to create all those missing css vars and make them specific for those elements or element categories
   * And then add the css reset block (to be reset to the previous state before hover)

eszkadev:
Old PR, we will need other solution anyway and has conflicts. Let's close for now.

PR #7302, opened 2023-09-25
persist slide number in impress and sheet number

Rash419:
in calc on reconnect

Change-Id: I7dafcd5f8d57ae79952153faf90b682f17a33cfd


gokaysatir (approved):
Thanks.

PR #7305, opened 2023-09-26
postmessage: fix get export formats

eszkadev:
- added get_export_formats postmessage invoke button in debug postmessage page
- fixed not present download entries when grouped_download_as is disabled

    This fixes regression in Get_Export_Formats where we
    didin't report possible formats when in coolwsd.xml
    option "group_download_as" was set to false.
    
    Regression comes from:
    commit 939a7a030109352a34d9b28a7dcd6f04325cc1b0
    Simplify notebookbarwriter and tab page containers.
    
    Where JSON for File tab was incorrectly built and
    some of the entries were skipped.

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -90,6 +90,12 @@
             });
       }
 
+      function getExportFormats() {
+        post({'MessageId': 'Get_Export_Formats',
+              'Values': null
+            });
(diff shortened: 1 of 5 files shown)


gokaysatir (approved):
Tested, thank you for the fix @eszkadev.

PR #7308, opened 2023-09-26
mobile: allow scroll outside data area

eszkadev:
mobile screen is very small and there was not
enought space to scroll to the place where we "extend" scrollable area, this caused that in spreadsheets we were not able to scroll outside last used cells

this patch make us able to scroll 4 lengths of screen outside data for non-desktop devices to fix that issue

regression commit 48ec2f5925edec5550a1320ae7815a9bd86c4e12 Limit scrollbar range to used area in the spreadsheet


mmeeks (approved):
Looks great to me - thanks Szymon =)

pedropintosilva:
don't merge this until @eszkadev says so

Rash419 (approved):
works as expected in mobile view, thanks

PR #7310, opened 2023-09-26
Correctly extract file name for exported file

eszkadev:
We assumed in the code that LO will return in a
LOK_CALLBACK_EXPORT_FILE url starting with "file:///tmp/" But that is not true for all the cases. Let's use
more generic approach to extract file name.


Ashod (approved):
Thanks @eszkadev. Looks reasonable, but I think we should verify (in WSD) that the path we get from the Kit matches the childroot. We can do that at the point of processing the `registerdownload:` message.

PR #7314, opened 2023-09-27
force the tabbed user interface for new installations and containers

timar:
Previously we respected the choice of integrator, i.e. the "default" was the integrator's preference, either compact mode or tabbed mode. But clearly the tabbed mode is the future, compact mode is legacy and not developed. Compact mode is not a good default. Tabbed mode is the preferred default, it's modern and accessible. Users who like compact mode better, can switch back easily.

Change-Id: If772bad748662090edaf57fc233415f74346da8b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/coolwsd.xml.in
+++ b/coolwsd.xml.in
@@ -211,7 +211,7 @@
     @WELCOME_CONFIG_FRAGMENT@
 
     <user_interface>
-      <mode type="string" desc="Controls the user interface style. The 'default' means: Take the value from ui_defaults, or decide for one of compact or tabbed (default|compact|tabbed)" default="default">default</mode>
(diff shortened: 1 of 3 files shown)


mmeeks (commented):
If there is no option in the UI_Defaults - what do we choose? - I'm thinking of our older integrations out of interest.
But not against this - we need to take action here I think.

timar:
> If there is no option in the UI_Defaults - what do we choose? - I'm thinking of our older integrations out of interest. But not against this - we need to take action here I think.

We defaulted to tabbed mode in that case. Without this PR you get compact mode only, if you ask for it, with this PR whatever you ask, you get tabbed mode, which is a bit rude, but if I understood properly, we wanted this.

In Nextcloud richdocuments the default UIMode was classic (compact), but since May 30, 2023, it is notebookbar (tabbed). 

So I can imagine that this PR is unnecessary.

mmeeks:
Well - I like all the code paths that default to tabbed - but the XML config thing prolly should be tweaked (?) =)

PR #7324, opened 2023-09-28
don't request new tiles unnecessarily on canonicalidchange

Rash419:
- it applies for 2 cases:
1. when document is loaded for the first time document already has tiles of respective theme
2. when Kit loads the document it sends canonicalidchange  unconditionally (https://github.com/CollaboraOnline/online/blob/5520965b151cb854acbac3f1bbb68aec5ac2e1f2/kit/Kit.cpp#L1708) we don't need to request new tiles for that also

Change-Id: I42cdc5a03e70c3d3d653f3124d3d5ed9382e22c0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

- [] I have run tests with `make check`

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -12,6 +12,7 @@ L.Control.UIManager = L.Control.extend({
 	busyPopupTimer: null,
 	customButtons: [], // added by WOPI InsertButton
 	modalIdPretext: 'modal-dialog-',
+	previousTheme: null,
 
 	onAdd: function (map) {
 		this.map = map;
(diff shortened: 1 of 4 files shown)


OFFTKP:
Theme will be more than just Light/Dark in the future so a bool state seems wrong. Can you also check if previews work fine with this commit? Open 2 browsers on the same document which has multiple slides, switch dark mode to one, type stuff, see if everything looks fine.

Rash419:
@OFFTKP Thanks for the review 

> Theme will be more than just Light/Dark in the future so a bool state seems wrong. 

Done

> Can you also check if previews work fine with this commit? Open 2 browsers on the same document which has multiple slides, switch dark mode to one, type stuff, see if everything looks fine.

Everything looks good to me!

Ashod (commented):
Thanks @Rash419. Good catch and improvement. 

Ideally, we will have a test that validates the number of tiles requested/rendered after a reconnection. This will also help defend against future regressions (and it would have caught this particular case of duplicate tiles).

OFFTKP (approved):

PR #7326, opened 2023-09-28
German shortcut changes for Calc

gokaysatir:
Change-Id: I56ff4672699860208b37c0aad575dc4ba294763b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -465,7 +465,7 @@ L.Map.Keyboard = L.Handler.extend({
 		}
 
 		// save-as for German shortcuts - F12.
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Instead of adding all that ifs can we do it somehow more generic ? with possibility to extend in the future for other languages?

like some array
```
[
{
    lang: 'de',
    shortcuts: [
        {altKey: false, shiftKey: true, keyCode: keyCodes.F12, unoCommand: '.uno:FunctionDialog'}, // <-- using pure uno command
        {app: ['text', 'spreadsheet'], altKey: true, shiftKey: false, keyCode: keyCodes.R, dispatchCommand: 'focus_formulabar'} // <-- using map.dispatch for custom logic
        ...
        ]
    }
]
```

gokaysatir:
@eszkadev thanks for the review and hopefully we will have it more generic in the near future :)

PR #7329, opened 2023-09-29
WIP : Add KIT in process mode

Darshan-upadhyay1110:

Change-Id: I153bdeed50e86c9a97a60591f2644eb6b4c51f00

 - have a mode KITINPROCESS (limited to --enable-debug) where everything is running as a single process
 - global function that will know if we're in kit-in-process mode,=> Util::isKitInProcess()

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -328,8 +328,11 @@ namespace Util
 
     bool isKitInProcess()
     {
-        const char* kitInProcess = std::getenv("KIT_IN_PROCESS");
-        return (kitInProcess != nullptr && strcmp(kitInProcess, "true") == 0);
+#ifdef KIT_IN_PROCESS
+        return true;
+#else
+        return false;
+#endif
     }
 
 #if !MOBILEAPP
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -325,6 +325,13 @@ namespace Util
         return std::getenv("DISPLAY") != nullptr;
     }
 
+
+    bool isKitInProcess()
+    {
+        const char* kitInProcess = std::getenv("KIT_IN_PROCESS");
+        return (kitInProcess != nullptr && strcmp(kitInProcess, "true") == 0);
+    }
+
 #if !MOBILEAPP
 
     static const char *startsWith(const char *line, const char *tag, std::size_t tagLen)
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -190,7 +190,7 @@ namespace Util
 
     bool windowingAvailable();
 
-#if !defined(BUILDING_TESTS) && !defined(KIT_IN_PROCESS) && !MOBILEAPP
+#if !defined(BUILDING_TESTS) && !MOBILEAPP
 
     /// Send a message to all clients.
     void alertAllUsers(const std::string& msg);
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -1300,6 +1300,8 @@ int main(int argc, char**argv)
      */
     bool isFuzzing();
 
+    bool isKitInProcess();
+
     /**
      * Splits string into vector<string>. Does not accept referenced variables for easy
      * usage like (splitString("test", ..)) or (splitString(getStringOnTheFly(), ..))
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -56,8 +56,9 @@ static bool NoSeccomp = false;
 static bool SingleKit = false;
 #endif
 #else
-static const bool NoCapsForKit = true; // NoCaps for in-process kit.
-static const bool NoSeccomp = true; // NoSeccomp for in-process kit.
+static bool NoCapsForKit = true; // NoCaps for in-process kit.
+static bool NoSeccomp = true; // NoSeccomp for in-process kit.
+static bool SingleKit = true;
 #endif
 
 static std::string UserInterface;
@@ -196,7 +197,6 @@ protected:
     }
 };
 
-#ifndef KIT_IN_PROCESS
 #ifndef __FreeBSD__
 static bool haveCapability(cap_value_t capability)
 {
@@ -275,7 +275,6 @@ static bool haveCorrectCapabilities()
     return getuid() == 0;
 }
 #endif // __FreeBSD__
-#endif
 
 /// Check if some previously forked kids have died.
 static void cleanupChildren()
@@ -360,6 +359,7 @@ static void cleanupChildren()
             cleanupJailPaths.erase(cleanupJailPaths.begin() + i);
     }
 }
+
 // lokit_main should only call once, there maybe chances createLibreOfficeKit called more then once
 // and that leads to execute multiple threads
 static bool lokitProcessIsRunning = false;
@@ -382,17 +382,19 @@ static int createLibreOfficeKit(const std::string& childRoot,
                                                       << spareKitId << '.');
     const auto startForkingTime = std::chrono::steady_clock::now();
     pid_t pid = 0;
-    if (Util::isKitInProcess() && !lokitProcessIsRunning)
+    if (Util::isKitInProcess())
     {
-        std::thread(
-            [childRoot, jailId, sysTemplate, loTemplate, queryVersion]
-            {
-                Util::setThreadName("kit_in_process");
-                lokit_main(childRoot, jailId, sysTemplate, loTemplate, true, true,
-                           queryVersion, DisplayVersion, spareKitId);
-            })
-            .detach();
-        lokitProcessIsRunning = true;
+        if (!lokitProcessIsRunning)
+        {
+            lokitProcessIsRunning = true;
+            std::thread(
+                [childRoot, jailId, sysTemplate, loTemplate, queryVersion]
+                {
+                    lokit_main(childRoot, jailId, sysTemplate, loTemplate, true, true,
+                            queryVersion, DisplayVersion, spareKitId);
+                })
+                .detach();
+        }
     }
     else
     {
@@ -403,10 +405,6 @@ static int createLibreOfficeKit(const std::string& childRoot,
 
             // Close the pipe from coolwsd
             close(0);
-            if (!Util::isKitInProcess())
-            {
-                UnitKit::get().postFork();
-            }
 
             if (std::getenv("SLEEPKITFORDEBUGGER"))
             {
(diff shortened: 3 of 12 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -1121,13 +1121,15 @@ bool DocumentBroker::download(const std::shared_ptr<ClientSession>& session, con
...
-                        return false;
+                    if (!Util::isKitInProcess()) {
+                        int process = Util::spawnProcess(command, args);
+                        int status = -1;
+                        const int rc = ::waitpid(process, &status, 0);
+                        if (rc != 0)
+                        {
+                            LOG_ERR("Conversion from " << extension << " to " << newExtension << " failed (" << rc << ").");
+                            return false;
+                        }
```
vmiklos, line 1132: Command used to be a separate process. It has a main function. Probably you need an else branch here, so that you invoke that main function on a thread?

configure.ac
```diff
@@ -192,6 +192,10 @@ AC_ARG_ENABLE([androidapp],
                               to work similarly to the iOS app, from the JavaScript and the pseudo WebSocket
                               message plumbing point of view.]))
 
+AC_ARG_ENABLE([kitinprocess],
+              AS_HELP_STRING([--enable-kitinprocess],
```
vmiklos, line 196: Nit: probably this is not needed, this mode could be enabled at runtime with a cmdline switch and just be enabled in debug mode. The more ifdefs we have, the more likely CI doesn't catch a build breakage.

kit/ForKit.cpp
```diff
@@ -383,50 +383,61 @@ static int createLibreOfficeKit(const std::string& childRoot,
     LOG_DBG("Forking a coolkit process with jailId: " << jailId << " as spare coolkit #"
                                                       << spareKitId << '.');
     const auto startForkingTime = std::chrono::steady_clock::now();
+    pid_t pid = 0;
+    if (Util::isKitInProcess()) {
```
mmeeks, line 387: Makes sense of course - we should of course guard against doing this twice - having two threads in lokit_main will be a nightmare =) can you add a LOG_ERR and perhaps a sleeping loop for that mode - I guess we have this pre-forking logic that will try to always launch a 2nd idle process when one is used so - sleeping rather than asserting makes sense for that.

mmeeks (commented):
Looks like a fine start - I expect we'll want a different coolwsd process - this is most likely the easiest way to get everything linked into one binary.

Create a cool-onedoc binary or something in the Makefile.am - that links in all of the coolwsd and coolforkit source files - and then tweak / unwind the conflicts / overlaps etc.

Remove the 'spawnProcess' that starts the forkit - and instead run its 'main' or some other entry point you can add.

To avoid both 'main' functions conflicting - it's likely we'll need to split out two very small 'coolwsd-main' and 'forkit-main' files that just call an extern "C" int coolwsd_main(int argv, char **argv); type function each - since we can't have two main functions in the same linked object.

Does that help ?

common/Util.hpp
```diff
@@ -1300,6 +1300,8 @@ int main(int argc, char**argv)
      */
     bool isFuzzing();
 
+    bool isKitInProcess();
```
hfiguiere, line 1303: shouldn't this me `const` ?

Darshan-upadhyay1110: just a function declaration so i think it is okay :)

hfiguiere: duh I misread the scope. nevermind.

kit/ForKit.cpp
Darshan-upadhyay1110: Question : i am not sure about which mode you are refering here. is it the KitInProcess where the new thread has been created and we call lokit_main ?

and the LOG_ERR is also for that main function call right ?

wsd/COOLWSD.cpp
```diff
@@ -3724,22 +3699,23 @@ class PrisonerRequestDispatcher final : public WebSocketHandler
             LOG_TRC("Child connection with URI [" << COOLWSD::anonymizeUrl(request.getUrl())
                                                   << ']');
             Poco::URI requestURI(request.getUrl());
-#ifndef KIT_IN_PROCESS
-            if (requestURI.getPath() == FORKIT_URI)
-            {
-                if (socket->getPid() != COOLWSD::ForKitProcId)
+            if(!Util::isKitInProcess()) {
```
Ashod, line 3702: Please use clang-format to have uniform formatting everywhere. You can configure vscode to apply formatting only on edited lines.

common/Util.cpp
```diff
@@ -325,6 +325,13 @@ namespace Util
         return std::getenv("DISPLAY") != nullptr;
     }
 
+
+    bool isKitInProcess()
+    {
+        const char* kitInProcess = std::getenv("KIT_IN_PROCESS");
+        return (kitInProcess != nullptr && strcmp(kitInProcess, "true") == 0);
```
Ashod, line 332: This can be static. Also, is there a reason to have it set to "true" and "false"? We can simply define the `KIT_IN_PROCESS` for 'true' and not define it for 'false'. And avoid this comparison (just check for `!= nullptr`).

kit/ForKit.cpp
```diff
@@ -383,18 +380,28 @@ static int createLibreOfficeKit(const std::string& childRoot,
...
+        std::thread ([childRoot, jailId, sysTemplate, loTemplate, queryVersion]
+            {
+                Util::setThreadName("kit_in_process");
+                lokit_main(childRoot, jailId, sysTemplate, loTemplate, NoCapsForKit, NoSeccomp,
+                   queryVersion, DisplayVersion, spareKitId);
+            }).detach();
+        lokitProcessIsRunning = true;
+    }
+    else {
+        pid = fork();
```
Ashod, line 395: Please run clang-format, or apply proper formatting. This is all wrong.

kit/ForKit.cpp
```diff
@@ -383,18 +380,28 @@ static int createLibreOfficeKit(const std::string& childRoot,
...
-        // Child
-
-        // Close the pipe from coolwsd
-        close(0);
+        std::thread ([childRoot, jailId, sysTemplate, loTemplate, queryVersion]
+            {
+                Util::setThreadName("kit_in_process");
+                lokit_main(childRoot, jailId, sysTemplate, loTemplate, NoCapsForKit, NoSeccomp,
+                   queryVersion, DisplayVersion, spareKitId);
+            }).detach();
```
Ashod, line 391: Why are we detaching? Why isn't the thread owned by coolwsd?

Can you share the design of this whole PR please?

kit/ForKit.cpp
```diff
@@ -423,11 +430,11 @@ static int createLibreOfficeKit(const std::string& childRoot,
             LOG_INF("Forked kit [" << pid << ']');
             childJails[pid] = childRoot + jailId;
         }
-
-#ifndef KIT_IN_PROCESS
-        UnitKit::get().launchedKit(pid);
-#endif
+        if (!Util::isKitInProcess()) {
```
Ashod, line 433: Please do not use javascript formatting in C++ code. Apply clang-format if it's too much work.

wsd/debug-main.cpp
```diff
@@ -0,0 +1,57 @@
...
+        Util::setThreadName("debug main kit_in_process");
+        int result = forkit_main(args.size(), argv);
+
+        // Fulfill the promise with the result.
+        resultPromise.set_value(result);
+    }).detach();
+
+    // Wait for the thread to finish and get the result using a future.
+    std::future<int> resultFuture = resultPromise.get_future();
+    int result = resultFuture.get();
```
Ashod, line 46: The proper way of waiting for a thread to finish is to join the thread.

There is no good reason to detach the thread and use a promise/future.
```
int result = 0;
std::thread worker = std::thread ([&] {
        Util::setThreadName("debug main kit_in_process");
        result = forkit_main(args.size(), argv);
    });

worker.join();
return result;
```

wsd/COOLWSD-main.cpp
```diff
@@ -0,0 +1,40 @@
+/*
```
Ashod, line 1: Not clear to me why we need all these new files. Is there an explanation on the design somewhere?

kit/ForKit.cpp
```diff
@@ -366,6 +361,8 @@ static void cleanupChildren()
     }
 }
 
+static bool lokitProcessIsRunning = false;
```
Ashod, line 364: Please add comments so it is clear what the various variables/functions do. It's not clear what this is supposed to represent and how it's supposed to work.

wsd/COOLWSD.cpp
```diff
@@ -6383,7 +6356,9 @@ void forwardSigUsr2()
 int main(int argc, char** argv)
 {
     SigUtil::setUserSignals();
-    SigUtil::setFatalSignals("wsd " COOLWSD_VERSION " " COOLWSD_VERSION_HASH);
+    if (!Util::isKitInProcess()) {
+        SigUtil::setFatalSignals("wsd " COOLWSD_VERSION " " COOLWSD_VERSION_HASH);
+    }
```
Ashod, line 6361: Why? Why remove the fatal signal in 'kit in process' mode?

Ashod (commented):
Thanks @Darshan-upadhyay1110. I left some comments and suggestions, as well as questions.

wsd/COOLWSD-main.cpp
eszkadev: @Ashod comment above will explain: https://github.com/CollaboraOnline/online/pull/7329#pullrequestreview-1655289137

common/Util.cpp
eszkadev: @Darshan-upadhyay1110 as you have separate binaries now: regular one and  "single process" one, that env variable is not needed I think.
You should make this true when you link single process code probably.

wsd/COOLWSD.cpp
Darshan-upadhyay1110: ohh yes, sorry we only need to remove fatal signal for Forkit and Kit not for COOLWSD ( pushing that change...)

kit/ForKit.cpp
Darshan-upadhyay1110: thanks @Ashod i updated the formatting mistakes but there might be some still present sorry for that. i am still working on this patch so will update all formatting mistakes  soon...

JaumePujantell:
Push my current local state (not working) so @caolanm  can see it.

mmeeks:
Eventually we merged this in another pull request; thanks Darshan =)

Issue #7321, opened 2023-09-28
Calc: tabbed view: Sparkline: Format options don't work


pedropintosilva:
### Steps to Reproduce

1. The issue is reproducible only using Tabbed View (notebookbar).
2. No option in Format->Sparklines submenu is working

### Expected Behavior

Compact mode works. It should also work on tabbed view.

Where:
https://staging-perf.eu.collaboraonline.com/nextcloud/index.php/s/tESPC8GKwiMTXaQ

timur-g:
I do not reporduce this locally not on staging.eu. There must be some detailll.

gmasei11:
> I do not reporduce this locally not on staging.eu. There must be some detailll.

The scenario from the description is not the right one. Please see the below one:
1. The issue is reproducible only using Tabbed View (notebookbar).
2. No option in Format->Sparklines submenu is working

pedropintosilva:
For me  even the insert didn't work, but let me edit the description

timur-g:
Thanks.

PR #7335, opened 2023-09-29
Calc: fix Format->Sparkline menu items in notebookbar

gmasei11:
Change-Id: Ic51d3d020568e4c9488cc8be71ea446e66227a25

* Resolves: #7321 
* Target version: master 

### TODO

- [ ] ...

### Checklist


quikee (approved):
Looks fine to me, but don't know why this is necessary considering there are other commands that are prefixed with ".uno:" just in the menus around. Probably we need to change something else also so we can consistently define the commands or make it work both both - with ".uno:" and without.

PR #7338, opened 2023-09-29
Fix compiler warnings when building on macOS Sonoma

plubius:
Note: no patches were needed to build LibreOffice branch distro/collabora/co-23.05 on macOS Sonoma.

### DIFF
--- a/common/MessageQueue.hpp
+++ b/common/MessageQueue.hpp
@@ -82,7 +82,7 @@ public:
     /// Thread safe remove_if.
     void remove_if(const std::function<bool(const Payload&)>& pred)
     {
-        std::remove_if(_queue.begin(), _queue.end(), pred);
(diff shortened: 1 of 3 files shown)


common/MessageQueue.hpp
```diff
@@ -82,7 +82,7 @@ class MessageQueue
     /// Thread safe remove_if.
     void remove_if(const std::function<bool(const Payload&)>& pred)
     {
-        std::remove_if(_queue.begin(), _queue.end(), pred);
+        (void)std::remove_if(_queue.begin(), _queue.end(), pred);
```
caolanm, line 85: I think this use of std::remove_if is really dubious without a paired std::erase to actually remove the shuffled unwanted element. Luckily it seems that this MessageQueue::remove_if isn't actually used and this is dead code. I think we should remove the dead code in a follow up

caolanm (approved):

PR #7340, opened 2023-09-30
jsdialog: use minial size of a dialog needed

eszkadev:
Use 'position: absolute' in jsdialog-window and jsdialog-container so we don't stretch the dialog. It will use minimal needed size.

this is backport of: https://github.com/CollaboraOnline/online/commit/afbf052bb5d1f9b74389c068552c40878301aa43

fixes super long warning dialogs:

error


Ashod (approved):
Thanks @eszkadev.

PR #7341, opened 2023-09-30
Fix typeerror on key event in interaction handler dialog

eszkadev:
This fixes error:
Map.Keyboard.js:364 Uncaught TypeError: Cannot read properties of undefined (reading '_docType')
    at NewClass._globalKeyEvent (Map.Keyboard.js:364:3

When we open document which requires some interaction from user before load and we press any key - that error appears. It can be seen with macro warning or broken file warning.


gokaysatir (approved):
Thanks :)

PR #7343, opened 2023-09-30
cid#318949 Logically dead code

caolanm:
assuming checkForIdle is true, then

if (timeoutMicroS < 0)
{
    // in this case timeoutMicroS is < 0 and block (b)
    // won't be executed
}
else
{
    if (checkForIdle)
        timeoutMicroS = 0;

    ...
    do
    {
        ...

        if (...)
            // if we break on first iteration then when
            // checkForIdle is true, timeoutMicroS is 0 and
            // the block(b) isn't executed.
            break;

        timeoutMicroS = ...
        ++eventsSignalled;
        // after this line eventsSignalled is not 0 and block (b)
        // won't be executed
    }
    while (...)
}

// block (b)
if (...  checkForIdle && eventsSignalled == 0 && timeoutMicroS > 0 && ...) {
    ...
}

Change-Id: I20d2308325683dd51e67e130ad1223332755a12b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2354,9 +2354,6 @@ public:
 
         auto startTime = std::chrono::steady_clock::now();
 
-        // handle processtoidle waiting optimization
-        bool checkForIdle = ProcessToIdleDeadline >= startTime;
-
         if (timeoutMicroS < 0)
         {
             // Flush at most 1 + maxExtraEvents, or return when nothing left.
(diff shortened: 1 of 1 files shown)


caolanm:
@mmeeks this seems to be unused

mmeeks:
Lets abandon this - the code has a purpose, even if it's not much used/loved currently. The aim is to significantly speedup the remaining 'sleeps' in our cypress & unit tests :-)

PR #7344, opened 2023-10-02
remove unused used method

caolanm:
which is a little dubious anyway

Change-Id: If12bd7807ca5f2cff0ca05cad4948962b0cf6688

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/MessageQueue.hpp
+++ b/common/MessageQueue.hpp
@@ -79,12 +79,6 @@ public:
         clear_impl();
     }
 
-    /// Thread safe remove_if.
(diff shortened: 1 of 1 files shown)


plubius (approved):
Approved. I verified that this patch compiles and runs on iOS.

vmiklos:
Great, I was wondering how this std::remove() is supposed to work when its return value is not used; when the function is actually invoked. If it's not invoked, that explains. :-)

Issue #7345, opened 2023-10-02
Impress: "Set Background Image" button does not work in tabbed view


cosminh11:
### Describe the Bug

For presentation files, the "Set Background Image" button from the "Layout" tab in the tabbed view has no effect.

### Steps to Reproduce

1. In tabbed view, go to 'Layout'
2. Click on 'Set Background Image'

### Expected Behavior

A file selector appears and I can select an image file for the background.

### Actual Behavior

No file selector appears.

### Desktop

 - **OS:** Ubuntu 22.04
 - **Browser:** Firefox
 - **Version:** 118.0.1

### Additional Context

The corresponding menu item in the compact view works as expected.

PR #7346, opened 2023-10-02
Enable "Set Background Image" button in tabbed view

cosminh11:
Change-Id: I9e6c5f3070683e1949d66355d8e0cf832580b912

* Resolves: #7345
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarImpress.js
+++ b/browser/src/control/Control.NotebookbarImpress.js
@@ -1473,9 +1473,9 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 							{
 								'id': 'selectbackground',
 								'class': 'unoselectbackground',
-								'type': 'menubartoolitem',
+								'type': 'toolitem',
 								'text': _UNO('.uno:SelectBackground', 'presentation'),
-								'command': '',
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.NotebookbarImpress.js
```diff
@@ -1473,9 +1473,9 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 							{
 								'id': 'selectbackground',
```
gokaysatir, line 1474: This is not related to your work but since you touched these lines, can you also re-set the id from "selectbackground" to something like "layout-selectbackground" (tab-name + function)? The function you added is id-agnostic, i think that change shouldn't be a problem.

There is an id-conflict with the "selectbackground" input item which is created inside the HTML *.m4 file. If we can also solve that, that'd be good.

Needs to be tested again of course.

gokaysatir (approved):
Thanks for the fix :)

browser/src/control/Control.NotebookbarImpress.js
cosminh11: OK, I changed that and tried again and it still works.

PR #7355, opened 2023-10-03
return false early on 'null' origin

caolanm:
Change-Id: I9cdb26c3d2787beac6e2344d5f768e0fa6993bd8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Lets see the impact =)

PR #7357, opened 2023-10-03
Add CTRL + ALT + N shortcut for new documents

gokaysatir:
Added CTRL + ALT + N shortcut.
* This sends the NewDocument postmessage to integrator along with the document type.
* Integrator should handle the message.

Change-Id: I8458f382771211b245ad4cedf501b73b50afc1c7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -734,6 +734,13 @@ L.Map.Keyboard = L.Handler.extend({
 			e.preventDefault();
 		}
 
+		// CTRL + ALT + N for new document. This needs to be handled by the integrator.
+		if (this._isCtrlKey(e) && e.altKey && e.key.toUpperCase() === 'N') {
+			this._map.fire('postMessage', {msgId: 'UI_CreateFile', args: { DocumentType: this._map.getDocType() }});
+			e.preventDefault();
+			return true;
+		}
+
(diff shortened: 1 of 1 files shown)


eszkadev:
please fix commit message:
- title is sign off
- 2 change ids

gokaysatir:
> please fix commit message:
> 
> * title is sign off
> * 2 change ids

I changed them thanks :)

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -734,6 +734,13 @@ L.Map.Keyboard = L.Handler.extend({
 			e.preventDefault();
 		}
 
+		// CTRL + ALT + N for new document. This needs to be handled by the integrator.
+		if (this._isCtrlKey(e) && e.altKey && e.key.toUpperCase() === 'N') {
+			this._map.fire('postMessage', {msgId: 'NewDocument', args: { type: this._map.getDocType() }});
```
eszkadev, line 739: It doesn't use standard convention we have for other commands: COOL -> Host UI_xxxx (but there are already 2 exceptions... so not sure :) )

We have already command to create new document: UI_CreateFile
How this is different?
https://sdk.collaboraonline.com/docs/postmessage_api.html#id6

gokaysatir: Yep, thanks for the catch :) Replaced the command with the existing one.

PR #7358, opened 2023-10-03
browser: make reconnnect experience smoother

Rash419:
- don't invalidate the tiles
- don't reload the document if serverId changes
- writer: fix cursor jumps to top the document

Change-Id: Ie42850e77f9011cd1c820115919f10283407f216

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -541,15 +541,12 @@ app.definitions.Socket = L.Class.extend({
 		var command = this.parseServerCmd(textMsg);
 		if (textMsg.startsWith('coolserver ')) {
 			// This must be the first message, unless we reconnect.
-			var oldId = null;
 			var oldVersion = null;
 			var sameFile = true;
 			// Check if we are reconnecting.
 			if (this.WSDServer && this.WSDServer.Id) {
 				// Yes we are reconnecting.
-				// If server is restarted, we have to refresh the page.
 				// If our connection was lost and is ready again, we will not need to refresh the page.
-				oldId = this.WSDServer.Id;
 				oldVersion = this.WSDServer.Version;
 
 				window.app.console.assert(this._map.options.wopiSrc === window.wopiSrc,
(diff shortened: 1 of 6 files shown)


browser/src/core/Socket.js
```diff
@@ -1409,20 +1402,21 @@ app.definitions.Socket = L.Class.extend({
 		}
 		else if (this._reconnecting) {
 			// we are reconnecting ...
-			this._reconnecting = false;
-			this._map._docLayer._resetClientVisArea();
-			this._map._docLayer._requestNewTiles();
+			this._map._docLayer._refreshTilesInBackground();
 			this._map.fire('statusindicator', {statusType: 'reconnected'});
-			this._map._isNotebookbarLoadedOnCore = false;
```
eszkadev, line 1416: This breaks notebookbar after reconnect. It requires to be inicialized properly in the core so our widgets like "style previews" will work.

browser/src/core/Socket.js
```diff
@@ -1409,20 +1402,21 @@ app.definitions.Socket = L.Class.extend({
...
 			var uiMode = this._map.uiManager.getCurrentMode();
-			this._map.fire('changeuimode', {mode: uiMode, force: !window.migrating});
+			this._map.fire('changeuimode', {mode: uiMode, force: false});
```
eszkadev, line 1408: we call `this.map.sendUnoCommand('.uno:ToolbarMode?Mode:string=notebookbar_online.ui');`  only onAdd currently in notebookbar. it has to be done here to be sure notebookbar's widgets work

eszkadev:
for me it clears all the tiles (makes screen white) for a short time, so it could be more smooth :)

browser/src/core/Socket.js
Rash419: Done thanks

eszkadev (changes requested):
still notebookbar doesn't work after reconnect

browser/src/core/Socket.js
```diff
@@ -1393,20 +1386,23 @@ app.definitions.Socket = L.Class.extend({
 		}
 		else if (this._reconnecting) {
 			// we are reconnecting ...
-			this._reconnecting = false;
```
eszkadev, line 1396: if this `app.socket._reconnecting = false` is removed, where it is now done?

browser/src/core/Socket.js
```diff
@@ -1393,20 +1386,23 @@ app.definitions.Socket = L.Class.extend({
...
-			this._map._docLayer._resetClientVisArea();
-			this._map._docLayer._requestNewTiles();
+			this._map._docLayer._refreshTilesInBackground();
 			this._map.fire('statusindicator', {statusType: 'reconnected'});
 			this._map._isNotebookbarLoadedOnCore = false;
 			var uiMode = this._map.uiManager.getCurrentMode();
-			this._map.fire('changeuimode', {mode: uiMode, force: !window.migrating});
+			if (uiMode === 'notebookbar') {
+				this._map.sendUnoCommand('.uno:ToolbarMode?Mode:string=notebookbar_online.ui');
```
eszkadev, line 1394: this could be done via some helper function, I will try to do it later

eszkadev (approved):

eszkadev:
now notebookbar issue is fixed

PR #7359, opened 2023-10-03
Add CTRL + ALT + O as open document shortcut.

gokaysatir:
* This needs to be handled by the integrator.

Change-Id: If57bdd1f7db496766fbb7082ffeff855aba4e97c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/map/handler/Map.Keyboard.js
```diff
@@ -741,6 +741,13 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// CTRL + ALT + O for new document. This needs to be handled by the integrator.
+		if (this._isCtrlKey(e) && e.altKey && e.key.toUpperCase() === 'O') {
+			this._map.fire('postMessage', { msgId: 'OpenDocument' });
```
eszkadev, line 746: It doesn't use standard convention we have for other commands:    COOL -> Host    UI_xxxx      (but there are already 2 exceptions... so not sure :) )

We have already command to create new document: `UI_CreateFile`
How this is different?
https://sdk.collaboraonline.com/docs/postmessage_api.html#id6

gokaysatir: Changed the command name to "UI_OpenDocument" in order to follow the rules, thanks :)

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -741,6 +741,13 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// CTRL + ALT + O for new document. This needs to be handled by the integrator.
```
eszkadev, line 744: copy paste from ctrl + alt + n? 
not for new document, but rather for opening other document?

eszkadev:
please prepare PR for SDK with new command

PR #7364, opened 2023-10-04
Private/caolanm/fixquarantine

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -298,6 +298,18 @@ static void cleanupChildren()
         const auto it = childJails.find(exitedChildPid);
         if (it != childJails.end())
         {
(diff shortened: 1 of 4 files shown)


caolanm:
Three problems as I see it:
  - if not currently uploading/saving then the quarantine was asked to save the original "_filename" which isn't the path on disk
  - forkLibreOfficeKit is run regularly and can remove a jail before the docbroker gets a chance to quarantine any document from it. This probably doesn't happen in a devel debug version with the sleep after crash to allow a debugger to be attached.
  - if a crash happens during load, before the user modifies it, then that won't be considered as a "data loss", so no quarantine happened for that case

mmeeks (approved):
Love it - thanks for the fixups Caolan ! =)

PR #7365, opened 2023-10-04
Tabbed: Calc: Scroll indicator is 3 pixels misaligned

pedropintosilva:
The scroll indicator set with an absolute position ends up being
positioned in the 0 y coordinate. Since the wrapper has a top padding
of 3 pixel it means the scroll indicator is not 3 pixel off.

Also, and since these measurements are set in the CSS (the height of
the wrapper and the padding of the other wrapper), better to just move
everything to the CSS and avoid unnecessary inline styles.

Change-Id: Icc1e51b2590ee27b318f9e4c3d550183b08cb49d


Darshan-upadhyay1110 (approved):
Thanks :)

PR #7367, opened 2023-10-04
Allow http Range header for media files

Minion3665:
The range header allows a client to specify that they would like only part of a file, e.g. only the first 1000 bytes of a video. When playing video on Apple devices (Macs, iPads, iPhones etc.) this is required to avoid Safari rejecting the video as broken. This is the first part of a fix to embedded videos on Safari.

This pull request does not implement specifying multiple ranges at once (e.g. bytes=0-9,-100 to get the first 10 and last 100 bytes).

This feature is necessary to play video on Apple devices, but it is not sufficient. In particular, I believe some further client-side fixes will be needed to make the video play properly.

Change-Id: Id89a06d374b7d0efbf2b3184d8618df61684dcb6

* Target version: master 

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4594,7 +4594,9 @@ private:
         {
             // Do things in the right thread.
             LOG_TRC_S("Move media request " << tag << " to docbroker thread");
-            docBroker->handleMediaRequest(socket, tag);
+
+            std::string range = request.get("Range", "none");
+            docBroker->handleMediaRequest(range, socket, tag);
         }
     }
 
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1555,6 +1555,10 @@ public:
         , _size(0)
         , _fd(-1)
         , _connected(false)
+        , _start(0)
+        , _end(-1)
+        , _startIsSuffix(false)
+        , _statusCode(http::StatusCode::OK)
     {
     }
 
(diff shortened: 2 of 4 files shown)


mmeeks (changes requested):
Headers etc. look good - but somehow I'd expect to have a seek in the asyncUpload that would not load the whole video into memory etc. and to honor _start and _end not just in the generated headers: or did I miss something ? =)

mmeeks:
Overall great catch and nice change though - but this code is @Ashod baby so prolly he has more requests =)

Minion3665:
> Headers etc. look good - but somehow I'd expect to have a seek in the asyncUpload that would not load the whole video into memory etc. and to honor _start and _end not just in the generated headers: or did I miss something ? =)

yep, you're absolutely right, thanks for catching that!

mmeeks (approved):
Looks great to me - but still, I'd like @Ashod 's take =)

wsd/COOLWSD.cpp
```diff
@@ -4594,7 +4594,7 @@ class ClientRequestDispatcher final : public SimpleSocketHandler
         {
             // Do things in the right thread.
             LOG_TRC_S("Move media request " << tag << " to docbroker thread");
-            docBroker->handleMediaRequest(socket, tag);
+            docBroker->handleMediaRequest(request, socket, tag);
```
Ashod, line 4597: Let's pass in the `range`. No reason to depend on `Poco::Net::HTTPRequest`  to read a header.

mmeeks:
So - we should get this merged for the release if we can to get testing - lets get it in now & improve on it in a follow-up if necessary :_)

Minion3665:
> @mmeeks 
> So - we should get this merged for the release ... lets get it in now

Alright, I've deleted the half-made unit test file so I can work on that in a followup. Would you like to wait for @Ashod to take another look at it after I addressed the comment on passing the header in directly or should we merge when green?

mmeeks:
Lets merge it as long as the unit test comes next :-) thanks Skyler.

PR #7371, opened 2023-10-05
exclude debug/dev files from production packages

timar:
Change-Id: Ic9b49a401b63af65f4e53e3eb2584b8258575c2a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Not beautiful - can we not use Makefile wildcards to get the files instead if we need to do something like this ? but anyhow - lets get it in for now.

PR #7375, opened 2023-10-06
browser: show error msg when socket connection is closed unexpectedly

Rash419:
Change-Id: Ib654d0e81d0796ba379e9482ad90e6cb8af5901d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1522,8 +1522,10 @@ app.definitions.Socket = L.Class.extend({
 		window.app.console.debug('_onSocketClose:');
 		if (!this._map._docLoadedOnce) {
 			var postMessageObj = {
+				errorType: 'websocketconnectionfailed',
 				success: false,
-				errorMsg: errorMessages.websocketconnectionfailed
+				errorMsg: errorMessages.websocketconnectionfailed,
+				result: '',
 			};
 			this._map.fire('postMessage', {msgId: 'Action_Load_Resp', args: postMessageObj});
 			this._map.fire('error', {msg: errorMessages.websocketconnectionfailed, cmd: 'socket', kind: 'closed', id: 4});
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1520,6 +1520,14 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		window.app.console.debug('_onSocketClose:');
+		if (!this._map._docLoadedOnce) {
+			var postMessageObj = {
+				success: false,
+				errorMsg: errorMessages.websocketconnectionfailed
+			};
+			this._map.fire('postMessage', {msgId: 'Action_Load_Resp', args: postMessageObj});
+			this._map.fire('error', {msg: errorMessages.websocketconnectionfailed, cmd: 'socket', kind: 'closed', id: 4});
+		}
 		if (this.ReconnectCount > 0)
 			return;
 
--- a/browser/src/errormessages.js
+++ b/browser/src/errormessages.js
@@ -27,6 +27,7 @@ errorMessages.docloadtimeout = _('Failed to load the document. This document is 
 errorMessages.docunloadingretry = _('Cleaning up the document from the last session.');
 errorMessages.docunloadinggiveup = _('We are in the process of cleaning up this document from the last session, please try again later.');
 errorMessages.clusterconfiguration = _('Your %productName cluster appear to be mis-configured or scaling rapidly - please contact your system administrator. Continuing with editing may result in multiple users not seeing each other, conflicts in the document storage and/or copy/paste problems. Expected serverId %0 for routeToken %1 but connected to serverId %2');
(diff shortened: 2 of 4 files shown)


mmeeks (changes requested):
That looks fine - but I expect that Nextcloud puts an overlay over the top of us - so we can't show our own error messages at all - will we get a decent helpful diagnostic error on top of that overlay ? can you test that. Also - we should prolly have a col.la/  link that goes to a forum post related to this topic - or add a hint as to what is missing in the reverse proxy configuration - or better a  perma-link to our SDK that shows how to configure this - thanks!

Rash419:
> That looks fine - but I expect that Nextcloud puts an overlay over the top of us - so we can't show our own error messages at all - will we get a decent helpful diagnostic error on top of that overlay ? 
`this._map.fire('postMessage', {msgId: 'Action_Load_Resp', args: postMessageObj});`
this postMessage can show overlay. let me add screenshot here
image

>  Also - we should prolly have a col.la/ link that goes to a forum post related to this topic - or add a hint as to what is missing in the reverse proxy configuration - or better a perma-link to our SDK that shows how to configure this 

Sure, added

pedropintosilva (approved):
Needs another +1

browser/src/map/handler/Map.WOPI.js
```diff
@@ -335,6 +335,11 @@ L.Map.WOPI = L.Handler.extend({
 				this._map._disableDefaultAction[msg.Values.action] = msg.Values.disable;
 			}
 		}
+		else if (msg.MessageId === 'Error_Messages') {
+			if (msg.Values.websocketconnectionfailed) {
+				errorMessages.websocketconnectionfailed = _(msg.Values.websocketconnectionfailed);
```
eszkadev, line 340: if any custom message will come here, we will not be able to translate it - because it will not be found  in our translations

pedropintosilva: @eszkadev right, I mean if the string comes from outside, sure we cannot guarantee it's translated. But probably that should already come translated from the integrator. (Don't know if it's possible but I would expect integrators to set the value of that message differently according to user's UI )

browser/src/errormessages.js
```diff
@@ -27,6 +27,7 @@ errorMessages.docloadtimeout = _('Failed to load the document. This document is
 errorMessages.docunloadingretry = _('Cleaning up the document from the last session.');
 errorMessages.docunloadinggiveup = _('We are in the process of cleaning up this document from the last session, please try again later.');
 errorMessages.clusterconfiguration = _('Your %productName cluster appear to be mis-configured or scaling rapidly - please contact your system administrator. Continuing with editing may result in multiple users not seeing each other, conflicts in the document storage and/or copy/paste problems. Expected serverId %0 for routeToken %1 but connected to serverId %2');
+errorMessages.websocketconnectionfailed = _('Failed to establish socket connection or socket connection closed unexpectedly. The reverse proxy might be misconfigured, please contact the administrator. For more info on proxy conifguration please checkout https://sdk.collaboraonline.com/docs/installation/Proxy_settings.html');
```
pedropintosilva, line 30: configuration not conifguration

browser/src/core/Socket.js
```diff
@@ -1516,6 +1516,14 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		window.app.console.debug('_onSocketClose:');
+		if (!this._map._docLoadedOnce) {
+			var postMessageObj = {
+				success: false,
+				errorMsg: errorMessages.websocketconnectionfailed
```
juliushaertl, line 1522: Any chance we could have a static identifier on the error type? With the current way we can still not render the link properly within the richdocuments error screen as.

I imagined something like this where any possible error type is also documented along the Action_Load_Resp in the sdk docs.

```suggestion
				errorMsg: errorMessages.websocketconnectionfailed,
				errorType: 'websocketconnectionfailed'
```

juliushaertl: With this we could actually decide on the error message in richdocuments without the need to pass around post messages (while i think this still makes sense for other cases to customize errors displayed within Collabora).

Just knowing the type of error allows to decide on different messages for admins or regular users and also be more flexible to show the url for example as a call to action button instead of the raw link.

browser/src/core/Socket.js
```diff
@@ -1520,6 +1520,15 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		window.app.console.debug('_onSocketClose:');
+		if (!this._map._docLoadedOnce) {
+			var postMessageObj = {
+				errorType: 'websocketconnectionfailed',
+				success: false,
+				errorMsg: errorMessages.websocketconnectionfailed
```
eszkadev, line 1527: missing `result` property:
```
success: <boolean>
result:  <string>
errorMsg: <string>
```
https://sdk.collaboraonline.com/docs/postmessage_api.html

eszkadev: apart of that one thing looks good :)

eszkadev (approved):
code looks good

Issue #7354, opened 2023-10-03
Calc: sheet tab: triggering context menu switches tabs


pedropintosilva:
### Describe the Bug

https://github.com/CollaboraOnline/online/assets/65948705/2b181f55-bd1e-4777-8d76-1212650e39ea

Right click (or long tab when using mobile or tablet) on a sheet tab triggers the contact menu **but** it also automatically switches the active tab. Result, user is now in a completely different sheet tab when the user just wanted to access the context menu of said sheet tab.

### Steps to Reproduce

1. Open a file in edit mode 
2. Right click (or long tap on mobile or tablet)
3. Context menu appears and at the same time user is now thrown to another sheet tab

### Expected Behavior

Do not switch sheet tabs when context menu is triggered. 

### Version

Master a8e97f6220790611f0626649fc3e69578ae9c470

pedropintosilva:
@bayramcicek can you take this one?

bayramcicek:
yes, working on it...

PR #7377, opened 2023-10-06
Calc: Don't switch tabs when context menu is triggered

bayramcicek:
- Trying to open context menu from a sheet tab, also changes the active tab. This occurs when right click on a sheet tab (or long tap on mobile or tablet)

- moveLeft and moveRight of inactive tabs are implemented.

- added _moveSheetLR and _moveOrCopySheet functions.

Change-Id: I5b25a8cb363b8e04fd6dcec1f500e7ff2982b678

* Resolves: #7354 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -103,7 +103,7 @@ L.Control.Tabs = L.Control.extend({
 			// no blacklisting available for this context menu so only add when needed
 			this._menuItem['.uno:Move'] = {
 				name: _UNO('.uno:Move', 'spreadsheet', true),
-				callback: function() {this._map.sendUnoCommand('.uno:Move');}.bind(this),
+				callback: (this._moveOrCopySheet).bind(this),
 				visible: areTabsMultiple
 			};
 
@@ -204,7 +204,6 @@ L.Control.Tabs = L.Control.extend({
 							.on('press', function (j) {
 								return function(e) {
 									this._tabForContextMenu = j;
-									this._setPart(e);
 									if (!this._map.isReadOnlyMode()) {
 										if (window.mode.isMobile()) {
 											window.contextMenuWizard = true;
@@ -226,9 +225,8 @@ L.Control.Tabs = L.Control.extend({
 							};
 						}(i).bind(this));
 						L.DomEvent.on(tab, 'contextmenu', function(j) {
-							return function(e) {
+							return function() {
 								this._tabForContextMenu = j;
-								this._setPart(e);
 							};
 						}(i).bind(this));
 					}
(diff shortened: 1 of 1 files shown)


eszkadev (changes requested):
it was introduced previously because if we right-clicked on non-selected sheet to open context menu - actions were applied to the selected sheet. It still happens for some actions:
"move sheet left", "move sheet right", "move or copy sheet"

steps to reproduce:
- have 3 sheets, 1st is scurrently selected
- right click on 3rd sheet and do "move sheet left"
result: sheet 1 is moved
expected: sheet 3 is moved

bayramcicek:
> it was introduced previously because if we right-clicked on non-selected sheet to open context menu - actions were applied to the selected sheet. It still happens for some actions: "move sheet left", "move sheet right", "move or copy sheet"
> 
> steps to reproduce:
> 
> * have 3 sheets, 1st is scurrently selected
> * right click on 3rd sheet and do "move sheet left"
>   result: sheet 1 is moved
>   expected: sheet 3 is moved

@eszkadev : I see that https://github.com/CollaboraOnline/online/pull/7377/files#diff-a3b5fb6d0c66140a66a5e27fed984875e9c8f1194a8dfb152a5716b17fc2ea6dL282 `_moveSheet` only moves the `selected sheet` to new index with `sendUnoCommand('.uno:Move?Copy:bool=false&UseCurrentDocument:bool=true&Index=' + newIndex);`  

There should be a way to move inactive sheets. _I'm not sure but_ it seems there should be some hack on the `core` side to make this happen, I wonder what do you think about that?

Thanks for the review.

bayramcicek:
158078: (WIP) sc: make inactive sheets movable/copyable | https://gerrit.libreoffice.org/c/core/+/158078

eszkadev:
please rebase so Unit & Cypers will pass

bayramcicek:
> fail:
> -  calc/sheet_operation_spec.js 
>     - Sheet Operations. Move sheet left/right)
> - most of the writer tests failed

- Since we changed the behavior of the sheet tabs and upstream patch is not merged, I think, the fail in the Calc is expected(?).
- and test for the Sheet Operations (sheet_operation_spec.js) should be rewritten/updated since it assumes that right click on a sheet tab will change tab page.

@eszkadev : what should be the next step. ?

> I think:
> - update sheet_operation_spec.js test to adapt the new sheet tab behavior.
> - merge the upstream patch (https://gerrit.libreoffice.org/c/core/+/158078)
> - rebase the online to see if CI passes 

should I update the sheet_operation_spec.js test?

browser/src/control/Control.Tabs.js
```diff
@@ -283,17 +281,27 @@ L.Control.Tabs = L.Control.extend({
...
+
 	_moveSheetLeft: function () {
-		var targetIndex = this._map._docLayer._partNames.indexOf(this._map._docLayer._partNames[this._map._docLayer._selectedPart]);
 		//core handles sheet with 1 base indexing
 		// 0 index means last sheet
-		if (targetIndex <= 0) return;
-		this._moveSheet(targetIndex);
+		var contextMenuTab = this._tabForContextMenu;
+		if (contextMenuTab <= 0) return;
+		this._moveSheetLR(contextMenuTab, contextMenuTab);
```
eszkadev, line 299: I don't understand how it works.
Current value and target are the same?

browser/src/control/Control.Tabs.js
```diff
@@ -283,17 +281,27 @@ L.Control.Tabs = L.Control.extend({
...
 	},
 
 	_moveSheetRight: function () {
-		var targetIndex = this._map._docLayer._partNames.indexOf(this._map._docLayer._partNames[this._map._docLayer._selectedPart]) + 3;
-		this._moveSheet(targetIndex);
+		var contextMenuTab = this._tabForContextMenu;
+		this._moveSheetLR(contextMenuTab, contextMenuTab + 3);
```
eszkadev, line 304: Why +3 ?

eszkadev:
> should I update the sheet_operation_spec.js test?

If we will merge core part first, then this PR will pass right?
So we will wait for core.

browser/src/control/Control.Tabs.js
bayramcicek: Actually it should be: `this._moveSheetLR(contextMenuTab, contextMenuTab - 1);`
But, decreasing the `contextMenuTab` by `1` is made in the core side here: https://gerrit.libreoffice.org/c/core/+/158078/5/sc/source/ui/view/tabvwshf.cxx#514
So, I did not used `contextMenuTab - 1` in the online side.

Seems like this is confusing. I can do the decreasing here, on the online side, if needed?

eszkadev: Yes, please do that so it will be obvious how to use the uno command

browser/src/control/Control.Tabs.js
bayramcicek: The old code, which is `var targetIndex = this._map._docLayer._partNames.indexOf(this._map._docLayer._partNames[this._map._docLayer._selectedPart]) + 3;` also use `+3`, and when I was trying to understand why there is `+3`; I figured out that:
- for moveRight, core side acts like 1-based index. so we need to add `+1`
- since we moveRight we need to add `+1`
- and at the core side, there is a `-1` for this value, so we add `+1` again.

I will send a new patchset that will remove the `-1` in the core, so we will only use `+2` for the moveRight.

bayramcicek:
FYI: I encountered an issue with selecting sheet tabs while testing the patch (it selects the previous/next tab, then selects the correct tab). I'm working on that too.

browser/src/control/Control.Tabs.js
eszkadev: So rather adjust `ContextMenuIndex` to be also compatible with `Index` param in the core...

bayramcicek: OK. I will do so.

bayramcicek:
**upstream**: 158078: sc: make inactive sheets movable and copyable | https://gerrit.libreoffice.org/c/core/+/158078

pedropintosilva (changes requested):
Pull gerrit change ... sorry
----
Thanks @bayramcicek I have tested all the actions in the context menu. There don't work as expected:

* Move sheet left
* Move sheet right
* Move or copy sheet...

1. If you have: [sheet1] **[sheet2]** [sheet3] [sheet4] [sheet5] `current sheet in bold`
2. Right Click in the [sheet4]
3. Choose any of the actions listed above
4. Notice: the actions affect the current sheet instead of the one that was clicked on
5.

pedropintosilva:
Would like to have +2 from @gokaysatir

gokaysatir (approved):
Thanks @bayramcicek :)

gokaysatir:
I tested and found no issues. I think we can merge this.

PR #7379, opened 2023-10-06
cid#323357 silence String not null terminated

caolanm:
it's a good catch, but inotify man page says it can't happen

Change-Id: I350f8fad2aa8308563ccb4f9f3ec9b7850a72fa7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665 (approved):
looks good with respect to the manpage and doesn't cause any issues in my testing, looks good to me :)

PR #7385, opened 2023-10-08
Private/caolanm/coverity22

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.hpp
+++ b/common/Log.hpp
@@ -488,8 +488,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
 #define LOG_ASSERT_INTERNAL(condition, message, LOG)                                               \
     do                                                                                             \
     {                                                                                              \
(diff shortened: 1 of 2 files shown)


common/Log.hpp
```diff
@@ -488,8 +488,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
 #define LOG_ASSERT_INTERNAL(condition, message, LOG)                                               \
     do                                                                                             \
     {                                                                                              \
-        auto&& cond##__LINE__ = !!(condition);                                                     \
-        if (!cond##__LINE__)                                                                       \
+        if (!(condition))                                                                          \
```
vmiklos, line 491: I assume there was some original context where `cond##__LINE__` was checked multiple times, and we wanted to evaluate `condition` only once, but seeing that we only check it once currently, this is fine.

vmiklos (approved):

PR #7387, opened 2023-10-09
restore TileCacheTests::testTileProcessed

caolanm:
with the original 25 expected results from 25 requests

from cid#318925 Structurally dead code

Change-Id: I236876fadd5b5c82ecdbcc52a13fc58e8513aa12

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/TileCacheTests.cpp
+++ b/test/TileCacheTests.cpp
@@ -1513,14 +1513,6 @@ void TileCacheTests::testTileWireIDHandling()
 
 void TileCacheTests::testTileProcessed()
 {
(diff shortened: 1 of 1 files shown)


caolanm:
this got disabled back in

commit 98ae6ac269b96f6a2b457b21f85dee642232922f
Date:   Tue Mar 14 21:17:13 2023 +0100

    disable fragile TileCacheTests::testTileProcessed()

it seems ok to me locally and here in ci now. Worth restoring it ?

vmiklos (approved):
Indeed passes at least against both core.git master and co-23.05, so why not.

PR #7388, opened 2023-10-09
docker: cleanup

Rash419:
- remove start-collabora-online.pl from-source

Change-Id: I4cc477733e8bd6a2c736e16bcb2d75aa295e9895


timar (approved):
This is fine, but `from-source` is out of date in many respect. 
I don't know in which direction should we improve it. Release packages are not dependent on system packages any more, they link libpng, Poco, OpenSSL and libzstd statically. But of course community members can build as they like, and use system packages.

PR #7389, opened 2023-10-09
private/pedro/remove unused localized string n improve pencil string

pedropintosilva:
- Fix Pencil button's title and aria-label attributes
- Collapse tabs: Remove extra unused localized string


gokaysatir (approved):
Checked, thanks :)

PR #7390, opened 2023-10-09
delayed messages can be applied when _map._docLayer becomes available

caolanm:
without a need for an additional timer to poll when that is possible.

Change-Id: Ia019d8d8516f8c9156f3e96be262138a16343983

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -348,8 +348,8 @@ app.definitions.Socket = L.Class.extend({
 					}
 					catch (e)
 					{
-						// unpleasant - but stops this one problem
-						// event stopping an unknown number of others.
+						// unpleasant - but stops this one problem event
+						// stopping an unknown number of others.
 						window.app.console.error('Exception ' + e + ' emitting event ' + evt.data);
 					}
 					finally {
@@ -1198,7 +1198,7 @@ app.definitions.Socket = L.Class.extend({
 			this._onHyperlinkClickedMsg(textMsg);
 		}
 
-		if (!this._isReady() || !this._map._docLayer || this._delayedMessages.length || this._handlingDelayedMessages) {
+		if (!this._map._docLayer || this._handlingDelayedMessages) {
 			this._delayMessage(textMsg);
 		} else {
 			this._map._docLayer._onMessage(textMsg, e.image);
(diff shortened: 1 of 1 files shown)


browser/src/core/Socket.js
```diff
@@ -1198,7 +1197,7 @@ app.definitions.Socket = L.Class.extend({
 			this._onHyperlinkClickedMsg(textMsg);
 		}
 
-		if (!this._isReady() || !this._map._docLayer || this._delayedMessages.length || this._handlingDelayedMessages) {
+		if (!this._map._docLayer) {
```
mmeeks, line 1200: Are we sure not using _isReady is a good idea ? =) if we don't want it - then we should prolly remove the function - since this is it's whole raison d'etre I think =)

browser/src/core/Socket.js
```diff
@@ -31,7 +31,6 @@ app.definitions.Socket = L.Class.extend({
 		this._map = map;
 		this._msgQueue = [];
 		this._delayedMessages = [];
-		this._handlingDelayedMessages = false;
```
mmeeks, line 34: I assume this is a re-entrancy guard to ensure that if more messages arrive while we are processing others (not sure if/how that can happen), that they are appended to the delay queue rather than being processed out of order.

browser/src/core/Socket.js
```diff
@@ -1318,42 +1317,24 @@ app.definitions.Socket = L.Class.extend({
...
-		}
+	_handleDelayedMessages: function(docLayer) {
 		var messages = [];
 		for (var i = 0; i < this._delayedMessages.length; ++i) {
 			var message = this._delayedMessages[i];
-			if (message)
-				messages.push(message.msg);
+			messages.push(message.msg);
 		}
 		this._delayedMessages = [];
```
mmeeks, line 1328: On that basis, I guess popping one message and processing it might be a better plan (?) - but then we would want the || this._delayedMessages.length on delay / queueing incoming messages (?)

browser/src/core/Socket.js
```diff
@@ -1403,6 +1384,9 @@ app.definitions.Socket = L.Class.extend({
 				});
 			}
 
+			if (docLayer !== null) {
+				this._handleDelayedMessages(docLayer);
```
mmeeks, line 1388: Nice to get rid of the timer =)

mmeeks (commented):
A few thoughts/comments =)

browser/src/core/Socket.js
caolanm: The this._map._docLayer is set in _onStatusMsg which is also using _isReady() (admittedly with another of these "try again later" timers) so _isReady() has to be true if this._map._docLayer is set.

browser/src/core/Socket.js
caolanm: yeah, but the question is if that is at all possible?

mmeeks: Fair cop - if you think not; lets simplify it; but mis-ordering of events in races/corner-cases adds -such- difficult to find/fix/reproduce bugs - that I'd love some vestigal warning / assert if its possible =) some defensiveness may be useful.

browser/src/core/Socket.js
mmeeks: I guess things like:

commit a93d8f1e1cadd6a80248f8e6cb9a8eae80f0b657
Author: Ashod Nakashian <ashod.nakashian@collabora.co.uk>
Date:   Wed Jul 3 00:29:43 2019 -0400

Suggest 'isReady' is useful in some cases. Bear in mind that we can start executing JS without loading all of it (I think) in the browser, and its order of initialization is not cleanly defined.

Then again the _isReady in _onStatusMsg seems a bit frightening - but the browser really can take quite some time to get its act together, and its not entirely predictable when that happens (I expect).

Not sure that overall this is a good idea. I forget what the goal was here - a  better understanding? removing a long timeout (100ms is a good one to loose of course =), or ? ...

caolanm: The original thing that invited scrutiny was  https://github.com/CollaboraOnline/online/issues/7164 and during that a question was, why only delay certain messages if they can't be applied yet, so we now delay all messages until they can be applied and not lose any.

But a follow up was to see if we really need another long timeout to apply them. Which is this proposed change which takes advantage that this._map._docLayer gets set when _isReady() is true to drop that timer.

mmeeks: Ah! ok - so I missed that _onStatusMsg is also doing a similar timeout spin (and we should prolly shorten the timeout to 10ms or so there) to check that we are really ready - and when done so it sets the ... ok :-) I get it.

And we're confident that there is always a _onStatusMsg coming through - since this is IIRC one of the very earliest messages coming in ...

Good :-) nice cleanup - sorry to miss that bit!.

mmeeks (approved):
Looks great to me =) thanks Caolan - good to see less code here ...

PR #7391, opened 2023-10-09
Change the local storage setting name 'CompactMode' to 'compactMode'

timar:
We want to re-force the choice, i.e. people who mistakenly set their UI to compact mode, now will get tabbed mode and can decide, if they like it better or not.

Change-Id: Ic5810321abf2cbc3996478816fe153d848c66c71

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -67,7 +67,7 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	shouldUseNotebookbarMode: function() {
-		var forceCompact = this.getSavedStateOrDefault('CompactMode', null);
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
Nice commit; Of course - keeping the StdlyCaps consistent but choosing another name might be less confusing in the long run - but this is fine.

pedropintosilva (approved):

pedropintosilva (changes requested):
I see @mmeeks .

Maybe best to use something like `CompactView` since anyhow we refer to that both in the code comments and in the UI as a view. This way we would keep the alternating caps format.

mmeeks (approved):
Lets get it in quick.

PR #7395, opened 2023-10-10
Enable drawing even if the first cell cursor message is empty.

gokaysatir:
Reason:
Sometimes cell address is at the end of a merged cells block. That kind of cell address is invalid. Example:
Merged cells: A1-A5.
When example merged block is selected, cell address should be A1. If it is A5, cell cursor is not drawn.
If cell cursor is not drawn "EMPTY" message is sent from core side. And we cannot enable drawing in this case.
So we enable drawing even if the cell cursor message is "EMPTY".

Change-Id: I456bc757725c9bfaede20068cf165310712488e7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks sensible

Issue #7084, opened 2023-08-15
Ctrl+f does not work on the Ukrainian keyboard layout.


AIlkiv:
### Describe the Bug

Ctrl+f (Ctrl+а) does not work on the Ukrainian keyboard layout.
I have checked other keyboard shortcuts, and they work correctly on the Ukrainian keyboard layout.
There is no such problem on English layouts.

### Steps to Reproduce

1. Open the document/table in the Collabora Office browser.
2. Switch the keyboard layout to Ukrainian.
3. Press Ctrl+F.
4. The search bar does not appear.
5. I switch to the English layout.
6. When I press Ctrl+F, the search bar appears.

### Expected Behavior

A search bar should appear.

### Desktop

 - **OS:**  Windows 11
 - **Browser:** Chrome 115.0.5790.171
 - **Collabora Office Version:**  23.05.2.2

oleua:
Still not working in Collabora 25.04.8.1 :(

COOLWSD version: 25.04.8.1(git hash: 278cb3481b (E))
LOKit version: Collabora Office 25.04.8.1(git hash: 489cb5d73b)
Served by: unknown
Server ID: ****
WOPI host: nextcloud 31.0.9.1 - richdocuments 8.7.5.3

PR #7396, opened 2023-10-10
Shortcuts for different keyboard layouts

AIlkiv:
* Resolves: #7084
* Target version: master 

### Summary

In certain sections of the code, the shortcuts was tied to the e.key value. However, this approach doesn't function correctly on keyboard layouts that significantly differ from English. Therefore, it has been rewritten to utilize 'e.keyCode,' which is not influenced by the keyboard layout.

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -721,35 +721,34 @@ L.Map.Keyboard = L.Handler.extend({
 		if (e.keyCode == this.keyCodes.CTRL)
 			return true;
 
-		if (e.type !== 'keydown' && e.key !== 'c' && e.key !== 'v' && e.key !== 'x' &&
+		if (e.type !== 'keydown' && e.keyCode !== this.keyCodes.C[DEFAULT] && e.keyCode !== this.keyCodes.V[DEFAULT] && e.keyCode !== this.keyCodes.X[DEFAULT] &&
 		/* Safari */ e.keyCode !== this.keyCodes.C[MAC] && e.keyCode !== this.keyCodes.V[MAC] && e.keyCode !== this.keyCodes.X[MAC]) {
 			e.preventDefault();
 			return true;
 		}
 
 		if (e.keyCode !== this.keyCodes.C[DEFAULT] && e.keyCode !== this.keyCodes.V[DEFAULT] && e.keyCode !== this.keyCodes.X[DEFAULT] &&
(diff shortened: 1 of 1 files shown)


AIlkiv:
@gokaysatir Thank you for your prompt response.

In the shortcuts help, this key combination is described as follows:

Screenshot 2023-10-10 114807

I have checked, and the keys generate the following codes:

Screenshot 2023-10-10 115443

Therefore, it is necessary to:
1. Change the shortcuts to 189.
2. Correct the shortcut help so that the key combination is listed as Ctrl + Shift + "-".

Did I understand you correctly?

gokaysatir:
Hello @AIlkiv and thank you again. Let me pull @pedropintosilva in :) I will consult also to him about this keyboard layout differences. You  have a valid point, your keyboard shows a direct key for question mark symbol.

gokaysatir:
Maybe we can leave that '?' part as is for now.

pedropintosilva (commented):
Everything seems fine, ctrl + shift + / works for me using American keyboard but I think it doesn't work for @gokaysatir

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -971,13 +970,13 @@ L.Map.Keyboard = L.Handler.extend({
 
 	readOnlyAllowedShortcuts: function(e) {
 		// Open keyboard shortcuts help page
-		if (this._isCtrlKey(e) && e.shiftKey && e.key === '?')
+		if (this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.FORWARDSLASH) // ?
```
pedropintosilva, line 973: This seems to not work with Turkish keyboard layout. Please revert the following occurrences of this line
```javascript
if (this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.FORWARDSLASH) { // ?
``` 
and then we can merge this PR :)

Nevertheless, I do agree, we could have a better shortcut for the keyboard shortcuts dialog. We need to think and test this further with multiple different keyboard layouts

Thank you so much for not only testing it deeply but also helping out !

AIlkiv: Thank you for investigating the matter. I concur that it's more accurate to retain the link to the char '?'. I have made the necessary changes to the code.

gokaysatir (approved):
Thanks for the improvements @AIlkiv :)

pedropintosilva (approved):

browser/src/map/handler/Map.Keyboard.js
pedropintosilva: Thank you!

PR #7397, opened 2023-10-10
cid#318830 help coverity see assert

caolanm:
and

cid#318839 Explicit null dereferenced
cid#318838 Explicit null dereferenced

Change-Id: I34685ea04722aeccc21910283fa4819dbf45c012

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/lokassert.hpp
+++ b/test/lokassert.hpp
@@ -78,7 +78,7 @@ std::string inline lokFormatAssertEq(const char* expected_name, const std::strin
     return oss.str();
 }
 
-#ifdef LOK_ABORT_ON_ASSERTION
+#if defined(LOK_ABORT_ON_ASSERTION) || defined(__COVERITY__)
 #define LOK_ASSERT_IMPL(X) assert(X);
 #else
 #define LOK_ASSERT_IMPL(X)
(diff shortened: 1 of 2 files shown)


test/lokassert.hpp
```diff
@@ -115,16 +115,15 @@ inline constexpr bool failed() { return false; }
...
+            if (!(condition))                                                                      \
             {                                                                                      \
                 std::ostringstream oss##__LINE__;                                                  \
                 oss##__LINE__ << message;                                                          \
                 const auto msg##__LINE__ = oss##__LINE__.str();                                    \
                 TST_LOG("ERROR: Assertion failure: "                                               \
                         << (msg##__LINE__.empty() ? "" : msg##__LINE__ + ". ")                     \
                         << "Condition: " << (#condition));                                         \
-                LOK_ASSERT_IMPL(cond##__LINE__);                                                   \
+                LOK_ASSERT_IMPL(!#condition); /* NOLINT(misc-static-assert) */                     \
```
vmiklos, line 126: Aha, we know that the condition is false here, so this is assert(false) with a better output.

And then clang-tidy would lint about this, but we silence it, similar to how common/Log.hpp does it.

OK, makes sense now.

vmiklos (approved):

PR #7410, opened 2023-10-11
different 'parts' in requestedTiles on switching calc tabs

caolanm:
can result in still unfulfilled requests for the first tab in requestedTiles when the new requests are compared with those for duplicates

Change-Id: Id8d40f0a7e7f9d32fed67a415756284d7f6a53d4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3129,6 +3129,16 @@ void DocumentBroker::handleTileRequest(const StringVector &tokens, bool forceKey
     _debugRenderedTileCount++;
 }
 
+void DocumentBroker::sendTileCombine(const TileCombined& newTileCombined)
+{
+    assert(!newTileCombined.hasDuplicates());
+
+    // Forward to child to render.
+    const std::string req = newTileCombined.serialize("tilecombine");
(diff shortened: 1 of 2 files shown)


caolanm:
@mmeeks So, it looks to me that you can click on tabs fast enough that the request for tiles from the new tab can arrive before the tiles from the old tab have been processed and the assert fires for unexpected mismatch of old and new tab "parts" when considering if the old requests can be dropped in favor of the new one. As it stands we presumably are inadvertently dropping some requests from the old tab we switched from.

The patch here isn't enough on its own so address the assert. Done this way there's an occasional later assert inDocumentBroker::sendRequestedTiles due to "duplicates" in the tileCombined  that is synthesized there because multiple "parts" with the same location end up in a single timecombine.

We could split by part into multiple timecombine in that sendRequestedTiles to keep things somewhat sensible as an option.

My guess is that all this worked when canceltiles worked and so clicking on a new spreadsheet tab cancelled the requests for the old tab.

mmeeks (approved):
Looks reasonable to me - thanks Caolan :-)

PR #7411, opened 2023-10-11
browser: notify user for unsupported video

lpranam:
detect if video is able to play via checking if any frames are decoded, in other cases(i.e: file not found or corrupted file) check for the errors in source and video

Change-Id: I22bfc836a98ec0eb2caa58b89332b2531d69d18f

* Resolves: # 
* Target version: master 

### Checklist


lpranam:
merged duplicate #7222

PR #7416, opened 2023-10-12
keep quarantine files on SIGABRT too

caolanm:
AdminModel::cleanupResourceConsumingDocs uses SIGABRT as first attept to kill misbehaving documents, we should give DocumentBroker a chance to quarantine documents killed off by SIGABRT as we do SIGSEGV/SIGBUS

Change-Id: Ic3a703572393050379b5a1444a5380bbeafcf2d3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -298,7 +298,9 @@ static void cleanupChildren()
         const auto it = childJails.find(exitedChildPid);
         if (it != childJails.end())
         {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Makes sense, I guess this also helps in case the crash-on-load is an assertion failure, not a crash.

PR #7417, opened 2023-10-12
admin:log activities irrespective of logging.level

Rash419:
- It logs activities like when admin logged in, authenticated metrics endpoint accessed, external monitor getting connected and admin actions like kill the document etc

Change-Id: I059f6b6ee0d8269aec7e3f521622773e348304a3

* Resolves: # 
* Target version: master

### DIFF
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -170,6 +170,12 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
             std::set<pid_t> pids = model.getDocumentPids();
             if (pids.find(pid) != pids.end())
             {
+                if (Admin::instance().logAdminAction())
+                {
+                    LOG_ANY("Admin request to kill document ["
+                            << COOLWSD::anonymizeUrl(model.getFilename(pid)) << "] with pid ["
+                            << pid << "] and source IPAddress [" << _clientIPAdress << ']');
+                }
                 SigUtil::killChild(pid, SIGKILL);
             }
             else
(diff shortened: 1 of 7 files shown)


timar (changes requested):
Can we log the IP address for all actions? E.g. when multiple admins are logged in, we wouldn't know who did what.

wsd/Admin.cpp
```diff
@@ -170,6 +170,11 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
             std::set<pid_t> pids = model.getDocumentPids();
             if (pids.find(pid) != pids.end())
             {
+                if (Admin::instance().logAdminAction())
+                {
+                    LOG_ANY("Admin request to kill document [" << model.getFilename(pid)
```
mmeeks, line 175: We need to use this: COOLWSD::anonymizeUrl(getFilename())) to follow the logging anonymization logic properly I think.

Otherwise this is all good :-)

Rash419: Done

timar (approved):

timar (changes requested):
Sorry, just one question. Why did you use false in getConfigValue() calls, when the default is true?

Rash419:
> Sorry, just one question. Why did you use false in getConfigValue() calls, when the default is true?

First I had false as a default value. I forgot to change after changing it to true. Thanks for noticing

Rash419:
CI broken: https://github.com/CollaboraOnline/online/pull/7446

eszkadev:
Please rebase. `impress/editable_area_spec.js ` will be passing then.

Other unstable tests are failing ~sometimes, but in general it should pass on second try:
writer/table_operation_spec.js
writer/top_toolbar_spec.js
writer/track_changes_spec.js
writer/undo_redo_spec.js

timar (approved):

PR #7418, opened 2023-10-12
JSDialog: Dialogs shouldn't be taller than viewport

pedropintosilva:
Don't allow dialogs to grow taller than the web browser's viewport
instead,
 - Set (for dialog's content) a max of 90% of the View height and subtract the dialog's
header. This way the header with the close button stays always visible
 - Set overflow accordingly so it can be  scrollable

Change-Id: I61a2a41d8b1198fe00b939427cde43d247dc08bb

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -73,6 +73,8 @@
 }
 
 .jsdialog-container .ui-dialog-content {
+	max-height: calc( 90vh - 34px);
+	overflow-y: auto !important;
 	background-color: var(--color-background-lighter) !important;
 	font-family: var(--jquery-ui-font);
 	line-height: 1.5;


GulsahKose (approved):

eszkadev:
For me doesn't work (no scrollbar, cannot scroll using mouse, not visible bottom of the dialog if window is small):

[resize-2023-10-12_15.30.37.webm](https://github.com/CollaboraOnline/online/assets/5307369/700647d2-c871-40de-a5c4-c0bf36e72838)

eszkadev (changes requested):
.

pedropintosilva:
> For me doesn't work (no scrollbar, cannot scroll using mouse, not visible bottom of the dialog if window is small):
> resize-2023-10-12_15.30.37.webm

thanks, please check again. It seems when I tested I was using a theme/customizations that somehow bypassed that. I found the culprit. `!important` needs to be added

eszkadev (approved):
now is ok

PR #7429, opened 2023-10-12
jsdialog: make sure all elements in frame are displayed

lpranam:
Change-Id: Ie0b949c8dcd820de380f9ee6b497631e594114f3

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -105,7 +105,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		this._controlHandlers['grid'] = this._gridHandler;
 		this._controlHandlers['alignment'] = this._alignmentHandler;
 		this._controlHandlers['buttonbox'] = this._buttonBox;
-		this._controlHandlers['frame'] = this._frameHandler;
+		this._controlHandlers['frame'] = JSDialog.frame;
 		this._controlHandlers['deck'] = this._deckHandler;
 		this._controlHandlers['panel'] = this._panelHandler;
 		this._controlHandlers['calcfuncpanel'] = this._calcFuncListPanelHandler;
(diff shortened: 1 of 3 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1082,7 +1082,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			$(frameChildren).addClass('expanded');
 
 			var children = [];
-			for (var i = 1; i < data.children.length; i++) {
+			for (var i = 0; i < data.children.length; i++) {
```
eszkadev, line 1085: 0 was label, and above code already added that label into the DOM, don't we get duplicated frame title in some cases?

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1073,22 +1073,26 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
-			var frame = L.DomUtil.create('div', 'ui-frame ' + builder.options.cssClass, container);
-			frame.id = data.id + '-frame';
-			var label = L.DomUtil.create('label', 'ui-frame-label ' + builder.options.cssClass, frame);
-			label.innerText = builder._cleanText(data.children[0].text);
-			label.id = data.children[0].id;
-			if (data.children[0].visible === false)
-				L.DomUtil.addClass(label, 'hidden');
-			builder.postProcess(frame, data.children[0]);
+			var isLabel = data.children[0].type === 'fixedtext';
+			if (isLabel) {
```
eszkadev, line 1077: we always should create the same label here, no matter if it is simple fixed text or test inside container.

Please prepare function for `extractLabelText(data.children[0])` which will return correct text, and use it for `label.innerText`

eszkadev:
unrelated test failure, will be fixed after merge of: https://github.com/CollaboraOnline/online/pull/7440

PR #7430, opened 2023-10-12
cool#6893 Avoid zeroing memory we will immediately overwrite in doRender

grandinj:
On a long running session flamegraph, we are spending 4% of the time zeroing this buffer, which we will immediately overwrite with data. Just skip the zeroing phase, and simplify the code.

Change-Id: I2b5e17e4d4cf3076d0aa244d96f9987eec3d2010

* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -334,6 +334,7 @@ COOL_JS_LST =\
 	src/control/jsdialog/Widget.Combobox.js \
 	src/control/jsdialog/Widget.DrawingArea.js \
 	src/control/jsdialog/Widget.FormulabarEdit.js \
+	src/control/jsdialog/Widget.Frame.js \
 	src/control/jsdialog/Widget.IconView.js \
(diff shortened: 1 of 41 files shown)


caolanm:
ci was broken recently, if you can rebase to current trunk after "Disable failing impress/editable_area_spec.js" I expect cypress will probably pass

mmeeks (changes requested):
So - all this unsigned char[] stuff looks nice - but previous profiles using C++ types on Android suggested that there was impossibly slow initialization of base types going on in loops left and right - everywhere; huge chunks of time spent in std::vector<char> in the destructor doing dumb stuff.
Hence using clean C stuff instead - at least calloc will use accelerated SIMD memory wiping, whereas whatever that android C++ compiler was doing was dumb beyond that of a rock. That bit me really hard.
So - I'd really prefer to use 'malloc' instead of 'calloc' if we want to save there :-)

Thanks!

PR #7433, opened 2023-10-13
Scripts: profile-cool: Add echo: openSUSE dependency

pedropintosilva:
Probably people should just run "cnf flamegraph.pl" in their own
distribution. Nevertheless and because there are multiple flamegraph
packages (unrelated to this one) on the official and community
repositories, probably best to mention the package where people can
get the mentioned perl profiler.

Change-Id: Icd4139fd01418892eaad7c23d6346d9fda543b7d

### DIFF
--- a/scripts/profile-cool
+++ b/scripts/profile-cool
@@ -12,6 +12,7 @@ fi
 if ! which flamegraph.pl > /dev/null 2>&1; then
     echo "no flamegraph.pl found"
     echo "On fedora install systemwide with: sudo dnf install flamegraph"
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
I confirm installing that package gives me a plausible /usr/bin/flamegraph.pl.

timar (approved):

pedropintosilva:
nah, the best is to simply clone https://github.com/brendangregg/FlameGraph/tree/master into `~/FlameGraph`.

`perl-Devel-NYTProf` doesn't offer `stackcollapse-perf.pl` which is also needed. And there is no `stackcollapse-perf.pl` anywhere on any repo be that the official or any of the other ones (searched via opi)

PR #7435, opened 2023-10-13
cool#7374 very little delta threading

caolanm:
We only run one task in each thread when run() is called, then complete the remainder of tasks in the main thread while the other threads wait.

https: //github.com/CollaboraOnline/online/issues/7374

Change-Id: Id4fc0a620d98b0bb55310b495eeff7411a4544b5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -124,7 +124,7 @@ public:
         while (!_shutdown)
         {
             _cond.wait(lock);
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Thanks Caolan :-) would be great to back-port to 22.05 and so on :-)

PR #7436, opened 2023-10-13
Fix CSS selector: make checkboxes visible in light mode

timar:
Change-Id: I6e8194ea456e4af3104754cf9219be22f135c042

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Good catch, thanks

Issue #4727, opened 2022-05-03
style previews white background


Andreas-Kainz:
The styles preview images has a white background which is ok, cause the background color is ordinary white. 
![2022-05-03 13_24_17-writer odt - Nextcloud](https://user-images.githubusercontent.com/8517736/166446584-c341edfc-d278-4ad0-986f-2355a8fce078.png)

Also MSO show a white background at the previews.
![Dark_Mode_in_Word_in_SV_NEW](https://user-images.githubusercontent.com/8517736/166446225-d487f609-1b2a-43dd-8632-dfe129053667.png)

But in general I would prefer to have a png with transparent background and the background color as rgb color. 

If the background color was set (#ffffff, #ff0000, ...), than the preview should show this background color, if it's not set (auto/none), than the preview should'n show the background color and stay transparent.

Andreas-Kainz:
@eszkadev is it possible to get the styles preview from core as png with transparent background color?

@pedropintosilva fyi

pedropintosilva:
Thanks @Andreas-Kainz for reporting it. But unfortunately this needs to be a change done in core. The image previews should come, ideally, without background so we can then simply change the background via css on online. Any online-only solution will otherwise bring confusing and ultimately distort the font rendering (sometimes just the color and contrast other times depending on the effects even the  sharpness)  
![image](https://github.com/CollaboraOnline/online/assets/65948705/257d9445-5e33-4945-a1ae-0f47ad1172e4)

And even then, with core change, it will not fix for styles with their own background. Plus it can be quite misleading changing the background to dark since that is not really true: if printed the default styles will come out without background on a white piece of paper. And that is probably why MS also chooses to have that in white.

Maybe an improvement would be to just make sure, when in dark mode, that those previews do not have pure %100 lightness white

juliushaertl:
I think this should be all fixed

sudwhiwdh:
@juliusknorr How can this issue be solved technically?

Andreas-Kainz:
The preview generated in core should have a transparent background when no background color was defined.

PR #7437, opened 2023-10-13
dark mode improvement of UI images

timar:
- Writer - Insert tab - Page Number page preview

image

- Writer - References tab - Table of Contents and Index - Columns preview 

image

- Calc - Autofilter - Two image buttons

image

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -97,3 +97,12 @@ button.jsdialog img {
 button.has-img {
 	min-width: auto !important;
 }
+
+/* Calc Auto-Filter buttons coming from core */
+[data-theme='dark'] button#select_current.has-img > img {
+	filter: invert(.9)
+}
+
+[data-theme='dark'] button#unselect_current.has-img > img {
+	filter: invert(.9)
+}
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1535,6 +1535,17 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	margin: auto;
 }
 
+/* Writer - Insert - Page Number - page preview image */
+[data-theme='dark'] #PageNumberDialog #previewImage {
+	filter: invert(.95)
+}
+
+/* Writer - Table of Contents, Index or Bibliography dialog, Columns tab */
+[data-theme='dark'] #ColumnPage #pageexample-img, #ColumnPage #valueset-img {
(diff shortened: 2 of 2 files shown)


timar:
Maybe I should split this, I see that style previews are controversial (#4727).

browser/css/btns.css
```diff
@@ -97,3 +97,12 @@ button.jsdialog img {
...
 }
+
+/* Calc Auto-Filter buttons */
+[data-theme='dark'] button#select_current.has-img > img {
+	filter: invert(.9)
+}
+
+[data-theme='dark'] button#unselect_current.has-img > img {
+	filter: invert(.9)
+}
```
pedropintosilva, line 108: Thank you . I would prefer if we would just change the path of src to use the dark subfolder but I see that these are base64 images (probably coming from core) . So I think these lines should get in.

browser/css/jsdialogs.css
```diff
@@ -1510,10 +1510,22 @@ input[type='number']:hover::-webkit-outer-spin-button {
...
 }
 
+[data-theme='dark'] .ui-drawing-area-container {
+	filter: invert(.9)
+}
+
+[data-theme='dark'] img.notebookbar {
+	filter: invert(.9)
+}
+
```
pedropintosilva, line 1528: This or any CSS color graphical effects cannot enter if they affect element that:
1. Can have any other color than black or white (otherwise it will be impossible to see/read)
2. Even if some elements can only be monochrome they should stay faithful to how the styles appear when printed (standard used also in MSO etc)

browser/css/jsdialogs.css
```diff
@@ -1510,10 +1510,22 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	margin: auto;
 }
 
+[data-theme='dark'] .ui-image {
+	filter: invert(.95)
+}
+
```
pedropintosilva, line 1516: This can enter but I'm afraid of being too generic and ending up inverting images that shouldn't be inverted (either because will look odd, or because they are already in dark mode). Better to add `#PageNumberDialog` before `.ui-image`

timar:
https://gerrit.libreoffice.org/c/core/+/158307 will solve the ToC preview.

timar:
@pedropintosilva I amended my patch. I updated the PR description.

pedropintosilva (approved):
Thanks!

PR #7438, opened 2023-10-14
ProcSMapsFile might not be a valid file descriptor

caolanm:
in which case don't send it, which then implies we might have an empty shareFDs which we can treat the same as a nonexisting shareFDs

Change-Id: I80a78a01c69dbee5ee28a64442a5069a6c2b4dbe

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (approved):

mmeeks (approved):
Looks fine to me. Thanks :-)

PR #7443, opened 2023-10-15
Small jsdialog cleanup

eszkadev:
- move mobile border selector widget to separate file
- reduce warnings in sidebar and formula bar we can ignore


gokaysatir (approved):
Thanks @eszkadev :)

PR #7444, opened 2023-10-15
different tile widths in requestedTiles on switching calc tabs

caolanm:
if we zoom in calc quickly that can result in still unfulfilled requests for the first set of tiles from the earlier zoom in requestedTiles when the new requests are compared with the existing one for duplicates, but they are of different tilewidth/tileheight and shouldn't match as a duplicate.

Change-Id: Ifc3f905c5304feb6c05696835dcb4eaed7f1e546

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3199,22 +3199,17 @@ void DocumentBroker::handleTileCombinedRequest(TileCombined& tileCombined, bool 
     // Drop duplicated tiles, but use newer version number
     else
     {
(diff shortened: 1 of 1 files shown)


caolanm:
The issue is very similar and fixed the same way as https://github.com/CollaboraOnline/online/pull/7410

vmiklos (approved):
Yes, this sounds reasonable. If tile width or height can differ in practice, then don't just warn but actually handle the difference.

PR #7446, opened 2023-10-16
Test fails on CI

eszkadev:
This reverts commit af318702d1c7fb00ed1962e993ee654c0c76e197.

From the CI history I see that first time when 4 tests failed is on that PR.
Which also passed only sometimes before merge:

https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/7848/

Test if tests ara failing without this change too. If yes then it means some core change caused it.

Failing tests:
writer/table_operation_spec.js
writer/top_toolbar_spec.js
writer/track_changes_spec.js
writer/undo_redo_spec.js

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -348,8 +348,8 @@ app.definitions.Socket = L.Class.extend({
 					}
 					catch (e)
 					{
-						// unpleasant - but stops this one problem event
-						// stopping an unknown number of others.
+						// unpleasant - but stops this one problem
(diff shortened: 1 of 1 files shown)


eszkadev:
5 times this PR passed without failing any test, what can confirm something is bad with the mentioned commit.
I will try to minimize revert in https://github.com/CollaboraOnline/online/pull/7460

eszkadev:
not needed fixed by https://github.com/CollaboraOnline/online/pull/7463

PR #7447, opened 2023-10-16
Private/caolanm/coverity25

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/lokassert.hpp
+++ b/test/lokassert.hpp
@@ -108,12 +108,18 @@ inline constexpr bool failed() { return false; }
 } // namespace detail
 } // namespace test
 
+#if !defined(__COVERITY__)
+#define LOK_ASSERT_MESSAGE_PRIOR_FAILURE failed()
+#else
+#define LOK_ASSERT_MESSAGE_PRIOR_FAILURE false
+#endif
+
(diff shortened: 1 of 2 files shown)


test/lokassert.hpp
```diff
@@ -108,12 +108,18 @@ inline constexpr bool failed() { return false; }
...
+#define LOK_ASSERT_MESSAGE_PRIOR_FAILURE false
+#endif
+
 /// Assert the truth of a condition, with a custom message.
 #define LOK_ASSERT_MESSAGE_IMPL(message, condition, silent)                                        \
     do                                                                                             \
     {                                                                                              \
         using namespace test::detail;                                                              \
-        if (!failed())                                                                             \
+        if (!LOK_ASSERT_MESSAGE_PRIOR_FAILURE)                                                     \
```
vmiklos, line 122: I think cppunit gets around this by throwing an exception when the "assert" macro fails, so you don't have to explicitly have a special "already failed" mode, which is much more clean. But anyway, this is good enough for now.

vmiklos (approved):

PR #7448, opened 2023-10-16
cool#7374 very little delta threading

caolanm:
We only run one task in each thread when run() is called, then complete the remainder of tasks in the main thread while the other threads wait.

https: //github.com/CollaboraOnline/online/issues/7374
Change-Id: Id4fc0a620d98b0bb55310b495eeff7411a4544b5 (cherry picked from commit d6c1916cb201dcd0c851b31ed776c1cbf24540a0)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -119,7 +119,7 @@ public:
         while (!_shutdown)
         {
             _cond.wait(lock);
-            if (!_shutdown && !_work.empty())
+            while (!_shutdown && !_work.empty())
                 runOne(lock);
         }
     }


mmeeks (approved):
Looks great - but lets leave this to get some testing of this on some live servers - there could be some fun existing threading problem that really hammering this exposes.

timar:
> Looks great - but lets leave this to get some testing of this on some live servers - there could be some fun existing threading problem that really hammering this exposes.

The good strategy is then merge, then see what happens on staging servers. Nothing will happen if we leave this here as-is.

Issue #7236, opened 2023-09-11
In Columns dialog, plus/minus in spinfield doesn't adjust the value


thebearon:
In Columns dialog, the number of columns can't be aligned using the +/- buttons.

- Open Columns dialog in Writer,
- Try increasing number of columns by pressing the `+` in the spinfield.

-> Number of columns stays the same.

It can be adjusted by editing the field manually.

CC @eszkadev.

timur-g:
Regression:
commit 4b9cbab33db05b5ed601fbeb97fa0afc315f5bac
Author: Szymon Kłos <szymon.klos@collabora.com>
Date: Fri Aug 11 09:17:01 2023 +0200

jsdialog: fix spinfield precision detection

timur-g:
Thanks.

PR #7449, opened 2023-10-16
Fix spinfield doesn't adjust the value

Darshan-upadhyay1110:
Change-Id: I740f86e9477656d371ba3b8793bff99671444d7c

    - Condition modified for getPrecision.
    - it was not handling the cases for dot value less then 0.
   
* Resolves: #7236
* Target version: master


eszkadev (approved):

pedropintosilva (approved):
Thanks

PR #7450, opened 2023-10-16
a11y: accessible description to describe how to turn on accessible mode

mcecchetti:
Set up accessible text through aria-description for describing how to
enable accessibility support for text content when it is disabled.

The description is reported by the screen reader on document load and
on explicit request to read current text content.

Change-Id: I439b7703c15d3b38c39181d27c11da438834c414

### DIFF
--- a/browser/src/layer/marker/TextInput.js
+++ b/browser/src/layer/marker/TextInput.js
@@ -8,7 +8,7 @@
  * text area itself.
  */
 
-/* global app */
+/* global app _ */
 
 L.TextInput = L.Layer.extend({
 	initialize: function() {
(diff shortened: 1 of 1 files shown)


mmeeks (changes requested):
This looks great in principle Marco ! =) but ...

I expect this breaks our translation tooling.
I would expect a translated string to have several string literals with _() around them.
If we need to substitute something into the string - we should have a %FOO% that is replaced into the string (concatenation seems cheap but won't fit all languages).
Can you tweak for that  :-)

mcecchetti:
Okay. I translated the single sentences.

mmeeks (approved):
Thanks Marco :-)

Issue #7196, opened 2023-08-31
Format Cell broken with custom text


relikd:
### Describe the Bug

Format Cell does not update cell if using static text in format. I knew this worked before, because my spreadsheet has some cells that were formatted with Collabora. Further, it also works if I reuse a previous Cell Format from the format history list. But as soon as I try to change the text, it will ignore my input and revert to the previous value.

### Steps to Reproduce

1. Enter a number in a new cell
2. Click on 'Format Cell'
3. Set new format to be `#" suffix"`

### Expected Behavior

The cell should now print "12 suffix"

### Actual Behavior

The cell is either malformed or does not update at all.

### Screenshots

https://github.com/CollaboraOnline/online/assets/1321720/a0807dc0-05b8-476e-9ed3-30d07e1c153a

### Desktop

 - **Device:** macOS
 - **Browser:** Firefox 116
 - **Version:** collabora/code:23.05.3.1.1

### Additional Context

N/A

Tex2002ans:
I can reproduce in:

> COOLWSD version: 23.05.3.1 (git hash: 0621aa48)
> LOKit version: Collabora Office 23.05.3.1 (git hash: aaeaa1f)
> Served by: Linux Mint 21.1
> Server ID: 7ace7a16

- - -

In LibreOffice 7.6.0:

- Step 4, you type: `#" suffix"`.
- Press OK.
- Right-Click > Format Cells
- Still see `#" suffix"` in "Format-Code" box.

In Collabora Online, you get:

- Step 4, you type: `#" suffix"`.
- Press OK.
- Right-Click > Format Cells
- Still see `#" suffix` in "Format-Code" box.
   - Notice the missing quote at the end!

So it looks like that Collabora version of the box might be doing extra parsing on the string and getting confused with the quotation marks (?).

- - -

**Side Note:** While testing this after Step 4, I also got a random crash.

Here is the log + backtrace:

<details>
<pre>
warn:editeng:3296:3289:editeng/source/misc/acorrcfg.cxx:108: Unable to iterate directory. com.sun.star.ucb.InteractiveAugmentedIOException message: "an error occurred during file opening" Code: 22
warn:vcl:3296:3289:vcl/source/window/floatwin.cxx:66: FloatWindow::FloatingWindow(): - pParent == NULL!
warn:vcl:3296:3289:vcl/source/window/floatwin.cxx:71: FloatWindow::FloatingWindow(): - pParent == NULL and no AppWindow exists
[...]

PR #7452, opened 2023-10-16
Fix Format Cell broken with custom text

Darshan-upadhyay1110:
Change-Id: I27cf86972e70d956b9005fc1ea0b65a622e898d7

* Resolves: #7196
* Target version: master


pedropintosilva (approved):
It works. Initially, when reading the code, I thought replaceAll could lead to unwanted results. However we do need that as replace only 1 pair is not enough. User can combine as much place holders as he wants when defining a custom format. Example:

> For example, if you want to apply different colors to different temperature data, enter:
> ```
> [BLUE][<0]#.0 "°C";[RED][>30]#.0 "°C";[BLACK]#.0 "°C"
> ```
file:///opt/collaboraoffice/help/en-US/text/shared/01/05020301.html?&DbPAR=CALC&System=UNIX

PR #7453, opened 2023-10-16
browser: fix white document background on reconnect with darkmode

Rash419:
Change-Id: Ic853ba5108a1f6f9db2a7f8ab5310fe77357b07c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1387,7 +1387,11 @@ app.definitions.Socket = L.Class.extend({
 		else if (this._reconnecting) {
 			// we are reconnecting ...
 			this._map._docLayer._refreshTilesInBackground();
-			this._map.fire('statusindicator', {statusType: 'reconnected'});
+			this._map.fire('statusindicator', { statusType: 'reconnected' });
+
+			var selectedMode = this._map.uiManager.getDarkModeState();
+			this._map.uiManager.activateDarkModeInCore(selectedMode);
+
 			var uiMode = this._map.uiManager.getCurrentMode();
 			if (uiMode === 'notebookbar') {
 				this._map.uiManager.notebookbar.resetInCore();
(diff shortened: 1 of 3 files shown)


browser/src/core/Socket.js
```diff
@@ -1533,6 +1534,7 @@ app.definitions.Socket = L.Class.extend({
 			}
 			if (this._map._docLayer._docType === 'presentation')
 				this._map._isCursorVisible = false;
+			this._map._docLayer._canonicalIdInitialized = false;
```
eszkadev, line 1537: I think it would be better to not modify "private" variable outside CanvasTileLayer.js.
Could you create method inside CanvasTileLayer.js which resets that value?
like: `resetCanonicalId()`

browser/src/core/Socket.js
```diff
@@ -1387,7 +1387,13 @@ app.definitions.Socket = L.Class.extend({
 		else if (this._reconnecting) {
 			// we are reconnecting ...
 			this._map._docLayer._refreshTilesInBackground();
-			this._map.fire('statusindicator', {statusType: 'reconnected'});
+			this._map.fire('statusindicator', { statusType: 'reconnected' });
+
+			var selectedMode = this._map.uiManager.getDarkModeState();
+			var cmd = { 'NewTheme': { 'type': 'string', 'value': '' } };
+			selectedMode ? cmd.NewTheme.value = 'Dark' : cmd.NewTheme.value = 'Light';
+			app.socket.sendMessage('uno .uno:ChangeTheme ' + JSON.stringify(cmd));
```
eszkadev, line 1395: This also could be inside some helper function and be used in Control.UIManager.js as well...
like `activateDarkMode(true / false)` .  You can put it in UIManager.

and be used in:` toggleDarkMode` https://github.com/CollaboraOnline/online/blob/bb246318fef5b409c237dc7bf34147a164c18b0d/browser/src/control/Control.UIManager.js#L118

and in `initDarkModeFromSettings` https://github.com/CollaboraOnline/online/blob/bb246318fef5b409c237dc7bf34147a164c18b0d/browser/src/control/Control.UIManager.js#L147

PR #7462, opened 2023-10-17
jsdialog: avoid error when container is missing

eszkadev:
TypeError was noticed where container was undefined. This was case where snackbar was closed.


gokaysatir (approved):
Thanks for the fix @eszkadev  :)

PR #7463, opened 2023-10-17
Assertion in DocumentBroker::sendRequestedTiles fails on running cypr…

caolanm:
…ess impress tests

make -C cypress_test check-desktop

asserts seen in cypress_test/cypress/wsd_logs/coolwsd_output.log of:

coolwsd: wsd/DocumentBroker.cpp:3134: void DocumentBroker::sendTileCombine(const TileCombined&): Assertion `!newTileCombined.hasDuplicates()' failed.

If we check for, and don't reuse, an old request with a different NormalizedViewId then we could end up with multiple requests with different NormalizedViewIds that end up in the same final tilecombine.

similarly there was no check for different modes ending up in the same tilecombine.

just split out the logic we have to see if two tiles have the same properties that appear as a shared set of properties for tilecombine and use that in the two relevant places.

Change-Id: Ieb2ee0e85f124dd57c6b050e5b669dd808cf6bbf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3206,10 +3206,7 @@ void DocumentBroker::handleTileCombinedRequest(TileCombined& tileCombined, bool 
             {
                 if(oldTile.getTilePosX() == newTile.getTilePosX() &&
                    oldTile.getTilePosY() == newTile.getTilePosY() &&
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Looks like a great fix - thanks Caolan :-)

PR #7464, opened 2023-10-17
jsdialog: make mobile border style control compatible

eszkadev:
- jsdialog requires to create in the handler single root node with id of the widget
- remove old style update by _refreshSidebar

### DIFF
--- a/browser/src/control/jsdialog/Widget.MobileBorderSelector.js
+++ b/browser/src/control/jsdialog/Widget.MobileBorderSelector.js
@@ -80,13 +80,16 @@ function _borderControlItem(parentContainer, data, builder, i, selected) {
 	});
 }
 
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
This works, we can use a more specific id for mainContainer if that dialog doesn't depend on that id. Because "borderstyle" is a generic one and it looks like there may be an id conflict in the future? Thanks for the improvements.

eszkadev:
> This works, we can use a more specific id for mainContainer if that dialog doesn't depend on that id. Because "borderstyle" is a generic one and it looks like there may be an id conflict in the future? Thanks for the improvements.

Main container id always has to be unmodified and set to value of "data.id". This is required to make updates and actions in the widget.

If we want more unique id's we need to change id in the core of particular widgets.

PR #7465, opened 2023-10-17
Private/caolanm/coverity26

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/tools/Replay.hpp
+++ b/tools/Replay.hpp
@@ -213,7 +213,7 @@ public:
         std::cerr << "Attempt connect to " << uri << " for trace " << _trace << "\n";
         getNextRecord();
         _start = std::chrono::steady_clock::now() + std::chrono::milliseconds(delayMs);
-        _nextPing = _start + std::chrono::milliseconds((long)(std::rand() * 1000.0) / RAND_MAX);
+        _nextPing = _start + std::chrono::milliseconds(Util::rng::getNext() % 1000);
         _lastTile = _start;
     }
 
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -2189,14 +2189,18 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
         else if (tokens.equals(0, "extractedlinktargets:"))
         {
             LOG_TRC("Sending extracted link targets response.");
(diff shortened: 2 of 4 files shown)


tools/Replay.hpp
```diff
@@ -213,7 +213,7 @@ class StressSocketHandler : public WebSocketHandler
         std::cerr << "Attempt connect to " << uri << " for trace " << _trace << "\n";
         getNextRecord();
         _start = std::chrono::steady_clock::now() + std::chrono::milliseconds(delayMs);
-        _nextPing = _start + std::chrono::milliseconds((long)(std::rand() * 1000.0) / RAND_MAX);
+        _nextPing = _start + std::chrono::milliseconds(Util::rng::getNext() % 1000);
```
vmiklos, line 216: Just commenting here to note down my understanding. We used to have a 0..1 random number here (rand() / RAND_MAX), so this used to be some time between 0 and 1 seconds. The new way just gives us a random integer, hence the %1000.

wsd/ClientSession.cpp
```diff
@@ -2189,14 +2189,18 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
         else if (tokens.equals(0, "extractedlinktargets:"))
         {
             LOG_TRC("Sending extracted link targets response.");
+            if (!_saveAsSocket)
```
vmiklos, line 2192: One should read these with 'git show -w' and then these are straightforward modifications.

wsd/Storage.cpp
```diff
@@ -832,6 +832,8 @@ WopiStorage::WOPIFileInfo::WOPIFileInfo(const FileInfo& fileInfo,
             anonObject->set("UserFriendlyName", COOLWSD::anonymizeUsername(_username));
         }
         anonObject->stringify(wopiResponse);
+
+        // coverity[freed_arg] - object/anonObject's nRefCount is > 1 here
```
vmiklos, line 836: The trouble here is that both object and anonObject goes out of scope, but it's fine, the first will decrease nRefCount to 1, the second will decrease to 0 & free, so there is no double-free.

vmiklos (approved):
I see nothing obviously wrong with this.

PR #7466, opened 2023-10-17
Fix JS error on translations.

gokaysatir:
queryselectorall('something').queryselectorall('something else') fails.

^ This JS error is fixed.

Also we don't have specific images for German help page. So we removed the "de" from that language list.

Change-Id: I69b56e8f4bd2b92835c4aca9c2284750f3df9455

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -520,10 +520,11 @@ L.Map.include({
 		}
 
 		//translatable screenshots
-		var supportedLanguage = ['fr', 'it', 'de', 'es', 'pt-BR'];
+		var supportedLanguage = ['fr', 'it', 'es', 'pt-BR'];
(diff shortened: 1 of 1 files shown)


pedropintosilva:
@gokaysatir could you please add some context to the commit message? Like what is fixing and why. Sorry I don't want to be annoying just trying to get nice commits that are easy to trace back : )

gokaysatir:
> @gokaysatir could you please add some context to the commit message? Like what is fixing and why. Sorry I don't want to be annoying just trying to get nice commits that are easy to trace back : )

You are right, thank you for the heads up. I updated the commit message and the description here.

eszkadev (approved):
fixes the issue

Issue #7468, opened 2023-10-17
Rename dialog should have in the input field the current document name


pedropintosilva:
### Describe the Bug

The rename dialog triggered by going to File tab > Rename, should have the input field completed with the current file name and extension. Ideally with the filename selected, so user can just start to type and rewrite to something else. If difficult, then we can have it with the whole thing selected.

Currently the dialog opens with the input field empty.

pedropintosilva:
as discussed with @juliushaertl and @jancborchardt

PR #7472, opened 2023-10-17
UI: Prefill rename dialog with the current document name

lpranam:
resolves: #7468

Change-Id: Icb6563f76cc0f9c7d81eac96fedc6e476dfdad46

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -152,7 +152,8 @@ L.Control.UIManager = L.Control.extend({
 	renameDocument: function() {
 		// todo: does this need _('rename document)
 		var docNameInput = this.documentNameInput;
-		this.showInputModal('rename-modal', _('Rename Document'), _('Enter new name'), '', _('Rename'),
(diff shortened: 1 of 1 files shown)


lpranam:
@pedropintosilva I wish we could keep the name(just name without extension) selected in some way. I might do that as follow-up

pedropintosilva:
> @pedropintosilva I wish we could keep the name(just name without extension) selected in some way. I might do that as follow-up

Yes that would be good, Or at least for now please make a PR that the whole thing is selected. Currently nothing is selected

pedropintosilva (approved):
Thanks

lpranam:
> > @pedropintosilva I wish we could keep the name(just name without extension) selected in some way. I might do that as follow-up
> 
> Yes that would be good, Or at least for now please make a PR that the whole thing is selected. Currently nothing is selected

that's complicated, that's why its gonna be a follow up 😂

pedropintosilva:
Yes that is what I meant make a PR, meaning the follow up xD

PR #7473, opened 2023-10-18
Fix media request error

eszkadev:
It fixes richdocumentscode case when trying to request media:
`[ websrv_poll ] ERR  #-1: Failed to open file [.../9a2gf9.ogg] for uploading| net/HttpRequest.hpp:1603`

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3300,7 +3300,16 @@ void DocumentBroker::handleMediaRequest(std::string range,
         {
             // For now, we only support file:// schemes.
             // In the future, we may/should support http.
-            const std::string path = getJailRoot() + url.substr(sizeof("file://") - 1);
+#if !MOBILEAPP
+            // We always extract media files in /tmp. Normally, we are in jail (chroot),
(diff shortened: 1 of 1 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -3300,7 +3300,12 @@ void DocumentBroker::handleMediaRequest(std::string range,
         {
             // For now, we only support file:// schemes.
             // In the future, we may/should support http.
-            const std::string path = getJailRoot() + url.substr(sizeof("file://") - 1);
+#if !MOBILEAPP
+            const std::string root = COOLWSD::NoCapsForKit ? "/" : getJailRoot();
```
Ashod, line 3304: This looks reasonable. I'd only add a comment to explain what is going on here. Something like:

> We always extract media files in /tmp. Normally, we are in jail (chroot), and this would need to be accessed from WSD through the JailRoot path. But, when we have `NoCapsForKit`, there is no jail, so the media file ends up in the host (AppImage) /tmp.
> Therefore, we need to change the path prefix here to account for this.

Ashod (approved):
Thanks @eszkadev. Looks good. One suggestion.

PR #7476, opened 2023-10-18
cid#318817 avoid Uncaught exception

caolanm:
and

cid#318819 Uncaught exception

take a reference in the COOLWSD/DocumentBroker ctor, so indicating it exists before the calls in the COOLWSD/DocumentBroker dtor, and so the Admin ctor doesn't throw during the COOLWSD/DocumentBroker dtor.

Change-Id: I8190cc3594a5f81fedd355aeadcca45e532bda90

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
I don't see anything in the Admin ctor that would require an already existing COOLWSDServer, so this looks OK.

PR #7481, opened 2023-10-18
Help window was too narrow

timar:
Change-Id: I245e0979cd069f13eb8237eb155aa1760289a180

* Resolves: # 
* Target version: master 

### Summary

I did not modify the keyboard shortcut window, because I found that it is better when keyboard shortcuts and their descriptions are close to each other.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1772,3 +1772,8 @@ kbd,
 #modal-dialog-copy_paste_download_progress .ui-dialog-content {
 	width: 450px;
 }
+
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
I think here we could use the viewport width (https://developer.mozilla.org/en-US/docs/Web/CSS/length#vw) but then again in this case % will have the same result (even though is related to its parent which is modal-dialog-online-help-content-box-overlay that is always filling up the screen, so good to go)

PR #7483, opened 2023-10-19
jsdialog: fix scrollbars inside dialogs

eszkadev:
This fixes side effect from commit a1b2091ae622437f531f1126a8de9c143b06be68
JSDialog: Dialogs shouldn't be taller than viewport

which introduced scrollbars to dialogs when not needed, examples:
   Writer -> Format -> Character
   Calc -> Data -> Validity

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -80,6 +80,19 @@
 	line-height: 1.5;
 }
 
+#SortKeyWindow {
+	/* As used in .jsdialog-container .ui-dialog-content 90 % of the viewport
+	   minus 352px:
+		- 34px the dialog's header
+		- 112px the next sibling
+		- 50px the dialog's buttons
+		- 80px status bar and calc sheet bar
+		- 38px formulabar
+		- 38px document-header
+   */
+	max-height: calc( 90vh - 352px);
+}
+
 .jsdialog-container .ui-dialog-content > div:not(.ui-scrollwindow) {
 	overflow: hidden; /* don't inherit scrollbar from parent */
 }
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -80,6 +80,10 @@
 	line-height: 1.5;
 }
 
+.jsdialog-container .ui-dialog-content > div:not(.ui-scrollwindow) {
+	overflow: hidden; /* don't inherit scrollbar from parent */
+}
+
 .jsdialog-container .lokdialog.ui-dialog-content.ui-widget-content {
 	overflow: hidden;
 }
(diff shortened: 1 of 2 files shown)


browser/css/jsdialogs.css
```diff
@@ -202,6 +206,10 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	overflow: auto;
 	max-height: 500px;
 	max-width: 700px;
+}
+/* - Also used in SpellingDialog where suggestion come in form
+     of an image*/
+#SpellingDialog-dialog-vbox .jsdialog.ui-scrollwindow {
```
eszkadev, line 212: This width: max-content is very important for all places where image is inside scroll window.
Now in places like: Writer -> Character dialog, preview on the bottom is cropped. 
Or Format -> Theme icon view.

It was introduced to fix exactly that bug.
Please do opposite: keep it as is, but change value to initial for sort dialog (as it is only example where it breaks anything).

eszkadev:
It doesn't fix the sort dialog case (double scrollbars). We need to limit inner scroll window height I think.

doublescroll

browser/css/jsdialogs.css
pedropintosilva: really? the commit from Darshan doesn't say nothing related to that, Just a subject and just about one single dialog.

eszkadev: commit doesn't say but it's true :)

pedropintosilva:
as I tried to explain, there is something here missing in our logic. It doesn't make sense to set max-content to every single scrollwindow , this has horrible effect on any case where we don't have inside a drawingarea and instead something else. I don't even understand how this got merged in the first place.

Here is what I propose:

Possibly in the `Widget.ScrolledWindow.js` we could 

Check if any of the children has css class 'ui-drawing-area-container'. If true we would add addiitonal 'has-ui-drawing-area' class to ' ui-scrollwindow-content', done.

browser/src/control/jsdialog/Widget.ScrolledWindow.js
```diff
@@ -30,11 +30,29 @@
...
+	for (var i in children) {
+		if (children[i].type === 'drawingarea')
+			return true;
+		if (_hasDrawingAreaInside(children[i].children))
+			return true;
+	}
+
+	return false;
+}
+
```
pedropintosilva, line 46: Thank you! I tried using the data.children but sometimes was getting nothing. I see the importance of the first check. I then tried to use content but I also see that it's much better to use the JSON object instead of the HTMLCollection item (data.children VS content)

pedropintosilva (approved):

PR #7486, opened 2023-10-19
Private/caolanm/coverity28

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -752,10 +752,10 @@ void FileServerRequestHandler::readDirToHash(const std::string &basePath, const 
             strm.zalloc = Z_NULL;
             strm.zfree = Z_NULL;
             strm.opaque = Z_NULL;
-            int result = deflateInit2(&strm, Z_DEFAULT_COMPRESSION, Z_DEFLATED, 31, 8, Z_DEFAULT_STRATEGY);
-            if (result != Z_OK)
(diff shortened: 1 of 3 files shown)


wsd/FileServer.cpp
```diff
@@ -786,7 +786,13 @@ void FileServerRequestHandler::readDirToHash(const std::string &basePath, const
                 strm.next_out = (unsigned char *)&cbuf[0];
                 strm.total_out = strm.total_in = 0;
 
-                deflate(&strm, Z_FINISH);
+                const int deflateResult = deflate(&strm, Z_FINISH);
+                if (deflateResult != Z_OK && deflateResult != Z_STREAM_END)
+                {
+                    LOG_ERR("Failed to deflate, result: " << deflateResult);
+                    free(cbuf);
+                    break;
```
vmiklos, line 794: It looks a bit inconsistent that here we go for "break" after logging the error, but above pre-existing code goes with "continue" after logging the error. I assume these typically don't fail in practice; so if they do, it would be better to also do "break" above? But that's pre-existing code. Just sharing my concerns. :-)

vmiklos (approved):
See inline comment for a nit, looks OK otherwise.

wsd/FileServer.cpp
caolanm: Yeah the outer continue on deflateInit2 failure matches the existing stat failure continue. Unsure if there is a great solution on failures of these.

PR #7487, opened 2023-10-19
Added the images for the Help guide in German language

Ezinnem:
Change-Id: Iff34a6d00cacfc0cb4597cc987e2bb8b3427b15f

* Resolves: The images for German localized Help
* Target version: master 

### Checklist


timar (changes requested):
- Checkboxes are invisible, you did not build from the latest source.
- Instead of Jane Smith, please use a German name, like Anna Schmidt.
- Guest is Gast in German
- In manage-changes.png the Aktion column is empty, this is a bug, so you have to re-take this screenshot after the bug is fixed (FYI @aszucs3)
- In selection-list.png Apkel should be Apfel.
- In thesaurus.png there is an empty button, a back arrow should be there. It is a bug, and you have to re-take this screenshot after the bug is fixed.

timar (changes requested):
Thank you, these are much better. Can you please re-take browser/images/help/de/manage-changes.png? The action icons have been merged I believe.

timar (approved):

Issue #7428, opened 2023-10-12
perf: calc - doc_setView shouldn't change toolbars


mmeeks:
Switching view should not of itself fiddle with toolbars or the notebookbar etc. it seems in the trace here:

https://user-images.githubusercontent.com/833656/274579731-f8946ec3-74da-4d50-91fc-518d6c14639c.svg

We have SfxDispatcher::Update_Impl being called on view switching; in every mouse move it updates the toolbar bits, but also in every paint tile, in every input event etc. - eg. the cost of getCommandValues is totally dominated by pointless doc_setView work being done:

![image](https://github.com/CollaboraOnline/online/assets/122848/eddf8bd7-a18d-43d4-b8b4-d2b3032d1fa0)

Quickly adding this up I see around 10% of total CPU cycles spent.

It seems that Document::renderTiles - does a doc_setView - costing 8bn cycles, which then does another doc_setView inside itself costing 11bn+ cycles.

Quite probably this is caused by the outer version not understanding canoncial-views, and creating extra work - nevertheless - perhaps an easy one-liner win to remove the doc_setView in Document::renderTiles (?) or ...

![image](https://github.com/CollaboraOnline/online/assets/122848/4e7a54b5-48d9-4fdf-97eb-cecdf27a4ee0)

Thanks ! =)

caolanm:
This is an odd one.

We do a _loKitDocument->setView in kit/Kit.cpp in renderTiles and locally with multiple connections to a calc document I see no sign of an inner setView from doc_paintPartTile getting called unlike seen above.

Also in the flamegraph I see the WeldedTabbedNotebookbar ctor getting called a load of times, eating most of the doc_setView time, something I don't see locally.

The WeldedTabbedNotebookbar is created from SfxNotebookBar::StateMethod which is called from ScDocShell::GetState. In ScDocShell::GetState SfxNotebookBar::StateMethod is called with bReloadNotebookbar of false, so presumably the creation must be due to the other condition of m_pNotebookBarWeldedWrapper.find(pViewShell) != m_pNotebookBarWeldedWrapper.end(). But we keep that mapping of viewshell of notebookbar until VclDisposeHdl is called.

Locally I see one notebookbar per view, created when a user joins and not torn down to be replaced until that user exits.

I could speculate that the first issue was related to the recently fixed DocumentBroker::sendRequestedTiles but I'm not sure the same could account for the second issue.

caolanm:
[...]

PR #7494, opened 2023-10-19
avoid recreating the core notebookbar when we don't have to

caolanm:
this is inspired by trying to reproduce:

https://github.com/CollaboraOnline/online/issues/7428

Where I see that we tear down and recreate the notebookbar when we switch from light to dark mode, and we recreate it when Zotero support is enabled.

Change-Id: I4a95af357a95931df12ff96d9378b235d27e3eaf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -15,6 +15,7 @@ L.Control.UIManager = L.Control.extend({
 	mobileWizard: null,
 	documentNameInput: null,
 	blockedUI: false,
+	refreshingNotebookbar: false,
 	busyPopupTimer: null,
 	customButtons: [], // added by WOPI InsertButton
 	previousTheme: null,
@@ -457,6 +458,9 @@ L.Control.UIManager = L.Control.extend({
 
 	refreshNotebookbar: function() {
 			var selectedTab = $('.ui-tab.notebookbar[aria-selected="true"]').attr('id') || 'Home-tab-label';
+
+			this.refreshingNotebookbar = true;
+			var wasNotebookbarLoadedOnCore = this.notebookbar && this.notebookbar._isNotebookbarLoadedOnCore;
 			this.removeNotebookbarUI();
 			this.createNotebookbarControl(this.map.getDocType());
 			if (this._map._permission === 'edit') {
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -457,6 +458,9 @@ L.Control.UIManager = L.Control.extend({
 
 	refreshNotebookbar: function() {
 			var selectedTab = $('.ui-tab.notebookbar[aria-selected="true"]').attr('id') || 'Home-tab-label';
+
+			this.refreshingNotebookbar = true;
+			var wasNotebookbarLoadedOnCore = this.notebookbar && this.notebookbar._isNotebookbarLoadedOnCore;
```
eszkadev, line 463: there is method for that: `notebookbar.isInitializedInCore()` to not expose internals outside

eszkadev (changes requested):
it breaks widgets in notebookbar, repro:
1. open writer
2. switch mode light/dark
3. switch notebookbar to home tab, try font selector, style previews, font size
Result: none of the widgets received update from the core so they are empty/have initial values

We would need to cache last values and inject into "template" from which they are created, or force widgets to send update message.

eszkadev:
btw, do we need to refresh notebookbar on theme switch at all?
I think it will apply dark theme automatically because different CSS rules will be applied. I don't think any widget has different content after rebuild. Maybe we can just skip it.

caolanm:
Checking just simply skipping this.refreshNotebookbar in refreshAfterThemeChange on a theme switch, then the notebookbar has the wrong icons. Looks to me that we set these icons with an explicit "dark" in their path (or not) during the building and they don't have css rules for dark/light so we have to destroy and recreate the browser notebookbar to get the right icons. If we then don't recreate the core one then the browser one has the empty menus etc as mentioned above.

var icon = data.icon ? data.icon : builder._createIconURL(data.command); is where we set the icons

eszkadev:
I see, I have idea how to reload icons without rebuilding notebookbar. I will try to do that.

eszkadev:
Reload of icons on theme change: https://github.com/CollaboraOnline/online/pull/7513

caolanm:
https://github.com/CollaboraOnline/online/pull/7513 is better, so close this

PR #7497, opened 2023-10-20
Private/caolanm/coverity29

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -312,7 +312,7 @@ private:
             headerLen += 8;
         }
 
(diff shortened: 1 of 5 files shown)


vmiklos (approved):
Makes sense, one part puzzles me, see inline comment.

caolanm:
The copy the warning about in this case is from the [=] to capture by value. Afterwards we are still copying by value, but a smaller std::string copy instead.

vmiklos:
I see, thanks!

PR #7500, opened 2023-10-20
Prevent mouse button up event to click on menu item.

GulsahKose:
In some case[*] when we right click on row header and release mouse button, first element of the context menu is invoked. To prevent this, I planned just prevent the mouseup event on context menu items but it prevented to do click event too.

[*] 1920x1080 resolution with %75 browser zoom

Change-Id: I28299e7a7cf83eaed27ef4ed6c3555fb2cc80682

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.ColumnHeader.ts
+++ b/browser/src/control/Control.ColumnHeader.ts
@@ -143,6 +143,8 @@ export class ColumnHeader extends Header {
 		this.context.strokeStyle = this._borderColor;
 		this.context.lineWidth = app.dpiScale;
 		this.context.strokeRect(startX - 0.5, 0.5, entry.size, this.size[1]);
+
+        this._map._contextMenu.stopRightMouseUpEvent();
 	}
 
 	getHeaderEntryBoundingClientRect (index: number): Partial<DOMRect> {
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -296,7 +296,19 @@ L.Control.ContextMenu = L.Control.extend({
 		}
 
 		return contextMenu;
-	}
+	},
+
+    // Prevents right mouse button's mouseup event from triggering menu item accidentally.
+    stopRightMouseUpEvent: function() {
+        var menuItems = document.getElementsByClassName('context-menu-item');
+
+        for (var i = 0 ; i < menuItems.length; i++) {
+            menuItems[i].addEventListener('mouseup', function(e) {
+                if (e.button == 2) // Is a right mouse button event?
+                    e.stopPropagation();
+            });
+        }
+    }
 });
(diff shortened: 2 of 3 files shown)


GulsahKose:
To reproduce bug everytime you can do
rightclick on row header
hold the mouse button
release when the pointer on menu item.

It shouldn't run the menu item but runs

pedropintosilva:
re-triggering the cypress...

gokaysatir (commented):
I think these changes effect more than 2 context menus. It seems that this can effect every context menu.

Though i agree with having a function that fixes all, it'd be better to put it in a generic place. Not into a specific file that does a specific job.

We also need to check other effected context menus if they work well after the change.

pedropintosilva:
@gokaysatir @GulsahKose minutes from 23 Oct
* Gokay suggests to use just the mouseUp event of right mouse button
* Gulsah has updated the PR

pedropintosilva:
> I think these changes effect more than 2 context menus. It seems that this can effect every context menu.
> 
> Though i agree with having a function that fixes all, it'd be better to put it in a generic place. Not into a specific file that does a specific job.
> 
> We also need to check other effected context menus if they work well after the change.

@GulsahKose
* please check other context menus and see if they are affected
* The method is defined in a generic place so maybe we just need to call it from a more generic place also , right @gokaysatir ?  (I wonder where)

GulsahKose:
Moved the function into Control.ContextMenu.js that I think more generic and related.
It affects all context-menu-items that already exist during creation of the row-col header context menus. But it seems not problematic to me. right mouse button up shouldn't trig any of them.
But to call the function i couldn't find a better place.

gokaysatir:
@pedropintosilva @GulsahKose  Thanks for digging the issue :) I think we can get this in. We have the notes for the functions.

@pedropintosilva my guess would be adding required functionality to callbacks or directly into the control class itself. Maybe we can turn back to this improvement in the future for a different solution. For now, since we are calling a function (not solving the issue while we create the menus), i think calling it from these places are good.

Thanks :)

gokaysatir (approved):

browser/src/control/Control.ColumnHeader.ts
```diff
@@ -143,6 +143,8 @@ export class ColumnHeader extends Header {
 		this.context.strokeStyle = this._borderColor;
 		this.context.lineWidth = app.dpiScale;
 		this.context.strokeRect(startX - 0.5, 0.5, entry.size, this.size[1]);
+
+        this._map._contextMenu.stopRightMouseUpEvent();
```
eszkadev, line 147: this is repeated a lot of times and always adds new event listeners - causing to clow down calc with every time a bit...

PR #7501, opened 2023-10-20
jsdialog: treegrid

eszkadev:
Use pure CSS to make margins for different levels.
Inspired by: https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/#ex_label
    
    - reuse arrow from tree view and expander widget
    - center verticaly entries in the tree grid / list view
    - add pointer cursor for expander arrow
    - do not allow to sort tree grid - it makes no sense
    
Tree grid is a kind of table with headers but rows can be expanded
and have subentries with different "level" in the tree.

It can be tested in Writer -> Review -> Manage Changes

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -370,6 +370,7 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	overflow-x: hidden;
 	max-width: 1000px;
 	min-width: 150px;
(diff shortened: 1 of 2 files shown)


eszkadev:
managechanges

aszucs3 (approved):
Thank you.
It is much better now. :)

PR #7509, opened 2023-10-23
a11y: impress: help screen reader to report shape selection

mcecchetti:
Avoid screen reader to wrongly report text when a shape or image is
selected:
- Got editable area to be made empty when user is not editing text
- Got default for any input to be prevented (except for some special
cases) when user is not editing text so editable area is kept empty

The selection action and the selected object name (e.g. "Rectangle",
"Presentation Title", etc.) are sent to the client.
That allows screen reader to report: "Presentation Title selected" or
"Rectangle unselected", according to the action type.
Selection text content is reported too when available.

Something alike is reported on cell navigation in a spreadsheet.

Change-Id: I75a8b66ef8cb7b24b28d749f0b24afe2587de45e

### DIFF
--- a/browser/src/layer/marker/A11yTextInput.js
+++ b/browser/src/layer/marker/A11yTextInput.js
@@ -19,6 +19,11 @@ L.A11yTextInput = L.TextInput.extend({
 		// Used for signaling when in a mobile device the user tapped the edit button
 		this._justSwitchedToEditMode = false;
 
+		// Used when editing a shape text content.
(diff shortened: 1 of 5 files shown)


gokaysatir (approved):
I approve that this works as intended.

mcecchetti:
Cypress failures (5) are puzzling. They are probably due to my patch, but locally tests are passing, which makes difficult to debug them.

PR #7511, opened 2023-10-23
wsd: log TCP_NODELAY error only once

Ashod:
It seems that when TCP_NODELAY fails to be set,
the failure is permanent. As such, there is no
point in filling the logs with the same error.

This patch logs the error only once, per process,
and supresses further logs from Socket::setNoDelay().

Change-Id: I52c6b8cca35a8c281b4c4639d61a7e2521775d49

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -203,7 +203,14 @@ public:
 #if !MOBILEAPP
         const int val = 1;
         if (::setsockopt(_fd, IPPROTO_TCP, TCP_NODELAY, (char *) &val, sizeof(val)) == -1)
-            LOG_SYS("Failed setsockopt TCP_NODELAY: " << strerror(errno));
+        {
+            static std::once_flag once;
+            std::call_once(once,
+                           [&]() {
+                               LOG_SYS("Failed setsockopt TCP_NODELAY. Will not report further "
+                                       "failures to set TCP_NODELAY");
+                           });
+        }
 #endif
     }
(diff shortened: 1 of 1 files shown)


net/Socket.hpp
```diff
@@ -203,7 +203,14 @@ class Socket
 #if !MOBILEAPP
         const int val = 1;
         if (::setsockopt(_fd, IPPROTO_TCP, TCP_NODELAY, (char *) &val, sizeof(val)) == -1)
-            LOG_SYS("Failed setsockopt TCP_NODELAY: " << strerror(errno));
+        {
+            static std::once_flag once;
+            std::call_once(once,
+                           [&]() {
+                               LOG_SYS("Failed setsockopt TCP_NODELAY. Will not report further "
```
Ashod, line 210: `errno` is already captured and logged by `LOG_SYS`, so I removed the duplicate `strerror`.

caolanm (approved):
fine by me, alternatively if we don't care at all if it works or not (void)::setsockopt will keep coverity quiet about an unchecked return

Ashod:
> fine my me, alternatively if we don't care at all if it works or not (void)::setsockopt will keep coverity quiet about an unchecked return

I think it could be useful to know when it failed and why. It's just that too many repetitive error logs is too noisy and a cause for complaint. (I didn't know that it failed on my VM, for example; EOPNOTSUPP.)

PR #7512, opened 2023-10-23
Misc Improvements and Kill Poco

Ashod:
- wasm: improved build instructions in readme
- wsd: log TCP_NODELAY error only once
- browser: add wasm.html to load documents via wasm
- wsd: http: use named HTTP status-code instead of naked int
- wsd: http: no need to pass the mime-type explicitly to sendFile
- wsd: helper to serialize pairs
- killpoco: replace HTTPResponse in sendFileAndShutdown
- wasm: support serving wasm files

### DIFF
--- a/browser/html/wasm.html
+++ b/browser/html/wasm.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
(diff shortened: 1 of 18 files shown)


mmeeks (approved):
All looks good, lets get it in ...

PR #7513, opened 2023-10-23
jsdialog: change icon theme without reloading whole components

eszkadev:
Previously we reloaded sidebar or notebookbar on theme change because we wanted to load icons to match new colors. This change makes icons aware of theme change so they automatically switch the URL. Reload of notebookbar or sidebar is not needed then.


caolanm:
Ah, really great. Not triggering a core notebookbar reload on this theme change is the ideal outcome for me.

caolanm (approved):

PR #7514, opened 2023-10-23
Idle status postmessages

eszkadev:
postmessage: handle Get_User_State
    
Get_User_State postmessage is used to query user status.
If user is active or idle and how many seconds ago there
was registered last activity.
    
{
    state: 'idle',
    elapsed: 2.3
}

-------------------------

postmessage: send User_Active when become active again
    
We send User_Idle when user become idle, but there was no
similar message when user become active again.

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -13,12 +13,18 @@ declare var postMobileMessage: any;
 class IdleHandler {
     _serverRecycling: boolean = false;
     _documentIdle: boolean = false;
+	_lastActivity: number = Date.now();
     _active: boolean = true;
     map: any;
-	dimId: 'inactive_user_message';
+	dimId: string = 'inactive_user_message';
 
 	isDimActive(): boolean {
-		return !!document.getElementById(this.dimId);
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):
Code looks good, but haven't tested yet 👍

pedropintosilva:
Waiting for @Rash419 to be merged

Rash419 (approved):
Works as expected : )

Rash419:
Should we also add button for 'Get_User_State' in our framed.doc.html ?

juliushaertl:
@eszkadev Quick followup question for integrating:

> Get_User_State postmessage is used to query user status.
> If user is active or idle and how many seconds ago there
> was registered last activity.

Can you tell what exactly last activity means in this case? Is it bound to mouse/keyboard events and is it also sent back if the user is still considered active?

eszkadev:
Yes, it is bound to mouse / keyboard activity. Timer is always reset on any activity and Get_User_State_Resp returns time elapsed between that time and Date.now()

hfiguiere:
> Should we also add button for 'Get_User_State' in our framed.doc.html ?

Done in https://github.com/CollaboraOnline/online/pull/7876

PR #7520, opened 2023-10-24
These warnings are frequent but harmless

caolanm:
Change-Id: Id2517f310a1df71dc1357d103ea074bdd12e7e65

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -207,8 +207,9 @@ public:
             static std::once_flag once;
             std::call_once(once,
                            [&]() {
-                               LOG_SYS("Failed setsockopt TCP_NODELAY. Will not report further "
-                                       "failures to set TCP_NODELAY");
+                               LOG_WRN("Failed setsockopt TCP_NODELAY. Will not report further "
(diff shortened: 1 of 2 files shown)


net/Socket.hpp
```diff
@@ -207,7 +207,7 @@ class Socket
             static std::once_flag once;
             std::call_once(once,
                            [&]() {
-                               LOG_SYS("Failed setsockopt TCP_NODELAY. Will not report further "
+                               LOG_WRN("Failed setsockopt TCP_NODELAY. Will not report further "
```
Ashod, line 210: In this case we will need to put back `strerror(errno)`, since `LOG_WRN` wouldn't do that.
```
LOG_WRN("Failed setsockopt TCP_NODELAY. Will not report further failures to set TCP_NODELAY: " << strerror(errno));
```
(Let clang-format do its thing.)

Ashod (approved):
Thanks @caolanm. One note about logging `errno`, but otherwise looks good.

PR #7531, opened 2023-10-25
postmessage: Don't output error for handled UI message before ready

hfiguiere:
Change-Id: Id357c54bce57464334a24115175087e4bbfa4aea

* Resolves: # 
* Target version: master 

### Summary

The warning is confusing if the UI message has been handled.

@eszkadev This is a follow up on your change.

### TODO

- [ ] ...

### Checklist


hfiguiere:
Maybe an alternative could be to return after each handled.

hfiguiere:
> Maybe an alternative could be to return after each handled.

I have now done that.

eszkadev (approved):

PR #7532, opened 2023-10-26
wsd: config cleanup and warning fixes

Ashod:
Change-Id: Idb66227d98f8fcca35307c3598a1fcc6b69e38c6

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -908,6 +908,7 @@ std::string COOLWSD::RouteToken;
 bool COOLWSD::SingleKit = false;
 bool COOLWSD::ForceCaching = false;
 #endif
+COOLWSD::WASMActivationState COOLWSD::WASMState = COOLWSD::WASMActivationState::Disabled;
 #endif
 std::string COOLWSD::SysTemplate;
 std::string COOLWSD::LoTemplate = LO_PATH;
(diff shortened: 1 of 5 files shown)


wsd/COOLWSD.cpp
```diff
@@ -2073,8 +2074,6 @@ void COOLWSD::innerInitialize(Application& self)
         { "hexify_embedded_urls", "false" },
         { "experimental_features", "false" },
         { "logging.protocol", "false" },
-        { "logging.anonymize.filenames", "false" }, // Deprecated.
```
Ashod, line 2076: These were causing a warning at every startup.

wsd/COOLWSD.cpp
```diff
@@ -2381,6 +2382,26 @@ void COOLWSD::innerInitialize(Application& self)
     if (getSafeConfig(conf, "storage.wopi.reuse_cookies", reuseCookies))
         LOG_WRN("NOTE: Deprecated config option storage.wopi.reuse_cookies - no longer supported.");
 
+    COOLWSD::WASMState = getConfigValue<bool>(conf, "wasm.enable", false)
```
mmeeks, line 2385: Looks like this won't compile since the variable is declared in ENABLE_DEBUG only (?) - I hate the conditionals - so perhaps better to pull it out of that conditional.

mmeeks (approved):
Looks fine when it compiles outside of ENABLE_DEBUG =)

PR #7538, opened 2023-10-26
sidebar: fix paragraph panel layout

eszkadev:
This fixes regression from commit c77f1041a58cd4606c69a0808888dc16dd654785 jsdialog: implement horizontal separator

Where horizontal separator widget was implemented
and now it appears in the sidebar, but as we have grid layout there - it occupies only half of the sidebar because we have 2 columns.

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -720,6 +720,13 @@ td[id^='tb_spreadsheet-toolbar_item']:focus table.w2ui-button div.w2ui-icon, td[
 #mobile-wizard.popup.snackbar #button {
 	font-size: 14px !important;
 }
+/* Separators */
+.mobile-wizard.ui-separator {
+	/* Hide separators coming from core when on mobile
+	since we are already adding border top on every
+	#mobile-wizard .mobile-wizard.ui-text */
+	display: none;
+}
 /*   - Import text */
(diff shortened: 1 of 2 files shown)


eszkadev:
a

eszkadev:
after fix:
a

pedropintosilva:
Thanks, the space is fixed. But why we need t have the separator in the sidebar?

image

pedropintosilva (changes requested):
There are more places where this happens 
image


I would prefer to simple avoid additional separators altogether in the sidebar

```css

.jsdialog.sidebar.ui-separator.horizontal.ui-grid-cell {
  display: none;
}
```

eszkadev:
We don't need, but it was always there (in the JSON generated by the core). Now as control is visible - it is there...
If you want we can hide it.

pedropintosilva (approved):

grandinj:
> There are more places where this happens image
> 
> I would prefer to simple avoid additional separators altogether in the sidebar
> 

@pedropintosilva possibly you should engage with the libreoffice design team, then this could be fixed in core?

pedropintosilva:
@grandinj : thanks for the reminder but I'm still not sure if in the long run we don't want/need them. So, and as per my code comments, better to hide them for now until we either style it better etc

PR #7539, opened 2023-10-26
missing continue keyword

timar:
Caused log ERR on startup:
[ websrv_poll ] ERR  unknown UI default's component UITheme| wsd/FileServerUtil.cpp:99

Change-Id: I82f036e393cfb271e2dec2c52c5dc512aed265c8


gokaysatir (approved):
I still couldn't reproduce the error. The lines below continue doesn't seem related to UITheme.

Issue #7527, opened 2023-10-25
Comments overlapping so i can't edit them


pedropintosilva:
Reported initially by @jospoortvliet https://github.com/nextcloud/richdocuments/issues/3254

----- 

**Describe the bug**
See this:
![Screenshot 2023-10-24 at 15 57 21](https://github.com/nextcloud/richdocuments/assets/551757/f2c3be2c-4ae1-4bf8-8b71-ae13ca600607)

Says it all, I guess. Firefox, on Mac OS, both latest version, on cloud.nextcloud.com

Happy to share the document.

jospoortvliet:
Thanks!

PR #7540, opened 2023-10-26
annotation: forcefully reselect comment before modifying

lpranam:
select and unselect usually updates the layout but,
in case if it failed we manually do the layout update

problem:
when already selected comment try to modify,
it will not rearrange replies in the thread this caused overlapping of the comments and bad user experience

fixes: #7527

Change-Id: Id56a5c35795e1ff75e8c3a92032afce6b8585fcb

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -589,9 +589,13 @@ export class CommentSection extends CanvasSectionObject {
 			}
 		}
 		else {
-			this.unselect();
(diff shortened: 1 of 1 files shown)


Rash419 (changes requested):
still in some cases there is overlap

Rash419 (changes requested):
I can still reproduce a overlapping case when replying to the comment

Rash419 (approved):
Works as expected, thanks

PR #7541, opened 2023-10-26
apply delayed messages after docLayer.map is set

caolanm:
which is done by addLayer, otherwise its possible to handle a onViewInfoMsg before map is set

Change-Id: If6e6d15c90de129614bcc2b64705d0b51b5c4e83

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1377,11 +1377,11 @@ app.definitions.Socket = L.Class.extend({
 				});
 			}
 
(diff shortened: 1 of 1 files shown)


caolanm:
@Rash419 you could reproduce this, does this solve it?

Rash419 (approved):
Tested, fixes the the problem for me! Thanks

Issue #7334, opened 2023-09-29
calc perf: Note listing


mmeeks:
From this week's calc flamegraph:

https://user-images.githubusercontent.com/833656/271274520-6345ee72-2a06-402e-a12c-72bd8d49df8d.svg

Shows a chunk of time when users join the sheet is interrogating the whole document as to whether notes are present:

![image](https://github.com/CollaboraOnline/online/assets/122848/6f266a80-7f2b-444d-abdc-6a2f1c609544)

6% of the time - is searching in empty columns far outside the range of the sheet that contains data I expect - in order to see if those columns have notes - inevitably they do not. 16000 is a lot of columns to check to see if there is something there - when there really isn't almost all the time.

I wonder if we should tweak our implementation to shrink the real sheet column and row count to something more reasonable - and then expand that on user interaction: adding row/columns with some '+' button, or pasting into them in some way (?) though that's a bigger pain.

Failing that, having and maintaining some 'last note column' or other better data-structure for notes than columnar storage might be helpful.

mmeeks:
There is a ridiculous amount of this search going on; I assume that the result is:

1697730362525 INCOMING: commandvalues: { "commentsPos": [ { "id": 1, "tab": 0, "cellRange": "0 0 0 0"}]} 

I seem to get this when switching tab, but worse - when I move between cells with the arrow keys (perhaps just when scrolling the window) - that's a pretty heinous cost to pay for this small operation - if that is indeed what's going on :-) @caolanm I would assume there is some legacy fetching or pushing going on there from when notes were pushed not in their entirity but cropped to the viewpane or (?).

mmeeks:
Seems like we do far too much work as we extend the document - by moving into 'blank space' - that's when we emit (from the client) this OUTGOING: commandvalues command=.uno:ViewAnnotationsPosition which is something of an anomaly: much of the data we need is helpfully pushed:

1697732170923 OUTGOING: key type=up char=0 key=1031
keyboard handler: keydown PageDown 0 undefined KeyboardEvent {isTrusted: true, key: 'PageDown', code: 'PageDown', location: 0, ctrlKey: false, …}
1697732170997 OUTGOING: key type=input char=0 key=1031

... getting page down ...

1697732147269 INCOMING: invalidateheader: all
1697732147269 OUTGOING: commandvalues command=.uno:ViewAnnotationsPosition	-> why is this not pushed ? ...
1697732147270 INCOMING: cellselectionarea: EMPTY
[...]

PR #7546, opened 2023-10-27
we don't need to round-trip through core to reposition notes

caolanm:
we know where they are by cell addresses, so defer getting the screen positions until we need them and we don't need to get core to trigger recalculating them. When we redraw the comments after the new geometry arrives then we can place them via the address.

When adding a new note we want to know the range of the potentially merged cells we are inserting into.

https://github.com/CollaboraOnline/online/issues/7334

https://gerrit.libreoffice.org/c/core/+/158560 needs to be applied to solve the problem described there that becomes apparent when this is in place.

Change-Id: I3228dc8fa8d47ba4e796e50427c125d7f78fe5fc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -21,9 +21,11 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		var commentList = app.sectionContainer.getSectionWithName(L.CSections.CommentList.name).sectionProperties.commentList;
 		var comment = null;
 
+		var cursorAddress = this._cellCursorXY;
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Looks fine to me; and we're robust over inserting / removing columns, and expanding/contracting row&column grouping and so on I assume ? =)

caolanm:
grouping/ungrouping looks ok, inserting rows and cols not so much. So https://gerrit.libreoffice.org/c/core/+/158718 to tell the client when a note has moved to another cell

PR #7547, opened 2023-10-27
Menu: Fix separators, chevron and focus

pedropintosilva:
- JSdialog: Fix ui-separator visual importance
- Multi-level menus (dropdowns): Fix chevron elements
- Menus (dropdowns) with separator: Fix hidden keyboard focus

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -218,6 +218,9 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 
 /* Separator */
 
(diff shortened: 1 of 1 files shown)


eszkadev:
looks much better, but still if 1st or last element on the list is focused, we see only 1 border line, not both on top and bottom

eszkadev (approved):

pedropintosilva:
> looks much better, but still if 1st or last element on the list is focused, we see only 1 border line, not both on top and bottom

True but we always had like that and also in other widgets I think. I can look into that but I prefer to do that in a different PR because it requires more testing (not just in that case)

PR #7551, opened 2023-10-27
apply delayed messages after docLayer.viewid is set

caolanm:
which is done by docLayer._onStatusMsg(), otherwise its possible to handle a onViewInfoMsg before viewid is set

Change-Id: Id68dba5f5cdc8efd0b5f6ed429ac4f105cef4866

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
Thanks

PR #7552, opened 2023-10-27
annotation: fix parent comment being deleted if reply cancelled

lpranam:
Change-Id: I080f817a3020249de4eefd12b3d23af1141a273d

* Target version: master 

### Checklist


Rash419 (approved):
works as expected, thanks

PR #7553, opened 2023-10-27
Fix: Select of content in Formula Bar is hard as it changes position

pedropintosilva:
- Fix illegal HTML id `autosummenu:AutoSumMenu`
- Update CSS rules: img if doesn't exist anymore, apply margin reset
to new button instead
- Also Add missing css class, we used to have that before the
Accessibility commit 1a2500c

Change-Id: If4c2fa9ff99a6c09f42d6ba4db3f1dd186e099ea

### DIFF
--- a/browser/src/control/Control.FormulaBar.js
+++ b/browser/src/control/Control.FormulaBar.js
@@ -139,7 +139,7 @@ L.Map.include({
 			target = mobileTopBar;
 		} else {
 			jsdialogFormulabar.hide('startformula');
-			jsdialogFormulabar.hide('AutoSumMenu');
+			$('#AutoSumMenu').hide();
 		}
 		target.show('cancelformula');
 		target.show('acceptformula');
@@ -165,11 +165,11 @@ L.Map.include({
 			mobileTopBar.show('undo');
 			mobileTopBar.show('redo');
 
-			$('#AutoSumMenuimg').css('margin-inline', '0');
+			$('#AutoSumMenu-button').css('margin-inline', '0');
 			$('#AutoSumMenu .unoarrow').css('margin', '0');
 
 			jsdialogFormulabar.show('startformula');
-			jsdialogFormulabar.show('AutoSumMenu');
+			$('#AutoSumMenu').show();
 
 			// clear reference marks
 			map._docLayer._clearReferences();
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.FormulaBar.js
```diff
@@ -139,7 +139,7 @@ L.Map.include({
 			target = mobileTopBar;
 		} else {
 			jsdialogFormulabar.hide('startformula');
-			jsdialogFormulabar.hide('AutoSumMenu');
+			$('#AutoSumMenu').hide();
```
pedropintosilva, line 142: not happy with this... would prefer to re-use the hide function but then again I don't know why we have in the same file jquery methods and then this. Nevertheless using out custom hide and place there the uno command didn't work

browser/src/control/Control.FormulaBarJSDialog.js
```diff
@@ -50,7 +50,8 @@ L.Control.FormulaBarJSDialog = L.Control.extend({
 							command: '.uno:FunctionDialog'
 						},
 						{
-							id: 'autosummenu:AutoSumMenu',
+							id: 'AutoSumMenu',
```
eszkadev, line 53: this will break menu, it will not work
after ":" we provide id of the menu which should be used

eszkadev: in the past we removed ":" from the id of HTML element, it was only used to identify menu...
I think that was changes and caused regression

eszkadev:
```
{
	id: 'autosummenu:AutoSumMenu',
	type: 'menubutton',
	command: '.uno:AutoSumMenu'
},
```
previously it used `command` because `id` was not unoCommand
(https://github.com/CollaboraOnline/online/commit/1a2500c8d31f12a8e23c648a0afb208e3d09bb05#diff-1951654bafeb4ed39d89489812ff8da269c905feabbe179ccfd312d1f2d19fb6L2485):

Proper fix should be:
1. if `id` has `:` - use first part before `:` as the id of the HTML element
2. change JSON `id: 'autosummenu:AutoSumMenu'` to `id: 'AutoSumMenu:AutoSumMenu'` so it will receive the same id as before

browser/src/control/Control.FormulaBar.js
eszkadev: it doesn't work due to my commit:
commit a7ece5d0d615ca1857b4d020844bfa098cf9fb07
jsdialog: reduce warnings in console for formulabar

I thought it is unwanted core message...

eszkadev (changes requested):
I propose other fix: https://github.com/CollaboraOnline/online/pull/7554
fixing root cause

Issue #7548, opened 2023-10-27
Select of content in Formula Bar is hard as it changes position


timur-g:
Online 23.05.5. 

Select of content in Formula Bar is hard as it changes position.
Best seen if ew double-click to select 1st word, here "test". If we select 3rd word "word" then 2 words are selected.

![image](https://github.com/CollaboraOnline/online/assets/7240617/e007b641-5f66-4ed4-b8bd-868e3a14c92a)

OK with 23.05.2, NOK with 23.05.4, regression. OK in Core, so Online issue.

timur-g:
commit 1a2500c8d31f12a8e23c648a0afb208e3d09bb05
Author: Gökay Şatır <gokaysatir@gmail.com>
Date:   Tue Aug 8 15:46:49 2023 +0300

    Merge accessibility definitions with notebookbar definitions.
    
I confirm by bisect.

eszkadev:
Fix merged

PR #7554, opened 2023-10-27
jsdialog: fix id for buttons with menu #7548

eszkadev:
- fixes regression from Accessibility commit 1a2500c
- it fixes formulabar control of AutoSum button #7548
- removes useless commit a7ece5d: jsdialog: reduce warnings in console for formulabar
- id with ":" contains "id : menuId", so we can identify menu which should be used, let's not use whole string as HTML element id - we don't want ":" there

### DIFF
--- a/browser/src/control/Control.FormulaBar.js
+++ b/browser/src/control/Control.FormulaBar.js
@@ -165,7 +165,7 @@ L.Map.include({
 			mobileTopBar.show('undo');
 			mobileTopBar.show('redo');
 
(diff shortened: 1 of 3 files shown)


pedropintosilva (changes requested):
This doesn't fix the ID being malformed (that's why I didn't simply change the case) and the css rules are not targeting the new button. See  #7553

eszkadev:
Good catch, fixed

pedropintosilva (approved):

PR #7556, opened 2023-10-27
postmessage: added message to collapse / extend notebookbar

hfiguiere:
The new messages are: Collapse_Notebookbar and Extend_Notebookbar As a side effect they also hide the classic toolbar the same way

Change-Id: Ic9d04876acb06f2885a6be1e171df7f87e513ed8

* Resolves: # 
* Target version: master 

### Summary

Add two postmessage to show hide the notebookbar. Due to the UI layout, they also hide
the classic toolbar in that mode.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -139,6 +139,12 @@
               'Values': {'Mode': value}});
       }
 
+      function CollapseNotebookbar(collapse) {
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Looks great to me thanks Hub ! =)

Issue #7166, opened 2023-08-25
JS: store damaged areas in incoming tiles since last render


mmeeks:
When a tile arrives and we apply it in browser/src/layer/tile/CanvasTileLayer.js - we should track initially which tiles have changed.

Then when we re-render over the visible HTML canvas object (essentially just a bitmap) - we should re-blit only those tiles in the bbox of all changes - prolly we can use a single simple rectangular clip of a bbox of all tile changes there.

This -should- queue much less work for the GPU to do, and potentially the browser can also damage track and composite a smaller section of the layers it works with. Most likely we will also need to do some JS clipping of what tiles are visible to make this perform nicely.

For bonus points we should get _applyDeltaChunk to track the actual pixels inside the tile that really changed if we have a delta - often that's a very small area indeed.

Thanks !

PR #7560, opened 2023-10-29
We don't always need to redraw the entire canvas

caolanm:
If we end up dirtying only due to a series of TileRange reasons then we only need to redraw the bounds of those tiles.

https://github.com/CollaboraOnline/online/issues/7166

checked: split calc sheets and rtl calc sheets

Change-Id: If468b735bdff85408155fb23ebf6db891a449d5e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -21,7 +21,7 @@ interface SectionCallbacks {
 	onMultiTouchMove?: (point: Array<number>, dragDistance: number, e: TouchEvent) => void;
 	onMultiTouchEnd?: (e: TouchEvent) => void;
 	onResize?: () => void;
-	onDraw?: (frameCount?: number, elapsedTime?: number) => void;
+	onDraw?: (frameCount?: number, elapsedTime?: number, subsetBounds?: cool.Bounds) => void;
 	onDrawArea?: (area?: cool.Bounds, paneTopLeft?: cool.Point, canvasContext?: CanvasRenderingContext2D) => void;
 	onNewDocumentTopLeft?: (size: Array<number>) => void;
 	onRemove?: () => void;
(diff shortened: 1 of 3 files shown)


browser/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -2028,16 +2052,36 @@ class CanvasSectionContainer {
...
 
+		var subsetBounds: cool.Bounds = null;
+		// if there is a tileSubset we only want to draw the miniumum region of its bounds
+		if (tileSubset) {
+			var tileSection = <TilesSection>(this.getSectionWithName(L.CSections.Tiles.name));
+			if (tileSection && this.shouldDrawSection(tileSection)) {
+				subsetBounds = tileSection.getSubsetBounds(this.context, tileSubset);
+			}
+			if (subsetBounds) {
+				this.context.save();
```
mmeeks, line 2070: My recollection is (but it is fuzzy) - that it paid dividends performance-wise not to do save/restores - and to avoid clipping wherever we could, and to not translate the canvas - but do the arithmetic ourselves. Worth checking if that is still so - hopefully not.

mmeeks (commented):
Looks good to me - did you test with RTL and also test with Calc's split-panes - which add some funkiness there :-)

caolanm:
> Looks good to me - did you test with RTL and also test with Calc's split-panes - which add some funkiness there :-)

Yeah, the Calc RTL sheet stuff necessitates the beforeDraw/afterDraw stuff to get that working. Split panes example behaves for me correctly when testing by debugging through the clipped clear background and repaints. Just the tile around changed char blanked out and repainted afterwards.

mmeeks (approved):
Looks good then - thanks =)

browser/src/layer/tile/CanvasSectionContainer.ts
caolanm: The translate is easily removable if needs be. If it was always just a solid background we could possibly not need a clip, but for calc we have that grid which gets drawn before we draw the cells so I'm not sure we can live without any clipping.

PR #7564, opened 2023-10-30
annotation: fix parent comment being deleted if reply cancelled

lpranam:
part 2

fix from #7552 stopped working, this patch amends it

Change-Id: I8c1227199112599fc2ade4e425452709e9cd05e5

* Target version: master 

### Checklist


Rash419 (approved):
Looks good, works as expected!

PR #7565, opened 2023-10-30
annotation: fix cancelled modification of autosaved comment

lpranam:
problem:
when a user modifies a comment and its autosaved,
if user then cancels the modification they were still added

Change-Id: I26c872f0957f0815882c4b1aedae83fe6736b7a2

* Target version: master 

### Checklist


Rash419 (approved):
Thanks

PR #7566, opened 2023-10-30
cypress: refector mismatching parameter in function call

lpranam:
added and rearranged new parameters in the reload function call, its safe because so far all the calls to reload are made using only three parameters...

Change-Id: I8fa8e4b1e20de0af16368cb23ca641b8d7186bfc

* Target version: master 

### Checklist


Ashod (approved):
Thanks @lpranam. I assume we are certain this won't break anything.

lpranam:
> Thanks @lpranam. I assume we are certain this won't break anything.

yes have checked all the instances...

PR #7570, opened 2023-10-30
Fix TypeErrors in Calc

eszkadev:
Fix Uncaught TypeError in combobox selection
    
    TypeError: Cannot set properties of undefined (setting 'selected') emitting event jsdialog: { "jsontype": "sidebar", "action": "action", "id": 0, "data": { "control_id": "applystyle", "position": "-1", "action_type": "select"}}
    
    When selecting whole column in Calc
    
----------------------------------

    Fix Uncaught TypeError in col/rows headers selection
    
    Uncaught TypeError: Cannot read properties of null (reading 'index')
        at RowHeader.onMouseUp (bundle.js:24879:87)
        at CanvasSectionContainer.propagateOnMouseUp (bundle.js:14350:24)
        at CanvasSectionContainer.onMouseUp (bundle.js:14442:6)

### DIFF
--- a/browser/src/control/Control.ColumnHeader.ts
+++ b/browser/src/control/Control.ColumnHeader.ts
@@ -254,7 +254,7 @@ export class ColumnHeader extends Header {
 			const entry = this._mouseOverEntry;
 			let modifier = 0;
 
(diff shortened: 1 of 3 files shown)


mmeeks (approved):
Looks reasonable.

PR #7573, opened 2023-10-30
net: don't try to set TCP_NODELAY on local Unix sockets.

mmeeks:
Change-Id: Ic0f720554080a01b7a46abbffda023834016c0ea

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/DelaySocket.cpp
+++ b/net/DelaySocket.cpp
@@ -48,7 +48,7 @@ class DelaySocket : public Socket {
     std::vector<std::shared_ptr<WriteChunk>> _chunks;
 public:
     DelaySocket(int delayMs, int fd) :
-        Socket (fd), _delayMs(delayMs),
+        Socket (fd, Socket::Type::Unix), _delayMs(delayMs),
         _state(ReadWrite)
 	{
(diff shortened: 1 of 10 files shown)


Ashod (approved):
Thanks @mmeeks. Looks reasonable. (Though I wish we never had Unix sockets to begin with.)

thebearon (changes requested):
Android/iOS app builds fail with this in `SocketPoll::insertNewFakeSocket(...)`:
```
  /home/collabora/jenkins/workspace/github_android_online_co-23.05/net/Socket.cpp:668:18: error: no matching function for call to 'create'
          socket = StreamSocket::create<StreamSocket>(std::string(), fd, true, websocketHandler);
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/collabora/jenkins/workspace/github_android_online_co-23.05/net/Socket.hpp:1199:37: note: candidate function template not viable: requires at least 5 arguments, but 4 were provided
      static std::shared_ptr<TSocket> create(std::string hostname, const int fd, Type type, bool isClient,
                                      ^
```

thebearon (approved):

PR #7575, opened 2023-10-30
default-ui: Allow skipping the local storage default

hfiguiere:
Pass SaveUIState=false to the ui_defaults

Change-Id: I2cceffed338440d1fee5b0818e072ecc02379468

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


hfiguiere:
Needs a bit more testing. The idea is to allow integration to always force the UI defaults.

hfiguiere:
Things like the sidebar being visible is cached in local storage. In some situations it can be desirable to always have the same layout whichever the user change in the course of use.

eszkadev (approved):
looks good

PR #7576, opened 2023-10-31
browser: wireId out of sync on reconnect

Rash419:
- regression from 762ce28e22f4c11e0f4d7c2ba0ba66144586f11c

Change-Id: I86b913213135982031d33b4b29f54df2e8320ea4

* Target version: master


mmeeks (changes requested):
I would set them both to 1 - I believe '0' is likely to be a special state. If the ids are the same it is in invalid; can you fix & re-validate ? =) thanks.

quikee (approved):

PR #7579, opened 2023-10-31
Exit cell edit mode before add a new sheet

GulsahKose:
To exit cell edit mode we send a KEY_RETURN event. That prevents to move last edited cell to new sheet.

Change-Id: I115be45c170789011d51d4d7914bd4825efb48f3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Parts.js
+++ b/browser/src/control/Parts.js
@@ -278,6 +278,11 @@ L.Map.include({
 	},
 
 	insertPage: function(nPos) {
+		// Make sure we exit cell edit mode.
+		if (this._docLayer.insertMode) {
+			this._docLayer.postKeyboardEvent('input', 0, 1280);
+		}
+
 		if (this.isPresentationOrDrawing()) {
 			app.socket.sendMessage('uno .uno:InsertPage');
 		}


pedropintosilva:
@GulsahKose what's the status of this PR. Would be awesome if we could have this fix in.

CI didn't pass but it seems unrelated writer/statusbar_spec.js   on desktop. Can you please run that test locally and see if it passes? Meanwhile you can retrigger the CI

GulsahKose:
Trying to run problematic cypess test locally with :
make -C cypress_test run-desktop spec=writer/statusbar_spec.js


But has a problem to connecting chromium browser as see:
cypress:network:connect received error on connect, retrying { iteration: 50, delay: 1000, err: Error: connect ECONNREFUSED 127.0.0.1:34555     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1187:16) { errno: -111, code: 'ECONNREFUSED', syscall: 'connect', address: '127.0.0.1', port: 34555 } } +1s

GulsahKose:
Problematic test passed locally

browser/src/control/Parts.js
```diff
@@ -278,6 +278,11 @@ L.Map.include({
 	},
 
 	insertPage: function(nPos) {
+		// Make sure we exit cell edit mode.
+		if (this._docLayer.insertMode) {
```
eszkadev, line 282: This applies only to Calc, but this Part.js is used also in Impress/Draw.
Could you move it below inside the if which checks if it is a spreadsheet?

Also question: couldn't be  that done in the core directly in handling of `.uno:Insert` command? I think it would be better, as there you have "real current state", no race could occur

GulsahKose: I agree Szymon, I'll handle this in core. Thanks

GulsahKose:
Implemented the patch in core. So I'm closing the PR thanks szymon: https://gerrit.libreoffice.org/c/core/+/160758

PR #7580, opened 2023-10-31
Stop onscreen keyboard when panning calc on tablet

Minion3665:
Before this change, panning a calc spreadsheet would cause the onscreen keyboard to pop up when in tablet mode. Dismissing the keyboard would not stop it from being popped up the next time you panned around. This made it very difficult to pan around spreadsheets in calc without either using an external keyboard, the mobile app or another device altogether.

This commit introduces a known regression: when you are on a tablet and have a physical external keyboard, you could previously select the cell and start typing. Now you must tap into the cell in the same way you would if you have an onscreen keyboard. This matches the behavior on mobile devices. I consider this regression a reasonable tradeoff, so notwithstanding I believe this should be merged.

~~A followup is planned to add a POST message so that integrators etc. can toggle this setting, as they may have better knowledge than us of whether someone is using an onscreen keyboard. I believe it's best to merge this without the followup because calc is borderline unusable on tablets (in browser, without an external keyboard). Additionally, getting this up quickly allows us to start testing it a bit earlier to find any urgent issues.~~ The POST messages have been added in this commit

### Checklist

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -120,6 +120,12 @@
             });
       }
 
+      function HintOnscreenKeyboard(hint) {
+        post({'MessageId': hint ? 'Hint_OnscreenKeyboard' : 'Hint_NoOnscreenKeyboard',
+              'Values': null
+            });
+      }
+
       function ShowMenubar(visible) {
         var messageId = visible ? 'Show_Menubar' : 'Hide_Menubar';
         post({'MessageId': 'Host_PostmessageReady'});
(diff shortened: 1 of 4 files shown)


hfiguiere:
If the regression is acceptable, LGTM, but ideally we should followup with a fix for the regression later on.

hfiguiere (approved):

Minion3665:
06539d7ae3bda4508fda0e7764da8597f36fb9a7..f62f365a0ed26b21e1f86b0e028cf80fc6b2b9df refactors out some window.ThisIsAMobileApp in TextInput.js, as `guessOnscreenKeyboard` defaults to `window.ThisIsAMobileApp || ...`. This is not entirely equivalent as it means that if we get a postmessage in the mobile app it can override our "ThisIsMobileApp" setting.

I see it as a good thing to respect the postmessage :)

hfiguiere:
The gist of this is that the issue exist on iPad / iOS in Safari (not the apps) and this fixes it.
If that doesn't break the rest, then we should be able to get this in.

hfiguiere:
The postmessage works as advertised (ie it sets the hint)

hfiguiere:
as agreed, merging.

PR #7587, opened 2023-11-01
Private/caolanm/avoid clipping tile layer

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.CornerHeader.ts
+++ b/browser/src/control/Control.CornerHeader.ts
@@ -66,6 +66,10 @@ export class CornerHeader extends CanvasSectionObject {
 	onMouseLeave(): void {
 		this.containerObject.getCanvasStyle().cursor = 'default';
 	}
+
+	shouldDrawForTileBounds (tileBounds: cool.Bounds) : boolean {
+		return tileBounds === null;
+	}
 }
 
 }
--- a/browser/src/control/Control.CornerHeader.ts
+++ b/browser/src/control/Control.CornerHeader.ts
@@ -67,6 +67,8 @@ export class CornerHeader extends CanvasSectionObject {
 		this.containerObject.getCanvasStyle().cursor = 'default';
 	}
 
+	// This doesn't overlap tiles at all, so if there are tile bounds we
+	// can skip redrawing it entirely
 	shouldDrawForTileBounds (tileBounds: cool.Bounds) : boolean {
 		return tileBounds === null;
 	}
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -594,6 +594,10 @@ export class Header extends CanvasSectionObject {
 	onNewDocumentTopLeft(): void {
 		return;
 	}
+
+	shouldDrawForTileBounds (tileBounds: cool.Bounds) : boolean {
+		return tileBounds === null;
+	}
 }
 
 export interface HeaderEntryData {
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -595,6 +595,8 @@ export class Header extends CanvasSectionObject {
 		return;
 	}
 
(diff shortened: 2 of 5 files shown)


browser/src/layer/tile/ScrollSection.ts
```diff
@@ -1050,6 +1050,10 @@ export class ScrollSection extends CanvasSectionObject {
 			this.performHorizontalScroll(delta[0]);
 		}
 	}
+
+	shouldDrawForTileBounds (tileBounds: cool.Bounds) : boolean {
+		return tileBounds === null;
```
mmeeks, line 1055: Are we sure this is right? I'd expect us to check if the scroll bar (?) or somesuch intersected with tileBounds, rather than just checking for null ?

mmeeks (commented):
I like the way it's going; but from a quick skim not entirely certain why we don't want to always clear the background of the small part we re-render on the canvas (?) and/or why we use the presence of the subsetBounds as a boolean to not do things. I assume that change (perhaps from another view) can intersect with lots of  other sections somehow (?)
But the general direction is  great - of avoiding re-rendering in lots of places if we can :-)

caolanm:
> I like the way it's going; but from a quick skim not entirely certain why we don't want to always clear the background of the small part we re-render on the canvas (?) 

Well, we do want to clear the bg, redraw the grid in the case of calc, paint tiles (and redraw the comment indicators on top of the tiles). In current trunk we just set a clip, draw the bg, full grid layer, tile layer, mostly as per normal and let the clip sort it out. However there is already a preexisting mode (used in zooming apparently) where we don't draw the bg, and the tile layer itself draws the bg (and calls the grid drawing) for each tile before it does the tile itself. So this version reuses that, which has the advantage of already being aware of calc split panes, rather than adding another way of doing things.

> and/or why we use the presence of the subsetBounds as a boolean to not do things.

The thought there is that we've a bunch of layers in calc, CornerHeader, RowHeader, ColumnHeader, which don't intersect with the tile layer at all, so if we are drawing just a tile update the presence of any tile subset to redraw is enough to elide the effort. If we reuse the "tile-layer draws the grid" path, then the GridLayer falls into that group as well. The scrollbar, in hindsight, does not fall into that category getting drawn over the tile layer, rather than beside it, so that won't fly as is.

mmeeks:
@Cwiiis - would be good to know what you think of clipping as a contributor to potential performance problems and/or getting some thoughts / testing / benchmarking of this. If the win is not significant would be good to drop it from the review queue I think =)

Cwiiis:
> @Cwiiis - would be good to know what you think of clipping as a contributor to potential performance problems and/or getting some thoughts / testing / benchmarking of this. If the win is not significant would be good to drop it from the review queue I think =)

I probably need to know a bit more about this code path, but just going on a cursory reading, this doesn't seem immediately intuitive to me. As long as clipping is a pixel-grid aligned rectangle, it shouldn't be expensive and I'd be surprised to see it show up in a profile. That said, if nothing is being drawn outside of the clipping path and that can be easily determined, it's sure to be a small win to not do it I suppose? I can do some before/after profiling here.

Cwiiis:
So I'm still interpreting the profiles and it took a while to figure out a process that was consistently repeatable that gives me useful results, but I do think this patch is beneficial, at least vs the commit that it's branched from.

Profiling just drawSections, which I believe is what this patch will affect primarily, there is significantly less time spent clearing the canvas. Then, within drawing the canvas, it seems to spend less time - I can't say just from the profile exactly why just yet and this could just be variation in my testing methodology (which is pretty far from scientific).

The reduction in clearing appears to be a real benefit.

caolanm:
we can ditch this with more recent developments

mmeeks:
Thanks Caolan - shrinks the list; turns out we were rendering the same scene 5+ times each so ;-) lower hanging fruit there I think.

Issue #5957, opened 2023-03-06
Hide_Menu_Item and Hide_Button not working


ErikDW1:
Hello I’m trying to hide some menu options by using the PostMessage API, however nothing seems to work.

I’m sending these messages right after receving the “App_LoadingStatus” “Document_Loaded” feedback:
{“MessageId”:“Hide_Menu_Item”,“SendTime”:1677600490227,“Values”:{“id”:“help”}}
{“MessageId”:“Hide_Menu_Item”,“SendTime”:1677600490227,“Values”:{“id”:“file”}}

Also I made a JS function called by a button as in the https://github.com/CollaboraOnline/online/tree/629b25b27a299ae5eea182e5e58a89a10678c680/browser/html/framed.doc.html, but it's not working too.
If I call the "Hide_Menubar" and the "Action_Save" action these work seamlessly.

The version I'm working with is the latest COOLWSD version: 22.05.10.8 [ef89c82]

ErikDW1:
I want also to add that an optimal way of handling UI customization would be through the "CheckFileInfo" response properties, as it is for Save and Print options.
This way we can have the ability to customize the UI according to the users' capabilities on our hosting platform (ex. I might want to display an user just the Review controls and nothing else, since it has the role "Reviewer" on my hosting platform).

nadirlouati:
yes i confirm your diagnosis

Override CheckFileInfo is only for DownloadAsPostMessage propertery for now
https://sdk.collaboraonline.com/docs/advanced_integration.html#override-checkfileinfo

robmv:
I can confirm this problem. Even if the user has switched to the compact view, the button is hidden, but if the user switch to the notebook mode and reverts back to compact, the button is again visible.

As @ErikDW1 said, I wish too UI customization could be done at CheckFileInfo time, even locking UNO actions, these can only be done currently by host for example, not based on the specific user requesting the editing.

useresd:
I confirm this bug. The tabbed view UI is not affected by the postMessage API for Hide_Menu_Item

dan-aridhia:
I can confirm this is also in 23.05.4.

hfiguiere:
I file #7517 recently.

nadirlouati:
any news ?

Issue #7517, opened 2023-10-23
Improve postMessage interface with tabbed UI


hfiguiere:
### Describe the Bug

When using the postMessage interface, there are error that pops in the console when clicking "Hide XXX Command" or "Show XXX Command".

### Steps to Reproduce

1. Go to the post message interface as shown in the make run output.
2. In the post message page, Click on "Tabbed Toolbar"
3. Open the Developer tools, JS Console
4. In the post message page, Click on "Hide Save Command"

### Expected Behavior

No message in the console.

### Actual Behavior

Console shows
```
Menu item with id "save" not found.
```

### Screenshots

Before:
![image](https://github.com/CollaboraOnline/online/assets/114441/4df9cc57-1eef-4305-a7e0-af0cdd878152)

After:
![image](https://github.com/CollaboraOnline/online/assets/114441/a3174931-e126-411d-b32d-491da974ebbf)

### Desktop

(Please complete the following information)
 - **OS:** Linux
 - **Browser:** Firefox
 - **Version:** stable

### Additional Context

Some other observations: the icon in the tabbed bar isn't hidden.

Also happen with "Hide Print Command".

In the compact toolbar (classic), it actually doesn't happen and the toolbar button are also hidden.

hfiguiere:
Another takeaway is that this feature doesn't work with the tabbed notebook interface. The messages `Hide_Button` or `Hide_Menu_Item` do nothing when using the tabbed notebook UI mode.

thebearon:
Yes, those commands have only been implemented for the compact view (menu, toolbar) so far.

hfiguiere:
Dupe of #5957

nadirlouati:
any news ?

PR #7523, opened 2023-10-24
postmessage: Add field to allow sending arbitrary postmessages

hfiguiere:
Change-Id: I13976c21ebea018150469dd56f93b3b1ab2105cc

* Resolves: # 
* Target version: master 

### Summary

This allow to send arbitrary post message. This is just UI in the postmessage interface page.

This is how I did test #7517 

### TODO

- [ ] ...

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id
- [x] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [x] Documentation (manuals or wiki) has been updated or is not required

PR #7589, opened 2023-11-01
notebookbar: handle show/hide button

hfiguiere:
https://github.com/CollaboraOnline/online/issues/5957
https://github.com/CollaboraOnline/online/issues/7517

Change-Id: I719f27063f01e4d76742358189b080087881ab7c

* Resolves: #5957
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -22,7 +22,6 @@ L.Control.Notebookbar = L.Control.extend({
 	HOME_TAB_ID: 'Home-tab-label',
 
 	additionalShortcutButtons: [],
-	hiddenShortcutButtons: [],
 
 	onAdd: function (map) {
 		// log and test window.ThisIsTheiOSApp = true;
@@ -281,14 +280,11 @@ L.Control.Notebookbar = L.Control.extend({
 			toolitems.push(item);
 		}
 
-		for (i in this.hiddenShortcutButtons) {
-			var toHide = this.hiddenShortcutButtons[i];
-			for (var j in toolitems) {
-				item = toolitems[j];
-				if (item.id == toHide) {
-					toolitems.splice(j, 1);
-					break;
-				}
+		for (var j in toolitems) {
+			item = toolitems[j];
+			if (!this.map.uiManager.isButtonVisible(item.id)) {
+				toolitems.splice(j, 1);
+				break;
 			}
 		}
 
@@ -327,18 +323,13 @@ L.Control.Notebookbar = L.Control.extend({
 		this.reloadShortcutsBar();
 	},
 
-	showShortcutsButton: function(buttonId, show) {
-		var i = this.hiddenShortcutButtons.indexOf(buttonId);
-		if (i > -1) {
-			if (show === true)
-				this.hiddenShortcutButtons.splice(i, 1);
-
-			this.reloadShortcutsBar();
-			return;
+	showNotebookbarButton: function(buttonId, show) {
+		var button = $('.notebookbar-scroll-wrapper #' + buttonId);
+		if (show) {
+			button.show();
+		} else {
+			button.hide();
 		}
-
-		this.hiddenShortcutButtons.push(buttonId);
-		this.reloadShortcutsBar();
 	},
 
 	setCurrentScrollPosition: function() {
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -17,6 +17,7 @@ L.Control.UIManager = L.Control.extend({
 	blockedUI: false,
 	busyPopupTimer: null,
 	customButtons: [], // added by WOPI InsertButton
+	hiddenButtons: {},
 	previousTheme: null,
 
 	onAdd: function (map) {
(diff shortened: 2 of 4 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -654,8 +655,19 @@ L.Control.UIManager = L.Control.extend({
...
 			this.showButtonInClassicToolbar(buttonId, show);
-		else
+		else {
 			this.notebookbar.showShortcutsButton(buttonId, show);
+			if (show) {
+				delete this.hiddenButtons[buttonId];
+			} else {
+				this.hiddenButtons[buttonId] = true;
+			}
+			this.refreshNotebookbar();
```
eszkadev, line 665: this operation has a big cost in sense of performance, it will reinitialize notebookbar in the core on the server, if we will add few buttons it will be done few times... (Control.Notebookbar.js `resetInCore`)

I wonder if we can avoid reset in the core and just rebuild HTML. It has to be tested with widgets like: style previews, font picker with previews - these notebookbar widgets are using core.

hfiguiere: I have taken that approach as:

1. being minimal in changes
2. meant to be a one time thing in the session
3. it worked

hfiguiere: I have reworked this. It should be faster.

browser/src/control/Control.UIManager.js
```diff
@@ -658,8 +659,21 @@ L.Control.UIManager = L.Control.extend({
 	showButton: function(buttonId, show) {
 		if (!this.notebookbar)
 			this.showButtonInClassicToolbar(buttonId, show);
-		else
+		else {
 			this.notebookbar.showShortcutsButton(buttonId, show);
```
eszkadev, line 663: btw: didi you look at that `showShortcutsButton` function?
it does `this.reloadShortcutsBar();` inside it should already do the show/hide logic.

But I see my mistake there: `this.hiddenShortcutButtons.push(buttonId);` is called unconditionaly, even if `show == true`. Maybe that is enough to fix and it will work?

eszkadev: and maybe we don't need that `hiddenShortcutButtons` inside notebookbar at all, as now you have this "global" `hiddenButtons` thing...

hfiguiere: I have now removed `showShortcutsButton` and just call `reloadShortcutsBar()` and use `UIManager.hiddenButtons`

browser/src/control/Control.UIManager.js
```diff
@@ -658,8 +659,21 @@ L.Control.UIManager = L.Control.extend({
...
+			var button = $('#' + buttonId);
```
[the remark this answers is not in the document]
hfiguiere: moved it to the Notebook bar. Also changed the selector to be more restrictive.

browser/src/control/Control.UIManager.js
```diff
@@ -656,10 +657,21 @@ L.Control.UIManager = L.Control.extend({
...
-		else
-			this.notebookbar.showShortcutsButton(buttonId, show);
+		} else {
+			if (show) {
+				delete this.hiddenButtons[buttonId];
+			} else {
+				this.hiddenButtons[buttonId] = true;
+			}
+			this.notebookbar.reloadShortcutsBar();
+			this.notebookbar.showButton(buttonId, show);
```
eszkadev, line 669: it that order correct? : `first RELOAD then SHOW`

browser/src/control/Control.Notebookbar.js
```diff
@@ -327,18 +323,13 @@ L.Control.Notebookbar = L.Control.extend({
...
-	showShortcutsButton: function(buttonId, show) {
-		var i = this.hiddenShortcutButtons.indexOf(buttonId);
-		if (i > -1) {
-			if (show === true)
-				this.hiddenShortcutButtons.splice(i, 1);
-
-			this.reloadShortcutsBar();
-			return;
+	showButton: function(buttonId, show) {
+		var button = $('.notebookbar-scroll-wrapper #' + buttonId);
```
eszkadev, line 327: I think this will break initial functionality: ability to show/hide shortcutsbar icons, shortcut icons are not inside: `.notebookbar-scroll-wrapper`

scroll wrapper

hfiguiere: I think there is some confusion here. This is only supposed to show/hide the notebook bar button. `showShortcutsButton` is removed because it no longer need to handle `hiddenShortcutButtons` and it is called in one place, and in that place we just call `reloadShoertcutsBar`

browser/src/control/Control.UIManager.js
hfiguiere: actually the order doesn't matter. `reloadShortcutsBar` replace `showShortcutsButton` since we have `hiddenButtons`, while `showButton` is about notebook bar buttons. Maybe I should just rename it to `showNotebookButton`.

Also it should be noted that the shortcut buttons have different ID than the notebook buttons counterparts.

hfiguiere: Just to summarize:
- set the `hiddensButtons` to the right values
- call `reloadShortcutsBar` which will handle the shortcut bar.
- call `showButton` (that I'll rename to be less confusing) to show / hide the notebookbar buttons.

eszkadev: I see! it's good then

eszkadev:
looks good, but please add at least 1/2 sentences what it does/changes in the commit message - so reader will not have to read whole GH issue conversation :)

hfiguiere:
I have augmented the commit message

eszkadev (approved):
Looks good now :)

eszkadev:
Things to consider:
- Could you prepare PR with testing code in framed.doc.html / postMessage tester and add cypress test for it maybe?
- now "id" is used to point the icon to show/hide, so it means "ids" in notebookbar become "published" to the integrators and we cannot change them from now....
- how to protect integrators from regressions because of changed id?
- "ids" are generated to be unique, it may happen that they will change because some buttons will be added/removed (number at the end of the name) so we prefer to use command identifiers which are added as a class. This has advantage because it finds all the instances of the button. Like: `.uno:Save` has always class `unoSave`.
 Could we handle also hiding/showing using that class, and recommend that in SDK in favor of using ids?
 - now "id" is used to point the icon to show/hide, so it means "ids" in notebookbar become "published" to the integrators and we cannot change them....

hfiguiere:
> Could you prepare PR with testing code in framed.doc.html / postMessage tester and add cypress test for it maybe?

framed.doc.html has a send message to sent raw messages. it was added with https://github.com/CollaboraOnline/online/pull/7523

This is what I used to test.

PR #7598, opened 2023-11-02
Fix select row icon

Darshan-upadhyay1110:
- SVG for lc_fromrow is not loading properly in browser
- fixed svg syntex to properly load by browser
- removed unused css class from mobile wizard
- that fixes the icon remains in light mode when we change theme to dark

Change-Id: Ifa1e2e5397e48d33ede8a6081808cd0051d9059a

* Resolves: # 
* Target version: master


eszkadev:
is "removed unused css class from mobile wizard" somehow related to the fixed icon visibility?

Darshan-upadhyay1110:
no that removed class thing is for the mobile view where icon has this wired white background in dark mode. it was a small change so i added here.
The broken icon is because svg is corrupted or something and not properly getting loaded to browser.

eszkadev:
I see, best to put separate topics in separate commits so then it is obvious what it does :) (this can be as it is, just for future)

Darshan-upadhyay1110:
Ahha i see, sorry fo the confusion :)

eszkadev (approved):

PR #7601, opened 2023-11-03
clipboard: do not put formulas into client clipboard

lpranam:
problem:
when user copy cell with formula and paste it outside of online, formulas are pasted instead of values.

Ideally we should always put real value in client clipboard

Change-Id: I93af2fffa501c3cdc3d8ecc78d3c44121784d3e8

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -787,6 +787,14 @@ L.Clipboard = L.Class.extend({
 
 	// sets the selection to some (cell formula) text)
 	setTextSelectionText: function(text) {
+		// Usually 'text' is what we see in the formulabar
+		// In case of actual formula we don't wish to put forumla into client clipboard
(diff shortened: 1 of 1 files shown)


browser/src/map/Clipboard.js
```diff
@@ -788,6 +788,14 @@ L.Clipboard = L.Class.extend({
 
 	// sets the selection to some (cell formula) text)
 	setTextSelectionText: function(text) {
+		// Usually 'text' is what we see in the formulabar
+		// In case of actual formula we don't wish to put forumla into client clipboard
+		// Putting formula in clipboard means user will paste formula outside of online
+		// Pasting inside online is handled by internal paste
+		if (text.startsWith('=')) {
```
eszkadev, line 795: can we do it only in calc?

gokaysatir (approved):
Thanks for the improvement :)

PR #7607, opened 2023-11-03
formulabar: optimization

eszkadev:
formulabar: modify DOM only when content is complete

createDocumentFragment creates out of the DOM nodes which we can use while building the new content of the formulabar. This reduces reflow and similar operations on node insertion.

------------------------------

perf: do not call addEventListener on every header update

This helps with performance of typing into spreadsheets.

Regression from commit https://github.com/CollaboraOnline/online/commit/46c1248c770c2409ebb3b38c796690220ea98880
Prevent right mouse button up event to click on menu item.

Do not add new event listener on every header update what
happens often but rather apply right click blocker when
context menu is opened. Original commit was trying to prevent
menu items activation on right click.

### DIFF
--- a/browser/src/control/jsdialog/Widget.FormulabarEdit.js
+++ b/browser/src/control/jsdialog/Widget.FormulabarEdit.js
@@ -101,12 +101,11 @@ function _appendNewLine(cursorLayer) {
 function _appendCursor(cursorLayer) {
 	var cursor = L.DomUtil.create('span', 'cursor');
 	cursorLayer.appendChild(cursor);
-	var blockOption = JSDialog._scrollIntoViewBlockOption('nearest');
-	cursor.scrollIntoView({behavior: 'smooth', block: blockOption, inline: 'nearest'});
+	return cursor;
 }
 
 function _setSelection(cursorLayer, text, startX, endX, startY, endY) {
-	cursorLayer.innerHTML = '';
+	var newCursorLayer = document.createDocumentFragment();
 
 	var reversedSelection = false;
 	if (endY == startY && endX < startX) {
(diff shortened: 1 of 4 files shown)


browser/src/control/jsdialog/Widget.FormulabarEdit.js
```diff
@@ -106,7 +106,7 @@ function _appendCursor(cursorLayer) {
 }
 
 function _setSelection(cursorLayer, text, startX, endX, startY, endY) {
-	cursorLayer.innerHTML = '';
+	var newCursorLayer = document.createDocumentFragment();
```
mmeeks, line 109: Surely we should use newCursorLayer in the code below to build the fragment? it seems we use cursorLayer (?)

mmeeks (changes requested):
Looked unexpected somehow (?) =)

browser/src/control/jsdialog/Widget.FormulabarEdit.js
eszkadev: done

browser/src/control/jsdialog/Widget.FormulabarEdit.js
```diff
@@ -172,8 +174,17 @@ function _setSelection(cursorLayer, text, startX, endX, startY, endY) {
 
 	// possible after cursor is added to the DOM
 	if (cursor) {
-		var blockOption = JSDialog._scrollIntoViewBlockOption('nearest');
-		cursor.scrollIntoView({behavior: 'smooth', block: blockOption, inline: 'nearest'});
+		if (scrollToCursorTimeout)
+			clearTimeout(scrollToCursorTimeout);
+
+		// put scrol at the end of the task queue so we will not scroll multiple times
```
mmeeks, line 180: Soo ... we have a nice event for this that doesn't require a timeout:
src/layer/tile/TileLayer.TableOverlay.js:               this._map.on('messagesdone', this._updateTableMarkers, this);
best to slave to that if we can in a follow-up commit.

mmeeks (approved):
Looks plausible to me - needs a follow-up; not tested, just read the code =)

browser/src/control/jsdialog/Widget.FormulabarEdit.js
mmeeks: Why is not using a timeout better ? it's more deterministic to rely on messagesdone and reduces the number of interleavings we see / test =)

eszkadev: It was intentional as that callback code is independent and result doesn't affect anything else in the app. 

I tried to avoid "mandatory" scroll after processing every messages chunk (in case of multiple very active users, where messages are produced faster than consumed). JS env task is possible to cancel in that case.

messagesdone: |messages|scroll|messages|scroll|messages|scroll|
setTimeout "0":  |messages|~~scroll~~|messages|~~scroll~~|messages|scroll|

But I did quick test with 6 editors emiting events by typing in parallel. I've seen no difference between `messagesdone` and `setTimeout` in profile  so probably it's hard to produce situation when it has impact.

eszkadev: followup: https://github.com/CollaboraOnline/online/pull/7615

PR #7615, opened 2023-11-06
formulabar: use messagesdone for scrolling

eszkadev:
I tried to avoid "mandatory" scroll after processing every messages chunk (in case of multiple very active users, where messages are produced faster than consumed). JS env task is possible to cancel in that case.

messagesdone: |messages|scroll|messages|scroll|messages|scroll|
setTimeout "0": |messages|scroll|messages|scroll|messages|scroll|

But I did quick test with 6 editors emiting events by typing in parallel. I've seen no difference between messagesdone and setTimeout in profile so probably it's hard to produce situation when it has impact.

Use messagesdone then.

### DIFF
--- a/browser/src/control/jsdialog/Widget.FormulabarEdit.js
+++ b/browser/src/control/jsdialog/Widget.FormulabarEdit.js
@@ -23,7 +23,7 @@
 
 /* global JSDialog */
 
-var scrollToCursorTimeout = null;
+var cursor = null; // used to indicate we need to scrollIntoView
 
 function _sendSelection(edit, builder, id, event) {
 	if (document.activeElement != edit)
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
So you're prolly right that if this scrolling can be triggered by any other code-path than a websocket message it makes more sense to leave it as a timeout - lets just abandon this =) Sorry!

browser/src/control/jsdialog/Widget.FormulabarEdit.js
```diff
@@ -259,6 +245,14 @@ function _formulabarEditControl(parentContainer, data, builder) {
 			L.DomUtil.addClass(cursor, 'hidden');
 	});
 
+	var blockOption = JSDialog._scrollIntoViewBlockOption('nearest');
+	builder.map.on('messagesdone', function() {
```
eszkadev, line 249: this should be deregistered at some point.... easier to have it as setTimeout as it was

PR #7616, opened 2023-11-06
non-calc: Revert "Stop onscreen keyboard when panning calc on tablet"

Minion3665:
This ignores the hints from commit
f62f365a0ed26b21e1f86b0e028cf80fc6b2b9df when not using calc. This is because they were causing an undefined reference in WASM. (window.keyboard.guessOnscreenKeyboard). To me, this looks like a bug in our WASM implementation, as I think WASM should get window globals, but as it's blocking us we're partially rolling this commit back.

I expect to revert this revert as soon as it no longer breaks WASM.

Change-Id: I91bab273d400bf7925e338631e74dabc8a01e529

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/marker/TextInput.js
+++ b/browser/src/layer/marker/TextInput.js
@@ -263,14 +263,23 @@ L.TextInput = L.Layer.extend({
 			return;
 		}
 
+
+		var isCalc = this._map._docLayer.isCalc();
+
 		// Trick to avoid showing the software keyboard: Set the textarea
 		// read-only before focus() and reset it again after the blur()
 		if (!window.ThisIsTheiOSApp && navigator.platform !== 'iPhone' && !window.mode.isChromebook()) {
-			if (window.keyboard.guessOnscreenKeyboard() && acceptInput !== true)
+			if ((
+				(isCalc && window.keyboard.guessOnscreenKeyboard())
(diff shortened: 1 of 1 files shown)


Minion3665:
Approved by @Ashod on another platform, will merge when CI greens
image

browser/src/layer/marker/TextInput.js
```diff
@@ -263,14 +263,23 @@ L.TextInput = L.Layer.extend({
...
 
+
+		var isCalc = this._map._docLayer.isCalc();
+
 		// Trick to avoid showing the software keyboard: Set the textarea
 		// read-only before focus() and reset it again after the blur()
 		if (!window.ThisIsTheiOSApp && navigator.platform !== 'iPhone' && !window.mode.isChromebook()) {
-			if (window.keyboard.guessOnscreenKeyboard() && acceptInput !== true)
+			if ((
+				(isCalc && window.keyboard.guessOnscreenKeyboard())
```
hfiguiere, line 273: why not `window.keyboard && window.keyboard.guessOnscreenKeyboard()` ?

Minion3665:
This isn't really needed for WASM, instead the issue was out-of-sync trees. As this is now pure negative, I'm closing it

Issue #7425, opened 2023-10-12
Chrome: Navigator Hyperlinks Don't Linebreak


Tex2002ans:
### Describe the Bug

In Chrome:

- The Navigator hyperlinks flow off the screen and don't linebreak.

In Firefox:

- The Navigator hyperlinks do linebreak.
   - (Correct!)

### Steps to Reproduce

0. Open [COOLWeekly139.odt](https://github.com/CollaboraOnline/online/files/12883255/COOLWeekly139.odt).
1. Go to "View" tab.
2. Press "Navigator" button.
3. Expand "Hyperlinks" category.

See the URLs.

### Screenshots

Chrome 117.0 vs. Firefox 118.0.1:

![Collabora 23 05 5 1 - Navigator Hyperlinks Chrome](https://github.com/CollaboraOnline/online/assets/5774324/63e055e9-b35e-4903-9e16-ce78a564d232) ![Collabora 23 05 5 1 - Navigator Hyperlinks Firefox](https://github.com/CollaboraOnline/online/assets/5774324/320e1553-7951-4110-ae88-ae3a8fc67bc3)

### Expected Behavior

Linebreaking like Firefox.

### Actual Behavior

In Chrome, Hyperlink text does not linebreak + goes off the edge of screen.

### Desktop

 - **OS:** Windows 10
 - **Browser:** Chrome
 - **Version:** Version 117.0.5938.150 (Official Build) (64-bit)

and:

 - **OS:** Windows 10
 - **Browser:** Firefox
 - **Version:** 119.0b8 (64-bit)

I noticed it in Staging during COOL Weekly Meeting:

> COOLWSD version: 23.05.5.1 (git hash: dffef0e9 (E))
> LOKit version: Collabora Office 23.05.5.1 (git hash: 9f03693)
> Served by: openSUSE Leap 15.5
> Server ID: 3a813eff

Ezinnem:
I can confirm this.

Darshan-upadhyay1110:
This is already fixed by above PR: https://github.com/CollaboraOnline/online/pull/7617

![image](https://github.com/user-attachments/assets/f4c6d1f2-afc1-4f5b-a6f2-be805233ac5d)

Working in latest version:

COOLWSD version: 24.04.5.1 (git hash: 1ecf31df)
LOKit version: Collabora OfficeDev 24.04.5.1 (git hash: e1cfc65)
Served by: Ubuntu 23.10
Server ID: f5be9861

PR #7617, opened 2023-11-06
fix:navigator hyperlinks line-breaking on chrome and firefox

aarsh21:
Change-Id: Icbd85bdd3ec400a016f2c9affdfca04b78711987

* Resolves: # [ Chrome: Navigator Hyperlinks Don't Linebreak #7425 ](https://github.com/CollaboraOnline/online/issues/7425)
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -395,6 +395,8 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	/* Move this whole block away and fix in the main control */
 	display: flex;
 	align-items: center;
+	width: 200px;
+	word-break: break-word;
 }
 
 span.jsdialog.sidebar.ui-treeview-entry.ui-treeview-notexpandable {


Rash419 (approved):
Looks good but I would I like to get +1 from @pedropintosilva

browser/css/jssidebar.css
```diff
@@ -395,6 +395,8 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	/* Move this whole block away and fix in the main control */
 	display: flex;
 	align-items: center;
+	width: 200px;
+	word-break: break-all;
```
pedropintosilva, line 399: # Breaking

Best to use `break-word` instead otherwise the the text will break everywhere making it hard to read

| break-all | break-word |
| ------------- | -----------------|
| `https://en.wikipedia.org/wiki/Sai` | `https://en.wikipedia.org` | 
| `ling_yacht` | `/wiki/Sailing_yacht` |

browser/css/jssidebar.css
```diff
@@ -395,6 +395,8 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	/* Move this whole block away and fix in the main control */
 	display: flex;
 	align-items: center;
+	width: 200px;
```
pedropintosilva, line 398: # Cropping

Awesome! Thanks!

pedropintosilva (changes requested):

browser/css/jssidebar.css
aarsh21: i used break-all cause it looked even on both the browsers chrome and firefox , but you're right thank you.

pedropintosilva (approved):

PR #7618, opened 2023-11-06
Fix video position in safari presentation mode

Minion3665:
https://github.com/CollaboraOnline/online/commit/237d9c084a7ebf4681e1e850c4c5df410f8fa67a fixed videos rendering in the
wrong place on Safari (and all iOS browsers) in edit mode.
Unfortunately, it made the oversight of missing presentation mode, which
gets its SVGs from a completely different place 🤦.

This commit fixes presentation mode.

After this commit, you will see the video in the right place. Please
note that the white border present in edit mode is not there in
presentation mode, so if you line the video edge up with something in
the presentation you may think the video is offset: I suggest lining
something in the video up rather than the edge of the video object.

Known caveats:
- This commit seems to be a few pixels off with larger border sizes.
  It's a lot better than before, but still not perfect

This commit does not include detection for the upcoming Layer-Based SVG
engine <https://wpewebkit.org/blog/05-new-svg-engine.html>. This means
that this commit will *break* video if this is enabled in Safari debug
mode.  If this commit is merged as-is we need to make a followup to fix
this as soon as possible or we will end up with a similar-but-opposite
bug when that flag is toggled on by default.

Follow-up-to: I205e692e7027ad917bd6f29aa96b0ac70a4c9e04
Change-Id: Ifeea18874d189c80cbb96029706a6f71d6125898

### DIFF
--- a/browser/src/map/handler/Map.SlideShow.js
+++ b/browser/src/map/handler/Map.SlideShow.js
@@ -134,6 +134,7 @@ L.Map.SlideShow = L.Handler.extend({
 		if ('processCoolUrl' in window) {
 			this._processSlideshowLinks();
 		}
+		this._processSlideshowVideoForSafari();
 
 		this._startPlaying();
 	},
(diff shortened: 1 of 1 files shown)


Minion3665:
as predicted

> This commit has only been tested on my mac, with my monitor. If the magic numbers depend on that this may not work at all for you. Safari video is inevitably fragile.

this is not working on my iPad... gonna try to unmagic it some more

eszkadev:
Tested, works well on Macbook Air M1 and iPad 9 in "Presentation mode" and "Edit mode" but not in "Present in Window" yet (which is present as default in Nextcloud app for iOS.

eszkadev (approved):

PR #7621, opened 2023-11-07
notebookbar: add Insert Frame to the Insert tab

vmiklos:
This allows the creation of floating tables when an entire table is
selected by the time we dispatch .uno:InsertFrame.

Change-Id: Ia0614d47275eb8d9127c496a0b06a8993c5f9989


vmiklos:
@pedropintosilva once you're back, could you please state if you're OK to put this new insert frame button on the insert tab in the notebookbar, under qr code? The other option I considered is fontwork, but that's tricky (hidden for non-ODF), so it's not ideal to put an always-on frame button under it. It looks like this:

image

Thanks.

pedropintosilva:
I have no objection. But @gokaysatir might want to check if this will affect the access keys (we are trying to set each to already known letters from MSO but there are a limit set of keys and positions )

vmiklos:
unit-copy-paste failed, let's try again.

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/8487/

And again unit-copy-paste is one of the failed tests. I wonder that broke there.

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/8491/

Progress, now only unit-bad-doc-load fails, but that passes for me locally.

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/8515/

Better, now we fail in cypress tests.

eszkadev:
pplease rebase, writer/statusbar_spec.js will be fixed

vmiklos:
> please rebase, writer/statusbar_spec.js will be fixed

Done. :-)

eszkadev:
We need also menubar entry in the compact mode: `browser/src/control/Control.Menubar.js`

vmiklos:

@pedropintosilva already said OK from the design point of view. I checked that the a11y shortcut is unique, so I believe it should be good enough.

vmiklos:
> We need also menubar entry in the compact mode: `browser/src/control/Control.Menubar.js`

Ah, let me fix that. I hoped we killed the menubar by now.

caolanm (approved):
I think its uncontroversial if it now passes ci

PR #7623, opened 2023-11-07
Add touch detection for pointer:course media query

Minion3665:
matchMedia allows you to check matches for CSS media queries from JavaScript. By checking if the primary pointer is 'coarse' we have a pretty good shot at guessing that it's a touch device (mice, etc. are 'fine' instead). This expands our current touch detection so it detects more screens, while making sure it doesn't detect devices which have a touchscreen but also a mouse/touchpad/whatever. This non-detection is important, because our touch support currently breaks things like the rightclick menu on calc.

Touchscreen detection is useful for devices that are being used as touch devices, because it enables some features like pinch-to-zoom and hold-for-rightclick-menu which improve usability when you are using a touchscreen. These are particularly important without a mouse.

Some devices are 2-in-1 laptops, which have a touchscreen and a touchpad, but can be collapsed so only the touchscreen is active. These devices will be detected as touch only when they are in their 'tablet' form. *As we setup parts of the page differently, the page will need to be refreshed if they change from tablet form to laptop form and vice-versa*. Hopefully in the future we will make a followup so that both input schemes can be enabled at the same time and extend this media query detection to (any-pointer: coarse)

Change-Id: Ied4f61a1ffb06bd63359bc6d81c6a6c23436cc23

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -108,7 +108,8 @@ window.app = {
 	var chromebook = window.ThisIsTheAndroidApp && window.COOLMessageHandler.isChromeOS();
 
 	var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window ||
-			(window.DocumentTouch && document instanceof window.DocumentTouch)) && !chromebook;
+			(window.DocumentTouch && document instanceof window.DocumentTouch) ||
+			window.matchMedia('(pointer: coarse)').matches) && !chromebook;
 
 	var isInternetExplorer = (navigator.userAgent.toLowerCase().indexOf('msie') != -1 ||
 				  navigator.userAgent.toLowerCase().indexOf('trident') != -1);


browser/js/global.js
```diff
@@ -108,7 +108,8 @@ window.app = {
 	var chromebook = window.ThisIsTheAndroidApp && window.COOLMessageHandler.isChromeOS();
 
 	var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window ||
-			(window.DocumentTouch && document instanceof window.DocumentTouch)) && !chromebook;
+			(window.DocumentTouch && document instanceof window.DocumentTouch) ||
+			window.matchMedia('(pointer: coarse)').matches) && !chromebook;
```
hfiguiere, line 112: How do we handle chromebooks with touchscreen?

hfiguiere (approved):
LGTM.

browser/js/global.js
Minion3665: > How do we handle chromebooks with touchscreen?

that's a good question! unfortunately I don't have a chromebook to test on, but from line 108 here
```js
var chromebook = window.ThisIsTheAndroidApp && window.COOLMessageHandler.isChromeOS();
```
chromebook detection is whether we're running on the android app on chromeos... therefore the detection in the browser would be handled exactly the same way as on a windows laptop that had a touchscreen (assuming ChromeOS doesn't appear as a touchscreen, which I guess it could)

Minion3665: I'm not sure how the android app would handle input on a chromebook, I remember being told that there was something a little different for the iOS app... maybe android has something similar?

PR #7624, opened 2023-11-07
fix; error TS2304: Cannot find name 'TilesSection'

caolanm:
doesn't happen in ci, but is happening in tinderbox and for some users.

Seeing as ScrollSection doesn't have the problem, follow the export and namespace pattern used there.

Change-Id: I4c4916ba7a3ad19275bd8b0f5bc53111b3b30cd5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
@Minion3665 I think you were able to reproduce this build failure. Does this solve it?

Minion3665 (approved):
This fixes the issue for me, and code looks fine... thanks!

PR #7626, opened 2023-11-07
'Tileprocessed message did not arrive in time' for slide preview

caolanm:
A slide preview tile falls into the special case with early return and no tileprocessed is sent back to the server, so eventually a 'Tileprocessed message did not arrive in time' will be reported

Change-Id: Iab85e5eca535ce377508e91cf67f1435e9242bc4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great; thanks Caolan =)

PR #7631, opened 2023-11-08
cypress: disbale failing Switching page test

eszkadev:
Fails only in cypress, cannot reproduce in manual testing. Probalby difference is enabled a11y in cypress?

Fails with: "Page 1 and 2 of 2" instead of "Page 2 of 2"


eszkadev:
we have better fix: https://github.com/CollaboraOnline/online/pull/7632

PR #7632, opened 2023-11-08
make this test more robust if partial page 1 is still visible when on…

caolanm:
… page 2

Change-Id: I6f5cc5a1d22b77fe3f9adebdb5fde086fa8e2a3b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

caolanm:
wrt private/eszkadev/cypress-disable-failing I can reproduce this locally and could bisect to core b9e5543be78b0b124bc087840f13ae1bfd2b155b but I think I have seen this intermittently in the past before that too.

eszkadev:
> wrt private/eszkadev/cypress-disable-failing I can reproduce this locally and could bisect to core b9e5543be78b0b124bc087840f13ae1bfd2b155b but I think I have seen this intermittently in the past before that too.

Yes, I saw it even before. It's better than just disabling so I will cancel my PR.

PR #7634, opened 2023-11-08
Copying the hyperlink issue:

gokaysatir:
* Add a paragraph.
* Select the whole paragraph and add comment.
* Add hyperlink to a word in the paragraph.
* Try to copy the hyperlink via its menu.
* It sometimes doesn't copy the hyperlink.

We are checking if a comment has focus or not before doing the copy operation. This check is for another issue. I changed the check with "app.view.commentHasFocus" which seems reliable. Then tested the previous fix.

Also commented out code is removed.
White space change is automatic.

Change-Id: Ibc2c5e550eb8a3e24b734e0371c87dac44b2ff6d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -484,7 +484,6 @@ L.Clipboard = L.Class.extend({
 		this._checkSelection();
 
 		var text = this._getHtmlForClipboard();
-		//this._stopHideDownload(); - this confuses the browser ruins copy/cut on iOS
 
 		var plainText = this.stripHTML(text);
 		if (ev.clipboardData) { // Standard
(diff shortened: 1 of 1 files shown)


lpranam (approved):
I could not reproduce the bug, but code look fine.

Thanks :)

mmeeks (commented):
Looks sensible to me - please add a the hash and author of the regression in the commit message.  It would be good of course to test copy/paste in comments and so on at the same time as this goes on.

mcecchetti (approved):
I tested it and it seems to have fixed both issues.

mcecchetti:
Thanks Gokay

Issue #7167, opened 2023-08-25
Extra invalidation regression tests ...


mmeeks:
We should expose in the JS client a total count of the number of empty deltas we have received, not just per-tile:
_applyDelta:
...
				tile.updateCount++;

Should bump a global counter somewhere. Prolly we should put this into an HTML element in the DOM somewhere easy for Cypress to find it as well: it would be nice to have it displayed - it is almost certainly a bad-ness metric.

If we invalidate something, and we generate an empty delta - something is wrong - not very wrong, but ... inefficient: we shouldn't be re-rendering identical tiles.

Then we should add code to our cypress tests to check that HTML element with the number in it to ensure that it doesn't change over eg. a new user joining the document, or a character being typed in writer at a certain point, or clicking in the header etc.

Thanks ! =)

caolanm:
This is done now, the specific examples are included. I note we are sending empty updates from at least grammar checking while editing, so there is room for improvement uncovered by these regression tests for the more gross fixed invalidations.

PR #7636, opened 2023-11-08
add regression tests via cypress for unwanted no-change invalida…

caolanm:
…tions

single user scenarios:

a) Click Empty Header: fails with core:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=f425c0aab6e33b8029bf7e87ac9b10b4b4489ac9
reverted

b) Enter Numbering Portion: fails with core:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=643c6b9a96020d5498d981c7ab081226415ab016
reverted

c) Click Existing Header: fails with core:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=ce948053b1fa9a2843886dce4642cc36c94dd0e0
reverted

multi-user scenarios:
    
d) Rejoin triggers invalidations: fails with core:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=f1633dc0e06b046a9ca0f2fa53e7858be1584ffb
reverted

e) Rejoin after save triggers constant invalidations: fails with core:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=770df258bc0a66ee7551a59807d79cebcdcbb85e
reverted

https://github.com/CollaboraOnline/online/issues/7167


mmeeks (changes requested):
Looks like we put a <div behind the canvas - if we do that un-conditionally this may be a problem for accessibility; I'm not an expert on this but I think we want something like getTestDiv()  or addPathTestDiv() - that stashes this stuff away somewhere hopefully somewhere where it is so hidden/invisible - that changing it will not trigger unwanted repaints / performance issues or a11y problems (?) =)

Otherwise- looks awesome; great to have tests for this !

mcecchetti (approved):
I see the empty delta div is created only on cypress tests, moreover a11y tests are passing so there is no interaction issue with a11y.

PR #7642, opened 2023-11-09
make this other test more robust if partial page 1 is still visible

caolanm:
when cursor is on page 2

Change-Id: Iae8d28ad54401fbbe78a2f2429f1a6912f1ea967

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
we made another similar test more robust with https://github.com/CollaboraOnline/online/pull/7632 and this one is the same pattern

gokaysatir (approved):
Thanks for the help :)

PR #7644, opened 2023-11-09
Remove /dark from images path and use filter invert property for dark  mode icons

Darshan-upadhyay1110:
- Current case we use icons for dark mode from dark folder inside images
- both the icons in Light/Dark are almost same
- we need to just change color of that icons when ever it changes the theme
- we can use filter: invert(1) to change svg icon for dark theme

Note: Currently in this PR for safe side i will not delete the dark icons folder inside images. will do it after proper testing

Change-Id: I158ddfc609bd26d3e0cf8940e183089439375c6f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thank you for the improvements. This looks even a bit better in some parts of the UI. Hopefully also easier to manage in case we want to update the theme colors.

pedropintosilva (changes requested):
We currently don't have monochrome icons which means that we cannot simply invert them, we need also to do the hue changes etc as you did but then the probability of breaking dark mode is really high on every icon update.

So I think we should not do this for all icons. I think we could invert for elements that are not really icons and style previews: example thinks that are monochrome such as checkboxes or radio buttons, or in Format > Columns the columns icon (which was already inverted by Andras and merged by me)

Darshan-upadhyay1110:
I forgot to update here. let me close this for now as it's too risky to invert all icons on mode change. will create new patch where we should only target the icons as mentioned in this comment ( CC @pedropintosilva )

> We currently don't have monochrome icons which means that we cannot simply invert them, we need also to do the hue changes etc as you did but then the probability of breaking dark mode is really high on every icon update.

> So I think we should not do this for all icons. I think we could invert for elements that are not really icons and style previews: example thinks that are monochrome such as checkboxes or radio buttons, or in Format > Columns the columns icon (which was already inverted by Andras and merged by me) but for example bullets preview in bullets dialog still need the inversion

PR #7645, opened 2023-11-10
context menu: allow frame properties for Writer text frames

vmiklos:
This is now safe to show, as it'll trigger the frame insert/properties
dialog, which was converted to jsdialog in core.git commit
355681eead2411d70caf4f52f1b802cf8c61a981 (sw floattable: make Insert
Frame dialog async and mark it as a jsdialog, 2023-11-06).

Change-Id: I910012286416231d607cf79cc9b3811803c52768


vmiklos:

Given that the "frame properties" and the "insert frame" dialog is the same code, this uno command is now safe (i.e. not problematic UX-wise) to expose.

caolanm (approved):
looks reasonable to me

PR #7649, opened 2023-11-12
NC: Accessibility toggle button not set to the right state on doc load

mcecchetti:
When a document is loaded the Accessibility Support toggle button was
not set to the right state.
The problem affeced Online when integrated in Nextcloud.

Moreover the button label has been renamed to 'Voice Over' in
accordance with what suggested by NGI audit

Change-Id: I5bc7d8348397691f469b6274af58e18a8d6d603c


gokaysatir (approved):
It now remembers the last state. Thanks @mcecchetti.

Issue #7638, opened 2023-11-09
Android: "File > Properties" Misaligned Headings


Tex2002ans:
### Describe the Bug

In **File > Properties**, the menu that pops up has uneven + squished-together-headings:

### Steps to Reproduce

In Writer, after opening a document:

1. Press three-line hamburger menu in upper-right corner.
2. **File > Properties**

The menu appears at the bottom of device.

### Screenshots

Landscape:

<img alt="Collabora 23 05 5 2 - File Properties Alignment Landscape" src="https://github.com/CollaboraOnline/online/assets/5774324/be09ad33-7a8f-4911-942c-49e9a7bd27e3" width="50%"/>

Portrait:

<img alt="Collabora 23 05 5 2 - File Properties Alignment" src="https://github.com/CollaboraOnline/online/assets/5774324/6594d746-89ef-4f92-a7b1-c05225c2c14e" width="50%"/>

### Expected Behavior

The selections for:

- General
- Description
- Font
- Statistics

should be evenly spaced/distributed.

### Actual Behavior

"Font" + "Statistics" are jammed together along far-right edge.

And "Font" is very hard to click.

### Smartphone

- **Device:** OnePlus 6T
- **OS:** Android 11 (OxygenOS 11.1.2.2)

> COOLWSD version: 23.05.5.2
> git hash: 8fe1f0fe
> LOKit version:
> Collabora Office 23.05.5.2
> git hash: a2baff6

### Additional Context

I'd expect:

- All 4 to be ~ the same size when pressing with finger
- And/OR you should be able to "swipe" left/right through them.
   - Currently you can only PRESS each one with your finger.

Also, looks like the unchecked checkboxes in that "Font" screen are black in Dark Mode too:

- Embed fonts in the document
- Only embed fonts that are used in documents

Very hard to see.

Darshan-upadhyay1110:
Not only a android bug but also Reproduce in mobile view.

PR #7651, opened 2023-11-13
Fix mobile-ui-tabs Misaligned Headings

Darshan-upadhyay1110:
Change-Id: Ib982a59c141d937c7f92eb9684b91fc7f2548df5

- will fix Misaligned ui-tabs in mobile wizard
- also added scrollable property that will handle more fields in tabs if it does not fit into screen

* Resolves: #7638 
* Target version: master 

### Summary

- Before
image

-After

image

### DIFF
--- a/browser/css/mobilewizard.css
+++ b/browser/css/mobilewizard.css
@@ -477,9 +477,10 @@ p.mobile-wizard.ui-combobox-text.selected {
 .ui-tabs.mobile-wizard {
 	text-align: center;
 	height: 42px;
-	display: table;
+	display: flex;
 	width: 100%;
 	padding: 0px;
+	overflow-x: auto;
 }
 
 #mobile-wizard-tabs {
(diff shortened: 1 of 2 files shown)


Darshan-upadhyay1110:
> disable that menu item completely when on mobile because as we stand it's unusable.

@pedropintosilva `menu item` means the Property tag inside `files` ?

pedropintosilva:
Yes

Darshan-upadhyay1110:
@pedropintosilva i have removed `Property` menu item for mobile.  Still there are many places where we have this Mialligned headings issue.

For Ex: On Mobile 
- go to Download As
- selcted Downaload as PDF 

Before this patch that dialog looks very odd.

so Pedro can we go with this flex property for better User interaction in mobile.


About 

> Also I have noticed that it seems we are ignoring .button-primary that is correctly assigned to the OK button on desktop but on mobile-wizard we loose that andthen every single pushbutton is styled as primary button.

I will fix that in other PR with better description so this will not overlap with current issue

pedropintosilva:
I'm also worried about that margin change

pedropintosilva (approved):

pedropintosilva:
Looks good but I didn't test the second commit , can we get @Rash419 to give a +2?

Rash419 (approved):
thanks

PR #7652, opened 2023-11-13
fix: improve prometheus-rules

genofire:
Change-Id: I5076c74c3beeada0978d5b29dcc4e531cfce9241

* Target version: master 

### Summary

ping @Rash419

we refactoring your default alerts there text and redefined the Alert DocumentsOpenSimultaneously: where on critical, it say how many of all documents are opened simultaneously on different pods procent (if the ingress does not work correct)

### Checklist


Rash419 (approved):
Looks good

PR #7653, opened 2023-11-13
fix(helm): fix haproxy balance (no not use post body for balancing - just WOPISrc)

genofire:
This PR correct the documentation for kubernetes (helm) of the issue #7654 (i do not know where to fix the official documentation: https://sdk.collaboraonline.com/docs/installation/Proxy_settings.html#example-with-haproxy)

Ping @Rash419 and @mmeeks

### Checklist

### DIFF
--- a/kubernetes/helm/collabora-online/Chart.yaml
+++ b/kubernetes/helm/collabora-online/Chart.yaml
@@ -4,7 +4,7 @@ type: "application"
 name: collabora-online
 description: Collabora Online helm chart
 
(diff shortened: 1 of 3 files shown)


Rash419:
why would removing check_post fix the problem ?
https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4.2-balance%20url_param

genofire:
@Rash419 let is discuss it in the issue #7654

PR #7656, opened 2023-11-13
Unify touchscreen handling, remove L.Browser.touch

Minion3665:
L.Browser.touch is sometimes nice, but it's ultimately a flawed concept to use it for input events. Using L.Browser.touch for input handicaps people with mice if it's too liberal in what it classes as touchscreens, and handicaps people with touchscreens if it's too conservative. There's also no sweet spot: it's impossible to choose correctly if someone is using both a touchscreen and a pointer device, as there's no right option!

Previously many of our event handlers and some of our UI was gated behind L.Browser.touch. This commit adds a new "window.touch" property which is used instead. It has functions to help with event detection, allowing you to easily make event handlers that work for only the input devices they are designed for, without gating them behind feature detection. This has the added bonus that - as you register all the events - switching between a touchscreen and pointer is now not only possible but already implemented!

For cases which don't have reasonable events to tag onto (e.g. the teardrop for cursor movement) this commit adds "hasPrimaryTouchscreen" and "hasAnyTouchscreen" which use the CSS media queries to detect if there's a touchscreen attached to your device (either as the primary input mechanism or at all). This works a lot more similarly to L.Browser.mode, but being dynamically updated allows you to effectively swich between touchscreen and not at-will. This still has all of the disadvantages that L.Browser.touch did when used to register event handlers, so my advice would be to avoid using it with events.

Change-Id: I9016fc15ad3ccb3664af348fdcdca006495b0778

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -107,13 +107,6 @@ window.app = {
 
 	var chromebook = window.ThisIsTheAndroidApp && window.COOLMessageHandler.isChromeOS();
 
-	if (window.uiDefaults['touchscreenHint'] !== undefined)
-		var touch = window.uiDefaults['touchscreenHint'];
-	else
-		var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window ||
-				(window.DocumentTouch && document instanceof window.DocumentTouch) ||
-				window.matchMedia('(pointer: coarse)').matches) && !chromebook;
-
 	var isInternetExplorer = (navigator.userAgent.toLowerCase().indexOf('msie') != -1 ||
 				  navigator.userAgent.toLowerCase().indexOf('trident') != -1);
 
@@ -238,10 +231,6 @@ window.app = {
 		// `true` when the browser run by cypress
 		cypressTest: cypressTest,
 
-		// @property touch: Boolean
-		// `true` for all browsers supporting [touch events](https://developer.mozilla.org/docs/Web/API/Touch_events).
-		touch: !!touch,
-
 		// @property msPointer: Boolean
 		// `true` for browsers implementing the Microsoft touch events model (notably IE10).
 		msPointer: !!msPointer,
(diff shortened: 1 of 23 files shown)


Minion3665:
@Ashod please may you make sure this doesn't break WASM when you have some time... I see no reason why it would be different from the keyboard stuff so I want to make sure

thanks!

Minion3665:
CI is passing so we're ready to get reviews. Please, if you have a device with a touchscreen, it would be great if you could test this PR. If you have a device that doesn't have a touchscreen, please test this PR. It touches a lot of mouse and touch input code, so the more people we get testing this, the better.

---

On a related note, if you review positively, please don't merge this yet. We believe it could cause issues with WASM as well as some other stuff we want to wait on confirmation for. Thanks!

Minion3665:
@Ashod has told me that this does not break WASM, it would be great if we could get some reviews on this so we can merge soon

browser/js/global.js
```diff
@@ -279,6 +268,51 @@ window.app = {
...
+				return !(e instanceof MouseEvent);
+			}
+
+			return e.pointerType === 'touch' || e.pointerType == 'pen' || e.pointerType == 'kinect';
+		},
+
+		/// a decorator that only runs the function if the event is a touch event
+		touchOnly: function(f) {
+			return function(e) {
+				if (!global.touch.isTouchEvent(e)) return;
```
hfiguiere, line 288: nit': I tend to really dislike this formatting.
```suggestion
				if (!global.touch.isTouchEvent(e)) {
					return;
				}
```

also elsewhere

hfiguiere: it does.

image

note: none of this is blocking either.

hfiguiere:
LGTM, but it would be great to have someone more familiar with this review.

eszkadev:
Funny bug: Open writer, insert shape, try to drag the shape. result: you will scroll the window with mouse cursor on movement. Later I clicked outside and then it moved the cursor and overlay with handles.
Was using regular case: chrome + mouse

[scroll-2023-11-22_12.59.41.webm](https://github.com/CollaboraOnline/online/assets/5307369/9b98e4a7-29f7-43f0-a6fb-279ad38ddbbe)

PR #7659, opened 2023-11-13
fix: Use proper command + c keyboard shortcut for copy actions on snackbar

juliushaertl:
The initial copy action already used Cmd+C but when a download is required the snackbar actions were still requiring Ctrl-C which is not common on macOS.

Change-Id: I24d9e01bc23a1dc32a1362f48775c6890f5a77c8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


juliushaertl:
🏓 Kind ping for review ;)

eszkadev (approved):
Thanks!

PR #7661, opened 2023-11-13
wopi: Added DisableInsertLocalImage

hfiguiere:
This disable being able to pick up a local image. Handle the UI as well. To be used with `EnableInsertRemoteImage`

Return `DisableInsertLocalImage`: true as part of the WOPI CheckFileInfo reply

Change-Id: I1925f804433a7bda6025cb65b0943d78927bea15

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -884,12 +884,16 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 	},
 
 	_insertGraphicControl: function(parentContainer, data, builder) {
-		var options = {hasDropdownArrow: builder.map['wopi'].EnableInsertRemoteImage};
+		var options = {
+			hasDropdownArrow: builder.map['wopi'].EnableInsertRemoteImage && !builder.map['wopi'].DisableInsertLocalImage
+		};
 		var control = builder._unoToolButton(parentContainer, data, builder, options);
 
 		$(control.container).unbind('click.toolbutton');
 		$(control.container).click(function () {
-			if (builder.map['wopi'].EnableInsertRemoteImage) {
+			var hasRemote = builder.map['wopi'].EnableInsertRemoteImage;
+			var hasLocal = !builder.map['wopi'].DisableInsertLocalImage;
+			if (hasRemote && hasLocal) {
 				var menu = [
 					{id: 'localgraphic', text: _('Insert Local Image')},
 					{id: 'remotegraphic', text: _UNO('.uno:InsertGraphic', '', true)}
@@ -909,8 +913,10 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 					onSelect: itemCallback
 				});
 				builder._makeW2MenuFocusable(builder, 'w2ui-overlay-insert-graphic-menu', menu, data.id, itemCallback);
-			} else {
+			} else if (hasLocal) {
 				L.DomUtil.get('insertgraphic').click();
+			} else if (hasRemote) {
+				builder.map.fire('postMessage', {msgId: 'UI_InsertGraphic'});
 			}
 		});
 		builder._preventDocumentLosingFocusOnClick(control.container);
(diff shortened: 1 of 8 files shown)


browser/src/control/Control.TopToolbar.js
```diff
@@ -434,9 +435,24 @@ L.Control.TopToolbar = L.Control.extend({
 			w2ui['editbar'].hide('savebreak');
 		}
 
-		if (e.EnableInsertRemoteImage === true && w2ui['editbar']) {
-			w2ui['editbar'].hide('insertgraphic');
-			w2ui['editbar'].show('menugraphic');
+		if (w2ui['editbar']) {
+			if (e.EnableInsertRemoteImage === true && !e.DisableInsertLocalImage) {
+				w2ui['editbar'].hide('insertgraphic');
+				w2ui['editbar'].hide('insertremotegraphic');
```
caolanm, line 441: it seems odd to have something called "EnableInsertRemoveImage" hiding something called "insertremovegraphic"?

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -884,12 +884,16 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
...
+		var options = {
+			hasDropdownArrow: builder.map['wopi'].EnableInsertRemoteImage && !builder.map['wopi'].DisableInsertLocalImage
+		};
 		var control = builder._unoToolButton(parentContainer, data, builder, options);
 
 		$(control.container).unbind('click.toolbutton');
 		$(control.container).click(function () {
-			if (builder.map['wopi'].EnableInsertRemoteImage) {
+			var hasRemote = builder.map['wopi'].EnableInsertRemoteImage;
+			var hasLocal = !builder.map['wopi'].DisableInsertLocalImage;
```
Ashod, line 895: We should use either positive or negative statements. Why not make both EnableInsert...Image ?

browser/src/map/handler/Map.WOPI.js
```diff
@@ -24,6 +24,7 @@ L.Map.WOPI = L.Handler.extend({
 	DownloadAsPostMessage: false,
 	UserCanNotWriteRelative: true,
 	EnableInsertRemoteImage: false,
+	DisableInsertLocalImage: false,
```
Ashod, line 27: `EnableInsertLocalImage: true,` ?

Ashod (commented):
Thanks @hfiguiere. Left a couple of notes.

browser/src/control/Control.TopToolbar.js
hfiguiere: I can rename it `EnableInsertLocalImage` and have it set to `true` by default. I thought `false` by default was fine, but either work.

browser/src/control/Control.NotebookbarBuilder.js
hfiguiere: I can rename it `EnableInsertLocalImage` and have it set to `true` by default. I thought `false` by default was fine, but either work.

browser/src/control/Control.TopToolbar.js
hfiguiere: but it shows menugraphic below. If we have the two options local and remote then we show the menu, otherwise we show one or the other.

Previously you couldn't disable the local image insertion, so there was only two choices. remoteimage enabled -> menu, otherwise just the button. The new button is added above (line 230).

hfiguiere:
Rename the WOPI parameter.

Also fixed a bug in classic mode: the command didn't work.

browser/src/control/Control.NotebookbarBuilder.js
thebearon: @Ashod  @caolanm There was a very good reason why the existing is `Enable` and this is `Disable`. `EnableInsertRemoteImage` concerns inserting images from the WOPI host, and needs additional support on the side of the WOPI host, while this one is for disabling a feature in COOL, because an integration doesn't want to allow inserting images from the users' computer.
If this patch has to go in, I would strongly suggest considering reverting to the original naming.

hfiguiere: It's back to `DisableInsertLocalImage`

hfiguiere:
CI is green

caolanm (approved):

PR #7627, opened 2023-11-07
browser: Add ui_default to hide the toolbar

hfiguiere:
Change-Id: I189224581451c9f5d534593696b060c8efb23414

* Resolves: # 
* Target version: master 

### Summary

Pass `{Text,Spreadsheet,Presentation,Draw}Toolbar=false` in the POST to hide the toolbar in compact mode.

### TODO

- [ ] Have it work with the notebook, but this is only the same as "click" on a tab.

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id
- [x] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #7664, opened 2023-11-14
Menubar & Tabs visibility logic fixes

Minion3665:
# Commit 9115d9df8588f4efe21531ff7d8d4fd4e7424e72

Previously, when using the Collapse_Notebookbar postmessage or equivalent ui_defaults (SpreadsheetToolbar=false, etc.), particularly in compact mode, it was possible to additionally hide the menu bar. As the button to show the menu bar is on the notebookbar, this meant that you couldn't reactivate either notebookbar or menubar until you refreshed the page. This is particularly annoying in integrators that may not provide an easy way to reload the page

This commit makes it so that hiding the menu bar automatically uncollapses the notebookbar and won't let it be collapsed again. Whether the notebook bar should be collapsed (the last thing done to it was a collapse) is remembered and restored after the menu bar is shown again, so if you send a postmessage that will affect the state of the notebookbar after the menu is shown (even though it will not affect the notebookbar's state immediately)

Caveats:
- If you are hiding the notebookbar to limit the control the user has, that's broken by this commit as it makes it impossible to hide both the menu and notebook bars at the same time.
- The notebook bar will be hidden again when re-showing the menu bar, however there still isn't a way to hide the notebook bar in normal use (i.e. without using either postmessage or ui_defaults) while in compact mode (although there is a workaround to show it- switching into tabbed mode and then back!). It might be nice to have one.

Other considered solutions:
- We could add a new button that allowed you to reopen the menu if both menu and notebookbar were hidden
  - Not sure there's much benefit to this over just doing what we're doing here, and it's harder to implement
- We could disable the button to hide the menu bar when the notebookbar is collapsed
  - As far as I know, there's no button in the UI to show the notebook bar. This would make it impossible to hide the menu bar if the notebookbar was hidden via postmessage or ui_defaults

Change-Id: Ieab6d72a6be181aba88e9a5b21dda16a369b9e54

# Commit de0410ecaee0a31f9c872d9ae7e5d7a524269308

Prevent hiding the menu bar in tabbed mode

Tabbed mode doesn't have a menu bar, instead it has tabs. These can't be
hidden. Unfortunately, the post messages to hide the menu bar have the
side effect of hiding the tabs. This commit prevents the tabs being
hidden when in tabbed mode, and shows the tabs again when switching from
compact mode into tabbed mode.

When switching back from tabbed into compact mode, the state that you
would like the menu bar to be in (hidden/shown) will be remembered and
restored. This includes any postmessages that were not acted on while in
tabbed mode.

Change-Id: I1177903fe965e354538e6e7bbc3c83af3177938e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -674,11 +674,19 @@ L.Control.UIManager = L.Control.extend({
 		obj.removeClass('w2ui-icon unfold');
 		obj.addClass('w2ui-icon fold');
 		$('#tb_editbar_item_fold').prop('title', _('Hide Menu'));
+
+		if (this._notebookbarShouldBeCollapsed)
+			this.collapseNotebookbar();
 	},
 
 	hideMenubar: function() {
 		if (this.isMenubarHidden())
 			return;
+
+		var notebookbarWasCollapsed = this.isNotebookbarCollapsed();
+		this.extendNotebookbar();  // The notebookbar has the button to show the menu bar, so having it hidden at the same time softlocks you
+		this._notebookbarShouldBeCollapsed = notebookbarWasCollapsed;
+
 		$('.main-nav').hide();
 		if (L.Params.closeButtonEnabled) {
 			$('#closebuttonwrapper').hide();
@@ -752,7 +760,9 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	collapseNotebookbar: function() {
-		if (this.isNotebookbarCollapsed())
+		this._notebookbarShouldBeCollapsed = true;
+
+		if (this.isNotebookbarCollapsed() || this.isMenubarHidden())
 			return;
 
 		this.moveObjectVertically($('#formulabar'), -1);
@@ -764,6 +774,8 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	extendNotebookbar: function() {
+		this._notebookbarShouldBeCollapsed = false;
+
 		if (!this.isNotebookbarCollapsed())
 			return;
 
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -429,6 +429,9 @@ L.Control.UIManager = L.Control.extend({
 		this.map._docLayer._requestNewTiles();
 
 		this.map.topToolbar.updateControlsState();
+
+		if (this._menubarShouldBeHidden)
+			this.hideMenubar();
 	},
 
 	createNotebookbarControl: function(docType) {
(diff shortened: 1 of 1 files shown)


pedropintosilva:
> This commit makes it so that hiding the menu bar automatically uncollapses the notebookbar and won't let it be collapsed again. Whether the notebook bar should be collapsed (the last thing done to it was a collapse) is remembered and 

This seems reasonable. Somehow I'm getting a weird behaviour where hiding the menubar still hides the tabs:

https://github.com/CollaboraOnline/online/assets/65948705/77ca917e-d537-4e35-a051-f1d0223a6798

I would prefer that if `Hide Menubar`, and since  there is no menu bar in the tabbed view, would do nothing to tabbed view.

Note: Collapse and expand tabs are already possible via post message or by pressing the active tab or by pressing the collapse tabs (view tab)

Minion3665:
> I would prefer if `Hide Menubar` would act solely on comapct view. Since there is no menu bar in the tabbed view, it shouldn't affect tabbed view at all.

sure, that makes sense. Would you like me to do that here or in a followup PR as it's a separate thing, @pedropintosilva?

pedropintosilva (changes requested):
Thanks! Now the hide menu bar doesn't affect tabbed view :+1: 

Still something looks out of place in the compact view

https://github.com/CollaboraOnline/online/assets/65948705/1e89c985-f354-4edf-9917-68c68a186256

Hide menu bar can now trigger `Extend Tabbed Toolbar`

I would to have both `Extend Tabbed Toolbar` and `Collapse Tabbed Toolbar` not affecting compact view at all.

Minion3665:
> I would prefer to have both `Extend Tabbed Toolbar` and `Collapse Tabbed Toolbar` not affecting compact view at all.

Are you sure? The commands are used elsewhere (e.g. https://github.com/CollaboraOnline/online/pull/7627) in order to hide the compact view toolbar.

I could make it so that there were various different states that were restored when switching back, but as I am aware of at least 1 instance of people using stuff around here to affect compact mode in the wild so I don't know that we should stop these functions affecting it.

@hfiguiere @pedropintosilva what do you think?

Minion3665:
looking again, maybe we could change this so that the ui_defaults added in #7627 don't affect notebookbar and the postmessages to collapse/expand don't affect the compact view (and then maybe add some more postmessages to collapse and expand for compact view)... would this be a good solution? I notice the collapse/expand postmessages for tabbed toolbar were only added 2 weeks ago so maybe it's not too late to change what they do (see https://github.com/CollaboraOnline/online/pull/7556)?

hfiguiere:
I think we can merge these improvements as is (unless @pedropintosilva disagrees)
We can still followup

hfiguiere (approved):

pedropintosilva (approved):
> I think we can merge these improvements as is (unless @pedropintosilva disagrees)
We can still followup

sure

PR #7666, opened 2023-11-15
Update bug_report.md with Collabora

timur-g:
When creating NEW bug in GITHUB, there is template text, it includes OS, Browser, Version, but missing is COLLABORA Version. I am seeing reports with all fillied, but missing that which is most important. In addiiton, I set Browser with Versoin to be single line with e.g. Chrome 114.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,43 +8,34 @@ assignees: ''
 ---
 
 ### Describe the Bug
-
 A clear and concise description of what the bug is.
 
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
the `Signed-off-by: Timur Gadžo <timur.gadzo@collabora.com>` was missing in the commit message and the change-id also.  

I have `git commit --amend  --author="Timur Gadžo <timur.gadzo@collabora.com>" --signoff` amend the commit, and force push

PR #7667, opened 2023-11-15
wsd: disable ctrl+s when HideSaveOption is true

Rash419:
Change-Id: I190a03abcf80176cc2da77fd32190c683852b9a7

* Resolves: follow up of https://github.com/CollaboraOnline/online/commit/6b06f1d8fec1788c18c5755d7e5c2046c3407cbf
* Target version: master

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -690,6 +690,11 @@ bool ClientSession::_handleInput(const char *buffer, int length)
                                 << "] has no write permissions in Storage and cannot save.");
             sendTextFrameAndLogError("error: cmd=save kind=savefailed");
         }
+        else if (_wopiFileInfo && _wopiFileInfo->getHideSaveOption())
+        {
+            LOG_WRN("Session [" << getId() << "] on document [" << docBroker->getDocKey()
+                            << "] has HideSaveOption set to true by CheckFileInfo, cannot save");
+        }
         else
         {
             // Don't save unmodified docs by default.


eszkadev (commented):
code looks ok, but let's double check with Ash if he doesn't have anything against

wsd/ClientSession.cpp
```diff
@@ -690,6 +690,11 @@ bool ClientSession::_handleInput(const char *buffer, int length)
                                 << "] has no write permissions in Storage and cannot save.");
             sendTextFrameAndLogError("error: cmd=save kind=savefailed");
         }
+        else if(_wopiFileInfo && _wopiFileInfo->getHideSaveOption())
```
Ashod, line 693: Please use clang-format or manually format code to be consistent with C++ code in the project.

There is always a space after `if` tokens.

wsd/ClientSession.cpp
```diff
@@ -690,6 +690,11 @@ bool ClientSession::_handleInput(const char *buffer, int length)
...
+        {
+            LOG_WRN("Session [" << getId() << "] on document [" << docBroker->getDocKey()
```
Ashod, line 695: I assume the UI will have blocked the save as well. So this warning is to signal that someone is trying to hack the UI?

Ashod (approved):
Thanks @Rash419. Looks good, but formatting can be improved.

wsd/ClientSession.cpp
Rash419: yeah thanks for noticing, I just switched editor I will configure clang-format, which clang formatting style do we use ?

wsd/ClientSession.cpp
Rash419: Yes

wsd/ClientSession.cpp
Rash419: before this I was using vscode and it was falling back to its own style
 https://code.visualstudio.com/docs/cpp/cpp-ide

PR #7671, opened 2023-11-16
browser: fix undefined property 'isPopup'

hcvcastro:
The PopUp dialog is not closed, and side effects
are unresponsive key input

Change-Id: Id72ef0c6d081aa73acb39a07eb3e8b33d0e8dc85

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -554,7 +554,7 @@ L.Control.JSDialog = L.Control.extend({
 		instance.updatePos = null;
 		instance.canHaveFocus = !instance.isSnackbar && instance.id !== 'busypopup' && !instance.isMention;
 		instance.isDocumentAreaPopup = instance.popupParent === '_POPOVER_' && instance.posx !== undefined && instance.posy !== undefined;
-		instance.isPopUp = instance.isModalPopUp || instance.isDocumentAreaPopup || instance.isSnackbar;
+		instance.isPopup = instance.isModalPopUp || instance.isDocumentAreaPopup || instance.isSnackbar;
 		instance.containerParent = instance.isDocumentAreaPopup ? document.getElementById('document-container'): document.body;
 		instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();
 		instance.haveTitlebar = (!instance.isModalPopUp && !instance.isSnackbar) || (instance.hasClose && instance.title && instance.title !== '');
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -606,7 +606,8 @@ L.Control.JSDialog = L.Control.extend({
 			// They are displayed before the document is loaded
 			// Spinning should be happening until the 1st interaction with the user
 			// which is the dialog opening in this case
-			this.map._progressBar.end();
+			if (this.map)
+				this.map._progressBar.end();
 
 			this.createContainer(instance, instance.overlay ? instance.overlay: instance.containerParent);
 			this.createDialog(instance);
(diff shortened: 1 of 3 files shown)


eszkadev:
Could you provide more detailed commit message? How to reproduce the error, in which dialog?

browser/src/control/Control.JSDialog.js
```diff
@@ -77,7 +77,7 @@ L.Control.JSDialog = L.Control.extend({
 			if (this.dialogs[id].timeoutId)
 				clearTimeout(this.dialogs[id].timeoutId);
 
-			if (this.dialogs[id].isPopup)
+			if (this.dialogs[id].isPopUp)
```
eszkadev, line 80: There is other occurrence of `isPopup` in the code, in general it was the original name of the property, but it was changed in the rework at some point: https://github.com/CollaboraOnline/online/commit/a2d666d53adcb23e60ed61328e8e3fe44fdc1af5

Can we rather change where it is set to `isPopUp` -> `isPopup` to be consistent with other occurrences?
Here: https://github.com/CollaboraOnline/online/blob/93825c688ccb0430bff1b5114996230a7615013c/browser/src/control/Control.JSDialog.js#L557C11-L557C19

hcvcastro:
Changes done.

eszkadev:
Thanks now looks good, can we also have test for that case in: `cypress_test/integration_tests/desktop/calc/autofilter_spec.js` ?

hcvcastro:
This is not an auto filter issue, perhaps ensure if popup dialog is closed

eszkadev:
As I understood we can test this bug with autofilter popup so that's why I proposed that. Also it would be good to put exact steps to reproduce the issue in the commit message. So later in the future someone will know how to test that.
like:
1. open calc
2. insert autofilter, click on dropdown arrow
...
Result: cannot type into the cell
Expected: it should something...

hcvcastro:
I guess there is no JSDialog unit test, I will investigate

eszkadev:
There is integration/cypress test for autofilter menu, in mentioned: cypress_test/integration_tests/desktop/calc/autofilter_spec.js :)

cypress_test/integration_tests/desktop/calc/help_dialog_update_spec.js
```diff
@@ -44,4 +45,30 @@ describe(['tagscreenshot'], 'Help dialog update', function() {
 		});
 	});
 
+	it('JSDialog popup dialog', function() {
```
eszkadev, line 48: nice test! but it seems to be unrelated to other tests in this file...
Can we put that into separate file? like   `jsdialog_spec.js`    so we will have there jsdialog specific tests

eszkadev:
your test nicely checks internals, but it would be good to have also integration test which also communicates with the core.
Could you please add one in `cypress_test/integration_tests/desktop/calc/autofilter_spec.js` which will click into `jsdialog-overlay` and will test if autofilter popup was closed and we can type into cell?

cypress_test/integration_tests/desktop/calc/help_dialog_update_spec.js
hcvcastro: I could not find a proper to file to add the unit test but "jsdilalog_spec.js" is Ok, I guess it is a matter to add the file and extra config file

eszkadev: @hcvcastro I think for cypress it is enough to add new file in the directory, it will be automatically executed during check

cypress_test/integration_tests/desktop/calc/help_dialog_update_spec.js
```diff
@@ -1,3 +1,4 @@
+/* -*- js-indent-level: 8 -*- */
```
eszkadev, line 1: unwanted change

hcvcastro: no, future indentation for emacs

hcvcastro:
I get confused with other comments here. As I wrote before this is not autofilter issue, this was a JSDialog no calling a proper closePopup method. if we want to check the the core responses, there should be a JSDialog C++ unit test.

eszkadev (approved):

PR #7679, opened 2023-11-17
fix: metrics change doc_pid to doc_info use pid as label

genofire:
@mmeeks i like to have the pid value as an label so it would be use on the other metrics for lookup with [promql group_left](https://prometheus.io/docs/prometheus/latest/querying/operators/#group-modifiers)

i believe it is usefull to rename that metrics afterwards my suggestion `doc_info` as an representer for all the metrics

After this change i will enrich the PrometheusRules / Alerts to show more information (based on the instance and pid label) - current i believe the following should work:
`doc_active_views * group_left(doc_info)`

### Checklist

### DIFF
--- a/kubernetes/helm/collabora-online/grafana_dashboards/ha-allocation.json
+++ b/kubernetes/helm/collabora-online/grafana_dashboards/ha-allocation.json
@@ -458,7 +458,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "count(doc_pid{env_id=~\"$env_id\"})by(host,env_id)",
+          "expr": "count(doc_info{env_id=~\"$env_id\"})by(host,env_id)",
           "legendFormat": "{{env_id}}: {{host}}",
           "range": true,
           "refId": "A"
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Seems reasonable to me as/when it passes CI =) Not a huge fan of changing the interface here, but it was never that beautiful.
Depending on feedback we should consider versioning the URL we get the metrics from I think.

mmeeks (changes requested):
So of course - the code also needs to compile =) 

wsd/AdminModel.cpp:1197:42: error: stray ‘\’ in program
                "pid=\"" << pid << "\"} 1"\n";
                                          ^
wsd/AdminModel.cpp:1197:44: error: missing terminating " character [-Werror]
                "pid=\"" << pid << "\"} 1"\n";

etc. ... =)

genofire:
(or `doc_open`)

mmeeks (approved):
Well - of course, its not beautiful to have this dummy '1' value - and no doubt a prometheus purist would want us to generate gigabytes of duplicate data for no benefit here =) and I'm not interested in being pushed in that direction - but this seems to me a reasonable approach for now =)

Thanks !

genofire:
it is very common in prometheus to have an info or up metric with the value if 1

the database / prometheus should compress it nice -but yes that is transmit some more bytes

mmeeks:
Good to know - thanks =)

PR #7684, opened 2023-11-20
annotation: reset comment id if autosave is cancelled

lpranam:
problem:
User A writes Comment 1 and saves
User A replies to himself, write something, click away to autosave, Cancel
user A modifies Comment 1 and OK - comment disappears, although it is there, seen after reload, but without modification.

Change-Id: I5d83936f26939b5a05a0ce3099c01923a55c9606

* Target version: master 

### Checklist


Rash419 (approved):
Looks good, Thanks

PR #7685, opened 2023-11-20
delta: fix threading gotcha when working.

mmeeks:
The ThreadPool::work function can get its condition signalled -very- late.

With bad timing, this can occur after all the work is done, and when the next batch of work is being fed into the pool.

This can mean that it takes work from the queue, and subverts the:

        bool useThreads = _threads.size() > 1 && _work.size() > 1;

check in ThreadPool::run - which can believe we are in a single threaded, single tile mode - and not wait for this thread to complete.

That's not good [!] so ensure that threads are only runnable during ThreadPool::run.

Change-Id: Ifebb0f15cbb4c22ef33ffba06e7c6c87493818be

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -34,10 +34,12 @@ class ThreadPool {
     std::vector<std::thread> _threads;
     size_t _working;
     bool   _shutdown;
+    std::atomic<bool> _running;
 public:
     ThreadPool()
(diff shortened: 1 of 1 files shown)


caolanm (approved):
looks sane to me

PR #7692, opened 2023-11-20
Private/mmeeks/kip 1

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -470,12 +470,8 @@ void forkLibreOfficeKit(const std::string& childRoot,
     // Cleanup first, to reduce disk load.
     cleanupChildren();
 
-#ifndef KIT_IN_PROCESS
-    (void) limit;
-#else
-    if (limit > 0)
+    if (Util::isKitInProcess() && limit > 0)
         ForkCounter = limit;
-#endif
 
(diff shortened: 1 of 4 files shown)


mmeeks:
Latest work is on mmeeks/private/kit-in-process branch - but splitting these bits out should make that work a lot smaller and simpler after re-basing; lets get this in first.

kit/ForKit.cpp
```diff
@@ -470,12 +470,8 @@ void forkLibreOfficeKit(const std::string& childRoot,
     // Cleanup first, to reduce disk load.
     cleanupChildren();
 
-#ifndef KIT_IN_PROCESS
-    (void) limit;
-#else
-    if (limit > 0)
+    if (!Util::isKitInProcess() && limit > 0)
```
JaumePujantell, line 473: The condition equivalent to the original code should be "if (Util::isKitInProcess() && limit > 0)".

mmeeks: ah good catch =) thanks will fix.

JaumePujantell (approved):
Looks good to me.

Darshan-upadhyay1110 (approved):
LGTM Thanks :)

PR #7694, opened 2023-11-21
context menu: allow unfloat table for Writer text frames

vmiklos:
This is safe to do as it's not interactive, so it happens synchronously.
The matching uno command was added on core.git co-23.05 in commit
1f5c20352725cd6133e68e80e8523d865006161f (sw floattable, delete UI: add
an uno command to unfloat frame from context menu, 2023-11-17).

Change-Id: I74736c7d589c2062a8e9255a42f81bf790b7d3e3


vmiklos:

This makes the delete/unfloat UI from core visible in Online as well.

caolanm (approved):

PR #7696, opened 2023-11-21
annotation: make sure parent-child relation is established

lpranam:
problem:
some times we order comments before parent-child relation is established, this caused reply count not being updated correctly for other users

steps to reproduce:
User A open ODT, let us say already with comment or add one, and keep Sidebar open so that comments are short. user B opens the same ODT with full comments. adds reply User A does not update ticket number, does not have +1

Change-Id: If3360c8dd938c6bd177764d3c1383d7f3f845990

* Target version: master 

### Checklist


Rash419 (approved):
looks good

PR #7702, opened 2023-11-23
browser: don't show socket error message if we are reconnecting

Rash419:
- specifically a case where coolwsd is cleaning up the document we don't want to give up on clientside and show the error message to user

Change-Id: Ie75fa3e19f97ca18acc8e881813d6c617fc9f7b7

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1526,7 +1526,7 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		window.app.console.debug('_onSocketClose:');
-		if (!this._map._docLoadedOnce) {
+		if (!this._map._docLoadedOnce && this.ReconnectCount === 0) {
 			var postMessageObj = {
 				errorType: 'websocketconnectionfailed',
 				success: false,


browser/src/core/Socket.js
```diff
@@ -1526,7 +1526,7 @@ app.definitions.Socket = L.Class.extend({
 
 	_onSocketClose: function () {
 		window.app.console.debug('_onSocketClose:');
-		if (!this._map._docLoadedOnce) {
+		if (!this._map._docLoadedOnce && this.ReconnectCount === 0) {
```
eszkadev, line 1529: I noticed that `ReconnectCount` is only increased in our code.
What about the case when we successfully reconnected to the server? shouldn't it be reset to 0 ?

Rash419: Yes it resets to zero here
https://github.com/CollaboraOnline/online/blob/3f46c1db445b9840318af0a4994f0c3ff1012285/browser/src/core/Socket.js#L1090

eszkadev: ah I missed that one

eszkadev (approved):
looks good

PR #7711, opened 2023-11-23
Make icon structure consistent

Darshan-upadhyay1110:
- All Icon HTML structure shoulb be same as other icons
- Most of the places we have this way to add icon in DOM
  - <Button>  <Button>
- This patch is to change for some icon with drop down which is not consistent with above structure

Change-Id: I1a82bc4b3dfbca15f62bfe111e36804c754da539

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Command.js
+++ b/browser/src/control/Control.Command.js
@@ -47,7 +47,7 @@ L.Map.include({
 			if (window.mode.isMobile()) {
 				var overlay = L.DomUtil.create('div', 'locking-overlay', DOMParentElement);
 				var lock = L.DomUtil.create('img', 'locking-overlay-lock', overlay);
-				L.LOUtil.setImage(lock, 'lc_lock.svg', this._docLayer._docType);
+				L.LOUtil.setImage(lock, 'lc_lock.svg', this);
 			}
 
 			$(DOMParentElement).click(function(event) {
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2802,7 +2802,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var icon = builder._createIconURL(data.command);
 			var buttonId = data.id ? data.id: id + '-button';
 			var button = L.DomUtil.create('button', 'ui-content unobutton', div);
-			button.style.background = 'url(' + L.LOUtil.getImageURL(icon.split('/').pop(), builder.map.getDocType()) + ')';
+			// Create an img element
+			var iconImg = document.createElement('img');
+			// Set the src attribute of the img element to the image URL
+			iconImg.src = L.LOUtil.getImageURL(icon.split('/').pop(), builder.map.getDocType());
+
+			// Append the img element to the button
+			button.appendChild(iconImg);
 			button.id = buttonId;
 			button.setAttribute('alt', id);
 
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -822,12 +822,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		case 'SdTableDesignPanel':
 		case 'ChartTypePanel':
 		case 'rotation':
-			iconURL = L.LOUtil.getImageURL('lc_'+ sectionTitle.id.toLowerCase() +'.svg', builder.map.getDocType());
+			iconURL = 'lc_'+ sectionTitle.id.toLowerCase() +'.svg';
 			break;
 		}
 		if (iconURL) {
 			var icon = L.DomUtil.create('img', 'menu-entry-icon', leftDiv);
-			L.LOUtil.setImage(icon, iconURL.split('/').pop(), builder.map.getDocType());
+			L.LOUtil.setImage(icon, iconURL, builder.map);
 			icon.alt = '';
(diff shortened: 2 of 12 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2802,7 +2802,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var icon = builder._createIconURL(data.command);
 			var buttonId = data.id ? data.id: id + '-button';
 			var button = L.DomUtil.create('button', 'ui-content unobutton', div);
-			button.style.background = 'url(' + L.LOUtil.getImageURL(icon.split('/').pop(), builder.map.getDocType()) + ')';
+			// Create an img element
+			var iconImg = document.createElement('img');
+			// Set the src attribute of the img element to the image URL
+			iconImg.src = L.LOUtil.getImageURL(icon.split('/').pop(), builder.map.getDocType());
```
eszkadev, line 2808: In other places we do that using: `L.LOUtil.setImage`, then we also check if image exists and in case of error - use followup

Darshan-upadhyay1110:
@eszkadev second commit is for adding support to handle icon refresh directly from `setImage` method

browser/src/core/LOUtil.js
```diff
@@ -108,9 +108,15 @@ L.LOUtil = {
...
 		return url;
 	},
-	setImage: function(img, name, doctype) {
-		img.src = this.getImageURL(name, doctype);
-		this.checkIfImageExists(img);
+	setImage: function(img, name, doctype, map) {
+		var setupIcon = function () {
+			img.src = this.getImageURL(name, doctype);
+			this.checkIfImageExists(img);
+			console.log('Hello', img, name, doctype);
```
eszkadev, line 115: debug log left

browser/src/core/LOUtil.js
```diff
@@ -108,9 +108,15 @@ L.LOUtil = {
 		url += path;
 		return url;
 	},
-	setImage: function(img, name, doctype) {
-		img.src = this.getImageURL(name, doctype);
-		this.checkIfImageExists(img);
+	setImage: function(img, name, doctype, map) {
```
eszkadev, line 111: `docType` can be extracted from `map.getDocType()`, maybe we can pass only map, for simpler usage?

browser/src/core/LOUtil.js
Darshan-upadhyay1110: sure i'll fix it

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2409,17 +2409,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
-				buttonImage.src = imagePath;
-				L.LOUtil.checkIfImageExists(buttonImage);
-			};
-
 			var buttonImage = L.DomUtil.create('img', '', button);
-			setupIcon();
-
-			builder.map.on('themechanged', setupIcon);
+			var icon = data.icon ? data.icon : builder._createIconURL(data.command);
+			L.LOUtil.setImage(buttonImage, icon.split('/').pop(), builder.map);
```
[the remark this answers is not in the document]
eszkadev: I think this is last thing to do, looks good now :)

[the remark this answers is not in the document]
Darshan-upadhyay1110: @eszkadev i am not sure we can remove split from here because icon also depends on data.icon and the value i don't know will be not url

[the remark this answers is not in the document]
eszkadev: ok, lets not change that then.
But I think it should be ok, I checked core - it doesn't send that icon property, only place is https://github.com/CollaboraOnline/online/blob/578dd1e7801f5dbfbbae6e41fa53b381e7c10de6/browser/src/control/Control.Notebookbar.js#L315
where integrator can put own button.

eszkadev:
without branding now it adds 2 icons for color control:
color control

eszkadev (approved):
Looks good, let's have also approval from @pedropintosilva

browser/src/control/Control.Command.js
```diff
@@ -47,7 +47,7 @@ L.Map.include({
 			if (window.mode.isMobile()) {
 				var overlay = L.DomUtil.create('div', 'locking-overlay', DOMParentElement);
 				var lock = L.DomUtil.create('img', 'locking-overlay-lock', overlay);
-				L.LOUtil.setImage(lock, 'lc_lock.svg', this._docLayer._docType);
+				L.LOUtil.setImage(lock, 'lc_lock.svg', this);
```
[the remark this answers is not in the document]
Darshan-upadhyay1110: we changed the `setImage` function and for that i have chnaged parameters in all places where we use setImage function

browser/src/core/LOUtil.js
```diff
@@ -108,9 +108,15 @@ L.LOUtil = {
...
-		this.checkIfImageExists(img);
+	setImage: function(img, name, map) {
+		var setupIcon = function () {
+			var docType = map.getDocType();
+			img.src = this.getImageURL(name, docType);
+			this.checkIfImageExists(img);
+		}.bind(this);
+		setupIcon();
+
+		map.on('themechanged', setupIcon, this);
```
pedropintosilva, line 119: oh, I se why

PR #7712, opened 2023-11-23
Followup to unified touch handling

Minion3665:
Touch unification (https://github.com/CollaboraOnline/online/commit/b3bff28bede07a48cdecefcda14a6fc9da86e4bb, change ID
I9016fc15ad3ccb3664af348fdcdca006495b0778) was a rework of the input
system to better support touch devices, but unfortunately it caused some
fairly serious regressions. This commit fixes the following:
- Triple/Quadruple clicking was no longer recognized on non-touch
  devices.
- There were some issues recognizing wrapped events
- Pens were considered to be touch devices, but this broke some
  remote/virtual machine setups. It's possible that this change will
  cause a regression for apple pencil users. I plan to get an Apple
  pencil to test whether preventing pens from being touch devices breaks
  the Apple pencil workflow instead
- manualDrag was taken to mean "is on a touch device". This is only
  almost true, causing some input to be incorrectly ingnored when it was
  not
- manualDrag and _manualDrag were confused. They affect different
  things, and we now recognize this

A second followup may be needed to fix the two instances of `this.dragging.disable`. One of them was present from the initial unification commit, the other was not. I don't know what issue they cause, and I expect this will take an amount of work that puts it out-of-scope for this commit.

### DIFF
--- a/browser/src/dom/Draggable.js
+++ b/browser/src/dom/Draggable.js
@@ -21,6 +21,18 @@ L.Draggable = L.Evented.extend({
 		}
 	},
 
+	_manualDrag: function() {
+		return false;
+	},
+
+	noManualDrag: window.memo.decorator(function(f) {
+		return function(e) {
+			if (!this._manualDrag(e)) {
+				return f.apply(this, arguments);
+			}
+		};
+	}),
+
 	initialize: function (element, dragStartTarget, preventOutline) {
 		this._element = element;
 		this._dragStartTarget = dragStartTarget || element;
@@ -38,17 +50,17 @@ L.Draggable = L.Evented.extend({
 	},
 
 	enable: function () {
-		if (this._manualDrag || this._enabled) { return; }
+		if (this._enabled) { return; }
 
-		L.DomEvent.on(this._dragStartTarget, L.Draggable.START.join(' '), this._onDown, this);
+		L.DomEvent.on(this._dragStartTarget, L.Draggable.START.join(' '), this.noManualDrag(this._onDown), this);
 
 		this._enabled = true;
 	},
 
 	disable: function () {
-		if (this._manualDrag || !this._enabled) { return; }
+		if (!this._enabled) { return; }
 
-		L.DomEvent.off(this._dragStartTarget, L.Draggable.START.join(' '), this._onDown, this);
+		L.DomEvent.off(this._dragStartTarget, L.Draggable.START.join(' '), this.noManualDrag(this._onDown), this);
 
 		this._enabled = false;
 		this._moved = false;
@@ -90,8 +102,8 @@ L.Draggable = L.Evented.extend({
 		this.startOffset = this._startPoint.subtract(new L.Point(startBoundingRect.left, startBoundingRect.top));
 
 		L.DomEvent
-			.on(document, L.Draggable.MOVE[e.type], window.touch.mouseOnly(this._onMove), this)
-			.on(document, L.Draggable.END[e.type], window.touch.mouseOnly(this._onUp), this);
+		 .on(document, L.Draggable.MOVE[e.type], this.noManualDrag(this._onMove), this)
+		 .on(document, L.Draggable.END[e.type], this.noManualDrag(this._onUp), this);
 	},
 
 	_onMove: function (e) {
@@ -178,8 +190,8 @@ L.Draggable = L.Evented.extend({
 
 		for (var i in L.Draggable.MOVE) {
 			L.DomEvent
-			    .off(document, L.Draggable.MOVE[i], this._onMove, this)
-			    .off(document, L.Draggable.END[i], this._onUp, this);
+			 .off(document, L.Draggable.MOVE[i], this.noManualDrag(this._onMove), this)
+			 .off(document, L.Draggable.END[i], this.noManualDrag(this._onUp), this);
 		}
 
 		L.DomUtil.enableImageDrag();
--- a/browser/src/layer/vector/SVGGroup.js
+++ b/browser/src/layer/vector/SVGGroup.js
@@ -9,7 +9,16 @@ L.SVGGroup = L.Layer.extend({
 
 	options: {
 		noClip: true,
+		manualDrag: window.touch.isTouchEvent,
 	},
+
+	noManualDrag: window.memo.decorator(function(f) {
+		return function(e) {
+			if (!this.options.manualDrag(e)) {
+				return f.apply(this, arguments);
+			}
+		};
+	}),
 
 	initialize: function (bounds, options) {
 		L.setOptions(this, options);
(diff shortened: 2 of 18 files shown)


browser/src/layer/vector/SVGGroup.js
```diff
@@ -9,6 +9,20 @@ L.SVGGroup = L.Layer.extend({
 
 	options: {
 		noClip: true,
+		manualDrag: window.touch.isTouchEvent,
+	},
+
+	_noManualDragMemo: {},
+	noManualDrag: function(f) {
```
eszkadev, line 16: can we create Util with this functionality?
I see it is reapeated in [Draggable.js](https://github.com/CollaboraOnline/online/pull/7712/files#diff-67a550f117b7181778af814172e9a1ad841e2c15bcc0b98568724284eb44fc7c)

browser/src/map/handler/Map.TouchGesture.js
```diff
@@ -98,44 +98,60 @@ L.Map.TouchGesture = L.Handler.extend({
...
+			var bound = L.bind(f, context);
+
+			if (touchOnly) {
+				bound = window.touch.touchOnly(bound);
+			}
+
+			this._boundEventListeners[L.stamp(f) + '-' + L.stamp(context) + '-' + L.stamp(touchOnly)] = bound;
+		}
+
+		return this._boundEventListeners[L.stamp(f) + '-' + L.stamp(context) + '-' + L.stamp(touchOnly)];
```
eszkadev, line 114: this is also the mentioned repeated piece?

eszkadev:
Nice, now VM / VNC works :)

browser/src/map/handler/Map.TouchGesture.js
Minion3665: yes, it's quite similar ... I'll make something to generalize all these 3, because it's quite a useful thing to have

browser/src/dom/Draggable.js
```diff
@@ -21,6 +21,18 @@ L.Draggable = L.Evented.extend({
 		}
 	},
 
+	_manualDrag: function() {
+		return false;
```
eszkadev, line 25: This is always false here. I see no places where it will be true. Is that expected? Or for future flexibility and extension?

browser/src/layer/vector/SVGGroup.js
```diff
@@ -9,8 +9,17 @@ L.SVGGroup = L.Layer.extend({
 
 	options: {
 		noClip: true,
+		manualDrag: window.touch.isTouchEvent,
```
eszkadev, line 12: why we need this? can't we use directly `window.touch.isTouchEvent` as before?
I see no current code which uses that in a different way.

eszkadev:
Chrome, responsive mode, drag a textbox in Impress - exception in the browser log:

[drag-2023-12-15_17.15.44.webm](https://github.com/CollaboraOnline/online/assets/5307369/dd9a7270-966f-4329-a308-66518e83aa60)

eszkadev:
Chrome, responsive mode, in Calc insert shape, try to drag - doesn't work

[drag-calc-2023-12-15_17.19.25.webm](https://github.com/CollaboraOnline/online/assets/5307369/1a8dbcfb-e09f-4ba5-9c29-0738cd8dbe57)

eszkadev:
Chrome, responsive mode, in Writer, insert shape, resize, try to drag - exception in browser console:

[Uploading drag-writer-2023-12-15_17.25.17.webm…]()

eszkadev:
Firefox responsive mode - cannot scroll in Writer using "pan" gesture

Minion3665:
> Chrome, responsive mode, drag a textbox in Impress - exception in the browser log:

> Chrome, responsive mode, in Writer, insert shape, resize, try to drag - exception in browser console:

Thanks for these, I've applied a fix

> Chrome, responsive mode, in Calc:

Still trying to reproduce this one, I'll keep you posted

> Firefox responsive mode - cannot scroll in Writer using "pan" gesture

This one's interesting... it seems that when using a real touchscreen panning works no matter whether we are in responsive mode or not. At a glance this makes me think that Firefox's fake "touch events" aren't being interpreted as such, I need to do further testing to confirm this

Minion3665:
> > Chrome, responsive mode, in Calc:
> > first double click in fromulabar / try to select in formulabar
> > insert shape, try to drag - doesn't work
>
> Still trying to reproduce this one, I'll keep you posted

@eszkadev  I'm struggling to reproduce this one... can you share some more details about your setup (e.g. is this over RDP again? can you share how you inserted the shape and selected the formulabar contents... I notice you've only filmed the dragging failing?)

eszkadev:
> > > Chrome, responsive mode, in Calc:
> > > first double click in fromulabar / try to select in formulabar
> > > insert shape, try to drag - doesn't work
> > 
> > 
> > Still trying to reproduce this one, I'll keep you posted
> 
> @eszkadev I'm struggling to reproduce this one... can you share some more details about your setup (e.g. is this over RDP again? can you share how you inserted the shape and selected the formulabar contents... I notice you've only filmed the dragging failing?)

Not it seems to work for me.

eszkadev:
Now I used Firefox responsive mode using VNC (so my mouse is detected as "pen"):
I can click in Calc - selects the cell.
Cannot pan. Event I have is:
```
additionalEvent: "panright"
angle: -54.93638994342833
center: {…}
changedPointers: (1) […]
deltaTime: 404
deltaX: 113
deltaY: -161
direction: 4
distance: 196.69773765857096
eventType: 4
isFinal: true
isFirst: false
maxPointers: 2
offsetDirection: 8
overallVelocity: -0.39851485148514854
overallVelocityX: 0.27970297029702973
overallVelocityY: -0.39851485148514854
pointerType: "touch"
pointers: (1) […]
preventDefault:emit/data.preventDefault()
rotation: 0
scale: 1
srcEvent: pointerup
target: div.leaflet-layer
timeStamp: 1702908209224
type: "pan"
velocity: 0
velocityX: 0
velocityY: 0
```

eszkadev:
Firefox on Mac, using mouse, responsive mode. The same problem: cannot pan.
Event data:
```
additionalEvent: "pandown"
angle: 123.6900675259798
center: {…}
changedPointers: (1) […]
deltaTime: 111
deltaX: -14
deltaY: 21
direction: 16
distance: 25.238858928247925
eventType: 2
isFinal: false
isFirst: false
maxPointers: 1
offsetDirection: 16
overallVelocity: 0.1891891891891892
overallVelocityX: -0.12612612612612611
overallVelocityY: 0.1891891891891892
pointerType: "touch"
pointers: (1) […]
preventDefault: preventDefault()
rotation: 0
scale: 1
srcEvent: pointermove
target: div.leaflet-layer
timeStamp: 1702908450580
type: "pan"
velocity: 0.5757575757575758
velocityX: -0.3939393939393939
velocityY: 0.5757575757575758
```

eszkadev (approved):
But these Firefox issues are not regression from this change and this PR fixes some problems - let's merge and get more testing.

PR #7714, opened 2023-11-23
calc: select whole row/column on mouse drag

lpranam:
problem:
In Calc click on a row header and drag the mouse. The selection does not follow the mouse, selection only appears when you release the mouse button

Change-Id: I510f35d9f61fca63a619e83c04ff340f9578f5d2

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -34,6 +34,7 @@ export class Header extends CanvasSectionObject {
 	_dragEntry: HeaderEntryData;
 	_mouseOverEntry: HeaderEntryData;
 	_startSelectionEntry: HeaderEntryData;
+	_lastSelectedIndex: number;
 	_hitResizeArea: boolean;
 	_menuItem: any;
 	_dragDistance: number[];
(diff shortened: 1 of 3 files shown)


browser/src/control/Control.Header.ts
```diff
@@ -543,6 +544,15 @@ export class Header extends CanvasSectionObject {
 		else { // We are in dragging mode.
 			this._dragDistance = dragDistance;
 			this.containerObject.requestReDraw(); // Remove previously drawn line and paint a new one.
+
+			if (this._lastSelectedIndex == this._mouseOverEntry.index)
+				return;
+			const modifier = this._lastSelectedIndex ? UNOModifier.SHIFT : 0;
+			this._lastSelectedIndex = this._mouseOverEntry.index;
+			if (this._isColumn)
```
eszkadev, line 552: It's a bit weird that in base class we check type and manually choose method :)
It could be `selectIndex(index)` method which has different implementation in row and column classes...
Could you do that? To not make this pattern more popular here.

eszkadev (approved):

PR #7716, opened 2023-11-24
formulabar: select formulas on mouseleave event

lpranam:
problem:
In a spreadsheet, click on a cell with content,
In the formula bar, select the content in a way that the mouse key is lifted outside the edit field Note how the formula bar doesn't get into edit mode, but you can still edit the content

Change-Id: Ib4f193b897d57e0ffa9c3bfdd598e0c3c4201195

* Target version: master 

### Checklist


eszkadev (approved):
thanks, that was missing

PR #7717, opened 2023-11-24
annotation: avoid checking uninitialized comment for editing

lpranam:
problem:
Uninitialized comments don't have properties set,
which made them seem like comments are being edited. This caused problem when checked if any comments are being edited, and couldn't unselect comments sometimes

Change-Id: I22a465627d361cb3fe42fb69e9c197f0b0e69755

* Target version: master 

### Checklist


lpranam:
it's now part of #7761

PR #7721, opened 2023-11-24
framed.doc: Relayout the post message test harness

hfiguiere:
Change-Id: Idbd1b5d89ea53d4f9925b727e8db2854ac23b17f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -15,6 +15,10 @@
     <title>Online Editor</title>
 
     <style>
+      :root {
+          font-family: "sans-serif";
+      }
+
       .alert {
         font-size: large;
         padding: 20px;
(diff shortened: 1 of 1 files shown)


hfiguiere:
This is how it looks:

image

mmeeks:
ooh - looks beautiful =) except that (somehow, one again) I see the classic toolbar which is an abomination =) how does that thing keep coming back like the bad penny =)

mmeeks (approved):
Looks really good to me Hubert =)

browser/html/framed.doc.html
```diff
@@ -320,76 +356,122 @@ <h4>Example</h4>
...
+          <div class="vbox">
+            <button onclick="CollapseNotebookbar(true); return false;">Collapse Tabbed Toolbar</button>
+            <button onclick="CollapseNotebookbar(false); return false;">Extend Tabbed Toolbar</button>
+          </div>
+        </div>
+      </div>
+      <div class="vbox framed">
+        <h3>Send UNO Commands</h3>
+        <div class="hbox">
+          <button onclick="sendBoldUNOCommand(); return false;">Send Bold UNO Command (and example without args)</button>
```
timar, line 462: *an* example without args

timar:
> ooh - looks beautiful =) except that (somehow, one again) I see the classic toolbar which is an abomination =) how does that thing keep coming back like the bad penny =)

No worries the default is notebookbar. Probably Hubert pressed a few buttons before teking the screenshot, e.g. there are two save icons.

hfiguiere:
> ooh - looks beautiful =) except that (somehow, one again) I see the classic toolbar which is an abomination =) how does that thing keep coming back like the bad penny =)

Probably because I was testing stuff and it persist into the local storage in the browser?

browser/html/framed.doc.html
hfiguiere: done

hfiguiere:
I have added a fix for the "Show Hide save". Print did work thanks to #7589

mmeeks (approved):
Looks great to me thanks Hubert =)

Issue #7369, opened 2023-10-05
Calc: groupping doesn't work 


pedropintosilva:
If you have multiple groups, one inside of the other then it seems impossible to show/hide all of them. It seems that only the top parent becomes responsive and all the others do not react to the click.

This was reported by Julius in the chat. Specifically: when trying to work with column groups

**But** both parent and child groups need to start at the same column.

1. Select A to F Columns
2. Data tab -> Group
3. Select A to C
4. Data Tab -> Group

Now collapsing the child group also collapses the parent group

pedropintosilva:
let me assigned to you @bayramcicek : )

bayramcicek:
notes:
- double clicking on a group's tail closes the groups correctly.
- but with expand (`+`) and collapse (`-`) buttons, only the top-parent is responsive to the click.

timur-g:
Fixed in 23.05.

PR #7722, opened 2023-11-25
Calc: make grouping responsive for each group level

bayramcicek:
Change-Id: I68572eb126acf4a00ed96af870c1f908083aec8b

* Resolves: #7369 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.ColumnGroup.ts
+++ b/browser/src/control/Control.ColumnGroup.ts
@@ -184,8 +184,8 @@ export class ColumnGroup extends GroupBase {
 						const startX = this.getRelativeX(group_.startPos);
 						const startY = this._levelSpacing + (this._groupHeadSize + this._levelSpacing) * group_.level;
 						const endX = startX + this._groupHeadSize;
-						const endY = group_.endPos + this._cornerHeaderWidth - this.documentTopLeft[1];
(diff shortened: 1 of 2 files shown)


bayramcicek:
I see that group-related codes are added by @gokaysatir here https://github.com/CollaboraOnline/online/pull/1221 . 
Adding @gokaysatir as a reviewer too.

gokaysatir (approved):
Thanks for the fix @bayramcicek :)

pedropintosilva:
I somehow encounter some weird behaviour: 

https://github.com/CollaboraOnline/online/assets/65948705/48c80089-db86-4985-a8e5-40e7da3f4444

pedropintosilva (approved):
So WRT the video I posted. I have double checked and that particular bug is already reproducible on master without this commit. Furthermore I have tested this PR and it fixes the problems reported in https://github.com/CollaboraOnline/online/issues/7369 : 
* With this PR: clicking in the control boxes (expander buttons) correctly expands/collapses the respective group instead of activating the Parent
* It also fixes hidden children control boxes (expanders). Before the mouse cursor would switch to the pointer hand when hovering above headers that were hidden (because they were collpased by their parent). Now the cursor doesn't change and there is no click event at all.

PR #7723, opened 2023-11-27
jsdialog: close dropdown properly

eszkadev:
Make sure we don't expect response from server for online-only component: dropdown which doesn't have corresponding widget in the core

fixes regression on close from:
commit a23adac78c34d3ea1717954fcab1a3caaeef0bbb
browser: fix undefined property 'isPopup'


hcvcastro:
Looks good

hcvcastro (approved):

PR #7727, opened 2023-11-27
Fix fuzzers build

vmiklos:
- the HttpEcho build broke with commit
  08d9081280f80094194bd8501030fdd2e6dde6af (net: don't try to set
  TCP_NODELAY on local Unix sockets., 2023-10-30)

- fix an unused variable error from commit
  0631593c96289de2cc26d8efdfd6155547fe740c (wasm: proxy wopi documents,
  2023-11-06).

- the undefined reference to COOLWSD::ForKitProcId probably went wrong
  in commit 3f46c1db445b9840318af0a4994f0c3ff1012285 (kit-in-process:
  pure re-factor to a run-time function to flag this., 2023-11-20)

Change-Id: I0cf06d188860bdb2f795485a91c7634b596255aa


vmiklos:
Now that fuzzers work, the admin fuzzer found a problem, adding a 2nd commit to fix that.

vmiklos:

This is just a fallout from me trying to teach Jaume how to set up fuzzers. First fix the build the fix a find where a bad message to the admin console can lead to std::terminate().

caolanm (approved):
seems sane, and safe if there is a throw

PR #7729, opened 2023-11-27
slide pane container height

hcvcastro:
- browser: fix slide pane height container
- cypress: add unit test to check slide pane height


caolanm (approved):
seems to make sense and has a welcome cypress test for it

PR #7730, opened 2023-11-28
fix(helm): aliases is an regex

genofire:
### Checklist

### DIFF
--- a/kubernetes/helm/collabora-online/README.md
+++ b/kubernetes/helm/collabora-online/README.md
@@ -46,8 +46,8 @@ In order for Collaborative Editing and copy/paste to function correctly on kuber
       hosts:
          - host: chart-example.local
             paths:
-            - path: /
-            pathType: ImplementationSpecific
+              - path: /
+                pathType: ImplementationSpecific
 
    image:
       tag: "latest"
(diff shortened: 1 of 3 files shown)


Rash419 (changes requested):
for example if I have

```
collabora:
  aliasgroups:
    - host: "http://192.168.1.112"
      aliases: ["test.collabora.online", "test.integrator.online"]
```
The output as per the current patch is 
`aliasgroup1: http://192.168.1.112|test.collabora.online|test.nextcloud.online`

But we want
`aliasgroup1: http://192.168.1.112, http://test.collabora.online|text.nextcloud.online`

Darshan-upadhyay1110:
Hi @genofire! I hope you're doing well. I noticed @Rash419 has suggested a few small changes on the patch. Could you kindly take a look at them when you have a moment? It would be wonderful to get this patch into master as soon as possible. Thanks a ton! 😊

genofire:
hmm, maybe i have not try it with `protocol://` on aliases.

so an example like on @Rash419 does not work on my privat setup.

eszkadev:
@Rash419 what should we do with this PR?

Issue #7701, opened 2023-11-23
Android: "Insert Cells" Menu Misaligned Radio Buttons


Tex2002ans:
### Describe the Bug

In Calc, the "Insert Cells" menu option radio buttons are all misaligned:

<img alt="Collabora 23 05 5 2 Android - Insert Cells Radio Buttons Misaligned" src="https://github.com/CollaboraOnline/online/assets/5774324/56bd766c-c01f-4d7f-a62a-f2f591d03cbe" width="40%" />

### Steps to Reproduce

In Calc:

1. **Press+Hold** on a cell until "Cut/Copy/Paste/Insert" menu shows up.
2. **Click** on "Insert..."

See the "Insert Cells" Selection menu with these options:

- Shift cells down
- Shift cells right
- Entire row
- Entire column

### Actual Behavior

Each of the checkboxes is aligned like:

          o
        o
      o 
    o

### Expected Behavior

Should all be right-aligned like:

        o
        o
        o
        o

### Smartphone

- **Device:** OnePlus 6T
- **OS:** Android 11 (OxygenOS 11.1.2.2)

> COOLWSD version: 23.05.4.1
> git hash: b3abc03d
> LOKit version:
> Collabora Office 23.05.4.1
> git hash: 966b9f8

### Additional Context

It looks like the text labels also have absolutely 0 margins along the left, so they appear right up against the left-edge of the screen.

PR #7736, opened 2023-11-28
Android: Insert Cells Menu Misaligned Radio Buttons

codewithvk:
Change-Id: Ie6bdb2423f92b5959a2233ee0dccc6f8e2e5dc35

* Resolves: #7701
* Target version: master

### DIFF
--- a/browser/css/mobilewizard.css
+++ b/browser/css/mobilewizard.css
@@ -836,7 +836,8 @@ div#mobile-wizard-content .spinfieldcontainer .spinfieldimage.disabled {
 #mobile-wizard #top{background: url('images/sc_aligntop.svg') no-repeat center;}
 #mobile-wizard #bottom{background: url('images/sc_alignbottom.svg') no-repeat center;}
 #mobile-wizard #standard{background: url('images/sc_alignverticalcenter.svg') no-repeat center;}
-.checkbutton{
+.checkbutton,
+.radiobutton{
 	clear: both;
 }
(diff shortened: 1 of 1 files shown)


Darshan-upadhyay1110 (commented):
image
image


Now it looks good in mobile.

A small nit pick please make sure we are consistent with element that has checkboxes.

here if you could change margin to  'margin: 10px 20px 10px 0px !important;` ( in the new css property as you have added)
Same as previous case.

Darshan-upadhyay1110:
@codewithvk i won't mind if you will add more consistency in element margin and padding in mobile view that will be great 😉 

For example: - make sure starting or ending of any element  in mobile wizard should be same. Either it is `radioButton` or `checkbox`

codewithvk:
> @codewithvk i won't mind if you will add more consistency in element margin and padding in mobile view that will be great 😉
> 
> For example: - make sure starting or ending of any element in mobile wizard should be same. Either it is `radioButton` or `checkbox`

That's a great suggestion! I will consider it and try to create a follow-up PR if needed.

Darshan-upadhyay1110 (approved):

codewithvk:
@Darshan-upadhyay1110 Found another interesting approach. It is much more consistent and better than the older ones. Could you please review it again? : )

Darshan-upadhyay1110 (approved):

gokaysatir (approved):
Checked, thanks for the fix :)

PR #7738, opened 2023-11-28
postmessage: Added Show_StatusBar and Hide_StatusBar

hfiguiere:
* Resolves: # 
* Target version: master 

### Summary

postmessage: Added `Show_StatusBar` and `Hide_StatusBar` to show / hide the status bar, like with the ui_defaults.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -184,6 +184,12 @@
           post({'MessageId': messageId});
       }
 
+      function ShowStatusBar(visible) {
+          var messageId = visible ? 'Show_StatusBar' : 'Hide_StatusBar';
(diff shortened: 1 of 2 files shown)


hfiguiere:
> do we need to send the Host_PostMessageReady repeatedly ?

In the SDK example I have handled differently and I think we can.

```javascript
function receiveMessage(event) {
    let msg = JSON.parse(event.data);
    if (msg.MessageId == 'App_LoadingStatus') {
        if (msg.Values) {
            if (msg.Values.Status == 'Document_Loaded') {
                post({'MessageId': 'Host_PostmessageReady'});
            }
        }
    }
}
```

PR #7743, opened 2023-11-29
typically contemporary browsers don't cache files over 50M by default

caolanm:
So warn when we serve something over that size that it won't typically be cached.

The default max individual file size that firefox caches is that of about:config browser.cache.disk.max_entry_size which is 50M out of the box

It's not quite as clear in chrome to see what the size it, but in practice I see the same behaviour by default.

Change-Id: Ifa484e08ad64a7f4e7970d828d4c209a46e26fb2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665 (approved):
Makes sense and looks good to me

mmeeks (changes requested):
Lets have a chat about this =)

PR #7744, opened 2023-11-29
Fix the thickness dialog is shown cut

GulsahKose:
Thickness dialog has an image in. When the image render is delayed we are getting image size 0x0. That wrong size caused wrong resizing of the parent containers. To show better result, we need to get image's actual size from core.

Change-Id: I207fd2fcca63deee21f51ef34c1ad670e15afd87

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

eszkadev:
requires: https://gerrit.libreoffice.org/c/core/+/160039

PR #7760, opened 2023-11-30
Sometimes we only request two rows of tiles in writer on an initial load

caolanm:
Sometimes we only request and only get two rows of tiles in writer on an initial load. Most frequently I see this with a non-debug online server via nextcloud of the default "Welcome to Nextcloud" docx in firefox in the 2nd or 3rd tab of opening the same document. While the page doesn't visually resize there is a resize delivered after the initial request of tiles.

Change-Id: Id263adbbd98ad0a7c6143d241d0d91e1c28eaac5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -79,14 +79,6 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			}
 		}.bind(this));
 
-		// This is called when page size is increased
-		// the content of the cells that become visible may stay empty
-		// unless we have the tiles in the cache already
-		// This will only fetch the tiles which are invalid or does not exist
-		map.on('sizeincreased', function() {
-			this._update();
-		}.bind(this));
-
 		app.sectionContainer.addSection(new app.definitions.AutoFillMarkerSection());
 
 		this.insertMode = false;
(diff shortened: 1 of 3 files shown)


Ashod (commented):
Could this be related to the threading issue, fixed in #7820? And if so, would this patch still be preferred?

caolanm:
Had this again today with (non-debug) online ba536c30 and core 2e57d87

Captured the deferred log from protocol dumping: [example.log](https://github.com/CollaboraOnline/online/files/13661972/example.log)

What I see as unusual vs the working case is the clientvisiblearea lines
1702477719537.OUTGOING.clientvisiblearea x=-8160 y=0 width=28785 height=1575 splitx=0 splity=0
which is unusually short (and not representative of my actual browser height)
and then
1702477721957.OUTGOING.clientvisiblearea x=-8160 y=0 width=28785 height=12075 splitx=0 splity=0
which is not followed by any tiles, so I end up with just one row at the top of the document

caolanm:
And I managed to capture a log with this patch applied where we still have that initial:

1702479723439.OUTGOING.clientvisiblearea x=-8160 y=0 width=28785 height=1575 splitx=0 splity=0
but end up with a full screen of tiles.

 [after.log](https://github.com/CollaboraOnline/online/files/13662200/after.log)

eszkadev (approved):

PR #7761, opened 2023-11-30
Annotation: revamp comment autosaving

lpranam:
* Target version: master 

### Summary
Reverts all the previous comment autosave commits and implements a new approach.

Original approach before comment autosave:
When we save a new comment, original DOM element which user was editing
is removed from the DOM and new comment DOM element is created from the data and message sent by core.

Previous comment autosave approach:
When we save a new comment, original DOM element which user was editing
is not removed but kept in editing mode. We register the comment message sent by the core
but react and create new comment DOM element only after user clicks on save or cancel.
This lazy initialization caused many regression due to juggling of the new and old comment DOM element.
All these problems were in JS side only, reloading doc would bring comment in correct state.
This approach gave smoother transition between different comment states without any flickers.

New comment autosave approach:
Simple as the original approach, when comment loses focus we save it as normal comment.
When new DOM element created via core comment message, we immidietly set it to edit mode.
This approach may have a little flicker when we replace old comment DOM element with new comment DOM element.
Overall this approach requires less condition checking which makes it easy to maintain and hopefully reduces any regression from previous method.

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -766,6 +766,11 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	font-size: var(--default-font-size);
 }
 
+.cool-annotation-autosavelabel {
+	font-size: var(--default-font-size);
+	opacity: 75%;
+}
+
 .cool-annotation-menubar {
 	margin: 0;
 	padding: 0;
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -766,11 +766,6 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	font-size: var(--default-font-size);
 }
 
-.cool-annotation-autosavelabel {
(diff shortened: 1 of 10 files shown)


lpranam:
@eszkadev isn't that a good thing? it is intentional. our competitors can do similar things.

However, on the other hand, I could not reproduce the overlapping of the comment (my branch was on top of the latest master just in case that affected, so I rebased the PR can you test again please?)

lpranam:
@eszkadev @thebearon pushed a new commit that should take care of the issues you mentioned :)

eszkadev:
When document is narrow and collapsed comments mode.:
1. insert comment
2. type something
3. click outside to autosave
Result: comment is half visible

collapsed

I saw bug report already so not caused by this PR probably.

eszkadev (approved):
Works reasonably well and is much simpler. Let's test this more with additional cypress tests from other PR you have then.

PR #7764, opened 2023-12-01
Add list of allowed data sources

eszkadev:
- uses new lok_allow setting to provide such list
- for backward compatibility uses also all post_allow and storage.wopi entries
- Used for example in: PostMessage Action_InsertGraphics, =WEBSERVICE() function, external reference in the cell


caolanm (approved):
With it in place if I visit https://www.collaboraoffice.com/ and select some text before the image and after, to include the image and paste that into the test build then it won't show the image. The clipboard data presumably contains the link to the image so blocked.

But just right click the image and paste just that then it will work, pasted as image data through browser presumably.

PR #7767, opened 2023-12-01
wsd: do not remove quarantine files explicitly

Ashod:
Let clearOldQuarantineVersions() remove quarantined
files only when necessary and per the config settings.

Change-Id: Ie1fb18c02d61a710546e9b5962ab8b7973c2066e


lpranam (approved):
code looks good :)

PR #7768, opened 2023-12-01
calc: split pane mis-rendering.

mmeeks:
Invalidations were not cropped to the split pane, such that large invalidations that crossed from the mane pane over another top/left split pane, would cause partial re-render, flicker and other problems.

Regression from: 0453140f6180c4f82ec66c9515014b7341d732c4

Potential fix is to have a complex multi-rectangle clip region, or to only do this for simple cases, which are common: of invalidation within a single split pane, doing a complete re-render elsewhere.

Change-Id: I255806286f6c80a233d4c6ade2bcb4e12f2bf753

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -718,7 +718,10 @@ class CanvasSectionContainer {
 	private setDirty(coords: any) {
 		if (this.dirty == DirtyType.All)
 			return;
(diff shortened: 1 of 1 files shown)


mmeeks:
After some review, should be included in the next stable release I think; this commit shipped in version 23.05.5.4.

caolanm (approved):

PR #7772, opened 2023-12-03
annotation: avoid wheel event handling for comments

### LINKED ISSUE #7735: Lengtlhy comments are cropped, mouse scrolling doesn't work
### Describe the Bug
Long comments gets cropped and the scrolling via mouse wheel doesn't work; scroll bar is almost invisible;
@nimishavijay could you pretty please drop here the screenshot you have showed me?

lpranam:
it was previously required to handle comment scrolling separately but not anymore it was introduced in a3e59d1

current problem:
when a comment is very long that it overflows and has a scroll bar, it is only possible to scroll using the scroll bar and mouse wheel didn't work

fixes #7735

Change-Id: I6d9f85c61b837f86e021a795e8af43b375a318d1

* Resolves: #7735
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentSection.ts
+++ b/browser/src/layer/tile/CommentSection.ts
@@ -146,13 +146,6 @@ export class Comment extends CanvasSectionObject {
 			},
 			this);
 
-		if ((<any>window).mode.isDesktop()) {
-			L.DomEvent.on(this.sectionProperties.container, {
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
I have tested. Now the scroll wheel works. I can scroll the comment.

PR #7775, opened 2023-12-04
Make sure to set the correct normalized view id for the tiles.

gokaysatir:
When user switches to dark theme, previously requested tiles' normalizedViewIds may need to be updated.

Issue:
Users sometimes were getting tiles from the wrong theme.

Change-Id: Ie3324eb3f4879da5bfd0738513409e190422a9c0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3198,6 +3198,12 @@ void DocumentBroker::handleTileCombinedRequest(TileCombined& tileCombined, bool 
     // Drop duplicated tiles, but use newer version number
     else
     {
+        // Make sure that the old request has the same normalizedviewid with the new request.
+        for (size_t i = 0; i < requestedTiles.size(); i++) {
+            if (requestedTiles[i].getNormalizedViewId() != session->getCanonicalViewId())
+                requestedTiles[i].setNormalizedViewId(session->getCanonicalViewId());
+        }
+
         for (const auto& newTile : tileCombined.getTiles())
         {
             bool tileFound = false;


mmeeks (changes requested):
We should **never** trust a 'normalizedViewId' that comes from a client over a websocket. In fact - it's so bad doing that, that we should add code that wipes if it is not already there in our incoming ClientSession. Any TileCombine that comes in should have its normalizedViewId reset to the view's normalizedViewId. We should only believe such a view-Id when it comes from Kit itself.

Can you check that this never happens. Also - I'd love a unit test.

Thanks Gokay != =)

gokaysatir:
> We should **never** trust a 'normalizedViewId' that comes from a client over a websocket. In fact - it's so bad doing that, that we should add code that wipes if it is not already there in our incoming ClientSession. Any TileCombine that comes in should have its normalizedViewId reset to the view's normalizedViewId. We should only believe such a view-Id when it comes from Kit itself.
> 
> Can you check that this never happens. Also - I'd love a unit test.
> 
> Thanks Gokay != =)

Thank you for the review. I checked the implementation and saw that normalizedViewId is set by ClientSession before the "handleTileCombinedRequest" is called. I also changed the code to get it from the session in order to be on the safe side.

I will prepare a unit test now, can you have a second look when you have time?

caolanm:
There's a few normalized view id mismatch issues around at the moment. Is there a reproducible scenario for this one?

There was the theme-change-didn't-do-anything issue for core:
https://gerrit.libreoffice.org/c/core/+/159992

And there is also the https://github.com/CollaboraOnline/online/pull/7780 reconnect case which has a howto reproduce, and I'd like to consider them all together.

gokaysatir:
> There's a few normalized view id mismatch issues around at the moment. Is there a reproducible scenario for this one?
> 
> There was the theme-change-didn't-do-anything issue for core: https://gerrit.libreoffice.org/c/core/+/159992
> 
> And there is also the #7780 reconnect case which has a howto reproduce, and I'd like to consider them all together.

Yes it is still reproducible with latest core and online branches.

Reproducer:
* Open 2 views of the document.
* Start typing on one of them.
* Stop typing and quickly click on Dark Mode button.
* Somehow difficult to reproduce at first try but certainly possible.

s4

gokaysatir:
If i am reading the situation correctly (a timing issue), it will be difficult to write a unit test for this. I already created the unit test but it doesn't always fail :) It sometimes fails. Anyway, maybe someone can have a better idea before i am finished with the test. I haven't sent it yet, test is a mess for now.

caolanm:
> Any TileCombine that comes in should have its normalizedViewId reset to the view's normalizedViewId.

We appear to be doing this, we always setNormalizedViewI to the ClientSession one, but if we end up with unprocessed requests, f.e. after hitting tilesOnFlyUpperLimit limit and then get a theme change then the contents of ClientSession::_requestedTiles still have the old id

I guess it would also work to change whatever is pending in _requestedTiles when NormalizedViewId/CanonicalViewId changes.

Had a stab at making a test for this, but proving elusive so far.

caolanm (approved):

PR #7776, opened 2023-12-04
Show info modal on network access error

eszkadev:
If administrator doesn't allow some URLs in the coolwsd.xml under net.lok_allow then when pasting webpage with images
- it might result in missing pictures. We should show the user what happened.


pedropintosilva:
Thank you

PR #7779, opened 2023-12-05
Fix icon wrong url caused by data.icon property

Darshan-upadhyay1110:
Change-Id: I44cf079b8dca1395d31e8b786ec69b74a76424cc

- some icons are not set correct because of typeo
- we need to use data. Image not the data. Icon

Change-Id: Ia112202fdebec2f3529f3bbd19d46084d4e0dc1a


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2410,8 +2410,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				button.setAttribute('aria-haspopup', true);
 
 			var buttonImage = L.DomUtil.create('img', '', button);
-			var icon = data.icon ? data.icon : builder._createIconURL(data.command);
-			L.LOUtil.setImage(buttonImage, icon.split('/').pop(), builder.map);
+			if (data.image) {
```
eszkadev, line 2413: I think we need also .icon   for  integrator's custome icons in notebookbar shortcuts bar:
https://github.com/CollaboraOnline/online/blob/f203a4d68f017a4c578deca317f559549a1b06c3/browser/src/control/Control.Notebookbar.js#L326

You can check that in the postMessage tester I think

eszkadev:
could you put example where to find broken case in the commit message? it is useful when in the future we will debug some related problem

eszkadev (approved):
looks good

thebearon:
This is a nitpick, not a big deal, but in the future if there is a regression, please mention the regressing commit hash (f798be095cdb39809f72a8e06e1b25ea1c29b0b0 in this case), and add a brief description of the issue it caused. Thanks!

Darshan-upadhyay1110:
Understood I will remember it, thanks :)

pedropintosilva (approved):

PR #7780, opened 2023-12-05
fix: tiles not updated after reconnection

Rash419:
from idle

The following scenario was affected:
- user A and B both goes idle
- user B starts typing
- user A reconnects, all tiles updated by user B were not getting reflected for user A

Mismatch of canonical ids was causing the issue. On reconnection, tilecombine messages were sent before uno:ChangeTheme can update the canonical id on server side, therefore the received tile cache from server were from old canonical id(https://github.com/CollaboraOnline/online/blob/4a8974d107896c7fb4b7c35e326b52a986144a8b/wsd/ClientSession.cpp#L1379).

By reseting the previousTheme on connection we make sure there after canonicalidchange we invalidate all the tiles

Change-Id: I874bfe0bd71d176bacf0c7aa768e49613535ebd5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1557,6 +1557,7 @@ app.definitions.Socket = L.Class.extend({
 				this._map._isCursorVisible = false;
 
 			this._map._docLayer._resetCanonicalIdStatus();
+			this._map.uiManager.previousTheme = null;
 		}
 
 		if (isActive && this._reconnecting) {


browser/src/core/Socket.js
```diff
@@ -1557,6 +1557,7 @@ app.definitions.Socket = L.Class.extend({
 				this._map._isCursorVisible = false;
 
 			this._map._docLayer._resetCanonicalIdStatus();
+			this._map.uiManager.previousTheme = null;
```
caolanm, line 1560: If _resetCanonicalIdStatus sets this._canonicalIdInitialized to false I then see that we have:

if (!this._canonicalIdInitialized) {
        this._canonicalIdInitialized = true;
        this._update();
} else if (viewRenderedState !== this._map.uiManager.previousTheme) {
        this._requestNewTiles();
        this._invalidateAllPreviews();
        this.redraw();
}

inside else if (textMsg.startsWith('canonicalidchange:')) 

does that mean we get two canonicalidchange messages on a reconnect?, one setting this._canonicalIdInitialized back to true, and then a second one which triggers the else this._requestNewTiles ... ?

Rash419: Yes we get two canonicalidchange message when we load/reconnect

Rash419: This might be interesting
https://github.com/CollaboraOnline/online/pull/7324

caolanm (approved):
Yeah, so debugged a bit more and I see that this works and isn't harmful.

I think there's an underlying oddness a viewid 1000 which I think is the reason we get stale tiles back and I'll have a look at a later follow up on that thought.

PR #7782, opened 2023-12-05
Private/mmeeks/moretiles

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3340,10 +3340,10 @@ void DocumentBroker::sendRequestedTiles(const std::shared_ptr<ClientSession>& se
 
     size_t tilesOnFlyUpperLimit = session->getTilesOnFlyUpperLimit();
 
-    // Drop tiles which we are waiting for too long
-    session->removeOutdatedTilesOnFly();
-
     auto now = std::chrono::steady_clock::now();
+
+    // Drop tiles which we are waiting for too long
+    session->removeOutdatedTilesOnFly(now);
 
     // All tiles were processed on client side that we sent last time, so we can send
     // a new batch of tiles which was invalidated / requested in the meantime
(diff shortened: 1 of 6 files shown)


mmeeks:
WIP patch - but would love feedback. I'm looking at re-working tileprocessed - so we can ignore deltas on the wire since they are rather small B/W wise; I'd like to get >=2 items down the connection if possible

wsd/DocumentBroker.cpp
```diff
@@ -3405,8 +3404,7 @@ void DocumentBroker::sendRequestedTiles(const std::shared_ptr<ClientSession>& se
                     tilesNeedsRendering.push_back(tile);
                     _debugRenderedTileCount++;
                 }
-                if (tileCache().subscribeToTileRendering(tile, session, now))
-                    beingRendered++;
+                tileCache().subscribeToTileRendering(tile, session, now);
```
caolanm, line 3407: presumably the return value from subscribeToTileRendering can be dropped again after that, as no more users of it

caolanm: Dropping the beingRendered would have, presumably, also gotten me out of the hole I found myself in during ed8f95c1bc5ac9a2e5c7dd2fb17d7a94f9d8b2b5 trying to get the empty-update tests in place.

I wonder should we continue to "subscribeToTileRendering" + "sendTileCombine" in handleTileCombinedRequest before calling "sendRequestedTiles" which also does those things? Is that a duplication of effort?

wsd/DocumentBroker.cpp
```diff
@@ -3340,38 +3340,23 @@ void DocumentBroker::sendRequestedTiles(const std::shared_ptr<ClientSession>& se
...
-              // If we delayed all tiles we don't send any tile (we will when next tileprocessed message arrives)
-              delayedTiles < requestedTiles.size())
+        while (!requestedTiles.empty() &&
+               session->getTilesOnFlyCount() < tilesOnFlyUpperLimit)
         {
             TileDesc& tile = *(requestedTiles.begin());
 
-            // We already sent out two versions of the same tile, let's not send the third one
-            // until we get a tileprocessed message for this specific tile.
-            if (session->countIdenticalTilesOnFly(tile) >= 2)
```
caolanm, line 3366: oddly enough TileCacheTests::testLimitTileVersionsOnFly apparently tries to test for this but seeing as ci is green the test didn't catch this change

caolanm (approved):

PR #7784, opened 2023-12-05
add hasWASMSupport to capabilities

caolanm:
so the launching site can query if wasm is enabled to see if it needs to insert the required headers, so we can then in turn make it optional on those being set to also require matching headers.

Change-Id: Icd73081809abb8098c21bc61a8357869db45ff6c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected

Issue #7406, opened 2023-10-11
Draw: Display Grid


hfiguiere:
### Is your feature request related to a problem?

In Draw in LibreOffice offline there is a grid.  View > Grid & Helpines > Display Grid, or the toolbar:

![image](https://github.com/CollaboraOnline/online/assets/114441/01a0f4a3-7cd6-4336-9bc2-f91a49c794ad)

The grid:

![image](https://github.com/CollaboraOnline/online/assets/114441/d7485091-e600-4e59-83c2-1b27e95f1f15)

It's not available in Draw online.

### Describe the solution you'd like

Allow display the grid. Also allow also snap to grid and other helpings.

### Describe alternatives you've considered
n/a

### Additional context

n/a

hfiguiere:
Possibly #4893 (Impress)

Ezinnem:
It could be regarded as an enhancement request

hfiguiere:
Possibly a dupe of #4556

JoshuaPettus:
Mentioned in the other thread,  also what goes hand and hand with the grid is the horizontal and vertical guide rulers.  Don't know how hard to implement but it is already part of LO for draw and impress

hfiguiere:
One other issue is that currently setting the grid in a document on deskop (LibreOffice) will mean the grids are displayed in Draw or Impress, with no possibility to remove them.

eszkadev:
implemented in https://github.com/CollaboraOnline/online/pull/10147

Issue #4556, opened 2022-04-06
Draw Grid and Helplines


JoshuaPettus:
I realize the draw portion of collaboraonline is extremely new, but one feature I would like to see from the regular LibreOffice is Draw's grid and helplines feature.  Honestly it's a must have if I'm going to use this as a real flowchart/drawing app.

Andreas-Kainz:
at least the setting isn't perfect in LibreOffice, but yes, without any snap option draw isn't usefull.

JoshuaPettus:
Oh definitely...  I feel in general Draw could use some more polish in LO to make it what it truly deserves be.  But it has been slowly making progress and I'm hopeful. I'd love to see this web version be as useful.

eszkadev:
implemented in: https://github.com/CollaboraOnline/online/pull/10147

PR #7786, opened 2023-12-05
draw: Added Grid

hfiguiere:
* Resolves: #7406 
* Also see: #4556
* Target version: master 

### Summary

This provide Grid visibility in both draw and impress, as well as enable snap to grid.

### TODO

- [x] Figure out how the have the action state report to the UI - this need a core patch
  - TBD
- [ ] Snap to grid doesn't seem to be implemented and maybe need to be done first
- [ ] Validate the location of the various UI items

### Checklist


eszkadev:
cc : @gokaysatir    it something from this PR useful now? I think we have competing solution merged already.
If not - can we close the PR? @hfiguiere

hfiguiere:
I don't see where the "GridVisible" command is made available beside the Shift F9 shortcut.

Also, it should be noted #9019 called for disabling it in LOKit as a workaround. 

So it remain relevant but #9019 should be fixed first.

gokaysatir:
I think we could use a section for the task? Also we can use simplePoint class instead of point i think.

I think i'd prefer to position the Path.Constraints.ts file as a section inside canvas->sections directory. If it cannot be the section itself but a helper, then we have "app" folder for middlewares. But it seems we need a section, either as a base class or with the content of Constraints.ts file.

If we can solve this without touching Path.Drag.js file, that'd be awesome.

All above should be possible. Can you check CanvasSectionContainer and a few sections if you are new to structure and share your comments? That's the new way we use and if you use too much of old structures, we may need to refactor your code in the future.

hfiguiere:
oh so you are actually talking about this PR https://github.com/CollaboraOnline/online/pull/7958

(this one is currently on top of it)

gokaysatir:
> oh so you are actually talking about this PR #7958 - "draw: constrain object move with Shift "
> 
> (this one is currently on top of it)

Ah, yes it seems, they seem to share code.

hfiguiere:
Note that #7958 is now broken since some of the front-end code got re-architectured.

eszkadev:
closing, we just merged: https://github.com/CollaboraOnline/online/pull/10147

PR #7789, opened 2023-12-06
notebookbar: don't send events to core

eszkadev:
switching tabs should not send events to the core
notebookbar is online-only widget


gokaysatir (approved):
Thanks for the improvement.

PR #7790, opened 2023-12-06
Do not refresh notebookbar when we start in read-only mode for txt an…

timar:
…d csv

This is a workaround for the bug, when a txt or a csv file was opened, and there were both the classic menu and the menu of the notebookbar at the same time at the top of the window. This bug occurred because regardless of the user's UI mode setting, the read-only mode has the classic menu. The fix also have a side effect, that is better than the original symptom. Now when a txt file is opened, and user starts editing it, the Zotero buttons won't be on the References tab. I think it's a good trade-off, but FIXME.
Eventually, when we won't have the classic UI mode any more, this problem would be solved better, for all corner cased (txt. csv, pdf, read-only shares etc.)

Change-Id: Ic3f95c02a0b75136a0b8069fa8297ed129b63865

* Resolves: # 
* Target version: master 

### Summary
image

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -136,7 +136,7 @@ L.Control.Zotero = L.Control.extend({
 	},
 
 	refreshUI: function () {
-		if (this.map.uiManager.notebookbar)
+		if (this.map.uiManager.notebookbar && !this.map._shouldStartReadOnly())
 			this.map.uiManager.refreshNotebookbar();
 		else
 			this.map.uiManager.refreshMenubar();


eszkadev:
wow, I see, as a temporary workaround it is ok.
But really we shouldn't have notebookbar in the readonly mode... I guess it comes from the recent "force notebookbar mode" changes?

timar:
> wow, I see, as a temporary workaround it is ok. But really we shouldn't have notebookbar in the readonly mode... I guess it comes from the recent "force notebookbar mode" changes?

I don't know. When you press the Edit button, the UI changes to notebookbar, if you had notebookbar set. So I thought the txt/csv case was special, because it starts as read-only, then user can enter editing mode.

PR #7797, opened 2023-12-07
Stop onscreen keyboards closing when used to move

Minion3665:
This relies on core change I5e6c93c64af0d201a8ec045fea5546e189baca74 (https://gerrit.libreoffice.org/c/core/+/160313) as the logic to keep the cell focused is from there.

Here we set the new option when there is an onscreen keyboard hint given by ui_defaults or a postmessage. As ui_defaults are set at page load, we need to wait to send our message until we're initializing everything else to set our options. Additionally, we need to use the socket directly rather than sending this through the map so that this bypasses things like readonly mode

Change-Id: Ie6d80c0e1ebca607b9d705b673ab8008fae39260

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -264,6 +264,18 @@ window.app = {
 		// alternatively, maybe someone else (e.g. an integrator) knows more about the situation than we do. In this case, let's
 		// let them override our default
 		hintOnscreenKeyboard: function(hint) {
+			if (global.app
+					&& global.L.Map
+					&& global.L.Map.THIS._docLayer.isCalc()
+					&& hint !== undefined) {
+				var command = {
+					Enable: {
+						type: 'boolean',
+						value: hint
+					}
+				};
+				global.L.Map.THIS.sendUnoCommand('.uno:MoveKeepInsertMode', command);
+			}
 			global.keyboard.onscreenKeyboardHint = hint;
 		},
 	};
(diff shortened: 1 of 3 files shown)


hfiguiere (approved):
LGTM

Minion3665:
> LGTM

Thanks hub! Reminder to people looking that this shouldn't be merged without its core patch being reviewed and submitted... thanks!

https://gerrit.libreoffice.org/c/core/+/160313

browser/src/core/Socket.js
```diff
@@ -1400,6 +1400,7 @@ app.definitions.Socket = L.Class.extend({
 			this._map.uiManager.closeAll();
 			this._map.setPermission(app.file.permission);
 			window.migrating = false;
+			window.keyboard.hintOnscreenKeyboard(window.keyboard.onscreenKeyboardHint);
```
eszkadev, line 1403: This is calc only as I read in the core commit message. This code is applied to all modules.
Maybe better to do this in CalcTileLayer ?  In onStatus handling also.

browser/js/global.js
```diff
@@ -264,6 +264,18 @@ window.app = {
...
+					&& global.L.Map
+					&& global.L.Map.THIS._docLayer.isCalc()
+					&& hint !== undefined) {
+				var command = {
+					Enable: {
+						type: 'boolean',
+						value: hint
+					}
+				};
+				global.app.socket.sendMessage('uno .uno:MoveKeepInsertMode ' + JSON.stringify(command));
```
eszkadev, line 277: There is a list of allowed commands for read only mode:
https://github.com/CollaboraOnline/online/blob/61aa037ad9315a95e932b894cca75346c65d14f8/browser/src/control/Toolbar.js#L367

so probably it can be sent via `map.sendUnoCommand('.uno:MoveKeepInsertMode', command)` with whitelisting it

eszkadev (approved):

browser/js/global.js
Minion3665: thanks, good call!

eszkadev (approved):

PR #7801, opened 2023-12-07
annotation: avoid updating empty comment section layout

lpranam:
This will help in performance

Change-Id: I09538376cbebf971a4cad605ae193400b651abb7

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -1606,6 +1606,8 @@ export class CommentSection extends CanvasSectionObject {
 	}
 
 	private update (): void {
+		if (this.sectionProperties.commentList.length === 0)
+			return;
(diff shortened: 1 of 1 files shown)


eszkadev (commented):
looks sensible, but @gokaysatir knows best the sections.
Isn't the update in doLayout of `app.view.size.pixels[1]` needed?

gokaysatir:
Indeed i didn't miss this one but i couldn't be sure. Can we delay it until we have a look at dolayout function? This may cause issues in the future.
I tested it and didn't see any issues but lastY variable is assigned inside the layout function regardless of the comment count.
So it may be an issue if all comments are removed and dolayout is not fired.
I may have not be able to test all the cases.

Let's keep this thought in mind until we do some refactoring in this section.

PR #7802, opened 2023-12-08
fix: persistent snackbar after the document get idle

Rash419:
Change-Id: I4ccc93dabb4dd260338d5a941b18ebaf8e42b1ad

- previously, when document was getting idle, it was
  showing a snackbar "The server is disconnected" which
  unnecessary plus the snackbar was persistent even
  after reconnection.
- There are two scenarios to be tested,
 1. It shouldn't show snackbar on idle
    - Set per_document / idle_timeout_secs to something small, eg. 30,
    - Open a document, and wait until it idles out.
    - make sure it doesn't show "The server is disconnected" snackbar
 2. It should show "The server is disconnected" snackbar when you shutdown
    the server, when you start the server again snackbar should disappear

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -87,7 +87,7 @@ L.Control.JSDialog = L.Control.extend({
 	closeAll: function(leaveSnackbar) {
 		var dialogs = Object.keys(this.dialogs);
 		for (var i = 0; i < dialogs.length; i++) {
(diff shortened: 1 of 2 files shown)


eszkadev:
Looks good, but please remove duplicated `Signed-off-by: Rash419 <rashesh.padia@collabora.com>` in commit message.
Also write one sentence what exactly it fixes (how to test), with snackbar message which was persistent, so someone will know how to test this.

eszkadev (approved):

PR #7803, opened 2023-12-08
wasm is derived from the mobile, but doesn't have on-screen keyboard

caolanm:
There's a bit of a mobile/wasm munge with wasm considered a subset of mobile in some places which should be unwound a bit, but for the purposes of usable wasm by default, don't assume that wasm has an onscreen keyboard so we don't lose keystrokes on assuming that there will be an onscreen keyboard appearing when we lose and regain focus.

Change-Id: I0baae0d414ce9aafd1c27d74bcdad2276d104ee5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Makes perfect sense to me =)

PR #7804, opened 2023-12-08
update wasm.README for distro-config

caolanm:
the default suggestion uses --enable-qt5 which isn't needed for LOKit and breaks finding custom notebookbar widgets so online notebookbar doesn't work with it.

Change-Id: I644924a0f8cdb01d4ec20d70ff7aedb27a2c899f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
needs https://gerrit.libreoffice.org/c/core/+/160458 in core

PR #7806, opened 2023-12-08
Make textbox draggable when text is edting

lpranam:
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/vector/SVGGroup.js
+++ b/browser/src/layer/vector/SVGGroup.js
@@ -3,7 +3,7 @@
  * L.SVGGroup
  */
 
-/* global _ */
+/* global _ $ */
 
 L.SVGGroup = L.Layer.extend({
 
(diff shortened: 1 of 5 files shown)


browser/src/layer/vector/SVGGroup.js
```diff
@@ -165,7 +165,7 @@ L.SVGGroup = L.Layer.extend({
 	_onDragStart: function(evt) {
 		if (!this._map || !this._dragShapePresent || !this.dragging)
 			return;
-		if (this._map._docLayer._cursorMarker && this._map._docLayer._cursorMarker.isVisible())
+		if (this._map._docLayer._cursorMarker && $('.leaflet-pane.leaflet-map-pane').css('cursor') === 'text')
```
eszkadev, line 168: This query can fail and cause error.
Also please don't use queries when possible. Can't we get object by getting the section reference and using it's member?

eszkadev:
My impression is that cursor change was always very slow/delayed. Is that reliable approach to rely on that?

eszkadev:
I see that in the core we have 2 modes too: editing text, just shape selection:
selection

It is possible to drag if user has a pointer on a border. Could it be possible to detect here the border too? And then allow dragging.

lpranam:
But the thing is in core you can drag also when you are editing text(which had border) if you start dragging from part of the box which does not have text.

eszkadev:
And that is my question: is that possible to do also in online? I guess we send wrong commands because we are in edit mode. But we could add some overlay element which will do different thing... and will move the shape by sending other command.

lpranam:
close in favor of https://github.com/CollaboraOnline/online/pull/8419

PR #7807, opened 2023-12-10
Update viewId in Writer after recover from idle

eszkadev:
There was a bug in Writer not allowing to successfully reconnect after idle state if we were not a first view in the initial document. Let's update view id even if document size wasn't changed. Added also additional error logging to be sure we never use -1 viewId which is a default "bad value" when we enter some unwanted state in the core.

Steps to reproduce:
1. set per_document.idle_timeout_secs to something small, eg. 30, so sessions time out quickly,
2. Open a document with 2 sessions (remember which one was loaded first - A, important!)
3. wait for it to idle out in both sessions
4. activate SECOND session (B) by click on the document area

Result:
in the browser console you can see an exception:
Exception TypeError: this.map._viewInfo[this.map._docLayer._viewId] is undefined emitting event viewinfo: [{"id":0,"userid":"2","username":"LocalUser#2","readonly":"0","color":411298}] cool.html:359:37

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -405,6 +405,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 				this.requestSheetGeometryData();
 			}
 			this._viewId = parseInt(command.viewid);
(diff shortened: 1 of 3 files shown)


caolanm (approved):
makes a lot of sense to me

PR #7810, opened 2023-12-11
Fix bug in undoing deletion of a parent comment.

gokaysatir:
Bug:
  When a parent and a child comments are removed then user reverts the action respectively:
    * Child comment is revived first and it cannot find its parent.

* Add a possibleParentCommentId to use in case also the parent comment is revived.
* Consider the child comment as a parent until parent appears.

Other changes:
* Remove duplicate call to updateIdIndexMap.
* Remove duplicate call to unselect.
* Remove duplicate call to adjustParentAdd.
* Reorder the 'remove' operation's rows: To avoid unselecting the comment after it is removed.

Change-Id: I2247b23c710236c578b5b5c78d2b7bb45f7a51b4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -999,14 +999,21 @@ export class CommentSection extends CanvasSectionObject {
 	public add (comment: any): cool.Comment {
 		var annotation = new cool.Comment(comment, comment.id === 'new' ? {noMenu: true} : {}, this);
 
-		if (!this.containerObject.addSection(annotation))
-			return;
+		/*
+			Remove if a comment with the same id exists.
+			When user deletes a parent and a child of that parent and undoes the operation respectively:
+				* The first undo: Core side sends the deleted child - this is fine.
+				* The second undo: Core side sends parent and child together - which is not fine. We already had the child with the first undo command.
+			So, delete if a comment already exists and trust core side about the ids of the comments.
+		*/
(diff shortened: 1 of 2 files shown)


browser/src/layer/tile/CommentListSection.ts
```diff
@@ -1062,6 +1068,18 @@ export class CommentSection extends CanvasSectionObject {
 		return index == -1 ? null : this.sectionProperties.commentList[index];
 	}
 
+	private checkIfCommentHasPreAssignedChildren(comment: CommentSection) {
```
eszkadev, line 1071: name could be better, it doesn't check anything (which would be getter/readonly operation)  but rather adds children to the old parent (does modification, which coudn't be guessed from the name)

eszkadev:
Tested in collapsed mode:
1. added structure
```
a
-1
--1a
-2
b
```
2. press ctrl+z to undo all
3. press ctrl+y to redo all
Result: lack of structure

[comments-2023-12-11_14.19.47.webm](https://github.com/CollaboraOnline/online/assets/5307369/3b3528e7-1586-44b1-b51f-e5746663c0f5)

eszkadev:
Removing just parent then undo - OK

Removing thread then undo - undo inserts comments one by one, final order is different ~OK

[comments-2023-12-11_14.23.19.webm](https://github.com/CollaboraOnline/online/assets/5307369/489cc70a-48ac-407a-904b-4c3756a6914a)

PR #7814, opened 2023-12-11
browser: fix sidebar height when the resize is triggered

hcvcastro:
Adjust height values of the slide sorter content element

Change-Id: I04fdb5f89f70b703385e61253bf8ea8a387a65ea


caolanm (approved):
looks reasonable to me

PR #7817, opened 2023-12-11
Refresh cell cursor after recover from idle in Calc

eszkadev:
We were using old cursor data after recovering the view from idle state. For the user there was impression that view jumped at some point - caused by other user action. But it was just an update to real cursor position which scrolled view to it, because previously we were showing some other place in the spreadsheet.

Steps to reproduce bug:

1. Open spreadsheet with 2 views
2. B should have cursor in the bottom part of sheet, let's say row 600
3. Wait for idle in both sessions
4. Open the same spreadsheet with some other user C
5. C does some action on row 300 and goes away (closes tab)
6. User B activates view and selects row 600
7. User A activates view (he should be at the top still)
8. User B changes font size using dropdown in notebookbar, or double clicks on any row header

Result: view A jumps to other place

When we initially connect to the spreadsheet then
requestCellCursor is called in onAdd. Also
_gotFirstCellCursor is not set.

Let's do the same on reconnect so we will have similar effect to clean startup.

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -90,10 +90,14 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		app.sectionContainer.addSection(new app.definitions.AutoFillMarkerSection());
 
 		this.insertMode = false;
+		this._resetInternalState();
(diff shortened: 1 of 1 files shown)


eszkadev:
we should prepare some unified App/Socket state machine for all the modules with clear places where reconnecting happens in the future...

Rash419 (approved):

PR #7819, opened 2023-12-11
annotation: don't change slide/sheet while editing comment 

lpranam:
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -3,7 +3,7 @@
  * L.Control.PartsPreview
  */
 
-/* global _ app $ Hammer w2ui _UNO */
+/* global _ app $ Hammer w2ui _UNO cool */
 L.Control.PartsPreview = L.Control.extend({
 	options: {
 		fetchThumbnail: true,
@@ -499,6 +499,11 @@ L.Control.PartsPreview = L.Control.extend({
 	},
 
 	_setPart: function (e) {
+		if (cool.Comment.isAnyEdit()) {
+			cool.CommentSection.showCommentEditingWarning();
+			return;
+		}
+
 		var part = this._findClickedPart(e.target.parentNode);
 		if (part !== null) {
 			var partId = parseInt(part) - 1; // The first part is just a drop-site for reordering.
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -3,7 +3,7 @@
  * L.Control.Tabs is used to switch sheets in Calc
  */
 
-/* global $ _ _UNO Hammer */
+/* global $ _ _UNO Hammer cool */
 L.Control.Tabs = L.Control.extend({
 	onAdd: function(map) {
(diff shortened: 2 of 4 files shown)


browser/src/control/Control.PartsPreview.js
```diff
@@ -499,6 +499,11 @@ L.Control.PartsPreview = L.Control.extend({
 	},
 
 	_setPart: function (e) {
+		if (cool.Comment.isAnyEdit()) {
```
eszkadev, line 502: All of these different actions from different components are using `this._map.setPart` in the end. Could it be possible to fix it in one place there?

Or these actions have other side effects that could make it more complicated?

browser/src/control/Control.Tabs.js
```diff
@@ -271,6 +271,11 @@ L.Control.Tabs = L.Control.extend({
 	},
 
 	_setPart: function (e) {
+		if (cool.Comment.isAnyEdit()) {
```
eszkadev, line 274: same here

eszkadev (commented):
what about Calc? did you test if the bug happens also there?

eszkadev:
btw: In Calc when I click outside "spreadsheet area", eg. "Bold" button in sidebar - comment is not autosaved

Maybe we should also block sending unoCommands when comment is unser editing? and autosave too.
example what we do for dialogs when they are already opened: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Toolbar.js#L402

browser/src/control/Control.PartsPreview.js
lpranam: It makes it more complicated as these methods are calling other functions like selectPart and deselectAll or _scrollViewToPartPosition. I am afraid I might go into a rabbit hole if we don't stop these actions at their starting point.

eszkadev: I see

lpranam:
> btw: In Calc when I click outside "spreadsheet area", eg. "Bold" button in sidebar - comment is not autosaved
> 
> Maybe we should also block sending unoCommands when comment is unser editing? and autosave too. example what we do for dialogs when they are already opened: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Toolbar.js#L402

This is an interesting case... comment never lost focus so never autosave. But on the other hand I don't wish to block the uno command as it defeats the purpose of the autosave which is to allow editing or doing anything else...

lpranam:
hmmm if it doesn't even lose focus why to force it to autosave, I was testing and I am tempted to keep it as it is.. what's your opinion on that @eszkadev ?

eszkadev:
I don't know, just pointing what I noticed. I think it's not directly related to this fix so no need to solve it here.

But please fix `ctrl + alt + page UP/DOWN` case

eszkadev (approved):
let's get this in and prepare followup for shortcuts later

lpranam:
I had added check in setPart so even shortcut would be blocked :)

PR #7820, opened 2023-12-11
wasm often doesn't get tiles after invalidates

caolanm:
ClientSession::handleTileInvalidation returns early with _tileWidthPixel (etc) of 0 because _tileWidthPixel (etc) never get set

These are supposed to get set from 'clientzoom' which the javascript side sends. clientzoom arrives, but the attempt is made to dispatch via a thread and during startup there are enough messages dispatched via threads to overwhelm wasm

```
bundle.js:56376 ================ handle_cool_message(): 'clientvisiblearea x=-3915 y=0 width=20640 height=7980 splitx=0 splity=0'
bundle.js:56377 Tried to spawn a new thread, but the thread pool is exhausted. This might result in a deadlock unless some threads eventually exit or the code explicitly breaks out to the event loop. If you want to increase the pool size, use setting `-sPTHREAD_POOL_SIZE=...`. If you want to throw an explicit error instead of the risk of deadlocking in those cases, use setting `-sPTHREAD_POOL_SIZE_STRICT=2`.
bundle.js:56376 ================ handle_cool_message(): 'clientzoom tilepixelwidth=256 tilepixelheight=256 tiletwipwidth=1536 tiletwipheight=1536'
bundle.js:56377 Tried to spawn a new thread, but the thread pool is exhausted. This might result in a deadlock unless some threads eventually exit or the code explicitly breaks out to the event loop. If you want to increase the pool size, use setting `-sPTHREAD_POOL_SIZE=...`. If you want to throw an explicit error instead of the risk of deadlocking in those cases, use setting `-sPTHREAD_POOL_SIZE_STRICT=2`.
bundle.js:56376 ================ handle_cool_message(): 'tilecombine...'
```

The 'clientzoom' never get processed and so
ClientSession::handleTileInvalidation returns early with _tileWidthPixel (etc) of 0 because _tileWidthPixel (etc) never get set.

this std::thread dispatching of messages is similar to seen in gtk/mobile.cpp in similar code, and also existed in android/lib/src/main/cpp/androidapp.cpp before getting replaced in:

commit 155718796ea0020a0cff2429813644a6e1c96794
Date:   Fri Jul 19 16:43:44 2019 +0200

    android: Fix a threading / ordering issue.

follow the same simpler android model here.

Change-Id: I3b5e57b3d91bb80fbae90b269300581825518d7f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wasm/wasmapp.cpp
+++ b/wasm/wasmapp.cpp
@@ -148,14 +148,11 @@ void handle_cool_message(const char *string_value)
         LOG_TRC_NOFILE("Actually sending to Online:" << fileURL);
         std::cout << "Loading file [" << fileURL << "]" << std::endl;
 
-        std::thread([]
-                    {
-                        struct pollfd pollfd;
-                        pollfd.fd = fakeClientFd;
-                        pollfd.events = POLLOUT;
-                        fakeSocketPoll(&pollfd, 1, -1);
-                        fakeSocketWrite(fakeClientFd, fileURL.c_str(), fileURL.size());
-                    }).detach();
+        struct pollfd pollfd;
+        pollfd.fd = fakeClientFd;
+        pollfd.events = POLLOUT;
+        fakeSocketPoll(&pollfd, 1, -1);
+        fakeSocketWrite(fakeClientFd, fileURL.c_str(), fileURL.size());
     }
     else if (strcmp(string_value, "BYE") == 0)
     {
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Thanks @caolanm. Good catch. I hadn't seen the "thread pool exhausted" case locally, but I suppose it's highly likely to cause dropped messages and pain. I assume writing directly is tested and works fine.

caolanm:
yeah, or at least I've been running it with this in place for the last day without incident

mmeeks (approved):
Wow - that is very thread'y - no idea why we would need that =) looks sensible to me.

PR #7823, opened 2023-12-12
fix: can't insert mention on pressing 'Enter' key

Rash419:
Change-Id: I2002721dbd92fc44e9b924e67107b627e7176b87

* Target version: master


eszkadev (approved):
looks good

PR #7824, opened 2023-12-12
Fix filter by color in Calc (auto-filter)

Darshan-upadhyay1110:
- We should fire `active` state call back for selected color value in filter by color
 - it will directly apply filter to columns
 - this change is consistent with LO
 
Change-Id: I90977cff67cda1bca78c7d60c132bd0685cadbaa

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -247,7 +247,7 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 	}
 
 	if (!disabled) {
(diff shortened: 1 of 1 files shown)


eszkadev (changes requested):
This affects all the tree views, it shouldn't be like that in all cases.
Could you check if there is maybe set "single click activation" in the core for that particular widget?
Then we should send in the JSON that information, and here make it conditional: regular case sends other message, singleClickActivation sends activate on first click.

eszkadev:
look at line:
```var singleClick = treeViewData.singleclickactivate === 'true' || treeViewData.singleclickactivate === true;```

it seems to be in JSON already. if yes, please use it and also update JSON documentation in the file header.

eszkadev (approved):
now is ok :)

PR #7825, opened 2023-12-12
wasm: limit threads using for scaling

caolanm:
disable this with VCL_NO_THREAD_SCALE for wasm

"Yacht.odt" document with lots of large images fails to render:

Tried to spawn a new thread, but the thread pool is exhausted. This might result in a deadlock unless some threads eventually exit or the code explicitly breaks out to the event loop. If you want to increase the pool size, use setting `-sPTHREAD_POOL_SIZE=...`. If you want to throw an explicit error instead of the risk of deadlocking in those cases, use setting `-sPTHREAD_POOL_SIZE_STRICT=2`.

Change-Id: I8b2f994d2d3d23f8a0043f0c7fefa4d5e250b8bc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -2771,6 +2771,9 @@ void COOLWSD::innerInitialize(Application& self)
         setenv("MAX_CONCURRENCY", std::to_string(maxConcurrency).c_str(), 1);
     }
     LOG_INF("MAX_CONCURRENCY set to " << maxConcurrency << '.');
+#elif defined(__EMSCRIPTEN__)
+    // disable threaded image scaling for wasm for now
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
Prolly we should set MAX_CONCURRENCY=1 as well (?)

caolanm:
Yeah, I wondered, on the other hand I originally tried just setenv MAX_CONCURRENCY 1 in COOLWSD::innerInitialize in an #else to the existing non-wasm/mobile case and that alone wasn't apparently sufficient, while this was, and then with this in place I didn't see further failures so stopped there.

mmeeks (approved):
=)

PR #7826, opened 2023-12-12
calc header: avoid selecting row/col while resizing them

lpranam:
problem:
while resizing row/column it would select all the rows and column user hovers over

regression from: 87a55af

Change-Id: If786c57222b9bb89ac3faa1e645acc223a8dcd7e

* Target version: master 

### Checklist


eszkadev (approved):
works as expected

PR #7829, opened 2023-12-13
Private/eszkadev/annotation fixes

eszkadev:
annotations: isAutoSaved helper

- we need to check if received comment is "our" autosaved one
- without that we were showing comments from others on the
  screen when we were in "autosaved mode"
- remove "interval" to make it more deterministic

This fixes regression from commit 93bdcc349b75ecc0a5268f2577504e8e5ba3dfee
Annotation: revamp comment autosaving

### DIFF
--- a/browser/src/layer/tile/CommentSection.ts
+++ b/browser/src/layer/tile/CommentSection.ts
@@ -754,6 +754,17 @@ export class Comment extends CanvasSectionObject {
 		}
 	}
 
+	// check if this is "our" autosaved comment
+	// core is not aware it's autosaved one so use this simplified detection based on content
+	public isAutoSaved (): boolean {
+		var autoSavedComment = app.view.commentAutoSave;
+		if (!autoSavedComment)
+			return false;
+
+		var authorMatch = autoSavedComment.sectionProperties.data.author === this.sectionProperties.data.author;
+		return authorMatch;
+	}
+
 	private hide (): void {
 		if (this.sectionProperties.data.id === 'new') {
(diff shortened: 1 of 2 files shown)


browser/src/layer/tile/CommentSection.ts
```diff
@@ -754,6 +756,18 @@ export class Comment extends CanvasSectionObject {
...
 
+	// check if this is "our" autosaved comment
+	// core is not aware it's autosaved one so use this simplified detection based on content
+	public isAutoSaved (): boolean {
+		var autoSavedComment = cool.CommentSection.autoSavedComment;
+		if (!autoSavedComment)
+			return false;
+
+		var authorMatch = autoSavedComment.sectionProperties.data.author === this.sectionProperties.data.author;
+		var textMatch = autoSavedComment.sectionProperties.data.text === this.sectionProperties.data.text;
```
lpranam, line 767: This makes reply autosave fail in writer. 

When we reply to any comment we edit into the original comment element in reply textbox, so original comment is set as `autoSavedComment` value and `this.sectionProperties.data.text` is new reply, so these values are not going to match for reply.

lpranam (changes requested):

browser/src/layer/tile/CommentSection.ts
eszkadev: ok, then we should match only author

eszkadev:
fails in `impress/annotation_spec.js`

lpranam:
fixed failing test

lpranam (approved):
Looks good to me

PR #7832, opened 2023-12-14
Fix view pushed up in Calc for autfilter popup

Darshan-upadhyay1110:
- fix view jump by adding display `hidden` in js-dialog overlay 

Change-Id: I4e48fd2d7363db53d4bf3a1d85f9897de3b1e91a

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -12,6 +12,7 @@
 	max-height: 100%;
 	z-index: 2000;
 	display: flex;
+	overflow: hidden;
 }
 
 .jsdialog-overlay.cancellable {
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -523,8 +523,11 @@ L.Control.JSDialog = L.Control.extend({
 		instance.posy = top + offsetY;
 
 		var width = instance.form.getBoundingClientRect().width;
+		var height = instance.form.getBoundingClientRect().height;
 		if (instance.posx + width > window.innerWidth)
 			instance.posx = window.innerWidth - width;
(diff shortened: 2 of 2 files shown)


Darshan-upadhyay1110:
@pedropintosilva i have tested  cases where we have Class  `jsdialog-overlay `
For ex : 
  - Format => Conditional format
  - Adding dropdown in Write using `Form`
  - some cases are in mobile-wizard that also looks good nothing breaks there for pop-ups
  
 All works okay
 
 Do you have any specific case to check for `jsdialog-overlay`  class ? let me know if any other than that it's good to go

eszkadev:
- all the popups in sidebar
- dropdowns like font, font size in notebookbar
- modal dialogs (warnings, errors)
- mobile-wizard
- autofilter
- mentions

eszkadev:
please add more detailed description for the commit message when possible: how to reproduce the problem it fixes

eszkadev:
I can confirm it fixes the "view pushing up" problem. But I see that this is caused because we stopped moving autofilter to be always fully visible. In the past it was not possible to have it like:
auto

eszkadev (approved):
For me it seems to be ok. @pedropintosilva ?

pedropintosilva (changes requested):
hm, but having it then cropped is no bueno

pedropintosilva (commented):
now it fixes for chrome and improves the situation for firefox. Meaning: for any browser it doesn't push the view but for FF it gets cropped

image

eszkadev (approved):

pedropintosilva (approved):
I think we should merge. What do you think?

bottom line: the popover still can get cropped but it is a great improvement from before (the whole UI was being pushed upwards

Issue #7709, opened 2023-11-23
Improve error message on invalid WOPI host


juliushaertl:
### Is your feature request related to a problem?

When setting up Collabora with Nextcloud there are plenty of configuration pitfalls. In case the connectivity works but coolwsd is rejecting the host because it is not an allowed wopi host, the error reporting could be improved.

### Describe the solution you'd like

- A clear error type for the post message, the current websocketconnectionfailed is not very specific and the proxy configuration hint is not helpful in this case
- A better error message than try again as retrying will not help here and the admin needs to take action to configure this properly

### Describe alternatives you've considered

None except to just rely on the logs.

### Additional context

Current error message:

<img width="526" alt="Screenshot 2023-11-23 at 10 38 11" src="https://github.com/CollaboraOnline/online/assets/3404133/77969c55-9793-4516-bd1c-7dae5dd266fd">

Current post message:

```
Action_Load_Resp  { errorType: "websocketconnectionfailed", success: false, errorMsg: "Failed to establish socket connection or socket connection closed unexpectedly. The reverse proxy might be misconfigured, please contact the administrator. For more info on proxy configuration please checkout https://sdk.collaboraonline.com/docs/installation/Proxy_settings.html", result: "" }
```

Log hint is the only valuable information right now:
```
nextcloud-collabora-1  | wsd-00001-00148 2023-11-23 09:38:00.606399 +0000 [ docbroker_00c ] ERR  No acceptable WOPI hosts found matching the target host [nextcloud] in config| wsd/Storage.cpp:290
nextcloud-collabora-1  | wsd-00001-00148 2023-11-23 09:38:00.606603 +0000 [ docbroker_00c ] ERR  loading document exception: No acceptable WOPI hosts found matching the target host [nextcloud] in config| wsd/DocumentBroker.cpp:2631
[...]

PR #7842, opened 2023-12-15
improve error message on invalid WOPI host

Rash419:
- before this patch when socket connection failed, we used to show the same error for socket connection failure and send the 'Action_Load_Resp' with same errorMsg
- now we show/send different message for unauthorized and loading failure

Resolves: #7709
Change-Id: I71b1b7f8e2eca93790d068583370787825911a41

* Target version: master

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1542,17 +1542,36 @@ app.definitions.Socket = L.Class.extend({
 		// Let onclose (_onSocketClose) report errors.
 	},
 
-	_onSocketClose: function () {
+	_onSocketClose: function (event) {
 		window.app.console.debug('_onSocketClose:');
 		if (!this._map._docLoadedOnce && this.ReconnectCount === 0) {
+			var errorMsg, errorType = '';
+			var reason = event.reason;
+			if (reason && reason.startsWith('error:')) {
+				var command = this.parseServerCmd(reason);
+				if (command.errorCmd === 'internal' && command.errorKind === 'unauthorized') {
+					errorType = 'websocketunauthorized';
+					errorMsg = errorMessages.unauthorized;
+				} else if (command.errorCmd === 'storage' && command.errorKind === 'loadfailed') {
+					errorType = 'websocketloadfailed';
+					errorMsg = errorMessages.storage.loadfailed;
+				} else {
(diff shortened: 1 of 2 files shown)


browser/src/core/Socket.js
```diff
@@ -1533,17 +1533,33 @@ app.definitions.Socket = L.Class.extend({
...
+				errorType: errorType, 
 				success: false,
-				errorMsg: errorMessages.websocketconnectionfailed,
+				errorMsg: errorMsg,
 				result: '',
 			};
-			this._map.fire('postMessage', {msgId: 'Action_Load_Resp', args: postMessageObj});
-			this._map.fire('error', {msg: errorMessages.websocketconnectionfailed, cmd: 'socket', kind: 'closed', id: 4});
+			this._map.fire('postMessage', { msgId: 'Action_Load_Resp', args: postMessageObj });
+			return;
```
eszkadev, line 1562: it that return intentional, we don't want to execute code below?

Rash419: Yes,  we execute this code only when `this._map._docLoadedOnce` is false, if there is document loaded there is nothing to cleanup.

browser/src/core/Socket.js
```diff
@@ -1533,17 +1533,33 @@ app.definitions.Socket = L.Class.extend({
...
+			var reason = event.reason;
+			if (reason && reason.startsWith('error:')) {
+				var command = this.parseServerCmd(reason);
+				if (command.errorCmd === 'internal' && command.errorKind === 'unauthorized') {
+					errorType = 'websocketunauthorized';
+					errorMsg = errorMessages.unauthorized;
+				} else if (command.errorCmd === 'storage' && command.errorKind === 'loadfailed') {
+					errorType = 'websocketloadfailed';
+					errorMsg = errorMessages.storage.loadfailed;
+				}
```
eszkadev, line 1549: else ?
we need something generic here I expect. Because it will pass nothing to the post message

Rash419: but I can't find anything other error being sent from the server, but we should handle here

eszkadev: yes but it might be changed in the future

eszkadev (approved):
Looks good

Issue #3491, opened 2021-10-26
Hide text boundaries by default


juliushaertl:
**Describe the bug**
When discussing the general UI with @jancborchardt and @pedropintosilva we noticed that the the text boundaries add some distraction to the document view and it would be good to hide them by default if possible.

![image](https://user-images.githubusercontent.com/3404133/138888477-ff55e5a8-1aa1-4ae7-ac8b-e34e554fead1.png)

Also it is already possible to get a boundary view with the "View > Formatting Marks" option:

![image](https://user-images.githubusercontent.com/3404133/138888632-8ae8f715-9b46-41d1-9412-6fd8aa2998c1.png)

core seems to have a dedicated option for this as "Text Boundaries", so it would make a cleaner document view when hiding those by default.

timar:
This fix does not work with CODE/COOL 23.05.
```
warn:configmgr:52415:52415:configmgr/source/xcuparser.cxx:337: unknown item "/org.openoffice.Office.UI/ColorScheme/ColorSchemes/org.openoffice.Office.UI:ColorScheme['Collabora Office']/DocBoundaries" in "file:///home/timar/MOVE/online/coolkitconfig.xcu"
warn:configmgr:52415:52415:configmgr/source/xcuparser.cxx:337: unknown item "/org.openoffice.Office.UI/ColorScheme/ColorSchemes/org.openoffice.Office.UI:ColorScheme['Collabora OfficeDev']/DocBoundaries" in "file:///home/timar/MOVE/online/coolkitconfig.xcu"
warn:configmgr:52415:52415:configmgr/source/xcuparser.cxx:337: unknown item "/org.openoffice.Office.UI/ColorScheme/ColorSchemes/org.openoffice.Office.UI:ColorScheme['LibreOffice']/DocBoundaries" in "file:///home/timar/MOVE/online/coolkitconfig.xcu"warn:configmgr:52415:52415:configmgr/source/xcuparser.cxx:337: unknown item "/org.openoffice.Office.UI/ColorScheme/ColorSchemes/org.openoffice.Office.UI:ColorScheme['LibreOfficeDev']/DocBoundaries" in "file:///home/timar/MOVE/online/coolkitconfig.xcu"
```
all items are unknown, the correct item seems to be:
```
[...]

PR #7845, opened 2023-12-17
Private/timar/coolkitconfig

timar:
* Resolves: #3491 
* Target version: master


vmiklos:
It looks like this means you can't do `coolconfig.xcu` edits locally without those changes showing up in `git diff` output. Is this a problem, or people never customize this in practice and we can just 'not care'?

timar:
> It looks like this means you can't do `coolconfig.xcu` edits locally without those changes showing up in `git diff` output. Is this a problem, or people never customize this in practice and we can just 'not care'?

I did not think about this use case. Do you use this file that way?

vmiklos:
> I did not think about this use case. Do you use this file that way?

No, I was just considering this a possibility. If we say that developers should not use custom coolwsd.xml settings for a longer period of time, then this is fine.

caolanm (approved):
looks reasonable to me

Issue #7813, opened 2023-12-11
Calc: sheet drop zone is too small


pedropintosilva:
### Describe the Bug

It's very difficult to drag and drop a calc sheet to be the first sheet. It seems the drop zone is incredibly tiny - and we should do something about that. Perhaps we should accept over the right-most movement icons.

https://github.com/CollaboraOnline/online/assets/65948705/50c57fac-d397-4b2b-8ad7-27f9477cbeb1

PR #7846, opened 2023-12-17
Calc: improve drag&drop functionality of sheet tabs

bayramcicek:
- drop zone indicator added to the left of all tabs
- additional drop zone indicator added to the right of the last tab
- added drop zone container at the end of tabs for the last drop zone indicator

- div#drop-zone-end-container element also adds a space to the end of the tabs so that the last tab can be seen easily after scrolling to the end.

Change-Id: I25a87bc1798b94c5f294532d308cd46292ef1795

* Resolves: #7813
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -180,14 +180,14 @@ L.Control.Tabs = L.Control.extend({
 					var menuData = L.Control.JSDialogBuilder.getMenuStructureForMobileWizard(menuItemMobile, true, '');
 				}
 
-				var frame = L.DomUtil.create('div', '', ssTabScroll);
-				frame.setAttribute('draggable', false);
-				frame.setAttribute('id', 'first-tab-drop-site');
-				this._addDnDHandlers(frame);
-
 				for (var i = 0; i < parts; i++) {
 					if (e.hiddenParts.indexOf(i) !== -1)
 						continue;
+
+					// create a drop zone indicator for the sheet tab
+					// put the indicator on the left of the tab
+					var dropZoneIndicator = L.DomUtil.create('div', 'tab-drop-area', ssTabScroll);
+					dropZoneIndicator.id = 'drop-zone-' + i;
 					var id = 'spreadsheet-tab' + i;
 					var tab = L.DomUtil.create('button', 'spreadsheet-tab', ssTabScroll);
 					if (window.mode.isMobile() || window.mode.isTablet()) {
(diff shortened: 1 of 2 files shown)


bayramcicek:
Implementation behavior on Light mode and Dark mode:

https://github.com/CollaboraOnline/online/assets/24255179/a2abe4f5-9ed4-407b-9d3f-b1e13136b247

browser/src/control/Control.Tabs.js
```diff
@@ -354,26 +366,35 @@ L.Control.Tabs = L.Control.extend({
 		// support duplication with ctrl in the future.
 		e.dataTransfer.dropEffect = 'move';
 
-		this.classList.add('tab-dropsite');
+		var currentTabId = e.target.id;
+
+		// check if the dragged tab is over the dropZoneEndContainer
+		if (currentTabId == 'drop-zone-end-container')
```
eszkadev, line 372: I think we can avoid this block of code below.
Drop zone is always before tab button.
We can use: `e.target.previousElementSibling.addClass(...)`,
no need for regex to get number, and searching in the DOM for a correct node :)

browser/src/control/Control.Tabs.js
```diff
@@ -354,26 +366,35 @@ L.Control.Tabs = L.Control.extend({
...
+			var dropZoneId = 'drop-zone-' + currentIndex;
+			$('#' + dropZoneId).addClass('tab-drop-area-active');
+		}
+
 		return false;
 	},
 
 	_handleDragLeave: function() {
-		this.classList.remove('tab-dropsite');
+		$('.tab-drop-area').removeClass('tab-drop-area-active');
```
eszkadev, line 384: please use `e.target.previousElementSibling` here too

browser/src/control/Control.Tabs.js
```diff
@@ -354,26 +366,35 @@ L.Control.Tabs = L.Control.extend({
...
-		var targetIndex = this._map._docLayer._partNames.indexOf(e.target.innerText);
 
-		this._moveSheet(targetIndex+2);
+		var targetIndex = this._map._docLayer._partNames.indexOf(e.target.innerText);
+		this._moveSheet(targetIndex + 1); // drop to left side of the tab
 	},
 
 	_handleDragEnd: function() {
-		this.classList.remove('tab-dropsite');
+		$('.tab-drop-area').removeClass('tab-drop-area-active');
```
eszkadev, line 397: please use `e.target.previousElementSibling` here too

eszkadev (commented):
Looks good and works. I just recommend a small simplification of a code.

eszkadev (approved):

PR #7848, opened 2023-12-18
Changes for handling readonly mode click events:

gokaysatir:
* These changes are made primarily for reading the hyperlink data from the document in readonly mode.
* Added readonlyclick event.
* Added readonlyhyperlinkclicked event.
* And the handlers for these.

When user clicks on document in readonly mode, a readonlyclick event is sent to the server. Childsession checks if there is a hyperlink at the clicked location. If so, the information of the hyperlink is sent to the client.

Change-Id: I9c104a1ec770491725337b9bbb1e2d1c2851321b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1925,6 +1925,11 @@ L.CanvasTileLayer = L.Layer.extend({
 			var json = JSON.parse(textMsg.substring('colorpalettes:'.length + 1));
 			app.colorPalettes['ThemeColors'].colors = json.ThemeColors;
 		}
+		else if (textMsg.startsWith('readonlyhyperlinkclicked: ')) {
+			var json = JSON.parse(textMsg.replace('readonlyhyperlinkclicked: ', ''));
+			var position = this._twipsToLatLng(json.position);
+			this._showURLPopUp(position, json.link);
+		}
 	},
 
 	_onTabStopListUpdate: function (textMsg) {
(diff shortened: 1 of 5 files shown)


gokaysatir:
This requires below PR:
* https://gerrit.libreoffice.org/c/core/+/160022/4

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2595,28 +2600,40 @@ L.CanvasTileLayer = L.Layer.extend({
 		link.innerText = url;
 		var copyBtn = L.DomUtil.createWithId('div', 'hyperlink-pop-up-copy', parent);
 		L.DomUtil.addClass(copyBtn, 'hyperlink-popup-btn');
-		copyBtn.setAttribute('title', _('Copy link location'));
+
+		if (app.file.permission !== 'readonly')
+			copyBtn.setAttribute('title', _('Copy link location'));
+		else
+			copyBtn.setAttribute('title', _('Select link text.'));
```
Ashod, line 2607: Is the '.' at the end necessary/intentional?

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,28 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
     return true;
 }
 
+bool ChildSession::readOnlyClickEvent(const StringVector& tokens)
+{
+    getLOKitDocument()->setView(_viewId);
+    int x = std::atoi(tokens[1].substr(2).data());
```
Ashod, line 1650: We have conversion helpers in Util. Best to reuse them. They handle invalid cases and support default value.

Also, we have utilities to parse 'key=value' cases. `substr(2)` is fragile. See other cases where we parse similar commands.

Finally, these can be `const`.

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,28 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
...
+    int y = std::atoi(tokens[2].substr(2).data());
+
+    char *test = getLOKitDocument()->hyperlinkInfoAtPosition(x, y);
+    std::string result(test);
```
Ashod, line 1654: Must test for `nullptr`. Otherwise, this will throw if `test` is nullptr.

Instead:
```
if (test == nullptr)
{
    return true;
}
```

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,28 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
     return true;
 }
 
+bool ChildSession::readOnlyClickEvent(const StringVector& tokens)
+{
+    getLOKitDocument()->setView(_viewId);
+    int x = std::atoi(tokens[1].substr(2).data());
+    int y = std::atoi(tokens[2].substr(2).data());
+
+    char *test = getLOKitDocument()->hyperlinkInfoAtPosition(x, y);
```
Ashod, line 1653: Please run clang-format on new code. vscode can be configured to run automatically on saving on modified parts only (as opposed to the full file).

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,28 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
...
+    std::string result(test);
+    std::free(test);
+
+    if (!result.empty())
+    {
+        std::string message = "readonlyhyperlinkclicked: { \"link\": \"%%LINK_HERE%%\", \"position\": { \"x\": %%X_HERE%%, \"y\": %%Y_HERE%% } }";
```
Ashod, line 1659: Much simpler, safer, faster, better, to simply use `std::ostringstream` and stream/concatenate the values directly into it.

Creating a template and then doing search/replace 3 times, with `const_cast` is not right.

Ashod (commented):
Thanks @gokaysatir. Left some comments and suggestions.

gokaysatir:
Thanks for the comments @Ashod :)

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,35 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
...
+    const int x = Util::safe_atoi(tokens[1].data(), tokens[1].size());
+    const int y = Util::safe_atoi(tokens[2].data(), tokens[2].size());
+    char* test = getLOKitDocument()->hyperlinkInfoAtPosition(x, y);
+
+    if (test != nullptr)
+    {
+        std::string result(test);
+        std::free(test);
+
+        if (!result.empty() && x != 0 && y != 0)
```
Ashod, line 1663: Nit: The `x != 0 && y != 0` could be done at line 1658, with the `test != nullptr`, instead of here.

kit/ChildSession.cpp
```diff
@@ -1639,6 +1644,35 @@ bool ChildSession::mouseEvent(const StringVector& tokens,
...
+        {
+            std::stringstream message;
```
Ashod, line 1665: `std::ostringstream` is better, as it only allows for writing. This protects against accidental reading, instead of writing.

Ashod (approved):
Thanks @gokaysatir. Looks much better. Left a couple of minor suggestions, but no blockers. Let's see if the CI is happy.

PR #7850, opened 2023-12-18
Add a lint workflow to fail on eslint warnings

Minion3665:
As discussed [on Matrix](https://col.la/suppresseslintindevelopment) the set of things we want to pick up in CI and development are different. Specifically, there are some things that only get in the way during development but should not be allowed in merged code. Examples are 'debugger' statemens, dead code (e.g.  resulting from 'if (false)') and unused variables.

After #7822 there are no more eslint warnings, so we can use it as a separate state for these issues that should only block in CI.

This commit adds a new workflow which will run eslint and fail if it receives any warnings, and changes those 3 errors so that they only emit warnings.

It's expected that there are more annoying warnings which I've missed, if you have any please make a followup change!

This commit explicitly does not deal with formatting-related issues (e.g. requiring single quotes for strings in eslint) as though they can be annoying there's no reason why they should change between development and master... another followup to improve linting experience could be to use an autoformatter so the computer fixes the formatting for you, however that's out-of-scope for this commit

Change-Id: I036afac5ef5056a9cc2effc21e31165aa1436ad2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -630,6 +630,14 @@ compile_commands: $(abs_srcdir)/compile_commands.json
 browser/node_modules: browser/package.json browser/archived-packages
 	@cd browser && npm install
 
+eslint: browser/node_modules
+	browser/node_modules/.bin/eslint browser/src browser/js browser/admin/src \
+		--max-warnings 0 \
+		--resolve-plugins-relative-to browser \
+		--ignore-path browser/.eslintignore \
+		--no-eslintrc \
+		--config browser/.eslintrc
+
 install-exec-hook:
 	cd $(DESTDIR)$(bindir) && \
 	$(LN_S) coolconfig loolconfig && \
@@ -671,7 +679,7 @@ check-for-system-nss:
 		fi \
 	fi
 
-check: check-for-system-nss check-recursive
+check: check-for-system-nss check-recursive eslint
 	$(GEN_COVERAGE_COMMAND)
 
 coverage-report:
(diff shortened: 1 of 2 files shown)


vmiklos (commented):
Looks reasonable to me (we have a similar setup for core: clang plugin warnings are enforced at CI time but you are not required to respect them during local development). But I prefer if @eszkadev have the final word on this, I don't do enough work on the JS side to decide.

Minion3665:
> I'm wondering if we could maybe put more strict eslint check into `make check` which is run on CI. This has advantage that doesn't waste CI time

This makes sense, I wonder if we should run it in both CI and a GitHub action ... that way if we're waiting in a queue for CI we'll still get input eslint failures quickly

eszkadev:
We can always do eslint before c++ unit-tests in make check

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,39 @@
...
+        pip install polib
+    - name: Configure
+      run: |
+        cd .. && wget https://github.com/CollaboraOnline/online/releases/download/for-code-assets/core-co-23.05-assets.tar.gz
+        tar xvf core-co-23.05-assets.tar.gz && rm core-co-23.05-assets.tar.gz && export LOCOREPATH=$(pwd)
+        cd - && ./autogen.sh
+        ./configure --enable-silent-rules --with-lokit-path=${LOCOREPATH}/include --with-lo-path=${LOCOREPATH}/instdir --enable-debug --disable-setcap
+    - name: Run ESLint
+      run: |
+        make eslint
```
eszkadev, line 39: do we need this as we will do it anyway on Jenkins?

eszkadev (approved):

PR #7854, opened 2023-12-19
Filter by color options should have radio buttons insted of Checkboxes

Darshan-upadhyay1110:
- Added radio buttons in  `filter by color` options which had Checkboxes
 - it should have radio buttons same as we have in LO core
 - it also makes sense because we only need to apply one filter at a time in cases like `filterByColor`

Steps to check patch

    - Open .ods file
    - add some dummy data in different rows, but column should be the same
    - click on the top cell of dummy data
    - got to `DATA` tab and apply `AutoFilter`
    - open auto filter popup and select filer by color
    - compare previous version with the current patch
    - check same thing in LO core

Change-Id: I00b4d84e5117b3c1f05fbc37c61331d2c3bc2a31

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -87,6 +87,17 @@ function _createCheckbox(parentContainer, treeViewData, builder, entry) {
 	return checkbox;
 }
 
+function _createRadioButton(parentContainer, treeViewData, builder, entry) {
+	var radioButton = L.DomUtil.create('input', builder.options.cssClass + ' ui-treeview-checkbox', parentContainer);
+	radioButton.type = 'radio';
+	radioButton.tabIndex = -1;
+
+	if (entry.state === 'true' || entry.state === true)
+		radioButton.checked = true;
+
+	return radioButton;
+}
+
 function _changeCheckboxStateOnClick(checkbox, treeViewData, builder, entry) {
 	if (checkbox.checked) {
 		var foundEntry = _findEntryWithRow(treeViewData.entries, entry.row);
@@ -184,10 +195,10 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 	var expander = L.DomUtil.create('div', builder.options.cssClass + ' ui-treeview-expander ', span);
 
 	if (entry.state !== undefined)
-		var checkbox = _createCheckbox(span, treeViewData, builder, entry);
+		var selectionElement = _createSelectionElement(span, treeViewData, entry,builder);
 
 	if (entry.selected && (entry.selected === 'true' || entry.selected === true))
-		_selectEntry(span, checkbox);
+		_selectEntry(span, selectionElement);
 
 	var text = L.DomUtil.create('span', builder.options.cssClass + ' ui-treeview-cell', span);
 	for (var i in entry.columns) {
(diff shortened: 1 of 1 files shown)


browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -362,7 +379,7 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 			td.innerText = entry.columns[i].text;
 
 		if (!disabled)
-			$(td).click(clickFunction);
+			$(parentContainer).click(clickFunction);
```
eszkadev, line 382: why is that needed?

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -331,14 +342,20 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 
 	if (entry.state !== undefined) {
 		var td = L.DomUtil.create('td', '', parentContainer);
-		var checkbox = _createCheckbox(td, treeViewData, builder, entry);
+		var toggleElement;
+		if (singleClick) {
```
eszkadev, line 346: here will be `data.checkboxtype === 'radio'` or something similar

browser/src/control/jsdialog/Widget.TreeView.js
Darshan-upadhyay1110: td only covers the label part of the option. it does not fire the `clickfunction` if we do check/uncheck the radio button.

`parentContainer` will cover that case where if user click on label or on checkbox, it will fire the `clickFunction`

Darshan-upadhyay1110:
CORE changes :  https://gerrit.libreoffice.org/c/core/+/161043

Before:
image

After:
image

Darshan-upadhyay1110:
image

Tested cypress on local, passing all the cases

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -394,6 +405,19 @@ function _hasIcon(columns) {
...
+function _createSelectionElement (parentContainer, treeViewData, entry, builder) {
+	var selectionElement;
+	switch (treeViewData.checkboxtype) {
+		case 'radio':
+			selectionElement = _createRadioButton(parentContainer, treeViewData, builder, entry);
+			break;
+		case 'checkbox':
+			selectionElement = _createCheckbox(parentContainer, treeViewData, builder, entry);
+			break;
+	}
```
eszkadev, line 417: Core could send empty checkboxtype.
can we do this like: if (checkboxtype === radio) createRadio else createCheckbox ?

Darshan-upadhyay1110: i am afraid we can't do if else because some cases does not have neither checkbox nor radio.

eszkadev: but this is done only when `entry.state` exists which means we need one of those

Darshan-upadhyay1110: @eszkadev yes you are right for those cases where no radio or checkbox the entry.state will be undefined and it will never call the _createSelectionElement

Darshan-upadhyay1110: Okay got it i am updating switch with if else

eszkadev (approved):

pedropintosilva (approved):
Thank you

PR #7855, opened 2023-12-19
Calc tabs: remove drop zone indicator after drop

bayramcicek:
- we need to remove 'tab-drop-area-active' class after a sheet tab is dropped, otherwise drop
zone indicator will still be shown after dropping

- this fixes the bug from f8433d323d2f49852501961c7949300b4a59a512 Calc: improve drag&drop functionality of sheet tabs #7846

Change-Id: Iff7ea455452c431907cc7534cbd111e7b4f037d8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -380,6 +380,7 @@ L.Control.Tabs = L.Control.extend({
 			e.stopPropagation();
 		}
 
+		e.target.previousElementSibling.classList.remove('tab-drop-area-active');
 		var targetIndex = this._map._docLayer._partNames.indexOf(e.target.innerText);
 		this._moveSheet(targetIndex + 1); // drop to left side of the tab
 	},


eszkadev:
How to reproduce? I don't see any issue. After drop it disappears for me.

bayramcicek:
> How to reproduce? I don't see any issue. After drop it disappears for me.

@eszkadev : -sometimes- It does not disappears after dropping on the same tab:

https://github.com/CollaboraOnline/online/assets/24255179/3e22fda7-94d0-435f-8262-0648839e9da4

IMHO, it will be also good to add `e.target.previousElementSibling.classList.remove('tab-drop-area-active');` inside `_handleDrop` function - to make sure that drop zone won't visible. With this patch, everything works fine for me.

eszkadev (approved):

PR #7857, opened 2023-12-19
Add prettier

Minion3665:
Prettier is a code formatter for javascript/typescript. It's relatively
common to see complaints about style in pull requests or inconsistent
style getting in to Collabora Online. Also, without a formatter it's
harder for editors to automatically format code in sensible and
consistent ways

Prettier doesn't have many options, but those which it does have I have
set to the closest values to current behavior (e.g. single quotes) in
order to minimize unnecessary changes. There'll still be a lot of
changes, but in my opinion this is preferable to no formatter as we only
pay the cost once but get benefits continuing on into the future

I have decided to format everything in a single commit, as it will make it easier to ignore in blames in the future and less annoying in the future. It's annoying now, and I'm sorry for that. When reviewing, you'll probably want to look at individual commits - I don't imagine the full diff view will be particularly helpful.


eszkadev (dismissed):
we can't do that big code refactor which will cause lots of conflicts when cherry-picking,

can we do this for new files only currently?

mmeeks:
Great initiative - but of course the blames are also really useful for assigning blame ;-) Luckily we will be moving more code to typescript incrementally, so would love to see a consistent style at that stage as well as for new files =)

Minion3665:
Thanks for the suggestions! I've added an ignore file containing everything that would be changed, I've opted to leave out these 3 files which already match the new formatting
```
$ make prettier-write
browser/node_modules/.bin/prettier \
	--config browser/.prettierrc \
	--ignore-path browser/.prettierignore \
	--ignore-path browser/.beforeprettier \
	--write \
	browser/src browser/js browser/admin/src
browser/src/copyright.js 68ms (unchanged)
browser/src/Leaflet.js 14ms (unchanged)
browser/admin/src/Admin.js 8ms (unchanged)
```

Minion3665:
this still has a big problem: node versions... our CI is currently using an old version of node that isn't supported by prettier. I'm talking to people about how we can upgrade that, but one option could be to use a github workflow for this. It's not preferable because it'll waste CI time (see also: #7850) but it's something we could do more immediately... then again, I'm not sure how hard node will be to update

timar:
I added Node v20 to the Tinderbox container, and I think I'll clone that for all CI for co-24.04 onwards.

Minion3665:
> I added Node v20 to the Tinderbox container, and I think I'll clone that for all CI for co-24.04 onwards.

the build is still failing for the same issue - I'm not sure if it's added and only not being used? It may be worth checking if 24.04 CI shares the same issue and fixing it there if it does

mmeeks:
Seems reasonable to me =) But I'll leave to a JS person. Thanks Skyler.

eszkadev (commented):
Nice, but maybe let's wait for CI test (unit + cypress step) to be sure it will not fail all the builds

Minion3665:
> Nice, but maybe let's wait for CI test (unit + cypress step) to be sure it will not fail all the builds

@eszkadev that makes sense, I've pulled the .editorconfig stuff into its own change so we can merge it first, as it doesn't have an effect on the build: https://github.com/CollaboraOnline/online/pull/8455

Minion3665:
CI is back up to having checks, and I've modified the commit to make sure it doesn't format any files before our CI runs prettier. In other words: it can now fail, so that's good. I think we should merge this soon.

I note also that when we do merge it we need to be sure that we either rerun CI for or manually format currently open and green PRs that add new files (or CI will start complaining about them after they get merged if they have formatting issues)

eszkadev (approved):

PR #7860, opened 2023-12-19
bind noManualDrag in Path.Drag.js

Minion3665:
Previously, noManualDrag was not bound to any object. This worked fine in SVGGroup.js because the 'this' that was inferred was the correct 'this', however in Path.Drag.js the 'this' inferred was incorrect. This code path was occasionally called when dragging which caused an intermittent error while dragging objects in draw

This is a followup to Iab0d3bcca588eaed14469597868a9c4e2dcf8488, which I believe introduced this regression

Change-Id: If8657133a1b6c000b1096fb2c822d02d728de80c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/vector/Path.Drag.js
+++ b/browser/src/layer/vector/Path.Drag.js
@@ -62,13 +62,13 @@ L.Handler.PathDrag = L.Handler.extend(/** @lends  L.Path.Drag.prototype */ {
 
 	},
 
-	noManualDrag: function(f) {
+	noManualDrag: window.memo.decorator(function(f) {
 		if ('noManualDrag' in this._path) {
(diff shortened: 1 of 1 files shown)


hfiguiere:
LGTM, it fix the issue I reported.

hfiguiere (approved):

hfiguiere (approved):
LGTM

hfiguiere:
New iteration tested as well. sadly cypress failure.

Minion3665:
> New iteration tested as well. sadly cypress failure.

indeed, I think the previous one passed because it subtly broke the event again (by binding to the wrong thing so a condition was never true)

PR #7861, opened 2023-12-19
Allow localhost name in lok_allow

eszkadev:
This can be useful for debug purpose. This allows to use integrator with `localhost` domain and still beeing able to paste/insert some remote content.


caolanm (approved):
if we already have 127.0.0.7 then no hardship in localhost

PR #7864, opened 2023-12-20
Fix view pushed up in Calc for autfilter popup

pedropintosilva:
- fix view jump by display hidden for auto-filter popup
Change-Id: I4e48fd2d7363db53d4bf3a1d85f9897de3b1e91a

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -12,6 +12,7 @@
 	max-height: 100%;
 	z-index: 2000;
 	display: flex;
+	overflow: hidden;
 }
 
 .jsdialog-overlay.cancellable {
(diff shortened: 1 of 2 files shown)


pedropintosilva:
Do not merge yet. Testing...

pedropintosilva:
I'm not able to reproduce the original bug that #7832 fixes on master. So probably there is some regression that entered on master and it's not present on this branch (23.05). So we should merge this

pedropintosilva:
Yes this fixes the same problem encountered on master (that was fixed by Pr #7832 )

How to reproduce the bug (before cherry-picking) : 
1. Insert content in 6 cells
2. Make sure the scroll position is in such a manner that we have one empty cell between our 6 stacked cells and the bottom:

image

------ 


| Before | After |
| --------- | ------- |
| Screenshot_20231220_110309 | Screenshot_20231220_110401 |

PR #7866, opened 2023-12-20
[Backport to 23.05] Filter by color options should have radio buttons insted of Checkboxes

pedropintosilva:
- Fixed filter by color options which had Checkboxes
 - it should have radio buttons same as we have in LO core
 - it also makes sense because we only need to apply on filter by setting at a time

Steps to check patch

    - Open .ods file
    - add some dummy data in diffent rows but column should be the same
    - click on the top cell of dummy data
    - got to `DATA` tab and apply `AutoFilter`
    - open auto filter popup and select filer by color
    - compare previous version with the current patch
    - check same thing in LO core

Change-Id: I00b4d84e5117b3c1f05fbc37c61331d2c3bc2a31

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -87,6 +87,17 @@ function _createCheckbox(parentContainer, treeViewData, builder, entry) {
 	return checkbox;
 }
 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
from https://github.com/CollaboraOnline/online/pull/7854

pedropintosilva:
Tested :+1:

PR #7870, opened 2023-12-20
Remove protocol when adding to lok_allow

eszkadev:
We automatically add allowed wopi hosts to lok_allow. But we expect to have plain host names, we need to remove protocol if it is present.

Example:
wopi - group - host: `https://.*\.mydomain.com`
we need to add: `.*\.mydomain.com` to lok_allow list.


eszkadev:
reviewed on master

PR #7872, opened 2023-12-20
missing tiles in multi-pane spreadsheets

caolanm:
here std::ceil is operating on the result of an int divided by an int so the input is already an effectively floor-ed int result of the division so the ceil doesn't do anything

so we end up filtering out some of the invalidates for tiles in frozen panes.

Change-Id: Ibfb25fdfdfb84735ca9410f8250d2e5b9fa1070e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Would love a unit test in a follow-up commit =)
Thanks.

PR #7873, opened 2023-12-20
browser: adjust initial annotation height

hcvcastro:
It is a bug from Firefox, according to records
https://bugzilla.mozilla.org/show_bug.cgi?id=292284

The vertical scroll bar requires minimum height to render.

Change-Id: I93b06c625bc98c307dde0c1d6331d457ff99baa5


caolanm (approved):
Why 54px specifically? just increase until it works, or is there some known calculation to get to that size?

hcvcastro:
>> Why 54px specifically? just increase until it works
yes, experimental data until vertical bar is shown.

PR #7875, opened 2023-12-20
calc: don't make app inactive on drag and drop

eszkadev:
When Pivot table dialog was used and we drag and drop elements from list to other list in Firefox - then we sent `userinactive` message because we lost focus of a window. Don't do that.


gokaysatir (commented):
I am getting below error on Firefox:

* Uncaught TypeError: document.querySelectorAll(...).length is not a function

gokaysatir (approved):

PR #7879, opened 2023-12-21
Add more cypress test in Autofilter (Calc)

Darshan-upadhyay1110:
Below cases are covered in current patch
- `Filter by Color` in autofilter should have radio buttons.
- Changes should be applied when we select any option from `filter by color` submenu.

Change-Id: I1a757e09be81df934b46d48f71e8be28fc48c984

### DIFF
--- a/cypress_test/integration_tests/desktop/calc/autofilter_spec.js
+++ b/cypress_test/integration_tests/desktop/calc/autofilter_spec.js
@@ -111,4 +111,37 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'AutoFilter', function() {
 
 		assertDataOnFilter(['Cypress Test', 'Status', 'Test 1', 'Pass', 'Test 2', 'Fail', 'Test 3', 'Pass', 'Test 5', 'Fail']);
 	});
+
+	// check if filter by color applied or not
+	it('Filter by color', function() {
+		// apply background color to some cells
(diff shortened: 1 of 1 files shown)


cypress_test/integration_tests/desktop/calc/autofilter_spec.js
```diff
@@ -111,4 +111,59 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'AutoFilter', function() {
...
+
+		//wait after apply filter by color filter
+		cy.wait(500);
+
+		// Find the table element with ID "background"
+		cy.cGet('table#background')
+		.find('input') // Find all input elements inside the table
+		.each(($input) => { // Iterate through each input element
+		  // Assert that each input is of type radio
+		  cy.wrap($input).should('have.attr', 'type', 'radio');
```
eszkadev, line 166: this check should be part of the other test - it will be faster, than doing the same 2 times

Darshan-upadhyay1110:
@eszkadev do you think this will be enough ? or any other case also included here

Darshan-upadhyay1110:
@eszkadev for more readability of code, I have not done the complicated thing like use `each` inside `find` and then conditionally click on first element .

eszkadev (approved):

Darshan-upadhyay1110:
@Rash419 can you please review this test and if it's okay we can add to master

PR #7881, opened 2023-12-21
calc: avoid jumping on large cell selection

eszkadev:
If we had very long or tall cell - maybe bigger than a screen on every selection we tried to show it on the screen what caused view to jump. If cell is bigger and partially visible left view at the same position to avoid "jumping".

Steps to reproduce:
1. merge many cells - longer than one screen width
2. scroll in the middle of merged cell
3. select the cell Result: jump to the begining
Expected: no jump if it is paritally visible


eszkadev:
It should work like core:
[merged-2023-12-19_09.44.43.webm](https://github.com/CollaboraOnline/online/assets/5307369/527e793c-55ca-4271-99ea-65c6da16666b)

gokaysatir (approved):

PR #7892, opened 2023-12-23
annotation: if comment is not initialised its not in editing mode

lpranam:
this caused problem when new doc is loading and we are settings up comments, but they are not initialised and marked as being in editing. i.e: this will cause problem with showing reply count(in collapsed view) initially

Change-Id: Ia688ca0bfc37dd925bd74dc5826d43fdb37c81db

* Target version: master 

### Checklist


gokaysatir (approved):
Nice catch, thanks :)

PR #7894, opened 2023-12-25
annotation: remove redundant condition

lpranam:
it did not make sense to check for selection if we are unselecting before selecting

problem:
when already selected comment try to modify,
it will not rearrange replies in the thread this caused overlapping of the comments and bad user experience

Change-Id: I276c34301b45951a7051018770e21b4970603861

* Target version: master 

### Checklist


gokaysatir (approved):
Thanks @lpranam :)

PR #7895, opened 2023-12-25
annotation: scroll to make selected comment visible

lpranam:
Change-Id: Ifc76f83ef1cca6cf6113f63bcb50312523ffbf52

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -52,7 +52,9 @@ class IdleHandler {
 			}
 		}
 
-		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen()) {
+		// Ideally instead of separate isAnyEdit check here, we could check isAnyEdit inside isAnyDialogOpen,
+		// but unfortunatly that causes problem in _deactivate and unnecessary 'userinactive' message is sent
+		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen() && !cool.Comment.isAnyEdit()) {
 			this.map.focus();
 		}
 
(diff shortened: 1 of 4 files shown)


gokaysatir (dismissed):
```
public getScreenTopBottom(): Array<number> {
	return [app.file.viewedRectangle[1], app.file.viewedRectangle[1] + app.file.viewedRectangle[3]];
}
```

Something like above would be better, it uses already assigned values and makes use of "app.someDocState" structure.

**But** "app.file.viewedRectangle" seems wrong until user scrolls the document. I don't know why and we will need to solve it sooner or later i think.

Anyway, apart from this side note, works beautifully.

If you can have time to investigate the issue i mentioned here, that'd be great, thanks for the improvements :)

gokaysatir:
On Calc, when i start a comment and scroll the view, comment sticks with its first position.

gokaysatir:
On Writer, if i start to modify a comment and then click somewhere, it doesn't show the replies until i click on the comment again. Also i need to scroll the view while modifying the comment if it is at the bottom of the screen.

gokaysatir:
@lpranam thank you for the PR. Writer works well. On Calc, all comments are visible at start. After user moves cursor over a comment and closes it, they re-appear with a scroll. Can you have a look?

I will check apps after we have a solution work that.

gokaysatir (changes requested):
Hello, can you check below scenario when you have time? I don't get this error on master branch.

Add a few comments to Calc.

Open in mobile view.

Click on the comments to navigate to their positions.

Error:

CommentListSection.ts:699 Uncaught TypeError: Cannot read properties of null (reading 'positionCalcComment')
    at CommentSection.select (CommentListSection.ts:699:46)
    at CommentSection.click (CommentListSection.ts:522:8)
    at Comment.onMouseClick (CommentSection.ts:821:45)
    at HTMLDivElement.handler (DomEvent.js:51:14)

gokaysatir (approved):
Thanks @lpranam :)

PR #7897, opened 2023-12-26
Handle LOK_CALLBACK_TOOLTIP in wsd

mikekaganski:
Send "tooltip: <payload>" to client for this callback type.
Client-side handling should be implemented separately.

Change-Id: I2d1a07dd36fa20a4b5ff785da788356e0f41d62f

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -3209,6 +3209,11 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
         sendTextFrame("corelog: " + payload);
         break;
     }
+    case LOK_CALLBACK_TOOLTIP:
+    {
+        sendTextFrame("tooltip: " + payload);
+        break;
+    }
     default:
         LOG_ERR("Unknown callback event (" << lokCallbackTypeToString(type) << "): " << payload);
     }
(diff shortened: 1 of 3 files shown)


mikekaganski:
This depends on https://gerrit.libreoffice.org/c/core/+/161302 (and its backport to 23.05).

kit/ChildSession.cpp
```diff
@@ -3209,6 +3209,11 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
         sendTextFrame("corelog: " + payload);
         break;
     }
+    case LOK_CALLBACK_TOOLTIP:
+    {
+        sendTextFrame("tooltip: " + payload);
```
Ashod, line 3214: Might need to add `"tooltip:"` in `HTTPWSTest::testInactiveClient()`, in case it is generated by a test document.

Ashod (approved):
Thanks @mikekaganski. Looks good. One suggestion.

kit/ChildSession.cpp
mikekaganski: Thank you!
Let me merge this, and then I will try to create a follow-up implementing your suggestion.

PR #7898, opened 2023-12-26
Refactor debug menu

neilguertincollabora:
Change-Id: I6e2520035d54b9de2b4cf8e015741aaae517fee1

### Summary
I'm trying to add additional debug tools for automated input but the current structure was not conducive to adding new tools. The mess clearly just comes from years of adding one tool at a time. I think it's time for a refresh.

Here are some issues this change addresses:
- Each tool getting added and activated in a different way
- Tools relying on other tools
- Required knowledge of leaflet layer internals
- Code in many far apart places
- Unclear or incorrect names
- Unclear ownership or intent of variables

Now, each tool gets added with a call to `_addDebugTool`, and its name, setup and teardown code are declared all in the same place. This will make it much easier to add or modify new tools, and to add new functionality to the debug menu

### Checklist

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -577,7 +577,7 @@ L.Map.include({
 
 	aboutDialogKeyHandler: function(event) {
 		if (event.key === 'd') {
-			this._docLayer.toggleTileDebugMode();
+			this._docLayer.toggleDebugMode();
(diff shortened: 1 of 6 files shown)


mmeeks (approved):
Looks a lot cleaner - thanks Neil ! =)

PR #7901, opened 2023-12-27
Private/gokay/impres cut

gokaysatir:
Core side seems to handle the mouseup event automatically for the right mouse button.
When we send another mouseup event, selected text can get unselected and context menu can't find a selected text.

Issue is:
* In Impress, select a text in a textbox and right click.
* Context menu cut operations don't work.

Cause:
* We send the right mouseup twice, the second event unselects the text and sets the cursor position to somewhere else.

Signed-off-by: Gökay Şatır <gokaysatir@gmail.com>
Change-Id: Idae749da42443ccdd7ef2fc795b042536fae0db6

### DIFF
--- a/browser/src/docstate.js
+++ b/browser/src/docstate.js
@@ -28,6 +28,18 @@ window.app = { // Shouldn't have any functions defined.
 		readOnly: true,
 		permission: 'readonly',
 		disableSidebar: false,
+		cursor: {
+			visible: false,
+
+			/*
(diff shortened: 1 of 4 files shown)


Ashod (approved):
Thanks @gokaysatir. We should have some Cypress tests for this.

gokaysatir:
Thanks for checking. I will prepare a test for this soon hopefully.

PR #7904, opened 2023-12-27
More work on the debug menu

neilguertincollabora:
### Summary

- Split the tools into categories (Display, Logging, Input)
- Made a tool for the tile overlay grid and turned it off by default
- Commented out Sidebar Rerendering and Always Active which as far as I could tell from a code search don't have any affect anymore

image

### Checklist


mmeeks (commented):
Changes look fine the last two 'Debug fixes' should be either squashed, or better squashed into the commits that created the problems (if they are in this set of changes). Worth using:

git rebase -i HEAD~4

and playing with that a bit. The git history is forever - and its best to have fewer, cleaner commits if we can, rather than having a trail of fixup commits =)

Otherwise changes look fine to me; thanks !

Thanks !

neilguertincollabora:
Squashed, rebased, and pushed, just waiting for the checks now

mmeeks (approved):
Thanks Neil =)

PR #7905, opened 2023-12-29
Memory Accounting

Ashod:
This applies the cgroup memory limit, if set,
such that if it is lower than the configured
memproportion percentage, we do not exceed it.
Otherwise, we risk running out of our cgroup
limit and by then it is too late to do anything
but die due to OOM.

This also moves the logging of the cgroup memory
stats from COOLWSD into Admin, to avoid duplicate
logging.

Also updated the description of memproportion
config entry to account for the cgroup logic.

- wsd: admin: minor clean-up of memory-tracking members
- wsd: admin: use the cgroup limit as available memory

### DIFF
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -496,26 +496,26 @@ bool AdminSocketHandler::handleInitialRequest(
 }
 
 /// An admin command processor.
-Admin::Admin() :
-    SocketPoll("admin"),
-    _forKitPid(-1),
-    _lastTotalMemory(0),
-    _lastJiffies(0),
-    _lastSentCount(0),
-    _lastRecvCount(0),
-    _cpuStatsTaskIntervalMs(DefStatsIntervalMs),
-    _memStatsTaskIntervalMs(DefStatsIntervalMs * 2),
-    _netStatsTaskIntervalMs(DefStatsIntervalMs * 2),
-    _cleanupIntervalMs(DefStatsIntervalMs * 10)
+Admin::Admin()
+    : SocketPoll("admin")
+    , _totalSysMemKb(Util::getTotalSystemMemoryKb())
+    , _totalAvailMemKb(_totalSysMemKb)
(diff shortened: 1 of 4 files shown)


wsd/COOLWSD.cpp
```diff
@@ -5962,29 +5962,6 @@ int COOLWSD::innerMain()
     Util::getVersionInfo(version, hash);
     LOG_INF("Coolwsd version details: " << version << " - " << hash << " - id " << Util::getProcessIdentifier() << " - on " << Util::getLinuxVersion());
 
-    std::size_t availableMemoryMb = Util::getTotalSystemMemoryKb()/1024;
```
Ashod, line 5965: `Admin` is initialized before we get here (it's static/singleton). So this was generating duplicate logs.

mmeeks (commented):
I think we should apply memproportion to the cgroup limit - ie take what we're given there and apply the * memLimit to it.

Otherwise this looks great =)

Ashod:
> I think we should apply memproportion to the cgroup limit - ie take what we're given there and apply the * memLimit to it.

Thanks. Updated.

wsd/Admin.cpp
```diff
@@ -514,15 +514,62 @@ Admin::Admin()
...
+    {
+        _totalAvailMemKb -= minHeadroomKb;
+    }
+
+    const size_t totalUsedMemKb = getTotalMemoryUsage();
+    _model.addMemStats(totalUsedMemKb);
+
+    LOG_INF("Total available memory: "
+            << _totalAvailMemKb << " KB, System memory: " << _totalSysMemKb
+            << " KB, configured memproportion: " << memLimit
```
mmeeks, line 563: We should dump the cgroup soft and hard numbers here, as well as the headroom number - so we can debug this algorithm in the field =)

mmeeks (approved):
Great - lets improve logging in a follow-up =)

Issue #7888, opened 2023-12-22
Dark mode shows white page on read onlymode


juliushaertl:
### Describe the Bug
Open a document without edit permissions in Nextcloud shows a broken dark mode where the page itself is still white

Also switching between dark/light mode or scrolling trying to rerender tiles does not change anything. 

Note I had another user in light mode joined the document.

### Steps to Reproduce
1. Open a file with read only permissions (Nextcloud indicates this to online with the UserCanWrite: false in the checkFileInfo repsponse
2. Switch to dark mode or have dark mode set from the integrator

### Expected Behavior
The page is also dark

### Actual Behavior
The page is white

### Screenshots
<img width="2560" alt="Screenshot 2023-12-22 at 09 44 03" src="https://github.com/CollaboraOnline/online/assets/3404133/2d910a9b-c685-41d7-84a3-a2b9222e3666">

### Additional Context
COOLWSD version: 23.05.6.3 (git hash: 3c909e4 (E))
LOKit version: Collabora Office 23.05.6.3 (git hash: f6f27ac)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: 4ab3d5e8

timur-g:
I saw ODT differently, was always dark for me. But there is a bug here anyway, so I will remove Unconfirmed. 
ODS has a different issue, document background changes light to dark, but font does not.

Darshan-upadhyay1110:
@juliushaertl thanks for reporting. I have added a fix for read-only mode. 

Thank you :)

PR #7906, opened 2023-12-29
Do not supress ChnageTheme event in readonly mode for any document

Darshan-upadhyay1110:
- allow `ChangeTheme` event to update document theme in read-only mode

Change-Id: Ic64bc61011a9e053c105e5c7540dddd4eb530681

* Resolves: #7888 
* Target version: master

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -505,8 +505,10 @@ L.Control.UIManager = L.Control.extend({
 
 	},
 	refreshToolbar: function() {
-		w2ui['editbar'].refresh();
-		w2ui['actionbar'].refresh();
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -505,8 +505,10 @@ L.Control.UIManager = L.Control.extend({
 
 	},
 	refreshToolbar: function() {
-		w2ui['editbar'].refresh();
-		w2ui['actionbar'].refresh();
+		if (w2ui['editbar'] && w2ui['actionbar']) {
```
eszkadev, line 508: maybe best will be to do it in 2 separate conditions, in case only one is ready

eszkadev (approved):

PR #7907, opened 2023-12-29
traceevent: re-factor and cleanup trace event emission.

mmeeks:
Share a single emission function. Events are re-parsed and emitted and adjusted in ClientSession; JSON escape event names.

Change-Id: If489d516bf20d5ec34cfe61f7e4fb09cf2e14bf2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -161,6 +161,18 @@ app.definitions.Socket = L.Class.extend({
 		}
 	},
 
+	sendTraceEvent: function(name, ph, timeRange, args, id, tid) {
+		if (timeRange === undefined)
+			timeRange = 'ts=' + Math.round(performance.now() * 1000);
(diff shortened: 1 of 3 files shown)


browser/src/core/Socket.js
```diff
@@ -1840,19 +1844,13 @@ app.definitions.Socket = L.Class.extend({
 		result.args = args;
 
 		if (this.traceEventRecordingToggle)
```
neilguertincollabora, line 1846: This if condition is redundant

neilguertincollabora (approved):

browser/src/core/Socket.js
mmeeks: True - but hey =) lets get this merged for now.

PR #7908, opened 2023-12-29
Set window.savedUIState to false in iOS app

plubius:
When building the iOS app, the %SAVED_UI_STATE% does not get replaced with a quoted string during the build and the resulting unquoted string causes document loading to stop.

Also, eliminate warnings when running in Xcode by adding missing CFBundleTypeRole entries in the iOS app's Info.plist.


timar (approved):

mmeeks:
Thanks Patrick! and a Happy New Year ! =)

PR #7910, opened 2024-01-01
Don't select columns, when clicks are over resize area

mikekaganski:
This allows to select several columns (or individual cells) at once,
and then double-click on the respective "border" area to resize them
automatically to their content (.uno:SetOptimalColumnWidthDirect).

Change-Id: I8b19ecb00a6285fec90d73ca09731b4393be6981

### DIFF
--- a/browser/src/control/Control.ColumnHeader.ts
+++ b/browser/src/control/Control.ColumnHeader.ts
@@ -180,6 +180,9 @@ export class ColumnHeader extends Header {
 		if (!this._mouseOverEntry)
 			return;
 
+		if (this._hitResizeArea)
+			return;
+
 		const col = this._mouseOverEntry.index;
 
 		let modifier = 0;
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
Double clicking on the column triggers the optimalColumnWidth function even before this patch. That's interesting. Though it doesn't seem intended, i am not sure if it is wanted. It is no harm to keep the behaviour IMHO. @pedropintosilva can i have your opinion?

* Double click on a column and its width is adjusted.
* Normally, one needs to double click on the resize area of column (the corner of it) to adjust its width.
* Shall we keep this behaviour?

Other than that, thanks for the improvements :)

pedropintosilva (approved):
thanks @mikekaganski , just tested!

meven:
FYI, I opened https://github.com/CollaboraOnline/online/pull/12452 that does exactly the same for rows.

Issue #7795, opened 2023-12-07
Human-Readable Dates/Times In "File > Repair"


Tex2002ans:
### Is your feature request related to a problem?

Pressing **File > Repair** on a document gives you 4 columns:

- Type
- What?
- Who?
- When?

The "When?" column is written using a full timestamp:

- `2023-12-07T01:53:40,596209456`
   - Unix? ISO 8601?

![Collabora 23 05 6 1 - File Repair Document Window](https://github.com/CollaboraOnline/online/assets/5774324/819f3cb5-d276-4900-b971-651686007a59)

### Describe the solution you'd like

Should probably be a given in a more human-readable/-understandable format:

- `2023/12/07 20:53:40`
- `12/07/2023 8:53:40 PM`
- `8:53:40 PM`
- `8:53 PM`

I'm suspecting hours/minutes/seconds should be enough for the normal person. Not down to the nanosecond level.

### Describe alternatives you've considered

None.

### Additional context

I came across this a few months back when explaining this menu to someone else:

- https://forum.collaboraonline.com/t/versioning-for-documents-in-cool/1986

And I just recently hit it on Android (23.05.5.2):

- https://forum.collaboraonline.com/t/how-do-you-undo-on-a-ipad/2232/2

The mobile version seems like it's an extremely poor idea, since there's very limited horizontal real-estate:

<img alt="Collabora 23 05 5 2 - Android - Repair Menu" src="https://github.com/CollaboraOnline/online/assets/5774324/a83a6c09-bbeb-4702-b000-9957c00f10c4" width="50%" />

codewithvk:
@Tex2002ans Hi, actually, the date format is ISO8601. I have converted it to local date and time with a human-readable format. PR: https://github.com/CollaboraOnline/online/pull/7911

Let me know your suggestion.

PR #7911, opened 2024-01-01
Make 'File > Repair' date and time human-readable

codewithvk:
Change-Id: Ib8b745fd10aca2eea2e829ad42cf1d91f7ff2129

* Resolves: #7795 
* Target version: master 

### Summary
We are receiving actions dates in ISO8601 format from Core. We can convert them to both local date and a human-readable format.

Before:
image

After: 
Screenshot from 2024-01-01 20-54-18


eszkadev (approved):
code looks ok

PR #7912, opened 2024-01-01
Private/timar/mobilefixes

timar:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -140,10 +140,6 @@ var Base64ToArrayBuffer = function(base64Str) {
     window.fullyLoadedAndReady = true;
   }, false);
 
-m4_ifelse(IOSAPP,[true],
-  [   window.savedUIState = false;],
-  [   window.savedUIState = %SAVED_UI_STATE%;])
-
 window.isLocalStorageAllowed = (function() {
   var str = 'localstorage_test';
   try {
@@ -362,6 +358,7 @@ m4_ifelse(MOBILEAPP,[true],
       window.checkFileInfoOverride = {};
       window.deeplEnabled = false;
       window.zoteroEnabled = false;
+      window.savedUIState = false;
       window.indirectionUrl='';],
      [window.host = '%HOST%';
       window.serviceRoot = '%SERVICE_ROOT%';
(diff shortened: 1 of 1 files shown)


plubius (commented):
This pull request generally works on iOS except for one item on both the read-only and notebook bar: PDF Document (.pdf) - Expert. For those items, the expected export dialog appears, but clicking OK hangs the app with the following animated graphic:

Screenshot 2024-01-03 at 9 49 34 AM

browser/html/cool.html.m4
```diff
@@ -356,6 +356,7 @@ m4_ifelse(MOBILEAPP,[true],
       window.frameAncestors = '';
       window.socketProxy = false;
       window.tileSize = 256;
+      window.groupDownloadAsForNb = 'true'
```
eszkadev, line 355: I think it should be just `true`, not `'true'`

eszkadev:
> This pull request generally works on iOS except for one item on both the read-only and notebook bar: PDF Document (.pdf) - Expert. For those items, the expected export dialog appears, but clicking OK hangs the app with the following animated graphic:
> Screenshot 2024-01-03 at 9 49 34 AM

this screenshot was made with patch in? I see not grouped download entries.

plubius:
> this screenshot was made with patch in? I see not grouped download entries.

Good catch. I forgot that I need to do a make clean, ./autogen.sh, ./configure to get the .m4 changes through on iOS. I ran into this when I was writing the iOS patch.

I'll retry with a clean build.

plubius:
> Good catch. I forgot that I need to do a make clean, ./autogen.sh, ./configure to get the .m4 changes through on iOS. I ran into this when I was writing the iOS patch.
> 
> I'll retry with a clean build.

So I now see the File > Download button in the notebook bar. I tested all of the items and only the "PDF export as expert" hangs the iOS app. All of the other items display a native save dialog and properly download the file.

IIRC, in previous versions of the iOS app, the "PDF export as expert" option would just do the same as the regular PDF export and an HTML dialog would not be displayed:

Screenshot 2024-01-03 at 10 42 15 AM

plubius (approved):
The hanging in the "busybox" in the iOS app is apparently an existing bug and occurs on master without this pull request. So, I think this pull request can be merged.

When I get some time, I'll if I can figure out why the busybox graphic never ends.

timar:
> The hanging in the "busybox" in the iOS app is apparently an existing bug and occurs on master without this pull request. So, I think this pull request can be merged.
> 
> When I get some time, I'll if I can figure out why the busybox graphic never ends.

I do not see the problem in released app. 
F4FAC5FC-0B4A-435F-AE19-88F19E73781F
We have to double check before next release.

plubius:
> I do not see the problem in released app.

I see the "busybox" graphic immediately after I press the Export button in the PDF dialog in the released Collabora Office version on my iPad.

I did some debugging last night and when I press the Export button, the JS code sends a "dialogevent 2 ..." message to the LibreOffice code but no messages are ever sent back to from LibreOffice to JS. I think that the JS code is waiting for some sort of "action completed" message.

PR #7913, opened 2024-01-01
Quarantine Improvements

Ashod:
Improves the quarantine path config and corrects the enabled/disabled state of quarantine.

    The quarantine path should be an absolute path.
    
    Unfortunately, because we had relative=true in
    the path config, we couldn't detect empty configs.
    This is because with relative=true the getter
    would create a path based on the current directory
    and the config value, which would default to
    "quarantine" when empty.
    This would result in /opt/cool/quarantine or
    /usr/bin/quarantine when in fact the path is
    really empty.
    
    Now, the config has relative=false and
    there is no default. In addition, we
    warn if the path is no absolute.
    

- wsd: quarantine: isQuarantineEnabled -> isEnabled and log it
- wsd: quarantine: absolute-path and better logging


timar (approved):
This is fine. I don't think we need to over-think this config change, wrt. upgrade, because it is a debug feature mostly used by us (developers) for crash reporting.

PR #7914, opened 2024-01-02
wsd: force autoSave on always_save_on_exit

Ashod:
This forces autoSave when always_save_on_exit
is true. This is needed so we can guarantee
that we don't have modification and that
we upload if there has every been one.
The latter case is checked in
DocumentBroker::needToUploadToStorage(),
which is called from
DocumentBroker::checkAndUploadToStorage().

A new test reproduces the issue and defends
the fix.

Change-Id: I0b2105a57cfd7049ba7b1f63e62a700fdc3744c2


eszkadev (approved):
Looks good

PR #7915, opened 2024-01-02
Private/mmeeks/debugfixes

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1843,8 +1843,7 @@ app.definitions.Socket = L.Class.extend({
 		result.active = true;
 		result.args = args;
 
-		if (this.traceEventRecordingToggle)
-			this.sendTraceEvent(name, 'S', undefined, args, result.id, result.tid);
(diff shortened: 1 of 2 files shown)


neilguertincollabora (approved):
I think it would be better to only build the state if the corresponding tool is on. As long as there is no need to turn a tool on and see data for events that happened previously, which I don't think there is.

So, for example, change `if (_debug)` to `if (_debugTileInvalidations)` instead, for anywhere that needs it.

The reason is that I'd like to keep the association between state and debug tool as tight as possible. There are a lot of them and it becomes harder to make changes if it's not clear where or why those variables are used.

I'm approving this anyways so you can merge it if you think this is the way to go.

mmeeks:
Yes - I don't believe it is worth the complexity of trying to work out what tool uses what setting - that's scattered across the code, and cleaning it up seems a fairly pointless waste of time I think; we can do it over time as we work on those features I guess. *Far* more important is the richer automated input, that is the top priority I think.

PR #7920, opened 2024-01-03
Make sure to clear "menu-opened" on popup close

mikekaganski:
Some popups call closePopover passing sendCloseEvent=false. This keeps the toolbutton's state as "opened"; and so, the next click on it would perform "close" cleanup. Only then a following click would activate the popup.

This was seen when clicking on the arrow of sidebar's Paragraph's Ordered List toolbar. Clicking on a popup's element to select a style, and then immediately clicking on the arrow again didn't open the popup.

Make sure that closePopover cleans the respective class from the element, even when not sending events. Simplify the arrow click handler, to not use a local variable, and rely on the class.

Change-Id: I923cbe28212e7fdedf91a78f64f9c1cae3d8b26f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2522,17 +2522,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var arrowbackground = L.DomUtil.create('div', 'arrowbackground', div);
 			var arrow = L.DomUtil.create('i', 'unoarrow', arrowbackground);
 			controls['arrow'] = arrow;
-			var menuIsOpened = false;
 			$(arrowbackground).click(function (event) {
 				if (!$(div).hasClass('disabled')) {
-					if (menuIsOpened) {
+					if ($(div).hasClass('menu-opened')) {
 						builder.callback('toolbox', 'closemenu', parentContainer, data.command, builder);
-						menuIsOpened = false;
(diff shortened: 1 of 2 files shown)


eszkadev:
as we discussed I prepared version without "state" to synchronise: https://github.com/CollaboraOnline/online/pull/7926

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2522,17 +2522,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var arrowbackground = L.DomUtil.create('div', 'arrowbackground', div);
 			var arrow = L.DomUtil.create('i', 'unoarrow', arrowbackground);
 			controls['arrow'] = arrow;
-			var menuIsOpened = false;
 			$(arrowbackground).click(function (event) {
 				if (!$(div).hasClass('disabled')) {
-					if (menuIsOpened) {
+					if ($(div).hasClass('menu-opened')) {
```
eszkadev, line 2527: I prefer to avoid this if

eszkadev (changes requested):

mikekaganski:
Thank you - since your change is superior, and fixes the cause of this problem, I'm glad to abandon this :-)

PR #7924, opened 2024-01-03
Core side handles the right click with mouse down event.

gokaysatir:
It internally handles the mouseup. So we don't send it again.

Without this change, context menus can't behave normal. Because mouse pointer is set to somewhere else with the mouse up event.

Change-Id: I719565b1ce192045250f9217e971725d8cc1003b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected on iPad

PR #7926, opened 2024-01-03
jsdialog: simplify dropdown closing

eszkadev:
This is partial revert of "jsdialog: use open/close instead of toggle for dropdowns" commit 0833ec64894fc6815748bb9ab7f7cc02073842a9.

Which introduced "state" variable which was desynchronised with core in some cases. Now directly call desired close command in closePopover.

This fixes bug:
1. Click on the sidebar's Paragraph's "Ordered list" dropdown arrow;
2. Click on any kind of the list (so that the current paragraph gets the chosen list style);
3. Click on the drop-down arrow again, expecting the drop-down to open. Result: you need to click two times to open menu again


mikekaganski (approved):
Looks nice, and works great in my testing. Thank you!

PR #7930, opened 2024-01-03
Add automated user debug tools

neilguertincollabora:
Change-Id: Ib6cb1080780f75dc0ea75930c757567e90c46bd8

### Summary
Typer: Type text into document. One keystroke every 50ms.
Move around sheet: For Calc only. Move to random corner or center of sheet and click on a cell there. Repeat every 1s. This expands the sheet very quickly.
Toggle dark mode: Toggles dark mode every 3s.
Add/Remove shapes: Adds a shape then removes it 2s later, repeat every 4s. This is a bit unstable because I couldn't reliably select the shape again, so it relies on the shape remaining selected for 2s. So it can't really be mixed with Typer or Move around sheet.
Zoomer: Select a random zoom level every 1s.

None of these makes for a complete or realistic user. Also having more than one of these selected is quite chaotic. It's fine if there are multiple windows and each has one tool going though.

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -5266,7 +5266,6 @@ L.CanvasTileLayer = L.Layer.extend({
 			},
 			onRemove: function () {
 				self._debugTileInvalidations = false;
-				clearTimeout(self._debugTyperTimeoutId);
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Code looks fine to me; thanks Neil.

neilguertincollabora:
Closing because I want to rework it to focus on randomizing settings instead.

PR #7934, opened 2024-01-04
Do not upload unmodified document

Ashod:
When always_save_on_exit=true we should
still not upload the document when it
isn't modified.

In this case, because we now always
save the document (forced) when
always_save_on_exit=true, and because
saving always generates a new file on
disk, with a new timestamp, we couldn't
detect that there are no modifications.

We now still force save, but ask Core
to skip it if the document is unmodified.

This is safe since we now always issue
the save, but rely on Core to do the
right thing. When the document is saved,
we do the normal upload as in that case
we know we have a new version of the
document, which must be uploaded.

Worth noting that the closedocument
command doesn't trigger the same path.
To reproduce the issue, we need a new
test that disconnects, instead of the
graceful closedocument command.

- wsd: logging and cosmetics
- wsd: do not upload unmodified document

### DIFF
--- a/test/UnitWOPISaveOnExit.cpp
+++ b/test/UnitWOPISaveOnExit.cpp
@@ -19,11 +19,7 @@
 
 #include <Poco/Net/HTTPRequest.h>
 
-#include "Util.hpp"
-#include "Log.hpp"
 #include "Unit.hpp"
-#include "UnitHTTP.hpp"
-#include "helpers.hpp"
 #include "lokassert.hpp"
 
 class UnitWOPISaveOnExit : public WOPIUploadConflictCommon
@@ -213,7 +209,7 @@ public:
     /// The document is loaded.
     bool onDocumentLoaded(const std::string& message) override
     {
-        LOG_TST("onDocumentLoaded: [" << message << ']');
+        LOG_TST("Got: [" << message << ']');
         LOK_ASSERT_STATE(_phase, Phase::WaitLoadStatus);
 
         LOG_TST("Modifying the document");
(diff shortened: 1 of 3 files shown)


eszkadev (approved):

test/UnitWOPISaveOnExit.cpp
```diff
@@ -213,7 +209,7 @@ class UnitSaveOnExitSaved : public WopiTestServer
     /// The document is loaded.
     bool onDocumentLoaded(const std::string& message) override
     {
-        LOG_TST("onDocumentLoaded: [" << message << ']');
```
eszkadev, line 216: isn't that better to have different strings for later debugging?
or `onDocumentLoaded` is already in the message?

mmeeks:
Are we certain that integrations are not using this up-load as a signal that the editing is finished; even if the document is unmodified ? =)

test/UnitWOPISaveOnExit.cpp
Ashod: It is already in the log. This is from before including the function name in the log. In this case it's duplicate and not helpful

Ashod:
> Are we certain that integrations are not using this up-load as a signal that the editing is finished; even if the document is unmodified ? =)

Good question! There is always that risk.

In this case, this particular issue is a regression introduced two days ago in PR #7914.

In all communication we make it clear that relying on such an upload is wrong, since that would be relying on a bug. Any such upload of unmodified documents would create duplicate versions in storage, waste disk, and confuse users (a bug). Though there has been requests to detect unloading of documents, and one can imagine the temptation of relying on always_save_on_exit.

Even if we had bugs and unwittingly allowed uploading of unmodified documents, I don't believe it ever worked reliably. We have tests that check that we don't do that (albeit, it apparently didn't cover disconnections).

So this shouldn't cause breakage in integration. At least I don't expect so.

PR #7937, opened 2024-01-04
Improvment in spinfield conatiner class.

Darshan-upadhyay1110:
Steps to reproduce:
    - Insert chart in odt
    - right click and select properties
    - see dialog looks odd

Problem:
- label element after spinfieldunit getting overlapped
- user can not see label tag after spinfield conatiner

Improvment:
- added css for spinfieldconatiner to fix overlapping issue.
- improve margin for more readability of .jsdialog.ui-text

Change-Id: If64d30d10021a023c76852647ef8aebdede0513a

Before:
image

After: 
image

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -649,6 +649,7 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 
 .ui-dialog-content .jsdialog.ui-text {
 	padding-inline-start: 0;
+	margin: auto 5px;
 }
 
 .jsdialog.ui-text {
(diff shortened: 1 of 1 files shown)


pedropintosilva:
note to self: this needs to be tested in different browsers since the current master branch has different results :  
image

----

pulling, testing

pedropintosilva (approved):
.spinfieldcontainer affects other places then dialogs, such as sidebar. It changes the alignment in the sidebar, but it doesn't seem to break. Please investigate further even after this PR is merged

browser/css/jsdialogs.css
```diff
@@ -657,6 +658,12 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 
 /* spinfield */
 
+.spinfieldcontainer {
+	margin-right: 5px;
```
eszkadev, line 662: it should be not left / right, but start / end. This is important for RTL mode cc: @pedropintosilva

Darshan-upadhyay1110: https://github.com/CollaboraOnline/online/pull/8026

PR #7942, opened 2024-01-04
Rename getFrame() to frame()

neilguertincollabora:
Change-Id: Ibff5c765ba37c1cf9946a75d05945186718e179a

### Summary
Shorter, easier to type, easier to read
Also removed reliance on setActiveFrame

### Checklist


cypress_test/integration_tests/common/helper.js
```diff
@@ -368,16 +368,16 @@ function checkIfDocIsLoaded(isMultiUser) {
 
 	if (isMultiUser) {
 		cy.cSetActiveFrame('#iframe1');
-		cy.getFrame().its('body').should('not.be.undefined');
+		cy.cframe('#iframe1').its('body').should('not.be.undefined');
```
gokaysatir, line 371: I guess we don't need to send the frameID since it is set at the line above? Same for line 375. If you agree then we can remove these with your next PR.

gokaysatir (approved):

PR #7949, opened 2024-01-05
Do not force background color for calc sheet tab when hover

Darshan-upadhyay1110:
- When we click on any sheet tab it has active + hover effect at same time
- by forcing !important in device-desktop.css it will break the visual effect on active tab when we hover over it

Before:

image

After:

image

Change-Id: I9e10f2a7e435b03e73c859ba742db0d7353aaa52


Darshan-upadhyay1110:
All button and secondary button hover will be effected by this change not worth to add in upstream. need more research on this. Closing this PR for now

PR #7954, opened 2024-01-05
Remove waits from cypress tests (make cypress tests 45% faster)

neilguertincollabora:
Change-Id: I6b7bbcfd580a7ddb454218eeba96019b408f53e5

### Summary
- Removed calls to cy.wait()
- Stubbed waitUntilIdle, which was usually acting as wait(1250)
- Added comments to some waits that are necessary (like holding for long press on mobile)
- Fixed failing tests by taking advantage of cypress's retry functionality. For example:
-- Add assertions that items have appeared, disappeared, or are no longer disabled
-- Replace .then() with .should()
-- Click on the underlying button instead of the div (better actionability detection)
-- Force click (skip actionability detection)
-- Use waitUntil to retry unsafe commands like clicking
-- Remove chained unsafe commands

### Not done
- I did not actually remove the waitUntilIdle or clickOnIdle functions because that would require touching many files.
- I focused on removing waits in helper files because they have the greatest impact. Waits in spec files are likely only called once and are more likely to be necessary.
- There are likely some flaky tests that I wasn't able to detect and fix.
- There are were some failing or flaky tests from before these changes that I did not attempt to fix.
- I did not run Nextcloud, proxy, or interference tests.

### Results
Time to run all desktop, mobile, and multiuser tests reduced from 85:11 to 47:01 (45% faster)

### Checklist

### DIFF
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -35,7 +35,6 @@ function clickOnFirstCell(firstClick = true, dblClick = false, isA1 = true) {
 	cy.log('Param - firstClick: ' + firstClick);
 	cy.log('Param - dblClick: ' + dblClick);
 
-	cy.wait(2000);
 	// Use the tile's edge to find the first cell's position
 	cy.cGet('#map')
 		.then(function(items) {
@@ -56,10 +55,6 @@ function clickOnFirstCell(firstClick = true, dblClick = false, isA1 = true) {
 			});
 	} else {
 		cy.cGet('.cursor-overlay .blinking-cursor').should('be.visible');
-
-		helper.doIfOnDesktop(function() {
-			cy.wait(500);
-		});
 	}
 
 	if (isA1)
@@ -204,6 +199,9 @@ function assertDataClipboardTable(expectedData) {
 		.should(function(cells) {
 			expect(cells).to.have.lengthOf(expectedData.length);
 		});
+
+	// Wait for clipboard to update anyways, in case previous contents were same length
+	cy.wait(200);
 
 	var data = [];
 
--- a/cypress_test/integration_tests/common/contenteditable_helper.js
+++ b/cypress_test/integration_tests/common/contenteditable_helper.js
@@ -19,7 +19,6 @@ function type(text, times = 1) {
 	}
 
 	cy.log('Clipboard - typing start.');
-	// cy.wait(50);
 	cy.get('@clipboard').type(input, {delay: 10, force: true});
 	cy.log('Clipboard - typing end.');
 }
@@ -77,14 +76,12 @@ function select(start, end) {
 }
 
 function checkHTMLContent(content) {
-	cy.wait(1000);
 	cy.get('@clipboard').should(($c) => {
 		expect($c).have.html($c.get(0)._wrapContent(content));
 	});
 }
 
 function checkPlainContent(content) {
-	cy.wait(1000);
 	cy.get('@clipboard').should('have.text', content);
 }
 
@@ -94,25 +91,21 @@ function checkSelectionRange(start, end) {
 		start = end;
 		end = t;
 	}
-	cy.wait(500);
 	cy.get('@clipboard').should('have.prop', 'isSelectionNull', false);
 	_checkSelectionStart(start);
 	_checkSelectionEnd(end);
 }
 
 function checkCaretPosition(pos) {
-	cy.wait(500);
 	_checkSelectionStart(pos);
 	_checkSelectionEnd(pos);
 }
 
 function checkSelectionIsNull() {
-	cy.wait(500);
 	cy.get('@clipboard').should('have.prop', 'isSelectionNull', true);
 }
 
 function checkSelectionIsEmpty(pos) {
-	cy.wait(500);
 	cy.get('@clipboard').should('have.prop', 'isSelectionNull', false);
 	_checkSelectionStart(pos);
 	_checkSelectionEnd(pos);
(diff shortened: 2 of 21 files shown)


mmeeks (commented):
Bold - but of course rather useful. I would suggest splitting this into smaller commits. Perhaps separate out all the improved waiting code commits without removing any of the waits, and get that to pass CI =)

Another thought: you should use 'cGet' and not .get - in almost all contexts, since AFAIR get() is not reliable with iframes, and we have had -lots- of nasty intermittent problems with that behavior.

The waitUntilIdle thing is ... well; we could spin that loop faster generally - but the problems of nodes being found and then removed by code rapidly mutating the DOM to create and destroy and re-arrange items - that we badly don't want to check in some half-done state are somewhat legion =) brave to remove that and hope.

So I'd split this into several separate pull requests and try to get each of them to pass CI =)

But really, Gokay is the expert here to date, might be worth a chat with him in your morning next week.

neilguertincollabora:
> The waitUntilIdle thing is ... well; we could spin that loop faster generally - but the problems of nodes being found and then removed by code rapidly mutating the DOM to create and destroy and re-arrange items - that we badly don't want to check in some half-done state are somewhat legion =) brave to remove that and hope.

waitUntilIdle was the single biggest contributor to the total wait time, called 500 times, that's over 10 minutes of waiting. Most of the time it seems to be used just as a general wait, and when tests did fail they could be fixed by waiting for something specific, like a button not having class 'disabled'. I did not see any failures caused by items being detached, though I do see from my before.log 28 instances where items were detached at some point. I'll look at those to see if there's something we need to do for those tests specifically. But I do believe that each case should be handled separately, instead of inserting a too-long wait all the time.

> Another thought: you should use 'cGet' and not .get - in almost all contexts, since AFAIR get() is not reliable with iframes, and we have had -lots- of nasty intermittent problems with that behavior.

The `.get('@input')` in this change is an alias, getting something that was already found with cGet. This should be safe, at least it has worked for me so far.
By the way I've already made some improvements to cGet and there are more big changes I'd like to make, but that's another discussion for another time.

gokaysatir:
We have only one commit and many different changes here. We suffered from tests' instability. It would be really more maintainable if this code was split into smaller commits.

Many of timeouts were put based on experience. I agree that it will be a huge forward step if can remove them. But a step by step approach would be much safer and i guess is no harm:)

neilguertincollabora:
Is there a way we can run more tests to gain more confidence on these changes? I see code for nextcloud and proxy tests but I don't know how to run them. And I tried to run the interference tests but they hung on my machine (even without these changes). Also, I see there was a failure in Jenkins but it doesn't seem to be a cypress failure so it's likely not related but I'd want to resolve that too.

> It would be really more maintainable if this code was split into smaller commits.

I don't see much value in that. I don't think it would reduce the overall number of failures or make them easier to investigate, only spread it out over a longer time. I did develop the change in parts - for example first remove the waits in desktop_helper.js and run just the desktop tests and resolve failures, then mobile, and so on - but it wasn't really in a commitable state until now.

gokaysatir:
Okay, let me check the PR tomorrow and i will also run the tests for a few times. The issue is that sometimes unstable tests pass and we merge them. Then they start to fail here and there. This is why we became cautious with these changes.

I hope i can confirm that these changes work - tomorrow.

mmeeks:
Well - splitting into smaller changes means less code thrash as we apply and revert your change. Other engineers can also see that it is the well-known "Test Foo" where we removed the waits that is causing the intermittency. We should try to do this incrementally with our safest things first, and get feedback / metrics on CI stability.

We can't afford to gum up Jenkins more than it is, and stop commits getting in by breaking the build - we're currently at a long-time, and some 50% intermittency, going to a slightly-shorter-time and 95% intermittency is not a good trade: we need to be merging lots of code fast =)

Might be good to focus first on fixing intermittency - to get a better idea of the horrible race conditions that exist out there (for which the waits are a bad fix) . *And* there is no harm in doing a wait, and then also doing a more correct check as you suggest - so I would (please cf. "New Starter reading ..." E-mail for the meaning of "I would") merge all the nicer loop-until-something-is-good stuff and then experiment with merging a removal of the sleep =)

Thanks.

neilguertincollabora:
I have made a few more updates and stabilized things even more.

I ran all of the tests 5 times today, and over all 5 runs there were a total of 4 failures (these do not cause the job to fail since they all passed on rerun). Compare this to the last 5 jobs in the tinderbox which have 22 test failures and 2 failed jobs.

These are stable enough to submit, and I am monitoring the tinderbox anyways so I will catch and fix any issues that were not found already.

neilguertincollabora:
All tests passed three times in a row locally with zero failures.
All tests passed twice in a row in CI with only three reruns total due to the existing document loading failure.

mmeeks:
I'm excited to get this in, but Szymon is probably right. Also we should increase parallelism here if we can =) lets bump that right up - my hope is with the C++ unit test fix, we can get that much higher;  lets remove:

test/Makefile.am:
AUTOMAKE_OPTION = serial-tests

and see if we can get it going; ultimately load from ourselves is probably the best stressor we can ask for - giving more speed left and right anyway =)

neilguertincollabora:
All jobs passed:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9954/
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9955/
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9956/
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9957/
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9958/

Excluding the known failure on loading documents, there was an average of 1.2 tests that had to be rerun per job. It seems there are more failures under load, but not by much.

eszkadev:
@neilguertincollabora Great! Then it seems to be better than last week state, where I had to trigger every PR twice.

eszkadev (approved):
For me looks promising :)

PR #7955, opened 2024-01-05
spdx: improve machine and human readability of headers.

mmeeks:
Change-Id: I2432e3c437f3a68c361da55d3aaf7a53c3f37838

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir:
Control.Scroll.Annotation.js file is still used :) When there are more than one comments in different slides of an Impress file, the buttons get visible. They provide navigating between slides which have comment(s).

gokaysatir (approved):

PR #7957, opened 2024-01-05
Implement LOK_CALLBACK_EXPORT_FILE for download requests in the iOS app

plubius:
Have the document's DocumentViewController display a save dialog and export a copy of the file written by the LibreOffice code.


plubius:
Note: for the "PDF export as expert" menu items to work in the iOS app, the following patch is needed in the LibreOffice build:

https://gerrit.libreoffice.org/c/core/+/161692

timar (approved):

Issue #7959, opened 2024-01-06
Calc control-header entry highlighting is broken 


get-state:
### Describe the Bug
In the control header for rows/columns hovering the mouse over the control header "entry" the entry gets highlighted correctly, but the highlight never gets removed. 

### Steps to Reproduce
1. Open Calc.
2. Hover mouse pointer over random column/row entries in the control header. 

### Expected Behavior
After hovering over another entry, the initial entry's "highlight" must be removed. Exiting the control header area should also remove the highlight if the entry is not selected.

### Actual Behavior
Hovering over a control header entry and leaving it never removes the highlight.

### Desktop
(Please complete the following information)
 - **Collabora version:** 23.05.6.4
 - **OS and version:**  Arch - Kernel 6.6.9
 - **Browser and version:** FF nightly 123.0.a1 (2023-12-21)

PR #7960, opened 2024-01-06
Fix control header entry highlighting

get-state:
The header entry initialization does not correctly set the "backgroundColor" or "hoverColor" and fails silently. Fix that by grabbing a known CSS property.

The "onMouseMove" callback incorrectly sets the previous entry to the resulting (new) entry before using it within a comparison. This causes the branch to never be taken. Fix it by saving the previous entry and use the previous entry when appropriate.

In "onLeave" set the "isOver" property to false so the redraw removes the highlight.

Change-Id: I801fd3e6b6759b0ab66a99d68cd0ef80f1f3737e
Signed-off-by: Mazin Alhaddad <mazin@getstate.dev>

* Resolves: #7959 
* Target version: master 

### Checklist

- [] Documentation (manuals or wiki) has been updated or is not required

### DIFF
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -42,6 +42,7 @@ export class Header extends CanvasSectionObject {
 	_headerInfo: HeaderInfo;
 	_dragEntry: HeaderEntryData;
 	_mouseOverEntry: HeaderEntryData;
+	_prevMouseOverEntry: HeaderEntryData;
 	_startSelectionEntry: HeaderEntryData;
 	_lastSelectedIndex: number;
 	_hitResizeArea: boolean;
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):
I checked and tested, thanks!!

But I would love to have a review and +2 from @mikekaganski and/or @gokaysatir on this

mikekaganski (approved):
Great - it makes the backgrounds non-transparent.
My change ( https://github.com/CollaboraOnline/online/pull/7987 ) is related, but different (except for the mouse events, where I did exactly the same) - it makes sure that even (semi)transparent background wouldn't re-introduce the problem. Please take a look! Thanks

Issue #7811, opened 2023-12-11
Calc: Impossible to create same level groups


pedropintosilva:
### Describe the Bug
Calc: Impossible to create same level groups

### Steps to Reproduce

1. Select a couple of cell (vertically)
1. Create a group
1. Select again cells (fewer) from the top 
1. Create a group
1. Select again cells but this time from the bottom end of the first group
1. Create a group -> group is not created, no error message nor any feedback is displayed

If the last action is repeated a child ends up to be created but on the next level.
 

https://github.com/CollaboraOnline/online/assets/65948705/8d45ab91-f574-4bbd-b2ce-db051a9b312d

I have double checked and that particular bug is already reproducible on master before https://github.com/CollaboraOnline/online/pull/7722 was merged.

### Version

* online: Master Head on 1bd1e5a8281915b66f029b14885c5cb78adff448
* core: eec05aa9cf4a283efa3be76ecf18aa77f714d26e

PR #7961, opened 2024-01-06
Calc grouping: find parent groups correctly

bayramcicek:
- groups was using indexes to find their parent groups.
  - this does not work properly and causes following issues: 
    - some groups are created but not drawn (so they are invisible)
    - some parent groups does not hide their child groups

- this patch:
  - removes the index-based search of parent groups.
  - implements startPos and endPos based search.

Change-Id: Ia3b85f3d9a032008f8abbc7b38ad2417aa0ec6c8

* Resolves: #7811
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.ColumnGroup.ts
+++ b/browser/src/control/Control.ColumnGroup.ts
@@ -194,7 +194,9 @@ export class ColumnGroup extends GroupBase {
 						const startY = this._levelSpacing + (this._groupHeadSize + this._levelSpacing) * group_.level;
 						const endX = startX + this._groupHeadSize;
 						const endY = startY + this._groupHeadSize;
-						if (this._isPreviousGroupVisible(group_.index, group_.level) && this.isPointInRect(point, startX, startY, endX, endY, mirrorX)) {
+						if (group_.level == 0 && this.isPointInRect(point, startX, startY, endX, endY, mirrorX))
+							return group_;
+						else if (this._isPreviousGroupVisible(group_.level, group_.startPos, group_.endPos, group_.hidden) && this.isPointInRect(point, startX, startY, endX, endY, mirrorX)) {
 							return group_;
 						}
 					}
--- a/browser/src/control/Control.GroupBase.ts
+++ b/browser/src/control/Control.GroupBase.ts
@@ -127,8 +127,8 @@ export abstract class GroupBase extends CanvasSectionObject {
 				if (!moved && lastGroupIndex[level] !== undefined) {
 					const prevGroupEntry = this._groups[level][lastGroupIndex[level]];
 					if (prevGroupEntry.hidden) {
-						if (startPos > prevGroupEntry.startPos && startPos < prevGroupEntry.endPos) {
-							startPos = prevGroupEntry.endPos;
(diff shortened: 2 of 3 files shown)


bayramcicek:
some recordings after the patch:

https://github.com/CollaboraOnline/online/assets/24255179/af829248-1cf8-4a6c-8af8-bebfc042e52c

https://github.com/CollaboraOnline/online/assets/24255179/c8ba9ea3-8349-4f48-b7e3-ecebd284685a

https://github.com/CollaboraOnline/online/assets/24255179/c3166b22-67bc-40e5-b8d5-4baf3fbb984f

gokaysatir:
Some part of this looks like regression. Anyway, 2 suggestions:
* I strongly recommend to simplify the "_isPreviousGroupVisible" and "_isParentGroupVisible" functions.
    * There is group-index property, which distinguishes the groups of the same level. I think we can make use of it, to avoid the issue in the images.
* We need Cypress tests for this feature. If it can cover some of the issues solved here, that'd be great.

There is an issue with the behaviour, please check the screen shots and let me know if anything is unclear.
* First 2 are from your PR.
* Last 2 are from desktop build.

gokaysatir (changes requested):
Please see my first comment.

bayramcicek:
(if I am not mistaken) I see that the issue on the screenshots are "overlapping issue" that is known issue - but I didn't report it yet. The issue is already exist on online master:

https://github.com/CollaboraOnline/online/assets/24255179/cdf6628d-0ea9-4c02-aab4-28b72a4cfbac

---

and I don't see any plus sign when I create the same groups as in the screenshots you've shared:
 (probably I followed the wrong steps?)

(we can also see the overlap issue here - which is not related to my patch)

https://github.com/CollaboraOnline/online/assets/24255179/23975089-8443-4e8a-a1db-29e7ca13661c

Thanks!

gokaysatir:
Okay, please add 2 groups to the same level and collapse the upper one. Then the drawing is faulty i think.

gokaysatir:
The 2 groups should touch btw. Like "A1:A20" and "A21:A40" etc.

bayramcicek:
> Okay, please add 2 groups to the same level and collapse the upper one. Then the drawing is faulty i think.

Ok. Thanks. Then I'm working on the overlapping issue...

bayramcicek:
about commit: https://github.com/CollaboraOnline/online/pull/7961/commits/ab306916d7ca78809e1e483593211a14fcd670d3 (Calc grouping: fix overlapping issue)

Behavior on Online vs LO:

**Online**: expanding order starts with the **first** group.
**LibreOffice**: expanding order starts with the **latest** group.

Both of them seem plausible for me. So IMHO, there is no problem about the expanding order.


https://github.com/CollaboraOnline/online/assets/24255179/3a57823e-1b31-46c4-8033-12289b7408dc

https://github.com/CollaboraOnline/online/assets/24255179/c4aa372f-9152-44ae-89a4-057b186ce393

gokaysatir (approved):
Thanks for the fixes @bayramcicek :)

pedropintosilva (approved):
Tested :+1: Thanks!

PR #7962, opened 2024-01-07
ensure that progressbar in mobile apps doesn't display 'Collabora Online'

timar:
ensure that progressbar in mobile apps doesn't display 'Collabora Online'

Change-Id: I972b3acdbabc24df8d51643215c7bf5c9d226a65


Minion3665 (approved):
Makes sense & works in my testing

PR #7963, opened 2024-01-07
Add color palette selector to Calc and Impress

eszkadev:
It will be possible to use "theme colors" then.
It requires core part: https://gerrit.libreoffice.org/c/core/+/161771
sc: announce theme colors with lok callback


eszkadev:
It passed previously but CodeQL cannot start. Tried in new PR - same effect.
Previous successful Cypress run: https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9318/

PR #7968, opened 2024-01-08
Set titlebar max width equal to dialog container width

GulsahKose:
Prevents to overflow when the range name is long in conditinal format dialog

Change-Id: Ib845bc84a85d90dec87115cdcd10ae8704163081

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.LokDialog.js
```diff
@@ -660,6 +660,9 @@ L.Control.LokDialog = L.Control.extend({
 		// Override default minHeight, which can be too large for thin dialogs.
 		L.DomUtil.setStyle(dialogContainer, 'minHeight', height + 'px');
 
+		// Title bar may overflow due to range name. So we should have max width
+		L.DomUtil.setStyle(dialogContainer.previousSibling, 'maxWidth', width - 36 + 'px');
```
GulsahKose, line 664: 36 is close button's size and margin. I think it doesn't worth to do complicated calculation here to find exact width . I'm open to suggestions if there is a simple way to get it.

pedropintosilva:
pulling and reviewing

pedropintosilva (changes requested):
Thanks @GulsahKose it works but now we have introduced a regression, see the following grey space (after the close icon)
image

The rational is correct, but probably wrong element. If the same is applied to its span child we get the desired result without regression:

image

browser/src/control/Control.LokDialog.js
eszkadev: to which object points dialogContainer.previousSibling ?

eszkadev:
How to reproduce the issue? would be good to put steps in commit message.

pedropintosilva:
> How to reproduce the issue? would be good to put steps in commit message.

agreed,

feel free to include @GulsahKose something along these lines:
```
1. Create a conditional format where the range is very long (inserted manually)
2. Home > Conditional formatting
3. Create
4. Scroll and paste the following in the Cell Range > Range: 
`N3:AL4,N25:AA25,AC25:AL25,X11:AL11,N6:P7,Q5:AL10,Q12:AL13,N5:N7,O5:P5,N14:AL19,N11:P13,A8:C11,N8:P10,C20:J20,O20:AL20,A20:A27,D27,D10,T21:AL24,A3:K7,A12:K19,H8:K10,H27:K27,A28:K89,N26:AL89,M5,D11:K11,Q11:V11,L3:L89,N21:S22,N24:S24,M23:S23,B21:K26`
5. The title never gets truncated and the dialog is too wide
```

browser/src/control/Control.LokDialog.js
GulsahKose: Titlebar

pedropintosilva (changes requested):
@GulsahKose This is what I see:

https://github.com/CollaboraOnline/online/assets/65948705/7948f086-eb74-457b-9fa8-d7d2d5ceb8eb

browser/src/control/Control.LokDialog.js
```diff
@@ -660,6 +660,12 @@ L.Control.LokDialog = L.Control.extend({
 		// Override default minHeight, which can be too large for thin dialogs.
 		L.DomUtil.setStyle(dialogContainer, 'minHeight', height + 'px');
 
+		// Title bar may overflow due to range name. So we should have max width
+		var titleBar = dialogContainer.previousSibling;
+		var leftPadding = window.getComputedStyle(titleBar).getPropertyValue('padding-left');
+		var rightPadding = window.getComputedStyle(titleBar).getPropertyValue('padding-right');
+		L.DomUtil.setStyle(titleBar, 'maxWidth', width - (leftPadding + rightPadding) + 'px');
```
pedropintosilva, line 667: The problem is here, leftPadding and rightPadding are coming with `px` suffix

pedropintosilva: so either you parse that before the calculation, which seems a bit ineffective since anyhow we end up adding the at the end. OR you simply set the style wrapped in `calc()` and let the css handle the calculation: so at the end the browser seems something like `style="max-width: calc(600px - (16px + 16px));"`

pedropintosilva (changes requested):

pedropintosilva:
```diff
-		L.DomUtil.setStyle(titleBar, 'maxWidth', width - (leftPadding + rightPadding) + 'px');
+		L.DomUtil.setStyle(titleBar, 'maxWidth', 'calc(' + width + 'px ' + '- (' + leftPadding + ' + ' + rightPadding + '))');
```

pedropintosilva:
> > How to reproduce the issue? would be good to put steps in commit message.
> 
> agreed,
> 
> feel free to include @GulsahKose something along these lines:
> 
> ```
> 1. Create a conditional format where the range is very long (inserted manually)
> 2. Home > Conditional formatting
> 3. Create
> 4. Scroll and paste the following in the Cell Range > Range: 
> `N3:AL4,N25:AA25,AC25:AL25,X11:AL11,N6:P7,Q5:AL10,Q12:AL13,N5:N7,O5:P5,N14:AL19,N11:P13,A8:C11,N8:P10,C20:J20,O20:AL20,A20:A27,D27,D10,T21:AL24,A3:K7,A12:K19,H8:K10,H27:K27,A28:K89,N26:AL89,M5,D11:K11,Q11:V11,L3:L89,N21:S22,N24:S24,M23:S23,B21:K26`
> 5. The title never gets truncated and the dialog is too wide
> ```

Also this ^ still missing a more contextualized commit message

browser/src/control/Control.LokDialog.js
pedropintosilva: so either you parse that before the calculation, which seems a bit ineffective since anyhow we end up adding the at the end. OR you simply set the style wrapped in calc() and let the css handle the calculation: so at the end the browser seems something like style="max-width: calc(600px - (16px + 16px));" `'calc(' + width + 'px ' + '- (' + leftPadding + ' + ' + rightPadding + '))');`

@eszkadev 
> Parsing will have non-visible impact (it's single operation). Such thing done thousand times per second would be a problem.
> I'm happy with both solutions. Correct paring or css calculation.
> we do 'px' removal in many places already...

pedropintosilva (approved):
Thanks!

browser/src/control/Control.LokDialog.js
```diff
@@ -660,6 +660,12 @@ L.Control.LokDialog = L.Control.extend({
 		// Override default minHeight, which can be too large for thin dialogs.
 		L.DomUtil.setStyle(dialogContainer, 'minHeight', height + 'px');
 
+		// Title bar may overflow due to range name. So we should have max width
+		var titleBar = dialogContainer.previousSibling;
+		var leftPadding = window.getComputedStyle(titleBar).getPropertyValue('padding-left').slice(0, -2);
+		var rightPadding = window.getComputedStyle(titleBar).getPropertyValue('padding-right').slice(0, -2);
+		L.DomUtil.setStyle(titleBar, 'maxWidth', width - (+leftPadding + +rightPadding) + 'px');
```
eszkadev, line 667: could it be just: `width - (leftPadding + rightPadding)` ?

browser/src/control/Control.LokDialog.js
```diff
@@ -660,6 +660,12 @@ L.Control.LokDialog = L.Control.extend({
 		// Override default minHeight, which can be too large for thin dialogs.
 		L.DomUtil.setStyle(dialogContainer, 'minHeight', height + 'px');
 
+		// Title bar may overflow due to range name. So we should have max width
+		var titleBar = dialogContainer.previousSibling;
+		var leftPadding = window.getComputedStyle(titleBar).getPropertyValue('padding-left').slice(0, -2);
```
eszkadev, line 665: In CSS there are units with more than 2 (rem) and less than 2 characters (%).
https://www.w3schools.com/cssref/css_units.php

To make code bulletproof we should take it into account.
BTW I see that this titlebar is using em as units. Here we add this to px.
```
.ui-dialog .ui-dialog-titlebar {
    padding: 0.4em 1em;
```

Maybe as we do it programatically anyway, can we declare here some constant variable: `titlePadding`, which we will set here and on creation of the dialog? So we will be sure it is equal.

browser/src/control/Control.LokDialog.js
GulsahKose: That gives wrong result unfortunately.

eszkadev: Because its string value. Better to convert type and make calculations on numbers, so it's more obvious.

browser/src/control/Control.LokDialog.js
GulsahKose: Couldn't understand clearly but. Although units defined em here, "window.getComputedStyle(titleBar).getPropertyValue returns "px". One day it slight probability can return something else. But can't we trust the returned value now?

browser/src/control/Control.LokDialog.js
GulsahKose: ok doing now

PR #7970, opened 2024-01-08
Add color palette selector to Calc and Impress 2

eszkadev:
It will be possible to use "theme colors" then.
It requires core part: https://gerrit.libreoffice.org/c/core/+/161771

sc: announce theme colors with lok callback


quikee (approved):
Looks good to me

PR #7971, opened 2024-01-08
Center header/group lines in the pixel grid

OFFTKP:
Round width and center lines of row/column headers so that they don't overflow in the pixel grid and cause antialiasing artifacts at zoom levels > 100%.

Also use constant linewidth of 1 for group level headers and of size 2 for group controls for the same reason.

110% zoom:
image

150% zoom:
image


gokaysatir (approved):
Column and row group drawings certainly look better. I couldn't see much changes in row & column headers, my poor eyes may have missed the difference:)

Anyway, thank you for the improvements.

mmeeks (approved):
Looks plausible still =) any chance of a screenshot at 120% browser zoom and 150% document zoom or somesuch ? =)

OFFTKP:
@mmeeks 
125% browser, 150% document:
image

150% browser, 150% document:
image

mmeeks:
Much better - you must think me some sort of ingrate =) but:

image

Is also an issue - and more noticable now we're making everything else prettier - visible in the 150% zoom. Any chance of tweaking those pixels too ?

PR #7972, opened 2024-01-08
Private/gokay/calc impress hyperlink info at position

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks @gokaysatir. Looks reasonable.

vmiklos:
Is disabling the notebookbar in read-only mode a short-term workaround? My understanding was that with time we want to drop the classic UI and just keep notebookbar; if that's true, then it's strange to go in the opposite direction in read-only mode.

CC @pedropintosilva

eszkadev:
Our readonly mode uses classic menu still - we will need to prepare our menu replacement to drop classic mode.
We had recently bug that it opened notebookbar in readonly mode, which was empty and was non-functional. Probably that is the reason.

PR #7975, opened 2024-01-08
Input tools

neilguertincollabora:
### Summary
Added two options:
Randomize user settings: Sets dark mode, zoom, spell check, position in sheet
Automated user input: loops over actions: type in cell/formula, delete cell/row, insert/type in shape
These replace the chaotic inputs from before

I encountered a crash somewhere around editing the text in the shape. I got the stack trace and am trying to get better reproduction steps.

It's not ready to submit yet but I want to get some feedback:
- What do you think of the approach, a toggle for starting the input and a toggle (button, really) for randomizing the settings
- Additional settings to toggle (I understand interface language is not possible)
- Additional tasks to do (I already plan to select a block of cells and change the row height)
- How do I programatically click in the sheet? (Just center is fine. I looked through the methods on _map and couldn't find anything useful)

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -5101,6 +5101,8 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._debugRenderCount = 0;
 		this._debugTimePING = this._debugGetTimeArray();
 		this._debugPINGQueue = [];
+		this._debugAutomatedUserQueue = [];
+		this._debugAutomatedUserTasks = {};
 
 		if (this.isCalc()) {
 			this._painter._addSplitsSection();
@@ -5176,7 +5178,7 @@ L.CanvasTileLayer = L.Layer.extend({
 	},
 
 	_addDebugTools: function () {
-		var self = this;
+		var self = this; // easier than using (function (){}).bind(this) each time
 
 		this._addDebugTool({
 			name: 'Screen Overlays',
@@ -5344,25 +5346,12 @@ L.CanvasTileLayer = L.Layer.extend({
 			category: 'Automated User',
 			startsOn: false,
 			onAdd: function () {
-				self._debugLorem = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.';
-				self._debugLorem += ' ' + self._debugLorem + '\n';
+				self._debugLorem = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.\n';
 				self._debugLoremPos = 0;
 				self._debugTyperTimeout();
 			},
 			onRemove: function () {
 				clearTimeout(self._debugTyperTimeoutId);
-			},
-		});
-
-		this._addDebugTool({
-			name: 'Add/Remove shapes',
-			category: 'Automated User',
-			startsOn: false,
-			onAdd: function () {
-				self._debugAddRemoveShapesTimeout();
-			},
-			onRemove: function () {
-				clearTimeout(self._debugAddRemoveShapesTimeoutId);
 			},
 		});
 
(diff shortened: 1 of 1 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -5399,26 +5413,74 @@ L.CanvasTileLayer = L.Layer.extend({
...
 
-	_debugAddRemoveShapesTimeout: function () {
-		var shapes = ['smiley','sun','moon','lightning','heart','flower'];
-		var shape = shapes[Math.floor(Math.random() * shapes.length)];
-		this._map.sendUnoCommand('.uno:SymbolShapes.'+shape);
-
-		setTimeout(L.bind(function() {app.socket.sendMessage('removetextcontext id=0 before=0 after=1');},this), 2000);
+	_debugAutomatedUserInputTimeout: function () {
+		switch (this._debugAutomatedUserInputCase) {
+			case 0:
```
mmeeks, line 5425: This looks fine; I'd prefer to have a set of tools that are added into this state machine and each of which has a mini-state-machine of its own - and the outer tool wrapper manages those to make sure the overall number range 1->100 is split into ten 0 -> 10 state machines (or for however many states they ask for). This infrastructure would then allow tools to be included into the larger state-machine loop at will with toggles =) clearly on registration they'd need to state how many states they want to request; and perhaps the time for each one - up to you =)

We should then do typing with a reasonable cadence - not throwing lots of keys in at a time which is unlike a real user - but typing one every 100ms or so.

Otherwise this looks promising =)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -5428,68 +5456,114 @@ L.CanvasTileLayer = L.Layer.extend({
...
-				this._debugTypeText('textinshape', 100);
-				break;
-			case 8:
-				console.log('Automated User: Type Escape');
-				this.postKeyboardEvent('input',0, 1281); // esc
-				break;
-			default:
-				this._debugAutomatedUserInputCase = 0;
+	_debugAutomatedUserTypeCellFormula: function () {
+		setTimeout(L.bind(function () {
```
mmeeks, line 5467: I would be inclined to manage the underlying timeout setting with a parent helper function - make the state machine more of a 'switch' or a if (state <10 { emit_one_character(string[state]); return 50; } - and make that function return the timeout for the next state switch. Hopefully make the code easier to read.

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -5428,68 +5456,114 @@ L.CanvasTileLayer = L.Layer.extend({
...
+			console.log('Automated User: Click somewhere visible');
+			var pos = this._pixelsToTwips(this._map._getCenterLayerPoint());
+			this._postMouseEvent('buttondown',pos.x,pos.y,1,1,0);
+			this._postMouseEvent('buttonup',pos.x,pos.y,1,1,0);
+		}, this), 0);
+
+		setTimeout(L.bind(function () {
+			console.log('Automated User: Type text');
+			this._debugTypeText('asdf\nqwer\n', 100);
```
mmeeks, line 5476: I'd want _debugTypeText hooked into the state machine logic really; typing big chunks of text as a ton of keystrokes in a single message, or as a single message with lots of keystrokes in it is a very unusual behavior; people type characters in one at a time - and we need that time to be realistic.
So the majority of _debugTypeText calls should send just 1 character I think.

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -5573,9 +5573,13 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._debugTyperTimeoutId = setTimeout(L.bind(this._debugTyperTimeout, this), 50);
 	},
 
-	_debugTypeText: function(text) {
+	_debugTypeText: function(text, delayMs) {
 		for (var i=0; i<text.length; i++) {
-			this._debugTypeChar(text.charCodeAt(i));
+			if (delayMs) {
+				setTimeout(L.bind(this._debugTypeChar, this, text.charCodeAt(i)), i*delayMs);
```
mmeeks, line 5653: Looks better, but how do we chain these things helpfully ? promises or something clever ? I guess the latest JS async support is not available in all browser and will not parse and/or fall in a heap for older users but - ... =)

mmeeks (commented):
Lots of overlapping patches; best to squash them down I think for the next review.

neilguertincollabora:
I updated the task queue and timeout system. I'm happy with how simple it turned out.
 - Tasks authors don't have to worry about timeouts at all.
 - There is only one timeout id used, making it easy to cancel at any time
Look at _debugAutomatedUserTypeCellFormula for an example of how a task would be written. Just a switch statement, and each phase returns the wait time afterwards.
All the timeout and task/phase management happens in _debugAutomatedUserTimeout.

mmeeks (approved):
Looks really good thanks Neil! =)

PR #7976, opened 2024-01-08
_debugGetTimeArray not found

caolanm:
since:

commit dbc4227fa373b6bdaedb727aef0b10a09e408a36
Date:   Tue Jan 2 12:37:45 2024 +0000

    debug menu: initialize state on enabling debug.

Change-Id: I600583b670a0cd5846c95d297703c9d79f98a71a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
There is only one _debugGetTimeArray(), the 2 other uses already are like this, so looks OK. (We should really have a sample commit that converts one .js file to .ts, similar to how dialogs were converted to .ui back in the day, I guess the TS compiler would catch this.)

PR #7977, opened 2024-01-08
Fix failing test desktop/calc/focus_spec.js

neilguertincollabora:
Change-Id: Ibcd6a2b3d27cbd56d8a24de5f12316bacbd2061f

* Resolves: # 
* Target version: master 

### Summary
The test is about typing into the formula bar. The issue was that clickOnFirstCell hasn't finished, so clicking on the formula bar didn't work. Specifically, the accept and cancel buttons weren't visible.

I added a short wait after calling clickOnFirstCell, and cleaned up the code a bit. Also uncommented the second half of the test.

### Checklist


gokaysatir (approved):
Test works better now, thanks for the improvement.

PR #7978, opened 2024-01-08
Fix failing test multiuser/writer/annotation_spec.js

neilguertincollabora:
Change-Id: Id1be5607e15ac23596e527c3e27fca83c91ee380

* Resolves: # 
* Target version: master 

### Summary
Many testpoints in this file were sporadically failing. Tests add a comment and then immediately modify it. I'm not sure how exactly, but starting to edit the annotation as it was animating from the top left to the right caused text to be inserted in the wrong spot.

I added a wait directly after saving the new annotations. I ran the tests several times with no failures. Tests failed about half the time before adding the wait.

### Checklist


gokaysatir (approved):
Cool. Now tests are passing without second tries. Which seems to save time, in spite of the cy.wait command.

PR #7990, opened 2024-01-09
context menu: allow edit section for Writer text sections

meven:
Change-Id: I40a40a99551a756ad234e73956169018b761fa42

* Target version: master 

### Summary

Allow edit section for Writer text sections, the dialog is made async in https://gerrit.libreoffice.org/c/core/+/161828.

### Checklist

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -55,7 +55,7 @@ L.Control.ContextMenu = L.Control.extend({
 			text: ['TableInsertMenu',
 				   'InsertRowsBefore', 'InsertRowsAfter', 'InsertColumnsBefore', 'InsertColumnsAfter',
 				   'TableDeleteMenu', 'SetObjectToBackground', 'SetObjectToForeground',
-				   'DeleteRows', 'DeleteColumns', 'DeleteTable',
+				   'DeleteRows', 'DeleteColumns', 'DeleteTable', 'EditCurrentRegion',
 				   'MergeCells', 'SetOptimalColumnWidth', 'SetOptimalRowHeight',
 				   'UpdateCurIndex','RemoveTableOf',
 				   'ReplyComment', 'DeleteComment', 'DeleteAuthor', 'DeleteAllNotes',
--- a/browser/src/unocommands.js
+++ b/browser/src/unocommands.js
@@ -117,6 +117,7 @@ var unoCommandsArray = {
 	'DrawText':{global:{context:_('Insert Text Box'),menu:_('~Text Box'),},},
 	'DuplicatePage':{presentation:{menu:_('Duplicate Page'),},},
 	'DuplicateSlide':{presentation:{menu:_('Duplicate ~Slide'),},},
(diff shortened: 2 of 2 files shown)


browser/src/control/Control.ContextMenu.js
```diff
@@ -55,7 +55,7 @@ L.Control.ContextMenu = L.Control.extend({
 			text: ['TableInsertMenu',
 				   'InsertRowsBefore', 'InsertRowsAfter', 'InsertColumnsBefore', 'InsertColumnsAfter',
 				   'TableDeleteMenu', 'SetObjectToBackground', 'SetObjectToForeground',
-				   'DeleteRows', 'DeleteColumns', 'DeleteTable',
+				   'DeleteRows', 'DeleteColumns', 'DeleteTable', 'EditCurrentRegion',
```
vmiklos, line 58: Once an uno command is allowed in this list, we should check if the dialog behaves correctly in the context of Online. This dialog has a frame called "Link", which allows the section to take content from some other file's section instead of providing own content. I would say this is confusing at best in the context of Online, since the file picker will just show you files in the chroot. The simplest would be to hide the entire Link frame in the Online case. Could you please do that?

That'll be again a core.git change: sw/source/ui/dialog/uiregionsw.cxx:393 has an example on how to hide some widgets in the "web" (Writer/Web, HTML edit mode), you could do something similar:

- check for comphelper::LibreOfficeKit::isActive() to see if this is Online or desktop
- rename frame2 to something like linkframe in the .ui file
- extend the ctor to have a reference to the link frame (probably you can use weld_frame() there)
- hide that entire frame in the LOK case

Could you please do this? Fine to do this as a separate core.git commit. Once that's done, this can also go in. Thanks. :-)

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9426/console

unit-bad-doc-load failed, which looks unrelated, resume.

browser/src/control/Control.ContextMenu.js
vmiklos: Thanks. Let's get to that once https://gerrit.libreoffice.org/c/core/+/161828 is in.

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9449/console

writer/invalidations_spec.js failed, looks unrelated, retry.

meven:
With dependencies in core merged:
https://gerrit.libreoffice.org/c/core/+/161828
https://gerrit.libreoffice.org/c/core/+/161878

Should be ready to land.

A question since we are adding a small feature to COOL, does it warrant a documentation update?

vmiklos (approved):
> Should be ready to land.

Indeed.

> A question since we are adding a small feature to COOL, does it warrant a documentation update?

The Help button already leads to https://help.collaboraoffice.com/23.05/en-US/text/swriter/01/04020100.html?System=UNIX&DbPAR=WRITER&HID=modules/swriter/ui/editsectiondialog/curname#bm_id8467970, which reuses help from the desktop, it's good enough, no need to document this further.

PR #7991, opened 2024-01-09
Fix failing test multiuser/writer/invalidations_spec.js

neilguertincollabora:
Change-Id: If34a07e6a70f2c4293e8a2eb9d5e3d5fd14258ac

* Resolves: # 
* Target version: master 

### Summary
Test was failing because second user wasn't completely reloaded.

### Checklist


gokaysatir (approved):
Thanks for the fix :)

PR #7992, opened 2024-01-09
highlight row/col header when editing

JaumePujantell:
The header highlighting was tied to the position of the cell cursor
so when editing a cell the highlights disapeared with the cell cursor.

This also checks the previous cursor position if there isn't a current one.

Change-Id: I93bf38b42cbaf091511c05e38bfab331e9532f4a


vmiklos (approved):
Looks OK to me. Things I considered:

- desktop Calc does this highlighting
- indeed highlighting was missing when cell editing
- indeed this commit fixes the problem

Did you consider if it's possible to cover this fix with a cypress test? But let's not block on that.

Issue #7403, opened 2023-10-10
Resume document message ignore click in message


hfiguiere:
### Describe the Bug

Resume document message ignore click in message

### Steps to Reproduce

1. Open a document
2. Let it open for while. switch away from the tab.
3. Reactivate the tab with the document
5. Click **in the message**

### Expected Behavior

It resumes.

### Actual Behavior

It does nothing.

### Screenshots

The message:

![image](https://github.com/CollaboraOnline/online/assets/114441/e571589c-0669-4163-93c3-511a2f13adf0)

### Desktop

(Please complete the following information)
 - **OS:** Linux
 - **Browser:** Firefox
 - **Version:** stable

### Smartphone

n/a

### Additional Context

Clicking elsewhere (outside the message) works.

Ezinnem:
Clicking on the document should make the document become active and ready for use.

@hfiguiere, if this issue is still reproducible for you, please can you share the Collabora Online you are using

hfiguiere:
This is on share.collabora.com

hfiguiere:
> Clicking on the document should make the document become active and ready for use.

But clicking in the message overlay doesn't. It's a natural user behaviour to click on the message that say "click to continue".

hfiguiere:
This boils down to have the modalpopup stop propagation of the click event while the overlay that is below it is the one that will dismiss it.

I think that an alert like that without a button should just let the click go through.

PR #7993, opened 2024-01-09
Click on a resume message to dismiss it

hfiguiere:
See https://github.com/CollaboraOnline/online/issues/7403

Change-Id: I20aa320a6efbbf8ce19eda34c3d5278a601e15fc

* Resolves: #7403 
* Target version: master 

### Summary

When you get the idle message, click on the message like you'd click in the document to dismiss. It feels more natural.

To trigger the message, in the JS Console, in the COOL frame, type:
```javascript
app.idleHandler._dim()
```

### TODO

- [ ] ...

### Checklist


hfiguiere:
This is the least invasive change. We could beautify this by using something else than `showInfoModal` for that specific case (idle), which is the only one that that do call it with one single parameter.

PR #7995, opened 2024-01-10
Make drawing area image render delay solution to specific

GulsahKose:
Line width dialog is affected from image render delay but spellcheck dialog is not. So make the solution specific to line width dialog.

Change-Id: I9456c347f0fa9da932e4df61f3324bb39544c06d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Thanks!

eszkadev (changes requested):
I don't like hardcoding special cases based on some id.
Is it possible to do it differently?

browser/src/control/jsdialog/Widget.DrawingArea.js
```diff
@@ -43,8 +43,9 @@ function _drawingAreaControl (parentContainer, data, builder) {
 	image.ondragstart = function() { return false; };
 	builder.map.uiManager.enableTooltip(image);
 
-	// If the image render is delayed, use width and height of the data
-	if (image.width == 0 && image.height == 0) {
```
eszkadev, line 47: I don't understand why it doesn't work in line width case?
Maybe we should use OR not AND in the condition?

eszkadev: It would be good to put more details into commit message, what exactly happens here

GulsahKose: That is a regression of https://github.com/CollaboraOnline/online/pull/7744

The issue is described here: https://github.com/CollaboraOnline/online/pull/7744

Actually it works on line width case but doesn't work on spellcheck dialog. Both dialogs are jsdialog have  drawing areas. Both image in the dialog delays at first time. Width and height is 0 at first time. And for the problematic spellcheck dialog data.image width and height is huge. So we see the following result:

Screenshot from 2024-01-10 15-10-11

Difference is line width dialog is relocating after then created. But spell check dialog is fixed we don't change the spellcheck dialog's position. Most probably that is the difference between two dialog. And couln't find a better way than id to differenciate them

eszkadev: Ok. So please always reference in the commit message previous commit which caused regression if you know about that. It's much easier to review and later if someone will have to work on that code again.

I see that when I sometimes click in spellchecking dialog, I got new jsdialog update and it is correctly shown.
Sometimes it adds width and height, sometimes not. Would be good to know why.

pedropintosilva:
We (Szymon, Gulsah and I) agreed that for now and if we test this a bit further and there is no bad consequences we should just merge this as it fixes something quite problematic

pedropintosilva:
I have tested

* Calc: sidebar (hunting for widgets with previews); dialogs (format cells and its tabs with previews) and everything seems ok
* On Writer I also didn't seem bad resizing **expect** for this glitch that it's only reproducible with this commit: 
| Without this commit| Local Master with this commit on top |
| ---------------------- |  ------------------------------------------------------ |
| https://github.com/CollaboraOnline/online/assets/65948705/21c76c6f-c90d-46cb-8665-2619ef2a8c3d | https://github.com/CollaboraOnline/online/assets/65948705/af776242-5711-4da3-92c1-1bcd2796b694|

@GulsahKose can you see if you can reproduce it (dialog resizes quickly with this commit but it might not be related)
* On Impress I saw this:
| Without this commit| Local Master with this commit on top |
| ---------------------- |  ------------------------------------------------------ |
| Screenshot_20240112_170049 | Screenshot_20240112_165840 |

Can you also test this ^ just to make sure ?

GulsahKose:
@pedropintosilva
I couldn't reproduce with b282e64cb62012e236470159effd914ac00505b9 with my commit
Screenshot from 2024-01-15 10-27-38

GulsahKose:
I also tested my patch with other dialogs and i found some nonconsistent bug on bullets part of writer sidebar:
Screenshot from 2024-01-15 11-13-48

@eszkadev @pedropintosilva container id is "valueset" for 3 bullet item. When i add the id in condition seems better i think. Should i include it into condition?

eszkadev (approved):
Great! code looks good :)

But one question: what about updates?
Update message doesn't have JSON with "parent"...   does it work when typing into the spellchecker?

I thought we need to check real DOM elements, `perent = parentContainer` then `parent = parentContainer.parentNode`, etc...

eszkadev:
>  does it work when typing into the spellchecker?
tested, and for me it is ok, so I think let's get this in as it fixes important issue

PR #8002, opened 2024-01-10
Rewinding and rereading proc/smaps_rollup doubles the values on some …

caolanm:
…kernels

So this only affects the case where we reuse the fd from opening smaps_rollup

Bug seen in 4.15.0 and not in 6.5.10, suspected to be fixed by: https://github.com/torvalds/linux/commit/258f669e7e88c18edbc23fe5ce00a476b924551f included in >= v4.19

Test for this in coolwsd and set a flag if it is broken

Change-Id: I0a4aca77b9d9201e4f70172340296e5eb5460229

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -2997,6 +2997,24 @@ void COOLWSD::innerInitialize(Application& self)
     StorageBase::initialize();
 
 #if !MOBILEAPP
+    // Check for smaps_rollup bug where rewinding and rereading gives
+    // bogus doubled results
+    if (FILE* fp = fopen("/proc/self/smaps_rollup", "r"))
+    {
+        std::size_t memoryDirty1 = Util::getPssAndDirtyFromSMaps(fp).second;
+        (void)Util::getPssAndDirtyFromSMaps(fp); // interleave another rewind+read to margin
+        std::size_t memoryDirty2 = Util::getPssAndDirtyFromSMaps(fp).second;
+        LOG_TRC("Comparing smaps_rollup read and rewind+read: " << memoryDirty1 << " vs " << memoryDirty2);
+        if (memoryDirty2 >= memoryDirty1 * 2)
+        {
+            // Believed to be fixed in >= v4.19, bug seen in 4.15.0 and not in 6.5.10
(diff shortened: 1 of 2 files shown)


wsd/COOLWSD.cpp
```diff
@@ -2997,6 +2997,23 @@ void COOLWSD::innerInitialize(Application& self)
...
 
 #if !MOBILEAPP
+    // Check for smaps_rollup bug there rewinding and rereading gives
+    // bogus doubled results
+    if (FILE* fp = fopen("/proc/self/smaps_rollup", "r"))
+    {
+        std::size_t nMemoryDirty1 = Util::getPssAndDirtyFromSMaps(fp).second;
+        std::size_t nMemoryDirty2 = Util::getPssAndDirtyFromSMaps(fp).second;
+        LOG_TRC("Comparing smaps_rollup read and rewind+read: " << nMemoryDirty1 << " vs " << nMemoryDirty2);
+        if (nMemoryDirty2 >= nMemoryDirty1 * 2)
```
mmeeks, line 3007: getPssAndDirty is unlikely to allocate too many pages (I guess) - and I expect it will work but I wonder how glibc and FILE * stuff buffering/allocation works etc. the >= *2 could potentially fail perhaps; would prefer a bit more slack myself for glibc to fool around allocating more space ;-) how about we do an extra defensive getPssAndDirtyFromSMaps() in between  nDirty1 and nDirty2 - and leave the >= 2 there ? =) would be more foolproof I think (?) =)

OH - and also, from a naming perspective - we don't hungarian mangle names in online be good to javaName them to stop the contagion spreading =)

mmeeks (approved):
Looks great - fine as is, but a tweak would make it perfect ? =)

mmeeks (approved):
Great - thanks =)

caolanm:
Funnily enough my  inclination was the other direction, to disable if the second call gave back results ~80% or so larger than the first, but checking that 3 calls doesn't double the values sounds good to me.

mmeeks:
Heh - pondered that; but <insert rounding problems - and more over-thinking> =) Thanks for getting it in.

Issue #7951, opened 2024-01-05
perf: calc avoid full invalidation on new user join


mmeeks:
Turn on 'invalidation' debugging in the help->triple click overlay.
Join the document from another view.
Notice the huge whole-document invalidation.

That shouldn't be there =)

Also - we should have a unit test for this that tests all components to ensure this doesn't happen / regression =)

caolanm:
First part merged as https://gerrit.libreoffice.org/c/core/+/161757

Second part as https://gerrit.libreoffice.org/c/core/+/161791 In writer the ViewOptions are in the ViewShell and are copied when a new ViewShell is created from another so the dark/light-mode and doc color are the same, in calc the ViewOptions exist in both the ViewShell and Document and a new ViewShell copies from the document not the old ViewShell. Setting the ViewOptions of a ViewShell in calc doesn't affect new ViewShells created from the first and the mismatch in ViewOptions triggers an invalidation. Disentangled the existing calc ViewOptions from the relatively new LoKIT options and move them into the ViewShell and out of the problematic ViewOptions.

There still remains a third invalidation on join however, lcl_ExtendTiledDimension invalidates two rects in a row and they are combined into a single rect which invalidates the whole typical initial viewbox. Which also suggests if user 1 is at some position in the doc outside the initial set of row/cols and then user 2 joins and scrolls to that position, outside their initial range, that user 1 would get an invalidation as user 2 enters that region. Some more debugging to do.

mmeeks:
Thanks Caolan ! =) Looks good - I suspect we're over-doing coalescing calc invalidations into huge rectangles IIRC we saw this with someone editing in A1, and someone else in S50 rather often we get large rectangular invalidations - but this is probably some other overly-optimistic merging location I guess.

caolanm:
https://gerrit.libreoffice.org/c/core/+/161791
and
https://gerrit.libreoffice.org/c/core/+/162412

PR #8004, opened 2024-01-10
cool#7951 add cypress test

caolanm:
add regression test for:
https://github.com/CollaboraOnline/online/issues/7951
to ensure no invalidations first user in a calc doc when second user joins

confirmed test fails with reverted core commit of:

commit 754cd4309b5b055a05b57f5d25ed128c439f460b
Date:   Mon Jan 8 15:04:07 2024 +0000

    Related: cool#7951 don't invalidate when creating a new view

Change-Id: I4e2bf09f8fbe36919a9826e8cf29cc5250520dc6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


cypress_test/integration_tests/multiuser/calc/invalidations_spec.js
```diff
@@ -0,0 +1,52 @@
...
+		helper.typeIntoDocument('X');
+		cy.cGet('#InsertMode').should('have.text', 'Insert');
+		helper.typeIntoDocument('{enter}');
+		cy.cGet('#InsertMode').should('have.text', '\u00a0Insert mode: inactive\u00a0');
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'A2');
+		helper.typeIntoDocument('{uparrow}');
+		// wait until round trip of cell address
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'A1');
+
+		cy.cGet('.empty-deltas').then(($before) => {
```
vmiklos, line 32: Why the dollar sign here, is that some kind of convention? It seems $ has no special meeting at a JS language level.

vmiklos (approved):
I don't see anything obviously problematic with this. Seeing this is mostly the same as the existing Writer equivalent, looks OK.

Inline comment is more about me trying to understand this JS code rather than about the code itself.

cypress_test/integration_tests/multiuser/calc/invalidations_spec.js
caolanm: This one is $before because the writer one I wrote in the past that I based this one off used $before, and that one was $before because I followed https://docs.cypress.io/guides/core-concepts/variables-and-aliases which used $btn as an example :-)

I think it is just convention and has no actual meaning.

vmiklos: I see, thanks for the cypress reference.

PR #8005, opened 2024-01-11
kit: Only log smaps_rollup open failure when it actually failed

thebearon:
...not when it wasn't even attempted.

Change-Id: Ide56c46b1a2d99c3639aace430cec099c6afcbf4

* Target version: master 

### Checklist


mmeeks (approved):
Fair enough - Kit warnings happen a lot more than coolwsd warnings =)
Thanks!

PR #8008, opened 2024-01-11
Set child popup position for autofilter

Darshan-upadhyay1110:
- Child popoup like filter by color or filter by condition has fixed potion on screen
- added calculation child popup in autofilter
Change-Id: I8adc33e782bd7a238231ecb80e533a11bc71e5b4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Darshan-upadhyay1110:
Please do not merge this yet. These is my final changes, but i think there is scope for some refactoring ( removing some duplication of code )

browser/src/control/Control.JSDialog.js
```diff
@@ -501,6 +501,13 @@ L.Control.JSDialog = L.Control.extend({
 		var offsetX = isSpreadsheetRTL ? 0 : app.sectionContainer.getSectionWithName(L.CSections.RowHeader.name).size[0];
 		var offsetY = app.sectionContainer.getSectionWithName(L.CSections.ColumnHeader.name).size[1];
 
+
+		var isDescendant = instance.lastFocusedElement.closest('table#menu.jsdialog.autofilter.ui-treeview');
```
eszkadev, line 505: doesn't that work only when focus in some specific place?

browser/src/control/Control.JSDialog.js
```diff
@@ -525,12 +532,36 @@ L.Control.JSDialog = L.Control.extend({
 		instance.posy = top + offsetY;
 
 		var width = instance.form.getBoundingClientRect().width;
-		var height = instance.form.getBoundingClientRect().height;
+		var autoFilterBottom = instance.posy + document.querySelector('.main-nav').clientHeight + document.getElementById('toolbar-wrapper').clientHeight + instance.form.getBoundingClientRect().height;
```
eszkadev, line 535: Can't we just use `instance.form.getBoundingClientRect().bottom`  ? better to let browser calculate that, than do it on your own. Especially that now it introduces dependency on "main-nav" and "toolbar-wrapper". We will want to change classic mode in the future - it will cause problems.

browser/src/control/Control.JSDialog.js
```diff
@@ -525,12 +532,36 @@ L.Control.JSDialog = L.Control.extend({
...
+		this.updatePosition(instance.container, instance.posx, instance.posy);
+	},
+
+	calculateSubmenuAutoFilterPosition: function(instance) {
+		// Get the position of the last focused element
+		var parentAutofilterEl = instance.lastFocusedElement.parentElement.getBoundingClientRect();
+		instance.posx = parentAutofilterEl.right + 20;
+		instance.posy = parentAutofilterEl.top - document.getElementById('toolbar-wrapper').offsetHeight;
+
+		var toolbarHeight= document.getElementById('spreadsheet-toolbar').offsetHeight;
```
eszkadev, line 554: can we just rely on parent popup position?
to not depend on any other toolbar / element/ etc ?

browser/src/control/Control.JSDialog.js
Darshan-upadhyay1110: Yes, this will be our key point that the current opened popup will surly the child autofilter popup.

Darshan-upadhyay1110: and also it will be obvious that the current popup is child popup only and only is the last focused this is at parent .

so i think it will be a good way for calculate child popup.

@eszkadev is there any possibility or way that the last focused thing will not be parent autofilter popup in case of opening a child autofilter popup ? ( i think it will not possible )

browser/src/control/Control.JSDialog.js
Darshan-upadhyay1110: i have studied the JSON structure of child and parent popup they seem different but i don't find any specific property from JSON which can distingué that this JSON is for child popup.

browser/src/control/Control.JSDialog.js
Darshan-upadhyay1110: 100 % agreed. I have added too much complicity. Surly update this thanks :)

browser/src/control/Control.JSDialog.js
```diff
@@ -525,12 +532,33 @@ L.Control.JSDialog = L.Control.extend({
...
+	},
+
+	calculateSubmenuAutoFilterPosition: function(instance) {
+		// Get the position of the last focused element
+		var parentAutofilterEl = instance.lastFocusedElement.parentElement.getBoundingClientRect();
+		instance.posx = parentAutofilterEl.right + 20;
+		instance.posy = parentAutofilterEl.top - document.getElementById('toolbar-wrapper').offsetHeight;
+
+		// set posx of instance (submenufilter) based on window width 
+		if (instance.posx + instance.content.clientWidth > window.innerWidth)
```
eszkadev, line 555: repeated `instance.content.clientWidth`, please use variable `width`. Also will be shorter, more easy to read.

browser/src/control/Control.JSDialog.js
eszkadev: things like: `document.querySelector('sss')` can fail when we remove some element in the future. It will cause error and half-baked dialog/popup. Please first get query result into variable, check if exists and then use. If not exist - handle somehow that case.

browser/src/control/Control.JSDialog.js
```diff
@@ -525,12 +532,33 @@ L.Control.JSDialog = L.Control.extend({
...
+			instance.posy = instance.posy - (autoFilterBottom - toolbarTop);
+
+		this.updatePosition(instance.container, instance.posx, instance.posy);
+	},
+
+	calculateSubmenuAutoFilterPosition: function(instance) {
+		// Get the position of the last focused element
+		var parentAutofilterEl = instance.lastFocusedElement.parentElement.getBoundingClientRect();
+		instance.posx = parentAutofilterEl.right + 20;
+		instance.posy = parentAutofilterEl.top - document.getElementById('toolbar-wrapper').offsetHeight;
```
eszkadev, line 552: `document.querySelector` can fail

browser/src/control/Control.JSDialog.js
eszkadev: But can we avoid using `document.querySelector` at all?
You are interested in map's top edge. Can't you access it using some variables and it's members? it's more safe to use and reliable (from: `this.map.xxxx`)

browser/src/control/Control.JSDialog.js
eszkadev: I think it might be not reliable. (we can always break this in the future).
Maybe try to guess that by looking at JSON structure?
Which widgets are visible, which not for example.

browser/src/control/Control.JSDialog.js
eszkadev: why we subtract `toolbar-wrapper`'s height?

browser/src/control/Control.JSDialog.js
```diff
@@ -525,9 +532,43 @@ L.Control.JSDialog = L.Control.extend({
 		instance.posy = top + offsetY;
 
 		var width = instance.form.getBoundingClientRect().width;
-		var height = instance.form.getBoundingClientRect().height;
+		var canvasEl = document.getElementById('document-canvas').getBoundingClientRect();
```
eszkadev, line 535: I prefer to get this directly:
`map._docLayer._canvas`

See: https://github.com/CollaboraOnline/online/blob/33634effc6dcca523feee7072e670aae2e4f7db2/browser/src/layer/tile/CanvasTileLayer.js#L990

browser/src/control/Control.JSDialog.js
```diff
@@ -525,9 +532,43 @@ L.Control.JSDialog = L.Control.extend({
...
+			return isSubmenuPopup ? (isSubmenuPopup.id === 'menu' && isSubmenuPopup.type === 'treelistbox') : false;
+		}
+		return false;
+	},
+
+	calculateSubmenuAutoFilterPosition: function(instance, parentAutofilter) {
+		// Get the position of the last focused element
+		var parentAutofilter = parentAutofilter.getBoundingClientRect();
+		instance.posx = parentAutofilter.right + 20;
+		instance.posy = parentAutofilter.top - document.getElementById('document-canvas').getBoundingClientRect().top;
```
eszkadev, line 563: `map._docLayer._canvas`

browser/src/control/Control.JSDialog.js
```diff
@@ -525,9 +532,43 @@ L.Control.JSDialog = L.Control.extend({
...
+		var rootChild = instance.children[0];
+		if (rootChild) {
+			var isSubmenuPopup = rootChild.children[0];
+			return isSubmenuPopup ? (isSubmenuPopup.id === 'menu' && isSubmenuPopup.type === 'treelistbox') : false;
+		}
+		return false;
+	},
+
+	calculateSubmenuAutoFilterPosition: function(instance, parentAutofilter) {
+		// Get the position of the last focused element
```
eszkadev, line 560: old comment about last focused element

browser/src/control/Control.JSDialog.js
```diff
@@ -525,9 +532,43 @@ L.Control.JSDialog = L.Control.extend({
...
+			instance.posy = instance.posy - (autoFilterBottom - canvasBottom);
+
+		this.updatePosition(instance.container, instance.posx, instance.posy);
+	},
+
+	isChildAutoFilter: function(instance) {
+		// JSON structure suggest that if children array's first element has id='menu' and widgetType = 'treelistbox' then it will definatly a child autofilter popup
+		var rootChild = instance.children[0];
+		if (rootChild) {
+			var isSubmenuPopup = rootChild.children[0];
```
eszkadev, line 553: `isSubmenuPopup` suggesst boolean value, maybe `firstWidget` ?

eszkadev:
Thanks, now works well and code looks ok.
But do we want 20px gap between parent and child menu? I think usually submenus are glued to the parent.
gap

Tested also RTL sheet with RTL UI (lang=ar), it seems to open submenu on the right side, not left. Also sometimes outside screen. We should adjust conditions to prevent that too. But I think we can do it in the followup.
rtl_ar

eszkadev (approved):

pedropintosilva:
> Thanks, now works well and code looks ok. But do we want 20px gap between parent and child menu? I think usually submenus are glued to the parent. 

@eszkadev : No gap please. Even if we are not using dropdown widget it should look like that at least

Darshan-upadhyay1110:
- removed the extra 20px ( Gap )
image

- handle the rtl calse. now user will see child popup on the left of parent popup if we have rtl mode enable
image

CC: @eszkadev @pedropintosilva

PR #8012, opened 2024-01-11
IdleHandler: sedn userinactive to server

eszkadev:
This is followup for commit add3b76eb5ad576332c8a7150c2ad07881ec224f
IdleHandler: bring back idle timeout for view

Which restored dimmed view, but I forogot to bring back sending userinactive message in some cases


quikee (approved):
Look OK

PR #8014, opened 2024-01-12
add missing icons (used by mobile UI)

timar:
Change-Id: I82b1fc1c59452c092f7ff250e3919496458fff46

### TODO

- insertsheetbefore, insertsheetafter, movesheetleft and movesheetright are custom commands in COOL therefore we did not have 100% matching icon from core. Maybe we need to draw new icons for them.


pedropintosilva (changes requested):
.

pedropintosilva (approved):
Thanks

pedropintosilva:
one more thing I actually couldn't find where `lc_movesheetleft.svg` and `lc_movesheetright.svg` is used on mobile because long pressing the sheet triggers a list without icons (even if the assets exist)

timar:
> one more thing I actually couldn't find where `lc_movesheetleft.svg` and `lc_movesheetright.svg` is used on mobile because long pressing the sheet triggers a list without icons (even if the assets exist)

It was a blind fix from my side. I'll check and Android snapshot after merge.

pedropintosilva:
> so this PR was not a complete failure. :)

far from it, Thanks! :)

PR #8007, opened 2024-01-11
fix error TS2694: Namespace 'cool' has no exported member 'TilesSecti…

caolanm:
…on'.

in make check-local

Change-Id: I9d6f130b7943165b3937cbbd70efcf0690c94b88

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

- [ ] Code is properly formatted
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #8018, opened 2024-01-12
Resolve TS build error

neilguertincollabora:
Change-Id: Ic25805de741d6d71153211de68fb0ccc88e62d40

### Summary
Follow up from https://github.com/CollaboraOnline/online/pull/8007/files

The error is TS2569: Type 'IterableIterator<any>' is not an array type or a string type. TS is complaining about the use of an iterator in the for loop where it expects an array or string.

The error message and everyone on stackoverflow recommend setting the compiler option --downlevelIteration. But wrapping in Array.from() is simpler.

Compiles on my machine with `./node_modules/typescript/bin/tsc mocha_tests/CanvasSectionContainer.test.ts --outfile mocha_tests/CanvasSectionContainer.test.js --module none --lib dom,es2016 --target ES5` when before it didn't. But I didn't run any other tests.

### Checklist

### DIFF
--- a/browser/src/layer/tile/TilesSection.ts
+++ b/browser/src/layer/tile/TilesSection.ts
@@ -131,7 +131,7 @@ export class TilesSection extends CanvasSectionObject {
 		var ctx = this.sectionProperties.tsManager._paintContext();
 
 		var bounds: cool.Bounds;
-		for (const coords of tileSubset.values()) {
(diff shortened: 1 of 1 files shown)


caolanm (approved):
The fix in https://github.com/CollaboraOnline/online/pull/8019 to address why this was *not* failing for me locally.

PR #8019, opened 2024-01-12
tsc can report an error but create js output, so 2nd make's 'succeed'

caolanm:
so on failure remove the output so a second attempt doesn't appear to succeed

Change-Id: Ibc1f7d46c2946cdc6a4dbcd6181872fb12e19fb6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
I expect this to fail until https://github.com/CollaboraOnline/online/pull/8018 gets merged

caolanm:
I expect this to fail until https://github.com/CollaboraOnline/online/pull/8018 gets merged

caolanm:
failed as hoped for, so now rebased on top of the fix and hopefully we pass cleanly

caolanm:
which it now does, so if something triggers a similar case again a rule like this would make it consistently fail

Minion3665 (approved):

PR #8021, opened 2024-01-12
convert-to: be more tolerant of unusual filenames.

mmeeks:
But re-write them anyway to be more sensible.

Change-Id: Ie146f4f84b539ab7e826a1c1b947497acde7e384

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -608,6 +608,34 @@ namespace Util
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


vmiklos (approved):
Looks OK to me, allowing / in filenames was indeed bad in the past.

PR #8022, opened 2024-01-14
Make minus/plus centered in group boxes

OFFTKP:
Centered the minus/plus in group boxes across zoom levels.


pedropintosilva:
re-triggering CI

gokaysatir:
It seems that some duplicates lines are found.

OFFTKP:
On it :)

mmeeks:
The expand/collapse is now beautiful:
image

There is some more polish we need before this goes in though I think; this change is not great:
image
to this:
image
for the grouping boxes.

Also dark mode is broken for grouping - but that can be a separate fix of course.

image

I wonder if for those grouped 1,2,3 etc. inside squares we should not try to use a browser font but a pre-rendered tiny image (?) not clear really.

pedropintosilva (approved):
This should be merged as it introduces improvements on how we render the minus and plus :

left: before ; right: after applying the commit

gokaysatir:
I'll check this soon.

gokaysatir (changes requested):
This really improves the view for desktop view. I think we still need to check the mobile view.

browser/src/control/Control.GroupBase.ts
```diff
@@ -54,8 +54,8 @@ export abstract class GroupBase extends app.definitions.canvasSectionObject {
 		this._groups = null;
 
 		// group control styles
-		this._groupHeadSize = Math.round(12 * app.dpiScale);
-		this._levelSpacing = app.roundedDpiScale;
+		this._groupHeadSize = 12;
```
eszkadev, line 57: this looks like problem when zooming in/out :)
we should use DPI probably

eszkadev:
@gokaysatir do you think you will find time at some point to finish this so we can close PR?

eszkadev:
@banobepascal could you check at some point if it is still valid issue? and we need that patch?

banobepascal:
> @banobepascal could you check at some point if it is still valid issue? and we need that patch?

I checked, and this patch doesn't improve scaling at different zoom levels. The current implementation already handles centering and zooming quite well  even up to `175%` browser zoom  so the patch doesn't seem necessary.

### CURRENT UI 
COOLWSD version: 25.04.3.1(git hash: e4ec936bd4)
LOKit version: Collabora Office 25.04.3.1(git hash: 531aea4c7d)

Browser Zoom: 175% 
Document Zoom: 150%

Screenshot from 2025-06-19 00-57-47

Browser Zoom: 150%
Document Zoom: 100%

Screenshot from 2025-06-19 01-01-24


### WITH THIS PR
Browser Zoom: 175%
Document Zoom: 150%

Screenshot from 2025-06-19 01-09-05


Browser  Zoom: 150%
Document Zoom: 100%
Screenshot from 2025-06-19 01-07-01

timar:
As @banobepascal said, this patch does not seem necessary.

mmeeks:
@banobepascal the existing '-' is ugly and non-centred in your shots, we want to go from something like left to right (though the size is different):

image

The minus should not be anti-aliased, the box around it should be a hair-line, and there should be the same pixel gap to the border on each side - so it doesn't look 'wrong' and lop-sided. I agree it is minor but ... why would we choose not to get that right ?

banobepascal:
> @banobepascal the existing '-' is ugly and non-centred in your shots, we want to go from something like left to right (though the size is different):

> The minus should not be anti-aliased, the box around it should be a hair-line, and there should be the same pixel gap to the border on each side - so it doesn't look 'wrong' and lop-sided. I agree it is minor but ... why would we choose not to get that right ?

Thanks for pointing that out. There was a slight imbalance in the pixel spacing of the " `−` ", which is easy to miss on smaller screens but becomes noticeable at higher zoom levels or on high DPI displays.

I’ve adjusted the drawing so both the  " `−` " and " `+` " are properly centered with even spacing on all sides, and also vertically aligned the level numbers within the group boxes.

PR: https://github.com/CollaboraOnline/online/pull/15177

PR #8027, opened 2024-01-15
Private/gokay/hyperlinkinfo readonly calc

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1886,6 +1886,7 @@ L.CanvasTileLayer = L.Layer.extend({
 		else if (textMsg.startsWith('readonlyhyperlinkclicked: ')) {
 			var jsonString = textMsg.replace('readonlyhyperlinkclicked: ', '');
 			var json = JSON.parse(jsonString);
+
 			var position = this._twipsToLatLng(json.position);
 			this._showURLPopUp(position, json.link);
 		}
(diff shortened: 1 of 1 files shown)


eszkadev:
Please explain in the commit message why we need different units. Maybe add link to core patch. Will be easier to debug later.

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1886,6 +1886,12 @@ L.CanvasTileLayer = L.Layer.extend({
 		else if (textMsg.startsWith('readonlyhyperlinkclicked: ')) {
 			var jsonString = textMsg.replace('readonlyhyperlinkclicked: ', '');
 			var json = JSON.parse(jsonString);
+
+			if (this._docType === 'spreadsheet') {
+				json.position.x *= app.pixelsToTwips;
+				json.position.y *= app.pixelsToTwips;
```
eszkadev, line 1892: core sends position in pixels? it seems to be incorrect, what if user has different DPI in the browser than server's view when generating message?

gokaysatir:
@eszkadev hello, thank you for checking.

Core side doesn't send pixel coordinates. Indeed the clicked coordinates are sent back by COOLWSD for hyperlink popup position.

This is the core side PR's link:
https://gerrit.libreoffice.org/c/core/+/162111

I needed to convert the twips to pixels, it works also on mobile view, which has a different DPI ratio.

We could do this on the core side too, i preferred to solve this on Online side, since the twips to pixels multiplier is ready there. It was quicker to solve this way.

gokaysatir:
@eszkadev so the core side was also a quick one :) If you can review below one when you have time, that'd be great. I checked with mobile, scrolled, zoomed views etc. It should be safe.

Thank you.

https://gerrit.libreoffice.org/c/core/+/162180

eszkadev (approved):
This is perfect now :)

eszkadev:
Tested on far distance cell, with different zoom levels, `CC1000` and it works correctly showing popup with hyperlink. Thanks

Issue #8023, opened 2024-01-15
Clipboard: Implement permissions for pasting in Chrome-based browsers


vmiklos:
### Is your feature request related to a problem?

Pressing ctrl-v pastes from the clipboard, but pressing the paste button, we just show a popup to press ctrl-v.

### Describe the solution you'd like

Chrome has some new clipboard API that allows asking for the clipboard permission, we should conditionally start using that. See <https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read>.

venky-17:
hey can u pls assign me this issue??

vmiklos:
> hey can u pls assign me this issue??

This issue is already assigned, please pick an other issue which is not yet assigned. Thanks.

vmiklos:
https://github.com/CollaboraOnline/online/pull/8061 switches to the improved new markup, described above.

vmiklos:
Next step is to actually use the new API in Chrome. To test the new functionality, have an `--enable-experimantal` build, and then the baseline gives you an annoying popup when you click the paste button on the toolbar, while the new code asks for this permission once, and otherwise it more or less "just works".

vmiklos:
https://github.com/CollaboraOnline/online/pull/8091 removes the old markup.

vmiklos:
#8106 adds tests, let's close this.

PR #8041, opened 2024-01-17
Clipboard: add new-style metadata that survives browser sanitize

vmiklos:
- this is a first step towards fixing https://github.com/CollaboraOnline/online/issues/8023
- this doesn't switch to the new API yet, but adds a new markup to recognize if the clipboard has our own HTML or something else
- the new markup is better, because <meta> tags are removed by the browser when using the new navigator.clipboard API, and this one is not
- this is meant to be a safe set of first commits, we already write the new markup, but we don't remove the old one yet, e.g. Android needs updating still
- on the other hand, this already provides enough markup so src/map/Clipboard.js can start experimenting with the new API, including the internal paste codepath

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -60,14 +60,18 @@ L.Clipboard = L.Class.extend({
 		document.onbeforepaste = beforeSelect;
 	},
 
-	// We can do a much better job when we fetch text/plain too.
-	stripHTML: function(html) {
-		var tmp = new DOMParser().parseFromString(html, 'text/html').body;
-		// attempt to cleanup unwanted elements
-		var styles = tmp.querySelectorAll('style');
+	// Attempt to cleanup unwanted elements
+	stripStyle: function(domNode) {
+		var styles = domNode.querySelectorAll('style');
 		for (var i = 0; i < styles.length; i++) {
 			styles[i].parentNode.removeChild(styles[i]);
 		}
+	},
+
+	// We can do a much better job when we fetch text/plain too.
+	stripHTML: function(html) {
+		var tmp = new DOMParser().parseFromString(html, 'text/html').body;
+		this.stripStyle(tmp);
 		return tmp.textContent.trim() || tmp.innerText.trim() || '';
 	},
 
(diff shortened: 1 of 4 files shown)


vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9640/consoleFull

but that suite has one known-unstable test and here the insert hyperlink one failed. Fixed now so that the content of <style> doesn't leak into the plain text selection result.

vmiklos:

This is meant to be a safe subset, we just change the marker format, no actual use of the new clipboard stuff yet. https://github.com/CollaboraOnline/online/issues/8023#issuecomment-1898616860 shows how the old vs new markup looks like.

browser/src/map/Clipboard.js
```diff
@@ -115,15 +119,15 @@ L.Clipboard = L.Class.extend({
 	_originWrapBody: function(body, isStub) {
 		var encodedOrigin = encodeURIComponent(this.getMetaURL());
 		var text =  '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">\n' +
-		            '<html>\n' +
+		            '<div id="meta-origin" data-coolorigin="' + encodedOrigin + '"><html>\n' +
```
caolanm, line 122: It isn't actually valid html though, right, so could stop working if the sanitizing gets tightened up?

reminds me of the bodge for similar with gtk3 by adding an extra mimetype to identify that the clipboard content was coming from "ourself": https://cgit.freedesktop.org/libreoffice/core/commit/?id=88f7aae022bedd61588424a11bbc033217ba4e43

caolanm (approved):

browser/src/map/Clipboard.js
vmiklos: Oh wow, I thought desktop is nice & simple, given we have a reliable way of getting notified of the ownership change & only browser is poor. Not. :-)

In any case, good point with the invalid HTML, I'll look at that. We had the markup in <head>, that is removed, so the first thought was to do this toplevel, which gives valid output currently, and I didn't tune this further. I'll check if it's possible to do this inside <body>, that would be very similar, but would meet both the sanitize and the validity requirements, I think.

PR #8048, opened 2024-01-18
fix: Impress: Dismiss snackbar on presentation window close

codewithvk:
Resolved an issue in Impress's 'Present in Window' feature where the snackbar would persist even after closing the presentation window or pressing the escape key. Now, the snackbar properly disappears as intended, ensuring a cleaner and more intuitive user interface experience.

Change-Id: Id491e39d88b433993b5ff83e10c3ca2b1eb8ab40

* Resolves: # 
* Target version: master 

### Summary
Screenshot from 2024-01-18 09-29-51

### DIFF
--- a/browser/src/map/handler/Map.SlideShow.js
+++ b/browser/src/map/handler/Map.SlideShow.js
@@ -150,13 +150,21 @@ L.Map.SlideShow = L.Handler.extend({
 			}
 
 			this._slideShowWindowProxy.focus();
-			this._slideShowWindowProxy.addEventListener('keydown', this._onCloseSlideWindow);
+			this._slideShowWindowProxy.addEventListener('keydown', this._onCloseSlideWindow.bind(this));
 
 			var slideShowWindow = this._slideShowWindowProxy;
 			this._map.uiManager.showSnackbar(_('Presenting in window'),
 				_('Close Presentation'),
 				function() {slideShowWindow.close();},
 				-1, false, true);
+
+			this._windowCloseInterval = setInterval(function() {
+				if (slideShowWindow.closed) {
+					clearInterval(this._windowCloseInterval);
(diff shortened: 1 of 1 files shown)


browser/src/map/handler/Map.SlideShow.js
```diff
@@ -150,7 +150,8 @@ L.Map.SlideShow = L.Handler.extend({
 			}
 
 			this._slideShowWindowProxy.focus();
-			this._slideShowWindowProxy.addEventListener('keydown', this._onCloseSlideWindow);
+			this._slideShowWindowProxy.addEventListener('keydown', this._onCloseSlideWindow.bind(this));
+			this._slideShowWindowProxy.addEventListener('beforeunload', this._closeSnackbar.bind(this));
```
codewithvk, line 154: Oddly, the `beforeunload` event is not working in Chrome.

codewithvk: Works fine for firefox ;)

codewithvk: I need your suggestion. Actually, when we close the slideshow window, we should use the `beforeunload` event to trigger an action when the window is closed. At that time, we can hide the snackbar. Unfortunately, `beforeunload` is not supported by a few older browsers, and or implementation varies across different browsers. In the previous commit code,
```js
this._slideShowWindowProxy.addEventListener('beforeunload', this._closeSnackbar.bind(this)); 
the _closeSnackbar: function(e) { this._map.uiManager.closeSnackbar(); }
```
, which works fine in Firefox but not in Chrome. I have implemented an approach that is not ideal, but it works fine. Could you please share your views or suggestions on this?

mmeeks:
Please back-port to the online branch cp-23.05 if suitable ?

PR #8052, opened 2024-01-18
Parallel Loading of Users' Requests

Ashod:
- wsd: use RequestVettingStation for async loading
- wsd: restructure RequestVettingStation interface 
- wsd: use a single shared_ptr instead of container for RequestVettingS… 
- wsd: extract ResourceAccessDetails from cool.html file request 
- wsd: make wopiInfo a member of RequestVettingStation 
- wsd: RequestDetails from document-load URI components 
- wsd: parallel load with cool.html 
- wsd: better logging in RequestVettingStation::handleRequest 
- wsd: split createDocBroker into createClientSession

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4138,7 +4138,6 @@ int COOLWSD::innerMain()
     std::string version, hash;
     Util::getVersionInfo(version, hash);
     LOG_INF("Coolwsd version details: " << version << " - " << hash << " - id " << Util::getProcessIdentifier() << " - on " << Util::getLinuxVersion());
-
 #endif
 
     initializeSSL();
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -147,6 +147,7 @@ using Poco::Net::PartHandler;
 
 #include <common/SigUtil.hpp>
 
+#include <RequestVettingStation.hpp>
 #include <ServerSocket.hpp>
 
 #if MOBILEAPP
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -62,6 +62,7 @@
 #include <chrono>
 #include <iostream>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <regex>
 #include <sstream>
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -491,9 +491,10 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request, http:
 #endif
 
 void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
-                                             const RequestDetails &requestDetails,
+                                             const RequestDetails& requestDetails,
                                              Poco::MemoryInputStream& message,
-                                             const std::shared_ptr<StreamSocket>& socket)
+                                             const std::shared_ptr<StreamSocket>& socket,
+                                             ResourceAccessDetails& accessDetails)
 {
     try
     {
@@ -511,9 +512,10 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
         {
             if (config.getBool("ssl.sts.enabled", false))
             {
-                const auto maxAge = config.getInt("ssl.sts.max_age", 31536000); // Default 1 year.
+                const auto maxAge =
+                    config.getInt("ssl.sts.max_age", 31536000); // Default 1 year.
                 response.add("Strict-Transport-Security",
-                             "max-age=" + std::to_string(maxAge) + "; includeSubDomains");
+                                "max-age=" + std::to_string(maxAge) + "; includeSubDomains");
             }
         }
 #endif
@@ -587,7 +589,7 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
             endPoint == "uno-localizations.json" ||
             endPoint == "uno-localizations-override.json")
         {
-            preprocessFile(request, requestDetails, message, socket);
+            accessDetails = preprocessFile(request, requestDetails, message, socket);
             return;
         }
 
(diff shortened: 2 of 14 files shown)


net/WebSocketHandler.hpp
```diff
@@ -252,8 +252,9 @@ class WebSocketHandler : public ProtocolHandlerInterface
             if (!_shuttingDown)
                 sendCloseFrame(statusCode, statusMessage);
             socket->closeConnection();
-            socket->getInBuffer().clear();
             socket->ignoreInput();
+            assert(socket->getInBuffer().empty() &&
```
mmeeks, line 256: Not convinced by this. A websocket can have lots of queued input, and yet we need to shut it down because one of the top messages is malformed eg. Would prefer the clear after the assert if we must have it - but it will cause grief I think.

wsd/FileServer.cpp
```diff
@@ -544,7 +543,7 @@ void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
 #if ENABLE_DEBUG
         if (Util::startsWith(relPath, std::string("/wopi/files"))) {
             handleWopiRequest(request, requestDetails, message, socket);
-            return;
+            return ResourceAccessDetails();
```
mmeeks, line 546: Not a huge fan of having this as a return value and all these un-related code-paths having this return.
Can we either split the function, or have an optional std::function callback that is usually null for the one path that matters, or pass a reference to an empty shared_ptr in for the path that matters or somesuch ? =)

wsd/COOLWSD.cpp
```diff
@@ -4280,7 +4281,34 @@ class ClientRequestDispatcher final : public SimpleSocketHandler
...
+                                       "Expected identical WOPISrc in the request as in cool.html");
+
+                        const std::string requestKey =
+                            RequestDetails::getRequestKey(details.wopiSrc(), details.accessToken());
+
+                        std::vector<std::string> options = {
+                            "access_token=" + details.accessToken(), "access_token_ttl=0"
+                        };
+
+                        const std::string documentLoadURI = RequestDetails::createDocumentLoadURI(
```
mmeeks, line 4300: Not a big fan of this string construction and re-parsing; is there a need for that? Can't we do this from the already broken down requestDetails and so on - since we turn this into another RequestDetails lower down ?

mmeeks (approved):
So - overall this looks good. Seems we do the async CheckFileInfo, but don't pre-download the file (or did I miss that ? =). I prolly need to re-read / read some more I guess.

I'm worried about:

disposition.setMove - we use this a fair bit, yet:

    // not the method you want.
    void setMove(MoveFunction moveFn)

concerns me - I'm hazy about its usage; can you check it's the right disposition where used vs. transfer etc.

I would like a clean, single pure code-movement commit that moves 

COOLWSD ClientRequestDispatcher - into its own file - COOLWSD.cpp is far too large: no other code-changes though - a pure code-move commit =)

Can you test copy/paste ?

Otherwise - don't want to block this getting in; looks like a good direction.

Thanks Ash! =)

mmeeks:
Ah - we should get the unit tests passing too I guess ;-)

net/WebSocketHandler.hpp
Ashod: `ignoreInput()` clears the buffer. Since I removed the redundant `socket->getInBuffer().clear();` (2 lines above), I put this assert to signal that removing the clear is safe and to defend against regressions.

Unless I missed your point.

wsd/FileServer.cpp
Ashod: Agreed. Reworked it.

wsd/COOLWSD.cpp
Ashod: Agreed. It was the simplest approach that really served as a POC. Should've passed `RequestDetails` directly. Will do in a follow up.

Ashod:
Thanks for the review and feedback!

> So - overall this looks good. Seems we do the async CheckFileInfo, but don't pre-download the file (or did I miss that ? =). I prolly need to re-read / read some more I guess.

Not yet! For that we need DocBroker to get created without a client WS, create the jail, spin up a Kit, fire the poll thread, and download, while at the same time not have a client WS (it captures it in the constructor). The tricky bit is error handling, as normally we would report the error to the client and close the socket and finally disbanding the DocBroker. But if we destroyed the DocBroker without the client WS, we would re-do all this work when the client WS is created (when the browser open the WS), which is wasteful. So there is a bit of reworking that needs to be done before downloading in parallel works.

Rest assured it is in the works.

> 
> I'm worried about:
> 
> disposition.setMove - we use this a fair bit, yet:
> 
> ```
> // not the method you want.
> void setMove(MoveFunction moveFn)
> ```
> concerns me - I'm hazy about its usage; can you check it's the right disposition where used vs. transfer etc.

Good point. Read that patch. A thorough review is in order. Will do.

> I would like a clean, single pure code-movement commit that moves
> 
> COOLWSD ClientRequestDispatcher - into its own file - COOLWSD.cpp is far too large: no other code-changes though - a pure code-move commit =)

Agreed. It has grown way too large and I wanted to move it. Perhaps a good idea to do it before this one lands.

> 
> Can you test copy/paste ?

Will do. I was relying on the unit-tests, but I think we need some manual testing too.

> 
> Otherwise - don't want to block this getting in; looks like a good direction.

I have broken this up into 
https://github.com/CollaboraOnline/online/pull/8179
and
https://github.com/CollaboraOnline/online/pull/8180

I'll rebase when the above two land. Meanwhile, will follow up on the feedback here.

> Thanks Ash! =)

Thank you!

net/WebSocketHandler.hpp
mmeeks: Ah - I missed that; this is fine :-) thanks.

wsd/FileServer.cpp
mmeeks: Didn't see that in this patch; but look forward to it =)

mmeeks:
> But if we destroyed the DocBroker without the client WS, we would re-do all this work when the client WS is created (when the browser open the WS), which is wasteful. So there is a bit of reworking that needs to be done before downloading in parallel works.

I don't think we need to worry about making an error case - which after all won't do a download anyway particularly optimal. We don't want to remember a previous error I think; just re-issue the request at the later point if necessary and it failed in the background and/or is not still downloading I think =)

Can we get some more of these re-factors, SPX cleanup etc. pushed separately too ? =) Thanks

wsd/FileServer.cpp
Ashod: Yes, was in the update that wasn't yet uploaded at the time. Now updated.

wsd/COOLWSD.cpp
Ashod: Done. Converted the static helper into a proper constructor and pass the created instance directly.

mmeeks:
Glad this got in - good stuff =) no I guess for pipelining the download too (?) =)

PR #8058, opened 2024-01-19
Make `make check-multi` act like `check-desktop`

neilguertincollabora:
And other changes to make the targets -desktop, -idle, -mobile, and -multi more similar

Change-Id: I8866f43987d76009ea1d7b5cabb5b05849ecea1a

* Resolves: # 
* Target version: master 

Tested targets `check`, `check-desktop`, `check-mobile`, `check-idle`, `check-multi`. Also tested all of the above with a spec given. Also tested an invalid spec.

The primary benefit of this is that multiuser tests report their failures in one table so you don't have to scroll through the whole log looking at each one.
The secondary benefit is that the file becomes easier to read and edit. I was recently digging through it trying to investigate a test failure and it was making things difficult.

### Checklist


eszkadev (changes requested):
It has conflicts, things were changed in the meantime (since Jun). Can we close this @neilguertincollabora  ?

neilguertincollabora:
Yes, I will revisit this later when I get some time

PR #8059, opened 2024-01-19
Close autofilter popovers when click or change the sheet tab

Darshan-upadhyay1110:
- jsdialog not getting close when we change tab
- this will close all dialogs when we click on tab or change the tab

To test this patch
    - add 2 or 3 tab sheet in calc
    - add `autofilter` in any calc sheet tab
    - open autofilter popup
    - change sheet tab

Before this patch:
    - popup not getting close when focus goes away.

After: - all dialogs should be closed now when change the sheet tab.

Change-Id: I4eb308e499853a6e1d86b8dc7f60aa938e9c03b9

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -29,6 +29,7 @@ L.Control.JSDialog = L.Control.extend({
 		this.map.on('jsdialogaction', this.onJSAction, this);
 		this.map.on('zoomend', this.onZoomEnd, this);
 		this.map.on('closealldialogs', this.onCloseAll, this);
+		this.map.on('closeAutoFilterDialog', this.closeAutoFilterDialogsOnTabChange, this);
 	},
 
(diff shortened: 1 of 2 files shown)


eszkadev:
We can't do it this way:

[formula-2024-01-19_10.13.05.webm](https://github.com/CollaboraOnline/online/assets/5307369/88415b65-ad20-45e5-833f-c6e55dbbceb3)

eszkadev (approved):
code looks good

eszkadev (changes requested):
I found one other case:
- when you see autofilter popup, click insert sheet
- you will be switched and popup is still there

Darshan-upadhyay1110:
I guess i have fire the event at wrong place. now it works. Close dialogs when we add class tab selected so it will be more generic than call inside `setPart`

CC: @eszkadev

eszkadev (approved):

PR #8060, opened 2024-01-19
Small tooltip refactor

eszkadev:
Use once to call event handler just one time
    
    Use dedicated option, don't do it manually.
    
    https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once

-----------------------------------------------------

Move document tooltip helper to UIManager
    
    so it is with all other showXXXX helpers


mikekaganski (approved):
Great - thank you! I needed to document it myself - you did it for me, thank you!

PR #8061, opened 2024-01-19
cool#8023 clipboard: switch marker from outside-html to inside-body

vmiklos:
As pointed out at
<https://github.com/CollaboraOnline/online/pull/8041#discussion_r1457933092>,
it's not valid to have a markup like <div><html>...</html></div>.

The old markup was <html><head><meta ...></head>...</html> and that got
stripped away.

Fix the problem by going with <html><body><div ...>...</div></body></html>
which survives browser sanitize and also is a valid HTML.

Both the browser side and the wsd side is updated to go with this
inside-body markup. This also allows simplifying setTextSelectionHTML(),
because the valid HTML means the browser automatically strips the
<style> element whtn converting to plain text. Also remove a leftover
debug code in ClientSession::preProcessSetClipboardPayload() that I
forgot to remove.

Change-Id: Icb0b26a37320ae98237222b450a861ebcb2f0c43

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -117,37 +117,36 @@ L.Clipboard = L.Class.extend({
 
 	// wrap some content with our stub magic
 	_originWrapBody: function(body, isStub) {
+		var lang = 'en_US'; // FIXME: l10n
 		var encodedOrigin = encodeURIComponent(this.getMetaURL());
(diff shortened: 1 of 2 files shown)


vmiklos:

At the end doing the div inside body works, it survives sanitize and looks valid as well. The PR message looks corrupted due to the HTML tags, but the commit message has it correctly.

caolanm (approved):
much more reassuring

PR #8062, opened 2024-01-19
lang ended up as "(null)" and docs from debug.html didn't load

caolanm:
Change-Id: Ie007dbdc2d4066c657d2a1cd05c6f4532404495b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


hcvcastro (approved):
Thanks

PR #8063, opened 2024-01-19
Private/caolanm/better session updates deltas display

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -5140,6 +5140,8 @@ L.CanvasTileLayer = L.Layer.extend({
 				this._debugInvalidateCount + ', recv-tiles: ' + this._debugLoadTile +
 						       ', recv-delta: ' + this._debugLoadDelta +
 						       ', recv-update: ' + this._debugLoadUpdate);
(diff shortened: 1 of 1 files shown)


caolanm:
maybe we should make the _debugLoadTile/_debugLoadDelta/_debugLoadUpdate init from the start of the session and always collect them even when not currently debugging so we can show up to date numbers when the debugging display is used, rather than starting from 0 from beginning of debugging.

Seems to be a little ambiguity about what a null img might mean, debugging code counts it as a "update" but presumably if there was a null img that would raise an exception currently so maybe it's worth throwing out the null check to make things more straight forward

mmeeks (approved):
Looks fine; I was hoping for a big, clearly labelled: "tile waste: 50%" type thing - the number of empty updates / number of tiles we've received is that number and - I expect when things go wrong that number goes up and up noticeably =)

caolanm:
Alright, so lets move the exact nums back to the bottom left and put a single "Tile update waste:" % num at the top left in bold and bigger font?

mmeeks (approved):
We should include loadTile in the %age too - it's fine if there is a new tile from fresh; but prolly we should include in that text a %age of deltas/updates to new tiles too - which is also a quality metric - anyhow - lets get this in. Thanks ! =)

PR #8072, opened 2024-01-23
IdleHandler: close idle mode on hover

eszkadev:
Some people say it's annoying that you need to click to enable the document. This patch makes it automatic so popup is closed when mouse enters the app area.


hcvcastro (approved):

hcvcastro:
It looks OK.

hcvcastro:
FreeBSD 13. I think is unrelated, Please merge if pass it

PR #8077, opened 2024-01-23
Add visual hint for freeze pane Lines

Darshan-upadhyay1110:
- added row/col divider with a transparent gradient over the area where cells are concealed
- shadow color will change according to UI theme (Dark/Light)

[Screencast from 2024-01-23 19-38-01.webm](https://github.com/CollaboraOnline/online/assets/61383886/c44ea050-2fa8-45d3-b213-ecaa0d28a30a)

Change-Id: I3fe6f3b35ec84cda5d93fe9a094765cfb75b906c

### DIFF
--- a/browser/src/layer/vector/CPath.ts
+++ b/browser/src/layer/vector/CPath.ts
@@ -30,6 +30,8 @@ abstract class CPath extends CEventsHandler {
 	point: cool.Point;
 	zIndex: number = 0;
 
+	isTopOrLeftOfSplitPane: boolean = true;
+
 	static countObjects: number = 0;
 	private id: number;
 	private isDeleted: boolean = false;
--- a/browser/src/layer/vector/CanvasOverlay.ts
+++ b/browser/src/layer/vector/CanvasOverlay.ts
@@ -470,6 +470,10 @@ class CanvasOverlay extends CanvasSectionObject {
 		if (path.fill) {
 			this.ctx.globalAlpha = path.fillOpacity;
 			this.ctx.fillStyle = path.fillColor || path.color;
+			if (!path.isTopOrLeftOfSplitPane) {
+				this.setBoxGradient(path);
+			}
+
 			this.ctx.fill(path.fillRule || 'evenodd');
 		}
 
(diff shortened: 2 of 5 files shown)


Darshan-upadhyay1110:
@pedropintosilva need your view on UI. Is box-shadow looks okay or need any design changes ?

Darshan-upadhyay1110:
i have amended some changes recently, this is the new look of canvas after changes.

[Screencast from 2024-01-25 13-00-44.webm](https://github.com/CollaboraOnline/online/assets/61383886/44b4ac7f-43ad-484b-9820-5b8cfc7ba64c)

mmeeks:
I'm getting extremely tired of this bug. Please read the very clear spec. then carefully compare the screen mockup I gave you with what you've produced. If it matches the mockup then we should merge it, if not - we should not. Note - it does not. And the bar is now hugely and disturbingly wide and over-writes the headers in a confusing way that we are trying to keep constant down all the columns: why !?

Darshan-upadhyay1110:
> I'm getting extremely tired of this bug. Please read the very clear spec. then carefully compare the screen mockup I gave you with what you've produced. If it matches the mockup then we should merge it, if not - we should not. Note - it does not. And the bar is now hugely and disturbingly wide and over-writes the headers in a confusing way that we are trying to keep constant down all the columns: why !?

Sorry meeks for the trouble i messed up with the coordinates in `createLinearGradient` which lead me to increase the line thickness ( not a good thing that i did :(  applogies )

I corrected those mistakes and updated the patch. The only thing is left now is to make sure the current state should look same as the mock-up image.

For my notes:

Current state:

image

Ref/Mock-up image:

image

mmeeks:
Pixels look better, what is a regular expression doing in the code? does it work in dark mode ? Let me drop myself as reviewer here. The thinner, and higher contrast we can make the splitter in the 1st instance (perhaps just 1 pixel) the better - and I think this gradient can make the 1-pixel grid line much more visible and effective and much better for layout; but - lets get something merged here.

Darshan-upadhyay1110:
>  what is a regular expression doing in the code? 

- It is use to set `addColorStop` values for gradient in [ CanvasOverlay.ts](https://github.com/CollaboraOnline/online/pull/8077/files#diff-5963e7a749945e06f5d7071658db98e95b990a019b1608ae7c6a4b02a94540dfR507)  from `background` property in cool.css
- it allows us to change any gradient effect directly from CSS without changing anywhere in Overlay file

>  does it work in dark mode ?

Yes 
image

> The thinner, and higher contrast we can make the splitter in the 1st instance (perhaps just 1 pixel) the better - and I think this gradient can make the 1-pixel grid line much more visible and effective and much better for layout; but - lets get something merged here.

Sure, I am working on that as well (I will complete that soon :) )

pedropintosilva (changes requested):
Thank you @Darshan-upadhyay1110 

The light mode seem to correspondent Michael's expectation. We should merge it. (personally I think the original mockup is too indiscreet [too much in your face] but we we can always adjust after getting user's feedback)

The dark mode is even more indiscreet. But We can merge. 
Note:
* Ideally, Dark mode related gradients shouldn't be just invert AKA light gradient they should be darker actually. The Problem is that using dark gradient here. But with the current implementation changing gradient will also affect the solid color. See the problem: [data-theme-dark-splitters-linear-gradient.patch.txt](https://github.com/CollaboraOnline/online/files/14110524/data-theme-dark-splitters-linear-gradient.patch.txt), [data-theme-dark-splitters-linear-gradient](https://github.com/CollaboraOnline/online/assets/65948705/a629f048-4b05-40e0-95ca-91c9ebcb2232)

But before we merge this please do fix formatting, by removing unnecessary tabs and spaces:
image

browser/src/layer/vector/CPath.ts
```diff
@@ -30,6 +30,8 @@ abstract class CPath extends CEventsHandler {
 	point: cool.Point;
 	zIndex: number = 0;
 
+	background:string;
```
eszkadev, line 33: space, `background: string;`

eszkadev:
Please check RTL sheet:
[rtl-2024-02-02_14.56.12.webm](https://github.com/CollaboraOnline/online/assets/5307369/4b113a4c-476b-44fd-acc9-c6dc2451627e)

browser/src/layer/vector/CanvasOverlay.ts
```diff
@@ -485,6 +489,53 @@ class CanvasOverlay extends CanvasSectionObject {
...
+		// last row geometry data will be a good for setting deafult raw height
+		const spanlist = this.map._docLayer.sheetGeometry.getRowsGeometry()._visibleSizes._spanlist;
+		const rowData = spanlist[spanlist.length - 1];
+
+		// Create a linear gradient based on the extracted color stops
+		// get raw data from sheet geometry. use index = 1
+		const deafultRowSize = rowData.data.sizecore;
+		// gradient width shoulb be half a default row hight.
+		const gradientWidth: number = Math.ceil(deafultRowSize / 2);
+		const colorStopsArray = [
```
eszkadev, line 505: move this directly to `createLinearGradientBasedOnColorStops`
no need to pass it, it is used only there

eszkadev: and then name could be eg. `createSplitLineGradient`

Darshan-upadhyay1110:
> Please check RTL sheet: [rtl-2024-02-02_14.56.12.webm](https://github.com/CollaboraOnline/online/assets/5307369/4b113a4c-476b-44fd-acc9-c6dc2451627e)

Gradient calculation for RTL is done

Thanks, @eszkadev for checking

eszkadev (approved):
Looks ok, RTL now works

pedropintosilva (approved):
Thanks. As an improvement for next next we should make the gradient darker (black) on dark mode or if it affects or introduces regression on light mode we should just remove it when on dark mode

PR #8080, opened 2024-01-23
sidebar: initialize after recover from idle

eszkadev:
After document idle (document become unloaded on the server) when we click in the popup and document was reloaded - if sidebar was present previously it should be reinitialized.

Without that it was visible but not interactive and it was not possible to hide it.

1. To test that use small timeout in the coolwsd.xml: per_document.idle_timeout_secs
2. Open sidebar
3. Wait for idle
4. Click in the idle popup

Sidebar should be interactive and it should be possible to hide it.


hcvcastro:
It looks OK

PR #8084, opened 2024-01-24
fix UI bugs in adminClusterOverview.html

Rash419:
* Resolves: # 
* Target version: master 

Before:
image

After:
image


Rash419:
cherry-picked to distro/collabora/co-23.05 branch: https://github.com/CollaboraOnline/online/pull/8085

pedropintosilva (approved):

pedropintosilva:
it seems good, I just wonder if wouldn't be better to try to reload that parent element instead of the whole page.

PR #8086, opened 2024-01-24
calc: don't jump on merged cell selection with split panes

eszkadev:
This issue can be reproduced in spreadsheet with split pane
where some cell is merged and longer than a visible area.
Then every click on that cell will move view to the
beginning of the cell - we don't want that.

This is followup for: https://github.com/CollaboraOnline/online/pull/7881


eszkadev:
It should work like core:
[merged-2023-12-19_09.44.43.webm](https://github.com/CollaboraOnline/online/assets/5307369/527e793c-55ca-4271-99ea-65c6da16666b)

(even with split panes)

quikee (approved):
Looks good to me

PR #8087, opened 2024-01-24
calc: avoid view jump when using formula wizard

eszkadev:
Steps to reproduce problem:
1.    open spreadsheet
2.    open function wizard (Fx icon next to the formulabar)
3.    move dialog so it is outside view partially
4.    select some function and click inside edit field for it's param
5.    click on the sheet content
Result: view jumps and moves everything up

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -398,6 +398,7 @@ L.Control.JSDialog = L.Control.extend({
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


eszkadev:
[formulabug-2024-01-18_19.57.25.webm](https://github.com/CollaboraOnline/online/assets/5307369/53d02c48-9d0c-4563-83fd-b8bced3572a7)

hcvcastro (approved):

hcvcastro:
It looks Ok.

PR #8088, opened 2024-01-24
calc: avoid view jump when focusing formula wizard

eszkadev:
Change from commit 85ff10db7a861ac0a848d6276347b8709d151a23
calc: avoid view jump when using formula wizard

wasn't enough to fix the issue. When we receive grab_focus action for widget which is hidden outside view - browser will move the app view leaving blank space on bottom.

Steps to reproduce problem:
    1.    open spreadsheet
    2.    open function wizard (Fx icon next to the formulabar)
    3.    move dialog so it is almost completely outside view (edit field
          with focus has to be outside view)
    4.    select some function and click inside edit field for it's param
    5.    click on the sheet content
    Result: view jumps and moves everything up
    
see: https://github.com/CollaboraOnline/online/pull/8087


gokaysatir:
Issue is reproducible with Firefox browser.

gokaysatir:
Solves the issue but we have a regression. This doesn't happen on master branch. Can you have a look when you have time?

eszkadev:
Solved

gokaysatir (approved):

PR #8089, opened 2024-01-24
Extract debug tools to new file core/Debug.js

neilguertincollabora:
Change-Id: I20f42d24a6af91bdcf79dc7fe19e1cf25f90bd89

* Resolves: # 
* Target version: master 

### Summary
Moves 600+ lines out of CanvasTileLayer.js into an new file core/Debug.js. Most of this code had nothing to do with the CanvasTileLayer, that's just where it was put because the first debug tool happened to have been put there.

The only changes are to move the code, move the ownership from the doc layer to the map, and rename variables and functions. There is no functional change.

I'm looking for feedback on the new file location (perhaps map/Debug.js would be better?), gotchas around adding a new js file (I added it to the two lists in the Makefile but I'm not sure it needs to be in the pot section), and anything else.

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -206,6 +206,7 @@ COOL_JS_LST =\
 	src/core/Events.js \
 	src/core/Socket.js \
 	src/core/Matrix.js \
+	src/core/Debug.js \
 	src/geometry/Point.ts \
 	src/geometry/Bounds.ts \
 	src/geometry/Transformation.ts \
@@ -772,6 +773,7 @@ pot:
 		src/control/jsdialog/Widget.MobileBorderSelector.js \
 		src/control/jsdialog/Widget.TreeView.js \
 		src/core/Socket.js \
+		src/core/Debug.js \
 		src/docstate.js \
 		src/errormessages.js \
 		src/layer/tile/CanvasTileLayer.js \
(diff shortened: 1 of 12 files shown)


hfiguiere:
Shouldn't `Debug.js` be `Debug.ts` and moved to TypeScript?

neilguertincollabora:
> Shouldn't `Debug.js` be `Debug.ts` and moved to TypeScript?

Maybe later. I want to keep this change as straightforward as possible. Just copy and paste, and rename where necessary.

mmeeks (approved):
Not sure consolidating all debugging code from across many different classes with different lifecycles into a single static shared class would be a generally positive move. But having a separate class for a parent classes debug logic is fine.

Can we note move this to TypeScript ? new modules should be that really. Anyhow - lets get it in not to hold up progress.

I guess this level of largescale cosmetic change however breaks our ability to back-port vital debug work to the stable branch: the last similar cosmetic re-factor caused a handful regressions which we can't take for the rather low level of benefit this brings.

So - we will not include this in 23.05 - so if you merge this, and work on it, you'll need to be responsible for back-porting, and carefully re-validating in 23.05 - to ensure that there are no additional regressions from the back-port that are not caught because you're working on 23.05. so ... in a nutshell, not a huge fan of the idea at this stage - but if you like it and are willing to continue to maintain the test automation separately in 23.05 - lets get it in.

PR #8090, opened 2024-01-24
fix fail unit test "builddir != srcdir"

hcvcastro:
Change-Id: I7a5b96da9e892c1948cbbc52c57344a2b67daab6

### DIFF
--- a/test/FileServeWhiteBoxTests.cpp
+++ b/test/FileServeWhiteBoxTests.cpp
@@ -348,7 +348,7 @@ void FileServeTests::testPreProcessedFileRoundtrip()
 {
     constexpr auto testname = __func__;
 
-    const Poco::Path path(TDOC "/../../browser/dist");
+    const Poco::Path path(TDIST);
 
     std::vector<std::string> files;
     Poco::File(path).list(files);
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -69,6 +69,9 @@ nginxconfig_DATA = etc/nginx/coolwsd.conf
 
 ACLOCAL_AMFLAGS = -I m4
 
+DIST_FOLDER ?= $(abs_builddir)/browser/dist
```
mmeeks, line 72: Not clear why we export this here, when we can calculate this from abs_top_builddir in the test/Makefile.am surely and not jam stuff into the environment (?) =)

Good fix though, can you simplify & push with that ?

Thanks!

hcvcastro: We have the same variable in browser/Makefile.am, it is configurable to custom dist folder, I cannot do it in
test/Makefile.am since browser project can have different value.

mmeeks: $(abs_top_builddir)/browser is not reliable ? - note the _top_ in the variable name (?) =)

hcvcastro: Yes, please check browser/Makefile.am:9

DIST_FOLDER ?= $(builddir)/dist

This variable can be customized, if not set default value

mmeeks (approved):
Thanks =)

PR #8092, opened 2024-01-25
sidebar: reinitialize after reconnect only

eszkadev:
It seems that commit 2d7743cb734b31b763f389d686ca349661de9739 sidebar: initialize after recover from idle

Introduced a regression, in multi user case sometimes sidebar is shown/hidden by other users. Also app was slow sometimes what could be a sidebar reloading all the time.

Revert previous single-init on docloaded and use dedicated reconnection code to reinitialize sidebar.

This should still fix the problem that when doc become unloaded with sidebar opened - after reconnecting it was not possible to close.


caolanm (approved):
looks very plausible to me

PR #8094, opened 2024-01-25
browser: fix "Uncaught TypeError: getSectionWithName"

hcvcastro:
Change-Id: I660ca7cbc9e6ce6eaa3834ebd3c2829c606bc045

### DIFF
--- a/browser/src/layer/tile/CommentSection.ts
+++ b/browser/src/layer/tile/CommentSection.ts
@@ -988,7 +988,7 @@ export class Comment extends CanvasSectionObject {
 	}
 
 	public static isAnyEdit (): boolean {
-		var section = app.sectionContainer ?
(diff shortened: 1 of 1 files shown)


caolanm (approved):
what was it actually an instanceof?

hcvcastro:
>> what was it actually an instanceof?

It is something like dynamic_cast in C++, but Javascript called prototypes

caolanm:
yeah, I get that :-). I mean it wasn't a CanvasSectionContainer so what was passed there?

hcvcastro:
>> yeah, I get that :-). I mean it wasn't a CanvasSectionContainer so what was passed there?

I have checked all assignments, but somewhere the app.sectionContainer is assigned an empty object
or other type during certain condition (reloading, reconnecting document, etc)

PR #8098, opened 2024-01-25
add pageup and page down to preview navigation

caolanm:
Change-Id: I8000946e5d4c28288d0ad778b243ce8442c3b8c0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (commented):
Looks plausible; quite probably I had focus in the slide sorter when I was playing with this =)

caolanm:
I think this is worth doing, but I think I can reproduce your original scenario without focus in this preview by generating a load of slide changes and getting the main pane into some sort of of state where it won't refresh without clicking on it.

mmeeks (approved):
Looks great - thanks Caolan ! =)

PR #8099, opened 2024-01-25
the viewrenderedstate=string string is only used in debugging

caolanm:
so we can pass through the full state that the view reports which makes it easy to see if a browser side is given a view whose render settings are unexpected.

e.g. clicking "formatting marks" and not getting a viewrenderedstate with 'P' or view#1 clicks their "formatting marks" and view#2 receives a 'P'ilcrow state.

Change-Id: I31d3a8397f02dedf2505bed5dd83576f3138b504

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1451,7 +1451,7 @@ private:
         std::string stateName;
         if (!viewRenderedState.empty())
         {
-            stateName = viewRenderedState.substr(viewRenderedState.find(';') + 1);
+            stateName = viewRenderedState;
         }
         else
         {


caolanm:
It is very convenient for debugging to see browser side what the reported render state of the server side view actually is. It just gets reported as "Light" or "Dark" and drops the Pilcrow, Spellchecking tags. The actual string doesn't currently get used for anything browser side, just a debugging message.

mmeeks (approved):
Love it =) I assume we show it in the debug overlay / display.

caolanm:
yeah, its there in bottom left now as:
"Canonical id changed to: 1003 for view id: 0 with view renderend state: PS;Dark"
vs before of:
"Canonical id changed to: 1003 for view id: 0 with view renderend state: Dark"

Issue #8031, opened 2024-01-16
Image size buttons have different collor and spacing in FireFox vs Chrome


sebastiaanveld:
### Describe the Bug
Image size buttons have different collor and spacing in FireFox vs Chrome. In Chrome the dark branding seems to be followed correctly, while in FireFox the collor looks off from the dark mode.

### Steps to Reproduce
1. Using Firefox on your desktop, open Nextcloud and open a document for editing in Dark Mode
2. Insert as new image, or use an existing image and use right moude button to view the image properties
3. The buttons of the image size are light, amost like in Light mode light. Also the spacing seems off.
4. Do the same in Chrome browser, bow the buttons of the image size are more in line with dark-mode and spacing seems right.

### Expected Behavior
Consistent look between browsers (Chrome vs Firefox), consitent spacing and consintent collor of the buttons.

### Actual Behavior
In FireFox the collor of the buttons seems too light for dark-mode. Also the spacing seems off for the metrics text.

### Screenshots
FireFox 121.0.1 (x64)

![image](https://github.com/CollaboraOnline/online/assets/37268999/28478830-10e2-424e-8e6f-55053586ddd1)

Chrome 120.0.6099.217 (x64)

![image](https://github.com/CollaboraOnline/online/assets/37268999/4e87a6a1-4d99-49ec-812f-217c68d136d5)

### Desktop
(Please complete the following information)
 - **Collabora version:** 
COOLWSDversion: 23.05.7.2 (git hash: f76e08a (E))
LOKit version: Collabora Office 23.05.7.2 (git hash: 6d9d5d1)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: c31b1370
 - **Nextcloud:** 27.1.4
 - **OS and version:** Windows 11 Ent, Dutch, x64
 - **Browser and version:** FireFox 121.0.1 (x64) vs Chrome 120.0.6099.217 (x64)

timar:
For me it looks like these up/down arrow widgets are coming from the browser.  `<input type="number">` looks like this in Firefox. 
For the padding of spinfield unit, I propose a patch.

PR #8100, opened 2024-01-25
fix padding of spinfield unit on Firefox #8031

timar:
Change-Id: Iff389334e534eb386904705b00b917a848a7154e

* Resolves: #8031
* Target version: master


pedropintosilva (approved):
Thank you, tested it works including with largest unit suffix `mm`

PR #8105, opened 2024-01-26
widgets: Add a time field

meven:
Change-Id: I9e3e280db4a0b1f3e3a9df5fe8b6e0bfd97ddaa5

* Resolves: # 
* Target version: master 

### Summary

It uses input[type=time] which might not be looking as good as we'd like.

image

Requires CORE patch: https://gerrit.libreoffice.org/c/core/+/161892 to expose the custom properties tab.

And https://gerrit.libreoffice.org/c/core/+/162598 to expose the field.

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -723,7 +723,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 .ui-listbox,
-.ui-combobox {
+.ui-combobox,
+.ui-timefield  {
 	box-sizing: border-box;
 	-webkit-appearance: none;
 	-moz-appearance: none;
@@ -777,7 +778,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 
 /* ui-combobox */
-.ui-combobox {
+.ui-combobox,
+.ui-timefield {
 	border: 1px solid var(--color-border);
 }
 
@@ -1906,4 +1908,9 @@ kbd,
 /* Modal info with 2 line message*/
 .jsdialog-container [id^='info-modal-label'] {
 	display: flex;
+}
+
+#DocumentPropertiesDialog #customprops input.ui-timefield {
+	margin-top: auto;
+	margin-bottom: auto;
 }
--- a/browser/src/control/jsdialog/Widget.Timefield.js
+++ b/browser/src/control/jsdialog/Widget.Timefield.js
@@ -1,0 +1,47 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * JSDialog.Timefield - input field for time data
+ *
+ * Example JSON:
+ * {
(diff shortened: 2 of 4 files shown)


browser/src/control/jsdialog/Widget.Timefield.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * JSDialog.Timefield - icon view widget
```
eszkadev, line 13: `icon view widget` -> `input field for time data`

browser/src/control/jsdialog/Widget.Timefield.js
```diff
@@ -0,0 +1,54 @@
...
+JSDialog.timeField = function (parentContainer, data, builder) {
+	var container = L.DomUtil.create('span', builder.options.cssClass + ' ui-timefield-container', parentContainer);
+	if (data.id !== undefined) {
+		container.setAttribute('id', data.id);
+	}
+
+	var inputTimeField = L.DomUtil.create('input', builder.options.cssClass + ' ui-timefield', container);
+	inputTimeField.setAttribute('type', 'time');
+	inputTimeField.setAttribute('step', 1); // forces the display of seconds
+	inputTimeField.id = data.id;
```
eszkadev, line 34: It is incorrect that 2 nodes have the same id (container and inputTimeField).
Please remove container and use only inputTimeField. As I understand there is only one element inside container - please simplify it.

browser/src/control/jsdialog/Widget.Timefield.js
```diff
@@ -0,0 +1,54 @@
...
+	inputTimeField.value = data.text;
+
+	inputTimeField.addEventListener('change', function (event) {
+		var timefield = event.target;
+
+		var attrdisabled = timefield.getAttribute('disabled');
+		if (attrdisabled !== 'disabled') {
+			var valueMSecs = timefield.valueAsNumber * 1000;
+			builder.callback('spinfield', 'change', timefield, timefield.value, builder);
+			console.log('changed', event, timefield.valueAsNumber, valueMSecs, timefield.value);
```
eszkadev, line 44: builder.callback will log the message we send, is that console.log needed? At least please make it console.debug so it will not appear in logs when not needed.

browser/css/jsdialogs.css
```diff
@@ -1906,3 +1906,9 @@ kbd,
 .jsdialog-container [id^='info-modal-label'] {
 	display: flex;
 }
+
+input.ui-timefield {
+	width: 7.3em;
+	height: 1.7em;
```
eszkadev, line 1912: is the height unified with other widgets: edit/input field, spinfields ?

eszkadev (approved):
Looks good

browser/css/jsdialogs.css
meven: No, I should probably use a min-height here.
I picked this height just so that it looks ok:
image

eszkadev: ok, maybe left it as it is now. We will create issue and unify fields height at some point I think.

meven: With unified:

image

And then the border tuned:

image

meven: I went with the last screenshot styling.

eszkadev (approved):

PR #8107, opened 2024-01-26
Revert "cool#8023 clipboard: remove old marker"

vmiklos:
This reverts commit cc12a1b060adb31afe037bb5e18f1cbe4885c3e6, because
there is not much short-term benefit of this and it may cause problems,
as it affects builds produced without --enable-experimental, too.

Change-Id: I4c27614e9ee26c8aaa2c510ee6bd115cd19a45f3


vmiklos:
CC @timar

mmeeks:
We do eg. dynamic cluster upgrades where some nodes have upgraded and others not - and best to keep that working really nicely inside a major version =)

Issue #8016, opened 2024-01-12
Faster bind mounting of jails - fork/exec pte mgmt ...


mmeeks:
Long running profiles of demo servers suggest that really quite a lot of time is spent in forkit and also in kit_spare doing fork / sys_execve -  a quick strace -e execve shows that all of these are coolmount commands

around five of these per jail:

[pid 1057514] 1705067637.961776 execve("/bin/sh", ["sh", "-c", "/opt/libreoffice/online/coolmount -b /opt/libreoffice/online/systemplate /opt/libreoffice/online/jails/1056869-c351f329/i9FcuX8Q6hJescJX"], 0x3117890 /* 129 vars */) = 0
[pid 1057514] 1705067637.963976 execve("/opt/libreoffice/online/coolmount", ["/opt/libreoffice/online/coolmount", "-b", "/opt/libreoffice/online/systemplate", "/opt/libreoffice/online/jails/1056869-c351f329/i9FcuX8Q6hJescJX"], 0x55bb36ee85e0 /* 129 vars */) = 0
[pid 1057514] 1705067637.970834 +++ exited with 0 +++
[pid 1057513] 1705067637.970872 --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=1057514, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---

Those spawn a shell and then coolmount, which is not ideal. This chews up quite a chunk of time:
![demo-2024-01-12](https://github.com/CollaboraOnline/online/assets/122848/699d99ae-8e85-40dd-a903-8194f80d036e)

Currently I believe the mounting is currently done by synchronously forking coolmount from the forkit process.

The cost here is mostly that the forkit has a very large number of shared libraries mapped and pre-initialized, which requires lots of work managing PTEs etc. as we double fork and exec etc.

Instead we should have a helper process communicated with via a UDS that checks PEERCREDS, and parent, and can only be run by coolforkit - and that has a simple text protocol to pass jail identifiers to mount, and that does those sys-calls without any forking in the client; can do that with a simple, synchronous IPC layer I think replacing the coolmount call.

Hopefully a simple enough re-work that would save lots of kernel-level thrash, and save ~10% of the CPU time we can see in the profile I think =)

Thanks!

mmeeks:
[...]

PR #8109, opened 2024-01-26
cool#8016 speed up spawning/execing child programs

grandinj:
using fork(), the kernel needs to copy the VM data structures, which can be quite large for the main COOL processes.

I looked into things like vfork(), clone() and using a helper-process, but it looks like posix_spawn is sufficient for our needs. Internally it uses clone(CLONE_VM) which means we avoid the VMA copies.

Also, simplify the arguments of spawnProcess,
we don't use the optional params anymore.

Change-Id: I8f943541dc4d7f56c2d36ccbf7b78c40ec14b8e0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -38,6 +38,7 @@
 #include <unistd.h>
 #include <dirent.h>
 #include <fcntl.h>
+#include <spawn.h>
 
 #include <atomic>
 #include <cassert>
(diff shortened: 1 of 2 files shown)


mmeeks (commented):
I like the code size reduction; do we get stdout,stderr inherited so we can get the error messages etc. out of coolmount and into the journal in case of failures ? =)

mmeeks:
strace shows: clone(child_stack=0x7fc31fe95ff0, flags=CLONE_VM|CLONE_VFORK|SIGCHLD... - which looks helpful =) would be nice if we can win with this; so that piece looks promising.

mmeeks:
Lets get this in and see =) thanks Noel!

mmeeks:
Interestingly when compiling locally I get this cryptic complaint:

common/Util.cpp: In function ‘int Util::spawnProcess(const string&, const StringVector&)’:
common/Util.cpp:226:18: error: ‘posix_spawn_file_actions_addclosefrom_np’ was not declared in this scope
         status = posix_spawn_file_actions_addclosefrom_np(&fileActions, STDERR_FILENO + 1);
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
common/Util.cpp:226:18: note: suggested alternative: ‘posix_spawn_file_actions_addchdir_np’
         status = posix_spawn_file_actions_addclosefrom_np(&fileActions, STDERR_FILENO + 1);
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                  posix_spawn_file_actions_addchdir_np

=)

mmeeks:
Hmm - I merged too soon; reverted the 2nd commit Noel - the API is too recent. Also for forkit (where this makes most difference and where we do lots of our forks) - we have a well defined, fully CLOEXEC situation I think - so we don't need to care there, more details in the ticket: https://github.com/CollaboraOnline/online/issues/8016#issuecomment-1913143112

mmeeks:
Pull request to improve things here: https://github.com/CollaboraOnline/online/pull/8112 =)

PR #8110, opened 2024-01-26
browser: fix cell selection if ESC is pressed

hcvcastro:
If a cell is edited and suddenly press ESC key,
followed by Shitf + Arrow down to select more cells,
so the "uno:InsertMode" state arrives late and
it cannot update the row header selection.

Change-Id: Ib0e4e0699336e07d96efd96559f7b47caa285921


caolanm (approved):
seems reasonable, if we get ci to agree

PR #8112, opened 2024-01-27
sockets should use CLOEXEC ticket cool#8016

mmeeks:
Change-Id: I5f59ed9446a1fda301bb4791f98ca7c1c137e606

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/FakeSocket.hpp
+++ b/net/FakeSocket.hpp
@@ -18,9 +18,12 @@
 #include <poll.h>
 
 #ifndef __linux__
-#ifndef SOCK_NONBLOCK
-#define SOCK_NONBLOCK 0x100
-#endif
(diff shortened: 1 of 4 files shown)


grandinj (approved):

grandinj:
Noting that if we want to be more cross-platform we could do a

      (void) fcntl(fd, F_SETFD, FD_CLOEXEC);

on the socket fd

mmeeks:
> Noting that if we want to be more cross-platform we could do a

Sure, but then there is a race between opening and CLOEXEC'ing - at least, if we fork/exec in some threaded program ;-) which of course is a bad idea anyway. We can of course do that but ...

PR #8115, opened 2024-01-29
cypress: fix writer/top_toolbar_spec.js insert special character

eszkadev:
In 24.04 we activated dialog which informs about idle dialog. It covers full screen so some tests fail on that.
Add {force: true} to ignore that in failing case.


neilguertincollabora:
Isn't this dialog a permanent change? So we would not need the FIXME comments to eventually undo the changes to the test.

eszkadev:
> Isn't this dialog a permanent change? So we would not need the FIXME comments to eventually undo the changes to the test.

No, we should make all dialogs Async so they will not show this warning. I left FIXME so when someone will see error ther - will know what is going on.

neilguertincollabora (approved):

PR #8121, opened 2024-01-29
Debug tool for incoming/outcoming/keyboard msgs

neilguertincollabora:
Change-Id: I157cf9b651069e8f4c81b4595cddf7748323c062

### Summary
Adds three new tools:
- Log incoming messages
- Log outgoing messages
- Log keyboard events

I left them on by default in debug mode but this change does disable the logging until debug mode is started.

### Checklist


mmeeks (approved):
Seems fine though you can filter in the JS console easily too =)

PR #8122, opened 2024-01-29
Update iOS settings panel

timar:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


plubius (approved):
Rebuilt both LibreOffice and Online with the changes and tested on my iPad. I see the new Version field set to 23.0.5.8.2 and the Template list URL is now longer in the app's settings.

PR #8123, opened 2024-01-30
wsd: Add config to customize OS version info

thebearon:
Change-Id: I12abdd0755883acb7bae126ac3c9c999598b4428

* Target version: master 

### Checklist


mmeeks (approved):
Looks ok; I'm not a huge fan of these extra attributes to enable things though. I would instead use the empty string as "use OS values" and note that if un-set it will use data from the OS - that's more future-proof: we will kill Poco one day, and the extra complexity of these at-enable things is the future's unwelcome past baggage =)

If you can cleanup & get into 23.05 - we're really trying to get people to move to 23.05 left and right of course [!] =)  if this can encourage them to upgrade that sounds like a good thing to me.

Thanks!

thebearon:
> Looks ok; I'm not a huge fan of these extra attributes to enable things though. I would instead use the empty string as "use OS values" and note that if un-set it will use data from the OS - that's more future-proof: we will kill Poco one day, and the extra complexity of these at-enable things is the future's unwelcome past baggage =)

Patch updated, thanks for the comment!

PR #8124, opened 2024-01-30
Calc: Selecting columns by dragging mouse deselect the first column

lpranam:
problem:
drag selection starting from the first row or column would result in first col or row being unselected. It was caused by index being 0 resulting in condition being false

Change-Id: I3619f52ccc7fdb2883af547f1c907c07ef27ddf9

* Target version: master 

### Checklist


Rash419 (approved):
Looks good me!
Works as expected :)

Issue #8029, opened 2024-01-15
Replace all menubartoolitems with customtoolitems


eszkadev:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @eszkadev

### Detailed description and rationale
There were in the past initial approach to share code between menubar and notebookbar by using `(big)menubartoolitem` element in the notebookbar JSON. Now we have better widget for that: `(big)customtoolitem` that will not use Menubar code at all so reduces external dependencies for us (to smartmenus library).

### Code pointers
- use `git grep menubartoolitem` command to identify existing items 
- we want to move implementation of command/action from menubar (https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.Menubar.js#L1841) to `dispatch` function https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Toolbar.js#L945
- `customtoolitem` runs `command` using `map.dispatch(command)` function where we need to put command implementation
- adjust Menubar code to also use `map.dispatch(command)` for item you moved
- JSON with notebookbar description is in `https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.NotebookbarWriter.js` for Writer and similar files for other apps

adighiero1:
hello could i be assigned to work on this?

eszkadev:
hi, @adighiero1 sure

adighiero1:
Hi @eszkadev. I was sick so I haven't been able to do it. can you tell me where the code is on your codebase that needs changing? I have the codebase on my computer right now and I'm getting familiar with the structure.

adighiero1:
Hi @eszkadev. I've been sick. I should have an update by monday.

PR #8125, opened 2024-01-30
browser: move  impl of savecomments command to Toolbar dispatch

meven:
Change-Id: Ifcb34977b72bb0eccd23667deb4750637810e569

* Resolves: # 
* Target version: master 

### Summary

cc #8029

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -1153,6 +1153,14 @@ L.Map.include({
 			this.sendUnoCommand('.uno:DeletePrintArea');
 			this.print();
 			break;
+		case 'savecomments':
+			if (this.isPermissionEditForComments()) {
+				this.fire('postMessage', {msgId: 'UI_Save'});
+				if (!this._disableDefaultAction['UI_Save']) {
+					this.save(false, false);
+				}
+			}
+			break;
 		default:
 			console.error('unknown dispatch: "' + action + '"');
 		}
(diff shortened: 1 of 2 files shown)


browser/src/control/Toolbar.js
```diff
@@ -1153,6 +1153,14 @@ L.Map.include({
 			this.sendUnoCommand('.uno:DeletePrintArea');
 			this.print();
 			break;
+		case 'savecomments':
+			if (this._map.isPermissionEditForComments()) {
```
eszkadev, line 1157: did you test this code? does it work?

I think `this` here is `map` object already

meven: I put this back in draft, it wasn't tested yet.
This is fixed now.

meven:
Do you know where I can test this @eszkadev?

I am surprised at how simple the change is, does this patch look as expected?

eszkadev:
https://github.com/CollaboraOnline/online/blob/543256280133e0a5d46bebc623796020425dd72a/browser/src/control/Control.Menubar.js#L543

you can find it in the menubar in classic mode in Impress/Draw

meven:
> https://github.com/CollaboraOnline/online/blob/543256280133e0a5d46bebc623796020425dd72a/browser/src/control/Control.Menubar.js#L543
> 
> you can find it in the menubar in classic mode in Impress/Draw

Only in readOnly mode though.

I commented the part that need it to be read-only and added a break in Toolbar.js,
I can confirm the code is called, and works as before.

meven:
> > https://github.com/CollaboraOnline/online/blob/543256280133e0a5d46bebc623796020425dd72a/browser/src/control/Control.Menubar.js#L543
> > 
> > you can find it in the menubar in classic mode in Impress/Draw
> 
> Only in readOnly mode though.
> 
> I commented the part that need it to be read-only and added a break in Toolbar.js, I can confirm the code is called, and works as before.

I can confirm with "&permission=readonly" too.

eszkadev (approved):

PR #8128, opened 2024-01-30
notebookbar: enable currency dropdown (backport to 23.05)

eszkadev:
jsdialog: openmenu action support in notebookbar
    
    In notebookbar we insert additional containers for rows and cells.
    When sending openmenu action we need to pass id of the real parent
    of a button which is a toolbox. Go up in the structure to get
    correct node.
    
    This is 23.05 only as in master we already have simplified structure.

--------------------------

notebookbar: enable dropdown for currency toolitem
    
    use toolbar from notebookbar_online.ui in the core


quikee (approved):
Looks OK

PR #8131, opened 2024-01-31
jsdialog: use correct parent for toolbox actions in notebookbar (23.05)

eszkadev:
Followup for commit e8357ceb287d506096f489f3ca3479ed228f7bf9
jsdialog: openmenu action support in notebookbar

This is 23.05 only as in master we already have simplified structure.

It fixes case: click on currency button (not arrow) in notebookbar to toggle.


quikee (approved):
Looks fine

PR #8141, opened 2024-01-31
Private/mmeeks/randombits

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


common/FileUtil.cpp
```diff
@@ -234,11 +234,24 @@ namespace FileUtil
     std::string getTempFileCopyPath(const std::string& srcDir, const std::string& srcFilename, const std::string& dstFilenamePrefix)
     {
         const std::string srcPath = srcDir + '/' + srcFilename;
-        const std::string dstFilename = dstFilenamePrefix + Util::encodeId(Util::rng::getNext()) + '_' + srcFilename;
```
caolanm, line 237: Breaks the !HAVE_STD_FILESYSTEM case which seems to be that of our current android/iOS ci build. I'll update this pr to fix that.

common/FileUtil.cpp
```diff
@@ -234,11 +234,24 @@ namespace FileUtil
...
+            dstPath = filesystem::temp_directory_path() / dstFilename;
+            try {
+                filesystem::copy(srcPath, dstPath);
+            }
+            catch (const std::exception& ex)
+            {
+                LOG_SYS("ERROR: unexpected conflict creating file: " << ex.what());
+                retry = true;;
+            }
+        } while (retry);
```
caolanm, line 254: are we asking for some sort of infinite loop if we run out of disk space or some other arbitrary error?

mmeeks: Well - currently we throw an exception and crash; also that exception is a horror:

terminate called after throwing an instance of 'std::experimental::filesystem::v1::__cxx11::filesystem_error'

And not worth catching IMHO ;-) and TBH - I have really no idea why we use this trendy C++ thing instead of the other code-path that (IIRC) was more reliable anyway for random reasons in the past :-)

Anyhow - if we get the random piece in prolly we don't need this other bit nearly as much.

caolanm: I see getTempFileCopyPath is only used in tests so doesn't get used outside of there, so that concern doesn't matter a whole pile from its usage. Can move it into tests later maybe.

common/Util.cpp
```diff
@@ -80,22 +79,25 @@ namespace Util
 {
     namespace rng
     {
-        static std::random_device _rd;
         static std::mutex _rngMutex;
-        static Poco::RandomBuf _randBuf;
 
         // Create the prng with a random-device for seed.
         // If we don't have a hardware random-device, we will get the same seed.
         // In that case we are better off with an arbitrary, but changing, seed.
```
neilguertincollabora, line 86: Update comment

neilguertincollabora (commented):
Have you been able to reproduce the failure and verify that this fixes it?

caolanm:
Checking ci with https://github.com/CollaboraOnline/online/pull/8144/commits/057514d21b33869adaa008f798974f8282c59d3e to dump std::rng::entropy then from https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-23.05/9924/console "entropy of 0" it is indeed 0 on that builder while 32 locally.

Ubuntu 18.04, not sure what version of libstdc++ is in play there, https://en.cppreference.com/w/cpp/numeric/random/random_device/entropy claims "The entropy of the Linux kernel device /dev/urandom may be obtained using [ioctl RNDGETENTCNT](https://man7.org/linux/man-pages/man4/random.4.html) — that is what std::random_device::entropy() in [GNU libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B11/random.cc#L188) uses as of version 8.1." so maybe there is something around that.

But neither here not there I guess. We are getting 0 on that ci, so that's the issue.

caolanm (approved):

mmeeks:
also 0 on my up-to-date laptop - openSUSE Leap 15.5
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 141
model name	: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
stepping	: 1
microcode	: 0x4e

mmeeks:
If we want a fallback we should use current time down to the nanosecond, not process-duration-elapsed-time =)

PR #8146, opened 2024-01-31
getTempFileCopyPath is only used by tests so move to test-only code

caolanm:
Change-Id: I7c8a122425c689edb3eeb8c80e08470eac347bcc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Love it =) - Can we move the low grade RNG across as well, I'd like not to have that in the product if possible.

PR #8152, opened 2024-02-01
Capture this explicitely by name

thebearon:
To avoid the following warning in C++20:
error: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Werror=deprecated]

Change-Id: I8b195d675ddc87fa6b16f7240a046fac9a7769e1

* Target version: master 

### Checklist

### DIFF
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -690,33 +690,33 @@ void Admin::pollingThread()
     }
 }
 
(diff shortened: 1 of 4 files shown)


vmiklos (approved):
Would be nice to have a syntax like `this, =` to capture `this` explicitly, but we don't have it, so this makes sense to me.

PR #8153, opened 2024-02-01
Bump baseline to C++20

thebearon:
Change-Id: I8201175e83eba549f22aaeb935cda46a712ec492

* Target version: master 

### Checklist


thebearon:
CI needs to be switched to test against core co-24.04, and thus have an environment that supports C++20.

timar (approved):

PR #8160, opened 2024-02-01
backporting: Avatar list: set Unknown user avatar with 0 margin, fix alignment

pedropintosilva:
It seems when users are unknown we can get a different JSON bject:
Object { ViewId: "8", UserName: "Unknown", Color: 13013504, … }

This is then flagged and set as a different (from the identified
avatars) html element in createAvatar from Control.UserList.js

Better to fix the alignment via CSS

Change-Id: I94fee62ae14bb47c6f989b4e0bb147f8ffc9ff39

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1649,6 +1649,10 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
 	margin-left: -10px;
 }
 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
backporting: https://github.com/CollaboraOnline/online/pull/8160 
[WAIT until ^ passes]

Issue #8157, opened 2024-02-01
Mouse-wheel exception


mmeeks:
From the server-side logs: not clear how it is generated...

```
"column": 26
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:   "line": 14052,
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:   "source": "
https://share.collaboraonline.com/browser/61a0df1/bundle.js
",
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:   "message": "Uncaught TypeError: Cannot read properties of undefined (reading 'length')",
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:   "vendor": "google inc.",
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:   "userAgent": "mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/120.0.0.0 safari/537.36",
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]: wsd-102222-108122 2024-02-01 11:36:07.791675 +0000 [ docbroker_0ae ] ERR  ToClient-30e1: jserror {
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]: | wsd/ClientSession.cpp:595
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:     at HTMLDivElement.<anonymous> (
https://share.collaboraonline.com/browser/61a0df1/bundle.js:17035:81
)
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:     at CanvasSectionContainer.onMouseWheel (
https://share.collaboraonline.com/browser/61a0df1/bundle.js:14138:69
)
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]:     at CanvasSectionContainer.propagateOnMouseWheel (
https://share.collaboraonline.com/browser/61a0df1/bundle.js:14052:26
)
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]: TypeError: Cannot read properties of undefined (reading 'length')
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]: Cannot read properties of undefined (reading 'length')
Feb 01 11:36:07 ip-172-31-14-76 coolwsd[102222]: } 
```

meven:
[...]

PR #8167, opened 2024-02-02
Fix Uncaught TypeError in propagateOnMouseWheel #8157

eszkadev:
No clear steps to reproduce. This patch at least prevents us from exception and processing events. Leave error in the console to see it is happening.


eszkadev:
Fixed by: https://github.com/CollaboraOnline/online/pull/8182

PR #8168, opened 2024-02-02
Fix Uncaught TypeError in propagateOnMouseWheel #8157

eszkadev:
No clear steps to reproduce. This patch at least prevents us from exception and processing events. Leave error in the console to see it is happening.

### DIFF
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -1277,7 +1277,11 @@ class CanvasSectionContainer {
 				propagate = false; // Window sections can not stop the propagation of the event for other window sections.
 		}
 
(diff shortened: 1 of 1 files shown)


eszkadev:
@gokaysatir any ideas, how it can happen?
https://github.com/CollaboraOnline/online/issues/8157

gokaysatir:
Comment on @meven's PR.

gokaysatir (approved):
I am indecisive with this PR. I think we can get this in with a note and roll back the changes after we become confident that the error is gone.

eszkadev:
Or maybe wait for the next community meeting, to see if error still is there as Meven's patch is in

eszkadev:
Fixed by https://github.com/CollaboraOnline/online/pull/8182

PR #8176, opened 2024-02-02
Add header to debug menu

neilguertincollabora:
### Summary
Instructions for exiting debug mode, just in case a user accidentally ends up here

image

### Checklist

### DIFF
--- a/browser/src/core/Debug.js
+++ b/browser/src/core/Debug.js
@@ -18,7 +18,7 @@
  * - Help > About > Triple Click (Toolbar.js aboutDialogClickHandler)
  */
 
-/* global app L */
+/* global app L _ */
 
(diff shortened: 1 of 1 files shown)


timar:
If it is for end users, I think the 2 added strings should be localizable.

neilguertincollabora:
> If it is for end users, I think the 2 added strings should be localizable.

Sure. How do I do that?

timar:
> > If it is for end users, I think the 2 added strings should be localizable.
> 
> Sure. How do I do that?

Add `_` to `/* global app L */` section, like `/* global app L _ */`
Then all localizable strings should be marked up as `_('string')`. That's all.

timar (approved):

PR #8177, opened 2024-02-02
Add formatting marks and sidebars to random user settings

neilguertincollabora:
Change-Id: I7692bb07adf7fc50ef72c4e933584573755295ad

### Summary
Adding more sidebars (Navigator for all document types, slide editing sidebars for Impress). The way sidebars are managed is very weird and requires lots of different cases.

Formatting marks is a straightforward toggle

The &randomUser=true url parameter doesn't open sidebars (and doesn't work for some other settings too). This will be fixed in a future change.

### Checklist


mmeeks (approved):
Looks great to me =)

PR #8179, opened 2024-02-04
More Parallel Loading Prerequisites

Ashod:
Non-functional changes in preparation for parallel loading. This should help reduce the review of #8052.

- wsd: exclude WopiStorage in mobile builds
- wsd: simplify WebSocketHandler::sendErrorAndShutdown
- wsd: support sendMessage with string literals
- wsd: optional timeout to getHttpSession helper
- wsd: refactor findOrCreateDocBroker

### DIFF
--- a/wsd/RequestVettingStation.cpp
+++ b/wsd/RequestVettingStation.cpp
@@ -15,7 +15,9 @@
 
 #include <COOLWSD.hpp>
 #include <TraceEvent.hpp>
+#if !MOBILEAPP
 #include <StorageConnectionManager.hpp>
+#endif // !MOBILEAPP
 #include <Exceptions.hpp>
 #include <Log.hpp>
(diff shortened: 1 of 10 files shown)


wsd/RequestVettingStation.cpp
```diff
@@ -15,7 +15,9 @@
 
 #include <COOLWSD.hpp>
 #include <TraceEvent.hpp>
+#if !MOBILEAPP
```
mmeeks, line 18: I hate more conditionals. Can't we stub this and/or have some no-op backend for the Mobile case? we really have waaay too many conditionals in the code, and it makes testing a nightmare. Do we really need these ?

mmeeks (approved):
Lots of nice cleanups here - we need to get rid of the mobile conditionals - but lets get this in with the other stuff and iterate =)

PR #8182, opened 2024-02-05
browser: CanvasSectionContainer initialize boundsList

meven:
Also move its declaration to the top of the class section with other class members.

Change-Id: I506423078423074171ab00efb6f2d3fab0c68796

* Resolves: # 
* Target version: master 

### Summary

Mostly a code smell fix and a formatting change, but it might help https://github.com/CollaboraOnline/online/issues/8157

https://github.com/CollaboraOnline/online/pull/8168 will prevent the actual issue.

### DIFF
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -229,6 +229,9 @@ class CanvasSectionObject {
 	sectionProperties: any = {};
 	callbacks: SectionCallbacks;
 
+	/// The sections those this section can propagate events to. Updated by container.
+	boundsList: Array<CanvasSectionObject> = [];
+
 	/// Paramaters: null (use sectionProperties).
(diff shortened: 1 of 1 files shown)


gokaysatir:
@eszkadev it seems possible to me when:
* drawingAllowed = false
* A new section is added.
* The new section is target to an event.

I think we can get this PR in. It will prevent above edge case if it doesn't solve the issue.

Adding a new comment and immediately scrolling the view may cause this. Or with multiple users, if one adds a comment and the other is scrolling at the same time. Hard to guess.

gokaysatir (approved):

eszkadev (approved):

eszkadev:
@meven can we have 23.05 backport?

meven:
> @meven can we have 23.05 backport?

https://github.com/CollaboraOnline/online/pull/8234

PR #8183, opened 2024-02-05
Load balancing fix

eszkadev:
Fix error on clipboard download fail

----------------------------------

Use encoded WOPISrc in ws requests
    
    This is an issue if we use encoded and decoded WOPISrc for the
    same document when using load balancers like HAProxy. It should
    be always encoded.
    
    What we did:
    
    cool.html:
    http://haproxy.local:9980/browser/2a43eb3236/cool.html?WOPISrc=http%3A%2F%2Fsomewopisrcfileaddress&title=title
    
    ws:
    ws://haproxy.local:9980/cool/http%3A%2F%2Fsomewopisrcfileaddress/ws?WOPISrc=http://somewopisrcfileaddress&compat=/ws
    
    These 2 were redirected to different nodes when balancing
    was done based on WOPISrc
    
This is regression from commit https://github.com/CollaboraOnline/online/commit/a722687c116f81be19423fc92630cb7d7209b65a
browser: simplify getParameterByName

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1045,7 +1045,7 @@ window.app = {
 	global.makeDocAndWopiSrcUrl = function (root, docUrlParams, suffix, wopiSrcParam) {
 		var wopiSrc = '';
 		if (global.wopiSrc != '') {
-			wopiSrc = '?WOPISrc=' + global.wopiSrc;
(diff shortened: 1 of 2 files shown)


quikee (approved):
Seems fine

PR #8188, opened 2024-02-05
refactor: split out CSS image visiting.

mmeeks:
We use this for the richdocumentscode case, but it is more generally useful, this should be a pure re-factor.

Change-Id: I3a7f09d3da902a85dc2b362f7a160f07c90b97c6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -813,59 +813,68 @@ window.app = {
 		this.getSessionId();
 	};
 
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Looks useful for preloading NB icons.

mmeeks:
Sure there is another WIP commit for that =) let me re-base ..

PR #8189, opened 2024-02-05
Private/mmeeks/imageprefetch

mmeeks:
Thoughts appreciated =)


eszkadev:
`No cache API` logged for me in Chromium 120 and Firefox 115 (which should be supported: https://caniuse.com/mdn-api_cache ). Is that my problem with browser settings? (I think I have defaults in Chromium)

mmeeks:
Needs doing - but we need someone to chase this - and it's not me.

PR #8190, opened 2024-02-05
browser: impress: override invalidate preview tiles

hcvcastro:
Invalidate tile preview when canonical view id
changes.

Change-Id: I62d2e20946ec863ceb71014bb0216675316a658b

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -55,6 +55,7 @@ L.Control.PartsPreview = L.Control.extend({
 
 		map.on('updateparts', this._updateDisabled, this);
 		map.on('updatepart', this._updatePart, this);
+		map.on('invalidateparts', this._invalidateParts, this);
 		map.on('tilepreview', this._updatePreview, this);
 		map.on('insertpage', this._insertPreview, this);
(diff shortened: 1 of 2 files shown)


caolanm (commented):
This looks reasonable. 

I see there is an existing  _invalidateAllPreviews in browser/src/layer/tile/CanvasTileLayer.js which calls a _invalidatePreviews which has similar this._map.getPreview code in it.

Does that get called in this case and not do what we want for some reason? or not get called at all?

hcvcastro:
> I see there is an existing _invalidateAllPreviews in browser/src/layer/tile/CanvasTileLayer.js which calls a _invalidatePreviews which has similar this._map.getPreview code in it.
> 
> Does that get called in this case and not do what we want for some reason? or not get called at all?

It is a long history, at the beginning the only preview was called for Impress only (this._map.getPreview),
and it was created a Control.PartsPreview.js, then I guess Writer start requesting previews but they coded in
CanvasTileLayer.js. If you trace logs for _InvalidatePreviews in CanvasTileLater.js for Impress it does not 
execute any invalidate previews from  Control.PartsPreview.js.

caolanm (approved):
alright, lets not block on it.

PR #8191, opened 2024-02-05
Fix warning for unocommands.py

hfiguiere:
This resolves the following Python warning.
unocommands.py:319: SyntaxWarning: invalid escape sequence '\('
  f.write('''};

Also make sure the output depends on the script

Change-Id: I11f1e028c9c132f4afc66f9b5a6137b478f00774

* Resolves: # 
* Target version: master 

### Summary

I was seeing this warning. Escaping rules applied. I compared the output and it is unchanged. I run Python 3.12.1 in case this started to only pop on recent Python.

### TODO

- [ ] ...

### Checklist


caolanm (approved):
Looks sane to me, so lets approve it pending ci. But tag @timar in case I missing something. I imagine this is for stripping out trailing Latin character accelerators within brackets seen after CJK translations

hfiguiere:
backport #8249

PR #8194, opened 2024-02-06
wsd: rescue the clipboard before autosaving on unload

Ashod:
This moves the clipboard resque logic earlier to
make sure it is cached before the Kit process
exits.

By putting the clipboard-fetching command before
the save-before-exit, we should get the clipboard
in most-all cases. There are still edge-cases
that can slip through, but they should be much
less frequent, and they will require much more
involved logic (tbd).

Change-Id: I0eff95c28524002a6131fa0a57ef8e09ae21ca59

### DIFF
--- a/common/Clipboard.hpp
+++ b/common/Clipboard.hpp
@@ -145,16 +145,16 @@ public:
         const auto it = _cache.find(key);
         if (it == _cache.end())
         {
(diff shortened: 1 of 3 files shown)


eszkadev (approved):
Fixes the issue, tested with HAProxy

PR #8199, opened 2024-02-06
signal activity dumping: enrich with session details and state.

mmeeks:
Cue up some basic state for being signal safe dumped on crash/abort.

Change-Id: Ibc6713aef2a0e7b878b178b05f0e13c8d40b47fc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -795,6 +795,8 @@ public:
             LOG_DBG("Have " << _sessions.size() << " active sessions after creating "
                             << session->getId());
             LOG_INF("New session [" << sessionId << "]");
+
+            updateActivityHeader();
             return true;
         }
         catch (const std::exception& ex)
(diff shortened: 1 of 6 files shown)


kit/Kit.cpp
```diff
@@ -1346,6 +1348,17 @@ class Document final : public DocumentManagerInterface
         return _docPasswordType;
     }
 
+    void updateActivityHeader() const override
+    {
+        // pre-prepare and set details in case of a signal later
+        std::stringstream ss;
+        ss << "Session count: " << _sessions.size() << "\n";
+        for (auto const it : _sessions)
```
Ashod, line 1356: `for (const auto& it : _sessions)`

Ashod (approved):
Thanks @mmeeks. Looks good to me.

kit/Kit.cpp
```diff
@@ -1346,6 +1348,17 @@ class Document final : public DocumentManagerInterface
         return _docPasswordType;
     }
 
+    void updateActivityHeader() const override
+    {
+        // pre-prepare and set details in case of a signal later
+        std::stringstream ss;
+        ss << "Session count: " << _sessions.size() << "\n";
+        for (const auto& it : _sessions)
+            ss << "\t" << it.second->getActivityState() << "\n";
```
caolanm, line 1357: We have some places where we, like here, just deref "it.second" but then we have some other places like drainQueue and findSessionByViewId kit/Kit.cpp where we null check it.second.

Seems that the null check ones are just excessively cautious and it.second is always dereferenceable.

caolanm (approved):

PR #8201, opened 2024-02-06
cypress, desktop/writer/top_toolbar_spec.js, 'Insert Special Character.': add conditional 3rd click

vmiklos:
My local run failed in this test, because scrolling to the right just
twice resulted in a partially clicked toolbar item, so:

cy.cGet('#Home-container .unospan-CharmapControl').click();

failed. Add a 3rd click if scrolling to the right is still possible.

Change-Id: I4caa3dc79df1bbbd5bd0c7e88ef1baea8ae4b17e


neilguertincollabora:
Two other ways this could be done that are slightly more robust:
- Use click({force: true}) to click the charmap button even though it is not visible (and include a comment to explain why the force is necessary)
- Use waitUntil to perform the click action until the scroll bar is all the way to the right

I think you should try the force click at least, since it's easy and can make the test faster. But the triple click is fine too so I'm approving this so you can merge it if the force doesn't work.

neilguertincollabora (approved):

PR #8209, opened 2024-02-07
Reduce noise of somewhat excessive trace debug.

mmeeks:
Given the surrounding debug we should be able infer the same state from its absence.

Change-Id: Id20de40f67ed52dc5535585aec9b5888131d6324

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -841,7 +841,10 @@ public:
 #endif
         }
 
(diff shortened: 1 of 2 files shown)


caolanm (approved):
Looks alright to me. std::vector::empty wouldn't really gain anything over std::vector::size but is my habit.

PR #8213, opened 2024-02-07
return false early on 'null' origin

caolanm:
Change-Id: I9cdb26c3d2787beac6e2344d5f768e0fa6993bd8 (cherry picked from commit d1e00dbe1bfd59afef707673f2893e90d35e0d46) (cherry picked from commit c536e34f4ef59a5ecd05bf3d4bcfb7d91844e42d)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Why not back-port this; surprising that origin === 'null' though.

caolanm:
a literal "null" appears to be the thing https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin

> The serialization of an origin is the string obtained by applying the following algorithm to the given origin:
>    If origin is an [opaque origin], then return "null".

PR #8215, opened 2024-02-07
we never have an empty std::unique_ptr for .second

caolanm:
and don't check for it in the other cases, so drop it in these ones

Change-Id: Ie343178869cadc127ee898968fef8f798b6068d5

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks reasonable -- if the value is nullptr then probably something is already broken earlier and this check was just defensive programming.

PR #8218, opened 2024-02-08
wsd: detect unencoded WOPISrc

Ashod:
We should always have an encoded WOPISrc.
We add detection logic to make sure
that all URIs that contain WOPISrc have
it encoded properly. We do this by
comparing the decoded WOPISrc with
the original URI.

Change-Id: Ia0c2a79b009ce105321ad35db3d4f81006e81cb3


eszkadev:
Thanks, tested and it correctly detects regression fixed by: https://github.com/CollaboraOnline/online/pull/8183

eszkadev (approved):
Works as expected

PR #8221, opened 2024-02-08
WIP: global.js: convert to typescript

Minion3665:


eszkadev:
Having Window interface seems to be very useful to avoid all the any casting we have...
Can we for example extract only that interface for now? In smaller PR so it will be merged as safe non-functional change :)

Minion3665:
Superseded by #9996

PR #8224, opened 2024-02-08
Reenable logging socket messages before debug mode

neilguertincollabora:
Change-Id: I2452d4c7e3e3152533b676fb2db81ef95d96fb54

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Debug.js
+++ b/browser/src/core/Debug.js
@@ -24,6 +24,7 @@ L.DebugManager = L.Class.extend({
 	initialize: function(map) {
 		this._map = map;
 		this.debugOn = false;
+		this.debugNeverStarted = true;
 	},
 
 	toggle: function() {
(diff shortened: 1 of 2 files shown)


eszkadev:
Works fine, but does it enable automatically only for DEBUG builds? (--enable-debug)
I think before it was disabled by default in release.

neilguertincollabora:
I believe the DEBUG build check happens in _logSocket, which I have not touched. Compared to before the toggle was added in https://github.com/CollaboraOnline/online/pull/8121 there's only an extra condition checked, so it shouldn't have enabled any logging where there wasn't before.

That said, I'm having build problems right now and can't actually verify that right now.

eszkadev (approved):

PR #8225, opened 2024-02-08
Make browser/ clean before building

Minion3665:
We create our bundle.js from our browser/dist/ directory, but this is not cleaned before building. Normally this is not an issue but occasionally we have a file that stops being there from one build to another (maybe it got deleted or there was a checkout). In this case the file persists in dist/ and still makes it into the new bundle.js. This is horrible in CI, but generally an exciting footgun at all times, and should be promptly removed.

Change-Id: Iadb69c5164cbc99062c14c8ce9de0ea5aa5d5680

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -468,7 +468,7 @@ define global_file
 		@$(NODE) node_modules/uglify-js/bin/uglifyjs $< --output $@)
 endef
 
-.PHONY: build-cool
+.PHONY: build-cool mostlyclean-local
 
 all-local: build-cool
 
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Looks ok and Android passed now

eszkadev:
rebased, few times in row it failed with unit-integration, was not able to find file Rectangle.js

Minion3665:
> rebased, few times in row it failed with unit-integration, was not able to find file Rectangle.js

looks like this was due to a leftover `tscompile.done`... as a bonus this file would have been left around after a `make -C browser clean-local` before too so this commit fixes that as well :)

eszkadev (changes requested):
We can't do this on every `make run` to not lost additional content we copied into dist

Minion3665:
Superseded by a CI change, I'll make another PR to fix the issue I noticed with `make clean-local` keeping tscompile.done

PR #8227, opened 2024-02-08
es2015: bump eslint version

eszkadev:
- allow ES2015 in JS

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -927,12 +927,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (builder.wizard) {
 			var that = this;
 			var functionName = data.functionName;
-			$(rightDiv).click(function() {
+			$(rightDiv).click(() => {
 				builder.wizard.goLevelDown(mainContainer);
 				if (contentNode.onshow)
 					contentNode.onshow();
 			});
-			$(leftDiv).click(function() {
+			$(leftDiv).click(() => {
 				if (functionName !== '') {
 					app.socket.sendMessage('completefunction name=' + functionName);
 					that.map.fire('closemobilewizard');
@@ -1076,7 +1076,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		$(contentDiv).hide();
 		if (builder.wizard) {
-			$(sectionTitle).click(function() { builder.wizard.goLevelDown(mainContainer); });
+			$(sectionTitle).click(() => { builder.wizard.goLevelDown(mainContainer); });
 		} else {
 			window.app.console.debug('Builder used outside of mobile wizard: please implement the click handler');
 		}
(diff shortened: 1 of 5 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1824,7 +1824,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === false || data.enabled === 'false')
 			$(listbox).attr('disabled', 'disabled');
 
-		$(listbox).change(function() {
+		$(listbox).change(() => {
 			if ($(this).val())
 				builder.callback('combobox', 'selected', data, $(this).val()+ ';' + $(this).children('option:selected').text(), builder);
 		});
```
Minion3665, line 1830: what is the `this` referring to here? I worry that by changing this to an arrow function you may have altered its meaning

Minion3665 (commented):
From a quick look it seems fine except for the bit I commented on, which I'd need to test but looks to me like it could cause trouble...

browser/src/control/Control.JSDialogBuilder.js
eszkadev: ah, good catch. It is a jQuery callback where `this` was a `listbox` / target of the callback.
I think they do `.bind(listbox)` on the passed function object so it should work.
I checked it in Format -> Character -> Style. Seems to work.

eszkadev: good catch, fixed

Minion3665: can you bind arrow functions? I was under the impression that this isn't a thing you can do and from a quick test node seems to agree on this...
image

Minion3665: in contrast, here's what happens if you do the same thing without using an arrow function
image

Minion3665 (approved):

vmiklos:
Nice! :-)

Do you plan to go further (I'm especially waiting for await in ES2007), or that'll be not for 24.04? Thanks.

Minion3665:
> Nice! :-)
> 
> Do you plan to go further (I'm especially waiting for await in ES2007), or that'll be not for 24.04? Thanks.

@eszkadev I would also really like async/await... shall we do another?

Minion3665:
image
https://caniuse.com/async-functions

caniuse looks fine

vmiklos:
We discussed this on Thur and it should be OK to update to ES2017. However, when I try it with a trivial patch:

[await.txt](https://github.com/CollaboraOnline/online/files/14334256/await.txt)

then eslint fails like this:

> /home/vmiklos/git/collaboraonline/online/browser/src/map/Clipboard.js
>   719:21  error  Parsing error: Unexpected token blob
>
> ✖ 1 problem (1 error, 0 warnings)

@Minion3665 any idea what might be the problem? Thanks.

hfiguiere:
`await` is only valid in an `async` block

hfiguiere:
So you'd change

```
	_navigatorClipboardGetTypeCallback: function(clipboardContent, blob, type) {
```
to
```
	_navigatorClipboardGetTypeCallback: async function(clipboardContent, blob, type) {
```

vmiklos:
Thanks, see https://github.com/CollaboraOnline/online/pull/8319

PR #8228, opened 2024-02-08
annotation: refresh comments on accept/reject all tracked changes

lpranam:
problem:
tracked comments deletion was not reflected with accept/reject all changes

Change-Id: I0964fbd220f450cb144a115e8203250467e7edae

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -2283,10 +2283,10 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 						'type': 'toolbox',
 						'children': [
 							{
-								'id': 'review-accept-all-tracked-changes',
-								'type': 'toolitem',
+								'id': 'acceptalltrackedchanges',
+								'type': 'customtoolitem',
 								'text': _UNO('.uno:AcceptAllTrackedChanges', 'text'),
-								'command': '.uno:AcceptAllTrackedChanges',
+								'command': 'acceptalltrackedchanges',
 								'accessibility': { focusBack: true, combination: 'A2', de: 'A2' }
 							}
 						]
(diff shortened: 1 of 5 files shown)


lpranam:
requires: https://github.com/CollaboraOnline/online/pull/8127

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -2283,8 +2283,8 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 						'type': 'toolbox',
 						'children': [
 							{
-								'id': 'review-accept-all-tracked-changes',
-								'type': 'toolitem',
+								'id': 'acceptalltrackedchanges',
+								'type': 'menubartoolitem',
```
eszkadev, line 2287: should be `customtoolitem` so it will use `dispatch` not `menubar` code

eszkadev: see https://github.com/CollaboraOnline/online/issues/8029

eszkadev (changes requested):

browser/src/layer/tile/CommentListSection.ts
```diff
@@ -1992,6 +1992,16 @@ export class CommentSection extends CanvasSectionObject {
 			comment.onCommentDataUpdate();
 		}
 	}
+
+	public rejectAllTrackedCommentChanges(): void {
+		for (var i = 0; i < this.sectionProperties.commentList.length; i++) {
+			var comment = this.sectionProperties.commentList[i];
+			if (comment.sectionProperties.data.layoutStatus === 3) {
+				comment.sectionProperties.data.layoutStatus = 2;
```
eszkadev, line 2000: what are these magic numbers?
best to have them named by some static const in the class

lpranam: it is explained here in the comments: https://github.com/CollaboraOnline/online/blob/71a427c465c3b952f032b283cedc9e52ea985fcd/browser/src/layer/tile/CommentSection.ts#L69

but yes good to have some class

eszkadev:
Would be good to clearly use enum also when assigning the value:
https://github.com/CollaboraOnline/online/blob/71a427c465c3b952f032b283cedc9e52ea985fcd/browser/src/layer/tile/CommentListSection.ts#L1219

eszkadev (approved):

PR #8232, opened 2024-02-09
Replace dead chat links

Minion3665:
The libera.chat matrix bridge shut down a while ago, so links to our room on matrix load forever. Additionally, we use matrix rather than the IRC room so we should not be advertising it when they will only need to be redirected to the right place if they join.

I run a matrix homeserver (clicks.codes) and have set up a room address on there. By telling people this room address instead of the libera.chat one they can join the existing room even though libera.chat is not hosting matrix.

Change-Id: I8a4224f1a856e7295325196f1b1a80f860ed7962

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks good, thanks.

PR #8252, opened 2024-02-12
disable getting the host printer queue

caolanm:
and default printer. In typical deployment these should be effectively no no-op, so mostly aligns debugging env with deployment env so whatever printer happens to be installed locally has no effect.

Change-Id: I09651d9af2de9817d4d278d25937c9ce0dc87352

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Thanks =)

PR #8253, opened 2024-02-12
admin: send shutdown_received before shutting down

Rash419:
- added a new RunState::Migrate use this state when indirection_endpoint is configured
- indirection server send shutdown message once migration of unsaved documents is complete

Change-Id: I2606b24efcf3b1080ead35bb8535c11b4419e939

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -50,6 +50,7 @@ namespace
 enum class RunState : char
 {
     Run = 0, //< Normal up-and-running state.
+    DelayedShutDown, //< Delayed shutdown used with indirection_endpoint
     ShutDown, //< Request to shut down gracefully.
     Terminate //< Immediate termination.
 };
@@ -70,6 +71,7 @@ static bool UnattendedRun = false;
 #if !MOBILEAPP
 static int SignalLogFD = STDERR_FILENO; //< The FD where signalLogs are dumped.
 static char* VersionInfo = nullptr;
+static std::atomic<bool> IndirectionServerEnabled(false);
 static char FatalGdbString[256] = { '\0' };
 #endif
 
@@ -81,6 +83,8 @@ namespace SigUtil
 bool getShutdownRequestFlag() { return RunStateFlag >= RunState::ShutDown; }
 
 bool getTerminationFlag() { return RunStateFlag >= RunState::Terminate; }
+
+bool getDelayedShutdownRequestFlag() { return RunStateFlag >= RunState::DelayedShutDown; }
 
 void setTerminationFlag()
 {
@@ -96,6 +100,12 @@ void setTerminationFlag()
 void requestShutdown()
 {
     RunState oldState = RunState::Run;
+
+    if (getDelayedShutdownRequestFlag())
+    {
+        oldState = RunState::DelayedShutDown;
+    }
+
     if (RunStateFlag.compare_exchange_strong(oldState, RunState::ShutDown))
         SocketPoll::wakeupWorld();
 }
@@ -326,12 +336,20 @@ void requestShutdown()
         const auto onrre = errno; // Save.
 
         bool hardExit = false;
-        const char* domain;
-        RunState oldState = RunState::Run;
-        if ((signal == SIGINT || signal == SIGTERM) &&
-            RunStateFlag.compare_exchange_strong(oldState, RunState::ShutDown))
+        const char* domain = "";
+        RunState oldState = RunStateFlag;
+        if (signal == SIGINT || signal == SIGTERM)
         {
-            domain = " Shutdown signal received: ";
+            if (IndirectionServerEnabled && RunStateFlag == RunState::Run)
+            {
+                domain = " Delayed shutdown signal received: ";
+                RunStateFlag.compare_exchange_strong(oldState, RunState::DelayedShutDown);
+            }
+            else if (RunStateFlag == RunState::Run || RunStateFlag == RunState::DelayedShutDown)
+            {
+                domain = " Shutdown signal received: ";
+                RunStateFlag.compare_exchange_strong(oldState, RunState::ShutDown);
+            }
         }
         else
         {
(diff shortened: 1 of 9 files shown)


quikee (approved):
Looks fine

wsd/Admin.cpp
```diff
@@ -348,6 +348,7 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
         const std::string& dockey = tokens[2];
         const std::string& routeToken = tokens[3];
         const std::string& serverId = tokens[4];
+        const std::string& autoSaveAndStop = tokens[5];
```
Ashod, line 351: What happens where there are only 5 tokens? Why aren't we checking for the correct number of tokens (on line 345) ?

Ashod (dismissed):
Thanks @Rash419.

I can see that the admin has a new command that does the `autoSaveAndStop` on a document. But I don't see what the new `RunState::DelayedShutDown` state supposed to do. Is this PR incomplete? As it is, this new state is only sending "shutdown_received". But that could be done just the same with the existing `ShutDown` state. So why do we need a new state at all? 

The most serious issue in this PR is the config reading in the signal handler. We cannot do that. So that must be move out. I suspect that the new `RunState::DelayedShutDown` state is there only to detect the config has the "indirection_endpoint.url" value. But then that could be done in Admin or AdminModel::sendShutdownReceivedMsg(). If so, then we need to check for ShutDown and "indirection_endpoint.url" config in Admin and remove the new signal-handler code completely.

common/SigUtil.cpp
```diff
@@ -326,12 +334,20 @@ void requestShutdown()
...
-        const char* domain;
-        RunState oldState = RunState::Run;
-        if ((signal == SIGINT || signal == SIGTERM) &&
-            RunStateFlag.compare_exchange_strong(oldState, RunState::ShutDown))
+        const char* domain = "";
+        RunState oldState = RunStateFlag;
+        if (signal == SIGINT || signal == SIGTERM)
         {
-            domain = " Shutdown signal received: ";
+            if (!config::getString("indirection_endpoint.url", "").empty() && RunStateFlag == RunState::Run)
```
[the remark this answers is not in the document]
mmeeks: Rash - to do that - you need to inject the setting in via eg. the COOLWSD. That means we need to check a global variable inside the thing here, that is set from code that happens outside a signal handler. So eg. see SigUtil::setVersionInfo - add an accessor that sets the boolean you want and have a std::atomic global bool inside SigUtil.cpp for that flag.

wsd/Admin.cpp
mmeeks: This should be safe - the StringVector returns "" for un-defined tokens - please check; and ensure for back-compat the code paths work nicely with ""

mmeeks (commented):
Lets get these cleaned up =) but then looking good - we need it into COOL-next =)

common/SigUtil.cpp
```diff
@@ -326,12 +336,20 @@ void requestShutdown()
...
-        const char* domain;
-        RunState oldState = RunState::Run;
-        if ((signal == SIGINT || signal == SIGTERM) &&
-            RunStateFlag.compare_exchange_strong(oldState, RunState::ShutDown))
+        const char* domain = "";
+        RunState oldState = RunStateFlag;
+        if (signal == SIGINT || signal == SIGTERM)
         {
-            domain = " Shutdown signal received: ";
+            if (IndirectionServerEnabled.load() && RunStateFlag == RunState::Run)
```
mmeeks, line 343: no need for the load - just use it as a boolean.

common/SigUtil.cpp
```diff
@@ -481,6 +499,11 @@ void requestShutdown()
         VersionInfo = strdup(versionInfo.c_str());
     }
 
+    void setIndirectionServerEnabled(const bool enabled)
+    {
+        IndirectionServerEnabled.store(enabled);
```
mmeeks, line 504: similarly assign with =

kit/ForKit.cpp
```diff
@@ -746,6 +746,7 @@ int forkit_main(int argc, char** argv)
         const auto conf = std::getenv("COOL_CONFIG");
         config::initialize(std::string(conf ? conf : std::string()));
         EnableExperimental = config::getBool("experimental_features", false);
+        SigUtil::setIndirectionServerEnabled(!config::getString("indirection_endpoint.url", "").empty());
```
mmeeks, line 749: Do we want and/or need this ? should a Kit be getting a SIGTERM and doing something - I think not; can we drop that ?

kit/Kit.cpp
```diff
@@ -2869,6 +2869,7 @@ void lokit_main(
         // Already set by COOLWSD.cpp
         SigUtil::setFatalSignals("kit startup of " COOLWSD_VERSION " " COOLWSD_VERSION_HASH);
         SigUtil::setUserSignals();
+        SigUtil::setIndirectionServerEnabled(!config::getString("indirection_endpoint.url", "").empty());
```
mmeeks, line 2872: And this too - the processes should be managed by coolwsd surely?

wsd/COOLWSD.cpp
```diff
@@ -2365,6 +2365,9 @@ void COOLWSD::innerInitialize(Application& self)
 
     setenv("COOL_LOGLEVEL", LogLevel.c_str(), true);
 
+    bool isIndirectionServerEnabled = !getConfigValue<std::string>(conf, "indirection_endpoint.url", "").empty();
+    SigUtil::setIndirectionServerEnabled(isIndirectionServerEnabled);
```
mmeeks, line 2369: This one is good =)

wsd/AdminModel.cpp
```diff
@@ -1353,4 +1353,13 @@ void AdminModel::routeTokenSanityCheck()
     notify(oss.str());
 }
 
+void AdminModel::sendShutdownReceivedMsg()
+{
+    ASSERT_CORRECT_THREAD_OWNER(_owner);
+
+    std::ostringstream oss;
+    oss << "shutdown_received";
```
mmeeks, line 1361: Should be documented in protocol.txt I guess.

mmeeks (changes requested):
Lets abandon this approach; and instead warn the user that they are being migrated if we a route_token configured in JS instead of being shutdown; and consider how best we can add intelligence into coolwsd to pick good workloads to migrate as we shutdown first.

PR #8254, opened 2024-02-12
Fix &debug=true and &randomUser=true url parameters

neilguertincollabora:
Randomized settings had been getting overwritten by later setup. This change moves the randomization to after the other setup.

Move debug toggle to after docloaded message
Move sidebar toggle even later than that

Change-Id: Ib72b5ef4c2c31f7707a384b2d819d5cde57a796d

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/core/Debug.js
+++ b/browser/src/core/Debug.js
@@ -16,6 +16,8 @@
  * - Ctrl+Shift+Alt+D (Map.Keyboard.js _handleCtrlCommand)
  * - Help > About > D (Toolbar.js aboutDialogKeyHandler)
  * - Help > About > Triple Click (Toolbar.js aboutDialogClickHandler)
+ * - &debug=true URL parameter (Map.js initialize)
+ * - &randomUser=true URL parameter (global.js)
  */
 
 /* global app L _ */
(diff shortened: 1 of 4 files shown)


browser/src/core/Debug.js
```diff
@@ -515,13 +517,28 @@ L.DebugManager = L.Class.extend({
 		}
 
 		// Toggle sidebar
+		if (!this._map._docLoadedOnce) {
+			// When first opening the document, initializeSidebar is called
```
mmeeks, line 521: good catch =) I hate 200ms delays though - possibly we should rework the sidebar delay logic; and/or put this in initializeSidebar itself ? =) I'm not optimistic about racing timers on loaded machines.

browser/src/core/Debug.js
```diff
@@ -515,13 +517,28 @@ L.DebugManager = L.Class.extend({
...
+			// 200ms after setup, which would overwrite our randomization.
+			// So in this case, wait for sidebar initialization and the
+			// response to complete so that we know the current state
+			setTimeout(this._randomizeSidebar.bind(this), 1000);
+		} else {
+			this._randomizeSidebar();
+		}
+
+		this._painter.update();
```
mmeeks, line 530: Does that do what you think it does? still no harm if it's only on a debug path.

browser/src/map/Map.js
```diff
@@ -154,6 +154,11 @@ L.Map = L.Evented.extend({
 		this._progressBar = L.progressOverlay(new L.point(150, 25));
 
 		this._debug = new L.DebugManager(this);
+		this.on('docloaded', function() {
```
mmeeks, line 157: if possible may be nicer to hook into an existing handler rather than having the "my timeout is bigger than yours" approach etc. Then we know what order handlers are executed in and so on; but not read carefully =)

mmeeks (approved):
Looks reasonable; not a fan of racing timers.

PR #8259, opened 2024-02-13
jsdialog: handle disable action on mobile

eszkadev:
- unify desktop and mobile toolbox code
- toolbox should apply disabled state on child nodes
- this is needed to handle disable action, previously we received JSON for complete widget


eszkadev:
requires: https://gerrit.libreoffice.org/c/core/+/163261

caolanm (approved):

PR #8263, opened 2024-02-13
jsdialog: disabled state

eszkadev:
Make sure all widgets handle "disabled" state. Improve styling.

Requires: https://gerrit.libreoffice.org/c/core/+/163261


gokaysatir (approved):
Checked & tested.

PR #8264, opened 2024-02-13
disable file renaming in readonly mode

codewithvk:
Change-Id: Iebfd5fd57a322b745a5a4791d48071473dcccc49

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (commented):
looks ok, but best to ask @Ashod who knows the code better

wsd/Storage.cpp
```diff
@@ -908,6 +908,13 @@ WopiStorage::WOPIFileInfo::WOPIFileInfo(const FileInfo& fileInfo,
     (void)uriObject;
 #endif
 
+    // disable file renaming in readonly mode.
+    if (!_userCanWrite)
+    {
+        _userCanNotWriteRelative = true;
+        _userCanRename = false;
```
eszkadev, line 915: is that setting per user or per file?
(it's important, if we have 2 users and one of them has edit permission, we should still allow that)
(maybe we should check user permission when command to rename / save as comes to the server?)

eszkadev (approved):

PR #8267, opened 2024-02-14
Private/gokay/readonly mode

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -370,7 +370,7 @@ L.Map.Keyboard = L.Handler.extend({
 	// any 'beforeinput', 'keypress' and 'input' events that would add
 	// printable characters. Those are handled by TextInput.js.
 	_onKeyDown: function (ev) {
-		if (this._map.uiManager.isUIBlocked() || (this._map.isReadOnlyMode() && !this.readOnlyAllowedShortcuts(ev))
+		if (this._map.uiManager.isUIBlocked()
 			|| ((this._map._docLayer._docType === 'presentation' || this._map._docLayer._docType === 'drawing') && this._map._docLayer._preview.partsFocused === true)
 		)
 			return;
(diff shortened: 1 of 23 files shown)


gokaysatir:
Requires below changes:
https://gerrit.libreoffice.org/c/core/+/163430

wsd/ClientSession.cpp
```diff
@@ -1120,7 +1120,7 @@ bool ClientSession::_handleInput(const char *buffer, int length)
         if (tokens.equals(0, "key"))
             _keyEvents++;
 
-        if (isEditable() && COOLProtocol::tokenIndicatesDocumentModification(tokens))
+        if (isEditable() && COOLProtocol::tokenIndicatesDocumentModification(tokens)) // check this function: tokenIndicatesDocumentModification
```
eszkadev, line 1123: forgotten todo? what is that comment about?

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -370,7 +370,7 @@ L.Map.Keyboard = L.Handler.extend({
 	// any 'beforeinput', 'keypress' and 'input' events that would add
 	// printable characters. Those are handled by TextInput.js.
 	_onKeyDown: function (ev) {
-		if (this._map.uiManager.isUIBlocked() || (this._map.isReadOnlyMode() && !this.readOnlyAllowedShortcuts(ev))
+		if (this._map.uiManager.isUIBlocked()
```
eszkadev, line 373: We removed this read-only filter for keyboard. What do we have now?
User cannot open any dialog using keyboard shortcut for example?

gokaysatir: Can you specify please? Keyboard events are filtered on the core side. We are planning to add some more checks but this should be working by now. If i missed anything, an example would work better for me.

wsd/ClientSession.cpp
gokaysatir: Yep, let me check why i did it, i will probably remove that comment :)

browser/src/map/handler/Map.Keyboard.js
eszkadev: no, just asking if we protect user from doing something by keyboard shortcuts. If it is filtered on core side - even better (it's always possible to emit something manually on the wire).

gokaysatir: Ah okay thanks.

eszkadev (approved):
looks fine then

mmeeks (approved):
given that master doesn't compile without this and it's been through some review cycles; lets get it in =)

PR #8281, opened 2024-02-15
globals: Make sure coolParams.get() return an empty string instead of null

hfiguiere:
This fixes a regression where `framed.doc.html` wouldn't load.

Change-Id: I57b4a617755be4615ec6ec734f6e084d57eaebce

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -80,7 +80,19 @@ window.app = {
 
 	global.setLogging(global.coolLogging != '');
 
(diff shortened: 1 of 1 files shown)


hfiguiere:
rebased

removed ES6 to backport this more easily.

eszkadev (approved):

PR #8296, opened 2024-02-16
getComputedStyle is very slow, but we don't need it for LTR comments

caolanm:
but it gets computed anyway, move it so its only called if actually needed

Change-Id: I187540df6def6b4f92199e273665e04ed493035a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
This is rather curious. Most of the comments are invisible / hidden and show up only as (incredibly ugly) red rectangles =) Why do we need to compute their positions in any way ?

PR #8298, opened 2024-02-16
calc: calculate correct position when scrolling to comment

eszkadev:
In calc all comments are hidden and only shown when mouse hovers the cell to which the comment is attached. Anchor was always a wrong value so when we were at some more distant position - it always jumped to random place on hover.

This is regression from:
commit 6550e713c4bf687931aa721f36d0a28ac70e94f0
annotation: calc scroll vertical to make selected comment visible which was followup for:
commit 2bf93e34c349cbce3bfb70c5b92aed61b749101a
annotation: writer scroll to make selected comment visible

We need to compare the same type of positions. Now all the values will be in core pixels and methods have comment which indicates that.

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -686,6 +686,10 @@ export class CommentSection extends CanvasSectionObject {
 			this.sectionProperties.commentList[i].sectionProperties.container.style.display = 'none';
 	}
 
+	private cssToCorePixels(cssPixels: number) {
+		return cssPixels * app.dpiScale;
+	}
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
Thanks for the fix. We can move the Calc case into its own function when we have time. Looks like select function is getting complicated.

eszkadev:
> Thanks for the fix. We can move the Calc case into its own function when we have time. Looks like select function is becoming complicated.

Thank you too. Sure we can work later on improving the code. I will try also to introduce test, but I'm not sure if hover is easy to reproduce.

PR #8301, opened 2024-02-16
wsd: Blank stored timestamp if an up-load times out.

mmeeks:
Quite possibly we did successfully up-load, or at least our timestamp may not match, but we cannot be sure enough to warn the user in this case. We cannot reliably get notified of remote alteration on the server of files, so wait until a more succesful save or new user join to get an improved timestamp.

Change-Id: I5c8124ed74ba5c26371768ac778b9670282c87a3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Didn't get to the unit test (typical) - just poking CI to see what's up here; it's not happy with the same code vs. 24.04 - perhaps there is some problem there ...

mmeeks:
Would be good to get this in for this evening's release ...

caolanm (approved):

PR #7124, opened 2023-08-21
browser: simplify logServer

hcvcastro:
Use "fetch()" function to save logging client data
to server side.

Change-Id: I8b6a94b7ad7ae02ca00666759e3695867df4d372
Signed-off-by: Henry Castro <hcastro@collabora.com>

PR #8302, opened 2024-02-18
Revert "Remove unused file."

plubius:
This reverts commit c69e14a5ddf25dd8ba8494008b112e20f839eacd.

The above commits causes document loading in the iOS app to always fail. Even the animated loading graphic never appears.

I don't know why removing that .js file causes this behavior, but I did see the following in the trace messages in Xcode.

  To Online: load url=null accessibilityState=false lang=en-US

It appears that the JS code is sending a null URL to load.

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -169,6 +169,7 @@ NODE_MODULES_JS =\
 	node_modules/fzstd/umd/index.js
 
 COOL_LIBS_JS =\
+	ResizeObserverPolyfill.js \
 	Autolinker.js \
 	select2.js \
 	sanitize-url.js
--- a/browser/js/ResizeObserverPolyfill.js
+++ b/browser/js/ResizeObserverPolyfill.js
@@ -1,0 +1,944 @@
+/* eslint-disable */
+
+/* Source:
+	https://github.com/que-etc/resize-observer-polyfill/blob/master/dist/ResizeObserver.js
+*/
+
+if (typeof ResizeObserver === 'undefined') {
+    (function (global, factory) {
+        typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
+        typeof define === 'function' && define.amd ? define(factory) :
+        (global.ResizeObserver = factory());
+    }(this, (function () { 'use strict';
+
+        /**
+         * A collection of shims that provide minimal functionality of the ES6 collections.
+         *
+         * These implementations are not meant to be used outside of the ResizeObserver
+         * modules as they cover only a limited range of use cases.
+         */
+        /* eslint-disable require-jsdoc, valid-jsdoc */
+        var MapShim = (function () {
+            if (typeof Map !== 'undefined') {
+                return Map;
+            }
+            /**
+             * Returns index in provided array that matches the specified key.
+             *
+             * @param {Array<Array>} arr
+             * @param {*} key
+             * @returns {number}
+             */
+            function getIndex(arr, key) {
+                var result = -1;
+                arr.some(function (entry, index) {
+                    if (entry[0] === key) {
+                        result = index;
+                        return true;
+                    }
+                    return false;
+                });
+                return result;
+            }
+            return /** @class */ (function () {
+                function class_1() {
+                    this.__entries__ = [];
+                }
+                Object.defineProperty(class_1.prototype, "size", {
+                    /**
+                     * @returns {boolean}
+                     */
+                    get: function () {
+                        return this.__entries__.length;
+                    },
+                    enumerable: true,
+                    configurable: true
+                });
+                /**
+                 * @param {*} key
+                 * @returns {*}
+                 */
+                class_1.prototype.get = function (key) {
+                    var index = getIndex(this.__entries__, key);
+                    var entry = this.__entries__[index];
+                    return entry && entry[1];
+                };
+                /**
+                 * @param {*} key
+                 * @param {*} value
+                 * @returns {void}
+                 */
+                class_1.prototype.set = function (key, value) {
+                    var index = getIndex(this.__entries__, key);
+                    if (~index) {
+                        this.__entries__[index][1] = value;
+                    }
+                    else {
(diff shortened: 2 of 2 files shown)


gokaysatir:
Hello @plubius, that file was added for IE11 compatibility. Once our support for that browser ended, we didn't need that file any more. It's really interesting to see that. A dump question but can there be anything with your local? Some cached files etc? If not, happy to approve the revert the changes of course.

plubius:
> Hello @plubius, that file was added for IE11 compatibility. Once our support for that browser ended, we didn't need that file any more. It's really interesting to see that error. A dump question but can there be anything with your local? Some cached files etc? If not, happy to approve to revert the changes of course.

I don't think we will need to revert the patch. I opened this PR just to note what I found in my latest iOS app build.

I'll debug further when I have some time, but for some unknown reason, removing that one .js file causes JavaScript to stop sending messages to the server. Problem is that I have been unable to connect the Safari debugger after opening a document to catch where the JS failure is. This is because the iOS app creates a new native window and adds a new WebKit instance to that window when opening document.

Background: the iOS app is all local. The server and the browser (really one WebKit view per document) run in one controlling local process.  Since the server code is just another thread in the app, the JS sends messages to the server code by writing to some standard WebKit JS variables instead of making an HTTP request. You can this with all of the "if (window.ThisIsTheiOSApp)" JS code.

I suspect that one of these "if (window.ThisIsTheiOSApp)" is now needed somewhere in the JS code. My first guess is that one of the following are causing this:

- The JS code is trying to make an HTTP request to a non-existent server instead of writing a message in an "if (window.ThisIsTheiOSApp)" block.
- The JS code is still trying to load the ResizeObserver class. I grep'd the browser/src directories for ResizeObserver and only found one case of this but wrapping it in a "if (!window.ThisIsTheiOSApp)" block didn't change anything.

I'll update if I find anything.

eszkadev:
@plubius recently we had some regression wrt connecting to the server due to https://github.com/CollaboraOnline/online/pull/7124
worth checking

gokaysatir:
Okay, thanks for the insights :)

plubius:
After some debugging, I am still convinced that the iOS app (and maybe the Android app) need to have the ResizeObserver class defined somewhere. I haven't been able to verify this, but the fact that adding back browser/js/ResizeObserverPolyfill.js fixes the bug makes me think that the regular Collabora Online is loading the ResizeObserver class from somewhere and the iOS app doesn't have access to the class.

I still see two places in the JS code that calls "new ResizeObserver()" so my question is where does Collabora Online expect this class to be defined? Maybe the iOS and Android builds just aren't copying the required files into browser/dist?

plubius:
@Tex2002ans Does the Android app nightly builds suffer from this bug? In the iOS app, I see all JavaScript activity stop once the empty bar with the close icon is drawn. The animated loading graphic never appears: Screenshot 2024-02-25 at 2 45 13 PM

plubius:
OK. Instead of reverting the entire patch, I reverted but deleted everything special from browser/js/ResizeObserverPolyfill.js so that it is just a simple "if (typeof ResizeObserver === 'undefined')" test (see below). WebKit has a ResizeObserver class per https://webkit.org/blog/9997/resizeobserver-in-webkit/ but apparently we have to force it to loaded on iOS? Weird, but when using my simplified browser/js/ResizeObserverPolyfill.js (see below), everything works again *and* the app.socket.sendMessage() call never gets executed.

Seems like some weird JS class loading behavior in at least WebKit. Is anyone seeing this with Collabora Online with a Safari client browser?:

% cat browser/js/ResizeObserverPolyfill.js
/* eslint-disable */

/* Source:
	https://github.com/que-etc/resize-observer-polyfill/blob/master/dist/ResizeObserver.js
*/

if (typeof ResizeObserver === 'undefined') {
    app.socket.sendMessage('In custom ResizeObserver loading code');
}

plubius:
> Seems like some weird JS class loading behavior in at least WebKit. Is anyone seeing this with Collabora Online with a Safari client browser?:

Disclaimer: I have no knowledge about JS class loading. Seems weird that the WebKit has a built-in ResizeObserver *but* we must force it to load for the few "new ResizeObserver()" calls that Collabora's JS makes. I know very little about the JS runtime, but shouldn't "new ResizeObserver()" be enough to load the class?

Maybe some JS experts can make sense of this (to me) very unexpected behavior in WebKit. Maybe I am missing something obvious.

gokaysatir:
Hello @plubius, thank you for digging the issue. As you noticed, the browser API should have the ResizeObserver class, and we shouldn't need to for it manually - i agree there are some weird things there.

Here is a browser compatibility page for the class:
* https://caniuse.com/resizeobserver

Maybe Safari version you have doesn't have that class or may be it's experimental or the webkit you use doesn't have that feature - though i checked the link you shared.

I don't know IOS side. Since you confirmed that there is no direct solution, let me approve the changes. We can try this removal again, after we are sure that it works everywhere :)

Thank you.

gokaysatir (approved):

plubius:
> I don't know the IOS side. Since you confirmed that there is no direct solution, let me approve the changes. We can try this removal again, after we are sure that it works everywhere :)

Thank you for approving this. After committing, the iOS app build with the master branch now loads documents on both my iPad and the emulator (i.e. Mac Catalyst) on my Mac.

Tex2002ans:
> Does the Android app nightly builds suffer from this bug? In the iOS app, I see all JavaScript activity stop once the empty bar with the close icon is drawn.

@plubius This nightly APK was broken with the "white screen":

- collabora-office-mobile-23.05-snapshot-arm64-v8a-**2024-02-18.apk**

This one was okay:

- 2024-02-04 APK

I haven't had the chance to test any others + I don't see any newer APKs built yet.

- - -

**Side Note:** I wrote little more info here (when a user first reported other issues with Android):

- https://www.reddit.com/r/libreoffice/comments/1aw0ys1/collabora_office_is_really_buggy_in_android/krgi1so/

PR #8307, opened 2024-02-19
zotero: fix missing field data in the dialog (backport)

lpranam:
problem:
in some types of items fields may be named differently, leading to fields not being found and empty dialog entries.

currently only itemType 'case' was culprit but in future there maybe some other types too

Change-Id: I24b58dc474a555b3f26c34bb4ae0b144c06abc57

### Checklist


eszkadev (approved):

Rash419 (approved):
Looks good

PR #8310, opened 2024-02-19
Show tooltips over the cell while typing a function in a cell.

GulsahKose:
Change-Id: I85354a6c17947824408ae0458f29b858f31ab0e3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1017,26 +1017,34 @@ L.Control.UIManager = L.Control.extend({
 		this.map.fire('hidebusy');
 	},
 
-	// Document area tooltip
+	// Calc function tooltip
 
-	/// Shows tooltip in the document area
-	/// tooltipInfo contains rectangle (position in twips) and text properties
-	/// elem has to be jQuery selector output eg. $('.leaflet-layer')
-	showDocumentTooltip: function(tooltipInfo, elem) {
-		var split = tooltipInfo.rectangle.split(',');
-		var latlng = this.map._docLayer._twipsToLatLng(new L.Point(+split[0], +split[1]));
-		var pt = this.map.latLngToContainerPoint(latlng);
+	/// Shows tooltip over the cell while typing a function in a cell.
+	/// tooltipInfo contains possible function list. If you type a valid
+	/// function it'll show the usage of the function. elem has to be
+	/// jQuery selector output eg. $('.leaflet-layer')
+	showDocumentTooltip: function(tooltipInfo, elem, pos) {
+		var pt = this.map.latLngToContainerPoint(pos);
+		pt.y -=35; //Show tooltip above the cursor.
 
-		elem.tooltip();
-		elem.tooltip('enable');
-		elem.tooltip('option', 'content', tooltipInfo.text);
-		elem.tooltip('option', 'items', elem[0]);
-		elem.tooltip('option', 'position', { my: 'left bottom',  at: 'left+' + pt.x + ' top+' + pt.y, collision: 'fit fit' });
-		elem.tooltip('open');
-		document.addEventListener('mousemove', function() {
-			elem.tooltip('close');
-			elem.tooltip('disable');
-		}, {once: true});
+		if ($('.ui-tooltip').length > 0) {
+			elem.tooltip('option', 'content', tooltipInfo);
+		}
+		else {
+			elem.tooltip({
+				tooltipClass: 'functiontooltip',
+				content: tooltipInfo,
+				items: elem[0],
+                position: { my: 'left top', at: 'left+' + pt.x +  ' top+' +pt.y, collision: 'fit fit' }
+			});
+			elem.tooltip('option', 'customClass', 'functiontooltip');
+			elem.tooltip('open');
+			elem.off('mouseleave');
+		}
+	},
+
+	hideDocumentTooltip: function(elem) {
+		elem.tooltip('option', 'disabled', true);
 	},
 
 	// Snack bar
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1158,10 +1158,6 @@ app.definitions.Socket = L.Class.extend({
 			// Switching modes.
 			window.location.reload(false);
 		}
-		else if (textMsg.startsWith('tooltip: ')) {
-			var tooltipInfo = JSON.parse(textMsg.substring(textMsg.indexOf('{')));
-			this._map.uiManager.showDocumentTooltip(tooltipInfo, $('.leaflet-layer'));
-		}
 		else if (!textMsg.startsWith('tile:') && !textMsg.startsWith('delta:') &&
 			 !textMsg.startsWith('renderfont:') && !textMsg.startsWith('windowpaint:')) {
 
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1772,6 +1772,9 @@ L.CanvasTileLayer = L.Layer.extend({
 		else if (textMsg.startsWith('calcfunctionlist:')) {
 			this._onCalcFunctionListMsg(textMsg.substring('calcfunctionlist:'.length + 1));
 		}
+		else if (textMsg.startsWith('tooltip:')) {
+			this._onCalcFunctionUsageMsg(textMsg.substring('tooltip:'.length + 1));
+		}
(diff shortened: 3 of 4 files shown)


GulsahKose:
Patch needs https://gerrit.libreoffice.org/c/core/+/163594 on core side. That is the reason of the fail.

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2539,6 +2555,59 @@ L.CanvasTileLayer = L.Layer.extend({
...
+
+		var remainingFuncCount = suggestionArray.length - maxSuggestion;
+		if (remainingFuncCount > 0)
+			resultText = resultText + ' and ' + remainingFuncCount + ' more';
+
+		resultText = resultText + ' : ' + currentFuncDescription;
+
+		text.innerText = resultText;
+
+		this._map.functionListPopup = new L.Popup({className: 'functionlist-popup', closeButton: false, closeOnClick: false, autoPan: false})
```
eszkadev, line 2594: We are trying to get rid of Leaflet. We should use own component, not `L.Popup`.
Could you try `showDocumentTooltip` created by Mike? https://github.com/CollaboraOnline/online/blob/0431c584653d49a66d3b85130ab8aa32e1374517/browser/src/control/Control.UIManager.js#L1025

browser/src/core/Socket.js
```diff
@@ -1158,10 +1158,9 @@ app.definitions.Socket = L.Class.extend({
 			// Switching modes.
 			window.location.reload(false);
 		}
-		else if (textMsg.startsWith('tooltip: ')) {
-			var tooltipInfo = JSON.parse(textMsg.substring(textMsg.indexOf('{')));
-			this._map.uiManager.showDocumentTooltip(tooltipInfo, $('.leaflet-layer'));
-		}
+		//else if (textMsg.startsWith('tooltip: ')) {
```
eszkadev, line 1161: please remove it completely, showDocumentTooltip was changed so it might be confusing that uncommenting will work

browser/src/control/Control.UIManager.js
```diff
@@ -1022,21 +1022,23 @@ L.Control.UIManager = L.Control.extend({
...
-		elem.tooltip('option', 'items', elem[0]);
-		elem.tooltip('option', 'position', { my: 'left bottom',  at: 'left+' + pt.x + ' top+' + pt.y, collision: 'fit fit' });
-		elem.tooltip('open');
-		document.addEventListener('mousemove', function() {
-			elem.tooltip('close');
-			elem.tooltip('disable');
-		}, {once: true});
+	showDocumentTooltip: function(tooltipInfo, elem, pos) {
+        var pt = this.map.latLngToContainerPoint(pos);
+        pt.y -=35; //Show tooltip above the cursor.
```
eszkadev, line 1027: 35 is standard cell height? we will need to adjust that:
- with real cursor height
- with DPI of a user

(We can do it later, in the followup)

eszkadev:
Please fix formatting: unified tabs / spaces with rest of the file. Some lines match some not.

browser/src/control/Control.UIManager.js
GulsahKose: It would be nice to fix it as followup.

GulsahKose:
Szymon actualy need a suggestion for 2 topic.
1-) Currently tooltip is open and it can be updated but i can not decide when we should close the tooltip. Which event should we listen on leaflet
2) Tooltip style comes from jquery lib, It defined in .ui-tooltip class. I need to modify(For example we should show tip in 1 line)  But I can not add a specific class name to our tooltip. I tried the suggestions here  but doesn't work. [AN idea  ](https://stackoverflow.com/questions/13160357/adding-custom-class-to-tooltip)
Any suggestion here?

pedropintosilva:
I also couldn't get the tooltip to show the additional help with regards to parameters (left core, right online):
https://github.com/CollaboraOnline/online/assets/65948705/fb4e95c2-01b3-485b-9f40-0d88654b2d4e

GulsahKose:
Pedro I'll easily implement the second type tooltip(includes parameters).  Now struggling with the tooltip behaviour. I'll try to close and open tooltip on time.

browser/src/control/Control.UIManager.js
```diff
@@ -1022,21 +1022,30 @@ L.Control.UIManager = L.Control.extend({
...
+				content: tooltipInfo,
+				items: elem[0],
+                position: { my: 'left top', at: 'left+' + pt.x +  ' top+' +pt.y, collision: 'fit fit' }
+			});
+			elem.tooltip('open');
+			elem.off('mouseleave');
+		}
+
+		document.addEventListener('keydown', function closeTooltip(e) {
+			if (e.key === 'Escape') {
```
pedropintosilva, line 1043: I think it would be good to add the edge browser case so, `e.key === 'Escape' || e.key === 'Esc'` also this doesn't work. It seems other keys are being detected (ctrl etc) but escape is not. I wonder if somehow it being caught earlier by the exit cell edit mode or some such , what do you think @eszkadev ?

pedropintosilva (changes requested):

browser/src/control/Control.UIManager.js
pedropintosilva: or maybe it's been directly send as keyboard events on the document via `postKeyboardEvent`

GulsahKose: @pedropintosilva @eszkadev I'll try to send hide message from core. I hope it will work.

browser/src/control/Control.UIManager.js
```diff
@@ -1022,21 +1022,30 @@ L.Control.UIManager = L.Control.extend({
...
+				elem.tooltip('close');
+				elem.tooltip('disable');
+				document.removeEventListener('keydown', closeTooltip);
```
eszkadev, line 1046: removing callback can be done using { once: true } parameter, look in addEventListener documentation

browser/src/control/Control.UIManager.js
eszkadev: maybe we can get this by: `editorGotFocus`
```
this.map.on('editorgotfocus'
```

It happens when we change cell, lost focus from editing area etc... It seems to work for me, but more testing would be useful

eszkadev: Also can we try to always create new tooltip? not modify existing one? I fear the position will be out of sync, when we will be shoing other popups in the future

GulsahKose:
I've updated core master patch too and currently we have better tooltip behaviour. But I'll update the patch again for the parameter tooltip and some style improvements on tooltip balloon.

GulsahKose:
With updated core patch https://gerrit.libreoffice.org/c/core/+/163651 I think I've completed the feature. Here is the screen record. 

https://github.com/CollaboraOnline/online/assets/2954981/7fe197e7-a5ce-4e72-805c-b121e941204f

browser/src/control/Control.UIManager.js
```diff
@@ -1022,21 +1022,28 @@ L.Control.UIManager = L.Control.extend({
...
-		elem.tooltip('enable');
-		elem.tooltip('option', 'content', tooltipInfo.text);
-		elem.tooltip('option', 'items', elem[0]);
-		elem.tooltip('option', 'position', { my: 'left bottom',  at: 'left+' + pt.x + ' top+' + pt.y, collision: 'fit fit' });
-		elem.tooltip('open');
-		document.addEventListener('mousemove', function() {
-			elem.tooltip('close');
-			elem.tooltip('disable');
-		}, {once: true});
+	showDocumentTooltip: function(tooltipInfo, elem, pos) {
```
eszkadev, line 1025: please update the comment above function so it will contain true information

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1964,27 +1967,45 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._map.fire('cellformula', {formula: formula});
 	},
 
+	_onCalcFunctionUsageMsg: function (textMsg) {
```
eszkadev, line 1970: in the followup PR we should move all these `_onCalcXXXXXX` functions into `_onMessage` in CalcTileLayer, as they are used only in Calc.

eszkadev (commented):
code is ok

pedropintosilva:
it works as expected with the latest https://gerrit.libreoffice.org/c/core/+/163594

pedropintosilva (approved):

PR #8311, opened 2024-02-19
calc: fix header context menu on touch devices

Minion3665:
Previously we did not register or trigger the context menu when long-pressing on the headers in calc on mobile devices. This commit registers and activates the context menu when we detect a longpress

To avoid harming devices that are both touchscreen and mouse we do not unregister the context menu after activating it (as on mouse-devices we register/unregister the context menu based on the position of the mouse). I believe this to never actually cause an issue however it's plausible that triggering the context menu again on an area that doesn't have a context menu could cause a context menu for a header to show. I still think this is significantly better than the status-quo

Change-Id: Ib5e9cbdb3d7e10ed58396e8559ab2230f2b76c4c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasSectionContainer.ts
+++ b/browser/src/layer/tile/CanvasSectionContainer.ts
@@ -576,6 +576,18 @@ class CanvasSectionContainer {
 		this.canvas.ontouchend = this.onTouchEnd.bind(this);
 		this.canvas.ontouchcancel = this.onTouchCancel.bind(this);
 
+		if (typeof Hammer !== 'undefined') {
+			// Mocha does not provide window but hammerjs requires it. As a workaround let's check if hammer exists before trying to create an instance
+			(new Hammer(this.canvas, {recognizers: [[Hammer.Press]]}))
+				.on('press', (window as any /* TODO: remove cast after gh#8221 */).touch.touchOnly((e: any) => {
+					const syntheticEvent = new MouseEvent('contextmenu', {
+						clientX: e.srcEvent.clientX,
+						clientY: e.srcEvent.clientY
+					});
+					this.canvas.dispatchEvent(syntheticEvent);
+				}));
+		}
+
 		// Some explanation first.
 		// When the user uses the mouse wheel for scrolling, different browsers use different technics for calculating the deltaY and deltaX values.
 		// For example FireFox uses "deltaMode=1" which corresponds to "lines". So it creates the event with the number of lines to scroll.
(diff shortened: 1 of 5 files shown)


browser/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -576,6 +576,18 @@ class CanvasSectionContainer {
 		this.canvas.ontouchend = this.onTouchEnd.bind(this);
 		this.canvas.ontouchcancel = this.onTouchCancel.bind(this);
 
+		if (typeof Hammer !== 'undefined') {
+			// Mocha does not provide window but hammerjs requires it. As a workaround let's check if hammer exists before trying to create an instance
+			(new Hammer(this.canvas, {recognizers: [[Hammer.Press]]}))
```
eszkadev, line 581: can we not use hammerjs?
there is `onLongPress` handler for canvas sections, can we use it?

Minion3665: So, @eszkadev, I considered this... unfortunately I didn't manage to get a longpress working with this handler - it seemed it wasn't called. As it's not a DOM event and I can't see anywhere it's called except via `onMouseMove` (which isn't triggered on touch) I presumed it's not for this and wrote it with hammer instead... do you believe this to be an incorrect assumption?

We could write a standard Javascript solution, but I wonder if it's worth it... do you think there is significant benefit to keeping hammer.js away from here given that we depend on it in online already? I guess maybe for mocha, but beyond that I'm not so sure

eszkadev: @gokaysatir is author of canvas section, maybe we need some trick to make it working. can you help here?

We tried to not rely on hammer here (in sections) as I know. Not introduce many new handlers fighting with each other, but rather create set of event handlers which propagate properly events. I think sections can be stacked on each other so it's important.

eszkadev: What is the status here @Minion3665 @gokaysatir ? Do we have replacement for hammerjs ?

gokaysatir: There was a missing commit there. Now it's ready to be tested again.

gokaysatir: I tested it on real device and it seemed to work without hammer.js. Using mostly current implementation of canvasSectionContainer.

Minion3665:
See instead #8510

PR #8312, opened 2024-02-19
globals: Make sure coolParams.get() return an empty string instead of…

hfiguiere:
* Resolves: #  backport of #8281 
* Target version: 23.05 

### TODO

- [ ] ...

### Checklist


hfiguiere:
Tested locally on 23.05

caolanm (approved):
looks reasonable to me, but I think we need to reviewers on this branch now

eszkadev (approved):

PR #8313, opened 2024-02-19
wsd: Blank stored timestamp if an up-load times out.

caolanm:
Quite possibly we did successfully up-load, or at least our timestamp may not match, but we cannot be sure enough to warn the user in this case. We cannot reliably get notified of remote alteration on the server of files, so wait until a more succesful save or new user join to get an improved timestamp.

Change-Id: I5c8124ed74ba5c26371768ac778b9670282c87a3
(cherry picked from commit 8acf957b4c862d2a5934d6a8e03f94b9ca9e604f)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Sorry! duplicated this, and fixed the unit tests too - which oddly don't fail in 23.05 - for unknown reasons; would expect we need the same patch there really.

PR #8315, opened 2024-02-19
Handle sperator case for tree Widget

Darshan-upadhyay1110:
- separator should be treated as UI not any object value

Before:
image

After:
image

Change-Id: Ie6182dda40b7fd3c0d1ea87c5df78a46239677ed

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -174,6 +174,9 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 	var disabled = treeViewData.enabled === 'false' || treeViewData.enabled === false;
 
 	var li = L.DomUtil.create('li', builder.options.cssClass, parentContainer);
+	if (_isSeparator(entry)) {
+		L.DomUtil.addClass(li,'context-menu-separator');
+	}
 
 	if (!disabled && entry.state == null) {
 		li.draggable = treeType === 'navigator' ? false: true;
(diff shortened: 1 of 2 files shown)


browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -174,6 +174,9 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 	var disabled = treeViewData.enabled === 'false' || treeViewData.enabled === false;
 
 	var li = L.DomUtil.create('li', builder.options.cssClass, parentContainer);
+	if (entry.text === 'Separator') {
```
eszkadev, line 177: Please add function `isSeparator` so it will be in one place.
There is no other property like `type` or similar? Do we need to detect by test?

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -217,7 +220,7 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 			var iconName = builder._createIconURL(iconId, true);
 			L.LOUtil.setImage(icon, iconName, builder.map);
 			L.DomUtil.addClass(span, 'ui-listview-expandable-with-icon');
-		} else if (entry.columns[i].text) {
+		} else if (entry.columns[i].text && entry.columns[i].text != 'Separator') {
```
eszkadev, line 223: !isSeparator()

browser/src/control/jsdialog/Widget.TreeView.js
Darshan-upadhyay1110: >Please add function isSeparator so it will be in one place.

Sure i will do that 

>There is no other property like type or similar?

No, it does not have any type property for separator 

```
entry : {
    columns: [{text: 'Separator'}]
    row: 4
    text: "Separator"
  }
```
 
 >Do we need to detect by test?
 
 Yes we need to add test for detect separator is handled in tree widget view

eszkadev (approved):

PR #8271, opened 2024-02-14
backport - Sheet protect status

hfiguiere:
* Resolves: #  See #551
* Target version: 23.05 

### Summary

The required LOKit patches have landed.

### TODO

- [ ] ...

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id
- [ ] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #8317, opened 2024-02-19
calc: Don't use es6 let and () =>

hfiguiere:
This was introduced by backport #8271 of the sheet protection

Change-Id: I07a9fd428b11042ce191a5c7a48d53400253c445

* Resolves: # 
* Target version: master 

### Summary

This is a cherry-pick to master from 23.05 where it was fixing a breakage not occuring on master, caused by ES6.

### TODO

- [ ] ...

### Checklist


caolanm (approved):
no sure if there was some idea around raising the baseline for co24, but doesn't hurt to sync

PR #8320, opened 2024-02-20
init document theme during load

caolanm:
like we do for the spelling setting, so we have the document in the desired state early before any rendering, dropping the early full document invalidations during initial setup

requires https://gerrit.libreoffice.org/c/core/+/163650 in the code to be useful

Change-Id: I6b762c95fd4c00c7da04cf89f7bbeef4bc57375d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1214,7 +1214,14 @@ window.app = {
 					if (spellOnline) {
 						msg += ' spellOnline=' + spellOnline;
 					}
-
+					var docTypes = ['text', 'spreadsheet', 'presentation', 'drawing'];
+					for (var i = 0; i < docTypes.length; ++i) {
+						var docType = docTypes[i];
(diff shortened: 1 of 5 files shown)


caolanm:
This is in relation to flicker seen on longer writer documents.

It looks a good idea to have document in the desired theme from the start, which is something we do already for for spelling mode, and init from the users settings. This needs https://gerrit.libreoffice.org/c/core/+/163650 alongside it.

If this passes muster, the follow up suggestion is the pairing of https://github.com/CollaboraOnline/online/pull/8329 and https://gerrit.libreoffice.org/c/core/+/163666 to let online know the viewRenderStart of the just constructed documents so we can elide unnecessary "canonicalidchange" messages that cause browser-side to re-request tiles it already asked for.

mmeeks (approved):
Looks good to me. OTOH - I'd really prefer that we cleanup the DocOptions to have some nice structured JSON thing that is more simple to extend in future =)

PR #8321, opened 2024-02-20
Impress: Allow to autostart in presentation mode

meven:
This is implemented in two cases:
 * using a request parameter '&startPresentation=true'
 *  * for ppsx/pps for which the backend sends a startPresentation stateChange event. This can be overridden passing
'&startPresentation=false' to allow editing the file.

The presentation can only start in present-in-window mode due to browser gating fullscreen by user activation.

Refactor SlideShow._onFullScreen to better handle the requestFullscreen permission restriction failing case. Also clean up old references to old restricted implementations.

Change-Id: I98fcb1a7da170fa1c6a7de75c680259163233c9d

Better used with https://gerrit.libreoffice.org/c/core/+/163652

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/map/handler/Map.SlideShow.js
```diff
@@ -52,37 +52,43 @@ L.Map.SlideShow = L.Handler.extend({
...
-			else if (this._slideShow.mozRequestFullScreen) {
-				this._slideShow.mozRequestFullScreen();
-			}
-			else if (this._slideShow.webkitRequestFullscreen) {
-				this._slideShow.webkitRequestFullscreen();
-			} else {
-				// fallback to "open in new tab"
-				L.DomUtil.remove(this._slideShow);
-				this._slideShow = null;
+		var onPresentation = () => {
```
hfiguiere, line 55: arrow functions are a problem if this is to be ported to 23.05

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -52,37 +52,43 @@ L.Map.SlideShow = L.Handler.extend({
...
+			this._presentInWindow = false;
+
+			this._startSlideNumber = 0; // Default: start from page 0
+			if (e.startSlideNumber !== undefined) {
```
hfiguiere, line 59:
```suggestion
			if (typeof e.startSlideNumber !== 'undefined') {
```

This is the only reliable way.

hfiguiere: yeah looks like it was in the previous code

browser/src/map/handler/Map.SlideShow.js
meven: Don't we support the same brower's versions?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#browser_compatibility
What am i missing ?

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -52,37 +52,45 @@ L.Map.SlideShow = L.Handler.extend({
...
+		if (!(this._cypressSVGPresentationTest || this._map['wopi'].DownloadAsPostMessage)) {
+			this._slideShow = L.DomUtil.create('iframe', 'leaflet-slideshow', this._map._container);
+			if (this._slideShow.requestFullscreen) {
 
-		this._startSlideNumber = 0; // Default: start from page 0
-		if (e.startSlideNumber !== undefined) {
-			this._startSlideNumber = e.startSlideNumber;
+				let that = this;
+				this._slideShow.requestFullscreen()
+					.then(doPresentation(that, e))
```
hfiguiere, line 84: This doesn't look right. `.then()` expect a function, not the return value of a `doPresentation()` call which is `undefined`.

Something like:

```suggestion
					.then(function () { doPresentation(that, e) })
```

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -52,37 +52,45 @@ L.Map.SlideShow = L.Handler.extend({
...
+					.catch((err) => {
```
hfiguiere, line 85: ES5 function should be used here too.

browser/src/map/handler/Map.SlideShow.js
hfiguiere: Or 

```suggestion
					.then(doPresentation.bind(that, e))
```

browser/src/map/handler/Map.StateChanges.js
```diff
@@ -54,6 +54,14 @@ L.Map.StateChangeHandler = L.Handler.extend({
 				$('.leaflet-pane.leaflet-map-pane').removeClass('bucket-cursor');
 		}
 
+		if (e.commandName === '.uno:StartWithPresentation' && (state === true || state === 'true')) {
+			let startPresentationParam = window.coolParams.get('startPresentation');
+			if (startPresentationParam === undefined || (startPresentationParam === 'true' || startPresentationParam == true)) {
```
hfiguiere, line 59:
```suggestion
			if (typeof startPresentationParam === 'undefined' || (startPresentationParam === 'true' || startPresentationParam == true)) {
```

browser/src/control/Control.UIManager.js
```diff
@@ -360,8 +360,14 @@ L.Control.UIManager = L.Control.extend({
 
 		this.map.on('changeuimode', this.onChangeUIMode, this);
 
-		if (typeof window.initializedUI === 'function')
+		if (typeof window.initializedUI === 'function') {
 			window.initializedUI();
+		}
+
+		var startPresentationGet = this.map.isPresentationOrDrawing() && window.coolParams.get('startPresentation');
+		if (startPresentationGet == true || startPresentationGet === 'true') {
```
hfiguiere, line 368:
```suggestion
		if (startPresentationGet === true || startPresentationGet === 'true') {
```

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -52,37 +52,45 @@ L.Map.SlideShow = L.Handler.extend({
...
+		if (!(this._cypressSVGPresentationTest || this._map['wopi'].DownloadAsPostMessage)) {
+			this._slideShow = L.DomUtil.create('iframe', 'leaflet-slideshow', this._map._container);
+			if (this._slideShow.requestFullscreen) {
 
-		this._startSlideNumber = 0; // Default: start from page 0
-		if (e.startSlideNumber !== undefined) {
-			this._startSlideNumber = e.startSlideNumber;
+				let that = this;
+				this._slideShow.requestFullscreen()
+					.then(function () { doPresentation(that, e); return true;})
```
hfiguiere, line 84: I don't think you need to return anything here.

browser/src/control/Control.UIManager.js
meven: I used '==' on purpose to be more input tolerant, i.e accepting 1 as true.

browser/src/map/handler/Map.StateChanges.js
meven: In fact window.coolParams.get return '' when the key is missing, defined in global.js

browser/src/map/handler/Map.SlideShow.js
meven: I had a warning, because the return of the `then` is used.

hfiguiere: it was because you passed the function result as a parameter expecting a function.

hfiguiere: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

It definitely return `undefined` in their example.

hfiguiere:
```suggestion
					.then(function () { doPresentation(that, e); })
```

hfiguiere:
I see the gerrit review is for 24.04, are we targetting this for 24.04 only?

hfiguiere:
'nit: one of the commit has a typo (as reflected in the PR title)

meven:
> There should also be a SDK documentation update to document how this works.

I am trying to find a place to document this, there does not seem to be a prefect place at the moment.

Other url query parameters such as closebutton, revisionhistory, randomUser, darkTheme, userinterfacemode, lang, dir, target, alwaysactive, permission, debug and timestamp seem to be not documented.

I think we can add a new page for those parameters, perhaps under "Get Started" > "Advanced integration".

meven:
> There should also be a SDK documentation update to document how this works.

I am trying to find a place to document this, there does not seem to be a prefect place at the moment.

Other url query parameters such as closebutton, revisionhistory, randomUser, darkTheme, userinterfacemode, lang, dir, target, alwaysactive, permission, debug and timestamp seem to be not documented.

I think we can add a new page for those parameters, perhaps under "Get Started" > "Advanced integration".

> > There should also be a SDK documentation update to document how this works.
> 
> I am trying to find a place to document this, there does not seem to be a prefect place at the moment.
> 
> Other url query parameters such as closebutton, revisionhistory, randomUser, darkTheme, userinterfacemode, lang, dir, target, alwaysactive, permission, debug and timestamp seem to be not documented.
> 
> I think we can add a new page for those parameters, perhaps under "Get Started" > "Advanced integration".

I created an issue in the documentation gitlab:
https://gitlab.collabora.com/productivity/sdk/-/issues/2

meven:
The core side enabling this (https://gerrit.libreoffice.org/c/core/+/163652/10) was merged.

meven:
@pedropintosilva would you be interested in reviewing this (after Hubert) ?

hfiguiere (approved):
pending CI.

Issue #8303, opened 2024-02-18
Document Not Reloading After Rename [ Only in Debug Mode ]


codewithvk:
#### Describe the Bug
When attempting to rename a document, the document shows a message indicating that the server has disconnected. This is most likely a regression because everything was working properly until commit hash `43db77e`.

#### Steps to Reproduce
1. Open the document in Nextcloud.
2. Attempt to rename the document.
3. Observe that the server has been disconnected.

#### Screenshots
![Peek 2024-02-18 19-56](https://github.com/CollaboraOnline/online/assets/61119120/6b640673-ff34-4165-9c4c-ea0cb0e42374)
![image](https://github.com/CollaboraOnline/online/assets/61119120/a93d7c75-c186-4416-8bd8-f04b44b3f2b7)

codewithvk:
Found regression :  [cee21f0](https://github.com/CollaboraOnline/online/pull/8218/commits/cee21f0cc42f4dd061d6fff7c18c4cc799731874)

PR #8326, opened 2024-02-20
Fix: Multiple Encoding of wopiSrc  During Document Renaming

codewithvk:
On the client side, when setting up the socket, we encode  beforehand, so there's no need to encode it again.

Change-Id: I09e59c6fdda4c93de5dd36fca5023ac88926c4d4
* Resolves: #8303
* Target version: master

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1294,7 +1294,7 @@ app.definitions.Socket = L.Class.extend({
 			var docUrl = url.split('?')[0];
 			this._map.options.doc = docUrl;
 			this._map.options.previousWopiSrc = this._map.options.wopiSrc; // After save-as op, we may connect to another server, then code will think that server has restarted. In this case, we don't want to reload the page (detect the file name is different).
-			this._map.options.wopiSrc = encodeURIComponent(docUrl);
+			this._map.options.wopiSrc = docUrl;
 			window.wopiSrc = this._map.options.wopiSrc;
 
 			if (textMsg.startsWith('renamefile:')) {


browser/src/core/Socket.js
```diff
@@ -1294,7 +1294,7 @@ app.definitions.Socket = L.Class.extend({
 			var docUrl = url.split('?')[0];
 			this._map.options.doc = docUrl;
 			this._map.options.previousWopiSrc = this._map.options.wopiSrc; // After save-as op, we may connect to another server, then code will think that server has restarted. In this case, we don't want to reload the page (detect the file name is different).
-			this._map.options.wopiSrc = encodeURIComponent(docUrl);
+			this._map.options.wopiSrc = docUrl;
```
Ashod, line 1297: I was expecting to see `encodeURIComponent()`. But, if we're sure that the issue was double-encoding and that this fixes it, then great. (Although I'm not sure how double-encoding can result in breaking the missing-encoding detection.)

Ashod (approved):
Thanks @codewithvk.

browser/src/core/Socket.js
codewithvk: We are checking wopiSrc here: https://github.com/CollaboraOnline/online/blob/master/wsd/RequestDetails.cpp#L33. That's where the code was breaking.

eszkadev: @Ashod this only proves that our protection works :)

eszkadev (approved):

PR #8329, opened 2024-02-20
set the initial rendering state before the first invalidateCanonicalId

caolanm:
now that we will typically init the document theme, etc. from the users last used settings, the current rendering state is likely the one that the user wants to use.

The browser-side throws away all tiles, etc on getting a canonidalidchange, so get the current document renderering state to describe this initial state so later LOK_CALLBACK_VIEW_RENDER_STATE events can be compared to this initial state and only emit canonidalidchange if there is a need.

requires https://gerrit.libreoffice.org/c/core/+/163666 to be merged first

Change-Id: I2902ff4e143f432755c335e1225149207fafdfc3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1777,6 +1777,13 @@ private:
                 viewCount << " view" << (viewCount != 1 ? "s." : "."));
 
         session->initWatermark();
+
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Thanks =)

PR #8330, opened 2024-02-20
cool#8328 - config header fixup.

mmeeks:
config.h should be included as the 1st line in each source file. It should not be included in headers.

config_version.h which changes on every commit should not be included widely - so remove it from the HttpRequest heeader to save tinderbox's ccache.

Fetch version info from helper methods rather than in-lining via defines, to better encapsulate.

Change-Id: If449a36f1ac61940f04d70d5f4180db389d9b4c4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


neilguertincollabora (approved):

neilguertincollabora:
Maybe you should add a comment to config_version.h explaining why it should not be included and what to do instead.

PR #8332, opened 2024-02-20

hcvcastro:
The UNO command is dispatched to change the theme when loading the
document, it causes serious problems the "kit" process changes the
canonical view id.

So this patch avoids sending UNO command and add a parameter to set
the initial theme.

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -264,7 +264,26 @@ window.app = {
 
 		// @property lang: String
 		// browser language locale
-		lang: navigatorLang
+		lang: navigatorLang,
+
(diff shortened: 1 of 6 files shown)


caolanm:
Because master changed and has an alternative "switch to document theme early" what I've done is to make this into two new pull requests:

https://github.com/CollaboraOnline/online/pull/8620 as this original patch, but just for 23.05 backport
and
https://github.com/CollaboraOnline/online/pull/8621 for master which has just the "drop additional redundant activateDarkModeInCore" which is the part of this that doesn't exist in master

PR #8333, opened 2024-02-20
Private/caolanm/reduce subarray usage

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me - thanks Caolan =) 
In retrospect almost certainly the big cost here is in the zstd de-compressor's use of subarray - and we should perhaps reach out to the author there again (?) =)

PR #8334, opened 2024-02-20
Private/mmeeks/unitfixes

mmeeks:
Szymon oddness here re: jsdialog - nowadays I get:

fullyLoadedAndReady]: jsdialog: { "id": "", "type": "messagebox", "text": "Collabora OfficeDev 24.04", "enabled": true, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "child<truncated 3785chars> 
dialogevent 6 {"id":"yes", "cmd": "click", "data": "2", "type": "responsebutton"} 
jsdialog: { "id": 6, "jsontype": "dialog", "action": "close"} 
jsdialog: { "id": "", "type": "messagebox", "text": "Warning", "enabled": true, "visible": false, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "child<truncated 2463chars> 
jsdialog: { "id": "", "type": "messagebox", "text": "Collabora OfficeDev 24.04", "enabled": true, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "children": [ { "id": "", "type": "container", "text": "", "enabled": true, "child<truncated 2481chars> 
dialogevent 7 {"id":"ok", "cmd": "click", "data": "1", "type": "responsebutton"} 
jsdialog: { "id": 7, "jsontype": "dialog", "action": "close"} 
error: cmd=load kind=faileddocloading 

Might be interesting.

### DIFF
--- a/test/UnitBadDocLoad.cpp
+++ b/test/UnitBadDocLoad.cpp
@@ -82,6 +82,7 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
         auto response = helpers::getResponseString(socket, "error:", testname);
         if (Util::startsWith(response, "error: cmd=notasync"))
         {
+            TST_LOG("Ignoring initial notasync error - waiting again");
             // Continue searching for the interesting failure.
             response =
                 helpers::getResponseString(socket, "error:", testname, std::chrono::seconds(30));
--- a/test/UnitBadDocLoad.cpp
+++ b/test/UnitBadDocLoad.cpp
@@ -65,28 +65,34 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
         // Send a load request with incorrect password
         helpers::sendTextFrame(socket, "load url=" + documentURL, testname);
 
+        std::map<std::string, std::string> items;
+
         // We receive jsdialog with question about repair
         const auto dialog = helpers::getResponseString(socket, "jsdialog:", testname);
         LOK_ASSERT_EQUAL(true, dialog.size() > 0);
 
-        // Click "Yes" in a dialog
-        helpers::sendTextFrame(socket, "dialogevent 1 {\"id\":\"yes\", \"cmd\": \"click\", \"data\": \"2\", \"type\": \"responsebutton\"}", testname);
+        // Extract all json entries into a map.
+        items = Util::JsonToMap(dialog.substr(sizeof("jsdialog:")));
+        auto firstId = items["id"];
 
-        // We receive jsdialog with warning that repair failed
-        const auto dialog2 = helpers::getResponseString(socket, "jsdialog:", testname);
-        LOK_ASSERT_EQUAL(true, dialog2.size() > 0);
+        // Click "Yes" in a dialog
+        helpers::sendTextFrame(socket, "dialogevent " + firstId + " {\"id\":\"yes\", \"cmd\": \"click\", \"data\": \"2\", \"type\": \"responsebutton\"}", testname);
+
+        // we can potentially receive multiple jsdialog update messages
+        std::string dialog2;
+        do {
+            dialog2 = helpers::getResponseString(socket, "jsdialog:", testname);
+            LOK_ASSERT_EQUAL(true, dialog2.size() > 0);
+            items = Util::JsonToMap(dialog2.substr(sizeof("jsdialog:")));
+        } while(items["id"] == firstId); // a duplicate update of existing dialog
+
+        // Now we received jsdialog with warning that repair failed
 
         // Click "Ok"
-        helpers::sendTextFrame(socket, "dialogevent 2 {\"id\":\"ok\", \"cmd\": \"click\", \"data\": \"1\", \"type\": \"responsebutton\"}", testname);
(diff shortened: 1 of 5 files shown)


neilguertincollabora:
Did you try opening corrupted.odt interactively? That wasn't working but I don't think it was related because it's been broken since before the test failed. Would be nice to get that fixed but we shouldn't delay this change for it.

test/UnitLoadTorture.cpp
```diff
@@ -76,41 +81,35 @@ int UnitLoadTorture::loadTorture(const std::string& name, const std::string& doc
...
-                    }
-                    else
-                    {
-                        // Correct back, since we aren't unloading just yet.
-                        ++num_of_views;
-                    }
+                    std::lock_guard<std::mutex> guard(stateLock);
+                    LOK_ASSERT("Duplicate view-id generated " && std::find(view_ids.begin(), view_ids.end(), viewid) == view_ids.end());
+                    view_ids.push_back(viewid);
                 }
```
neilguertincollabora, line 91: Leftover {} ?

test/UnitLoadTorture.cpp
```diff
@@ -23,19 +23,20 @@
 /// Load torture testcase.
 class UnitLoadTorture : public UnitWSD
 {
-    int loadTorture(const std::string& name, const std::string& docName,
-                    const size_t thread_count, const size_t max_jitter_ms);
+    void loadTorture(const std::string& name, const std::string& docName,
+                    const size_t thread_count, const size_t max_jitter_ms = 100);
```
neilguertincollabora, line 27: Can we set max_jitter to 100 all the time and remove the extra parameter? I see it is called with 75 below, but I don't think 75 vs 100 would make much difference.

test/UnitLoadTorture.cpp
```diff
@@ -76,41 +81,35 @@ int UnitLoadTorture::loadTorture(const std::string& name, const std::string& doc
...
+
+                // Get all the views loaded - lean on test timeout to interrupt.
+                while (num_to_load > 0)
+                    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+
+                // delay randomly to close in random order:
+                const auto ms
+                    = (max_jitter_ms > 0
+                       ? std::chrono::milliseconds(Util::rng::getNext() % max_jitter_ms)
+                       : std::chrono::milliseconds(0));
```
neilguertincollabora, line 108: Randomness is generally a bad idea in tests because it affects reproducibility. Obviously there is some inherent randomness because it's multithreaded, but if we can avoid adding more that would be nice.

Edit: Actually I don't see where this value is used at all

test/UnitBadDocLoad.cpp
```diff
@@ -82,6 +82,7 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
         auto response = helpers::getResponseString(socket, "error:", testname);
         if (Util::startsWith(response, "error: cmd=notasync"))
         {
+            TST_LOG("Ignoring initial notasync error - waiting again");
```
neilguertincollabora, line 97: Since the alert is no longer thrown on document load, can we get rid of this extra section?

eszkadev:
It will be good to understand why jsdialog messages have been changed. I will look.
(on core version upgrades we often have changes that break things, so not always it is a good solution to just change test)

eszkadev:
> It will be good to understand why jsdialog messages have been changed. I will look. (on core version upgrades we often have changes that break things, so not always it is a good solution to just change test)

Solved in core: https://gerrit.libreoffice.org/c/core/+/163679

test/UnitLoadTorture.cpp
mmeeks: the braces scope the lock guard.

test/UnitLoadTorture.cpp
mmeeks: Quite possibly; but there was probably some thought here.

test/UnitLoadTorture.cpp
mmeeks: Good point we're missing a sleep =) Of course cleaning this up to have a nice fixed order that is pathological would be great but ...

test/UnitBadDocLoad.cpp
mmeeks: Sure why not - with the other patch included.

test/UnitBadDocLoad.cpp
```diff
@@ -65,23 +65,36 @@
...
 
+        // Extract all json entries into a map.
+        items = Util::JsonToMap(dialog.substr(sizeof("jsdialog:")));
+        auto firstId = items["id"];
+
         // Click "Yes" in a dialog
-        helpers::sendTextFrame(socket, "dialogevent 1 {\"id\":\"yes\", \"cmd\": \"click\", \"data\": \"2\", \"type\": \"responsebutton\"}", testname);
+        helpers::sendTextFrame(socket, "dialogevent " + firstId + " {\"id\":\"yes\", \"cmd\": \"click\", \"data\": \"2\", \"type\": \"responsebutton\"}", testname);
+
+        // FIXME: unexpected - we receive multiple jsdialog update messages
```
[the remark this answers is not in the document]
mmeeks: removed the fixme - added a robustness comment =)

neilguertincollabora (approved):

PR #8338, opened 2024-02-21
Don't reinit header entry styles without a mode change.

gokaysatir:
Change-Id: I82bdbaea429d4b2fec8cda9c431a215feb47a129

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
@caolanm might be rather helpful perf-wise on 23.05 too - if Szymon is happy =)

caolanm (approved):
looks reasonable to me, and switching dark/light mode, or start a new session with saved dark mode auto applied on new document seem to work as expected.

PR #8339, opened 2024-02-21
private/marco/calc tab switching restore pos

mcecchetti:
- calc: on tab swithing do not replay print twips messages
- calc: on tab switching document view can be restored to a wrong position

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -332,6 +332,7 @@ L.Control.Tabs = L.Control.extend({
 
 	//selected sheet is moved to new index
 	_moveSheet: function (newIndex) {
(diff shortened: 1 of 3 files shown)


mcecchetti:
Hi @caolanm ,
I requested your review since you're the author of `don't try and reselect graphic on previous tab on switching to new tab` (https://github.com/CollaboraOnline/online/commit/aa00f92b99939b62a3676c12018ecc7d637e67b5). First commit fixes a similar problem with cell selection.  It was causing an unexpected jump on tab switching. I fixed the problem in a more radical way, why to replay any kind of message on the wrong part ? I hope there is no side effect.  
The second commit is still about restoring the sheet view at the correct position, after a tab switch.

caolanm (approved):
that seems fairly reasonable to me, tab switching might never have been in mind with that original replay concept

PR #8349, opened 2024-02-22
Fix sporadically failing autofilter test, simplify autofilter and other calc tests

neilguertincollabora:
* Resolves: # 
* Target version: master 

### Summary
Commit 1:
The autofilter test was failing quite often in the tinderbox. It regularly failed and then passed on rerun locally.
I could not figure out exactly why, but I added a wait so that it will pass.

Commit 2:
I also cleaned up the verification of sheet contents, replacing the custom autofilter function with the one in calc_helper. And I wrapped the function with one whose name says what it does (assert the sheet contents) instead of how it does it (look at the clipboard).

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -194,20 +194,26 @@ function ensureViewContainsCellCursor() {
 	});
 }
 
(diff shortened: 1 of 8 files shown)


eszkadev (approved):
Looks reasonable

PR #8351, opened 2024-02-22
Replace Util::startsWith/endsWith w/ C++20 equivalent

thebearon:
Replace a few cases of Util::toString(...), too.

Change-Id: I53c5dbbdf2f60710e94add9dd137aa9becb7bd3e

* Target version: master 

### Checklist

### DIFF
--- a/test/UnitSession.cpp
+++ b/test/UnitSession.cpp
@@ -123,8 +123,8 @@ UnitBase::TestResult UnitSession::testHandshake()
         wsSession->poll(
             [&](const std::vector<char>& message)
             {
-                const std::string msg = Util::toString(message);
-                if (!Util::startsWith(msg, "error:"))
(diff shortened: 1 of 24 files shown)


test/UnitSession.cpp
```diff
@@ -123,8 +123,8 @@ UnitBase::TestResult UnitSession::testHandshake()
         wsSession->poll(
             [&](const std::vector<char>& message)
             {
-                const std::string msg = Util::toString(message);
-                if (!Util::startsWith(msg, "error:"))
+                const std::string msg(std::string(message.begin(), message.end()));
```
vmiklos, line 126: Would it work to just go with `const std::string msg(message.begin(), message.end())`?

vmiklos (approved):
Looks OK in general, I just didn't expect you'll also replace some toString() calls as well.

Issue #8342, opened 2024-02-21
Improve UX of the paste special fallback dialog


vmiklos:
### Describe the Bug
In case navigator.clipboard.read() fails for paste special, then we show a dialog which does work, but it has poor UX and can easily confuse users.

### Steps to Reproduce
1. Open Calc
2. Copy a cell
3. Click the paste dropdown on the notebookbar
4. Click the paste special menu item

### Expected Behavior
A clear dialog that explains it's time to press Ctrl-V.

### Actual Behavior
A dialog that looks like just a warning to be ignored, and a confused user when pressing OK doesn't result in a paste special.

### Screenshots

![image](https://github.com/CollaboraOnline/online/assets/13838/89ac2345-aecf-4681-8abc-09c8a7f0ced4)

### Desktop

Tested on Linux with Firefox, Chrome doesn't show this dialog anymore (with experimental settings enabled).

### Suggested fixes

1. 'Ok' legend needs renaming to 'Cancel' in 23.05 - no translation change needed and 'Cancel paste special' in 24.04
2. We should bold the 'Please press now:' - which at least in English should say: "Please press" so it is far more visible.
3. We should loose the excessive white-space at the bottom of the dialog if we can.
4. Probably we should have a "Paste from this document" button as well which is likely the majority use-case and shouldn't need access to the clipboard

Tex2002ans:
Yep, full agree!

- - -

I think it could just be reworded to:

> Your browser has very limited access to the keyboard.
>
> Press `Ctrl` + `V` to see more options.

Then you could just:

- Press `Cancel` (or click out of popup) if you want to get out.
- Press `Ctrl+V` if you want to go further.
   - Bringing you to the "Paste Special" dialog.

- - -

**Side Note:** I just wrote a little about the clunky Calc "Paste Special" dialog here too:

- https://github.com/CollaboraOnline/online/issues/8033#issuecomment-1936949598

vmiklos:
New look after #8366:

![image](https://github.com/CollaboraOnline/online/assets/13838/913e50fb-6258-49a5-bf9f-aeba878bd502)

vmiklos:
The above two PRs are now in, closing.

PR #8366, opened 2024-02-23
Improve the paste special fallback dialog, step 2

vmiklos:
The 4 commits complete points 1-4 from https://github.com/CollaboraOnline/online/issues/8342 . This contains new l10n strings, so not to be backported as-is.


vmiklos:

See https://github.com/CollaboraOnline/online/issues/8342#issuecomment-1965948324 for a screenshot, this does the rest of the sub-tasks from the issue. You can test this in Firefox, ideally you don't hit this dialog in Chrome.

caolanm (approved):

PR #8367, opened 2024-02-23
ui-tooltip: don't inherit the same shadow values as regular overlays

pedropintosilva:
what: jquery ui-tooltip visible when user hovers buttons in the
notebookbar on when user starts a formula in calc (helpful floating
tooltip appears)

Before this commit, ui-tooltip was being set with a harsh
shadow making it more of a main element than a simple tooltip. Best to
don't inherit that strong shadow from .ui-widget-shadow (used in
overlays) and instead set a new one just for tooltip elements.

Change-Id: Ibf3ad525f97decc1d82ca7172ca1fb2bb7c1a6f2

### DIFF
--- a/browser/css/jquery-ui-lightness.css
+++ b/browser/css/jquery-ui-lightness.css
@@ -938,6 +938,9 @@ body .ui-tooltip,
 	color: var(--color-main-text);
 	background: var(--color-background-lighter);
 }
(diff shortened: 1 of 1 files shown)


pedropintosilva:
Backport https://github.com/CollaboraOnline/online/pull/8368

Rash419 (approved):

PR #8372, opened 2024-02-23
Private/caolanm/reduce subarray usage23

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
The idea here is that this is exactly the same before and after, except reducing the use of subarrays

vmiklos (approved):
Aha, this is a backport of https://github.com/CollaboraOnline/online/pull/8333 . I don't see anything obviously problematic with this, should help perf.

eszkadev (approved):

PR #8375, opened 2024-02-23
Keyboard shortcuts dialog: don't allow width to change after opened

pedropintosilva:
With the inclusion of the searching (actually filtering) input field
the dialog would change width mid typing. Example: 1. open dialog: the
width would be x; 2 type something in the search box: the width would
be < x. Better to set a fixed width so if the biggest element is
hidden the dialog doesn't change in size.

Change-Id: I6662433388efa66812317f60a4e0a6bf36e5e98a

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1910,6 +1910,12 @@ kbd,
 	min-width: 75%;
 }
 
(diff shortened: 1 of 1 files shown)


Darshan-upadhyay1110 (approved):
Tested looks good. Thanks 👍🏼

Issue #6941, opened 2023-07-19
Dragging slides sometimes results in wrong slide order


siccovansas:
**Describe the bug**
This doesn't happen every time, but sometimes if I drag a slide to another location it ends up at the top of all slides (see video below). Other times the slide at the top is moved to another location after I dragged another slide (so I didn't event touch this top slide). It seems that this maybe only happens when there are 'many' slides and a scrollbar is shown?

**To Reproduce**
1. Create a new presentation
2. Create enough slides to show a scrollbar (8 or 9 on an HD screen)
3. Move slides around. Most of the times to works fine. But sometimes the slides end up in a wrong order. (maybe you first also have to close and open the presentation?)

**Expected behavior**
Slides are moved to the location where they are dragged to. And slides that aren't dragged stay at their location.

**Actual behavior**
Slides sometimes end up in the wrong location.

**Screenshots**
[Video of slide 4 ending up in the wrong location (top) after dragging it two times](https://github.com/CollaboraOnline/online/assets/696839/ada53edc-13a6-4969-86c9-877500e519f3).

**Desktop (please complete the following information)**
 - OS: Ubuntu 22.04.2
 - Browser: Firefox 115.0.2
 - Nextcloud: 26.0.3
 - Collabora Online - Built-in CODE Server: 23.5.104

gitgt:
I reproduce often, on 2nd or 3rd move, but I cannot write exact steps. @Ezinnem please see.

hfiguiere:
I think it happens when you drag the slide between two (while there is the highlight). If you drag it over the one above the insertion place, it will work.

See attached video that show when the bug occur:

[Screencast from 2024-02-01 16-29-30.webm](https://github.com/CollaboraOnline/online/assets/114441/ab7d3a43-7039-4dab-b9dd-c163fe85daa4)

Pay close attention to the position of the mouse cursor.

PR #8388, opened 2024-02-23
impress: Make sure that drop a slide between slide reorder correctly.

hfiguiere:
We need to get the right node in the control

Change-Id: Ib37c6411667fcbf4714e6f6448292c84adc78f0e

* Resolves: #6941 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
I couldn't reproduce the issue but i think this improvement will work. Thank you.

PR #8400, opened 2024-02-26
wsd: fix assertion and better logging

Ashod:
In some cases the WOPISrc is inconsistently
encoded. So now we decode before comparing.

Also, improve related log entries.

Change-Id: Ifd6cca76c91a34761dd0ce8f6d78c3af5d47fd43


neilguertincollabora (approved):
I tested the change locally and everything looks good!

PR #8403, opened 2024-02-27
Fix cypress impress text position test failures

neilguertincollabora:
Tests desktop/impress/top_toolbar_spec.js and mobile/impress/apply_font_text_spec.js have been failing after a change in core that slightly alters the position of some text elements.

These tests fail when run against core 24.04 but not against 23.05. So there must have been a core that caused them to fail. I have not tried to find the change.

I'm not sure whether to consider this a regression in core or a flaky test here in online. Nevertheless, I'd like to submit these changes to get the tests passing as quickly as possible.

These positions were stable for several years and I don't know why they have changed now.

The new x value of "top_toolbar_spec.js/Apply left/right element" is the same in Jenkins and on my machine, but the y values of "Apply superscript" and "Apply subscript" are different (8711 vs 8745, originally 8734), so I have allowed a range of values. The mobile test hasn't run on Jenkins (because it stops after the desktop tests fail) so I don't know if they will need a range too. It might be good to modify all these tests to use a range, if we see them continue to fail.

I am also including some minor cleanup in separate commits in this pull request.

### Checklist

### DIFF
--- a/cypress_test/integration_tests/desktop/impress/top_toolbar_spec.js
+++ b/cypress_test/integration_tests/desktop/impress/top_toolbar_spec.js
@@ -17,8 +17,6 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Top toolbar tests.', funct
 		} else {
 			desktopHelper.hideSidebar();
 		}
-
-		impressHelper.selectTextShapeInTheCenter();
 	});
 
 	afterEach(function() {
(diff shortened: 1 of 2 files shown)


vmiklos:
@neilguertincollabora could you please try to get CI green? Than happy to take a look at the actual changes. Thanks.

vmiklos (approved):
Looks like the possibly controversial part of this is that something change on the core side for the generated SVG so there is now some <1px difference between expected vs actual. This is probably just some noise: while it would be ideal to not have change detector tests at all, the seen difference is acceptable, so this looks good to me in general.

PR #8405, opened 2024-02-27
keep focus on input field on document load

JaumePujantell:
If the document reloads while the user is on an input field, do not
take the focus away to the document.

Change-Id: If5e97a666ddef6e7c28a30c8bd8ef8bca5f5b336

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -80,7 +80,7 @@ class IdleHandler {
 			}
 		}
 
-		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen()) {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
This does seem to fix the problem. Here is what I tested:

1) Load a test document with lots of text, font size is set to 15pt
2) ctrl-a to select all text
3) try to change font size using the home tab -> font size dropdown
4) tripple-click on the '15pt' to get it selected
5) try to change the size to '12 pt', so type '1' as a start. The focus used to go back to the body text so "2" replaced the in-doc selection, but it doesn't do that anymore.

Some nits:

1) If the commit doesn't have an issue associated with it, please describe it in the commit message, see https://github.com/CollaboraOnline/online/blob/master/README.CONTRIBUTING.md#other-information-to-put-into-your-commit-message (the intro is missing in your case)

2) Did you consider writing a cypress test to cover your fix?

Thanks.

PR #8412, opened 2024-02-27
Improve following UI

### LINKED ISSUE #7602: Improve 'Following' UI
### Is your feature request related to a problem?

Here are 3 users in a document
3 users in a document

Here's what it looks like when I follow one of them
it looks like the same screenshot, but I followed one of them- I promise...

The only way I can see which one I followed is by clicking on users again
a close-up of the users popup, showing I'm following LocalUser#3

This problem is exacerbated when using the always follow the editor feature
a closeup of the always follow the editor checkbox

Because then there's not a good way to know who you jumped to beyond looking who's editing nearby
a screenshot of the users popup, showing no "following" text

### Describe the solution you'd like

Other apps, (example: Figma) do this way better: look at this, it's clear who I'm following right from the top corner, not to mention the document border
a screenshot of figma, showing off its awesome following UI

No matter who you're following, they go right to the top of the list... it's beautiful!
image

### Describe any other solutions you've considered

I imagine an easier (although worse) solution would be something that just showed you were following someone, even that isn't clear currently without clicking on users 

### Additional context

Figma doesn't stand alone, VSCode also does something similar

vscode, showing a "following" banner and a green pin to show you're following someone
a closeup of the green pin
a closeup of the following banner

### LINKED ISSUE #7603: Allow unfollowing users by clicking on them again
### Is your feature request related to a problem?

When I want to unfollow someone, I often try to do it by clicking on the user I'm following
Screenshot from 2023-11-03 10-11-32

Unfortunately, that's not possible... instead you have to click on yourself
Screenshot from 2023-11-03 10-12-05

### Describe the solution you'd like

Presumably this was meant as a "follow yourself" thing, and it's fine but it'd be nice to be able to click on the user you're following to cancel as well. I'd like this to be implemented as an additional option

### Describe alternatives you've considered

- We could have a separate unfollow button instead
- We could make it clearer that clicking on yourself is the unfollow option (e.g. by adding some text underneath that said "stop following")
Frame 1

### Additional context

This feature is related to #7602, in that they are both pain points with the following UI currently

Minion3665:
Fixes #7602, Fixes #7603

The following UI has historically been functional but poor - you haven't been able to see who you were following without opening the user list, and it was similarly hard to *un*follow them

This PR contains a set of improvements for the following UI. Its highlights are reordering the user list so the most important bits are up-top, allowing you to click on various easy-to-reach places to stop following someone and showing you who you're following right next to the avatar list. For more details, please look through the commits.


browser/src/control/Control.UserList.ts
```diff
@@ -434,6 +461,22 @@ class UserList extends L.Control {
...
+		const followingChip = document.getElementById('followingChip');
+
+		const following = this.getFollowedUser();
+
+		if (following === undefined || !this.map._docLayer._followUser) {
+			followingChip.style.display = 'none';
+			return;
+		}
+
+		followingChip.innerHTML = this.options.followingChipText.replace('%user', following[1].username);
```
eszkadev: can we not use innerHTML but create nodes and use innerText? (username is something what comes from user/integration input)

browser/src/control/Control.UserList.ts
```diff
@@ -483,25 +507,35 @@ class UserList extends L.Control {
...
+
+		if (this.getFollowedUser() === undefined) {
+			displayedAvatarCount = this.options.userLimitHeader;
+		} else {
+			displayedAvatarCount = this.options.userLimitHeaderWhenFollowing;
+		}
+
+		const topAvatarZIndex = displayedAvatarCount;
+
+		followingChip.innerHTML = '<div id="text">' + this.options.followingChipText.replace('%user', following[1].username);
```
eszkadev, line 547: still innerHTML, commented that on previous commit

browser/src/control/Control.UserList.ts
Minion3665: Sure, that makes sense

We directly insert the username as HTML in the user join/leave messages

```ts
		if (type === 'join') {
			message = this.options.userJoinedPopupMessage.replace('%user', username);
		} else {
			message = this.options.userLeftPopupMessage.replace('%user', username);
		}

		$('#tb_actionbar_item_userlist')
			.w2overlay({
				class: 'cool-font',
				html: message,
				style: 'padding: 5px'
			});
```

so it is sanitized with `escapeHtml` as soon as we get it, so I believe it not to be a vulnerability, but I agree it seems rather like it could become a problem with someone making a mistake. I'll clean it and the other bits up

browser/src/control/Control.UserList.ts
```diff
@@ -157,29 +162,34 @@ class UserList extends L.Control {
...
+
+		if (img === undefined) {
+			img = L.DomUtil.create('img', 'avatar-img') as HTMLImageElement;
+		}
+
+		const defaultImage = L.LOUtil.getImageURL('user.svg', this.map._docLayer._docType);
+
 		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
-			img = L.DomUtil.create('img', 'avatar-img');
 			img.src = extraInfo.avatar;
```
eszkadev, line 179: it would be good for a11y to add also `alt` with user name

eszkadev (commented):
nice, a lot of cleanup. only 2 nitpicks:
- alt for avatar image
- innerHTML -> innerText

browser/src/control/Control.UserList.ts
```diff
@@ -0,0 +1,581 @@
...
+		if (followedUser === undefined) {
+			return undefined;
+		}
+
+		return [this.map._docLayer._followThis, followedUser];
+	}
+
+	unfollowAll() {
+		this.map._docLayer._followEditor = false;
+		this.map._docLayer._followThis === -1;
```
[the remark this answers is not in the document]
Minion3665: Fair enough, I guess it doesn't

pedropintosilva:
awesome!

Minion3665:
(this forcepush brought to you by "stepping through the commits and running prettier-write for each of them, then checking out the path in the commit to avoid conflicts", or in other words it is purely formatting)

Minion3665:
> awesome! Thank you @Minion3665 ❤️
> 
> I wonder if we could do something additionally when the window is narrow: image
> 
> Thoughts?

It's a decent idea, I've not done anything special for when the window is too small yet, but it would be quite nice to have *something*

I suspect this could be added to some media query, so long as we don't mind having to open the menu to stop following. It'd probably be a little more annoying to make, say, clicking on the avatar of the followed user unfollow

Also: following "the editor" is a bit different to following a specific user. Right now I'm using you as the user in the chip, this is a bit lazy. If we were going to have a condensed UI this should almost certainly be done in some different way, to avoid making it unclear who you were following (e.g. we could have a "the editor" pseudouser in the avatar preview with its own profile image)

Minion3665:
> > awesome! Thank you @Minion3665 ❤️
> > I wonder if we could do something additionally when the window is narrow: image
> > Thoughts?
> 
> It's a decent idea, I've not done anything special for when the window is too small yet, but it would be quite nice to have _something_
> 
> I suspect this could be added to some media query, so long as we don't mind having to open the menu to stop following. It'd probably be a little more annoying to make, say, clicking on the avatar of the followed user unfollow
> 
> Also: following "the editor" is a bit different to following a specific user. Right now I'm using you as the user in the chip, this is a bit lazy. If we were going to have a condensed UI this should almost certainly be done in some different way, to avoid making it unclear who you were following (e.g. we could have a "the editor" pseudouser in the avatar preview with its own profile image)

regardless, I guess maybe this is enough that we should get it in a separate PR, these changes are already a huge improvement, let's wait for the final bits of review rather than implementing new features in #8412 now

eszkadev:
I found something:
- A: in writer follow the editor
- B: session does something on other page (so we jump with session A to B)
- A: types something on keyboard
Result: A didn't jump to current cursor, header still shows "you are following the editor chip", but when B does something - we don't jump to it

Previously when you started to type - following was stopped, but you saw what you are typing I think.

eszkadev (commented):
looks ok for me

eszkadev (approved):

PR #8413, opened 2024-02-27
Fix view jumps when switching tabs in calc

eszkadev:
Disable multi-sheet selection keyboard shortcut - here is minified CO-23.05 version.
In master it's different: https://github.com/CollaboraOnline/online/pull/8427/commits/87018a49a01279e379c34cdc66f49a901d963147
    
commits:

---
    
     Calc: disable mutli-sheet selection keyboard shortuct

     - it doesn't work / is implemented
     - it makes view jumps in some cases

----

    Don't switch to the same sheet again
    
    This prevent us from requesting sheet geometry we
    already have + we block a bug with jumping view.
    
    Steps to reproduce:
    1. open spreadsheet with 2 tabs
    2. switch to tab 2
    3. click again on tab 2
    4. scroll up with mouse wheel
    Result: after some time we jumped back (every time we scrolled)

-----

    This caused problems with view jumping when user
    clicked on the tab he already was on.

------

    cypress: test for jumps when switching tabs in calc
    
    Protect us from regressions.
    
    We had issues with jumping view in calc, it was related to
    switching tabs (not all was visible in the UI).
    
    One case was repeated selection of the same tab.
    Second one was using a keyboard shortcut to select multiple
    sheets (ctrl + shift + page up/down).


cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js
```diff
@@ -0,0 +1,82 @@
...
+		// go to sheet 1
+		cy.cGet('#spreadsheet-tab0').click();
+		desktopHelper.assertScrollbarPosition('vertical', 35, 45);
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'G45');
+
+		// use shortcut to go to 2nd sheet
+		helper.typeIntoDocument('{ctrl}{alt}{pageUp}');
+
+		// click on sheet 2 tab
+		cy.cGet('#spreadsheet-tab1').click();
```
neilguertincollabora, line 45: Why change to second sheet both ways?

cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js
```diff
@@ -0,0 +1,82 @@
...
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'F720');
+		desktopHelper.assertScrollbarPosition('vertical', 320, 330);
+
+		// at this point after scroll up or down we got view jump
+		cy.cGet('input#addressInput').type('{selectAll}A2{enter}');
+		desktopHelper.assertScrollbarPosition('vertical', 15, 25);
+	});
+
+	it('Check multiple sheet selection and view position', function() {
```
neilguertincollabora, line 54: Same name as other test point. Also it looks like they test the same thing.

cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js
```diff
@@ -0,0 +1,82 @@
...
+		// go to sheet 1
+		cy.cGet('#spreadsheet-tab0').click();
+
+		desktopHelper.assertScrollbarPosition('vertical', 35, 45);
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'G45');
+
+		// select sheet 2
+		helper.typeIntoDocument('{ctrl}{shift}{pageDown}');
```
neilguertincollabora, line 62: ctrl+shift+pagedown moves tabs in my browser. Did you mean ctrl+alt+pagedown?

cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+		// we still have selected 2 sheets
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'F720');
+
+		// go to sheet 2
+		cy.cGet('#spreadsheet-tab1').click();
+		cy.cGet('input#addressInput').should('have.prop', 'value', 'F720');
+
+		// at this point after scroll up or down we got view jump
+		// if we go to any cell, we jump back to F720
```
neilguertincollabora, line 78: Describe what the behavior should be, not what the bug did.

vmiklos (approved):
Looks OK, and most importantly repairs CI for 23.05.

Tomaz: can you also review, please? Thanks.

quikee (approved):
Looks fine

PR #8416, opened 2024-02-28
Handle specific tooltip message for function usage tooltip

GulsahKose:
Using same callback caused a regression about showing tooltip while reviewing a change in writer. So we have to use a special message for function usage tooltip.

Change-Id: I6f769a43af6a01ed8e13abfd9c35a57ccd47c1ac

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1027,13 +1027,35 @@ L.Control.UIManager = L.Control.extend({
 		this.map.fire('hidebusy');
 	},
 
+	// Document area tooltip
+
+	/// Shows general tooltips in the document area
+	/// tooltipInfo contains rectangle (position in twips) and text properties
+	showDocumentTooltip: function(tooltipInfo) {
+		var split = tooltipInfo.rectangle.split(',');
+		var latlng = this.map._docLayer._twipsToLatLng(new L.Point(+split[0], +split[1]));
+		var pt = this.map.latLngToContainerPoint(latlng);
+		var elem = $('.leaflet-layer');
+
+		elem.tooltip();
+		elem.tooltip('enable');
+		elem.tooltip('option', 'content', tooltipInfo.text);
+		elem.tooltip('option', 'items', elem[0]);
+		elem.tooltip('option', 'position', { my: 'left bottom',  at: 'left+' + pt.x + ' top+' + pt.y, collision: 'fit fit' });
+		elem.tooltip('open');
+		document.addEventListener('mousemove', function() {
+			elem.tooltip('close');
+			elem.tooltip('disable');
+		}, {once: true});
+	},
+
 	// Calc function tooltip
 
 	/// Shows tooltip over the cell while typing a function in a cell.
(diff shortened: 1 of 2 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1773,7 +1773,12 @@ L.CanvasTileLayer = L.Layer.extend({
 			this._onCalcFunctionListMsg(textMsg.substring('calcfunctionlist:'.length + 1));
 		}
 		else if (textMsg.startsWith('tooltip:')) {
-			this._onCalcFunctionUsageMsg(textMsg.substring('tooltip:'.length + 1));
+			if (textMsg.startsWith('tooltip: formulausage:'))
```
eszkadev, line 1776: We do core change anyway. Can we simplify that and just use JSON always? That will also prevent us from unexpected new types of tooltips.

```
var tooltipInfo = JSON.parse(textMsg.substring('tooltip:'.length + 1));
if (tooltipInfo.type === 'formulausage')
    onCalcFunctionUsage(tooltipInfo.text);
else if (tooltilInfo.type === 'trackchange')
   showDocumentTooltip(tooltipInfo);
else
    console.error('unknown tooltip type');
```

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1969,12 +1974,12 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	_onCalcFunctionUsageMsg: function (textMsg) {
 		var pos = this._lastVisibleCursorRef._northEast;
-		this._map.uiManager.showDocumentTooltip(textMsg, $('.leaflet-layer'), pos);
+		this._map.uiManager.showFormulaTooltip(textMsg, $('.leaflet-layer'), pos);
```
eszkadev, line 1977: btw, I think we can remove all these: `$('.leaflet-layer')` and just get that element directly inside called function.
It is always the same as these tooltips are for "canvas area", no need for a parameter now.

browser/src/control/Control.UIManager.js
```diff
@@ -1027,13 +1027,35 @@ L.Control.UIManager = L.Control.extend({
...
+			elem.tooltip('disable');
+		}, {once: true});
+	},
+
 	// Calc function tooltip
 
 	/// Shows tooltip over the cell while typing a function in a cell.
 	/// tooltipInfo contains possible function list. If you type a valid
 	/// function it'll show the usage of the function. elem has to be
 	/// jQuery selector output eg. $('.leaflet-layer')
```
eszkadev, line 1057: please update comment in case params will be changed

pedropintosilva:
note: https://gerrit.libreoffice.org/c/core/+/164093

browser/src/layer/tile/CanvasTileLayer.js
eszkadev: sure, thanks

eszkadev (approved):
looks good :)

PR #8417, opened 2024-02-28
notebookbar: event prevent default for insert annotation button (backport)

lpranam:
problem:
if writer->insert tab-> insert comment button is clicked, if button is clicked from icon is works fine,
but if button is clicked on label event is triggered twice, which caused js errors

Change-Id: I3956083b02204733abc3543c73f6019d3a3b4c67

### Checklist


Rash419 (approved):
Looks good to me!

eszkadev (approved):

eszkadev (approved):

PR #8418, opened 2024-02-28
Close all dialogs on document idle

Darshan-upadhyay1110:
- All dialogs should be closed if document become idle
- There are 2 cases for which all dialogs are being closed before socket closed or before screen dim 1. Document idle : we disconnect from server 2. View idle: only screen dim, server is still connected

To test:
    - add auto filter in calc sheet
    - open autofilter popup
    - wait for document to become idle

Note: change value of `idle_timeout_secs` in coolwsd.xml to test

Change-Id: If76f03d830b8b3744fd2b54d131c5fa85a024c5c

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -124,6 +124,7 @@ class IdleHandler {
 	}
 
 	_dim() {
+		this.map.fire('closealldialogs');
 		const message = this.getIdleMessage();
 
 		window.app.console.debug('IdleHandler: _dim()');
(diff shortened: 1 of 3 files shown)


eszkadev (changes requested):
It blocks user input on "per view idle", because we don't send message to the server. (after activation it is not possible to type into cells). (it would be nice to have cypress test for this one case as it is seriour issue, here: https://github.com/CollaboraOnline/online/blob/master/cypress_test/integration_tests/idle/calc/idle_spec.js )

"per document idle" works correctly.

eszkadev (approved):

Issue #7821, opened 2023-12-11
Draw: Select+Drag make drag without feedback


hfiguiere:
### Describe the Bug
When you select + drag an object, the drag (move) happen without feedback.

### Steps to Reproduce
1. Open a draw document with some content
2. Click and drag on a shape to move it

### Expected Behavior
The shape follow the mouse cursor and when you release it get moved there.

### Actual Behavior
When you release the mouse it will have moved as expected, but otherwise you don't have feedback.

### Screenshots

[Screencast from 2023-12-11 15-39-13.webm](https://github.com/CollaboraOnline/online/assets/114441/29f8302a-3df2-445c-94f1-bd7d79072469)

The video show the select + drag vs just drag.

### Desktop

COOLWSD version: 23.05.6.1 (git hash: 65a85cda)
LOKit version: Collabora Office 23.05.6.2 (git hash: a6b78ba)
Served by: Fedora Linux 38 (Container Image)
Server ID: 55888ee5

(Please complete the following information)
 - **Collabora version:** [e.g. 23.05.5.3]
 - **OS and version:** [e.g. Windows 11]
 - **Browser and version:** [e.g. Chrome 114.0]

hfiguiere:
This video show the mouse cursor. Same scenario as above.

[Screencast from 2023-12-11 15-43-29.webm](https://github.com/CollaboraOnline/online/assets/114441/5c47ef50-1aa1-494b-85ba-6e5923c6fcf7)

hfiguiere:
It looks like this is more than just a cosmetic issues.

PR #8419, opened 2024-02-28
canvas: svg: show textbox move animation while edit

lpranam:
problem:
if a textbox/shape is being edited and try to move it while still in editing, it was not animated, so user gets illusion that shape was not moved, but in fact it moved and directly appears at the last moved position without animation

Change-Id: I830e368e162b3b38edea0d9240cf82800fa8080a

* Target version: master 

### Checklist


lpranam:
requires https://gerrit.libreoffice.org/c/core/+/163982

browser/src/layer/vector/SVGGroup.js
```diff
@@ -174,7 +174,7 @@ L.SVGGroup = L.Layer.extend({
 	_onDragStart: function(evt) {
 		if (!this._map || !this._dragShapePresent || !this.dragging)
 			return;
-		if (this._map._docLayer._cursorMarker && this._map._docLayer._cursorMarker.isVisible())
+		if (this._map._docLayer._cursorMarker && this._map._docLayer._cursorMarker.isVisible() && this.options.isText)
```
caolanm, line 177: what does this new bit do?

lpranam: if we return from here, it prevents drag animation and allows us to show selection. this new isText option also new and is added in this PR.

hfiguiere: #7821 (which is I believe the same issue) also applies to non text objects. When I had a look at it it's where I found the code path did diverge.

caolanm (approved):
Well, lets not block then.

PR #8420, opened 2024-02-29
keep focus on input field on document load

JaumePujantell:
If the document reloads while the user is on an input field, do not
take the focus away to the document.

Change-Id: If5e97a666ddef6e7c28a30c8bd8ef8bca5f5b336

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -80,7 +80,7 @@ class IdleHandler {
 			}
 		}
 
-		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen()) {
+		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen() && $('input:focus').length === 0) {
 			this.map.focus();
 		}
 


vmiklos (approved):
Looks OK (I reviewed this on master). Next steps:

- please wait for CI to pass. If it fails, you'll need to rebase, Szymon is working on turning the CI back to green at https://github.com/CollaboraOnline/online/pull/8413
- once you have a green CI, please also request a review from Szymon (all 4 recent changes to this file are from him), and then he can approve + merge it

Thanks!

browser/src/control/Control.IdleHandler.ts
```diff
@@ -80,7 +80,7 @@ class IdleHandler {
 			}
 		}
 
-		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen()) {
+		if (window.mode.isDesktop() && !this.map.uiManager.isAnyDialogOpen() && $('input:focus').length === 0) {
```
eszkadev, line 83: we have helper method which is able to detect any input (we have many variants):
https://github.com/CollaboraOnline/online/blob/993af840c00bb329926b14eb27455de9622488bd/browser/src/layer/tile/CanvasTileLayer.js#L3974
I think we need to move it to better place for sharing...

eszkadev (approved):
Looks ok, let's get this in for sync with master. We can do a followup with `_isAnyInputFocused`

PR #8422, opened 2024-02-29
annotation: avoid resetting comment show state if being edited

lpranam:
problem:
when try to insert comment(inside text with existing comment) and comment field is empty, click on some other comment(which is on same text), it will cause an empty comment insertion in DOM which can not be removed until doc reload.
This subsequently prevented new comment insertion.

Change-Id: I29294dcfbbaddddaa768e08b47b855b3e8c7f454

* Target version: master 

### Checklist


gokaysatir (approved):
Thanks @lpranam :)

PR #8426, opened 2024-02-29
calc: tab drag&drop does not work properly

mcecchetti:
This is a regression from 74a4484e "calc: on tab switching document
view can be restored to a wrong position"

Moreover the passed index was wrong when tab is dragged on the right
wrt to current tab position. Now the correct zero based index is
always computed.

Change-Id: I7395773da5a846ce61d8508bbd1d6e2e833d4511


Darshan-upadhyay1110 (approved):
Tested, it works. Thanks!
CC: @mcecchetti

mcecchetti:
Thanks  Darshan!

PR #8427, opened 2024-02-29
 Fix view jumps when switching tabs in calc

eszkadev:
Add KeyboardShortcuts and disable multi-sheet selection
    
    This will be a class for storing keyboard shortcuts in
    a unified way. We can easily define shortcuts for different
    languages so we don't have to write long if-statements
    which can cause problems if done not very carefully.
    
    Added tests protecting from regressions.

----

    Don't switch to the same sheet again
    
    This prevent us from requesting sheet geometry we
    already have + we block a bug with jumping view.
    
    Steps to reproduce:
    1. open spreadsheet with 2 tabs
    2. switch to tab 2
    3. click again on tab 2
    4. scroll up with mouse wheel
    Result: after some time we jumped back (every time we scrolled)

-----

    This caused problems with view jumping when user
    clicked on the tab he already was on.

------

    cypress: test for jumps when switching tabs in calc
    
    Protect us from regressions.
    
    We had issues with jumping view in calc, it was related to
    switching tabs (not all was visible in the UI).
    
    One case was repeated selection of the same tab.
    Second one was using a keyboard shortcut to select multiple
    sheets (ctrl + shift + page up/down).


vmiklos:
<https://cpci.cbg.collabora.co.uk:8080/job/github_linux_release_co-24.04/126/console>:

> stderr: fatal: unable to access 'https://github.com/CollaboraOnline/online.git/': Failed connect to github.com:443; Connection refused

Let's try once more.

browser/src/control/Control.Tabs.js
```diff
@@ -324,9 +324,15 @@ L.Control.Tabs = L.Control.extend({
 	},
 
 	_setPart: function (e) {
-		var part =  e.currentTarget.id.match(/\d+/g)[0];
-		if (part !== null) {
-			this._setPartIndex(parseInt(part));
+		var part = e.currentTarget.id.match(/\d+/g)[0];
```
vmiklos, line 327: Nit: would it be possible to avoid regex matching here? For example put the part number in a data-* attribute and then just get it instead of parsing it out of the id?

vmiklos (approved):
Looks OK in general, see inline comment for a nit.

PR #8430, opened 2024-03-01
calc: tab drag&drop does not work properly

mcecchetti:
This is a regression from 74a4484e "calc: on tab switching document
view can be restored to a wrong position"

Moreover the passed index was wrong when tab is dragged on the right
wrt to current tab position. Now the correct zero based index is
always computed.

Change-Id: I7395773da5a846ce61d8508bbd1d6e2e833d4511

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -338,7 +338,11 @@ L.Control.Tabs = L.Control.extend({
 
 	//selected sheet is moved to new index
 	_moveSheet: function (newIndex) {
-		this._map._docLayer._sheetSwitch.updateOnSheetMoved(this._map.getPartNumber(), newIndex - 1);
+		var currentTab = this._map.getCurrentPartNumber();
+		// odd but true: the base index changes according to the tab is dragged
+		// on the left or on the right wrt the current tab position
+		var newIndexZeroBased = newIndex > currentTab ? newIndex - 2 : newIndex - 1;
+		this._map._docLayer._sheetSwitch.updateOnSheetMoved(currentTab, newIndexZeroBased);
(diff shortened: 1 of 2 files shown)


mcecchetti:
Already merged in master

vmiklos:
Can you please rebase? The tip of the branch should have passing CI, Szymon just fixed it.

mcecchetti:
@vmiklos it's already re-based on Szymon fix infact cypress tests passed

vmiklos:
> @vmiklos it's already re-based on Szymon fix infact cypress tests passed

I see. Could you at least fix up your commit messages to please the DCO check? I think that's the only required check that fails for you. Thanks.

mcecchetti:
> I see. Could you at least fix up your commit messages to please the DCO check? I think that's the only required check that fails for you. Thanks.
I rebased, anyway I don't understand the check complain, the `Signed-off-by` line is present.

vmiklos:
> I rebased, anyway I don't understand the check complain, the Signed-off-by line is present.

Now the DCO check is green. :-)

eszkadev (approved):
Code looks ok

quikee (approved):
Well I guess this is fine for a quick fix, but looking at the code in SheetSwitch.ts the centerOfSheet map should be changed to some dynamic array structure (vector or list like - whatever the JS supports) so we don't need to change indices in the map when we move something around (would be handled by the dynamic array structure).
That code there  (updateOnSheetInsertion, updateOnSheetInsertion, updateOnSheetMoved) seems to me quite fragile and unnecessary.

eszkadev:
@quikee centerOfSheet ia a Map<number, any> already, we should probably use some key which is unique identifier of a "part" not the index

quikee:
@eszkadev yes, it's a Map - that's the issue - the behavior however aligns with an array

quikee:
Hm OK, the sheet name is enforced to be unique, so we could also use that and a Map....

PR #8433, opened 2024-03-01
wsd: move deprecated logic under legacy_server config

Ashod:
We no longer send LOOL-WOPI headers, unless
the configuration specifically flags for
legacy servers. But we always send COOL-WOPI
even to legacy servers, to help them upgrade
seamlessly.

Change-Id: Ifc919ed8f6665cd8f846117ef4e8b7ef09fbd563


hfiguiere (approved):
LGTM but the iOS build failed.

PR #8435, opened 2024-03-01
annotation: prevent page up/down button from messing the view

lpranam:
problem:
when pressed page up/down button inside the annotation textarea, entire view is forcefully pushed upwards to put the cursor at the top of view. This is a chrome bug reported at: https://issues.chromium.org/issues/41417806

Change-Id: I30048374e7b8330cd8c865010e3b4237d355391e

* Target version: master 

### Checklist


eszkadev (approved):
Works as expected, thanks

eszkadev (changes requested):
now I noticed that it is enough to use any key...
eg. just type something and it also happens

eszkadev (approved):

eszkadev:
Ok, let's use it. Later we can continue improvements

PR #8437, opened 2024-03-01
annotation: fixed resolved comments are not hidden completly (backport)

lpranam:
problem:
when we turn off the option to show the resolved comments, author image was not hidden and stuck on screen which can't be interacted or scrolled

Change-Id: Id0fb30c787b6e8c288df1142b9f882b7633f3dc9

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -605,6 +605,10 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	background: var(--color-background-lighter) !important;
 }
 
+.cool-annotation-collapsed-show .cool-annotation-img {
+	visibility: visible;
+}
+
 #mobile-wizard-popup {
 	background-color: var(--color-background-lighter) !important;
 }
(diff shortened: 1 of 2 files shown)


browser/src/layer/tile/CommentListSection.ts
```diff
@@ -1809,8 +1809,10 @@ export class CommentSection extends CanvasSectionObject {
 						this.unselect();
 					}
 					this.sectionProperties.commentList[idx].hide();
+					L.DomUtil.removeClass(this.sectionProperties.commentList[idx].sectionProperties.container, 'cool-annotation-collapsed-show');
```
eszkadev, line 1812: it cannot be done inside `hide()` ?

browser/src/layer/tile/CommentListSection.ts
```diff
@@ -1809,8 +1809,10 @@ export class CommentSection extends CanvasSectionObject {
...
 				} else {
 					this.sectionProperties.commentList[idx].show();
+					L.DomUtil.addClass(this.sectionProperties.commentList[idx].sectionProperties.container, 'cool-annotation-collapsed-show');
```
eszkadev, line 1815: it cannot be done inside `show()` ?

lpranam:
Done :)

gokaysatir (approved):
Same goes also here, if we can prepare a follow up for reply count of resolved threads:)

Thanks for the fix.

eszkadev:
Cypress fails on annotation tests - seems to be related then

eszkadev (approved):
looks ok

eszkadev:
note: in the future best would be to create method for:
`this.sectionProperties.data.resolved === 'false' || this.sectionProperties.commentListSection.sectionProperties.showResolved`
it is repeated many times.

something like `shouldBeVisible()`, this makes code more clear, readable and in case we will need modify condition -> it's single place to change

PR #8441, opened 2024-03-02
private/ash/cfi pre

Ashod:
- wsd: test: refactor WopiTestServer to override uploading
- wsd: expose http::Session's socket FD
- wsd: new CheckFileInfo class to do async requests
- wsd: use new CheckFileInfo class in RequestVettingStation
- wsd: log the previous upload attempt result


mmeeks (approved):
Looks fine to me - thanks Ash.

PR #8446, opened 2024-03-04
Modals should always be placed on top of tooltips

pedropintosilva:
Before this commit, already opened ui-tooltips were being placed on top
of everything, including modal dailogs. Fix that, set jquery css
line to the same z-index used by non-modal dialogs.

Additionally: update browser/README file so it documents this change
and clean up vex legacy bits.

Change-Id: Ie7123822d96c594433ac197aeb9c651aa9b14461

### DIFF
--- a/browser/README
+++ b/browser/README
@@ -387,10 +387,10 @@ x				menu-nb-hamburger > ul (calculated from jquery.smartmenus.js)
 -------------------------------------------
 		on the top
 -------------------------------------------
-1105			dialogs
+1105			dialogs, tooltips
 1001			mobile-edit-button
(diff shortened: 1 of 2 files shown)


pedropintosilva:
How to test it 

on calc:

* Open a spreadsheet
* Add sum with number (3) make the tooltip appear
* open the help button
  * Without this PR: See the tooltip is not closing and messes with the Help modal


This PR fixes the visibility order. This way user can read help dialog then close it and the tooltip nicely still there (provided that the cursor is still editing that formula) so the user can resume where he left of

gokaysatir (approved):
Works, thanks :)

eszkadev (approved):

PR #8449, opened 2024-03-04
cool#8023 browser: try to use the new Chrome clipboard in non-experimental mode

vmiklos:
Now that
<https://sdk.collaboraonline.com/docs/advanced_integration.html#clipboard-handling>
is figured out, this should be less risky.

Change-Id: I549adad6386d328d7dc5f7e4eeee16446665df82


mmeeks:
Lets get it in - thanks Miklos =)

mmeeks (approved):

PR #8450, opened 2024-03-04
browser: fix regression print range

hcvcastro:
Do not send the command ".uno:DeletePrintArea" to server side,
it disables printing range feature.

Change-Id: I8bb1ec9ab01793e6bf3b74b8cb121a5fbad3b8e2

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -1270,7 +1270,6 @@ L.Map.include({
 			this.uiManager.toggleWasm();
 			break;
 		case 'print-active-sheet':
-			this.sendUnoCommand('.uno:DeletePrintArea');
 			var currentSheet = this._docLayer._selectedPart + 1;
 			var options  = {
 				ExportFormFields: {
@@ -1290,7 +1289,6 @@ L.Map.include({
 			this.print(options);
 			break;
 		case 'print-all-sheets':
-			this.sendUnoCommand('.uno:DeletePrintArea');
 			this.print();
 			break;
 		case 'savecomments':


caolanm:
Deleting the print range does seem a bit drastic, but I'm not sure what the original intent was. Does this then break an expected use case?

hcvcastro:
let's redirect the question to @Darshan-upadhyay1110 ^

Darshan-upadhyay1110:
We can not remove this.sendUnoCommand('.uno:DeletePrintArea'); from the print options case and the reason is as follows:

- Consider you have multiple sheets in current document
- if you try to define or add print range in any one sheet
- then try to print that current sheet, it will work OK
- but here is the catch, just go to other sheet where print range is not defined and try to print that
- it will only print the content of current sheet which is resided in the defined range of the other sheet where you have defined range
    - for EX: if you defined range in sheet 2 let's say on B4. Then try to print data of Sheet 1 it will only print the content of resides that range defined in sheet 2. All the data outside that range in sheet 1 will not be printed.
- That is why I have deleted the range before print active sheet or current sheet

Addition Note: https://github.com/CollaboraOnline/online/pull/7035#issuecomment-1672704969
CC: @hcvcastro @caolanm

Darshan-upadhyay1110:
@hcvcastro  I am thinking what if we just

1. Add new option Print define range in print dropdown
2. it will not use the delete range command and also only print the range data of current sheet

caolanm:
Attached is a simple ods created from scratch, The first sheet was defaulted to "all sheet" print range. I changed the second sheet to have a print range for the area of cells "sheet 2 print" and the third was changed to be "none".

[simpleprintrange.ods](https://github.com/CollaboraOnline/online/files/14492842/simpleprintrange.ods)

With Henrys patch applied then these are the outputs for attempting "print active sheet" on each sheet and then "all sheets"

[sheet1.pdf](https://github.com/CollaboraOnline/online/files/14492890/sheet1.pdf)
[sheet2.pdf](https://github.com/CollaboraOnline/online/files/14492894/sheet2.pdf)
[sheet3.pdf](https://github.com/CollaboraOnline/online/files/14492899/sheet3.pdf)
[allsheets.pdf](https://github.com/CollaboraOnline/online/files/14492901/allsheets.pdf)

Sheet 1 pdf gets all content on that sheet, Sheet 2 gets the content from the sheet 2 print range and sheet 3 is blank where there was no print range. And allsheets contains Sheet 1 content + Sheet 2 print range content. 

That seems fairly reasonable. I don't seem to see a quirk where if I e.g. define a print range on sheet 2 then if affects  what gets printed from sheet 1? Maybe this example is too simple or there is a quirk using xlsx format?

Darshan-upadhyay1110:
@caolanm I tested, and it seems issue of content not being printed is solved :). Sorry for the noise

Darshan-upadhyay1110 (approved):
Tested and works fine, Thanks @hcvcastro

caolanm (approved):

PR #8451, opened 2024-03-04
Move freeze options to view tab and add freeze-panes options to column/row header context menu

Darshan-upadhyay1110:
- move freeze options from layout tab to view tab
- options should be consistent for both compact and tabbed view
- Add freeze-panes options to column/row headers


eszkadev (approved):
code looks ok

pedropintosilva (approved):
image

^Thanks

PR #8454, opened 2024-03-05
UI:annotation: fixed selection indication for track change buttons

lpranam:
Change-Id: I0e4f40d1f2ad50c323e9780e05451d7fe9e6e4c9

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1776,6 +1776,14 @@ L.Control.Menubar = L.Control.extend({
 								$(aItem).removeClass(constChecked);
 							}
 						}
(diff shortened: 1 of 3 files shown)


eszkadev (changes requested):
quick git grep and it is used also in Menubar:
`browser/src/control/Control.Menubar.js:                 this._map.dispatch('showresolvedannotations');`
`dispatch()` is a thing for sharing between notebookbar an classic mode, it has to be always checked if both are working.

eszkadev:
As I understand that is regular uno command but we need to do something additional on the JS side.
So it's not the usual use-case of `dispatch()` which was dedicated for online-only commands.
I think we need "interceptor" for uno commands as well... but that something for the future.

Here it will be enough to just fix classic mode entries.

eszkadev (approved):
looks ok

PR #8460, opened 2024-03-05
lok_allow: remove also port

eszkadev:
Noticed in some k8s deployments where alias groups were defined with port. For lok_allow we need only host name. Avoid adding empty hosts.


eszkadev:
backport passed tests: https://github.com/CollaboraOnline/online/pull/8459

caolanm (approved):
looks sane to me

PR #8461, opened 2024-03-05
Fix processing of URL parameters (in proxy.php case)

timar:
The URLSearchParams() function was confused in the proxy.php case, because the global.location.search contained two '?' marks, like ?req=/browser/dist/cool.html?WOPISrc=...&lang=..&etc... In the normal case the string started with ?WOPISrc=

Change-Id: I566935656be6622d698ea2ba3c949455877526f6

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -80,8 +80,9 @@ window.app = {
 
 	global.setLogging(global.coolLogging != '');
 
+	var gls = global.location.search;
 	var coolParams = {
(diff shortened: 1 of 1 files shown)


timar:
@thebearon could you please test this with proxy.php before merging?

hfiguiere (approved):

thebearon (approved):
Thanks for the fix, Andras!

PR #8463, opened 2024-03-06
Fix for unit-httpws failure

neilguertincollabora:
Fix race condition in testSaveOnDisconnect by waiting for kit process to stop instead of just counting total kit processes

Change-Id: I6bda6b114070123a1366bc04eac1873f19928ac0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/helpers.hpp
+++ b/test/helpers.hpp
@@ -13,6 +13,7 @@
 
 #include <test/testlog.hpp>
 #include <test/lokassert.hpp>
+#include <test/test.hpp>
 
 #include <Poco/BinaryReader.h>
 #include <Poco/Dynamic/Var.h>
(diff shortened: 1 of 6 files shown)


mmeeks (approved):
Looks ok - but leaves behind the race condition in all sorts of other tests.
Also it breaks the: "Did we leave any other processes behind" check of the previous code - which is worth having: to check we don't launch tons of processes by mistake.

That's fine in this one place - but I'd like a follow-on that fixes both testCountHowManyCoolKits and also testNoExtraCoolKitsLeft - to both use the new getKitPids() method - hopefully that will fix a large number of similar race conditions in tests left & right and make things more robust.
We should back-port this to 23.05 too - the same thing must cause grief there =)

Otherwise nice patch; thanks ! =)

Ashod (commented):
Thanks @neilguertincollabora.

PR #8467, opened 2024-03-06
WOPI: Action_Export responses

eszkadev:
PostMessage test page: add export with notify set
    
    This allows to test responses.
    
    Signed-off-by: Szymon Kłos <szymon.klos@collabora.com>
    Change-Id: Idfee8632ac0aeb8a105555e7c9d527288568091e

-----------------------

    Notify WOPI host about status after export
    
    It requires to add `Notify: true` in the Action_Export
    postmessage. Result will be sent with `Action_Save_Resp`
    
    Signed-off-by: Szymon Kłos <szymon.klos@collabora.com>
    Change-Id: Iaeee2e22bf1a2ae934eb657116c1878cb6ed50f6

----------------------

    Make exportas error more informative for user
    
    Replace "The server encountered a xxx error while parsing
    the yyy command." with more informative "Document cannot
    be exported. Please try again." on export issues.

PR #8468, opened 2024-03-06
WOPI: Action_Export responses (master)

eszkadev:
CI running on 23.05: https://github.com/CollaboraOnline/online/pull/8467

------------------

    PostMessage test page: add export with notify set
    
    This allows to test responses.

-----------------------

    Notify WOPI host about status after export
    
    It requires to add `Notify: true` in the Action_Export
    postmessage. Result will be sent with `Action_Save_Resp`

----------------------

    Make exportas error more informative for user
    
    Replace "The server encountered a xxx error while parsing
    the yyy command." with more informative "Document cannot
    be exported. Please try again." on export issues.

----------------

Errors can be tested with core patch:
```
diff --cc sfx2/source/doc/objserv.cxx
index b5bde70a3ca6,b5bde70a3ca6..b0c5c8ad2568
--- a/sfx2/source/doc/objserv.cxx
+++ b/sfx2/source/doc/objserv.cxx
@@@ -275,7 -275,7 +275,7 @@@ bool SfxObjectShell::APISaveAs_Impl(std
                                      const css::uno::Sequence<css::beans::PropertyValue>& rArgs)
  {
      bool bOk = false;
--
++    return bOk;
      if ( GetMedium() )
      {
          OUString aFilterName;
@@@ -897,6 -897,6 +897,11 @@@ void SfxObjectShell::ExecFile_Impl(SfxR
          case SID_SAVEASREMOTE:
          case SID_SAVEDOC:
          {
++            rReq.SetReturnValue( SfxBoolItem(0, false ) );
++
++            Invalidate();
++            return;
++
              // so far only pdf and epub support Async interface
              if (comphelper::LibreOfficeKit::isActive() && rReq.GetCallMode() == SfxCallMode::ASYNCHRON
                  && (nId == SID_EXPORTDOCASEPUB || nId == SID_EXPORTDOCASPDF))
```

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -138,6 +138,12 @@
             });
       }
 
+      function exportAsPdfWithNotify() {
+        post({'MessageId': 'Action_Export',
+              'Values': { 'Format': 'pdf', Notify: true }
+            });
+      }
+
       function getExportFormats() {
         post({'MessageId': 'Get_Export_Formats',
               'Values': null
@@ -476,7 +482,9 @@
             <button onclick="startPresentation(); return false;">Start presentation</button>
 
             <button onclick="exportAsHtml(); return false;">Export as HTML</button>
+            <button onclick="exportAsPdfWithNotify(); return false;">Export as PDF with Notify set</button>
             <button onclick="getExportFormats(); return false;">Get export formats</button>
+            <div></div>
 
             <button onclick="ShowSave(false); return false;">Hide Save Commands</button>
             <button onclick="ShowSave(true); return false;">Show Save Commands</button>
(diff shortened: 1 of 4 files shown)


browser/html/framed.doc.html
```diff
@@ -476,6 +482,7 @@ <h3>Various other messages to post</h3>
             <button onclick="startPresentation(); return false;">Start presentation</button>
 
             <button onclick="exportAsHtml(); return false;">Export as HTML</button>
+            <button onclick="exportAsPdfWithNotify(); return false;">Export as PDF with Notify set</button>
             <button onclick="getExportFormats(); return false;">Get export formats</button>
```
hfiguiere, line 486: I would add an empty `<div>` here to make the next button pushed to the next row. This would keep the logical grouping.

eszkadev: done

PR #8470, opened 2024-03-06
Make JSDialogs fully reachable when wider than the screen

Minion3665:
[Improve screen-edge jsdialog drag ergonomics](https://github.com/CollaboraOnline/online/commit/b4dc25401442955925833803663047a3ad014f5b) 

When trying to drag a jsdialog offscreen, we previously froze the
dialog. This was frustrating if you dragged a jsdialog to the edge of
the screen, because you then couldn't continue to drag it at all without
moving it away from the edge of the screen, even if the direction you
tried to drag it would not move it offscreen.

An example is if I dragged a jsdialog to the right edge of the screen,
it would be harder to drag it up as it would "stick" to the edge. I
would expect the upwards component of the movement to be honored, even
if the rightwards component could not be.

[jsdialogs: allow limited dragging offscreen](https://github.com/CollaboraOnline/online/commit/5fb75a11edd29cdd4c101dc7c7f43b836bb24a04)

You should be able to drag part of a JSDialog offscreen, so long as you
are also dragging another part of them onscreen. This allows you to
access the whole jsdialog even if you have a small device, which
otherwise would not be possible.

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -780,9 +780,9 @@ L.Control.JSDialog = L.Control.extend({
 		builder.executeAction(dialogContainer, innerData);
 	},
 
-    _clamp: function(value, min, max) {
-        return Math.min(Math.max(value, min), max);
-    },
+	_clamp: function(value, min, max) {
+		return Math.min(Math.max(value, min), max);
+	},
 
 	onPan: function (ev) {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):

Minion3665:
CodeQL C++ is broken in master, and this doesn't change any C++... I'm going to merge it without worrying about the failing irrelevant CodeQL

PR #8472, opened 2024-03-06
Fix click in impress annotation tests

neilguertincollabora:
There are now two .avatar-img elements after recent user list changes

Change-Id: Ia229422b4dc3e90ede240e36927b4615091cb97d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665 (approved):
Looks good, and I know why it's needed given my changes - as well as the new features a lot of refactoring went on and I started using avatar-imgs for everything rather than having some stuff use a background image instead...

Basically, what I'm saying is this makes a lot of sense :)

Minion3665:
(trivial rebase, I expect it to fail CodeQL C++ which is broken in master but pass everything else. It doesn't modify any C++ so in that case I will go ahead and merge)

PR #8481, opened 2024-03-07
Stop following text overflowing chip

Minion3665:
~~Depending on the page font size, 0.8em may resolve to a different size.~~
~~If this is larger the text will overflow the --btn-size heighted container.~~

~~By sizing based on --btn-size instead, we can avoid this issue. A smaller button can have smaller text.~~

With some styles the following text was too large for the chip.
Unfortunately, sizing it to be consistent made it unreadably small. To
solve this, we can show only the first line of the text in the following
chip and show the second as a tooltip instead.

Change-Id: I4b5306370c31f4d0b13e1a91bf17a9d491807c45


mmeeks:
My suspicion is that in the end we will get some a11y person complaining about minimum font sizes, and we'll have to go with a Figma-style thing ;-) but - lets see for now; it looks nice :-)

pedropintosilva:
image
 it looks super small to the point that I think it would be better to drop the second line and have it appearing as a tooltip when hovering the chip : 
image

pedropintosilva (approved):

PR #8483, opened 2024-03-07
Stop following anyone when you start typing

Minion3665:
When you are following someone and start typing, it's a pretty good indication that you don't want to be following that person, particularly because if they edit you will then be jumped over to where they are

We also need to stop unfollowAll from always rerendering the user list, as when we rerendered on every keypress there was a noticable performance degredation. Instead, we can make it only rerender when we're following someone, which is a lot better.

Change-Id: I110ca74859b7cee03c6a58c87d3975275f29760d


eszkadev (approved):
looks ok

Issue #8465, opened 2024-03-06
Better plain text copy


vmiklos:
### Describe the Bug
Currently we give HTML to the browser when it comes to copy&paste, and then the browser converts that HTML to plain text. This is worse than our own plain text export, so we should use that instead.

### Steps to Reproduce
1. Create a new Writer document with a normal paragraph and 3 bullets.
2. Select all.
3. Copy
4. Paste somewhere that wants a plain text format.

### Expected Behavior

Something similar to Collabora Office:

```
Bullets:
    • a
    • b
    • c
```

### Actual Behavior

HTML -> plain text conversion is done by browser, e.g. Firefox produces:

```
Bullets:

	a
	b
	c
```

### Additional Context

A related problem is that when we determine if the current selection is simple or complex, then Writer marks inline images as complex (good), but anchored images are still simple (bad). That should be fixed first, given it allows easier testing that way.

This is a general problem, affecting both Firefox and Chrome.

vmiklos:
https://gerrit.libreoffice.org/c/core/+/164453 for the related problem around a simple selection that has an image.

vmiklos:
#8569 fixes the complex case, this should be OK now.

PR #8489, opened 2024-03-08
cool#8465 clipboard: improve handling of plain text copy

vmiklos:
simple case,Currently the current selection is always requested as HTML by the
browser, and then we ask the browser to convert it to plain text.

The problem is that e.g. Writer can produce much better plain text from
its model, compared to the plain text by the browser, e.g. bullet
characters for bullet points.

Fix the problem by:

- CanvasTileLayer.js, _onTextSelectionMsg(): requesting both HTML and
  plain text. Use ',' as a separator, as that's already established,
  e.g. the HTTP Accept header does that already

- Switching the textselectioncontent protocol message from just HTML to
  JSON that contains both HTML and plain text. This is produced in
  ChildSession::getTextSelection() and parsed in CanvasTileLayer.js,
  _onMessage()

- Clipboard.js, setTextSelectionHTML(): allowing setting both HTML and
  plain text.

- ClientSession::postProcessCopyPayload(): knowing if the content to be
  processed is HTML-in-JSON or just HTML, do additional escaping in the
  JSON / textselectioncontent case, but leave the other clipboardcontent
  case unchanged.

So far this only handles the simple case, the behavior for complex
selections are left unchanged for now. The payload is also unchanged
when a single format is requested, as many tests depend on test.

Change-Id: I2fe1378a8d50b7901ac9e808eb78858cd8ff8575


vmiklos:

#8465 has some before/after "screenshots" to see what's the benefit. Initially I wanted to go all-in for the JSON usage for the textselectioncontent message, but so many c++ tests depend on it (5 suites) that I decided to just use it in the new multi-type case (at least not change it in this commit). Local run of cypress tests don't seem to find new failures (both before and after is some 4 suites failing out of the 62 ones).

caolanm (approved):
ok then

PR #8488, opened 2024-03-08
Private/gokay/comment editing readonly

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

- [ ] Code is properly formatted
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #8491, opened 2024-03-08
Add sample URLs for readonly mode to console output.

gokaysatir:
Change-Id: I2e10a25850a64bb666d0473d8881878ff41ad6b1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -768,7 +768,7 @@ inline std::string getLaunchBase(bool asAdmin = false)
     return oss.str();
 }
 
-inline std::string getLaunchURI(const std::string &document)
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks great; do we propagate this to kit & core though ? and overwrite what we get from CheckFileInfo if readonly=true (but never if it is false =)

gokaysatir:
> Looks great; do we propagate this to kit & core though ? and overwrite what we get from CheckFileInfo if readonly=true (but never if it is false =)

Yes we will be propagating the readonly flag to core side, but it is in another commit :) Waiting for approval.

* https://github.com/CollaboraOnline/online/pull/8488

PR #8492, opened 2024-03-08
Mobile-view: paragraph section fixes

Darshan-upadhyay1110:
- Fix image path for base spin field icons
- Do not hide or show elements which are force to be hidden in mobile view

### DIFF
--- a/browser/src/control/Control.MobileWizardWindow.js
+++ b/browser/src/control/Control.MobileWizardWindow.js
@@ -228,12 +228,12 @@ L.Control.MobileWizardWindow = L.Control.extend({
 
 		$('#mobile-wizard .ui-effects-placeholder').hide();
 
-		var nodesToHide = $(contentToShow).siblings().not('.mobile-wizard-scroll-indicator');
+		// do not select already hidden nodes at first place
+		var nodesToHide = $(contentToShow).siblings(':visible').not('.mobile-wizard-scroll-indicator');
 
 		var parent = $(contentToShow).parent();
 		if (parent.hasClass('toolbox'))
-			nodesToHide = nodesToHide.add(parent.siblings().not('.mobile-wizard-scroll-indicator'));
-
+			nodesToHide = nodesToHide.add(parent.siblings(':visible:not(.mobile-wizard-scroll-indicator)'));
 		var duration = 10;
 		if (animate) {
 			nodesToHide.hide('slide', { direction: 'left' }, duration);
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.MobileWizardWindow.js
```diff
@@ -299,13 +299,21 @@ L.Control.MobileWizardWindow = L.Control.extend({
 				this._customTitle ? this._setCustomTitle(this._customTitle) : this._setTitle(this._mainTitle);
 
 			var currentNode = $('.ui-explorable-entry.level-' + this._currentDepth + '.mobile-wizard:visible');
-			var headers = currentNode.siblings();
+			// select only those nodes which are updated on Level down
+			var headers = currentNode.siblings().filter(function() {
+				return $(this).attr('style') === 'display: none;';
```
pedropintosilva, line 304: is this ok

browser/src/control/Control.MobileWizardWindow.js
```diff
@@ -299,13 +299,21 @@ L.Control.MobileWizardWindow = L.Control.extend({
...
 			var currentHeader = currentNode.children('.ui-header');
 			headers = headers.add(currentHeader);
 
 			var parent = currentNode.parent();
 			if (parent.hasClass('toolbox'))
-				headers = headers.add(parent.siblings());
+				parent.siblings().each(function() {
+					// select only those nodes which are updated on Level down
+					if ($(this).attr('style') === 'display: none;') {
+						headers = headers.add($(this));
```
pedropintosilva, line 314: Is this ok? Meaning, will it always work? wouldn't be better to `$(this).style.display === 'none'`

Darshan-upadhyay1110: `$(this).style.display === 'none'` will not work as it consider all the elements which have display = none ( including the elements which we forcefully hide by css )

this condition will only consider elements which have inline property display = none

Darshan-upadhyay1110: As i am seeing you are right if in future there are addition inline css line added then this fail 🤔

pedropintosilva:
image
 looks good, I need to check the second one, please backport to 23.05 so CI can start

pedropintosilva (approved):
Looks good but it would be nice to wait and see if https://github.com/CollaboraOnline/online/pull/8517 passes the CI also could you give an example of 

> Do not hide or show elements which are force to be hidden in mobile view

Darshan-upadhyay1110:
> example of Do not hide or show elements which are force to be hidden in mobile view

/* Paragraph properties */
 #paraspacingimg,
 #aboveparaspacingimg,
 #belowparaspacingimg,
 #firstlineindentimg,
 #indentimg,
 #beforetextindentimg,
 #aftertextindentimg
 
 https://github.com/CollaboraOnline/online/blob/master/browser/css/mobilewizard.css#L1234

eszkadev (approved):
looks ok

PR #8500, opened 2024-03-09
cypress: master fixes of check-desktop

eszkadev:
For me it fixes `make check-desktop` tests. Some tests I disabled some are fixed.

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -357,7 +357,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		}
 
 		if (data.enabled === false) {
-			div.setAttribute('disabled', '');
+			div.disabled = true;
 			spinfield.setAttribute('disabled', '');
 		}
 
@@ -510,7 +510,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		toolbox.id = data.id;
 
 		if (data.enabled === false) {
-			toolbox.setAttribute('disabled', '');
+			toolbox.disabled = true;
 			for (var index in data.children) {
 				data.children[index].enabled = false;
 			}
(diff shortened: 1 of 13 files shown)


neilguertincollabora:
I think we should not disable mobile and multiuser tests in the make file. To run the desktop tests only in CI, use make check-desktop (or from top level directory: make -C cypress_test check-desktop)

eszkadev:
> I think we should not disable mobile and multiuser tests in the make file. To run the desktop tests only in CI, use make check-desktop (or from top level directory: make -C cypress_test check-desktop)

but how it is defined what CI does? (I don't know)
If that's Jenkins config - I think better to do it here, so together with fixing commit we can enable that just for one PR. Not all pending.

neilguertincollabora (approved):

eszkadev:
> It is defined in Jenkins, and it's super easy to change. It probably needs to be updated to run the desktop cypress tests anyways.
> 
> I'm concerned that this hides new failures in check-mobile for developers who run make check locally. It doesn't affect me though, since my script runs check-desktop, check-mobile, and check-multi separately anyways.
> 
> I don't feel too strongly about it. You can submit it this way if you want.

Developers can still make `check-mobile` (which fails anyway currently). I wanted to make it simpler for fixing (no need to modify Jenkins config, but only push makefile change together with cypress fixes in single PR).

The goal is to have something what can pass on CI as quickly as we can. Even if it is a smaller subset.

neilguertincollabora:
Sounds good.

eszkadev:
It failed previously with `assert expected **<div#sidebar-panel>** to be **visible**`
https://github.com/CollaboraOnline/online/blob/ad3fca4ea32cdb0ddbe5e57309570900e90a0143/cypress_test/integration_tests/common/helper.js#L337

eszkadev:
It proves, build is just slow due to poco (wait on load fixes sidebar detection). We need to fix poco, then it will work.

eszkadev:
Unit Tests against Core co-24.04 (With Cypress desktop checks) - passed
Unit Tests against Core co-24.04 (Without Cypress checks) - old one, didn't run

PR #8510, opened 2024-03-11
Merge onContextMenu and onLongPress events.

gokaysatir:
Change-Id: Icd4452a30479591035282a789ad6c9ae742c1b0b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/leaflet.css
+++ b/browser/css/leaflet.css
@@ -1060,6 +1060,9 @@ input.clipboard {
 	bottom: 0;
 	right: 0;
 	z-index: 9;
+	user-select: none;
+	-webkit-touch-callout: none !important;
+	-webkit-user-select: none !important;
 }
 
 .leaflet-pane-splitter {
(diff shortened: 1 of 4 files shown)


Minion3665 (approved):
This seems slightly unreliable for me, I think there may occasionally be an issue if we have not selected the column header? However it's far better so I think we should merge it and make a followup

gokaysatir:
> This seems slightly unreliable for me, I think there may occasionally be an issue if we have not selected the column header? However it's far better so I think we should merge it and make a followup

Thank you for checking. If we don't select the column header then nothing happens for now. Because the handler is not passing the context menu events to CanvasSectionContainer. contextmenu events are not mirrored for some reason.

Also, even if context menu event is sent to the CSC, then the target section will handle it, not the header section. Other sections i think don't have a handler for contextmenu for now. Or only 1 has it or so.

gokaysatir:
> This seems slightly unreliable for me, I think there may occasionally be an issue if we have not selected the column header? However it's far better so I think we should merge it and make a followup

By the way, one last bit: Our canvas and the map's HTML element overlap "almost" entirely. Only the Calc headers don't oeverlap. That's why contextmenu events is not handled by Hammer.js there. Only canvas gets the events there.

Issue #8358, opened 2024-02-22
Calc: Row of Sheet Names Should Be "Mouse-Scrollable" + "Allow Press+Hold" To Move Faster Through Long Lists


Tex2002ans:
### Is your feature request related to a problem?

With many sheets, it's extremely awkward to scroll through the list.

Currently, you can only:

- **Click one-at-a-time** on the "Scroll Left" and "Scroll Right" buttons.
   - Causes sheet names to "move left/right ~6 characters".
      - With long sheet names, this also becomes a bit unwieldy.

To move far, you have to:

- Click-click-click multiple times.

See sample document with:

- [40.Sheets.ods](https://github.com/CollaboraOnline/online/files/14374994/40.Sheets.ods)

and the screenshot below:

![Collabora 24 04 0 0 - Calc - Row of Sheet Names](https://github.com/CollaboraOnline/online/assets/5774324/55411297-178a-4b35-9873-93bd9f37c6a8)

### Describe the solution you'd like

In LibreOffice 24.2, you are able to:

#### A. Press+Hold on "Scroll" buttons

1. **Press+Hold** on the "Scroll to previous sheet" or "Scroll to next sheet" buttons.
   - This constantly moves you in the left/right direction.

#### B. Middle Mouse-Scroll

1. Hover mouse into Sheet Name row.
   - For example, over "Sheet5".
2. Spin middle **mouse-wheel up/down**.
   - Quickly moves you left/right throughout the list.

Same exact thing should be possible in Online Calc.  :)

### Describe alternatives you've considered

Learning keyboard shortcuts to flip through Previous/Next Sheets.

### Additional context

I tested in today's:

> COOLWSD version: 24.04.0.0 (git hash: ff61e2a5)
> LOKit version: Collabora Office 23.05.9.1 (git hash: 0313617)
> Served by: Linux Mint 21.1
> Server ID: 8a645485

and:

> Version: 24.2.0.3 (X86_64) / LibreOffice Community
> Build ID: da48488a73ddd66ea24cf16bbc4f7b9c08e9bea1
> CPU threads: 8; OS: Windows 10.0 Build 22631; UI render: Skia/Raster; VCL: win
> Locale: en-US (en_US); UI: en-US
> Calc: CL threaded

Darshan-upadhyay1110:
@rparth07 i see you did some work on this issue. No we have fully replaced the w2ui library so you can add your solution now.
If you want to add a solution, i can assign you this issue...

[...]

PR #8514, opened 2024-03-11
Calc: added scrolling to row of sheet names

rparth07:
Change-Id: If0c4b45d256e98eb71d8935f506d1ff3c0e8b84d

* Resolves: #8358
* Target version: master 

### Summary
The solution introduces mouse scrolling and press + hold functionality for scrolling through records. It optimizes event handling by triggering the scrolling action after a delay when the left mouse button is pressed and held down. This is achieved by setting a timeout to execute the scrolling function after 500 milliseconds of continuous pressing. The scrolling action continues until the mouse button is released. In the scrolling function, it calls onClick handler to avoid code duplication at 100 millisecond interval.

### Checklist

### DIFF
--- a/browser/css/spreadsheet.css
+++ b/browser/css/spreadsheet.css
@@ -16,7 +16,13 @@
 	height: 100%;
 	overflow-y: hidden;
 	overflow-x: auto;
+	scrollbar-width: none; /* Firefox */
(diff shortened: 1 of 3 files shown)


eszkadev (commented):
Thank you for a contribution, unfortunately this is no longer valid code as we rework w2ui usage...

PR #8519, opened 2024-03-12
fix prettier builddir != srcdir

hcvcastro:
- config: fix eslint 'prettier' package extension
- browser: admin: ignore eslint transpilation code
- config: add 'prettier' command before bundeling process
- browser: fix format style
- browser: simplify eslint call
- browser: create eslint rule
- browser: simplify 'prettier' call
- browser: create prettier rule
- browser: create 'prettier-write' rule

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -481,8 +481,8 @@ endef
 
 define run_prettier_check
 	@$(NODE) node_modules/.bin/prettier \
-		--ignore-path $(srcdir)/browser/.prettierignore \
-		--ignore-path $(srcdir)/browser/.beforeprettier \
+		--ignore-path $(srcdir)/.prettierignore \
+		--ignore-path $(srcdir)/.beforeprettier \
 		$(1)
 endef
 
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -471,6 +471,14 @@ define global_file
 		@$(NODE) node_modules/uglify-js/bin/uglifyjs $< --output $@)
 endef
 
+define run_eslint_check
+	@$(NODE) node_modules/eslint/bin/eslint.js \
+		--resolve-plugins-relative-to $(abs_builddir) \
+		--ignore-path $(srcdir)/.eslintignore \
+		--config .eslintrc \
+		--no-eslintrc $(1)
+endef
+
 .PHONY: build-cool
 
 all-local: build-cool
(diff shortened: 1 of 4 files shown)


browser/Makefile.am
```diff
@@ -471,6 +471,21 @@ define global_file
...
+		--resolve-plugins-relative-to $(abs_builddir) \
+		--ignore-path $(srcdir)/.eslintignore \
+		--config .eslintrc \
+		--no-eslintrc $(1)
+endef
+
+define run_prettier_check
+	@$(NODE) node_modules/.bin/prettier \
+		--ignore-path $(srcdir)/browser/.prettierignore \
+		--ignore-path $(srcdir)/browser/.beforeprettier \
```
Minion3665, line 485: Should this be like this instead?
```suggestion
		--ignore-path $(srcdir)/.prettierignore \
		--ignore-path $(srcdir)/.beforeprettier \
```

configure.ac
```diff
@@ -1403,6 +1403,8 @@ AC_SUBST(ENABLE_SETCAP)
 
 AC_CONFIG_LINKS([discovery.xml:discovery.xml])
 AC_CONFIG_LINKS([browser/package.json:browser/package.json])
+AC_CONFIG_LINKS([browser/.eslintrc:browser/.eslintrc])
+AC_CONFIG_LINKS([browser/.prettierrc:browser/.prettierrc])
```
Minion3665, line 1407: Should we be linking .beforeprettier here too?

Minion3665 (changes requested):
Mostly good, however we can't merge it just yet because it seems that .beforeprettier isn't being respected ... I wonder if this is down to an extra browser/ in the path we give to prettier?

Overall though, happy to see this - it very nearly works so thank you!

browser/Makefile.am
hcvcastro: Ok, I will read documentation if I find some parameter

configure.ac
hcvcastro: yes, but prefer to check documentation first

browser/Makefile.am
hcvcastro: Oh, yes, you are right

Minion3665 (commented):
Tentative yes, it seems to be working fine for me but I'm still nervous about the AC_CONFIG_LINKS section. Let's resolve that conversation and then I will approve

hcvcastro:
sure, you can check the automake documentation https://www.gnu.org/software/automake/manual/1.10.3/automake.html

configure.ac
Minion3665: so, I've read the documentation, I'm still not entirely sure why .eslintrc/.prettierrc are included and .eslintignore/.prettierignore/.beforeprettier are not, is there something I'm missing here?

Minion3665: nevermind, maybe this is fine, we do `--ignore-path $(srcdir)/.beforeprettier \` and the like so that includes the srcdir there... alright, I think this is fine then

Minion3665 (approved):

PR #8531, opened 2024-03-13
annotation: fix scaled comment positioning

lpranam:
problem:
when using display scaling/zooming clicking on comment may cause overlapping on comments if the previous comment from clicked comment is long

Change-Id: I6e9eec3ba5c3fb9f871ce42bb0a554d7e2cc3c5e

* Target version: master 

### Checklist


gokaysatir (approved):
Nice catch :) Thanks @lpranam.

PR #8537, opened 2024-03-13
Profile Zones

Ashod:
- wsd: better construction of ProfileZone
- wsd: better encapsulation in ProfileZone
- wsd: emit ProfileZone recording on demand
- wsd: streamline TraceEvent strings
- wsd: better empty string operations
- wsd: support ProfileZone in CheckFileInfo
- wsd: avoid duplicate string copying for ProfileZone
- wsd: corrections to TraceEvent
- wsd: add ProfileZone to WopiStorage::uploadLocalFileToStorageAsync

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -64,7 +64,7 @@ static std::atomic<bool> ForwardSigUsr2Flag(false); //< Flags to forward SIG_USR
 #endif
 
 static size_t ActivityStringIndex = 0;
-static std::string ActivityHeader = "";
+static std::string ActivityHeader;
 static std::array<std::string, 16> ActivityStrings;
 static bool UnattendedRun = false;
 #if !MOBILEAPP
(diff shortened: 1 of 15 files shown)


mmeeks (approved):
Looks fine to me. Some follow-ons:

1. I'm not hyper-enthusiastic about generating & building all these std::strings when recordingOn is not on =). I wonder - can we improve that like we do with logging - with macros ?

2. I think -every- remote http request we do should have a profile-zone; so can we just include that into the async API implementation there: take a const char * for a description we have for the thing, and dump some of the debugging internals into that as we build th ezone.

3. this 'getThreadId' - we should use the faster & better Util::getThreadId() method that uses a thread local variable to avoid the getpid() system call. then we can (I guess) ignore the recordingOn conditional to avoid that syscall.

And I assume that you did test this in the Chrome trace profile viewer to see what is going on ? =)

mmeeks:
CI broken - let get this merged.

PR #8544, opened 2024-03-14
Convert ColorPicker.js to ColorPicker.ts

Minion3665:
I only did the things necessary for typescript to stop complaining, or
rather: there is more we could do such as modernizing the class to avoid
using L...extend, etc.

However, I wanted to have a good example of a simple-to-convert
typescript file, and this is that


browser/src/control/ColorPicker.ts
```diff
@@ -378,11 +385,11 @@ L.ColorPicker = L.Class.extend({
...
 	},
 
-	_getSampleElement: function (index, type) {
+	_getSampleElement: function (index: number, type: ColorTypes) {
 		var sampleId;
-		if (type === L.ColorPicker.BASIC_COLOR) {
+		if (type === "BASIC_COLOR") {
 			sampleId = this._basicColorSampleIdTag + index;
-		} else if (type === L.ColorPicker.TINT) {
+		} else if (type === "TINT") {
```
eszkadev, line 392: do we allow " now ?

eszkadev:
yes, it would be good to also remove L.Class.extend as we probably don't use it's features anyway.
If that would be present in this PR we can mail the team with this example

browser/src/control/ColorPicker.ts
Minion3665: no, but I'm used to doing it and forgot to run the formatter - I'll add it to a hook or something

Minion3665:
> yes, it would be good to also remove L.Class.extend as we probably don't use it's features anyway. If that would be present in this PR we can mail the team with this example

Makes sense, I have done that - I was worried about it getting noisy and overly complicated but I did find some times where I had incorrectly typed something so I guess I should've done it from the start. Ahh well, at least it's done now :)

eszkadev (approved):
Awesome, thanks

PR #8536, opened 2024-03-13
Revert "cypress: disable mobile, idle, multi tests for now"

eszkadev:
This reverts commit aa259d73378c19317ca08573536d3cc69ebf17ee.

PR #8545, opened 2024-03-14
Fix Cypress mobile tests

neilguertincollabora:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -44,8 +44,8 @@ L.Clipboard = L.Class.extend({
 		div.setAttribute('contenteditable', 'true');
 		div.setAttribute('type', 'text');
 		div.setAttribute('style', 'position: fixed; left: 0px; top: -200px; width: 15000px; height: 200px; ' +
-				 'overflow: hidden; z-index: -1000, -webkit-user-select: text !important; display: block; ' +
-				 'font-size: 6pt">');
+				 'overflow: hidden; z-index: -1000; -webkit-user-select: text !important; display: block; ' +
+				 'font-size: 6pt');
 
 		// so we get events to where we want them.
 		var parent = document.getElementById('map');
@@ -54,6 +54,9 @@ L.Clipboard = L.Class.extend({
 		if (L.Browser.cypressTest) {
 			this._dummyPlainDiv = document.createElement('div');
 			this._dummyPlainDiv.id = 'copy-plain-container';
+			this._dummyPlainDiv.style = 'position: fixed; left: 0px; top: -400px; width: 15000px; height: 200px; ' +
+				'overflow: hidden; z-index: -1000; -webkit-user-select: text !important; display: block; ' +
+				'font-size: 6pt';
 			parent.appendChild(this._dummyPlainDiv);
 		}
 
--- a/cypress_test/Makefile.am
+++ b/cypress_test/Makefile.am
@@ -106,8 +106,8 @@ MULTIUSER_TESTS_DONE= \
 check-local: do-check
 	$(if $(wildcard $(ERROR_LOG)),$(error CypressError: some tests failed!))
 
-# TODO: enable other tests $(MOBILE_TEST_FILES_DONE) $(IDLE_TEST_FILES_DONE) $(MULTIUSER_TESTS_DONE)
-do-check: $(DESKTOP_TEST_FILES_DONE)
+# TODO: enable multiuser tests $(MULTIUSER_TESTS_DONE)
+do-check: $(DESKTOP_TEST_FILES_DONE) $(MOBILE_TEST_FILES_DONE) $(IDLE_TEST_FILES_DONE)
 	$(V)$(KILL_COMMAND) || true
 	$(if $(HEADLESS_BUILD),$(V)pkill Xvfb,)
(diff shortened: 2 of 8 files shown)


eszkadev:
could you put these fixes into https://github.com/CollaboraOnline/online/pull/8536    branch? (only for testing, we will merge it here)
so we will check on CI? because here we don't run mobile check

neilguertincollabora:
> could you put these fixes into #8536 branch? (only for testing, we will merge it here) so we will check on CI? because here we don't run mobile check

Done. I did run them locally, obviously, but it's good to test them here in CI too.

eszkadev:
Failed on desktop step:
    impress/apply_paragraph_props_text_
    writer/invalidations_spec.js

Retriggered

cypress_test/integration_tests/common/impress_helper.js
```diff
@@ -124,7 +124,7 @@ function removeShapeSelection() {
 
 		return cy.cGet('.leaflet-overlay-pane svg')
 		.then(function(overlay) {
-				return overlay.children('g').length === 0;
```
gokaysatir, line 127: This change doesn't seem reliable to me. Shape selection may persist after the click and "cursor" won't exist. This will return true if the shape is selected + its cursor is not visible.

Checking 'g' element is more reliable here. Because it exists if the shape is selected.

If we want to distinguish the "shape selected" and "shape text being edited" cases, i think we need another function. Or if this is used only by one place, then we can rename this to something like "cancel shape text edit" etc etc.

cypress_test/Makefile.am
```diff
@@ -106,8 +106,7 @@ MULTIUSER_TESTS_DONE= \
 check-local: do-check
 	$(if $(wildcard $(ERROR_LOG)),$(error CypressError: some tests failed!))
 
-# TODO: enable other tests $(MOBILE_TEST_FILES_DONE) $(IDLE_TEST_FILES_DONE) $(MULTIUSER_TESTS_DONE)
```
gokaysatir, line 109: Shall we keep ToDo for multiuser tests?
"MULTIUSER_TESTS_DONE"

gokaysatir (commented):
Thanks @neilguertincollabora :) A few comments, can you check when you have time?

cypress_test/integration_tests/common/impress_helper.js
neilguertincollabora: @lpranam and I have both been struggling with this. There is sometimes an extra g element that means the check was never returning true. This was the only reliable query I could find.

What this really needs is some changes to the DOM to make shape selection/text selection/no selection more detectable, then updates to the tests to use a new simpler query. That's too much to do here in this change right now.

browser/src/map/Clipboard.js
```diff
@@ -45,7 +45,7 @@ L.Clipboard = L.Class.extend({
 		div.setAttribute('type', 'text');
 		div.setAttribute('style', 'position: fixed; left: 0px; top: -200px; width: 15000px; height: 200px; ' +
 				 'overflow: hidden; z-index: -1000, -webkit-user-select: text !important; display: block; ' +
```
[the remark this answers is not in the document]
neilguertincollabora: Fixed

cypress_test/Makefile.am
gokaysatir: I see, ooks like multiuser tests failed for this PR but it may be a temp one i guess.

cypress_test/integration_tests/common/impress_helper.js
gokaysatir: Let me also check this, i am wondering about the extra g element. Maybe we can find a formula for that. We have time for now, since the tests haven't passed yet.

gokaysatir: I think there are 2 scenarios for this:
* Text edit is enabled on the selected shape. A click to somewhere else disables the text edit but doesn't de-select the shape.
* A click to somewhere else selects another shape.
* These 2 scenarios would still show a g element in the div.
* If text selection is not disabled, then there are 2 g elements. But i think the extra g element you saw was due to selecting another shape.

I think we can use:
* cy.cGet('body').type('{esc}'); -> To disable text editing on shape.
* Then one more:
* cy.cGet('body').type('{esc}'); -> To deselect the shape.

This should be safe and shouldn't cause a leftover element.

neilguertincollabora: Did you try this? I couldn't get the escape keypress to work.

Also I don't think it's from selecting another shape. As far as I can tell there's only one shape and the click is in empty space outside the slide anyway.

gokaysatir: We can merge this and you may prefer a follow up for this if you wish? Since this passed CI. Maybe a follow up will be more comfortable?

neilguertincollabora: Yeah let's merge this to get mobile tests running for all the other CI jobs. Then I'll try this.

gokaysatir: Okay :)

gokaysatir (approved):
Thanks for the improvements & fixes :)

PR #8547, opened 2024-03-15
browser: do not hide the document title when resize

hcvcastro:
Change-Id: I55f5aa0ccd8938feea3b64c5d1422bb8502adb2b

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -68,12 +68,6 @@
 	align-self: center;
 }
 
-@media screen and (max-width: 1238px) {
-	.main-nav.hasnotebookbar:not(.readonly) #document-titlebar {
-		display: none;
-	}
-}
-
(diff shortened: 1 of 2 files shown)


eszkadev:
@hcvcastro I think it was intentional, by @pedropintosilva

hcvcastro:
I asked to review it. it has side effects that it is marked as a bug.

pedropintosilva:
thanks but I really think that it's better to hide elements when there is no space so we can have room to interact with other buttons. With this PR we enhance the situation for one user (now user can see the title in narrow window ) while making it unusable for other users, screenshot with this patch:

image

hcvcastro:
Ok, I think you can continue to follow-up patch to fix the remaining issues =)

pedropintosilva:
fixed with https://github.com/CollaboraOnline/online/pull/9119

PR #8549, opened 2024-03-15
Fix desktop/calc/focus_spec.js (23.05)

neilguertincollabora:
Wait for formula bar buttons before typing

Change-Id: I0603fee2acd7ba3090f7b8536a5e5d6f18a52179

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):

mmeeks:
Neil assures me that "It is a test fix, I verified that the test is passing, and that all the other failures in the job are unrelated. Meanwhile, this test is causing other jobs to fail so I'd like to not have to wait to retry for a passing job." =)

PR #8550, opened 2024-03-15
Fix desktop/calc/focus_spec.js

neilguertincollabora:
Wait for formula bar buttons before typing

This was only failing in 23.05 (https://github.com/CollaboraOnline/online/pull/8549) but the change should be made in 24.04 too

Change-Id: I4871a0bb05472892934a23306c18ff9541f4e5b4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the improvement :)

PR #8554, opened 2024-03-18
fix: data loss when shutting down while using typer

Rash419:
- Docbroker only uploads the document when the document is already saved and there are no further modifications.
- But when using typer once dockerbroker saves the document and tries to upload there are already new changes to the document. Therefore, docbroker keeps on saving this new changes and keeps on skiping the upload as there are new changes to the document; until it times out

Change-Id: I427d37a6228299006530daddebdf4365af63588b

* Resolves: # 
* Target version: master

### Checklist


mmeeks (approved):
Looks like a really good fix; lets get it into 23.05 too =)

Thanks !

Rash419:
Backport to co-23.05 branch: https://github.com/CollaboraOnline/online/pull/8558

PR #8555, opened 2024-03-18
private/jaume/reduce mobileapp

JaumePujantell:
- add isMobileApp util function to reduce ifdefs
- reduce uses of MOBILEAPP on some files


vmiklos (commented):
Please make sure CI is green, then I'm happy to read this from start to end. I'm not good at catching problems that are caught by automated tests. :-) Thanks.

JaumePujantell:
@vmiklos now it passed all the tests, the CodeQL errors are from code that already existed but is marked as new due to indentation changes or being move to a new file.

common/Log.cpp
```diff
@@ -529,10 +529,8 @@ namespace Log
         Static.setName(name);
         std::ostringstream oss;
         oss << Static.getName();
-#if !MOBILEAPP // Just one process in a mobile app, the pid is uninteresting.
-        oss << '-'
-            << std::setw(5) << std::setfill('0') << getpid();
-#endif
+        if (!Util::isMobileApp())
```
vmiklos, line 532: Nit: next time it's useful to keep such comments, this looks relevant.

common/Unit.cpp
```diff
@@ -695,10 +698,6 @@ UnitKit::~UnitKit() {}
 
 UnitKit& UnitKit::get()
 {
-#if MOBILEAPP
-    if (!GlobalKit)
-        GlobalKit = new UnitKit("UnitKit");
-#endif
```
vmiklos, line 701: Why don't you keep this inside an `if (Util::isMobileApp())` block?

Ah, I see, now calls to `get()` are changed to be `!Util::isMobileApp() && ...`, which means you don't need this.

wsd/RequestDetails.cpp
```diff
@@ -72,11 +72,9 @@ RequestDetails::RequestDetails(Poco::Net::HTTPRequest &request, const std::strin
         _proxyPrefix = it->second;
     it = request.find("Upgrade");
     _isWebSocket = it != request.end() && Util::iequal(it->second, "websocket");
-#if MOBILEAPP
     // request.getHost fires an exception on mobile.
```
vmiklos, line 75: Nit: also keep this one?

vmiklos (approved):
Looks good to me in general. I'm not sure if it's useful to not create the Unit instance in the mobile case -- now you have to put lots of "not mobile" conditions before calling into these hooks. But that's just a small detail, in general it's great to reduce the ifdef forest.

Please merge this yourself once you did a smoketext (short manual test) on Android (just make sure it's not completely broken, like loading an ODT should work, you should be able to type a few characters, no obvious behavior change before/after this PR). Thanks.

common/Unit.cpp
JaumePujantell: Actually it's because now isKitInProcess is always true in mobile, so next if is enough.

PR #8556, opened 2024-03-18
Watchdog to call probe-point when SocketPolls are slow.

mmeeks:
Change-Id: Id58efab0c0fee88a3da094765eb6b7641aec6e02

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -357,6 +357,7 @@ shared_headers = common/Common.hpp \
                  common/CommandControl.hpp \
                  common/Simd.hpp \
                  common/ThreadPool.hpp \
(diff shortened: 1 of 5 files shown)


vmiklos (approved):
Looks good to me in general. Some comments:

- if it turns out that the 100ms limit depends on machine speed, then perhaps COOL_WATCHDOG could have an int value and that would specify what should be the custom limit, instead of the 100ms default
- nothing jumps out as I code-read this, but probably good to wait for CI to finish before merging.

caolanm (approved):

PR #8557, opened 2024-03-18
add test for font input focus

JaumePujantell:
Add a cypress test for the change in commit dfe1e1b5

Change-Id: Id829d271efe7ccebf58c9f83cc67e6b1cf8b72e9


vmiklos (approved):
Looks good, thanks. I assume you tried to run the test without the fix to make sure it fails with the old/bad code. :-)

If so, next time it's useful to add a comment in the code to document how it failed.

PR #8559, opened 2024-03-18
clipboard: do no compare wth invalid metapath

JaumePujantell:
Copying from a document, closing it, reopening it and then pasting
did nothing.

The short-circuit for internal paste gave a false positive.

Change-Id: Ib21faf270d2ed03af2954f79bee8c1efbb0fd0ca

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -100,6 +100,8 @@ L.Clipboard = L.Class.extend({
 	getMetaPath: function(idx) {
 		if (!idx)
 			idx = 0;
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Looks OK, thanks.

`cypress_test/integration_tests/desktop/calc/clipboard_spec.js` has some infrastructure to set a dummy clipboard and then paste that to the document, would it be possible to cover your fix with a test there?

I guess your problem is the 'HTML paste, external case' test, just you want to adjust what's your HTML input and make sure the HTML gets pasted, internal paste doesn't kick in.

PR #8564, opened 2024-03-18
config: remove scheme document background colors

hcvcastro:
The desktop app writer, calc and impress does not change
the document background color when switch to "Dark".

If we change the color we should change all document property
styles and text background colors.

Change-Id: Ia7e192c2c21fc6e00d7e15de4f72a6300f5bf425

### DIFF
--- a/coolkitconfig.xcu
+++ b/coolkitconfig.xcu
@@ -62,11 +62,6 @@
 <!-- Light Theme -->
 <item oor:path="/org.openoffice.Office.UI/ColorScheme/ColorSchemes">
     <node oor:name="Light" oor:op="replace">
-        <node oor:name="DocColor">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>16777215</value>
-            </prop>
-        </node>
         <node oor:name="DocBoundaries">
             <prop oor:name="Color" oor:op="fuse">
                 <value>12632256</value>
@@ -381,11 +376,6 @@
 <!-- Dark Theme -->
 <item oor:path="/org.openoffice.Office.UI/ColorScheme/ColorSchemes">
     <node oor:name="Dark" oor:op="replace">
-        <node oor:name="DocColor">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>1842204</value>
-            </prop>
-        </node>
         <node oor:name="DocBoundaries">
             <prop oor:name="Color" oor:op="fuse">
                 <value>8421504</value>


pedropintosilva:
Note: this PR makes the dark mode button affect **only** ui. Document page etc remains the same AKA white page surrounded by dark mode UI which makes it harder to read and eye jump from UI <-> document

hcvcastro:
Yes, it is same as desktop, also notice that user document properties preferences (styles, fonts, colors), the user should
change the preference document background color.

derei:
> Yes, it is same as desktop, also notice that user document properties preferences (styles, fonts, colors), the user should change the preference document background color.

Agreed. But even at this stage, where dark mode leaves page white, should be merged. And in the future, settings can be added to manage document BG color.

Ornanovitch:
> Note: this PR makes the dark mode button affect **only** ui. Document page etc remains the same AKA white page surrounded by dark mode UI which makes it harder to read and eye jump from UI <-> document

This is a legit point of view, but:

1. There are [a lot](https://github.com/CollaboraOnline/online/issues?q=is%3Aissue+is%3Aopen+dark%20mode) of bugs with the "full" dark mode
2. Users who regularly switch from local editor (with dark UI) to online editor (with dark UI) do not expect the document itself to be re-colored, moreover when this behavior tends to be buggy
3. "white page surrounded by dark mode UI" is a pretty common thing: PDF viewers, draw editors, web pages... There is absolutely nothing shocking in this.

IMO, the "full" dark mode should follow the local LO behavior: a different option than the UI one, because the goal is simply not the same.

pedropintosilva:
Thanks @derei and @Ornanovitch for chiming in. The goal ultimately and already for some time is to have 2 toggles: one allowing for UI mode switch and another for Document mode switch (invert colours). So, I think we all agree on that. WTR a lot of bugs, yes I'm ware there are but we have been fixing them and finding that the majority of them came from the files and the templates that come with that and I have been thinking in a way to communicate that to the user or even better fix it.

I also don't see any reason, specially for 24.04, for this to don't be merge but it needs to pass the tests first.

Ornanovitch:
Thanks for those explanations @pedropintosilva! And thanks for the precious work of all the CO team on this great piece of software :)

pedropintosilva (commented):
Thank you @hcvcastro. I think the idea is good but the place to do it seems not the best one. We should follow https://github.com/CollaboraOnline/online/issues/9059 's plan. We shouldn't do this by changing the configuration file. We should instead make it possible to turn on/on on the user side and getting those from core side.  I think @GulsahKose is already working on this.

-----

Also thanks for the https://github.com/CollaboraOnline/online/pull/8692 (font color)

Issue #7515, opened 2023-10-23
Fix scroll direction in right-to-left mode


Minion3665:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @Minion3665 

### Detailed description and rationale
This is issue #7298

Calc has a right-to-left mode, but unfortunately it still contains many bugs. This specific task focuses on a scroll direction bug, particularly evident on a touch device, where scrolling is not switched when RTL mode is enabled. This means that swiping left on a mobile device or scrolling left with a mouse that can scroll left will still increase the column numbers you can see in the document, even though the document is actually moving right

### Code pointers
- This is believed to be a Javascript bug even though there is a similar issue in LibreOffice core: <https://bugs.documentfoundation.org/show_bug.cgi?id=93352>. We think that because we don't believe we do a roundtrip to core every time we want to scroll
- You may want to use the `isCalcRTL` function from `browser/src/layer/tile/CanvasTileLayer.js` https://github.com/CollaboraOnline/online/blob/686f9d7ccb1ba9afebfac9b6a1f8e422c2dec0ce/browser/src/layer/tile/CanvasTileLayer.js#L7250-L7252 to detect if you're in right-to-left mode

krunal-ctrl:
Hey @Minion3665, I would like to work on this, Can you assign me this issue?

Minion3665:
> Hey @Minion3665, I would like to work on this, Can you assign me this issue?

Awesome and thanks! I've assigned you to the issue, please feel free to ping me if you'd like any help/have any questions

krunal-ctrl:
Hey @Minion3665 ,  So far I have find the code responsible for drag event  and we need to set X position as opposite direction, here's what i have implemented 


```diff --git a/browser/src/dom/Draggable.js b/browser/src/dom/Draggable.js
index a442ec798e..fb27b51254 100644
--- a/browser/src/dom/Draggable.js
+++ b/browser/src/dom/Draggable.js
@@ -131,8 +131,8 @@ L.Draggable = L.Evented.extend({
 				offset.y = 0;
 			}
 		}
-		if (!offset.x && !offset.y) { return; }
-		if (L.Browser.touch && Math.abs(offset.x) + Math.abs(offset.y) < 3 && !e.autoscroll) { return; }
+		if (!offset.x && !offset.y && !this._map._docLayer.isCalcRTL()) { return; }
[...]

Issue #7298, opened 2023-09-25
calc: RTL mode scrolls the wrong direction


Minion3665:
### Describe the Bug

When you're in calc RTL mode, sliding your finger to the left swipes towards an increasing column number, the same as in LTR mode

### Steps to Reproduce

1. Open calc
2. Go to layout
3. Enable Right To Left mode
4. Swipe your finger towards the left
5. Notice as this moves the page in an unintuitive direction

### Expected Behavior

The page scrolls in the direction you move your finger

### Actual Behavior

The page scrolls the opposite way (i.e. away from your moving finger as if you were moving a scrollbar)

### Smartphone

 - **Device:** iPad generation 5
 - **OS:** iPadOS 16.6.1
 - **Browser:** Safari, also happens in the mobile app
 - **Version:** Commit d9d9ec6c4186a6a72f92be58f3b27f584709cc43, but this will also occur with the latest master which is 9b5b8c85c5345217e393d926599e3e7fd8ab3329

### Additional Context
iOS in the mobile app has an additional bug that causes it to display a black overlay ontop of the calc cells when in RTL mode. That is not a part of this ticket. You can still see the column numbers to test.

Minion3665:
Also reproducible on android... looks like the problem isn't just iOS

Minion3665:
Also reproducible on desktop... looks like the problem isn't just mobile

Minion3665:
I was not able to reproduce this on writer. It seems like the problem may be limited to calc

Minion3665:
Problem also appears to occur in LibreOffice (although there the scrollbar isn't flipped either)

Minion3665:
This was resolved as https://github.com/CollaboraOnline/online/issues/7515

PR #8565, opened 2024-03-18
Fix panning with RTL

Minion3665:
Resolves: #7515, #7298
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665:
Thanks for the review and heads-up Gökay! I'll merge this and make a new task for those I think, since as they both affect mobile view only and this is already a good improvement on what we have

PR #8575, opened 2024-03-19
unipoll cleanup - remove obsolete DocumentBroker mutex.

mmeeks:
Remove some pre-unipoll legacy superstition =)


Ashod (approved):
Thanks, @mmeeks. These are indeed pre-unipoll locks that weren't removed.

PR #8576, opened 2024-03-19
polls: switch compiled in delays to 60 seconds.

mmeeks:
Forkit forking children is done in response to socket messages, and parent process death should kill us too.

In general if we are relying on a poll to spin to achieve something, we have a performance bug; this should exacerbate them to flush them out.

Change-Id: I60d1c3b3c2532bbd686a3d3cfdea10f2a541a19a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


common/Unit.cpp
```diff
@@ -537,6 +537,8 @@ void UnitBase::exitTest(TestResult result, const std::string& reason)
 
     // Notify inheritors.
     onExitTest(result, reason);
+
+    SocketPoll::wakeupWorld();
```
Ashod, line 541: Both `UnitKit::onExitTest()` and `UnitWSD::onExitTest()` already do call `SocketPoll::wakeupWorld()`. I just double-checked.

wsd/DocumentBroker.hpp
```diff
@@ -159,6 +159,11 @@ class ChildProcess final : public WSProcess
     void setSMapsFD(int smapsFD) { _smapsFD = smapsFD;}
     int getSMapsFD(){ return _smapsFD; }
 
+    void moveSocketFromTo(const std::shared_ptr<SocketPoll> &from, SocketPoll &to)
```
Ashod, line 162: Can be static?

Ashod (approved):
Thanks, @mmeeks. Looks reasonable.

common/Unit.cpp
mmeeks: Fair cop - dropped that.

test/KitPidHelpers.cpp
```diff
@@ -120,10 +122,60 @@ void helpers::waitForKitPidsReady(
...
+    {
+        std::set<pid_t> result;
+        for (auto pid : a)
+        {
+            if (b.find(pid) != b.end())
+                result.insert(pid);
+        }
+        return result;
+    }
+}
```
neilguertincollabora, line 145: Use std::set_intersection

test/KitPidHelpers.cpp
```diff
@@ -120,10 +122,60 @@ void helpers::waitForKitPidsReady(
...
+
+        TST_LOG("Current kit processes: " << getPidList(current));
+    }
+
+    TST_LOG("Finished waiting for all previous kit processes to close"
+            << " before: " << getPidList(before)
+            << " current: " << getPidList(getKitPids())
+            << " intersection: " << getPidList(inters));
+
+    LOK_ASSERT("Timed out waiting for these kit processes to close" && pass);
```
neilguertincollabora, line 180: Use `LOK_ASSERT_MESSAGE_SILENT("message",pass)`

This will hide confusing "Timed out" messages from the log in the passing case where it didn't time out.

Or use `if(pass) { TST_LOG() } else { LOK_ASSERT_FAIL() }`, like in waitForKitProcessCount, so you can also customize the other output for passing/failing, and skip logging the intersection which will always be empty for the passing case.

test/KitPidHelpers.cpp
```diff
@@ -58,7 +59,8 @@ void helpers::logKitProcesses(const std::string& testname)
             << " Spare Kits: " << getPidList(spareKitPids));
 }
 
-void helpers::waitForKitProcessCount(
+namespace {
+void waitForKitProcessCount(
```
neilguertincollabora, line 63: Please re-add a comment about -1 being a valid input here, that ignores counting that type of pid.

Also, please remove the comments showing the default argument values below. I just noticed those are still there but are incorrect, I must have forgotten to remove them after moving things around before submitting this file last time.

mmeeks: Inlined the method instead, and removed quite some complexity there that we don't need; thanks.

common/Common.hpp
```diff
@@ -37,6 +37,7 @@ constexpr int COMMAND_TIMEOUT_MS = COMMAND_TIMEOUT_SECS * 1000;
 constexpr int CHILD_TIMEOUT_MS = COMMAND_TIMEOUT_MS;
 constexpr int CHILD_REBALANCE_INTERVAL_MS = CHILD_TIMEOUT_MS / 10;
 constexpr int POLL_TIMEOUT_MICRO_S = (COMMAND_TIMEOUT_MS / 5) * 1000;
+constexpr int POLL_FORKIT_TIMEOUT_SECS = 60;
```
Ashod, line 40: Note for the future: We should use `std::chrono` directly here for all these and avoid all the conversions in the code.

mmeeks: Yep - ideally we should have these as global variables that can be tweaked too - but that creates a load of linking / stubbing problems for different impls I think.

wsd/COOLWSD.cpp
```diff
@@ -924,7 +833,105 @@ class PrisonPoll : public TerminatingPoll
...
+
+#if MOBILEAPP
+#ifndef IOS
+std::mutex COOLWSD::lokit_main_mutex;
+#endif
+#endif
+
+std::shared_ptr<ChildProcess> getNewChild_Blocks(SocketPoll &destPoll, unsigned mobileAppDocId)
+{
+    (void)mobileAppDocId;
```
Ashod, line 846: `[[maybe_unused]]` next to the parameter.

wsd/COOLWSD.cpp
```diff
@@ -924,7 +833,105 @@ class PrisonPoll : public TerminatingPoll
...
+    const auto startTime = std::chrono::steady_clock::now();
+
+    std::unique_lock<std::mutex> lock(NewChildrenMutex);
+
+#if !MOBILEAPP
```
Ashod, line 851: I think it would be better if we split this function into mobile and non-mobile versions. The implementations are vastly different and it doesn't help us to have both together.

wsd/COOLWSD.hpp
```diff
@@ -198,13 +198,13 @@ class WSProcess
...
     std::shared_ptr<WebSocketHandler> getWSHandler() const { return _ws; }
-    std::shared_ptr<StreamSocket> getSocket() const { return _socket; };
+    std::shared_ptr<StreamSocket> getSocket() const { return _socket.lock(); };
 
 private:
     std::string _name;
     std::atomic<pid_t> _pid; //< The process-id, which can be access from different threads.
-    std::shared_ptr<WebSocketHandler> _ws;
-    std::shared_ptr<StreamSocket> _socket;
+    std::shared_ptr<WebSocketHandler> _ws;  // FIXME: should be weak ? ...
```
Ashod, line 206: `WebSocketHandler` is created in `ChildProcess`'s ctor. There are no other references to it than this.

Weak works for `_socket` because it's referenced in the `SocketPoll`. So weak is good there.

wsd/DocumentBroker.hpp
Ashod: I see now `getSocket()`.

neilguertincollabora (approved):

PR #8577, opened 2024-03-19
remove-w2ui #6: convert sheets toolbar to JSDialog

eszkadev:


eszkadev:
3 times in a row only `calc/focus_spec.js` failed - which also fails in other PRs - we can merge

gokaysatir (commented):
* On mobile readonly mode, i cannot see the tablist.
* There is JS error when adding new tabs and also while loading the document:
`"cannot call methods on tooltip prior to initialization; attempted to call method 'option' emitting event calcfunctionlist"`
* "Protect Sheet" doesn't work but i think that's another issue.

Other than these, it's beatiful:) Thanks.

gokaysatir:
I think below one should be Calc only, or it causes errors on loading of different formats (Writer for example):
* `this.map.uiManager.sheetsBar.show();`

eszkadev:
> * On mobile readonly mode, i cannot see the tablist.

Fixed.


>     * There is JS error when adding new tabs and also while loading the document:
>       `"cannot call methods on tooltip prior to initialization; attempted to call method 'option' emitting event calcfunctionlist"`
> 
>     * "Protect Sheet" doesn't work but i think that's another issue.

These 2 are present even before patch. I will try to fix/report later

eszkadev:
only calc/focus_spec.js failed, which is broken

gokaysatir (approved):
Thanks @eszkadev :)

eszkadev:
First https://github.com/CollaboraOnline/online/pull/8583 has to be merged

pedropintosilva (approved):
tested, it works, thanks!

PR #8579, opened 2024-03-19
Fix iOS build error due to call to method not available on mobile

plubius:
Just an #ifdef for mobile platforms.


plubius:
Looks like a later commit fixed this and master now builds without this patch.

PR #8580, opened 2024-03-20
Rewrite desktopHelper.insertComment()

neilguertincollabora:
The annotation tests have always been very flaky and very hard to debug. This should make them simpler and more stable.

Replace four helper functions (insertMultipleComment setupUIforCommentInsert createComment saveComment) with one (insertComment)
Simplify code flow
Remove unneeded docType conditionals
Remove unneded arguments from every call
Use force click to remove extra toolbar clicks
Add more relevant assertions

Apologies for the hard to follow diff in desktop_helper.js because it matched on a random `else`.

The diff in the spec files is all like this:
```
- insertMultipleComment('calc', 1, false, '[id=insert-insert-annotation]');
+ desktopHelper.insertComment();
```

except for the autosave tests, which do not click the save button:
```
- setupUIforCommentInsert('calc');
- createComment('writer', 'Test Comment', false, '[id=insert-insert-annotation]');
+ desktopHelper.insertComment(undefined, false);

- cy.cGet('#annotation-content-area-1').should('have.text','Test Comment');
+ cy.cGet('#annotation-content-area-1').should('have.text','some text0');
```

Change-Id: Ia4bc4ac964fd0ee754a481843d55682b87b63086

### DIFF
--- a/cypress_test/integration_tests/common/desktop_helper.js
+++ b/cypress_test/integration_tests/common/desktop_helper.js
@@ -310,91 +310,45 @@ function assertImageSize(expectedWidth, expectedHeight) {
 	cy.log('<< assertImageSize - end');
 }
 
-function createComment(docType, text, isMobile, selector) {
-	cy.log('>> createComment - start');
+function insertComment(text = 'some text0', save = true) {
+	cy.log('>> insertComment - start');
 
-	if (docType === 'draw') {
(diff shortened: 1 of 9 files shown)


gokaysatir (approved):
Thank you for the detailed improvements. We really needed this.

PR #8582, opened 2024-03-20
fixed js console error

lpranam:
problem:
in impress get into textbox editing and then try to drag and move text box, you get js error because map is null

regression from 9cf64b0726a5b8850db02363f125c9a8d285e4e4

Change-Id: I586c5f242241aa50f62c49371c994aeb7c370525

* Target version: master 

### Checklist


gokaysatir (approved):
Thanks for the fix.

PR #8593, opened 2024-03-21
Related: cool#8023 wsd, ClientSession: fix bogus <body> warning with plain text

vmiklos:
E.g. unit-copy-paste emitted a warning like this:
wsd-1576653-1576734 2024-03-20 21:28:53.448488 +0000 [ docbroker_002 ] DBG  ToClient-005: Missing <body> ...

This is a real problem in case the clipboard has HTML, but it's fine
when the clipboard is just plain text.

Fix the noise by returning early when the body of the clipboardcontent
message contains just plain text.

This works because the body never starts with plain text when HTML is
also available: the JSON case is not an issue and the entire clipboard
dump lists HTML before plain text.

Change-Id: I1ff51249fb710e3a9c127917aae4178c22738668

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -1609,6 +1609,12 @@ void ClientSession::postProcessCopyPayload(const std::shared_ptr<Message>& paylo
 {
     // Insert our meta origin if we can
     payload->rewriteDataBody([this](std::vector<char>& data) {
+            if (Util::findInVector(data, "clipboardcontent: content\ntext/plain") == 0)
+            {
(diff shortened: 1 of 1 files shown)


mmeeks (approved):

vmiklos:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/251/console

calc/focus_spec.js failed again. Trying locally, `make -C cypress_test check-desktop spec=calc/focus_spec.js` also fails for me. If I revert this PR locally, it's also failing, so it's not a new problem, let's merge this PR.

Issue #8045, opened 2024-01-17
Image properties "Rotation Angle" popup sticks on the screen after closing the Image properties menu


sebastiaanveld:
### Describe the Bug
Image properties Roration ange popup sticks on the screen after closing the Image properties menu. 
Performing this action multiple times result in also multiple "Rotation angle" popups show on the screen and stay there.

### Steps to Reproduce
1. Go to Nextcloud, open a document for editing
2. Insert an Image
3. Select Image properties
4. Select the Rotation tab
5. Change the Rotation Angle bij dragging on the rotation wheel, the popup "Rotation angle" pops up
6. Close the properties dialog
7. The Rotation angle" popup still shows on the screen

### Expected Behavior
When using the "Rotation angle" wheel the popup should show and fade when the wheel is released.

### Actual Behavior
Change the Rotation Angle bij dragging on the rotation wheel, the popup "Rotation angle" pops up and stays on the screen. When performing this action multiple times also multiple "Rotation angle" popups show on the screen and stay there.

### Screenshots

![image](https://github.com/CollaboraOnline/online/assets/37268999/19d8144b-86f4-4026-82b5-dc4e93fbe479)

![image](https://github.com/CollaboraOnline/online/assets/37268999/6b850971-71de-4dee-b284-15cdb67a6214)

### Desktop
(Please complete the following information)
 - **Collabora version:** COOLWSD version: 23.05.7.2 (git hash: f76e08a (E))
LOKit version: Collabora Office 23.05.7.2 (git hash: 6d9d5d1)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: 298b6d4d
 
 - **OS and version:** Windows 11, Dutch, latest patches
 - **Browser and version:** Firefox 121.0.1 x64

timur-g:
Thanks. Seems about popup in 23.05, it was not there in 22.05, so I marked regression. Minor, disappears when saved.

PR #8594, opened 2024-03-21
Fix permanent tooltips on dialogs

GulsahKose:
Position and Size dialog of the shape has a rotation angle widget in rotation tab. User can drag move instead only one click on the rotation angle widget. It causes a permanent tooltip on document. So It is better to handle mousedown instead click to close tooltip.

Change-Id: Idbea0bb8a56568a92f3f4d4c9ee6df1da01b7c24

* Resolves: #8045
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1714,7 +1714,7 @@ L.Control.UIManager = L.Control.extend({
 		var elem = $(element);
 		if (window.mode.isDesktop()) {
 			elem.tooltip();
-			elem.click(function() {
+			elem.on("mousedown", function() {
 				$('.ui-tooltip').fadeOut(function() {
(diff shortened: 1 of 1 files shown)


pedropintosilva:
we can wait until the CI passes to review

lpranam:
hmmm... I am seeing no different before and after this patch :(

GulsahKose:
@lpranam It would be niceto  see a screen record. It works on my side. Maybe you find another use case from drag on angle widget

eszkadev:
please rebase to pass CI

lpranam (approved):
Original problem is gone, tooltip still gets stuck sometimes inside the dialog but that is not blocker for this PR :)

eszkadev (approved):
thanks

PR #8597, opened 2024-03-21
browser: fix dialog warning alerts

hcvcastro:
* Enable security.enable_macros_execution in coolwsd.xml.
* Open a file with a macro.
* Click Help.

Expected result: the popup dialog should not close

Change-Id: I3981c1ddbb3782dd9ee43dc0c9dce282d2f21392


gokaysatir (approved):
This is sensible. I think we will need better ways to decide on when to close the current dialogs and when not to. I can't think of any better ways for now. I think this improves our behaviour.

Thanks :)

PR #8598, opened 2024-03-21
make profiling script find all coolwsds

caolanm:
there might be coolwsds launched from another service which doesn't match -u coolwsd, so use -t coolwsd instead to capture all of them

then their output might be interleaved, so switch to another log format that retains the name[pid] and stable sort on that column so we have all the matching lines for a name[pid] contiguous but ordered within that by time. After the sort, cut to drop the columns that are additional over the 'cat' format.

Change-Id: Id1d1c37001249ff348d6c09c005fb0836133258c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/scripts/profile-cool
+++ b/scripts/profile-cool
@@ -31,7 +31,13 @@ pkill -SIGUSR1 kitbroker
 echo Waiting 1 second for results
 sleep 1
 echo ---systemd journal---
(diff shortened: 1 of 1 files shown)


caolanm:
This updates this profiling script to handle two coolwsd (the -t instead of -u) services running like in a haproxy testing scenario. And then adds some script goo (commented inline) to make it understand interleaved output of multiple coolwsds.

ci failures aren't relevant here as its not use by ci

vmiklos (approved):
Looks OK.

PR #8599, opened 2024-03-21
remove-w2ui #7: PresentationBar

eszkadev:
- it is on the left below slides list
- handle disabled state in CSS for regular unotoolbuttons


gokaysatir (approved):
Thank you for the improvements.

On mobile, long press on slide in the slide sorter open context menu. I think we don't need it and it is not entirely visible anyway.

There is a hamburger menu item instead of the context menu on mobile.

Not sure if this is related to this PR.

PR #8604, opened 2024-03-22
improvment in filter by color submenu

Darshan-upadhyay1110:
Add color image preview in autofilter dialog for filter by color option

- it's not user-friendly to just see color code with a radio button.
- this patch will show color image preview for filter by color options according to the color code
- it will apply only for dialogs with filter using color codes in `Autofilter`
-Fix radio buttons are not vertically aligned in center

Some snap shot :
image
image

image
image

### DIFF
--- a/browser/src/control/jsdialog/Widget.TreeView.js
+++ b/browser/src/control/jsdialog/Widget.TreeView.js
@@ -29,7 +29,7 @@
  *     type: 'treelistbox',
  *     headers: [ { text: 'first column' }, { text: 'second' }],
  *     entries: [
- *         { row: 0, columns [ { text: 'a' }, { collapsed: 'collapsedIcon.svg' } ] },
+ *         { row: 0, columns [ { text: 'a' }, { collapsed: 'collapsedIcon.svg' }, { collapsedimage: '<BASE64 encoded PNG>' } ] },
  *         { row: 1, columns [ { text: 'a2' }, { expanded: 'expandedIcon.svg' }, selected: true ]}
  *     ]
  * }
@@ -41,7 +41,7 @@
  *     headers: [ { text: 'first column' }, { text: 'second' }],
  *     entries: [
  *         { row: 0, columns [ { text: 'a' }, { collapsed: 'collapsedIcon.svg' } ] },
- *         { row: 1, columns [ { text: 'a' }, { collapsed: 'collapsedIcon.svg' } ],
+ *         { row: 1, columns [ { text: 'a' }, { collapsed: 'collapsedIcon.svg' }, { expandedimage: '<BASE64 encoded PNG>' } ],
  * 			   children: [
  *                 { row: 2, columns [ { text: 'a2' }, { expanded: 'expandedIcon.svg' }, selected: true ]}
  *             ]
@@ -100,6 +100,15 @@ function _createRadioButton(parentContainer, treeViewData, builder, entry) {
 		radioButton.checked = true;
 
 	return radioButton;
+}
+
+function _createImageColumn(parentContainer, builder, imageUrl) {
+	var colorPreviewButton = L.DomUtil.create('img', builder.options.cssClass + ' ui-treeview-checkbox', parentContainer);
+	colorPreviewButton.src = imageUrl
+	colorPreviewButton.style.setProperty('outline', '1px solid var(--color-btn-border)');
+	colorPreviewButton.style.setProperty('vertical-align', 'middle');
+
+	return colorPreviewButton;
 }
 
 function _changeCheckboxStateOnClick(checkbox, treeViewData, builder, entry) {
@@ -209,7 +218,10 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 
 	var text = L.DomUtil.create('span', builder.options.cssClass + ' ui-treeview-cell', span);
 	for (var i in entry.columns) {
-		if (entry.columns[i].collapsed || entry.columns[i].expanded) {
+		var pngImage = entry.columns[i].collapsedimage ? entry.columns[i].collapsedimage : entry.columns[i].expandedimage;
+		if (pngImage) {
+			_createImageColumn(text, builder, pngImage);
+		} else if (entry.columns[i].collapsed || entry.columns[i].expanded) {
 			var icon = L.DomUtil.create('img', 'ui-listview-icon', text);
 
 			if (treeType === 'navigator')
@@ -374,8 +386,10 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 			var expander = L.DomUtil.create('div', builder.options.cssClass + ' ui-treeview-expander', td);
 			expander.addEventListener('click', function () { _expandTreeGrid(parentContainer); });
 		}
-
-		if (entry.columns[i].collapsed || entry.columns[i].expanded) {
+		var pngImage = entry.columns[i].collapsedimage ? entry.columns[i].collapsedimage : entry.columns[i].expandedimage;
+		if (pngImage) {
+			_createImageColumn(td, builder, pngImage);
+		} else if (entry.columns[i].collapsed || entry.columns[i].expanded) {
 			var icon = L.DomUtil.create('img', 'ui-listview-icon', td);
 			var iconId = _getCellIconId(entry.columns[i]);
 			L.DomUtil.addClass(icon, iconId + 'img');
@@ -385,7 +399,7 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 			td.innerText = entry.columns[i].text;
 
 		if (!disabled)
-			$(parentContainer).click(clickFunction);
+			$(td).click(clickFunction);
 	}
 
 	if (!disabled) {
@@ -412,7 +426,8 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 
 function _hasIcon(columns) {
 	for (var i in columns)
-		if (columns[i].collapsed !== undefined)
+		if (columns[i].collapsed !== undefined || columns[i].expanded !== undefined
+			|| columns[i].collapsedimage !== undefined || columns[i].expandedimage !== undefined)
 			return true;
 	return false;
 }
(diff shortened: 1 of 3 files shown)


pedropintosilva (commented):
It works but wouldn't it be better to avoid misuse radio buttons ?


I wonder if wouldn't be better to simply change the bellow code
```
function _createSelectionElement (parentContainer, treeViewData, entry, builder) {
	var selectionElement;
	var checkboxtype = treeViewData.checkboxtype;
	if (checkboxtype == 'radio') {
		selectionElement = _createRadioButton(parentContainer, treeViewData, builder, entry);
	}
	else {
		selectionElement = _createCheckbox(parentContainer, treeViewData, builder, entry);
	}
	return selectionElement;
}
```
so in the case it's a a we are in presence of radio within auto filter (or even better another a treeViewData with checkboxtype = radio and colorpreview = true or somesuch) then we _createColorSqr

Thoughts @eszkadev ?

But if the current PR is ok with Szymon then we can merge it and have this improvement in

eszkadev (changes requested):
Please try to not use any information from outside "data". Widget creation should be independent from the outside world.

If something is visible in the core - it means we somehow know about it there. We should put additional marker in the JSON and then use it here.

Darshan-upadhyay1110:
> Please try to not use any information from outside "data". Widget creation should be independent from the outside world.
> 
> If something is visible in the core - it means we somehow know about it there. We should put additional marker in the JSON and then use it here.

Added color-preview flag from core which will help to identify whether we need to add preview or not 
https://gerrit.libreoffice.org/c/core/+/165298

CC: @eszkadev

mmeeks:
@Darshan-upadhyay1110 would love to get this merged; marketing are missing it - bit it seems the core change is stalled in gerrit; what is the status here ?

Darshan-upadhyay1110:
> @Darshan-upadhyay1110 would love to get this merged; marketing are missing it - bit it seems the core change is stalled in gerrit; what is the status here ?

Sorry @mmeeks i was a bit busy with accessibility & other avatar image stuff, I will complete it by end of this week. Sorry for the delay 🙇🏼

Darshan-upadhyay1110:
There is a discussion with @eszkadev about current implementation and we have a better approach.

- will send direct image URL and not type like is it color preview or checkbox or radio button 
-  by using this image url we will display that color code

That is what a high level plan on this patch

mmeeks:
Wow - we propose a "colored image server" that serves arbitrary colored images with RGB URLs ? =) sounds like an unusual way to do this efficiency wise.

eszkadev:
> Wow - we propose a "colored image server" that serves arbitrary colored images with RGB URLs ? =) sounds like an unusual way to do this efficiency wise.

TreeView / ListView widget can have column with "image" type. It can be used as a button or some kind of a preview for an entry.

It is used in the autofilter popup to show preview. Having generic mechanism to show it is what I want - then all the TreeViews which use "image" columns will work too. (We need that also for conditional formatting dialog to implement mockup we have in plan. cc @codewithvk )

I want to avoid creating custom JSON for every widget instance and adding new code on the online side for every instance to handle that.

mmeeks:
> > Wow - we propose a "colored image server" that serves arbitrary colored images with RGB URLs ? =) sounds like an unusual way to do this efficiency wise.
> 
> TreeView / ListView widget can have column with "image" type. It can be used as a button or some kind of a preview for an entry.
> 
> It is used in the autofilter popup to show preview. Having generic mechanism to show it is what I want - then all the TreeViews which use "image" columns will work too. (We need that also for conditional formatting dialog to implement mockup we have in plan. cc @codewithvk )
> 
> I want to avoid creating custom JSON for every widget instance and adding new code on the online side for every instance to handle that.

Fair enough; does that mean a PNG or somesuch base64'd or a URL or ? ... I worry about the size of the JSON we serve - but that's a different problem I guess. Anyhow - lets do it and re-think later =)

eszkadev:
> 
> Fair enough; does that mean a PNG or somesuch base64'd or a URL or ? ... I worry about the size of the JSON we serve - but that's a different problem I guess. Anyhow - lets do it and re-think later =)

I edited to put URL explanation, here is it again:
URL here means - put base64 encoded image from the JSON into SRC of a img tag :)

These are very small images (icons) - shouldn't be a big problem. When it is a "stock" image we send an ID only and use some svg on the online side for it.

But sometimes we have "previews" which are rendered on the core, they are custom bitmaps - we have to send them.

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -102,6 +102,15 @@ function _createRadioButton(parentContainer, treeViewData, builder, entry) {
 	return radioButton;
 }
 
+function _createColorPreviewButton(parentContainer, builder, entry) {
+	var colorPreviewButton = L.DomUtil.create('input', builder.options.cssClass + ' ui-treeview-checkbox', parentContainer);
+	colorPreviewButton.type = 'color-preview';
```
eszkadev, line 107: I don't think it is correct value. I think we can only use standard types.
https://www.w3schools.com/html/html_form_input_types.asp

Probably it should be an img

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -102,6 +102,15 @@ function _createRadioButton(parentContainer, treeViewData, builder, entry) {
 	return radioButton;
 }
 
+function _createColorPreviewButton(parentContainer, builder, entry) {
```
eszkadev, line 105: please use some generic name (it's not only for color previews):
_createImageColumn for example

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -102,6 +102,15 @@ function _createRadioButton(parentContainer, treeViewData, builder, entry) {
...
+	var colorPreviewButton = L.DomUtil.create('input', builder.options.cssClass + ' ui-treeview-checkbox', parentContainer);
+	colorPreviewButton.type = 'color-preview';
+    colorPreviewButton.style.setProperty("background-image", "url('" + entry.image + "')");
```
eszkadev, line 108: indent

eszkadev: please use single quotes as in the past - we will switch to it again probably. to keep consistent with other examples here

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -420,6 +429,11 @@ function _hasIcon(columns) {
 function _createSelectionElement (parentContainer, treeViewData, entry, builder) {
 	var selectionElement;
 	var checkboxtype = treeViewData.checkboxtype;
+
+	if(entry.image) {
```
eszkadev, line 433: Please not we use `_createSelectionElement` only when `entry.state` is used in the JSON.
But it can be also used in other cases without `entry.state`. Better to put it before  `_createSelectionElement` in the parent function.

Darshan-upadhyay1110:
CORE changes to test this patch: https://gerrit.libreoffice.org/c/core/+/166378 

CC: @eszkadev

Darshan-upadhyay1110:
image

I see some style improvement we can do. But for 24.04 the main thing to add color-preview is done. so if all okay, we can merge this patch.

CC: @eszkadev, @pedropintosilva

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -352,7 +364,10 @@ function _headerlistboxEntry(parentContainer, treeViewData, entry, builder) {
 	var disabled = treeViewData.enabled === 'false' || treeViewData.enabled === false;
 	var singleClick = _isSingleClickActivate(treeViewData);
 	var selectionElement;
-	if (entry.state !== undefined) {
+	if (entry.image) {
+		selectionElement = _createImageColumn(parentContainer, builder, entry);
```
eszkadev, line 368: it should be created in the loop where we process columns - so we will put it in the correct place (sometimes on the start sometimes at the end or in the middle)

browser/src/control/jsdialog/Widget.TreeView.js
```diff
@@ -201,8 +210,11 @@ function _treelistboxEntry(parentContainer, treeViewData, entry, builder, isTree
 
 	var expander = L.DomUtil.create('div', builder.options.cssClass + ' ui-treeview-expander ', span);
 
-	if (entry.state !== undefined)
-		var selectionElement = _createSelectionElement(span, treeViewData, entry,builder);
+	var selectionElement;
+	if (entry.image)
+		selectionElement = _createImageColumn(parentContainer, builder, entry);
```
eszkadev, line 215: please put it in the columns processing loop - also look at the core patch comment wrt that

pedropintosilva:
I'm afraid cypress test `calc/autofilter_spec.js ` needs to be updated, as of not it fails 
```javascript
cy:command ✔  cGet	table#background
      cy:command ✘  find	img
      cy:command ✔  fail:	
                    Test failed: integration_tests/desktop/calc/autofilter_spec.js / AutoFilter / Filter by color
                    
                    Timed out retrying after 10000ms: Expected to find element: `img`, but never found it. Queried from:
                    
                                  > cy.get(#coolframe, [object Object]).its(0.contentDocument, [object Object]).find(table#background, [object Object])
                    
                    /home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04_cypress_desktop/cypress_test/integration_tests/desktop/calc/autofilter_spec.js:94:14
                      92 |         // Find the table element with ID "background"
                      93 |         cy.cGet('table#background')
                    > 94 |             .find('img') // Find all input elements inside the table
                         |              ^
                      95 |             .first() // Select the first input element
                      96 |             .click(); // Click on ...
```

eszkadev:
> I'm afraid cypress test `calc/autofilter_spec.js ` needs to be updated, as of not it fails
Works locally, it requires core patch to work.

pedropintosilva:
> > Works locally, it requires core patch to work.

Ah, of course , for got that! I'm just now started the tests locally. Thanks for clarifying.

eszkadev (approved):

PR #8613, opened 2024-03-23
calc: fixed bogus cell view cursor overlay

mcecchetti:
On switching tab not clear other view messages saved for replay.
Added cypress unit tests for checking the cell view cursor bounds.

Change-Id: I788828783a86038bb9d7f26648925b0e03bd560a


eszkadev (approved):
Looks ok, no incorrect selection after applying, test also works (fails before, pass after patch)

PR #8614, opened 2024-03-24
Fixed Autosaved comment are half visible

lpranam:
problem:
when in small window, comment was autosaved,
it wasn't collapsed but would hang on the edge of the screen half visible.
We can't collapse the comment directly as user will not be able to find again
which comment was being edited.

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/layer/tile/CommentListSection.ts
+++ b/browser/src/layer/tile/CommentListSection.ts
@@ -211,6 +211,8 @@ export class CommentSection extends CanvasSectionObject {
 		for (var i: number = 0; i < this.sectionProperties.commentList.length; i++) {
 			if (this.sectionProperties.commentList[i].sectionProperties.data.id !== 'new')
 				this.sectionProperties.commentList[i].setCollapsed();
+			if (this.sectionProperties.commentList[i].isRootComment())
+				this.collapseReplies(i, this.sectionProperties.commentList[i].sectionProperties.data.id);
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


cypress_test/integration_tests/desktop/writer/annotation_spec.js
```diff
@@ -122,6 +123,35 @@ describe(['tagdesktop'], 'Collapsed Annotation Tests', function() {
 		cy.cGet('.cool-annotation-content-wrapper').should('not.exist');
 	});
 
+	it.only('Autosave Collapse', function() {
```
gokaysatir, line 126: I think this "only" selector is forgotten.

browser/src/layer/tile/CommentListSection.ts
```diff
@@ -670,19 +670,26 @@ export class CommentSection extends CanvasSectionObject {
...
+		rootComment.updateThreadCountIndicator();
 	}
 
 	private collapseReplies(rootIndex: number, rootId: number) {
 		var lastChild = this.getLastChildIndexOf(rootId);
 
-		for (var i = lastChild; i > rootIndex; i--)
+		for (var i = lastChild; i > rootIndex; i--) {
 			this.sectionProperties.commentList[i].sectionProperties.container.style.display = 'none';
+			this.sectionProperties.commentList[i].updateThreadCountIndicator();
```
gokaysatir, line 691: Do we need this line inside for loop?

browser/src/layer/tile/CommentSection.ts
```diff
@@ -274,7 +274,7 @@ export class Comment extends CanvasSectionObject {
 		imgAuthor.setAttribute('width', this.sectionProperties.imgSize[0]);
 		imgAuthor.setAttribute('height', this.sectionProperties.imgSize[1]);
 
-		if (this.sectionProperties.docLayer._docType === 'text') {
+		if (this.sectionProperties.docLayer._docType !== 'spreadsheet') {
```
gokaysatir, line 277: This worries me somehow. Impress and Draw don't have reply counts and this seems structurally wrong. Also there may be some other code which checks reply count div (didn't check though).

gokaysatir (changes requested):
@lpranam i am compiling now and it may work which is fine. But i couldn't feel comfortable with the changes. Using reply count for multiple purposes doesn't seem fine to me:(

UpdateReplyCount functionality seems a good one i think.

I think we need incremental steps. First we need to have updateReplyCount function for Writer.

Then rename reply-count ... container to something like collapsed-mode-info.

Then we need to separate the functions like "updatereplycount" and "updatecollapsedmodeinfo" etc.

This seems like this change also works for being-modified comments. Is that intended?

Also that !=="spreadsheet" and many added if statements look error prone to me.

We have something good here, please let's turn this into more structured shape and improve step by step. We can even use different PRs i think.

browser/src/layer/tile/CommentSection.ts
lpranam: it is intended because even in Draw and Impress comments get collapsed so it is helpful to indicate when a comment is collapsed while it is being edited. On the other hand we do not collapse comments in calc

lpranam:
> This seems like this change also works for being-modified comments. Is that intended?

its on purpose, basically if textbox in comment is open for new comment, autosaved, reply or modify basically in any case it should indicate that.

gokaysatir:
When i start a reply to a root comment and click on the document before finishing it: It shows the reply count and saves the unfinished reply.
After this, if i resize the page, it shows the replies in non-collapsed mode as shown below.

gokaysatir:
Oh, i think only resizing window shows the replies non-collapsed. Doesn't need the first step.

gokaysatir (changes requested):
Let me check this again after you have time to have a look at the last issue.

I checked in general and couldn't find any other issues.

gokaysatir (changes requested):
When i start to modify a comment and type something and click on the document comment is not collapsed:

gokaysatir:
Below error continues, i tried again today. Please let me know if this should be gone:

> When i start to modify a comment and type something and click on the document comment is not collapsed:
> 
>

gokaysatir (approved):

PR #8620, opened 2024-03-25

caolanm:
The UNO command is dispatched to change the theme when loading the document, it causes serious problems the "kit" process changes the canonical view id.

So this patch avoids sending UNO command and add a parameter to set the initial theme.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -265,7 +265,26 @@ window.app = {
 
 		// @property lang: String
 		// browser language locale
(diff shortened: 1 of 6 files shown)


caolanm:
This builds on https://gerrit.libreoffice.org/c/core/+/163671 to set the desired render state early

eszkadev (approved):
Looks good

PR #8621, opened 2024-03-25
drop additional activation of dark theme

caolanm:
we now load the document in the initial theme, so we don't need this additional activation of the dark theme.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
Since https://github.com/CollaboraOnline/online/commit/ed5924c4943683f1bb5a6eeb715d5c00135c52ef and some earlier work we put the document view into the user's desired theme early so we don't need this additional activation of the dark theme

eszkadev (approved):

PR #8622, opened 2024-03-25
android: fix build error

Rash419:
- alertAllUsersInternal is not used for mobile apps

Change-Id: I44ecc10e64b98be58d5774e2ba14744815f75658

* Resolves: # 
* Target version: master

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4383,8 +4383,6 @@ int COOLWSD::innerMain()
 #endif
     }
 
-    COOLWSD::alertAllUsersInternal("close: shuttingdown");
-
     // Lots of polls will stop; stop watching them first.
     SocketPoll::shutdownWatchdog();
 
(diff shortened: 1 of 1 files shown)


timar (approved):

timar:
Interestingly CI passes even without this PR. See https://cpci.cbg.collabora.co.uk:8080/job/github_android_app_co-24.04/24/
Also, with your change the order of notifying users and shutting down is the opposite of original.

Rash419:
> Interestingly CI passes even without this PR. See https://cpci.cbg.collabora.co.uk:8080/job/github_android_app_co-24.04/24/ Also, with your change the order of notifying users and shutting down is the opposite of original.

@timar make sense because we are now using `Util::isMobile()` to early return
https://github.com/CollaboraOnline/online/blob/5c7f26ed2215d45b1768a23e527f3bab526f09f4/wsd/COOLWSD.cpp#L354
This PR is not needed, closing this PR

PR #8630, opened 2024-03-26
Show cursor in readonly mode.

gokaysatir:
Add a guard for potentially undefined functions in scroll section. Remove unused variable DEFAULT.
Allow arrow and home, end keys in readonly mode.

Change-Id: Ie695ebf2cd14f49fbb4db2b9ba8f81df5b289591

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -328,7 +328,7 @@ L.Map.Keyboard = L.Handler.extend({
 	},
 
 	_setPanOffset: function (pan) {
-		var keys = this._panKeys = {},
+		var keys = {},
 		    codes = this.navigationKeyCodes,
 		    i, len;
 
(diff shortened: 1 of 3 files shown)


browser/src/map/handler/Map.Keyboard.js
```diff
@@ -502,11 +502,8 @@ L.Map.Keyboard = L.Handler.extend({
 		var charCode = ev.charCode;
 		var keyCode = ev.keyCode;
 
-		var DEFAULT =0;
-		//var MAC=1; use this when you encounter a MAC value
-
 		if ((this.modifier == UNOModifier.ALT || this.modifier == UNOModifier.SHIFT + UNOModifier.ALT) &&
-		    keyCode >= this.keyCodes.NUM0[DEFAULT]) {
+		    keyCode >= this.keyCodes.NUM0[0]) {
```
eszkadev, line 506: This is magic number, best to use labels to avoid confusion what does it mean.
1 was for MAC binding
0 was for non-MAC code

eszkadev: I see its declaration repeated few times in many function.
We can instead move it to the top of a file as a "global" variable.

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -594,18 +591,9 @@ L.Map.Keyboard = L.Handler.extend({
...
-				}
-				map.fire('scrollby', {x: this._panKeys[key][0], y: this._panKeys[key][1]});
-			}
-			else if (key in this._panKeys && ev.shiftKey &&
-					!docLayer._textCSelections.empty()) {
-				// if there is a selection and the user wants to modify it
-				if (keyEventFn) {
-					keyEventFn('input', charCode, unoKeyCode);
-				}
+			if (['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'Home', 'End'].includes(ev.code)) {
```
eszkadev, line 594: I think `_panKeys` is not used at all? (no assignments in the code I can see).

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -594,18 +591,9 @@ L.Map.Keyboard = L.Handler.extend({
...
+				keyEventFn('input', charCode, unoKeyCode);
+				keyEventFn('up', charCode, unoKeyCode);
```
eszkadev, line 596: On keydown we send `input` and `up`.
What does happen on `keyup` ?

gokaysatir: I think we only send it as "input", as is. I guess it's not keydown and it's not keyup. Also handled as input on the core side.

I must have forgotten this while testing. Thank you.

browser/src/map/handler/Map.Keyboard.js
gokaysatir: Yes a forgotten local variable. I cleaned it.

browser/src/map/handler/Map.Keyboard.js
gokaysatir: I took the chane back. It seems that NUM0 doesn't need to be an array. We can check this later again and possibly turn it into a number from array. It was unnecessary for this PR anyway.

PR #8633, opened 2024-03-26
Fix LOUtil.checkIfImageExists reprocessing errors 

Minion3665:
Previously, we would sometimes get the same error event mulitple times,
leading to us erroneously believing that an image which loaded correctly
was invalid. This caused images to sometimes disappear, particularly
when switching to dark mode.

Additionally, some images are critical to layout so even if they fail to
load we should not 'display: none' them. We can set their src to an
invisible pixel to get a similar effect. We must assume that they have
their width/height correctly set elsewhere or things such as the broken
image would also break the layout.

I also reverted commit 1bb727c, as it was a less general fix for the same issue


Darshan-upadhyay1110 (approved):
The code looks good. I've tested the patch, and it works as expected.
Thanks, @Minion3665

PR #8639, opened 2024-03-27
Modals should always be placed on top of tooltips

pedropintosilva:
Before this commit, already opened ui-tooltips were being placed on top
of everything, including modal dailogs. Fix that, set jquery css
line to the same z-index used by non-modal dialogs.

Additionally: update browser/README file so it documents this change
and clean up vex legacy bits.

Change-Id: Ie7123822d96c594433ac197aeb9c651aa9b14461


pedropintosilva:
this is a forwardport of https://github.com/CollaboraOnline/online/pull/8446

gokaysatir (approved):

PR #8640, opened 2024-03-27
User avatar should be consistent after mode change

Darshan-upadhyay1110:
- User avatar should first consider if there is already extract info available or not
- if we already have that image source data then please do not set default value
- also when mode change we do preform some refresh function on all icons
    - `map.on('themechanged', setupIcon, this);`
- in that case as well, first check if we have user extract info or not
- and based on that set image source URL

Change-Id: Ifbd543322f5222fa7717b40a5ae565ed2f3c9891

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/LOUtil.js
+++ b/browser/src/core/LOUtil.js
@@ -118,6 +118,25 @@ L.LOUtil = {
 
 		map.on('themechanged', setupIcon, this);
 	},
+	setUserImage: function(img, map) {
+		var docType = map.getDocType();
+		// set avatar image if it exist in user extract info
+		var defaultImage = L.LOUtil.getImageURL('user.svg', docType);
+		var myViewId = map._docLayer._viewId;
+		var extraInfo = map._viewInfo[myViewId].userextrainfo;
+		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
+			// set user avatar
+			img.src = extraInfo.avatar;
+			// Track if error event is already bound to this image
+			img.addEventListener('error', function () {
+				img.src = defaultImage;
+				this.checkIfImageExists(img, true);
+			}.bind(this), {once:true});
+			return;
+		}
+		img.src = defaultImage;
+		this.checkIfImageExists(img, true);
+	},
 	getImageURL: function(imgName, docType) {
 		var defaultImageURL = this.getURL('images/' + imgName);
 		if (window.isLocalStorageAllowed) {
--- a/browser/src/control/Control.UserList.ts
+++ b/browser/src/control/Control.UserList.ts
@@ -163,17 +163,7 @@ class UserList extends L.Control {
 			img = L.DomUtil.create('img', 'avatar-img') as HTMLImageElement;
 		}
 
-		const defaultImage = L.LOUtil.getImageURL(
-			'user.svg',
-			this.map._docLayer._docType,
-		);
-
-		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
-			img.src = extraInfo.avatar;
-			img.addEventListener('error', () => (img.src = defaultImage));
-		} else {
-			img.src = defaultImage;
-		}
+		L.LOUtil.setUserImage(img, this.map);
 
 		img.alt = this.options.userAvatarAlt.replace('%user', username);
 
@@ -182,8 +172,6 @@ class UserList extends L.Control {
 		img.style.backgroundColor = 'var(--color-background-lighter)';
 
 		img.setAttribute('data-view-id', viewId.toString());
-
-		L.LOUtil.checkIfImageExists(img, true);
 
 		return img;
 	}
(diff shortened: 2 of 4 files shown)


Darshan-upadhyay1110:
Basically my idea is to replace avatar source in common place like setImage() method so that will be more consistent and fewer chances to break even if we change the mode to dark or light 

Thoughts @Minion3665 ?

Darshan-upadhyay1110:
> I wonder if we could do a setUserImage function instead?

Interesting approach. Yup that would work as well . Let me try and test it

browser/src/core/LOUtil.js
```diff
@@ -118,6 +118,21 @@ L.LOUtil = {
...
+		var extraInfo = map._viewInfo[myViewId].userextrainfo;
+		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
+			// set user avatar
+			img.src = extraInfo.avatar;
+			img.addEventListener('error', () => (img.src = defaultImage));
+			return;
+		}
+		img.src = defaultImage;
+		this.checkIfImageExists(img, true);
+	},
```
Minion3665, line 135: I notice that there's nothing for the 'themechanged' event here, I'm undecided as to whether we should add it (we don't currently have a light/dark mode avatar difference) so I definitely don't want to block on it, but I would like to flag it to make sure it's not an oversight

browser/src/core/LOUtil.js
```diff
@@ -118,6 +118,21 @@ L.LOUtil = {
...
+	setUserImage: function(img, map) {
+		var docType = map.getDocType();
+		// set avatar image if it exist in user extract info
+		var defaultImage = L.LOUtil.getImageURL('user.svg', docType);
+		var myViewId = map._docLayer._viewId;
+		var extraInfo = map._viewInfo[myViewId].userextrainfo;
+		if (extraInfo !== undefined && extraInfo.avatar !== undefined) {
+			// set user avatar
+			img.src = extraInfo.avatar;
+			img.addEventListener('error', () => (img.src = defaultImage));
```
Minion3665, line 130: we do a `checkIfImageExists` for the default image normally, but this won't happen if there was meant to be a user image and we fallback to default, this is probably an oversight

Minion3665 (commented):
Pretty good, just a couple of comments (1 blocking, 1 not)

browser/src/core/LOUtil.js
Darshan-upadhyay1110: Added `checkIfImageExist` and a safeguard that will prevent `error` event registration if it already has one
CC: @Minion3665

browser/src/core/LOUtil.js
```diff
@@ -118,6 +118,28 @@ L.LOUtil = {
...
+			img.src = extraInfo.avatar;
+			// Track if error event is already bound to this image
+			if (!img.errorEventBound) {
+				img.errorEventBound = true;
+				img.addEventListener('error', function() {
+					img.src = defaultImage;
+					this.checkIfImageExists(img, true);
+				}.bind(this));
+			}
+			return;
```
Minion3665, line 138: setUserImage is unlikely to be called twice, so this doesn't matter... I presume this is to avoid problems with the error event executing multiple times in which case you likely want something that modifies the event itself like so...
```suggestion
			// Track if error event is already bound to this image
			img.addEventListener('error', () => {
				if (e.loUtilProcessed) {
					return;
				}

				img.src = defaultImage;

				e.loUtilProcessed = true;
				// FIXME: remove this error event listener here to avoid future conflicts with the checkIfImageExists listener

				this.checkIfImageExists(img, true);
			});
			return;
```

That also has the advantage of hooking into the same thing that the checkIfImageExists error event does, so the same event retriggering also won't cause an issue there.

You would also need to do the thing in the FIXME comment to avoid future errors rerunning this error event listener.

Minion3665 (changes requested):
Thanks for the work on this, I think your solution with the error listener doesn't quite do the right thing, but I'm otherwise happy here...

browser/src/core/LOUtil.js
```diff
@@ -118,6 +118,31 @@ L.LOUtil = {
...
+			img.src = extraInfo.avatar;
+			// Track if error event is already bound to this image
+			img.addEventListener('error', function errorHandler (e) {
+				if (e.loUtilProcessed) {
+					return;
+				}
+				img.src = defaultImage;
+				e.loUtilProcessed = true;
+				// remove this error event listener here to avoid future conflicts with the checkIfImageExists listener
+				img.removeEventListener('error', errorHandler);
```
Minion3665, line 138: are you sure this will reference the right thing? I think that binding it before you add event listener will stop this from working... you may need to assign it to a variable instead

Minion3665 (changes requested):
Generally very good, I'm dubious of the line I commented on working as expected, other than that I'm happy for this to be merged

Minion3665 (approved):
This now looks good to me (p.s. Thanks for your help earlier Szymon, the once suggestion is a good one!)

eszkadev (approved):
looks good, I like it has dedicated method now :)

PR #8643, opened 2024-03-27
browser: fix scroll tab when loading document

hcvcastro:
If the element 'spreadsheet-tab-scroll' does not exist
when loads document, then scroll to the selected tab.

Change-Id: Iec86225f7ea97a149d455210921daaf91a4e761a


lpranam (approved):
Works as expected :)

PR #8645, opened 2024-03-27
Change convoluted url and name in windowed presentation

NickWingate:
Open an empty popup with a blank url, then set the content of the popup to be our slideURL wrapped inside an iframe with a user friendly title.
Focus the iframe on load so that arrowkey navigation works and forward the eventlisterner of the iframe to the parent window to watch for escape key calls

Change-Id: Ie1063095db51119fd45453b8b4b180fcacb6d7c1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.SlideShow.js
+++ b/browser/src/map/handler/Map.SlideShow.js
@@ -160,8 +160,8 @@ L.Map.SlideShow = L.Handler.extend({
 					_('OK'), null, false);
 			}
 
-			this._slideShowWindowProxy.document.write(htmlContent)
-			this._slideShowWindowProxy.document.close()
+			this._slideShowWindowProxy.document.documentElement.innerHTML = htmlContent;
+			this._slideShowWindowProxy.document.close();
 			this._slideShowWindowProxy.focus();
 
 			this._slideShowWindowProxy.onload = this._handleSlideWindowLoaded.bind(this);
--- a/browser/src/map/handler/Map.SlideShow.js
+++ b/browser/src/map/handler/Map.SlideShow.js
@@ -3,7 +3,7 @@
  * L.Map.SlideShow is handling the slideShow action
  */
 
-/* global _ */
+/* global _ sanitizeUrl */
 L.Map.mergeOptions({
 	slideShow: true
 });
@@ -211,13 +211,18 @@ L.Map.SlideShow = L.Handler.extend({
 	},
 
 	_generateSlideWindowHtml: function(title, slideURL) {
+		var sanitizer = document.createElement('div');
+		sanitizer.innerText = title;
+
+		var sanitizedTitle = sanitizer.innerHTML;
+		var sanitizedUrl = sanitizeUrl(slideURL);
 		return `
 		<!DOCTYPE html>
 		<html lang="en">
 		<head>
 			<meta charset="UTF-8">
 			<meta name="viewport" content="width=device-width, initial-scale=1.0">
-			<title>${title}</title>
+			<title>${sanitizedTitle}</title>
 			<style>
 				body, html {
 					margin: 0;
@@ -233,7 +238,7 @@ L.Map.SlideShow = L.Handler.extend({
 			</style>
 		</head>
 		<body>
-			<iframe src="${slideURL}"></iframe>
+			<iframe src="${sanitizedUrl}"></iframe>
 		</body>
 		</html>
 		`;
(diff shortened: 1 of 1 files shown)


browser/src/map/handler/Map.SlideShow.js
```diff
@@ -148,7 +148,7 @@ L.Map.SlideShow = L.Handler.extend({
 		// Windowed Presentation
 		if (this._presentInWindow) {
 
-			var popupTitle = "Windowed Presentation: " + this._map['wopi'].BaseFileName;
+			var popupTitle = _("Windowed Presentation: ") + this._map['wopi'].BaseFileName;
```
pedropintosilva, line 151: Tested on FF and chrome, works

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -190,6 +201,44 @@ L.Map.SlideShow = L.Handler.extend({
...
+				}
+				iframe {
+					width: 100%;
+					height: 100%;
+					border: none;
+				}
+			</style>
+		</head>
+		<body>
+			<iframe src="${slideURL}"></iframe>
```
pedropintosilva, line 236: The "url mask" works on chrome but not firefox which seems odd :

image

pedropintosilva:
didn't test the Impress: Present in window: Next/Prev slide key scroll window (HiDPI screen) yet

pedropintosilva:
Thanks @NickWingate ! this is awesome news, I have added some comments from my tests above I hope it helps.

browser/src/map/handler/Map.SlideShow.js
NickWingate: Firefox has automatic url navigation when we use `document.write(...)` and changes the url to the parent window. document.write may also cause additional problems (see https://developer.mozilla.org/en-US/docs/Web/API/Document/write) Solution is to use `document.documentElement.innerHTML = ...` changing only the innerHTML of the popup.

pedropintosilva (approved):

pedropintosilva:
I have rebased and forced push as there were many cypress changes that were meanwhile merged on master

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -147,16 +147,27 @@ L.Map.SlideShow = L.Handler.extend({
 	_startPlaying: function() {
 		// Windowed Presentation
 		if (this._presentInWindow) {
-			this._slideShowWindowProxy = window.open(this._slideURL, '_blank', 'popup'); // do we need to set this to null when closed or is that already done?
+
+			var popupTitle = "Windowed Presentation: " + this._map['wopi'].BaseFileName;
```
eszkadev: This has to be translated string. Please use `_('....')`.
Also everywhere we used `'` for strings here. Can we be consistent inside single file? (somehow we have enabled `"` in master now)

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -147,16 +147,27 @@ L.Map.SlideShow = L.Handler.extend({
...
+					_('Presentation was blocked. Please allow pop-ups in your browser. This lets slide shows to be displayed in separated windows, allowing for easy screen sharing.'), '',
+					_('OK'), null, false);
 			}
 
+			this._slideShowWindowProxy.document.write(htmlContent)
+			this._slideShowWindowProxy.document.close()
 			this._slideShowWindowProxy.focus();
-			this._slideShowWindowProxy.addEventListener('keydown', this._onCloseSlideWindow.bind(this));
+
+			this._slideShowWindowProxy.onload = this._handleSlideWindowLoaded();
```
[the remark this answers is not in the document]
eszkadev: `... = this._handleSlideWindowLoaded.bind(this)`

eszkadev (changes requested):
added some comments

browser/src/map/handler/Map.SlideShow.js
```diff
@@ -190,6 +201,44 @@ L.Map.SlideShow = L.Handler.extend({
...
+	_handleSlideWindowLoaded: function() {
+		const iframe = this._slideShowWindowProxy.document.querySelector('iframe');
+
+		if (iframe) {
+			iframe.contentWindow.focus();
+			iframe.contentWindow.addEventListener('keydown', this._onSlideWindowKeyPress.bind(this));
+		}
+	},
+
+	_generateSlideWindowHtml: function(title, slideURL) {
```
eszkadev, line 213: we need to sanitize these variables to avoid injecting HTML

eszkadev (approved):

PR #8646, opened 2024-03-27
Direct Label Association

Darshan-upadhyay1110:
A11y improvement

Problem

-  The labels for the `select` elements are being associated with a `div` element that surrounds the corresponding `select` element, instead of directly linking it to the `select` element.

-  This creates an unclear semantic relationship between the label and the actual form element, affecting usability for users with assistive technologies

-  this was not only issue with `select` element but for all input selection elements in JSDialogBuilder

Solution

-  `lable` and it's corresponding `input/select` element should have same `for` and `id` values
-   this patch will make `for` attr of `label` equals to `id` attr value of 'input/select' elements

Change-Id: If341174edcdb72af6fde3da14d5ed36ba0c3ae95

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1345,7 +1345,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		var radiobuttonLabel = L.DomUtil.createWithId('label', data.id + '-label', container);
 		radiobuttonLabel.textContent = builder._cleanText(data.text);
-		radiobuttonLabel.htmlFor = data.id;
+		radiobuttonLabel.htmlFor = data.id + '-input';
 
 		radiobutton.setAttribute('aria-labelledby', radiobuttonLabel.id);
 
@@ -1385,7 +1385,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		var checkboxLabel = L.DomUtil.create('label', builder.options.cssClass, div);
 		checkboxLabel.id = data.id + '-label';
 		checkboxLabel.textContent = builder._cleanText(data.text);
-		checkboxLabel.htmlFor = data.id;
+		checkboxLabel.htmlFor = data.id + '-input';
 
 		checkbox.setAttribute('aria-labelledby', checkboxLabel.id);
 
@@ -1801,7 +1801,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		var fixedtext = L.DomUtil.create('label', builder.options.cssClass, parentContainer);
 
 		if (data.labelFor)
-			fixedtext.htmlFor = data.labelFor;
+			fixedtext.htmlFor = data.labelFor + '-input';
 
 		if (data.text)
 			fixedtext.textContent = builder._cleanText(data.text);


Darshan-upadhyay1110:
Note: here at many places we have '-input' string appended, which is added in this commit to make sure to get focus back after widget update.

https://github.com/CollaboraOnline/online/commit/8cd4bbd8b96952ef4ad35502b02b876bff9f92ad

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1517,7 +1517,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 	_editControl: function(parentContainer, data, builder, callback) {
 		var edit = L.DomUtil.create('input', 'ui-edit ' + builder.options.cssClass, parentContainer);
 		edit.value = data.text;
-		edit.id = data.id;
+		edit.id = data.id + '-input';
```
eszkadev, line 1520: This cannot be done. It will break updates for JSDialog. Main container/element of every widget has to have id, exactly like in `data.id`

eszkadev (changes requested):

browser/src/control/Control.JSDialogBuilder.js
Darshan-upadhyay1110: @eszkadev to bind label with with input element we need remove `-id` which are appended here in this commit to make as same as lable `htmlFor` https://github.com/CollaboraOnline/online/commit/8cd4bbd8b96952ef4ad35502b02b876bff9f92ad

And to handle that duplication of id we should add '-input' in main conatiner div and not for input fields

Thoughts ?

Darshan-upadhyay1110: Possibly we also can not use `arie-labelledby` everywhere because it conflicts with label and element connection.

Is it okay if we change ID of parent-container div and not append `-input` into to all id's of  `input` elements ?

eszkadev: 1. widget's main container always should have id = data.id
2. widget's main container's children can have any id you can imagine, but it has to be unique
3. you cannot modify parent inside widget's "function", `parentContainer` should only be used as a pointer where to put your widget. we can modify everything below in the tree (children) not above current widget (parent)

Darshan-upadhyay1110: @eszkadev understood your points, but my question why is it OK to change element id of

 `_radiobuttonControl` https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L1334, 
`_checkboxControl`,
https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L1382

AND NOT OK to change id of 

_editControl: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L1599

> This cannot be done. It will break updates for JSDialog. Main container/element of every widget has to have id, exactly like in data.id

@eszkadev maybe i have added more confusion here but if widget will update for other control functions even if we changed id's of these elements then isn't it also ok for "edit Control".  It is also as same as other input tags, just difference is it's a editable input field 


If we should not change id here then there will be too much inconstancy for `input` elements, and it is very hard to establish relation between label and input tag

eszkadev:
So maybe let's unify them, add additional "parent" for edit's input, and call input "xxx-input"

eszkadev (approved):

PR #8647, opened 2024-03-27
calc: Fix pinch zoom positioning

Minion3665:
Previously, the document would not always properly follow your finger
when you tried to pan while in a pinch-zoom. In particular, we would
sometimes get this disatrously wrong (e.g. at some zoom levels we would
move our fingers one way and have the document go the other!).

I have rewritten the _getZoomDocPos function to avoid this, in
particular by adding a parameter to keep track of where our zoom started
from which lets us know where we need to zoom around. In the
mouse/button zoom case this doesn't change, but in the touchscreen zoom
case it informs us how much the viewport should have been moved.

I've also changed the surrounding code to use the new logic, in particular
the code which positions the mouse cursor

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -667,12 +667,6 @@ L.TileSectionManager = L.Class.extend({
 			yMin = splitPos.y;
 		}
 
-
-		pinchCenter = {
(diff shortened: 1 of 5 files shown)


eszkadev (approved):
Thanks, improves experience with touch device and zooming a lot

PR #8649, opened 2024-03-27
Related: cool#8648 clipboard: extract parseClipboard() from Control.DownloadProgress.js

vmiklos:
This way it can be reused in future code in Clipboard.js when
navigator.clipboard.write() is used, without duplication.

Change-Id: I843577e55802bcb7ad6cbdc5f83919172bc7d3e0


vmiklos:

I plan to add some _navigatorClipboardWrite() in Clipboard.js later, that would also use this new function.

caolanm (approved):

PR #8653, opened 2024-03-27
do a syscall from the watchdog

caolanm:
perf probe -x /usr/bin/coolwsd watchdog_probe
is not (yet?) universally available, at least
Ubuntu 22.04.4 doesn't link per to libtraceevent
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2019247

so use a (obscure) syscall to use as an event instead

Change-Id: I1e4b7ab702b1a25b2ecd627dea4580eef10f8a7c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1343,6 +1343,13 @@ extern "C" {
         // connect perf probes here
         LOG_TRC("Watchdog triggered");
         volatile int i = 42; (void)i;
+#if defined(__linux__) && !defined(__ANDROID__)
+        const struct timeval times[2] = {};
+        // call something fairly obscure that perf can trigger on.  futimesat
+        // look a good candidate (calling "futimesat" typically results in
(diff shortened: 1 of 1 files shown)


caolanm:
Hacking "if (msSinceEpoc - snapshot > MsToTrigger)" to "if (true)", and:

perf record -e syscalls:sys_enter_futimesat  -Fmax --call-graph dwarf,65528 -o perf.data.watchdog -a sleep 60s
perf script --force --no-inline -i perf.data.watchdog | stackcollapse-perf.pl | flamegraph.pl > /tmp/watchdog.svg

does at least capture watchdog_probe, but only the stack with that call, which might be more the artificial nature of the test, but maybe indicative that it's only whatever is calling watchdog_probe that is captured

watchdog

mmeeks (approved):
Love it - thanks =)

PR #8654, opened 2024-03-27
remove-w2ui #10: TopToolbar

eszkadev:


gokaysatir (approved):
Cool :) There is a JS error related to scrolls. I think it'd be good to solve it within this PR. Other than that, works seamlesly.

eszkadev:
> Cool :) There is a JS error related to scrolls. I think it'd be good to solve it within this PR. Other than that, works seamlesly.

yes, I fixed it in the last push. Thank you

PR #8662, opened 2024-03-29
Watchdog: re-direct USR2 when enabled to snapshot a late thread.

caolanm:
By tracking the thread-id, we can deliver a SIGUSR2 to the right thread at the right time; this avoids perf polling our uninteresting watchdog thread.

In that thread use Caolan's suitably obscure futimestat system-call, so that we can record based on that to see only slow things:

perf record -e syscalls:sys_enter_futimesat -ag --call-graph dwarf,65528

Change-Id: Iad05d8589fdc9541a7d0599f63625d2cde5fdf89
(cherry picked from commit 588aabb7c33242d82aa01f24b6a603c25fed08ca)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -671,6 +671,15 @@ namespace Util
         return replace(r, "\n", " / ");
     }
 
+    void killThreadById(int tid, int signal)
+    {
+#if defined __linux__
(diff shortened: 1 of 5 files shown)


mmeeks (approved):
Thanks for the back-port =)

PR #8663, opened 2024-03-29
writer: resize comments based on space

aszucs3:
Make comments window bigger, if needed, and possible. Selected commits grow up to MAX size, even if there is no space for it.

There is 2 constant limitation:
MAX - --annotation-max-size this is the maximum size commentText window can grow. If the comment have more text, than it will have scrollbar.

MIN - --annotation-min-size if there is not enought space, but comment have more text, this is the minimum size of commentText window.

note: commentText window can be smaller then MIN, if it has less text.

During doLayout:
first, in resetCommentsSize comments size resets to default. (max-height = MIN size.)

after comments positioned with loopUp / loopDown,
resizeComments check all comments (maybe it could be optimized)

If a comment have more text as its actual size, then it check the space between this, and next comment.. and resize this comment.

If there is not enought space after the comment, then it check for more space between this, and previous comment, and resize + move up the comment as much as needed.

Change-Id: Ie7b0971febd8c965b01ebe23125eac4eded1a742

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -36,6 +36,8 @@
 
 	/* Annotations */
 	--annotation-input-size: 100px;
+	--annotation-max-size: 1000;
+	--annotation-min-size: 100;
 }
 .focus-hidden:focus {
 	outline: none;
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
looks ok

lpranam:
Everything works fine but there is one issue with positioning.

Comment box should start from where the comment is inserted. 

I.e: in below screenshot long comment should be aligned with GlobalSessionMap. Otherwise it feels very confusing 

Screenshot_20240404_191516

lpranam (changes requested):
As explained in the other comment :)

lpranam (approved):

PR #8665, opened 2024-03-29
Private/mmeeks/bgsave

mmeeks:
A first cut; which should do nothing until you enable it =)

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -543,9 +543,20 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             }
             else if (tokens[1].find(".uno:Save") != std::string::npos)
             {
-                // Disable processing of other messages while saving document
-                InputProcessingManager processInput(getProtocol(), false);
-                return unoCommand(tokens);
+                static bool doBgSave = !!getenv("COOL_BGSAVE");
+
+                bool saving = false;
+                if (doBgSave)
+                    saving = !saveDocumentBackground(tokens);
+
+                if (!saving)
+                { // fallback to foreground save
+
+                    // Disable processing of other messages while saving document
+                    InputProcessingManager processInput(getProtocol(), false);
+                    return unoCommand(tokens);
+                }
+                return true;
             }
 
             return unoCommand(tokens);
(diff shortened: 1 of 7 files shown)


kit/Kit.cpp
```diff
@@ -1285,6 +1309,170 @@ void Document::updateActivityHeader() const
...
+
+        // hand parentSocket to the main poll
+        KitSocketPoll::getMainPoll()->insertNewSocket(parentSocket);
+        parentWs.reset();
+
+        getLOKit()->setForkedChild(true);
+
+        const auto now = std::chrono::steady_clock::now();
+        LOG_TRC("Background save process " << getpid() << " fork took " <<
+                std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() << "ms");
```
Ashod, line 1446: Chrono serializes with the units already. Removing `.count() << "ms"` should do the right thing and log "ms" as expected).

kit/ChildSession.cpp
```diff
@@ -822,6 +830,38 @@ bool ChildSession::loadDocument(const StringVector& tokens)
     return true;
 }
 
+// attempt to shutdown threads, fork and execute in the background
+bool ChildSession::saveDocumentAsync(const StringVector &tokens)
+{
+#if MOBILEAPP
```
Ashod, line 836: Perhaps we can do this above, at `static bool doBgSave = !!getenv("COOL_BGSAVE");`. It should make this already-complex function simpler.

kit/ChildSession.cpp
```diff
@@ -543,9 +543,17 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             }
             else if (tokens[1].find(".uno:Save") != std::string::npos)
             {
-                // Disable processing of other messages while saving document
-                InputProcessingManager processInput(getProtocol(), false);
-                return unoCommand(tokens);
+                static bool doBgSave = !!getenv("COOL_BGSAVE");
```
Ashod, line 546: Would we have a separate command for sync-saving when cleaning up DocBroker?

kit/ChildSession.cpp
```diff
@@ -543,9 +543,17 @@ bool ChildSession::_handleInput(const char *buffer, int length)
...
+
+                // Attempt to save in the background.
+                if (doBgSave && !saveDocumentAsync(tokens))
+                { // fallback to synchronous save
+
+                    // Disable processing of other messages while saving document
```
Ashod, line 552: I thought ".uno:Save" returned after queuing it, and it was handled in the background. Perhaps I misunderstand how saving is handled in Core.

kit/ChildSession.hpp
```diff
@@ -90,6 +90,13 @@ class ChildSession final : public Session
         return _docManager->sendFrame(msg.data(), msg.size(), WSOpCode::Text);
     }
 
+    bool sendSaveFrame(const std::string& msg)
+    {
+        bool ret = sendTextFrame(msg);
+        _docManager->handleSaveMessage(msg);
```
Ashod, line 96: Isn't it simpler to just do this line above, at line 3007, if `saveCommand` is true, after we `sendTextFrame`? Unless we expect this to be called from multiple places, or to become more complex, that is.

Ashod (approved):
Thanks, @mmeeks. Good stuff. I guess we'll enable this under 'experimental' at some point. Would be interesting to see it gets exercised in our tests. Left some questions/suggestions.

kit/Kit.cpp
mmeeks: oh nice =) fixed.

kit/ChildSession.cpp
mmeeks: Ah perhaps - we want a parameter here too for the save uno command to enable this code-path I think; TBD.

kit/ChildSession.cpp
mmeeks: Hmm; synchronous save is perhaps misleading though it mostly happens in that callback - the final status thing comes later (I think).

kit/ChildSession.hpp
mmeeks: Sure - why not; fixed.

Ashod:
Thanks for the update, @mmeeks. Good improvements.

PR #8667, opened 2024-03-29
remove-w2ui #10: TopToolbar (copy)

eszkadev:
Copy of https://github.com/CollaboraOnline/online/pull/8654
Which passed CI, but there was second cypress step added which is "Expected — Waiting for status to be reported" and required ...
I will test it here.


eszkadev:
It seems the old step is still required (but doen't run), new one runs but is not required. Configuration has to be changed.

PR #8670, opened 2024-03-30
SocketPoll: better re-entrancy protection.

mmeeks:
Remove toErase list; instead null socket pointers earlier to make things more deterministic.

Simplify toErase path, by just removing null sockets.

Check _socketPoll array bounds to cope with a re-entrant mutation - imperfect; but the fd comparison will help.

Change-Id: I27d81358a7d80b939b50ce4ccb1b2178a091a360

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -509,15 +509,19 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         if (_pollStartIndex > size - 1)
             _pollStartIndex = 0;
 
-        std::vector<int> toErase;
-
+        size_t itemsErased = 0;
         size_t i = _pollStartIndex;
         for (std::size_t j = 0; j < size; ++j)
         {
-            if (!_pollSockets[i])
+            if (i >= _pollSockets.size())
+            {
+                // re-entrancy hazard
+                LOG_DBG("Unexpected socket poll resize");
+            }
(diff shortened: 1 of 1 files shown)


net/Socket.cpp
```diff
@@ -551,25 +560,24 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
...
-            LOG_TRC("Removing " << toErase.size() << " socket" << (toErase.size() > 1 ? "s" : "")
-                                << " of " << _pollSockets.size() << " total");
-            LOG_ASSERT(_pollSockets.size() > 0);
-            LOG_ASSERT(toErase.size() <= _pollSockets.size());
-            std::sort(toErase.begin(), toErase.end(), [](int a, int b) { return a > b; });
-            for (const int eraseIndex : toErase)
+            LOG_TRC("Scanning to removing " << itemsErased << " defunct sockets from "
+                    << _pollSockets.size() << " sockets");
+
+            for (auto it = _pollSockets.begin(); it != _pollSockets.end();)
```
Ashod, line 575: Alternatively, we can start at the end and decrement. It's faster.

Ashod (approved):
Thanks, @mmeeks. Great cleanup. I never liked `toErase`; felt unnecessary.

PR #8673, opened 2024-03-31
remove-w2ui #11: Mobile top bar

eszkadev:
- converts mobile top bar to JSDialogs

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -164,7 +164,7 @@ textarea.cool-annotation-textarea {
 		top: -1px;
 	}
 }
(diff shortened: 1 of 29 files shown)


gokaysatir:
Rebased to resolve conflict.

gokaysatir:
Calc:
    * Click on formula bar. Type something. Click on check box.
        "this.sendUnoCommand is not a function" error.

Other than this, everything seems to work. Thanks :)

This issue is not a blocker and we can solve it with a one line fix i think. I will merge if it passes the CI.

gokaysatir (approved):

PR #8676, opened 2024-04-02
Remove Redundant Linkage: `aria-labelledby`

Darshan-upadhyay1110:
A11y improvements

Problem:

    - At various places, form elements are linked to the same label both through the 'for' attribute in the
associated label, and through the 'aria-labelled by' attribute in the element itself.
    -  This double linkage
can lead to confusion and redundancy, especially when using assistive technologies.

What should we do:

    - Remove Redundant Linkage: In most cases, the 'for' attribute in the label is sufficient to
    establish a clear linkage. The 'aria-labelled by' attribute can be removed in this case to avoid
    redundancy.

    - Consistent Use of Attributes: Ensure consistent use of attributes throughout the website. Use
    either 'for' or 'aria-labelled by' depending on the context and requirements, but avoid double
    linkage

Solution :

    - as we made label `for` attribute value consistent with `id` value for its corresponding `input/select` element
        - https://github.com/CollaboraOnline/online/pull/8646
    - now we can remove 'aria-labelled by' as there is already linkage present
    - there are some cases where labels are not present, we must use `aria-labelledby` in that case.
    
Change-Id: I389d271280a80835fd7345e8221e3777f8f8ed51

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -326,8 +326,6 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		spinfield.tabIndex = '0';
 		controls['spinfield'] = spinfield;
 
-		if (data.labelledBy)
-			spinfield.setAttribute('aria-labelledby', data.labelledBy);
 
 		if (data.unit && data.unit !== ':') {
 			var unit = L.DomUtil.create('span', builder.options.cssClass + ' spinfieldunit', div);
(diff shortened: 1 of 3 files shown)


eszkadev (approved):
looks ok

Darshan-upadhyay1110:
@eszkadev first need to merge https://github.com/CollaboraOnline/online/pull/8646 so that HTMLfor and input id changes got in :). then we can use this PR to remove are-lablledby :)

PR #8679, opened 2024-04-02
indireciton: migration on shutdown

Rash419:
Change-Id: Ia6e29a988d85fbb6f622f4d456ad64c6f747cb4e

* Target version: master

### DIFF
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -720,6 +720,11 @@ void Admin::pollingThread()
         }
         poll(closeMonitorMsgTimeout - elapsedMicroS);
     }
+
+    // if monitor closes early we set the _migrateMsgReceived for each docbroker
+    // so that docbroker can cleanup the documents
+    if (_closeMonitor)
+        COOLWSD::setAllMigrationMsgReceived();
 }
 
 void Admin::modificationAlert(const std::string& docKey, pid_t pid, bool value){
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -365,6 +365,8 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
             oss << "\"routeToken\"" << ':' << '"' << routeToken << '"' << ',';
             oss << "\"serverId\"" << ':' << '"' << serverId << '"' << '}';
             COOLWSD::alertUserInternal(dockey, oss.str());
+            if (SigUtil::getShutdownRequestFlag())
+                COOLWSD::setMigrationMsgReceived(dockey);
         }
         else
         {
@@ -408,6 +410,10 @@ void AdminSocketHandler::handleMessage(const std::vector<char> &payload)
             LOG_DBG("Invalid auth token");
             sendTextFrame("InvalidAuthToken " + id);
         }
+    }
+    else if(tokens.equals(0, "closemonitor"))
+    {
+       _admin->setCloseMonitorFlag();
     }
 }
 
@@ -688,6 +694,32 @@ void Admin::pollingThread()
         LOG_TRC("Admin poll for " << timeout);
         poll(timeout); // continue with ms for admin, settings etc.
     }
+
+    if (!COOLWSD::IndirectionServerEnabled)
+        return;
+
+    _model.sendShutdownReceivedMsg();
+
+    static const std::chrono::microseconds closeMonitorMsgTimeout = std::chrono::seconds(
+        COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180));
+
+    std::chrono::time_point<std::chrono::steady_clock> closeMonitorMsgStartTime =
+        std::chrono::steady_clock::now();
+    while (!_closeMonitor)
+    {
+        LOG_DBG("Waiting for migration to complete before closing the monitor");
+        const auto now = std::chrono::steady_clock::now();
+        const auto elapsedMicroS =
+            std::chrono::duration_cast<std::chrono::microseconds>(now - closeMonitorMsgStartTime);
+        if (elapsedMicroS > closeMonitorMsgTimeout)
+        {
+            LOG_WRN("Timed out waiting for the migration server to respond within the configured "
+                    "timeout of "
+                    << closeMonitorMsgTimeout);
+            break;
+        }
+        poll(closeMonitorMsgTimeout - elapsedMicroS);
+    }
 }
 
 void Admin::modificationAlert(const std::string& docKey, pid_t pid, bool value){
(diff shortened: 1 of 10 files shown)


wsd/Admin.cpp
```diff
@@ -685,6 +691,28 @@ void Admin::pollingThread()
...
         poll(timeout); // continue with ms for admin, settings etc.
     }
+
+    if (!COOLWSD::IndirectionServerEnabled)
+        return;
+
+    _model.sendShutdownReceivedMsg();
+
+    constexpr size_t sleepMs = 1000;
+    const auto migrationMsgTimeout =
```
Ashod, line 701: Can be `static`.

wsd/Admin.cpp
```diff
@@ -685,6 +691,28 @@ void Admin::pollingThread()
...
+        COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180) * 1000;
+    size_t count = migrationMsgTimeout / sleepMs;
+    for (size_t i = 0; i < count; ++i)
+    {
+        poll(std::chrono::seconds(0));
```
Ashod, line 706: Why not pass `sleepMs` to `poll`? If it has no events, it will sleep for the full duration. If some even comes through, it'll wake. Of course you would also need to replace the fixed-iterations loop with a duration loop (i.e. you would need to track how long it has passed since you started the loop, and break it when you exceed `migrationMsgTimeout`).

wsd/Admin.cpp
```diff
@@ -685,6 +691,28 @@ void Admin::pollingThread()
...
+        if (_closeMonitor)
+        {
+            LOG_INF("Received closemonitor message");
+            break;
+        }
+
+        LOG_DBG("Waiting for migration to complete before closing the monitor");
+        std::this_thread::sleep_for(std::chrono::milliseconds(sleepMs));
+    }
```
Ashod, line 715: It would be best to log a warning when the wait fails. To do that easily, you can check for `_closeMonitor` after the loop. If it's not set, log a warning.

wsd/AdminModel.cpp
```diff
@@ -1355,4 +1355,13 @@ void AdminModel::routeTokenSanityCheck()
     notify(oss.str());
 }
 
+void AdminModel::sendShutdownReceivedMsg()
+{
+    ASSERT_CORRECT_THREAD_OWNER(_owner);
+
+    std::ostringstream oss;
+    oss << "shutdown_received";
```
Ashod, line 1363: Why not just `notify("shutdown_received");` ? Surely we don't need to stream a string literal then convert it to a string.

wsd/DocumentBroker.cpp
```diff
@@ -338,6 +338,10 @@ void DocumentBroker::pollThread()
     const auto limStoreFailures =
         COOLWSD::getConfigValue<int>("per_document.limit_store_failures", 5);
 
+    bool waitingForMigrationMsg = false;
+    std::chrono::time_point<std::chrono::steady_clock> migrationMsgStartTime;
+    const auto migrationMsgTimeout = std::chrono::milliseconds(COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180) * 1000);
```
Ashod, line 343: Can be `static`.

wsd/Admin.hpp
```diff
@@ -248,6 +252,8 @@ class Admin : public SocketPoll
 
     // map to make sure only connection with unique monitor uri exists
     std::map<std::string, std::shared_ptr<MonitorSocketHandler>> _monitorSockets;
+
+    bool _closeMonitor = false;
```
Ashod, line 256: Technically, this should be `atomic_bool`. Same for `_migrateMsgReceived`.

Ashod (commented):
Thanks, @Rash419. Good stuff. Left some comments and suggestions, but overall looks good. I think it's worth avoiding the hard `sleep` and use `poll(timeout)` instead.

wsd/Admin.cpp
Rash419: > It would be best to log a warning when the wait fails. To do that easily, you can check for `_closeMonitor` after the loop. If it's not set, log a warning.

`LOG_WRN("Timeout waiting for 'closemonitor' message");` 
Is this warning fine or you are talking about other thing ?

wsd/Admin.hpp
Rash419: > Technically, this should be `atomic_bool`. Same for `_migrateMsgReceived`.

If we make _migrateMsgReceived atomic bool do we need this lock  ? 
https://github.com/CollaboraOnline/online/pull/8679/files#diff-e65ca5973009b9c5a78569b89969010bfa8b8459e26ca0a251f89f49a08a3afbR3320

Ashod: Absolutely! Changing `_migrateMsgReceived` (which is a new variable) doesn't change existing `DocBrokers` and the need for the lock. That lock is to protect `DocBrokers`.

wsd/Admin.cpp
```diff
@@ -685,6 +691,35 @@ void Admin::pollingThread()
         LOG_TRC("Admin poll for " << timeout);
         poll(timeout); // continue with ms for admin, settings etc.
     }
+
+    if (!COOLWSD::IndirectionServerEnabled)
+        return;
+
+    _model.sendShutdownReceivedMsg();
+
+    const auto static closeMonitorMsgTimeout = std::chrono::microseconds(
```
Ashod, line 700: Nit: It is customary to put `static` first.

wsd/Admin.cpp
```diff
@@ -685,6 +691,35 @@ void Admin::pollingThread()
...
+        COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180) * 1000000);
```
Ashod, line 701: Why multiply with 1000000 and do the conversion to microseconds manually, when you can just as easily create `chrono::seconds`?
```
static const auto closeMonitorMsgTimeout = std::chrono::seconds(COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180));
```

If you need it as microseconds:
```
static const std::chrono::microseconds closeMonitorMsgTimeout = std::chrono::seconds(COOLWSD::getConfigValue<int>("indirection_endpoint.migration_timeout_secs", 180));
```

Ashod (commented):
Thanks, @Rash419. Looks better. We're very close, I feel. Just left a few more suggestions to iron out some minor points.

Ashod (approved):
Thanks, @Rash419. Looks good. I wonder what we do in the way of tests. I don't see any automated tests.

Rash419:
> Thanks, @Rash419. Looks good. I wonder what we do in the way of tests. I don't see any automated tests.

@Ashod I am also not sure. I do have tests on indirection server side. Do you have any example unit test on coolwsd in your mind ? Can you give some pointer ?

PR #8681, opened 2024-04-02
Private/caolanm/backportearlytheme23

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -265,29 +265,7 @@ window.app = {
 
 		// @property lang: String
 		// browser language locale
-		lang: navigatorLang,
-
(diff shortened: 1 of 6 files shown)


hcvcastro:
@caolanm rebase please ?

hcvcastro (approved):

caolanm:
So I guess this is what a backport of the alternate would look like now, but why do we want to do this. Does it solve a specific problem? Or do we want to drop this entirely from 23.05 ?

caolanm:
It seems we want to drop this for 23.05 after all and go back to 23.05.09 state, so I'll close this again in favor of a raw revert of https://github.com/CollaboraOnline/online/pull/8827

PR #8685, opened 2024-04-02
kit: set creation date when online creates file from template

lpranam:
problem:
when online created file using WOPI clients, creation dates were never set. in online files are created using templates, even empty files are created using an empty template

Change-Id: I312d7af17bf322ca55b6e20bba344e4fc6628e66

* Target version: master 

### Checklist


caolanm:
I imagine https://gerrit.libreoffice.org/c/core/+/165705 is the core side for this

PR #8687, opened 2024-04-03
fix : Present in the window button hides only for mobile apps

codewithvk:
Change-Id: I8b2824f6a17091d1a5202051cc95b6a53013f0cf

* Resolves: # 
* Target version: master


Minion3665 (approved):
Looks good and works in my testing, thanks very much!

PR #8688, opened 2024-04-03
Hide unwanted toolbar btns for readonly in mobile & Add reaonly option for other application in mobile view

Darshan-upadhyay1110:
Hide disabled elements on mobile top-toolbar
- we disabled some specific button on mobile top toolbar
- but just disabling it won't hide that from UI
- maybe it is better to not show those buttons
- it also occupies much space in top view

Add read-only option in JSON for all application
- Read-only option was only present in writer before this patch
- this will add option for all other Apps: Calc,Draw,Impress
- There is slight issue while getting the read-only flag for MobileTopBar . 
                - `app.map. isReadOnlyMode ()`: Still not work because at point of getToolItems() no permission had be setted in permission.js file for function `isReadOnly()`
                - it means this._permission still undefined when we cal `app.map. isReadOnlyMode ()` from Control.MobileTopBar.js

### DIFF
--- a/browser/src/control/Control.MobileTopBar.js
+++ b/browser/src/control/Control.MobileTopBar.js
@@ -22,24 +22,14 @@ class MobileTopBar extends JSDialog.Toolbar {
 	}
 
 	getToolItems() {
-		var isReadOnlyMode = app.map ? app.map.isReadOnlyMode() : true;
-		var canUserWrite = !app.isReadOnly();
-
 		if (this.docType == 'text') {
 			return [
 				{type: 'toolitem', id: 'signstatus', command: '.uno:Signature', w2icon: '', text: _UNO('.uno:Signature'), visible: false},
 				{type: 'toolitem',  id: 'undo', text: _UNO('.uno:Undo'), command: '.uno:Undo', enabled: false},
 				{type: 'toolitem',  id: 'redo', text: _UNO('.uno:Redo'), command: '.uno:Redo', enabled: false},
-				{type: 'spacer', id: 'before-PermissionMode'},
-				{
-					type: 'container',
-					id: 'permissionmode-container',
-					children: [
-						{type: 'htmlcontent', id: 'PermissionMode', htmlId: 'permissionmode', text: '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp', isReadOnlyMode: isReadOnlyMode, canUserWrite: canUserWrite, visible: false},
-						{type: 'spacer', id: 'after-PermissionMode', visible: false},
-					],
-					vertical: false,
-				},
+				{type: 'spacer', id: 'before-permissionmode'},
+				this._generateHtmlItem('permissionmode'),
+				{type: 'spacer', id: 'after-permissionmode'},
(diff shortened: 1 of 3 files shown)


browser/src/control/Toolbar.js
```diff
@@ -1085,7 +1085,7 @@ L.Map.include({
 			if ($('.leaflet-cursor').is(':visible'))
 				return;
 
-			if (window.mode.isMobile()) {
+			if (window.mode.isMobile() && this.isEditMode()) {
```
eszkadev, line 1088: what about readonly mode on mobile? it will try to show/hide some buttons in formulabar. is that correct?

browser/src/control/Control.MobileTopBar.js
```diff
@@ -71,7 +72,17 @@ L.Control.MobileTopBar = L.Control.extend({
...
+				{
+					type: 'container',
+					id: 'permissionmode-container',
+					children: [
+						{type: 'htmlcontent', id: 'PermissionMode', htmlId: 'permissionmode', text: '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp', isReadOnlyMode: isReadOnlyMode, canUserWrite: canUserWrite, visible: false},
+						{type: 'spacer', id: 'after-PermissionMode', visible: false},
+					],
+					vertical: false,
+				},
+				{type: 'spacer', id: 'after-PermissionMode'},
```
eszkadev, line 85: dup

browser/src/control/Control.MobileTopBar.js
```diff
@@ -81,7 +92,17 @@ L.Control.MobileTopBar = L.Control.extend({
...
```
eszkadev, line 105: dup

browser/src/control/Control.MobileTopBar.js
```diff
@@ -92,7 +113,17 @@ L.Control.MobileTopBar = L.Control.extend({
...
```
eszkadev, line 126: dup

browser/src/control/Toolbar.js
Darshan-upadhyay1110: yes there are undo and redo buttons inside this scope  we need to hide it

browser/src/control/Control.MobileTopBar.js
eszkadev: instead of copy pasting the same item, maybe please create function for it, similar like it is done in: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.StatusBar.js#L193

browser/src/control/Toolbar.js
eszkadev: no, I mean "else" block below will be called in case of "readonly" or "non mobile"

Darshan-upadhyay1110: yes correct

Darshan-upadhyay1110:
Note for me: need to add some change in cypress_test/integration_tests/mobile/writer/toolbar_spec.js because not buttons are hidden and not disabled

eszkadev (approved):

PR #8689, opened 2024-04-03
Add simple geometry classes.

gokaysatir:
Change-Id: Ic1819ff187edc9c0d33f77b3b82054cb8b27c6dc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/geometry.ts
+++ b/browser/src/core/geometry.ts
@@ -1,0 +1,226 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+    Notes about the design:
+        * Because there are more then one definitions of rectangle and point, we need a prefix. The prefix is "Simple".
+        * This file is meant to be the base for geometry classes.
+        * Needs to keep things simple and maintainable:
+            * Interoperability between classes is important. These classes don't get other classes as inputs.
+            * There shouldn't be something like "rectangle.testSomething(otherRectangle)", instead "rectangle.testSometghing(otherRectangle.toArray())"
+            * We need this approach to keep things maintainable. These classes are not base for others for now. We shouldn't force types.
+            * Function inputs are primitives, like number, array of numbers, array of arrays of numbers, ..
+        * We have 3 types of coordinate units in Collabora Online.
+            * CSS pixels.
+            * Core pixels.
+            * twips.
+        * Core pixels are indeed equal to Canvas pixels. So:
+            * Core pixels = Canvas pixels.
+        * We are using term "core pixels" in many places.
+        * Why are there a CSS pixels and core / canvas pixels?
+            * Because now many devices have extreme pixel densities.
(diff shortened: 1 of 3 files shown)


eszkadev:
can we have mocha test together with it? so it will at least protect basic operations

gokaysatir:
> can we have mocha test together with it? so it will at least protect basic operations

Yes we will hopefully, but i am planning to add & modify mocha tests after some PRs.

browser/src/core/geometry.ts
```diff
@@ -0,0 +1,226 @@
...
+	public get cHeight(): number { return Math.round(this._height * app.twipsToPixels / app.dpiScale); }
+	public set cHeight (height: number) { this._height = Math.round(height * app.dpiScale * app.pixelsToTwips); }
+
+	public get cArea(): number { return this.cWidth * this.cHeight; }
+	public get cCenter(): number[] { return [(this.cX1 + this.cX2) / 2, (this.cY1 + this.cY2) / 2]; }
+
+	public cToArray(): number[] { return [this.cX1, this.cY1, this.cWidth, this.cHeight]; }
+
+	// CSS pixel checkers for coordinate match.
+	public cContainsPoint (point: number[]): boolean { return (Math.round(point[0]) >= this.cX1 && Math.round(point[0]) <= this.cX2 && Math.round(point[1]) >= this.cY1 && Math.round(point[1]) <= this.cY2); }
```
eszkadev, line 212: we have SimplePoint class, why do we pass regular array? Wouldn't be better to pass SimplePoint, so we will be sure inside it will use correct value?

with untyped array of just number we be still affected by programmer mistakes in conversions

gokaysatir: I wanted not to enforce it until we have consistent instances accross app. I added this thought in the comment at the start of the file.

eszkadev: Doesn't that mean we will need 2 times iterate over the full code?
1. use this Simple classes but with plain "numbers" and "arrays"
2. after all is completed, again change arrays into new types

I believe source of some of our problems is that we can use any incompatible numbers and still get "some" result.

So I would recommend to:
- on input allow plain numbers as you have, it should be always TWIP as you say
- on output (all cX(), pX(), x() methods) should return typed value. eg `type CssPixel = number;`
  it makes readability better because you don't need to look at comment to understand which unit it is,
  also will help to warn about incorrect usage

Thanks to that, if we switch code to use of new class, it will be done in single step

eszkadev: But sure, it is good enough now to at least have this version

eszkadev (approved):

browser/src/core/geometry.ts
gokaysatir: > Doesn't that mean we will need 2 times iterate over the full code?
   I don't think the we spend on this will be longer. There is no consensus yet between class types. I understand your concerns, but this way seems safer to me.

> we have SimplePoint class, why do we pass regular array? Wouldn't be better to pass SimplePoint, so we will be sure inside it will use correct value?
   Where there is a ts file, a number array or point class instance doesn't matter. Type is checked. If it is a JS file, type doesn't matter this time, it's not checked. Same thing with or without point class.

> on output (all cX(), pX(), x() methods) should return typed value. eg type CssPixel = number;
> it makes readability better because you don't need to look at comment to understand which unit it is,
> also will help to warn about incorrect usage
   This is a good idea:) But i still need to check this. Seems that CSSPixels and numbers can be added but we need to be sure that we won't get unnecessary linting warnings when  we add +1 to a CSSPixel type. I will check this. Let me note and we will return to this one.

PR #8690, opened 2024-04-03
On mobile platforms, the default poll time needs to be short

plubius:
It needs to be short as it directly affects the time needed to load a document.

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2244,8 +2244,11 @@ std::shared_ptr<KitSocketPoll> KitSocketPoll::create() // static
     std::shared_ptr<KitSocketPoll> result(new KitSocketPoll());
 
 #ifdef IOS
-    std::unique_lock<std::mutex> lock(KSPollsMutex);
-    KSPolls.push_back(result);
+    {
+        std::unique_lock<std::mutex> lock(KSPollsMutex);
+        KSPolls.push_back(result);
+    }
+    KitSocketPoll::KSPollsCV.notify_all();
 #endif
     return result;
 }
@@ -2393,7 +2396,7 @@ bool pushToMainThread(LibreOfficeKitCallback cb, int type, const char *p, void *
 #ifdef IOS
 
 std::mutex KitSocketPoll::KSPollsMutex;
-// std::condition_variable KitSocketPoll::KSPollsCV;
+std::condition_variable KitSocketPoll::KSPollsCV;
 std::vector<std::weak_ptr<KitSocketPoll>> KitSocketPoll::KSPolls;
 
 #endif
@@ -2422,13 +2425,16 @@ int pollCallback(void* pData, int timeoutUs)
         if (p)
             v.push_back(p);
     }
-    lock.unlock();
     if (v.empty())
     {
-        std::this_thread::sleep_for(std::chrono::microseconds(timeoutUs));
+        // Remove any stale elements from KitSocketPoll::KSPolls and
+        // block until an element is added to KitSocketPoll::KSPolls
+        KitSocketPoll::KSPolls.clear();
+        KitSocketPoll::KSPollsCV.wait(lock, []{ return KitSocketPoll::KSPolls.size(); });
     }
     else
     {
+        lock.unlock();
         for (const auto &p : v)
             p->kitPoll(timeoutUs);
     }
--- a/kit/Kit.hpp
+++ b/kit/Kit.hpp
@@ -165,7 +165,7 @@ public:
 
 #ifdef IOS
     static std::mutex KSPollsMutex;
-    // static std::condition_variable KSPollsCV;
+    static std::condition_variable KSPollsCV;
     static std::vector<std::weak_ptr<KitSocketPoll>> KSPolls;
 
     std::mutex terminationMutex;


mmeeks (changes requested):
Hi Patrick!

Thanks for the patch - it's exactly the sort of thing I'm trying to hunt down with my timeout change.
No interactive process should depend on what is effectively a 'sleep (5)' =) an average  2.5 seconds delay should not be in a hot code-path - so  ... can you get a stack track of all threads while it is waiting there ? or a trace log ?

Then by doing a wakeup or re-working this (I hope) we can save 2.5 seconds on average from load time which would be great :-)

Thanks !

plubius:
Unfortunately, after switching back to master, one or more of today's commits is causing the iOS to crash due to a nullptr in thread #20, name = 'docbroker_001' (see attached file). I'll see if I can see what is causing the new crash and see how things behave without my patch after I fix the crash:
[crashlog.txt](https://github.com/CollaboraOnline/online/files/14856238/crashlog.txt)

plubius:
> Unfortunately, after switching back to master, one or more of today's commits is causing the iOS to crash due to a nullptr in thread #20, name = 'docbroker_001' (see attached file). I'll see if I can see what is causing the new crash and see how things behave without my patch after I fix the crash: [crashlog.txt](https://github.com/CollaboraOnline/online/files/14856238/crashlog.txt)

Found it. Commit 96bfccd5ae06e67e5bb1a9541a970f563e09bc63 is passing a nullptr for wopiFileInfo in wsd/RequestVettingStation.cpp:389:

docBroker->addSession(clientSession, std::move(*wopiFileInfo));

Before that commit, wopiFileInfo was not deferenced so no crash.

mmeeks:
Wow - thanks for the trace; after quarter of an hour of head-scratching, I'm not much wiser here - although the FakeSocket stuff is really quite something ;-)
I can only assume that one of the polling loops we have is doing some work that needs to get done before we can progress, and it does that only after a timeout and not before it (somehow) I read through thread #4 and the main poll, but the loop does basically nothing except:

while (...) {
        mainWait.poll(waitMicroS);
        // Wake the prisoner poll to spawn some children, if necessary.
        PrisonerPoll->wakeup();
} 

which calls:

void PrisonPoll::wakeupHook()

which does nothing notable on mobile; so I can only assume it is another thread.

Bit of a pain - but could you try clobbering the various poll timeout callbacks to see if shortening one of them solves the problem; I'm most suspicious of this one:

int pollCallback(void* pData, int timeoutUs)
{
    if (timeoutUs < 0)
        timeoutUs = SocketPoll::DefaultPollTimeoutMicroS.count();
 
Perhaps there is some problem in the headless core / Kit mainloop integration that would be a great catch if it is there. If shortening just that one timeoutUs thing fixes stuff that'd be great.

Other thing is to tail the debug log output and see what happens immediately after it wakes up and gets going again.

Sorry to be a pain !

plubius:
timeoutUs in pollCallback in kit/Kit.cpp was the culprit. I've pushed an intermediate commit that just sets the "sleep if there are no documents" to 5 seconds. Not ideal, but no worse than before.

Unless there is a hurry to release a master iOS build, I'll see if I convert the sleep to a wait condition.

kit/Kit.cpp
```diff
@@ -2425,7 +2425,11 @@ int pollCallback(void* pData, int timeoutUs)
     lock.unlock();
     if (v.empty())
     {
-        std::this_thread::sleep_for(std::chrono::microseconds(timeoutUs));
```
mmeeks, line 2428: Oh wow - this is not great =) Indeed - this seems really weirdly broken =) I didn't spend nearly enough time reviewing the SocketPoll usage in the iOS app I think. It seems totally bonkers to do a whole set of separate back-to-back blocking polls in a loop - that's incomprehensible. If we have two polls - one app will only wakeup either after timeout or the other one ... anyhow; its a nonsense. This needs re-thinking; so good thing we caught it here =)

The easy hack here is to replace this 'sleep_for' which doesn't really belong in the code with:

fakeSocketPoll(nullptr, 0, timeout-in-milliseconds)

That should get woken up as we expect and not block like a sleep -> giving a perf. boost.

kit/Kit.cpp
```diff
@@ -2425,7 +2425,11 @@ int pollCallback(void* pData, int timeoutUs)
...
+        // On iOS, we cannot wait very long otherwise we are waiting here
+        // when the app opens a document on another thread.
+        // TODO: replace sleeping with a wait condition that can be
+        // interrupted if the user initiates opening a document.
+        std::this_thread::sleep_for(std::chrono::seconds(5));
     }
     else
     {
```
mmeeks, line 2435: The loop and kitPoll below is sometihng of an abomination. We should add a SocketPoll method for mobile - that can be passed this vector of SocketPolls - will suck all the FakeSocket fd's out of those - and then end up shoving them all into a int fakeSocketPoll(struct pollfd *pollfds, int nfds, int timeout) with a suitable timeout.
Then - inside that loop we can do this p->kitPoll(0); thing with a very short timeout to re-poll and dispatch the items per SocketPoll. Not beautiful, but far less problematic with multiple documents I think.

kit/Kit.cpp
```diff
@@ -2422,17 +2425,16 @@ int pollCallback(void* pData, int timeoutUs)
...
     if (v.empty())
     {
-        // On iOS, we cannot wait very long otherwise we are waiting here
-        // when the app opens a document on another thread.
-        // TODO: replace sleeping with a wait condition that can be
-        // interrupted if the user initiates opening a document.
-        std::this_thread::sleep_for(std::chrono::seconds(5));
+        // Remove any stale elements from KitSocketPoll::KSPolls and
+        // block until an element is added to KitSocketPoll::KSPolls
+        KitSocketPoll::KSPolls.clear();
```
plubius, line 2432: Seems no elements are ever removed from KitSocketPoll::KSPolls. So my assumption here is that if a call to lock() on an element returns a nullptr, it must be a stale entry.

kit/Kit.cpp
plubius: Implemented the condition in my third patch (none of the first two patch changes remain). Conveniently, there already was a static condition for iOS that was just commented out.

Now, when the iOS app is at the file browser, it blocks and waits forever until an element is added to KitSocketPoll::KSPolls. I noticed that no elements are removed from KitSocketPoll::KSPolls so I added some code to remove any stale elements before blocking and waiting.

PR #8691, opened 2024-04-03
Eliminate crash on mobile by checking if wopiFileInfo is a nullptr

plubius:


mmeeks (commented):
Ooh - hmm; not adding the session looks risky on the face of it; no client-session no document editing I think.

@Ashod can you handle?

plubius:
> Ooh - hmm; not adding the session looks risky on the face of it; no client-session no document editing I think.

Good point. Even after the lengthy timeout, the document never loads. So I set the timeout to 5 seconds like in pull request #8690 and it won't load either with this patch.

mmeeks:
Drat I was hoping for a nice, as yet un-improved performance win in #8690 :-) without a ClientSession for the view there is no view sadly; so another fix would be needed.

Ashod (commented):
Thanks, @plubius for reporting this. The fix is in making sure the `wopiFileInfo` is always created. As Michael points out, we need the `addSession` call, and since "WopiFileInfo" would not be available on mobile, not calling it would mean nothing gets loaded. I'll submit a file shortly and add you.

plubius:
Closing as this bug is fixed by pull request #8697.

PR #8692, opened 2024-04-03
Font Color (Dark view)

hcvcastro:
- **kit: fix missing theme option**
- **config: fix font color value (Dark Theme)**

### DIFF
--- a/coolkitconfig.xcu
+++ b/coolkitconfig.xcu
@@ -417,7 +417,7 @@
         </node>
         <node oor:name="FontColor">
             <prop oor:name="Color" oor:op="fuse">
-                <value>0</value>
+                <value>0xFFFFFF</value>
             </prop>
         </node>
         <node oor:name="Links">
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1569,6 +1569,7 @@ private:
         const std::string& macroSecurityLevel = session->getMacroSecurityLevel();
         const bool accessibilityState = session->getAccessibilityState();
         const std::string& userTimezone = session->getTimezone();
+        const std::string& theme = session->getTheme();
 
 #if !MOBILEAPP
         consistencyCheckFileExists(uri);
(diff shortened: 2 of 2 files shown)


kit/Kit.cpp
```diff
@@ -1593,6 +1594,9 @@ class Document final : public DocumentManagerInterface
         if (!userTimezone.empty())
             options += ",Timezone=" + userTimezone;
```
caolanm, line 1596: I get the first commit, seems very plausible

coolkitconfig.xcu
```diff
@@ -604,7 +604,7 @@
         </node>
         <node oor:name="CalcText">
             <prop oor:name="Color" oor:op="fuse">
-                <value>15658734</value>
```
caolanm, line 607: But I don't really get the second commit. Do they have a known effect, I mean "CalcText" of 15658734 is 0xEEEEEE which is pretty close to 0xFFFFFF already, and "CalcValue" of 7512015 is a bright blue of 0x729FCF and I am unsure where they get used.

Do we need the second commit to fix the known issue?

hcvcastro: I did not test CalcText and CalcValue how is used, but I changed to be sure all white.

caolanm: Searching in coolkitconfig.xcu I see that the "Light" and "Dark" colors for "CalcText" and "CalcValue" are already quite different from eachother, while "FontColor" is 0 in both. I suggest to drop those two hunks about "Calc*" we're not sure what they affect, seeing as they are already quite different in Dark/Light they are probably ok.

hcvcastro: Ok, I will drop CalcText and CalcValue, but FontColor it fixes the Writer Text font color case.

caolanm: yeah, it looks odd the have the same "FontColor" in Light and Dark.

caolanm (approved):

Issue #8686, opened 2024-04-02
UserCanWrite = false and UserCanNotWriteRelative = true still offer the edit button


hfiguiere:
### Describe the Bug
The button to edit a file when opening a .csv or .txt is still present when UserCanWrite = false and UserCanNotWriteRelative = true. And it does nothing

### Steps to Reproduce
1. Open a .txt or .csv. The WOPI CheckFileInfo should have UserCanWrite = false and UserCanNotWriteRelative = true
2. Click on the edit button (bottom right corner, pencil icon)
3. Nothing happen

### Expected Behavior
The worst: an error.
The best: the button is hidden as no action can be taken.

### Actual Behavior
Nothing happen

### Desktop
(Please complete the following information)
 - **Collabora version:** `master` commit ad5826e4c355b64f68f378d3c9c1f682febaa2a8 and 23.05
 - **OS and version:** Linux
 - **Browser and version:** Firefox

### Smartphone
(Please complete the following information)
 - **Device:** [e.g. iPhone 14]
 - **OS:** [e.g. iOS 16.5]
 - **Browser and version:** [e.g. Safari 16.4]

PR #8693, opened 2024-04-03
permission: Don't show the edit button if read-only

hfiguiere:
If wopi.UserCanWrite tells us we can't write the file no reason to show the edit button that would do nothing anyway.

Fix https://github.com/CollaboraOnline/online/issues/8686

Change-Id: I4f6d1f8d8be520e422e260f2e532cb86541e8d0f

* Resolves: #8686
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Permission.js
+++ b/browser/src/control/Permission.js
@@ -68,7 +68,7 @@ L.Map.include({
 				button.on('click', function () {
 					that._requestFileCopy();
 				});
-			} else if (!this.options.canTryLock && (window.mode.isMobile() || window.mode.isTablet())) {
+			} else if ((!window.ThisIsAMobileApp && !this['wopi'].UserCanWrite) || (!this.options.canTryLock && (window.mode.isMobile() || window.mode.isTablet()))) {
 				$('#mobile-edit-button').hide();
 			}
 


browser/src/control/Permission.js
```diff
@@ -68,7 +68,7 @@ L.Map.include({
 				button.on('click', function () {
 					that._requestFileCopy();
 				});
-			} else if (!this.options.canTryLock && (window.mode.isMobile() || window.mode.isTablet())) {
+			} else if (!this['wopi'].UserCanWrite || (!this.options.canTryLock && (window.mode.isMobile() || window.mode.isTablet()))) {
```
codewithvk, line 71: Suggestion: For check read-only mode, we can use `map.isReadOnlyMode()`

Darshan-upadhyay1110:
Not sure if my commit is related or nor but i had added same thing yesterday for read-only mode. But it was for mobile view
https://github.com/CollaboraOnline/online/pull/8688/commits/f31af7722cc59524d1b7a8f6c7ffe48879004a79

hfiguiere:
> @hfiguiere Not sure if my commit is related or nor but i had added same thing yesterday for read-only mode. But it was for mobile view [f31af77](https://github.com/CollaboraOnline/online/commit/f31af7722cc59524d1b7a8f6c7ffe48879004a79)

I saw it yesterday and it's not related.

hfiguiere:
I think it might break with mobile.

browser/src/control/Permission.js
hfiguiere: Now I question the state of `this.wopi` on Mobile.

hfiguiere: Updated the test.

hfiguiere:
> I think it might break with mobile.

Updated the PR

browser/src/control/Permission.js
codewithvk: I'm actually surprised. If we are in readonly mode, then why should we check the condition of UserCanWrite? By reading the code, it seems illogical. Correct me if I'm wrong.

hfiguiere: you can be in readonly mode because you open a .csv or .txt. See `readonlyStartingFormats`.

Then when you click the edit button, if `this['wopi'].UserCanNotWriteRelative` is false, then you get offered to save as ODF or to edit with possible losses. But the user can't write at all, there is no point.

The original issue is "we open a .cvs and the edit button is there, but it does nothing", whith `UserCanNotWriteRelative` set to true, and `UserCanWrite` to false.

caolanm (approved):
looks ok, if you can get it to pass ci

hfiguiere:
The same error occur on master, locally.

hfiguiere:
> The same error occur on master, locally.

https://github.com/CollaboraOnline/online/issues/8768

caolanm:
passes ci now at least

PR #8694, opened 2024-04-04
uno: .uno:SetDocumentProperties: add file path parameter

lpranam:
problem:
in LOK/online to support async save, files in jails may have different extensions (ie: .upload, .uploading)
this caused problem to detect files as original file name may not exist. As result property like file size were always set to 0

Change-Id: Ia4b39c04412e191742b98046c5c31e3964db7ce9

* Target version: master 

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -546,6 +546,21 @@ bool ChildSession::_handleInput(const char *buffer, int length)
                 assert(false);
                 return false;
             }
+            else if (tokens[1].find(".uno:SetDocumentProperties") != std::string::npos)
+            {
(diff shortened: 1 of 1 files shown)


caolanm (commented):
I think this is basically a good route to go after all, but see comments in https://gerrit.libreoffice.org/c/core/+/165769 I think it would be better to send FileSize instead of FilePath.

Maybe we can get the size in a more robust way from DocBroker, but even if we do a loop over options, the time window that we get that wrong is smaller here than in core I think.

caolanm (approved):
We could wish for a more straight forward way to know the current file name/file info. But looks reasonable.

caolanm:
rats, there was a conflict with another change :-(,  a trivial conflict, but we can't use the online conflict resolution because there is then no change-id on the merge fix. 

And because this is in your repo and not in a private/whatever/bar branch I can't fix it on my side. Can you do (yet another) rebase to get this rebased on top of the change that trivially "conflicts"

PR #8700, opened 2024-04-04
Use SimplePoint for cell address.

gokaysatir:
Use SimpleRectangle for cell cursor.
Add .clone function to SimplePoint.
Use app.file.calc.cellCursor.visible for checking the cursor visibility instead of null variables. Replace map._cellCursorPixels with app.file.calc.cellCursor.rectangle Remove map._cellCursorXY with app.file.calc.cellCursor.address Replace map._validatedCellXY with SimplePoint.

Change-Id: I9b1e2c92e5029b2f184055083d35200bb5e6b8c9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -702,8 +702,8 @@ export class HeaderInfo {
 		const cellSelections: cool.Rectangle[] = this._map._docLayer._cellSelections;
 		let currentIndex = -1;
 
-		if (app.file.calc.cellCursor.visible) {
-			currentIndex = this._isColumn ? app.file.calc.cellCursor.address.x: app.file.calc.cellCursor.address.y;
+		if (app.calc.cellCursorVisible) {
+			currentIndex = this._isColumn ? app.calc.cellAddress.x: app.calc.cellAddress.y;
 		}
 
 		const startPx = this._isColumn === true ? section.documentTopLeft[0]: section.documentTopLeft[1];
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -702,12 +702,8 @@ export class HeaderInfo {
 		const cellSelections: cool.Rectangle[] = this._map._docLayer._cellSelections;
 		let currentIndex = -1;
 
-		if (this._map._docLayer._cellCursorXY) {
-			currentIndex = this._isColumn ? this._map._docLayer._cellCursorXY.x: this._map._docLayer._cellCursorXY.y;
-		}
-
-		if (currentIndex === -1 && this._map._docLayer._prevCellCursorXY) {
-			currentIndex = this._isColumn ? this._map._docLayer._prevCellCursorXY.x : this._map._docLayer._prevCellCursorXY.y;
+		if (app.file.calc.cellCursor.visible) {
+			currentIndex = this._isColumn ? app.file.calc.cellCursor.address.x: app.file.calc.cellCursor.address.y;
 		}
 
 		const startPx = this._isColumn === true ? section.documentTopLeft[0]: section.documentTopLeft[1];
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -42,16 +42,14 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		}
 
 		if (!comment) {
-			var pixelStart = new L.Point(Math.ceil(this._cellCursorPixels.getX1()),
-						     Math.ceil(this._cellCursorPixels.getY1()));
+			var pixelStart = new L.Point(app.file.calc.cellCursor.rectangle.pX1, app.file.calc.cellCursor.rectangle.pY1);
 			var rangeStart = this.sheetGeometry.getCellFromPos(pixelStart, 'corepixels');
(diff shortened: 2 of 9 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2514,18 +2507,14 @@ L.CanvasTileLayer = L.Layer.extend({
...
+			let tempRectangle = this._cellCursorTwips.toRectangle();
+			app.file.calc.cellCursor.rectangle = new app.definitions.simpleRectangle(tempRectangle[0], tempRectangle[1], tempRectangle[2], tempRectangle[3]);
 			app.file.calc.cellCursor.visible = true;
+
 			app.sectionContainer.onCellAddressChanged();
 			if (autofillMarkerSection)
-				autofillMarkerSection.calculatePositionViaCellCursor([this._cellCursorPixels.getX2(), this._cellCursorPixels.getY2()]);
-
-			this._cellCursorXY = new L.Point(parseInt(strTwips[4]), parseInt(strTwips[5]));
+				autofillMarkerSection.calculatePositionViaCellCursor([app.file.calc.cellCursor.rectangle.pX2, app.file.calc.cellCursor.rectangle.pY2]);
```
eszkadev, line 2517: long names to type `app.file.calc.cellCursor.rectangle` maybe we should keep it as a member of tile layer `this.cellCursor`? (just a note to consider in later PRs for other variables, for readability)

or we can put getter here:   `cellCursor() { return app.file.calc.cellCursor; }` so it will be short

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -4774,7 +4762,7 @@ L.CanvasTileLayer = L.Layer.extend({
 	},
 
 	_addDropDownMarker: function () {
-		if (this._validatedCellXY && this._cellCursorXY && this._validatedCellXY.equals(this._cellCursorXY)) {
+		if (this._validatedCellAddress && app.file.calc.cellCursor.visible && this._validatedCellAddress.equals(app.file.calc.cellCursor.address.toArray())) {
```
eszkadev, line 4765: SimpleRectangle has method: `equals`, can we avoid `toArray` this way?

eszkadev: (seen few times in this PR)

browser/src/control/Control.Header.ts
```diff
@@ -702,12 +702,8 @@ export class HeaderInfo {
...
 		let currentIndex = -1;
 
-		if (this._map._docLayer._cellCursorXY) {
-			currentIndex = this._isColumn ? this._map._docLayer._cellCursorXY.x: this._map._docLayer._cellCursorXY.y;
-		}
-
-		if (currentIndex === -1 && this._map._docLayer._prevCellCursorXY) {
-			currentIndex = this._isColumn ? this._map._docLayer._prevCellCursorXY.x : this._map._docLayer._prevCellCursorXY.y;
+		if (app.file.calc.cellCursor.visible) {
+			currentIndex = this._isColumn ? app.file.calc.cellCursor.address.x: app.file.calc.cellCursor.address.y;
```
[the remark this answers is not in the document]
gokaysatir: Yes, i tested the previous fix but it is not working any more. Also checked the rational behind it. This is not the improvement to previous fix, these lines changes nothing but removes obsolete code. Previous fix should be replaced by a new solution and that's another topic.

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -972,8 +970,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 	_onTextSelectionMsg: function (textMsg) {
 		L.CanvasTileLayer.prototype._onTextSelectionMsg.call(this, textMsg);
 		// If this is a cellSelection message, user shouldn't be editing a cell. Below check is for ensuring that.
-		if ((this.insertMode === false || this._map._isCursorVisible == false) &&
-		    this._cellCursorXY && this._cellCursorXY.x !== -1) {
+		if ((this.insertMode === false || this._map._isCursorVisible == false) && app.file.calc.cellCursor.visible) {
```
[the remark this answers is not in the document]
gokaysatir: Yes, not now unfortunately. Because there is something wrong with the cell cursor for now. It's already in master - cell cursor is visible while editing a cell - i'm not sure if that's expected / wanted. But you are right, let me note this for the future.

browser/src/layer/tile/CanvasTileLayer.js
gokaysatir: Maybe i can replace this like below? That'll be shorter. I am also uncomfortable with the long names.

* app.file.calc.cellCursor.rectangle.
* app.calc.cellRectangle

browser/src/layer/tile/CanvasTileLayer.js
gokaysatir: If i do this, i will probably need to convert L.Point to SimplePoint in multiple places (in place or change the variable, doesn't matter). I am not sure how frequently it will happen. That's why i am trying to push this to future adjustments :D Not that i don't agree.

I am currently working on a little more complicated part of cell cursor. Let me re-think while i work on the next PR.

I need/want to feel more comfortable with variable interoperability before i enforce things.

eszkadev (approved):

PR #8701, opened 2024-04-04
Draggable, TouchGesture: Fix panning with RTL

Minion3665:
Previously when we were in RTL mode panning would be flipped. This
commit unflips it

This is a nontrivial backport of #8565, nontrivial only because that contained some es6-specific code that cannot be backported

I also didn't backport removing the dead code, because I don't want to make any backport that's unnecessary to fix something

Change-Id: Iea4d16918b054d355e6d8695e0dc1d6ededd6793* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/dom/Draggable.js
+++ b/browser/src/dom/Draggable.js
@@ -120,6 +120,10 @@ L.Draggable = L.Evented.extend({
 		    newPoint = new L.Point(first.clientX, first.clientY),
 		    offset = newPoint.subtract(this._startPoint);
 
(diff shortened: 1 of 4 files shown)


gokaysatir (approved):

pedropintosilva:
I would like to have this in eventually but I wonder if it's reasonable to try to do it now. What do you think @eszkadev @timar ?

eszkadev (approved):

PR #8706, opened 2024-04-04
remove-w2ui: cleanup, create shared base class for toolbars

eszkadev:
- base class for toolbars
- use ES6 classes not leaflet's objects


gokaysatir (approved):
Code looks fine and apps seem to work, thanks @eszkadev :)

PR #8709, opened 2024-04-04
After fork we don't have a watchdog thread

caolanm:
So watchdog won't fire for a stalling kit.

After a fork the child has only one thread, but a copy of the watchdog object.

Stop the watchdog thread before fork, let the child discard its copy of the watchdog that is now in a discardable state.

And allow it to create a new one on the next SocketPoll ctor.

Change-Id: I7dc166dca3996401fbdc20cd7643f944662454c8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -49,6 +49,7 @@
 #include <common/SigUtil.hpp>
 #include <common/security.h>
 #include <common/ConfigUtil.hpp>
+#include <common/Watchdog.hpp>
 #include <kit/DeltaSimd.h>
 
 static bool NoCapsForKit = false;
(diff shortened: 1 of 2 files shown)


kit/ForKit.cpp
```diff
@@ -437,6 +448,10 @@ static int createLibreOfficeKit(const std::string& childRoot,
 
             SigUtil::setSigChildHandler(nullptr);
 
+            // Throw away inherited watchdog, which will let a new one for this
+            // child be created on demand
+            SocketPoll::PollWatchdog.reset();
```
mmeeks, line 453: Hmm - it might be nicer to just startThread() again in both child and parent - to save a potential race (which will never happen ;-) this should work just fine.

mmeeks (approved):

mmeeks:
Needed on 23.05 too I think =)

caolanm:
yes, waiting for ci before bothering https://github.com/CollaboraOnline/online/pull/8710

PR #8710, opened 2024-04-05
After fork we don't have a watchdog thread

caolanm:
So watchdog won't fire for a stalling kit.

After a fork the child has only one thread, but a copy of the watchdog object.

Stop the watchdog thread before fork, let the child discard its copy of the watchdog that is now in a discardable state.

And allow it to create a new one on the next SocketPoll ctor.

Change-Id: I7dc166dca3996401fbdc20cd7643f944662454c8 (cherry picked from commit bb7cd9f35704d1391874d3e302dc7a66570cd242)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Watchdog.hpp
+++ b/common/Watchdog.hpp
@@ -52,7 +52,8 @@ public:
 
     ~Watchdog()
     {
-        joinThread();
+        if (_thread)
+            joinThread();
     }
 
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Thanks ! =)

PR #8712, opened 2024-04-05
Mend checks for randomness.

mmeeks:
I was rather surprised by this. We were missing some critical #undefs in config.h - Bayram - something to watch out for in the future =)

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -115,24 +115,16 @@ namespace Util
             std::vector<char> v(length);
 
             int len = 0;
-            /* use "/dev/[u]random" by default in case of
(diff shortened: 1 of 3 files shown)


mmeeks (commented):
Ah - thanks for the cleanup ! =) I have a larger change for master I've just pushed that will need to overlap with this.

mmeeks:
This is probably the wrong way to do this in retrospect; lets avoid getrandom - and just open the random device just once.

mmeeks:
Turns out there may be a better idea that avoids using these calls - coming in master; lets hold for that.

PR #8715, opened 2024-04-05
Private/mmeeks/norandom dev

mmeeks:
Needs co-23.05 patches too for core - but even then, I'm getting test failures - hey ho.

https://gerrit.libreoffice.org/c/core/+/165824
https://gerrit.libreoffice.org/c/core/+/165820

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -118,7 +118,7 @@ namespace Util
             std::vector<char> v(length);
 
             int len = -1;
-#if !HAVE_GETENTROPY && !HAVE_SYS_RANDOM_H
+#if !HAVE_GETENTROPY && !HAVE_SYS_RANDOM_H && !MOBILEAPP
 #  error "getentropy or getrandom are required for succesful function";
 #endif
 
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -117,25 +117,33 @@ namespace Util
         {
             std::vector<char> v(length);
 
-            int len = 0;
-            /* use "/dev/[u]random" by default in case of
-            getentropy() or getrandom() fails,
-            or they don't available */
-            bool useDevRandom = true;
-#ifdef HAVE_GETENTROPY
-            len = getentropy(v.data(), length);
-
-            // getentropy() works, no need to use "/dev/[u]random"
-            if (len != -1)
-                useDevRandom = false;
-#elif defined HAVE_SYS_RANDOM_H
-            len = getrandom(v.data(), length, GRND_NONBLOCK);
-
-            // getrandom() works, no need to use "/dev/[u]random"
-            if (len != -1)
-                useDevRandom = false;
+            int len = -1;
+#if !HAVE_GETENTROPY && !HAVE_SYS_RANDOM_H
+#  error "getentropy or getrandom are required for succesful function";
 #endif
-            if (useDevRandom)
+
+#if HAVE_GETENTROPY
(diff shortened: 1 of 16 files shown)


caolanm:
@mmeeks I think the android failures are that it is detecting the host getentropy/getrandom (or something like that) and until now were relying on the conditionals not getting defined, and some other similar problem for other failures that I added commits to in https://github.com/CollaboraOnline/online/pull/8712

common/Util.cpp
```diff
@@ -117,25 +117,33 @@ namespace Util
...
-            int len = 0;
-            /* use "/dev/[u]random" by default in case of
-            getentropy() or getrandom() fails,
-            or they don't available */
-            bool useDevRandom = true;
-#ifdef HAVE_GETENTROPY
-            len = getentropy(v.data(), length);
+            int len = -1;
+#if !HAVE_GETENTROPY && !HAVE_SYS_RANDOM_H
+#  error "getentropy or getrandom are required for succesful function";
```
caolanm, line 122: I think this is problematic for the iOS/Android cases, and seeing as the /dev/urandom fallback is restored, I guess drop this, or amend to be Linux only?

mmeeks: Ah yes, I restored the fallback - we should not error out on iOS or Android I think - they're single-user and don't do the forkit stuff. We have a run-time error, but it'd be nicer to compile fail for Linux if it some ancient Linux I think.

caolanm:
"COOL release (from tarball) co-24.04" is "Building remotely on oraclelinux7-a"computer/oraclelinux7-a)
"Unit Tests and Cypress Tests against core co-24.04" is "Building remotely on "Oracle Linux 7" 

man pages for getrandom and getentropy say they were added in glibc 2.25 but RHEL7 was "based on glibc 2.17"

timar:
> "COOL release (from tarball) co-24.04" is "Building remotely on oraclelinux7-a" "Unit Tests and Cypress Tests against core co-24.04" is "Building remotely on "Oracle Linux 7"
> 
> man pages for getrandom and getentropy say they were added in glibc 2.25 but RHEL7 was "based on glibc 2.17"

This is a problem, as our baseline is RHEL7 :(

mmeeks:
> > "COOL release (from tarball) co-24.04" is "Building remotely on oraclelinux7-a" "Unit Tests and Cypress Tests against core co-24.04" is "Building remotely on "Oracle Linux 7"
> > man pages for getrandom and getentropy say they were added in glibc 2.25 but RHEL7 was "based on glibc 2.17"
> 
> This is a problem, as our baseline is RHEL7 :(

Hmm - so, what do the kernels look like? I wonder if we can wrap the syscall ourselves and be alright (?) =) quite possibly glibc moves rather slower than the underlying kernels here. https://djalil.chafai.net/blog/2014/10/13/linux-kernel-3-17/ suggests this was 2014 - which being a decade ago -must- be in enterprise kernels by now surely !?

mmeeks:
In extremis of course we can open the /dev/urandom file-descriptor before forking and chrooting - and have it as a global handle that we push into COOL, and forkit and NSS, and so on and read from but never close; quite possibly that is a simpler approach and/or something we'll have to do for NSS et. al.

timar:
@mmeeks kernel is OK.

mmeeks:
So this is obsoleted by: https://github.com/CollaboraOnline/online/pull/8719 which is far simpler.

PR #8719, opened 2024-04-06
Private/mmeeks/random dev

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -113,43 +113,44 @@ namespace Util
             return _rng();
         }
 
+        int getURandom()
+        {
+            static int urandom = open("/dev/urandom", O_RDONLY);
+            if (urandom < 0)
+            {
+                LOG_SYS("Failed to source hard random numbers");
+                fprintf(stderr, "No adequate source of randomness");
+                abort();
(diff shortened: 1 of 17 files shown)


common/Util.cpp
```diff
@@ -113,43 +113,44 @@ namespace Util
...
         std::vector<char> getBytes(const std::size_t length)
         {
             std::vector<char> v(length);
 
-            int len = 0;
-            /* use "/dev/[u]random" by default in case of
-            getentropy() or getrandom() fails,
-            or they don't available */
-            bool useDevRandom = true;
-#ifdef HAVE_GETENTROPY
```
caolanm, line 125: Then we can drop the HAVE_SYS_RANDOM_H and HAVE_GETENTROPY include guard above, the configure check for those and the config.h.in mentions

mmeeks: Yes - I believe so - of course we need the core commits before this one will pass too of course.

timar (approved):
With this patch unit tests pass in the Incus container, too. So it is definitely an improvement, and needs more testing which is easier after merge.

PR #8720, opened 2024-04-06
Enable new copy handler on mobile platforms

plubius:
At least on iOS, the new copy handler works as intended and sends an uno message to the LibreOffice code.

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -890,11 +890,6 @@ L.Clipboard = L.Class.extend({
 	// Pull UNO clipboard commands out from menus and normal user input.
 	// We try to massage and re-emit these, to get good security event / credentials.
 	filterExecCopyPaste: function(cmd) {
-		if (window.ThisIsAMobileApp) {
-			// We do native copy/paste in the iOS and Android cases
-			return false;
-		}
-
 		if (this._map['wopi'].DisableCopy && (cmd === '.uno:Copy' || cmd === '.uno:Cut')) {
 			// perform internal operations
 			app.socket.sendMessage('uno ' + cmd);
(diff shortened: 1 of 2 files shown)


plubius:
Notes: It seems the copy button in the default layout's toolbar now goes through this code.  The Compact layout's Edit > Copy menu item still works so it must be going through a different path.

mmeeks:
Nice! thanks for that - we're all travelling to Cambridge for the week (and sorry to miss you) - but hopefully Miklos will have some time to review this; he's been re-writing the copy/paste code for the web recently.

plubius:
> Nice! thanks for that - we're all travelling to Cambridge for the week (and sorry to miss you) - but hopefully Miklos will have some time to review this; he's been re-writing the copy/paste code for the web recently.

No hurry. There are two more known bugs when pasting that I found that I'll take a look at if I have some time this week:
- In the default toolbar's Paste button, issue #5839 (Paste button is disabled) has reappeared when the system clipboard is empty
- Paste button in all default toolbar and in the Compact layout's Edit > Paste menu item both display the "Please use the copy/paste buttons on your on-screen keyboard." dialog

So, looks like we lost some of the mobile-specific code during the copy/paste rewrite.

vmiklos:
> Paste button in the default toolbar and in the Compact layout's Edit > Paste menu item both display the "Please use the copy/paste buttons on your on-screen keyboard." dialog

Is it possible the ios app has to opt in to allow the webview to read the clipboard? Integrators do this: https://sdk.collaboraonline.com/docs/advanced_integration.html#allow-the-clipboard-permission-query

For the actual changes, these look OK to me, thanks.

vmiklos (approved):

plubius:
> Is it possible the ios app has to opt in to allow the webview to read the clipboard? Integrators do this: https://sdk.collaboraonline.com/docs/advanced_integration.html#allow-the-clipboard-permission-query

The iOS app expects copying and pasting to be done in the "server" in the vcl C++ native code since the server is really just another thread in the same process. IIRC, the JavaScript code would send a .uno:Paste message via app.socket.sendMessage(). I assume, but can't confirm,  that Android uses the same process.

PR #8723, opened 2024-04-08
Lat / Lng Removal

gokaysatir:
Change-Id: I43f9e744882c8d9158e07cc8f66a029c42f0a9fc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/SplitPanesContext.ts
+++ b/browser/src/layer/SplitPanesContext.ts
@@ -113,6 +113,7 @@ export class SplitPanesContext {
 			changed = true;
 		}
 
+		app.calc.splitCoordinate.pX = newX;
 		this._updateXSplitter();
 
 		if (!noFire)
@@ -139,6 +140,7 @@ export class SplitPanesContext {
 			changed = true;
 		}
 
+		app.calc.splitCoordinate.pY = newY;
 		this._updateYSplitter();
 
 		if (!noFire)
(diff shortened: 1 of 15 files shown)


gokaysatir:
@timar i don't think this can pass CI, fyi. It'd be surprising.

browser/src/layer/SplitPanesContext.ts
```diff
@@ -195,6 +195,60 @@ export class SplitPanesContext {
...
+
+			// There is vertical split, narrow down the initial view.
+			viewRectangles[0].pX1 = 0;
+			viewRectangles[0].pX2 = this._splitPos.x;
+		}
+
+		if (this._splitPos.y) {
+			const bottomLeftPane = viewRectangles[0].clone();
+			bottomLeftPane.pY1 = this._splitPos.y;
+
```
eszkadev: no `viewRectangles.push` ?

browser/src/layer/tile/CanvasSectionContainer.ts
```diff
@@ -1667,6 +1670,9 @@ class CanvasSectionContainer {
 		this.canvas.style.width = cssWidth.toFixed(4) + 'px';
 		this.canvas.style.height = cssHeight.toFixed(4) + 'px';
 
+		app.canvasSize.pX = newWidth;
```
eszkadev, line 1673: above we have some calculations we could already avoid by using SimplePoint created from newWidth / newHeight probably

browser/src/layer/tile/CellCursorSection.ts
```diff
@@ -0,0 +1,56 @@
...
+			drawingOrder: L.CSections.AutoFillMarker.drawingOrder,
+			zIndex: L.CSections.AutoFillMarker.zIndex,
+			interactable: true,
+			sectionProperties: {},
+		});
+
+		this.documentObject = true;
+
+		this.sectionProperties.viewId = viewId;
+        this.sectionProperties.cursorWeight = 2;
```
eszkadev, line 32: indent

browser/src/layer/SplitPanesContext.ts
eszkadev: fixed in the next commit

browser/src/layer/tile/CanvasSectionContainer.ts
gokaysatir: Those lines are for canvas resizing and carefully set for crisp images. Our simplePoint and simpleRectangle classes round the values. Canvas sizing uses floating numbers. I think they are good for now, unless we are really sure that it wouldn't break image quality. I also commented about this in geometry.ts file. Thanks for the comment.

gokaysatir:
cell_cursor test was failing with 1px difference. I updated the range.

PR #8726, opened 2024-04-09
fix(browser): Limit size of user list avatars

juliushaertl:
Change-Id: I9750f345a98e89711f679f8d4c5ff9e2d99e0648

* Target version: master 

### Summary

Fix the avatar size in the user list popover menu by setting a fixed size

| Before | After |
|---|---|
| Screenshot 2024-04-09 at 10 07 27 | Screenshot 2024-04-09 at 10 07 21 |

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1697,6 +1697,11 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
 	grid-template-columns: 0fr 1fr;
 }
 
+.user-list-item img {
+	width: 32px;
+	height: 32px;
+}
+
 #follow-editor {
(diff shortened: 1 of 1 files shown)


pedropintosilva (changes requested):
Could we instead use the css vars (so it adapts to any integrator) and maybe place it near to the place where we are already setting other avatars? -> toolbar.css after `#userListHeader .avatar-img`

```css
.user-list-item img {
	width: var(--btn-size);
	height: var(--btn-size);
}
```

juliushaertl:
Has meanwhile been addressed in the way @pedropintosilva suggested in https://github.com/CollaboraOnline/online/pull/8761

PR #8732, opened 2024-04-10
Fix header avatar have the same image of current UserList

Darshan-upadhyay1110:
Problem :
 - Open the document which have some comment from other User
 - we can see that they are same as the logged in user avatar

Regression
    - this is a regression from commit https://github.com/CollaboraOnline/online/pull/8640/commits/e855aa6114bda90fdd64fbcbb32a77adcf7938f4
    - we were passing url of user which is logged in and that should not happening for other user avatar image

Solution: - while setUserImage function pass viewId
    - this will help to identify the user Info from map

Change-Id: I5ee8b41c2e58a4ba15ecf551b264f944aa926da8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/LOUtil.js
+++ b/browser/src/core/LOUtil.js
@@ -118,12 +118,11 @@ L.LOUtil = {
 
 		map.on('themechanged', setupIcon, this);
 	},
-	setUserImage: function(img, map) {
+	setUserImage: function(img, map, viewId) {
 		var docType = map.getDocType();
 		// set avatar image if it exist in user extract info
 		var defaultImage = L.LOUtil.getImageURL('user.svg', docType);
-		var myViewId = map._docLayer._viewId;
-		var extraInfo = map._viewInfo[myViewId].userextrainfo;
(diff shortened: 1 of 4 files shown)


browser/src/core/LOUtil.js
```diff
@@ -118,11 +118,10 @@ L.LOUtil = {
 
 		map.on('themechanged', setupIcon, this);
 	},
-	setUserImage: function(img, map) {
+	setUserImage: function(img, map, myViewId) {
```
Minion3665, line 121: can we rename this viewId, since as it isn't always ours?

Minion3665 (changes requested):
Mostly good, nice small change ... if we name this one bit better then I am happy to see this merged :)

browser/src/core/LOUtil.js
Darshan-upadhyay1110: yes we can use just `viewId` which will make sense. Thanks i will
 change it :)

Minion3665 (approved):

PR #6436, opened 2023-05-25
feat: Add UITheme option for light/dark to ui_defaults

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>
Change-Id: Ief028dcec0daa2ae448f83df993e0e5ade594a7c

* Resolves: #6432
* Target version: master 

### Summary

This adds the option to provide a default theme through the form parameters ui_defaults. As the frontend code currently only uses a boolean flag I mapped just UITheme=dark to enable the dark mode, all other values will use light by default.

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id
- [x] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #8735, opened 2024-04-10
cool#8734 browser: allow enabling dark mode from a URL parameter

vmiklos:
One way for integrations to ask for dark mode is to set URL parameters.
This works for e.g. the language, but didn't work for dark mode.

There was already a parameter for this, but it was limited to Android.

Fix the problem by exposing the parameter for non-Android as well, and
also extend global.js to default to this, but keep the local storage as
an option to customize the setting if the user wants to.

Tested with 'make run', <input type=hidden> way may need more work on
top of this.

Change-Id: I78c4115a8e337ef2b1ea9e498db2572965762bef

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -362,9 +362,8 @@ m4_ifelse(IOSAPP,[true],
 m4_ifelse(ANDROIDAPP,[true],
      [window.userInterfaceMode = window.coolParams.get('userinterfacemode');])
 
-m4_ifelse(ANDROIDAPP,[true],
-     [var darkTheme = window.coolParams.get('darkTheme');
-      if (darkTheme) {window.uiDefaults = {'darkTheme': true};}])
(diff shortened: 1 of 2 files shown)


vmiklos:

As for the concept (default to false, then ask integration, final say by user pref if set), I heard @pedropintosilva and @eszkadev are OK with this.

Testing this will be a bit more tricky, as current we don't have any cypress tests for theme/dark mode I fear, but should be doable in a separate PR, given enough time.

caolanm (approved):

pedropintosilva:
<3

juliushaertl:
I might miss something but isn't this already covered by https://github.com/CollaboraOnline/online/pull/6436 ?

vmiklos:
@juliushaertl thanks, I have to confess I wasn't aware of that, and seeing that this PR's way is already used for Android made me more confident. I've asked @caolanm to take a look, it's likely a regression on our side.

PR #8736, opened 2024-04-10
Coverity Fixes

Ashod:
- wsd: redunce string concatenation churn
- wsd: test: move single-use local map
- wsd: remove unused eof from the TileQueue

### DIFF
--- a/test/FileServeWhiteBoxTests.cpp
+++ b/test/FileServeWhiteBoxTests.cpp
@@ -437,7 +437,7 @@ void FileServeTests::testPreProcessedFileSubstitution()
         { "BUYPRODUCT_URL", "https://buy.ourproduct.com/'" }
     };
 
-    preProcessedFileSubstitution(testname, variables);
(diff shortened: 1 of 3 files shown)


test/FileServeWhiteBoxTests.cpp
```diff
@@ -437,7 +437,8 @@ void FileServeTests::testPreProcessedFileSubstitution()
         { "BUYPRODUCT_URL", "https://buy.ourproduct.com/'" }
     };
 
-    preProcessedFileSubstitution(testname, variables);
+    preProcessedFileSubstitution(testname, std::move(variables));
+    preProcessedFileSubstitution(testname, std::move(variables));
```
caolanm, line 441: dup line here I think

Ashod: Evidently didn't save after undoing.

caolanm (approved):

PR #8737, opened 2024-04-11
welcome window: allow to translate tab numbers

meven:
Change-Id: I01d62428ec98b051a7b131901bab3c346e7697d0

The numbers can be translated as well.


meven:
> Out of curiosity, did anyone request this?

I don't remember, that's something I noticed.

Issue #8648, opened 2024-03-27
Copy&paste: less push of the current selection to the clients


vmiklos:
### Is your feature request related to a problem?

Currently the simple selection is sent to clients proactively because that's the model needed for Firefox.

The task is to switch to async clipboard when we know the client is Chrome: in that case we can do an async network request to get the current selection and write that to the clipboard, no need to proactively push data to the client (even when it won't be used).

### Describe the solution you'd like

If we switch to navigator.clipboard.write (when available), then no need to press ctrl-c / click copy twice when the content is complex enough, which would be great.

### Describe alternatives you've considered

Stick to the subset provided by Firefox, which means only really copy to the clipboard, when all content is available synchronously, which is poor UX.

vmiklos:
https://github.com/CollaboraOnline/online/pull/8649 is related to this.

vmiklos:
#8695 does this mostly. What's remaining is to get rid of the implicit clipboard fetch also when running tests, i.e.

```diff
diff --git a/browser/src/layer/tile/CanvasTileLayer.js b/browser/src/layer/tile/CanvasTileLayer.js
index 71c8f486e8..a0654807c3 100644
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -3267,7 +3267,7 @@ L.CanvasTileLayer = L.Layer.extend({
 			this._map.removeLayer(this._map._textInput._cursorHandler); // User selected a text, we remove the carret marker.
 			// Keep fetching the text selection during testing, for now: too many tests
 			// depend on this behavior currently.
-			if (navigator.clipboard.write && !L.Browser.cypressTest) {
+			if (navigator.clipboard.write) {
 				this._map._clip.setTextSelectionType('text');
 			} else {
 				if (this._selectionContentRequest) {
```

vmiklos:
The actual change is in, only tests need adapting now.

PR #8739, opened 2024-04-12
cool#8648 clipboard: fix insert hyperlink dialog

vmiklos:
Type a word into Writer, select it, Ctrl-K, the link text should default
to the selection, but it was empty.

This happens because Toolbar.js getTextForLink() doesn't have the
pre-fetched clipboard anymore, so we return an empty string there.

Fix the problem by triggering a text selection fetch (and leaving the
clipboard alone), and then handling the hyperlink dialog on the
CanvasTileLayer.js side.

Once this is done, the usual dummy clipboard registration & copy() calls
need adding and the test passes even with the patch from
<https://github.com/CollaboraOnline/online/issues/8648#issuecomment-2037278091>.

Change-Id: Ibd8313588e705c4b027643b452099bbf02cfe8ec


vmiklos:

This was the last testcase from the suite, it was an actual breakage.

PR #8742, opened 2024-04-12
clean up and simplify logging framework

hcvcastro:
AutoPtr release the Channel instance

Backtrace 25378 - wsd 23.05.10.1 6fe1a82ff4:
./coolwsd(_ZN7SigUtil13dumpBacktraceEv+0x83)[0x555fb7731bff]
./coolwsd(+0x51ab3c)[0x555fb7731b3c]
/lib/x86_64-linux-gnu/libc.so.6(+0x3c050)[0x7f9d3245b050]
/home/hcastro/build/online/install/poco/v-1.9.1/lib/libPocoFoundationd.so.61(_ZN4Poco6Logger3logERKNS_7MessageE+0x5c)[0x7f9d32bcb23a]
./coolwsd(_ZN12StreamSocket17writeOutgoingDataEv+0x88e)[0x555fb777873a]
./coolwsd(_ZN12StreamSocket4sendEPKcib+0x7a)[0x555fb74c5456]
./coolwsd(_ZN12StreamSocket4sendERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEb+0x4b)[0x555fb74c54a5]
./coolwsd(_ZN24FileServerRequestHandler19preprocessAdminFileERKN4Poco3Net11HTTPRequestERK14RequestDetailsRKSt10shared_ptrI12StreamSocketE+0x1700)[0x555fb769afc4]
./coolwsd(_ZN24FileServerRequestHandler13handleRequestERKN4Poco3Net11HTTPRequestERK14RequestDetailsRNS0_17MemoryInputStreamERKSt10shared_ptrI12StreamSocketE+0x1119)[0x555fb7688311]
./coolwsd(_ZN23ClientRequestDispatcher21handleIncomingMessageER17SocketDisposition+0xc58)[0x555fb76041f0]
./coolwsd(_ZN12StreamSocket10handlePollER17SocketDispositionNSt6chrono10time_pointINS2_3_V212steady_clockENS2_8durationIlSt5ratioILl1ELl1000000000EEEEEEi+0xa6b)[0x555fb7776d97]
./coolwsd(_ZN10SocketPoll4pollEl+0x18c5)[0x555fb776480b]
./coolwsd(_ZN10SocketPoll4pollENSt6chrono8durationIlSt5ratioILl1ELl1000000EEEE+0x2e)[0x555fb74c4be0]
./coolwsd(_ZN10SocketPoll13pollingThreadEv+0x24)[0x555fb757920e]
./coolwsd(_ZN10SocketPoll18pollingThreadEntryEv+0x2c4)[0x555fb776291e]
./coolwsd(_ZSt13__invoke_implIvM10SocketPollFvvEPS0_JEET_St21__invoke_memfun_derefOT0_OT1_DpOT2_+0x66)[0x555fb7782c6c]
./coolwsd(_ZSt8__invokeIM10SocketPollFvvEJPS0_EENSt15__invoke_resultIT_JDpT0_EE4typeEOS5_DpOS6_+0x37)[0x555fb7782b3b]
./coolwsd(_ZNSt6thread8_InvokerISt5tupleIJM10SocketPollFvvEPS2_EEE9_M_invokeIJLm0ELm1EEEEvSt12_Index_tupleIJXspT_EEE+0x43)[0x555fb7782a61]
./coolwsd(_ZNSt6thread8_InvokerISt5tupleIJM10SocketPollFvvEPS2_EEEclEv+0x18)[0x555fb77829e4]
./coolwsd(_ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJM10SocketPollFvvEPS3_EEEEE6_M_runEv+0x1c)[0x555fb7782944]
/lib/x86_64-linux-gnu/libstdc++.so.6(+0xd44a3)[0x7f9d326d44a3]
/lib/x86_64-linux-gnu/libc.so.6(+0x89134)[0x7f9d324a8134]
/lib/x86_64-linux-gnu/libc.so.6(+0x1097dc)[0x7f9d325287dc]

Change-Id: I54afb7fcb1aa05c6dddceb0a8b366985bbe83448

### DIFF
--- a/common/Log.hpp
+++ b/common/Log.hpp
@@ -31,6 +31,18 @@
 
 #include "Util.hpp"
 
+enum Level
+{
+    FTL = 1, // Fatal
+    CTL,     // Critical
+    ERR,     // Error
+    WRN,     // Warning
+    NTC,     // Notice
+    INF,     // Information
+    DBG,     // Debug
+    TRC      // Trace
+};
+
 namespace Log
 {
     /// Initialize the logging system.
@@ -260,13 +272,13 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
 
 #ifdef __ANDROID__
 
-#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                               \
+#define LOG_LOG(LOG, NAME, LVL, STR)                                     \
     ((void)__android_log_print(ANDROID_LOG_DEBUG, "coolwsd", "%s %s", LVL, STR.c_str()))
 
 #else
 
-#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                           \
-    LOG.log(Poco::Message(NAME, STR, Poco::Message::PRIO_##PRIO))
+#define LOG_LOG(LOG, NAME, LVL, STR)                                     \
+    LOG.log(Poco::Message(NAME, STR, static_cast<Poco::Message::Priority>(LVL)))
 
 #endif
 
@@ -282,13 +294,13 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         LOG.flush();                                                                               \
     } while (false)
 
-#define LOG_BODY_(LOG, PRIO, LVL, X, PREFIX, END)                        \
+#define LOG_BODY_(LOG, LVL, X, PREFIX, END)                        \
     char b_[1024];                                                                                 \
-    std::ostringstream oss_(Log::prefix<sizeof(b_) - 1>(b_, LVL), std::ostringstream::ate);        \
+    std::ostringstream oss_(Log::prefix<sizeof(b_) - 1>(b_, #LVL), std::ostringstream::ate);       \
     PREFIX(oss_);                                                                                  \
     oss_ << std::boolalpha << X;                                                                   \
     END(oss_);                                                                                     \
-    LOG_LOG(LOG, LOG.name(), PRIO, LVL, oss_.str())
+    LOG_LOG(LOG, LOG.name(), LVL, oss_.str())
 
 #define LOG_ANY(X)                                                                                 \
     char b_[1024];                                                                                 \
@@ -313,7 +325,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, trace))                                                          \
         {                                                                                          \
-            LOG_BODY_(log_, TRACE, "TRC", X, logPrefix, LOG_END);                                  \
+            LOG_BODY_(log_, TRC, X, logPrefix, LOG_END);                                           \
         }                                                                                          \
     } while (false)
 
@@ -323,7 +335,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, trace))                                                          \
         {                                                                                          \
-            LOG_BODY_(log_, TRACE, "TRC", X, logPrefix, LOG_END_NOFILE);                           \
+            LOG_BODY_(log_, TRC, X, logPrefix, LOG_END_NOFILE);                                    \
         }                                                                                          \
     } while (false)
 
@@ -333,7 +345,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, debug))                                                          \
         {                                                                                          \
-            LOG_BODY_(log_, DEBUG, "DBG", X, logPrefix, LOG_END);                                  \
+            LOG_BODY_(log_, DBG, X, logPrefix, LOG_END);                                           \
         }                                                                                          \
     } while (false)
 
@@ -343,7 +355,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, information))                                                    \
         {                                                                                          \
-            LOG_BODY_(log_, INFORMATION, "INF", X, logPrefix, LOG_END);                            \
+            LOG_BODY_(log_, INF, X, logPrefix, LOG_END);                                           \
         }                                                                                          \
     } while (false)
 
@@ -353,7 +365,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, information))                                                    \
         {                                                                                          \
-            LOG_BODY_(log_, INFORMATION, "INF", X, logPrefix, LOG_END_NOFILE);                     \
+            LOG_BODY_(log_, INF, X, logPrefix, LOG_END_NOFILE);                                    \
         }                                                                                          \
     } while (false)
 
@@ -363,7 +375,7 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n = 
         auto& log_ = Log::logger();                                                                \
         if (LOG_CONDITIONAL(log_, warning))                                                        \
         {                                                                                          \
-            LOG_BODY_(log_, WARNING, "WRN", X, logPrefix, LOG_END);                                \
+            LOG_BODY_(log_, WRN, X, logPrefix, LOG_END);                                           \
         }                                                                                          \
     } while (false)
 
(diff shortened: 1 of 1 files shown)


mmeeks:
Seems a very surprising result to me. Is there a good explanation for why this happens ? I'd really like to see a much more detailed github bug report referencing your full analysis here before this goes in.

Though of course I hate coolwsd crashes.

Thanks !

hcvcastro:
when LOG_ANY is invoked then Log::logger().getChannel() return a Channel pointer
AutoPtr<Channel> own the pointer and release it (refcount ==0)

mmeeks:
Please file a github ticket. Please explain in the ticket - in detail - your thinking and analysis here. In particular why AutoPtr's reference counting is not protecting us and why it causes harm, why this has worked for so long, what particular scenario the problem happens in ? what other threads are doing (if there are any) during the crash.
If this is a really non-obvious fix, then I'd want to see what type is returned here in the code instead of hiding that behind the 'auto' keyword, and why this new type is better than a smart-pointer.
It's great that you're chasing a really important coolwsd crash but I'd like to review the thinking here in detail.
Thanks!

hcvcastro:
Ok, I will file a ticket,:

Also check Poco Documentation

// Foundation/src/Logger.cpp
Channel* Logger::getChannel() const
{
	return _pChannel;
}

getChannel returns a pointer, not smart pointer

hcvcastro:
created ticked

https://github.com/CollaboraOnline/online/issues/8743

mmeeks (changes requested):
LOG_ANY is there for a reason - for auditing messages. Lets not remove it.

common/Log.hpp
```diff
@@ -260,13 +260,13 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
...
-#define LOG_LOG(LOG, PRIO, LVL, STR)                                                               \
+#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                               \
     ((void)__android_log_print(ANDROID_LOG_DEBUG, "coolwsd", "%s %s", LVL, STR.c_str()))
 
 #else
 
-#define LOG_LOG(LOG, PRIO, LVL, STR)                                                               \
-    LOG.log(Poco::Message(LOG.name(), STR, Poco::Message::PRIO_##PRIO))
+#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                           \
+    LOG.log(Poco::Message(NAME, STR, Poco::Message::PRIO_##PRIO))
```
mmeeks, line 269: How does this remove a Poco dependency  ? =) I was hoping we could come up with a new enumeration to replace Poco::Message::PRIO_ and use LVL as the enumeration value - switch that from a string in the macro and stringify that for where it is used to simplify. Not sure what the Name is good for (?)

hcvcastro: Oh, LOG_ANY does not have a name (empty string), I cannot call macro LOG_LOG without a name , but the LVL enumeration it is a good idea, reworking ...

common/Log.hpp
```diff
@@ -311,69 +311,76 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
     do                                                                                             \
     {                                                                                              \
         auto& log_ = Log::logger();                                                                \
+        auto& name_ = log_.name();                                                                 \
```
mmeeks, line 328: So - this is extra cost in all the hot paths - if we do this it really should be in the (very unusual) if () condition where it is actually used.

common/Log.hpp
```diff
@@ -311,69 +311,76 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
...
         if (LOG_CONDITIONAL(log_, trace))                                                          \
         {                                                                                          \
-            LOG_BODY_(log_, TRACE, "TRC", X, logPrefix, LOG_END_NOFILE);                           \
+            LOG_BODY_(log_, name_, TRACE, "TRC", X, logPrefix, LOG_END_NOFILE);                    \
         }                                                                                          \
     } while (false)
 
 #define LOG_DBG(X)                                                                                 \
```
mmeeks, line 346: I'm far from happy with the degree of duplication here - though it pre-dates you.

We should be able to macro generate the body of all these LOG_DBG, LOG_TRC macros with another level of macro wrapper - and make this far more maintainable.

Can you squash this patch into the other large scale change here - and do that simplificaiton ? =)

mmeeks (changes requested):
Looking lots better; can we do another cleanup here - the code is still baroque =)

Thanks !

Ashod:
Thanks @hcvcastro for the patch.

As I wrote in the bug report, I need to understand why and how you hit this crash when this code has been working in production without incident for years. Without understanding the issue, it's impossible to have confidence in the fix. (I'm also very much worried about introducing regressions here, performance and functional.)

common/Log.hpp
```diff
@@ -282,22 +294,35 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
...
-    LOG_LOG(LOG, PRIO, LVL, oss_.str())
+    LOG_LOG(LOG, LOG.name(), LVL, oss_.str())
 
 #define LOG_ANY(X)                                                                                 \
-    char b_[1024];                                                                                 \
-    std::ostringstream oss_(Log::prefix<sizeof(b_) - 1>(b_, "INF"), std::ostringstream::ate);      \
-    logPrefix(oss_);                                                                               \
-    oss_ << std::boolalpha << X;                                                                   \
-    LOG_END(oss_);                                                                                 \
-    Poco::AutoPtr<Poco::Channel> channel = Log::logger().getChannel();                             \
```
Ashod, line 299: Perhaps the better fix is to check that `channel` here is not null?

common/Log.hpp
```diff
@@ -294,6 +294,17 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
         LOG.flush();                                                                               \
     } while (false)
 
+#define LOG_BODY__(LVL, X, TYPE, PREFIX, SUFFIX)                                                   \
```
Ashod, line 297: Two macros that differ by the number of underscores in the name is downright unfriendly. 

Can we find a more sensible name please? Perhaps `LOG_BODY_IMPL` or something like that ?

common/Log.hpp
```diff
@@ -260,13 +272,13 @@ static constexpr std::size_t skipPathPrefix(const char (&s)[N], std::size_t n =
...
-#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                               \
+#define LOG_LOG(LOG, NAME, LVL, STR)                                     \
     ((void)__android_log_print(ANDROID_LOG_DEBUG, "coolwsd", "%s %s", LVL, STR.c_str()))
 
 #else
 
-#define LOG_LOG(LOG, NAME, PRIO, LVL, STR)                           \
-    LOG.log(Poco::Message(NAME, STR, Poco::Message::PRIO_##PRIO))
+#define LOG_LOG(LOG, NAME, LVL, STR)                                     \
+    LOG.log(Poco::Message(NAME, STR, static_cast<Poco::Message::Priority>(LVL)))
```
Ashod, line 281: We really haven't removed any dependencies here, we just made our dependency implicit and fragile.

Sorry, I don't see the benefit of redefining the Priority enum and then casting it. We are far better off using the enum from Poco directly and explicitly. This kind of a change is how we can shoot ourselves in the foot, for no gains whatsoever (even this code is more complex than the old).

common/Log.hpp
```diff
@@ -31,6 +31,18 @@
 
 #include "Util.hpp"
 
+enum Level
+{
+    FTL = 1, // Fatal
```
Ashod, line 36: If we must use our own enum, then we certainly want to assign the values from Poco's `Priority` enum. This is very fragile as it is.

Ashod (commented):
Thanks, @hcvcastro. I left some comments inline.

Very odd that for a PR that is labelled "remove unused LOG_ANY" it doesn't seem that we remove it. Did I miss something? If `LOG_ANY` is unused, how did we get the crash? If it's used, why is this labelled "remove unused" ?

common/Log.hpp
mmeeks: @Ashod I want to remove the Poco headers from our code - so we don't accidentally use new features there, and we have better control over what we use. I'd like to kill Poco from the implementation too. We can have a no-op mapping function inside the Log.cxx if you're concerned that the numbers may drift =) once we start logging with Poco performance is terrible anyway - several redundant system calls per log message I'm afraid.

common/Log.hpp
mmeeks: Again - not having the Poco header included is the end goal here; but we can have an (internal) mapping function that will evaporate away at compile time I think.

common/Log.hpp
hcvcastro: I did it in my initial patches, if you can check the history

common/Log.hpp
hcvcastro: Sure, I will look another name

common/Log.hpp
hcvcastro: The "enum" data type are for future internal implementation logging framework, it coincides from API POCO Message Priority, I do not think they will modify the order, maybe add new value, by the time they add something we will get rid of the dependency.

mmeeks:
Merged in logging-cleanup commit; thanks Henry.

PR #8752, opened 2024-04-12
use SigUtil::setUserSignals before creating SocketPoll

caolanm:
The SocketPoll ctor which may, depending on COOL_WATCHDOG env variable, want to override the SIG2 handler so set user signal handlers before that otherwise tthat choice is overwritten

Change-Id: I305570ab8becb41f0696e60908c1ca26fd9ba14a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Very sensible =) thanks !

PR #8753, opened 2024-04-13
Enable new paste handler on mobile platforms

plubius:
At least on iOS, the new paste handler needs to send an uno message to the LibreOffice code.

Also, reenable fix for GitHub issue #5839.

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2381,23 +2381,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 					}
 
 					if (disabled) {
-						if (data.command === '.uno:Paste') {
-							// Fix GitHub issue #5839 and never disable Paste toolbar button
-							// Behave the same as Contol.Menubar and never
-							// disable Paste toolbar button. Native clients
-							// that run LibreOffice locally may send a
-							// "statechanged: .uno:Paste=disabled" message when
-							// opening a document if the system clipboard is
-							// empty, So, we ignore such messages or else the
-							// current document's Paste toolbar button will
-							// never be enabled.
-							$(div).removeClass('disabled'); // TODO: remove after css rework
-							div.setAttribute('disabled', '');
-							window.app.console.log('do not disable paste based on server side data');
-						} else {
-							$(div).addClass('disabled'); // TODO: remove after css rework
-							div.setAttribute('disabled', '');
-						}
+						$(div).addClass('disabled'); // TODO: remove after css rework
+						div.setAttribute('disabled', '');
 					}
 					else {
 						$(div).removeClass('disabled'); // TODO: remove after css rework
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -896,6 +896,12 @@ L.Clipboard = L.Class.extend({
 			return true;
 		}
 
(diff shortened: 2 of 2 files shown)


plubius:
Note: issue #5839 is a corner case where the system clipboard (and the server's internal clipboard for online) is empty. Maybe this never occurs in online, but if a document is opened in the iOS app when the system clipboard is empty, the paste toolbar button and menu items will be disabled until you reopen the document.

plubius:
This patch is now ready for review. I completely removed the fix for issue #5839. I will submit a patch to LibreOffice's vcl/ios code shortly to fix issue #5839.

plubius:
> This patch is now ready for review. I completely removed the fix for issue #5839. I will submit a patch to LibreOffice's vcl/ios code shortly to fix issue #5839.

This fix for issue #5839 is in the following LibreOffice patch:

https://gerrit.libreoffice.org/c/core/+/166073

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/650/console fails with:

> /bin/sh: ./coolwsd: No such file or directory

Let's retry in case that was a problem with CI itself.

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2381,23 +2381,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 					}
 
 					if (disabled) {
-						if (data.command === '.uno:Paste') {
-							// Fix GitHub issue #5839 and never disable Paste toolbar button
```
vmiklos, line 2385: I had a concern that you do this unconditionally (even when this is not ios/android), but seems the fix was added for mobile and the scenario (start Writer, copy something on the clipboard, use paste button on the notetebookbar) does work even after this fix with a desktop browser, so probably we're OK.

vmiklos (approved):

vmiklos:
Non-required multi-user test looks known-broken, the rest is fine.

browser/src/control/Control.JSDialogBuilder.js
plubius: Back in 23.05, I assumed that the ".uno:Paste=enabled" message was never sent by LibreOffice on all platforms. I don't have a Linux machine so I had to make a guess and I guessed wrong. Fortunately, after digging into the LibreOffice code, it was clear that the iOS clipboard code in vcl/ios just had an incomplete implementation.

Might be worth checking if Android works too. I don't know if the Android native clipboard code detects changes in the system clipboard like Linux, macOS, and now iOS. But I think an easy way to test this is to do the following (either in Online or Android):
- Open a Writer document in editable mode (doesn't matter if the clipboard is empty or not)
- Switch to another app/application and copy something
- Back in the Writer document, press the Paste button in the Notebookbar and then select Paste Special in the popup menu
- If the pasted content matches what you copied from another app/application, then issue #5839 is not occurring

When issue #5839 occurred on iOS, either the Paste button would remain disabled (if the clipboard started empty) or the pasted content would be what was previously on the clipboard before copying from another app/application.

vmiklos: Thanks, makes sense. I think Android/browsers don't have anything similar to the iOS PasteboardChangedEventListener / the usual desktop-style "owner changed" events, so what we do instead is to put a marker inside the HTML: if we get back our own HTML, we can do a much better internal paste, otherwise we do a HTML paste.

PR #8756, opened 2024-04-13
fix: duplicate `You` on reconnect

Rash419:
- it might happen viewinfo message arrives before status message, viewinfo message update view list with old _viewId because new _viewId after reconnection is not avaialable yet
- this patch fixes this by reseting the _viewId on socket close and call updateviewlist event after client gets new _viewId from status message

Change-Id: I88a6ade574faa1368b1635db891fe87f89fa080e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -959,7 +959,7 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	onUpdateViews: function () {
-		if (!this.map._docLayer)
+		if (!this.map._docLayer || !this.map._docLayer._viewId)
 			return;
 
 		var myViewId = this.map._docLayer._viewId;
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1178,6 +1178,14 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._canonicalIdInitialized = false;
 	},
 
+	_resetViewId: function () {
+		this._viewId = undefined;
+	},
+
+	_getViewId: function () {
+		return this._viewId;
+	},
+
 	_updateTileTwips: function () {
 		// smaller zoom = zoom in
 		var factor = Math.pow(1.2, (this._map.options.zoom - this._tileZoom));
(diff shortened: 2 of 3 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -959,10 +959,10 @@ L.Control.UIManager = L.Control.extend({
 	},
 
 	onUpdateViews: function () {
-		if (!this.map._docLayer)
+		var myViewId = this.map._docLayer._viewId;
```
eszkadev, line 962: please check `this.map._docLayer` before using

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1178,6 +1178,14 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._canonicalIdInitialized = false;
 	},
 
+	_resetViewId: function () {
+		this._map._docLayer._viewId = undefined;
```
eszkadev, line 1182: isn't `this` the same as `this._map._docLayer` here?

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1178,6 +1178,14 @@ L.CanvasTileLayer = L.Layer.extend({
...
+	},
+
+	_getViewId: function () {
+		return this._map._docLayer._viewId;
```
eszkadev, line 1186: `this`

eszkadev (approved):
looks ok

eszkadev:
I was not able to test this, somehow I cannot reconnect back to document in my local instance

Rash419:
> I was not able to test this, somehow I cannot reconnect back to document in my local instance

interesting, what is happening ?

Rash419:
> > I was not able to test this, somehow I cannot reconnect back to document in my local instance
> 
> interesting, what is happening ?

I think you need to immediately restart the coolwsd otherwise interval to try to reconnect increases exponentially

eszkadev:
> > > I was not able to test this, somehow I cannot reconnect back to document in my local instance
> > 
> > 
> > interesting, what is happening ?
> 
> I think you need to immediately restart the coolwsd otherwise interval to try to reconnect increases exponentially

might be it, I just have reconnecting screen and nothing happens

PR #8767, opened 2024-04-15
logging preview parts

hcvcastro:
- **browser: fix queue preview part duplicate**
- **browser: logging warning if empty rects**
- **browser: logging to check request preview parts**
- **browser: logging fetched preview parts**

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -376,9 +376,6 @@ L.Control.PartsPreview = L.Control.extend({
 	},
 
 	_layoutPreview: function (i, img, bottomBound) {
-		var topBound = this._previewContTop;
-		var previewFrameTop = 0;
-		var previewFrameBottom = 0;
 		if (i > 0) {
 			if (!bottomBound) {
 				var previewContBB = this._partsPreviewCont.getBoundingClientRect();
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.PartsPreview.js
```diff
@@ -393,16 +393,15 @@ L.Control.PartsPreview = L.Control.extend({
 			previewFrameBottom = previewFrameTop + this._previewFrameHeight;
 		}
 
-		var imgSize;
-		if (i === 0 || (previewFrameTop >= topBound && previewFrameTop <= bottomBound)
```
mmeeks, line 397: Hmm - just a thought - with a 1000 slide file, do we then request all previews ? there is great benefit in requesting only previews around where we are looking at least in the 1st instance - to avoid thrashing a ton of huge images in and scaling them - there can be a large expense here. What is the performance impact ? (didn't read in great detaul but ... ;-)

browser/src/control/Control.PartsPreview.js
```diff
@@ -633,6 +633,7 @@ L.Control.PartsPreview = L.Control.extend({
 			if (this._previewTiles[e.id]) {
 				this._previewTiles[e.id].src = e.tile.src;
 				this._previewTiles[e.id].fetched = true;
+				console.log('PREVIEW: part fetched : ' + e.id);
```
mmeeks, line 636: There is an app specific wrapper: window.app.console.log app. that can be disabled IIRC - we should use that consistently everywhere surely ?

browser/src/control/Parts.js
```diff
@@ -181,7 +181,7 @@ L.Map.include({
 
 	_addPreviewToQueue: function(part, tileMsg) {
 		for (var tile in this._previewQueue)
-			if (tile[0] === part)
+			if (this._previewQueue[tile][0] === part)
```
mmeeks, line 190: Ah - interesting; a C++ vs. JS iterator confusion on JS arrays ? - I wonder if there are others of those around (?) =)

mmeeks (changes requested):
Looks reasonable generally; needs a bit of cleanup I think =) and for back-porting there is just a small fix needed there I think (?)
Thanks Henry !

browser/src/control/Control.PartsPreview.js
hcvcastro: No, it will not request all 1000 preview, it just queue a list. Please read the function
"_processPreviewQueue",  it will request only the visible parts, in your example it will 
request at least 10 previews.

browser/src/control/Control.PartsPreview.js
hcvcastro: Sure, I will change it

browser/src/control/Parts.js
hcvcastro: Yes, it happened to me, I did not see another, but we will know =)

mmeeks (approved):
Thanks Henry =)

mmeeks:
Seems we need to fix: 

/home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04/browser/src/control/Control.PartsPreview.js
  380:7  warning  'topBound' is assigned a value but never used            no-unused-vars
  382:7  warning  'previewFrameBottom' is assigned a value but never used  no-unused-vars

✖ 2 problems (0 errors, 2 warnings)

caolanm:
FWIW There is a real warning here that fails

browser/src/control/Control.PartsPreview.js
  380:7  warning  'previewFrameTop' is assigned a value but never used  no-unused-vars

PR #8771, opened 2024-04-15
Change eslint config to new format & fix errors

Minion3665:
Eslint config has changed its format to use Javascript. This is pretty useful to us as it allows us to define the configuration in terms of .beforeprettier (and therefore include stylistic rules in files that are not touched by prettier because they are too early).

This needs some significant migration in places, in particular
- Some versions need to be changed
- "Globals" from bundled files need to be explicitly specifed in the config
- We need to ignore unused variables, as unless we use import/export we cannot know if they are really used somewhere else

I also fixed several warnings & errors involving incorrect conditions, which were revealed as we brought stricter eslint configs than had been active to some places

Change-Id: I78c15efcc26b9d613ae178da4784a96a00a7d6ee

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
Is that something still up-to-date but we need to rebase? Or it was old approach?

Minion3665:
> Is that something still up-to-date but we need to rebase? Or it was old approach?

Some of it is probably useful (e.g. updating the eslint config format) but I think if we want to do it we need to get it in in smaller pieces, this PR seems painful. I'll close it f

PR #8772, opened 2024-04-15
try adding {force: true} to leaflet-layer clicks

caolanm:
error is:

1) Joining a document should not trigger an invalidation
       Join after document save and modify:
     CypressError: Timed out retrying after 10050ms: `cy.click()` failed because this element:

`<div class="leaflet-layer" style="position: absolute; width: 642px; height: 484px;"></div>`

is being covered by another element:

`<div class="jsdialog ui-button-box end" id="">...</div>`

Fix this problem, or use {force: true} to disable error checking.

Change-Id: I8b0dde030b2aadf554db75872facb6c5c96a321c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


hfiguiere:
This patch allow the test to pass.

caolanm:
All tests are green now with this change.

timar (approved):

PR #8775, opened 2024-04-16
wsd: support synchronous CheckFileInfo

Ashod:
In certain scenarios, such as with Proxy.php,
we don't have CheckFileInfo in advance and
need to request it before proceeding further.

This adds a wrapper to wait for the
CheckFileInfo request to complete.

Change-Id: I8afa955f841899eb667c02168a29a68c1dffb21b

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -20,6 +20,7 @@
 #include <ios>
 #include <fstream>
 #include <memory>
+#include <stdexcept>
 #include <string>
 #include <sstream>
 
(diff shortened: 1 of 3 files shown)


timar (approved):

wsd/DocumentBroker.cpp
```diff
@@ -878,6 +881,20 @@ bool DocumentBroker::download(
     {
         LOG_DBG("CheckFileInfo for docKey [" << _docKey << ']');
         std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
+        if (!wopiFileInfo)
+        {
+            auto poller = std::make_shared<TerminatingPoll>("CFISynReqPoll");
+            poller->startThread();
+            CheckFileInfo checkFileInfo(poller, session->getPublicUri(), [](CheckFileInfo&) {});
```
caolanm, line 888: This dereferences "session", but later on on line 903 there is a check of session against null., which suggests it could be null here too?

PR #8785, opened 2024-04-17
statusbar: improve formatting

eszkadev:
statusbar: use ui-badge for items
    
    Statusbar items which are not active previously were grey.
    Recover that styling and do cleanup to not specify CSS in the
    JS code.
    
------------------------------------------

    statusbar: better strings formatting
    
After:
status-badge

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -236,6 +236,18 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	height: 1px;
 }
 
+/* Badge */
+.ui-badge {
+	padding: 5px !important;
+	color: var(--color-main-text);
+	background-color: var(--color-background-darker);
+	border-width: 0px;
+	border-radius: var(--border-radius);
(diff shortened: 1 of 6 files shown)


pedropintosilva (changes requested):
Awesome Szymon, Thanks!

A note, would be nice to don't style them all as disabled. What I mean is that some of those are not status but navigational labels. Example: Sheet 1 of 1 is not a status label so doesn't have enable/disabled or on/off status so it should be just text without background (we can now or later simple change the text color to somehting like var(--color-text-lighter)). PermissionMode also should not be treated as an off/on (there is Edit and Read-only mode but it's not really perceived as boolean for the user and rather just different modes)

eszkadev:
Ok, fixed. I misunderstood. I thought it is only for visual grouping of words for single item :)
I removed `!important` where I can for easier modification later. Now it takes style from `disabled` rules.

Now it is like in 23.05:
status-badge2

pedropintosilva (approved):

PR #8790, opened 2024-04-17
on joining an existing session calc/writer can use new users spell pref

caolanm:
but impress/draw should continue to use the existing document preference. Writer and Calc have spell as a pre-view feature, while impress/draw still has it per document.

Change-Id: Ieb8c307b13f2ee010e09058a1ba63c7808637487

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
Noticed this on joining a calc session that "spellchecking" was always on regardless of the existing view state or the default settings for the joining user

PR #8791, opened 2024-04-17
comments: fix avatar when author is not in document

Minion3665:
In e855aa6114bda90fdd64fbcbb32a77adcf7938f4 a regression was introduced where a comment avatar was set to your avatar if the user was not in the document

This commit returns it to the previous behavior, which had a "default" avatar instead. It might instead be nice to embed avatars in the document to avoid the avatars being missing, however the previous behavior is far less confusing than the regressed behavior and is far easier to get to than embedding avatars in the document

Change-Id: I82834089062b1e7c2570d9fd78c7e1a3077f4c96

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


thebearon (approved):
I tested it, and works well for me, thanks Skyler!

PR #8794, opened 2024-04-17
if a sheet is removed or moved, re-request comment positions

caolanm:
a problem since:

commit 72b2ce4bb649aeb36182303783f064b0b5d5e35a
Author: Caolán McNamara <caolan.mcnamara@collabora.com>
Date:   Wed Oct 25 19:19:08 2023 +0100
    we don't need to round-trip through core to reposition notes

Change-Id: If00ae17a8e80ca3f8d2a530efa6cdba90207be56

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -440,7 +440,12 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this._documentInfo = textMsg;
 			var partNames = textMsg.match(/[^\r\n]+/g);
 			// only get the last matches
+			var oldPartNames = this._partNames;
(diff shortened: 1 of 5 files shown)


caolanm:
We used to fetch comments all the time on every change, that was changed to minimize that, but cut too deep and we don't "know" a sheet was deleted, so restore this refresh when the number or order of sheets change.

Ideally we should have a more direct "sheet was deleted" or "sheet was moved" message, but in the interim simply refresh comments if the list of sheets change

vmiklos (approved):
Yes, I think that's the way you can do it currently. Once you insert or delete a sheet, you get a new status: incoming websocket message with the sheet names and you need to compare it with your local state to figure out the insertion / deletion. So this looks OK to me.

PR #8796, opened 2024-04-18
Hide outline format button for mobile

Darshan-upadhyay1110:
- Hide `Set outline format` uno button
- we will add this option again after fixing the dialog issue on mobile view for Outline option in `Bullets and numbering`

Change-Id: I270b78b7b9f00fd7b0f12c348794e3d3da3e653e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

eszkadev:
is that only in mobile-wizard?

then we can do it directly in MobileWizardBuilder.js -> no need to check if it is a mobile

Darshan-upadhyay1110:
> is that only in mobile-wizard?
> 
> then we can do it directly in MobileWizardBuilder.js -> no need to check if it is a mobile

Sure i have added hide setOutline to mobileWizard

Thanks @eszkadev

PR #8797, opened 2024-04-18
wsd: allow for recovering from unloading document

Ashod:
When the last connection is closed, we no longer
mark the document as unloading, until we save
and unload and have nothing further to do.

This allows others to connect to the document
and "recover" it.

A previously failing and disabled test is now
re-enabled and passes.

Change-Id: I425a9286ec5df56064132e5e4227227761c65830


wsd/DocumentBroker.cpp
```diff
@@ -2918,9 +2918,9 @@ std::size_t DocumentBroker::removeSession(const std::shared_ptr<ClientSession>&
                 // Don't destroy just yet, wait until save and upload are done.
                 // Notice that the save and/or upload could have been triggered
                 // earlier, and not necessarily here when removing this last session.
-                _docState.setUnloadRequested();
```
mmeeks, line 2921: This is the only place this is set; but it is checked in a lot of places - we should remove the state if we're going down this path; but then we need to check that all the 'markToDestroy' code-paths are instead gated on something else: eg. "no open sessions" at those points instead (?) =) or I fear we won't destroy the document (?)

Anyhow - looks encouragingly simpler than expected which is good.

mmeeks (changes requested):
As per comment =) looks promising in principle.

mmeeks (approved):
Looks a lot more robust ! =) Thanks Ash.

PR #8799, opened 2024-04-18
default to using available compiler hardening flags

caolanm:
can be disabled with --disable-hardening-flags, defaults on except for wasm and sanitizer cases

Change-Id: Id3ccb7492b73dfa6c7bbd5dd0419927032fc9bdf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -376,18 +376,19 @@ namespace FileUtil
             return nullptr;
         }
 
-        auto data = std::make_unique<std::vector<char>>(st.st_size);
+        auto remainingSize = st.st_size;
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Great; thanks ! =)

caolanm:
what should happen is that the "whether $CC supports -D_FORTIFY_SOURCE=2" should fail for a configure with those CFLAGS set and default to the rest on and that off. I think what is wrong there, checking a similar setup locally, is that there isn't enough in the test to trigger the failures.

https://github.com/CollaboraOnline/online/pull/8820 is something for that

PR #8800, opened 2024-04-18
Add aria lables to more option button in sidebar

Antony007:
### Summary

Add aria labels to 'More option' button in sidebar

### Changes

- Add aria>label data to toolbutton struct when creating more option button for Panel
- Removed alt attribute from Tool Button since buttons don't support alt attribute
- Added condition to set alt for button image from aria>label, if present

### Commit Info

Change-Id: Ic608ee29891154e0f46523a1f5b1e5c249c54c6a

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2322,7 +2322,11 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				label.textContent = builder._cleanText(data.text);
 				button.setAttribute('alt', label.textContent);
 				if (buttonImage !== false) {
-					buttonImage.alt = label.textContent;
+					if(data.aria) {
+						buttonImage.alt = data.aria.label;
+					} else {
+						buttonImage.alt = label.textContent;
+					}
 				}
 				builder._stressAccessKey(label, button.accessKey);
 				controls['label'] = label;
--- a/browser/src/control/jsdialog/Widget.SidebarContainers.ts
+++ b/browser/src/control/jsdialog/Widget.SidebarContainers.ts
@@ -67,6 +67,14 @@ JSDialog.panel = function (
 			{
 				type: 'toolitem',
 				command: expanderData.command,
+				aria: {
+					label: expanderData.children[0].text
+						? _('More options for {name}').replace(
+								'{name}',
+								expanderData.children[0].text,
+							)
+						: '',
+				},
 				icon: builder._createIconURL('morebutton'),
 			},
 			builder,


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -920,7 +920,17 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
 			var iconParent = expander.children('.ui-expander').get(0);
 			var icon = L.DomUtil.create('div', 'ui-expander-icon-right ' + builder.options.cssClass, iconParent);
-			builder._controlHandlers['toolitem'](icon, {type: 'toolitem', command: data.command, icon: builder._createIconURL('morebutton')}, builder);
+			builder._controlHandlers['toolitem'](
+				icon,
+				{
+					type: 'toolitem',
+					command: data.command,
+					aria : {
+						label: (data.children[0].text) ? 'More options for ' + data.children[0].text : '',
```
mmeeks, line 929: Hmm - surely for such things we should translate the text, and if we are translating it we should not concatenate strings like that, but do some %1 type substitution into a translated string using a stock helper (?).

eszkadev: you can translate string with: `_('some string')`

browser/src/control/jsdialog/Definitions.Types.ts
```diff
@@ -15,6 +15,7 @@
 
 // common for all widgets
 interface WidgetJSON {
+	text: string;
```
eszkadev, line 18: can we avoid this? it is not in all widgets

browser/src/control/jsdialog/Widget.SidebarContainers.ts
```diff
@@ -67,6 +67,14 @@ JSDialog.panel = function (
 			{
 				type: 'toolitem',
 				command: expanderData.command,
+				aria: {
+					label: data.children[0].text
+						? _('More options for %name').replace(
+								'%name',
+								data.children[0].text,
```
eszkadev, line 74: please use `expanderData` instead of `data`, there we already casted children to Array of any. `text` property should be valid then

browser/src/control/jsdialog/Widget.SidebarContainers.ts
```diff
@@ -67,6 +67,14 @@ JSDialog.panel = function (
 			{
 				type: 'toolitem',
 				command: expanderData.command,
+				aria: {
+					label: data.children[0].text
+						? _('More options for %name').replace(
```
timar, line 72: Please use {name} instead of %name, cf. bbe25379d5e23d1e0950d2cd8f5e209f3705334f

eszkadev (approved):
looks good to me

PR #8801, opened 2024-04-18
Fix broken frozen columns mid-pinch-zoom

Minion3665:
Previously the following issues were present when columns were frozen
- The frozen column was allowed to move, this was exacurbated when I made it easier to move in my 'calc: pinch-zoom:' commits, (3ba8257c6432a5a8f42901c9955c11f83a9a720a...b378a0fd778760dfb0f457c796769d14aaad11f6). This is now fixed
- A regression was also introduced in that range where the selected cell marker would become offset if you placed it in a frozen section then pinched to zoom. This is also fixed

Change-Id: I07f70a32e5cb68aeb759fd6cad626b2d4c0eed72

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -640,6 +640,8 @@ L.TileSectionManager = L.Class.extend({
 	 * @param paneBounds {{min: {x: number, y: number}, max: {x: number, y: number}}} The edges of the current pane
 	 * Traditionally this is the map border at the start of the pinch
 	 *
+	 * @param freezePane {{freezeX: boolean, freezeY: boolean}} Whether the pane is frozen in the x or y directions
+	 *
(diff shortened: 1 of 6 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -688,6 +690,14 @@ L.TileSectionManager = L.Class.extend({
 			Math.max(documentTopLeft.y, pinchStartCenter.y + (centerOffset.y - paneSize.y * panePortion.y) / scale)
 		);
 
+		if (freezePane.freezeX) {
```
eszkadev, line 693: can't we get that information from split pos?

eszkadev: I see now, it only is needed for some panes

eszkadev (approved):
It works much better. But still it has issues:
like: during panning we still see old freeze "lines", while preview renders cells with "new, adjusted ones"

PR #8802, opened 2024-04-18
Prevent opening multiple tooltips on drawing area widgets

GulsahKose:
We see multiple tooltip problem on drawing area widgets. Example case:
    Add basic shape into a writer document
    Right click on shape and open Position and Size dialog
    Select the Rotation tab
    Hold the rotation angle widget and drag around itself
When you inspect the "Rotation Angle" tooltip, count may be 600 We expect only 1.

So we prevent to show another tooltip if we have already one.

Change-Id: I2f548f8c7afce8095e83a6faa3722e164f990210

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1710,7 +1710,11 @@ L.Control.UIManager = L.Control.extend({
 	enableTooltip: function(element) {
 		var elem = $(element);
 		if (window.mode.isDesktop()) {
-			elem.tooltip();
+			if (this._tooltip) {
+				$(".ui-tooltip").remove();
+				this._tooltip = undefined;
+			}
+			this._tooltip = elem.tooltip();
 			elem.on("mousedown", function() {
 				$('.ui-tooltip').fadeOut(function() {
 					$(this).remove();


browser/src/control/Control.UIManager.js
```diff
@@ -1710,12 +1710,17 @@ L.Control.UIManager = L.Control.extend({
...
-					$(this).remove();
+			if ($('.ui-tooltip').length <= 0) {
+				elem.tooltip();
+				elem.on("mousedown", function() {
+					$('.ui-tooltip').fadeOut(function() {
+						$(this).remove();
+					});
 				});
-			});
+			} else {
```
eszkadev, line 1720: I believe it's better to show the last tooltip (it might have newer information).
Just:
1. hide old if exists
2. show new one
Is it possible to do it that way?

pedropintosilva: @eszkadev that's a good idea, but in practice and if it's makes less code, I would just remove/hide all because what are the odds of a user wanting to constantly see the tooltip? Just thinking out loud here but, I would rather not see multiple tooltips that actually make it harder to operate with the thing I'm trying to work with (be that the rotation widget, or maybe I'm just trying to look closely to a preview) and if the user really wants I assume the user will just move away and back in to see the tooltip.

eszkadev: yes, exactly what I wanted to say: we should have only one tooltip: last one ->  when new tooltip appears: remove all old, insert one new

GulsahKose:
Problematic cypress test passes locally:
CYPRESS_BROWSER="/usr/bin/google-chrome" make -C cypress_test UPDATE_SCREENSHOT=true check-desktop spec=calc/autofilter_spec.js

I've triggered twice on CI but still failing.

browser/src/control/Control.UIManager.js
```diff
@@ -1710,6 +1710,9 @@ L.Control.UIManager = L.Control.extend({
 	enableTooltip: function(element) {
 		var elem = $(element);
 		if (window.mode.isDesktop()) {
+			if($('.ui-tooltip').length > 0)
+				$(".ui-tooltip").remove();
+
 			elem.tooltip();
```
eszkadev, line 1716: will it work to save result in variable and later just use it to close using an JS API?

```
if (this._tooltip) {
    this._tooltip.tooltip('close');
    this._tooltip = undefined;
}
this._tooltip = elem.tooltip();
```

browser/src/control/Control.UIManager.js
```diff
@@ -1710,7 +1710,13 @@ L.Control.UIManager = L.Control.extend({
...
 		var elem = $(element);
 		if (window.mode.isDesktop()) {
-			elem.tooltip();
+			if (this._tooltip) {
+				this._tooltip.tooltip();
+				this._tooltip.tooltip('close');
+				this._tooltip = undefined;
+			}
+
+			this._tooltip = elem.tooltip();
```
eszkadev, line 1719: can we do the same in below "else" block? (looks similar case).
I'm sorry for that many comments :)

GulsahKose: Szymon I've modified the else block but ı've detected some difference between mine and your version. Let me send screen record

GulsahKose: When we remove all the tooltips with $(".ui-tooltip").remove(); 
Tooltip is less blinking 
[gulsah.webm](https://github.com/CollaboraOnline/online/assets/2954981/2b0a50dd-319b-478c-967d-0a7e520ffbab)


But if we use this._tooltip.tooltip('close'); We see more tooltip blinking:
[szymon.webm](https://github.com/CollaboraOnline/online/assets/2954981/901a93be-d9f8-4bae-847a-1a9fe9157cf4)


Should i use first version in the if block. What do you think?

browser/src/control/Control.UIManager.js
```diff
@@ -1710,12 +1710,19 @@ L.Control.UIManager = L.Control.extend({
...
-			elem.tooltip();
-			elem.on("mousedown", function() {
-				$('.ui-tooltip').fadeOut(function() {
-					$(this).remove();
+			if (this._tooltip) {
+				this._tooltip.tooltip();
+				this._tooltip.tooltip('close');
+				this._tooltip = undefined;
+			}
+			else {
```
eszkadev, line 1718: it was ok before. I meant other else block: 	line 1720

eszkadev: mobile version

eszkadev: also do we need to do `this._tooltip.tooltip()` before close?

GulsahKose: Szymon we have to initialize with this._tooltip.tooltip(). Without code doesn't work and throws an error message to debugger console.

It would be  nice handle mobile version in separate patch because I don't know much about mobile side. It can take some time to test on mobile.

eszkadev: ok, then please do the less blinking version as you had. Sorry for the noise.

browser/src/control/Control.UIManager.js
eszkadev: yes, please go with the best version you had.

eszkadev (approved):
looks ok

PR #8804, opened 2024-04-18
Lat / Lng Removal

gokaysatir:
Remove _cellCursor and use app.calc.cellCursorRectangle -> SimpleRectangle.

Turn _prevCellCursor,_cellCursorOnPgUp, _cellCursorOnPgDn into SimpleRectangle.

Remove horizontalDirection and verticalDirection calculations. They are not used. Also remove them from _onUpdateCellCursor function.

Use app.calc.cellCursorVisible instead of empty rectangle checks. Set variables to null to re-init them instead of assigning them empty rectangles.

Push lat/lng usage through markers (leaflet things).

Change-Id: I4b258a72f1ac3c94ccf25a07d0e7dab232165b90

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -3877,7 +3877,7 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	// Scrolls the view to selected position
 	scrollToPos: function(pos) {
-		if (pos.pX) // Turn into lat/lng if required.
+		if (pos.pX) // Turn into lat/lng if required (pos may also be a simplePoint.).
 			pos = this._twipsToLatLng({ x: pos.x, y: pos.y });
 
 		var center = this._map.project(pos);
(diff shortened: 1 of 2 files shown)


mmeeks:
Looks like a nice cleanup to me, but would appreciate Szymon's take :-)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -4677,9 +4666,9 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	// TODO: unused variables: horizontalDirection, verticalDirection
```
eszkadev, line 4667: please remove TODO which is now done

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -3891,6 +3877,9 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	// Scrolls the view to selected position
 	scrollToPos: function(pos) {
+		if (pos.pX) // Turn into lat/lng if required.
```
eszkadev, line 3880: please add comment above function that it can take old lat/lng and also new coordinate object as input so it will be obvious

PR #8805, opened 2024-04-18
browser: preview: remove unused code

hcvcastro:
The code is old and it is not used anymore by
new changes did it.

Change-Id: Id163cba7f586e3a083bff4cbb169a5d16e4b0cce

### DIFF
--- a/browser/src/control/Control.PartsPreview.js
+++ b/browser/src/control/Control.PartsPreview.js
@@ -88,7 +88,7 @@ L.Control.PartsPreview = L.Control.extend({
 					}, this), 500);
 				}
 
-				var bottomBound = this._getBottomBound();
+				this._getBottomBound();
 
 				// Add a special frame just as a drop-site for reordering.
 				var frameClass = 'preview-frame ' + this.options.frameClass;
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.PartsPreview.js
```diff
@@ -88,7 +88,7 @@ L.Control.PartsPreview = L.Control.extend({
 					}, this), 500);
 				}
 
-				var bottomBound = this._getBottomBound();
+				this._getBottomBound();
```
mmeeks, line 91: Unusual to have a getFoo() function that doesn't return anything; is that expected and/or can we rename it or ? ... needs someone like Gokay to review this not me of course =)

gokaysatir: I think this one is forgotten while checking the unused variables :) 

@hcvcastro let me approve this PR so you don't lose time.

This will be complete if you can have time for a one line follow up for the forgotten line.

Thanks :)

gokaysatir (approved):

browser/src/control/Control.PartsPreview.js
hcvcastro: Yes, I am still cleaning more code, I will simplify and rename functions

PR #8807, opened 2024-04-18
writer: resize comments based on space

aszucs3:
Make comments window bigger, if needed, and possible. Selected commits grow up to MAX size, even if there is no space for it.

There is 2 constant limitation:
MAX - --annotation-max-size this is the maximum size commentText window can grow. If the comment have more text, than it will have scrollbar.

MIN - --annotation-min-size if there is not enought space, but comment have more text, this is the minimum size of commentText window.

note: commentText window can be smaller then MIN, if it has less text.

During doLayout:
first, in resetCommentsSize comments size resets to default. (max-height = MIN size.)

after comments positioned with loopUp / loopDown,
resizeComments check all comments (maybe it could be optimized)

If a comment have more text as its actual size, then it check the space between this, and next comment.. and resize this comment.

If there is not enought space after the comment, then it check for more space between this, and previous comment, and resize + move up the comment as much as needed.

Change-Id: Ie7b0971febd8c965b01ebe23125eac4eded1a742

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -36,6 +36,8 @@
 
 	/* Annotations */
 	--annotation-input-size: 100px;
+	--annotation-max-size: 1000;
+	--annotation-min-size: 100;
 }
 .focus-hidden:focus {
 	outline: none;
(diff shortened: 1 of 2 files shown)


eszkadev:
@aszucs3 in the future: please rebase and force-push if something doesn't pass - it's better method to pass CI
Reopening doesn't rebase - and it is possible some tests are broken at the point your change is

lpranam (approved):

PR #8808, opened 2024-04-18
bgsave: have a single source for InputProcessing enable & disable.

mmeeks:
Somehow this state can get confused in a bgsave process:

   Kit Document:
         ...
         inputProcessingEnabled: false
         ...
   SocketPoll:
     Poll [kit] with 1 socket - wakeup rfd: 39 wfd: 45
             fd        events        rbuffered        wbuffered        rtotal        wtotal
             52        0x1        process             0             0         r:    825

'process' should read 'ignore' for disabled input.

Change-Id: I787eebe6fda3ae1b527d7605b8813fa764e81890

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -711,7 +711,6 @@ Document::Document(const std::shared_ptr<lok::Office>& loKit,
       _editorChangeWarning(false),
       _lastMemTrimTime(std::chrono::steady_clock::now()),
       _mobileAppDocId(mobileAppDocId),
-      _inputProcessingEnabled(true),
       _duringLoad(0)
 {
     LOG_INF("Document ctor for [" << _docKey <<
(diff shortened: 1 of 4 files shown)


caolanm (approved):

net/WebSocketHandler.hpp
```diff
@@ -691,6 +691,15 @@ class WebSocketHandler : public ProtocolHandlerInterface
         return sendFrame(socket, data, len, WSFrameMask::Fin | static_cast<unsigned char>(code), flush);
     }
 
+    virtual bool processInputEnabled() const override
+    {
+        std::shared_ptr<StreamSocket> socket = _socket.lock();
+        if (socket)
+            return socket->processInputEnabled();
+
+        return true;
```
Ashod, line 700: Enabled even when we have no socket?

kit/Kit.cpp
```diff
@@ -2090,13 +2089,12 @@ void Document::checkIdle()
...
 
-void Document::enableProcessInput(bool enable)
+bool Document::processInputEnabled() const
 {
-    LOG_TRC("Document - input processing now: " <<
-            (enable ? "enabled" : "disabled") <<
-            " was " <<
-            (_inputProcessingEnabled ? "enabled" : "disabled"));
-    _inputProcessingEnabled = enable;
+    bool enabled = !_websocketHandler || _websocketHandler->processInputEnabled();
```
Ashod, line 2094: I suppose it's intentional that when we have no socket we assume input processing is enabled.

net/WebSocketHandler.hpp
mmeeks: Sure - not processing input is a sufficiently bad state to be in that we don't really want to go there I think. =)

kit/Kit.cpp
mmeeks: And yes; perhaps un-necessary, but I was bitten hard by this non processing of input ;-)

PR #8810, opened 2024-04-18
bgsave: don't try to save twice in the same background process.

mmeeks:
Get the polarity of the check on whether we succeeded to spawn a background save process right, so we don't spawn two saves concurrently, and block input processing in the wrong place.

Also a few obvious sillies fixed, and some more assertions for good measure; was not good:

[ kitbgsv_001_001 ] TRC  ToMaster-002: Finished synchronous background saving ...| kit/ChildSession.cpp:882
[ kitbgsv_001_001 ] TRC  ToMaster-002: saveDocumentBackground returns succesful start.| kit/ChildSession.cpp:887
[ kitbgsv_001_001 ] TRC  Document - input processing now: disabled was enabled| kit/Kit.cpp:2095
[ kitbgsv_001_001 ] TRC  ToMaster-002: uno command .uno:Save {... notify: true| kit/ChildSession.cpp:1944

Change-Id: Ie06a74a538bdd5038ca9c94b422f27cff9e4a82e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/MessageQueue.hpp
+++ b/common/MessageQueue.hpp
@@ -77,7 +77,7 @@ public:
         return _queue.empty();
     }
 
-    bool size()
+    size_t size()
     {
         return _queue.size();
     }
--- a/common/MessageQueue.hpp
+++ b/common/MessageQueue.hpp
@@ -77,7 +77,7 @@ public:
         return _queue.empty();
     }
 
-    size_t size()
+    size_t size() const
     {
         return _queue.size();
     }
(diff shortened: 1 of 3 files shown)


common/MessageQueue.hpp
```diff
@@ -77,7 +77,7 @@ class MessageQueue
         return _queue.empty();
     }
 
-    bool size()
+    size_t size()
```
caolanm, line 80: failed to see bool vs size_t the last time, might as well suggest it could be const this time

caolanm (approved):

common/MessageQueue.hpp
mmeeks: =)

Issue #8734, opened 2024-04-10
Dark mode: not possible to override from the URL


vmiklos:
### Describe the solution you'd like

1. Append `&darkTheme=true` at the end of the URL.
2. Expected result: both the JS / CSS frame and tiles appear in dark mode.
3. Actual result: not possible to overwrite the dark theme easily before the UI loads.

### Describe alternatives you've considered

Load the document and then press the toggle on the notebookbar.

vmiklos:
See #8735

PR #8813, opened 2024-04-18
backport 23.05 equivalent of honoring window.uiDefaults.darkTheme

caolanm:
if it exists as the default dark/light theme to use if there is no explicit user setting for it

Related to: https://github.com/CollaboraOnline/online/issues/8734

similar issue as in 24.04 arises from similar 23.05 change of:

commit 100def4c80d6e467c32e2a3300e695905c358838
Date:   Tue Feb 20 16:23:37 2024 -0400

Change-Id: I49596768ece16ceee1a74740877550122722d51a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -280,6 +280,9 @@ window.app = {
 					if (docTheme) {
 						themes.push(global.L.Browser.docs[theme] + ':' +
 							    (docTheme === 'true' ? 'Dark' : 'Light'));
+					} else if (window.uiDefaults) {
+						themes.push(global.L.Browser.docs[theme] + ':' +
+							    (window.uiDefaults.darkTheme ? 'Dark' : 'Light'));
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Thanks a lot, I tested this with `&ui_defaults=UITheme%3Ddark` and fixes the issue.

vmiklos:
CC @meven quite possibly it's better to document this syntax in the SDK, and not `&darkTheme=true`. Currently both works, but the previous is more generic.

meven:
> CC @meven quite possibly it's better to document this syntax in the SDK, and not `&darkTheme=true`. Currently both works, but the previous is more generic.

Agreed `window.uiDefaults` is a much better way to pass-in default parameters.

PR #8820, opened 2024-04-19
compile check of -D_FORTIFY_SOURCE=2 not failing with CFLAGS=-O0

caolanm:
need to include some typical headers to get it to issue a failure diagnostic

Change-Id: I37d6e0f342698507952a8006b88d56e55060270e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
does this work for your use case?

mmeeks (approved):
Looks great to me =) thanks!

vmiklos:
> does this work for your use case?

yes, thanks. I used to get:

> checking whether ccache g++-12 supports C++20... no

and now things are back to working.

PR #8821, opened 2024-04-19
cool#8806 clipboard: don't assume that navigator.clipboard is always defined

vmiklos:
Trying to paste from the notebookbar in a setup which is not localhost
but uses http resulted in a JS exception:

> Clipboard.js:863 Uncaught TypeError: Cannot read properties of undefined (reading 'read')

We assumed that navigator.clipboard is always defined, but then its
'read' and 'write' properties may be undefined.

Fix the problem by abstracting this, and check for navigator.clipboard
as well in global.js.

Can be tested by disabling https in coolwsd.xml and changing localhost
to the real local IP in the make run URL.

Change-Id: I1dab14e311066fa640dbba65dfcc35526a6886d7

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -101,6 +101,11 @@ window.app = {
 	    ie = 'ActiveXObject' in global,
 
 	    cypressTest = ua.indexOf('cypress') !== -1,
+	    // Firefox has undefined navigator.clipboard.read and navigator.clipboard.write,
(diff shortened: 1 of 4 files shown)


vmiklos:

This addresses the initially reported problem (http and non-localhost -> broken clipboard); I will keep the issue open because @pedropintosilva now found another related one, but I suggest to deal with that in a follow-up PR.

caolanm (approved):

PR #8823, opened 2024-04-19
Private/pedro/insert annotation more consistency writer impress

pedropintosilva:
Overall idea: Facilitate user's spatial memory by making sure Insert comment is reachable at similar spots between writer and impress and between home and insert tab with no scrolling

Commits

-----

Writer, Impress Home tab: Promote insert annotation to bigtoolitem

- Promote to big button
  - Use the same from the previous position to make the whole tab more
  - compact by changing the insert image to a simple toolitem
- Change location in the home tab so it's easier to reach and it's
more consistent between writer and impress
- Use the same accesskey (accessibility, aka shortcut) between both apps

-----

Writer, Impress Insert tab: Annotation & Remotelink to similar place

- Make the Insert tab : insert annotation in similar place between
apps
- Use the same accesskey (accessibility, shortcut) in between apps and
the same as the one used inside of Home tab
  - Writer: Insert tab
    - Insert annotation: Add missing DE accesskey
    - Insert remotelink: Add missing accesskey
- Make both insert tabs more compact by making sure hyperlink and
remotelink stacked on both apps

### DIFF
--- a/browser/src/control/Control.NotebookbarImpress.js
+++ b/browser/src/control/Control.NotebookbarImpress.js
@@ -696,6 +696,13 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 				'vertical': 'true'
 			},
 			{
+				'id': 'home-insert-annotation',
+				'type': 'bigtoolitem',
+				'text': _UNO('.uno:InsertAnnotation'),
+				'command': '.uno:InsertAnnotation',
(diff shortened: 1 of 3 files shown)


pedropintosilva:
there is no change in functionality here. Just 1. Moving the insert comment so it's prioritized 2. Add missing  accesskeys and 3. Move other buttons into icons only or/and stacked to save space

eszkadev (approved):

PR #8827, opened 2024-04-19
Private/caolanm/revertearlytheme23

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos:
@caolanm what's the thinking here, is this about "if the first 2 commits are reverted, then the 3rd is not needed, and that's a better fix", or something else?

caolanm:
qa requests a revert of the early theme setting in co-23. Without it then the 3rd patch isn't required

vmiklos (approved):
I see, makes sense then.

PR #8832, opened 2024-04-19
Cypress: Clean up document loading and test setup

neilguertincollabora:
This is a big change, sorry. I tried to make smaller chunks but there was no easy way. It should be manageable if you review it one commit at a time.

The commits do work independently and I did make sure they pass CI after each commit. However I didn't want to submit them separately because the first two leave warnings and messy code that are cleaned up by the last one.

## The problem
- Many helper functions have a huge list of parameters
  - The list is copied at least five times
  - The list has changed over time without updating all of the tests that call helpers using it
    - So some tests currently open documents with the wrong parameters
  - noRename and noFileCopy do *almost* the same thing, and subsequentLoad is similar but totally different
- The helper functions have misleading names
  - "beforeAll", "loadTestDocNoIntegration", and "reload" are all used in different places to reopen a document
- There is lots of boilerplate test setup code, which makes it harder to write new tests

### Commit 1: Remove afterEach/afterAll
Skip closing the document at the end of the test.
 - Cypress strongly recommends NOT cleaning up after tests ([The docs explain why](https://docs.cypress.io/guides/references/best-practices#Using-after-Or-afterEach-Hooks))
 - Documents get closed automatically after leaving the page anyways 
   - There is a C++ test explicitly for this, no need to test it with every cypress test
 - Many tests verified the wrong document was closed anyways (the original file instead of the copied file with a random name).
 - Removes a small amount of boilerplate from every test
 - Removes the need for most tests to know the name of the new random file name (necessary for later cleanup)
 - This makes failure investigations slightly easier because the page stays open after the test finishes when using make run

### Commit 2: Replace beforeAll with setupAndLoadDocument
Add convenience function "setupAndLoadDocument" used by most tests.
 - Does the same thing as beforeAll but simpler and more explicitly
 - Uses a relative path like 'calc/hello.ods' instead of separate parameters 'hello.ods' and 'calc'.
 - Applied to most test files, where it was straightforward to do so
 
### Commit 3: Remove beforeAll and rewrite document loading helpers
Separate the document setup and load.
- Splits huge parameter list and removes the need for some of the parameters altogether

Modify all remaining tests to use the new functions.
- Mostly documents that need to reload the file
- Also spec files without beforeEach because the individual testpoints use different documents
- Also tests that skip documentChecks because of an interaction before load

Functions changed in helper.js:
  - getRandomFileName, generateDocumentURL, generateDocumentURI: inlined
  - logLoadingParameters: deleted
  - loadTestDocNoIntegration: renamed to loadDocumentNoIntegration, made private
  - loadTestDocNextcloud: renamed to loadDocumentNextcloud
  - loadTestDoc: split functionality to setupDocument and loadDocument
  - checkIfDocIsLoaded: inlined. documentChecks is now public.
  - reload: renamed to reloadDocument

## Risks
- Many tests are currently skipped
  - I did run these where possible. The document loading worked, but many were failing for other reasons. Without full qualification in CI there is a chance that some minor issue won't be caught. But it would be fixed when the test is re-enabled anyways.
- Nextcloud/php-proxy tests
  - I did not run these tests. They are already broken, but there is a small chance I broke them more.

## Example
Here is an example to show you how this change affects a basic test.
Screenshot from 2024-04-24 14-03-48

- beforeEach: clearer
- afterEach: deleted
- filename variables: not needed

### DIFF
--- a/cypress_test/integration_tests/common/desktop_helper.js
+++ b/cypress_test/integration_tests/common/desktop_helper.js
@@ -420,10 +420,12 @@ function pressKey(n, key) {
 	cy.log('<< pressKey - end');
 }
 
-function openReadOnlyFile(type, filename) {
+function openReadOnlyFile(subFolder, fileName) {
 	cy.log('>> openReadOnlyFile - start');
 
-	var testFileName = helper.loadTestDocNoIntegration(filename, type, false, false, false);
+	var newFileName = helper.setupDocument(fileName, subFolder);
+	// Skip document checks because sidebar does not appear
+	helper.loadDocument(newFileName, subFolder, true);
 
 	//check doc is loaded
 	cy.cGet('.leaflet-canvas-container canvas', {timeout : Cypress.config('defaultCommandTimeout') * 2.0});
@@ -433,7 +435,7 @@ function openReadOnlyFile(type, filename) {
 	cy.cGet('#PermissionMode').should('be.visible').should('have.text', ' Read-only ');
 
 	cy.log('<< openReadOnlyFile - end');
-	return testFileName;
+	return newFileName;
 }
 
 function checkAccessibilityEnabledToBe(state) {
--- a/cypress_test/integration_tests/common/helper.js
+++ b/cypress_test/integration_tests/common/helper.js
@@ -545,12 +545,6 @@ function beforeAll(fileName, subFolder, noFileCopy, isMultiUser, subsequentLoad,
 	return loadTestDoc(fileName, subFolder, noFileCopy, isMultiUser, subsequentLoad, hasInteractionBeforeLoad, noRename);
 }
 
-// testState no longer used, but not removed from all calls to afterAll yet
-// eslint-disable-next-line no-unused-vars
-function afterAll(fileName, testState) {
-	closeDocument(fileName);
-}
-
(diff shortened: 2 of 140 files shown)


neilguertincollabora:
> I like this :D
> 
> If we want to be sure we don't leave zombie kits - we can always check that after all the tests So we will have single place at the end where we open admin console and check if number of opened documents i 0. @neilguertincollabora what do you think?

We could do this! I just quickly added this after all hook to support/index.js and it seems to work:
```
after(function() {
	// Make sure all documents have been automatically closed
	cy.visit('http://admin:admin@' + Cypress.env('SERVER') + ':' +
		Cypress.env('SERVER_PORT') +
		'/browser/dist/admin/admin.html');
	cy.get('#doclist').should('be.empty');
});
```

I'm not sure it's worth it though. We would lose the ability to interact with documents after they finish in make run. And I think it would be confusing if it ever fails sporadically, because the error message is not very descriptive, and the logs don't get printed with cypress-terminal-report, and it's not clear where the after all hook is running from.

Is failing to close documents something we've seen before, and something that really needs to be locked down? Maybe we can do it from somewhere else, like at the end of make check?

eszkadev (approved):

PR #8833, opened 2024-04-20
Private/mmeeks/bgsave12

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
A couple for you here Ash - not that clear on what is going on there =)

Ashod (approved):
Thanks, @mmeeks. As discussed offline, the defaults don't matter since these two are unrelated flags. We can indeed have both true (for PDF, f.e.). We set these flags through `setWritable(bool)`, which takes care of these cases. See `DocumentBroker::download()` for details.

PR #8834, opened 2024-04-21
cid#360697 Ignoring number of bytes read

caolanm:
Change-Id: I24c1bf9f103f3bf1325246a0223919ff1dbf62b4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Makes sense, it's good to fix the noise so the rest stands out more clearly.

PR #8839, opened 2024-04-22
private/pedro/add missing tooltips add cursor pointer to btn

pedropintosilva:
- Fix unobutton's lack of pointer cursor
- Use the same tooltip style and framework for doc name & close button


eszkadev (approved):
looks good

PR #8843, opened 2024-04-22
includes should be in AC_LANG_PROGRAM preamble

caolanm:
no the body

Change-Id: I9377142c676efc3f0d658efa01e06a91f3f0924d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
AC_LANG_PROGRAM puts its 2nd arg into the function body it creates, and the 1st arg before the body, so having the include in the "body" isn't right

vmiklos (approved):
Aha, reading https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Generating-Sources.html, the old code put the include inside main(), which worked just by accident.

PR #8845, opened 2024-04-22
Introduce HTMLObjectSection

gokaysatir:
Change-Id: I90719321781afcd29b72aa6d8eaafb186b4238dc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -225,6 +225,7 @@ COOL_JS_LST =\
 	src/layer/tile/TilesSection.ts \
 	src/layer/tile/AutoFillMarkerSection.ts \
 	src/layer/tile/CellCursorSection.ts \
+	src/layer/tile/HTMLObjectSection.ts \
 	src/layer/vector/CEventsHandler.ts \
 	src/layer/vector/CPointSet.ts \
 	src/layer/vector/CPath.ts \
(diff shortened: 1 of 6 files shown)


eszkadev (changes requested):
I found 2 problems:
- it doesn't scale a button when zoomed in
- when button is shown and we zoom in/out - it shows button at wrong position

[dropdown-2024-04-23_22.40.37.webm](https://github.com/CollaboraOnline/online/assets/5307369/91ac7953-b3d4-4061-9029-0360fd3357c5)

browser/Makefile.am
```diff
@@ -226,6 +226,7 @@ COOL_JS_LST =\
 	src/layer/tile/TilesSection.ts \
 	src/layer/tile/AutoFillMarkerSection.ts \
 	src/layer/tile/CellCursorSection.ts \
+	/src/layer/tile/HTMLObjectSection.ts \
```
hfiguiere, line 229: shouldn't this be `src/layer/tile/HTMLObjectSection.ts \` (no starting `/` ?

gokaysatir: Thank you for checking.

eszkadev:
Thanks, now it works correctly with the zoom. But still it is very small (16x16), maybe we should scale that too?

eszkadev (approved):
code looks ok

gokaysatir:
> Thanks, now it works correctly with the zoom. But still it is very small (16x16), maybe we should scale that too?

I agree, just couldn't be sure. Old code works that way - unscaled. Maybe we can consult @pedropintosilva and prepare a follow up? Thanks for the review btw :)

PR #8847, opened 2024-04-23
cool#8648 clipboard: stop fetching the clipboard on text selection create in tests

vmiklos:
This was just a workaround to keep tests working till they are fixed to
not assume we do an implicit 'copy' as soon as we have a text selection.

Change-Id: I37d16ee2841c41593a5af2fb6ceba025fe1c79e0


vmiklos:

Finally we can stop cheating when we're under test.

caolanm (approved):

PR #8848, opened 2024-04-23
Allow to embed the presentation in window

eszkadev:
Similar like in:
commit d9d13d7092d2d597eb34d657945f2678756fb800
wasm: support serving wasm files

add headers required when we use WASM so we are able to embed presentation in window.

Without that we get:
not-set Cross-Origin-Embedder-Policy


caolanm (approved):
yeah, that seems to fit with the overall "every frame has to agree to be part of the embedding party" pita

PR #8851, opened 2024-04-23
wsd: allow for recovering from unloading document

Ashod:
When the last connection is closed, we flag the
document for unloading, until we save and
unload and have nothing further to do. However,
if during that time a new client joins the
document, we reset the unloading state and
recover.

A previously failing and disabled test is now
re-enabled and passes.

(cherry picked from commit 1e206ac0016f08d19daa0e34b8f042bfe7eba610)

Change-Id: I425a9286ec5df56064132e5e4227227761c65830

### DIFF
--- a/test/httpwstest.cpp
+++ b/test/httpwstest.cpp
@@ -47,8 +47,7 @@ class HTTPWSTest : public CPPUNIT_NS::TestFixture
 
     CPPUNIT_TEST(testExoticLang);
     CPPUNIT_TEST(testSaveOnDisconnect);
-    // This test is failing
-    //CPPUNIT_TEST(testReloadWhileDisconnecting);
+    CPPUNIT_TEST(testReloadWhileDisconnecting);
     CPPUNIT_TEST(testInactiveClient);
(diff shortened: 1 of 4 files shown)


eszkadev (approved):
I can confirm it fixed my issue

mmeeks:
Hmm - are we convinced there is not quite some risk here? I'm not convinced that this doesn't need more testing in 24.04 first (?)

mmeeks:
Not convinced. I thought we found more follow on problems and fixed them around this; I think at this point it's a reasonable thing to ask people to update to 24.04 and get various other benefits, rather than potentially de-stabilizing 23.05. We can resolve a ticket with "please upgrade to 23.05".

I'd prefer to close this unless there is lot of screaming =)

Ashod:
> I thought we found more follow on problems and fixed them around this

Not to my knowledge. This just improves the chances of recovering an unloading document, rather than turning the client down.

Anyway, I don't know of anyone asking for this, so best to close. I believe you had asked for improving recovery, but perhaps I misremember.

PR #8857, opened 2024-04-23
do invalidation tests without spellchecking

caolanm:
which may kick in at an arbitrary point. We shouldn't invalidate if there is no change, but that's a TODO.

Change-Id: I910c96f2c2a0efab393ccea52c9f25d5804cd967

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


neilguertincollabora:
Please add more information to the comment about why this is important for this test. Something like "Turn off spellchecking by default. Spellchecking adds an extra empty update when it kicks in after a change".

neilguertincollabora (approved):

PR #8859, opened 2024-04-23
Private/caolan/extend delta test string

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
don't want this anymore, private/caolan/make_header_invalidation_test_reliable does what I need instead

PR #8864, opened 2024-04-24
css: fix insert shape popup

eszkadev:
insert shape popup now shows all the items in one column. fix it so we have rows


eszkadev:
CodeQL stopped working also in other PRs

gokaysatir (approved):

PR #8865, opened 2024-04-24
Part Preview update

hcvcastro:
- **browser: preview: remove more unused code**
- **browser: preview: simplify "layoutPreview"**
- **browser: preview: change place hoder to svg**


browser/src/control/Control.PartsPreview.js
gokaysatir, line 1: I thought we would remove line 91. The lonely this._getBottomBound call :) Let me also check other changes here.

gokaysatir (approved):

PR #8867, opened 2024-04-24
backport: iOS: stop keyboard pushing content offscreen

Minion3665:
Previously using the onscreen keyboard would shrink the visual viewport while leaving the elements at the same size, meaning some of the content had to be offscreen. As we scrolled the page to have the cursor in view, this would move the notebook bar offscreen.

This commit does 2 things:
- First, we stick the viewport to the top of the screen with CSS, this avoids the possibility of the notebook bar being moved offscreen
- Then, to avoid the possibility of the cursor leaving the screen or the bottom toolbar being offscren, we use the VisualViewport API to force the logical viewport to have the same size as the visual one

Together, these vastly improve the iOS keyboard popup experience

Change-Id: I07726bee8b61334f6a32e873ab2d5428fa60dca3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
seems to be ok

PR #8871, opened 2024-04-25
cool#8648 clipboard: fix idle tests

vmiklos:
This is not part of CI at the moment, so I didn't notice it till Szymon
pointed it out.

	make -C cypress_test check-idle spec=calc/idle_spec.js

passes with this, which is the only suite in 'make check-idle'.

Change-Id: Ieeb15653abcc016bc7de87115df136dfe3f27d0d


vmiklos:
CC @eszkadev @neilguertincollabora

PR #8872, opened 2024-04-25
Prevent opening multiple tooltip on drawing area widgets

GulsahKose:
We see multiple tooltip problem on drawing area widgets. Example case:
    Add basic shape into a writer document
    Right click on shape and open Position and Size dialog
    Select the Rotation tab
    Hold the rotation angle widget and drag around itself
When you inspect the "Rotation Angle" tooltip, count may be 600 We expect only 1.

So we prevent to show another tooltip if we have already one.

Change-Id: I2f548f8c7afce8095e83a6faa3722e164f990210

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
thank you, works as expected

PR #8878, opened 2024-04-25
Replace textselection markers with textselection handlers.

gokaysatir:
Change-Id: I2de0c4720250022d5547f3f657e277949721ba92

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -3260,8 +3259,20 @@ L.CanvasTileLayer = L.Layer.extend({
 				this._selectionContentRequest = setTimeout(L.bind(function () {
 					app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');}, this), 100);
 			}
+
+			this._selectionHandles.start.showSection = true;
```
eszkadev: to separate method: `showSection()`

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -4858,63 +4807,52 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._textSelectionStart = null;
 		this._textSelectionEnd = null;
 		this._selectedTextContent = '';
-		for (var key in this._selectionHandles) {
-			this._map.removeLayer(this._selectionHandles[key]);
-			this._selectionHandles[key].isDragged = false;
-		}
+
+		this._selectionHandles.start.showSection = false;
```
eszkadev: reuse separate method: `hideSection()`

browser/.beforeprettier
```diff
@@ -133,7 +133,10 @@
 /src/layer/SplitPanesContext.ts
 /src/layer/tile/AutoFillMarkerSection.ts
 /src/layer/tile/CellCursorSection.ts
+/src/layer/tile/CellSelectionHandleSection.ts
```
eszkadev, line 136: It was proposed to clarify folder structure a bit.
Maybe we could from now put new files related to canvas sections put into `src/canvas` ?
It will be shorter and will separate new files from leftovers from the leaflet.
It's no longer about separate tiles :)

browser/src/layer/tile/CalcValidityDropDownSection.ts
```diff
@@ -0,0 +1,26 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+class CalcValidityDropDown extends HTMLObjectSection {
+
+	constructor (sectionName: string, documentPosition: cool.SimplePoint, visible: boolean = true) {
+		super(sectionName, 16, 16, documentPosition, 'spreadsheet-drop-down-marker', visible);
+	}
+
+    onClick(point: number[], e: MouseEvent): void {
```
eszkadev, line 18: indent

browser/src/layer/tile/HTMLObjectSection.ts
```diff
@@ -36,14 +36,25 @@ class HTMLObjectSection extends CanvasSectionObject {
...
+		// canvas-container and canvas overlap entirely. We can append the html object to canvas-container.
+		const tempFunction = function(elementToAdd: any) {
+			const container = document.getElementById('canvas-container');
+			if (container) {
+				container.appendChild(elementToAdd);
+			}
+			else {
+				setTimeout(() => {
+					tempFunction(elementToAdd);
+				}, 100);
```
eszkadev, line 54: can we make it deterministic? without timeout?
Like call from some specific place outside, which will be executed AFTER we already initialized required elements?

browser/css/device-desktop.css
```diff
@@ -9,7 +9,7 @@
 	margin-left: 1px !important;
 	transform: translateY(-12px);
 }
-#tb_formulabar_item_formula .inputbar_selection_handles .leaflet-selection-marker-start{
+#tb_formulabar_item_formula .inputbar_selection_handles .text-selection-handle-start{
```
eszkadev, line 12: `#tb_formulabar_item_formula` doesn't exists anymore after w2ui removing.
We can drop this rule then

browser/css/device-desktop.css
```diff
@@ -1,15 +1,15 @@
 /* CSS specific for desktop browsers. */
 
 /* Related to selectionMarkers.css on formulabar*/
-#tb_formulabar_item_formula .inputbar_selection_handles *{
+.inputbar_selection_handles *{
```
eszkadev, line 4: we don't use tuneled formulabar - we don't have selection handles

so we can remove these 2 rules (as they even were not used for a long time)

browser/css/device-desktop.css
```diff
@@ -1,15 +1,6 @@
...
-.inputbar_selection_handles *{
-	background-size: 20px;
-	background-position-y: top;
-	width: 20px;
-	background-repeat: no-repeat;
-	margin-left: 1px !important;
-	transform: translateY(-12px);
-}
-.inputbar_selection_handles .text-selection-handle-start{
+.text-selection-handle-start{
```
eszkadev, line 3: I think it should be compleely removed, previously it was applied only to formulabar, now for all `.text-selection-handle-start`

browser/css/selectionMarkers.css
```diff
@@ -18,10 +18,7 @@
 	height: 44px;
 	background-image: url('images/cursor-handler.svg');
 }
-.inputbar_selection_handles * {
-	z-index: 1;
-}
-#tb_formulabar_item_formula .inputbar_selection_handles{
+#tb_formulabar_item_formula {
```
eszkadev, line 21: remove this rule, that element no longer exists

cypress_test/integration_tests/mobile/calc/insertion_wizard_spec.js
```diff
@@ -77,6 +77,8 @@ describe(['tagmobile', 'tagnextcloud'], 'Calc insertion wizard.', function() {
 		calcHelper.selectEntireSheet();
 		helper.copy();
 
+		return;
```
eszkadev, line 80: `return` in cypress test, before the end?

browser/css/device-desktop.css
gokaysatir: This one seems used. Let's check this later after this PR.

eszkadev: It wan't possible that `#tb_formulabar_item_formula .inputbar_selection_handles .leaflet-selection-marker-start{` was used -> `#tb_formulabar_item_formula` doesn't exists anymore.

It was used when we had tunneled formulabar, then we creatred handles for mobile I guess.

eszkadev:
now circles in the corner are visible but lost tyling: are transparent, then after usage - blue.
Before change they were always white

gokaysatir:
> now circles in the corner are visible but lost tyling: are transparent, then after usage - blue. Before change they were always white

Yep, that was intentional to make the user feel that they are moving the circles. But I'll send a rebase anyway, let me cancel it. Maybe we can beautify it after consulting to Pedro :)

eszkadev:
> > now circles in the corner are visible but lost tyling: are transparent, then after usage - blue. Before change they were always white
> 
> Yep, that was intentional to make the user feel that they are moving the circles. But I'll send a rebase anyway, let me cancel it. Maybe we can beautify it after consulting to Pedro :)

ok, it's fine. But having invisible handles in Dark Mode is bad and a regression (they are transparent not white)

gokaysatir:
> I see a problem on tablet (iPad):
> 
> * in Calc when I click on a cell to select it - I see only blue rectangle handle, but not white circles in the corner
> * I see sometimes circles but without white color inside - its visible in dark mode

Thank you for checking, i also corrected that one. I hope it is now ready to be merged?

eszkadev (approved):

PR #8881, opened 2024-04-25
Private/mmeeks/doccleanup

mmeeks:
Improved unit test framework coming next for the removed test - this is mostly to check CI =)

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -181,6 +181,13 @@ namespace Log
         }
     };
 
+    void postFork()
+    {
+        /// after forking we can end up with threads that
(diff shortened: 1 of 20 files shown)


vmiklos (approved):
I like the approach in general. Mocking individual classes is a lot of maintenance, I rather prefer doing that only at the edge of the system (like network, filesystem, time, etc), and this PR moves in that direction.

As you suggest, best to only merge this PR once there is a replacement for the removed test.

vmiklos:
I don't see anything obviously problematic with the commits pushed since https://github.com/CollaboraOnline/online/pull/8881#pullrequestreview-2024235958

PR #8882, opened 2024-04-25
Cypress: Join subFolder and fileName into filePath

neilguertincollabora:
Removes the need to add subFolder information separately or duplicate it. For example, calc/macro_spec.js changes from
```
var newFileName = helper.setupDocument('macro.ods','calc');
helper.loadDocument(newFileName,'calc',true);
```
to
```
var newFilePath = helper.setupDocument('calc/macro.ods');
helper.loadDocument(newFilePath,true);
```

Follow up of https://github.com/CollaboraOnline/online/pull/8832 

Change-Id: I361d523f564e7b9a46dd2398080eef110eeae79c

### DIFF
--- a/cypress_test/integration_tests/common/desktop_helper.js
+++ b/cypress_test/integration_tests/common/desktop_helper.js
@@ -420,12 +420,12 @@ function pressKey(n, key) {
 	cy.log('<< pressKey - end');
 }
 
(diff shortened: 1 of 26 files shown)


eszkadev (approved):
looks ok

PR #8884, opened 2024-04-26
macro: open document without macros enabled

lpranam:
problem:
when user selected disable macro in the opening dialog
document used to close and return to integrator
but now cancel button in csv import dialog is hidden,
so no need to distinguish there and we can continue to load without macros
In addition we still close the file if user clicks "X" button on title bar

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -282,8 +282,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			// returns docnotloaded error. Instead of this we can return to the integration
 			if (!builder.map._docLoaded &&
 				 !window._firstDialogHandled &&
-				 ((object.id === 'cancel' || eventType === 'close') ||
-				 (objectType === 'responsebutton' && (data == 0 || data == 7)))) {
+				 (eventType === 'close' ||
+				 (objectType === 'responsebutton' && data == 7))) {
 				window.onClose();
 			}
 			switch (typeof data) {


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -76,6 +76,7 @@
 		this.windowId = options.windowId;
 		this.dialogId = null;
 		this.callback = options.callback ? options.callback : this._defaultCallbackHandler;
+		this.dialogId = options.dialogId
```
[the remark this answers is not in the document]
eszkadev: missing semicolon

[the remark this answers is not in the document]
eszkadev: duplicated property, above you have dialogId = null

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -277,11 +278,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			var encodedCommand = data.replace(' ', '%20');
 			builder.map.sendUnoCommand(encodedCommand);
 		} else if (object) {
-			// CSV and Macro Security Warning Dialogs are shown before the document load
+			// CSV Dialog is shown before the document load
 			// In that state the document is not really loaded and closing or cancelling it
 			// returns docnotloaded error. Instead of this we can return to the integration
 			if (!builder.map._docLoaded &&
 				 !window._firstDialogHandled &&
+				 !builder.dialogId.includes('MacroWarn') &&
```
eszkadev, line 286: I don't like idea of hardcoding ids. It can be easily broken in the future.
What will happen on "X" / close button in the dialog? From comment I read we wanted to go back to the integrator. Now it will not work with Macro dialog, right?

I think the problem where is that we explicitly wanted to go back to the integration on "cancel" button too.

Is it possible to just change that condition below?

Please test that with CSV import dialog as well. Also on mobile which doesn't use JSdialog but MobileWizard component.

lpranam: Currently it won't go back to integrator bit will open doc with macro disabled...

lpranam: Agreed about hardcoded id... Do you have nay suggestions on what approach we could use here?

eszkadev: I would prefere to just change the condition below:
- if code is "close" -> always go to integrator / close COOL, execute this code block

- if code is "cancel" - in macro dialog it should just open with disabled - so don't enter this code block below, allow to send the jsdialog message
- if code is "cancel" in csv import it makes no sense to me... so we should hide cancel button in the csv dialog if it is opened as the initial dialog before loading the document. So user can import the CSV successfully or just go back to the file explorer.  We don't have B plan. cc: @pedropintosilva 

That will make it working without hardcoding anything

lpranam: took your suggestions.

corresponding core patch: https://gerrit.libreoffice.org/c/core/+/166764

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -282,8 +282,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			// returns docnotloaded error. Instead of this we can return to the integration
 			if (!builder.map._docLoaded &&
 				 !window._firstDialogHandled &&
-				 ((object.id === 'cancel' || eventType === 'close') ||
-				 (objectType === 'responsebutton' && (data == 0 || data == 7)))) {
+				 eventType === 'close') {
```
eszkadev, line 285: please also handle `responsebutton` case, it's present in some cases.
`-7` is `close` I think: https://opengrok.libreoffice.org/xref/core/vcl/source/window/builder.cxx?r=174430d7#set_response

eszkadev (approved):

PR #8885, opened 2024-04-26
Don't warn about version mismatch for admin console

thebearon:
It's only relevant for regular URLs that contain version hash.

Change-Id: I73afaa0559c3dddc09a8ed5f786fb5126e1c8f0c

* Target version: master 

### Checklist


Ashod (approved):
Thanks, @thebearon. Should reduce the noise.

PR #8902, opened 2024-04-26
cool#8806 clipboard: fix paste special of images in Firefox

vmiklos:
Copy an image to the clipboard, click Paste -> Paste special on the
notebookbar in Firefox: the image is pasted, but the paste special
dialog doesn't close.

What happens here is that Firefox lacks navigator.clipboard.read(), so
the relevant code is Clipboard.js dataTransferToDocumentFallback(), not
the new _navigatorClipboardGetTypeCallback(). Here e.g. copy from
desktop Writer (html+text) would lead to a _doInternalPaste(), which
would call _checkAndDisablePasteSpecial() for us, but the image case
returns early, so the closing doesn't happen.

Fix the problem by explicitly calling _checkAndDisablePasteSpecial()
before returning early.  In contrast with the other uses of
_checkAndDisablePasteSpecial(), we don't need to call the paste or paste
special uno command (since _asyncReadPasteImage() takes care of pasting
with a websocket comment), so it's fine to ignore the return value.

No testcase, our cypress tests run on Chrome, and this happens only with
Firefox.

Change-Id: I751d234edfdf9e8fb37f2bf2d6b3f3d8e65a27f4

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -467,6 +467,11 @@ L.Clipboard = L.Class.extend({
 						this._asyncReadPasteImage(dataTransfer.items[t].getAsFile());
 				}
 			}
+
+			// If any paste special dialog is open, close it here, because we won't call
+			// _doInternalPaste() that would do the closing.
+			this._checkAndDisablePasteSpecial();
+
 			return;
 		}
 


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_mobile/357/consoleFull failed the same way, but this passes for me locally, hmm.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_mobile/359/console again in impress/apply_font_text_spec.js, hmmm.

vmiklos:
Quoting: <https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1045/console>:

> Test failed on unit-wopi-save-on-exit after 130s.

Retry.

vmiklos:

An edge case: paste special, firefox, image paste.

caolanm (approved):

PR #8904, opened 2024-04-26
allow a test to start without sidebar

caolanm:
instead of starting with sidebar and then removing it

Change-Id: If2ea433cd3d5fc18ad549bd3df095de15c126ad0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


neilguertincollabora:
Another option I had to use for the annotation tests was to leave the sidebar out for the test but then expand the viewport so that the sidebar doesn't cover any content.

caolanm:
If I can do this then it will help to get a stable count of tiles requested between runs in future writer invalidation test improvements.

neilguertincollabora (changes requested):

neilguertincollabora (approved):

PR #8905, opened 2024-04-26
Make freeze shadow move with a pinch-zoom

Minion3665:
Previously the freeze shadow would stay where it was, creating a hard
edge and confusion about where was frozen. With this change, the freeze
shadow moves with the zoom.

Previously when panning with zoom, the freeze shadow would stay in its
initial state (either hidden or shown) whether or not it should actually
be shown at the current position.

Previously when there was a frozen pane we calculated an incorrect
size on the unfrozen pane, leading to slight blank edges when you were
mid-zoom-out.

This pull request fixes all 3 of these issues


eszkadev (approved):
it works on iPad, thank you :)

PR #8913, opened 2024-04-27
wsd: test: give up attempting to connect if the test finished

Ashod:
Some tests are designed to exercise the reconnection
logic. Once the test validates that the correct
reconnection logic is used, the test is considered
finished. In those cases, we need to break the
test framework's retry logic.

Change-Id: Ie0bb5dcc430a954910c2af4817a500f5b576bd4c

### DIFF
--- a/test/helpers.hpp
+++ b/test/helpers.hpp
@@ -11,6 +11,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Thanks Ash =)

Issue #8103, opened 2024-01-26
Double Space in "Find and Replace" = Searches for Single Space


Tex2002ans:
### Describe the Bug

In the "Find and Replace" menu, if you search for TWO spaces, it will find ONE space instead.

### Steps to Reproduce

In Writer:

0. Open [sample ODT document](https://github.com/CollaboraOnline/online/files/14059379/This.is.a.sample.sentence.with.double.spaces.odt). Or just use this sample text:

- `This is a sample sentence with  double  spaces  in it.`

1. In "Home" tab:
   - Click on "Find and Replace" button.
2. In "Find and Replace" popup:
   - **Find:** `  `
      - = TWO spaces
3. Press "Find" or "Find All" button.

### Expected Behavior

"Find" or "Find All" button should only find:

- **3 hits** of double spaces.

### Actual Behavior

It finds:

- **13 hits** of space.

### Desktop

- **Collabora version:** 23.05.7.5
- **OS and version:** Linux Minut 21.1
- **Browser and version:** Firefox 122.0 (64-bit)
 
```
COOLWSD version: 23.05.7.5 (git hash: db903a3d)
LOKit version: Collabora Office 23.05.7.5 (git hash: e8d97f2)
Served by: Linux Mint 21.1
Server ID: 84b0e70c
```

### Additional Context

Same exact thing happens in the "Replace" box as well (Step 2).

Also odd, it looks like the second you click out of the Find/Replace boxes, they "gobble up" any spaces at the end... 1 space before/after is fine, but 2+ spaces at beginning/end get shrunk to 1.

So if you type this into the box:

- `with  `
   - with + TWO spaces

and leave the box, it automatically turns into:

- `with `
   - with + ONE space

so this doesn't let you search OR replace with TWO OR MORE spaces at the beginning/end of your terms.

- - -

**Side Note:** I use this type of search and replace often. TWO spaces -> ONE space, since it's a very common typo (especially hidden at end of lines/paragraphs).

PR #8918, opened 2024-04-28
Tokenize Quotes

NickWingate:
* Resolves: #8103
* Target version: master 

### Summary

When ignoreQuotes = false delimiters within quotes (' or ") aren't tokenised.
Causes problems especially as space is the default delimiter, so if the input string has any `key="string"` pattern tokens a space will cause issues

For #8103 
Unocommand was being incorrectly tokenized as space is the default delimiter:

`uno .uno:ExecuteSearch {"SearchItem.SearchString":{"type":"string","value":" "}, ...`
`                                                                           ^^^`

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/StringVector.hpp
+++ b/common/StringVector.hpp
@@ -79,6 +79,52 @@ public:
             tokens.emplace_back(start - data, end - start);
     }
 
+    // Tokenize delimited values outside of quotes until we hit new-line or the end.
+    static void tokenize_outside_quotes(const char* data, const std::size_t size,
+                                        const char delimiter, std::vector<StringToken>& tokens)
+    {
+        if (size == 0 || data == nullptr || *data == '\0')
+            return;
+
+        // if delimiter is a quote default back to standard tokenize
+        if (delimiter == '\'' || delimiter == '"')
+        {
+            tokenize(data, size, delimiter, tokens);
+            return;
+        }
+
+        tokens.reserve(16);
+
+        const char* start = data;
+        const char* end = data;
+        char quote = '\0';
+
+        for (std::size_t i = 0; i < size && data[i] != '\n'; ++i, ++end)
+        {
+            if ((data[i] == '\'' || data[i] == '"') &&
+                (i == 0 || data[i - 1] != '\\')) // check for escaped quote
+            {
+                if (quote == '\0')
+                    quote = data[i];
+                else if (quote == data[i])
+                    quote = '\0';
+            }
+            if (quote == '\0' && data[i] == delimiter)
+            {
+                if (start != end && *start != delimiter)
+                    tokens.emplace_back(start - data, end - start);
+
+                start = end;
+            }
+            else if (*start == delimiter)
+                ++start;
+        }
+
+        if (start != end && *start != delimiter &&
+            *start != '\n') // tokenizes the end of the string
+            tokens.emplace_back(start - data, end - start);
+    }
+
     // call func on each token until func returns true or we run out of tokens
     template <class UnaryFunction>
     static void tokenize_foreach(UnaryFunction&& func, const char* data, const std::size_t size, const char delimiter = ' ')
@@ -133,6 +179,29 @@ public:
         return StringVector(s, std::move(tokens));
     }
 
+    /// Tokenize single-char delimited values outside of quotes until we hit new-line or the end.
+    static StringVector tokenize_outside_quotes(const char* data, const std::size_t size,
+                                                const char delimiter = ' ')
+    {
+        if (size == 0 || data == nullptr || *data == '\0')
+            return StringVector();
+
+        std::vector<StringToken> tokens;
+        tokenize_outside_quotes(data, size, delimiter, tokens);
+        return StringVector(std::string(data, size), std::move(tokens));
+    }
+
+    /// Tokenize single-char delimited values outside of quotes until we hit new-line or the end.
+    static StringVector tokenize_outside_quotes(const std::string& s, const char delimiter = ' ')
+    {
+        if (s.empty())
+            return StringVector();
+
+        std::vector<StringToken> tokens;
+        tokenize_outside_quotes(s.data(), s.size(), delimiter, tokens);
+        return StringVector(s, std::move(tokens));
+    }
+
     /// Tokenize by the delimiter string.
     static StringVector tokenize(const std::string& s, const char* delimiter, int len = -1)
     {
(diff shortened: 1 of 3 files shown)


common/StringVector.hpp
```diff
@@ -51,19 +52,35 @@ class StringVector
     }
 
     /// Tokenize delimited values until we hit new-line or the end.
+    /// Ignoring delimiters that are within quotation marks, unless ignoreQuotes is set to true
     static void tokenize(const char* data, const std::size_t size, const char delimiter,
-                         std::vector<StringToken>& tokens)
+                         std::vector<StringToken>& tokens, bool ignoreQuotes = true)
```
Ashod, line 57: I'm not a fan of reusing function with default parameters. They make maintenance harder later on.

I also worry about the cost of the new logic. These tokenerizers are heavily used for each and every command, and multiple times at that. Perhaps we should consider separating the cases where we expect quotes vs those that we never support quotes and have separate implementations?

common/StringVector.hpp
```diff
@@ -111,64 +128,84 @@ class StringVector
...
         if (s.empty() || len == 0 || delimiter == nullptr || *delimiter == '\0')
             return StringVector();
 
         if (len < 0)
             len = std::strlen(delimiter);
 
-        std::size_t start = 0;
-        std::size_t end = s.find(delimiter, start);
+        std::string delimiterString(delimiter);
+        if(delimiterString.find('\'') != std::string::npos || delimiterString.find('"') != std::string::npos)
```
Ashod, line 163: Please use clang-format (you can configure your editor to run it on *modified* lines upon save).

common/StringVector.hpp
```diff
@@ -111,64 +128,84 @@ class StringVector
...
-        start = end + len;
-
         while (end != std::string::npos)
         {
-            end = s.find(delimiter, start);
-            tokens.emplace_back(start, end - start);
-            start = end + len;
+            end = s.find(delimiter, searchStart);
+
+            if (end != std::string::npos) {
```
Ashod, line 183: Please observe the coding convention used throughout the C++ codebase. Use clang-format to help with formatting.

common/StringVector.hpp
```diff
@@ -111,64 +128,84 @@ class StringVector
...
     {
         if (s.empty() || len == 0 || delimiter == nullptr || *delimiter == '\0')
             return StringVector();
 
         if (len < 0)
             len = std::strlen(delimiter);
 
-        std::size_t start = 0;
-        std::size_t end = s.find(delimiter, start);
+        std::string delimiterString(delimiter);
```
Ashod, line 162: We should use `strchr` to search for a single character in a C-string (https://en.cppreference.com/w/c/string/byte/strchr). Creating a `std::string` is unnecessarily inefficient in this case.

common/StringVector.hpp
```diff
@@ -111,64 +128,84 @@ class StringVector
...
         while (end != std::string::npos)
         {
-            end = s.find(delimiter, start);
-            tokens.emplace_back(start, end - start);
-            start = end + len;
+            end = s.find(delimiter, searchStart);
+
+            if (end != std::string::npos) {
+                // s.begin() + std::string::npos causes issues
+                singleQuotesBefore = std::count(s.begin(), s.begin() + end, '\'');
```
Ashod, line 185: Very costly to do this when `ignoreQuotes` is true. Also, isn't it better to do `find_first_of("\'\"", searchStart, end - searchStart + 1);`

(Actually, the more efficient way to implement support for quotes is to do the search inline. That is, to go through each character and see if it's one of the quotes or the delimiter, and act accordingly. That would at least be a single linear search, where now this algorithm is at least quadratic. But, perhaps it's best to split this into two implementations, the old one that doesn't support quotes, and a new one that might be inefficient but supports quotes.)

common/StringVector.hpp
```diff
@@ -111,64 +128,84 @@ class StringVector
...
+                doubleQuotesBefore = std::count(s.begin(), s.begin() + end, '"');
+            }
+            // if no. quotes before the delimiter is odd then it is within quotes
+            if (ignoreQuotes || (singleQuotesBefore % 2 == 0 && doubleQuotesBefore % 2 == 0) ||
```
Ashod, line 189: A better algorithm is to find either the delimiter or a quote. If the quote is before the delimiter, then search for the matching quote before searching for the delimiter again, or the end of string. There shouldn't be a need to count quotes (note that counting is a side-effect of the unnecessary quadratic algorithm).

Ashod (changes requested):
Thanks, @NickWingate. This is a good bug to fix.

I worry about the efficiency of the code, though. Given how frequently the tokenizers are called, we really can't afford to have an algorithm that's not linear. I left some notes and tips, but please feel free to reach out if need further help. Re-writing these functions to be optimal might not be trivial, so I suggest splitting them, since it seems we only care about supporting quotes in exactly one case. That single case can be less efficient, until we improve the implementation. I just don't want to make all the command processing code slower because of that single case.

Ashod:
> Made a new function `tokenize_outside_quotes` for handling cases where the data may have quotes. Also changed to a new algorithm as suggested: go character by character and when a quote is found stop looking for the delimiter until quote is closed. When the delimiter is character this is simple, when the character is a string we use `s.substr(i, len) == delimiter` to compare.

Thank you, @NickWingate, for the update.

I went through the code, but, before commenting on the code, I'm wondering why there is no test with the reported issue of searching for `"  "` (two spaces). Shouldn't one expect that if this resolves the reported issue, and assuming the issue is related to a bug in the tokenizer, that a test would include a reproducer of said issue?

Also, I don't understand why we have implemented so many variations of the tokenizer when it would seem that we only use one. Perhaps I missed something, but I don't see the code using all these variations.

Finally, if we have to implement a tokenizer that takes quotes into account, why do we care to implement a special version for when there are no quotes? Is that an optimization? Have we measured the three versions?

Thank you.

NickWingate:
>I went through the code, but, before commenting on the code, I'm wondering why there is no test with the reported issue of searching for `" "` (two spaces). Shouldn't one expect that if this resolves the reported issue, and assuming the issue is related to a bug in the tokenizer, that a test would include a reproducer of said issue?

Yes, good point. I will add the test.

>Also, I don't understand why we have implemented so many variations of the tokenizer when it would seem that we only use one. Perhaps I missed something, but I don't see the code using all these variations.

No, I just created further overloads of `tokenize_outside_quotes` where the original `tokenize` has these overloads already so if someone has a similar issue, but when using string delimiter instead of a char delimiter they can use this function without rewriting the `tokenize`. It seemed sensible at the time, but if you think this is just bloat I can remove it.

>Finally, if we have to implement a tokenizer that takes quotes into account, why do we care to implement a special version for when there are no quotes? Is that an optimization? Have we measured the three versions?

I did this as suggested:
>Re-writing these functions to be optimal might not be trivial, so I suggest splitting them, since it seems we only care about supporting quotes in exactly one case

In each `tokenize_outside_of_quotes` we check that if the delimiter is/contains a quote then default back to the normal `tokenize`. I thought this was easiest solution for now but if you have a better idea I can implement that instead?

caolanm:
The delim tokens are collapsed together. so using tokens.cat(' '...) doesn't get the original string back .

The StringVector keeps the original string, so if we have something like substrFromToken(int index) and used that to get the "remainder" of the string instead of tokens.cat(' ', 2) (in dialogEvent and unoCommand) then that would probably work too.

caolanm (approved):
lets merge this as is then

PR #8923, opened 2024-04-29
Custom properties dialog: fix grid and misalignments

pedropintosilva:
- Fix alignment between rows by:
  - Setting some of the columns' width
  - Removing margin from all buttons & set width. Add column gap
- Fix vertical alignment
- Fix header alignment (Name, Type, Value) so it aligns to the content
- Make "Add property" occupy full width (ideally this button would be
  part of the inner grid "last row" but for now this is already an
  improvement)
- Remove lengthly css targets and the use of nth-child

Change-Id: I85d9bdbcc365c735757ee086f6e4912d694a011c

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -2054,36 +2054,24 @@ kbd,
 }
 
 /* DocumentPropertiesDialog Custom props */
-#DocumentPropertiesDialog #customprops .jsdialog .ui-grid .menubutton {
+#DocumentPropertiesDialog #headerbox,
+#DocumentPropertiesDialog #properties > .jsdialog.ui-grid {
+	grid-template-columns: 128px 128px auto 26px !important;
+	align-items: center;
+	margin-block-start: 4px;
+}
+
+#DocumentPropertiesDialog #properties > .jsdialog.ui-grid > .ui-grid-cell > .ui-grid-cell {
+	align-items: center;
+	grid-column-gap: 4px;
+}
+
+#DocumentPropertiesDialog #customprops button {
 	line-height: 16px;
-}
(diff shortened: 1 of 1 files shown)


browser/css/jsdialogs.css
```diff
@@ -2054,36 +2054,24 @@ kbd,
...
-	margin-bottom: auto;
-}
-
-#DocumentPropertiesDialog #customprops #properties > .ui-grid {
-	display: flex;
-}
-
-#DocumentPropertiesDialog #customprops #properties > .ui-grid > div:nth-child(1) {
-	width: 150px;
+#headerbox,
```
eszkadev, line 2057: can we add `#DocumentPropertiesDialog` at the beginning? then it will be obvious for which dialog it is, and also will prevent applying the rule somewhere else. these ids are not that very unique

pedropintosilva: added, and also in the other cases

meven:
I would keep the namespacing with `#DocumentPropertiesDialog` a `#properties` can appear easily anywhere.

This needs some testing with all the different kind of field custom props can have. (date datetime, text, number).

eszkadev (approved):

pedropintosilva:
> I would keep the namespacing with `#DocumentPropertiesDialog` a `#properties` can appear easily anywhere.
Agreed, that's what I did, thanks for the heads up

> 
> This needs some testing with all the different kind of field custom props can have. (date datetime, text, number).

Yup I tested, we have still that button that appears blank but that is outside of the scope of this PR (I wonder if we have a GH issue for that, I remember we discussed but I don't know if I wrote somewhere)

PR #8933, opened 2024-04-30
iOS: update Info.plist

Minion3665:
There are various settings in Info.plist (generated from Info.plist.in at configure time) which are incorrect. Particularly important are the API and encryption declarations (which will be required to upload builds from tomorrow), and making the Collabora Office folder visible in iCloud (which stops files created from the recents tab being hidden)

Signed-off-by: Skyler <skyler.grey@collabora.com>

PR #8935, opened 2024-04-30
iOS: update Info.plist -> master

Minion3665:
There are various settings in Info.plist (generated from Info.plist.in at configure time) which are incorrect. Particularly important are the API and encryption declarations (which will be required to upload builds from tomorrow), and making the Collabora Office folder visible in iCloud (which stops files created from the recents tab being hidden)

This is a forward-port of #8933 

Signed-off-by: Skyler <skyler.grey@collabora.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,3 +167,6 @@ Makecache
 
 # autoconf stuff
 autogen.input
+
+# MacOS file manager metadata
+.DS_Store
--- a/ios/Mobile/Info.plist.in
+++ b/ios/Mobile/Info.plist.in
@@ -260,6 +260,12 @@
 	<string>$(PRODUCT_NAME) requires access to the camera in order for you to be able to take photos to be inserted while editing a document.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>$(PRODUCT_NAME) requires access to the photo library in order for you to be able to insert images into documents.</string>
+	<key>NSPrivacyAccessedAPITypeReasons</key>
+	<array>
+		<string>AC6B.1</string><!-- Allow reading com.apple.configuration.managed preferences (to use username if it is set in a profile) -->
+		<string>35F9.1</string><!-- Allow reading boot time to use timers -->
+		<string>3B52.1</string><!-- Allow reading file metadata when the user opens a file in Collabora Mobile -->
+	</array>
 	<key>UIAppFonts</key>
 	<array>
 		<string>share/fonts/truetype/Alef-Bold.ttf</string>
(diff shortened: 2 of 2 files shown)


mmeeks (changes requested):
ITSAppUsesNonExemptEncryption must be YES.

Minion3665:
```bash
[skyler@canvasback online]$ git range-diff daad3f3~4..daad3f3 HEAD~4..HEAD
1:  9c1775ef91 = 1:  d81c6cb85b iOS: Change document editing roles & ranks
2:  c5b40ccddf = 2:  24d7786aa3 Add macOS metadata to .gitignore
3:  7822b81982 ! 3:  aafbba9278 iOS: Add declarations for encryption and APIs
    @@ Metadata
     Author: Skyler <skyler.grey@collabora.com>
     
      ## Commit message ##
    -    iOS: Add declarations for encryption and APIs
    +    iOS: Add declarations for APIs
     
    -    To comply with export regulations, we need to tell Apple whether we use
    -    non-exempt encryption. We can do this manually through AppStore Connect
    -    but if we instead specify this in the Info.plist we will not have to do
    -    so repeatedly.
    -
    -    Similarly, some APIs require declaration of reason, and after May 1st
    -    this would prevent us from uploading new builds. This commit adds the
    -    required compliance information to explain why we use the APIs we do.
    +    Some APIs require declaration of reason, and after May 1st this would
    +    prevent us from uploading new builds. This commit adds the required
    +    compliance information to explain why we use the APIs we do.
     
         Please see https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api
         for more details on APIs which require reasoning to use
    @@ Commit message
         Change-Id: Ie6b05b754b2f3fd8e3e9b49d8e3b0581f63a8a1d
     
      ## ios/Mobile/Info.plist.in ##
    -@@
    -   <string>24.04.1</string>
    -   <key>CFBundleVersion</key>
    -   <string>@IOSAPP_BUNDLE_VERSION@</string>
    -+  <key>ITSAppUsesNonExemptEncryption</key>
    -+  <string>NO</string>
    -   <key>LSRequiresIPhoneOS</key>
    -   <true/>
    -   <key>LSSupportsOpeningDocumentsInPlace</key>
     @@
        <string>$(PRODUCT_NAME) requires access to the camera in order for you to be able to take photos to be inserted while editing a document.</string>
        <key>NSPhotoLibraryUsageDescription</key>
4:  daad3f3634 = 4:  4734fe7326 Make Collabora Office folder visible in iCloud
```

PR #8945, opened 2024-05-01
fix: can't reconnect and get to edit mode after full server shutdown

Rash419:
- regresssion from eadaf3f60694c8a3855fe1c842c5070b646e06f3

Change-Id: If2b20c7960e3c242064fc04425b4f893612f6362

* Resolves: # 
* Target version: master


eszkadev (approved):
Fixes reconnection and looks reasonable

Darshan-upadhyay1110 (approved):

PR #8947, opened 2024-05-01
_sessionUserInfo tracks disconnected sessions

caolanm:
as well as current ones.

i.e.

/// User Info container used to store user information /// till the end of process lifecycle - including
/// after any child session goes away

so add some info in the log to flag which ones are still connected and which ones are not.

https: //github.com/CollaboraOnline/online/issues/8943

Change-Id: I5350c04d1a7bb8095464881fba97e5910f71ffb3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1985,6 +1985,8 @@ bool Document::forwardToChild(const std::string& prefix, const std::vector<char>
 
                 _deltaGen->setSessionCount(count);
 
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Makes sense to me; thanks ! =) although its not entirely clear why we can't get this from the core on demand really - prolly it better belongs there in the long run. But good for now.

PR #8949, opened 2024-05-01
impress: restrict presentation in readonly mode

hfiguiere:
When export is disabled or watermarks are enabled, presentation in readonly mode is disabled

Change-Id: I4f7aff9f670f7523dfcf396f6009a272df9d5af8

This is a followup to #8831 

* Resolves: 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (changes requested):
Some good locations; but this needs another tweak. We should be deciding policy and enforcing in the server; what does tha tmean
* We should send a 'hidePresentation' flag or somesuch to the client which we synthesize on the server side, and apply in the client - rather than having a different policy algorithm in JS there.
* We should ensure that the policy is hard-enforced on the server; it should be impossible to inject the relevant presentation commands, they should be filtered on the server side if some nicely named function returns true for this user; we should belt+braces ensure that the download logic also refuses to download any (presentation) SVG in this case too.
Can you tweak ? =)

hfiguiere:
I hooked into filterMessage to prevent the download as on the server side.

mmeeks (approved):
Love it - thanks Hubert =)

hfiguiere:
Latest change is in `common/MobileApp.hpp` to fix the mobile build

PR #8953, opened 2024-05-02
Default to four pre-spawned children.

mmeeks:
They should be cheap memory wise, and save quite a bit of latency.

Keep 1 for debug mode - to make attaching to the right process easier.

Change-Id: I230be6ae61874f36b47ffc93d888d455bea3a7d8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
existing deployments will typically retain their config so we'll have to remember to bump this in our testing envs to match the new default.

Issue #8066, opened 2024-01-20
(calc): no possibility to hide the gridlines


Gitter61:
There is no option to hide the grid lines. The instructions are incorrect: there is no "Extras" "Options" menu item

Tex2002ans:
Sound like a great enhancement request. Thanks. :)

> The instructions are incorrect: there is no "Extras" "Options" menu item

What instructions?

Can you please point to them / link or explain where it says that?

PR #8956, opened 2024-05-02
calc: Properly handle the disable state

hfiguiere:
Doesn't hide the grid when the Function wizard is open anymore when disabling the ToggleSheetGrid button.

This is a follow up for issue #8066.

Requires https://gerrit.libreoffice.org/c/core/+/167370
**But this should land first.**

Change-Id: I79ba99716eb1103e7239fe5cc302c64827c9ef53

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -841,10 +841,14 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this.insertMode = e.state.trim() === '' ? false: true;
 		}
 		else if (e.commandName === '.uno:ToggleSheetGrid') {
-			let newState = e.state.trim() === 'true';
-			if (this._sheetGrid != newState) {
-				this._sheetGrid = newState;
-				this._painter._sectionContainer.requestReDraw();
+			let trimmedState = e.state.trim();
+			// Disabled mean we don't change the sheet grid state.
+			if (trimmedState != 'disabled') {
+				let newState = trimmedState === 'true';
+				if (this._sheetGrid != newState) {
+					this._sheetGrid = newState;
+					this._painter._sectionContainer.requestReDraw();
+				}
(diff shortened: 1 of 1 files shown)


eszkadev (commented):
Isn't that possible to avoid sending the message from the core? What is the root cause of hiding grid when opening the dialog?

hfiguiere:
there is large number of `statechanged` event that are sent. Not sure where and why, I suspect it's to set the UI state when the dialog is up.

Just adding `.uno:ToggleSheetGrid` to the list of state sent made this happen automatically, and it's necessary for the toolbar to be updated.

eszkadev:
For me this looks like masking the effect, we should fix the real issue. It is easy to be out of sync with "toggle" command by ignoring it.

State value is delivered by this code in core:
https://opengrok.libreoffice.org/xref/core/sc/source/ui/view/tabvwshf.cxx?r=3432dec5#601
Probably `GetShowGrid()` returns wrong value:
https://opengrok.libreoffice.org/xref/core/sc/source/ui/inc/viewdata.hxx?r=a7094c81&fi=GetShowGrid#464

All places where this "UNO command / sfx slot" used: https://opengrok.libreoffice.org/search?full=FID_TAB_TOGGLE_GRID&project=core

hfiguiere:
> State value is delivered by this code in core:
> https://opengrok.libreoffice.org/xref/core/sc/source/ui/view/tabvwshf.cxx?r=3432dec5#601

It always return 1.

hfiguiere:
And this code path isn't hit until AFTER I close the dialog. By that time the grid has been hidden and shown again.

hfiguiere:
Update the patch. This will need https://gerrit.libreoffice.org/c/core/+/167370 to work, but this should land first.

eszkadev (approved):
looks good now :)

PR #8957, opened 2024-05-02
Private/mmeeks/bgsave14

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
@Minion3665 Can you test this on Android - I did a somewhat grisly hack there to switch to the new, cleaner progress: message which is un-tested on Android.

@pedropintosilva - I also switched our loading animation to use a progress bar - which looks prettier & 'faster' somehow I think; there is a hook there for having a 'setValue' progress which will go one way 0 -> 100.

Possibly we should re-consider the Nextcloud overlay at some point, and/or polish ourselves to the point that they are happy to show our UI starting and loading with real progress and (hopefully) helpful error messages for users.

Minion3665:
> Can you test this on Android

Sure, I'll look at that right away. Thanks for pinging me!

Minion3665:
> > Can you test this on Android
> 
> Sure, I'll look at that right away. Thanks for pinging me!

I'm once again having some build issues with Android (my previous build was 23.05, but this is against master) ... I'll keep you posted on testing this

Minion3665:
@mmeeks 

I've managed to test this on Android now, it seems that, as you feared, the load progress bar breaks.

It still loads the document and the progress bar wasn't excellent before (only really showing movement just before the document opened) so, as this is important, you have my go-ahead to merge regardless of this regression. Please make a ticket about it too though.

eszkadev:
> Seems like CI breakage is unrelated; even a couple of simple comment changes breaks CI though less badly (but this JSDialog linking stuff is ...) #8965 @eszkadev any ideas ?- badly need this and more in for Monday.

This linking error was one-time thing it seems. Some poluted build probably. On matrix I will put my notes about failing CI.

caolanm (approved):

PR #8958, opened 2024-05-02
add a test for joining doc and disconnecting

caolanm:
a normal case, where user#2 joins the document successfully and then their connection closes; where the remaining client gets the notification that there is only one client in the session.

an abnormal case, failing test currently disabled, where user#2 joins the document, "load url" is launched, but the connection is immediately dropped; the other client gets a notification that a user has joined to make it a 2 user session, but no notification that the user has dropped.

Change-Id: I8271fb3d4def917acdff0fae6a3a52b5d02ef7af

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -86,6 +86,7 @@ all_la_unit_tests = \
 	unit-prefork.la \
 	unit-bad-doc-load.la \
 	unit-hosting.la \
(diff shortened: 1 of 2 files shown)


mmeeks:
Can't possibly break Cypress AFAICS =) so ...

PR #8963, opened 2024-05-03
Fix buttons visibility in formulabar on cancelformula

eszkadev:
Make code more deterministic by removing old hacks with timeout and cursor detection from:
- 6037d52
- b62320c

Bug was that on desktop when we were in formula edit mode and we clicked cancel button - it was not showing again the regular buttons but instead it was showing the accept/cancel buttons.

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -1081,36 +1081,27 @@ L.Map.include({
 	},
 
 	onFormulaBarBlur: function() {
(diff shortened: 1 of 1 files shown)


pedropintosilva:
tested, it works <3

eszkadev:
> tested, it works <3

@pedropintosilva passed CI after rebase

pedropintosilva (approved):

PR #8965, opened 2024-05-03
Private/mmeeks/commentfixes

mmeeks:
Lets see how CI copes with some plain comment changes ;-)


Ashod (approved):
Thanks, @mmeeks. Do we not want to prefix every privileged-code line in security.h with `/*WARNING*/` prefix, as done in ForKit.cpp?

PR #8967, opened 2024-05-03
clean up previews

hcvcastro:
- **browser: preview: fix time out resize event**
- **browser: preview: simplify 'isPreviewVisible'**


gokaysatir (approved):
Thanks for the simplifications :)

PR #8979, opened 2024-05-06
UI: improve combobox appearence

meven:
Since f7491e1f62de5401529db40b8b899e5c0563badc all combobox started having a with and min-width of 100px, this wasn't meant to be. Restrict this width setting to .jsDialog.

Add 5 px left padding to combobox, this looks better.

Set the font combobox width to 5em instead of 50px.

After:

Change-Id: I0f0247471376e4633f3bbffab89285e16f78cde9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -379,11 +379,14 @@ button.ui-tab.notebookbar {
 }
 
 #fontsizecombobox.notebookbar {
-	width: 50px !important;
+	width: 5em;
 }
 
 .notebookbar.ui-combobox * {
 	line-height: 22px;
+}
+.ui-combobox-content {
+	padding-inline: 5px;
 }
 
 /* Styles preview */
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -753,8 +753,6 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	-webkit-appearance: none;
 	-moz-appearance: none;
 	appearance: none;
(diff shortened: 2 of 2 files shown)


browser/css/notebookbar.css
```diff
@@ -379,12 +379,15 @@ button.ui-tab.notebookbar {
...
-	width: 50px !important;
+	width: 5em;
 }
 
 .notebookbar.ui-combobox * {
 	line-height: 22px;
 }
+.ui-combobox-content {
+	padding-left: 5px;
+}
```
pedropintosilva, line 390: Thanks we really needed padding here. I wonder where did you remove it. It seems it was not in f7491e1f62de5401529db40b8b899e5c0563badc :thinking:

pedropintosilva: Could change it to `padding-inline` so both we had padding on both sides and so languages such as `&lang=ar` (LTR) also get the fix? And maybe increase the padding to `8px`

pedropintosilva (changes requested):

browser/css/notebookbar.css
pedropintosilva: Either way it should be added on both sides. Thus my suggestion for padding-inline

meven: > Thus my suggestion for padding-inline

I have done that.

pedropintosilva (approved):

PR #8981, opened 2024-05-06
Convert Control.Mention.js to Control.Mention.ts

GulsahKose:
Change-Id: I575d9e14562b1a77f30d355024a2106d7955fe6a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/.beforeprettier
+++ b/browser/.beforeprettier
@@ -21,7 +21,6 @@
 /src/control/Control.LanguageDialog.js
 /src/control/Control.Layers.js
 /src/control/Control.LokDialog.js
-/src/control/Control.Mention.ts
 /src/control/Control.Menubar.js
 /src/control/Control.MobileBottomBar.js
 /src/control/Control.MobileSlide.js
--- a/browser/.beforeprettier
+++ b/browser/.beforeprettier
@@ -21,7 +21,7 @@
 /src/control/Control.LanguageDialog.js
 /src/control/Control.Layers.js
 /src/control/Control.LokDialog.js
-/src/control/Control.Mention.js
+/src/control/Control.Mention.ts
 /src/control/Control.Menubar.js
 /src/control/Control.MobileBottomBar.js
 /src/control/Control.MobileSlide.js
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -382,6 +382,7 @@ COOL_JS_LST =\
 	src/control/Control.NotebookbarBuilder.js \
 	src/control/Control.Layers.js \
 	src/control/Control.Sidebar.js \
+	src/control/AutoCompletePopup.ts \
 	src/control/Control.Mention.ts \
 	src/control/Control.Zotero.js \
 	src/control/Search.js \
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -382,7 +382,7 @@ COOL_JS_LST =\
 	src/control/Control.NotebookbarBuilder.js \
 	src/control/Control.Layers.js \
 	src/control/Control.Sidebar.js \
-	src/control/Control.Mention.js \
+	src/control/Control.Mention.ts \
 	src/control/Control.Zotero.js \
 	src/control/Search.js \
 	src/control/Permission.js \
--- a/browser/src/control/AutoCompletePopup.ts
+++ b/browser/src/control/AutoCompletePopup.ts
@@ -44,7 +44,11 @@ interface Entry {
 	row: string;
 }
 
-interface MentionWidget extends WidgetJSON {
+interface TextWidget extends WidgetJSON {
+	text: string;
+}
+
+interface TreeWidget extends WidgetJSON {
 	text: string;
 	singleclickactivate: boolean;
 	fireKeyEvents: boolean;
@@ -67,12 +71,11 @@ abstract class AutoCompletePopup {
 	protected map: ReturnType<typeof L.map>;
 	protected newPopupData: PopupData;
 	protected data: MessageEvent<any>;
+	protected popupId: string;
 
-	constructor(map: ReturnType<typeof L.map>) {
+	constructor(popupId: string, map: ReturnType<typeof L.map>) {
 		this.map = map;
-	}
-
-	onAdd(id: string, bAutoComplete: boolean): void {
+		this.popupId = popupId;
 		this.newPopupData = {
 			children: [
 				{
@@ -85,13 +88,16 @@ abstract class AutoCompletePopup {
 			] as Array<WidgetJSON>,
 			jsontype: 'dialog',
 			type: 'modalpopup',
-			isAutoCompletePopup: bAutoComplete,
 			cancellable: true,
 			popupParent: '_POPOVER_',
 			clickToClose: '_POPOVER_',
-			id: id,
+			id: this.popupId,
 		} as PopupData;
+
+		this.onAdd();
 	}
+
+	abstract onAdd(): void;
 
 	getCurrentCursorPosition(): Point {
 		var currPos = {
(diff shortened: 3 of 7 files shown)


browser/.beforeprettier
```diff
@@ -21,7 +21,7 @@
 /src/control/Control.LanguageDialog.js
 /src/control/Control.Layers.js
 /src/control/Control.LokDialog.js
-/src/control/Control.Mention.js
+/src/control/Control.Mention.ts
```
eszkadev, line 24: can we remove that so we will get new formatting?
then please do: `make prettier-write` to apply unified formatting

browser/src/control/Control.Mention.ts
```diff
@@ -93,25 +138,27 @@ L.Control.Mention = L.Control.extend({
 				entries.push(entry);
 			}
 
-			var data;
-			var control = {
+			var data: Data;
+			var control: Control = {
 				'id': 'mentionList',
```
eszkadev, line 154: please remove not needed `'` here in the property name:
`'id':` -> `id: `

browser/src/control/Control.Mention.ts
```diff
@@ -93,25 +138,27 @@ L.Control.Mention = L.Control.extend({
...
 				'enabled': true,
 				'singleclickactivate': false,
-				'fireKeyEvents': true
+				'fireKeyEvents': true,
+				'entries': []
 			};
 			// update the popup with list if mentionList already exist
 			if (L.DomUtil.get('mentionList')) {
 				data = {
 					'jsontype': 'dialog',
```
eszkadev, line 165: same here `'`

browser/src/control/Control.Mention.ts
```diff
@@ -121,20 +168,24 @@ L.Control.Mention = L.Control.extend({
...
 				'type': 'fixedtext',
 				'text': 'no search results found!',
 				'enabled': true,
+				'singleclickactivate': undefined,
+				'fireKeyEvents': undefined,
+				'entries': undefined
 			};
 			if (L.DomUtil.get('fixedtext')) {
 				data = {
 					'jsontype': 'dialog',
```
eszkadev, line 193: nad here `'`

browser/src/control/Control.Mention.ts
```diff
@@ -121,20 +168,24 @@ L.Control.Mention = L.Control.extend({
 			data.children[0].children[0] = control;
 			data.children[0].children[0].entries = entries;
 		} else {
-			var control = {
+			var control: Control = {
 				'id': 'fixedtext',
```
eszkadev, line 183: here too `'`

browser/src/control/Control.Mention.ts
```diff
@@ -59,23 +59,34 @@ interface PopupData {
...
+    typingMention: boolean;
+}
+
+class Mention {
+	map: ReturnType<typeof L.map>;
+	newPopupData: PopupData;
+	firstChar: string;
+	users: any;
+	itemList: Array<any>;
+	_map: ReturnType<typeof L.map>;
```
eszkadev, line 72: we don't need duplicated `map` and `_map` -> please leave only one: `map` and use it everywhere

browser/src/control/Control.Mention.ts
```diff
@@ -59,23 +59,34 @@ interface PopupData {
...
+	data: MessageEvent<any>;
+
+	onAdd (map: ReturnType<typeof L.map>) {
```
eszkadev, line 75: We need `constructor()`, I guess we want to do in it what we did in `onAdd`.
Also please look where this component was used outside this file. Probably we do something like:
`map.addControl(mention)` - with regular class we don't want that (as it no longer uses Leaflet Class parent class).
`addControl` calls onAdd

GulsahKose: I didn' touch here now, because we still getting different errors. When they are solved it will be easier to change here

GulsahKose:
@eszkadev We pushed Base class too here. But we are getting following error on runtime:

Uncaught ReferenceError: AutoCompletePopup is not defined
Uncaught TypeError: L.control.mention is not a function

How can we solve that?

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,219 @@
...
+ */
+/*
+ * Control.Mention
+ */
+
+/* global app */
+
+// TODO: duplicate from MessageRouter
+
+class Mention extends AutoCompletePopup {
```
eszkadev, line 19: it is defined under `L.Control.AutoCompletePopup`

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,224 @@
...
+		if (text.length > 1) {
+			this.itemList = this.users.filter((element: any) => {
+				// case insensitive
+				return element.username.toLowerCase().includes(text.toLowerCase());
+			});
+		} else {
+			this.itemList = this.users;
+		}
+
+		if (this.itemList.length !== 0) {
```
eszkadev, line 89: this code below can be shared too -> move to base class

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,224 @@
...
+		data.posx = framePos.x;
+		data.posy = framePos.y;
+		this.map.fire('jsdialog', {
+			data: data,
+			callback: this.callback.bind(this),
+		});
+	}
+
+	closeMentionPopup(ev: CloseMessageEvent) {
+		var closePopupData = {
```
eszkadev, line 176: can be shared -> to base class

browser/src/control/AutoCompletePopup.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+interface CloseMessageEvent extends FireEvent {
+	typingMention?: boolean;
+}
+
+abstract class AutoCompletePopup {
+	map: ReturnType<typeof L.map>;
+
+	constructor(map: ReturnType<typeof L.map>) {
+		this.map = map;
```
eszkadev, line 71: `this.onAdd(map);` can be here -> it will be used in all implementations

browser/Makefile.am
```diff
@@ -382,7 +382,8 @@ COOL_JS_LST =\
 	src/control/Control.NotebookbarBuilder.js \
 	src/control/Control.Layers.js \
 	src/control/Control.Sidebar.js \
-	src/control/Control.Mention.js \
+	src/control/Control.Mention.ts \
+	src/control/AutoCompletePopup.ts \
```
eszkadev, line 386: put base class above, so it will be defined before use

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,196 @@
...
+	map: ReturnType<typeof L.map>;
+	newPopupData: PopupData;
+	firstChar: string;
+	users: any;
+	itemList: Array<any>;
+	data: MessageEvent<any>;
+
+	constructor(map: ReturnType<typeof L.map>) {
+		super(map);
+		this.onAdd(map);
```
eszkadev, line 29: now this onAdd is not needed - it's in the base class already

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,190 @@
...
+		data.posy = framePos.y;
+		this.map.fire('jsdialog', {
+			data: data,
+			callback: this.callback.bind(this),
+		});
+	}
+
+	closeMentionPopup(ev: CloseMessageEvent) {
+		var closePopupData = super.getClosePopupData();
+		this.map.fire('jsdialog', { data: closePopupData, callback: undefined });
```
eszkadev, line 148: these 2 lines can be shared too, it can be just single `super.sendClosePopup()` which will create and send the message

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	closeMentionPopup(ev: CloseMessageEvent) {
+		var closeMentionPopupData = {
+			jsontype: 'dialog',
+			type: 'modalpopup',
+			action: 'close',
+			id: 'mentionPopup',
+		} as PopupData;
+		super.closeMentionPopup(closeMentionPopupData);
```
eszkadev, line 153: pass the `id` only to the parent

browser/src/control/Control.Mention.ts
```diff
@@ -0,0 +1,189 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * Control.Mention
+ */
+
+/* global app */
+
+// TODO: duplicate from MessageRouter
```
eszkadev, line 17: this todo was about interfaces, remove here

eszkadev:
it only fails on mobile with:
```
cy:command ✘  assert	expected **<tspan.TextPosition>** to have attribute **y** with the value **'3285'**, but the value was **'3286'**
      cy:command ✔  fail:	
                    Test failed: integration_tests/mobile/impress/apply_font_shape_spec.js / Apply font on selected shape. / Apply superscript on text shape.
```

which is known problem: https://github.com/CollaboraOnline/online/pull/8999

-> good to merge

PR #8983, opened 2024-05-07
try rm -f coolwsd instead of rm coolwsd

caolanm:
Making check in .
make[1]: Entering directory '/home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04' fatal: ref HEAD is not a symbolic ref
Cleaning up...
/bin/sh: ./coolwsd: No such file or directory
rm: cannot remove './coolwsd': No such file or directory make[1]: *** [Makefile:7155: /home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04/jails] Error 1 make[1]: *** Waiting for unfinished jobs....

Change-Id: I31748354bb55f26a8a5b8475a4daa74b3ba5d729

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -452,7 +452,7 @@ else
 # Use coolwsd to cleanup jails, if any. If it fails, we may have a broken/old coolwsd binary, remove it to rebuild.
 # A CI box may switch branches without cleaning up the binaries, if coolwsd from a broken branch is used here
 # it will fail all subsequent builds, until it's rebuilt from the new branch. So removing coolwsd after failing is needed.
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
All looks great to me. The idea that we can call one tool to find a free port, and then call another tool later passing that parameter is a big race though.

I could swear we got this right in the past; but perhaps the C++ unit tests avoided the need for that fix by being in-proc.

WSD will pick a free port if it can't get the one it wants when testing - it just increments the port number; if we can dump that to a temporary file-name and read from that file life would be much more reliable & better =)

Thanks!

PR #8986, opened 2024-05-07
Stop page scrolling

Minion3665:
Sometimes an input we are focusing on is considered out of the viewport by the browser. This is not always the case, and, even when it is, it's not trivial to move it back in.

Instead, this change detects scrolling and re-scrolls the view to 0, 0.

Doing so is not without consequence, as there is a slight jitter before we can scroll back, however this is substantially better than having the controls off-the-page.

Change-Id: Ie5003f9244ce1c0e8a183de5fdbcf3a880b13a13

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1285,6 +1285,7 @@ window.app = {
 	function handleViewportChange(event) {
 		var visualViewport = event.target;
 
+		window.scroll(0, 0);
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Idea seems to be fine, needs some testing on real devices

PR #9000, opened 2024-05-08
MessageQueue cleanup, and avoid accidental use in tests.

mmeeks:
The MessageQueue is not a generic message queue, it has much more un-anticipated functionality; so don't use it where we don't need to.

In particular unexpected re-writing and merging of messages during tests is probably not a great idea.

Change-Id: I657738307e611be18f5f83e11c055bf8a88826da

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/MessageQueue.cpp
+++ b/common/MessageQueue.cpp
@@ -26,6 +26,8 @@
 
 void TileQueue::put_impl(const Payload& value)
 {
+    StringVector tokens = StringVector::tokenize(value.data(), value.size());
+
     const std::string firstToken = COOLProtocol::getFirstToken(value);
 
     if (firstToken == "tilecombine")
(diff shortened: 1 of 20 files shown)


test/WebSocketSession.hpp
```diff
@@ -247,7 +256,7 @@ class WebSocketSession final : public WebSocketHandler
     {
         {
             std::unique_lock<std::mutex> lock(_outMutex);
-            _outQueue.put(std::vector<char>(msg.data(), msg.data() + msg.size()));
+            _outQueue.emplace_back(std::vector<char>(msg.data(), msg.data() + msg.size()));
```
caolanm, line 259: nit: emplace_back doesn't need the intermediate temp so could be:
_outQueue.emplace_back(msg.data(), msg.data() + msg.size());

common/MessageQueue.cpp
```diff
@@ -26,6 +26,8 @@
 
 void TileQueue::put_impl(const Payload& value)
 {
+    StringVector tokens = StringVector::tokenize(value.data(), value.size());
+
```
caolanm, line 33: is the firstToken = ... equivalent to tokens[0] of this tokenize?

caolanm (approved):

PR #9001, opened 2024-05-08
Do not return null character from TraceFileRecord toString

neilguertincollabora:
This is something I fixed while writing the performance tests. I am no longer using trace files for the performance tests but I think it is still good to fix this so I am submitting it separately.

Change-Id: I0458ca2cb4214cc6d235a1f1322282e7b271e3f1


mmeeks (approved):
ta.

Issue #8744, opened 2024-04-12
Firefox perf: ensure canvases on pre-loading ...


mmeeks:
In Chrome canvases are constructed and de-compressed very quickly - something we often need to do while rendering.

However in firefox - this can be more expensive.

I imagine we could 'ensure' those canvases much earlier - during pre-fetch, when the user is not expecting silky-smooth scrolling. I expect at the moment we pre-fetch around the sheet but leave the "de-compress and turn it into a canvas" logic until later.

Doing that in advance and keeping up there as we scroll - would most likely let us get ahead and be much smoother on scroll.

Should also be reasonably easy - calling 'ensureCanvas' at suitable places.

Thanks!

PR #9003, opened 2024-05-08
Private/caolan/ensure canvas on preload

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
wrt https://github.com/CollaboraOnline/online/issues/8744 it looks like we always ensureCanvas when we receive the tile data rather than only on rendering it, so maybe there isn't as much obvious gain as initially thought. But we could ensureCanvas when we are requesting tiles for preFetch (or alternatively always when using _addTiles for simplicity)?

mmeeks (approved):
Looks very sensible to me =) thanks I particularly like the idea of ensuring the canvas when we request it to make sure it's backed by the time we get deltas back etc. =)

PR #9004, opened 2024-05-08
Various mountpoint handling improvements

Ashod:
- mount: reduce failure noise
- mount: avoid hard-coding
- wsd: mount: safeRemoveDir will not error when unmount fails
- mount: remove test-mount directory


mmeeks (approved):
Looks good; be good to get rid of the frenetic error logging =)

Ashod:
> Looks good; be good to get rid of the frenetic error logging =)

Yes, reducing them, at least. The hope for having them is to give us insight when there is an unexpected failure. But, can be too noisy, evidently.

PR #9008, opened 2024-05-09
calc: don't jump to own cursor if not looking

eszkadev:
If we have cell selected on the bottom of the sheet and we scroll up to watch something else so we don't see the cursor: if other user will add row above - we will jump to our own cursor. It's because it changed the address: like A500 to A5001.

Let's don't do that. Solution could be to enter "non following" mode when we scroll outside our own cursor. So we will not jump to our cell if not needed. Then if we click or do any key press - we should cancel that mode and go back to "following own cursor mode".

This can be additionaly marked by something on the user list.


mmeeks (commented):
Looks like a promising approach - but unfinished - I guess needs you back for that :-)

eszkadev:
Do not merge, full version for master in other PR: https://github.com/CollaboraOnline/online/pull/9170

eszkadev:
old PR

PR #8730, opened 2024-04-10
mount: smarter umount handling

Ashod:
We fall back to MNT_FORCE only when
MNT_DETACH fails. Also improves when
we log the error to stderr.

Change-Id: I8b8b2a863b0339ddd489f1d0150446e6bae82a8a
Signed-off-by: Ashod Nakashian <ashod.nakashian@collabora.co.uk>

PR #9012, opened 2024-05-09
mount: better check the target jail directory exits

meven:
paranthesis made the check running havoc.

https://github.com/CollaboraOnline/online/pull/8730#issuecomment-2102569063

Change-Id: Ib68d765e88d344234c70bcf331cb1c1ac6779f5b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/tools/mount.cpp
+++ b/tools/mount.cpp
@@ -170,7 +170,7 @@ int main(int argc, char** argv)
         const char* target = argv[2];
 
         struct stat sb;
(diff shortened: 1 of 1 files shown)


meven:
https://en.cppreference.com/w/cpp/language/operator_precedence operator precedence shouldn't make this necessary, could it be a compiler optimization ?

The issue is random. I don't reproduce it anymore shuffling between branches.

Still i feel this a code improvement.

meven:
This patch does not solve any problem.
The run issue were sporadic and didn't need this to go away.

PR #9015, opened 2024-05-09
Private/mmeeks/queuecleanup2

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2682,9 +2682,7 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
         do
         {
             int realTimeout = timeoutMicroS;
-            if (_document &&
-                (_document->hasCallbacks() ||
-                 (_document->hasQueueItems() && _document->processInputEnabled())))
+            if (_document && _document->needsQuickPoll())
                 realTimeout = 0;
 
             if (poll(std::chrono::microseconds(realTimeout)) <= 0)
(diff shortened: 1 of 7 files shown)


kit/Kit.cpp
```diff
@@ -2681,7 +2682,9 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
         do
         {
             int realTimeout = timeoutMicroS;
-            if (_document && _document->hasQueueItems() && _document->processInputEnabled())
+            if (_document &&
+                (_document->hasCallbacks() ||
+                 (_document->hasQueueItems() && _document->processInputEnabled())))
```
caolanm, line 2687: Reassure me that this  (a || b && c) should not be (a || b) && c, we want this as it is written?

kit/KitQueue.hpp
```diff
@@ -42,11 +42,53 @@ class KitQueue
...
+        std::string _payload;
+
+        Callback() : _view(-1), _type(-1) { }
+        Callback(int view, int type, const std::string payload) :
+            _view(view), _type(type), _payload(payload) { }
+
+        static std::string toString(int view, int type, const std::string payload)
+        {
+            std::ostringstream str;
+            str << view << " " << lokCallbackTypeToString(type) << " "
```
caolanm, line 57: iOS build seems to suggest that #include <LibreOfficeKit/LibreOfficeKitEnums.h> is newly needed to build

kit/Kit.cpp
mmeeks: Yes - it is correct; I can split it into two happier named booleans if that helps =)

mmeeks: Split to its own function & commented it :-)

caolanm: great :-)

caolanm (approved):

PR #9016, opened 2024-05-09
preview clean up

hcvcastro:
- **browser: preview: simplify _onScroll function**
- **browser: preview: remove unused code**


browser/src/control/Control.PartsPreview.js
```diff
@@ -617,29 +617,10 @@ L.Control.PartsPreview = L.Control.extend({
 
 	_onScroll: function (e) {
 		setTimeout(L.bind(function (e) {
```
caolanm, line 607: e becomes unused after this, so there's a warning that the build catches now I think

browser/src/control/Control.PartsPreview.js
```diff
@@ -617,29 +617,10 @@ L.Control.PartsPreview = L.Control.extend({
...
-				if (img && img.parentNode && !img.fetched) {
-					var previewFrameBB = img.parentNode.getBoundingClientRect();
-					if (this._direction === 'x') {
-						if ((previewFrameBB.left >= topBound && previewFrameBB.left <= bottomBound)
-						|| (previewFrameBB.right >= topBound && previewFrameBB.right <= bottomBound)) {
-							this._map.getPreview(i, i, this.options.maxWidth, this.options.maxHeight, {autoUpdate: this.options.autoUpdate});
-						}
-					} else if ((previewFrameBB.top >= topBound && previewFrameBB.top <= bottomBound)
-						|| (previewFrameBB.bottom >= topBound && previewFrameBB.bottom <= bottomBound)) {
+				if (this._isPreviewVisible(i)) {
```
caolanm, line 609: I presume the idea is to match the logic of what we put into the queue of previews to fetch with the logic of what we use to take previews out of the queue and request them from the server?

caolanm:
ci is failing because Control.PartsPreview.js  607:31  warning  'e' is defined but never used:  no-unused-vars

browser/src/control/Control.PartsPreview.js
hcvcastro: Thanks, fixed. I do not why my local build show as error

browser/src/control/Control.PartsPreview.js
hcvcastro: well, after reading the logic, it is the same. yes

browser/src/control/Control.PartsPreview.js
```diff
@@ -615,36 +615,17 @@ L.Control.PartsPreview = L.Control.extend({
 		}
 	},
 
-	_onScroll: function (e) {
-		setTimeout(L.bind(function (e) {
-			var scrollOffset = 0;
-			if (e) {
-				var prevScrollY = this._scrollY;
-				var rectangle = e.target.getBoundingClientRect();
-				this._scrollY = this._direction === 'x' ? -rectangle.left : -rectangle.top;
```
caolanm, line 624: if we remove this._scrollY then there are no uses of _scrollY left and we should remove the remaining initialization of this._scrollY = 0 earlier in this file.

caolanm (approved):
I imagine the original code intends to request additional previews in the direction that the scroll is going so they will be considered when the queue is processed. Though I see that the queue get processed in getPreview which the scroll timer calls anyway, and that calls _processPreviewQueue which 

So I can see the logic in just requesting everything missing in the scroll timer  and let _processPreviewQueue figure out what to actually request.

Trying to figure out why it is the way it is suggests the pr of https://github.com/CollaboraOnline/online/pull/1100 is where the current approach came from.

I wonder will it make a difference though. Is it possible for the slide panel area to start off smaller than its final size, emit the initial "scroll" when its small, then resize to its final size and not request further previews to fill itself? Or some scenario like that.

I see we discard entries put into the queue because they are invisible at "skip this! we can't see it", I wonder if we should hold onto those and put them back into the end of the queue instead of throwing them away.

Though I also wonder what happens if we hit the "return" at "wait until the queue is empty" what is supposed to guarantee there will be another _processPreviewQueue called?

browser/src/control/Control.PartsPreview.js
hcvcastro: Thanks, I will remove it

hcvcastro:
> Though I also wonder what happens if we hit the "return" at "wait until the queue is empty" what is supposed to 
> guarantee there will be another _processPreviewQueue called?

good questions, I did the same on my side, but let me explain the background history.
In the beginning the previews were implemented using a third party library (I forget its name),
so we did  a lot of hacks to pre-fetch previews, then the library was removed and it remains
unused a lot of code that I have been removing.
 
The _processPreviewQueue is a new changes approach to request the previews, I think
the onScroll preview was trying to adapt to the old code approach that I see unnecessary now,
if you checks _processPreviewQueue, even if you push non visible previews, it will be removed
after execution (in batch of 3).

PR #9017, opened 2024-05-09
UiManager: default to notebookbar by default

meven:
In case of misconfiguration (missing userInterfaceMode) or "default" userInterfaceMode, which is the default, use notebookbar.

Change-Id: I00ee65edb21fcf80724bc83f947341dfdc5c8ed6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Tested it works, thanks

pedropintosilva (changes requested):
I have tested and it seems to work.  Nevertheless when debugging both
the new `(window.userInterfaceMode === 'classic' && forceCompact == null) || forceCompact === true;` and the previous ` (window.userInterfaceMode === 'notebookbar' && forceCompact === null) || forceCompact === false` I get similar results when testing with `classic` string in the xml or even removing the whole sub `mode` node from xml and xml.in


I wonder if it wouldn't be better to use notebookbar when `classic` is passed (when `window.userInterfaceMode === 'classic'` is true)... :thinking: this way we ensure that old value - that might have been forgotten in the sysadmin backed xml - is never user and instead we default to Notebookbar. If they really want that they can always change it to compact in the xml

meven:
> I wonder if it wouldn't be better to use notebookbar when `classic` is passed (when `window.userInterfaceMode === 'classic'` is true)... 🤔 this way we ensure that old value - that might have been forgotten in the sysadmin backed xml - is never user and instead we default to Notebookbar. If they really want that they can always change it to compact in the xml

I think that's a good option, fixing another notebookbar-not-by-default issue.

meven:
> > I wonder if it wouldn't be better to use notebookbar when `classic` is passed (when `window.userInterfaceMode === 'classic'` is true)... 🤔 this way we ensure that old value - that might have been forgotten in the sysadmin backed xml - is never user and instead we default to Notebookbar. If they really want that they can always change it to compact in the xml
> 
> I think that's a good option, fixing another notebookbar-not-by-default issue.

We might want to update the documentation then:
https://sdk.collaboraonline.com/docs/theming.html?highlight=compact
https://sdk.collaboraonline.com/docs/installation/Configuration.html?highlight=compact

pedropintosilva (approved):

meven:
There is a cypress test failing: `impress/apply_font_shape_spec.js         00:48       12       10        1        1      `

The assertion failing is:
```
      cy:command ✔  cGet	.leaflet-pane.leaflet-overlay-pane g.Page .TextPosition
      cy:command ✘  assert	expected **<tspan.TextPosition>** to have attribute **y** with the value **'3286'**, but the value was **'3285'**
      cy:command ✔  fail:	
                    Test failed: integration_tests/mobile/impress/apply_font_shape_spec.js / Apply font on selected shape. / Apply superscript on text shape.
                    
                    Timed out retrying after 10000ms: expected '<tspan.TextPosition>' to have attribute 'y' with the value '3286', but the value was '3285'
                    
                    /home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04_cypress_mobile/cypress_test/integration_tests/mobile/impress/apply_font_shape_spec.js:100:14
                       98 |         triggerNewSVG();
                       99 |         cy.cGet('.leaflet-pane.leaflet-overlay-pane g.Page .TextPosition')
                    > 100 |             .should('have.attr', 'y', '3286');
                          |              ^
                      101 |         cy.cGet('.leaflet-pane.leaflet-overlay-pane g.Page .TextParagraph .TextPosition tspan')
                      102 |             .should('have.attr', 'font-size', ...
          cy:log ✱  Finishing test: integration_tests/mobile/impress/apply_font_shape_spec.js / Apply font on selected shape. / Apply superscript on text shape.
```
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_mobile/987/consoleFull

It could be that this MR, changed the ui mode used in tests.
Looking into it.

pedropintosilva:
I think there is probably another problem in the test  cypress_test/integration_tests/mobile/impress/apply_font_shape_spec.js that is unrelated to this. I think there is a fix already: cypress: quick fix for rounding error #9167

eszkadev:
please rebase now, it should work

PR #9025, opened 2024-05-10
Add performance test framework

neilguertincollabora:
Files changed:
- PerfTest.hpp/cpp: Class for performance tests. Also contains derived classes CyclePerfTest and MessagePerfTest.
- PerfTestSocketHandler.hpp/cpp: Class for connecting and sending messages to coolwsd.
- COOLPerfTest.cpp: Main function. This will definitely change. If you have feedback it might help with future designs, but it's not necessary to review because it's basically temporary code anyways.
- SamplePerfTest.cpp: Example of what a performance test might look like. Temporary code.
- Makefile.am, .gitignore: Changes to support coolperftest executable

Here's the design:
Individual performance tests (like SamplePerfTest) will derive from PerfTest, CyclePerfTest, or MessagePerfTest. They have to implement the runTest method. Within it, they can use some functions to control the test. Like:
- startMeasurement, stopMeasurement
- connect, loadDocument, connectAndLoad, disconnect
- waitForMessage, waitForIdle, sleep
- sendMessage

The PerfTestSocketHandler mostly runs in a separate thread using a SocketPoll (_poll.startThread) and getPollEvents and handleMessage. Data is passed to the PerfTest thread using MessageQueues _incomingMessages and _outgoingMessages. MessagePerfTest needs to count messages so MessagePerfTestSocketHandler shares message count data using raw pointers.

There is also a minor change needed to core, which creates a .uno:WaitForIdle message, which the server responds to when it's otherwise idle (using a task that runs at the lowest priority). In init.cxx doc_postUnoCommand:
```
    else if (gImpl && aCommand == ".uno:WaitForIdle")
    {
        IdleTask::waitUntilIdleDispatched();
        tools::JsonWriter aJson;
        aJson.put("commandName", pCommand);
        aJson.put("success", true);
        pDocument->mpCallbackFlushHandlers[nView]->queue(LOK_CALLBACK_UNO_COMMAND_RESULT, aJson.finishAndGetAsOString());
        return;
    }
```

coolwsd must be started separately, and if using CyclePerfTest, coolwsd-inproc must be used instead, using the extra option `--pidfile=coolwsd.pid`. This is a bit annoying since it has to be restarted for every test, so if you know any way to improve that I'd love to hear it.

Also, to use perf, you need to run `sudo sysctl kernel.perf_event_paranoid=2` (if on Debian/Ubuntu at least). Also you need to install FlameGraph and add it to your path.

Then `make coolperftest && ./coolperftest`. The server and test arguments are hardcoded for now.

The result for now is just some output. This is likely to change when I make some system to run and process test results. The raw data from perf also gets saved.
```
{
    cycles: 3132925858,
    dataFile: sample100.data,
    flamegraph: sample100.svg,
    name: sample100,
    time: 2.532000,
}
```

Please give me any feedback you have, from typos to C++ tips to design alternatives. Thanks!


vmiklos (commented):
Failed to provide my input in the right textarea...

perftest/PerfTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+    // Use paste to format list of numbers for bc
+    // Use bc to sum numbers. bc, as an arbitrary precision calculator, is
+    // necessary because cycle counts can be >2^31 (too big for awk's sum)
+    std::string command = "perf report -i "+_fileName+".data --no-inline | grep \"# Event count (approx.): \" | awk '{print $5}' | paste -s -d+ | bc";
+    std::string cycles = getStringPopen(command);
+    addResult("cycles",cycles);
+
+    // Create flamegraph
+    std::string flamegraphCommand = "perf script -i "+_fileName+".data --no-inline | stackcollapse-perf.pl | flamegraph.pl > "+_fileName+".svg";
+    int systemResult = system(flamegraphCommand.c_str());
```
vmiklos, line 215: Just to comment on point 3): much better, now I think only this one remains for system(). I think you could replace this one with a combination of pipe(), fork(), dup2(), close() and execvp() or something similar. Could you please look at that? bash is terrible language, let's not introduce it here (think of escaping special filenames, etc). Thanks. :-)

perftest/COOLPerfTest.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <Poco/URI.h>
+
+#include <net/Ssl.hpp>
+#if ENABLE_SSL
+#  include <SslSocket.hpp>
+#endif
+#include <Log.hpp>
+
+#include <perftest/PerfTest.hpp>
+#include <perftest/SamplePerfTest.hpp>
```
vmiklos, line 28: "perftest/COOLPerfTest.cpp currently includes perftest/SamplePerfTest.hpp" still needs to be resolved, I guess.

perftest/COOLPerfTest.cpp
```diff
@@ -0,0 +1,99 @@
...
+class COOLPerfTest : public Poco::Util::Application
+{
+public:
+    COOLPerfTest() {}
+protected:
+    int  main(const std::vector<std::string>& args) override;
+};
+
+// coverity[root_function] : don't warn about uncaught exceptions
+int COOLPerfTest::main(const std::vector<std::string>& args)
```
vmiklos, line 40: Just `/*args*/` here or actually take that single argument?

perftest/PerfTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+    waitpid(child_pid, nullptr, 0);
+    addResult("dataFile",_fileName+".data");
+
+    // Get result
+    // Use perf report to get summary statistics from the data file
+    // Use grep and awk to get the cpu cycle counts from the report output
+    // Use paste to format list of numbers for bc
+    // Use bc to sum numbers. bc, as an arbitrary precision calculator, is
+    // necessary because cycle counts can be >2^31 (too big for awk's sum)
+    std::string command = "perf report -i "+_fileName+".data --no-inline | grep \"# Event count (approx.): \" | awk '{print $5}' | paste -s -d+ | bc";
```
vmiklos, line 209: Ah, this is not system() but needs similar fixing to avoid start programming in bash-in-C++.

perftest/COOLPerfTest.cpp
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+        std::cerr << "Did not finish measurement";
+        if (!perfTest.isStarted()) {
+            std::cerr << " (Never started)";
+        }
+        std::cerr << std::endl;
+        return EX_SOFTWARE;
+    }
+
+    perfTest.printResults();
```
vmiklos, line 91: See above, this needs to at least go to some CSV file so we can plot charts later, I think.

perftest/PerfTest.hpp
```diff
@@ -0,0 +1,94 @@
...
+public:
+    CyclePerfTest(const std::string &name, const std::string &server);
+    void startMeasurement() override;
+    void stopMeasurement() override;
+private:
+    pid_t getCoolwsdPid();
+    std::string getStringPopen(const std::string &command);
+};
+
+class MessagePerfTest : public PerfTest
```
vmiklos, line 80: Can we avoid MessagePerfTest for the scope of this PR?

perftest/PerfTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+        LOG_DBG("PerfTest waitForMessage Found message: " + substring + " after " + std::to_string(seconds) + "s");
+    } else {
+        abort("Timed out waiting for message \"" + substring + "\" after " + std::to_string(seconds) + "s");
+    }
+}
+
+void PerfTest::waitForIdle(size_t timeout)
+{
+    LOG_TRC("PerfTest waitForIdle");
+    sendMessage("uno .uno:WaitForIdle");
```
vmiklos, line 146: Would it work to not send a specific comment, just make sure nothing arrives on the socket for e.g. 1s?

vmiklos (commented):
Glad to see CI now passes for your changes. I tried to break up my longer comment from Fri into smaller, hopefully actionable comments. Please take a look. Thanks!

perftest/COOLPerfTest.cpp
neilguertincollabora: Got dlopen working, like in the unit tests. This was a pain to get working, but before I got it I also tried to extend the header includes and it got really messy.

I had come up with these requirements earlier, and dlopen meets them much better than header includes would have.
- Ability to run one performance test
- Ability to run all performance tests
- Adding a test should only require touching one other file, and only a few lines there
- Adding a test should require little or no boilerplate
- Test files should be able to go in nested directories

perftest/PerfTest.cpp
vmiklos: One more tip here: you probably only need to execute 'perf report' here. If you use code from common/Util.hpp and StringVector::tokenize(), then you can avoid the external grep / awk / etc calls.

vmiklos (commented):
Thanks, I like the direction this PR is going. I turned my initial comment at https://github.com/CollaboraOnline/online/pull/9025#issuecomment-2104742251 into a checklist, I hope that helps tracking what's still to be done. I would not worry about CI too much: it passed once, but points 2 -> 5 still need fixing, I would say.

Perhaps the most important is the data collection part: once you have the data in (say) a CSV, you can tweak how that data is collected for new runs / how it's presented.

Again, it's fine to keep it simple: a static HTML linking to the CSV (that we can open in Calc to chart it) is an acceptable start. Thanks!

Makefile.am
```diff
@@ -721,6 +732,82 @@ check-recursive: eslint
...
+perftest_pid_file = $(perftest_workdir)/coolwsd.pid
+perftest_coolwsd_log = $(perftest_workdir)/coolwsd.log
+perftest_coolwsd_output = $(perftest_workdir)/coolwsd_output.log
+perftest_ldflags = -module -rpath /dummy
+all_perftests_la =
+
+all_perftests_la += perftest/sample/SamplePerfTest.la
+perftest_sample_SamplePerfTest_la_SOURCES=perftest/sample/SamplePerfTest.cpp
+perftest_sample_SamplePerfTest_la_LDFLAGS=$(perftest_ldflags)
+all_perftests_la += perftest/example/ExamplePerfTest.la
```
vmiklos, line 745: Hmm, now we have both sample and example perf tests. :-) I would suggest to came up with a real workload, just a few, 2-4 ones and use them to demonstrate that the framework is working. As mentioned earlier, a minimal feature set is fine, but you usually should not add code that you'll remove in the next commit.

In general, please try to address the unchecked points from https://github.com/CollaboraOnline/online/pull/9025#issuecomment-2104742251, I would like to approve this rather sooner than later, but not at the cost of just forgetting about all review feedback.

As mentioned offline yesterday, I can live with the current "call out to bash" code for now, if they are clearly annotated with some `// TODO(neilguertin) don't call out to bash` markup. Thanks!

perftest/tests/sample/SamplePerfTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <perftest/PerfTest.hpp>
+
+class SamplePerfTest : public MessagePerfTest
```
vmiklos, line 14: Is it useful to derive your actual perf test suite from MessagePerfTest? It seemed to me that we're primarily interested in the number of CPU cycles the test uses, so CyclePerfTest would enough and you could drop MessagePerfTest from this PR.

For points 2) -> 6) from the checklist, I don't see a change with this update. Thanks.

perftest/tests/load/LoadCalcLarge.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+public:
+    LoadCalcLarge(const std::string resultsDir, const std::string &server) :
+        CombinedPerfTest(__func__, resultsDir, server)
+    {
+    }
+
+    void runTest() {
+        startMeasurement();
+        connectAndLoad("perftest/data/new-hello-world.fodt");
```
vmiklos, line 24: Some inconsistency here: the test is called "large" but the test doc is called "new", I assume it would be good to go with one term everywhere.

vmiklos (commented):
Great, now I see no "sample" in the Makefiles. :-)

This PR is growing: would it make sense to break out a separate PR from this, just to add your test documents? That could go in fast, is not controversial in any way and would reduce the size of this PR.

Anyhow, it seems to me this is slowly going in the right direction. Thanks.

perftest/PerfTest.hpp
```diff
@@ -0,0 +1,136 @@
...
+// Measure both Cycles and Messages
+// Must reimplement one class to avoid diamond inheritance
+// Reimplement MessagePerfTest because it is simpler
+// Fortunately we do not have to reimplement MessagePerfTestSocketHandler
+class CombinedPerfTest : public CyclePerfTest
+{
+private:
+    // Shared with MessagePerfTestSocketHandler
+    std::atomic<bool> _measuring = false;
+    std::atomic<unsigned int> _messageCount = 0;
```
mmeeks, line 112: Seems deeply inferior to tools/Reply.hpp which we could have used - which has per message statistics, latency histograms and far more ... hmm ? why did we duplicate that?

perftest/PerfTestSocketHandler.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+    LOG_TRC("PerfTestSocketHandler getPollEvents");
+    if (_connecting) {
+        LOG_TRC("PerfTestSocketHandler getPollEvents (Waiting for outbound connection)");
+        return POLLOUT;
+    }
+
+    while (!_outgoingMessages.empty()) {
+        std::string message = _outgoingMessages.front();
+        _outgoingMessages.pop();
```
mmeeks, line 107: How do we avoid pushing all the session's events down the socket in one whack ? That is not going to simulate real use; it seems you removed the timestamp handling from the trace replay code here - why ?

perftest/PerfTestSocketHandler.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+void MessagePerfTestSocketHandler::handleMessage(const std::vector<char> &data)
+{
+    PerfTestSocketHandler::handleMessage(data);
+    if (*_measuring) {
+        const std::string firstLine = COOLProtocol::getFirstLine(data.data(), data.size());
+        StringVector tokens = StringVector::tokenize(firstLine);
+        if (tokens.equals(0, "tile:") || tokens.equals(0,"invalidatetiles:")) {
+            (*_messageCountTile)++;
```
mmeeks, line 175: That is really odd C++ compared with how the Replay.hpp manages a Stats class; this is quite 'different'. A more normal C++ idiom would be to use a reference with '&' and get rid of the (* ) - around the member.

mmeeks (commented):
Had a quick read since the profile results were quite unexpected; we should talk it seems.

mmeeks (changes requested):
Lets talk about this before merging it.

PR #9030, opened 2024-05-10
Add vertical ruler for writer

Darshan-upadhyay1110:
- add new LOK callback for vertical & horizontal ruler update

Change-Id: I3c0e26f22072de4612e128d58ac41b629be82807

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Darshan-upadhyay1110:
CORE Patch: https://gerrit.libreoffice.org/c/core/+/167481

gokaysatir (changes requested):
Okay, for vertical ruler: We only have header and footer margins that are draggable as far as i can see.

So i guess we shouldn't have code for break points inside VRuler file?

And also it seems to me that vertical ruler goes off screen as i play with the header footer settings :) It feels like something related to left / right margins are forgotten. This is a guess. Can you check that vertical ruler doesn't go off screen and is readable.

I'd check the code from scratch and make some polishing (if required).

Thanks @Darshan-upadhyay1110 :) We are so close.

Darshan-upadhyay1110:
> So i guess we shouldn't have code for break points inside VRuler file?

Yes we do not need break point for vertical ruler

> And also it seems to me that vertical ruler goes off screen as i play with the header footer settings :) It feels like something related to left / right margins are forgotten. This is a guess. Can you check that vertical ruler doesn't go off screen and is readable.

Can you please give me hint, how did you change the header footer settings ? What action you performed that made ruler goes out of screen ?

gokaysatir:
> Can you please give me hint, how did you change the header footer settings ? What action you performed that made ruler goes out of screen ?

Of course, i used page settings and set the header & footer margins.

browser/src/control/VRuler.js
```diff
@@ -0,0 +1,396 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * Ruler Handler
+ */
+
+/* global L _ */
+L.Control.VRuler = L.Control.extend({
```
eszkadev, line 16: In the new code please avoid using Leaflet's code (L.Control / Class etc...).
We should use native `class` and also TypeScript.
Then we will be able to convert horizontal ruler and make some base class for both to share code.

browser/src/control/VRuler.ts
```diff
@@ -0,0 +1,462 @@
...
+			tabs: [],
+			unit: null,
+			DraggableConvertRatio: null,
+			timer: null,
+			showruler: true,
+			position:'topleft',
+			disableMarker: false
+		};
+
+		constructor(map: ReturnType<typeof L.map>, options: Options) {
```
eszkadev, line 87: did you run: `make prettier-write` to improve formatting?

browser/.beforeprettier
```diff
@@ -75,6 +75,7 @@
 /src/control/Parts.js
 /src/control/Permission.js
 /src/control/Ruler.js
+/src/control/VRuler.ts
```
eszkadev, line 78: please not add new files to ignore list, we want new formatting to be applied

browser/src/control/VRuler.ts
Darshan-upadhyay1110: no, i did not run that. i just do select all then format code by right click in VS Code

browser/.beforeprettier
Darshan-upadhyay1110: okay

Darshan-upadhyay1110: i see now code is formatted, I will push that change

eszkadev (approved):
Looks good, lets get this in and then do a followup :)

eszkadev:
One UX comment:
- dragging blue arrow works
- dragging grey line shows tooltip but doesn't move margin (does selection instead)

I would expect to have both of them draggable and working.

eszkadev:
@gokaysatir you requested changes, please re-check

gokaysatir (approved):

PR #9031, opened 2024-05-10
browser: preview: remove remaining unused code

hcvcastro:
Change-Id: I36efb34ea813fc110fdad3b1165c37f7f2b9d683


browser/src/control/Control.PartsPreview.js
```diff
@@ -60,6 +60,7 @@ L.Control.PartsPreview = L.Control.extend({
 		map.on('deletepage', this._deletePreview, this);
 		map.on('scrolllimits', this._invalidateParts, this);
 		map.on('scrolltopart', this._scrollToPart, this);
+		map.on('beforequestpreview', this._beforeRequestPreview, this);
```
caolanm, line 63: can we take the opportunity to use 'beforerequestpreview' instead of 'beforequestpreview' to avoid confusion later

hcvcastro: Ups, typo, thanks

browser/src/control/Control.PartsPreview.js
```diff
@@ -559,6 +561,11 @@ L.Control.PartsPreview = L.Control.extend({
 		this._width = window.innerWidth;
 	},
 
+	_beforeRequestPreview: function (e) {
+		if (e.part !== undefined && e.part >= 0 && e.part < this._previewTiles.length)
+			this._previewTiles[e.part].src = document.querySelector('meta[name="previewImg"]').content;
```
caolanm, line 566: With this patch, what I am seeing in the presentation30.odp test case we have used before is that there is a "flicker" in the top preview when loading a document, or flipping between dark/light modes.

[Screencast from 2024-05-13 20-18-29.webm](https://github.com/CollaboraOnline/online/assets/833656/560cad50-96a7-4549-b18e-c67ee8cc367f)

I wonder if we really want to set this placeholder on "beforerequestpreview", do we need it? Or if we do, do we always need it if we already have some preview content which is probably the same or similar to what the new content will be?

hcvcastro: It is a visual indicator when client side requests the preview, the objective it is to know which previews are requested or not, in case a preview is not requested, it will show a smile placeholder. And in the video it looks like it is requesting the same preview 2 times. I will look at it.

hcvcastro: The smile placeholder will show only on initial document loading if not request to server side.

caolanm (approved):
ok, I think it looks good, lets get on with it

PR #9032, opened 2024-05-10
Private/mmeeks/logfixup

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -50,6 +50,9 @@ constexpr int LOG_FILE_FD = STDERR_FILENO;
 
 } // namespace
 
+/// Which log areas should be disabled
+bool AreasDisabled[Log::AreaMax] = { false, };
+
 /// Wrapper to expose protected 'log' and genericise
 class GenericLogger : public Poco::Logger
 {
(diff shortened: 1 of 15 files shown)


mmeeks:
Thoughts appreciated; its a very substantial reduction in noise by default; then again - more to do there I think; and possibly we need to add more helpful text message printing for send/recv - lets see.

common/Log.cpp
```diff
@@ -665,14 +668,23 @@ namespace Log
...
             static_cast<Poco::Message::Priority>(logger().getLevel()));
-        return logLevel >= static_cast<int>(l);
+
+        if (logLevel < static_cast<int>(l))
+            return false;
+
+        bool disabled = AreasDisabled[static_cast<size_t>(a)];
+
+        // Areas shouldn't disable warnings & errors
+        assert(!disabled || logLevel > static_cast<int>(Level::WRN));
```
caolanm, line 685: seeing as this will be configurable from coolwsd.xml do we want end users with an NDEBUG build to be able to disable areas while using a warning/error debuglevel?

caolanm (approved):
Though I have no particular strong feelings on logging

PR #9033, opened 2024-05-10
Make headers move on pinch-zoom

Minion3665:
While zooming, things like the cursor move along with the zoom. The headers instead stayed fixed until the zoom is complete. This is inconsistent with panning, and worsens your experience when zooming

Additionally, I refactored the code in the HeaderInfo update() function a little since as so much of it needed to change anyway

Change-Id: I100de8f1fd62b831c285c9acab456260dbb77ead

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665:
-2, do not merge. It seems like this might be causing some issues with stopping/starting zooms on iOS

gokaysatir (approved):
Thanks :)

PR #9034, opened 2024-05-10
cool.html: append WOPISrc parameter to all JS, CSS & SVG loads

Rash419:
Change-Id: I1dd3902403a93914a8ad03b24bed5524a59e5aca

* Target version: master

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -129,14 +129,14 @@ m4_ifelse(EMSCRIPTENAPP,[true],[
 
 m4_ifelse(BUNDLE,[],
   <!-- Using individual CSS files -->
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Looks great to me - thanks Rash =) Did you check we clobber every URL by checking in the browser Network tab as things load ? =)

Rash419:
> Looks great to me - thanks Rash =) Did you check we clobber every URL by checking in the browser Network tab as things load ? =)

@mmeeks I did. Do icons url also needs to be appended with WOPISrc  ?

mmeeks:
Re: icons - yes, all URLs should have that.

Rash419:
> Re: icons - yes, all URLs should have that.

OK, done

mmeeks (changes requested):
So - I think we should abandon this for now - I suspect this will break browser caching; and that we should come up with a better approach & design here.

mmeeks:
Also a duplicate pull request it seems; lets abandon this one for now.

PR #9039, opened 2024-05-13
add auto fill popup: Copy cells and Fill series

bayramcicek:
AutoFill popup has two options:
- Copy cells
- Fill series

requires: https://gerrit.libreoffice.org/c/core/+/167726
5ab9b63e6624947a6076a01f9bab0d7a925e0e40

Change-Id: I08ad88afad1bb6bd2d746a9a8964f1b4446ae7c7

* Resolves: # 
* Target version: master 

### Summary
- This enhancement is for co-23.05

- Added UI
- close AutoFill popup when
  - sheet tab is clicked
  - new sheet is inserted
  - sheet tab is switched using Ctrl+Alt+PageUp/PageDown keyboard shortcuts
- Added functionalities for "Copy cells" and "Fill series".
- close AutoFill popup on some contextmenu entries from sheet tabs
  - insert sheet before/after
  - delete / hide / show sheet

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -602,7 +602,8 @@ L.Control.JSDialog = L.Control.extend({
 			var dialog = this.dialogs[autoFilterDialogId];
 
 			// Check if the current dialog has the isAutofilter property set to true
-			if (dialog.isAutofilter) {
+			// and if it is not an autoFillPopup
+			if (dialog.isAutofilter && (autoFilterDialogId !== 'autoFillPopup')) {
 				// Call this.close(key, true) for the current dialog
 				this.close(autoFilterDialogId, true);
 			}
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -600,10 +600,17 @@ L.Control.JSDialog = L.Control.extend({
 		for (var i = 0; i < dialogKeys.length; i++) {
 			var autoFilterDialogId = dialogKeys[i];
 			var dialog = this.dialogs[autoFilterDialogId];
+			var isAutoFillFromOnMouseUp = this.map._docLayer.isAutoFillFromOnMouseUp;
 
 			// Check if the current dialog has the isAutofilter property set to true
-			// and if it is not an autoFillPopup
-			if (dialog.isAutofilter && (autoFilterDialogId !== 'autoFillPopup')) {
+			if (dialog.isAutofilter) {
+
+				// don't hide AutoFill popup if it comes from onMouseUp event (see AutoFillMarkerSection.ts)
+				if (isAutoFillFromOnMouseUp && (autoFilterDialogId === 'autoFillPopup')) {
+					this.map._docLayer.isAutoFillFromOnMouseUp = false;
+					continue;
+				}
+
 				// Call this.close(key, true) for the current dialog
 				this.close(autoFilterDialogId, true);
 			}
--- a/browser/src/control/Control.SheetsBar.js
+++ b/browser/src/control/Control.SheetsBar.js
@@ -69,9 +69,7 @@ L.Control.SheetsBar = L.Control.extend({
 		}
 
 		if (id === 'insertsheet') {
-			if (this.map._docLayer.isAutoFillPopupOpen)
-				this._map.fire('closeautofillpopup');
-
+			this.map._docLayer.isAutoFillFromOnMouseUp = false;
 			var nPos = $('#spreadsheet-tab-scroll')[0].childElementCount;
 			this.map.insertPage(nPos);
 			this.map.insertPage.scrollToEnd = true;
--- a/browser/src/control/Control.SheetsBar.js
+++ b/browser/src/control/Control.SheetsBar.js
@@ -69,6 +69,9 @@ L.Control.SheetsBar = L.Control.extend({
 		}
 
 		if (id === 'insertsheet') {
+			if (this.map._docLayer.isAutoFillPopupOpen)
+				this._map.fire('closeautofillpopup');
+
 			var nPos = $('#spreadsheet-tab-scroll')[0].childElementCount;
 			this.map.insertPage(nPos);
 			this.map.insertPage.scrollToEnd = true;
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -334,9 +334,6 @@ L.Control.Tabs = L.Control.extend({
 		if (part !== this._map._docLayer._selectedPart) {
 			this._setPartIndex(part);
 		}
-
-		if (this._map._docLayer.isAutoFillPopupOpen)
-			this._map.fire('closeautofillpopup');
 	},
 
 	//selected sheet is moved to new index
@@ -429,10 +426,14 @@ L.Control.Tabs = L.Control.extend({
 
 	_showSheet: function() {
 		this._map.showPage();
+		// required for AutoFill popup on empty cells
+		this._map._docLayer.isAutoFillFromOnMouseUp = false;
 	},
 
 	_hideSheet: function() {
 		this._map.hidePage(this._tabForContextMenu);
+		// hiding the active sheet should close the AutoFill popup
+		this._map._docLayer.isAutoFillFromOnMouseUp = false;
 	},
 
 	_handleDragStart: function(e) {
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -334,6 +334,9 @@ L.Control.Tabs = L.Control.extend({
 		if (part !== this._map._docLayer._selectedPart) {
 			this._setPartIndex(part);
 		}
+
+		if (this._map._docLayer.isAutoFillPopupOpen)
+			this._map.fire('closeautofillpopup');
 	},
 
 	//selected sheet is moved to new index
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -376,6 +376,7 @@ COOL_JS_LST =\
 	src/control/Control.Sidebar.js \
 	src/control/AutoCompletePopup.ts \
 	src/control/Control.Mention.ts \
+	src/control/Control.AutoFillOptions.ts \
 	src/control/Control.Zotero.js \
 	src/control/Search.js \
 	src/control/Permission.js \
--- a/browser/src/control/Control.AutoFillOptions.ts
+++ b/browser/src/control/Control.AutoFillOptions.ts
@@ -21,7 +21,6 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
 	users: any;
 	itemList: Array<any>;
 	data: MessageEvent<any>;
-	autoFillPopupPoint: Point;
 
 	constructor(map: ReturnType<typeof L.map>) {
 		super('autoFillPopup', map);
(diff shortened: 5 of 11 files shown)


bayramcicek:
I'm getting the following issue:

https://github.com/CollaboraOnline/online/assets/24255179/a0369470-2cf8-4295-9242-df670f7368bb

With empty cells `AutoFill` popup appears - whick is ok. However, if there is a data in the cell; popup will close itself (I see that when we fill the other empty cells with data, it sends the `close` event type to the popup, so it closes itself immediately after opening - you may not realize the opening of the popup but I can see that from the logs.)

bayramcicek:
Current behavior: 

https://github.com/CollaboraOnline/online/assets/24255179/79ea4212-080e-4f50-96e8-139aef82478e

caolanm:
Worth trying a sequence like 1,3,9 and compare how it used to work with what might happen now.

I think that for "copy" we want to get repeated 1,3,9
i.e. 1,3,9:1,3,9:1,3,9

and for "fill series" we want 1,3.9, extended to
1,3,9:2,4,10:3,5,11

like what the original cases +- holding Ctrl does, is that what we expect?

also, I think that when the new buttons are clicked there is already data entered into the cells from the existing autofill triggered from releasing the mouse so the fill series operates on this new content and not the original content.

Not sure if its the best solution, but dispatching a "this.map.sendUnoCommand('.uno:Undo');" before would make the next fill-series/auto-fill operate on the original data.

I'm not sure there are combinations to the uno "FillSeries" that can exactly match the two behaviours of "AutoFill" with/without ctrl pressed?

gokaysatir (changes requested):
Hello, thanks for the improvements :) Can you check @caolanm's comment and the behaviour of the new menu, is there anything we can improve more?

bayramcicek:
> Hello, thanks for the improvements :) Can you check @caolanm's comment and the behaviour of the new menu, is there anything we can improve more?

Thanks for the review!

Now I see that when multiple cells are selected & `autoFill` used by dragging, all of them are copied or filled by a given pattern (as in the video below)
 
However my patch does copy only the first cell, and if `Fill series` used, increment starts from the first cell (by 1 default ) (we don't remember the selected cells & pattern at this point)

working on it...

https://github.com/CollaboraOnline/online/assets/24255179/d0599461-98c8-45dd-8dbf-8dd35ac50631

bayramcicek:
> Worth trying a sequence like 1,3,9 and compare how it used to work with what might happen now.
> 
> I think that for "copy" we want to get repeated 1,3,9 i.e. 1,3,9:1,3,9:1,3,9
> 
> and for "fill series" we want 1,3.9, extended to 1,3,9:2,4,10:3,5,11
> 
> like what the original cases +- holding Ctrl does, is that what we expect?
> 
> also, I think that when the new buttons are clicked there is already data entered into the cells from the existing autofill triggered from releasing the mouse so the fill series operates on this new content and not the original content.
> 
> Not sure if its the best solution, but dispatching a "this.map.sendUnoCommand('.uno:Undo');" before would make the next fill-series/auto-fill operate on the original data.
> 
> I'm not sure there are combinations to the uno "FillSeries" that can exactly match the two behaviours of "AutoFill" with/without ctrl pressed?

Thanks for review & suggestions @caolanm, @gokaysatir .

I've uploaded a patchset here that requires core patch https://gerrit.libreoffice.org/c/core/+/167726 .

I see that it works as expected now (exp: for complicated series like 1:3:900):
Ready for review.

https://github.com/CollaboraOnline/online/assets/24255179/914d60c7-34c4-4f09-8f56-70c2961808df

Tex2002ans:
> Worth trying a sequence like 1,3,9 and compare how it used to work with what might happen now.
> 
> I think that for "copy" we want to get repeated 1,3,9 i.e. 1,3,9:1,3,9:1,3,9
> 
> and for "fill series" we want 1,3.9, extended to 1,3,9:2,4,10:3,5,11

Yes... Fill Series is sometimes an odd one, especially when multi-selecting.

@bayramcicek I wrote a few posts about different cases + a breakdown of how it works a year ago:

- [/r/LibreOffice: "Is there any way to make Fill Down increment numbers the same way auto-fill handle does?"](https://old.reddit.com/r/libreoffice/comments/ydz7fg/is_there_any_way_to_make_fill_down_increment/itygvp5/)
- [/r/LibreOffice: "How does the little black square in LibreOffice Calc "think"?"](https://old.reddit.com/r/libreoffice/comments/132o6zy/how_does_the_little_black_square_in_libreoffice/ji8mvgc/)

The video that ultimately "made it click" for me was this fantastic one I watched *years* ago:

- ["You Suck at Excel" with Joel Spolsky](https://www.youtube.com/watch?v=0nbkaYsR94c)
    -  (And that same user helpfully recognized and linked me to!)

bayramcicek:
> > Worth trying a sequence like 1,3,9 and compare how it used to work with what might happen now.
> > I think that for "copy" we want to get repeated 1,3,9 i.e. 1,3,9:1,3,9:1,3,9
> > and for "fill series" we want 1,3.9, extended to 1,3,9:2,4,10:3,5,11
> 
> Yes... Fill Series is sometimes an odd one, especially when multi-selecting.
> 
> @bayramcicek A year ago, I wrote a few posts about different cases + a breakdown of how it works:
> 
>     * [/r/LibreOffice: "Is there any way to make Fill Down increment numbers the same way auto-fill handle does?"](https://old.reddit.com/r/libreoffice/comments/ydz7fg/is_there_any_way_to_make_fill_down_increment/itygvp5/)
> 
>     * [/r/LibreOffice: "How does the little black square in LibreOffice Calc "think"?"](https://old.reddit.com/r/libreoffice/comments/132o6zy/how_does_the_little_black_square_in_libreoffice/ji8mvgc/)
> 
> 
> The video that ultimately "made it click" for me was this fantastic one I watched _years_ ago:
> 
>     * ["You Suck at Excel" with Joel Spolsky](https://www.youtube.com/watch?v=JxBg4sMusIg)
>       
>       * (And that same user helpfully recognized and linked me to it!)

Thank you @Tex2002ans for the useful links & info.

bayramcicek:
### Update
 We may add this feature only for `online master`. This patch is for `co-23.05`. -If everything seems ok- let's **not merge** this and wait until we decide on which branch we need this feature.

browser/src/control/Control.SheetsBar.js
```diff
@@ -69,6 +69,9 @@ L.Control.SheetsBar = L.Control.extend({
 		}
 
 		if (id === 'insertsheet') {
+			if (this.map._docLayer.isAutoFillPopupOpen)
```
gokaysatir, line 72: Do we need this if we close the popup on tab change? Also, i can click on the sheet button and add a new sheet. Popup doesn't close in that case. But it closes if we close the popup on tab change.

browser/src/control/Control.SheetsBar.js
```diff
@@ -69,6 +69,9 @@ L.Control.SheetsBar = L.Control.extend({
...
+				this._map.fire('closeautofillpopup');
```
gokaysatir, line 73: Do we need to fire an event here? Maybe we can use objectInstance.openAutoFillPopup()?

Also, 'openautofillpopup' event does the same thing it seems. And i couldn't see its usage.

browser/src/control/Control.Tabs.js
```diff
@@ -334,6 +334,9 @@ L.Control.Tabs = L.Control.extend({
 		if (part !== this._map._docLayer._selectedPart) {
 			this._setPartIndex(part);
 		}
+
+		if (this._map._docLayer.isAutoFillPopupOpen)
```
gokaysatir, line 338: Maybe we won't need this if we close the popup on tab change.

browser/src/control/Control.JSDialog.js
```diff
@@ -602,7 +602,8 @@ L.Control.JSDialog = L.Control.extend({
 			var dialog = this.dialogs[autoFilterDialogId];
 
 			// Check if the current dialog has the isAutofilter property set to true
-			if (dialog.isAutofilter) {
+			// and if it is not an autoFillPopup
+			if (dialog.isAutofilter && (autoFilterDialogId !== 'autoFillPopup')) {
```
gokaysatir, line 606: Do we need to keep it open here? Maybe we don't need this change? Can you comment on this.

browser/src/control/Control.SheetsBar.js
bayramcicek: I think you mean ` objectInstance.closeAutoFillPopup()` ?.  I'd like to close `AutoFill` popup here if we already have it open. Could you please provide further explanation on that :) Thanks.

browser/src/control/Control.Tabs.js
bayramcicek: (added a comment on `tab change` situation in prior comments)

gokaysatir (commented):
I am getting below error when i use "Arrow Down" and then "TAB" keys:

`Widget.TreeView.js:584 Uncaught TypeError: Cannot read properties of undefined (reading 'parentNode')`

gokaysatir (approved):

gokaysatir (approved):
Why don't we set "isAutoFillFromOnMouseUp" to false on MouseDown? Would we still need to set it in other places like "hideSheet" etc if we did set it on MouseDown?

Also "isAutoFillFromOnMouseUp" property belongs to "AutoFillMarkerSection" not to "CanvasTileLayer" i think.

And one last thing:
* instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();
* If we can find a more reliable way to detect the "AutoFilter" of Calc, then we won't need to modify Control.JSDialog.js i think.

Having only one key "Arrow Down" for interaction with pop up items doesn't seem enough to me. But that's something @pedropintosilva knows better :)

Can you check these when you have time? We can also have a follow up for these if you wish.

Thanks @bayramcicek :)

bayramcicek:
> Why don't we set "isAutoFillFromOnMouseUp" to false on MouseDown? Would we still need to set it in other places like "hideSheet" etc if we did set it on MouseDown?
> 
> Also "isAutoFillFromOnMouseUp" property belongs to "AutoFillMarkerSection" not to "CanvasTileLayer" i think.
> 
> And one last thing:
> 
>     * instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();
> 
>     * If we can find a more reliable way to detect the "AutoFilter" of Calc, then we won't need to modify Control.JSDialog.js i think.
> 
> 
> Having only one key "Arrow Down" for interaction with pop up items doesn't seem enough to me. All arrow keys could work there. But that's something @pedropintosilva knows better :)
> 
> Can you check these when you have time? We can also have a follow up for these if you wish.
> 

I noted the reviews and would like to add improvements with a follow-up patch asap.

Thanks for the reviews @gokaysatir  o/.

bayramcicek:
> I am getting below error when i use "Arrow Down" and then "TAB" keys:
> 
> `Widget.TreeView.js:584 Uncaught TypeError: Cannot read properties of undefined (reading 'parentNode')`

This is fixed by @GulsahKose on `master`: https://github.com/CollaboraOnline/online/pull/9071/files#diff-8b226a03c86b8abe2744e941ebd655e85a96001273479a40214593a8f63f9928R599 . Thanks.

--

I will forward-port my PR to `master` asap (it is decided that we'll move this PR to `master`), so we also won't have this bug on `autoFill` popup.

bayramcicek:
closing this PR as this is ported to `master` : https://github.com/CollaboraOnline/online/pull/9135

PR #9041, opened 2024-05-13
fix(zoom,pan): Make screen edge offset document

Minion3665:
Zooming always leads to same backwards and forwards, even when you have
hit the screen edge. At the edge of the screen, you can move no further,
which causes the illusion of it being "sticky" on the return journey.

By adding an offset when we reach the edge, we can avoid the stickiness.

I have made
panning past the document edge create an offset too. This provides
similar benefits to the fluidity of scrolling at document edges.

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -672,9 +672,11 @@ L.TileSectionManager = L.Class.extend({
 			yMin = splitPos.y;
 		}
 
-		const documentTopLeft = new L.Point(xMin, yMin);
+		const minTopLeft = new L.Point(xMin, yMin);
 
 		const paneSize = paneBounds.getSize();
+
+		pinchCenter = pinchCenter.subtract(this._offset);
 
 		let centerOffset = {
 			x: pinchCenter.x - pinchStartCenter.x,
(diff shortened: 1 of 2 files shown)


mmeeks:
Thanks Skyler, Szymon is on vacation for two weeks - Gokay is the backup reviewer =)

gokaysatir (changes requested):
Seems to work well on iPad device. But i cannot see the console there.

On desktop, there is a JS error which breaks the view. Can you have a look at that?

Also, this seems to change the "_onPreDrag" function for all draggables? Not sure if that's the case but i guess it'd be good to ensure that this doesn't break the coordinates of objects while dragging etc.

Thanks :)

Minion3665:
> On desktop, there is a JS error which breaks the view. Can you have a look at that?

sure! I've not been able to reproduce a JS error on desktop from this, are there any specific steps I need to take to trigger this?

> Also, this seems to change the "_onPreDrag" function for all draggables? Not sure if that's the case but i guess it'd be good to ensure that this doesn't break the coordinates of objects while dragging etc.

Thanks, good call, I'll look closer into this...

Minion3665:
> Also, this seems to change the "_onPreDrag" function for all draggables?

Had some time to take a closer look... are you sure? `_onPreDrag` is changed in `browser/src/map/handler/Map.Drag.js`. My understanding was that that is only used for dragging the map, rather than all draggables.

gokaysatir:
> > Also, this seems to change the "_onPreDrag" function for all draggables?
> 
> Had some time to take a closer look... are you sure? `_onPreDrag` is changed in `browser/src/map/handler/Map.Drag.js`. My understanding was that that is only used for dragging the map, rather than all draggables.

I am not sure, it looked to me like and i wanted to inform. Didn't dive into it but i will check & test after the first issue is resolved.

gokaysatir:
> sure! I've not been able to reproduce a JS error on desktop from this, are there any specific steps I need to take to trigger this?
 
No specific steps, just zoom in - zoom out. Let me know if you can't reproduce and i'll debug it.

Minion3665:
> > sure! I've not been able to reproduce a JS error on desktop from this, are there any specific steps I need to take to trigger this?
> 
> No specific steps, just zoom in - zoom out. Let me know if you can't reproduce and i'll debug it.

Thanks! I reproduced it, it turns out that I was using a desktop device that had a touchscreen, and if you ever touchscreen zoom everything was set up properly. Oops.

I've pushed up my changes, I've also rebased and since as GitHub doesn't show you a range diff, here's one I made earlier
```diff
[skyler@canvasback online]$ git range-diff a40a20dfa486a8f56326dc545c7a8a2a32ad52de~..aac82a18aab67b547e50b2645a4bf167b9c46e65 c9fa489e377f0e926f71382e6feb288b0089a866~..ecc3c1e57a000353931e74840d7d147b90ac2317
1:  a40a20dfa4 ! 1:  c9fa489e37 fix(zoom): Make screen edge offset document
    @@ browser/src/layer/tile/CanvasTileLayer.js: L.TileSectionManager = L.Class.extend
                        center: this._map.project(this._map.unproject(newPaneCenter, this._map.getZoom()), this._map.getScaleZoom(scale))
                };
        },
    -
    - ## browser/src/map/handler/Map.TouchGesture.js ##
    -@@ browser/src/map/handler/Map.TouchGesture.js: L.Map.TouchGesture = L.Handler.extend({
    -           this._pinchStartCenter = { x: e.center.x, y: e.center.y };
    -           const _pinchStartLatLng = this._map.mouseEventToLatLng({ clientX: e.center.x, clientY: e.center.y });
    -           this._map._docLayer.preZoomAnimation(_pinchStartLatLng);
    -+          this._map._docLayer._painter._offset = new L.Point(0, 0);
    -   },
    +@@ browser/src/layer/tile/CanvasTileLayer.js: L.CanvasTileLayer = L.Layer.extend({
    + 
    +   preZoomAnimation: function (pinchStartCenter) {
    +           this._pinchStartCenter = this._map.project(pinchStartCenter).multiplyBy(app.dpiScale); // in core pixels
    ++          this._painter._offset = new L.Point(0, 0);
      
    -   _onPinch: function (e) {
    +           if (this.isCursorVisible()) {
    +                   this._cursorMarker.setOpacity(0);
2:  aac82a18aa = 2:  ecc3c1e57a fix(pan): Make screen edge offset document
```

gokaysatir (approved):
Thanks for the improvements.

PR #9042, opened 2024-05-13
Private/caolan/dns perf

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -21,6 +21,7 @@
 
 #include <Poco/Exception.h>
 #include <Poco/Net/DNS.h>
+#include <Poco/Net/NetException.h>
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
Looks plausible to me - thanks Caolan =)
Ideally of course we'd have a (single) thread with a queue of these resolutions if we don't have the cache entry that can add a callback into the SocketPoll to continue the process as/when they are looked up but ... =)

I also wonder if we are not doing DNS lookups on IP addresses where we should/could just compare them directly to the incoming IPs ;-) did you look at the data coming through that ?

Anyhow - lets get it in & watchdog it =)

caolanm:
yeah, plugging away at the threaded dns idea, with at least the capabilities hotspot case implemented

PR #9047, opened 2024-05-13
Private/mmeeks/clipboardfix

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -13,7 +13,7 @@
  * local & remote clipboard data.
  */
 
-/* global app _ brandProductName $ ClipboardItem */
+/* global app _ brandProductName $ ClipboardItem Promise */
 
 // Get all interesting clipboard related events here, and handle
 // download logic in one place ...
@@ -37,6 +37,10 @@ L.Clipboard = L.Class.extend({
 		this._dummyDiv = div;
 		this._dummyPlainDiv = null;
 		this._dummyClipboard = {};
+
+		// Tracks waiting for UNO commands to complete
+		this._commandCompletion = [];
+		this._map.on('commandresult', this._onCommandResult, this);
 
 		div.setAttribute('id', this._dummyDivName);
 		div.setAttribute('style', 'user-select: text !important');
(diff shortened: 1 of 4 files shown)


browser/src/map/Clipboard.js
```diff
@@ -828,26 +851,44 @@ L.Clipboard = L.Class.extend({
...
+			try {
+				var result = fetch(url).then(response => response.text());
+				resolve(result);
+			} catch (err) {
+				reject(err);
+			}
+		});
+
+		var awaitPromise = function(url, mimetype, shorttype) {
+			return new Promise((resolve, reject) => {
```
vmiklos, line 881: Oh nice, I didn't consider that the callback can be shared to fetch the data only once.

wsd/protocol.txt
```diff
@@ -242,9 +242,12 @@ dialog <command>
 
     <command> is unique identifier for the dialog that needs to be painted.
 
-uno <command>
+uno <command> [serial=text] [arguments]
```
vmiklos, line 245: Is this 'serial=' correct? I don't see any output from 'git grep serial=', but perhaps I miss something.

vmiklos (approved):
The actual race fix looks good to me, thanks for noticing this. I'm not yet merging, possibly you want to add more commits to this PR first. I also didn't check if the CI failure is unrelated or not.

browser/src/map/Clipboard.js
mmeeks: =)

mmeeks: Ah yes, this was my 1st thought at adding a new reply with a serial in it so we can undo any races related to which UNO command was invoked and which the reply is for - which would be by far more ideal, but is a bit vicious to get threaded through the core & API & so on ;-)

mmeeks:
CI failure is clearly un-related - this code-change can't affect C++ unit tests =)

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1450/console fails in unit-integration, let's see if this is a one-off, retry.

PR #9056, opened 2024-05-15
Introduce new formula auto complete feature

GulsahKose:
When we start to type a formula in a cell eg: =S
We will see the possible function names in a popup. If we move with "ArrowDown and ArrowUp" and enter, or direcly select a function with mouse, function name will be automatically inserted into cell. Then we will see the usage details of the selected function in another dialog.

Change-Id: Ic06ca0d904770e95bb85ccdf9a73f60b7d5c5eb0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Really clean code :) Thanks for the improvements @GulsahKose.

There are some JS errors and some usage behaviours that we can improve. But we can also do them on a follow up PR. I'll send some more info internally.

browser/src/layer/marker/TextInput.js
```diff
@@ -1112,6 +1112,27 @@ L.TextInput = L.Layer.extend({
 				this._map.fire('closementionpopup', { 'typingMention': false });
 			}
 		}
+
+        var formulaAutocompletePopup = L.DomUtil.get('formulaautocompletePopup');
```
eszkadev, line 1116: I would like to unify the code with mention popup above - it's copy pasted with different id.
We can use function with id parameter for now.

browser/src/control/Control.JSDialog.js
```diff
@@ -700,7 +700,7 @@ L.Control.JSDialog = L.Control.extend({
 			instance.updatePos = this.setPosition.bind(this, instance);
 
 			// Special case for nonModal dialogues. Core side doesn't send their initial coordinates. We need to center them.
-			if (instance.nonModal) {
+			if (instance.nonModal && !(instance.startX && instance.startY)) {
```
eszkadev, line 703: why it is needed? how to test?
best to do such modifications of common code in separate commit with description

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2014,43 +2014,23 @@ L.CanvasTileLayer = L.Layer.extend({
 	},
 
 	_onCalcFunctionUsageMsg: function (textMsg) {
-		var pos = this._lastVisibleCursorRef._northEast;
-		this._map.uiManager.showFormulaTooltip(textMsg, pos);
+		var formulaAutocompletePopup = L.DomUtil.get('formulaautocompletePopup');
+		if (formulaAutocompletePopup)
+			this._map.fire('closeformulapopup');
```
eszkadev, line 2019: do we need to check if it exists here? can we just fire the event, and check if it exists inside event handler? (so we will not repeat that code again)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2014,43 +2014,23 @@ L.CanvasTileLayer = L.Layer.extend({
...
-				this._openMobileWizard(data);
-			}
-			else {
-				var pos = this._lastVisibleCursorRef._northEast;
-				var tooltipinfo = this._getFunctionList(textMsg);
-				this._map.uiManager.showFormulaTooltip(tooltipinfo, pos);
-			}
+			var formulaAutocompletePopup = L.DomUtil.get('formulaautocompletePopup');
+			if (formulaAutocompletePopup)
+				this._map.fire('closeformulapopup');
```
eszkadev, line 2027: is that `if` needed? can we just fire event?

browser/src/control/AutoCompletePopup.ts
```diff
@@ -20,6 +20,7 @@ interface WidgetJSON {
 	enabled: boolean | undefined; // enabled state
 	visible: boolean | undefined; // visibility state
 	children: Array<WidgetJSON> | undefined; // child nodes
+	title?: string;
```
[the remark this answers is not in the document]
GulsahKose: I send title as empty string here. If I don't use the title for formula usage dialog writes "Undefined dialog" as title. To remove that i think i have to use.

browser/src/control/Control.JSDialog.js
GulsahKose: We have tree type of dialog:
    * Modal (isModalPopUp = true): non-movable + overlay + dimmed background.
    * Nonmodal: movable + no dim + no overlay (user can interact with the document).
    * Popup (Non-dialog) (isDocumentAreaPopup = true): overlay + no dim.
But i need a new type: no dim + no overlay + no titlebar(same like popup)

"Overlay" prevents  selecting a cell, Formula usage dialog has position info but not popup. Popups covers a layer and disable to selecting a cell. So I've solved the positioning problem like that.

browser/src/layer/tile/CanvasTileLayer.js
GulsahKose: Done with https://github.com/CollaboraOnline/online/pull/9201

browser/src/layer/tile/CanvasTileLayer.js
GulsahKose: Done with https://github.com/CollaboraOnline/online/pull/9201 New PR because it decided not merge 23.05 branch So I'll abort this PR after the conversation.

GulsahKose:
@eszkadev I've replied your questions, If all is okey, we can abort this PR, Because formula autocomplete won't merged 23.05

browser/src/control/Control.JSDialog.js
eszkadev: so can we explicitly define such type and name it? so we will have it documented in the comment and `instance.isYourNewType` property set ?

GulsahKose: I'm not sure that is a new type just a nonmodal dialog without title but have position info

GulsahKose: But maybe we can use isFormulaUsage same like isAutofilter?

eszkadev: I prefer some generic name, for a "popup without overlay" - so we can later reuse the same thing without confusion ion other cases.

PR #9057, opened 2024-05-15
Status bar: Fix MenuButtons being generated always with image

pedropintosilva:
Before this commit menu buttons were always generated with img even if
it's not applicable. Example, the StateTableCellMenu (located in Calc
status bar that allows setting a state according to the function
chosen in the submenu): This text-only menu was generated with img
and thus throwing a 404 error in the console.

- Allow any unoToolButton to be set without image via option
- Add image: false option when setting StateTableCellMenu in the
statusbar

Change-Id: I3fc3b5954f80229835526819b9b80d2a650bd922


pedropintosilva:
Note: it seems it fails on unit test but it seems unrelated (I will probably re-try it) locally cypress tests pass .

gokaysatir (approved):
Thanks for the improvement :)

PR #9058, opened 2024-05-15
Reenable screenshots on cypress failure

neilguertincollabora:
Also remove SCREENSHOT_FOLDER environment variable, this information is available from Cypress.config(), and if it needs to change it should be set there.

Screenshots were disabled in https://github.com/CollaboraOnline/online/commit/2ec6bd70a2d5ca43e2dddb3c6837e5d4b263a8c0, in July 2023.

I find they are useful, especially for sporadic failures or failures in CI only, where it is not so easy to just rerun the test with screenshots on. Doubly so for sporadic failures in CI.

Change-Id: Idbfb627fa818a6864da355a990dbac2589f8df37


gokaysatir (approved):
Thanks @neilguertincollabora :) @hcvcastro can you also comment when you have time? In case we may miss some important point.

mmeeks:
Well - we have lots of sporadic failures of Cypress - so lets enable this if it helps.

PR #9066, opened 2024-05-15
postmessage: Fix show hide / button / command in Notebookbar

hfiguiere:
This is a regression from commit 9923e695959
This is because jQuery .child() is only for the direct descendants. Use .find() instead which will search down the DOM tree.

This restore postmessage Hide_Command / Hide_Button and Show.

Change-Id: Ibe84c520c9786c6c0026d3507a27a8d3580b38a0

* Resolves: # 
* Target version: master (24.04)

### Summary

This is in 24.04 only

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -336,7 +336,7 @@ L.Control.Notebookbar = L.Control.extend({
 	},
 
 	showNotebookbarButton: function(buttonId, show) {
-		var button = $(this.container).children('#' + buttonId);
+		var button = $(this.container).find('#' + buttonId);
 		if (show) {
 			button.show();
 		} else {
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.Notebookbar.js
```diff
@@ -336,22 +336,22 @@ L.Control.Notebookbar = L.Control.extend({
 	},
 
 	showNotebookbarButton: function(buttonId, show) {
-		var button = $(this.container).children('#' + buttonId);
+		var button = $(this.container).find('#' + buttonId);
```
hfiguiere, line 339: the selector before was using the "descendant" combinator ` `.

hfiguiere:
The test failure is an assertion in coolwsd. Unrelated.

gokaysatir (approved):
Nice catch, thanks :) We could also use this.container.querySelector("...") i guess.

gokaysatir (approved):

gokaysatir (approved):

hfiguiere:
> Nice catch, thanks :) We could also use this.container.querySelector("...") i guess.

or `.getElementById` in that case.

PR #9067, opened 2024-05-15
jsdailog:css: increase width to fit content in manage changes dialog

lpranam:
increased jsdialog width to fit content when necessary. max width is set to 1000px because some other dialog(i.e: Zotero) already uses tree widget width 800px which makes the entire dialog approx 990px.

removed margin from icon to make it look even spaced in listview (i.e: manage changes, zotero add citation)

Change-Id: I225b5b5d42d573cb4afcdb428e2eb3b81aa99073

* Target version: master 

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -42,6 +42,15 @@
 	min-width: 400px;
 }
 
+.jsdialog-container:not(.snackbar) td[role='gridcell'] {
+	padding-left: 5px;
+	padding-right: 5px;
+}
+
+#AcceptRejectChangesDialog {
+	width: max-content;
+	max-width: 1000px;
+}
 .jsdialog.one-child-popup {
 	border: 0;
 	margin: 0 !important;


pedropintosilva:
Thank you @lpranam but with this commit specifically when using max-content which padding and surrounding context, granted you used also max-width to limit the side effects of that. Nevertheless the content is not wrapping (until it reaches the max-width). This in turn seems to make every jsdialog quite wide.

| With this commit | Before |
| --------------------------| --------- |
| image | image |

lpranam:
Now it should only affect manage changes dialog :)

lpranam:
Now it should only affect manage changes dialog :) @pedropintosilva

pedropintosilva (changes requested):
thanks.
It still affects other places introducing the lack of space regression in places like navigator or function wizard:

image

pedropintosilva (approved):
Thanks

PR #9068, opened 2024-05-16
browser: display unauthorized error message

Ashod:
It seems there are two handler for this message.
Since we moved the authentication logic ahead
of the web-socket upgrade, the handler for
the unauthorized error doesn't display the
message. This fixes the it so now we correctly
display the unauthorized message.

Change-Id: Ic62476c74ce1583a2d7f33e1233e6fafd43d3bc4


thebearon (approved):
Solves the issue for me, thanks Ash!

gokaysatir (approved):

PR #9069, opened 2024-05-16
Private/caolan/async perf

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/AsyncDNS.hpp
+++ b/net/AsyncDNS.hpp
@@ -1,0 +1,64 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <condition_variable>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <queue>
+#include <string>
+#include <thread>
+
+namespace net
+{
+
+class DNSResolver;
+
+class AsyncDNS
+{
+public:
+    AsyncDNS();
+    ~AsyncDNS();
+
+    static void startAsyncDNS();
+    static void stopAsyncDNS();
+
+    typedef std::function<void(const std::string& hostName, const std::string& exception)> DNSThreadFn;
+
+    static void canonicalHostName(const std::string& addressToCheck, const DNSThreadFn& cb);
+
+private:
+    std::atomic<bool> _exit;
+    std::unique_ptr<DNSResolver> _resolver;
+    std::unique_ptr<std::thread> _thread;
+    std::mutex _lock;
+    std::condition_variable _condition;
+    struct Lookup
+    {
+        std::string query;
+        AsyncDNS::DNSThreadFn cb;
+        // for now just canonicalHostName lookups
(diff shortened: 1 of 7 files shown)


net/NetUtil.cpp
```diff
@@ -138,6 +154,88 @@ bool isLocalhost(const std::string& targetHost)
...
+    {
+        while (_lookups.empty() && !_exit)
+            _condition.wait(guard);
+
+        if (_exit)
+            break;
+
+        Lookup lookup = _lookups.front();
+        _lookups.pop();
+        std::string hostToCheck, exception;
```
mmeeks, line 196: Seems we hold the lock while we do the slow synchronous lookup (?) =)

net/NetUtil.cpp
```diff
@@ -138,6 +154,88 @@ bool isLocalhost(const std::string& targetHost)
...
+            exception = "net::canonicalHostName(\"" + lookup.query + "\") failed: " + exc.displayText();
+        }
+
+        lookup.cb(hostToCheck, exception);
+    }
+}
+
+void AsyncDNS::addLookup(const std::string& lookup, const DNSThreadFn& cb)
+{
+    std::unique_lock<std::mutex> guard(_lock);
```
mmeeks, line 213: Given the above - superficially it looks like this will still block the main thread while trying to push a lookup if the other thread is doing a slow sync. lookup (?) - which might defeat the parallelism surely ?

wsd/ClientRequestDispatcher.cpp
```diff
@@ -370,6 +372,114 @@ getConvertToBrokerImplementation(const std::string& requestType, const std::stri
...
+        assert(_asyncCb);
+        _selfLifecycle = shared_from_this();
+        dispatchNextLookup();
+    }
+
+    void dispatchNextLookup()
+    {
+        net::AsyncDNS::DNSThreadFn pushHostnameResolvedToPoll = [this](const std::string& hostname,
+                                                                       const std::string& exception) {
+            COOLWSD::getWebServerPoll()->addCallback([this, hostname, exception]() {
```
mmeeks, line 444: Great - glad we push into the right thread at the end =) I wonder if API-wise we might want to take a SocketPoll & - as well as a function - so it's clear in the API that we will always want a callback to happen in the right thread and it's not possible to re-use the async-DNS stuff outside that safety-net ;-) but - fine for now I guess.

net/AsyncDNS.hpp
```diff
@@ -0,0 +1,64 @@
...
+        // for now just canonicalHostName lookups
+    };
+    std::queue<Lookup> _lookups;
+
+    void resolveDNS();
+    void addLookup(const std::string& lookup, const DNSThreadFn& cb);
+
+    void startThread();
+    void joinThread();
+};
```
mmeeks, line 60: Would like to have a dumpState() mechanism that chains to the classes waiting on things too - so we can see all the pending work in the queue, and where it goes to etc. on the console as/when we're head-scratching over some odd behavior here :-)

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1284,7 +1377,7 @@ void ClientRequestDispatcher::handlePostRequest(const RequestDetails& requestDet
         requestDetails.equals(1, "get-thumbnail"))
     {
         // Validate sender - FIXME: should do this even earlier.
-        if (!allowConvertTo(socket->clientAddress(), request))
+        if (!allowConvertTo(socket->clientAddress(), request, nullptr))
```
mmeeks, line 1380: Surely this is still rather synchronous with a nullptr callback ?

mmeeks (commented):
Very quick skim - and a few comments =) Prolly I mis-understood some things and we should have a call =)

mmeeks (approved):
Looks good to me =)

PR #9070, opened 2024-05-16
browser: fix print menu for Calc

JaumePujantell:
Changed the menu entries to conform to the new format used by
JSDialog menu since 07da50a.

Change-Id: I69109675ec69d66eb25ae128a6bb102e4628c705


gokaysatir (approved):
Thanks for the fix :)

Issue #9072, opened 2024-05-16
Better clipboard paste error handling on Chrome for cross-origin iframes


vmiklos:
### Describe the Bug

23.05 used to have suboptimal copy, but it did work to some extent. Now if an integration is not updated to allow clipboard write for cross-origin iframes, copy doesn't work at all on 24.04.

### Steps to Reproduce
1. Open e.g. a Writer document in COOL, in a setup where COOL and the integration is served from different domains. E.g. nextcloud with its richdocuments.git, stable28 branch, with commit 6e29846aeda3bb933935400bb7334a3a71445211 reverted. But https://github.com/owncloud/richdocuments/issues/545 is probably a similar environment.
2. Select some text.
3. Press 'copy' on the notebookbar.

### Expected Behavior

If the integration doesn't allow new-style copy, a popup should appear encouraging the user to try again via the keyboard. We should also remember that new-style copy failed and not try again.

### Actual Behavior

An error is logged on the JS dev console, but otherwise nothing.

### Desktop

Tested on Chrome, with old enough (see above) Nextcloud integration, COOL and Nextcloud served from different domains.

### Additional Context

An alternative would be to just remember the failure for the duration of the edit session (global JS variable). That would have the benefit that in case the integration gets fixed, new-style copy gets a chance again in the next edit session. The cost would be that every time the first copy would be broken. Seems the cost outweights the benefit, so for now we save a setting and never try again, even not after a browser reload.

vmiklos:
The above PR fixes this for me.

chris-fj:
I wanted to share that I updated to [`24.04.5.2.1`](https://hub.docker.com/layers/collabora/code/24.04.5.2.1/images/sha256-b2cdde5cabb3cfd59933daf3ce55990b258a4a1391d31d54833e436e7b7fc26f?context=explore) along with ownCloud 10.14 and the clipboard issue is gone. Thanks for your work here

PR #9073, opened 2024-05-16
cool#9072 browser: better error handling when navigator.clipboard.write is advertised but fails

vmiklos:
In case an integration doesn't allow clipboard interaction via
<https://sdk.collaboraonline.com/docs/advanced_integration.html#allow-the-clipboard-permission-query>,
we end up in a situation where navigator.clipboard.write is not
undefined, but is broken.

By the time navigator.clipboard.write() fails, it's too late to fall
back to the old copy code, as the security context is already gone.

Fix the problem by improving the failure handling: show the popup to try
again copying, remember that navigator.clipboard.write() failed and
prefetch the text selection, so next time a copy is tried (via the
keyboard), the selection doesn't need re-creating. (Normally the
selection change would trigger the prefetch, so select->copy->copy-again
would still fail.)

Note that paste doesn't have a similar problem, since there we always
try the old paste code first, and only use the new paste code if the old
one fails.

Change-Id: I12ea4810395970421000d213744d1838213c7a07

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -470,6 +470,11 @@ window.app = {
 		}
 	})();
 
+	if (global.isLocalStorageAllowed && global.localStorage.getItem('hasNavigatorClipboardWrite') === 'false') {
+		// navigator.clipboard.write failed on us once, don't even try it.
+		global.L.Browser.hasNavigatorClipboardWrite = false;
+	}
+
 	global.deviceFormFactor = global.mode.getDeviceFormFactor();
 
 	if (global.ThisIsTheiOSApp) {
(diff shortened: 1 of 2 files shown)


vmiklos:
Rebase to be on top of 873485f479ffa4fcdc607b6427475fd201ff0f1b

vmiklos:

https://github.com/CollaboraOnline/online/issues/9072 has some details on this, I went the more conservative route and even persist the disable flag in local storage (vs only setting the JS global to false, and try again on next doc edit). This should help with owncloud and other integrations which are left untouched at the moment and would see this as a 23.05 -> 24.04 regression. Nextcloud is not affected, I tested the change with a local revert.

caolanm (approved):

PR #9077, opened 2024-05-16
js: show progress bar during background save

Rash419:
Change-Id: I3608bf597ab8d2c7660a1f884cf2c763c7652415

[Screencast from 2024-05-16 21-09-10.webm](https://github.com/CollaboraOnline/online/assets/26241069/b91591a4-c0ce-43af-aba7-a6699e167385)

* TODO:
   - [x]  firefox colors are not correct, need to fix that

* To Test:
  1. Make autosave interval to something like 30 seconds
  2. Open a huge document
  3. Edit the document and wait for background autosave to happen

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -27,6 +27,30 @@
 	to { left: 100%; }
 }
 
+#document-name-input-progress-bar {
+	width: 100%;
+	height: 2px;
+	border: none;
+	background-color: var(--color-background-dark);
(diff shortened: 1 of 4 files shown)


mmeeks:
Save torture is not happy currently - but the other tests are fine; I need to get to fixing that.
Lets push it =)

mmeeks (approved):

PR #9083, opened 2024-05-16
cool.html: append WOPISrc parameter to all JS, CSS & svg loads

caolanm:
Change-Id: I1dd3902403a93914a8ad03b24bed5524a59e5aca

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (changes requested):
Lets not do this for now - I suspect it breaks our download file caching for browsers, and we need to think this through more carefully I believe.

caolanm:
https://github.com/CollaboraOnline/online/pull/9034 is the "real" one, I created this duplicate to experiment with getting ci to pass

PR #9089, opened 2024-05-17
wsd: added config option for background manualsave

Rash419:
Change-Id: I76de35fe9b4135b74a96436f0c9a3245f4e00bd9

* Target version: master

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -188,6 +188,7 @@ DocumentBroker::DocumentBroker(ChildType type, const std::string& uri, const Poc
     , _mobileAppDocId(mobileAppDocId)
     , _alwaysSaveOnExit(COOLWSD::getConfigValue<bool>("per_document.always_save_on_exit", false))
     , _backgroundAutoSave(COOLWSD::getConfigValue<bool>("per_document.background_autosave", true))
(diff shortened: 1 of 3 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -2743,7 +2744,8 @@ bool DocumentBroker::sendUnoSave(const std::shared_ptr<ClientSession>& session,
 
     // Note: It's odd to capture these here, but this function is used from ClientSession too.
     bool autosave = isAutosave || (_unitWsd && _unitWsd->isAutosave());
-    bool background = forceBackgroundSave || (autosave && _backgroundAutoSave);
```
mmeeks, line 2746: So - for the very last save of the document - we don't want to do a background save. Can you check from the logs that if you modify a document, then close the window without saving - that the final 'force' save is not a background save - even with the settings turned on.

Thanks!

mmeeks (changes requested):
We also need to dig into the wopi save unit test to see what is going on there. Perhaps it is best to get this setting included - but disabled initially as/when you've cleaned up the comment =)

Thanks!

mmeeks:
I'll work on finishing and merging this then.

Rash419:
closing this in favor of https://github.com/CollaboraOnline/online/pull/9098

PR #9091, opened 2024-05-17
private/pedro/status bar fixes

pedropintosilva:
- Toolbar-down (status bar): Fix hover status
- Toolbar-down (status bar): Don't over style disable status


pedropintosilva:
This does affect functionality and also doesn't affect positioning or visibility so if needed we can just skip cypress test.

gokaysatir (approved):
Thanks @pedropintosilva :)

PR #9092, opened 2024-05-17
Fix excessive logging when a session hard quits very early on.

mmeeks:
Enable Caolan's unit test, pass the proper 'disconnect' message to the Kit so it can close the underlying window / resource for the view.

Potentially this also removes 'phantom' users in the user-list which might be another symptom of this issue.

Change-Id: Ib0d0c5cefa7033fff5827d0a825a932cc12f8323

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
unsurprisingly passes for me locally while it never did before, so that's great

PR #9104, opened 2024-05-19
Private/gokay/cell cursors

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/canvas/sections/ShapeHandleRotationSubSection.ts
```diff
@@ -0,0 +1,119 @@
...
+		this.sectionProperties.lastDraggingDistance = null;
+
+		this.getHTMLObject().style.borderRadius = '50%';
+		this.getHTMLObject().style.border = '1px solid black';
+		this.getHTMLObject().remove();
+		document.getElementById('map').appendChild(this.getHTMLObject());
+		this.mirrorEventsFromSourceToCanvasSectionContainer(this.getHTMLObject());
+	}
+
+	mirrorEventsFromSourceToCanvasSectionContainer (sourceElement: HTMLElement): void {
```
eszkadev, line 31: this mirror thing possibly could be shared inside some base class

eszkadev (approved):
code looks ok

browser/src/canvas/sections/ShapeHandleRotationSubSection.ts
gokaysatir: Yes, i didn't want to do that now because we will remove them all :) Thank you for checking.

eszkadev:
But I get a lot of:
```
Uncaught TypeError: app.map.uiManager.getDarkModeState is not a function
``` 
in calc (debug.html)

gokaysatir:
> But I get a lot of:
> 
> ```
> Uncaught TypeError: app.map.uiManager.getDarkModeState is not a function
> ```
> 
> in calc (debug.html)

Thanks again :) fixed.

PR #9105, opened 2024-05-20
kit: Log UnitBase initialization on a more appropriate level

thebearon:
Change-Id: Id4dd61173271fcae5ea941680a39ee9889a3969c

* Target version: master 

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -724,7 +724,7 @@ int forkit_main(int argc, char** argv)
         return EX_USAGE;
     }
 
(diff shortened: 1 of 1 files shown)


mmeeks (commented):
Hmm - I guess we want this to come out on any server. It should never appear in a production environment. For a customer - this is an error - if it is ever seen - and a security error if that is ever compiled into a production build.

That is if UnitTestLibrary is non-empty.

I guess that's why it is ERR - but ... YMMV.

timar:
This PR is obsoleted by 6ed575f6af2ece458912102bfa89ced8188a47d5

PR #9116, opened 2024-05-20
Toolbar-down: Fix Search input field and other elements alignment

pedropintosilva:
Change-Id: Ib1e1b1247dc42e721921c6bf1b68726cf0c916e5


pedropintosilva:
This PR also doesn't change any functionality. It's CSS only and only changing the vertical alignment of items through flex and only for toolbar down so Cypress tests are useless when it comes to test this

PR #9117, opened 2024-05-20
Writer: Accessibility Check: Add missing CSS class

pedropintosilva:
Now that accessibility checker has change its command it's best to add
back in the CSS class that was being added automatically (based on the
uno command) for various reasons:
- Main one: the ".unoSidebarDeck.A11yCheckDeck" is less than ideal
class name (it includes a dot in the middle)
- Integrators might have customizations targeting the previous known
css class

Change-Id: I7685d000a2110872cdbe83323136b7a6be604ca3


pedropintosilva:
This PR doesn't need to wait for cypress test only for CodeQL test to pass because it doesn't remove any CSS class it only adds one.

Issue #9108, opened 2024-05-20
Zotero add reference dialog closes unexpectedly when setting the language


pedropintosilva:
### Steps to Reproduce
1. Make sure the Zotero API works  is passed
   * Note that the API key works and its access is correctly recorded on zotero side: "Collabora Online 	5 mins ago"
3. Click on References > Add Citation
4. In the dialog set "Language" to something else that is not "undefined"
5. See error

![Screenshot_20240520_113415](https://github.com/CollaboraOnline/online/assets/65948705/9950faec-9ab3-47e2-b06c-636bd761a061)

```javascript
11:36:23.844 control: 'combobox' id:'zoterolocale' event: 'selected' state: '1;English (US)' cool.html:324:37
11:36:23.844 1716197783845 OUTGOING: dialogevent ZoteroDialog {"id":"zoterolocale", "cmd": "selected", "data": "1;English (US)", "type": "combobox"} cool.html:324:37
11:36:23.885 1716197783885 INCOMING: error: cmd=dialogevent kind=syntax
```

### Version
5ec388e8b0d758fc13b0bce81a376d1a96dc391c (HEAD -> master, origin/master

Issue #9109, opened 2024-05-20
Zotero: Error when attempting to view library items


pedropintosilva:
### Attempting to click on empty folders
1. Go to References > Add citations
1. Wait for the elements to load
2. See error

![image](https://github.com/CollaboraOnline/online/assets/65948705/889553db-ccf2-4425-abf6-b852fa09794f)

how it looks on zotero:

![image](https://github.com/CollaboraOnline/online/assets/65948705/66e108d9-ecf5-404e-8e71-e9b8605072f7)

### Attempting to click on empty folders
1. Go to References > Add citations
3. Open folder and then empty sub folder
1. See error

![image](https://github.com/CollaboraOnline/online/assets/65948705/187d42ba-e0fb-473d-b472-efc564e46de0)

```javascript
11:51:30.971 Uncaught TypeError: targetEntry.children is undefined
    _expandCollection Control.Zotero.js:1067
    _onAction Control.Zotero.js:1111
    expandFunction Widget.TreeView.js:251
    handler DomEvent.js:51
    _on DomEvent.js:86
    on DomEvent.js:21
    _treelistboxEntry Widget.TreeView.js:265
    _treelistboxEntry Widget.TreeView.js:259
    _treelistboxEntry Widget.TreeView.js:259
    _treelistboxControl Widget.TreeView.js:725
    treeView Widget.TreeView.js:808
    build Control.JSDialogBuilder.js:3303
    _updateWidgetImpl Control.JSDialogBuilder.js:3129
    updateWidget Control.JSDialogBuilder.js:3153
    onJSUpdate Control.JSDialog.js:737
    fire Events.js:152
    _fetchCollectionAndPushTo Control.Zotero.js:1055
    promise callback*_fetchCollectionAndPushTo Control.Zotero.js:1040
    _expandCollection Control.Zotero.js:1074
    _onAction Control.Zotero.js:1111
    expandFunction Widget.TreeView.js:251
    handler DomEvent.js:51
    _on DomEvent.js:86
    on DomEvent.js:21
    _treelistboxEntry Widget.TreeView.js:265
    _treelistboxEntry Widget.TreeView.js:259
    _treelistboxControl Widget.TreeView.js:725
    treeView Widget.TreeView.js:808
    build Control.JSDialogBuilder.js:3303
    _updateWidgetImpl Control.JSDialogBuilder.js:3129
    updateWidget Control.JSDialogBuilder.js:3153
    onJSUpdate Control.JSDialog.js:737
[...]

PR #9123, opened 2024-05-21
zotero: fix: can't select language or store as option in citation style

Rash419:
- problem: in combobox widget we were always passing the `_defaultCallback` which will send `dialogevent` to server but in zotero's case the callback is different
- also fixes some `undefined type` errors

Change-Id: I7bd7aff6a1a3a9715123c1335ea83ab13fd7b22a

* Resolves: #9108 #9109 
* Target version: master


pedropintosilva (approved):
tested, it works

gokaysatir (approved):

PR #9124, opened 2024-05-21
wsd: cache WASM files and serve statically

Ashod:
This grows the largest file we cache to accommodate for WASM files.


caolanm (approved):

caolanm (approved):
seems reasonable

PR #9127, opened 2024-05-21
mobile: add save progress bar on top toolbar

Rash419:
Change-Id: I14c8a715e922d00b2cfd8f67afd7fc2df2dd9d38

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -1113,3 +1113,27 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 	-webkit-user-select: text;
 	user-select: text;
 }
+
+.progress-bar {
+	width: 100%;
+	height: 2px;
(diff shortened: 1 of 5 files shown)


gokaysatir (approved):
Cool, thanks @Rash419 :)

pedropintosilva (changes requested):
thanks @Rash419 but it seems we have a refactoring + code change in the same commit which makes hard now and more important in the future when trying to track down commit history. Could you please separate those. 1st the refactoring and then the code change?

pedropintosilva (approved):
never mind, I don't want to block this. So let's merge it. Still maybe a follow up would be to avoid the position shift in the document page


https://github.com/CollaboraOnline/online/assets/65948705/05bdf587-c4a5-4b13-b3b0-e24fc9ce01c4

PR #9129, opened 2024-05-21
Private/gokay/cursor visibility

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
What I don't really understand is what is different vs 23.05. As far as I can tell the core there also sends out "cursorvisible: false" in the zoom, but the cursor seems to remain visible in 23.05. Did we fix something to honour the cursorvisible more honestly in the online side at some point.

mcecchetti (approved):
So, after first typed character text cursor is visibile again.

PR #9130, opened 2024-05-21
Private/caolan/rejig slurp delay

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -437,11 +437,19 @@ app.definitions.Socket = L.Class.extend({
 	// buffer of web-socket messages in the client that we can't
 	// process so - slurp and the emit at idle - its faster to delay!
 	_slurpMessage: function(e) {
+		this._extractTextImg(e);
+
+		// Some messages - we want to process & filter early.
+		var docLayer = this._map ? this._map._docLayer : undefined;
+		if (docLayer && docLayer.filterSlurpedMessage(e))
+			return;
+
+		console.log('Predict ' + (docLayer ? docLayer.predictTilesToSlurp() : -1) + ' tiles');
+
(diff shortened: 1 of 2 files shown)


caolanm:
Can try scaling delay based on num of expected tiles to a max of 50ms.

wrt. better,  not overwhelmed, maybe slightly better on first start. Otherwise I don't know if it makes a discernible difference.

original, [Screencast from 2024-05-28 20-39-02.webm](https://github.com/CollaboraOnline/online/assets/833656/907a7cb3-e715-4df7-b0ec-e24852806af9)

after, [Screencast from 2024-05-28 20-39-02.webm](https://github.com/CollaboraOnline/online/assets/833656/eaa63836-573b-4cb1-9f4c-d992be64455e)

mmeeks (approved):
All looks very plausible; I assume when we have slurped all the predicted tiles - that we get back to a 1ms delay - and/or cancel the timer and get on with it - which is what we should do =)
Thanks!

mmeeks:
Wrt. the videos - do we manage to predict any tiles incoming as we press page-down; IIRC we somehow managed not to predict anything there (I think) - perhaps we are not counting missing tiles or something in the calculation or ... (?) could that be related; at least I didn't see much difference in the videos ;-) but conceptually it seems a sensible approach.

caolanm:
yeah, if the prediction is 0 then if there is an earlier longer-scheduled timer existent, it's rescheduled to as soon as possible.

wrt videos, I don't see much of a difference either. On adding some bespoke logging, then in a local server case I can see some initial long delay timer launches at first start, but otherwise they seem to get rescheduled to short generally, so I feel the actual change is modest.

PR #9132, opened 2024-05-22
annotation: show dialog for conflicting comment modification

lpranam:
problem:
when we imported comments we cleared all the comments and inserted every comments from start.
This caused problem if some action from another user caused to import the comments for all users(i.e: show/hide changes) This means if comments are in autosave state, import will fail

Change-Id: I888dfeb53fb215f7c7a0507593a06b09da7caf64

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -868,6 +868,8 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 			}
 		};
 
+		var removedComment = this.getComment(id);
+		removedComment.sectionProperties.selfRemoved = true;
 		if (app.file.fileBasedView) // We have to set the part from which the comment will be removed as selected part before the process.
 			this.map.setPart(this.sectionProperties.docLayer._selectedPart, false);
 
@@ -1254,7 +1256,10 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 	public onACKComment (obj: any): void {
 		var id;
 		const anyEdit = Comment.isAnyEdit();
-		if (anyEdit && anyEdit.sectionProperties.data.id === obj.comment.id && CommentSection.autoSavedComment !== anyEdit) {
+		if (anyEdit
+			&& !anyEdit.sectionProperties.selfRemoved
+			&& anyEdit.sectionProperties.data.id === obj.comment.id
+			&& CommentSection.autoSavedComment !== anyEdit) {
 			this.handleCommentConflict(obj, anyEdit);
 			return;
 		}
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -1214,25 +1214,48 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		}
 	}
 
+	public overWriteCommentChanges(obj: any, editComment: Comment) {
+		this.clearAutoSaveStatus();
+		editComment.onCancelClick(null);
+		this.onACKComment(obj);
+	}
+
+	public handleCommentConflict(obj: any, editComment: Comment) {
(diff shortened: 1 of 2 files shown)


gokaysatir (changes requested):
Thanks for the catch. I can't feel comfortable with how we catch events here. User can also press ESC key for example. Pending comments will not be executed in that case.

I think below approaches might be also useful for the users:
* We can show a non-blocking pop-up to the user when we want to reload the comments. Pop-up will say something like "Comments content has changed. List will be reloaded." etc.

* 1- We can use a timer to check if comment editing continues which calls "isAnyEdit" with an interval.
* 2- Or we can put the "doPendingCommentImport" function call into the root places we finish editing the comment. Like "onSaveComment" instead of "handleSaveCommentButton".
* 3- Or we can check how "isAnyEdit" is working. Then we can add event listeners to "modifyNode", "replyNode" etc and we can be sure that comment editing is done.
* I'd go with either 1 or 3. Either way, we need to be sure our method is reliable. Or user's comment list may not get refreshed in corner cases.

* If finding a reliable method becomes a challenge, we can also prefer putting there a button for manual refresh. That'd also be better than non-refreshing comment list.

Thank you.

gokaysatir (changes requested):
Yep, thanks @lpranam :) Also thanks for the detailed "someone else is touching this comment too" message.

I think we can polish this a little more.

* Users are prompted with a message for their own changes (overwrite message). We need to find a way to distinguish them from the others.
* I think we cannot ask if user wants to overwrite if their comment is deleted. Even user wants to overwrite a deleted comment, they can't.
* Also, what happens if user doesn't update the comments and keep modifying their comment and save? When is the list updated in that case?

After we polish & beautify these cases, we'll also need to consult @pedropintosilva. Since we are adding a new popup.

Thanks :)

lpranam:
> * Users are prompted with a message for their own changes (overwrite message). We need to find a way to distinguish them from the others.

I tried this thing but it is not possible as of now. For example, if user A authored a comment and is editing it and user B modifies it, we will never know with the current limitation that it was modified by user B, we only get comment data which says the author is user A. So as of now, it is hard to distinguish who modified the comment.


> * I think we cannot ask if user wants to overwrite if their comment is deleted. Even user wants to overwrite a deleted comment, they can't.
Thanks for spotting, updated

> * Also, what happens if user doesn't update the comments and keep modifying their comment and save? When is the list updated in that case?
AFAIK the list update doesn't contain any changes in the comments it is more of a change in the view (i.e: show tracked changes) that triggers this action so it's safe to proceed in this manner and accommodate the user change first and then update the view.

gokaysatir:
> > * Users are prompted with a message for their own changes (overwrite message). We need to find a way to distinguish them from the others.
> 
> I tried this thing but it is not possible as of now. For example, if user A authored a comment and is editing it and user B modifies it, we will never know with the current limitation that it was modified by user B, we only get comment data which says the author is user A. So as of now, it is hard to distinguish who modified the comment.

Hmm, i see. But still, we cannot get this in. Because i am modifying a comment, noone does anything else and i get below message. We either need to solve this or rethink the solution. We may postpone the popup message for now and focus on solving the "cancelled edit session" issue. Can you think of something when you have time? Thank you.

gokaysatir (approved):

Let me also add @pedropintosilva as reviewer.

Darshan-upadhyay1110:
I tested, this patch looks good to me. Just a small thing I want to mention, we use `Title Case` for Dialog header. Here I see that has not been followed.

Change `Comments updated` to `Comments Updated`

@lpranam can you please change that part and rebase the PR. Then we are good to merge it

Darshan-upadhyay1110 (approved):

Darshan-upadhyay1110:
Cypress fails seems to false positive i have triggered it again. 

@gokaysatir @lpranam please feel free to merge it after green check from CI.

Darshan-upadhyay1110 (approved):

PR #9135, opened 2024-05-22
[forward-port] add auto fill popup: Copy cells and Fill series

bayramcicek:
AutoFill contextmenu has two options:

- Copy Cells
- Fill Series

- requires 2 upstream patches: 
  - https://gerrit.libreoffice.org/c/core/+/167726 (MERGED to `master`, `23.05`, `24.04`) - 5ab9b63e6624947a6076a01f9bab0d7a925e0e40
  - https://gerrit.libreoffice.org/c/core/+/168086 - f7345a0d4b24210d27326e5d9acd5ba92d468f1d

* Resolves: # 
* Target version: master
* see also: (`23.05`) https://github.com/CollaboraOnline/online/pull/9039

### Summary

- Added UI
- close AutoFill contextmenu when
  - switch tabs by clicking on a different tab button
  - switch tabs by `Ctrl+Alt+PgUp/Down`
- Added functionalities for "Copy Cells" and "Fill Series".
- close AutoFill contextmenu when contextmenu opens from sheet tabs

### TODO

- [ ] ...

### Checklist


bayramcicek:
@eszkadev : Hi. I'd like to have your review on this PR. 

If there are small changes needed I can do it immediately. Otherwise; If there are no blockers, let's merge, and I'd like to implement your reviews with a follow-up patch since I work on another accessibility bug now. 

Thanks in advance :)

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * Control.AutoFillOptions - class for auto fill popup
```
eszkadev, line 12: best to write how to found it so someone new will know

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -770,6 +770,10 @@ L.Map.Keyboard = L.Handler.extend({
 						else {
 							partToSelect = currentSelectedPart != 0 ? currentSelectedPart - 1 : parts - 1;
 						}
+
+						if (this._map._docLayer.isAutoFillPopupOpen)
+							this._map.fire('closeautofillpopup');
```
eszkadev: please just fire an event and do the "if ( ... ) " in the handler

browser/src/control/Control.Tabs.js
```diff
@@ -324,6 +324,9 @@ L.Control.Tabs = L.Control.extend({
 		if (part !== this._map._docLayer._selectedPart) {
 			this._setPartIndex(part);
 		}
+
+		if (this._map._docLayer.isAutoFillPopupOpen)
```
eszkadev: please just fire an event and do the "if ( ... ) " in the handler

browser/src/control/Control.JSDialog.js
```diff
@@ -602,7 +602,8 @@ L.Control.JSDialog = L.Control.extend({
 			var dialog = this.dialogs[autoFilterDialogId];
 
 			// Check if the current dialog has the isAutofilter property set to true
-			if (dialog.isAutofilter) {
+			// and if it is not an autoFillPopup
+			if (dialog.isAutofilter && (autoFilterDialogId !== 'autoFillPopup')) {
```
eszkadev: why is that needed? what it does? it doesn't look right

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -76,20 +75,20 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
 		data.posx = framePos.x;
 		data.posy = framePos.y;
 		this.sendJSON(data);
+
+		this.map._docLayer.isAutoFillPopupOpen = true;
```
eszkadev: Please do not move implementation details outside of the class. We can store that information here - or not store but just check.

Please add `isOpen()` method to the base class, which can be used instead maybe?

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -51,8 +51,8 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
 	}
 
 	getPopupEntries(): any[] {
-		const entries: any[] = [{ text: '', columns: [{ text: 'Copy cells' }], row: '' },
-		{ text: '', columns: [{ text: 'Fill series' }], row: '' }];
+		const entries: any[] = [{ text: 'copy', columns: [{ text: 'Copy cells' }], row: '0' },
```
eszkadev: Labels should be alwasy translated:
`'abc' -> _('abc')`

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -113,7 +113,10 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
 			}
 
 		} else if (eventType === 'keydown') {
-			// TODO: handle arrowDown key
+			if (object.key !== 'Tab' && object.key !== 'Shift') {
+				this.map.focus();
+				return true;
+			}
```
eszkadev, line 120: it shouldn't be much different that in the base class - can we call the base class method?

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -51,8 +51,10 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
 	}
 
 	getPopupEntries(): any[] {
-		const entries: any[] = [{ text: 'copy', columns: [{ text: 'Copy cells' }], row: '0' },
-		{ text: 'fill', columns: [{ text: 'Fill series' }], row: '1' }];
+		const entries: any[] = [
+			{ text: 'copy', columns: [{ text: 'Copy cells' }], row: '0' },
```
eszkadev, line 55: translation

browser/src/control/Control.JSDialog.js
```diff
@@ -600,10 +600,17 @@ L.Control.JSDialog = L.Control.extend({
 		for (var i = 0; i < dialogKeys.length; i++) {
 			var autoFilterDialogId = dialogKeys[i];
 			var dialog = this.dialogs[autoFilterDialogId];
+			var isAutoFillFromOnMouseUp = this.map._docLayer.isAutoFillFromOnMouseUp;
```
eszkadev, line 603: I don't think we should have `isAutofilter == true` for that popup....
How we set this up? is that needed? maybe just don't use it for autofill?

browser/src/control/Control.AutoFillOptions.ts
bayramcicek: done

browser/src/map/handler/Map.Keyboard.js
bayramcicek: this is removed in the new commits: no need

browser/src/control/Control.Tabs.js
bayramcicek: this is removed in the new commits: no need x 2

browser/src/control/Control.AutoFillOptions.ts
bayramcicek: this is removed in the new commits: no need x 4

browser/src/control/Control.AutoFillOptions.ts
bayramcicek: I only changed `text: 'Copy cells'` and `text: 'Fill series'` since they are only visible to user.
(`text: 'copy'` and `text: 'fill'` stay the same as before)

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -95,11 +95,21 @@ class AutoFillOptions extends L.Control.AutoCompletePopup {
...
 			if (index == 0) {
-				// Copy cells
-				this.map.sendUnoCommand('.uno:FillSeries?FillStep:string=0');
+				/*
+					- Copy cells
+					undo needed here, "Copy cells" should operate on the original content,
+					not the new content of the cells.
+				*/
+				this.map.sendUnoCommand('.uno:Undo');
+				this.map.sendUnoCommand('.uno:AutoFill?Copy:bool=true');
```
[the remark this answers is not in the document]
bayramcicek: `undo` calls are removed on the latest commit

browser/src/control/Control.AutoFillOptions.ts
bayramcicek: I moved `if (object.key !== 'Tab' && object.key !== 'Shift')` to the base class.

browser/src/control/Control.AutoFillOptions.ts
bayramcicek: done

browser/src/control/Control.JSDialog.js
bayramcicek: I think this is very useful since when sheet tabs are changed, it automatically closes the tabs (including mine - autoFillPopup) so we don't need to fire `close` for every situation that we need to close popup. 

This code block also runs after when we use `autoFill` with `AutoFill marker box`. Since I open the popup before the cells are filled with data, this for loop closes `autoFill` popup immediately after it opens, therefore we can't see the popup. 

That's why I want to prevent that `this.close(autoFilterDialogId, true)` call only if we use `autoFill` with `AutoFill marker box`. (which triggers with `onMouseUp` in `browser/src/canvas/sections/AutoFillMarkerSection.ts`)

eszkadev: I agree with the effect, but I'm concerned about code, if it is not misleading :)

My question is: why    `isAutofilter` is set for auto fill popup which is not `autofilter` ?

bayramcicek: To be honest. I actually don't know why :) I think mentionPopup and formulaAutoComplete popup is also `true` for ` isAutofilter`.

eszkadev: `instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();`  in Control.JSDialog.js

Could you put your check for `id` there? so it will NOT be `isAutofilter == true` ?

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -3204,6 +3204,12 @@ L.CanvasTileLayer = L.Layer.extend({
 			var topLeftPixels = this._twipsToCorePixels(topLeftTwips);
 			var offsetPixels = this._twipsToCorePixels(offset);
 			this._cellAutoFillAreaPixels = L.LOUtil.createRectangle(topLeftPixels.x, topLeftPixels.y, offsetPixels.x, offsetPixels.y);
+
+			// if comes from onMouseUp event of AutoFillMarkerSection.ts, open AutoFill popup with location data
+			if (this._map._docLayer.isAutoFillFromOnMouseUp) {
+				// TODO: don't open AutoFill popup if empty cells are selected
```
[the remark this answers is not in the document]
bayramcicek: will be done in a follow up asap.

browser/src/control/Control.JSDialog.js
bayramcicek: Thanks, I have missed that part. 
Done with -> https://github.com/CollaboraOnline/online/pull/9135/commits/1bdaba888f3964d8c74bc2fc1572d3545109d7ea

browser/src/control/Control.JSDialog.js
```diff
@@ -606,6 +606,16 @@ L.Control.JSDialog = L.Control.extend({
 				// Call this.close(key, true) for the current dialog
 				this.close(autoFilterDialogId, true);
 			}
+
+			if (autoFilterDialogId === 'autoFillPopup') {
```
eszkadev, line 610: this is no longer needed as we fixed problem with `isAutofilter`

eszkadev: or am I wrong?

bayramcicek: This is needed. I don't want to put `closePopup` manually everywhere when there is a need to close popup on tab change. Here is easy, just don't close the autoFilter popup only if we drag with autoFilterMarker. otherwise close it. :)

eszkadev: ah, it's needed... then please add `instance.isAutoFillPopup` property and detect it here. (We want to avoid using the hardcoded id in many places).
Also change name of the function to `closePopupsOnTabChange` so it will tell the truth.

browser/src/control/Control.JSDialog.js
```diff
@@ -639,7 +649,7 @@ L.Control.JSDialog = L.Control.extend({
 		instance.isDocumentAreaPopup = instance.popupParent === '_POPOVER_' && instance.posx !== undefined && instance.posy !== undefined;
 		instance.isPopup = instance.isModalPopUp || instance.isDocumentAreaPopup || instance.isSnackbar;
 		instance.containerParent = instance.isDocumentAreaPopup ? document.getElementById('document-container'): document.body;
-		instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();
+		instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc() && instance.id !== 'autoFillPopup';
```
eszkadev, line 652: here add `instance.isAutoFillPopup`, and also use it in `isAutofilter` to not duplicate detection

bayramcicek: done

browser/src/control/Control.JSDialog.js
bayramcicek: Thanks ! 
Done

eszkadev (approved):
code looks ok now

bayramcicek:
> autofill-2024-05-24_16.17.05.webm
> 
>     1. click on a corner of a cell
> 
>     2. click anywhere
>        Result: autofill popup appears
> 
> 
> Is that expected?

No. I have to prevent `isAutoFillFromOnMouseUp = true` if there is no dragging. `onMouseUp` called when we do that. then when we select a random cell it thinks it comes from autoFill.

bayramcicek:
> autofill-2024-05-24_16.17.05.webm
> 
>     1. click on a corner of a cell
> 
>     2. click anywhere
>        Result: autofill popup appears
> 
> 
> Is that expected?

Done

eszkadev (approved):
Looks ok as a first step

eszkadev:
I think detection based on booelan when we used "handle" is fragile.
I managed to not show the auto fill popup if I moved cursor very fast.

Can we later trigger it from the core? so we don't need that boolean anymore? What do you think?

bayramcicek:
> I think detection based on booelan when we used "handle" is fragile. I managed to not show the auto fill popup if I moved cursor very fast.
> 
> Can we later trigger it from the core? so we don't need that boolean anymore? What do you think?

it is because of the issue on onMouseUp event in the `browser/src/canvas/sections/AutoFillMarkerSection.ts`:

>> I see that mouseButtonDown/Up called only if cursor is over the AutoFillMarker box. The crosshair can also seen outside of the AutoFillMarker box:
>> Therefore, as far as I can see/understand;
>> - if mouse cursor is very close to the AutoFillMarker box, the crosshair become visible and AutoFill operation executes.
>> - BUT mouseButtonDown/Up is not called there.


> This is interesting. I think we can make the size of the autofill section larger. Hopefully something like that will solve the issue.

bayramcicek:
> I think detection based on booelan when we used "handle" is fragile. I managed to not show the auto fill popup if I moved cursor very fast.
> 
> Can we later trigger it from the core? so we don't need that boolean anymore? What do you think?

I'm doing some changes in core - which seems will makes us get rid of that boolean...

bayramcicek:
> I think detection based on booelan when we used "handle" is fragile. I managed to not show the auto fill popup if I moved cursor very fast.
> 
> Can we later trigger it from the core? so we don't need that boolean anymore? What do you think?

@eszkadev : I added a core change : https://gerrit.libreoffice.org/c/core/+/168086 which helped us to NOT depend on `OnMouseUp` and `isAutoFillFromOnMouseUp`. 

(Tests are failed here because the patch is requires core change)

Ready for review. Thanks :)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1439,6 +1439,9 @@ L.CanvasTileLayer = L.Layer.extend({
 		else if (textMsg.startsWith('cellautofillarea:')) {
 			this._onCellAutoFillAreaMsg(textMsg);
 		}
+		else if (textMsg.startsWith('cellautofillpopuponmouseup:')) {
```
eszkadev, line 1442: more generic name please (no on mouse up in the name)

eszkadev:
Thank you, now it looks more bulletproof :)

browser/src/control/Control.AutoFillOptions.ts
```diff
@@ -0,0 +1,116 @@
...
+			this.closeMentionPopup({ typingMention: false } as CloseMessageEvent);
+		} else if (eventType === 'select' || eventType === 'activate') {
+			this.map.fire('closeautofillpopup');
+
+			if (index == 0) {
+				// Copy cells
+				this.map.sendUnoCommand('.uno:AutoFill?Copy:bool=true');
+			} else if (index == 1) {
+				// Fill series
+				this.map.sendUnoCommand('.uno:AutoFill?Copy:bool=false');
```
eszkadev, line 106: These actions will only do "uno command" calls? We don't need anything additional there?
Now I see we could reuse ... existing context menu  for it then if we don't need any custom content in that popup.

But this PR makes some good cleanup in TextInput.js

browser/src/layer/tile/CanvasTileLayer.js
bayramcicek: DONE

browser/src/control/Control.AutoFillOptions.ts
eszkadev: Please see example CALLBACK and it's JSON when right click on cell:
```
contextmenu: {
    "menu": [
        {
            "text": "~Cut",
            "type": "command",
            "command": ".uno:Cut",
            "enabled": "true"
        },
```

could you check if we can just reuse LOK_CALLBACK_CONTEXT_MENU  which already shows a menu and it can be shown in the correct place?

Then we will avoid custom callback.

eszkadev: You can just send `command` field without `text`, in the online we can get correct translatable string for it.

context menu is implemented in `browser/src/control/Control.ContextMenu.js`

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -3216,6 +3216,7 @@ L.CanvasTileLayer = L.Layer.extend({
 			var topLeftPixels = this._twipsToCorePixels(topLeftTwips);
 			var offsetPixels = this._twipsToCorePixels(offset);
 			this._cellAutoFillAreaPixels = L.LOUtil.createRectangle(topLeftPixels.x, topLeftPixels.y, offsetPixels.x, offsetPixels.y);
+			this._map._docLayer.autoFillLocation = { data: topLeftPixels };
```
eszkadev, line 3219: not used

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -88,6 +88,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		app.sectionContainer.addSection(new app.definitions.AutoFillMarkerSection());
 
 		this.insertMode = false;
+		this.autoFillLocation = {};
```
eszkadev, line 91: not used

eszkadev (commented):
Thanks, now I'm happy with the solution. Thanks for that many updates :)

browser/src/layer/tile/CalcTileLayer.js
bayramcicek: DONE

browser/src/layer/tile/CanvasTileLayer.js
bayramcicek: DONE

bayramcicek:
> Thanks, now I'm happy with the solution. Thanks for that many updates :)

Thanks for the reviews and support! :)

eszkadev (approved):

PR #9136, opened 2024-05-22
cool#9120 - use a simple hash to avoid most tile: comparisons.

mmeeks:
Parsing a string to a TileDesc and then comparing is inefficient enough, without doing that in an N^2 loop for the SenderQueue.

Still N^2 - and of dubious value in the world of deltas; but in almost all cases a simple string + integer compare.

Change-Id: Ibe8230679e27b98cfa95567fea700e2f7d5ac09c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/SenderQueue.hpp
+++ b/wsd/SenderQueue.hpp
@@ -90,11 +90,24 @@ private:
         {
             // Remove previous identical tile, if any, and use most recent (incoming).
             const TileDesc newTile = TileDesc::parse(item->firstLine());
+            uint32_t newTilePosHash = newTile.equalityHash();
+            // store a hash of position for this tile.
+            item->setHash(newTilePosHash);
+
             const auto& pos = std::find_if(_queue.begin(), _queue.end(),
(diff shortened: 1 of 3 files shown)


wsd/SenderQueue.hpp
```diff
@@ -90,11 +90,24 @@ class SenderQueue final
         {
             // Remove previous identical tile, if any, and use most recent (incoming).
             const TileDesc newTile = TileDesc::parse(item->firstLine());
+            uint32_t newTilePosHash = newTile.equalityHash();
+            // store a hash of position for this tile.
+            item->setHash(newTilePosHash);
```
caolanm, line 95: "Item" is only ever a std::shared_ptr<Message> so pointee of const Item& item can be modified, and this is the only place something can get put into the queue.

Not entirely sure that SenderQueue makes sense as a template where only one type could work, but that's not particularly relevant to the change.

caolanm (approved):

PR #9137, opened 2024-05-23
formulabar: fix css for accept / cancel formula

eszkadev:
When editing formula buttons were misaligned. Now they are stretched to fit the empty space.

After:
buttonsalign


eszkadev:
I was focused on showing the buttons in correct place ASAP. Pixel-perfection in sizing can be improved too later

pedropintosilva (changes requested):
This problem doesn't happen everywhere(without integrator and with some integrators). It seems to be specific to one of the integrator's customizations

PR #9148, opened 2024-05-24
Move hello-world.odg to hello.odg

thebearon:
So it behaves similar to other hello/hello-world documents.

Change-Id: I529b5bec81c1bf5085300f8c52311db023c6b09a

* Target version: master 

### Checklist


vmiklos (approved):
Makes sense, so the -world variant can be edited for testing purposes without changing a tracked file.

PR #9149, opened 2024-05-24
Add missing dark mode icons for 24.04

Darshan-upadhyay1110:
- Added missing icons for dark mode
- Statusbar: Set image to false for zoom menubutton
- Handle case for items only available in dark mode

Change-Id: Ibf42fc04f953769790e9bc4145d8fe4f6e994298

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/core/LOUtil.js
```diff
@@ -148,6 +164,7 @@ L.LOUtil = {
 				return this.getURL('images/dark/' + imgName);
 			}
 		}
+		defaultImageURL = this.isDarkModeItem(imgName) ? 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=' : defaultImageURL;
```
eszkadev, line 167: Can we add "label" here what is that base64 image?
`var dummyEmptyImg = 'data:image...';`
It will be more obvious :)

eszkadev (approved):
Very nice clean PR :)

pedropintosilva (approved):
Thank you I have took the liberty to add another commit before all of them and to remove existing icons and some of the icons you have added in the dark folder since they are unnecessary.

If you are going to do some changes, first please reset your local branch and git pull -r and then change it. And if that case please feel free to fix the typo in

> // Some items will only present in dark mode so we will not check wrrors for those in other mode

PR #9150, opened 2024-05-24
jsdialog: cleanup and deduplicate

eszkadev:
- no functional changes, just a cleanup
- move types into single file and remove duplicates
- move snackbar internals into separate TS file (to be continued)


GulsahKose (approved):
Look ok to me

PR #9156, opened 2024-05-24
private/pedro/ux fix color picker and remove legacy nb home css

pedropintosilva:
- CharBackColor: Remove legacy CSS rule, no need for extra margin
- Notebookbar: Home: Remove unnecessary CSS and fix fontnamecombobox
- Color pickers: Fix image size and alignment

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -14,6 +14,7 @@
 	--cell-cursor-selection-border-color: var(--color-primary);
 
 	--btn-size: 24px;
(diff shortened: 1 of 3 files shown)


pedropintosilva:
WRT [Color pickers: Fix image size and alignment](https://github.com/CollaboraOnline/online/pull/9156/commits/a6f4584a126f799ab39fc6f44f80a81185d24f24) : this way we can easily target all color pickers both in the tab and in the sidebar

PR #9164, opened 2024-05-27
Close mobile icon should be hidden in desktop

Darshan-upadhyay1110:
- if we switch to compact mode & server get reconnected, some icons are visible in desktop which should be only visible in mobile view.
- in control.toolbar.js we call updateItem method for icon in mobile - `toolbar.updateItem({id: 'closemobile', type: 'customtoolitem', w2icon: 'editmode'});`
- after this updateItem call we should also check all condition about that item like - item should be visible in mobile or not - item should be available in tablet or not etc
- this patch will do that checks

Change-Id: I11adc3d94cacd4afe85e177d050ccc73f3112872

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/jsdialog/Component.Toolbar.ts
+++ b/browser/src/control/jsdialog/Component.Toolbar.ts
@@ -132,6 +132,7 @@ class Toolbar {
 
 	updateItem(data: ToolbarItem) {
 		this.builder.updateWidget(this.parentContainer, data);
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
thanks, looks good :)

eszkadev:
unrelated Cypress fail

PR #9165, opened 2024-05-27
Fix state update of input element

Darshan-upadhyay1110:
- Input elements in jsDialog does not change state after CORE update message
- this is because in previous commits we changed input element structure.
- we should synchronize the child elements with parent.
- this patch will Synchronize disable state of parent with it's child elements

Change-Id: Ice94de732dc7392f3c79100e6a2d60480294eb0e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1603,6 +1603,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === 'false' || data.enabled === false)
 			$(edit).prop('disabled', true);
 
+		JSDialog.SynchronizeDisabledState(container, [edit]);
+
 		edit.addEventListener('keyup', function(e) {
 			var callbackToUse =
 				(e.key === 'Enter' && data.changedCallback) ? data.changedCallback : null;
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -3108,6 +3108,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		var separatorPos = elementId.indexOf(':'); // delete menuId
 		if (separatorPos > 0)
 			elementId = elementId.substr(0, separatorPos);
+		var isEditContainer = document.getElementById(rawId);
```
eszkadev, line 3111: best to use naming `isXXXX` for boolean values

eszkadev: I think we shouldn't add such logic to the function "removeMenuId" (it's not obvious that it will add something for an element, especially that it will no longer point to the root node).

Can we try to adjust executeAction instead? I belive `case 'setText':` needs some adjustments.

eszkadev (approved):

eszkadev:
unrelated Cypress fail

PR #9167, opened 2024-05-28
cypress: quick fix for rounding error

eszkadev:
unblock CI from:
cy:command ✘  assert	expected **<tspan.TextPosition>** to have attribute **y** with the value **'3286'**, but the value was **'3285'**


Darshan-upadhyay1110 (approved):
Tested, Works good.
image

Thanks

PR #9172, opened 2024-05-28
Formula Wizard: better alignment of label that under the Structure tab

bayramcicek:
- the label between Structure tab and TreeView was misaligned. This is corrected with this commit.
- label "Structure:" changed to "Functions in use and their validity:" with https://gerrit.libreoffice.org/c/core/+/168161 (6c15d246afa1774298e1438f2117b593a065bfe4)

Change-Id: I197fb84154226709395f4f2789fc288367c00f05

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1933,10 +1933,12 @@ input[type='number']:hover::-webkit-outer-spin-button {
 }
 #FormulaDialog .jsdialog.ui-tabpage > .ui-grid {
 	grid-template-rows: 0 auto !important;
+	grid-gap: 0px;
 }
 #FormulaDialog #StructPage {
 	display: grid;
 	grid-template-rows: auto 1fr;
+	grid-row-gap: 4px;
 }
 #FormulaDialog #StructPage .ui-treeview-entry {
 	background-color: transparent;
@@ -1967,6 +1969,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #FormulaDialog .ui-tabs-content.jsdialog {
 	grid-auto-columns: 250px;
 	margin-right: 0.2em;
+	margin-top: 0px;
 }
 
 div [id^='TemplateDialog'] {


bayramcicek:
### before:

Screenshot_20240528_181744-before

### after:

Screenshot_20240528_182943-after

bayramcicek:
core: https://gerrit.libreoffice.org/c/core/+/168161

browser/css/jsdialogs.css
```diff
@@ -1933,11 +1933,15 @@ input[type='number']:hover::-webkit-outer-spin-button {
...
 	grid-template-rows: 0 auto !important;
+	grid-gap: 0px;
 }
 #FormulaDialog #StructPage {
 	display: grid;
 	grid-template-rows: auto 1fr;
 }
+#FormulaDialog #StructPage .ui-treeview#struct {
+	margin-top: 4px;
+}
```
pedropintosilva, line 1944: I think we can remove this and instead use the existing declaration `#FormulaDialog #StructPage` and add a `grid-row-gap: 4px;`

browser/css/jsdialogs.css
```diff
@@ -1967,6 +1971,10 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #FormulaDialog .ui-tabs-content.jsdialog {
 	grid-auto-columns: 250px;
 	margin-right: 0.2em;
+	margin-top: 0px;
+}
+#FormulaDialog .ui-tab[id^='tabcontrol'] {
+	margin-bottom: 0px;
```
pedropintosilva, line 1977: Let's not add this. I would prefer to don't add special margin for special cases around tab control, at least for now

pedropintosilva (changes requested):

browser/css/jsdialogs.css
bayramcicek: DONE.

pedropintosilva (approved):

PR #9180, opened 2024-05-29
Private/mmeeks/perf simple

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/StringVector.hpp
+++ b/common/StringVector.hpp
@@ -32,8 +32,9 @@ struct StringToken
 };
 
 /**
- * Safe wrapper around an std::vector<std::string>. Gives you an empty string if you would read past
- * the ends of the vector.
+ * Safe wrapper around an std::vector<std::string>. Gives
+ * you an empty string if you would read past the ends
+ * of the vector.
  */
 class StringVector
 {
--- a/common/Util-desktop.cpp
+++ b/common/Util-desktop.cpp
@@ -458,6 +458,7 @@ std::chrono::microseconds SysStopwatch::elapsedTime() const
     uint64_t totalUs = (nowCPU - _startCPU) + (nowSys - _startSys);
     return std::chrono::microseconds(totalUs);
 }
+
 }
 
 /* vim:set shiftwidth=4 softtabstop=4 expandtab: */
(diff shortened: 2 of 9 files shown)


common/Util-desktop.cpp
```diff
@@ -424,4 +425,57 @@ void alertAllUsers(const std::string&) {}
...
+    sys = 0;
+#if defined __linux__
+    FILE* cg = fopen("/proc/self/stat", "r");
+    if (cg != nullptr)
+    {
+        char line[4096] = { 0 };
+        if (fgets(line, sizeof(line), cg))
+        {
+            StringVector bits = StringVector::tokenize(line, strlen(line), ' ');
+            cpu = std::atol(bits[13].c_str());
```
vmiklos, line 461: Aha, man 5 proc explains these are the utime and stime fields, so this measures time, but only for this process, unlike measuring with std::chrono::system_clock::now().

common/StringVector.hpp
```diff
@@ -32,8 +32,9 @@ struct StringToken
 };
 
 /**
- * Safe wrapper around an std::vector<std::string>. Gives you an empty string if you would read past
- * the ends of the vector.
+ * Safe wrapper around an std::vector<std::string>. Gives
```
vmiklos, line 35: Nit: not sure this is intended, the header in that file says we wrap at 100 chars, the old one seems to be within that limit.

vmiklos (approved):
Looks reasonable to me, asking the kernel to measure the process's time for us seems to be a google middle ground between something not time based (e.g. valgrind pseudo cycles) and just measuring system time. Let's wait till CI passes, though.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1782/console says unit-save-torture failed, retry to see if that's a one-off.

test/UnitPerf.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <thread>
+
+/// Save torture testcase.
+class UnitPerf : public UnitWSD
+{
+    void testPerf();
+
+    void configure(Poco::Util::LayeredConfiguration& config) override
+    {
+//        pfm_initialize();
```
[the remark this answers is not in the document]
vmiklos: Ah yes, are the pfm bits something that you first considered, then concluded that /proc/[pid]/stat has enough info for us, or it's rather the other way around, we read /stat as that works already, but would be good to investigate pfm in the future?

mmeeks:
We will need the pfm bits I expect to read the performance counters and get something reasonably accurate; and to do all of this in the same process I think - the /proc stuff is ok-ish; but not terribly helpful I think.

Issue #9166, opened 2024-05-28
ui_defaults=SavedUIState=false causes setting change inconsistencies


Minion3665:
The most obviously chaotic example is with light/dark mode... As the theme switching code reads from ui_defaults to determine what its next state should be, when you have SavedUIState=false set you can only switch to whatever the state *not* in your ui_defaults is.

In practical terms, this means that if your state is light, you can switch to dark but not back to light.

When you refresh the page the UI loads in light theme, but alas a second related issue strikes: [the code which reads the document loading theme doesn't read your ui_defaults first even when savedUIState=false](https://github.com/CollaboraOnline/online/blob/e674970dcb611cc85726d18ea6aa84077e34832e/browser/js/global.js#L276-L287), so the UI is light but the document is dark

These are caused by inconsistent use of ui_defaults leading to UI inconsistency (e.g. different priority ordering of different preference locations in different places), and code breaking when you fracture its assumptions (e.g. "localStorage will always be in sync with my current theme"). My proposed fix will make some helper functions that won't violate these assumptions even if localStorage was bypassed, and will make it a lot easier to remain consistent in the use of ui_defaults.

---

I should make a change to SavedUIState=false such that it still lets you make changes to the UI state, they just are discarded on reload. I believe this is more similar to the original intent

The way I believe would be best to implement this is to change how we access localstorage so we use a wrapper which can also ignore values in a ui_defaults dependent way. This would also make behavior more consistent if we were unable to use the real localstorage on a particular browser

Minion3665:
> I may have to move prefs into global.js for the initialization order I'll need

checked this, I would need to move both `global.keyboard` and some socket initialization out. I could move `global.keyboard` out, but since as I couldn't move socket initialization out I won't do so now ... may be a good idea to do so in future for unrelated (global.js size) reasons

PR #9181, opened 2024-05-29
fix: fix UI preference inconsistencies

Minion3665:
Add helper functions for fetching preferences from localstorage and
ui_defaults. By having these in unified helpers, we can be certain that
we're not making differences in the priority or use of different stores
which can cause UI inconsistency.

BREAKING-CHANGE: This changes the naming of many of the preferences in localStorage, effectively resetting lots of UI preferences

Change-Id: I4cc9cc885aa8cd0b40646629cd73bef236963a94
Fixes: #9166

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -281,6 +281,113 @@ window.app = {
 		lang: navigatorLang
 	};
 
+	global.prefs = {
+		_localStorageChanges: {}, // TODO: change this to new Map() when JS version allows
+		canPersist: (function() {
+			var str = 'localstorage_test';
+			try {
+				global.localStorage.setItem(str, str);
+				global.localStorage.removeItem(str);
+				return true;
+			} catch (e) {
+				return false;
+			}
+		})(),
+
+		/// Similar to using window.uiDefaults directly, but this can handle dotted keys like "presentation.ShowSidebar" and does not allow partially referencing a value (like just "presentation")
+		_getUIDefault: function(key, defaultValue = undefined) {
+			const parts = key.split('.');
+			let result = global.uiDefaults;
+
+			for (const part of parts) {
+				if (!Object.prototype.hasOwnProperty.call(result, part)) {
+					return defaultValue;
+				}
+
+				if (typeof result === 'string') {
+					return defaultValue;
+				}
+
+				result = result[part];
+			}
+
+			if (typeof result !== 'string') {
+				return defaultValue;
+			}
+
+			return result;
+		},
+
+		get: function(key, defaultValue = undefined) {
+			if (key in global.prefs._localStorageChanges) {
+				return global.prefs._localStorageChanges[key];
+			}
+
+			const uiDefault = global.prefs._getUIDefault(key);
+			if (
+				global.prefs._getUIDefault('SavedUIState', 'true').toLowerCase() === 'false' &&
+				uiDefault !== undefined
+			) {
+				return uiDefault;
+			}
+
+			if (global.prefs.canPersist) {
+				const localStorageItem = global.localStorage.getItem(key);
+
+				if (localStorageItem) {
+					return localStorageItem;
+				}
+			}
+
+			if (uiDefault !== undefined) {
+				return uiDefault;
+			}
+
+			return defaultValue;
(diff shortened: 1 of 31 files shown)


Minion3665:
(forcepushes were a minor nontrivial rebase, here's the range-diff)
```bash
[skyler@canvasback online]$ git range-diff 2fa476e815e55c53e478c744000190806ca7a7c5~..f02315d22cdf61a6cbcc5119b48bdca1a3c0b3e0 HEAD~2..HEAD
1:  2fa476e815 ! 1:  635ada2f4f fix: fix UI preference inconsistencies
    @@ browser/src/control/Control.UIManager.js: L.Control.UIManager = L.Control.extend
        },
      
        shouldUseNotebookbarMode: function() {
    --          var forceCompact = this.getSavedStateOrDefault('compactMode', null);
    -+          var forceCompact = window.prefs.getBoolean('compactMode', null);
    -           return (window.userInterfaceMode === 'notebookbar' && forceCompact === null)
    -                   || forceCompact === false;
    -   },
    +-          let forceCompact = this.getSavedStateOrDefault('compactMode', null);
    ++          let forceCompact = window.prefs.getBoolean('compactMode', null);
    +           // all other cases should default to notebookbar
    +           let shouldUseClassic = (window.userInterfaceMode === 'compact' && forceCompact == null) || forceCompact === true;
    +           return !shouldUseClassic;
     @@ browser/src/control/Control.UIManager.js: L.Control.UIManager = L.Control.extend({
                this.map.fire('darkmodechanged');
        },
2:  f02315d22c = 2:  3dc9638ffe refactor: remove per-app darkmode vestiges
```

browser/js/global.js
```diff
@@ -281,6 +281,113 @@ window.app = {
 		lang: navigatorLang
 	};
 
+	global.prefs = {
```
eszkadev, line 284: I think currently for globals we use:
docstate.js and docstatefunctions.js which are available under `app.something`
But it's fine for now. Later we will rework them to TS and probably merge anyway.

Minion3665: Alright

I imagine if we decide to do that, we'll need to add it to [the app object at the top of global.js](https://github.com/CollaboraOnline/online/blob/e4d2095c71f5f8ee861d36c3d51f0bde7de5e75b/browser/js/global.js#L8) anyway, since as it is used in initialization of other globals

caolanm:
So, we go from a per-application dark/light mode to a global one? I have no particular feelings on way or the other, but I presumed that was a deliberate choice at one point? Maybe @pedropintosilva knows if that is something we're ok with to change.

Though having the ability for *spellchecking* on for writer and off for calc looks a more useful thing than per-application dark mode to me.

caolanm (approved):
but code looks fine if that's what we want to do

Minion3665:
> So, we go from a per-application dark/light mode to a global one? I have no particular feelings on way or the other, but I presumed that was a deliberate choice at one point? Maybe @pedropintosilva knows if that is something we're ok with to change.
> 
> Though having the ability for _spellchecking_ on for writer and off for calc looks a more useful thing than per-application dark mode to me.

I looked at this, the commits where dark mode was introduced, and the ui_defaults built up around it, and I'm doubtful that it was deliberate. In our ui_defaults API, we don't have a per-application dark mode. In the place where we began storing the dark mode pref, and from my reading of the patch and surrounding code it looks more likely to be an accidental omission rather than an intentional choice.

I did decide to check this, though. I really want to avoid removing code where I don't know why it was introduced, and while I *think* it was oversight, it's possible I'm overlooking a reason. I asked about it, and Pedro didn't respond. Michael did and the verdict was pretty much "I'm not sure, the only reason I can think of to have them separate would be if we wanted to use them separately in ui_defaults, removing per-app dark mode makes sense to me".

Nevertheless, I'm happy to wait for @pedropintosilva to weigh in so long as we still get it merged in time for the next release. If that looks like it won't happen, I'll go ahead and merge it anyway.

---

Per-application spellchecking is a good shout; I'll do that in a separate PR if we want that, though

Minion3665:
Heading home now, let's get this merged

PR #9182, opened 2024-05-29
Text droplet (Teardrop): hide it when on desktop

pedropintosilva:
Change-Id: I6544845526c2e70bee7d6e3b193e9fb4931422c6

### DIFF
--- a/browser/css/device-desktop.css
+++ b/browser/css/device-desktop.css
@@ -1,5 +1,13 @@
 /* CSS specific for desktop browsers. */
 
+/* Hide text droplets if accurate pointing device such as a mouse is detected */
+@media (pointer: fine) {
+	.text-selection-handle-start,
+	.text-selection-handle-end {
+		display: none;
+	}
+}
+
 #coolwsd-version span:before,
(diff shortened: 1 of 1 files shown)


thebearon:
I wanted to check this by emulating a desktop-sized touch device, and used the Asus Zenbook Fold preset in Chrome.
What I saw:
- the caret had a teardrop marker, (expected)
- the selection had no teardrop markers. (**not** expected)

Not sure if the device emulator works properly, but would be worth to double-check this (and I wonder how we differentiate between touch vs. non-touch compared to the usual desktop-mobile segmentation).

thebearon:
I guess the touch part can be handled in a follow-up some other time. However, there seem to be tests depending on the droplet that are now failing.

pedropintosilva:
> I guess the touch part can be handled in a follow-up some other time. However, there seem to be tests depending on the droplet that are now failing.

Good catch, thanks! I have forced pushed and rectified that.

thebearon (approved):
Works well, thanks!

PR #9183, opened 2024-05-29
Put following state into docstate, fix following on other sheet

eszkadev:
Needed for jumping fixes


Minion3665 (approved):
Looks fine, and appears to work in my testing ... anything I was dubious of I quickly found to be inherited from the previous following system

PR #9184, opened 2024-05-30
annotation: match the reply order to core

lpranam:
problem:
Core behaviour:
if comment A has 3 replies (i.e: in order B, C, D) when user again replies to comment A
a new reply will be inserted after A but before B.

But in online same reply will be displayed after D. On reload of the document or after save-as when comments are reloaded, this order will change as comments in the doc is saved according to the core behaviour.

Change-Id: I3bf4e06c4b6f48eb086c7eaebe102eac0f90681d

* Target version: master 

### Checklist


gokaysatir (approved):
Thanks for the improvement :)

PR #9191, opened 2024-05-30
fix(impress): Allow rightclick on selected videos

Minion3665:
Previously you were unable to open the context menu on desktop (via right click) if the video had been selected (with left click) beforehand. This was frustrating if you knew about it, and confusing if you didn't

We can fix this by sending the rightclick event to core if it's triggered on a video.

Change-Id: Icb275854c8dfb0349ce6ef401b14adeb152e53f6

### DIFF
--- a/browser/src/layer/vector/SVGGroup.js
+++ b/browser/src/layer/vector/SVGGroup.js
@@ -91,7 +91,7 @@ L.SVGGroup = L.Layer.extend({
 
 		var videoContainer = svgLastChild.querySelector('body');
 		var videos = svgLastChild.getElementsByTagName('video');
-		this.addVideoSupportHandlers(videos);
+		this.addVideoEventHandlers(videos);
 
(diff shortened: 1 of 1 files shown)


gokaysatir (commented):
There is a weird behaviour. Different browsers work differently here.

On FireFox, context menu is not opened. On Chrome, it is opened.

With this PR. FireFox still doesn't open the context menu and Chrome gets the context menu event twice.

I am working on a PR that will remove SVGGroup usage. Can we wait for that one for now?

I'm not sure if new PR may help with the issue but we may avoid duplicate work.

Minion3665:
> There is a weird behaviour. Different browsers work differently here.

Interesting, I only tested this on Firefox, if Chrome behaves differently let's not merge this

> I am working on a PR that will remove SVGGroup usage. Can we wait for that one for now?

Sure, thanks!

PR #9186, opened 2024-05-30
Improvement to conditional formatting

codewithvk:
Previously, the menu and submenu of conditional formatting options were technically correct but not very readable. We replaced 'Condition...' with 'Highlight cells with...'.

Screenshot from 2024-05-30 11-14-45
Screenshot from 2024-05-30 11-14-19
Screenshot from 2024-05-30 11-14-06

Change-Id: I92a326e1394d1b40211748062c5bcddea049ffb4

* Resolves: # 
* Target version: master

PR #9194, opened 2024-05-31
Cypress: Mobile calc: nonexistent #test-div-cell_selection_handle_start

pedropintosilva:
I have noticed by looking at one of the PRs
https://github.com/CollaboraOnline/online/pull/9186 that this
unrelated test is failing.

I have tested locally and indeed, sometimes it fails. It seems it
tries to check for '#test-div-cell_selection_handle_start' but I
couldn't find it either. But I did find '.text-selection-handle-start'

Change-Id: I1daec0df5b8135935127b1de1348458b288c2d51

---

Here is the breakdown (v passes, x fails):

Mobile: calc/cell_appearance_spec.js
 Apply left and right border
- [ v ] cGet input#addressInput-input
   - log<< removeTextSelection - end
   [Comment] ^ from calc_helper.js:removeTextSelection()
- [ x ] cGet#test-div-cell_selection_handle_start
  - expected #test-div-cell_selection_handle_start to exist in the DOM
  [Comment] ^ Called in the calc_helper.js:helper.doIfOnMobile()

----

### DIFF
--- a/cypress_test/integration_tests/common/calc_helper.js
+++ b/cypress_test/integration_tests/common/calc_helper.js
@@ -156,7 +156,7 @@ function selectEntireSheet() {
 		});
 
 	helper.doIfOnMobile(function() {
-		cy.cGet('#test-div-cell_selection_handle_start').should('exist');
+		cy.cGet('.text-selection-handle-start').should('exist');
 	});
 
 	var regex = /^A1:(AMJ|XFD)1048576$/;


codewithvk (approved):

gokaysatir:
This issue may mean that the object is visible but the section visibility is set to false.
I'll need to check this.

gokaysatir (changes requested):
The issue might have been solved with the recent improvements, mentioned PR seems to have passed the cypress tests.

I think we don't need this PR.

pedropintosilva:
@gokaysatir at the time I inspected the DOM looking at the whole HTML structure (including hidden elements) and nowhere we were creating an element with a **css class** `text-selection-handle-start` but we do create an element with the **id** `text-selection-handle-start`

gokaysatir:
> @gokaysatir at the time I inspected the DOM looking at the whole HTML structure (including hidden elements) and nowhere we were creating an element with a **css class** `text-selection-handle-start` but we do create an element with the **id** `text-selection-handle-start`

I think you meant with class, not id? Because your patch replaces id with class name.

Anyway, i see your point, so it means that the object is not visible. Because if the object is not visible "the test div gets deleted" if i remember correctly. There may be some divs with classes but doesn't matter, the visibility check can be done correctly with test-divs.

I think CI is fine so far, can we keep this out for some time? If we see failures then we can always get this in.

pedropintosilva:
> > @gokaysatir at the time I inspected the DOM looking at the whole HTML structure (including hidden elements) and nowhere we were creating an element with a **css class** `text-selection-handle-start` but we do create an element with the **id** `text-selection-handle-start`
> 
> I think you meant with class, not id? Because your patch replaces id with class name.
> 
> Anyway, i see your point, so it means that the object is not visible. Because if the object is not visible "the test div gets deleted" if i remember correctly. There may be some divs with classes but doesn't matter, the visibility check can be done correctly with test-divs.
> 
> I think CI is fine so far, can we keep this out for some time? If we see failures then we can always get this in.

yes my mistake while typing, amended my comment. 

Understood. Closed this PR

PR #9197, opened 2024-05-31
Unload Kit Session/View

Ashod:
wsd: always exit Kit when unloading

We never re-use a Kit process. Therefore,
when we unload a document, we must also
exit the Kit process.

Alternatively, when there is at least
1 session, we keep the document loaded
and the Kit process so that a new
view can be created.


mmeeks (approved):
Looks plausible - thanks Ash =)

Issue #7984, opened 2024-01-09
Formula to Value


neilguertincollabora:
### Is your feature request related to a problem?

In LibreOffice Calc, you can replace a formula with its value by
Data > Calculate > Formula to Value

This option does not exist in Collabora Online, or if it does, it's in a different place.

There is a workaround, but it's not obvious. The workaround is to copy the cells and then paste as plain text.

### Describe the solution you'd like

There should be a Formula to Value button under the Data tab in Collabora Online.

timur-g:
I marked as Enancement. But for copy/paste of formula, Ctrl+V also pastes formula, while Cltr+Shift+V pastes value.

PR #9199, opened 2024-05-31
calc: add support for .uno:ConvertFormulaToValue command

rparth07:
Change-Id: Ia2f48eca778c396ae1fa17afe2d92d9631aadad5

* Resolves: #7984
* Target version: master 

### Summary
- Added support to replace a formula with its value in Data > Formula to Value
- Here, is the LOKit patch for the same: https://gerrit.libreoffice.org/c/core/+/168387

### Checklist


hfiguiere:
shouldn't there be LOKit patch to enable the command status?

rparth07:
> shouldn't there be LOKit patch to enable the command status?

Do you have any reference on how to enable command status using LOKit patch? I'm not much aware about it.

rparth07:
> > shouldn't there be LOKit patch to enable the command status?
> 
> Do you have any reference on how to enable command status using LOKit patch? I'm not much aware about it.

After conducting some research, I discovered the following approach: 
https://gerrit.libreoffice.org/c/core/+/132585

To enable the command, it appears that I need to modify two files in the core:
1. desktop/source/lib/init.cxx
2. sfx2/source/control/unoctitm.cxx

Could you please confirm if this is the correct way?

browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -1735,11 +1735,35 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'vertical': 'true'
 			},
 			{
-				'id': 'data-calculate',
```
codewithvk, line 1738: @pedropintosilva Are we really using `Recalculate` in Online? I think it is useful once we have a check menu of `AutoCalculate`.

browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -1735,11 +1735,35 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
...
+								'type': 'toolitem',
+								'text': _UNO('.uno:ConvertFormulaToValue', 'spreadsheet'),
+								'command': '.uno:ConvertFormulaToValue',
+								'image': false,
+								'accessibility': { focusBack: true,	combination: 'R', de: null }
+							}
+						]
+					}
+				],
+				'vertical': 'true'
```
codewithvk, line 1766: I would prefer to have a submenu for Recalculate and Formula to Value under the Calculate menu, just like we have in LO, but I would love to hear feedback from @pedropintosilva .  

Screenshot from 2024-06-01 13-43-23

rparth07:
@eszkadev and @pedropintosilva would love to hear your thoughs on this.

rparth07:
here, is the LOKit patch to enable the uno command: https://gerrit.libreoffice.org/c/core/+/168387

Darshan-upadhyay1110:
hello @rparth07 i just encounter the same thing you are having.And thanks to @eszkadev i solved it :)

Follow these steps.

Remove your hard coded change in unocommand.js
then run `    scripts/unocommands.py --update /path/to/online /path/to/loffice `

This will simply fix your issue.

- Please close your CORE pr as it will be not need now after above steps.

Thanks 🙇🏼

pedropintosilva (approved):
Thanks!

I think we should merge this , it just needs an icon.

Later we can think if we should have a dropdown or now

Darshan-upadhyay1110:
@rparth07 Please rebase this PR. 
CC: @pedropintosilva

rparth07:
> @rparth07 Please rebase this PR. CC: @pedropintosilva

Rebased!

PR #9201, opened 2024-06-01
Reduce code duplication

GulsahKose:
Follow-up commit 2acc5a94c9d883d1dc92ca52cb9a92116d177009

Move the "if exist" conditions into event handler

Change-Id: Iefdb36d95f53af79dcc3e7694a68e5cac3b1dfa1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.FormulaUsagePopup.ts
+++ b/browser/src/control/Control.FormulaUsagePopup.ts
@@ -43,14 +43,7 @@ class FormulaUsagePopup extends L.Control.AutoCompletePopup {
 	onAdd() {
 		this.newPopupData.isAutoCompletePopup = true;
 		this.map.on('openformulausagepopup', this.openFormulaUsagePopup, this);
-		this.map.on('closeformulausagepopup', this.closeFormulaUsagePopup, this);
 		this.map.on('sendformulausagetext', this.sendFormulaUsageText, this);
-	}
-
-	closeFormulaUsagePopup() {
-		var formulaUsagePopup = L.DomUtil.get('formulausagePopup');
-		if (formulaUsagePopup)
-			super.closePopup();
 	}
 
 	openFormulaUsagePopup(ev: FireEvent) {
--- a/browser/src/control/Control.FormulaUsagePopup.ts
+++ b/browser/src/control/Control.FormulaUsagePopup.ts
@@ -43,8 +43,14 @@ class FormulaUsagePopup extends L.Control.AutoCompletePopup {
 	onAdd() {
 		this.newPopupData.isAutoCompletePopup = true;
 		this.map.on('openformulausagepopup', this.openFormulaUsagePopup, this);
-		this.map.on('closeformulausagepopup', this.closePopup, this);
+		this.map.on('closeformulausagepopup', this.closeFormulaUsagePopup, this);
 		this.map.on('sendformulausagetext', this.sendFormulaUsageText, this);
+	}
+
+	closeFormulaUsagePopup() {
+		var formulaUsagePopup = L.DomUtil.get('formulausagePopup');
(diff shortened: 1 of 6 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1741,20 +1741,14 @@ L.CanvasTileLayer = L.Layer.extend({
...
 	_onCalcFunctionListMsg: function (textMsg) {
 		if (textMsg.startsWith('hidetip')) {
-			var formulaAutocompletePopup = L.DomUtil.get('formulaautocompletePopup');
-            if (formulaAutocompletePopup)
-                this._map.fire('closeformulapopup');
-            var formulaUsagePopup = L.DomUtil.get('formulausagePopup');
-            if (formulaUsagePopup)
-                this._map.fire('closeformulausagepopup');
+            this._map.fire('closeformulapopup');
+            this._map.fire('closeformulausagepopup');
```
eszkadev, line 1751: Is there any good reason to have separate events for all these popups?
Can we maybe call just "close all popups" event? And then all popups will do just closePopup on that signal.
(we even have already such event somewhere and we can try to reuse that)

browser/src/control/Control.FormulaUsagePopup.ts
```diff
@@ -43,10 +43,16 @@ class FormulaUsagePopup extends L.Control.AutoCompletePopup {
 	onAdd() {
 		this.newPopupData.isAutoCompletePopup = true;
 		this.map.on('openformulausagepopup', this.openFormulaUsagePopup, this);
-		this.map.on('closeformulausagepopup', this.closePopup, this);
+		this.map.on('closeformulausagepopup', this.closeFormulaUsagePopup, this);
 		this.map.on('sendformulausagetext', this.sendFormulaUsageText, this);
 	}
 
+	closeFormulaUsagePopup() {
+		var formulaUsagePopup = L.DomUtil.get('formulausagePopup');
```
eszkadev, line 51: can be that if condition doen inside super.closePopup ?
possibly we already have that id in the base class?

eszkadev: then we will not need also previous similar code in the autocomplete popup too

GulsahKose: But closePopup is a base class for AutoFill Mention apart from function autocomplete and usage, If i put a condition into base class it will look like following
if (id==formulausagePopup || formulaautocompletePopup)
    do the close action
But autofill and mention will fail this case. I have to add all ids. Does it make sense?
Actually it will work without any condition but it can trig many unnecessary operation.
I think this is best version of code. What do you think?

browser/src/layer/tile/CanvasTileLayer.js
GulsahKose: closeformulausagepopup is not a popup, It is a dialog. I'll try your suggestion, if it works i'll use it

browser/src/control/Control.FormulaUsagePopup.ts
eszkadev: I mean we can store "id" as a member in the base class when we create it.
Then base class will know the id and we will not have to add "if"s.
We will just do something like: `L.DomUtil.get(this.baseId)`

eszkadev: and subclasses can adjust the functionality:
```
closePopup() {
    super.closePopup();
    doAdditionalStuffInThisSubclass();
}
```

eszkadev (approved):
updated & rebased

PR #9202, opened 2024-06-03
Add an endpoint to check connectivity to WOPI server

meven:
The endpoint is accessible at /hosting/wopiAccessCheck

You send it a POST with input json:
{
	"callbackUrl: "<url-to-my-wopi-host>"
}

For instance:
curl -ki https://localhost:9980/hosting/wopiAccessCheck --header "Content-Type: application/json" -d '{"callbackUrl":"https://cool.local"}'

Returns json such as:
{
	"status": 0,
	"details": "OK"
}
With status and details giving hints to what went wrong if it did.

The callbackUrl needs just to be an HTTPS endpoint, it does not need to be the wopi service but can be only a healthcheck url for instance.

Change-Id: Ibdef0bbf0d2fd98e3d293a06dfcfc79478d618e5

cc @juliushaertl 

Superseeds https://github.com/CollaboraOnline/online/pull/4353

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1453,6 +1453,30 @@ private:
         return _response->state() == Response::State::Complete;
     }
 
+    void callOnFinished()
+    {
+        if (!_onFinished)
+            return;
+
+        LOG_TRC("onFinished calling client");
+        std::shared_ptr<Session> self = shared_from_this();
+        try
+        {
+            [[maybe_unused]] const long references = self.use_count();
+            assert(references > 1 && "Expected more than 1 reference to http::Session.");
+
+            _onFinished(self);
+
+            assert(self.use_count() > 1 &&
+                    "Erroneously onFinish reset 'this'. Use 'addCallback()' on the "
+                    "SocketPoll to reset on idle instead.");
+        }
+        catch (const std::exception& exc)
+        {
+            LOG_ERR("Error while invoking onFinished client callback: " << exc.what());
+        }
+    }
+
     /// Set up a new request and response.
     void newRequest(const Request& req)
     {
@@ -1473,26 +1497,8 @@ private:
             assert(_response->state() != Response::State::Incomplete &&
                    "Unexpected response in Incomplete state");
             assert(_response->done() && "Must have response in done state");
-            if (_onFinished)
-            {
-                LOG_TRC("onFinished calling client");
-                auto self = shared_from_this();
-                try
-                {
-                    [[maybe_unused]] const auto references = self.use_count();
-                    assert(references > 1 && "Expected more than 1 reference to http::Session.");
 
-                    _onFinished(std::static_pointer_cast<Session>(self));
-
-                    assert(self.use_count() > 1 &&
-                           "Erroneously onFinish reset 'this'. Use 'addCallback()' on the "
-                           "SocketPoll to reset on idle instead.");
-                }
-                catch (const std::exception& exc)
-                {
-                    LOG_ERR("Error while invoking onFinished client callback: " << exc.what());
-                }
-            }
+            callOnFinished();
 
             if (_response->header().getConnectionToken() == Header::ConnectionToken::Close)
             {
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1236,7 +1236,7 @@ public:
     void setFinishedHandler(FinishedCallback onFinished) { _onFinished = std::move(onFinished); }
 
     /// The onConnectFail callback handler signature.
-    using ConnectFailCallback = std::function<void()>;
+    using ConnectFailCallback = std::function<void(const std::shared_ptr<Session>& session)>;
 
     void setConnectFailHandler(ConnectFailCallback onConnectFail) { _onConnectFail = std::move(onConnectFail); }
 
@@ -1378,6 +1378,11 @@ public:
         {
             socket->closeConnection();
         }
+    }
+
+    net::AsyncConnectResult connectionResult()
+    {
+        return _result;
     }
 
     /// Returns the socket FD, for logging/informational purposes.
@@ -1609,8 +1614,37 @@ private:
 
             if (!socket->send(_request))
             {
+                _result = net::AsyncConnectResult::SocketError;
                 LOG_ERR("Error while writing to socket");
             }
+        }
+    }
+
+    std::shared_ptr<Session> shared_from_this()
+    {
+        return std::static_pointer_cast<Session>(ProtocolHandlerInterface::shared_from_this());
+    }
+
+    void callOnConnectFail()
+    {
+        if (!_onConnectFail)
+            return;
+
+        std::shared_ptr<Session> self = shared_from_this();
+        try
+        {
+            [[maybe_unused]] const long references = self.use_count();
+            assert(references > 1 && "Expected more than 1 reference to http::Session.");
+
+            _onConnectFail(self);
+
+            assert(self.use_count() > 1 &&
+                    "Erroneously onConnectFail reset 'this'. Use 'addCallback()' on the "
+                    "SocketPoll to reset on idle instead.");
+        }
+        catch (const std::exception& exc)
+        {
+            LOG_ERR("Error while invoking onConnectFail client callback: " << exc.what());
         }
     }
 
@@ -1623,7 +1657,10 @@ private:
         {
             LOG_TRC("onHandshakeFail");
             _handshakeSslVerifyFailure = socket->getSslVerifyResult();
+            _result = net::AsyncConnectResult::SSLHandShakeFailure;
         }
+
+        callOnConnectFail();
     }
 
     void onDisconnect() override
@@ -1659,7 +1696,7 @@ private:
         return socket; // Return the shared pointer.
     }
 
-    void asyncConnectCompleted(SocketPoll& poll, std::shared_ptr<StreamSocket> socket)
+    void asyncConnectCompleted(SocketPoll& poll, const std::shared_ptr<StreamSocket> &socket, net::AsyncConnectResult result)
     {
         assert((!socket || _fd == socket->getFD()) &&
                "The socket FD must have been set in onConnect");
@@ -1667,14 +1704,12 @@ private:
         // When used with proxy.php we may indeed get nullptr here.
         // assert(socket && "Unexpected nullptr returned from net::connect");
         _socket = socket; // Hold a weak pointer to it.
+        _result = result;
 
         if (!socket)
         {
             LOG_ERR("Failed to connect to " << _host << ':' << _port);
-
-            if (_onConnectFail)
-                _onConnectFail();
-
+            callOnConnectFail();
             return;
         }
 
@@ -1689,9 +1724,9 @@ private:
     {
         _socket.reset(); // Reset to make sure we are disconnected.
 
-        auto pushConnectCompleteToPoll = [this, &poll](std::shared_ptr<StreamSocket> socket) {
-            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket)]() {
-                asyncConnectCompleted(poll, socket);
+        auto pushConnectCompleteToPoll = [this, &poll](std::shared_ptr<StreamSocket> socket, net::AsyncConnectResult result ) {
+            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket), &result]() {
+                asyncConnectCompleted(poll, socket, result);
             });
         };
 
@@ -1744,6 +1779,7 @@ private:
     ConnectFailCallback _onConnectFail;
     std::shared_ptr<Response> _response;
     std::weak_ptr<StreamSocket> _socket; ///< Must be the last member.
+    net::AsyncConnectResult _result; // last connection tentative result
 };
 
 /// HTTP Get a URL synchronously.
--- a/wsd/ClientRequestDispatcher.cpp
+++ b/wsd/ClientRequestDispatcher.cpp
@@ -23,6 +23,8 @@
 #include <Exceptions.hpp>
 #include <FileServer.hpp>
 #include <HttpRequest.hpp>
+#include <NetUtil.hpp>
+#include <JsonUtil.hpp>
 #include <JailUtil.hpp>
 #include <ProofKey.hpp>
 #include <ProxyRequestHandler.hpp>
--- a/wsd/ClientRequestDispatcher.cpp
+++ b/wsd/ClientRequestDispatcher.cpp
@@ -16,6 +16,7 @@
 #endif
 
 #include <common/Anonymizer.hpp>
+#include <common/StateEnum.hpp>
 #include <Admin.hpp>
 #include <COOLWSD.hpp>
 #include <ClientSession.hpp>
@@ -870,6 +871,8 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
                 servedSync = handleWopiDiscoveryRequest(requestDetails, socket);
             else if (requestDetails.equals(1, "capabilities"))
                 servedSync = handleCapabilitiesRequest(request, socket);
+            else if (requestDetails.equals(1, "wopiAccessCheck"))
+                handleWopiAccessCheckRequest(request, message, socket);
             else
                 HttpHelper::sendErrorAndShutdown(http::StatusCode::BadRequest, socket);
         }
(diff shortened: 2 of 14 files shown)


meven:
```
  /home/collabora/jenkins/workspace/github_android_app_co-24.04/wsd/ClientRequestDispatcher.cpp:59:10: fatal error: 'Poco/Net/SSLException.h' file not found
  #include <Poco/Net/SSLException.h>
           ^~~~~~~~~~~~~~~~~~~~~~~~~
```

wsd/ClientRequestDispatcher.cpp
```diff
@@ -9,10 +9,15 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+#include <Poco/Net/HTTPRequest.h>
+#include <Poco/Net/HTTPResponse.h>
```
mmeeks, line 13: We don't want to use Poco ever if we can avoid it for various reasons.
see net/HttpRequest.hpp - and 'asyncRequest' =)

wsd/ClientRequestDispatcher.cpp
```diff
@@ -49,9 +54,15 @@
 #include <Poco/MemoryStream.h>
 #include <Poco/Net/HTMLForm.h>
 #include <Poco/Net/NetException.h>
+#include <Poco/Net/DNS.h>
 #include <Poco/Net/PartHandler.h>
+#if !MOBILEAPP
+#include <Poco/Net/SSLException.h>
```
mmeeks, line 60: Again we don't want lots of Poco-ness - and particularly synchronous DNS is a nightmare.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -990,6 +1003,155 @@ void ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+    else if (uri.getScheme() != "https")
+    {
+        result = CheckStatus::NotHttps;
+    }
+    else
+    {
+        // request the url
+        try
+        {
+            const auto hostAddress(Poco::Net::DNS::resolve(uri.getHost()));
```
mmeeks, line 1083: This is synchronous and blocks the thread - poke Caolan to see how to use his caching async DNS thread.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -990,6 +1003,155 @@ void ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+
+            Poco::Net::HTTPSClientSession httpSession(uri.getHost(), 443);
+            httpSession.setConnectTimeout(Poco::Timespan(0, 300));
+            Poco::Net::HTTPRequest httpRequest(Poco::Net::HTTPRequest::HTTP_GET,
```
mmeeks, line 1087: This is synchronous and blocks our whole thread; as above see HttpRequest::asyncRequest.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -990,6 +1003,155 @@ void ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+            LOG_ERR_S("Wopi Access Check request error, query to callback ["
+                      << uri.toString() << "] failed:" << exception.what() << exception.className()
+                      << exception.name() << exception.message());
+
+            result = CheckStatus::UnspecifiedError;
+        }
+    }
+
+    // construct the result
+    Poco::JSON::Object::Ptr status = new Poco::JSON::Object;
```
mmeeks, line 1138: For generating JSON an ostringstream + streaming data into it is vastly faster, and in many ways better; we have helpers to generate JSON / escaping - lets use those and not Poco wherever possible.

mmeeks (changes requested):
Good first cut - glad it works; needs re-working =) Thanks!

wsd/ClientRequestDispatcher.cpp
```diff
@@ -990,6 +1003,155 @@ void ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+        }
+        catch (Poco::Exception& exception)
+        {
+            LOG_ERR_S("Wopi Access Check request error, query to callback ["
+                      << uri.toString() << "] failed:" << exception.what() << exception.className()
+                      << exception.name() << exception.message());
+
+            result = CheckStatus::UnspecifiedError;
+        }
+    }
```
juliushaertl, line 1135: As far as I see this only checks the connectivity, ideally it would also verify the wopi aliasgroups if the host is in the allow list and provide a proper error code if that fails

meven:
The code is now ported to use async requests, but lacks detailed error reporting, cool internal API does not surface hostname errors, connection errors.

net/HttpRequest.hpp
```diff
@@ -1623,24 +1657,35 @@ class Session final : public ProtocolHandlerInterface
...
 
     void onDisconnect() override
     {
+
+        bool completed = false;
         // Make sure the socket is disconnected and released.
         std::shared_ptr<StreamSocket> socket = _socket.lock();
         if (socket)
         {
+            completed = socket->getOutBuffer().empty() && socket->getInBuffer().empty();
```
meven, line 1674: I am using this heuristic `outbuffer.empty() && inBuffer.empty()` to determine if the request was completed, this seems brittle, suggestion welcome to improve on those.

mmeeks: Very skeptical. Can @Ashod or @caolanm look over this.

My expectation is that we don't need to deeply change the async socket code - to add complexity here. Possibly we need to add a more detailed way of fetching errors on socket connections - I can believe a single new virtual function to do that and perhaps a member to store the last error or whatever might make sense as a minimal change - but not read it. 

Separately  I would  be very surprised if an empty outgoing buffer meant anything at all. It just means some combination of kernel and/or underlying SSL layer was able to take the data.

Please adapt to whatever @Ashod recommends =)

meven: To add more context: currently the code will mark any request's response `_state` to  `Error` status in `Response::finish` by way of `HttpRequest::onDisconnect` when the socket is closed in `Socket::handlePoll`.

But most of the time a disconnection is the normal/complete case at the end of the request.
`events & POLLHUP` might be more appropriate and we'd need to differentiate the two cases of onDisconnect.

meven: HTTP/1.1 wants optionally a "Connection: close" header to signal the connection/socket will be closed.  https://datatracker.ietf.org/doc/html/rfc2616#section-14.10
We interpret this by calling the `onDisconnect()` callback in `newRequest` that will mark the response in error with `_response->finish`.

I have been testing with a HTTP/1 server `openssl s_server` and that ends up, calling `onDisconnect` from `StreamSocket::handlePoll`

meven: This is a minor concern that shouldn't affect much the feature and can be fixed separately.

vmiklos (commented):
1) CI still needs to pass

2) I see that the PR has many commits, with commit messages that suggest you would squash this before merging. But if you do so, you'll get a quite large commit. It would be nice to split this to pieces: A) the net/ infrastructure you add (e.g. add socket to the connect fail callback) B) add the new endpoint with a single detected problem C) add the rest of the problems

3) I see that all the sync poco network code is gone, good.

4) The sync poco DNS code is also gone, good.

5) For something as simple as a JSON dictionary with a single entry, I would suggest to use escapeJSONValue() and build the reply manually, no need to go via poco JSON.

meven:
>     2. I see that the PR has many commits, with commit messages that suggest you would squash this before merging. But if you do so, you'll get a quite large commit. It would be nice to split this to pieces: A) the net/ infrastructure you add (e.g. add socket to the connect fail callback) B) add the new endpoint with a single detected problem C) add the rest of the problems

Now split in three first is A, the two seconds are C (mostly optional just code improvement) and the fourth commit is B, the actual endpoint.
> 
>     5. For something as simple as a JSON dictionary with a single entry, I would suggest to use escapeJSONValue() and build the reply manually, no need to go via poco JSON.

Taken care of. Thanks.

net/HttpRequest.hpp
```diff
@@ -1380,6 +1380,11 @@ class Session final : public ProtocolHandlerInterface
         }
     }
 
+    net::asyncConnectResult connectionResult()
```
vmiklos, line 1383: Nit: a followup PR could rename this to Async, our types always start with an uppercase letter.

net/HttpRequest.hpp
```diff
@@ -1609,11 +1614,35 @@ class Session final : public ProtocolHandlerInterface
...
             }
         }
     }
 
+    void callOnConnectFail()
+    {
+        if (!_onConnectFail)
+            return;
+
+        auto self = shared_from_this();
```
vmiklos, line 1634: Nit: https://github.com/CollaboraOnline/online/blob/cb100b9c4a5a8ab93e0edbcbb624b0b68cf2fb23/README.CONTRIBUTING.md#c encourages conservative use of 'auto', you could spell out the type here, so...

net/HttpRequest.hpp
```diff
@@ -1609,11 +1614,35 @@ class Session final : public ProtocolHandlerInterface
...
+        try
+        {
+            [[maybe_unused]] const auto references = self.use_count();
+            assert(references > 1 && "Expected more than 1 reference to http::Session.");
+
+            _onConnectFail(std::static_pointer_cast<Session>(self));
```
vmiklos, line 1640: ... here it would be more clear why you need a cast. I would expect this is Session already.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -869,6 +870,8 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
                 servedSync = handleWopiDiscoveryRequest(requestDetails, socket);
             else if (requestDetails.equals(1, "capabilities"))
                 servedSync = handleCapabilitiesRequest(request, socket);
+            else if (requestDetails.equals(1, "wopiAccessCheck"))
```
vmiklos, line 873: Good, this is consistent with the existing /cool/getMetrics style.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1097,6 +1100,255 @@ bool ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+    Poco::URI uri;
+
+    std::string text(std::istreambuf_iterator<char>(message), {});
+    LOG_TRC("Wopi Access Check request text: " << text);
+
+    std::string callbackUrlStr;
+    try
+    {
+        parsingResult = jsonParser.parse(text);
+        jsonObject = parsingResult.extract<Poco::JSON::Object::Ptr>();
```
vmiklos, line 1144: Nit: you could use JsonUtil::parseJSON() instead of working with poco directly + handling exceptions.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1097,6 +1100,255 @@ bool ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+        LOG_WRN_S("Wopi Access Check request error, json object expected got ["
+                  << text << "] on request to URL: " << request.getURI() << exception.what());
+
+        HttpHelper::sendErrorAndShutdown(http::StatusCode::BadRequest, socket);
+        return false;
+    }
+
+    try
+    {
+        callbackUrlStr = jsonObject->getValue<std::string>("callbackUrl");
```
vmiklos, line 1157: Nit: JsonUtil::findJSONValue() is better to use.

net/HttpRequest.hpp
meven: self is of type `std::shared_ptr<ProtocolHandlerInterface> `, `shared_from_this` will only return super-class instance, here `ProtocolHandlerInterface`, I anc add an explicit `std::shared_ptr<Session> Session::shared_from_this` implementation.

vmiklos: Ah, fair enough. Other code does this conversion in a dedicated function, see e.g. 'client_from_this'. Fine to do that in a followup PR.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1098,6 +1101,247 @@ bool ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+    assert(socket && "Must have a valid socket");
+
+    LOG_DBG("Wopi Access Check request: " << request.getURI());
+
+    Poco::MemoryInputStream startmessage(&socket->getInBuffer()[0], socket->getInBuffer().size());
+    StreamSocket::MessageMap map;
+
+    // Poco::JSON
+    Poco::JSON::Parser jsonParser;
+    Poco::Dynamic::Var parsingResult;
```
vmiklos, line 1134: Nit: I believe jsonParser and parsingResult is now unused.

vmiklos (commented):
OK, CI is green, that's promising.

wsd/ClientRequestDispatcher.cpp
meven: and also `Poco::URI uri;`

mmeeks: We are trying to remove Poco:: dependencies incrementally in new code of course so ... ;-)

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1096,6 +1101,247 @@ bool ClientRequestDispatcher::handleWopiDiscoveryRequest(
...
+            status = CheckStatus::CertificateValidation;
+
+            LOG_DBG("Result ssl: " << probeSession->getSslVerifyMessage());
+        }
+
+        sendResult(status);
+    };
+
+    httpProbeSession->setFinishedHandler(std::move(finishHandler));
+    httpProbeSession->asyncRequest(httpRequest, *COOLWSD::getWebServerPoll());
```
mmeeks, line 1340: Good this is now async.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1096,6 +1101,247 @@ bool ClientRequestDispatcher::handleWopiDiscoveryRequest(
     return true;
 }
 
+STATE_ENUM(CheckStatus,
```
mmeeks, line 1104: Lets have a comment here stating: "/// NB. These names are part of the published API - and should not be renamed or altered but can be expanded"

mmeeks (approved):
Ok - this is a lot better, looks async - which is nice ;-) and I guess will really help if we can get the Nextcloud (and other integration) side working nicely - with some SDK documentation for the test end-point and so on :-)

There is a follow-on fix there that would be good to have.

Thanks Meven, Miklos & all.

wsd/ClientRequestDispatcher.cpp
eszkadev: with clang  it stops the build due to unused functions:
```
wsd/ClientRequestDispatcher.cpp:1106:1: error: unused function 'toStringShort' [-Werror,-Wunused-function]
STATE_ENUM(CheckStatus,
^
./common/StateEnum.hpp:63:31: note: expanded from macro 'STATE_ENUM'
    static inline std::string toStringShort(NAME e)                                                \
                              ^
wsd/ClientRequestDispatcher.cpp:1106:1: error: unused function 'toString' [-Werror,-Wunused-function]
./common/StateEnum.hpp:76:31: note: expanded from macro 'STATE_ENUM'
    static inline std::string toString(NAME e)                                                     \
                              ^
wsd/ClientRequestDispatcher.cpp:1106:1: error: unused variable 'CheckStatusMax' [-Werror,-Wunused-const-variable]
./common/StateEnum.hpp:80:25: note: expanded from macro 'STATE_ENUM'
    static const size_t NAME##Max = COUNT_ARGS(__VA_ARGS__);                                       \
                        ^
<scratch space>:107:1: note: expanded from here
CheckStatusMax
^
3 errors generated.

```

eszkadev: fix: https://github.com/CollaboraOnline/online/pull/10785

juliusknorr:
Very nice to see this merged, thanks a lot.

I started a first draft PR to integrate with richdocuments (https://github.com/nextcloud/richdocuments/pull/4361). 

One thing I noticed would be that we have not reliable way to detect if the collabora server supports this check endpoint. Can we maybe expose it in /hosting/discovery similar to how it is done for capabilities:

Screenshot 2024-12-27 at 15 49 59

mmeeks:
@meven can you implement that and get it into green - it should be the work of a few seconds. Thanks ! =)

vmiklos:
FWIW I see that online.git commit 0026f62181872ed50ef01d457a4c54d0be38140f implements this now.

PR #9207, opened 2024-06-03
browser: toolbar: avoid forcing to clear references

hcvcastro:
1) press F2 to edit a formula cell with defined range
2) The reference it is marked and painted
3) The formula bar lost focus and force to clear the mark

The references will be cleared due to response from server message.

Change-Id: I32fbacc8d846cf8ba8b3371f3386d07e9cebe470


eszkadev (approved):
looks ok,

PR #9208, opened 2024-06-03
Use Workers to do tile decompression and unpremultiply

Cwiiis:
* Resolves: # 
* Target version: master 

### Summary

Two patches; the first fixes a bug where if we receive new keyframe tiledata and need to ensure the canvas, we end up decompressing/unpremultiplying/putting the old data unnecessarily (I think). The second uses a worker to offload the majority of tile image data decompression and unpremultiplying from the main thread and improves response during large update operations (such as scrolling through a document).

The code is not properly formatted yet, I'm opening this PR now and will update it with fixes once I figure out what those fixes are...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -12,7 +12,7 @@
  * L.Socket contains methods for the communication with the server
  */
 
-/* global app JSDialog _ $ errorMessages Uint8Array brandProductName */
+/* global app JSDialog _ $ errorMessages Uint8Array brandProductName Map */
 
 app.definitions.Socket = L.Class.extend({
 	ProtocolVersionNumber: '0.1',
@@ -41,6 +41,24 @@ app.definitions.Socket = L.Class.extend({
 		this._msgQueue = [];
 		this._delayedMessages = [];
 		this._handlingDelayedMessages = false;
+		if (window.Worker) {
+			window.app.console.info('Creating SocketWorker');
+			this._worker = new Worker('src/core/SocketWorker.js');
+			this._worker.addEventListener('message', (e) => this._onWorkerMessage(e));
+			this._workerTimer = null;
+			this._workerPendingEvents = [];
+			this._workerMap = new Map();
+			this._workerId = 0;
+
+			const bufferSize = window.tileSize * window.tileSize * 4 * 4;
+			const nBuffers = 128;
+
+			this._workerBuffers = [];
+			this._workerBuffersPendingSlurp = [];
+			for (var i = 0; i < nBuffers; ++i) {
+				this._workerBuffers.push(new Uint8Array(bufferSize));
+			}
+		}
 	},
 
 	getWebSocketBaseURI: function(map) {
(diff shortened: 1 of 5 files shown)


browser/src/core/SocketWorker.js
```diff
@@ -0,0 +1,112 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/* eslint-disable no-inner-declarations */
+/* eslint no-unused-vars: ["warn", { "argsIgnorePattern": "^_" }] */
+/* global importScripts Uint8Array cool */
+
+if ('undefined' === typeof window) {
+	importScripts('../../node_modules/fzstd/umd/index.js');
+	importScripts('../layer/tile/CanvasTileUtils.js');
```
mmeeks, line 18: I suspect that this has not been tested with a bundlified product build (?) or do we distribute CanvasTileUtils.js on its own ?

browser/src/layer/tile/CanvasTileUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+			byteLength: number,
+			byteOffset: number = 0,
+		): void {
+			for (var i = byteOffset; i < byteOffset + byteLength; i += 4) {
+				// premultiplied rgba -> unpremultiplied rgba
+				var alpha = data[i + 3];
+				if (alpha === 0) {
+					data.fill(0, i, i + 3);
+				} else if (alpha !== 255) {
+					data[i] = Math.ceil((data[i] * 255) / alpha);
```
mmeeks, line 25: We should probably use a 64k byte lookup table here as some of the core LibreOffice code does =)

browser/src/layer/tile/CanvasTileUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+				var pixOffset = y * width * 4 + outputOffset;
+				var pixSrc = offset;
+
+				for (var x = 0; x < width; ++x) {
+					if (bitToCheck > 128) {
+						bitToCheck = 1;
+						lastMask = data[rleMaskOffset++];
+					}
+					if (!(lastMask & bitToCheck)) {
+						this.lastPixel.set(data.subarray(pixSrc, pixSrc + 4));
```
mmeeks, line 69: I thought we decided that subarray was unfeasibly slow on firefox, and we didn't want to over-use it ?

browser/src/core/Socket.js
```diff
@@ -459,7 +484,14 @@ app.definitions.Socket = L.Class.extend({
 	// process so - slurp and then emit at idle - its faster to delay!
 	_slurpMessage: function(e) {
 		this._extractTextImg(e);
+		if (e.image && e.image.rawData && this._worker) {
+			this._queueWorkerTileProcessing(e);
```
mmeeks, line 488: Am I reading this right that we take 'tile:' messages only - and we potentially then emit them out of order with subsequent 'delta:' and 'update:' messages which could arrive before the tiles are de-compressed by the thread - that is likely to lead to corrupted pixels.
Hmm.

mmeeks (changes requested):
Ok - this needs more work; I don't think we can merge it as-is I'm afraid.

Cwiiis:
I'm going to close this as I have a newer, better version of the patch in a different branch. I'll likely need advice resolving some of these comments - I'll try to flag these issues explicitly.

PR #9221, opened 2024-06-05
cool#9219 kit: fix setclipboard command with HTML payload

vmiklos:
Have two COOL servers, copy on one instance, paste an on other instance,
nothing happens on Ctrl-V.

It appears the JS code in Clipboard.js
_dataTransferDownloadAndPasteAsync() gets the correct HTML, but when it
tries to contact the "copy" server for the underlying full clipboard,
some security policy prevents the JS code from doing that. Which means
1) perhaps we should do that HTTP fetch on the server, not in JS and 2)
the fallback code should at least work via normal HTML.

Fix 2) by extending ChildSession::setClipboard() to recognize if the
data has the typical HTML header, and only start the usual parsing in
case the data we got is not bare HTML.

The 1) still needs fixing.

Change-Id: I55a102c2a39d21ac898f82041760a220e46784cb

### DIFF
--- a/test/UnitCopyPaste.cpp
+++ b/test/UnitCopyPaste.cpp
@@ -363,6 +363,17 @@ public:
                                   existing + newcontent + '\n'))
             return;
 
+        // Test setting HTML clipboard:
+        std::string html("<!DOCTYPE html><html><body>myword</body></html>");
+        // Intentionally no buildClipboardText() here, just raw HTML.
+        if (!setClipboard(_clipURI, html, HTTPResponse::HTTP_OK))
+            return;
+        // This failed with: ERROR: Forced failure: Missing clipboard mime type, because we tried to
+        // parse HTML when we expected a list of mimetype-size-bytes entries.
+        if (!fetchClipboardAssert(_clipURI, "text/html", html))
+            return;
+
+        // Setup state that will be also asserte in postCloseTest():
         LOG_TST("Setup clipboards:");
         if (!setClipboard(_clipURI2, buildClipboardText("kippers"), HTTPResponse::HTTP_OK))
             return;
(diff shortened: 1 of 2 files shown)


test/UnitCopyPaste.cpp
```diff
@@ -374,6 +374,15 @@ class UnitCopyPaste : public UnitWSD
         if (!fetchClipboardAssert(_clipURI, "text/plain;charset=utf-8", "herring"))
             return;
 
+        // Test setting HTML clipboard:
+        std::string html("<!DOCTYPE html><html><body>myword</body></html>");
+        if (!setClipboard(_clipURI, buildClipboardText(html, "text/html"), HTTPResponse::HTTP_OK))
```
vmiklos, line 379: Ah, if I use buildClipboardText(), then the payload will be in the mimetype-size-bytes format, the test should avoid that to actually cover the new functionality.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1902/console fails in unit-wopi-languages, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1903/console fails in unit-save-torture, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/1908/console now fails in unit-wopi-fail-upload

vmiklos:
Aha, rebase on top of f5bfde5e40be5bb8525c4ca79b1e171c1a9fb3b6 helped

vmiklos:

This is the fallback part; the "possibly move http fetch to server" part first needs bisecting so we understand the problem better, I guess.

caolanm (approved):

PR #9225, opened 2024-06-06
.uno:BackColor is deprecated

timar:
See https://cgit.freedesktop.org/libreoffice/core/commit/?h=distro/collabora/co-24.04&id=be3d34b5d6b97c3eb12ab3f84ce2da2ef965a928

Change-Id: I0e33d6b9e3e38ccbfcb95f6703cf0395d4280842


timar:
I don't understand why this change causes cypress test failure (after several retries, so it look suspicious). 
Here is a screenshot from cypress, I don't see the problem.
image

timar:
https://gerrit.libreoffice.org/c/core/+/168519 is needed to pass 'COOL 24.04 unit tests w/o cypress'.

timar:
> I don't understand why this change causes mobile cypress test failure (after several retries, so it looks suspicious). Here is a screenshot from cypress, I don't see the problem. image

Scratch this, strings changed, I had to modify the cypress test.

eszkadev (approved):

PR #9228, opened 2024-06-06
fix(contextMenus): move above UI elements

Minion3665:
On small window sizes, a context menu may be forced to open in such a way that it goes over the same region as the notebookbar or spreadsheet tabs.

In that case, the context menu should be above the other element, or some of the context menu options could be unreachable.

Change-Id: Id269e3fb9bc2251677a0792d83fc51bd56b9cb21


eszkadev (approved):
looks good

PR #9229, opened 2024-06-06
Don't show audit snackbar if all entries are all 'ok' state

caolanm:
I see this on a local nextcloud where the audit dialog only has one entry which is "ok".

Change-Id: I55f12125cbc0ca4c88c9220ed114d455e84639e0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks great to me =)

PR #9230, opened 2024-06-06
fix(mobileWizard): Always start at the top

Minion3665:
Previously, when entering or exiting mobile wizard sublevels you would start scrolled down to the place you were before. For example, if you scrolled down a few items on the main context menu of an image in order to reach "arrange", you would always end up right at the bottom of the "arrange" menu.

By always scrolling up to the top, we can prevent you from ever having content hidden above you.

Change-Id: I02365ecef859fb6f199d14a12a49937d5a30b8e7


gokaysatir (approved):
Setting the lower level scroll to 0 seems to be what we need.

I think ideal solution for upper levels, would be to remember "upper level scroll position" and set it back when user goes one level up.

But, since this solves the issue, let me approve.

Thanks :)

PR #9234, opened 2024-06-07
cool#9219 clipboard: support URLs in the payload of the setclipboard command

vmiklos:
Copying the full clipboard (including ODF) from a remote server into the
JS client failed with:

Refused to connect to '...' because it violates the following Content Security Policy directive: "connect-src ..."

If this happens, still send the original URL to the server, so it can
download the full clipboard, which gives better result for pasting,
compared to going via HTML only.

Not done in this commit: 1) the JS still tries to get the clipboard
itself, we download on the server only on failure and 2) the server does
a blocking HTTP download, which may not be wanted.

Change-Id: I9ed8c8050f9dbe71dc8fcd36c9d8fa6e8bba44f6

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -319,6 +319,7 @@ void ClientSession::handleClipboardRequest(DocumentBroker::ClipboardRequest     
         {
             preProcessSetClipboardPayload(*data);
 
+#if !MOBILEAPP
             if (data->starts_with('{'))
             {
                 // We got JSON, extract the URL and the UNO command name.
@@ -357,6 +358,7 @@ void ClientSession::handleClipboardRequest(DocumentBroker::ClipboardRequest     
                 }
             }
             else
+#endif
             {
                 // List of mimetype-size-data tuples, pass that over as-is.
                 docBroker->forwardToChild(client_from_this(), "setclipboard\n" + *data, true);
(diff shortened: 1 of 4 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_multi_user/1254/ e.g. the multi-user calc/annotation_spec.js fails, but `make -C cypress_test check-multi spec=calc/annotation_spec.js` passes locally.

vmiklos:
@caolanm could you please review this at some stage? Thanks.

If http fetch in the browser fails, send the URL to the server and fetch there, as that's a less locked down environment. Can be tested with a remote staging COOL vs local make run, copy on remote, paste on local. ODF paste is ~perfect if the copied content has e.g. a footnote and looks worse when we go via HTML.

wsd/ClientSession.cpp
```diff
@@ -318,7 +318,59 @@ void ClientSession::handleClipboardRequest(DocumentBroker::ClipboardRequest
...
+                        const std::shared_ptr<const http::Response> httpResponse =
+                            session->response();
+                        if (httpResponse->statusLine().statusCode() != http::StatusCode::OK)
+                        {
+                            LOG_ERR("Clipboard download request failed");
+                            return;
+                        }
+
+                        std::string body = httpResponse->getBody();
+                        docBroker->forwardToChild(client_from_this(), "setclipboard\n" + body,
```
vmiklos, line 346: I'll make this more strict to require that the body is not a random format, but to some extent it's in the mimetype-length-data tuples format we expect here.

caolanm (approved):
lets merge work to date anyway

PR #9236, opened 2024-06-07
fix(comments): avoid comment collapse during edit

Minion3665:
This is particularly relevant on touch devices, where some events can be triggered a bit differently to desktop. Previously it was sometimes hard to reply to or edit a comment as upon hitting the modify/reply buttons we would consider the comment deselected and collapse it.

The easiest way I have found to deal with this is to prevent the collapse during an edit

Change-Id: I740b1c260e9ee63bac6e667f99ed6da141d71a9c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -200,7 +200,20 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		}
 	}
 
+	private isEditing(): boolean {
+		const sections = this.containerObject.sections;
+		const textBoxes = sections
+			.flatMap((section: any) => [section.sectionProperties.nodeModifyText, section.sectionProperties.nodeReplyText])
+			.filter((textBox: any) => textBox !== undefined);
+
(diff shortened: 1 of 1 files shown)


gokaysatir (commented):
I think there is an "isAnyEdit" function for checking if any comment is in edit mode.

After finishing the editing, comments don't collapse. So collapsing is not delayed but cancelled with this commit i guess?

It may be better to cancel the collapsing for only the edited comment. Then we need to make sure it is collapsed after editing is done.

Minion3665:
> @gokaysatir:
> I think there is an "isAnyEdit" function for checking if any comment is in edit mode.

Correct, although I believe that's any edited comment in any group, not just a comment in the group you're currently editing

> After finishing the editing, comments don't collapse. So collapsing is not delayed but cancelled with this commit i guess?

This is not the behavior I'm seeing on iPadOS, it's definitely unintentional... are you able to provide me with some more details (e.g. device) to help me reproduce that?

Thanks!

PR #9237, opened 2024-06-07
wsd: test: unit-test validation in debug only

Ashod:
Unit-tests are only ever run in debug builds by
desing (there is check in test/run_unit.sh to
build with --enable-debug to run the tests).

In our release buids, we have seen cases where
the process exists with an error code because
the unit-test logic thinks it ran a test and failed.
It also generates error log and other log entries
when initializing and uninitializing the unit-
test logic.

This excludes unit-test logs and the failure
check in non-debug builds to avoid such errors
and reduce noise.

Change-Id: I9f472172b72770551a491ff6e19ab9b997fcf365


common/Unit.cpp
```diff
@@ -388,7 +388,11 @@ int UnitBase::uninit()
         dlclose(DlHandle);
     DlHandle = nullptr;
 
+#ifndef NDEBUG
```
mmeeks, line 391: surely ENABLE_DEBUG ? wow - this is broken:

Makefile.am-if !ENABLE_DEBUG
Makefile.am:AM_CPPFLAGS += -DNDEBUG
Makefile.am-endif

Why do we add a weird synonym of another conditional with a different name; NDEBUG looks un-necessary and grim.

common/Unit.cpp
```diff
@@ -388,7 +388,11 @@ int UnitBase::uninit()
...
     return GlobalResult == TestResult::Ok ? EX_OK : EX_SOFTWARE;
+#else // !NDEBUG
+    return EX_OK; // Always success in release.
```
mmeeks, line 394: Can we return much earlier here - the dlopen thing is potentially regression / risky - we should never dlopen if we're not in debug mode ?

mmeeks (approved):
Looks ok; lets kill NDEBUG in its entirety though - and if we can expand the scope that woudl be better. I wonder what made this regress though.

common/Unit.cpp
Ashod: Done. Excluded `dlopen` and `dlclose` altogether.

common/Unit.cpp
```diff
@@ -352,6 +356,8 @@ void UnitBase::rememberInstance(UnitType type, UnitBase* instance)
 
 int UnitBase::uninit()
 {
+    // Only in debug builds do we support tests.
+#ifdef ENABLE_DEBUG
```
mmeeks, line 360: #if ENABLE_DEBUG - it is always defined =)

mmeeks (changes requested):
we have to have #if ENABLE_DEBUG everywhere- ENABLE_DEBUG is unconditionally defined either to 0 or 1 as is best-practice for compile conditionals; #ifdef should not be used.

Ashod:
> we have to have #if ENABLE_DEBUG everywhere- ENABLE_DEBUG is unconditionally defined either to 0 or 1 as is best-practice for compile conditionals; #ifdef should not be used.

Of course! Mechanical replace. Fixed.

caolanm:
https://cpci.cbg.collabora.co.uk:8080/job/github_linux_release_co-24.04/1903/console

tools/COOLWebSocket.hpp:55:41: error: unused parameter 'testname' [-Werror=unused-parameter]
tools/COOLWebSocket.hpp:103:38: error: unused parameter 'testname' [-Werror=unused-parameter]
tools/COOLWebSocket.hpp:131:38: error: unused parameter 'testname' [-Werror=unused-parameter]

I guess [[maybe_unused]] on those is what we want now TST_LOG_NAME is no op ? or (void)NAME; in TST_LOG_NAME?

caolanm (approved):

PR #9239, opened 2024-06-10
fix(calc): Prevent following view jumps

Minion3665:
Previously we were mixing up lat/lng, which made calc jump to unexpected places when the person we're following's cursor is offscreen.

Change-Id: I4db4c93118fbc3ffd04c7152652ccd2b26209b5d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
good catch!

PR #9243, opened 2024-06-10
fix: sometimes userlist is empty on reconnect

Rash419:
- before setting the viewId we check for _documentInfo in onStatusMsg, it might happen that on reconnect documentInfo remains same.
 - steps to reproduce:
   1. Open writer document 4 times in different tabs
   2. Kill the server (Ctrl + C)
   3. Run the server again and wait of clients to reconnect
   4. Check all the tabs, there must one tab where you should see the userlist empty. If not, kill and run server again...
- this patch fixes it by reseting the _documentInfo on socket disconnection

Change-Id: I7be045f2b2433b9e12b6749fefbdd53ab5f0a0cd

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1608,6 +1608,7 @@ app.definitions.Socket = L.Class.extend({
 
 			this._map._docLayer._resetCanonicalIdStatus();
 			this._map._docLayer._resetViewId();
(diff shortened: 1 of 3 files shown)


eszkadev:
please rebase, we fixed ci

eszkadev (approved):
looks ok

PR #9245, opened 2024-06-10
browser: fix PageDown cursor position

hcvcastro:
Scroll the new position if the cursor is not
visible.

Change-Id: Ib5a77fefa5ec16defa0695806c770f21beec4243


caolanm (approved):
Seems reasonable

PR #9246, opened 2024-06-10
Private/caolan/resize observer warning

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
@gokaysatir in https://github.com/CollaboraOnline/online/pull/2881 apparently there was 'ResizeObserver loop limit exceeded' as a cypress failure, so presumably in 2021 there was some some problem in cypress with a ResizeObserver. Any memory of what that problem might have been?

I'm wondering because there are occasional sightings of ResizeObserver warnings in deployment logs and I wonder if there is any connection.

hcvcastro (approved):

PR #9259, opened 2024-06-12
Show server audit if deprecated is_admin is used

eszkadev:
use optional isAdminUser to detect when it is missing


caolanm (approved):
looks reasonable to me

PR #9282, opened 2024-06-13
wsd: add 'wopi_host_id' to about dialog

hcvcastro:
Change-Id: I04004f416e7bc6cb2c265fe3ebc08f2c250e8009


mmeeks (approved):
Thanks Henry - great work.
I think some of our partners are not hyper-eager for us to expose this in the UI by default; can you show it only in our tripple click / debugging mode ? =)
But of course for feedback reporting it is invaluable =)
Fine to do the version hide/show in a follow-on =) Thanks!

hcvcastro:
Thanks, I will do next commit

hcvcastro:
oh, it is not rebased, I will modify now =)

mmeeks (approved):
Looks sensible; thanks =)

PR #9299, opened 2024-06-17
invalidatetiles not logged like the other incoming messages

caolanm:
its skipped because its filtered on slurp
add explicit logging for it in its handler

_logSocket is currently only called in two places that both check the same debugging flags so move that check inside _logSocket.

Change-Id: Icacd8b99b4aaf5feaa468f89b3273101d6722ca6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -176,11 +176,9 @@ app.definitions.Socket = L.Class.extend({
 	},
 
 	_doSend: function(msg) {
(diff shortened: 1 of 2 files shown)


caolanm:
invalidatetiles isn't logged to console, presumably since the slurp filtering, and I was confused by the apparent lack of invalidations getting to the browser

mmeeks (approved):
Looks great to me; thanks Caolan =) should prolly put it in the CODE branch too.

PR #9300, opened 2024-06-17
Private/caolan/make header invalidation test reliable

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
a) we're double-fetching via prefetch the fractional rows that lie part way into the viewport that are already queried because we have to get the full tiles anyway if any part is inside the viewport, that can give empty updates when the prefetch timer happens to fire

b) the header/footer indicators are close enough in the original example doc that they lie in the same tile so it depends on timing if their invalidations are merged together or managed as one rendered tile for the first one + an empty update for the 2nd one, just made the header bigger to test what we're actually interested in for this example

vmiklos (approved):
Looks reasonable to me, thanks.

PR #9304, opened 2024-06-18
Context menu fixes

Minion3665:
This is a trivial backport of my context menu fixes, #9228 and #9230.

I didn't include #9277, as it is not yet merged.

---

On small window sizes, a context menu may be forced to open in such a
way that it goes over the same region as the notebookbar or spreadsheet
tabs.

In that case, the context menu should be above the other element, or
some of the context menu options could be unreachable.

Change-Id: Id269e3fb9bc2251677a0792d83fc51bd56b9cb21

---

Previously, when entering or exiting mobile wizard sublevels you would
start scrolled down to the place you were before. For example, if you
scrolled down a few items on the main context menu of an image in order
to reach "arrange", you would always end up right at the bottom of the
"arrange" menu.

By always scrolling up to the top, we can prevent you from ever having
content hidden above you.

Change-Id: I02365ecef859fb6f199d14a12a49937d5a30b8e7

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -159,6 +159,7 @@ L.Control.ContextMenu = L.Control.extend({
 				selector: '.leaflet-layer',
 				className: 'cool-font',
 				trigger: 'none',
+				zIndex: 1500,
 				build: function() {
 					return {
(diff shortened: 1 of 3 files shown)


eszkadev (approved):
looks good

PR #9307, opened 2024-06-18
Fix failing test multiuser/writer/annotation_spec.js

neilguertincollabora:
Increased wait before inserting comment
Wait was added in a5e6a81cddb9084dc6189901787ebf96324bc1d2

Change-Id: Ia2aa4c901fc57d6700b43c7e0db0380055877081


vmiklos (approved):
I don't find it elegant to increase the timeout, but the test now does pass for me, so that's nice.

The "COOL 24.04 Cypress (multi-user)" job is also green, while it's red at at e.g. #9340 so let's get this in.

PR #9308, opened 2024-06-18
Use dedicated signal for version bar initialization

eszkadev:
It will look like this in the browser logs:
- IN viewinfo: ...
- IN adminuser: ...

and if we are the admin:
- OUT versionbar
- IN versionbar: ... < latest version ... if needed show the message

Thanks to that snackbar will be shown on the load, not when next user joins.


eszkadev:
Just a small update for better events ordering

hcvcastro (approved):

hcvcastro:
Looks good

PR #9310, opened 2024-06-19
Fix video playback with load balancer (master)

eszkadev:
It's important to have correct WOPISrc for load balancers. We need to decode "&" before other params like ServerId and Tag so load balancer will not use it as a part of WOPISrc.

In other case when I had 2 instances of COOL behind HAProxy sometimes request to get the video was forwarded to the wrong instance.


Rash419 (approved):
I get 200 response from server but for some reason I can't click on the play. It might not be related to this patch
image

eszkadev:
It is probably related to the new canvas code (changed handles to modify shapes?)

PR #9311, opened 2024-06-19
Fix video playback with load balancer (23.05)

eszkadev:
It's important to have correct WOPISrc for load balancers. We need to decode "&" before other params like ServerId and Tag so load balancer will not use it as a part of WOPISrc.

In other case when I had 2 instances of COOL behind HAProxy sometimes request to get the video was forwarded to the wrong instance.


Rash419 (approved):
Thanks

PR #9312, opened 2024-06-19
tests: try to catch endless loop waiting for modified status.

mmeeks:
Change-Id: I7d925e4703cdf9f252b1b3ffcd6716e4b3c87235

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitSaveTorture.cpp
+++ b/test/UnitSaveTorture.cpp
@@ -117,10 +117,11 @@ namespace {
         std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
         while (true)
         {
-            if (std::chrono::steady_clock::now() - start > timeout)
+            if (std::chrono::duration_cast<std::chrono::seconds>(
+                    std::chrono::steady_clock::now() - start).count() > timeout.count())
             {
                 LOK_ASSERT_FAIL("Timed out waiting for modified status change");
-                break;
+                return false; // arbitrary but why not.
             }
             std::vector<char> message
                 = wsSession->waitForMessage("statechanged:", timeout, name);
@@ -173,7 +174,7 @@ void UnitSaveTorture::testModified()
 
 void UnitSaveTorture::testTileCombineRace()
 {
-    std::string name = "testModified";
+    std::string name = "testTileCombineRace";
     std::string docName = "empty.ods";
 
     std::string documentPath, documentURL;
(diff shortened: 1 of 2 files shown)


mmeeks:
Again - not compiled, nevermind run - but if we need something urgently ... =)

vmiklos (approved):
My understanding is that test/Makefile.am defines LOK_ABORT_ON_ASSERTION unconditionally:

https://github.com/collaboraonline/online/blob/c3539a6a3fc957343bacfb5e94ebe4a69db817a5/test/Makefile.am#L118

and then once LOK_ABORT_ON_ASSERTION is defined, LOK_ASSERT means an assert() call, so the code you add here is not reachable:

https://github.com/collaboraonline/online/blob/c3539a6a3fc957343bacfb5e94ebe4a69db817a5/test/lokassert.hpp#L86

at the same time, I guess we have the LOK_ABORT_ON_ASSERTION condition for a reason, so in theory it makes sense to not assume that a failing LOK_ASSERT() is terminating the process.

The code builds & unit-save-torture passes with this.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1410/console failed in e.g. impress/slide_operations_spec.js, but it looks just unstable, given that only C++ test code changed, retry.

mmeeks:
Its far from clear how we are looping here then =) Is it possible that we run the tests in a non-debug build; where assert(foo); is compiled out and doesn't in fact fail somehow ?

vmiklos:
That would be a possibility, but reading https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/2322/console, we have these flags:

> checking whether to enable debug build... yes
> ...
>     C++ compiler flags        -g -O2 -Wall -Wextra -Wshadow -Wundef -Werror  -grecord-gcc-switches -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -fstack-clash-protection -fcf-protection -std=c++20

and 'man assert' says:

> If the macro NDEBUG is defined at the moment <assert.h> was last included, the macro assert() generates no code

and our Makefile.am has this:

https://github.com/CollaboraOnline/online/blob/967b85ec12ecb412070275bf0589819c3460707c/Makefile.am#L84-L86

so a brief code reading suggests our CI properly does not define NDEBUG, so we should assert. Perhaps the trouble would be our own SIGABRT handling code that should be disabled while running tests? I'm not sure.

PR #9323, opened 2024-06-21
Handle ".uno:DocumentStatus" instead of SlideMasterPage special-handling

mikekaganski:
Since other view modes are needed, too, not only master page view, we need
to send / handle status updates not only in case of SlideMasterPage update.
The new LOK_CALLBACK_STATUS_UPDATE serves for that.

Change-Id: If7bc7ae124d161658b06784f3e86d85f3c76f3c6


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/2490/console failed in unit-save-torture, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1532/console failed in impress/annotation_spec.js and impress/slide_operations_spec.js, retry.

test/UnitInsertDelete.cpp
```diff
@@ -52,7 +52,7 @@ void getPartHashCodes(const std::string& testname, const std::string& response,
     TST_LOG("Status reports " << totalParts << " parts.");
 
     Poco::RegularExpression endLine("[^\n\r]+");
-    Poco::RegularExpression number("^[0-9]+$");
+    Poco::RegularExpression number("^-?[0-9]+$");
```
eszkadev, line 55: this is not needed, it will be fixed soon: https://gerrit.libreoffice.org/c/core/+/169691

eszkadev (changes requested):
not needed change

test/UnitInsertDelete.cpp
mikekaganski: I don't see why. The change didn't make it unsigned, as far as I see.

eszkadev: Ah, right. I looked at getId(), not getUniqueId() -> needs to be changed

mikekaganski: So please let's not block on this. If a future change makes them unsigned, this test may be reverted to the previous state, maybe with an explicit comment requiring them to be unsigned.

eszkadev (approved):
ok, so I will update that after gerrit change will go in

PR #9326, opened 2024-06-21
Fix failing test mobile/writer/table_properties_spec.js

neilguertincollabora:
typeIntoInput was failing. Solution: Wait for the field to fully load before typing

Change-Id: I9c9fc50a1bfa3a579278bfaaf1a5a5e43b5f9f02


vmiklos (approved):
cy.cGet() instead of an arbitrary wait looks OK to me, thanks.

vmiklos:
`make -C cypress_test check-mobile spec=writer/table_properties_spec.js` passes for me and the above result doesn't even contain the 'check-mobile' result, let's get this in. Thanks.

PR #9332, opened 2024-06-24
Fix: Cursor shrinks on 200% browser zoom on every cell edit

eszkadev:


mmeeks:
Would be great to get into the stable 24.04 too if we can - its deeply annoying =)

gokaysatir (approved):

eszkadev:
cc @timar

PR #9335, opened 2024-06-24
Scroll to cursor only if we want to follow ourself

eszkadev:
This is followup to the commit d3ed75d0a2f7adecf682b9fd4803f7b734bdb437 User following: don't mark myself as followed

- detection of "visible cursor" both for cell cursor and text cursor
- only scroll to the cursor if we really follow ourself (when not Off)


gokaysatir (approved):
Thanks for the fix.

PR #9339, opened 2024-06-24
compile fixes

hcvcastro:
- **wsd: fix unused parameter ‘subjectHash’**
- **wsd: fix ‘ssl’ has not been declared**

PR #9343, opened 2024-06-25
fix: Remove build error when building without ssl

Minion3665:
Previously getSslCert took a parameter that was only used when SSL was
enabled, causing an unused parameter when SSL was disabled. As we have
-Werror=unused-parameter, this caused the build to fail

Similarly, ClientSession::loadDocument used ssl:: when it couldn't be
defined because of a lack of SSL support in the build

Change-Id: Ie0b4a5f3531a8fb6aba66f352ba109fbffd7687a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1276,12 +1276,15 @@ public:
 #endif
     }
 
+#if ENABLE_SSL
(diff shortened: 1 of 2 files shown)


timar:
Prefer #9339  because it came in first.

PR #9345, opened 2024-06-25
Private/gokay/views graphic selections

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
nice, tested on iPad, but found some issues:
- shape preview doesn't follow what will be the end result (markers show real end-size, but preview can be outside)
- when editing chart and selecintg some bar - it shows rotation handle which shouldn't be there
- while dragging the fontwork or shape - page is also moved in Writer

eszkadev (approved):

PR #9346, opened 2024-06-25
Make annotation icon partially transparent

quikee:
There is now an annotation object associated with an annotation, which is rendered in core (like other objects), so make the annotation icon partially transparent to make the annotation object visible.

Change-Id: I11fe2ce0fac005ba9b79bc474a71f79fbf09baa2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -42,6 +42,7 @@
 
 	/* Ruler */
 	--ruler-height: 20px;
+	--canvas-container-y: var(--ruler-height);
 }
 .focus-hidden:focus {
 	outline: none;
--- a/browser/css/leaflet.css
+++ b/browser/css/leaflet.css
@@ -882,7 +882,7 @@ input.clipboard {
 	margin: 0px; /*Ruler styling*/
 }
 .cool-ruler {
-	background-color: var(--color-main-background);
+	background-color: transparent;
 	height: var(--ruler-height);
 	width: 100vw;
 	margin: 0px !important;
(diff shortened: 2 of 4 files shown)


eszkadev:
cc: @pedropintosilva This doesn't look good in my opinion:

annotation

I think our policy is to send information to the online and draw "markers" on the canvas ourself so we can style them, handle click events properly, etc.
Now we have duplicated markers. Can we just not draw markers in the core as it was before?

pedropintosilva:
I would prefer to avoid transparency because I fear that if 2 or more markers (be that with the same icon or different) are stacked up, the transparency will not help. Plus, the problems we take in with contrast (the color behind and if the document is being viewed with inverted colors or not).

## For now

### What are they

I'm trying to understand what is the purpose of those additional new markers. It looks like they are referent to the user name that did them? If that's so, and since they already come with that additional text and different color: wouldn't it be better to:
* Just render one icon per annotation (note.svg)
* Serve that note.svg in the html structure itself (right now is being set via css background-image property)
  * So, we can then set its fill color via JS according to what the core says (that color that is visible in the new marker)

### Behavior
I have seen some weird behavior related to positioning. When dragging an annotation: it seems that sometimes that new marker is left in a previous position:

image

----

## Ideally (general notes, long term and not specific related to this)

It would be great if we could use the same user pattern that we use for comments by just have one icon when multiple annotations are in the same spot1 with the number of annotations present. When clicked the list of all annotations in that spot would be revealed. So:
* Loading a document:
  * 2 annotations or more that have the same icon on core -> render in online with only one icon + total number of annotations 
  * 2 annotations or more that have different icon from core -> render in online with only one **different/new**2 icon + total number of annotations 
* Insert new annotation with Collabora Online above or very near to another annotation:
  * Works as before but once that comment/annotation is saved the multiple icons are replaced with only one icon + total of annotation indicator as in the previous situation


----

1. x,y + safety margin, so they if if they are not in the exact coordinates it would count as if they were)
2. This new icon would be a generic/group icon that help the user to understand that we are in presence of a spot with different annotation "types"

PR #9340, opened 2024-06-24
Fix mobile build breakage

plubius:
The iOS and Android platforms run the LibreOffice backend in the client process so there is no SSL sockets. Instead, all client and server communication is done through a buffer shared between threads.

PR #9348, opened 2024-06-25
Fix mobile app builds

timar:
Change-Id: Iacc79a7a25cc35fb7ca9067cb0308528cb4da055

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks OK.

Also note that @plubius has a branch at https://github.com/CollaboraOnline/online/tree/fixiosbuildbreakage, with similar changes, though this one seems to be fixing also kit/Kit.cpp, which is not done there.

timar:
Prefer #9340

PR #9351, opened 2024-06-26
Ruler: Clip it, don't spill over other UI elements

pedropintosilva:
Before this commit the vertical ruler when the page is zoomed in would
expand beyond its parent. The result: ruler was present under
tabs. Example:
- Integrator has tab container margin (ruler would be visible full
height)
- User customized the color of the tab container making it transparent

Change-Id: I71d17c18c899cdb19b2c7e3bfd9903ba87221a78


Darshan-upadhyay1110 (approved):
Thanks :)

PR #9353, opened 2024-06-26
Dark mode: Fix dropdown shadow: use correct css var

pedropintosilva:
Before this commit things like status language dropdown etc were being
rendered with a light color

Change-Id: Ie2af6f4f88ff1b99fa6ee658a8dc8e700deac977


eszkadev (approved):

pedropintosilva:
This is a css var change of a color in CSS. no need to keep waiting or retry tests

PR #9356, opened 2024-06-26
Unify placeholder style

timar:
gettext tools gave false error messages in some
cases like %0. The variable between curly brackets is better.

Change-Id: I6226454bfc46f95c8d75c13043d3a7a8aba5b078

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar:
> I don't see anything obviously problematic with this.
> 
> Do we care about the translated strings invalidated by this? Perhaps 4 messages are not a concern and translators just deal with the fallout, I'm not sure.

I patched po files, too, at the same time I locked translations on Weblate. Hopefully we can get this in soon, and I can unlock translations on Weblate. There are a lot more than 4 messages, that were changed. I even fixed wrong placeholder 2x in translations.

PR #9357, opened 2024-06-27
postmessage demo: insert graphic test

eszkadev:
Add button in the post message tester to verify if inserting an image works properly. I used resource from our SDK so it is external domain, if not allowed it shows an error. We need to add domain to coolwsd.xml in the lok_allow section to allow importing the image.


mmeeks (approved):
Looks nice =)

PR #9358, opened 2024-06-27
postmessage demo: insert graphic test (master)

eszkadev:
Add button in the post message tester to verify if inserting an image works properly. I used resource from our SDK so it is external domain, if not allowed it shows an error. We need to add domain to coolwsd.xml in the lok_allow section to allow importing the image.

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -104,6 +104,12 @@
             });
       }
 
+      function insertGraphic(text) {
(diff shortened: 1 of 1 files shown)


eszkadev:
@hfiguiere could you review?
use case: test inserting from integration on a different domain, we can test with and without coolwsd.xml setting of `lok_allowed.host` to verify domain allowing functionality

hfiguiere (approved):

PR #9365, opened 2024-06-27
Stylesview: Fix dark mode background and border

pedropintosilva:
Before this commit there were
- Odd border artifact
- The styles previews were being rendered and looked like detached
rectangles (with different background than the background of the main
container)

This commit fixes those while making use of css vars. Additionally,
decrease the brightness of the whole stylesview to 90% (100% on hover
state) so the user doesn't have a very bright component always up
there.

Change-Id: I1bb21871d8a866f8f4308efda937904ac97504bb


pedropintosilva:
this change is css only and only changes color values

PR #9367, opened 2024-06-27
calc: fix copy hyperlink outside of edit mode

JaumePujantell:
Right now .uno:CopyHyperlinkLocation only works on edit mode or
inside a draw object, but LOK_CALLBACK_HYPERLINK_CLICKED is
dispatched outside of edit mode and the copy command from
the pop-up is ignored.

needed to discriminate the correct field, since there could be
multiple hyperlink fields inside the same cell.

Change-Id: Ie814867e0a4a319e78381b68343847491034d85f

### DIFF
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -1404,7 +1404,14 @@ function setupToolbar(e) {
 				//click pos tweak
 				cellCursor._northEast.lng = cellCursor._southWest.lng;
 				map._docLayer._closeURLPopUp();
-				map._docLayer._showURLPopUp(cellCursor._northEast, e.url);
+				var linkPosition;
+				if (strTwips.length > 7) {
+					linkPosition = {
+						x: parseInt(strTwips[6]),
+						y: parseInt(strTwips[7])
+					};
+				}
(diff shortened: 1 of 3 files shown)


JaumePujantell:
The corresponding core patch is: https://gerrit.libreoffice.org/c/core/+/169507

browser/src/control/Control.Toolbar.js
```diff
@@ -1404,7 +1404,14 @@ function setupToolbar(e) {
 				//click pos tweak
 				cellCursor._northEast.lng = cellCursor._southWest.lng;
 				map._docLayer._closeURLPopUp();
-				map._docLayer._showURLPopUp(cellCursor._northEast, e.url);
+				var linkPosition;
+				if (strTwips.length > 7) {
+					linkPosition = {
+						x: parseInt(strTwips[6]),
+						y: parseInt(strTwips[7])
```
vmiklos, line 1411: Looks OK, but this gets very complicated at this stage, could you add a comment above this block and describe the purpose of the 8 numbers you get here as a parameter?

Also, we have a cypress test suite at cypress_test/integration_tests/desktop/calc/clipboard_spec.js, do you think you could cover your fix with a test there? Thanks.

vmiklos (approved):

PR #9369, opened 2024-06-27
accessibility tooltip

hcvcastro:
- **browser: tooltip: add util function**
- **browser: tooltip: add Control.Tooltip.js file**
- **browser: tooltip: attach listener to buttons**

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2321,28 +2321,17 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				var label = L.DomUtil.create('label', 'ui-content unolabel', button);
 				label.htmlFor = buttonId;
 				label.textContent = builder._cleanText(data.text);
-				button.setAttribute('alt', label.textContent);
-				if (buttonImage !== false) {
-					if(data.aria) {
-						buttonImage.alt = data.aria.label;
-					} else {
-						buttonImage.alt = label.textContent;
-					}
-				}
 				builder._stressAccessKey(label, button.accessKey);
 				controls['label'] = label;
 				$(div).addClass('has-label');
 			} else if (builder.options.useInLineLabelsForUnoButtons === true) {
 				$(div).addClass('no-label');
 			} else {
-				div.title = data.text;
-				button.setAttribute('alt', data.text);
-				if (buttonImage !== false) {
-					buttonImage.alt = data.text;
-				}
 				builder.map.uiManager.enableTooltip(div);
 				$(div).addClass('no-label');
 			}
+
+			div.setAttribute('data-cooltip', builder._cleanText(data.text));
 
 			if (builder.options.useInLineLabelsForUnoButtons === true) {
 				$(div).addClass('inline');
@@ -2454,8 +2443,20 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			e.stopPropagation();
 		};
 
+		var mouseEnterFunction = function () {
+			if (builder.map.tooltip)
+				builder.map.tooltip.beginShow(div);
+		};
+
+		var mouseLeaveFunction = function () {
+			if (builder.map.tooltip)
+				builder.map.tooltip.beginHide(div);
+		};
+
 		$(controls.button).on('click', clickFunction);
 		$(controls.label).on('click', clickFunction);
+		$(div).on('mouseenter', mouseEnterFunction);
+		$(div).on('mouseleave', mouseLeaveFunction);
 
 		div.addEventListener('keydown', function(e) {
 			switch (e.key) {
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2331,6 +2331,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				$(div).addClass('no-label');
 			}
 
+			if (buttonImage !== false) {
+				if(data.aria) {
+					buttonImage.alt = data.aria.label;
+				} else {
+					buttonImage.alt = builder._cleanText(data.text);
+				}
+			}
+
 			div.setAttribute('data-cooltip', builder._cleanText(data.text));
 
 			if (builder.options.useInLineLabelsForUnoButtons === true) {
--- a/browser/src/control/Control.Tooltip.js
+++ b/browser/src/control/Control.Tooltip.js
@@ -43,6 +43,7 @@ class Tooltip {
 	beginHide(elem) {
 		if (this._cancel) return;
 
+		clearTimeout(this._showTimeout);
 		clearTimeout(this._hideTimeout);
 		if (this._current)
 			this._hideTimeout = setTimeout(
--- a/browser/src/control/Control.Tooltip.js
+++ b/browser/src/control/Control.Tooltip.js
@@ -51,24 +51,55 @@ class Tooltip {
 			);
 	}
 
+	/**
+	 * Calculate one of the 8 different position
+	 * of the tooltip, around the elem parameter
+	 *
+	 * - bottom-right, bottom-left, top-right, top-left
+	 * - and their left/right aligned versions
+	 *
+	 * @param elem - element that the cursor is over
+	 * @param popup - tooltip rectangle
+	 * @param index - used to determine one of the 8 tooltip location
+	 * @returns tooltip rectangle with its location calculated
+	 */
 	position(elem, popup, index) {
 		let rect = new DOMRect();
 		switch (index) {
-			case 0:
-				rect.x = elem.left + elem.width / 2 - popup.width / 2;
-				rect.y = elem.bottom;
-				break;
-			case 1:
-				rect.x = elem.right;
-				rect.y = elem.top + elem.height / 2 - popup.height / 2;
-				break;
-			case 2:
-				rect.x = elem.left + elem.width / 2 - popup.width / 2;
-				rect.y = elem.top;
-				break;
-			case 3:
+			case 0: // below cursor, bottom-right (aligned to left)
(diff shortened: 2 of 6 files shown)


browser/src/control/Control.Tooltip.js
```diff
@@ -0,0 +1,122 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * L.Control.Tooltip - tooltip manager
+ */
+
+/* global */
+
+L.Control.Tooltip =  L.Control.extend({
```
eszkadev, line 18: We currently try to get rid of Leaflet. Also we want new code in TypeScript.
Could you convert that into regular `class` (not inheriting from L.Control), and make this file `.ts` ?

eszkadev:
I think we already have at least 2 implementations of tooltips. Why this one is special? Cannot be other one reused?

Our goal is to unify things and reduce duplicated implementations. Best would be to reuse native browser tooltip (https://www.w3schools.com/jsref/prop_html_title.asp) or use jQuery one (`elem.tooltip()`) which is used in many places. Later we need to convert all of them into single type.

This reduces effort needed for maintaining the app in a long run.

hcvcastro:
> I think we already have at least 2 implementations of tooltips. Why this one is special? Cannot be other one reused?

I did not find implementations, or perhaps I missed, where are those implementations?

> Our goal is to unify things and reduce duplicated implementations. Best would be to reuse native browser tooltip
> (https://www.w3schools.com/jsref/prop_html_title.asp) or use jQuery one (elem.tooltip()) which is used in many places.
> Later we need to convert all of them into single type.

If those exist, it does not resolve accessibility issues, they disappear when mouse out.

> This reduces effort needed for maintaining the app in a long run.

I agree

browser/src/control/Control.Tooltip.js
hcvcastro: Sorry, I am not fan on proprietary library, but I fan standard an free software. I can change it
to class standard javascript

eszkadev: Yes, I mean: please use `class` from JavaScript - ECMAScript 6/2015 standard :)

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2449,8 +2441,18 @@ L.Control.JSDialogBuilder = L.Control.extend({
...
+		};
+
+		var mouseLeaveFunction = function () {
+			builder.map.tooltip.beginHide(div);
+		};
+
 		$(controls.button).on('click', clickFunction);
 		$(controls.label).on('click', clickFunction);
+		$(controls.button).on('mouseenter', mouseEnterFunction);
+		$(controls.button).on('mouseleave', mouseLeaveFunction);
```
eszkadev, line 2455: shoultn't we do this on the container - `$(div)` ? so label also will trigger the tooltip?

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2320,21 +2320,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 				var label = L.DomUtil.create('label', 'ui-content unolabel', button);
 				label.htmlFor = buttonId;
 				label.textContent = builder._cleanText(data.text);
-				button.setAttribute('alt', label.textContent);
```
eszkadev, line 2323: no tooltip data in this case?

browser/src/control/Control.JSDialogBuilder.js
hcvcastro: yes, I have tested to div container initially, I was not sure if the mouse will trigger,  I will set back to div again

browser/src/control/Control.JSDialogBuilder.js
hcvcastro: I was focused on buttons first wave, I have to change other places too. Thanks, if you know of other places too, it will help

eszkadev: I think it is important. Screen readers will read "alt" for images. Even if we add new tooltip it still has to be present I belive. In other case we will not pass automatic a11y checkers.

eszkadev: I checked with Firefox a11y checker and it seems to be valid issue:

a11y

hcvcastro: > I think it is important. Screen readers will read "alt" for images. Even if we add new tooltip it still has to be 
> present I belive. In other case we will not pass automatic a11y checkers.

It will conflict with native tooltips browser. Please write your concerns in the task description, it is not the scope of my task.

hcvcastro: > I checked with Firefox a11y checker and it seems to be valid issue:

Also, it is not my scope of my task. Please file a new issue in the accessibility task.

eszkadev: Thank you for the fix, now it will work as before

eszkadev (commented):
I approve the code, @pedropintosilva please look from UX / UI prespective and merge

pedropintosilva:
Thanks @hcvcastro !

Some tooltips are not readable (are cropped): 

image
image

NotebookbarBuilder is still using the other tooltips:
image

Additionally, sometimes the tooltip stays visible even when I interacting with the document. This can be very confusing for the user (The user triggers a tooltip by just moving the mouse from the UI to the document and doesn't notice that something appear and stayed there; Then looks again and sees a "Button" there (which is actually the tooltip)


https://github.com/CollaboraOnline/online/assets/65948705/088172bd-6509-407d-b3c8-7b8d4c913792

(all screenshots and record done in Firefox)

hcvcastro:
> Some tooltips are not readable (are cropped):

@eszkadev, @pedropintosilva Thanks, but I am not a Graphical Design expert, I did what I can play with those colors and fonts. My job is done,  @pedropintosilva  I will ask you to beautify the tooltips in your commits.

hcvcastro:
> NotebookbarBuilder is still using the other tooltips:

Unfortunately it is out of the scope of my task. Please file the new issue ticket

hcvcastro:
> Additionally, sometimes the tooltip stays visible even when I interacting with the document.

Firefox, of course, Ok, I will test it and see what I can do. Thanks

bayramcicek:
### Styling
- changed text color, bg color, padding
- added box-shadow
- didn't touch the font

| Before | After |
| --- | ---|
| before-styling | after-styling |

### Cropping/overflow issue

- add 8 different position for tooltips
  - bottom-right, bottom-left, top-right, top-left and their left/right aligned versions
  - pick one of them which is visible and is not overflow from the window
  - tooltips at the bottom and left/right aligned ones take precedence

| Before | After | 
| ---------------------------- | -------- |
| before-find-and-replace | after-find-and-replace |
| before-navigator | after-navigator |
| before-zoom-in | after-zoom-in |

### Tooltip stays visible if the cursor moves very fast over the elements/icons

> Additionally, sometimes the tooltip stays visible even when I interacting with the document.

```js
this._options = L.extend({ timeout: 150 }, options);
```
- when I put `timeout` to `0` the issue is gone. So, most likely there are some issues on the `mouseEnter & leave` functions about the `timeout` and I see that `mouseLeaveFunction` does not prevent `show(elem)` function from running, we need to prevent that.
```txt
mouseEnterFunction-->
beginShow-->>
---mouseLeaveFunction
<<--beginHide
show-->               <---- this should not run (we need to stop timeout somewhere and prevent this show function from running)
```

**Now**: working on this issue...

hcvcastro:
[bayramcicek](https://github.com/bayramcicek) Thanks for the improvements, nice looking tooltip =)

bayramcicek:
> ### Styling
> 
> * changed text color, bg color, padding
> * added box-shadow
> * didn't touch the font
> 
> | Before | After |
> |---- | ----- |
> | before-styling  | after-styling |
> ### Cropping/overflow issue
> 
> * add 8 different position for tooltips
>   * bottom-right, bottom-left, top-right, top-left and their left/right aligned versions
>   * pick one of them which is visible and is not overflow from the window
>   * tooltips at the bottom and left/right aligned ones take precedence
> 
> 
> | Before | After |
> |---- | ----- |
> | before-find-and-replace |	after-find-and-replace |
> | before-navigator |	after-navigator |
> | before-zoom-in  |	after-zoom-in |
> ### Tooltip stays visible if the cursor moves very fast over the elements/icons
> 
> > Additionally, sometimes the tooltip stays visible even when I interacting with the document.
> 
> ```js
> this._options = L.extend({ timeout: 150 }, options);
> ```
> 
> * I see that when mouse moves very fast over an icon/button; it is so fast that this move happens under `150 ms`, Therefore, after mouse left the icon, `show()` function still waits to be executed after `150 ms`; then it shows the tooltip even if the cursor is not over the icon. So, most likely there are some issues on the `mouseEnter & leave` functions about the `timeout` and I see that `mouseLeaveFunction` does not prevent `show(elem)` function from running, we need to prevent that.
> 
> 
> ```
> mouseEnterFunction-->
> beginShow-->>
> ---mouseLeaveFunction
> <<--beginHide
> show-->               <---- this should not run (we need to stop timeout somewhere and prevent this show function from running)
> ```
> 
> **Now**: working on this issue...

### Fix: Tooltip stays visible if the cursor moves very fast over the elements/icons
- cancel "_showTimeout" before it triggers show() - if beginHide() is called

### Before:

https://github.com/user-attachments/assets/a1236983-1241-4eb5-8445-5afcc6e22745

### After:

 https://github.com/user-attachments/assets/bf8491b1-17fd-4345-a574-f2d3bd10c355 


Ready for review. Thanks!

pedropintosilva (approved):
UI wise looks good

PR #9373, opened 2024-06-28
feat: record the performance of CPU,Network,Latency

elliotfreebairn1:
Change-Id: I415d414c832a00d2b55815df89f28a2d836a1290

* Resolves: # 
* Target version: master 

### Summary
The performance of the CPU is measured by the run time in ms, Network by the incoming and outgoing bandwidth in kB/s and latency is split into buckets of ping and tiles in ms. The data is stored in separate CSV files which will be updated each unit test allowing the performance data to be in a form which can be re-used, graphed or at-least monitored in the future.

### Checklist

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -14,3 +14,7 @@ gdb.txt
 ./unithttplib
 run_unit.sh
 run_unit_standalone.sh
+CPU.csv
+Network.csv
+PingLatency.csv
+TileLatency.csv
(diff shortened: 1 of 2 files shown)


elliotfreebairn1:
@mmeeks Hi, this is what i have got so far. I was a-bit unsure what data especially in regards to the networking performance you wanted, so i only added the incoming and outgoing bandwidth for now.

mmeeks (approved):
Looks fine to me; of course - quite how we aggregate and present these over time is not totally clear to me but =) perhaps a next step with @timar that would be interesting to setup; or perhaps we can just suck that from the tinderbox somehow =)

mmeeks:
Cypress tests are clearly unrelated; so lets push.

PR #9374, opened 2024-06-28
Accessibility Check btn: Fix label

pedropintosilva:
Use the previously translated string "Accessibility Check" instead of
the one that comes from the uno command "Open Accessibility Check Deck"

Change-Id: Id4ace21a933e31405a76eda7a530c05aea4d1c21


pedropintosilva:
to fix the new regression: image

PR #9376, opened 2024-06-28
CAP_MKNOD is no longer used

timar:
Since commit 144b701453de72c7c5a741cc148a42b25c309ad4 
cool#8703 - Drop random node creation and rely on inherited fd.

Change-Id: Iea3610989fa9eb46c41a9d3d2d6627ffa479cbd1


mmeeks (approved):
Oh - yes; and we hope to drop a lot more soon too with Kay / Caolan's namespace work =)

mmeeks:
Unrelated bogus cypress failures.

PR #9377, opened 2024-06-28
cypress: fix apply font in top_toolbar_spec

eszkadev:
probably broken due to on demand rendering rework


mmeeks:
Seems failures are in other un-related tests =) thanks Szymon.

PR #9380, opened 2024-06-29
Locking Documents

### LINKED ISSUE #9185: Latest version (24.04.3.1.1) does not send WOPI Lock/Unlock request
### Describe the Bug
Latest version (24.04.3.1.1) does not send WOPI Lock/Unlock request while user open/save/close a file, even if `SupportsLocks` is set as `true` from WOPI server in `GetFileInfo` response. It did carry a `X-WOPI-Lock` header in `PUT` request while saving file, but this did not help on protecting file been modified by other apps.

I tried a very old version (22.05.9.2), the lock works as expected.

### Steps to Reproduce
1. Setup WOPI server with `SupportsLocks=true, SupportsUpdate=true, UserCanWrite=true`;
2. Open a document, no LOCK request sent from client;
3. Save the document, no LOCK request sent from the client;
4. Close the document, no UNLOCK request sent from the client.

### Expected Behavior
Send Lock request right after user open the document, maintain the lock valid by refreshing lock request as long as the document is opened; Send Unlock request after all user close the document and all changes are saved.

### Actual Behavior
No lock related request is sent.

### Additional Context
Attach log snippet of one user opens a document.
[coolwsd.log](https://github.com/CollaboraOnline/online/files/15493159/coolwsd.log)

Ashod:
This resolves #9185.

The issue relates to loading asynchronously. This ahead-of-time loading is done before the client opens the WebSocket. This meant that the locking function didn't have the necessary details to decide whether locking was needed or not, and so it was skipped. The issue wouldn't happen if by the time we downloaded the document the WebSocket was created. As such, this is a time-sensitive bug. However, reports indicate that it's not uncommon that we weren't taking the lock.

The issue is resolved with this PR and now we take the lock when the _first_ editor joins (and as before, we unlock when the _last_ editor leaves). An improved unit-test asserts this behavior.

- wsd: all view-only extensions accept comments
- wsd: minor refactor around on-demand CFI
- wsd: always set view-file-extension flag
- wsd: reliable locking with async loading

### DIFF
--- a/test/UnitWOPILock.cpp
+++ b/test/UnitWOPILock.cpp
@@ -148,7 +148,9 @@ public:
 /// when the first view is read-only.
 class UnitWopiLockReadOnly : public WopiTestServer
 {
-    STATE_ENUM(Phase, LoadViewer, LoadEditor, Lock, WaitModify, Unlock, WaitUnload, Done) _phase;
+    STATE_ENUM(Phase, Connect, FirstCheckFileInfo, Lock, LoadViewer, WaitModify, Upload, Unlock,
+               WaitUnload, Done)
+    _phase;
 
     std::string _lockState;
     std::string _lockToken;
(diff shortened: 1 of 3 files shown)


mmeeks (approved):
OK - lets get this in - but it shouldn't be shipped in 24.04 until we've got through a red/green/stabilization cycle. Too risky IMHO =)

PR #9387, opened 2024-07-01
2d transition

codewithvk:
Change-Id: I8278a5c6ba57f1414ca890828fe38433ff75ec77

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/transition/transitions/FadeTransition.ts
+++ b/browser/src/control/transition/transitions/FadeTransition.ts
@@ -1,0 +1,17 @@
+class FadeTransition extends Transition2d {
+    private effectTransition : number = 0;
+    constructor(canvas: HTMLCanvasElement, vertexShaderSource: string, fragmentShaderSource: string, image1: HTMLImageElement, image2: HTMLImageElement) {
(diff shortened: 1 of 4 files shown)


browser/src/control/transition/transitions/FadeTransition.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    public renderUniformValue(): void {
+        this.gl.uniform1i(this.gl.getUniformLocation(this.program, 'effectType'), this.effectTransition);
+    }
+
+    public start(selectedEffectType : number): void {
+        this.effectTransition = selectedEffectType;
+        this.startTransition();
+    }
+}
```
eszkadev, line 17: please put this class into some global scope, example:
https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Widget.Containers.ts#L169

eszkadev: here you can initialize "global namespace", as some dict:
https://github.com/CollaboraOnline/online/blob/master/browser/src/docstate.js#L147

PR #9395, opened 2024-07-01
add 'mount_namespaces' option to use linux mount namespaces

caolanm:
if experimental_features and mount_namespaces is enabled and it is possible to "unshare" then instead of coolforkit exec coolforkitns which doesn't have any capabilities set and inprocess mount inside a namespace instead of calling coolmount

Change-Id: I48bef12b9156f41c78221e750a30aacee8a737a9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Wow - this looks great =) lets get it into red and iterate =)

PR #9400, opened 2024-07-02
Fix failing impress annotation tests

neilguertincollabora:
There is some instability with annotation ids. Sometimes it has the id annotation-content-2 instead of annotation-content-1. (or -3 instead of -2). The fix is to make the test look for either id.

Change-Id: Ie3759b5001f7194efea2d613752363bb0462f61f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):

PR #9399, opened 2024-07-02
Savemodified: Fix status, use svg instead of pseudo class

pedropintosilva:
- Remove unused CSS (w2ui)
- Before this commit the savemodified was not being reflect in the
  icon (didn't change)

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>
Change-Id: I33fa543b6f5e47fbca2fc87bdd4e822fbbac7337

PR #9406, opened 2024-07-03
Correct background URL for save modified icon in classic menu

ramchandra-kc:
* Resolves: No issue is created for this minor but
* Target version: master 

### Summary
When we edit a document while using the classic menu as a menu bar, then the icon that represents the document is modified and empty. The reason behind this is a very minor issue which is resolved by updating the URL of the background.

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1057,7 +1057,7 @@ button.leaflet-control-search-next
 .w2ui-icon.sign_not_ok{ background: url('images/sign_not_ok.svg') no-repeat center; }
 .w2ui-icon.save{ background: url('images/lc_save.svg') no-repeat center; }
 .w2ui-icon.saveas{ background: url('images/lc_saveas.svg') no-repeat center; }
-.w2ui-icon.savemodified{ background: url('images/savemodified_large.svg') no-repeat center; }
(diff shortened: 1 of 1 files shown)


pedropintosilva:
Thank you @ramchandra-kc for your first PR! Unfortunately the current PR doesn't fix it. We have another problem here related to the w2ui icons which we don't use anymore. Furthermore I have pushed the respective fix yesterday into the following PR https://github.com/CollaboraOnline/online/pull/9399 which not got merged.

Nevertheless we do still have many of these w2ui css rules that should be re-tested (to see if they are not really used somewhere else) and then remove. Would you be interested in taking that task?

ramchandra-kc:
> Thank you @ramchandra-kc for your first PR! Unfortunately the current PR doesn't fix it. We have another problem here related to the w2ui icons which we don't use anymore. Furthermore I have pushed the respective fix yesterday into the following PR #9399 which not got merged.
> 
> Nevertheless we do still have many of these w2ui css rules that should be re-tested (to see if they are not really used somewhere else) and then remove. Would you be interested in taking that task?

Hello @pedropintosilva, thank you for reviewing my PR, I would love to work on the issue you mentioned. Please provide me with some instructions and I can start immediately.

PR #9408, opened 2024-07-03
Private/gokay/latlng

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -2374,7 +2374,9 @@ L.CanvasTileLayer = L.Layer.extend({
 			return;
 		}
 
-		app.definitions.otherViewCursorSection.addOrUpdateOtherViewCursor(viewId, obj.rectangle.split(','), parseInt(obj.part), (obj.mode ? parseInt(obj.mode): 0));
+		const username = this._map._viewInfo[viewId].username;
+		const mode = obj.mode ? parseInt(obj.mode): 0;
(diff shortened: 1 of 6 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -4457,7 +4368,6 @@ L.CanvasTileLayer = L.Layer.extend({
 		if (this._docType === 'spreadsheet') {
 			map.on('zoomend', this._onCellCursorShift, this);
 		}
-		map.on('zoomend', L.bind(this.eachView, this, this._viewCursors, this._onUpdateViewCursor, this, false));
```
eszkadev, line 4466: this handler is removed, but do we have some replacement which happens after zoom change?

gokaysatir: > this handler is removed, but do we have some replacement which happens after zoom change?

Thanks for checking. Yes we a call under "_onZoomEnd" function.

eszkadev:
This PR works nice, but I found bug with user label. When using different sheets, we see label from the user who isn't on our sheet.

[latlng-2024-07-04_10.12.20.webm](https://github.com/CollaboraOnline/online/assets/5307369/5cccce57-6d2f-4537-9f7f-ce86b65d948e)

eszkadev (approved):
looks ok

gokaysatir:
> This PR works nice, but I found bug with user label. When using different sheets, we see label from the user who isn't on our sheet.
> 
>  latlng-2024-07-04_10.12.20.webm

Thanks for the catch :) Fixed.

PR #9409, opened 2024-07-03
Initialize storage.ssl.* config items

timar:
We need to initialize these, because we saw in the wild that a customer removed the empty defaults from coolwsd.xml and coolwsd stopped with fatal error.

Change-Id: Ifc75e3eb2fe50c2028f84d8f61ba5a034b14b4cf


caolanm (approved):
I suppose that's reasonable

PR #9422, opened 2024-07-03
Use a small section for rendering invalidation rectangles.

gokaysatir:
Change-Id: I837b71bab42070d744f6a61f463c315c9d4c413c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/.beforeprettier
+++ b/browser/.beforeprettier
@@ -133,6 +133,7 @@
 /src/canvas/sections/AutoFillMarkerSection.ts
 /src/canvas/sections/HTMLObjectSection.ts
 /src/canvas/sections/URLPopUpSection.ts
+/src/canvas/sections/InvalidationRectangleSection.ts
 /src/canvas/sections/ShapeHandlesSection.ts
 /src/canvas/sections/ShapeHandleScalingSubSection.ts
 /src/canvas/sections/ShapeHandleCustomSubSection.ts
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -231,6 +231,7 @@ COOL_JS_LST =\
 	src/canvas/sections/AutoFillMarkerSection.ts \
 	src/canvas/sections/HTMLObjectSection.ts \
 	src/canvas/sections/URLPopUpSection.ts \
+	src/canvas/sections/InvalidationRectangleSection.ts \
 	src/canvas/sections/ShapeHandlesSection.ts \
 	src/canvas/sections/ShapeHandleCustomSubSection.ts \
 	src/canvas/sections/ShapeHandlePolySubSection.ts \
--- a/browser/src/canvas/sections/InvalidationRectangleSection.ts
+++ b/browser/src/canvas/sections/InvalidationRectangleSection.ts
@@ -9,6 +9,8 @@
 */
 
 const sectionName = 'TileInvalidationRectangle';
+let counter = 0; // Unique counter.
+let sectionCount = 0;
 
 class InvalidationRectangleSection extends CanvasSectionObject {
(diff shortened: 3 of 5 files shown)


neilguertincollabora (commented):
I like that this is simpler than the previous implementation.

Look at the keypressqueue stuff. I don't think I ever figured out exactly what it was for or how it worked but I think it might have had something to do with the fade out that is now gone. See if you can delete it.

Also it looks like _tileInvalidationLayer is unused now.

Also make sure that the rectangles are deleted when the debug tool is turned off.

gokaysatir:
> I like that this is simpler than the previous implementation.
> 
> Look at the keypressqueue stuff. I don't think I ever figured out exactly what it was for or how it worked but I think it might have had something to do with the fade out that is now gone. See if you can delete it.
> 
> Also it looks like _tileInvalidationLayer is unused now.
> 
> Also make sure that the rectangles are deleted when the debug tool is turned off.

* I think keyPressQueue is still used.
* Removed the unused _tileInvalidationLayer.
* Checked that rectangles are deleted after turning the debug tools off. Also checked turning only invalidation rectangles off. It seems it's working. Can you describe the case if it's still not working for you?


And thanks for the review :)

neilguertincollabora:
I've played with it more now, and I have some minor issues:
- The 200ms delay before deleting the section seems unnecessary and makes it feel unresponsive.
- The leftover boxes in tiles that weren't refreshed are a little annoying. I guess this is a consequence of using the canvas that would need to be redrawn instead of a separate leaflet layer. Why is drawing on the canvas necessary? The more I think about it, the more I believe that as a debug visualization it should be a separate layer overlay, not part of the canvas/document.
- If two rectangles get invalidated in short succession (like when changing a cell that is used in a formula for another cell), only the second one shows. Would it be possible to display either the last 5 invalidations or the last 5 seconds of invalidations? I think this would be slightly more useful for debugging, but since I haven't actually done any tile invalidation debugging I can't say for sure.
- The red rectangle can be hard to see if there is already a rectangle there, like a cell selection or a text box. The semi-transparent fill in the old version was great for drawing attention to the spot.

These are just opinions though. I'm happy to approve if you can't or don't want to make changes.

gokaysatir:
> I've played with it more now, and I have some minor issues:
> 
> * The 200ms delay before deleting the section seems unnecessary and makes it feel unresponsive.
> * The leftover boxes in tiles that weren't refreshed are a little annoying. I guess this is a consequence of using the canvas that would need to be redrawn instead of a separate leaflet layer. Why is drawing on the canvas necessary? The more I think about it, the more I believe that as a debug visualization it should be a separate layer overlay, not part of the canvas/document.
> * If two rectangles get invalidated in short succession (like when changing a cell that is used in a formula for another cell), only the second one shows. Would it be possible to display either the last 5 invalidations or the last 5 seconds of invalidations? I think this would be slightly more useful for debugging, but since I haven't actually done any tile invalidation debugging I can't say for sure.
> * The red rectangle can be hard to see if there is already a rectangle there, like a cell selection or a text box. The semi-transparent fill in the old version was great for drawing attention to the spot.
> 
> These are just opinions though. I'm happy to approve if you can't or don't want to make changes.


* We need to keep the delay, or there would be a null pointer. Also, delay is now 1 second :) I think that won't be a big problem.
* Leftover boxes should be gone now.
* Now we draw the rectangles and for every invalidation and delete them after 1 second. But we keep the last rectangle.
* Rectangles are semi-transparent now.


Thanks for the review, can you have a second look when you have time?

neilguertincollabora (approved):
I like this much better!

PR #9430, opened 2024-07-04
browser: accessibility: set menubutton alt attribute

hcvcastro:
Set alt attribute to empty string.

Change-Id: I6732f70135a3ce60fb015e9fbeb72a3d93c2d687

### DIFF
--- a/browser/src/control/jsdialog/Widget.MenuButton.js
+++ b/browser/src/control/jsdialog/Widget.MenuButton.js
@@ -146,6 +146,7 @@ function _menubuttonControl (parentContainer, data, builder) {
 		if (data.image) {
 			var image = L.DomUtil.create('img', '', button);
 			image.src = data.image;
+			image.setAttribute('alt', '');
 		}
 		var label = L.DomUtil.create('span', '', button);
 		label.innerText = data.text ? data.text : '';


browser/src/control/jsdialog/Widget.MenuButton.js
```diff
@@ -146,6 +146,8 @@ function _menubuttonControl (parentContainer, data, builder) {
 		if (data.image) {
 			var image = L.DomUtil.create('img', '', button);
 			image.src = data.image;
+			if (button.id.indexOf('color') !== -1)
+				image.setAttribute('alt', '');
```
eszkadev, line 150: I think it should be the same as label of the icon: `data.text ? data.text : ''`
So the screen reader will get correct text.

hcvcastro: Unfortunately, the proposed solution, it is out of the scope of the task description. Please I advise you to change the task text description.

browser/src/control/jsdialog/Widget.MenuButton.js
```diff
@@ -146,6 +146,8 @@ function _menubuttonControl (parentContainer, data, builder) {
 		if (data.image) {
 			var image = L.DomUtil.create('img', '', button);
 			image.src = data.image;
+			if (button.id.indexOf('color') !== -1)
```
eszkadev, line 149: Please do not hardcode anything based on the ID of a widget.

Also, this only takes into account one case... when `color` is present. `alt` attribute should be always set, because browsers will warn about that in a11y check.

Please do this unconditionally.

PR #9440, opened 2024-07-05
wsd: fix async request

hcvcastro:
It will block main thread if fails.

Change-Id: I420c19f315b70b04e5e73c3dc7c6a93f3d165e34

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -208,6 +208,7 @@ std::string COOLWSD::BuyProductUrl;
 std::string COOLWSD::LatestVersion;
 std::mutex COOLWSD::FetchUpdateMutex;
 std::mutex COOLWSD::RemoteConfigMutex;
+std::shared_ptr<http::Session> FetchHttpSession;
 #endif
 
 // Tracks the set of prisoners / children waiting to be used.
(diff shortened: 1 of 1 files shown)


wsd/COOLWSD.cpp
```diff
@@ -4167,19 +4167,23 @@ void COOLWSD::processFetchUpdate()
         http::Request request(uriFetch.getPathAndQuery());
         request.add("Accept", "application/json");
 
-        const std::shared_ptr<const http::Response> httpResponse =
-            sessionFetch->syncRequest(request);
-        if (httpResponse->statusLine().statusCode() == http::StatusCode::OK)
-        {
-            LOG_DBG("Infobar update returned: " << httpResponse->getBody());
+        sessionFetch->setFinishedHandler([](const std::shared_ptr<http::Session>& httpSession) {
```
Ashod, line 4170: `sessionFetch` is local. It will get destroyed when we return. We need a long-lived instance and we need to check if there is a request in flight before we create a new one.

Ashod (commented):
Thanks, @hcvcastro. Please see the issue noted.

wsd/COOLWSD.cpp
mmeeks: @Ashod If we should never create sessions on the stack - can you privatize or protect the constructor, and have a static ::create method that returns a shared_ptr or somesuch ? =) hopefully that would stop others doing this and armor us a bit.

Thanks !

Ashod: @mmeeks, it is precisely as you say: `http::Session` is private and can only be created through a static that returns `shared_ptr`. The issue here was that the `shared_ptr` was a local and would get destroyed once we returned from this function.

mmeeks: Hmm - why then does the Response or whatver is holding things there not hold a reference on the session ? - requiring that to be owned by the std::function closure seems not to be a winning strategy ;-)

PR #9441, opened 2024-07-05
Use link color comes from core instead config

GulsahKose:
This patch is dependent to core patch that sends true link color for online cliens according to they are in dark mode or bright mode.

Change-Id: I7b819e9619a1254176424d85af2fd5c12c28f483

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/coolkitconfig.xcu
+++ b/coolkitconfig.xcu
@@ -101,14 +101,6 @@
                 <value>0</value>
             </prop>
         </node>
-        <node oor:name="Links">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>128</value>
-            </prop>
-            <prop oor:name="IsVisible" oor:op="fuse">
-                <value>true</value>
-            </prop>
-        </node>
         <node oor:name="LinksVisited">
             <prop oor:name="Color" oor:op="fuse">
                 <value>204</value>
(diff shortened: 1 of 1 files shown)


eszkadev:
best to put links to related tickets in the description so it is easy to find it

GulsahKose:
If you mean that, I'll update the commit message with core patch hash when core patch is merged.

pedropintosilva:
https://gerrit.libreoffice.org/c/core/+/169984

coolkitconfig.xcu
```diff
@@ -420,14 +412,6 @@
                 <value>0xFFFFFF</value>
             </prop>
         </node>
-        <node oor:name="Links">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>1939955</value>
```
eszkadev, line 425: we want to keep it here and not hardcode in the core I think

eszkadev (changes requested):

eszkadev:
i think it will be fixed in core-only patch

GulsahKose:
> i think it will be fixed in core-only patch

Yes i testednow only core patch works too.

eszkadev:
core patch will be enough

PR #9443, opened 2024-07-05
feat: recording CPU usage and peak memory usage

elliotfreebairn1:
Change-ID:Iee66f82e1026eae517f4452decf074fc0959bce0

* Resolves: 
* Target version: master 

### Summary
The CPU usage is calculated by the difference between the start up stats
compared to when the unit test has completed. Peak memory usage is measured
by comparing the memory of the process initially compared to, post loading a
document. Both metrics are recorded into seperate CSV files.

### Checklist

### DIFF
--- a/tools/Replay.hpp
+++ b/tools/Replay.hpp
@@ -92,21 +92,30 @@ struct Histogram {
 
         if (file.tellp() == 0)
         {
-            file << "Total Items,Too Long";
-            for (size_t i = 0; i < _buckets.size(); ++i) {
-                size_t bucketUpperLimit = (i < 10) ? (incLowMs * (i + 1)) : (maxLowMs + (i + 1 - 10) * incHighMs);
-                file << ",<" << bucketUpperLimit << "ms";
-            }
+            file << "Total Items, <10ms, <100ms, >100ms";
             file << "\n";
         }
 
-        file << _items << ",";
-        file << _tooLong << ",";
+        size_t totalTiles = _items;
+        size_t subTenCount = _buckets[0];
+        size_t subOneHundredCount = 0;
+        size_t overOneHundredCount = _tooLong;
 
-        for(size_t i = 0; i < _buckets.size(); i++)
+        for(size_t i = 1 ; i < _buckets.size(); i++)
         {
-            file << _buckets[i] << ",";
+            if(i < 10)
+            {
+                subOneHundredCount += _buckets[i];
+            }else
+            {
+                overOneHundredCount += _buckets[i];
+            }
         }
+
+        file << totalTiles << ",";
+        file << subTenCount << ",";
+        file << subOneHundredCount << ",";
+        file << overOneHundredCount << ",";
         file << "\n";
     }
 
@@ -120,7 +129,6 @@ struct Stats {
         _tileCount(0),
         _connections(0)
     {
-        initialCPUStat = GetProcessTimes();
         _startUpMemoryUsage = GetMemoryUsage();
         _peakMemoryUsage = 0;
     }
@@ -135,17 +143,6 @@ struct Stats {
     size_t _peakMemoryUsage;
     size_t _startUpMemoryUsage;
 
-    // cpu metrics breakdown
-    struct CPUStat
-    {
-        double userCPUTime;
-        double kernelCPUTime;
-        double elapsedTime;
-    };
-
-    CPUStat initialCPUStat;
-
-
     // message size breakdown
     struct MessageStat {
         size_t size;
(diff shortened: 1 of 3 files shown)


elliotfreebairn1:
@mmeeks Hi, I was a bit unsure about logic of when and where to record memory usage and CPU usage so this maybe not be completely correct. I've also charted the data and will post the spreadsheet to the original ticket. Thanks!

tools/Replay.hpp
```diff
@@ -108,6 +154,68 @@ struct Stats {
     std::unordered_map<std::string, MessageStat> _recvd;
     std::unordered_map<std::string, MessageStat> _sent;
 
+    CPUStat GetProcessTimes() {
+
+        std::ifstream fileStat("/proc/" + std::to_string(getpid()) + "/stat");
```
mmeeks, line 159: Hmm - this seems unhelpful. We measure higher resolution times in the SysStopwatch. We could split out the time there and get a much harder and hopefully more accurate number. Can you remove this lot and serialize the original tests's numbers from _timer.elapsedTime() or better.

Thanks.

tools/Replay.hpp
```diff
@@ -108,6 +154,68 @@ struct Stats {
...
+        CPUStat cpuStat = {
+            userProcessTimeInSeconds,
+            kernelProcessTimeInSeconds,
+            elaspedTime
+        };
+
+        return cpuStat;
+    }
+
+    size_t GetMemoryUsage()
```
mmeeks, line 199: Fair enough I guess - it is amazing that this jitters so much though [!].

mmeeks: I really love numbers that don't jitter. Can we addup 'Dirty' instead of PSS - which may be messed about by other processes starting/stopping that share things, and see if that helps jitter ? =)

tools/Replay.hpp
```diff
@@ -163,6 +271,17 @@ struct Stats {
...
+
+        CPUStat finalCPUStat = GetProcessTimes();
+        double userCPUTimeDiff = finalCPUStat.userCPUTime - initialCPUStat.userCPUTime;
+        double kernelCPUTimeDiff = finalCPUStat.kernelCPUTime - initialCPUStat.kernelCPUTime;
+        double elapsedTimeDiff = finalCPUStat.elapsedTime - initialCPUStat.elapsedTime;
+        double cpuUsagePercentage = ((userCPUTimeDiff + kernelCPUTimeDiff) * 100) / elapsedTimeDiff;
+
+        std::cout << "Peak memory usage: " << _peakMemoryUsage << "kB";
+
+        std::cout << "Percentage of CPU: " << cpuUsagePercentage << "%\n";
```
mmeeks, line 283: A %age like this is not useful; please drop. We want raw CPU cycle times spent as per the timer wrapper I included.

tools/Replay.hpp
```diff
@@ -174,12 +293,78 @@ struct Stats {
             " server sent " << Util::getHumanizedBytes(_bytesRecvd) <<
             " (" << recvKbps << " kB/s) to " << _connections << " connections.\n";
 
+        _pingLatency.dumpLatencyToCSV("Ping");
+        _tileLatency.dumpLatencyToCSV("Tile");
+
+        dumpStressToCSV(runMs);
+        dumpNetworkStatsToCSV(recvKbps,sentKbps);
```
mmeeks, line 300: We don't care about the time it took; adding Kbps - adds significant jitter to what -should- be a nice hard number.

Lets log the total _bytesRecvd and _bytesSent - those are the useful things that (I hope) - should be stable =)

tools/Replay.hpp
```diff
@@ -163,6 +271,17 @@ struct Stats {
...
+        double userCPUTimeDiff = finalCPUStat.userCPUTime - initialCPUStat.userCPUTime;
+        double kernelCPUTimeDiff = finalCPUStat.kernelCPUTime - initialCPUStat.kernelCPUTime;
+        double elapsedTimeDiff = finalCPUStat.elapsedTime - initialCPUStat.elapsedTime;
+        double cpuUsagePercentage = ((userCPUTimeDiff + kernelCPUTimeDiff) * 100) / elapsedTimeDiff;
+
+        std::cout << "Peak memory usage: " << _peakMemoryUsage << "kB";
+
+        std::cout << "Percentage of CPU: " << cpuUsagePercentage << "%\n";
+
         std::cout << "Stress run took " << runMs << " ms\n";
```
mmeeks, line 285: I do like the idea of logging the runMs - that should be jittery - but not that much, and if wiggling a lot provides a good clue something went wrong.

tools/Replay.hpp
```diff
@@ -81,6 +85,31 @@ struct Histogram {
             std::cout << "< " << std::setw(4) << ms << " ms |" << std::string(chrs, '-') << "| " << _buckets[i] << "\n";
         }
     }
+
+    void dumpLatencyToCSV(std::string fileName)
```
mmeeks, line 89: This is good - fine to have the data; but the graph is rather unexpected.

It seems we have two states - a slow and a fast one - sometimes we take >300ms to process things, and then we get many fewer tiles, and sometimes we are much quicker and get more high speed tiles. This points to an interesting potential problem.

We should log in a summary CSV - the number of tiles we got, and the number sub 10ms, the number up to 100ms, and the number over that: so four numbers:  #tiles, sub10ms, sub100ms, other

Hopefully we can come up with a single summary CSV that has our highest quality & most useful metrics in it =)

mmeeks (changes requested):
Nice spreadsheet; some feedback on the patches & data though as appended =)
Thanks !

elliotfreebairn1:
@mmeeks Just came back from holiday yesterday. The jittering has reduced a bit with some of these changes but i feel they're still definitely better ways i can measure some of these metrics which i'll keep on researching. Here is an updated spreadsheet with the charts: 
[NewPeformanceCharted.ods](https://github.com/user-attachments/files/16335922/NewPeformanceCharted.ods)

tools/Replay.hpp
```diff
@@ -92,21 +92,30 @@ struct Histogram {
...
+        size_t overOneHundredCount = _tooLong;
 
-        for(size_t i = 0; i < _buckets.size(); i++)
+        for(size_t i = 1 ; i < _buckets.size(); i++)
         {
-            file << _buckets[i] << ",";
+            if(i < 10)
+            {
+                subOneHundredCount += _buckets[i];
+            }else
```
mmeeks, line 109: please do read your diff before pushing =)

tools/Replay.hpp
```diff
@@ -137,6 +151,26 @@ struct Stats {
...
     std::unordered_map<std::string, MessageStat> _sent;
 
+    size_t GetMemoryUsage()
+    {
+        std::ifstream smapsFile("/proc/" + std::to_string(getpid()) + "/smaps");
+
+        std::string line;
+        size_t totalDirtyPss = 0;
+
+        while (std::getline(smapsFile, line)) {
```
mmeeks, line 161: whatever the majority bracing style is in the rest of the file: we should use that - please check =)

mmeeks (approved):
Looks great; I give my approval - but would love some stylistic cleanup there =)

We should work out what machine we run this on and start thinking about adding some web graphs here @timar any thoughts where we can host that ?

elliotfreebairn1:
@mmeeks I've made the changes but these failing tests seem to be weird, as no changes were made to the code and it looks like CodeQL has been failing a lot of tests since Tuesday. Might have to wait until it's fixed or there is an issue with my commit i can't see?

mmeeks:
Yep - that's unrelated; lets get this in; thanks Elliot =)

PR #9446, opened 2024-07-05
Private/mmeeks/bgsave21

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -518,6 +518,36 @@ extern "C" {
     }
 }
 
+// Define in order to dump a trace whenever a thread is created.
+// #define THREAD_TRACKER
+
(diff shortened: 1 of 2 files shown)


mmeeks:
Hard to believe; but chatgpt has this to say ;-)

**Visibility in /proc**

- The thread entry in the /proc filesystem (under /proc/[pid]/task/[tid]) is removed almost immediately after pthread_join completes and the kernel cleans up the thread.

- There might be a very brief delay (on the order of milliseconds) due to the time it takes for the kernel to process the cleanup, but for practical purposes, this happens as part of the pthread_join cleanup process.

caolanm (approved):

Issue #9383, opened 2024-06-29
Error make on FreeBSD


ev256545:
### Describe the Bug
make on FreeBSD ends with error
`
common/SigUtil.cpp:28:12: fatal error: 'sys/prctl.h' file not found
\#  include <sys/prctl.h>
`
### Steps to Reproduce
1. env HOME=/tmp/coolhome gmake -j`sysctl -n hw.ncpu`

### Expected Behavior
make on FreeBSD ends without error

### Actual Behavior
make on FreeBSD ends with error

### Desktop
(Please complete the following information)
 - **Collabora version:** git master Jun 29, 2024
 - **OS and version:** FreeBSD 14.0-RELEASE

### Additional Context
Commits a810221 (Apr 16, 2024) not take other behavior for FreeBSD

PR #9452, opened 2024-07-08
cool#9383: build error on FreeBSD

timar:
Change-Id: I287c938bfcdbbdb76dd8f1253e52bb167b440a22

* Resolves: #9383 
* Target version: master

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -24,7 +24,7 @@
 #include <sys/uio.h>
 #include <unistd.h>
 
-#if !defined(ANDROID) && !defined(IOS)
+#if !defined(ANDROID) && !defined(IOS) && !defined(__FreeBSD__)
 #  include <sys/prctl.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
There is a bit of risk that this will lead to processes that linger around because we don't die when our parent dies, but I guess this at least fixes the build. And if that concern turns into a practical problem, we can deal with it in some way.

JAicewizard:
There is an alternative, see https://man.freebsd.org/cgi/man.cgi?procctl(2) `PROC_PDEATHSIG_CTL` released in FreeBSD 11.2 (7 years ago) which has reached EOL already. I think we should be good to start using that? I will open a PR for it.

vmiklos:
Sounds good, in general we don't support COOL on EOL operating systems. Thanks.

JAicewizard:
This was already merged in #12499, didn't reference this PR so thats why it doesn't show up here

PR #9456, opened 2024-07-08
Renaming and Other Improvements

Ashod:
- wsd: skipped quarantines is not a warning
- killpoco: Poco::URI::decode
- wsd: test: move hex tests to UtilTests
- wsd: test: move isValidUtf8 test to UtilTests
- wsd: remove redundant initialization
- wsd: test: better rename test
- wsd: new CharacterConverter utility
- wsd: use CharacterConverter

### DIFF
--- a/common/CharacterConverter.hpp
+++ b/common/CharacterConverter.hpp
@@ -1,0 +1,89 @@
+/* -*- Mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; fill-column: 100 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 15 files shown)


caolanm (approved):

common/CharacterConverter.hpp
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        // Flush.
+        if (iconv(_iconvd, nullptr, nullptr, &out, &out_left) == static_cast<size_t>(-1))
+        {
+            LOG_ERR("Failed to flush [" << source << ']');
+            return source;
+        }
+
+        return std::string(&buffer[0], buffer.size() - out_left);
```
caolanm, line 79: I tend to see buffer.data() instead of &buffer[0] to avoid the empty case triggering debug asserts, but that's only an aside.

caolanm:
"CodeQL scanning" can be ignored. There has been a temp fix pushed to workaround that failure to master now.
"COOL 24.04 unit tests w/o cypress" I restarted to see if that's the usual intermittent failure.

common/CharacterConverter.hpp
Ashod: Good point. Will improve. Thanks!

PR #9458, opened 2024-07-08
Attempted blind fix for transparent tiles / race during load.

mmeeks:
Only seen in the wild so far on Wieland Lindenthal's browser & server combinations; but plausibly (somehow) we could be rendering blank tiles during a document load progress callback if we've gone really wrong.

Also cleanup the (non functioning) _duringLoad count to improve debugging cleanliness.

Change-Id: I9f32d4fa28fa4a60d8439c55b67058bd0c27ec2f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
looks sensible

caolanm:
writer/statusbar_spec.js failure is reproducible with this in place and not when not applied, via:

make run-desktop spec=writer/statusbar_spec.js

looks like an async dialog warning that appears in the non-working case, so presumably alertNotAsync that uses _duringLoad is indicated

I think the original std::unique_ptr...(nullptr, { _duringLoad-- }) simply didn't call that custom dtor decrement in the past, given the nullptr arg, so I imagine _duringLoad was always >= 0 (for the person loading the document?) so with this change that is fixed and an async warning dialog appears.

caolanm:
https://gerrit.libreoffice.org/c/core/+/170253 to make that offending dialog async

mmeeks:
And merged - thanks Caolan =)

caolanm:
make run-desktop spec=writer/top_toolbar_spec.js

shows that insert, special character is likewise non-async

caolanm:
https://gerrit.libreoffice.org/c/core/+/170273 for that next hurdle

mmeeks:
@timar would be good to have in the next release too - though these two core async'ing patches are needed too obviously =)

PR #9460, opened 2024-07-09
slideshow: Code refactoring for Fade transition type 

codewithvk:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/slideshow/PerformTransition.ts
+++ b/browser/src/slideshow/PerformTransition.ts
@@ -10,99 +10,99 @@
 declare var SlideShow: any;
 
 enum TransitionType {
-    BARWIPE,
-    PINWHEELWIPE,
-    SLIDEWIPE,
-    RANDOMBARWIPE,
-    CHECKERBOARDWIPE,
-    FOURBOXWIPE,
-    IRISWIPE,
-    FANWIPE,
-    BLINDSWIPE,
-    FADE,
-    DISSOLVE,
-    PUSHWIPE,
-    ELLIPSEWIPE,
-    BARNDOORWIPE,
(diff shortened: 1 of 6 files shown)


browser/src/slideshow/PerformTransition.ts
```diff
@@ -114,24 +114,17 @@ SlideShow.PerformTransition = function (
...
-				} else if (
-					transitionSubType == TransitionSubType.FADEOVERCOLOR &&
-					slideInfo.transitionDirection
-				) {
-					subTypeIndex = 1;
-				} else {
-					subTypeIndex = 0;
-				}
-			}
-			new SlideShow.FadeTransition(canvas, image1, image2).start(subTypeIndex);
```
eszkadev, line 129: can we have just `slideInfo` passed into transition itself?

`new SlideShow.FadeTransition(canvas, image1, image2, slideInfo).start()`

then every transition will be able to decide inside what to do with it

browser/src/slideshow/transition/helper/PerformFadeTransition.ts
```diff
@@ -0,0 +1,41 @@
...
+
+declare var SlideShow: any;
+
+enum FadeSubType {
+	FADEOVERBLACK,
+	FADEOVERWHITE,
+	SMOOTHLY,
+}
+
+function PerformFadeTransition(
```
eszkadev, line 19: This class/function is not needed. This code can be included inside `FadeTransition`.
But if any type can use 2 different types depending on some properties -> then we should create such intermediate object

PR #9465, opened 2024-07-09
Private/gokay/shape handles

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
it's green week, we have to wait ~2 weeks to merge

gokaysatir:
> all the handles doesn't work when trying to use them second time

This one is interesting. Reported also before and i couldn't reproduce this. I'll check again. Thanks for the review. I'll also check the rotation handle usage.

gokaysatir:
I increased the radius of rotation handle. I hope it is easier to pick it now. I can't reproduce the issue with second usages of handles.

eszkadev:
Rotation still doesn't work propery. It just works sometimes. When I try 10 times, 5 of them fail to drag...

gokaysatir:
> Rotation still doesn't work propery. It just works sometimes. When I try 10 times, 5 of them fail to drag...

I see, let's get this in since this improves the situation. Then i'll prepare a follow up and check this issue deeply.

eszkadev (approved):

PR #9467, opened 2024-07-10
calc: fix edit and remove hypelink outside of edit mode

JaumePujantell:
Right now .uno:EditHyperlink and .uno:RemoveHyperlink only work on
edit mode or inside a draw object, but LOK_CALLBACK_HYPERLINK_CLICKED
is dispatched outside of edit mode.

needed to discriminate the correct field.

Change-Id: I8499b430d5e62b904caba5bb5c7c34e784a8b850

### DIFF
--- a/browser/src/canvas/sections/URLPopUpSection.ts
+++ b/browser/src/canvas/sections/URLPopUpSection.ts
@@ -74,29 +74,30 @@ class URLPopUpSection extends HTMLObjectSection {
 				app.map.sendUnoCommand('.uno:JumpToMark?Bookmark:string=' + encodeURIComponent(this.sectionProperties.url.substring(1)));
 		};
 
+		var params: any;
+		if (linkPosition) {
+			params = {
+				PositionX: {
+					type: 'long',
+					value: linkPosition.x
+				},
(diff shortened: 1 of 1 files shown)


browser/src/canvas/sections/URLPopUpSection.ts
```diff
@@ -74,29 +74,30 @@ class URLPopUpSection extends HTMLObjectSection {
 				app.map.sendUnoCommand('.uno:JumpToMark?Bookmark:string=' + encodeURIComponent(this.sectionProperties.url.substring(1)));
 		};
 
+		var params: any;
```
vmiklos, line 77: Nit: if you touch this line anyway, then better to use const or let, 'var' is legacy.

vmiklos (approved):

Did you consider adding a cypress test for this change? I believe that cypress_test/integration_tests/desktop/calc/cell_cursor_spec.js may be a good place to add such a test. Again, the hyperlink edit is a bit more tricky to test, but removing a hyperlink would be probably reasonably straightforward. If it's easier, you can also add a new suite under desktop/calc. Thanks.

PR #9471, opened 2024-07-10
fix: coolwsd getting killed on kit segfault

Rash419:
- regression from a866719

Change-Id: I21b20aabb4dffad3011b75de8d1f373d495a697d


mmeeks (approved):
Great catch ! =) @timar badly need this in the stable release.

PR #9475, opened 2024-07-11
following: instant jump to followed user

eszkadev:
When we decide to follow someone - don't wait for the interaction. Just jump to the followed cursor.


mmeeks (approved):
Seems reasonable =) thanks Szymon.

PR #9476, opened 2024-07-11
following users fixes

eszkadev:
this fixes the case:
- open writer
- press multiple times ctrl + Enter

sometimes you got new page but view was on the old page, so you were not seeing the cursor


gokaysatir (approved):
Thanks for the fix :)

PR #9482, opened 2024-07-11
Private/caolan/usermount5

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3096,6 +3096,8 @@ void lokit_main(
                 return true;
             };
 
+            bool usingMountNamespace = false;
+
             // Copy (link) LO installation and other necessary files into it from the template.
             bool bindMount = JailUtil::isBindMountingEnabled();
             if (bindMount)
@@ -3107,7 +3109,6 @@ void lokit_main(
 #endif // CODE_COVERAGE
 
 #ifndef __FreeBSD__
-                bool usingMountNamespace = false;
                 const uid_t origuid = geteuid();
                 const gid_t origgid = getegid();
 
@@ -3220,12 +3221,23 @@ void lokit_main(
             }
 
 #ifndef __FreeBSD__
-            dropCapability(CAP_SYS_CHROOT);
-            dropCapability(CAP_FOWNER);
-            dropCapability(CAP_CHOWN);
+            if (usingMountNamespace)
+            {
+                // Drop all capabilities, we have the full set in our namespace
+                cap_t caps = cap_init();
+                cap_set_proc(caps);
+                cap_free(caps);
+            }
+            else
+            {
+                dropCapability(CAP_SYS_CHROOT);
+                dropCapability(CAP_FOWNER);
+                dropCapability(CAP_CHOWN);
+            }
 #endif
-
-            LOG_DBG("Initialized jail nodes, dropped caps.");
+            char *capText = cap_to_text(cap_get_proc(), nullptr);
+            LOG_DBG("Initialized jail nodes, dropped caps. Final caps are: " << capText);
+            cap_free(capText);
         }
         else // noCapabilities set
         {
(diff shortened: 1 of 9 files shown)


caolanm:
still a WIP so not there yet,

These namespace things can't be created when there are threads running, and the Poco::Log rotateOnOpen option triggers one of these, via an archiver/zip-thing if it finds an existing log. I've just temp bodged that off here for simplicity[1].

mounts done by a child are local to it and so not visible to the parent, so the /tmp mounted in the child isn't seen at /path/to/jails/foo/jailid/tmp by the parent. If the parent uses where it is really mounted from of /path/to/jails/tmp/cool-jailid then things are straightforward. I've made this tweakery dependent on using mount-namespaces for now at least.

[1] One possibility might be if archive mode is used to open the log and close it again to force the archive to do its work and join that thread, losing presumably the neat ability to zip that in a thread while continuing to load. And then reopen the log again which probably doesn't then need to launch a thread.

wsd/Storage.cpp
```diff
@@ -74,15 +74,44 @@ bool StorageBase::SSLEnabled = false;
...
     std::string localPath = _jailPath;
     if (localPath[0] == '/')
     {
         // Remove the leading /
         localPath.erase(0, 1);
     }
 
+    // Use where mountJail of kit/Kit.cpp mounts /tmp for this path *from* rather than
+    // where it is mounted *to*, so this process doesn't need the mount visible to it
+    if (COOLWSD::EnableMountNamespaces && !localPath.empty())
```
mmeeks: We should prolly do this un-conditionally in the next red release =)

kit/Kit.cpp
```diff
@@ -3190,12 +3221,23 @@ void lokit_main(
             }
 
 #ifndef __FreeBSD__
-            dropCapability(CAP_SYS_CHROOT);
-            dropCapability(CAP_FOWNER);
-            dropCapability(CAP_CHOWN);
+            if (usingMountNamespace)
```
mmeeks, line 3224: Cap dropping is tricky and seems to be copy/pasted below (?) =)

wsd/COOLWSD.cpp
```diff
@@ -2560,9 +2552,40 @@ void COOLWSD::innerInitialize(Application& self)
 #endif // CODE_COVERAGE
 
     // Setup the jails.
+    const bool UseMountNamespaces = getConfigValue<bool>(conf, "mount_namespaces", false);
```
mmeeks, line 2555: Bit of a surprise here; do we really need to do this inside coolwsd ? I'd expect this to be specific to the forkit; or are we containing the whole coolwsd now as well ? =) [ which makes good sense if I think about it too ;-]

mmeeks (approved):
Looks really good; a few comments below though =)

wsd/Storage.cpp
caolanm: yeah, needs checking for existing fallback non-mount cases where I think we use link as to what links we are using and perhaps need to align the link fallback with the mount case so its always the same.

kit/Kit.cpp
caolanm: can certainly bundle these together inside a utility

wsd/COOLWSD.cpp
caolanm: In coolwsd we use setupChildRoot and that has a testing bind mount+umount of cool_test_mount to see if the full combination works and if it fails then disables bindmounting globally. So stuck that into the same type of combination of namespaces as used in the forkit.

mmeeks: Makes sense I guess =) So we're not changing coolwsd's view of the world / capabilities etc.

Quite possibly coolwsd should (after sucking in its file data) enter a namespace with no system file-system visibility either =) but that's for later - I wonder what best practice is there these days.

PR #9483, opened 2024-07-12
Fix branding of COOL Lite when key is not present or expired

timar:
regression from 246e87fea21d6121fd4bd2daeff8a4f92082477e 
we've got [ERRFMT].js and [ERRFMT].css

Change-Id: Ic59257fedb6d9ae6712f7f9ff16022c01f70112a


vmiklos (approved):
Pff, indeed Poco::format() works with std::string, but not with string literals..

PR #9484, opened 2024-07-12
allow logRotateOnOpen but sync archiving to proceed without additiona…

caolanm:
…l thread

Change-Id: I34d3cdc6e1753c8d8410db2b760035c75df020b0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -610,10 +610,41 @@ namespace Log
 
         if (logToFile)
         {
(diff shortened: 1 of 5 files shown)


mmeeks (approved):
Grim, but lets get it in =) Really I'd love to rip & replace the rather unhelpful poco channel / logging stuff and finally bin that. It is at least all wrapped and hidden in Log.hpp these days which is good.

PR #9486, opened 2024-07-12
Fix branding of COOL Lite in Admin Console

timar:
I missed this case in the previous commit.

Change-Id: I48829591ce20bcfabb239e510041799c41ea50dc


vmiklos (approved):
Brief `git grep 'Poco::format.*[A-Z][A-Z]'` suggests this was indeed the last broken case.

PR #9490, opened 2024-07-12
Warn on low thread count systems.

mmeeks:
It pays from a latency perspective to have fewer machines or containers each with more threads, rather than more of them with fewer threads.

Change-Id: Id791438d080989ffcf6b079e551f622cf92353e2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -2666,6 +2666,14 @@ void COOLWSD::innerInitialize(Application& self)
         setenv("MAX_CONCURRENCY", std::to_string(maxConcurrency).c_str(), 1);
     }
     LOG_INF("MAX_CONCURRENCY set to " << maxConcurrency << '.');
+
+    // It is worth avoiding configuring with a large number of under-weight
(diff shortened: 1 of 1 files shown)


wsd/COOLWSD.cpp
```diff
@@ -2666,6 +2666,14 @@ void COOLWSD::innerInitialize(Application& self)
         setenv("MAX_CONCURRENCY", std::to_string(maxConcurrency).c_str(), 1);
     }
     LOG_INF("MAX_CONCURRENCY set to " << maxConcurrency << '.');
+
+    // It is worth avoiding configuring with a large number of under-weight
+    // containers / VMs - better to have fewer, stronger ones.
+    if (nThreads < 4)
+        LOG_WRN("Fewer threads than recommended. Having at least four threads for "
```
Ashod, line 2673: Good to include `nThreads` in the log.

Ashod (approved):
Thanks, @mmeeks. Minor suggestion, but looks good.

PR #9492, opened 2024-07-12
use buildLocalPathToJail instead of direct JAILED_DOCUMENT_ROOT concat

caolanm:
use buildLocalPathToJail instead of direct JAILED_DOCUMENT_ROOT concat
    
e.g. download as pdf/save as pdf
    
move buildLocalPathToJail to FileUtil instead of JailUtil given
that the JailUtil code isn't built on mobile

Change-Id: I6f8ddaf1053d084ec8c1f22adf830a677e80cadc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -390,6 +390,43 @@ namespace FileUtil
         return (readFile(path, *data, maxSize) >= 0) ? std::move(data) : nullptr;
     }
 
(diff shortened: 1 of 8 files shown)


caolanm:
Some testing finds these save as pdf, etc cases

mmeeks (approved):
Looks very sensible =)

Thanks!

PR #9493, opened 2024-07-12
default make run to use mount namespaces

caolanm:
which don't use coolmount and coolforkit, but instead use linux mount spaces with coolforkitns instead.

That doesn't need sudo, so drop automatically calling that in make to setcaps on coolmount + coolforkit for an easier developer experience.

Change-Id: I6f8ddaf1053d084ec8c1f22adf830a677e80cadc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -478,12 +478,6 @@ SYSTEM_STAMP = @SYSTEMPLATE_PATH@/system_stamp
 CAPABILITIES = $(if @ENABLE_SETCAP@,true,false)
 RUN_GDB = $(if $(GDB_FRONTEND),$(GDB_FRONTEND),gdb --tui --args)
 
-if ENABLE_SETCAP
-SET_CAPS_COMMAND=sudo @SETCAP@ cap_fowner,cap_chown,cap_sys_chroot=ep coolforkit && sudo @SETCAP@ cap_sys_admin=ep coolmount
-else
-SET_CAPS_COMMAND=echo "Skipping capability setting"
-endif
-
 if ENABLE_LIBFUZZER
 CLEANUP_DEPS=
 else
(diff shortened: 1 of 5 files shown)


caolanm:
Here we still default to namespace mount off except for "make run" where we default it on. And drop the SET_CAPS_COMMAND step with the sudo.

caolanm:
I think we can skip this for now, it was just to make things more convenient for dev

caolanm:
Here we try using these linux mount namespaces always by default for make run, etc for development

vmiklos (approved):
Nice. :-)

Couple of observations:

1) Works for me on openSUSE Leap 15.5, which is kernel 5.14, i.e. far from the latest one.

2) Interestingly still need to use 'sudo gdb', otherwise I get:

```
warning: "target:/home/vmiklos/git/collaboraonline/online-24.04/coolforkitns": could not open as an executable file: Operation not permitted.

warning: `target:/home/vmiklos/git/collaboraonline/online-24.04/coolforkitns': can't open to read symbols: Operation not permitted.

warning: Could not load vsyscall page because no executable was specified
0x00007f688362bb96 in ?? ()
```

While normally on gdb attach we break in ppoll. (Ignore the patch, 24.04 there means it's online.git master against core.git co-24.04.)

3) Otherwise debugging, including dumping content to /tmp works fine, e.g I can still do:

```
print (void)chdir("/tmp")
print SwRootFrame::s_pLast->dumpAsXml(0)
```

and that produces a layout.xml I can check from outside the namespace, based on what 'find . -name layout.xml' prints.

caolanm:
So for me on fedora this is already similarly the case for all process I attach to by default so I didn't pay any heed that this is a slightly different scenario with the same workaround of using root

So, for me first
```
sudo echo 0 > /proc/sys/kernel/yama/ptrace_scope
```
to allow attaching to processes at all.

Then what seems to work for me is:
```
(gdb) set sysroot /
(gdb) attach <pid>
```

or as a one-liner
```
$ gdb -iex "set sysroot /" --pid <pid>
```

Issue #9487, opened 2024-07-12
Writer: No Images in stylesview of home page for distro/collabora/co-24.04/


ramchandra-kc:
### For branch distro/collabora/co-24.04/ the images in the styles view are missing.

When we do a simple make run, and open the default .odt file, the styles view does not have images.

It is working with the master branch, but not with distro/collabora/co-24.04/ branch. 

### Steps to Reproduce
1. git checkout distro/collabora/co-24.04/
2. Configure and run
3. Go to HomeTab
4. See stylesview

### Expected Behavior
The images should be seen in the styles view.

### Actual Behavior
There are no images, only the text.

### Screenshots
Expected / running on master branch 
![image](https://github.com/user-attachments/assets/6409570f-e732-48ca-a094-cc6003beb03e)
Actual / running on distro/collabora/co-24.04/:
![image](https://github.com/user-attachments/assets/3c579ef7-5e6a-4260-b011-45d2006c1001)

### Desktop
(Please complete the following information)
 - **Collabora version:24.04** [e.g. 23.05.5.3]
 - **OS and version:Windows 11**
 - **Browser and version:Chrome** 
 - **Branch: distro/collabora/co-24.04/**
 
### Additional Context

I tried debugging, the main issue lies with the messageData we get.
I have used the same instdir and include folder for the both (i.e daily-built archive of LibreOffice core)
In distro/collabora/co-24.04, I get 4 logs:
![image](https://github.com/user-attachments/assets/2f7d64b1-b3a9-4f8b-8176-4370a5fc1c94)
But while running through master:
![image](https://github.com/user-attachments/assets/67600ee1-963d-45ea-84cb-d80bc61c40da)

PR #9494, opened 2024-07-12
backport style previews to co-24.04

eszkadev:
Fixes issue https://github.com/CollaboraOnline/online/issues/9487


timar:
The question is, whether you want an out-of-band COOL 24.04.4.x release for this immediately. If not, we'll merge everything from master to co-24.04 soon, because we are at the end of green CODE period.

eszkadev:
> The question is, whether you want an out-of-band COOL 24.04.4.x release for this immediately. If not, we'll merge everything from master to co-24.04 soon, because we are at the end of green CODE period.

ok, if everything goes in then it's fine :) closing

PR #9496, opened 2024-07-12
default to using mount namespaces if experimental enabled

caolanm:
Change-Id: I5ce91a2aaefd8c84eecb6aa5d66f492cb8a3becd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -66,14 +66,14 @@ bool enterMountingNS(uid_t uid, gid_t gid)
     if (unshare(CLONE_NEWNS | CLONE_NEWUSER) != 0)
     {
         // having multiple threads is a source of failure f.e.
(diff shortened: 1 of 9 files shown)


caolanm:
Fundamentally it seems that the bind remount readonly doesn't work in our ci container/vm, with or without namespaces enabled.

But with namespaces we assumed that the bind would work and that the tmp/cool-ever mounted as child-tmp is valid for the parent to use to put stuff into the child /tmp, which it didn't so we can't. So we should at least check that the full bind/remount/unmount works before considering BindMounting working, and disable the Namespace stuff if the BindMount is failing

timar (approved):
Implicit +1 from @mmeeks from yesterday.

PR #9500, opened 2024-07-15
Accessibility: Add ARIA roles for buttons

Darshan-upadhyay1110:
Changes are based on below points to improve accessibility

- The element that serves as the toolbar container has role toolbar.
- If the toolbar has a visible label, it is referenced by aria-labelledby on the toolbar element. Otherwise, the toolbar element has a label provided by aria-label.
- If the controls are arranged vertically, the toolbar element has aria-orientation set to vertical. The default orientation is horizontal.

Change-Id: I38f234461d4bf7982eec01ffe866c390256f50a0

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -216,7 +216,7 @@ m4_ifelse(MOBILEAPP,[true],
       </div>
      </nav>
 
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
looks ok

PR #9501, opened 2024-07-15
Remove the resizeDetector and use document-container for the event hand…

gokaysatir:
…ling.

Change-Id: I6990480da8ba3a058e4e5d32407bf38d341e8ee4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/.beforeprettier
+++ b/browser/.beforeprettier
@@ -92,6 +92,7 @@
 /src/core/Socket.js
 /src/core/Util.js
 /src/docstate.js
+/src/docevents.ts
 /src/dom/DomEvent.js
 /src/dom/DomEvent.MultiClick.js
 /src/dom/DomEvent.Pointer.js
--- a/browser/src/docevents.ts
+++ b/browser/src/docevents.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+    This is for listening to / firing document events.
+    Will be useful for custom events but native events should also be bound via this bridge for maintainability.
+    Search "app.events.on" and "app.events.fire" for examples.
+
+    For firing custom events:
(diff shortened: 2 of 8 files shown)


eszkadev (changes requested):
it's green week, let's wait for red. I'm blocking it for 1 day, so it is marked as reviewed

mmeeks (commented):
Would be lovely to see this go =)

eszkadev:
If you make window small, then open sidebar, then resize to bigger window. Sidebar stays small:
sidebar

browser/src/docevents.ts
```diff
@@ -0,0 +1,50 @@
...
+        // Resize event for a div element is not fired. We'll use a resize-observer for catching the event.
+        new ResizeObserver(this.fire.bind(this, 'resize')).observe(this.container);
+    }
+
+    public fire(eventType: string, details: any) {
+        const newEvent = new CustomEvent(eventType, details);
+        this.container.dispatchEvent(newEvent);
+    }
+
+    public on(eventType: string, callback: eventCallback) {
```
eszkadev, line 41: we need also `off` I think

browser/.beforeprettier
```diff
@@ -92,6 +92,7 @@
 /src/core/Socket.js
 /src/core/Util.js
 /src/docstate.js
+/src/docevents.ts
```
eszkadev, line 95: We can have later more and more "global app helpers".
Can we have directory for them maybe?
I think "app" would be good, it fits into the name we use for accessing them so can be quickly found by newcomers.
Also in the whole app we have convention to use CamelCase naming for files. Only in the main directory are some exceptions. I would prefer to keep all unified: `DocEvents.ts`

so: `app/DocEvents.ts` would be nice to begin the trend (later at new Big version we could move also docstate and docstatefunctions)

eszkadev: and please not include that file here, it is meant to keep only OLD files which we don't want to format with prettier

browser/src/docevents.ts
gokaysatir: I couldn't decide on this. Not sure if we have a real need for off functionality here. Let's keep that out for now, if we see a scenario that needs, then we can always add it quickly.

browser/.beforeprettier
gokaysatir: And thanks for checking:)

eszkadev (approved):

PR #9502, opened 2024-07-15
private/pedro/closebutton fix size margin add separator

pedropintosilva:
- **Close button: Fix tooltip**
- **Close button: introduce separator and fix size and margin**

### DIFF
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -1071,6 +1071,7 @@ function setupToolbar(e) {
 		map.uiManager.enableTooltip($('#closebutton'));
 	} else if (!L.Params.closeButtonEnabled) {
 		$('#closebuttonwrapper').hide();
+		$('#closebuttonwrapperseparator').hide();
 	} else if (L.Params.closeButtonEnabled && !window.mode.isMobile()) {
 		$('#closebuttonwrapper').css('display', 'block');
 	}
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -1067,8 +1067,8 @@ function setupToolbar(e) {
 
 	if (map.options.wopi && L.Params.closeButtonEnabled && !window.mode.isMobile()) {
 		$('#closebuttonwrapper').css('display', 'block');
-		$('#closebuttonwrapper').prop('title', _('Close document'));
-		map.uiManager.enableTooltip($('#closebuttonwrapper'));
+		$('#closebutton').prop('title', _('Close document'));
+		map.uiManager.enableTooltip($('#closebutton'));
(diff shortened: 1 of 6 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -790,6 +790,7 @@ L.Control.UIManager = L.Control.extend({
 		$('.main-nav').hide();
 		if (L.Params.closeButtonEnabled) {
 			$('#closebuttonwrapper').hide();
+			$('#closebuttonwrapperseparator').hide();
```
eszkadev, line 793: you forgot to add also "show" in above function showMenubar

browser/src/control/Control.Toolbar.js
```diff
@@ -1067,10 +1067,11 @@ function setupToolbar(e) {
...
 		$('#closebuttonwrapper').css('display', 'block');
-		$('#closebuttonwrapper').prop('title', _('Close document'));
-		map.uiManager.enableTooltip($('#closebuttonwrapper'));
+		$('#closebutton').prop('title', _('Close document'));
+		map.uiManager.enableTooltip($('#closebutton'));
 	} else if (!L.Params.closeButtonEnabled) {
 		$('#closebuttonwrapper').hide();
+		$('#closebuttonwrapperseparator').hide();
 	} else if (L.Params.closeButtonEnabled && !window.mode.isMobile()) {
 		$('#closebuttonwrapper').css('display', 'block');
```
eszkadev, line 1076: don;t we need display block here too?
possibly the default should be hidden state so we will not show it on mobile... have to check

browser/src/control/Control.UIManager.js
eszkadev: it works here on mode switch, but it is needed to make this functions work properly if used outside of that UI switch (eg. postmessage) I think

pedropintosilva: good catch, thanks!

browser/src/control/Control.UIManager.js
```diff
@@ -767,6 +767,7 @@ L.Control.UIManager = L.Control.extend({
 		$('.main-nav').show();
 		if (L.Params.closeButtonEnabled && !window.mode.isTablet()) {
 			$('#closebuttonwrapper').show();
+			$('#closebuttonwrapperseparator').show();
```
pedropintosilva, line 770: added `$('#closebuttonwrapperseparator').show();` to `showMenubar()` function

browser/src/control/Control.UIManager.js
pedropintosilva: added `$('#closebuttonwrapperseparator').show();` to `showMenubar()` function

eszkadev (approved):
looks good

PR #9505, opened 2024-07-15
calc: fix edit and remove hypelink outside of edit mode

JaumePujantell:
Right now .uno:EditHyperlink and .uno:RemoveHyperlink only work on
edit mode or inside a draw object, but LOK_CALLBACK_HYPERLINK_CLICKED
is dispatched outside of edit mode.

needed to discriminate the correct field.

Change-Id: I6ce3c21c60ea85dbe2e6e2523710ae1dcdff78a6


timar:
I think this can be merged, the core part is in already and the code looks good.

timar (approved):

timar:
Oh, I see it was a backport. Anyway... still good.

PR #9508, opened 2024-07-15
jsdialog: add tooltip to menubutton

eszkadev:
Fox example font color item on the notebookbar didn't have tooltip with the name of a button.


gokaysatir (approved):
Thanks for the improvement :)

PR #9509, opened 2024-07-15
fix EOL issue in transform Document Structure

aszucs3:
Send encoded JSON string instead of decoded.
This way it will not have special characters, and it will reach the core side even with a \n or \r inside ot it.

This way we can set multi-line text to a contentControl.

Change-Id: I8b32aeef522faf7f4597405bec9bbb37b99d53cf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
core side is in place now, so this looks sensible to me

PR #9511, opened 2024-07-16
fix UnitSaveTorture with linux namespaces

caolanm:
Make unit-crash work reliably again

Change-Id: I9a6f9e8db8f6edced18fe20367afc4d599a29ed2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
The "COOL 24.04 unit tests w/o cypress" has become particularly unreliable in ci with unit-crash failing a lot.

This is because we now detect earlier that bind mounting will fail by additionally in advance checking remount as well as mount+unmount. Bind mounting has basically always not worked on this ci config, but only at the remount stage, so coolwsd thought it worked and coolforkit discovered it didn't.

If coolwsd thinks bind mounting works, it reduces its timeout to restart coolforkit after a crash, so now it doesn't. unit-crash simulates such a crash, and that longer timeout for coolforkit to restart is 20 seconds, while the tests wait 10 seconds. So unit-crash fails more now that it did.

mmeeks (approved):
Looks very plausible - thanks ! =)

PR #9516, opened 2024-07-16
fix(hooks): Replace git diff-index call

Minion3665:
git-diff-index isn't always provided separately to the main git binary, so we can't necessarily use that as the command. We should always be able to use `git diff-index`, as the git binary will find files named `git-foo` when you run command `git foo`, so this will work whether the binary is separate or not

Change-Id: Id705855eedfd7e12c0706a9c7638e0bfb911b3fd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -40,7 +40,7 @@ sub check_whitespaces($)
             print STDERR "$filename:$lineno:$line\n";
         }
     }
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Yes, 'git diff-index' should work for everyone who was not living under a rock in the past 10 years. :-)

PR #9520, opened 2024-07-16
Private/caolan/usermount9

caolanm:
### Summary

Allow Namespace use even if BindMounting failed. In advance of potential fix for that problem to test standalone that a Namespace + Failed Bindmount works.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3052,6 +3052,10 @@ void lokit_main(
             jailLOInstallation.makeDirectory();
             const std::string loJailDestPath = jailLOInstallation.toString();
 
+            const std::string tempRoot = Poco::Path(childRoot, "tmp").toString();
+            const std::string tmpSubDir = Poco::Path(tempRoot, "cool-" + jailId).toString();
+            const std::string jailTmpDir = Poco::Path(jailPath, "tmp").toString();
+
             // The bind-mount implementation: inlined here to mirror
             // the fallback link/copy version bellow.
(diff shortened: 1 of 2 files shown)


caolanm:
Out lxc based ci container allows mount + unmount, but blocks our current remount flags, so right now whether using these linux namespaces or not (coolmount case), bind mounting fails in our ci and we end up using fallback linking.

Currently if bind mounting fails then we disable linux namespaces.

But we don't need to do that. If the linux namespace works, but bind mounting fails, we can continue to take advantage of the namespace not needing our coolforkit to have caps to chroot etc with the fallback filesystem linking

Typically for namespaces we have mounted /tmp in the child from a parent dir and for namespaces we always put files into that parent dir because the child mounts are hidden in the parent env. So if the mount didn't happen, link the fallback child /tmp dir to the same location it would normally have been mounted from and we can use the same location in the parent whether or not the bind mount from, or link to, worked.

vmiklos (approved):
I don't see anything obviously problematic with this.

> Out lxc based ci container allows mount + unmount, but blocks our current remount flags

And I assume this remount on the bind mount is also a leftover, or do we have a case where a remount is truly needed?

caolanm:
We want the remount, because we remount some of these bind mounts as readonly which is desirable. Its an annoying quirk that it has to be done in two passes, bind mount, then remount readonly and the 2nd pass fails due to some lxc apparmor rules. More to follow on that :-)

PR #9521, opened 2024-07-16
Private/caolan/remount flags

caolanm:
### Summary

Shrink our remount flags to the subset that we need to avoid container environment denials

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/CoolMount.cpp
+++ b/common/CoolMount.cpp
@@ -29,13 +29,11 @@
 
 #define MOUNT mount_wrapper
 #define MS_MGC_VAL 0
-#define MS_SILENT 0
 #define MS_NODEV 0
 #define MS_UNBINDABLE 0
 #define MS_BIND 1
 #define MS_REC 2
 #define MS_REMOUNT 4
-#define MS_NOATIME 8
 #define MS_NOSUID 16
 #define MS_RDONLY 32
 
@@ -81,9 +79,6 @@ int mount_wrapper(const char *source, const char *target,
 
     if(mountflags & MS_REMOUNT)
         freebsd_flags |= MNT_UPDATE;
-
-    if(mountflags & MS_NOATIME)
-        freebsd_flags |= MNT_NOATIME;
 
     if(mountflags & MS_NOSUID)
         freebsd_flags |= MNT_NOSUID;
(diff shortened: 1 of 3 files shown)


caolanm:
Regardless of whether it is linux mount namespace, or traditional coolmount, in our lxc container the readonly remount attempt (of our previously successful bind mount) fails with permission denied so we end up without bind mounting used.

This turns out to be because lxc has default apparmor rules of
https://github.com/lxc/lxc/blob/main/config/apparmor/abstractions/container-base
where the flags we use for remount don't match any of the allowed patterns. 
Where the closest match is:  "mount options=(ro,remount,bind,nodev,nosuid)"

And we additionally add MS_NOATIME, MS_SILENT and MS_REC

We already know, putting lxc aside, that in the linux namespace mount case an additional MS_NOATIME, etc. will result in EPERM on remounting something hosted in a toplevel [rel]atime mount. man 2 mount has 'An attempt was made to modify (MS_REMOUNT) the MS_RDONLY, MS_NOSUID, or MS_NOEXEC flag, or one of the "atime" flags (MS_NOATIME, MS_NODIRATIME, MS_RELATIME) of an existing mount, but the mount is locked'.

Otherwise, the MS_REC "recursive" only matters for the original mount and not the remount. And MS_SILENT is obscure and doesn't really matter. So if we use a more standard set of mount options we work out of the box with the default lxc configuration.

caolanm:
@timar this is the lxc mystery mount failure error FWIW

vmiklos (approved):
I like this, using the same flags in the namespace_mount case vs not simplifies the picture.

I would naively expect that in theory we could get the mount flags right for the first time and we don't ever need to remount, but possibly I miss something. And it's just a related thought, it's not something that needs instant simplification anyway.

caolanm:
I'ts unfortunately inherent that it takes two passes, one to create a bind mount, and another to make it readonly. man 2 mount, "Creating a bind mount" using MS_BIND on creation ignores flags like MS_RDONLY :-(

vmiklos:
Ah, I missed you can't do a read-only bind mount out of the box, thanks for clarifying.

PR #9522, opened 2024-07-16
Private/caolan/usermount10

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
cypress tests now work fine in mount namepaces, so don't need to force that off anymore

vmiklos (approved):
Nice, the less special-casing we need for the tested wsd, the better.

PR #9529, opened 2024-07-17
Statusbar: Zoom: set different icons

pedropintosilva:
Better to opt for a more minimal icon set for the zoom controls in the
statusbar otherwise we end up remanding too much attention for what it
should be instead a secondary control set.
- Improve minus and plus svgs and use them instead of the
normal (lc_zoomout and lc_zoomin) that are used in view tab
- Create zoomreset icon to be also used only on the statusbar

Change-Id: I24afe1b9e8e6098d1cb93d6d9ad246af8381609d


browser/src/control/Control.StatusBar.js
```diff
@@ -248,10 +248,10 @@ class StatusBar extends JSDialog.Toolbar {
 			{type: 'customtoolitem',  id: 'next', command: 'next', text: _UNO('.uno:PageDown', 'text')},
 			{type: 'separator', id: 'prevnextbreak', orientation: 'vertical'},
 		].concat(window.mode.isTablet() ? [] : [
-			{type: 'customtoolitem',  id: 'zoomreset', command: 'zoomreset', text: _('Reset zoom')},
-			{type: 'customtoolitem',  id: 'zoomout', command: 'zoomout', text: _UNO('.uno:ZoomMinus')},
+			{type: 'customtoolitem',  id: 'zoomreset', command: 'zoomreset', text: _('Reset zoom'), image: 'images/zoomreset.svg'},
+			{type: 'customtoolitem',  id: 'zoomout', command: 'zoomout', text: _UNO('.uno:ZoomMinus'), image: 'images/minus.svg'},
 			{type: 'menubutton', id: 'zoom', text: '100', selected: 'zoom100', menu: this._generateZoomItems(), image: false},
-			{type: 'customtoolitem',  id: 'zoomin', command: 'zoomin', text: _UNO('.uno:ZoomPlus')}
+			{type: 'customtoolitem',  id: 'zoomin', command: 'zoomin', text: _UNO('.uno:ZoomPlus'), image: 'images/plus.svg'}
```
eszkadev, line 254: just 'plus' doesn't work? or maybe it was other property name like: "img:".
Paths are generated by the toolitem widget, would be good to avoid hardcoding here

pedropintosilva: yes I did a mistake, it should be img instead, thanks

pedropintosilva:
image

pedropintosilva:
also cleaned/minified svgs, existing svgs were quite big (via scour)

eszkadev (approved):

PR #9532, opened 2024-07-17

plubius:
The iOS app (and presumably the Android app) do not have jails. Instead, they read and write files directly from the user's local file system.

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1162,11 +1162,12 @@ bool DocumentBroker::doDownloadDocument(const Authorization& auth,
         LOG_WRN("Failed to process plugins on file [" << localPath << ']');
         return false;
     }
+
+    const std::string localFilePath = Poco::Path(FileUtil::buildLocalPathToJail(COOLWSD::EnableMountNamespaces, getJailRoot(), localPath)).toString();
+#else  //!MOBILEAPP
+    const std::string localFilePath = Poco::Path(getJailRoot(), localPath).toString();
 #endif //!MOBILEAPP
 
-    const std::string localFilePath = Poco::Path(FileUtil::buildLocalPathToJail(COOLWSD::EnableMountNamespaces,
-                                                                                getJailRoot(),
-                                                                                localPath)).toString();
     std::ifstream istr(localFilePath, std::ios::binary);
     Poco::SHA1Engine sha1;
     Poco::DigestOutputStream dos(sha1);


caolanm:
There's a bunch of these buildLocalPathToJail uses now, so if we need special !MOBILE handling it probably needs to go inside the buildLocalPathToJail itself?

But what I would *expect* to happen currently is that buildLocalPathToJail has an initial arg of COOLWSD::EnableMountNamespaces of false, and with false it should already work the same as the past, like the else branch of this patch.

And we only set EnableMountNamespaces to true inside COOLWSD::setupChildRoot which is inside a #if !MOBILEAPP block, so I wonder those beliefs have gone wrong?

plubius:
> There's a bunch of these buildLocalPathToJail uses now, so if we need special !MOBILE handling it probably needs to go inside the buildLocalPathToJail itself?
> 
> But what I would _expect_ to happen currently is that buildLocalPathToJail has an initial arg of COOLWSD::EnableMountNamespaces of false, and with false it should already work the same as the past, like the else branch of this patch.
> 
> And we only set EnableMountNamespaces to true inside COOLWSD::setupChildRoot which is inside a #if !MOBILEAPP block, so I wonder those beliefs have gone wrong?

I will take a closer look at this today. I admit I didn't really debug this issue too deeply. Just found it via a git bisect so I assumed that the COOL server has its own file storage hierarchy.

caolanm:
Yeah, I've merged this stuff very recently so I'm wondering if I accidentally fixed this in a recent commit or if I'm missing a MOBILE scenario

plubius:
So I found the cause: without my patch using COOL's master branch, the following debug patch fixes the problem on iOS:

[debugios.patch](https://github.com/user-attachments/files/16287561/debugios.patch)

On iOS, Poco::File(rootPath).createDirectories() throws an "Access to file denied" exception when the rootPath already exists which is most (maybe all?) of the time.

So do we need do anything other than catch and log the exception? If no, I'll can replace the fprintf with a LOG_WARN() and submit a new patch.

caolanm:
Aha!, I see now.

How about adding a final ", bool createDir = false)" arg to buildLocalPathToJail and use true in the one place which used to do an explicit createDirectories in the past, which is the one of "StorageBase::getLocalRootPath", leaving the rest to default to false. And make the createDirectories() conditional on that and we are then as we always were.

plubius:
> How about adding a final ", bool createDir = false)" arg to buildLocalPathToJail and use true in the one place which used to do an explicit createDirectories in the past, which is the one of "StorageBase::getLocalRootPath", leaving the rest to default to false. And make the createDirectories() conditional on that and we are then as we always were.

Actually, the problem is that "rootPath" in this code is a garbage path on iOS:

jails/92621-a028d88a/jail//Users/pluby/Library/Containers/E3D583D5-4C8C-4F4D-8FCB-BDFCBB2B9120/Data/tmp/1/Blank.odt

The leading "jails/92621-a028d88a/jail/" makes no sense but "/Users/pluby/Library/Containers/E3D583D5-4C8C-4F4D-8FCB-BDFCBB2B9120/Data/tmp/1" does get created somewhere else in the code.

So, what we really need to do get rid of the leading "jails/92621-a028d88a/jail/". That's probably why my original patch worked: it eliminated prepending the jails string.

plubius:
Closing this pull request as the fix is now in pull request #9545.

PR #9540, opened 2024-07-18
private/pedro/private/pedro/status bar padding n search

pedropintosilva:
- **Status bar: Search input field should match the set height**
- **Status bar: Make the bar slimmer**
- **Status bar: Search input field add missing font-size**

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -398,6 +398,7 @@ nav.spreadsheet-color-indicator ~ #sidebar-dock-wrapper {
 }
 
 #search-input {
+	font-size: var(--default-font-size);
 	box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


Darshan-upadhyay1110 (approved):
Looks good! Thanks, Pedro :)

Darshan-upadhyay1110:
@pedropintosilva i guess we need to adjust `sheet_switch_spec.js` test case. As we made the bar slimmer so now we also need to change bound values for `assertScrollbarPosition`.

https://github.com/CollaboraOnline/online/blob/master/cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js#L24

image

So now we have to increase `upperBound` to maybe 335 currently it is 330.

PR #9545, opened 2024-07-18

plubius:
The iOS app (and presumably the Android app) do not have jails. Instead, they read and write files directly from the user's local file system.

Also, running the iOS in trace mode is nearly unusable so move the key "To JS" and "To Online" iOS messages to LOG_DBG and run the app with "debug" instead of "trace" by default.


caolanm (approved):

caolanm:
@Minion3665 if you could double check that all is well with this on android too

Minion3665:
Seems good for android to me, thanks!

PR #9546, opened 2024-07-19
chore(android): Update Android Gradle Plugin

Minion3665:
We've previously held off from updating this as the Valencian translation has been causing https://issuetracker.google.com/issues/293294923, which stopped the build unless you removed the translation

Unfortunately, I was getting some other build errors that were fixed by updating, so I have finally bitten the bullet and written a bit of Makefile to avoid the valencian translation being included in the Android app

Change-Id: Ic0ad52c1b31314a727bb6e74b88b0bb54c8b5775

---

@timar I'm not certain that weblate will like me deleting the `strings.xml` files in `android/` - do you have any recommendations to avoid it recreating them in the future?

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665:
The reason that CI is failing here is a Java version difference. I've built the Android app on bee, after switching the Java version, and the output is available as a snapshot.

When we merge this, we need to make sure to immediately switch the Java version.

(I wonder if we could in the future use something to keep track of package versions in the repo so changes like this don't cause an issue?)

Minion3665:
Realized Andras is on holiday, @pedropintosilva do you have any weblate-related thoughts? :)

pedropintosilva:
> @timar I'm not certain that weblate will like me deleting the strings.xml files in android/ - do you have any recommendations to avoid it recreating them in the future?

I have no idea : /

Rash419:
I also struggled with ca-valencian when I tried to update gradle verison to 8.0. Thanks for doing it. PR looks good but I have no idea about weblate question.

Minion3665:
> I also struggled with ca-valencian when I tried to update gradle verison to 8.0. Thanks for doing it. PR looks good but I have no idea about weblate question.

I'm not sure that this is a particularly easy-to-answer question then ... I would like to look into this a little, and then if we can't figure out what will happen I think we should merge this after the Android release (and on CODE "red") anyway ... I don't want to block updating forever on the chance that we might upset weblate ... it's somewhat of a pain to get all the old things needed to build Android now, and I want it to be possible for contributors to relatively easily build Android.

If weblate does cause us issues, we can give instructions on how to delete these files, do so in CI, etc. as we look for a solution

---

@pedropintosilva are there any permissions that I can get on weblate that would allow me to view more configuration than I can currently? Thanks!

Minion3665:
RE CI changes for this, we can't update the default Java version for everything since as some things depend on an older version, so I will set JAVA_HOME in the job ... I think that's the best without using something that can lock system packages to different versions

PR #9549, opened 2024-07-19
Statusbar: Zoom: set different icons via data.icon

pedropintosilva:
Better to opt for a more minimal icon set for the zoom controls in the
statusbar otherwise we end up remanding too much attention for what it
should be instead a secondary control set.
- Improve minus and plus svgs and use them instead of the
normal (lc_zoomout and lc_zoomin) that are used in view tab
- Create zoomreset icon to be also used only on the statusbar

Change-Id: I24afe1b9e8e6098d1cb93d6d9ad246af8381609d


pedropintosilva:
this is a backport of https://github.com/CollaboraOnline/online/pull/9529

Darshan-upadhyay1110 (approved):

PR #9550, opened 2024-07-19
embedded videos: fix path to the tmp directory

eszkadev:
recently we introduced multiple changes to the paths used by COOL server, it has to be adjusted so we will find the video file

request before:
`/jails/16613-ae50568c/HCUlXOt8uWVwjWey/tmp/5aempc.mp4`
request after:
`/jails/16613-ae50568c/tmp/cool-HCUlXOt8uWVwjWey/5aempc.mp4`


caolanm:
I think we want to use buildLocalPathToJail here

caolanm (approved):

PR #9552, opened 2024-07-19
private/pedro/private/pedro/Font of user tooltips has become serif

pedropintosilva:
- **Username popups: Do not rely on browser's font defaults**
- **Username popups: Add round corners and shadow**


gokaysatir (approved):

PR #9555, opened 2024-07-19
slideshow: fix videos

eszkadev:
fix in the fallback version of slideshow...
requires also: https://github.com/CollaboraOnline/online/pull/9550
so the request for a video will be successful

### DIFF
--- a/browser/src/slideshow/SlideShowPresenter.ts
+++ b/browser/src/slideshow/SlideShowPresenter.ts
@@ -152,11 +152,9 @@ class SlideShowPresenter {
 		window.addEventListener('keydown', this._onCanvasKeyDown.bind(this));
 
 		try {
-			this._slideRenderer = new SlideRendererGl();
-			this._slideRenderer.setup(canvas);
+			this._slideRenderer = new SlideRendererGl(canvas);
 		} catch (error) {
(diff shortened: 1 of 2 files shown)


browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -152,9 +152,11 @@ class SlideShowPresenter {
 		window.addEventListener('keydown', this._onCanvasKeyDown.bind(this));
 
 		try {
+			this._slideRenderer = new SlideRendererGl();
 			this._slideRenderer.setup(canvas);
 		} catch (error) {
-			this._slideRenderer.setup2d(canvas);
+			this._slideRenderer = new SlideRenderer2d();
+			this._slideRenderer.setup(canvas);
```
codewithvk, line 159: Is there any reason to create a setup method? If not, then we can accept the canvas as a constructor parameter so we don't need to call the setup method every time.

codewithvk (approved):
Code looks good.

browser/src/slideshow/SlideShowPresenter.ts
eszkadev: good point

Issue #7193, opened 2023-08-31
"Insert > Symbol" Dialog: Odd Alignment for Recent Characters


Tex2002ans:
### Describe the Bug

The "Recent Characters" little preview boxes in the:

- **Insert > Symbol**

dialog are placed very strangely:

![Collabora 23 05 3 - Insert Special Characters - Recent Characters Oddly Spaced](https://github.com/CollaboraOnline/online/assets/5774324/415cb762-7f51-4155-9c71-fdb5458c456d)

compared to the same menu in LibreOffice:

![LibreOffice 7 6 - Insert Special Characters - Recent Characters](https://github.com/CollaboraOnline/online/assets/5774324/25d66595-3909-4d7a-90d4-e109ca7371d7)

### Steps to Reproduce

1. Go to "Insert" Tab.
2. Press "Symbol" button.
3. Choose a few characters in the dialog by double-clicking on them.
4. Observe the "Recent Characters" put *huge* gaps between the characters.

### Actual Behavior

Boxes being centered(?).

### Expected Behavior

Left-aligned boxes.

### Desktop

- **OS:** Linux Mint 21.1
- **Browser:** Firefox
- **Version:** 117.0 (64-bit)

> COOLWSD version: 23.05.2.2 (git hash: 6d7efe98)
> LOKit version: Collabora Office 23.05.2.3 (git hash: 6d3919c)
> Served by: Linux Mint 21.1
> Server ID: ee10cf42

### Additional Context

None.

Some observations though:

- That menu could probably use a little touch-up though, even along the right-hand side.
- The dialog seems "very wide". On my widescreen, it made it a little awkward to place the dialog so I could still see what's happening in the document itself.
    - I suspect this'll be worse on narrower/vertical screens, or if you had COOL open along one-half of a screen.
- The scrollbar where all the characters are located is super-thin/nearly-invisible. If I didn't see the original dialog, I probably wouldn't have known you could scroll it.

Andreas-Kainz:
There are two options to fix this UX bug

1. define a fix number of items for recent and favorite characters to show, so add a new character will replace an "empty" character element. 
2. make the grid left align and add some grid-gap in the css file 

Nr. 1. need code change, and is a larger fix. I can offer an PR for 2. whit align the items (doesn't meter how much items will be there) well.

![Screenshot_20231227_003830](https://github.com/CollaboraOnline/online/assets/8517736/ac8c2213-42ed-4349-b43f-8b7a51336278)
[...]

PR #9558, opened 2024-07-21
Fix special characters widget

rparth07:
* Resolves: #7193
* Target version: master 

### Summary
- Correct the alignment issues for the "Recent Characters" and "Favorite Characters" sections to match it with libreoffice
- Resolve a potential bug where special characters are not displaying correctly due to an incorrectly named property.

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -182,6 +182,13 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	grid-gap: 4px;
 }
 
+/* Special Characters dialog */
+#SpecialCharactersDialog #viewgrid,
+#SpecialCharactersDialog #favgrid {
+	grid-gap: 4px;
+	width: fit-content;
+}
+
 /* Find and replace dialog */
 #search_grid, #replace_grid {
 	grid-template-columns: minmax(100px, min-content) auto !important;
--- a/browser/src/control/jsdialog/Widget.ScrolledWindow.js
+++ b/browser/src/control/jsdialog/Widget.ScrolledWindow.js
@@ -77,7 +77,7 @@ function _scrolledWindowControl(parentContainer, data, builder) {
 		scrollwindow.style.overflowX = 'scroll';
 
 	var realContentHeight = scrollwindow.scrollHeight;
-	var realContentWidth = scrollwindow.scrollwidth;
+	var realContentWidth = scrollwindow.scrollWidth;
 
 	var margin = 15;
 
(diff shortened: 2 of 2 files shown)


browser/css/jsdialogs.css
```diff
@@ -182,6 +182,10 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	grid-gap: 4px;
 }
 
+#viewgrid {
```
eszkadev, line 185: please add more precise selector, so it will also contain dialog's id (names could be repeated somewhere else)

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -3107,6 +3107,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			control.style.gridColumn = 'span ' + parseInt(data.width);
 		}
 
+		if(control && (data.id === 'viewgrid' || data.id === 'favgrid')) {
```
eszkadev, line 3110: please do not hardcode rules based on widget id

eszkadev: can't we do that in CSS?

eszkadev (changes requested):

browser/css/jsdialogs.css
rparth07: Thanks! updated the selector to include the dialog's Id.

browser/src/control/Control.JSDialogBuilder.js
rparth07: I’ve removed the hardcoded rules based on the widget Id and moved the relevant styling into the CSS file.

codewithvk (commented):
better to use general class ui-grid-cell

Something like this : ``` #SpecialCharactersDialog  .ui-grid-cell```

rparth07:
> better to use general class ui-grid-cell
> 
> Something like this : ` #SpecialCharactersDialog .ui-grid-cell`

We need to make sure our CSS changes target only `#viewgrid` and `#favgrid` without affecting other elements with the 
`.ui-grid-cell` class. Using a more general selector might unintentionally alter the styles of other elements that share the same class, which could lead to unintended layout changes and inconsistencies.

Would love to hear thoughts from @eszkadev :)

codewithvk:
@rparth07 
what I mean was something like this:
```
#SpecialCharactersDialog #box1 > .ui-grid-cell,
#SpecialCharactersDialog #box2 > .ui-grid {
	grid-gap: 4px;
	width: fit-content;
}
```

rparth07:
> @rparth07 what I mean was something like this:
> 
> ```
> #SpecialCharactersDialog #box1 > .ui-grid-cell,
> #SpecialCharactersDialog #box2 > .ui-grid {
> 	grid-gap: 4px;
> 	width: fit-content;
> }
> ```

Using IDs ensures precise and maintainable styling, targeting only the specific elements without unintended side effects. It provides clarity and reduces dependency on the HTML structure. Let me know if you have a different perspective.

eszkadev (approved):

PR #9561, opened 2024-07-22
JsDialogBuilder: add press and hold event for next and previous related commands

rparth07:
Change-Id: I3d284b87968d92bb356f90bd9c489cf79af13e15

* Resolves: #8358
* Target version: master 

### Summary
- add press and hold feature for next and previous related commands
- The solution introduces mouse scrolling and press + hold functionality for scrolling through records. It optimizes event handling by triggering the scrolling action after a delay when the left mouse button is pressed and held down. This is achieved by setting a timeout to execute the scrolling function after 500 milliseconds of continuous pressing. The scrolling action continues until the mouse button is released.

### Checklist


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2515,6 +2515,23 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		$(control.label).unbind('click');
 
 		if (!builder.map.isLockedItem(data)) {
+			//Allow "Press+Hold" To Move Faster for only "next" and "previous" related commands
+			if(data.id && (data.id.includes('next') || data.id.includes('prev'))) {
```
eszkadev, line 2519: Widget implementation should never contain any hardcoded ids.
Instead we would to handle "properties" of the widget.
So please put some property into `data` and based on that use new functionality.
Like in data `{... presAndHold: true; ... }`  then here  `if (data.pressAndHold) { ...} `

eszkadev (changes requested):
please remove ids from the implementation

browser/src/control/Control.JSDialogBuilder.js
rparth07: Thanks for pointing that out! I’ve updated the code to use properties from the data object instead of hardcoded IDs. The pressAndHold property is now used to control the functionality as you suggested.

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2524,6 +2541,16 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		builder._preventDocumentLosingFocusOnClick(control.container);
 	},
 
+	_handlePressAndHold: function(data) {
```
eszkadev, line 2544: can we put this inside `_mapDispatchToolItem` ? 
we are moving with widgets to be real classes/objects, I fear someone will forgot this when moving from this file.

`var handlePressAndHold = function(data) { ... };`

rparth07: sure, got it!

eszkadev (approved):
beautiful, thank you it's nice improvement :)
I think we will wait a bit with merging to put this in the next CODE version

PR #9562, opened 2024-07-22
[Backport] Fix dropdow arrows changing depending on users font stack

pedropintosilva:
With *45bd9c6 merged an improved was introduced but it still didn't
fix the alignment for all cases. Example if an integrator uses a
different font stack or the user doesn't have the 1st and the 2nd
default font it would fall back to something else. These differences
in font would change how the dropdown is aligned due to line-height.

On top of that, we were not accounting for the transparent border when
aligning. Set margin-top to 4px so to offset the existent transparent
border: 4px.

*45bd9c6:
- Vertical align dropdown arrows
- 45bd9c6009baf33eb7a4225c969c474edddaaa6c

Change-Id: Ie695fa3a4a3a7edf3ec2f15ba2565bb2079d16fa

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1271,6 +1271,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	border-top: 5px solid var(--color-main-text);
 	display: inline-block !important;
 	margin-left: 3px;
+	margin-top: 4px;
 }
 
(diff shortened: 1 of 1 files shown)


pedropintosilva:
this is a backport of https://github.com/CollaboraOnline/online/pull/9539

eszkadev (approved):

pedropintosilva:
retried `COOL 24.04 unit tests w/o cypress `

PR #9564, opened 2024-07-22
Private/caolan/backporting ns

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
backporting these fixes for these mount namespaces looks like a good idea

eszkadev (approved):

PR #9568, opened 2024-07-22
Fix canvas margin when ruler is toggled on

pedropintosilva:
Before this commit:
- Canvas margin would be ignored when showruler is triggered
  - Because of this the bg ruler couldn't be transparent (otherwise
  the canvas scrollbar would be visible overlapping it)

With this commit: Whenever ruler is visible -> re set the canvas
container y position

Change-Id: Id320a025675b92eb3b32734308a91dd6dd34dc43

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -451,9 +451,11 @@ L.Control.Notebookbar = L.Control.extend({
 	onRulerChange: function() {
 		if (this.map.uiManager.isRulerVisible()) {
 			$('#showruler').addClass('selected');
+			document.documentElement.style.setProperty("--canvas-container-y", "var(--ruler-height)");
 		}
 		else {
 			$('#showruler').removeClass('selected');
+			document.documentElement.style.setProperty("--canvas-container-y", "0");
 		}
 	},
 
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -42,6 +42,7 @@
 
 	/* Ruler */
 	--ruler-height: 20px;
+	--canvas-container-y: var(--ruler-height);
 }
 .focus-hidden:focus {
 	outline: none;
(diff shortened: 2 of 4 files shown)


pedropintosilva:
I just merged this because I saw everything green (approved and checks) but I now see that you @eszkadev didn't manage to also review it. Could you do it just for extra sanity

Darshan-upadhyay1110:
> I just merged this because I saw everything green (approved and checks) but I now see that you @eszkadev didn't manage to also review it. Could you do it just for extra sanity

@pedropintosilva i have tested the patch, looks good to me.

browser/src/control/Control.Notebookbar.js
```diff
@@ -451,9 +451,11 @@ L.Control.Notebookbar = L.Control.extend({
 	onRulerChange: function() {
 		if (this.map.uiManager.isRulerVisible()) {
 			$('#showruler').addClass('selected');
+			document.documentElement.style.setProperty("--canvas-container-y", "var(--ruler-height)");
```
eszkadev, line 454: I think we should prepare some helper for such things:
`app.setCssVariable("name", "value")`
and
`app.setGlobalCssClass("class-name")` - which will put class into root element in the DOM, like we use to later style app depending on app-type, mode, etc

pedropintosilva: understood

pedropintosilva: but it that case we can get away without any new helper I can just apply a new class that would probably be more useful for other things in the long run (so CSS knows when the ruler is on or off just by the class selector

eszkadev: yes, but helper will be nice - one unified way to add these "flags", so they are in one place and it is easy to target them then in CSS.
Also easier to find them in the code by searching the function name.
It's just an idea for later.

PR #9569, opened 2024-07-22
Status bar: Fix language separator (divider) not being rendered and dim toolbar-down, toolbar-up, ruler's border

pedropintosilva:
- **Status bar: Fix language separator (divider) not being rendered**
- **Toolbars and ruler: Dim toolbar-down, toolbar-up, ruler's border**


pedropintosilva:
Good catch. Yes I will do that!

eszkadev (approved):
looks ok now

PR #9572, opened 2024-07-22
systemplate and its immediate subdirs should be owned by 'cool'

caolanm:
otherwise post-install they can't be updated by coolforkit if they fall out of sync.

f.e.

service coolwsd stop
rm /opt/cool/systemplate/etc/hosts
service coolwsd start

```
[ coolforkit ] DBG  File contents mismatch: [/opt/cool/systemplate//etc/hosts] missing, 0 bytes, modified at 0 =/= [/etc/hosts]: exists, 423 bytes, modified at 1721313302| common/FileUtil.hpp:301
[ coolforkit ] INF  No write access to path [/opt/cool/systemplate]: Permission denied| common/FileUtil.cpp:305
[ coolforkit ] WRN  The systemplate directory [/opt/cool/systemplate] is read-only, and at least [/opt/cool/systemplate//etc/hosts] is out-of-date. Will have to copy sysTemplate to jails. To restore optimal performance, make sure the files in [/opt/cool/systemplate/etc] are up-to-date.| common/JailUtil.cpp:557
[ coolforkit ] WRN  Failed to update the dynamic files in [/opt/cool/systemplate]. Will disable bind-mounting in this run and clone systemplate into the jails, which is more resource intensive.| common/JailUtil.cpp:496
```
And in a podman scenario it's practically guaranteed that /etc/hosts etc are out of date at launch time.

So set the default owner of /opt/cool/systemplate and its immediate subdirs to cool:cool. The contents can remain at whatever owner they are created by to retain hard links.

Change-Id: Ib4a28c46bb2c0368693bf6b7ea6a61f48d33c15a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -490,14 +490,13 @@ void setupDynamicFiles(const std::string& sysTemplate)
     const std::string etcSysTemplatePath = Poco::Path(sysTemplate, "etc").toString();
     LinkDynamicFiles = true; // Prefer linking, unless it fails.
 
-    if (!updateDynamicFilesImpl(sysTemplate))
+    const bool uptodate = updateDynamicFilesImpl(sysTemplate);
+    if (!uptodate)
     {
         // Can't copy!
         LOG_WRN("Failed to update the dynamic files in ["
                 << sysTemplate
-                << "]. Will disable bind-mounting in this run and clone systemplate into the "
-                   "jails, which is more resource intensive.");
-        disableBindMounting(); // We can't mount from incomplete systemplate.
+                << "]. Will clone dynamic elements of systemplate to the jails.");
         LinkDynamicFiles = false;
     }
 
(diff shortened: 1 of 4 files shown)


mmeeks (approved):
So - this looks plausible; I'd be happy if we had a UID that didn't allow overwriting of any files in /etc etc. [sic] when we've dropped capabilities inside the namespace.

We should probably also ensure that those files are owned by root or somesuch in the case where we have capabilities - can you check that ?

mmeeks (approved):
Looks very sensible =)

caolanm:
How about this instead. We consider the systemplate out of date if the etc files are outdated. So rather than change any existing ownerships of the shared stuff to 'cool', instead mount the out of date systemplate as / anyway but mount an up to date per-jail etc copy over the out of date one.

PR #9576, opened 2024-07-22
cals: fix jump after search

eszkadev:
calc: fix view jump after search
    
    _searchRequested variable was set on wrong object
    also when we requested search - allow to jump

---------------

Fix jsdialog error on search


thebearon (approved):
Thanks for the fix, Szymon!

PR #9582, opened 2024-07-23
UI: replace conditional format dialog with conditiona manager

lpranam:
Change-Id: I3e8ccdb50702bb940db290186d646d5a23f7169c

* Target version: master 

### Checklist


eszkadev:
```
Checking formatting...
[warn] src/control/jsdialog/Definitions.Menu.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

eszkadev (approved):
@pedropintosilva do you agree we can do that now?

PR #9584, opened 2024-07-24
calc: jump to cell on address input

eszkadev:
To test:
type into "address" input next to the the formula bar
view should jump to the target cell


Darshan-upadhyay1110 (approved):
Tested and works fine 👌🏼. Thanks, @eszkadev

PR #9586, opened 2024-07-24
calc: jump to cell on address input (backport 24.04)

eszkadev:
To test:
type into "address" input next to the the formula bar
view should jump to the target cell


eszkadev:
not needed, have better fix: https://github.com/CollaboraOnline/online/pull/9625

PR #9588, opened 2024-07-24
drop storage.ssl.key_file_path

caolanm:
do we really want ssl *client* private certs

Change-Id: I4af2714acb78a2721465724cf726abb437badeaf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
lets drop this, I guess there is a theoretical possibility that the wopi server could enforce that the client provides a cert

PR #9590, opened 2024-07-24

Minion3665:
This is a backport of #9545

The iOS app (and presumably the Android app) do not have jails. Instead,
they read and write files directly from the user's local file system.

Also, running the iOS in trace mode is nearly unusable so move the key
"To JS" and "To Online" iOS messages to LOG_DBG and run the app with
"debug" instead of "trace" by default.

Signed-off-by: Patrick Luby <guibomacdev@gmail.com>
Change-Id: Ic8aabc09b297b15207be49c0e674e8006bcaa367


eszkadev (approved):
looks reasonable

PR #9591, opened 2024-07-24
SlideShow: adjust in-window presenter

eszkadev:


browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -214,12 +254,70 @@ class SlideShowPresenter {
...
-	_onFullScreen() {
+	_doInWindowPresentation() {
+		const popupTitle =
+			_('Windowed Presentation: ') + this._map['wopi'].BaseFileName;
+		const htmlContent = this._generateSlideWindowHtml(popupTitle);
+
+		this._slideShowWindowProxy = window.open('', '_blank', 'popup');
+
+		if (!this._slideShowWindowProxy) {
+			this._map.uiManager.showInfoModal(
```
codewithvk, line 270: I can see, `showInfoModal` is deprecated. ref: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L1212 

Can we use JSDialog.showInfoModalWithOptions here?

codewithvk (approved):
Otherwise, Code looks good.

browser/src/slideshow/SlideShowPresenter.ts
eszkadev: good catch, yes it should  be, but I just moved the existing code from the previous slideshow here.
let's do it later

PR #9593, opened 2024-07-24
Fix canvas margin when ruler is toggled on

pedropintosilva:
Before this commit:
- Canvas margin would be ignored when showruler is triggered
  - Because of this the bg ruler couldn't be transparent (otherwise
  the canvas scrollbar would be visible overlapping it)

With this commit: Whenever ruler is visible -> re set the canvas
container y position

Change-Id: Id320a025675b92eb3b32734308a91dd6dd34dc43

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -42,6 +42,7 @@
 
 	/* Ruler */
 	--ruler-height: 20px;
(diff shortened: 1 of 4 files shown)


eszkadev (approved):
looks ok

PR #9597, opened 2024-07-24
uno: fix CallPythonScript calling js error

lpranam:
we send "vnd.sun.star.script" message without uno prefix via uno command

Change-Id: If0e103bd74ac7f34a57559b2eebc1001352b0057

* Target version: master 

### Checklist


eszkadev (approved):
looks reasonable

PR #9599, opened 2024-07-24
private/pedro/pushbutton fix input focus offset

pedropintosilva:
- **Offset outline in input fields placed inside of a dialog**
- **PushButtonControl: Fix alignment (regression from locking overlay)**


lpranam (approved):

pedropintosilva:
This is a front end only PR and thus no need to block this because of CPP analyze

PR #9600, opened 2024-07-24
Calc: sheet tabs: center tabs

pedropintosilva:
Change-Id: I467acc1ec81fdccd48d76a52bb7bbc39d28e1ba6

### DIFF
--- a/browser/css/spreadsheet.css
+++ b/browser/css/spreadsheet.css
@@ -2,13 +2,14 @@
 	margin: 0;
 	padding: 0;
 	position: relative;
-	height: 39px;
+	height: 38px;
 	overflow: hidden;
 	width: 100%;
 	white-space: nowrap;
 	text-align: start;
 	background-color: transparent;
 	vertical-align: middle;
+	align-content: center;
 }
 
 .spreadsheet-tab-scroll {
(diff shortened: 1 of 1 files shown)


bayramcicek (approved):
|before|after|
|-|-|
| tabs-before | tabs-after |

-compared to previous one- looks good to me on my side. 
it seems(?) tabs are not exactly centered vertically and there is a bit more space on the top(I may be wrong); but this can be ignored(?) I think. (tried to fix this but couldn't find where the issue)

Thanks.

pedropintosilva:
> before 	after
> tabs-before 	tabs-after
> 
> -compared to previous one- looks good to me on my side. it seems(?) tabs are not exactly centered vertically and there is a bit more space on the top(I may be wrong); but this can be ignored(?) I think. (tried to fix this but couldn't find where the issue)
> 
> Thanks.

Thank you for testing it ! I have found another problem. Not really CSS alignment but just the height that we were using, changed 39 to 38 . I have also added a more detailed commit message

bayramcicek (approved):

pedropintosilva:
This is a front end only PR, no need to block this because of cpp analyze test

PR #9605, opened 2024-07-24
assume mount namespace of true if nocaps requested

caolanm:
if they work, then use those as achieving the goal of running without "outside" capabilities.

Change-Id: I6d6f03bfd386efdab098d02f5cb46ddebbc3d6ea

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks very sensible =) thanks Caolan !

caolanm:
@hcvcastro FYI as I think this is your typical local setup, so you might experience anything going wrong first.

PR #9607, opened 2024-07-25
following: fix status update in writer

eszkadev:
We want to follow our own cursor if we see it on the screen. We used calc's cell cursor always, but in Writer we need text cursor.

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -998,8 +998,14 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._move();
 		this._moveInProgress = false;
 
-		if (app.calc.cellCursorVisible || app.file.textCursor.visible) {
-			var cursorPos = this._map._docLayer._twipsToLatLng({ x: app.calc.cellCursorRectangle.x2, y: app.calc.cellCursorRectangle.y2 });
+		var isCellCursorVisible = app.calc.cellCursorVisible;
+		var isTextCursorVisible = app.file.textCursor.visible;
+		if (isCellCursorVisible || isTextCursorVisible) {
+			if (isCellCursorVisible)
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks very sensible - thanks Szymon - @thebearon can we get it into the COOL release too if its affected ?

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -998,8 +998,14 @@ L.CanvasTileLayer = L.Layer.extend({
...
 
-		if (app.calc.cellCursorVisible || app.file.textCursor.visible) {
-			var cursorPos = this._map._docLayer._twipsToLatLng({ x: app.calc.cellCursorRectangle.x2, y: app.calc.cellCursorRectangle.y2 });
+		var isCellCursorVisible = app.calc.cellCursorVisible;
+		var isTextCursorVisible = app.file.textCursor.visible;
+		if (isCellCursorVisible || isTextCursorVisible) {
+			if (isCellCursorVisible)
+				var cursorPos = this._map._docLayer._twipsToLatLng({ x: app.calc.cellCursorRectangle.x2, y: app.calc.cellCursorRectangle.y2 });
+			else
+				cursorPos = this._map._docLayer._twipsToLatLng({ x: app.file.textCursor.rectangle.x2, y: app.file.textCursor.rectangle.y2 });
```
Minion3665, line 1007: I understand that the `var` is hoisted here, so this is evaluated the same as declaring at the top of the function ... but it's still a little strange. May you declare it above (possibly using let) or add another var to the other cursorPos?

eszkadev: We have many places like that (eslint didn't allow to put 2 times `var` as yes it's hoisted so one time is enough anyway), in the new TS code I use const/let, in old one - I use var and consistent formatting with existing code, so in case of backport it's easier (no ES6 in the previous releases).

Thank you for a comment, but I think we will slowly extract pieces outside old JS files and then we apply prettier and const/let.
For example I have plan to create separate class for handling following to avoid current duplication in map._setFollowing and app.setFollowing at the same time...

PR #9608, opened 2024-07-25
following: fix status update in writer (backport 24.04)

eszkadev:
We want to follow our own cursor if we see it on the screen. We used calc's cell cursor always, but in Writer we need text cursor.


thebearon (approved):
Looks good, thanks!

PR #9609, opened 2024-07-25
[Backport] Notebookbar: Promote Reset zoom to bigmenubartoolitem

pedropintosilva:
The zoomreset button was occupying 100% of the vertical space but was
being render in one line instead of a big icon. Better to render it as
a big icon to aid in grouping.

Note on future implementations: Ideally we would actually have in one
line and we would replace the zoom out and zoom in btn with one
menu (something like `ZoomMenu`) that would allow the user to change
the zoom level (similar as already happens in the status bar). The
menu definition would be moved to menuDefinitions.

Change-Id: Ia136aa95e410f192984629feeb8eb05922372226


eszkadev (approved):

pedropintosilva:
backport of #9596

PR #9610, opened 2024-07-25
[24.04] Revert "Fix canvas margin when ruler is toggled on"

pedropintosilva:
This reverts commit 2e00fea4b6aba889834b3ee037565930d42f32bd.

So to fix the text cursor being in wrong place:
1. User A switches ruler on
2. User B Switches ruler off
3. Users click in line 1 but the blinking cursor is rendered in the
line 2

It seems we cannot change the top position of
.leaflet-canvas-container without introducing a regression affecting
cursors. Probably browser/src/layer/marker/Cursor.ts need to be
updated also and maybe other markers also such as form controls.

Change-Id: Ice70463934b0065b5eda4429c4a51995aeb9a4df

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -42,7 +42,6 @@
 
 	/* Ruler */
 	--ruler-height: 20px;
-	--canvas-container-y: var(--ruler-height);
 }
 .focus-hidden:focus {
 	outline: none;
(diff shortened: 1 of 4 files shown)


pedropintosilva:
orginal PR:
* Master: https://github.com/CollaboraOnline/online/pull/9568
* 24.04: https://github.com/CollaboraOnline/online/pull/9593

PR #9615, opened 2024-07-25
Status bar: Presentation there is no insertmode nor selection status

pedropintosilva:
So, don't show those items by default. For the other apps we already
specifically show them within the switch case in onDocLayerInit()

Change-Id: If25b09a8a37bcd5270f421f1d07d6e381be1d3d2


eszkadev (approved):

pedropintosilva:
cypress has tested, this is not related to core so we can merge it

PR #9617, opened 2024-07-25
[Backport] Calc: sheet tabs: center tabs

pedropintosilva:
- Don't rely on margin to align it
  - We are using flex box, add missing align property
- Fix height of containers: Use even numbers otherwise we end up with
  - non equal margins: (39-32)/2 = 3.5. Use 38 instead.

Change-Id: I467acc1ec81fdccd48d76a52bb7bbc39d28e1ba6

### DIFF
--- a/browser/css/spreadsheet.css
+++ b/browser/css/spreadsheet.css
@@ -2,13 +2,14 @@
 	margin: 0;
 	padding: 0;
 	position: relative;
-	height: 39px;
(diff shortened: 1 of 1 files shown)


pedropintosilva:
backport of https://github.com/CollaboraOnline/online/pull/9600

bayramcicek (approved):
Tested and looks good to me. Thanks.

PR #9618, opened 2024-07-25
if an "advance load" fails. e.g. due to a timeout, it cannot be reloaded

caolanm:
...for the lifetime of coolwsd. The docbroker sticks around, in "isAlive" state with _stop of false.

WopiStorage::downloadDocument (wsd/wopi/WopiStorage.cpp) throws a StorageConnectionException exception so we leave pollThread on the failed downloadAdvance attempt without any cleanup and presumably then appear to later attempts as an active docbroker whose session should be joined to.

Change-Id: I16bc2648657ba0d1f3903f049483f8ff9c88a19e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -313,8 +313,26 @@ void DocumentBroker::pollThread()
     // Download and load the document.
     if (_initialWopiFileInfo)
     {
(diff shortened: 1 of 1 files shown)


caolanm:
I'll do a test as a follow up

PR #9620, opened 2024-07-25
insertfile: show more logs when we fail

eszkadev:
Also log arguments used

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1634,6 +1634,8 @@ bool ChildSession::insertFile(const StringVector& tokens)
 
     SigUtil::addActivity(getId(), "insertFile " + type);
 
+    LOG_TRC("InsertFile with arguments: " << type << ": " << (data.empty() ? name : std::string("binary data")));
+
     if (type == "graphic" || type == "graphicurl" || type == "selectbackground")
     {
         std::string url;
@@ -1661,6 +1663,8 @@ bool ChildSession::insertFile(const StringVector& tokens)
                     return false;
                 }
             }
+            else
+                sendTextFrameAndLogError("error: cmd=insertfile kind=syntax");
         }
         else
         {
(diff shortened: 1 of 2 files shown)


eszkadev:
It would be good to merge this or adjust & merge something as a replacement to help debugging such fails - for a next CODE.

net/WebSocketHandler.hpp
```diff
@@ -784,7 +784,10 @@ class WebSocketHandler : public ProtocolHandlerInterface
                   [[maybe_unused]] unsigned char flags, bool flush = true) const
     {
         if (!socket || data == nullptr || len == 0)
+        {
+            LOG_DBG("Socket or data missing. Cannot send WS frame");
```
Ashod, line 788: I wonder how noisy this might be in some cases. We have to keep an eye, in case it spams when we shutdown a socket or some edge-case like that.

Regarding insertfile, that command shouldn't have anything special that relates to the socket's functionality. I seriously doubt if this log will help at all. It's unlikely that everything is perfectly fine and we lose the socket only when insertfile is executed.

kit/ChildSession.cpp
```diff
@@ -1634,6 +1634,8 @@ bool ChildSession::insertFile(const StringVector& tokens)
 
     SigUtil::addActivity(getId(), "insertFile " + type);
 
+    LOG_TRC("InsertFile with arguments: " << type << ": " << (data.empty() ? name : std::string("binary data")));
```
Ashod, line 1637: Why not put this at the top of the function?

Ashod (approved):
Thanks, @eszkadev. I'm not sure how much this will help, but let's hope these logs reveal something new.

kit/ChildSession.cpp
eszkadev: because I want to log params which we get in above lines

PR #9624, opened 2024-07-26
docker: update URLs used (22.05)

thebearon:
Change-Id: I39cc4329fcb640f853d691663fe3ef92619eafbe (cherry picked from commit 43563142cea05309058216a2204a1b75e724096c)

* Target version: 22.05


timar (changes requested):
We have different plans, so we will have to revert these patches on other branches, too. collaboraoffice.com will remain the download site for packages.

Issue #9631, opened 2024-07-26
Insert comment box looks odd in dark mode


pedropintosilva:
Originally reported by @jazevedo-coll 

![image](https://github.com/user-attachments/assets/3e618b1a-384f-4624-bd89-f5d104fd6ebe)

COOLWSD version: 24.04.4.5snapshot (git hash: eb91527) 
LOKit version: Collabora Office 24.04.4.20240711 (git hash: c16803a) 
Served by: Ubuntu 22.04.4 LTS Server ID: cf49b7f3

Steps to reproduce.

Switch to dark mode on the document.
insert a comment
Fill out the comment and save
Return to the comment and then it is normal

This happens on both Firefox and chromium

thebearon:
I believe this can be closed after the fixes.

PR #9635, opened 2024-07-26
Annotations (comment): do not change background on hover

pedropintosilva:
We already have high brightness on light mode and low brightness on
dark mode for comment background -> no need to try to increase it
further on hover since there is no high/low values to claim.

It fixes weird hover effect on the whole card.

Also related with
https://github.com/CollaboraOnline/online/issues/9631

Change-Id: I102bf161f90e6059bd6d981abb00d2abc854193d


pedropintosilva:
this is a background color removal property no need to wait for any tests

bayramcicek (approved):
Tested and LGTM. Thanks!

**Before**:

https://github.com/user-attachments/assets/d8421325-e88f-4951-9828-36e85d0e5135

**After**:


https://github.com/user-attachments/assets/698d1faf-3c89-4c10-a26f-c33f35eeb8cf

PR #9639, opened 2024-07-26
following: simplify followUser

eszkadev:
The UI component should only trigger map._setFollowing with correct parameters. Don't do own view jump management. This will help to finally centralize the following feature.

This patch fixes the issue with jumping to own cursor on any key press, example:
1. open writer with many pages
2. scroll to other page so you don't see the cursor
3. pres CTRL on the keyboard Result: jump back to the cursor
Expected: nothing happens

Keypress should activate following own cursor, but wait with the jump for a signal from the core

### DIFF
--- a/browser/src/control/Control.UserList.ts
+++ b/browser/src/control/Control.UserList.ts
@@ -121,7 +121,7 @@ class UserList extends L.Control {
 		app.setFollowingOff();
 	}
 
-	followUser(viewId: number) {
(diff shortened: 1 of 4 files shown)


quikee (commented):
Looks good to me

eszkadev:
unrelated mobile fails due to core's commit a26c07e37ed2e5996a5f1e07a23fa4f7ddb96534

PR #9641, opened 2024-07-26
Fix view jumping on keypress in writer (backport 24.04)

eszkadev:
cypress: check new page view jump in writer
    
    ---------------------------------------------------------------

    following: simplify followUser
    
    The UI component should only trigger map._setFollowing
    with correct parameters. Don't do own view jump management.
    This will help to finally centralize the following feature.
    
    This patch fixes the issue with jumping to own cursor on any
    key press, example:
    1. open writer with many pages
    2. scroll to other page so you don't see the cursor
    3. pres CTRL on the keyboard
    Result: jump back to the cursor
    Expected: nothing happens
    
    Keypress should activate following own cursor, but wait with
    the jump for a signal from the core

### DIFF
--- a/browser/src/control/Control.UserList.ts
+++ b/browser/src/control/Control.UserList.ts
@@ -121,7 +121,7 @@ class UserList extends L.Control {
 		app.setFollowingOff();
 	}
 
-	followUser(viewId: number) {
+	followUser(viewId: number, instantJump: boolean = true) {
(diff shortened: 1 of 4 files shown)


eszkadev:
requires https://github.com/CollaboraOnline/online/pull/9608 to work

thebearon (approved):
Thanks, Szymon!

PR #9645, opened 2024-07-29
Font of user tooltips has become serif (24.04)

thebearon:
* Target version: 24.04

### Checklist


Darshan-upadhyay1110 (approved):
Thanks, @thebearon !

PR #9647, opened 2024-07-29
wsd: avoid nested startActivity() in startSwitchingToOffline()

vmiklos:
Similar to commit d04d6bd392c849e868dc1224c82788cf1fc3b2ad (wsd: avoid
nested startActivity(), 2024-07-17).

Change-Id: Icff6f509100c672b2696f76b3c6cb43a94a08af6


mmeeks (approved):
Unusual wasm-only code-path - lets get it in =)

PR #9651, opened 2024-07-29
Private/caolan/add test for poisoned load

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -10,6 +10,7 @@
 #include <atomic>
 #include <cassert>
 #include <chrono>
+#include <map>
 #include <memory>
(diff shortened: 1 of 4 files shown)


caolanm:
CodeQL is failing for every build recently. Otherwise this adds a test for https://github.com/CollaboraOnline/online/pull/9618 Its a pain to get the right async download code path triggered from a unittest so I've split that out and explicitly call it when testing wants it.

vmiklos (approved):
Looks OK; and probably codeql cpp should not run tests in the first place, but that's not for this PR.

Issue #9627, opened 2024-07-26
Calc: Invert Background does not alter the background color of cells


pedropintosilva:
Changing of the cell background when "Invert Background" is selected in Calc.

Steps:

1. Open .ods file. Switch to Dark mode
2. Select "Invert Background".

| note | Screen cast |
| -------| ---------------- |
| Notice that the text color in the cells changes, but the background color of the cells does not change | [Screencast_from_18-06-2024_07_25_11.webm](https://github.com/user-attachments/assets/a6702c46-10da-4faf-a8b8-5d471589b441) | It works better in writer | [Screencast_from_20-06-2024_11_48_53.webm](https://github.com/user-attachments/assets/8c28556a-22c8-4151-aec0-0b96180075c1) |

pedropintosilva:
a fix got in https://github.com/CollaboraOnline/online/pull/9344 by @GulsahKose 

but @timur-g reported:

| 1st user | ![timur1](https://github.com/user-attachments/assets/d58dc89c-1784-4c2e-8666-1892b99ab9c0) | change light OK | ![timur2](https://github.com/user-attachments/assets/39634031-75fe-46eb-8842-2e2ac4a1fb70) | 1st dark, 2nd NOK | ![timur3](https://github.com/user-attachments/assets/c6febbf0-eec8-495e-a047-14882eb62aa3) | light OK | ![timur4](https://github.com/user-attachments/assets/cc524702-968f-4748-996e-a4c854e4e523) | 1st light, 2nd OK | ![timur5](https://github.com/user-attachments/assets/75d7778a-e584-470a-9ebd-456a66a9f996) | light NOK | ![timur6](https://github.com/user-attachments/assets/6f308083-9613-4d5a-8d71-6a0c264cca51)

This is improved and works for single user. While additional users are separated, it does not seem well, as shown in screenshots of the same file open by 2 users, also tested with 3.

timur-g:
Fully fixed in 24.04.6. Ref https://github.com/CollaboraOnline/online/pull/9697.

PR #9652, opened 2024-07-29
fix(invert): Specify desired theme

Minion3665:
This change makes our calls to .uno:InvertBackground specify the desired
theme (requires core change I05486860c5f562c3cfa59b4a7fc492d48913a181),
saves the background theme for different color themes, applying it when
they are loaded, and sends the previous theme on document load (requires
core change I8c22c03d3b4589736d48509004f7789dd5166389)

We need to send the desired background theme manually, as otherwise Core
looks at a global document background theme, meaning that switching your
background theme actually switches to whatever wasn't the last background
theme set by anyone

It's also nice to preserve this inverted document background, both when
switching in/out of dark mode and when reloading the document entirely.

Storing the desired preference on a per-theme basis avoids us having a
dark background in light mode when we enabled inversion in dark mode,
but means we need to enable the preference whenever we switch theme.

---

Additionally, I noticed another issue with our saving/loading, so here's a more general fix to that:

fix(prefs): Cache localStorage values on read

Sometimes we'll read a value, change it in another tab, and later read
an inconsistent value from preferences. This can do things like making
the invert background button need to be pressed twice to change whether
the background is inverted, if you change it in another tab

It's better to only use localStorage for saving/loading values and use
something else to read from where possible as this makes sure the
preference values we read will be stable unless we change them

To reflect it no longer storing only changes, I've changed the name of
the _localStorageChanges to _localStorageCache

---

Restoring the theme on load/theme change doesn't seem to work properly in calc, however as there are other similar issues with invert background in calc (e.g. #9627) I think it's better to get this in and then make another change fixing the calc issues here...

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -45,7 +45,9 @@
 	--color-warning: #eca700;
 	--color-success: #46ba61;
 
-	--color-cursor-blink-background: #fff;
-
 	--color-box-shadow: rgba(0, 0, 0, 0.5);
+}
+
+[data-bg-theme='dark'] {
+	--color-cursor-blink-background: #fff;
 }
--- a/browser/css/color-palette.css
+++ b/browser/css/color-palette.css
@@ -46,7 +46,9 @@
 	--color-warning: #eca700;
 	--color-success: #46ba61;
 
-	--color-cursor-blink-background: #000;
-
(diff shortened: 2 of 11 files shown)


caolanm (approved):
I think this is all sane, we can push when the core changes are merged

Minion3665:
Core changes are in... I will rerun CI one more time to make sure that we don't break anything (although stuff did pass locally, so I think it should be good...)

PR #9653, opened 2024-07-29
github, codeql: no need to run tests

vmiklos:
Those are invoked in other checks (on Jenkins) and recently one unit
test started to fail only here, so disable that till the problem is
understood there.

Change-Id: I57704c72ed239aadc460de1d2fd67b5cf0bc3597

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -77,7 +77,7 @@ jobs:
 
     - if: matrix.language == 'cpp'
       name: Build
-      run: make -j `nproc`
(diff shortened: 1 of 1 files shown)


vmiklos:
@caolanm could you please review this? I assume that this codeql thing is not for the tests, but for the action after the build.

CC @Ashod

caolanm (approved):
FWIW I tried on Friday to pull the image that CodeQL appears to use and tried to build manually in it to reproduce this. But I couldn't get the same failure of FileServeTests::testPreProcessedFileSubstitution

PR #9657, opened 2024-07-29
SlideShowPresenter: some fixes

mcecchetti:
Draw a white rectangle behind a slide or shape with a not
solid color will not look correct if the slide has no background.

Set canvas to null when presentation is over also for the present in a
window case or we can't play presentation a second time.

Change-Id: Ice9b66b1f262eca8758366ba4badbd44f0244248


mmeeks (approved):
Looks fine to me =) but of course needs to pass CI.
Thanks Marco.

PR #9658, opened 2024-07-29
Convert Control.Sidebar.js to Control.Sidebar.ts

NickWingate:
Change-Id: Iba2142d44f29ad0eb9293859f45c0718d3a223b4

* Resolves: # 
* Target version: master 

### Summary

```js
	setTabs: function(/*tabs*/) {
	},

	selectedTab: function() {
		// implement in child classes
	},
```
doesn't seem to have any purpose so I also removed in this patch since typescript-eslint doens't like empty functions

### TODO

- [ ] ...

### Checklist


gokaysatir (changes requested):
I think you also need to modify "Sidebar.js" in ".beforeprettier" file?

Other than that, this looks good, thanks for the improvements :)

NickWingate:
> I think you also need to modify "Sidebar.js" in ".beforeprettier" file?
Ah missed that, good catch. Thanks!

browser/.beforeprettier
```diff
@@ -44,7 +44,7 @@
 /src/control/Control.SearchBar.js
 /src/control/Control.Selection.js
 /src/control/Control.SheetsBar.js
-/src/control/Control.Sidebar.js
+/src/control/Control.Sidebar.ts
```
eszkadev, line 47: Please remove this, beforeprettier should be only used for old files.
We want to force new styling in the new files

eszkadev (changes requested):

browser/Makefile.am
```diff
@@ -395,7 +395,7 @@ COOL_JS_LST =\
 	src/control/Control.NotebookbarDraw.js \
 	src/control/Control.NotebookbarBuilder.js \
 	src/control/Control.Layers.js \
-	src/control/Control.Sidebar.js \
+	src/control/Control.Sidebar.ts \
```
eszkadev, line 398: There should be second list of files, please add that one too.
(It is  used for scanning for a translations)

browser/src/control/Control.Sidebar.ts
```diff
@@ -9,44 +9,59 @@
...
+ * Control.Sidebar
  */
 
-/* global $ app */
-L.Control.Sidebar = L.Control.extend({
+/* global app */
+interface SidebarOptions {
+	animSpeed: number;
+}
+class Sidebar extends L.Control {
```
eszkadev, line 19: no need for `extends L.Control` we want to completely drop Leaflet's dependency

eszkadev: it will be needed to also modify where we initialize the component then,
instead of `map.addControl(sidebar)`, we need to create sidebar object and then call `onAdd` method manually, or put it inside constructor.

eszkadev (approved):
looks good now, thanks :)

gokaysatir (approved):

Issue #9632, opened 2024-07-26
Dark mode does not set light text in read-only marked document


pedropintosilva:
Originally reported by @timur-g 

Dark mode was fixed for read-only shared document #7888 with https://github.com/CollaboraOnline/online/pull/7906. it may open Dark or Light, depending on COOL. But mode change does not work.

Mode change is still NOK for read-only marked and so opened document, sample from:

![image](https://github.com/user-attachments/assets/31c55687-40dd-40d5-85b2-6d4dbc658f85)

Darshan-upadhyay1110:
@timur-g for read-only mode it works fine can you please explain a bit more is there any specific thing you did, and then it was not working ?

[Screencast from 2024-07-26 18-04-33.webm](https://github.com/user-attachments/assets/3a21ceb8-1feb-4e60-9ffa-f681ab4448c0)

timur-g:
As noted in the OP, this is not about read-only from WOPI, it is about read-only marked let's say XLSX from MSO.

Darshan-upadhyay1110:
> As noted in the OP, this is not about read-only from WOPI, it is about read-only marked let's say XLSX from MSO.

I see I get it now . thanks @timur-g

PR #9660, opened 2024-07-29
Fix event binding issue by initializing event handlers once

Darshan-upadhyay1110:
- For 'marked read only' file somehow menubar item does not get initialized.

This patch:

- Introduced `isEventBound` function to check if an event with a specific type and namespace is already bound to an element.
- Simplified the `isEventBound` function to use the `Array.prototype.some` method for namespace checking.
- Created `bindEventIfNotBound` function to bind events only if they are not already bound.
- Updated event binding for `#main-menu` to use `bindEventIfNotBound`, preventing duplicate bindings for 'select.smapi', 'beforeshow.smapi', 'click.smapi', and 'keydown' events.

Change-Id: I26c95a7e366270b0020ce0fb27055f5acf9c1539

* Resolves: #9632 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1497,6 +1497,11 @@ L.Control.Menubar = L.Control.extend({
 	},
 
 	_onRefresh: function() {
+		if (!this._initialized) {
+			this._initialized = true;
+			this._onDocLayerInit();
+		}
+
 		// clear initial menu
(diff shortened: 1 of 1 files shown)


hcvcastro:
minor suggestion, use initialize function derived from L.Control class

hcvcastro (approved):

PR #9661, opened 2024-07-29
Revert "Move definitions and JS code into global.js from cool.html.m4"

plubius:
This reverts commit 8f0f931041b0b849a3e568862eb9de9641958906 which causes mobile apps to fail to open any documents and, instead, show a help page.


plubius:
Here's what the reverted commit causes when you open a document in the iOS app:

Screenshot 2024-07-29 at 4 02 43 PM

mmeeks:
Ah - that is indeed unfortunate; this improves our CSP headers in a helpful way however; of course that it breaks mobile sucks. @Minion3665 can you take a look ? I imagine it will bust Android too - and @gokaysatir will be back tomorrow to have a poke - hopefully we can get it looked at in the next days.

I wonder what we could do to get some more effective unit tests running on mobile though - it's odd that we get green for something fundamentally non-working; any thoughts? =)

Minion3665:
> @Minion3665 can you take a look ? I imagine it will bust Android too

Yep, it does

> I wonder what we could do to get some more effective unit tests running on mobile though - it's odd that we get green for something fundamentally non-working; any thoughts? =)

I think this is due to mobile unit tests only making sure
- That it runs in mobile mode on the browser (mobile cypress)
- That the mobile app builds (Android app co-24.04, iOS app co-24.04)

I think we should look into something that can automate presses and make assertions in a similar way to cypress for the real app in a simulator -- or, I guess, on hardware, but that's probably harder ... for iOS the simulator doesn't work, but we could presumably run it in "made for iPad" mode on our Apple-Silicon Mac ...

plubius:
> I think we should look into something that can automate presses and make assertions in a similar way to cypress for the real app in a simulator -- or, I guess, on hardware, but that's probably harder ... for iOS the simulator doesn't work, but we could presumably run it in "made for iPad" mode on our Apple-Silicon Mac ...

I usually build and run with the "made for iPad" on my Silicon Mac and it works quite well. A lot less painful than building and testing with a real device.

Also, I am happy to abandon this pull request if you are able to get Android working. I can then try getting iOS to work. I just posted a full revert to highlight the problem.

Minion3665:
> Also, I am happy to abandon this pull request if you are able to get Android working. I can then try getting iOS to work. I just posted a full revert to highlight the problem.

Android seems to (mostly) work after this pull request ... https://github.com/CollaboraOnline/online/pull/9669 ... there's still a little bit of weirdness in the about dialog but document loading/editing/etc. looks like it works from a quick test
image

mmeeks:
Thanks Skyler; so lets close this one for now =) Would be good to chase the help about dialog bits down too if you can.

PR #9664, opened 2024-07-30
fix: Cell header context menu

elliotfreebairn1:
Previously, when a row or column header context menu is openend, you could access the spreadsheet toolbar (switch sheets, scroll right etc..) and the context menu would stay open.

This was also an issue with the cell context menu aswell but skyler fixed it by changing the z-index of the its overlay. Howevever, the row and column header context menu's were left unchanged, so now the overlay for the context menu is above the spreadsheet toolbar.

Change-Id: I8dfe2304d86c15d1b343890797b9b2600ed61e31

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -491,7 +491,7 @@ export class Header extends app.definitions.canvasSectionObject {
 		$.contextMenu({
 			selector: '#canvas-container',
 			className: 'cool-font',
-			zIndex: 10,
+			zIndex: 1500,
 			items: this._menuItem,
 			callback: function() { return; }
 		});


Minion3665 (changes requested):
thanks very much!

The code looks good, and the commit message is *nearly* good - the one thing I would say is that it's a good idea to reference the change you're talking about, so for this case it would be nice to put "Id269e3fb9bc2251677a0792d83fc51bd56b9cb21" or something similar in your commit message ... that way people can easily find the context :)

Please may you do that, and then let's get this in!

Minion3665 (approved):
LGTM

PR #9665, opened 2024-07-30
CheckFileInfo Fortification

Ashod:
- wsd: better CheckFileInfo failure handling
- wsd: simplify DocBroker creation on async CFI
- wsd: detect invalidating http::Session in callbacks
- wsd: reset upload session on idle
- wsd: reset FetchHttpSession on idle
- browser: sorted packages
- wsd: use vector::data()


net/HttpRequest.hpp
```diff
@@ -688,6 +689,14 @@ class Request final
     /// and/or to interrupt transmission.
     int64_t readData(const char* p, int64_t len);
```
mmeeks, line 691: love more state dumping =)

net/HttpRequest.hpp
```diff
@@ -1366,7 +1412,15 @@ class Session final : public ProtocolHandlerInterface
                 LOG_TRC("onFinished calling client");
                 try
                 {
-                    _onFinished(std::static_pointer_cast<Session>(shared_from_this()));
+                    auto self = shared_from_this();
```
mmeeks, line 1415: Seems sensible; better outside the try ? =)

wsd/COOLWSD.cpp
```diff
@@ -4251,7 +4259,14 @@ void COOLWSD::processFetchUpdate()
         FetchHttpSession->setFinishedHandler([](const std::shared_ptr<http::Session>& httpSession) {
             std::shared_ptr<http::Response> httpResponse = httpSession->response();
 
-            FetchHttpSession.reset();
+            // Retire.
```
mmeeks, line 4262: Wow - really unhappy about this now I look. Why is this processFetchUpdate happening  in the 'mainWait' SocketPoll - and then using the WebServerPoll() to do its work ? that seems nonsensical.

Surely we should do the work in the mainWait poll in COOLWSD.cpp to avoid cross-thread horrors (and fouling up the webseverpoll with that ;)

wsd/RequestVettingStation.cpp
```diff
@@ -261,9 +261,13 @@ void RequestVettingStation::handleRequest(const std::string& id,
                         }
                     }
                     else if (_checkFileInfo == nullptr ||
-                             _checkFileInfo->state() == CheckFileInfo::State::None)
+                             _checkFileInfo->state() != CheckFileInfo::State::Pass)
```
mmeeks, line 264: Hmm - that's quite a big change; to double check the TimedOut state != Pass - and will now not go to the  // E.g. Timeout path (?)

wsd/RequestVettingStation.cpp
```diff
@@ -298,27 +302,16 @@ void RequestVettingStation::checkFileInfo(const Poco::URI& uri, bool isReadOnly,
...
-            if (_ws)
+
+            if (createDocBroker(docKey, url, uriPublic))
             {
-                if (createDocBroker(docKey, url, uriPublic))
+                assert(_docBroker && "Must have docBroker");
+                if (_ws)
                 {
-                    assert(_docBroker && "Must have docBroker");
+                    // If we don't have the WebSocket, deffer creating the client session.
```
mmeeks, line 311: defer ? =)

wsd/RequestVettingStation.cpp
```diff
@@ -298,27 +302,16 @@ void RequestVettingStation::checkFileInfo(const Poco::URI& uri, bool isReadOnly,
...
                     createClientSession(docKey, url, uriPublic, isReadOnly);
                 }
             }
```
mmeeks, line 314: Good to have a similar erro LOG_DBG as below here to describe the 'else' state here for !ws ?

wsd/wopi/WopiStorage.cpp
```diff
@@ -756,13 +756,18 @@ void WopiStorage::uploadLocalFileToStorageAsync(const Authorization& auth, LockC
...
-             asyncUploadCallback, profileZone = std::move(profileZone)](
-                const std::shared_ptr<http::Session>& httpSession)
+             asyncUploadCallback, profileZone = std::move(profileZone),
+             &socketPoll](const std::shared_ptr<http::Session>& httpSession)
         {
             profileZone->end();
 
             // Retire.
-            _uploadHttpSession.reset();
+            socketPoll.addCallback(
```
mmeeks, line 765: Really not a huge fan of this sort of thing. We should hold references properly on the things we own. I wonder if one of these trendy: "Don't need to copy stuff" plugins is getting rid of life-saving references up the stack somewhere ? =) but ...
Adding callbacks to de-reference stuff is liable to be far less reliable, and suck during shutdown / exception paths I think - where we most likely don't invoke the callbacks before cleanup.

Surely this is not necessary; I'd be ok with an assertion that we're in the right thread through ;-)

wsd/wopi/WopiStorage.cpp
```diff
@@ -803,11 +808,13 @@ void WopiStorage::uploadLocalFileToStorageAsync(const Authorization& auth, LockC
         LOG_ERR(wopiLog << " cannot upload file to WOPI storage uri [" << uriAnonym
                         << "]. Error: " << ex.displayText()
                         << (ex.nested() ? " (" + ex.nested()->displayText() + ')' : ""));
+        _uploadHttpSession.reset();
```
mmeeks, line 811: Fair enough etc. =)

mmeeks (changes requested):
Thanks Ash - needs a bit more work; some of this is obviously great to have in - and would go in green, others are more edgy and might want to be in red in a week+ - any sense in splitting them that way ?

wsd/COOLWSD.cpp
Ashod: Oh, I didn't realize we were in `mainWait`. Sorry, I'm not the original author of this feature.

I'm trying to avoid resetting what seems like the main reference to the Session from which the callback is done.
My concern is that if this is the last reference, we can segfault when we return from the callback and the local reference gets released _before_ the last member access. Something that the compiler controls and we can't force it to see this hard dependency.

Will fix the mismatch of polls. Thanks for catching it!

wsd/RequestVettingStation.cpp
Ashod: This function is called when cool.html is loaded and the WebSocket is created. By this point, `_checkFileInfo` was already working in the background getting CheckFileInfo.

If we had timed out, I want to try again. Hence the change to `!= Pass`. It intentionally includes the `TimedOut` state, so we issue a new request.

The old code gave up this opportunity to retry a timed-out request and rejected the attempt to load the document by returning an error and closing the socket.

The design is to do CheckFileInfo opportunistically on the GET of cool.html. But, unless we have a valid response, we always try again here, when the WS is created. This change shouldn't be risky.

I'll change it to explicitly retry if we haven't tried yet, or we timed out. All other cases go to the error handler.

mmeeks (approved):
Looks really good; and safer - I suspect we can get this in on green =)

mmeeks:
Clearly need to fix the iOS & Android builds first though =)

wsd/COOLWSD.cpp
mmeeks: In these cases - we should always hold references to 'this' all the way down the stack; and ideally in the SocketPoll / Socket infrastructure above too.

wsd/RequestVettingStation.cpp
mmeeks: Ah - makes sense =)

wsd/wopi/WopiStorage.cpp
mmeeks: Glad this got removed =)

wsd/COOLWSD.cpp
Ashod: Agreed. This is why I added the assertions on the `use_count()` of the `shared_ptr`, so we can catch when there is a violation. Not sure if there is a better way to enforce what you suggest otherwise.

PR #9666, opened 2024-07-30
ci: github action for nightly docker

juliushaertl:
Change-Id: I56d28f4b2428b248c52d550825d1ac337a536bdf

### Summary

Provides a separate docker build method to build a container isolated from the host system. The goal is to have a nightly docker image on the github container registry that can be used for easier testing and running automated tests in integrators.

For now only the built from the precompiled core assets is covered.

Note I've tested building in the github action in my local fork https://github.com/juliushaertl/online/actions/runs/10083478724 but not in here yet.

### Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,58 @@
+name: Build nightly docker image
+
+on:
+  schedule:
+    - cron: "30 0 * * *"
+
+jobs:
+  docker:
+    name: Build docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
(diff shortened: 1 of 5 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,59 @@
+name: Build nightly docker image
+
+on:
+  schedule:
+    - cron: "30 0 * * *"
+  push:
```
juliushaertl, line 6: To be removed before merge as then schedule will trigger only on the default github branch

juliushaertl:
@pedropintosilva Maybe you can request review if anyone else is also relevant for this

caolanm:
I wonder if we should have an --enable-experimental in here somewhere by default in the online configure line, or use ONLINE_EXTRA_BUILD_OPTIONS of --enable-experimental in the build.sh because I think we enable that for our normal code builds

juliushaertl:
Sure if it is closer to what code builds. I can add that.

PR #9669, opened 2024-07-30
fix(android): Allow document loading

Minion3665:
In I3155e799530d4c307569e88612a13570d0913e03, the mobile app was broken in 2 ways. This change fixes both of them. Here they are:

- First, certain elements were excluded on mobile. We presumably meant to check whether they were present before reading their value, but instead looked at whether they had a value ... this caused an error where we attempted to get a property on "undefined", crashing out of our initialization
- Second, we hardcoded the help information by catting `html/cool-help.html` into an input's value attribute in m4. Unfortunately, we neglected to add any form of escaping to speech marks, which caused it to flow out of its attribute... polluting the rest of the page with broken help information. I've used sed to replace speech marks with escaped HTML entities

Change-Id: I74d84afc3d333d4158f3be9c103f2c3131f26299

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -36,7 +36,7 @@ m4_ifelse(EMSCRIPTENAPP,[true],[m4_define([MOBILEAPP],[true])])
 m4_ifelse(MOBILEAPP, [true],
 [
   <input type="hidden" id="init-app-type" value="mobile" />
-  <input type="hidden" id="init-help-file" value="m4_syscmd([cat html/cool-help.html])" />
+  <input type="hidden" id="init-help-file" value="m4_syscmd([cat html/cool-help.html | sed 's/"/\&quot;/g'])" />
 ],
 [
   <input type="hidden" id="init-welcome-url" value="%WELCOME_URL%" />
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -255,9 +255,9 @@ class InitializerBase {
 		this.uriPrefix = document.getElementById('init-uri-prefix').value;
 		this.brandingUriPrefix = this.uriPrefix;
 
-		window.welcomeUrl = document.getElementById("init-welcome-url").value ? document.getElementById("init-welcome-url").value: "";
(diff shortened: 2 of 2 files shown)


mmeeks (approved):
Thanks Skyler! looks very sensible =)

plubius:
Hmm. I am seeing the following when I open a document. Seems like some sort of Online page:

Screenshot 2024-07-30 at 1 07 35 PM

plubius:
OK. Part of the problem was that the iOS Xcode file wasn't including global.js in the app so I fixed that in pull request #9673.

So now the iOS app gets a little farther and starts to populate the menu bar but never loads the document:

Screenshot 2024-07-30 at 7 40 52 PM

It looks like some sort of infinite JavaScript loop as I see the following debug message repeatedly in the Xcode console and the "y=" value in each message block growing increasing negative:

2024-07-30 20:01:39.706970 -0400 [ main ] DBG  To Online: clientvisiblearea x=-3450 y=-43560 width=19695 height=103515 splitx=0 splity=0| DocumentViewController.mm:305
2024-07-30 20:01:39.707221 -0400 [ lokit_runloop ] DBG  #20: child_ws: recv [child-001 clientvisiblearea x=-3450 y=-43560 width=19695 height=103515 splitx=0 splity=0 | KitWebSocket.cpp:58
2024-07-30 20:01:39.707495 -0400 [ main ] ERR  Error from WebView: jserror {
  "userAgent": "mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/605.1.15 (khtml, like gecko)",
  "vendor": "apple computer, inc.",
  "message": "Script error.",
  "source": "",
  "line": 0,
  "column": 0
}

plubius:
Forgot to add that the jserror after every clientvisiblearea message maps to the following error in the JavaScript console:

<pre>ResizeObserver loop completed with undelivered notifications.</pre>

PR #9671, opened 2024-07-30
clipboard: mobile: fixed js error

lpranam:
in mobile when user copied anything using context menu, user used to get unwanted message due to incorrectly triggered keydown event. message said "use on screen keyboard for copy/paste" which is not even possible in case of copying

Change-Id: I586021db1d1551d53993ba0791071ede8e742681

* Target version: master 

### Checklist


Rash419 (approved):
Looks good to me : )

PR #9673, opened 2024-07-31
Include global.js in iOS app

plubius:


Minion3665 (approved):
definitely an improvement, thanks Patrick!

Issue #7935, opened 2024-01-04
Spell correction annoyingly turned on for read-only users


RadiologieSLK:
### Describe the Bug
Hello,

we have an .xlsx spreadsheet on our nextcloud to be shown permanently on an always-on display as a daily overview. For this the sheet is shared with a user which has only read-rights. He cannot change the spell correction option, so by default all words (we use only abreviations) are underlined with red. Sometimes we have to restart the computer/browser meaning, even if I disable spell correction by login in with the admin-user (write-rights) and disable the spell correction, it will return once we restart the computer/browser.

Is there a way to disable spell correction permanently? I tried to deactivate it and saved the file, this is not enough.

Thank you!

See here:
https://forum.collaboraonline.com/t/disable-spell-correction-permanently/2275

### Expected Behavior
We would like to have the spell correction turned off for specific users (or all) by default. Having a button for read-only users to switch off the correction would be a start, but better would be a completely disabled spell check.

Thank you

c3systems:
Just to chime in on this one - I think everyone out there would appreciate a default behavior of spellcheck being off for any document opened in Read Only mode.  If you can't act on a spelling mistake, there's no reason to have it red-squigglied.  

Especially in documents with tons of technical terms, there's no hope to keep that dictionary up to date.  The read-only users have no ability to change anything, so, there's really no need to squiggle the heck out of their viewing experience.

- humble $0.02.  thanks for all you and the team do - this is an amazing platform.

PR #9674, opened 2024-07-31
Added method to disable spell check if document is read-only

rparth07:
Change-Id: I04eba16400825869522012eb16ba033200effeaa

* Resolves: #7935
* Target version: master 

### Summary
1. disableSpellCheckIfReadOnly Method:
    - Created a private method called disableSpellCheckIfReadOnly that updates the spellOnline property based on the isReadOnly status.
    - This method sets spellOnline to "false" if _isReadOnly is true, and keeps the current value otherwise.
    - Called disableSpellCheckIfReadOnly at the end of the parseDocOptions method to ensure spellOnline is set correctly based on the initial document options.

### Checklist

### DIFF
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -246,6 +246,17 @@ void Session::parseDocOptions(const StringVector& tokens, int& part, std::string
                 _docOptions += tokens.cat(' ', offset + 1);
         }
     }
+
+    // Disable spell check if the document is read-only
+    disableSpellCheckIfReadOnly();
+}
+
+void Session::disableSpellCheckIfReadOnly()
+{
+    if (_isReadOnly)
+    {
+        _spellOnline = "false";
+    }
 }
 
 void Session::disconnect()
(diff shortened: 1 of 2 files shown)


rparth07:
### Questions for Review

- **Approach:** Is the use of the `disableSpellCheckIfReadOnly` method appropriate for managing the `spellOnline` property based on `isReadOnly`? Are there any concerns with this approach?
- **Constructor Call:** Should we consider calling `disableSpellCheckIfReadOnly` in the constructor?
- **Handling in setReadOnly:** Should we set spellOnline within setReadOnly to ensure it reflects based upon current isReadOnly status whenever setReadOnly is called?
- **Additional Considerations:** Are there any additional improvements or considerations for managing `spellOnline` in relation to `isReadOnly`?

cc: @eszkadev
Your feedback and suggestions are greatly appreciated.

gokaysatir (approved):
I think it's fine to set it with the document options. Thanks for the improvements :)

PR #9675, opened 2024-07-31
Add missing trailing slash

pedropintosilva:
By looking at the top of the document "`<!DOCTYPE html>`" it seems we
don't need trailing slash for self-closing tags but maybe the parser
needs (?). This seems to be super safe and it would be nice to see if
it fixes the "fails to build a proper html tag, unclosed link tag" problem.

- For consistency (in other lines we use it)
- For XHTML-compatibility

Change-Id: I00bf34104a5b56d094196cb2b912e1dc308b6253

### DIFF
--- a/browser/src/slideshow/transition/CombTransition.ts
+++ b/browser/src/slideshow/transition/CombTransition.ts
@@ -1,3 +1,5 @@
+/** */
+
 /*
  * Copyright the Collabora Online contributors.
  *
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -61,7 +61,7 @@ m4_ifelse(BUNDLE,[],
   <!-- Using individual CSS files -->
   m4_foreachq([fileCSS],[COOL_CSS],[<link rel="stylesheet" href="][m4_ifelse(MOBILEAPP,[],[%SERVICE_ROOT%/browser/%VERSION%/])][fileCSS" />
 ]),
-[<link rel="stylesheet" href="][m4_ifelse(MOBILEAPP,[],[%SERVICE_ROOT%/browser/%VERSION%/][bundle.css"])])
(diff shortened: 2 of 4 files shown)


pedropintosilva:
attempt to fix 

```html
<link rel="stylesheet" href="/browser/747f5e21ff/bundle.css" <link="">
```

pedropintosilva:
now testing with `--enable-bundle` flag locally as in 
```
./configure CC="ccache gcc-12" CXX="ccache g++-12" --enable-silent-rules --with-lokit-path=/home/pedro/LibreOffice/core-24-x/include --with-lo-path=/home/pedro/LibreOffice/core-24-x/instdir --enable-cypress --enable-debug --enable-bundle --disable-ssl

```

pedropintosilva:
there is another parser problem that appear with or without this PR in:

```bash
Parse error at /home/pedro/LibreOffice/online-master/browser/release/bundle.js:131962,1
 * Copyright the Collabora Online contributors.

```

When I inspect the said fail I see missing forward slashes at the beginning :

Screenshot_20240731_103606

thebearon (approved):
The change from @pedropintosilva fixes the original issue for me, and the added changes fix the unrelated uglify JS issues (similar to those fixed in #9442).

browser/src/slideshow/transition/CombTransition.ts
```diff
@@ -1,3 +1,5 @@
+/** */
+
```
codewithvk, line 2: @pedropintosilva  I didn't understand why you added `/** */` here. Is there any reason? Please explain to me.

thebearon: @codewithvk I was the one who added them, based on c37bbf69c5eba8e15a348cea03e0133029e804b4, please see the commit message there.

codewithvk: Thanks @thebearon ! Nice work 🥇

PR #9676, opened 2024-07-31
private/marco/slideshow rework bugfix 2

mcecchetti:
- **slideshow rework: displayed slides has not the right aspect ratio**
- **slideshow rework: not possible to navigate when presenting in a window**

### DIFF
--- a/browser/src/slideshow/SlideCompositor.ts
+++ b/browser/src/slideshow/SlideCompositor.ts
@@ -19,21 +19,15 @@ declare var SlideShow: any;
 abstract class SlideCompositor {
 	_slideShowPresenter: SlideShowPresenter = null;
 	_presentationInfo: PresentationInfo = null;
-	_width: number = 0;
-	_height: number = 0;
 	_initialSlideNumber: number = 0;
 	_onGotSlideCallback: VoidFunction = null;
 
 	constructor(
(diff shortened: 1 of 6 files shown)


browser/src/slideshow/SlideCompositor.ts
```diff
@@ -55,13 +49,7 @@ abstract class SlideCompositor {
...
 	}
 
-	protected _getSlideWidth() {
-		return this._width;
-	}
-
-	protected _getSlideHeight() {
-		return this._height;
-	}
+	public abstract getCanvasSize(): Array<number>;
```
codewithvk, line 52: Suggestion: Instead of Array<number>, can we use [number, number] with a comment indicating which element is the width and which is the height? This will show a more strict type.

codewithvk (approved):
Code OK.

browser/src/slideshow/SlideCompositor.ts
quikee: Let's change that another time...

quikee (approved):
Looks great

quikee:
I think you need to use the prettier tool ... this is why the build fails

mcecchetti:
> I think you need to use the prettier tool ... this is why the build fails

yep, it seems so; thanks

PR #9678, opened 2024-07-31
fix(iOS): Allow document loading

Minion3665:
Commit 8f0f931 from #9442 broke both mobile apps in several ways...

In addition to the fixes which android needed (see #9669), there were 2 things broken in the iOS app only. Here they are:
- global.js wasn't included in the app resources (fixed by Patrick Luby in #9673)
- A garbled m4 macro caused some css to fail in loading, and -- my current understanding is -- as we use things like the page height in rendering this caused us to have severe issues when running our JavaScript. That's fixed in this commit


plubius (approved):
Tested the patch in my local master build (with LibreOffice build with distro/collabora/co-24.04) and documents now load normally.

Thanks Skyler!

PR #9683, opened 2024-08-01
wsd: update outdated doc wrt gdb and capabilities

vmiklos:
Mostly copied from
<https://github.com/CollaboraOnline/online/pull/9493#issuecomment-2237455110>.

Change-Id: Ia94a3b0f2d431c8b1d44a0953704c5de62bc1176

### DIFF
--- a/wsd/README.md
+++ b/wsd/README.md
@@ -216,9 +216,19 @@ or
 so that you have time to attach to the process.
 
 Then run coolwsd, and attach your debugger to the process you are
-interested in. Note that as the coolforkit executable file has
-capabilities set, so when debugging that you need to run the debugger
-with super-user privilege.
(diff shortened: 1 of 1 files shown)


vmiklos:
@caolanm could you please review? Thanks.

Copying the useful info from the PR comment to somewhere visible to 'git grep'.

caolanm (approved):

grandinj:
Nice!

Maybe we need a 'make debugrun' or 'make gdb' target that does this, and also uses pidof to get the PID (assuming that is possible in the namespaced world)

vmiklos:
Could you try if 'make run-gdb' does what you need? At least for me, the sysroot stuff is only needed when attaching to an already running process, plain gdb works out of the box for the 'gdb --args' case (since we don't require capabilities).

PR #9684, opened 2024-08-01
Widget: enable iconview with text

codewithvk:
Change-Id: I26aa8adc8bfdd8706fcb2dc28ad6c6af8d01196e

* Target version: master 

### Summary
- Previously, we only showed icons in cases where we wanted to display the IconView from the core. Now, if the IconView contains text, the text will be displayed below the icon from the IconView.

Screenshot from 2024-08-01 18-22-46

### TODO

- [ ] Need to check everyplace use IconView. Make sure IconView is not broken.

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1060,6 +1060,13 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	flex: 0 0 99%;
 }
 
+.icon-view-item-container {
+	max-width: 50px;
+	display: inline-block;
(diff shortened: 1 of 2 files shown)


eszkadev:
this should be chacked with: Fontwork dialog and Math sidebar

codewithvk:
> this should be chacked with: Fontwork dialog and Math sidebar

@eszkadev  This is the new PR for this Fix : https://github.com/CollaboraOnline/online/pull/9701. Would better to add comment there :)

PR #9686, opened 2024-08-01
continue with coolwsd-systemplate-setup if already mounted

caolanm:
like how clean-local already ignores failure to entirely remove the old systemplate.

This can happen if a non-namespace mount was left behind by a non-namespace ver of online. Then a namespace version lacks the privileges to remove the classic mounts.

Change-Id: I3aade8540b1cace405f13fa3b41a8e91b50ab06c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -488,8 +488,13 @@ endif
 # the systemplate directory, which we can't rm if it's mounted.
 $(SYSTEM_STAMP): ${top_srcdir}/coolwsd-systemplate-setup $(CLEANUP_DEPS)
 	$(CLEANUP_COMMAND)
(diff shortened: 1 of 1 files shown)


caolanm:
In clean-local we already carry on if we can't umount the classic file system mounts, so do the same here.

This one can arise if you switch to an older branch, build (where it will use sudo setcaps coolmount), run coolwsd, kill -9 it, or whack ctrl+c a few times to get the mounts left behind. Then switch to master, build and make run where coolmount won't have caps and can't remove the remnant system-level mounts. But it doesn't really matter, the template can still be updated, and the later namespace mounts themselves are autoremoved on process exit so they won't stack up to clutter the system mount table any further.

vmiklos (approved):
Continuing but printing a warning sounds like a reasonable compromise to me.

PR #9691, opened 2024-08-02
private/pedro/Backport PR 9561 n 9662

pedropintosilva:
- **JsDialogBuilder: add press and hold event for next and previous related commands**
- **UI: PushButtonControl: Fix alignment in mobile**

### DIFF
--- a/browser/css/spreadsheet.css
+++ b/browser/css/spreadsheet.css
@@ -17,7 +17,13 @@
 	height: 100%;
 	overflow-y: hidden;
 	overflow-x: auto;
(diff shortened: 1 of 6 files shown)


pedropintosilva:
I have tested both on computer and mobile. Didn't encounter any regression and it fixes the problems detailed in the commit messages. The commit `UI: PushButtonControl: Fix alignment in mobile` is the essential one (otherwise mobile dialogs are butchered) but the other commit needs to be cherry-picked to circumvent conflicts (plus it fixes https://github.com/CollaboraOnline/online/issues/8358 for 24.04 branch)

PR #9693, opened 2024-08-02
fix(jsdialog): Trigger change without keyup

Minion3665:
In 3bb07cdd528f37ec45af488ec6e58608cd041915, we changed the 'change' event for multiline text inputs to use 'keyup' instead. Unfortunately, that change forgets about ways in which the input field can change text without there being a 'keyup' event.

Of particular relevance to me is on iOS, where pasting into one a field won't trigger a 'keyup' event, leading to, e.g., QR codes to fail generation with pasted URLs

This patch is a bit unfortunate in that it makes things which *do* trigger 'keyup' events (which is most text edits) send double events for these text fields... I think there's some duplicate-event filtering that could be done, but I think it's better to get this in before worrying about it.

Change-Id: I188c13cfe7f5dbe2600551f4023d1c9d98bbbba2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/jsdialog/Widget.MultilineEdit.js
+++ b/browser/src/control/jsdialog/Widget.MultilineEdit.js
@@ -56,13 +56,16 @@ function _multiLineEditControl(parentContainer, data, builder, callback) {
 		edit.disabled = true;
 	}
 
-	edit.addEventListener('keyup', function() {
+	function _keyupChangeHandler() {
 		if (callback)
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
works as expected

eszkadev:
I noticed other issue:  when I open qr dialog and just press ok - it crashes

Issue #9694, opened 2024-08-02
Android: document loading progress bar doesn't move


Minion3665:
On Android the document loading progress bar no longer moves along as the document loads. Instead, it stays at 0% until the document loads, when it disappears. It would be nice to get this feedback of the document loading back...

---

### To reproduce

- Open any document on Android
- Notice that the progress bar stays completely empty while the document is loading

### Expected

- The progress bar moves along as the document loads

### Video

https://github.com/user-attachments/assets/de324ea1-474f-4614-a481-66ffab0cf81d

---

It seems like a regression coming from https://github.com/CollaboraOnline/online/pull/8957

PR #9695, opened 2024-08-02
fix(android): Restore progress bar

Minion3665:
On Android the document loading progress bar had regressed so it no
longer moved along as the document loaded. Instead, it stayed at 0%
until the document loaded, when it disappeared without having moved at
all...

...this was a known regression from
https://github.com/CollaboraOnline/online/pull/8957 -- known as in
"known before we merged it". We decided background save was important
enough that breaking the -- poor as discussed in the next paragraphs --
progress bar was worth merging it, and we would fix the regression
later. This is that fix.

The fix here still freezes for a surprisingly long time on "empty" -
enough that I wonder if there's something freezing the UI thread during
load - however it is at a similar level to before the regression and is
a definite improvement over the current state...

...I will continue looking to make the progress bar "look alive" earlier
during document load...

Fixes: https://github.com/CollaboraOnline/online/issues/9694
Change-Id: Idc30105cdd2f650bc5d39deae3437aa96bfe32d1

### DIFF
--- a/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
+++ b/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
@@ -80,6 +80,8 @@ import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
+import org.json.JSONException;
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks very sensible- thanks Skyler ! =)

Can't possibly break wopi-save-as either - so lets merge.

PR #9696, opened 2024-08-02
apparmor profile for coolwsd

caolanm:
See: https://wiki.debian.org/AppArmor/Contribute/FirstTimeProfileImport

Change-Id: I2451cfc222b6745307554fde37870717204e8a99

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/debian/apparmor-profile
+++ b/debian/apparmor-profile
@@ -1,0 +1,13 @@
+# vim:syntax=apparmor
+# AppArmor policy for coolwsd
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+profile coolwsd /usr/bin/coolwsd flags=(unconfined) {
+  userns,
+
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/coolwsd>
+}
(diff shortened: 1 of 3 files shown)


caolanm:
In a Ubuntu 24.04 virtualbox install manually installing these rules was sufficient to get coolwsd with namespaces allowed to get past https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2046844

The actual generation of .deb is untested however

mmeeks (approved):
Looks great to me =) thanks Caolan.

mmeeks:
I wonder if we can add some helpful docs to the configure.ac to get people to detect Ubuntu and/or this problem and get them to set that up ? =)

caolanm:
The catch with the apparmor profile is that it's tied to the install location, so the above should be fine for the install from .deb but leave the dev self-build make run case as a problem.

It's probably possible to use setfattr to tag a self-build coolwsd to use some existing profile that is allowed to do usernamespaces, but that needs root, so it's back to using sudo then.

But I see that Ubuntu podman (unsurprisingly) has apparmor permission to create these namespaces. So what works for me is to leverage Ubuntu podman with "podman unshare ./coolwsd" to launch a self-build coolwsd in an env where it is allowed to use namespaces by apparmor with a default Ubuntu 24.04 install in VirtualBox.

https://github.com/CollaboraOnline/online/pull/9716 for that idea

Issue #9538, opened 2024-07-18
Save "Invert Background" preference in JS storage


Rash419:
### Describe the Bug
Save "Invert Background" state in js storage

### Steps to Reproduce
1. `View` -> `Dark Mode`
2. You will see `Invert Background` button
2. Click you will see document background becomes light
3. Reload the document it will reset to dark

### Expected Behavior
It should save the  `Invert Background` preference on client side

### Actual Behavior
It doesn't save the `Invert Background`, resets on document reload

Minion3665:
This was mostly fixed in https://github.com/CollaboraOnline/online/pull/9652, but I have an additional pull request to fix it for calc here: https://github.com/CollaboraOnline/online/pull/9697

PR #9697, opened 2024-08-02
fix(calc): Restore inverted document on load

Minion3665:
Previously, we didn't have a way to restore the document background color on load for Calc, leading to a document that was inverted according to our LocalStorage and Core, but which didn't have the background to show it. This led to text rendering issues, as well as needing multiple presses to switch into invert background mode properly...

...in this change, I've simplified the calc background code to rely on our existing css data-bg-theme property, and made it consistently set the background color to --color-background-document ... it's possible this causes problems with changing themes alongside LOK_CALLBACK_DOCUMENT_BACKGROUND_COLOR, but I'm confident this is an improvement over the previous behavior as that should also have been true of it...

...this brings Calc invert backgrounds to parity with the other apps, in terms of saving/loading them properly...

Change-Id: I45b2fe174f1e7d0ed8fb6bc582a5932c86d4d3d1
Fixes: https://github.com/CollaboraOnline/online/issues/9538

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -12,8 +12,6 @@
 	--color-text-lighter: #fff; /* secondard text, disabled */
 	--color-status-badge: #d6d6d6;
 
-	--color-canvas-dark: #141414;
-	--color-background-document: #121212;
 	--color-main-background: #121212;
 	--color-background-dark: #1E1E1E;  /* select */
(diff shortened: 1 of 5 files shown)


caolanm (approved):
This seems reasonable to me

PR #9701, opened 2024-08-03
Widget: enable iconview with text

codewithvk:
Change-Id: I44e3e40b3be1dcfd0d099939e6f6a9a4e574fdbd

* Resolves: # 
* Target version: master 

### Summary
- Previously, we only showed icons in cases where we wanted to display the IconView from the core. Now, if the IconView contains text, the text will be displayed below the icon from the IconView.

Screenshot from 2024-08-01 18-22-46

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1044,6 +1044,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	border: 1px solid transparent;
 	border-radius: var(--border-radius);
 	padding-top: 5px;
+	text-wrap: nowrap;
+	font-family: var(--jquery-ui-font);
 }
 
 .ui-iconview-entry.selected {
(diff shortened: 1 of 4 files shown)


codewithvk:
This PR is failing. We are using style preview with a text field from the core side, which causes the Iconview to be displayed with text here.    

Screenshot from 2024-08-03 23-17-02

Code pointer: stylespreview.ui

I think we are using IconView with text in sidebargallery.ui, Not sure where we have this in UI. We have to fix it as well.

codewithvk:
> stylespreview.ui

Fix from core side : https://gerrit.libreoffice.org/c/core/+/171491

codewithvk:
Fix for Failing CI: https://gerrit.libreoffice.org/c/core/+/171491. @quikee We need to merge so that the CI green here. :)

eszkadev (approved):

eszkadev:
Added some cleanup and fix for a bug we detected recently

Issue #9621, opened 2024-07-25
perf: connection re-use ...


mmeeks:
When we receive incoming connections - we serve a single request - and then shut them down. This can waste lots of latency in terms of socket setup and shutdown - particularly with SSL; this is often hidden by a proxy of course, but it would be nice to not have to handshake again to serve javascript & UI images.

So - we should fix this - at least for the FileServer use-case.

git grep -20 handleFaviconRequest

Will show you code pointers for how this happens; of course:

    HttpHelper::sendFileAndShutdown(socket, faviconPath, response);

Is the culprit that we should probably keep; if we get an error or a malformed request on a socket we should shut it down - to hand the latency back to the bad actor I think ;-)

But I guess we need another method that will not add the Connection: Close, and ensure the socket ends up back in a good state for handling the next request - hopefully not horribly hard.

mmeeks:
As a bonus - we should really remove the Poco message parsing code:

    Poco::MemoryInputStream startmessage(&socket->getInBuffer()[0], socket->getInBuffer().size());

and instead remove our Poco dependency; we should re-use @Ashod 's net/HttpRequest.hpp classes I expect to store the headers. There is a http::Request::readData() that should do lots of this for us I think.

Unit tests required of course. Long term we want to loose the Poco dependency of course.

Sven - would be a good one for you too I think =)

sgothel:
Besides review and discussion, the replacement of Poco http request message parsing
starting with Ash's net/HttpRequest .. is also on the TODO.

I started with wrong unit test derived from UnitWOPIFileUrl,
this patch uses UnitSession and covers document + favicon download 
while keeping the connection/socket alive.

mmeeks:
And - of course; it turns out that this is a nice-to-have; but the vast majority of our file server doesn't go through this method I think =)

See:

 void FileServerRequestHandler::handleRequest(const HTTPRequest& request,
                                             const RequestDetails& requestDetails,
                                             Poco::MemoryInputStream& message,
                                             const std::shared_ptr<StreamSocket>& socket,
                                             ResourceAccessDetails& accessDetails)

Which is where the real file-serving code-path is for most of what we send - prolly should focus on that as well =)

PR #9703, opened 2024-08-05
Async Locking Prerequisites

Ashod:
**Possibly for RED.**

Some Kill-Poco and refactored locking to prepare for async locking.

- wsd: updateStorageLockState -> lockDocumentInStorage
- wsd: use an enum for lock states
- wsd: use LockState enum in LockContext
- wsd: encapsulate supportsLocks in LockContext
- wsd: refactor AsyncUpload for reuse
- wsd: use STATE_ENUM
- killpoco: document locking without poco
- killpoco: remove unused http helpers
- wsd: move and update StorageConnectionManager::initialize()
- wsd: use StorageConnectionManager::getHttpSession()

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -142,6 +142,7 @@ using Poco::Net::PartHandler;
 #  include <SslSocket.hpp>
 #endif
 #include "Storage.hpp"
+#include <wsd/wopi/StorageConnectionManager.hpp>
 #include "TraceFile.hpp"
 #include <Unit.hpp>
 #include <Util.hpp>
@@ -1157,7 +1158,7 @@ public:
                 try
                 {
                     std::shared_ptr<http::Session> httpSession(
-                            StorageBase::getHttpSession(remoteServerURI));
+                            StorageConnectionManager::getHttpSession(remoteServerURI));
                     http::Request request(remoteServerURI.getPathAndQuery());
 
                     //we use ETag header to check whether JSON is modified or not
@@ -1805,7 +1806,7 @@ private:
     bool downloadPlain(const std::string& uri)
     {
         const Poco::URI fontUri{uri};
-        std::shared_ptr<http::Session> httpSession(StorageBase::getHttpSession(fontUri));
+        std::shared_ptr<http::Session> httpSession(StorageConnectionManager::getHttpSession(fontUri));
         http::Request request(fontUri.getPathAndQuery());
 
         request.set("User-Agent", http::getAgentString());
@@ -1819,7 +1820,7 @@ private:
     bool eTagUnchanged(const std::string& uri, const std::string& oldETag)
     {
         const Poco::URI fontUri{uri};
-        std::shared_ptr<http::Session> httpSession(StorageBase::getHttpSession(fontUri));
+        std::shared_ptr<http::Session> httpSession(StorageConnectionManager::getHttpSession(fontUri));
         http::Request request(fontUri.getPathAndQuery());
 
         if (!oldETag.empty())
@@ -1844,7 +1845,7 @@ private:
     bool downloadWithETag(const std::string& uri, const std::string& oldETag)
     {
         const Poco::URI fontUri{uri};
-        std::shared_ptr<http::Session> httpSession(StorageBase::getHttpSession(fontUri));
+        std::shared_ptr<http::Session> httpSession(StorageConnectionManager::getHttpSession(fontUri));
         http::Request request(fontUri.getPathAndQuery());
 
         if (!oldETag.empty())
@@ -2939,6 +2940,9 @@ void COOLWSD::innerInitialize(Application& self)
 
 #if !MOBILEAPP
     net::AsyncDNS::startAsyncDNS();
+
+    LOG_TRC("Initialize StorageConnectionManager");
+    StorageConnectionManager::initialize();
 #endif
 
     PrisonerPoll = std::make_unique<PrisonPoll>();
@@ -4255,7 +4259,7 @@ void COOLWSD::processFetchUpdate(SocketPoll& poll)
         uriFetch.addQueryParameter("product", config::getString("product_name", APP_NAME));
         uriFetch.addQueryParameter("version", COOLWSD_VERSION);
         LOG_TRC("Infobar update request from " << uriFetch.toString());
-        FetchHttpSession = StorageBase::getHttpSession(uriFetch);
+        FetchHttpSession = StorageConnectionManager::getHttpSession(uriFetch);
         if (!FetchHttpSession)
             return;
 
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1112,7 +1112,7 @@ void DocumentBroker::lockIfEditing(const std::shared_ptr<ClientSession>& session
         {
             LOG_DBG("Locking docKey [" << _docKey << "], which is editable");
             std::string error;
-            if (!updateStorageLockState(*session, /*lock=*/true, error))
+            if (!updateStorageLockState(*session, StorageBase::LockState::LOCK, error))
             {
                 LOG_ERR("Failed to lock docKey [" << _docKey << "] with session ["
                                                   << session->getId()
@@ -1538,8 +1538,8 @@ bool DocumentBroker::lockDocumentInStorage(const Authorization& auth, std::strin
     assert(_lockCtx->_supportsLocks && "Expected to have lock support");
     assert(!_lockCtx->_isLocked && "Expected not to have locked already");
 
-    const StorageBase::LockUpdateResult result =
-        _storage->updateLockState(auth, *_lockCtx, /*lock=*/true, _currentStorageAttrs);
+    const StorageBase::LockUpdateResult result = _storage->updateLockState(
+        auth, *_lockCtx, StorageBase::LockState::LOCK, _currentStorageAttrs);
     error = _lockCtx->_lockFailureReason;
 
     switch (result)
@@ -1563,7 +1563,8 @@ bool DocumentBroker::lockDocumentInStorage(const Authorization& auth, std::strin
     return false;
 }
 
-bool DocumentBroker::updateStorageLockState(ClientSession& session, bool lock, std::string& error)
+bool DocumentBroker::updateStorageLockState(ClientSession& session, StorageBase::LockState lock,
+                                            std::string& error)
 {
     if (session.getAuthorization().isExpired())
     {
@@ -1571,7 +1572,7 @@ bool DocumentBroker::updateStorageLockState(ClientSession& session, bool lock, s
         return false;
     }
 
-    if (lock && session.isReadOnly())
+    if (lock == StorageBase::LockState::LOCK && session.isReadOnly())
     {
         // Readonly sessions cannot lock, only editors can.
         error = "Readonly session";
@@ -1589,28 +1590,30 @@ bool DocumentBroker::updateStorageLockState(ClientSession& session, bool lock, s
             return true; // Not an error.
             break;
         case StorageBase::LockUpdateResult::OK:
-            LOG_DBG((lock ? "Locked" : "Unlocked") << " docKey [" << _docKey << "] successfully");
+            LOG_DBG((lock == StorageBase::LockState::LOCK ? "Locked" : "Unlocked")
+                    << " docKey [" << _docKey << "] successfully");
             return true;
             break;
         case StorageBase::LockUpdateResult::UNAUTHORIZED:
-            LOG_ERR("Failed to " << (lock ? "Locked" : "Unlocked") << " docKey [" << _docKey
+            LOG_ERR("Failed to " << (lock == StorageBase::LockState::LOCK ? "Locked" : "Unlocked")
+                                 << " docKey [" << _docKey
                                  << "]. Invalid or expired access token. Notifying client and "
                                     "invalidating the authorization token of session ["
                                  << session.getId() << "]. This session will now be read-only");
             session.invalidateAuthorizationToken();
-            if (lock)
+            if (lock == StorageBase::LockState::LOCK)
             {
                 // If we can't unlock, we don't want to set the document to read-only mode.
                 session.setLockFailed(error);
             }
             break;
         case StorageBase::LockUpdateResult::FAILED:
-            LOG_ERR("Failed to " << (lock ? "Locked" : "Unlocked") << " docKey [" << _docKey
-                                 << "] with reason [" << error
+            LOG_ERR("Failed to " << (lock == StorageBase::LockState::LOCK ? "Locked" : "Unlocked")
+                                 << " docKey [" << _docKey << "] with reason [" << error
                                  << "]. Notifying client and making session [" << session.getId()
                                  << "] read-only");
 
-            if (lock)
+            if (lock == StorageBase::LockState::LOCK)
             {
                 // If we can't unlock, we don't want to set the document to read-only mode.
                 session.setLockFailed(error);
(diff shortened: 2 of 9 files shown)


mmeeks:
@sgothel I forget where we're at here. @Ashod are you working on this - IIRC there is/was a github task assigned for something similar here to Sven; would be good to de-conflict that if so =)

Meanwhile - @sgothel might be good to review Ash's patch if you have time =)

sgothel:
Great, this review shall help me with #9621 (as http retooling is also involved) .. reading.

Ashod:
> @Ashod are you working on this - IIRC there is/was a github task assigned for something similar here to Sven; would be good to de-conflict that if so =)

For async locking? I wasn't aware. I have more patches to come.

wsd/DocumentBroker.cpp
```diff
@@ -1145,7 +1145,7 @@ void DocumentBroker::lockIfEditing(const std::shared_ptr<ClientSession>& session
         {
             LOG_DBG("Locking docKey [" << _docKey << "], which is editable");
             std::string error;
-            if (!updateStorageLockState(Authorization::create(uriPublic), error))
+            if (!lockDocumentInStorage(Authorization::create(uriPublic), error))
             {
```
sgothel, line 1149: here you seem to change semantics? not update the lock state but lock it (initially?)
Intended? @Ashod (potential issue or a bug fix of yours?)

Ah, I see - you changed the name (below): `updateStorageLockState` -> `lockDocumentInStorage`. 
OK, assuming it matches semantics better.

wsd/Storage.cpp
```diff
@@ -464,70 +380,6 @@ void LocalStorage::uploadLocalFileToStorageAsync(const Authorization& /*auth*/,
...
-    auto httpSession = http::Session::create(uri.getHost(), protocol, uri.getPort());
-
-    static int timeoutSec = COOLWSD::getConfigValue<int>("net.connection_timeout_secs", 30);
-    httpSession->setTimeout(std::chrono::seconds(timeoutSec));
-
-    return httpSession;
-}
-
-#endif // !MOBILEAPP
-
```
sgothel, line 530: .. all replaced by your own work in http namespace, great.

wsd/Storage.hpp
```diff
@@ -239,16 +192,14 @@ class StorageBase
...
-            FAILED
-        };
+        STATE_ENUM(Result,
+                   OK = 0, //< Uploaded successfully
+                   DISKFULL, //< Unused.
+                   TOO_LARGE, //< 413
+                   UNAUTHORIZED, //< 401, 403, 404
+                   DOC_CHANGED, /**< Document changed in storage */
+                   CONFLICT, //< 409
+                   FAILED);
```
sgothel, line 202: STATE_ENUM produces an int8_t enum class, OK I guess. Would use uint8_t at least though (actually uint32_t anyways due to min-arch probably is 32-bit).

STATE_ENUM also takes care for a 'const char*' conversion, nice. We use a static local immutable string array - nice + the mandatory toString(..). Nice.

wsd/Storage.hpp
```diff
@@ -286,19 +237,19 @@ class StorageBase
...
     public:
-        enum class State
-        {
+        STATE_ENUM(
+            State,
             None, //< No async upload in progress or isn't supported.
             Running, //< An async upload request is in progress.
             Error, //< Failed to make an async upload request or timed out, no UploadResult.
             Complete //< The last async upload request completed (regardless of the server's response).
-        };
```
sgothel, line 299: I also prefer CamelCase for these constants (above all capital though)

wsd/Storage.hpp
```diff
@@ -308,13 +259,16 @@ class StorageBase
...
 
         /// Returns the result of the async upload.
-        const UploadResult& result() const { return _result; }
+        const TResult& result() const { return _result; }
 
     private:
         State _state;
-        UploadResult _result;
+        TResult _result;
     };
```
sgothel, line 267: so we prefer having flexible TResult types, nice

wsd/Storage.hpp
```diff
@@ -561,4 +513,47 @@ class LocalStorage : public StorageBase
     static std::atomic<unsigned> LastLocalStorageId;
 };
 
+/// Represents whether the underlying file is locked
+/// and with what token.
+struct LockContext
```
sgothel, line 518: But does it also actually lock the file/context?
Also what locking policy is being used? 
Like: One context per one thread at any given time? Or globally?

wsd/Storage.hpp
```diff
@@ -561,4 +513,47 @@ class LocalStorage : public StorageBase
...
+
+    /// do we need to refresh our lock ?
+    bool needsRefresh(const std::chrono::steady_clock::time_point& now) const;
+
+    void dumpState(std::ostream& os) const;
+
+private:
+    const std::chrono::seconds _refreshSeconds;
+};
+
```
sgothel, line 558: I miss synchronization here as locking is usually implying multiple threads concurrently acting. If so, either we would need atomics (best for non-blocking locks) or at least thread locks (perhaps with blocking and waiting condition_variable etc)? @Ashod

wsd/wopi/StorageConnectionManager.cpp
```diff
@@ -11,6 +11,10 @@
 
 #include <config.h>
 
+#if MOBILEAPP
+#error "Mobile doesn't need or support WOPI"
+#endif
+
```
sgothel, line 17: me still have to read about this protocol (spec from MS available)

wsd/wopi/StorageConnectionManager.cpp
```diff
@@ -150,3 +149,89 @@ StorageConnectionManager::getHttpSession(const Poco::URI& uri, std::chrono::seco
...
+    // Init client
+    Poco::Net::Context::Params sslClientParams;
+
+    // false default for upgrade to preserve legacy configuration
+    // in-config-file defaults are true.
+    SSLAsScheme = COOLWSD::getConfigValue<bool>("storage.ssl.as_scheme", false);
+
+    // Fallback to ssl.enable if not set - for back compatibility & simplicity.
+    SSLEnabled = COOLWSD::getConfigValue<bool>("storage.ssl.enable",
+                                               COOLWSD::getConfigValue<bool>("ssl.enable", true));
```
sgothel, line 169: Here we still use Poco? I used Botan (C++) once or maybe OpenSSL?

wsd/wopi/StorageConnectionManager.hpp
```diff
@@ -49,104 +49,11 @@ class StorageConnectionManager final
...
-            LOG_ERR("Failed to initialize Client SSL.");
-        else
-            LOG_INF("Initialized Client SSL.");
-#endif
-#else
-        FilesystemEnabled = true;
-#endif
-    }
-#endif
-
```
sgothel, line 149: @Ashod .. so you moved this up to the C++ implementation? This removed code was excluded anyways .. so some 'hidden' initialize() implementation used before?

wsd/wopi/StorageConnectionManager.hpp
```diff
@@ -171,7 +78,6 @@ class StorageConnectionManager final
     /// Saves new URI when resource was moved
     // void setUri(const Poco::URI& uri) { _uri = sanitizeUri(uri); }
 
-    static bool FilesystemEnabled;
```
sgothel, line 174: I assume this is a system property alike flag? Potentially immutable after setup?
Then we could use a singleton (via a local static instance) of a property class object,
which also takes care of the mutability etc? @Ashod (just brainstorming)

sgothel (approved):
Consider my questions only as newbie-to-the-codebase discussions.

Please judge yourself whether my interpretation is correct and act accordingly.
I welcome a discussion 'offline' to help my understanding of the codebase.
Thank you.

wsd/wopi/WopiStorage.cpp
```diff
@@ -342,70 +342,61 @@ StorageBase::LockUpdateResult WopiStorage::updateLockState(const Authorization&
 
     try
     {
-        std::unique_ptr<Poco::Net::HTTPClientSession> psession(getHTTPClientSession(uriObject));
+        std::shared_ptr<http::Session> httpSession = getHttpSession(uriObject);
```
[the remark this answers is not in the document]
Ashod: `http::Session` isn't intended to be "shared" in the sense of multi-threading, rather we use `shared_ptr` because the reference to the instance is passed in the callback, when async requests are used. So `unique_ptr` wouldn't do.

wsd/wopi/WopiStorage.cpp
```diff
@@ -342,70 +342,61 @@ StorageBase::LockUpdateResult WopiStorage::updateLockState(const Authorization&
...
-        Poco::Net::HTTPResponse response;
-        std::istream& rs = psession->receiveResponse(response);
+        httpHeader.setContentLength(0);
 
-        std::ostringstream oss;
-        Poco::StreamCopier::copyStream(rs, oss);
-        std::string responseString = oss.str();
+        const std::shared_ptr<const http::Response> httpResponse =
+            httpSession->syncRequest(httpRequest);
+        const std::string responseString = httpResponse->getBody();
```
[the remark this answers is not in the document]
Ashod: The reason for this separation is to make it easier to replace Poco. There are few things in the API that I'd also like to change. But being able to do a drop-in replacement with easier reviews is a good win.

Having said that, the only things we care from the URI when creating a session are the host and port. The request doesn't need to know either of these, so there is no hard-dependency between session and request. (F.e. two different sessions to two different hosts both can re-use the same request for "index.html".) So session and request are fairly independent objects.

wsd/wopi/WopiStorage.cpp
```diff
@@ -342,70 +342,61 @@ StorageBase::LockUpdateResult WopiStorage::updateLockState(const Authorization&
...
+            lockCtx._lockFailureReason = failureReason;
+            failureReason = ", failure reason: \"" + failureReason + "\"";
+        }
 
-                return LockUpdateResult::UNAUTHORIZED;
-            }
+        if (httpResponse->statusLine().statusCode() == http::StatusCode::Unauthorized ||
+            httpResponse->statusLine().statusCode() == http::StatusCode::Forbidden ||
+            httpResponse->statusLine().statusCode() == http::StatusCode::NotFound)
+        {
```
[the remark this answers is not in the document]
Ashod: There aren't many cases with multiple cases treated identically, this being one. And while readability is important, performance is too. This doesn't seem too bad to me, as it's transparent.

wsd/wopi/WopiStorage.cpp
```diff
@@ -342,70 +342,61 @@ StorageBase::LockUpdateResult WopiStorage::updateLockState(const Authorization&
...
-        LOG_ERR("Cannot " << wopiLog << " uri [" << uriAnonym << "]. Error: " << pexc.displayText()
-                          << (pexc.nested() ? " (" + pexc.nested()->displayText() + ')' : ""));
+
+        LOG_ERR("Un-successful " << wopiLog << " with HTTP status "
+                                 << httpResponse->statusLine().statusCode() << failureReason
+                                 << " and response: " << responseString);
+        return LockUpdateResult::FAILED;
     }
     catch (const BadRequestException& exc)
     {
```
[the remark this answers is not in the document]
Ashod: That's correct. But we have different use-cases. We tend to prefer status-codes, as they are more linear and clearer (among other advantages). But, in some cases we need to stop/abort a request and for those we use exceptions. A good example is authorization failure, or hitting a hard-limit (out of disk, etc).

Finally, some libraries (like Poco) use exceptions. Many try/catch clauses exist to protect against those exceptions, lest they leak and break the world.

wsd/COOLWSD.cpp
```diff
@@ -4255,7 +4259,7 @@ void COOLWSD::processFetchUpdate(SocketPoll& poll)
         uriFetch.addQueryParameter("product", config::getString("product_name", APP_NAME));
         uriFetch.addQueryParameter("version", COOLWSD_VERSION);
         LOG_TRC("Infobar update request from " << uriFetch.toString());
-        FetchHttpSession = StorageBase::getHttpSession(uriFetch);
+        FetchHttpSession = StorageConnectionManager::getHttpSession(uriFetch);
         if (!FetchHttpSession)
             return;
```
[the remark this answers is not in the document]
Ashod: Yes, `StorageBase` has too much functionality and the connection stuff is moved into the `StorageConnectionManager`. This is pure non-functional refactoring.

wsd/DocumentBroker.cpp
Ashod: Good that you're checking/catching subtle things like this. I like your approach.

`updateStorageLockState` is (was) overloaded. The one with 2 parameters only locked, and was called only at the start (without a client session). This is just renaming that overload, since it was really locking and that is a more accurate name (also, overloading in this case seems unnecessary source of confusion).

wsd/DocumentBroker.cpp
```diff
@@ -1553,27 +1553,26 @@ bool DocumentBroker::updateStorageLockState(const Authorization& auth, std::stri
...
+                                            std::string& error)
 {
     if (session.getAuthorization().isExpired())
     {
         error = "Expired authorization token";
         return false;
     }
 
-    if (lock && session.isReadOnly())
+    if (lock == StorageBase::LockState::LOCK && session.isReadOnly())
```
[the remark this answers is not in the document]
Ashod: `lock` could be `const`, yes (perhaps it should). But I'm not a fan of reverse-comparisons. It's far more natural to ask "is X zero?", "is Y empty?", "is Z equal to 5?" rather than "is 0 X?", "is empty Y?", and "is 5 equal to Z?"

I know some prefer the reverse approach, because it protects against accidentally assigning instead of comparing (i.e. missing a second '='), but modern compilers warn hard on this, and past the first few years of experience most people learn not to do that mistake anyway. But the cost to the reader (and we read far more than we write) to read the const first, think "I'm comparing 0 to something..." then see the variable think about its type and meaning, only to remember what constant we were comparing it with, well, it's overkill. (Also, in many cases we compare variables to variables. What to do in those cases?)

wsd/Storage.hpp
Ashod: We have a few all-caps constants. That's probably the influence of http status codes, and the fact that these codes were defined in Poco and we ported them over.

wsd/Storage.hpp
Ashod: This is to lock a document in storage. The policy is exclusive access while the document is open for editing.

wsd/Storage.hpp
Ashod: As stated above, this has nothing to do with threads. It's part of an external storage protocol to lock the document until the last editor disconnects.

wsd/wopi/StorageConnectionManager.cpp
Ashod: Our sockets use OpenSSL. Where we use Poco (not many cases left), Poco's sockets do their own SSL (also using OpenSSL). This is the config for whether HTTPS is available for the storage communications.

wsd/wopi/StorageConnectionManager.hpp
Ashod: Yes, now it's enabled and replaced by the code moved (which was more recent).

wsd/wopi/StorageConnectionManager.hpp
Ashod: This is only meaningful in debug builds, for debugging.

Ashod:
Thanks, @sgothel, for the thorough review and good comments/suggestions. I tried to answer with as much details as possible. Hope my responses make sense and clarify things. If not, we'll catch up offline to discuss.

sgothel:
> Thanks, @sgothel, for the thorough review and good comments/suggestions. I tried to answer with as much details as possible. Hope my responses make sense and clarify things. If not, we'll catch up offline to discuss.

thank you - yes, this helped me getting more insight.

mmeeks (approved):
All looks good to me - but only on Red =) thanks Ash & Sven.

Issue #9276, opened 2024-06-13
UnitPerf CSV generation


mmeeks:
We have a rudimentary performance unit test in tests/UnitPerf.cpp

It generates some numbers - and histogram and so on - which are pretty;

But we want to start capturing and logging that data in a form we can easily re-use, consume and chart - and/or at least monitor.

If you make 'testPerf' fail - by crashing it 'assert(false);' or somesuch; then the Makefile will kindly remind you of how to run just this one test - which is what you want.

Then - it would be ideal to have a set of CSV files - and prolly we should do this in the normal unit tests anyway for good measure and we should in each case use the git hash as the primary key / first item.

We should prolly split CPU, vs, Latency vs. Network - and generate 3x separate CSVs.

CPU should have the run-time in it, and as we go forward - more and more accurate CPU metrics - ideally from the libpfm API not the SysStopwatch class. But for now just getting something we can graph is key.

For Latency - we have a histogram we should horizontal-ize into CSV

And for Network - we should dump incoming  & outgoing bandwidth, and then have some defined column headers for each type of thing, and dump the breakdown there so we can see it over time. I expect bandwidth to be the most reliable indicator here - and the others to jitter unhelpfully between runs =)

@Minion3665 can help with code pointers I expect.

Thanks !

amkarn258:
Hi @mmeeks ,

Please assign this to me if no one else is assigned yet. I would like to contribute

mmeeks:
Hi Mayank - thanks so much for getting involved! I filed this for an intern - Elliot over the summer - but as long as you commit your code to a branch regularly, no doubt you could work together with him to improve this :-) I don't expect Elliot to start looking at this for another week or so - so - go for it ! =)

mmeeks:
@elliotfreebairn1 so - some other thoughts for expansion:

1. how jittery are the numbers - for the same commit ? can you do some stats on that & build a nice spreadsheet & attach here ?
2. can we record other interactive traces - and/or refresh the traces we have to make them re-playable, I suspect our existing traces are rather out of date
3. can we connect more traces into the performance testing framework; ideally we'd generate different metrics for each of them https://perf.libreoffice.org/ has some examples of how that might look - and we could re-use / build on that framework.

Thanks ! =)

elliotfreebairn1:
@mmeeks 
For the 1st point, do you mean running the same unit tests on this device a fair few times and analysing the variations in the data?

Minion3665:
[...]

PR #9709, opened 2024-08-05
 feat: Changes to UnitPerf measuring

elliotfreebairn1:
* Resolves: # 
* Target version: master 

### Summary
Working on issue #9276

Signed-off-by: elliotfreebairn <elliot.freebairn@collabora.com>
Change-Id: I868a0a1c91b30392ab002ce149cefae0dd3036cc

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitPerf.cpp
+++ b/test/UnitPerf.cpp
@@ -10,6 +10,7 @@
  */
 
 #include <config.h>
+#include <config_version.h>
 
 #include <Unit.hpp>
 #include <Util.hpp>
@@ -31,9 +32,7 @@
 /// Save torture testcase.
 class UnitPerf : public UnitWSD
 {
-    void testPerf();
-
-    void dumpCPUTimeToCSV(long cpuTime);
+    void testPerf(std::string testType, std::string fileType, std::string tracesStr);
 
     void configure(Poco::Util::LayeredConfiguration& config) override
     {
@@ -47,24 +46,29 @@ class UnitPerf : public UnitWSD
 public:
     UnitPerf();
     void invokeWSDTest() override;
+    void onDocumentLoading() override;
+    void onDocumentLoaded() override;
     std::unique_ptr<Util::SysStopwatch> _timer;
+    std::shared_ptr<Stats> stats;
 };
 
-void UnitPerf::testPerf()
+void UnitPerf::testPerf(std::string testType, std::string fileType, std::string traceStr)
 {
-    auto stats = std::make_shared<Stats>();
+    stats = std::make_shared<Stats>();
+    stats->setTypeOfTest(testType);
 
     TerminatingPoll poll("performance test");
 
-    std::string docName = "empty.odt";
+    std::string docName = "empty." + fileType;
 
     std::string filePath, dummy;
+
     helpers::getDocumentPathAndURL(docName, filePath, dummy, "testPerf");
 
-    const std::string tracePath;
+    const std::string tracePath = TDOC + traceStr;
     StressSocketHandler::addPollFor(
         poll, helpers::getTestServerURI("ws"),
-        filePath, TDOC "/../traces/perf-writer.txt",
+        filePath, tracePath,
         stats);
 
     do {
@@ -74,19 +78,7 @@ void UnitPerf::testPerf()
     stats->dump();
 }
 
-void UnitPerf::dumpCPUTimeToCSV(long cpuTime)
-{
-    std::ofstream file("CPUUsage.csv", std::ios::out | std::ios::app);
 
-    if(file.tellp() == 0)
-    {
-        file << "CPU Time";
-        file << "\n";
-    }
-
-    file << cpuTime;
-    file << "\n";
-}
 
 UnitPerf::UnitPerf() : UnitWSD("UnitPerf")
 {
@@ -102,14 +94,27 @@ void UnitPerf::invokeWSDTest()
     std::cerr << "startup: " << _timer->elapsedTime().count() << "us\n";
     _timer->restart();
 
-    testPerf();
+    testPerf("writer", "odt", "/../traces/perf-writer.txt");
+
+    testPerf("calc", "ods", "/../traces/perf-calc.txt");
 
     long cpuTime = _timer->elapsedTime().count();
-    dumpCPUTimeToCSV(cpuTime);
 
     std::cerr << "test: " << cpuTime << "us\n";
 
     exitTest(TestResult::Ok);
+}
+
+//Called when document loading process starts e.g. setup finishes
+void UnitPerf::onDocumentLoading()
+{
+    stats->perfSetupLoadEditPhases("Setup");
+}
+
+//called when document has been loaded into core
+void UnitPerf::onDocumentLoaded()
+{
+    stats->perfSetupLoadEditPhases("Load");
 }
 
 UnitBase* unit_create_wsd(void) { return new UnitPerf(); }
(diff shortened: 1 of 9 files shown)


elliotfreebairn1:
This is not fully finished as i haven't created a impress and draw trace but i wanted to see if i'm along the right path especially on if the hook is in the right place and if i messed up anywhere in here. 
Here is screenshot of what the outcome of a unit test would look like:

test/UnitPerf.cpp
```diff
@@ -86,13 +94,29 @@ void UnitPerf::invokeWSDTest()
...
+
+    std::cerr << "test: " << cpuTime << "us\n";
 
     exitTest(TestResult::Ok);
 }
 
+//Called when document loading process starts e.g. setup finishes
+void UnitPerf::onDocumentLoading()
+{
+    stats->perfSetupLoadEditPhases("Setup");
```
mmeeks, line 111: Hmm; sounds like stats->endPhase(foo) might be a better name (?) I assume that ends it (?)

tools/Replay.hpp
```diff
@@ -216,32 +267,104 @@ struct Stats {
         dumpMap(_recvd);
     }
 
-    void dumpStressToCSV(size_t runMs)
+
+    //loading phase will record the set up times and loaded phase records the document loading times
```
mmeeks, line 271: //<space>text =)

tools/Replay.hpp
```diff
@@ -216,32 +267,104 @@ struct Stats {
...
+    void perfSetupLoadEditPhases(std::string phase)
```
mmeeks, line 272: So - I guess, it's perhaps a better self documenting API to have an enumeration of phase type; there are some nice macros Ash has that allow you to get a nice string name from each Enum value - we can then use as the string for that; and then it's more clear what the phases are and that we move through them; checkout: common/Log.hpp and the STATE_ENUM macro =)

tools/Replay.hpp
```diff
@@ -216,32 +267,104 @@ struct Stats {
...
+        _start = std::chrono::steady_clock::now();
+
+        size_t cpuTime = _timer->elapsedTime().count();
+        _timer.reset(new Util::SysStopwatch);
 
-        if (file.tellp() == 0)
+        perfStatsList.push_back(GetStressStats(runMs, phase));
+        perfStatsList.push_back(GetCPUUSageStats(cpuTime, phase));
+
+        if(phase == "Edit")
```
mmeeks, line 284: enums more reliable for comparison;

tools/Replay.hpp
```diff
@@ -216,32 +267,104 @@ struct Stats {
...
+        PerfMetricInfo cpuStats = {testPhase, "CPU Usage (us)", cpuUsage};
+        return cpuStats;
+    }
+
+    std::vector<PerfMetricInfo> GetNetworkStats(size_t recievedKb, size_t sentKb, std::string testPhase)
+    {
+        std::vector<PerfMetricInfo> networkStatsList;
+
+        networkStatsList.push_back(PerfMetricInfo{testPhase, "Bytes recieved (kB)",  recievedKb});
+        networkStatsList.push_back(PerfMetricInfo{testPhase,"Bytes sent (kB)", sentKb});
```
mmeeks, line 327: one space after comma, not none, not two ;-)
Please read your diff carefully before committing.

tools/Replay.hpp
```diff
@@ -216,32 +267,104 @@ struct Stats {
...
+    }
+
+    void dumpPerfStatsToCSV(std::vector<PerfMetricInfo> perfData)
+    {
+        std::ofstream file("PerformanceMetricsSummary.csv", std::ios::out | std::ios::app);
 
         if(file.tellp() == 0)
         {
-            file << "Incoming bandwidth (kB/s)" << ",Outgoing bandwidth (kB/s)";
+            file << "Commit Hash" << "," << "Date" << "," << "Test" << "," << "Phase" << "," << "Metric" << "," << "Value";
```
mmeeks, line 344: All Capitalized Except value =)

wsd/DocumentBroker.cpp
```diff
@@ -873,6 +873,11 @@ bool DocumentBroker::download(
     {
         _docState.setStatus(DocumentState::Status::Downloading);
 
+        if(_unitWsd != nullptr)
+        {
+            _unitWsd->onDocumentLoading();
```
mmeeks, line 878: Looks reasonable I guess =)

mmeeks (commented):
Looking reasonable =) needs a bit of fixup, some rebasing / squashing fixes into commits and so on - but getting somewhere good :-)

test/UnitPerf.cpp
elliotfreebairn1: Yeah that would make sense, this gets called when the document is getting downloaded from storage (wopi host), so i assume from the beginning of the unit test to then is the setup phase?

tools/Replay.hpp
elliotfreebairn1: I've tried to implement this and it does work but i'm unsure if this is the correct way to implement the Enum value?

mmeeks: Looks lovely =)

tools/Replay.hpp
```diff
@@ -113,11 +119,13 @@ struct Histogram {
             }
         }
 
-        file << totalTiles << ",";
-        file << subTenCount << ",";
-        file << subOneHundredCount << ",";
-        file << overOneHundredCount << ",";
-        file << "\n";
+        std::vector<PerfMetricInfo> latencyStatsList;
+        latencyStatsList.push_back(PerfMetricInfo {testPhase,typeOfLatency + " Total tiles", totalTiles});
```
mmeeks, line 123: consistency 'Phase,typeOf' - no space - =)

tools/Replay.hpp
```diff
@@ -29,6 +29,20 @@
 
 #include <iostream>
 #include <fstream>
+#include <config_version.h>
+#include <ctime>
+#include <Util.hpp>
+#include <common/Log.hpp>
+
+
```
mmeeks, line 37: Unusually large to have 3x spaces - vertical space is a non-renewable resource to be used tastefully =)

tools/Replay.hpp
```diff
@@ -152,6 +165,8 @@ struct Stats {
     std::unordered_map<std::string, MessageStat> _recvd;
     std::unordered_map<std::string, MessageStat> _sent;
 
+    std::vector<PerfMetricInfo> perfStatsList;
```
mmeeks, line 168: Is that a member - if so it needs an _ prefix so code-readers can tell that -cf. all other member variables =)

tools/Replay.hpp
```diff
@@ -256,46 +268,106 @@ struct Stats {
...
+    }
+
+    void dumpPerfStatsToCSV(std::vector<PerfMetricInfo> perfData)
+    {
+        std::ofstream file("PerformanceMetricsSummary.csv", std::ios::out | std::ios::app);
 
         if(file.tellp() == 0)
         {
-            file << "Bytes recieved (kB)" << ", Bytes sent(kB)";
+            file << "COMMIT HASH" << "," << "DATE" << "," << "TEST" << "," << "PHASE" << "," << "METRIC" << "," << "value";
```
mmeeks, line 346: Hmm - now all of these legends are ALLCAPS - shouting at me, but the last item is still lower-case "value" - why ? =) I was suggesting that if you have: "Foo, Baa, baz" - something is wrong with baz - instead we got "FOO, BAA, baz" ;-) Can we have: "Foo, Baa, Baz" ?

mmeeks (approved):
Better; it can go in without further review; just trying to build your sub-clinical neurosis & code skimming speed with a few whitespace comments that jarred as I read =)

elliotfreebairn1:
Thank you for picking them out, i definitely need to switch on my brain when it comes to code skimming. Will fix these then merge :)

PR #9711, opened 2024-08-05
cool#9710 admin console: expose poco version

vmiklos:
It can happen that performance depends on what poco version is used, in
this case exposing this information helps debugging.

Of course it won't help with old versions, but at least start showing
this info, to help in the future.

Change-Id: I32439d70022b918094a7110bfce3aff29ce8b35d


vmiklos:

Just a small debug helper, suggested by Andras. I intentionally made the JS code check for the member, so in case one day we would drop the poco dependency, then we can safely remove that JSON key without worrying about backwards compatibility.

caolanm (approved):

PR #9716, opened 2024-08-06
test and workaround Ubuntu 24.04 apparmor defaults for make run

caolanm:
With Ubuntu 24.04 apparmor rules default to denying user namespaces. https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2046844

podman though is allowed by default, so podman unshare can (currently) be used to create an initial namespace env to work within.

Change-Id: Ic2a9f5b9ff8bb403c88e48e05a52ec44d3501a82

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
@grandinj If you reenable/restore the apparmor defaults on your Ubuntu 24.04, does this workaround the issue for make run?

grandinj:
Yup, this does the trick, thanks!

grandinj (approved):

PR #9718, opened 2024-08-06
arial-label

hcvcastro:
- **browser: accessibility: fix 'arial-label' property to user list button**
- **browser: accessibility: fix 'aria-label' to all buttons**


eszkadev (approved):
looks good

PR #9722, opened 2024-08-06
duplicate _update() calls

caolanm:
_onStatusMsg(textMsg) is only called from one place and there it is always followed by a call to _update();

since:

commit dffaeb228bb787af2be6b7a88048fc97c1df07d5
AuthorDate: Mon Nov 14 13:44:41 2022 +0100

    masterpage: correclty refresh after switching mode

    Mode is sent via status message so handle it there to avoid
    race

but each _onStatusMsg impl might also call _update() so there are duplicate tilecombine requests

Change-Id: I6403852321cb727750aba4874d053a2d3c37abc4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -459,7 +459,6 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 				source: 'status'
 			});
 			this._resetPreFetching(true);
-			this._update();
 			if (firstSelectedPart) {
 				this._switchSplitPanesContext();
 			}
--- a/browser/src/layer/tile/ImpressTileLayer.js
+++ b/browser/src/layer/tile/ImpressTileLayer.js
@@ -280,7 +280,6 @@ L.ImpressTileLayer = L.CanvasTileLayer.extend({
 			}
 			this._selectedMode = (command.mode !== undefined) ? command.mode : 0;
 			this._resetPreFetching(true);
-			this._update();
 			var partMatch = textMsg.match(/[^\r\n]+/g);
 			// only get the last matches
 			var newPartHashes = partMatch.slice(partMatch.length - this._parts);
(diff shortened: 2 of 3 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -5238,8 +5243,20 @@ L.CanvasTileLayer = L.Layer.extend({
 			rectangles.push(rectQueue);
 		}
 
-		for (var r = 0; r < rectangles.length; ++r)
+		for (var r = 0; r < rectangles.length; ++r) {
+
+			// While we are actively scrolling, filter out duplicate
+			// (still) missing tiles requests during the scroll.
+			if (this._moveInProgress) {
+				var rectStr = JSON.stringify(rectangles[r]);
```
mmeeks, line 5251: Quick skim - but surely we should track the tiles we're asking for and not the rectangles here and ensure we don't fetch a tile 'pixel' from that 'bitmap' twice (?)

mmeeks (commented):
Looks good in concept; is this a regression fix - so green; or better for a red release ? =)

browser/src/layer/tile/CanvasTileLayer.js
caolanm: this looked simple

caolanm:
As far as I can tell there is no recent regressions here, first one might be seen as a regression from 2022 I think. The other looks like how it has ~always been

browser/src/layer/tile/CanvasTileLayer.js
mmeeks: I -think- in the tilecombine logic we already have tile-by-tile de-duplication & aggregation logic that we could re-factor and re-use, but was just a quick skim =)

caolanm:
split this into the uncontroversial bit and the other bit to handle in follow up

PR #9725, opened 2024-08-07
slideshow: pause all videos when changing the slide

quikee:
Make sure the video is paused when we change the slide, so that we don't keep playing the video in the background.

Change-Id: I39b9118f7606a997f4cdf2eb4e3f1b736812074f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


codewithvk (approved):
Code looks good to me

PR #9734, opened 2024-08-08
Fix other views' cursor visibility.

gokaysatir:
showcursor variable is set by the core side. Extra assignment is removed.

Change-Id: I010be4f9f540ba53c982d9b21315c020c7c4147d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/OtherViewCursorSection.ts
+++ b/browser/src/canvas/sections/OtherViewCursorSection.ts
@@ -86,7 +86,6 @@ class OtherViewCursorSection extends HTMLObjectSection {
             OtherViewCursorSection.sectionPointers.push(section);
         }
 
-        section.sectionProperties.showCursor = true;
         section.setShowSection(section.checkMyVisibility());
         section.onNewDocumentTopLeft();
         section.adjustHTMLObjectPosition();


eszkadev (approved):
Looks good. The property is set inside: `_onViewCursorVisibleMsg` in `CanvasTileSection`.

BTW. it would be good to not expose implementation details outside. Could we create public method `showCursor()` in OtherViewCursorSection which will be used in favor of direct property assignment. It's easier to spot the place where it happens by grep.
Also if we will need to modify the code in the future it's in the single place which has to know all the property names.

gokaysatir:
> Looks good. The property is set inside: `_onViewCursorVisibleMsg` in `CanvasTileSection`.
> 
> BTW. it would be good to not expose implementation details outside. Could we create public method `showCursor()` in OtherViewCursorSection which will be used in favor of direct property assignment. It's easier to spot the place where it happens by grep. Also if we will need to modify the code in the future it's in the single place which has to know all the property names.

Thanks for the comment. I agree, that's why i am adding  public static methods here and there in the classes :) This needs some time, that may find itself a place in a middleware. There are multiple variables that sets the cursors visibility.

Issue #9740, opened 2024-08-08
Mobile: Calc: Dragging to fill the column pops up an empty menu


Minion3665:
# Steps to reproduce
- Open a .ods file
- add a number in a cell.
- Select and Drag the cell selection downwards to fill other cells

# Expected

The selection fills downwards into cells you drag over, and there is no other effect

# Actual

The selection fills, but an empty menu pops-up

# Versions

OK in 23.05.7.4, git hash: c36099e3
Not OK in 24.05.99-752, git hash: 5af63cc3

# Videos

Previous:
https://github.com/user-attachments/assets/b2891bc4-4a90-425f-b67b-b13019b9a050

Current behavior:
https://github.com/user-attachments/assets/fc8cc145-a703-4245-8358-fe7d1a1f1d49

Minion3665:
This does not require using the mobile app - even the mobile view in Firefox triggers this bug
![image](https://github.com/user-attachments/assets/f04f4275-4562-4fc0-8e15-3baaa52a3008)

PR #9741, opened 2024-08-08
fix(contextMenu): Avoid displaying empty menus

Minion3665:
It's possible sometimes for us to filter out everything in a context menu with _createContextMenuStructure. If we do that on mobile, we will end up with an empty mobile wizard, so it's much better to skip the context menu altogether...

Fixes: #9740
Change-Id: Iae65aa95a38c4e46f4bea544bd0b41195000cba1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks good

PR #9755, opened 2024-08-09
jsdialog: css: use CSS grid in icon view

eszkadev:
By default use well aligned 3 column grid.
In styles preview use old flex layout so
we don't change the behavior we had.

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1019,10 +1019,10 @@ input[type='number']:hover::-webkit-outer-spin-button {
 
 .ui-iconview {
 	overflow: auto;
-	display: inline-flex;
-	display: -ms-inline-flexbox;
-	flex-wrap: wrap;
 	justify-content: space-between;
+	display: grid;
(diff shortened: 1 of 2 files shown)


eszkadev:
icongrid

codewithvk (dismissed):
Can we add scrollbar like this ?

#transitions_icons {
    max-height: 40vh;
    overflow: scroll;
}

eszkadev:
> Can we add scrollbar like this ?
> 
> #transitions_icons { max-height: 40vh; overflow: scroll; }

done, height limit to 40% of the screen

pedropintosilva (approved):
Tested, it seems good. Both on desktop and laptop (not possible to access this on mobile). I have also added additional transition thumbnails to, it works no matter the number even or uneven . Thanks

pedropintosilva:
> > Can we add scrollbar like this ?
> > #transitions_icons { max-height: 40vh; overflow: scroll; }
> 
> done, height limit to 40% of the screen

also with this we avoid for the majority of cases 2 scroll bars which is awesome, thanks

PR #9763, opened 2024-08-09
SlideShow: reset cached slides to show updated content

eszkadev:
Introduce container for cached content. Reset data when we receive new presentation info. This should be improved in the future to detect and reset only modified slides. Also we want to put into cach all other layers.


quikee (approved):
looks good to me

PR #9766, opened 2024-08-09
private/pedro/status bar search btns disable status

pedropintosilva:
- **Status bar: Enable/disable search buttons**
- **Status bar: Make cancel search icon more subtle**

### DIFF
--- a/browser/images/dark/lc_cancel.svg
+++ b/browser/images/dark/lc_cancel.svg
@@ -1,1 +1,1 @@
-<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><g fill="none" stroke="#ed3d3b" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><path d="m4 20 16-16"/><path d="m20 20-16-15.9999998"/></g></svg>
+<svg version="1.1" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="m7 17 10-10" fill="none" stroke="#d4d4d4" stroke-linecap="round" stroke-width="1.4"/><path d="M 17,17 7,7" fill="none" stroke="#d4d4d4" stroke-linecap="round" stroke-width="1.4"/></svg>
--- a/browser/images/lc_cancel.svg
+++ b/browser/images/lc_cancel.svg
@@ -1,14 +1,1 @@
-<?xml-stylesheet type="text/css" href="icons.css" ?>
-<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-  <g id="background"
-	 class="icn icn--text-color"
-     fill="none"
-     stroke="#696969"
-     stroke-linecap="round"
-     stroke-linejoin="round"
-     stroke-width="2"
-     >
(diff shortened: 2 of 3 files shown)


browser/src/control/Control.StatusBar.js
```diff
@@ -106,7 +106,9 @@ class StatusBar extends JSDialog.Toolbar {
 		var searchInput = L.DomUtil.get('search-input');
 		if (e.count === 0) {
 			this.enableItem('searchprev', false);
+			this.enableItem('searchprev-button', false);
```
eszkadev, line 109: I think it is a workaround. We should always use "id" in this case "searchprev" only. Probably we have some bad code in the enable/disable implementation

pedropintosilva:
This seems quite serious to don't merge. I prefer to have functional search buttons ASAP then wait for the ideal approach here.

gokaysatir (approved):
Yep, sorry @pedropintosilva. Let me approve the request. Then we'll have a follow up for even better improvements :)

eszkadev (changes requested):
sorry, I really think it's some fundamental problem we should fix...

eszkadev:
I think we fail here: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2942
the example item is: `{type: 'customtoolitem',  id: 'searchprev', command: 'searchprev', text: _UNO('.uno:UpSearch'), enabled: false, pressAndHold: true},`
which has implementation in: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2499
which uses `_unoToolButton`
and this leads to https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2291  where `-button` is present

And that is not the root node. The rule is: we only execute action on the root node of widget (in this example: `div`).
So it is not applied to the button itself. To solve that problem we have helper which synchronizes state with the root node:
eg. `JSDialog.SynchronizeDisabledState(div, [spinfield]);`

eszkadev (approved):

pedropintosilva:
Cypress (desktop)
```
 Table operations
       Set minimal row height.:
     AssertionError: Timed out retrying after 10000ms: expected '[ <td>, 5 more... ]' not to have attribute 'height'
```

unrelated, re-try

pedropintosilva:
Thanks, but now without my changes and with your commit @eszkadev I now see the next and prev buttons again disabled:

image

I wonder,  and since @gokaysatir was already reviewing, if you @eszkadev would prefer @gokaysatir to take this from you

eszkadev:
yes, @gokaysatir  please check

gokaysatir:
@eszkadev can you have a look at the last commit when you have time? I think "disabled" class usage is being removed from some of the widgets. unotoolbuttons seem to be one of them.

gokaysatir (approved):

eszkadev:
yes, it seems we use attribute now: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2978

eszkadev:
thanks, seems to work now :)

PR #9769, opened 2024-08-09
test: Make cypress fail on exception

Minion3665:
In I636e07a93d2ed17d56b2a9fc11d66b7e4c3c19f3, Michael Meeks added a new
popup to show when we have an exception in an event. This is excellent,
and is catching errors. Unfortunately, the messages we get in our
Cypress tests when this fails are... not amazing.

They don't include the error, and instead the new popup just makes some
actions that we wanted to perform silently fail to happen. This causes
later assertion errors, but they are disconnected from the root cause -
the error. It's much neater to throw here if we're in cypress, failing
the test.

Change-Id: Ia729489ba2adc3be0e9286f942d128acafd29858* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/debug.html
+++ b/browser/html/debug.html
@@ -43,6 +43,29 @@
        var form = document.getElementById("form");
        form.action = constructWopiUrl();
        form.submit();
+
+        window.addEventListener(
+            'message',
+            (event) => {
+                if (event.origin !== window.origin) {
+                    return;
+                }
+
+                if (typeof event.data !== 'object' || event.data === null) {
+                    return;
+                }
+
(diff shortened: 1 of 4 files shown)


Minion3665:
so, this naïve solution to this doesn't quite work ... I think due to iframes not bubbling errors ... back to the drawing board, possibly this needs something in the tests themselves (e.g. how https://www.npmjs.com/package/cypress-fail-on-console-error looks to do this from their usage examples, but I guess this is also unlikely to work with iframes...)

eszkadev (approved):
Nice! and fails fast

PR #9771, opened 2024-08-09
split out a deprecated subpackage of the helpers not needed with namespaces

caolanm:
Default namespaces to on, and move non-namespaces helpers to a deprecated subpackage

- centos
  - 8.0 works with namespaces out of the box
  - 7.9 (EOL*) not enabled by default, possible with echo 10000 > /proc/sys/user/max_user_namespaces
    * RHEL7 is EOM
- debian
  - 11 (bullseye) works with namespaces out of the box
  - 10 (buster) EOL, not enabled by default, possible with sudo sysctl -w kernel.unprivileged_userns_clone=1
  - 9 (stretch) EOL, not enabled by default, possible with sudo sysctl -w kernel.unprivileged_userns_clone=1
- ubuntu
  - 16.04 works with namespaces out of the box

Change-Id: Iffabea0a871d23797c1482cb8d2ca7de3e98f34a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,11 +83,15 @@ browser/dist
 browser/compilets
 browser/typescript_js
 
+coolforkit
+coolforkitns
+coolforkit-caps
+coolforkit-nocaps
+coolforkit-ns
 connect
 lokitclient
 coolwsd
 loolwsd
-coolforkit
 coolmount
 coolmap
 coolconvert
@@ -95,14 +99,12 @@ coolsocketdump
 coolstress
 coolconfig
 loolconfig
-coolforkit-nocaps
 loadtest
 unittest
 clientnb
 libsimd.a
 coolwsd-inproc
 coolbench
-coolforkitns
 
 # Fuzzer
 *_fuzzer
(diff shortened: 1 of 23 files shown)


caolanm:
This defaults mount namespaces to on and moves coolmount and coolforkit to a "coolwsd-deprecated" subpackage rpm/deb.

Are we happy with that name or do we want something else?
Do we want to continue with the "coolforkitns/coolforkit" names or take the opportunity to rename to use "coolforkit" for the namespace one and say "coolforkit-caps" for the fallback one.

mmeeks (approved):
Looks great to me; good to have new names for support I think - so lets not re-use coolforkit - having a suffix in each case is fine - whichever you think best. A -deprecated suffix on the package sounds sensible to me unless @timar objects =)

timar (approved):

weberhofer:
It would be great to explicitly name the necessary changes in the release notes. https://www.collaboraonline.com/collabora-online-24-04-release-notes/ 

My  Debian 11 instance installed in Spring 2024 did not have the required setting in sysctl. After upgrading from Debian 11 to 12 the setting has not been set.

I found #10152 and installed the deprecated package which solved the problem for me. Didn't have time to check the kernel.unprivileged_userns_clone=1 solution yet.

Issue #9750, opened 2024-08-08
Help > Server Audit Does Nothing and Should Be Disabled/Hidden in Read-Only Mode


Darshan-upadhyay1110:
Description:

- The "Server Audit" option under Help does nothing in read-only mode. 
- It should be removed or hidden in the UI to avoid confusion in read only mode.

![image](https://github.com/user-attachments/assets/533a4070-3365-457a-a3c4-65b49bd236e3)

PR #9774, opened 2024-08-10
menubar: disable "serveraudit" option in help menu items in read only mode

rparth07:
Change-Id: Iafeeac96893816b813ff3d6fe0c5a845adcb759d

* Resolves: #9750
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1866,10 +1866,7 @@ L.Control.Menubar = L.Control.extend({
 					if (id === 'insertcomment' && (self._map.getDocType() !== 'drawing' && !app.isCommentEditingAllowed()))
 						found = false;
 					if (!found) {
-						$(aItem).addClass('disabled');
-						aItem.title = _('Read-only mode');
-					} else {
-						$(aItem).removeClass('disabled');
+						$(aItem).hide();
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


Darshan-upadhyay1110 (commented):
@rparth07 

However, would it be possible to remove the element from the list in read-only mode instead of just disabling it? My concern is that if we add more options in the future, disabling them could become a never-ending cycle. It would be more effective to hide the element entirely rather than just disabling it.

@eszkadev please correct me if i an wrong here  :)

Darshan-upadhyay1110 (commented):
This one similar to https://github.com/CollaboraOnline/online/pull/9775.

eszkadev:
yes, would be good to also remove that item

Darshan-upadhyay1110 (changes requested):
@rparth07 please rebase and if possible try to remove the items which are not needed in RO. 

ATB,
Darshan

rparth07:
On the similar line, Should we also remove all uno command elements instead of disabling them in Read Only mode? 
cc: @eszkadev @Darshan-upadhyay1110

eszkadev:
hiding not disabling patch - is fine
removing serveraudit in readonly - I think it could be useful for admin tom check also RO case

Darshan-upadhyay1110:
Hi @rparth07! Could you please rebase and resolve the conflicts? 

>Also, regarding the patch for hiding (not disabling) – that looks good. As for removing serveraudit in read-only mode, I believe it might still be helpful for admins to check in read-only cases as well. 

Please review this comment and make the necessary changes. Thanks so much! 😊

rparth07:
> hiding not disabling patch - is fine removing serveraudit in readonly - I think it could be useful for admin tom check also RO case

@eszkadev Thank you for reviewing the changes. Could you please confirm if the intended behavior is that, In read-only mode, for admin users, the 'Server Audit' option should be disabled while for other users, it should be hidden?

cc: @Darshan-upadhyay1110

eszkadev:
we need to solve merge conflict, @rparth07 are you still working on it? :)

rparth07:
> we need to solve merge conflict, @rparth07 are you still working on it? :)

Thank you for the update. I’ll take care of resolving the merge conflicts.

rparth07:
I resolved the merge conflicts and enabled serveraudit option for admin. 
Can you please review @eszkadev @Darshan-upadhyay1110 
Thanks :)

Darshan-upadhyay1110 (approved):
Looks good :)

Admin user read-only mode:
image



Non-admin user read-only mode: `server-audit` option is hidden
image

eszkadev:
Thank you both, for fix and checking :)

Issue #9746, opened 2024-08-08
Remove "Hide Menubar" Option from View Tab in read only mode


Darshan-upadhyay1110:
Description: Hiding the menubar in read-only mode makes it difficult to revert to the normal view. This option should be hidden in read-only mode.

![image](https://github.com/user-attachments/assets/a7528dd4-7b41-43fd-b0f1-034d1454b2f4)

CodingwithKetan:
I’d like to work on this issue. Could you please assign it to me?

Darshan-upadhyay1110:
> I’d like to work on this issue. Could you please assign it to me?

@CodingwithKetan Sure. Thanks :)

Darshan-upadhyay1110:
Hello @CodingwithKetan any luck here ?

CodingwithKetan:
> Hello @CodingwithKetan any luck here ?

Hi @Darshan-upadhyay1110 ,

I’m currently diving into the code structure and flow of control to get a better understanding. I’m reviewing the architecture and how different components interact with each other. This is taking some time, but I’m making progress and will update you once I have a clearer picture and the next steps.

Thanks for your patience!

PR #9775, opened 2024-08-10
menubar: disabled hideMenuBar button in readonly mode

CodingwithKetan:
Change-Id: I5658869c7f901dfdec26c7329f48da54bb179c0e

* Resolves: #9746 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -1345,7 +1345,7 @@ L.Control.Menubar = L.Control.extend({
 			'downloadas-odp', 'downloadas-ppt', 'downloadas-pptx', 'downloadas-odg', 'exportpdf' , // file menu
 			!window.ThisIsAMobileApp ? 'exportdirectpdf' : 'downloadas-pdf', !window.ThisIsAMobileApp ? 'exportepub' : 'downloadas-epub', // file menu
 			'downloadas-ods', 'downloadas-xls', 'downloadas-xlsx', 'downloadas-csv', 'closedocument', // file menu
-			'fullscreen', 'zoomin', 'zoomout', 'zoomreset', 'showstatusbar', 'togglemenubar', 'showresolved', 'toggledarktheme', // view menu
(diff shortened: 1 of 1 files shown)


Darshan-upadhyay1110 (commented):
@CodingwithKetan, great patch to fix the issue! 

However, would it be possible to remove the element from the list in read-only mode instead of just disabling it? My concern is that if we add more options in the future, disabling them could become a never-ending cycle. It would be more effective to hide the element entirely rather than just disabling it.

@eszkadev please correct me if i an wrong here  :)

Darshan-upadhyay1110:
Also i have just did a small try on code it seems easy to this hide task so this can be still a `easy hack` 😉

CodingwithKetan:
> Also i have just did a small try on code it seems easy to this hide menubar option and not just disable it so this can be still a `easy hack` 😉

Thank you @Darshan-upadhyay1110 for the feedback! I see your point about the potential issue with simply disabling elements. I’ll update the patch to remove the element from the list in read-only mode instead of just disabling it.

CodingwithKetan:
> @CodingwithKetan, great patch to fix the issue!
> 
> However, would it be possible to remove the element from the list in read-only mode instead of just disabling it? My concern is that if we add more options in the future, disabling them could become a never-ending cycle. It would be more effective to hide the element entirely rather than just disabling it.
> 
> @eszkadev please correct me if i an wrong here :)



> Also i have just did a small try on code it seems easy to this hide menubar option and not just disable it so this can be still a `easy hack` 😉

Thank you @Darshan-upadhyay1110  for the feedback! I see your point about the potential issue with simply disabling elements. I’ll update the patch to remove the element from the list in read-only mode instead of just disabling it.

eszkadev (approved):
looks ok, but yes, would be good to just remove that item from RO view

Darshan-upadhyay1110:
> looks ok, but yes, would be good to just remove that item from RO view

Sure let's merge this PR and as followup work we will do the next steps :)

Darshan-upadhyay1110:
Thanks, @CodingwithKetan looking forward to more commits in upstream. Checkout the issues section, a bunch of issues are waiting for a contributor.

Happy hacking 😉

PR #9777, opened 2024-08-12
cool#9621: Have WSD Keep Download-Socket open for file and favicon.ico requests (ClientRequestDispatcher); ..

sgothel:
### Summary

WSD's ClientRequestDispatcher shall keep the socket open on successful document and favicon download requests.

HttpHelper::sendFile(..) is refactored from HttpHelper::sendFileAndShutdown(..) w/o the socket closing (shutdown) and used in WSD's ClientRequestDispatcher accordingly.

WebSocketHandler: Added public isConnected() StreamSocket queries, allowing our unit tests to validate the connected state.

#### Keep-Alive Utilization
FileServer.cpp's FileServerRequestHandler::handleRequest(..) also keeps the socket alive (ENABLE_DEBUG and env 'COOL_SERVE_FROM_FS') with this patch.

Check whether this is desired, as not covered by a unit test!

+++

Analyze whether the following maintained HttpHelper::sendFileAndShutdown(..) use-cases shall be converted to keep-alive as well:
- WSD's ClientSession::handleKitToClientMessage()

+++

Unit tests w/ WopiTestServer still use HttpHelper::sendFileAndShutdown(..):
- test/UnitWOPIFileUrl.cpp:145
- test/UnitWOPITemplate.cpp:81

Currently the whole WopiTestServer code closes the sockets even after serving a request successfully.
We might want to change this behavior towards keep-alive as well? This would also allow us to employ 'local' performance unit tests with such a WopiTestServer.

#### Poco HTTP Message Parsing
Replace Poco http request message parsing
starting with Ash's net/HttpRequest

Change-Id: I83c09e0a8101931efece42e42e5a8e3abd7e696f

* Resolves: #9621
* Target version: master 

### Checklist

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -364,7 +364,8 @@ public:
         ConnectionToken,
         None, //< No `Connection` header token set
         Close, //< `Connection: close` [RFC2616 14.10](https://www.rfc-editor.org/rfc/rfc2616#section-14.10)
-        KeepAlive //< `Connection: Keep-Alive` Obsolete [RFC2068 19.7.1](https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1)
+        KeepAlive, //< `Connection: Keep-Alive` Obsolete [RFC2068 19.7.1](https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1)
+        Upgrade //< `Connection: Upgrade` HTTP/1.1 only [RFC2817](https://www.rfc-editor.org/rfc/rfc2817)
     );
 
     /// Describes the header state during parsing.
@@ -470,18 +471,21 @@ public:
             return ConnectionToken::Close;
         } else if( Util::iequal("keep-alive", token) ) {
             return ConnectionToken::KeepAlive;
+        } else if( Util::iequal("upgrade", token) ) {
+            return ConnectionToken::Upgrade;
         } else {
             return ConnectionToken::None;
         }
     }
     void setConnectionToken(ConnectionToken token) {
-        if( ConnectionToken::Close == token ) {
-            set(CONNECTION, "close");
-        } else if( ConnectionToken::KeepAlive == token ) {
-            set(CONNECTION, "Keep-Alive");
-        } else {
-            remove(CONNECTION);
+        std::string value;
+        switch( token ) {
+            case ConnectionToken::Close: value = "close"; break;
+            case ConnectionToken::KeepAlive: value = "Keep-Alive"; break;
+            case ConnectionToken::Upgrade: value = "Upgrade"; break;
+            default: remove(CONNECTION); return;
         }
+        set(CONNECTION, value);
     }
 
     /// Adds a new "Cookie" header entry with the given content.
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -360,10 +360,11 @@ public:
     static constexpr const char* CONNECTION = "Connection";
 
     /// Describes the `Connection` header token value
-    STATE_ENUM(ConnectionToken,
-               None,        //< No `Connection` header token set
-               Close,       //< `Connection: close` [RFC2616 14.10](https://www.rfc-editor.org/rfc/rfc2616#section-14.10)
-               KeepAlive    //< `Connection: Keep-Alive` Obsolete [RFC2068 19.7.1](https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1)
+    STATE_ENUM(
+        ConnectionToken,
+        None, //< No `Connection` header token set
+        Close, //< `Connection: close` [RFC2616 14.10](https://www.rfc-editor.org/rfc/rfc2616#section-14.10)
+        KeepAlive //< `Connection: Keep-Alive` Obsolete [RFC2068 19.7.1](https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1)
     );
 
     /// Describes the header state during parsing.
@@ -464,7 +465,7 @@ public:
 
     bool hasConnectionToken() const { return has(CONNECTION); }
     ConnectionToken getConnectionToken() const {
-        const std::string token = get(CONTENT_LENGTH);
+        const std::string token = get(CONNECTION);
         if( Util::iequal("close", token) ) {
             return ConnectionToken::Close;
         } else if( Util::iequal("keep-alive", token) ) {
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1472,7 +1472,7 @@ private:
                 }
             }
 
-            if (_response->get("Connection", "") == "close")
+            if( _response->header().getConnectionToken() == Header::ConnectionToken::Close )
             {
                 LOG_TRC("Our peer has sent the 'Connection: close' token. Disconnecting.");
                 onDisconnect();
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -31,6 +31,7 @@
 #include <NetUtil.hpp>
 #include <net/Socket.hpp>
 #include <utility>
+#include <algorithm>
 #if ENABLE_SSL
 #include <net/SslSocket.hpp>
 #endif
@@ -356,6 +357,15 @@ public:
     static constexpr int64_t MaxFieldLen = MaxNameLen + MaxValueLen;
     static constexpr int64_t MaxHeaderLen = MaxNumberFields * MaxFieldLen; // ~1.18 MB.
 
+    static constexpr const char* CONNECTION = "Connection";
+
+    /// Describes the `Connection` header token value
+    STATE_ENUM(ConnectionToken,
+               None,        //< No `Connection` header token set
+               Close,       //< `Connection: close` [RFC2616 14.10](https://www.rfc-editor.org/rfc/rfc2616#section-14.10)
+               KeepAlive    //< `Connection: Keep-Alive` Obsolete [RFC2068 19.7.1](https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1)
+    );
+
     /// Describes the header state during parsing.
     STATE_ENUM(State, New,
                Incomplete, //< Haven't reached the end yet.
@@ -364,8 +374,10 @@ public:
                Complete //< Header is complete and valid.
     );
 
-    using Container = std::vector<std::pair<std::string, std::string>>;
-    using ConstIterator = std::vector<std::pair<std::string, std::string>>::const_iterator;
+    using Pair = std::pair<std::string, std::string>;
+    using Container = std::vector<Pair>;
+    using Iterator = std::vector<Pair>::iterator;
+    using ConstIterator = std::vector<Pair>::const_iterator;
 
     ConstIterator begin() const { return _headers.begin(); }
     ConstIterator end() const { return _headers.end(); }
@@ -380,29 +392,37 @@ public:
         _headers.emplace_back(std::move(key), std::move(value));
     }
 
-    /// Set an HTTP header field, replacing an earlier value, if exists.
+    /// Set an HTTP header field, replacing an earlier value, if exists (case insensitive).
     void set(const std::string& key, std::string value)
     {
-        for (auto& pair : _headers)
-        {
-            if (pair.first == key)
-            {
-                pair.second.swap(value);
-                return;
-            }
+        const Iterator e = _headers.end();
+        const Iterator it = std::find_if(_headers.begin(), e,
+                                    [&key](const Pair &pair) -> bool { return Util::iequal(pair.first, key); });
+        if( e != it ) {
+            it->second.swap(value);
+        } else {
+            _headers.emplace_back(key, std::move(value));
         }
-
-        _headers.emplace_back(key, std::move(value));
     }
 
+    // Returns true if the HTTP header field exists (case insensitive)
     bool has(const std::string& key) const
     {
-        for (const auto& pair : _headers)
-        {
-            if (Util::iequal(pair.first, key))
-                return true;
-        }
+        const ConstIterator e = end();
+        return e != std::find_if(begin(), e,
+                            [&key](const Pair &pair) -> bool { return Util::iequal(pair.first, key); });
+    }
 
+    /// Remove the first matching HTTP header field (case insensitive), returning true if found and removed.
+    bool remove(const std::string& key)
+    {
+        const ConstIterator e = end();
+        const ConstIterator it = std::find_if(begin(), e,
+                                    [&key](const Pair &pair) -> bool { return Util::iequal(pair.first, key); });
+        if( e != it ) {
+            _headers.erase(it);
+            return true;
+        }
         return false;
     }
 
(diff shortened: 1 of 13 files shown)


test/UnitSession.cpp
```diff
@@ -232,6 +233,122 @@ UnitBase::TestResult UnitSession::testSlideShow()
...
+    int dlIter = -1;
+    try
+    {
+        std::string documentPath, documentURL;
+        std::string response;
+        helpers::getDocumentPathAndURL("setclientpart.odp", documentPath, documentURL, testname);
+
+        // Reused http wsd session + socket-poll to send commands to server
+        std::shared_ptr<SocketPoll> wsdSocketPoll = std::make_shared<SocketPoll>(testname);
+        wsdSocketPoll->startThread();
```
mmeeks, line 253: We can run the socket logic on the current thread if we set:
    bool runOnClientThread()
on the SocketPoll - which can make the tests simpler to follow.
Use 'SocketPoll::poll()' in a loop to wait for your condition then and avoid threads / races =)

test/UnitSession.cpp
```diff
@@ -232,6 +233,122 @@ UnitBase::TestResult UnitSession::testSlideShow()
...
+            {
+                const std::string path = "/favicon.ico";
+                Poco::Net::HTTPRequest requestICO(Poco::Net::HTTPRequest::HTTP_GET, path);
+                TST_LOG("Favicon requesting from " << path);
+                dlSession->sendRequest(requestICO);
+
+                Poco::Net::HTTPResponse responseICO;
+                std::istream& rs = dlSession->receiveResponse(responseICO);
+                LOK_ASSERT_EQUAL(Poco::Net::HTTPResponse::HTTP_OK /* 200 */, responseICO.getStatus());
+                LOK_ASSERT_EQUAL(std::string("image/vnd.microsoft.icon"), responseICO.getContentType());
```
mmeeks, line 279: We should assert the non-presence of the "Connection: close" header I think.

test/UnitSession.cpp
```diff
@@ -232,6 +233,122 @@ UnitBase::TestResult UnitSession::testSlideShow()
...
+        helpers::getDocumentPathAndURL("setclientpart.odp", documentPath, documentURL, testname);
+
+        // Reused http wsd session + socket-poll to send commands to server
+        std::shared_ptr<SocketPoll> wsdSocketPoll = std::make_shared<SocketPoll>(testname);
+        wsdSocketPoll->startThread();
+        std::shared_ptr<http::WebSocketSession> wsdSession = helpers::loadDocAndGetSession(
+            wsdSocketPoll, Poco::URI(helpers::getTestServerURI()), documentURL, testname);
+
+        // Reused http download-session for document and favicon download from server
+        std::unique_ptr<Poco::Net::HTTPClientSession> dlSession(
```
mmeeks, line 258: Surely we should use Ash's new HTTP code here and elsewhere.

Also we don't need to load a document in order to do these HTTP checks. I'd suggest we don't and instead focus on the various kinds of special files we serve: capabilities, prometheus metrics etc. etc.

mmeeks (commented):
Looks good generally, would be great to ensure that we are not keeping connections open and sending "Connection: close" down them - and of course, we need to have a follow-on that audits all the sites where we sendFileAndShutdown which I expect are many and varied =)

test/UnitSession.cpp
sgothel: `Connection: close` in the http header seems to be for client request only, signaling the server that no 'keep alive' is desired. 
While such a server response doesn't seem to be viable or possible here, we might want to use it in the WSD request handling to keep the connection open (as per this CR/PR) IFF not set. 
Hence we would still close the socket in the WSD if client optionally requests `Connection: close`, I earmark this and check with the WSD ClientRequestDispatcher code.

test/UnitSession.cpp
sgothel: I will add more test cases and expand its use case, i.e. covering the noted `FileServerRequestHandler::handleRequest(..)` and learning about the different served files. Thx!

Probably revisiting the unit test `UnitWOPIFileUrl` if suitable, as I tossed this one out after my initial attempt.
IMHO I will use the WOPI file request as used in `UnitWOPIFileUrl`, but repackage it in this linear UnitSession logic w/o the provided  WopiTestServer code.

sgothel: .. also will add the new HTTP code usage on top of the revised patch as intended earlier, to have the commits more 'on the point'.

test/UnitSession.cpp
sgothel: looking into it.. thx

test/UnitSession.cpp
sgothel: note: this unit tests checks whether the 2 sockets are still open in subsequent requests.

test/UnitSession.cpp
sgothel: yeah, while working on the changeset .. I also realized its better to put the new HTTP code in and enhance the http::Header with get/set methods for `Connection`. All in progress.

net/HttpRequest.hpp
```diff
@@ -441,6 +462,27 @@ class Header
     /// Return true iff Transfer-Encoding is set to chunked (the last entry).
     bool getChunkedTransferEncoding() const { return _chunked; }
 
+    bool hasConnectionToken() const { return has(CONNECTION); }
+    ConnectionToken getConnectionToken() const {
+        const std::string token = get(CONTENT_LENGTH);
```
mmeeks, line 467: Content-Length    = "Content-Length" ":" 1*DIGIT

Surely - we don't expect the content_length token to be anything but a number ? but I didn't read the get() impl.

test/UnitSession.cpp
```diff
@@ -189,27 +193,33 @@ UnitBase::TestResult UnitSession::testSlideShow()
         Poco::URI::encode(documentPath, ":/?", encodedDoc);
         const std::string ignoredSuffix = "%3FWOPISRC=madness"; // cf. iPhone.
         const std::string path = "/cool/" + encodedDoc + "/download/" + downloadId + '/' + ignoredSuffix;
-        std::unique_ptr<Poco::Net::HTTPClientSession> session(
-            helpers::createSession(Poco::URI(helpers::getTestServerURI())));
-        Poco::Net::HTTPRequest requestSVG(Poco::Net::HTTPRequest::HTTP_GET, path);
+
+        TerminatingPoll dlSocketPoller(testname+"-dl");
+        dlSocketPoller.runOnClientThread();
```
mmeeks, line 198: Better =)

mmeeks (commented):
Looks plausible; but I'm going to be out from tomorrow for some time - so Ash or Caolan will need to read/review updates.

net/HttpRequest.hpp
sgothel: ;-) .. reading your comment after my fix. Yeah: CONNECTION it is now. *slap* myself.
https://github.com/CollaboraOnline/online/commit/9998e90717a8a4083cdec3b82c9cc6d95e57634d

sgothel:
Last 2 commits reflect input given from @Ashod . 

TODO: The actual FileServer handling `FileServerRequestHandler::handleRequest(..)` 
.. coming up. Should be the last piece for this pull request.

sgothel:
rebased to master 42e9dac882d0558307d49c8d739726ab5691ea3a

sgothel:
OK, last commits done, rebased and forced pushed .. ready. @caolanm @Ashod @mmeeks

net/HttpRequest.hpp
```diff
@@ -441,6 +464,30 @@ class Header
     /// Return true iff Transfer-Encoding is set to chunked (the last entry).
     bool getChunkedTransferEncoding() const { return _chunked; }
 
+    bool hasConnectionToken() const { return has(CONNECTION); }
+    ConnectionToken getConnectionToken() const {
+        const std::string token = get(CONNECTION);
+        if( Util::iequal("close", token) ) {
+            return ConnectionToken::Close;
+        } else if( Util::iequal("keep-alive", token) ) {
```
Ashod, line 472: Ditto as above. No need for all these `else if` after returning. Just have these as `if/return` cases, without all the `else`s. More readable as the structure of the code reflects the intent and functionality.

net/HttpRequest.hpp
```diff
@@ -441,6 +464,30 @@ class Header
...
+            return ConnectionToken::KeepAlive;
+        } else if( Util::iequal("upgrade", token) ) {
+            return ConnectionToken::Upgrade;
+        } else {
+            return ConnectionToken::None;
+        }
+    }
+    void setConnectionToken(ConnectionToken token) {
```
Ashod, line 480: Nit: blank line between clauses and functions.

net/HttpRequest.hpp
```diff
@@ -441,6 +464,30 @@ class Header
...
+        std::string value;
+        switch( token ) {
+            case ConnectionToken::Close: value = "close"; break;
```
Ashod, line 483: Nit: More readable to break these on multiple lines. The reader's time and attention are more valuable resources than a line or two on the screen (let alone in the file). (Don't force me to parse 3 statements on one line, please.)

net/HttpRequest.hpp
```diff
@@ -1428,7 +1476,7 @@ class Session final : public ProtocolHandlerInterface
                 }
             }
 
-            if (_response->get("Connection", "") == "close")
+            if( _response->header().getConnectionToken() == Header::ConnectionToken::Close )
```
Ashod, line 1479: Somehow the space between the `if` and the paren went away. `if`, `for`, `while`, etc. are not functions. It's customary to leave a space to differentiate between them and function calls.

(Again, making it easier for maintainers to read and parse code quickly and will lower chance of confusion/misreading is a good thing.)

Ashod (commented):
Thanks, @sgothel. Good stuff. I left a few more comments, but most of my feedback was communicated offline, so not much new here.

One thing I wonder is how long the browser keeps this connection open. Could it be indefinite (as long as the tab is open) ? I think we should check with different browsers, so we can timeout and close them if they end up lingering for too long. We don't serve http files after loading the document, but the browser doesn't know that.

Of course CI also needs to be green, but I didn't check why there are failures. Let me know if need help looking into the reasons.

wsd/ClientRequestDispatcher.cpp
```diff
@@ -763,8 +763,6 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
 
                     launchAsyncCheckFileInfo(_id, accessDetails, RequestVettingStations);
                 }
-
-                socket->shutdown();
```
caolanm, line 767: cypress tests pass/fail with this shutdown present/omitted

caolanm (changes requested):
try the links from the "make run" prompt and they have the same issue I believe

wsd/ClientRequestDispatcher.cpp
caolanm: using:

if (request.getMethod() == Poco::Net::HTTPRequest::HTTP_POST)
    socket->shutdown();

here I think worksaround the problem, which I think is that ClientRequestDispatcher::handleIncomingMessage is using socket->eraseFirstInputBytes(map); to remove consumed data, but in the post we have body content that isn't consumed by that and gets left behind

net/Socket.cpp
```diff
@@ -1303,13 +1306,18 @@ bool StreamSocket::parseHeader(const char *clientName,
 {
     assert(map._headerSize == 0 && map._messageSize == 0);
 
+    constexpr std::chrono::duration<float, std::milli> delayMax =
```
Ashod, line 1309: Why convert micros into fractional millis? Seems having duration values as microsecond integers is more straightforward than this conversion.

net/Socket.cpp
```diff
@@ -1303,13 +1306,18 @@ bool StreamSocket::parseHeader(const char *clientName,
...
+
     // Find the end of the header, if any.
     static const std::string marker("\r\n\r\n");
-    auto itBody = std::search(_inBuffer.begin(), _inBuffer.end(),
-                              marker.begin(), marker.end());
+    auto itBody = std::search(_inBuffer.begin(), _inBuffer.end(), marker.begin(), marker.end());
     if (itBody == _inBuffer.end())
     {
-        LOG_TRC(clientName << " doesn't have enough data for the header yet.");
+        LOG_TRC("parseHeader: " << clientName << " doesn't have enough data for the header yet. delay " << delayMs.count() << "ms");
```
Ashod, line 1320: If `delayMs` is in a standard unit (i.e. `chrono::millseconds`, `chrono::microseconds`, etc.), you don't need the `.count()` nor the `<< "ms"`, instead, just stream `delayMs` and you get the units too! (Yet another reason to avoid floating-point duration).

net/Socket.cpp
```diff
@@ -1405,39 +1414,83 @@ bool StreamSocket::parseHeader(const char *clientName,
...
 
                 itBody+=2;
                 chunk++;
             }
-            LOG_TRC("Not enough chunks yet, so far " << chunk << " chunks of total length " << (itBody - _inBuffer.begin()));
+            LOG_TRC("parseHeader: Not enough chunks yet, so far " << chunk << " chunks of total length " << (itBody - _inBuffer.begin()) << ", delay " << delayMs.count() << "ms");
             return false;
         }
     }
+    catch (const Poco::Net::NotAuthenticatedException& exc)
```
Ashod, line 1434: Are these exceptions coming from the form parser? I'm rather surprised, if so.

net/Socket.cpp
```diff
@@ -1303,13 +1306,18 @@ bool StreamSocket::parseHeader(const char *clientName,
 {
     assert(map._headerSize == 0 && map._messageSize == 0);
 
+    constexpr std::chrono::duration<float, std::milli> delayMax =
+        std::chrono::duration_cast<std::chrono::milliseconds>(SocketPoll::DefaultPollTimeoutMicroS);
```
Ashod, line 1310: I expect we'd want a separate value for the `delayMax`, other than `DefaultPollTimeoutMicroS`, as I don't see how a socket-specific timeout would be related to our poll-timeout, which can be arbitrarily small or long (indeed, it can even be infinite!).

wsd/ClientRequestDispatcher.cpp
```diff
@@ -763,6 +764,10 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
 
                     launchAsyncCheckFileInfo(_id, accessDetails, RequestVettingStations);
                 }
+                if (request.getMethod() == Poco::Net::HTTPRequest::HTTP_POST)
```
Ashod, line 767: Would be good to have a comment explaining what's going on here. This is completely non-intuitive and we spent quite a bit of time to figure this workaround out. I know I'll likely forget what this is doing in a few short weeks.

Ashod (approved):
Thanks @sgothel. Left some comments and suggestions, but they can be addressed in a follow up. I think this is good to go in, if @caolanm is happy.

caolanm (approved):

caolanm:
yeah, lets do follow ups separately and get this in for maximum testing exposure

PR #9778, opened 2024-08-12
play, pause the video during slideshow, transition, unmute sound

quikee:
During slideshow the video could still continue playing if the slide with a video was the last one.

Also stop rendering when all the videos are paused.

Additionally unmute the video, so we have the sound from the video during the slideshow.

Change-Id: I32f3f405b0c92e446f84ba4f4f3f60f9dd294eb1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks ok

PR #9779, opened 2024-08-12
build(android): Simplify configure abi selection

Minion3665:
Previously it was quite cumbersome to specify a different architecture than armeabi-v7a, as you had to specify *both* a core in the armeabi section and one for whatever architecture you wanted to build (these could be the same core, the armeabi one didn't need to actually be armeabi...)

This was made more confusing by the requirement to specify cores in the right position - requiring you to put colons between cores even when you were only planning to build a single ABI, but only if that ABI wasn't armeabi-v7a...

For example, to build arm64-v8a you would have to do something like

--with-lo-builddir=$(realpath ../android-core):$(realpath ../android-core)::

This is now changed, so specifying a single core is equivalent to specifying it 4 times, i.e. these are both the same thing

--with-lo-builddir=$(realpath ../android-core)

or

--with-lo-builddir=$(realpath ../android-core):$(realpath ../android-core):$(realpath ../android-core):$(realpath ../android-core)

This is a lot neater, and should allow us to make some nicer instructions on our build page
(https://collaboraonline.github.io/post/build-code-android/) which don't care about the architecture you're building for. This is made all the more important as the vast majority of modern devices are arm64-v8a rather than armeabi-v7a

---

We may additionally want to make a followup so that...

--with-lo-builddir=$(realpath ../android-core):$(realpath ../android-core)::

...is equivalent to...

--with-lo-builddir=$(realpath ../android-core):$(realpath ../android-core)::

...but this is less urgent

Change-Id: I875451423cb773df5198074587fe4d4365b30347

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,8 +213,9 @@ AC_ARG_WITH(android-abi,
                             by colons).  The supported ABIs are: armeabi-v7a, arm64-v8a, x86 and x86_64.
                             Please note that you need to specify the parameters for --with-lo-builddir,
                             --with-poco-includes and --with-poco-libs in the order of armeabi-v7a:arm64-v8a:x86:x86_64.
-                            For example, when you use --with-android-abi=x86_64,
-                            you have to specify --with-lo-builddir=:::/path/to/x86-64/builddir]),
+                            For example, when you use --with-android-abi="x86_64 x86",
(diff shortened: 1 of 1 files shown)


Rash419 (approved):
Looks good to me, thanks

PR #9781, opened 2024-08-12
chore: discourage .beforeprettier changes

Minion3665:
When we added prettier, we didn't want to format all the files in the repository then and there, so the compromise we came to was the .beforeprettier file

.beforeprettier is there to ignore files that were there "before prettier", i.e. legacy files that do not need to be formatted

Unfortunately, we didn't make this intent as clear as we could have, and the file has been changed several times since (see https://github.com/CollaboraOnline/online/commits/master/browser/.beforeprettier), including during migrations of files from JavaScript to Typescript and additions of new files, both of which should result in the file being formatted.

This change adds a comment to the top of the file, explaining that people shouldn't add new content to it. It also makes a CODEOWNERS file to ping myself and @eszkadev for review in the case that someone does try to change this file, allowing us to more easily spot changes that do so.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,3 @@
+# Automatically mention people if .beforeprettier is being changed, as this
+# should be rare but sometimes people mistakenly add new stuff...
+/browser/.beforeprettier    @minion3665 @eszkadev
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
looks good

PR #9782, opened 2024-08-12
Calc: update geometry and position on new sheet insertion

eszkadev:
calc: use hashes for parts
    
part number is not enough when adding/deleting/moving the sheets

fixes regression from commit 703dae0ea2322cee3f7af3d293f6ab7307f4d91a

1. Open Calc
2. in the current sheet scroll down a but and make some rows bigger
3. right-click on the sheet tab, insert sheet before

Result: resized rows are still visible
Expected: new sheet has equal small rows

---------------------------------------

impress: move part hashes to the docstate

// unify code with calc

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -1511,7 +1511,7 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 	private adjustCommentFileBasedView (comment: any): void {
 		// Below calculations are the same with the ones we do while drawing tiles in fileBasedView.
 		var partHeightTwips = this.sectionProperties.docLayer._partHeightTwips + this.sectionProperties.docLayer._spaceBetweenParts;
-		var index = this.sectionProperties.docLayer._partHashes.indexOf(String(comment.parthash));
+		var index = app.impress.partHashes.indexOf(String(comment.parthash));
 		var yAddition = index * partHeightTwips;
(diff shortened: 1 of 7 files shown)


gokaysatir (changes requested):
When there is only one sheet and if user adds a sheet before that one, grid is not updated.

* Open a spreadsheet with one sheet-tab.
* Resize a column
* Add a sheet before that one.
* The resized column is preserved in the new sheet.

Can you have a look?

eszkadev:
Sorry, I didn't comment. If tested together with https://github.com/CollaboraOnline/online/pull/9784 then it works.
Now rebased.

thebearon:
https://github.com/CollaboraOnline/online/pull/9784 has been reverted via #9793.

eszkadev:
Sorry, I didn't comment. If tested together with https://github.com/CollaboraOnline/online/pull/9784 then it works.
Now rebased.

> When there is only one sheet and if user adds a sheet before that one, grid is not updated.
> 
>     * Open a spreadsheet with one sheet-tab.
> 
>     * Resize a column
> 
>     * Add a sheet before that one.
> 
>     * The resized column is preserved in the new sheet.
> 
> 
> Can you have a look?

this issue is fixed now, PR ready for review

gokaysatir (approved):

thebearon:
> Sorry, I didn't comment. If tested together with #9784 then it works. Now rebased.

Let me note that this was reverted via #9793, and reapplied via #9853.

PR #9786, opened 2024-08-12
use more async dns

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/AsyncDNS.hpp
+++ b/net/AsyncDNS.hpp
@@ -20,10 +20,13 @@
 #include <string>
 #include <thread>
 
+#include <NetUtil.hpp>
+
 namespace net
 {
 
 class DNSResolver;
+class HostEntry;
 
 class AsyncDNS
 {
@@ -36,11 +39,11 @@ public:
 
     static void dumpState(std::ostream& os);
 
-    typedef std::function<void(const std::string& hostName, const std::string& exception)> DNSThreadFn;
+    typedef std::function<void(const HostEntry& hostEntry)> DNSThreadFn;
     typedef std::function<std::string()> DNSThreadDumpStateFn;
 
-    static void canonicalHostName(const std::string& addressToCheck, const DNSThreadFn& cb,
-                                  const DNSThreadDumpStateFn& dumpState);
+    static void lookup(const std::string& searchEntry, const DNSThreadFn& cb,
+                       const DNSThreadDumpStateFn& dumpState);
 
 private:
     std::atomic<bool> _exit;
@@ -53,7 +56,6 @@ private:
         std::string query;
         AsyncDNS::DNSThreadFn cb;
         AsyncDNS::DNSThreadDumpStateFn dumpState;
-        // for now just canonicalHostName lookups
     };
     std::queue<Lookup> _lookups;
     Lookup _activeLookup;
--- a/net/AsyncDNS.hpp
+++ b/net/AsyncDNS.hpp
@@ -16,6 +16,7 @@
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <queue>
 #include <string>
 #include <thread>
@@ -42,7 +43,9 @@ public:
     typedef std::function<void(const HostEntry& hostEntry)> DNSThreadFn;
     typedef std::function<std::string()> DNSThreadDumpStateFn;
 
-    static void lookup(const std::string& searchEntry, const DNSThreadFn& cb,
+    static void lookup(const std::string& searchEntry,
+                       const std::optional<std::string>& port,
+                       const DNSThreadFn& cb,
                        const DNSThreadDumpStateFn& dumpState);
 
 private:
@@ -54,6 +57,7 @@ private:
     struct Lookup
     {
         std::string query;
+        std::optional<std::string> port;
         AsyncDNS::DNSThreadFn cb;
         AsyncDNS::DNSThreadDumpStateFn dumpState;
     };
@@ -61,7 +65,9 @@ private:
     Lookup _activeLookup;
 
     void resolveDNS();
-    void addLookup(const std::string& lookup, const DNSThreadFn& cb,
+    void addLookup(const std::string& lookup,
+                   const std::optional<std::string>& port,
+                   const DNSThreadFn& cb,
                    const DNSThreadDumpStateFn& dumpState);
 
     void startThread();
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1591,33 +1591,38 @@ private:
         return socket; // Return the shared pointer.
     }
 
+    void asyncConnectCompleted(SocketPoll& poll, std::shared_ptr<StreamSocket> socket)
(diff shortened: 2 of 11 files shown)


caolanm:
The idea here is link Session::asyncRequest with the async dns to do the getaddrinfo and initial connect in the async-dns thread.

And replace the Poco-usage in the async dns with direct getaddrinfo (not poco-like throwing any exceptions on failure, just hold onto the first failure and make it available for reporting at the consumer)

net/AsyncDNS.hpp
```diff
@@ -42,7 +43,9 @@ class AsyncDNS
     typedef std::function<void(const HostEntry& hostEntry)> DNSThreadFn;
     typedef std::function<std::string()> DNSThreadDumpStateFn;
 
-    static void lookup(const std::string& searchEntry, const DNSThreadFn& cb,
+    static void lookup(const std::string& searchEntry,
+                       const std::optional<std::string>& port,
```
Ashod, line 47: Wouldn't an empty `port` imply it's missing? That is, an empty string is already a natural sentry, instead of `std::optional`.

Edit: It's unclear from getaddrinfo(3) what happens with a non-null `service` that is empty (for `AI_CANONNAME | AI_ADDRCONFIG`). It seams it wouldn't care for those cases, but it still seems cleaner to _not_ pass in empty strings. So avoiding `optional` and checking only for empty string to pass the pointer or nullptr seems most reasonable here.

net/NetUtil.cpp
```diff
@@ -155,15 +158,16 @@ static HostEntry resolveDNS(const std::string& addressToCheck, std::vector<DNSCa
...
+                                 return entry.queryAddress == addressToCheck &&
+                                        entry.queryPort == port;
                                });
     if (findIt != querycache.end())
         return findIt->hostEntry;
 
     // lookup and cache
-    HostEntry hostEntry(addressToCheck);
-    querycache.push_back(DNSCacheEntry{addressToCheck, hostEntry, now});
+    HostEntry hostEntry(addressToCheck, port ? port->c_str() : nullptr);
```
Ashod, line 169: This would become `!port.empty() ? port.c_str() : nullptr`. No need for optional.

net/HttpRequest.hpp
```diff
@@ -1168,6 +1168,11 @@ class Session final : public ProtocolHandlerInterface
     /// regardless of the reason (error, timeout, completion).
     void setFinishedHandler(FinishedCallback onFinished) { _onFinished = std::move(onFinished); }
 
+    /// The onConnectFail callback handler signature.
+    using ConnectFailCallback = std::function<void()>;
```
Ashod, line 1172: Wouldn't it be helpful to pass in the `http::Session` in question? Can be captured in a lambda, but seems cleaner that we pass it as context. (C.f. `FinishedCallback` on line 1164.)

test/HttpRequestTests.cpp
```diff
@@ -307,7 +310,11 @@ void HttpRequestTests::testSimpleGet()
 
         std::unique_lock<std::mutex> lock(mutex);
 
-        LOK_ASSERT(httpSession->asyncRequest(httpRequest, pollThread));
+        httpSession->setConnectFailHandler([this]() {
+            LOK_ASSERT(false);
```
Ashod, line 314: `LOK_ASSERT_FAIL("Unexpected connection failure");` might be more helpful (with a similar message).

net/HttpRequest.hpp
```diff
@@ -1621,11 +1621,13 @@ class Session final : public ProtocolHandlerInterface
     {
         _socket.reset(); // Reset to make sure we are disconnected.
 
-        auto callback = [this, &poll](std::shared_ptr<StreamSocket> socket) {
-            asyncConnectCompleted(poll, socket);
+        auto pushConnectCompleteToPoll = [this, &poll](std::shared_ptr<StreamSocket> socket) {
+            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket)]() {
```
Ashod, line 1625: I assume this is because we don't want races between the poll thread and the asyncDNS thread. A comment would be a helpful reminder in that case, I think.

Nit: Also, since we already have `selfLifecycle`, it might be simpler to use that instead of capturing (and using) `this`.

Ashod (approved):
Thanks, @caolanm. Left some comments, but nothing that can't be done in a follow-up. Looks good.

net/NetUtil.cpp
```diff
@@ -34,14 +34,116 @@ namespace net
...
+
+HostEntry::HostEntry(const std::string& desc, const char* port)
+    : _requestName(desc)
+    , _errno(0)
+    , _eaino(0)
+{
+    initFromHostName(desc, port);
+}
+
+HostEntry::~HostEntry()
```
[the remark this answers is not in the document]
caolanm: yeah, left over from an earlier iteration which was more baroque and had a dtor body

net/NetUtil.cpp
```diff
@@ -34,14 +34,116 @@ namespace net
...
+    if (!result)
+    {
+        _errno = errno;
+        LOG_WRN("inet_ntop failure: " << errorMessage());
+        return std::string();
+    }
+    return std::string(result);
+}
+
+void HostEntry::initFromHostName(const std::string& host, const char* port)
```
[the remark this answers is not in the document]
caolanm: yes, will merge

net/AsyncDNS.hpp
caolanm: yeah, fine by me

net/NetUtil.cpp
caolanm: sure

test/HttpRequestTests.cpp
caolanm: will do

net/HttpRequest.hpp
caolanm: wrt selfLifeCycle, it's a std::shared_ptr<ProtocolHandlerInterface> which doesn't have asyncConnectCompleted, so it turns into something like

std::static_pointer_cast<Session>(selfLifecycle)->asyncConnectCompleted(poll, socket);

which a little unfriendly to read, so I'm inclined towards the suggestion of https://devblogs.microsoft.com/oldnewthing/20190104-00/?p=100635 to "capture both a strong reference and a raw this. The strong reference keeps the this alive, and you use the this for convenient access to members."

net/NetUtil.cpp
```diff
@@ -34,14 +34,116 @@ namespace net
...
+            std::string address = makeIPAddress(ai->ai_addr);
+            if (!good())
+                break;
+            _ipAddresses.push_back(address);
+        }
+        return;
+    }
+
+    setEAI(rc);
+    LOG_SYS("Failed to lookup host " << errorMessage());
```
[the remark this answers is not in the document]
caolanm: sure, I'll reshuffle this

PR #9790, opened 2024-08-13
duplicate check for failed fd

caolanm:
there are two back to back "if (fd < 0)" checks here

Change-Id: I4eb39673ec6dca1d3678228ba21f06579aee4961

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -325,7 +325,7 @@ connect(const std::string& host, const std::string& port, const bool isSSL,
                 }
 
                 int res = ::connect(fd, ai->ai_addr, ai->ai_addrlen);
-                if (fd < 0 || (res < 0 && errno != EINPROGRESS))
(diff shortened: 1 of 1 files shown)


net/NetUtil.cpp
```diff
@@ -325,7 +325,7 @@ connect(const std::string& host, const std::string& port, const bool isSSL,
                 }
 
                 int res = ::connect(fd, ai->ai_addr, ai->ai_addrlen);
-                if (fd < 0 || (res < 0 && errno != EINPROGRESS))
+                if (res < 0 && errno != EINPROGRESS)
                 {
```
sgothel, line 329: 1st if(fd<0) test is a few line above. Looks good!

sgothel (approved):
Looks good

Issue #9727, opened 2024-08-07
UI defaults not applied


juliushaertl:
### Describe the Bug

With the current code build and master branch settings handed over by integrators through ui_defaults are not applied.

One example is the saveAsMode that we send over in richdocuments

### Steps to Reproduce

1. Send a ui_default setting along the request to browser.html

### Expected Behavior

The setting is picked up

### Actual Behavior

It is not working

## Versions
productVersion: 24.04.6.1
productVersionHash: 8ec5fda

### Additional Context

Digging into it a bit this seems to have been caused by https://github.com/CollaboraOnline/online/pull/9442 where just inserting the raw json into the data-ui-defaults attribute results in the first `"`from the json cosing the attribute:

https://github.com/CollaboraOnline/online/blob/07f903660df88e3e03dcae72b56577fae70efd5e/browser/html/cool.html.m4#L250
![image](https://github.com/user-attachments/assets/1e17a20d-f56f-4250-80de-b489cf6a0b1b)
![image](https://github.com/user-attachments/assets/ba4e23b8-f191-4407-924e-6398f702ab7f)

gokaysatir:
@juliushaertl hello, thank you again for the heads up. I prepared a fix. Can you have a look whenever you can have time? I also tested and it seems to work normal locally.

gokaysatir:
https://github.com/CollaboraOnline/online/issues/9727

gokaysatir:
I think we can close this one now.

PR #9791, opened 2024-08-13
Use base64 string for UI_DEFAULTS variable. 

gokaysatir:
It is assigned JSON object which uses quotes.
Also parse the variable at initialization.

Change-Id: Ie87869bdf2d6e278144c0cb1a0b5519a645fca9c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -399,7 +399,7 @@ class BrowserInitializer extends InitializerBase {
 		window.enableDebug = element.dataset.enableDebug.toLowerCase().trim() === "true";
 		window.frameAncestors = decodeURIComponent(element.dataset.frameAncestors);
 		window.socketProxy = element.dataset.socketProxy.toLowerCase().trim() === "true";
-		window.uiDefaults = element.dataset.uiDefaults;
+		window.uiDefaults = JSON.parse(atob(element.dataset.uiDefaults));
 		window.checkFileInfoOverride = element.dataset.checkFileInfoOverride;
 		window.deeplEnabled = element.dataset.deeplEnabled.toLowerCase().trim() === "true";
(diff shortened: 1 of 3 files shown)


gokaysatir:
Fixes: https://github.com/CollaboraOnline/online/issues/9727

browser/html/cool.html.m4
```diff
@@ -247,7 +247,7 @@ m4_ifelse(MOBILEAPP, [true],
       data-enable-debug = "%ENABLE_DEBUG%"
       data-frame-ancestors = "%FRAME_ANCESTORS%"
       data-socket-proxy = "%SOCKET_PROXY%"
-      data-ui-defaults = "%UI_DEFAULTS%"
+      data-ui-defaults = '%UI_DEFAULTS%'
```
juliushaertl, line 250: I guess this does the trick, but a safer way could be to use base64 encoding of the json string and decode that in the global.js then which could help prevent future issues in case the UI defaults json contains any single quote

gokaysatir: That's right, though it's hard to guess if we need single quote in UI defaults where almost everything is either true, false or one word.

gokaysatir: Let me update and thanks for the comment :)

gokaysatir:
Now we have safer string :)

gokaysatir:
@Ashod can you comment on the change in FileServer.cpp when you have time? We use JSON for UI_DEFAULTS and wanted to be safe against single and double quotes.

eszkadev (approved):

Ashod:
Thanks @gokaysatir. Looks good.

PR #9799, opened 2024-08-13
slideshow rework: bug: some transitions are wrong

mcecchetti:
pushWipe transition subtypes are routed to the wrong gl transition
horizontal and vertical comb are inverted
fade transition subtyped are routed to the wrong gl transition (index
start from 1 instead of 0)

Change-Id: Ibe1e63a2678af1e5c298272f63f188234e10bbb5


mcecchetti:
wrong commit

PR #9800, opened 2024-08-13
slideshow rework: bug: some transitions are wrong

mcecchetti:
pushWipe transition subtypes are routed to the wrong gl transition
horizontal and vertical comb are inverted
fade transition subtyped are routed to the wrong gl transition (index
start from 1 instead of 0)

Change-Id: I516fafc1dcb39120ecec96df8d6834a13840cb3e

### DIFF
--- a/browser/src/slideshow/LayersCompositor.ts
+++ b/browser/src/slideshow/LayersCompositor.ts
@@ -15,8 +15,7 @@
  */
 
 declare var app: any;
-// declare var SlideShow: any;
-declare namespace SlideShow {}
+declare var SlideShow: any;
 
 class LayersCompositor extends SlideShow.SlideCompositor {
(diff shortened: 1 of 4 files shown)


codewithvk (approved):
Thanks!

codewithvk:
@mcecchetti There is one more issue i found here for fade transition, Minor.
Screenshot from 2024-08-13 23-07-28
Screenshot from 2024-08-13 23-07-46
Here for first slide I have applied fade through black and for second white through white but both have  transitionDirection same, I think we have to change it from core, So as of now both show through white transition, Would nice if you can fix this as I'm working on something else, thanks!

mcecchetti:
I don't think it's related to the direction, usually it's used in clipping transitions. Anyway, thanks for reporting, I will analyze the code more deeply.

Issue #9801, opened 2024-08-13
ChangeBezier command not properly labelled in Draw/Impress notebook bar


hfiguiere:
### Describe the Bug
ChangeBezier command not properly labelled in Draw/Impress notebook bar

![image](https://github.com/user-attachments/assets/587967e2-c0dc-45e3-b5ac-5dee2565529b)

Also the contextual menu is incorrect in Draw (but ok in Impress)

### Steps to Reproduce
1. Opem Draw
2. Select a shape
3. 

### Expected Behavior
The button has the proper name

![image](https://github.com/user-attachments/assets/629f50cb-254f-4d3a-8fdf-ff98cb286107)

### Actual Behavior
It is called "ChangeBezier" like the UNO command

### Desktop
(Please complete the following information)
 - **Collabora version:** master
 - **Browser and version:** Firefox

### Additional Context
This regressed from 4f0b165b2cbc5c8702f0ee6efbab152526ec1cb3
Turns out the `ChangeBezier` command should be `global` but the update script doesn't, and set it to `presentation`.

hfiguiere:
Setting the `ChangeBezier` to global fix the label

```diff
diff --git a/browser/src/unocommands.js b/browser/src/unocommands.js
index d59cf08ca2..7a1e3641b8 100644
--- a/browser/src/unocommands.js
+++ b/browser/src/unocommands.js
@@ -33,7 +33,7 @@ var unoCommandsArray = {
 	'CellVertCenter':{global:{context:_('Center Vertically'),menu:_('Center'),},},
 	'CellVertTop':{global:{context:_('Align Top'),menu:_('Top'),},},
 	'CenterPara':{global:{context:_('Align Center'),menu:_('Center'),},},
-	'ChangeBezier':{presentation:{menu:_('To ~Curve'),},},
+	'ChangeBezier':{global:{menu:_('To ~Curve'),},},
 	'ChangeCaseRotateCase':{global:{menu:_('Cycle Case'),},},
 	'ChangeCaseToLower':{global:{menu:_('~lowercase'),},},
 	'ChangeCaseToSentenceCase':{global:{menu:_('~Sentence case'),},},
```

which is the value it was at originally.

hfiguiere:
PR #9802 fixes the button. Not the context menu in Draw.

hfiguiere:
The following patch in unocommands solve the context menu problem.
However, it is overridden wtih `./scripts/unocommands.py --update`.

```diff
[...]

PR #9802, opened 2024-08-13
impress/draw: Fix ChangeBezier button label

hfiguiere:
Fixes https://github.com/CollaboraOnline/online/issues/9801

Change-Id: Ic1bc7f9ac55a6da7d19e0bd8e85ed67386bb8bae

* Resolves: #9801 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
Given the uno:ChangeBezier in "presentation" only this make sense.  I see uno:ChangeBezier in core is in DrawImpressCommands.xcu and not GenericCommands.xcu, but I see that only sd/sdi/sdraw.sdi then has "SfxVoidItem ChangeBezier SID_CHANGEBEZIER" in it. So I guess this only exists in draw/impress then.

PR #9803, opened 2024-08-13
annotation: avoid creating comment insert box if annotations disabled

lpranam:
problem:
some elements can not have comments, when these elements get selection, comment button in menus were disabled but shortcuts still worked, this ensures no element for comment insert is created in first place

Change-Id: Ib76366c23e9ad7f6fc29f98add65b0840d00ead9

* Target version: master 

### Checklist


caolanm (approved):
looks reasonable to me

Issue #9616, opened 2024-07-25
perf: invalidation overlay handling is horribly slow


mmeeks:
Ctrl-Alt-Shift-D - and enable "Tile Invalidations" - and when you have lots of users using calc - and lots of invalidations (10's per second) of multiple users typing, we end up with a totally bogged down firefox, and an unusable experience.

Also - not seeing any invalidations rendered ;-)

Firefox profile is here:

https://share.firefox.dev/3Wh3OiB

It shows that all the time is spent in adding and removing canvas sections 

![image](https://github.com/user-attachments/assets/edfba7e9-728c-4d08-b8c1-b0ec918e4fcf)

I would imagine that having a single section with a list of things to render might be more helpful.

Then again - it seems deeply concerning that sections render as they are adjusted. I would expect all rendering to occur in a single pass at the end of the slurped message emission:

			if (this._map._docLayer) {
				// Resume with redraw if dirty due to previous _onMessage() calls.
				this._map._docLayer.resumeDrawing(true);
			}
			// Let other layers / overlays catch up.
			this._map.fire('messagesdone');

Would suggest that we should not be doing drawing here anyway - surely (!?) at least until resumeDrawing is called or 'messagesdone' is emitted.

What's up there ? =)

gokaysatir:
https://github.com/CollaboraOnline/online/pull/9809

PR #9809, opened 2024-08-14
Private/gokay/tile invalidations

gokaysatir:
Fixes: https://github.com/CollaboraOnline/online/issues/9616

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/.beforeprettier
+++ b/browser/.beforeprettier
@@ -139,7 +139,6 @@
 /src/canvas/sections/AutoFillMarkerSection.ts
 /src/canvas/sections/HTMLObjectSection.ts
 /src/canvas/sections/URLPopUpSection.ts
-/src/canvas/sections/InvalidationRectangleSection.ts
 /src/canvas/sections/ShapeHandlesSection.ts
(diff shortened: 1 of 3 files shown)


gokaysatir:
Fixes: https://github.com/CollaboraOnline/online/issues/9616

mmeeks (commented):
Looks better; I would somehow have assumed we have one section, and an an elapsed time on each rectangle (?) =) Anyhow - over to Szymon - looks low risk to me - but good to get in for Red I guess.

gokaysatir:
> Looks better; I would somehow have assumed we have one section, and an an elapsed time on each rectangle (?) =) Anyhow - over to Szymon - looks low risk to me - but good to get in for Red I guess.

We have elapsed time but we also limit the maximum number of rectangles. So it also shortens the timeout of rectangles. We can increase or entirely remove the limit of rectangles of course.

There is one section, but if there is no rectangles left, it deletes itself. So we don't bother with deleting it.

Thanks for having a look.

eszkadev (approved):

PR #9811, opened 2024-08-14
Meven/button notebookbar pointer

meven:
* Target version: master 

### Summary
Currently only the icon of the notebookbar buttons that have the pointer (hand) cursor, but neither the hovered region nor the button label.

This allows to have the pointer cursor on the whole notebookbar buttons.

image

improves upon https://github.com/CollaboraOnline/online/commit/41063ceafbd6234ae180c55ea8f501de51f5ab2f

### Checklist

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -197,7 +197,6 @@ html[data-theme='dark'] .savemodified.unotoolbutton .unobutton img {
 }
 
 .unotoolbutton.notebookbar .unobutton {
-	cursor: pointer;
 	box-sizing: border-box;
 	padding: 0;
 	width: var(--btn-size);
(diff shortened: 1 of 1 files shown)


pedropintosilva (approved):

pedropintosilva:
looks good, thanks. And I'm assuming hovering the `unotoolbutton.notebookbar .unobutton` will still show the `cursor: pointer;` right?

meven:
> looks good, thanks. And I'm assuming hovering the `unotoolbutton.notebookbar .unobutton` will still show the `cursor: pointer;` right?

Yes it will, in dom, the .unobutton is a child of 

Also I found this a branded css file, that will may become obsolete 
```
.unotoolbutton:hover * {
  cursor: pointer;
}
```

Testing the compact mode, I think we should generalize it to `.unotoolbutton` and not just `.notebookbar` as it is done in the branded css.

meven:
This one had a regression, the button itself does not have cursor anymore (buttons need to have explicit cursor set).

See https://github.com/CollaboraOnline/online/pull/9827 for a fix.

Issue #9813, opened 2024-08-14
ChangeBezier context menu label incorrect in Draw


hfiguiere:
### Describe the Bug
ChangeBezier context menu label incorrect in Draw

### Steps to Reproduce
1. Open Draw
2. Select a shape
3. Right click

### Expected Behavior
It shows Convert > To Curve

### Actual Behavior
It shows Convert > ChangeBezier

### Screenshots
![image](https://github.com/user-attachments/assets/97d050e2-60b7-4543-bf47-f34c46fe642b)

### Desktop
(Please complete the following information)
 - **Collabora version:** master
 - **Browser and version:** Firefox

### Additional Context

Adjacent to #9801. Albeit the same cause.

hfiguiere:
The reason is that the UNO command is defiend for `presentation` not for `drawing`.

PR #9814, opened 2024-08-14
draw: just get the UNO commands for presentation

hfiguiere:
Now the ChangeBezier uno command has a proper context menu item in draw

Fixes https://github.com/CollaboraOnline/online/issues/9813

Change-Id: I834c19a80e6d8c6a279ae298e78ba650bbea1fb1

* Resolves: #9813 
* Target version: master 

### Summary

Turns out presentation and drawing have the same set of commands, but we distinguish on the doc type.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/unocommands.js
+++ b/browser/src/unocommands.js
@@ -532,6 +532,9 @@ window._UNO = function(string, component, isContext) {
 	if (entry === undefined) {
 		return command;
 	}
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks ok, we don't have special field for draw in `unocommands.js` so reusing presentation one is fine

PR #9824, opened 2024-08-15
fix(android): don't start drag on unmoving tap

Minion3665:
### fix(android): don't start drag on unmoving tap

Previously in impress we would always start a drag no matter how little
a tap had moved. This commonly inhibited us from tapping into a text
field to edit it, particularly when we didn't have something else
selected first.

I wondered for a little while about detecting whether the tap brought up
the keyboard, but it seems that by the time we know that it's already
too late. This tap distance detection seems reliable while also being
very simple to implement, and a drag of less than 0.1 would be an
unrealistically small drag anyway.

---

Additionally, 

### fix(cursor): hide on cursorvisible: false

Previously when we got a `cursorvisible: false` message from COOLWSD, we
would set `app.file.textCursor.visible` to `false` and then attempt to
update the cursor.

Unfortunately, updating the cursor was disabled when
`app.file.textCursor.visible` was false, leading to the cursor not being
hidden. This could lead to issues, e.g. where on Android your keyboard
and cursor would be shown but you would not be able to type anything,
leading to frustration...

We can still short-circuit through most of the cursor update logic, but
we should make sure to run `this._updateCursorAndOverlay` (which
actually hides the cursor, and is normally run after the rest of the
cursor invalidation/updating)

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -3143,8 +3143,13 @@ L.CanvasTileLayer = L.Layer.extend({
 	// Update cursor layer (blinking cursor).
 	_onUpdateCursor: function (scroll, zoom, keepCaretPositionRelativeToScreen) {
 
-		if (!app.file.textCursor.visible ||
-			this._map.ignoreCursorUpdate()) {
+		if (this._map.ignoreCursorUpdate()) {
+			return;
+		}
+
+		if (!app.file.textCursor.visible) {
+			this._updateCursorAndOverlay();
+			app.definitions.otherViewCursorSection.updateVisibilities(true);
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -3143,8 +3143,13 @@ L.CanvasTileLayer = L.Layer.extend({
...
 
-		if (!app.file.textCursor.visible ||
-			this._map.ignoreCursorUpdate()) {
+		if (this._map.ignoreCursorUpdate()) {
+			return;
+		}
+
+		if (!app.file.textCursor.visible) {
+			this._updateCursorAndOverlay();
+			app.definitions.otherViewCursorSection.updateVisibilities(true);
```
gokaysatir, line 3152: Do we need to update other views' cursors here? Their message is separate i think.

gokaysatir (approved):
Please check if we really need to call "updateVisibilities" function there. If not, you can remove that with a follow up if you wish to get this in now.

PR #9830, opened 2024-08-16
Private/gokay/misc

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/map/Clipboard.js
```diff
@@ -86,17 +86,49 @@ L.Clipboard = L.Class.extend({
...
+			if (startIndex !== -1) {
+				let endIndex = htmlString.indexOf('"', startIndex + 'style="'.length + 1);
+				let subString = htmlString.substr(startIndex, endIndex - startIndex + 1);
+				htmlString = htmlString.replace(subString, "");
+			}
 		}
+
+		// For single quotes.
+		startIndex = 1;
+		while (startIndex !== -1) {
```
eszkadev, line 116: could be a helper function, it is the same as above code block only different `'` vs `"`

eszkadev (changes requested):
I tested with NC where I set custom "theme color" for the user. Without patch that color was used to highlight active buttons in the notebookbar. With the patch it doesn't work, always the standard color is used.

eszkadev (approved):
thank you, theming was fixed

PR #9832, opened 2024-08-16
Accessiblity: FIX Lack of keyboard control for the color picker

Darshan-upadhyay1110:
- Added arrow key navigation to move focus across color picker items.
- Enabled vertical and horizontal focus shifting between custom input fields, color palette items, and recent colors.
- Improved accessibility by handling keyboard events for color selection and navigation.
- Updated `createColor`, `moveFocus`, and `updatePalette` functions to manage focus and selection using keyboard inputs.

Change-Id: Id1ae12360feba1f54091a3285b59ef77cfc17501

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -103,49 +103,266 @@ function createColor(
...
+	}
+
+	return null;
+}
+
+// Helper function to find the first focusable element within an element
+function findFocusableWithin(
+	element: HTMLElement,
+	direction: string,
+): HTMLElement | null {
```
eszkadev, line 335: we already have something similar: https://github.com/CollaboraOnline/online/blob/0e109494b536a52e11c04d8a7017c285a9bf004f/browser/src/control/jsdialog/Util.FocusCycle.js#L18

can we reuse old one?

browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -103,49 +103,266 @@ function createColor(
...
+	}
+	// Update the recent colors list
+	const recentRow = palette[palette.length - 1];
+	if (recentRow.indexOf(colorCode) !== -1)
+		recentRow.splice(recentRow.indexOf(colorCode), 1);
+	recentRow.unshift(colorCode);
+	window.prefs.set('recentColor', JSON.stringify(recentRow));
+}
+
+function moveFocus(
```
eszkadev, line 207: this should be moved to some helper/util class which can be reused on any other popup, example: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Util.FocusCycle.js

Please create new file there: eg. `control/jsdialog/Util.KeyboardGridNavigation.ts`

eszkadev:
nice it will be useful for all other grids I think, I wold like to have it as a separate util we can reuse for other components just by using single function: `JSDialog.KeyboardGridNavigation(div)`

browser/src/control/jsdialog/Widget.ColorPicker.ts
eszkadev: all "focus" related things could go to `Util.FocusCycle.js` and new helper will be only using them and will be for "keyboard handling" part

eszkadev (changes requested):
some code cleanup would be nice

Darshan-upadhyay1110:
> some code cleanup would be nice

Thanks, @eszkadev, for review I am doing the changes. It's just last week I am bit busy with other stuff so did not get chance this push change in this PR

Darshan-upadhyay1110:
Last commit is WIP i need to figure out about cyclic focus for last to first element and first to last element inside super parent element.

/me super parent element is like parent of a parent :)

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,166 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * JSDialog.ScrollableBar - helper for creating toolbars with scrolling left/right
+ */
+
+/* global JSDialog $ */
```
eszkadev, line 16: no need for that in ts?

but please add `declare var JSDialog: any;` instead

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,166 @@
...
+// declare var JSDialog: any;
+
+function moveFocus(
+	parentContainer: HTMLElement,
+	currentElement: HTMLElement,
+	direction: 'next' | 'previous',
+	axis: 'horizontal' | 'vertical',
+	nextElement?: Element,
+) {
+	const focusableElements = Array.from(
```
eszkadev, line 27: we should use single function  everywhere for getting focusable elements, there is already util for that:
https://github.com/CollaboraOnline/online/blob/f55c1a1403fcd350e81fdb98bd6c3a6b05cfda2b/browser/src/control/jsdialog/Util.FocusCycle.js#L18

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,166 @@
...
+	direction: string,
+): HTMLElement | null {
+	const focusableElements = Array.from(element.querySelectorAll('*'));
+	return direction === 'next'
+		? (focusableElements.find(isFocusable) as HTMLElement | null)
+		: (focusableElements.reverse().find(isFocusable) as HTMLElement | null);
+}
+
+// Utility function to check if an element is focusable
+function isFocusable(element: HTMLElement) {
```
eszkadev, line 143: we already have util like that, please reuse: https://github.com/CollaboraOnline/online/blob/f55c1a1403fcd350e81fdb98bd6c3a6b05cfda2b/browser/src/control/jsdialog/Util.FocusCycle.js#L18

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,166 @@
...
+
+	return focusableElements.some((selector) => element.matches(selector));
+}
+
+function getRowColumn(element: HTMLElement): [number, number] {
+	const index = element.getAttribute('index');
+	if (!index) return [-1, -1]; // we will never have this kind of index this is why we are pssing nagative values here
+	const [row, column] = index.split(':').map(Number);
+	return [row, column];
+}
```
eszkadev, line 166: at the end please add "exported" / public functions:
`JSDialog.nameOfFuntion = localNameOfFunction;`
and in the other files please use that `JSDialog.nameOfFunction` to call it.

I can imagine we should need only one function to be exported, so we can add this functionality to the widget just by one line:
`JSDialog.KeyboardGridNavigation(<some grid container>, <some additional args>);`

browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -445,6 +293,7 @@ function createPaletteSwitch(
 			event.preventDefault(); // Prevent default behavior
```
eszkadev, line 293: above keyboard event listener also could be moved to util (it doesn't use widget-specific things, just our helper functions)

browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -135,15 +135,16 @@ function createColor(
 				event.preventDefault();
```
eszkadev, line 135: above keyboard event listener also could be moved to util (it doesn't use widget-specific things, just our helper functions)

eszkadev: handleColorSelection will be passed as: `activateCallback`

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
Darshan-upadhyay1110: yes we have but it returns list of focusable element so i think if there is a function which can help to find the current element is focusable or not can be more helpful. But yes, i will this into `util.FocusCycle.js`

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
Darshan-upadhyay1110: > I can imagine we should need only one function to be exported, so we can add this functionality to the widget just by one line:
> `JSDialog.KeyboardGridNavigation(<some grid container>, <some additional args>);`

I am doing R and D on this suggestion

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
Darshan-upadhyay1110: I see at many places we are using same function but then after to get first focusable element we do like
```
FocusableElements[0].focus

```

This seems not okay because we are adding many code duplication here so by this `isFocusable` function we can directly get our first focusable element.

What you say @eszkadev on this ?

eszkadev: yes, please try to reuse as much as possible. If we need: `getFirstFocusableElement()` we casn add it.
best to keep everything in the util file you mentioned so everything is in one place

Darshan-upadhyay1110: sure @eszkadev now all function are in Util.FocusCycle.js

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
Darshan-upadhyay1110: Done 🎉

browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -110,29 +110,15 @@ function createColor(
 	if (themeData) color.setAttribute('theme', themeData);
 
 	color.innerHTML = isCurrent ? '&#149;' : '&#160;';
-
-	color.addEventListener('click', (event: MouseEvent) =>
+	// Assuming 'color' is your target HTMLElement
+	color.addEventListener('click', function (event: MouseEvent) {
```
eszkadev, line 114: please use arrow functions in the new code:` () => { }`

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -164,4 +131,17 @@ function getRowColumn(element: HTMLElement): [number, number] {
...
+	container: HTMLElement,
+	options?: any,
+) {
+	container.addEventListener('keydown', (event: KeyboardEvent) => {
+		const activeElement = document.activeElement as HTMLElement;
+		handleKeyboardNavigation(
+			event,
+			activeElement,
+			activeElement.parentElement,
+			...options,
```
eszkadev, line 144: can we explicitly pass the arguments?
it will help to catch mistakes in the future if we use types

eszkadev: it's the only call, so it's not a lot of work

eszkadev:
> Hii @eszkadev i have added the generic function for any grid implementation as `JSDialog.KeyboardGridNavigation`.
> 
> /me need more views on this solution. i have made 3 commits, but after your approval i will squash them into one. Those are just for my understanding and as phase wise implementation.
> 
> Thanks

please do not squash, it is easier to review and later check code if it is in smaller pieces (especially if they are logically split into sensible parts)

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
Darshan-upadhyay1110: > can we explicitly pass the arguments?
@eszkadev can you give an example of passing args explicitly ?
I kind of not sure about that.....

browser/src/control/jsdialog/Widget.ColorPicker.ts
```diff
@@ -367,5 +389,15 @@ JSDialog.colorPicker = function (
 		);
 	});
 
+	// Apply keyboard navigation to the color picker widget
+	JSDialog.KeyboardGridNavigation(container, [
```
eszkadev, line 393: do not pas array here

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
eszkadev: just don't use `options?: any` - where you put array of params and then you unfold it
please add names and types for all args you pass here:
```
JSDialog.KeyboardGridNavigation = function (
	container: HTMLElement,
        handleColorSelection: ...,
        builder: ...,
        data: ...,
        palette: ...
```

browser/src/control/jsdialog/Util.FocusCycle.js
```diff
@@ -63,5 +63,63 @@ function makeFocusCycle(container, failedToFindFocusFunc) {
...
+	);
+	if (focusableInSibling) return focusableInSibling;
+
+	// Depending on the direction, find the next or previous sibling
+	const sibling =
+		direction === 'next'
+			? element.nextElementSibling
+			: element.previousElementSibling;
+
+	if (sibling) {
```
eszkadev, line 87: no need for this check, we can just return findFocusableElement as it returns null for non existing element

browser/src/control/jsdialog/Util.FocusCycle.js
```diff
@@ -63,5 +63,63 @@ function makeFocusCycle(container, failedToFindFocusFunc) {
...
+		// Recursively check the next or previous sibling of the current sibling
+		return findFocusableElement(sibling, direction);
+	}
+
+	return null;
+}
+
+// Utility function to check if an element is focusable
+// This function diffrent from getFocusableElements. This only checkes the current element is focusable or not.
+function isFocusable(element) {
```
eszkadev, line 97: can we put that function close to the other function which generates list of focusable elements inside container?
so we will not forget to update it in case we will modify the other, we should think also how to share the conditions between them in the future, if that possible (not here, in the followup)

Darshan-upadhyay1110: sure let me change the position

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,153 @@
...
+	handleSelection?: any,
+	builder?: any,
+	widgetData?: any,
+	palette?: any,
+) {
+	switch (event.key) {
+		case 'Enter':
+		case ' ':
+			if (handleSelection && !(currentElement instanceof HTMLSelectElement)) {
+				handleSelection(currentElement, builder, widgetData, palette);
```
eszkadev, line 31: palette is specific for the color picker, can we somehow avoid it? so it will not confuse others if we will reuse this util?

eszkadev:
looks awesome now :) thanks, nice helper which can be later reused, just added some questions if we can it even improve, but no need to do all of them now

Darshan-upadhyay1110:
> looks awesome now :) thanks, nice helper which can be later reused, just added some questions if we can it even improve, but no need to do all of them now

Thanks for the review, i am trying to address all your queries and will improve them in this PR

eszkadev:
works nice in general, I found small problem: when I typed something into "custom color input", I cannot move cursor using keyboard arrows

eszkadev (approved):
let's merge this as it works fine, just a small fix can be added later

browser/src/control/jsdialog/Util.KeyboardGridNavigation.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * JSDialog.ScrollableBar - helper for creating toolbars with scrolling left/right
```
eszkadev, line 13: copy-pasted wrong comment cc @Darshan-upadhyay1110

Darshan-upadhyay1110: https://github.com/CollaboraOnline/online/pull/9940

This PR will fix both the issue

PR #9834, opened 2024-08-16
fix(android): Stop saving readonly docs

Minion3665:
When we switch apps, we previously always saved the document. This causes trouble with some documents, such as PDFs, that can't be saved. When we try, the save fails and we cause an error popup.

Instead, let's not save the document when we switch apps and aren't in edit mode. To avoid missing any saves, we should also save the document when we leave edit mode.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
thanks, code looks sensible but I don't have android build at hand to test

Minion3665:
> thanks, code looks sensible but I don't have android build at hand to test

thanks! In that case, I think I'd like @Rash419 to take a look too - although I think this should be fairly standard :)

PR #9840, opened 2024-08-19
cool#9621: Resolve POST fileserver request

sgothel:
* Resolves: #9621 POST fileserver request
* Target version: master 

Change-Id: I7db359aeb7773733f817a31e8777e076b8cd4408

### Summary
cool#9621: Resolve POST fileserver request, commit 6e4d2b2350100367f780664a4c5af1f5d7702488
    
Commit 6e4d2b2350100367f780664a4c5af1f5d7702488 mitigated the issue of
wrongly handled POST requests in ClientRequestDispatcher::handleIncomingMessage()
by simply shutting down the socket (non keep-alive socket).
   
This patch resolves the issue by removing the complete HTTP request
including the optional HTTP POST body from the socket's buffer,
i.e. removing map._messageSize bytes.
    
It also includes a fix in StreamSocket::parseHeader()
of only adjusting map._messageSize w/ contentLength if the latter is valid.
    
Hence we also allow reusing the given socket connection
for POST HTTP requests including mixing POST and GET.    

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1330,21 +1330,28 @@ bool StreamSocket::parseHeader(const char *clientName,
     {
         request.read(message);
 
-        LOG_INF("parseHeader: " << clientName << " HTTP Request: " << request.getMethod() << ' ' << request.getURI()
-                           << ' ' << request.getVersion() << ' '
-                           << [&](auto& log) { Util::joinPair(log, request, " / "); });
-
         const std::streamsize contentLength = request.getContentLength();
         const auto offset = itBody - _inBuffer.begin();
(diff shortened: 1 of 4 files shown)


caolanm (commented):
this looks good to me, but lets move the "commit XXX" out of the first line of the commit message, can keep it in the body of the commit message

caolanm (approved):
looks good to me

caolanm:
I think the cypress failures are a "real" problem, but I think that will be solved by https://github.com/CollaboraOnline/online/pull/9848

Ashod (approved):
Thanks, @sgothel. Looks very good.

PR #9839, opened 2024-08-19
Adopt WopiStorage::uploadLocalFileToStorageAsync() to ConnectFailCallback

sgothel:
Change-Id: I65ee775c584317d6dff720d009de8fe9874bfc8c

### Summary

Commit f6ce37b579c1753e61a71788a78bb25d5d60eb1d created the change and regression,
master branch failed to compile

PR #9841, opened 2024-08-19
Fix build in clang

eszkadev:
Some older PR was not compatible.

There was old check of success introduced in
commit 76fded3ef0065a04753422ba8b2910802943efcb
wsd: check for asyncRequest success for uploading

but in the meantime we have new callback for connection
fail what requires change from us:

commit f6ce37b579c1753e61a71788a78bb25d5d60eb1d
hook Session::asyncRequest up to use net::asyncConnect
calls _onConnectFail on async connect fail instead of returning false on
sync connect fail

So let's not use scoped callback trigger but rather do it manually
on returns and in the fail callback. Successfull one is already
done this way. Notice: on success we were firing it two times:
on Running state and Complete. Running doesn't happen if we fail
on connection...

### DIFF
--- a/wsd/wopi/WopiStorage.cpp
+++ b/wsd/wopi/WopiStorage.cpp
@@ -624,18 +624,13 @@ void WopiStorage::uploadLocalFileToStorageAsync(const Authorization& auth, LockC
 
     // TODO: Check if this URI has write permission (canWrite = true)
 
-    // Always invoke the callback with the result of the async upload.
-    ScopedInvokeAsyncRequestCallback<AsyncUploadCallback, StorageBase::AsyncUpload>
-        scopedInvokeCallback(
-            asyncUploadCallback,
-            StorageBase::AsyncUpload(
-                StorageBase::AsyncUpload::State::Error,
-                StorageBase::UploadResult(StorageBase::UploadResult::Result::FAILED)));
-
     //TODO: replace with state machine.
     if (_uploadHttpSession)
     {
         LOG_WRN("Upload is already in progress.");
+        asyncUploadCallback(
+            AsyncUpload(AsyncUpload::State::Error,
+                UploadResult(UploadResult::Result::FAILED, "Already in progress.")));
         return;
     }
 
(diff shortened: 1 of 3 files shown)


caolanm (approved):
seeing as we're not currently building, lets get this in and if we need to adjust further lets do that on a usable build

wsd/wopi/WopiStorage.cpp
```diff
@@ -784,12 +779,14 @@ void WopiStorage::uploadLocalFileToStorageAsync(const Authorization& auth, LockC
...
+            asyncUploadCallback(AsyncUpload(AsyncUpload::State::Error,
+                            UploadResult(UploadResult::Result::FAILED, "Connection failed.")));
+        });
+
         // Make the request.
-        if (_uploadHttpSession->asyncRequest(httpRequest, socketPoll))
-        {
-            scopedInvokeCallback.setArg(
-                AsyncUpload(AsyncUpload::State::Running, UploadResult(UploadResult::Result::OK)));
-        }
```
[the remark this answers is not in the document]
sgothel: so this will be re-added by .. ?
guess I shelf my patch then and rebase my other work. OK.

wsd/wopi/WopiStorage.cpp
```diff
@@ -784,12 +779,14 @@ void WopiStorage::uploadLocalFileToStorageAsync(const Authorization& auth, LockC
...
+        _uploadHttpSession->asyncRequest(httpRequest, socketPoll);
```
sgothel, line 789: Should happen before line 789 .. (who?)

// Notify client via callback that the request is in progress...
asyncUploadCallback(
            AsyncUpload(AsyncUpload::State::Running, UploadResult(UploadResult::Result::OK)));

caolanm: Yes, seeing as there is the near duplicate alt pr of https://github.com/CollaboraOnline/online/pull/9839 but that retains the "State:Running" case, let me rebase and reuse that to reintroduce this

PR #9842, opened 2024-08-19
cypress: fix errors due to missing a11y handling

eszkadev:
regression from commit e36d6ffb9b13ffd440744425da128a9446530f01 fix: fix UI preference inconsistencies

where we introduced second place where enableAccessibility is controlled. This caused that we were out of sync in few places so we used regular TextInput while having a11y enabled...

We got bunch of this._map._textInput.onAccessibilityFocusChanged is not a function

This was visible only after we added additional dialogs/errors on unhandled exceptions inside cypress recently.

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -691,6 +691,27 @@ function getInitializerClass() {
 		},
 	};
 
+	global.getAccessibilityState = function () {
+		var isCalcTest =
+			global.docURL.includes('data/desktop/calc/') ||
+			global.docURL.includes('data/mobile/calc/') ||
+			global.docURL.includes('data/idle/calc/') ||
+			global.docURL.includes('data/multiuser/calc/');
+
+		// FIXME: a11y doesn't work in calc under cypress
+		if (L.Browser.cypressTest && isCalcTest)
+			global.enableAccessibility = false;
+
+		if (L.Browser.cypressTest) {
+			global.prefs.set('accessibilityState', global.enableAccessibility);
+			global.app.accessibilityState = global.enableAccessibility;
+		}
+
+		var accessibilityState = global.prefs.getBoolean('accessibilityState');
+		global.app.accessibilityState = accessibilityState;
+		return accessibilityState;
+	};
+
 	// Renamed in 24.04.4.1
 	const prefDocTypes = ['text', 'spreadsheet', 'presentation', 'drawing'];
 	for (const docType of prefDocTypes) {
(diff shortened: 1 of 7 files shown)


browser/js/global.js
```diff
@@ -1595,11 +1595,16 @@ function getInitializerClass() {
...
 				if (L.Browser.cypressTest && isCalcTest)
 					global.enableAccessibility = false;
 
+				if (L.Browser.cypressTest)
+					window.prefs.set('accessibilityState', global.enableAccessibility);
+
 				var accessibilityState = global.prefs.getBoolean('accessibilityState');
-				accessibilityState = accessibilityState || (L.Browser.cypressTest && !isCalcTest);
+				if (accessibilityState === undefined)
+					accessibilityState = global.enableAccessibility;
```
Minion3665, line 1607: You can specify the default value of something in the prefs handler like so
```js
global.prefs.getBoolean('accessibilityState', global.enableAccessibility);
```

The default default value is `false`, making your code equivalent to 

```js
var accessibilityState = global.prefs.getBoolean('accessibilityState');

if (false) // accessibilityState will never be `undefined`
	accessibilityState = global.enableAccessibility;
```

The reason it wasn't like that before was that accessibilityState was *always* true if it was in localStorage, although that probably wasn't the best for a cypress test anyway

browser/js/global.js
```diff
@@ -1595,11 +1595,16 @@ function getInitializerClass() {
 					global.docURL.includes('data/idle/calc/') ||
 					global.docURL.includes('data/multiuser/calc/');
 
+				// FIXME: it doesn't work in calc under cypress
```
Minion3665, line 1598: slight nit: may you clarify what "it" is here ... I presume you mean "accessibility mode" but I'm not entirely sure

browser/src/map/Map.js
```diff
@@ -317,7 +312,7 @@ L.Map = L.Evented.extend({
 	},
 
 	initTextInput: function(docType) {
-		var hasAccessibilitySupport = window.enableAccessibility && this._accessibilityState;
+		var hasAccessibilitySupport = window.enableAccessibility;
```
Minion3665, line 315: I'm not sure that this doesn't change behaviour outside of cypress... it looks to me like `_accessibilityState` is whether the user has toggled accessibility on and `enableAccessibility` is whether it's enabled in the server ... given that I didn't change that in my patch, I'm surprised you need to here

Was this intentional?

Minion3665 (changes requested):
Mostly small things - the big one is the dropping of `&& this._accessibilityState` in line 315 of `Map.js`

Minion3665 (approved):
There are still things I think could make this nicer, but they're now firmly in the realm of "would be a nice refactor/is a nitpick" - in other words, I'm happy to get this in, thanks!

eszkadev:
yes, but for now I just want to have cypress green locally :)
it fixes desktop, probably need to disable it for mobile as a11y is not useful there

PR #9843, opened 2024-08-19
Private/caolan/async dns tidy

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -102,27 +102,26 @@ HostEntry::HostEntry(const std::string& desc, const char* port)
 
     addrinfo* ainfo = nullptr;
     int rc = getaddrinfo(desc.c_str(), port, &hints, &ainfo);
-    _ainfo.reset(ainfo, freeaddrinfo);
-    if (rc == 0)
+    if (rc != 0)
     {
-        for (const addrinfo* ai = _ainfo.get(); ai; ai = ai->ai_next)
-        {
-            if (ai->ai_canonname)
-                _canonicalName.assign(ai->ai_canonname);
-
-            if (!ai->ai_addrlen || !ai->ai_addr)
-                continue;
-
-            std::string address = makeIPAddress(ai->ai_addr);
-            if (!good())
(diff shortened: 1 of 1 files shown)


net/AsyncDNS.hpp
```diff
@@ -66,7 +66,7 @@ class AsyncDNS
 
     void resolveDNS();
     void addLookup(const std::string& lookup,
-                   const std::optional<std::string>& port,
+                   const std::string& port,
                    const DNSThreadFn& cb,
                    const DNSThreadDumpStateFn& dumpState);
```
sgothel, line 72: mandatory port arg, good.

net/NetUtil.cpp
```diff
@@ -129,23 +96,46 @@ HostEntry::HostEntry(const std::string& desc, const char* port)
...
-    initFromHostName(desc, port);
-}
+    addrinfo hints;
+    std::memset(&hints, 0, sizeof(hints));
+    hints.ai_flags = AI_CANONNAME | AI_ADDRCONFIG;
 
-HostEntry::~HostEntry()
-{
+    addrinfo* ainfo = nullptr;
+    int rc = getaddrinfo(desc.c_str(), port, &hints, &ainfo);
```
sgothel, line 104: ::getaddrinfo(..) ? (global namespace)

sgothel: personally I force myself to use ::system_call(..) to hint to me it is such a global function and not a method by any chance. Just taste.

net/AsyncDNS.hpp
sgothel: OK, port string can be empty(), so still optional.

sgothel (approved):
looks good. removal of optional reduces the argument check as string.empty() check was missing. Plus added assertion message is helpful.

net/NetUtil.cpp
caolanm: yeah, reasonable

PR #9845, opened 2024-08-19
Impress UI: Turn notes view and normal view buttons into big tool ite…

gokaysatir:
…m buttons.

Change-Id: I7088382d9319c4b8119708c55c2e7c027bc2efe1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarImpress.js
+++ b/browser/src/control/Control.NotebookbarImpress.js
@@ -447,22 +447,18 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 				'accessibility': { focusBack: true, combination: 'CU', de: null }
 			},
 			{
-				'type': 'container',
-				'children': [
-					{
-						'id': 'view-normal-view',
-						'type': 'toolitem',
-						'text': _('Normal View'),
-						'command': '.uno:NormalMultiPaneGUI',
-					},
-					{
-						'id': 'view-notes-view',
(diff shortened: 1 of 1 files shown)


browser/images/lc_view-normal-view.svg
eszkadev, line 1: we should prefer filename based on uno command, it is the priority to get icon name and also used by other components like sidebar. id defined in nb can not match

eszkadev (changes requested):

browser/images/lc_view-normal-view.svg
gokaysatir: But it doesn't match with the uno-command-names?

pedropintosilva:
I have amended the commit to don't rename the svg. So it says the same name as the uno command. I have tested.

eszkadev (approved):
looks ok

gokaysatir:
> this is a good improvement, thanks. As a follow up PR I would love to
> 
> * Have solely one button 'Notes View' that is a toggle (on/off). Otherwise, it's quite cumbersome for the user to understand what to press to be able to exit the notes view
> * When switching between views the Active tab shouldn't change.
> 
> Note: Currently I'm unable to exist the notes view no matter if I press the notes or the normal view buttons (on Firefox, master)

Thanks @pedropintosilva :) I will create an internal task for the additional improvements.

PR #9846, opened 2024-08-19
fix: Slideshow repeat show last slide after pause countdown for few seconds

codewithvk:
fix: Slideshow repeat show last slide after pause countdown for few seconds

Change-Id: I99b8458f72c67ce9bb5c088c4e1ed6d2ff9273a5

* Resolves: # 
* Target version: master


quikee (approved):
Looks fine

PR #9848, opened 2024-08-19
on reconnect of calc, annotations requested before geometry is available

caolanm:
this activate is called when we reconnect, but with the right timing we can have no docLayer here, so we can't tell that this is calc, and so end up requesting annotations without requesting calc geometry. The annotations then arrive before geometry does and on positioning the comments we fail with exception on use of the undefined geometry.

reproducible with:

make run-mobile spec=calc/annotation_spec.js

cy: command ✘  uncaught exception    TypeError: Cannot read properties of undefined (reading 'getCellRect')
    cy:command ✔ error:    Cannot read properties of undefined (reading 'getCellRect')
        TypeError: Cannot read properties of undefined (reading 'getCellRect')
          at NewClass._cellRangeToTwipRect (.../src/layer/tile/CanvasTileLayer.js:1032:53)
          at CommentSection.adjustCommentNormal (.../src/canvas/sections/CommentListSection.js:1362:79)

Change-Id: I5c868bec008668a5f7f33c3e32d83f40b356c566

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.IdleHandler.ts
+++ b/browser/src/control/Control.IdleHandler.ts
@@ -54,6 +54,14 @@ class IdleHandler {
 		return (Date.now() - this._lastActivity) / 1000;
 	}
 
+	refreshAnnotations() {
+		var docLayer = this.map._docLayer;
+		if (docLayer.isCalc() && docLayer.options.sheetGeometryDataEnabled) {
+			docLayer.requestSheetGeometryData();
+		}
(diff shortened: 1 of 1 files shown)


sgothel:
Using this as a base for my https://github.com/CollaboraOnline/online/pull/9840

sgothel (approved):
looks good to me, but me no JS expert

PR #9850, opened 2024-08-20
a11y: simplify state handling

eszkadev:
- don't use additional variable to store state, use prefs
- convert menubaritem in Nb to customtoolitem
- use dispatcher for sharing the code

followup for: https://github.com/CollaboraOnline/online/pull/9842


Minion3665 (approved):
thanks very much, this is looking a lot neater :)

PR #9855, opened 2024-08-20
survive if config.xml is missing indirection_endpoint.geolocation_set…

caolanm:
…up.enable

which an old config will not have and results in failure to connect as exception is thrown

Change-Id: Ic2c9f66aa180251c90424578eb6cc2cd5aaecb09

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
Thanks for the patch. I missed it : )

timar (commented):
I'd also add the default to COOLWSD::innerInitialize()

caolanm:
> I'd also add the default to COOLWSD::innerInitialize()

added that now

Issue #7591, opened 2023-11-02
"Format > Columns" Menu: Style Dropdown Text Misaligned


Tex2002ans:
### Describe the Bug

In **Format > Columns**, the "None" text in the Style dropdown is oddly misaligned:

![Collabora 23 05 2 - Layout Columns - Columns Menu - Misaligned Styles Dropdown](https://github.com/CollaboraOnline/online/assets/5774324/a74445e0-03a9-497d-9ac6-183dff6b3b6c)

### Steps to Reproduce

In Writer:

1. Press **Format > Columns** button.

See misaligned text.

### Expected Behavior

"None" to vertically align in center, just like the other dropdowns with text.

### Actual Behavior

"None" looks like it's at the top of the box.

### Desktop

- **OS:** Linux Mint 21.1
- **Browser:** Firefox
- **Version:** 119.0
 
> COOLWSD version: 23.05.5.3 (git hash: 557f191d)
> LOKit version: Collabora Office 23.05.5.3 (git hash: 894db9a)
> Served by: Linux Mint 21.1
> Server ID: 36f66248

### Additional Context

That specific dropdown is where you choose what line types as a separator, so the graphics-mixed-with-text might be throwing this alignment off.

Darshan-upadhyay1110:
Adding this in `easy hack` section
Potential mentor: @pedropintosilva

PR #9863, opened 2024-08-20
fix format > columns menu misaligned style text

ChaseNelson:
Change-Id: Icdac9332fcd9584dc004f98a52772da041ff8496

* Resolves: #7591
* Target version: master 

### Summary

When "None" is selected for the Columns style, the text is misaligned with the arrow.

image


Darshan-upadhyay1110 (approved):
Looks good to me. Thanks !

Darshan-upadhyay1110:
@pedropintosilva i have tested with multiple JS dialogs and this seems not creating any regression. Instead, it fixes this same issues in many dialogs, so i think this is a safe patch and okay to merge it.

PR #9866, opened 2024-08-21
Fix comment box cut-off issue when zooming in

Darshan-upadhyay1110:
- Adjusted the collapsedMarginToTheEdge so comment box does not get cut after document zoom
- better to have value same for all devices

Change-Id: I626388cdbf1db5bedf5f3fc5fe180e3d15a9ecfc

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (changes requested):
Thanks @Darshan-upadhyay1110 :)

I think it'd be better to play with the value of "collapsedMarginToTheEdge" variable, until we get a good margin from right side.

The variable name seems correct but its value needs to be modified.

Darshan-upadhyay1110:
24.04: Backport => https://github.com/CollaboraOnline/online/pull/9897
23.05: backport =>  https://github.com/CollaboraOnline/online/pull/9898

PR #9867, opened 2024-08-21
Async Lock Refreshing

Ashod:
- wsd: async updateLockState
- wsd: disable editing when document locking fails
- wsd: LockUpdateResult captures the requested lock state
- wsd: read the readonly flag after writing
- wsd: async lock/unlock documents
- wsd: better logging of http status in WopiStorage

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -1604,8 +1604,9 @@ bool ClientSession::filterMessage(const std::string& message) const
 void ClientSession::setReadOnly(bool bVal)
 {
     Session::setReadOnly(bVal);
-    // Also inform the client
-    const std::string sPerm = bVal ? "readonly" : "edit";
+
+    // Also inform the client.
+    const std::string sPerm = isReadOnly() ? "readonly" : "edit";
     sendTextFrame("perm: " + sPerm);
 }
 
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1120,9 +1120,10 @@ void DocumentBroker::lockIfEditing(const std::shared_ptr<ClientSession>& session
             std::string error;
             if (!updateStorageLockState(*session, StorageBase::LockState::LOCK, error))
             {
-                LOG_ERR("Failed to lock docKey [" << _docKey << "] with session ["
-                                                  << session->getId()
-                                                  << "] after downloading: " << error);
+                LOG_ERR("Failed to lock docKey ["
+                        << _docKey << "] with session [" << session->getId()
+                        << "] before downloading. Session will be read-only: " << error);
+                session->setWritable(false);
             }
         }
 
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -471,8 +471,10 @@ void DocumentBroker::pollThread()
             _admin.addBytes(getDocKey(), deltaSent, deltaRecv);
         }
 
-        if (_storage && _lockCtx->needsRefresh(now))
+        if (_storage && !_lockStateUpdateRequest && _lockCtx->needsRefresh(now))
+        {
             refreshLock();
+        }
 #endif
 
         LOG_TRC("Poll: current activity: " << DocumentState::name(_docState.activity()));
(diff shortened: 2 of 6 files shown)


wsd/Storage.hpp
```diff
@@ -532,8 +533,14 @@ class LocalStorage : public StorageBase
...
     void updateLockStateAsync(const Authorization&, LockContext&, LockState, const Attributes&,
-                              const AsyncLockStateCallback&) override
+                              SocketPoll&,
+                              const AsyncLockStateCallback& asyncLockStateCallback) override
     {
+        if (asyncLockStateCallback)
+        {
+            asyncLockStateCallback(AsyncLockUpdate(AsyncLockUpdate::State::Complete,
+                                                   LockUpdateResult(LockUpdateResult::Status::OK)));
+        }
```
sgothel, line 551: Is this a temporary patch to be completed at a later time?
- Why is this just claiming AsyncLockUpdate::State::Complete, disregarding the LockState?
- Why is LockState (and others) ignored? LockState and AsyncLockUpdate::State have different semantics.

wsd/wopi/WopiStorage.cpp
```diff
@@ -423,10 +423,123 @@ StorageBase::LockUpdateResult WopiStorage::updateLockState(const Authorization&
...
 
-void WopiStorage::updateLockStateAsync(const Authorization& /*auth*/, LockContext& /*lockCtx*/,
-                                       LockState /*lock*/, const Attributes& /*attribs*/,
-                                       const AsyncLockStateCallback& /*asyncLockStateCallback*/)
+void WopiStorage::updateLockStateAsync(const Authorization& auth, LockContext& lockCtx,
+                                       LockState lock, const Attributes& attribs,
+                                       SocketPoll& socketPoll,
+                                       const AsyncLockStateCallback& asyncLockStateCallback)
 {
+    auto profileZone = std::make_shared<ProfileZone>(
```
sgothel, line 431: I have not read the WOPI spec thoroughly, hence I can't properly review this yet. 
Other that this - the async change and its pattern looks good!

In general, as discussed, I like us to add the used section of the specification in the header API doc, allowing us to quickly validated and also know the impact etc.

wsd/Storage.hpp
```diff
@@ -526,20 +532,22 @@ class LocalStorage : public StorageBase
...
+                              const Attributes&, SocketPoll&,
                               const AsyncLockStateCallback& asyncLockStateCallback) override
     {
         if (asyncLockStateCallback)
         {
-            asyncLockStateCallback(AsyncLockUpdate(AsyncLockUpdate::State::Complete,
-                                                   LockUpdateResult(LockUpdateResult::Status::OK)));
+            asyncLockStateCallback(AsyncLockUpdate(
+                AsyncLockUpdate::State::Complete,
+                LockUpdateResult(LockUpdateResult::Status::OK, requestedLockState)));
```
sgothel, line 550: Here the given LockState is being used, great. See remark to wsd/Storage.hpp earlier.

wsd/DocumentBroker.cpp
```diff
@@ -471,8 +471,10 @@ void DocumentBroker::pollThread()
             _admin.addBytes(getDocKey(), deltaSent, deltaRecv);
         }
 
-        if (_storage && _lockCtx->needsRefresh(now))
+        if (_storage && !_lockStateUpdateRequest && _lockCtx->needsRefresh(now))
+        {
```
sgothel, line 475: Remark: So we need refreshLock() also if the _lockStateUpdateRequest is not yet set, i.e. desired if not required.

refreshLock() then retries to use a writable and authorized session and propagates down to (old: updateStorageLockState()) now updateStorageLockStateAsync().
Within this path I assume _lockStateUpdateRequest will be set eventually? 
Yes, updateStorageLockStateAsync() assigns the single _lockStateUpdateRequest entity.

sgothel (approved):
Assuming this is a WIP for now, looks good.
Would love to discuss added questions/remarks for a better picture.

wsd/DocumentBroker.cpp
sgothel: Additional thought: Feasible to have the _lockCtx hold _lockStateUpdateRequest, so it also could use it for the `needsRefresh()` reply?

wsd/Storage.hpp
Ashod: Capturing the lock state is added later, in a separate commit. If you looked at individual commits, it might not have been there. Notice that both your comments saying 'it's missing' and, later, that 'it's provided' are on the same function.

I also checked and couldn't find a case that was missing the lock state.

wsd/Storage.hpp
Ashod: See reply to related comment.

wsd/DocumentBroker.cpp
Ashod: I think the `_lockCtx` should be part of the Storage instance. 

What this is doing is simply checking that there isn't already an async request in flight. If there is, we don't need to issue a second one.

Perhaps we can simplify it, but I rather deal with `_lockCtx` more intelligently first.

wsd/ClientSession.cpp
```diff
@@ -1604,8 +1604,9 @@ bool ClientSession::filterMessage(const std::string& message) const
 void ClientSession::setReadOnly(bool bVal)
 {
     Session::setReadOnly(bVal);
-    // Also inform the client
-    const std::string sPerm = bVal ? "readonly" : "edit";
+
+    // Also inform the client.
+    const std::string sPerm = isReadOnly() ? "readonly" : "edit";
```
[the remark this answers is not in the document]
Ashod: I didn't follow your points very well. 

All this is doing is not assuming that `Session::setReadOnly()` will succeed in setting `bVal` as we want it. This is somewhat incomplete, since for now we always do. I have a follow up patch to check if the user has permission on not. 

But, conceptually, this change is reasonable: try to change the value of something, and use the value as it _exists_, _**not** as requested_.

wsd/wopi/WopiStorage.cpp
Ashod: Thanks. Understood. Although this code doesn't make any WOPI changes (it's changing an existing synchronous request to be async), more documentation is good.

Ashod (commented):
Thanks, @sgothel. This is ready for review/merging. As a feature, it doesn't have anything missing to make it fully functional. Left some responses to your detailed and thoughtful comments--thank you.

caolanm:
Alright, lets push during this red block for max length for testing

PR #9871, opened 2024-08-21
Chart Wizard: fix the up arrow image

meven:
Change-Id: I5e96ab2e7fe4d016674134b0b2da12a9644d392b

### Summary
Before:
image

After:

image


eszkadev (approved):

meven:
Just added a default height and width to have it render correctly (as the DOWN arrow does).

PR #9872, opened 2024-08-21
Attempt to use a more precise filter for time field.

gokaysatir:
Core side PR: https://gerrit.libreoffice.org/c/core/+/172239

The problem:
When there is no unit for the number field, it is treated as time field.
Users with "," as decimal separator have issues when the field type is not set properly.

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1478,12 +1478,6 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		value = parseFloat(data.value);
 
-		// time formatter or empty field
-		if (data.unit === ':' || (!data.unit && !data.text)) {
-			controls.spinfield.type = undefined;
-			value = data.text;
-		}
-
 		$(controls.spinfield).val(value);
 
 		return false;


eszkadev (changes requested):
This change breaks spinfields: no up/down arrow is shown, eg in the Impress sidebar.
Please also add more details into commit message - where to test and what it does (which cases it fixes).
If value is preformatted in the core - we probably need to show it as regular text, not spinfield. But regular spinfields need to be real spinfield to show correct ui on mobile.

spinfield

eszkadev:
> Please also add more details into commit message - where to test and what it does (which cases it fixes). If value is preformatted in the core - we probably need to show it as regular text, not spinfield. But regular spinfields need to be real spinfield to show correct ui on mobile.

could you add details how to test, what it changes?

gokaysatir:
> > Please also add more details into commit message - where to test and what it does (which cases it fixes). If value is preformatted in the core - we probably need to show it as regular text, not spinfield. But regular spinfields need to be real spinfield to show correct ui on mobile.
> 
> could you add details how to test, what it changes? + link to core change or at least title

Yep, thanks for checking :) Added more detailed explanation.

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1479,7 +1479,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		value = parseFloat(data.value);
 
 		// time formatter or empty field
-		if (data.unit === ':' || (!data.unit && !data.text)) {
+		if (data.hasTimeFormatter) {
```
eszkadev, line 1482: did you find any field which uses this condition?

I looked again and it seems widget with `"type"="time"` uses `JSDialog.timeField`, and it's the only widget which can trigger this condition... so probably it's leftover code from before we introduced `JSDialog.timeField`

eszkadev: File -> Properties -> Custom properties with field type date time has it it seems

PR #9878, opened 2024-08-21
during a scroll don't reissue tilecombines requests already issued

caolanm:
since the start of the scroll, reset on end of scroll

Change-Id: I2f3e79d9f34ddcdd5b7ad64f35f4d81ee1fadd6d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
Here we track what tiles we requested during a scroll so we don't issue a request for it more than once per scroll to cut down unnecessary requests

eszkadev (approved):
looks ok

PR #9881, opened 2024-08-22
annotation: after importing comment set showResolved stat

lpranam:
problem:
once comments are imported(i.e after reload) regardless of the show resolved comments, they were always shown

Change-Id: Ieb425faf2566bf7694723bea37dc9660f4c0ef5b

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -2137,6 +2137,9 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 			this.update();
 		}
 
+		var showResolved = this.map.stateChangeHandler.getItemValue('ShowResolvedAnnotations');
+		app.map.showResolvedComments(showResolved === true || showResolved === 'true');
+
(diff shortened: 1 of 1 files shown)


pedropintosilva:
Adding @eszkadev as @gokaysatir is out

browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -2137,6 +2137,9 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 			this.update();
 		}
 
+		var showResolved = this.map.stateChangeHandler.getItemValue('ShowResolvedAnnotations');
```
eszkadev, line 2140: don't have to be with `.uno`? like  in: https://github.com/CollaboraOnline/online/blob/f55c1a1403fcd350e81fdb98bd6c3a6b05cfda2b/browser/src/docdispatcher.ts#L538

lpranam: nop, `getItemValue` adds it if the command doesn't contain `.uno:` prefix.

eszkadev (approved):

PR #9888, opened 2024-08-23
backport: fix(mobile): reorder stylesheets

Minion3665:
As we insert other stylesheets after the head (and therefore later than
stylesheets in the head), they take precedence over stylesheets in the
head.

Therefore, we can't insert our device and branding stylesheets in the
end of the <head>. We need to insert them after the latest other
stylesheet.

This appears to a regression from https://github.com/CollaboraOnline/online/pull/9442

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -303,8 +303,12 @@ class InitializerBase {
 			brandingLink.setAttribute("href", this.brandingUriPrefix + theme_prefix + 'branding-desktop.css');
 		}
 
-		document.getElementsByTagName("head")[[0]].appendChild(link);
-		document.getElementsByTagName("head")[[0]].appendChild(brandingLink);
+		const otherStylesheets = document.querySelectorAll('link[rel="stylesheet"]');
+		const lastOtherStylesheet = otherStylesheets[otherStylesheets.length - 1];
+
+		lastOtherStylesheet
+			.insertAdjacentElement('afterend', link)
+			.insertAdjacentElement('afterend', brandingLink);
 	}
 
 	initializeViewMode() {


browser/js/global.js
```diff
@@ -303,8 +303,11 @@ class InitializerBase {
 			brandingLink.setAttribute("href", this.brandingUriPrefix + theme_prefix + 'branding-desktop.css');
 		}
 
-		document.getElementsByTagName("head")[[0]].appendChild(link);
-		document.getElementsByTagName("head")[[0]].appendChild(brandingLink);
+		const otherStylesheets = document.querySelectorAll('link[rel="stylesheet"]');
+		const lastOtherStylesheet = otherStylesheets[otherStylesheets.length - 1];
+
+		lastOtherStylesheet.insertAdjacentElement('afterend', brandingLink);
+		lastOtherStylesheet.insertAdjacentElement('afterend', link);
```
eszkadev, line 310: don't we want branding at the very end?

Minion3665: that's correct, I think perhaps I should change those statements to make it more obvious what was going on - while I was writing it I was iffy, now I'm even less sure it's readable

To explain...:
```
lastOtherStylesheet
```
(branding stylesheet gets inserted)
```
lastOtherStylesheet
brandingStylesheet
```
(device specific stylesheet gets inserted)
```
lastOtherStylesheet
deviceSpecificStylesheet
brandingStylesheet
```

notice how, by inserting after the child, we have flipped the stylesheets

...

contrast this with if we were inserting inside <head>, where we would always be inserting at the end, and you'll be able to see why the branding/device specific lines were flipped here

Minion3665: I've rewritten that to be clearer

eszkadev (approved):

PR #9889, opened 2024-08-23
indirection: expose server_name when provided

Rash419:
- this is useful for controller, collabora online vm setup
- usually when running the collabora online in k8s environment we pass POD_NAME env variable. POD_NAME is needed as a env variable because on every restart POD_NAME changes that's not the case for COOL running in the vms
- why don't use POD_NAME for VMs as well ? In container you only create one user and COOL process run under same user. If we set env variable in container it sets env varialble for that user only and COOL process can access it. For VMs, that's not the case we might have multiple users. so you need to set global env variable to make it. Instead of doing that I prefer providing the name inside coolwsd.xml

Change-Id: I92c204a0315c74ab121b5aa48795646b9dede39a

### DIFF
--- a/coolwsd.xml.in
+++ b/coolwsd.xml.in
@@ -325,6 +325,7 @@
       <geolocation_setup>
         <enable desc="Enable geolocation_setup when using indirection server with geolocation configuration" type="bool" default="false">false</enable>
       </geolocation_setup>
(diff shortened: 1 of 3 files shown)


quikee (approved):
looks fine

PR #9891, opened 2024-08-23
backport: fix(mobile): correct egregious dark UI colors

Minion3665:
This is a backport of #9890

Some of the coloring issues on mobile at the moment are truly egregious.
They leave dark mode text on a light mode background (or vice-versa),
leading to the text being entirely unreadable.

This happens because we have some colors hardcoded. Replacing them with
variables fixes this.

There are still some hardcoded colors after this change. I'm not sure
that this fixes everything, but it fixes the most obvious issues.

Change-Id: Ie1579751fb4154d1816dfd84becc726d06bc7876


eszkadev (approved):
looks ok

PR #9896, opened 2024-08-26
Related: cool#9735 kit: use the new registerAnyInputCallback() LOK API

vmiklos:
core.git has an Application::AnyInput() mechanism to inform idle jobs
when they should stop their work (and continue that at a later stage),
but this was not aware of pending online.git work. This lead to e.g.
performing all the idle layout after pasting into a numbered list in a
Writer document, instead of first rendering tiles for the visible area,
and only then doing the rest of the layout.

This is a problem for interactivity, e.g. the above scenario with 300
Writer pages results in a 274 ms hang till the first tile is rendered.

Fix the problem by using the new registerAnyInputCallback() LOK API, so
interested core.git code can ask our side if there are pending input
events.

The first naive implementation resulted in interrupting core.git idles
far too often, so apart from calling poll() on the KitSocketPoll, so
check if there are pending callbacks or tiles to be rendered, and only
report pending inputs in that case. This seems to result in an event
order that is similar to what happens for desktop Writer already.

Change-Id: I9a50d8b1ca59224084d5fa58f9ab29c9f29ef27b

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2873,6 +2873,24 @@ int pollCallback(void* pData, int timeoutUs)
 #endif
 }
 
+bool anyInputCallback(void* data)
+{
+    auto kitSocketPoll = reinterpret_cast<KitSocketPoll*>(data);
+    int ret = kitSocketPoll->poll(std::chrono::microseconds(0));
+    std::shared_ptr<Document> document = kitSocketPoll->getDocument();
+    if (document)
+    {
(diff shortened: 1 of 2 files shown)


vmiklos:

This is the final piece to actually use the infrastructure built on the core side, that paste scenario goes from a 963 ms lag to a 14 ms lag with this for me locally (core.git c025051b17a3851fc87ef334e16f603f13395f5a has the data).

caolanm (approved):

Issue #9749, opened 2024-08-08
Add toggle `Show menubar` icon option in compact view


Darshan-upadhyay1110:
In the top toolbar of compact view, there should be an option/icon to toggle the menubar. The {type: 'button', id: 'fold', img: 'fold'} from Control.TopToolbar.js could be used for this.
 
 Please check in version 23.05, it was like this in image below.
 
 
![image](https://github.com/user-attachments/assets/7e6f094c-a323-4f53-906c-58855c073c9a)

This option will help to toggle menubar.

Darshan-upadhyay1110:
Code pointers :

As I am to go through the code, i found some code pointers that can help here....

- Checkout the file Control.topToolbar.js file and find out why `fold` icon is not enabled ( hint : if you inspect in dev tools the `fold` icon element is there, but it had class `hidden`)
- add missing entry and all will start to work again 😉 

Also there can a new problem you will face after fixing this. (hint: overlapping images ), which is also easy to solve

CodingwithKetan:
Hi @Darshan-upadhyay1110, I'm interested in contributing to it. I wanted to check if anyone else is currently working on it. If not, could you please assign it to me?

PR #9900, opened 2024-08-26
toolbar: make fold btn visible in compact mode

aarsh21:
Change-Id: I703acfb96491e264f498cc0945ef5d052d4f46ae

* Resolves: #9749 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1016,7 +1016,7 @@ button.leaflet-control-search-next
 .w2ui-icon.link{ background: url('images/lc_inserthyperlink.svg') no-repeat center; }
 .w2ui-icon.insertsymbol{ background: url('images/lc_insertsymbol.svg') no-repeat center; }
 .w2ui-icon.edit{ background: url('images/lc_editdoc.svg') no-repeat center; }
-.w2ui-icon.fold{ background: url('images/fold.svg') no-repeat center center / 19px; opacity: 0.8; filter: alpha(opacity=80);}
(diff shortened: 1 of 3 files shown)


Darshan-upadhyay1110 (commented):
Hii @aarsh21 thanks for the fix. I have tested your patch it seems good.

But I saw a strange behavior there in icon. At the very first time when fold icon loads is looking different compare to what we have using CSS property.

is there any way you can direct change the url path of fold and not add or remove class?  is it possible ?

Let me know your findings here, other than that it's a good fix for now :)

CC: @eszkadev

aarsh21:
@Darshan-upadhyay1110 I can do that. but i thought it'd be still better to use unfold and fold classes.And path to fold icon were given wrong path but still it was displaying icon. That's still confusing.
- I will change the path directly in the image tag as you said. I will make them consistent.
- What if we just rotate the icon 180 degrees :D.  ( might be a dumbway ).

Darshan-upadhyay1110:
> @Darshan-upadhyay1110 I can do that. but i thought it'd be still better to use unfold and fold classes.And path to fold icon were given wrong path but still it was displaying icon. That's still confusing.
> 
> * I will change the path directly in the image tag as you said. I will make them consistent.
> * What if we just rotate the icon 180 degrees :D.  ( might be a dumbway ).

@pedropintosilva any thoughts on the last suggestion here ? Looks okay to me

Darshan-upadhyay1110 (approved):

pedropintosilva (approved):
Thank you!

pedropintosilva:
This seem quite safe to me merged

PR #9902, opened 2024-08-26
Use $host instead of $http_host

Glandos:
See https://trac.nginx.org/nginx/ticket/2468#comment:1

* Resolves: # 
* Target version: master 

### TODO

- The documentation on https://sdk.collaboraonline.com/docs/installation/Proxy_settings.html#reverse-proxy-with-nginx-webserver should be updated, but I wasn't able to find a repo for it.

### Checklist

### DIFF
--- a/etc/nginx/coolwsd.conf
+++ b/etc/nginx/coolwsd.conf
@@ -1,19 +1,19 @@
     # static files
     location ^~ /browser {
         proxy_pass http://localhost:9980;
-        proxy_set_header Host $http_host;
+        proxy_set_header Host $host;
     }
 
     # WOPI discovery URL
     location ^~ /hosting/discovery {
         proxy_pass http://localhost:9980;
-        proxy_set_header Host $http_host;
+        proxy_set_header Host $host;
     }
 
     # Capabilities
     location ^~ /hosting/capabilities {
         proxy_pass http://localhost:9980;
-        proxy_set_header Host $http_host;
+        proxy_set_header Host $host;
     }
 
     # main websocket
@@ -21,7 +21,7 @@
         proxy_pass http://localhost:9980;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "Upgrade";
-        proxy_set_header Host $http_host;
+        proxy_set_header Host $host;
         proxy_read_timeout 36000s;
     }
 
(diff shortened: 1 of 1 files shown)


meven:
If I understand correctly, this specifically fix HTTP/3 while having no downside/consequence for HTTP/1 or 2.

Correct @Glandos ?

Glandos:
You're right @meven 
I've been experimenting with HTTP/3 with nginx, and it seems that the `Host` header was not mandatory in HTTP/2, but still sent by most clients. Now that it doesn't exist anymore, the `:authority` pseudo-header is **not** copied as `Host` header, and `$http_host` only looks for this header, while `$host` uses the resolved value, whatever the protocol and header used:
- HTTP/1.1 will be resolved with `Host` header
- HTTP/2 will use `:authority` header, because `Host` header can be present but **must** be the same value
- HTTP/3 will use `:authority` because `Host` header isn't there at all

meven:
Could you add some context to your commit message:

nginx proxy example: Use $host instead of $http_host

Glandos:
What's the best way for you? Rewrite the branch with a new commit and push-force? Edit my first comment so that GitHub will make it available? Open a new PR?

meven:
> What's the best way for you? Rewrite the branch with a new commit and push-force? Edit my first comment so that GitHub will make it available? Open a new PR?

You can amend and force push.

meven:
I will be taking care of the documentation update.

timar:
@Glandos  Also please make sure that your [commits are signed-off](https://github.com/CollaboraOnline/online/blob/master/CONTRIBUTING.md#sign-your-work) and contain [Change-Id](https://github.com/CollaboraOnline/online/blob/master/.git-hooks/commit-msg#L82) (please install the git hook for that).

Glandos:
Hi,
Given the fact that I only used GitHub edition feature to replace a single variable name, it's going to be too complicated here for this simple contribution. It implies a `git clone`, installing hooks, amending, `Signed-Off-By`. Not that I don't know how to do it, but workflow is way too complex, I won't go through it for now, even if I don't exclude using it for a later change.

Hence, I release all my modifications and comments of this request (https://github.com/CollaboraOnline/online/pull/9902) to the public domain. You're allowed to remove my name/pseudonym of it too, since it can't even qualify as a creative work.

timar:
> Hi, Given the fact that I only used GitHub edition feature to replace a single variable name, it's going to be too complicated here for this simple contribution. It implies a `git clone`, installing hooks, amending, `Signed-Off-By`. Not that I don't know how to do it, but workflow is way too complex, I won't go through it for now, even if I don't exclude using it for a later change.
> 
> Hence, I release all my modifications and comments of this request (#9902) to the public domain. You're allowed to remove my name/pseudonym of it too, since it can't even qualify as a creative work.

Thanks! See cee77a8c57

PR #9903, opened 2024-08-26
calc: fix: named-range can't be used in function like SUM

Rash419:
Change-Id: Ia07c579b3bc81ce73f40e86e6e07a02098e7bf4c

* Target version: master 

- I tested the following with following partial test cases:
1. =SUM(te;te)
2. =SUM(te
3. =SUM(te; MAX(testrangename1;testrangename2)
4. = SUM(te; test1)

### DIFF
--- a/browser/src/control/Control.FormulaAutoCompletePopup.ts
+++ b/browser/src/control/Control.FormulaAutoCompletePopup.ts
@@ -60,15 +60,90 @@ class FormulaAutoCompletePopup extends L.Control.AutoCompletePopup {
 		return entries;
 	}
 
+	getAutocompleteText(
+		currentCellFormula: string,
+		functionName: string,
+		endIndex: number,
+	): string {
+		// Step-1: Find indexes of all the '(', ';', '-', '+', '*', '/'
+		const openBracketIndex: number[] = [];
+		const semicolonIndex: number[] = [];
+		const plusIndex: number[] = [];
+		const multiplyIndex: number[] = [];
+		const divideIndex: number[] = [];
+		const minusIndex: number[] = [];
+		const equalIndex: number = 0;
+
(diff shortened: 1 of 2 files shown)


gokaysatir (changes requested):
The second use of the named range seems to be wrong. Maybe we can use something like checking the chars from the end, one by one and then find the mismatching char etc?

Anyway, can you check the issue and if it's a real one, find another way, thanks :)

Rash419:
Hey @gokaysatir, its working perfectly fine for check this out:
[Uploading Screencast from 2024-08-30 15-56-18.webm…]()

Rash419:
@gokaysatir There are lot of cases your approach might not work as well
For example:
present partial formula: `=SUM(te; tes)`
functionName: `test4`
if user edit the first argument according to your approach checking from the last will  autocomplete on 2nd argument instead of first. I think we need think of different approach

gokaysatir:
> @gokaysatir There are lot of cases your approach might not work as well For example: present partial formula: `=SUM(te; tes)` functionName: `test4` if user edit the 1st argument according to your approach checking from the last will autocomplete on 2nd argument instead of 1st. I think we need think of different approach

Agreed, thanks for evaluating it :)

gokaysatir (approved):
Thank you for the improvements. There is also one more use case:
* User can set a cell's value to a constant, say "lambda" for an engineering formula.
* Then they can pick that cell value and add it. They will be free from writing its value every time.
* Can you check the image below?




Since this improves the infrastructure here and solves the issues partially, let me approve. We will need a follow up for the remaining things.

Thanks :)

gokaysatir (approved):
Works with commands like below:
=SUM(test;   test2) / (MAX(test3;10) + test3)

I think it's ready. Thanks @Rash419 :)

PR #9905, opened 2024-08-26
browser: accessibility: add label to combobox

hcvcastro:
Change-Id: Ib462aa34f450a84ad27c1a37ed554dff772c1c27


browser/src/map/handler/Map.StateChanges.js
```diff
@@ -3,13 +3,18 @@
...
 /*eslint no-extend-native:0*/
 L.Map.mergeOptions({
 	stateChangeHandler: true
 });
 
 L.Map.StateChangeHandler = L.Handler.extend({
+	Labels : {
+		'.uno:CharFontName' : _('Font:'),
+		'.uno:FontHeight' : _('Size:'),
+		'.uno:StyleApply' : _('Style:')
```
eszkadev, line 16: we usually try to avoid `:`

browser/src/map/handler/Map.StateChanges.js
```diff
@@ -3,13 +3,18 @@
...
  * LOK_CALLBACK_STATE_CHANGED callback
  */
-/* global $ app */
+/* global $ app _ */
 /*eslint no-extend-native:0*/
 L.Map.mergeOptions({
 	stateChangeHandler: true
 });
 
 L.Map.StateChangeHandler = L.Handler.extend({
```
eszkadev, line 12: StateChangeHandler seems to be not a perfect place for such things.
Let's keep it doing only one thing - storing status values for a command.
We use `_UNO('command')` for getting the translated "label" for uno command, which seems to already exist and do that correctly :) can we reuse it or there is some problem?

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -183,6 +183,12 @@ JSDialog.combobox = function (parentContainer, data, builder) {
 	var container = L.DomUtil.create('div', 'ui-combobox ' + builder.options.cssClass, parentContainer);
 	container.id = data.id;
 
+	var labelText = L.Map.StateChangeHandler.prototype.Labels[data.command];
```
eszkadev, line 186: in general - JSDialog widgets should be built only based on data available inside JSON.
For correct behavior on updates, etc. We need them to be stateless.

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -183,6 +183,12 @@ JSDialog.combobox = function (parentContainer, data, builder) {
...
+	if (labelText) {
+		var label = L.DomUtil.create('label', '', container);
+		label.textContent = labelText;
```
eszkadev, line 189: Why do we want that? combobox is the combobox field (input + dropdown).
If we want some label, we should put separate label widget next to it (in the .ui file in core, in the JSON).

browser/src/map/handler/Map.StateChanges.js
hcvcastro: Let me check, I did read it. It translate with hotkeys underscore I think.

browser/src/control/jsdialog/Widget.Combobox.js
hcvcastro: Occurs that the stateless does not have label info data, so If you read the task description,
I am doing a short path (hack) to resolve the problem.

browser/src/control/jsdialog/Widget.Combobox.js
hcvcastro: I did it, but for some reason, the parent container is removed and it messes up the combo box.
Somewhere it is not using well the combox which it is out of the scope of the task.

On the other hand, the label requires the LO Core .ui file changes that I am trying to avoid to do it.

browser/src/map/handler/Map.StateChanges.js
hcvcastro: Thanks

browser/src/control/jsdialog/Widget.Combobox.js
eszkadev: it's acctually what we should do (addin in .ui file).
Container can be replaced when we do some updates to the widget.
see `update` or `action` commands handling `executeAction` or `updateWidget` in JSDialogBuilder.
they work correctly as I know, but if you changed the structure here... action might not be able to change eg. text

hcvcastro: Ok, Please clarify in the task description that the changes also require changes in LibreOffice desktop accessibility.

eszkadev (changes requested):
marking as reviewed, we need input from @pedropintosilva

Darshan-upadhyay1110 (commented):
I think we can avoid adding label here.

Why ?
 
" If aria-label or aria-labelledby attributes are implemented correctly and provide clear, descriptive information about the dropdown's purpose, the controls should meet accessibility standards, even without visible labels."

We already have aria label associated with the input element here so i do not think wee need more for accessibility.

If i consider space to fit label here the i would suggest to only use aria-label and not add visible label for input field here...

hcvcastro:
> I think we can avoid adding label here.
> 
> Why ?
> 
> " If aria-label or aria-labelledby attributes are implemented correctly and provide clear, descriptive information about the dropdown's purpose, the controls should meet accessibility standards, even without visible labels."
> 
> We already have aria label associated with the input element here so i do not think wee need more for accessibility.
> 
> If i consider space to fit label here the i would suggest to only use aria-label and not add visible label for input field here...

The customer's request requires adding label visibility.

pedropintosilva:
I think we really need this change as in the toolbar, there are two fields used to select the font type and font size. These fields do not have visible labels. Although the toolbar follows common patterns seen elsewhere and users can usually recognize the purpose of the fields, visible labels should still be provided to avoid accessibility barriers.

but there are style issues and missing css classes. I have rebased this commit and add another one see the new PR https://github.com/CollaboraOnline/online/pull/14669

And what about the ui files topic? as in

> it's acctually what we should do (addin in .ui file).
> said @eszkadev

PR #9910, opened 2024-08-27
wsd: no need to create existing mount-point

Ashod:
**Backported**

The mount-point must exist already, since
we always mount read-only. There is no
directory to create, but the subsequent
chmod always fails, since the mounted
directory is read-only. Since we started
warning when chmod fails recently,
6208b37a327266c583f47cdd13948d57ecfcce05,
these previously-silent failures became
noisy.

Change-Id: I9efaa89182c016e9a7a5d36cc3da5bfa7ee599c2
(cherry picked from commit c1ced10f17d1190f259030d93b8002aee177ce6b)


caolanm (approved):
sounds reasonable to me for backport

Issue #9833, opened 2024-08-16
Handle limited open Connections due to keepalive connections (cool#9621)


sgothel:
### Is your feature request related to a problem?

Issue #9621 resolved by pull request #9777 keeps connections alive to enhance performance.
This may result in a shortage of available connection on the server and should be addressed

### Describe the solution you'd like

We should track inactivity of each socket and close the oldest-inactive sockets
either (a) in general and/or (b) when available connection count is below a threshold.

### Describe alternatives you've considered

Alternatively, we may need to close high-frequency short lived connections.

mmeeks:
Yep - clearly rather related, and important to get this nailed before widely deploying the connection re-using code =)

While we're there - we should shut prolly also detect and shut connections that are not making progress - 'slow' connections and so on that are drip-feeding us data.

Thanks!

PR #9916, opened 2024-08-28
Handle limited open Connections due to keepalive connections

sgothel:
* Resolves: #9833
* Target version: master 

### Summary
    
- ProtocolHandlerInterface::checkTimeout(..)
  is now called via StreamSocket::checkRemoval()
  to ensure tests even w/o input data (was: handlePoll())
    
- ProtocolHandlerInterface::checkTimeout(..)
  - Add bool return value: true -> shutdown connection, caller shall stop processing
  - Implemented for http::Session
    - Timeout (30s) net::Defaults::HTTPTimeout with missing response
  - Implemented for WebSocketHandler
    - Timeout (2s = net::Defaults::WSPingTimeout)
      after missing WS native frame ping/pong (server only)
    
- StreamSocket -> Socket (properties moved)
  - bytes sent/received
  - closed state
    
- Socket (added properties)
  - creation- and last-seen -time
  - socket type and port
  - checkRemoval(..)
    - called directly from SocketPoll::poll()
    - only for IPv4/v6 network connections
    - similar to ProtocolHandlerInterface::checkTimeout(..)
    - added further criteria (age, throughput, ..)
      - Timeout (64s = net::Defaults::SocketPollTimeout)
        if (now - lastSeen) > timeout
      - Timeout (net::Defaults::MinBytesPerSec)
        if Throughput < MinBytesPerSec (disabled by default)
      - TODO: Add maximimal IPv4/IPv6 socket-count criteria, drop oldest.
      - age via Socket::creationTime is currently not used as requested
    
- SocketPoll::poll()
  - Additionally erases if !socket->isOpen() || socket->checkRemoval()

#### Changes to original PR
- net::Config -> net::Defaults, without any xml config file side-effects
  - only allow net::Defaults to be changed in unit tests via dedicated callback
- removed Socket age criteria via Socket::creationTime and net::defaults::MaxDuration
  - Socket::creationTime still relevant for minimum throughput criteria
  - dropped the net::defaults::MaxDuration 
- max-connections
  - removed under- overflow checks
  - using SC atomic ops (not just relaxed)
- enhanced API doc of Util::TimeAverage (clarification)
- unit tests in one place / commit
- more unification/squashing to single semantic commits
- more sensible order of commits
- removed dead code
- dropped noexcept patch

Further:
- cleaned up the commit messages
- removed the COOLWSD.cpp default config reordering
- reordered the 'Unify socket stats', as it must be upfront the actual connection count limitation
  and before moved _lastSeenHTTPHeader
- fixed another intermediate breakage in the commit chain
- bisectable proof (but no unit test runs ofc) .. via
  - passed `git rebase -i --exec "make -j 20"` for all commits
- Use scoped enums with efficient types (STATE_ENUM)
  - now using desired STATE_ENUM

Unit tests:   +7 files, 879 insertions(+), remaining changed lines +800, -222

### Checklist
- [X] Passed certain unit tests and cypress desktop annotations.js

### DIFF
--- a/net/DelaySocket.cpp
+++ b/net/DelaySocket.cpp
@@ -23,9 +23,12 @@ std::once_flag Delay::DelayPollOnceFlag;
 /// Reads from fd, delays that and then writes to _dest.
 class DelaySocket : public Socket {
     int _delayMs;
-    enum State { ReadWrite,      // normal socket
-                 EofFlushWrites, // finish up writes and close
-                 Closed };
+    STATE_ENUM(State,
+               ReadWrite,      // normal socket
+               EofFlushWrites, // finish up writes and close
+               Closed
+              );
+
     State _state;
     std::shared_ptr<DelaySocket> _dest; // our writing twin.
 
@@ -50,7 +53,7 @@ class DelaySocket : public Socket {
 public:
     DelaySocket(int delayMs, int fd) :
         Socket (fd, Socket::Type::Unix), _delayMs(delayMs),
-        _state(ReadWrite)
+        _state(State::ReadWrite)
 	{
 //        setSocketBufferSize(Socket::DefaultSendBufferSize);
 	}
@@ -104,30 +107,30 @@ public:
     {
         switch (newState)
         {
-        case ReadWrite:
+        case State::ReadWrite:
             assert (false);
             break;
-        case EofFlushWrites:
-            assert (_state == ReadWrite);
+        case State::EofFlushWrites:
+            assert (_state == State::ReadWrite);
             assert (_dest);
             _dest->pushCloseChunk();
             _dest = nullptr;
             break;
-        case Closed:
-            if (_dest && _state == ReadWrite)
+        case State::Closed:
+            if (_dest && _state == State::ReadWrite)
                 _dest->pushCloseChunk();
             _dest = nullptr;
             shutdown();
             break;
         }
-        DELAY_LOG('#' << getFD() << " changed to state " << newState << '\n');
+        DELAY_LOG('#' << getFD() << " changed to state " << toStringShort(newState) << '\n');
         _state = newState;
     }
 
     void handlePoll(SocketDisposition &disposition,
                     std::chrono::steady_clock::time_point now, int events) override
     {
-        if (_state == ReadWrite && (events & POLLIN))
+        if (_state == State::ReadWrite && (events & POLLIN))
         {
             auto chunk = std::make_shared<WriteChunk>(_delayMs);
 
@@ -139,7 +142,7 @@ public:
             } while (len < 0 && errno == EINTR);
 
             if (len == 0) // EOF.
-                changeState(EofFlushWrites);
+                changeState(State::EofFlushWrites);
             else if (len >= 0)
             {
                 DELAY_LOG('#' << getFD() << " read " << len
@@ -153,14 +156,14 @@ public:
             else if (errno != EAGAIN && errno != EWOULDBLOCK)
             {
                 DELAY_LOG('#' << getFD() << " error : " << Util::symbolicErrno(errno) << ": " << strerror(errno) << '\n');
-                changeState(Closed); // FIXME - propagate the error ?
+                changeState(State::Closed); // FIXME - propagate the error ?
             }
         }
 
         if (_chunks.empty())
         {
-            if (_state == EofFlushWrites)
-                changeState(Closed);
+            if (_state == State::EofFlushWrites)
+                changeState(State::Closed);
         }
         else // Write if we have delayed enough.
         {
@@ -171,7 +174,7 @@ public:
                 if (chunk->getData().empty())
                 { // delayed error or close
                     DELAY_LOG('#' << getFD() << " handling delayed close\n");
-                    changeState(Closed);
+                    changeState(State::Closed);
                 }
                 else
                 {
@@ -193,7 +196,7 @@ public:
                                       << chunk->getData().size()
                                       << " queue: " << _chunks.size() << " error: "
                                       << Util::symbolicErrno(errno) << ": " << strerror(errno) << '\n');
-                            changeState(Closed);
+                            changeState(State::Closed);
                         }
                     }
                     else
@@ -214,10 +217,10 @@ public:
         if (events & (POLLERR | POLLHUP | POLLNVAL))
         {
             DELAY_LOG('#' << getFD() << " error events: " << events << '\n');
-            changeState(Closed);
+            changeState(State::Closed);
         }
 
-        if (_state == Closed)
+        if (_state == State::Closed)
             disposition.setClosed();
     }
 };
--- a/net/DelaySocket.cpp
+++ b/net/DelaySocket.cpp
@@ -51,8 +51,8 @@ class DelaySocket : public Socket {
 
     std::vector<std::shared_ptr<WriteChunk>> _chunks;
 public:
-    DelaySocket(int delayMs, int fd) :
-        Socket (fd, Socket::Type::Unix), _delayMs(delayMs),
+    DelaySocket(int delayMs, int fd, std::chrono::steady_clock::time_point creationTime) :
+        Socket(fd, Socket::Type::Unix, creationTime), _delayMs(delayMs),
         _state(State::ReadWrite)
 	{
 //        setSocketBufferSize(Socket::DefaultSendBufferSize);
@@ -260,8 +260,9 @@ int Delay::create(int delayMs, int physicalFd)
         int internalFd = pair[0];
         int delayFd = pair[1];
 
-        auto physical = std::make_shared<DelaySocket>(delayMs, physicalFd);
-        auto internal = std::make_shared<DelaySocket>(delayMs, internalFd);
+        const std::chrono::steady_clock::time_point now = std::chrono::steady_clock::now();
+        auto physical = std::make_shared<DelaySocket>(delayMs, physicalFd, now);
+        auto internal = std::make_shared<DelaySocket>(delayMs, internalFd, now);
         physical->setDestination(internal);
         internal->setDestination(physical);
 
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1434,7 +1434,8 @@ private:
         while (!_response->done())
         {
             const auto now = std::chrono::steady_clock::now();
-            checkTimeout(now);
+            if (checkTimeout(now))
+                return false;
 
             const auto remaining =
                 std::chrono::duration_cast<std::chrono::microseconds>(deadline - now);
@@ -1694,17 +1695,21 @@ private:
         net::asyncConnect(_host, _port, isSecure(), shared_from_this(), pushConnectCompleteToPoll);
     }
 
-    void checkTimeout(std::chrono::steady_clock::time_point now) override
+    bool checkTimeout(std::chrono::steady_clock::time_point now) override
     {
         if (!_response || _response->done())
-            return;
+            return false;
 
+        const std::chrono::microseconds timeout = getTimeout();
         const auto duration =
             std::chrono::duration_cast<std::chrono::milliseconds>(now - _startTime);
-        if (now < _startTime || duration > getTimeout() || SigUtil::getTerminationFlag())
+
+        if (now < _startTime ||
+            (timeout > std::chrono::microseconds::zero() && duration > timeout) ||
+            SigUtil::getTerminationFlag())
         {
-            LOG_WRN("Timed out while requesting [" << _request.getVerb() << ' ' << _host
-                                                   << _request.getUrl() << "] after " << duration);
+            LOG_WRN("CheckTimeout: Timeout while requesting [" << _request.getVerb() << ' ' << _host
+                                                               << _request.getUrl() << "] after " << duration);
 
             // Flag that we timed out.
             _response->timeout();
@@ -1714,7 +1719,9 @@ private:
             // no good maintaining a poor connection (if that's the issue).
             onDisconnect(); // Trigger manually (why wait for poll to do it?).
             assert(isConnected() == false);
+            return true;
         }
+        return false;
     }
 
     int sendTextMessage(const char*, const size_t, bool) const override { return 0; }
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1199,9 +1199,9 @@ public:
     }
 
     /// Returns the default timeout.
-    static constexpr std::chrono::milliseconds getDefaultTimeout()
+    static std::chrono::milliseconds getDefaultTimeout()
     {
-        return std::chrono::seconds(30);
+        return std::chrono::duration_cast<std::chrono::milliseconds>( net::Defaults::get().HTTPTimeout );
     }
 
     /// Returns the current protocol scheme.
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -153,6 +153,8 @@ STATE_ENUM(FieldParseState,
 /// See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
 enum class StatusCode : unsigned
 {
+    None = 0, // Undefined status (unknown)
+
     // Informational
     Continue = 100,
     SwitchingProtocols = 101,
--- a/net/NetUtil.hpp
+++ b/net/NetUtil.hpp
@@ -39,6 +39,8 @@ public:
     /// http::Session timeout in us (30s default). Zero disables metric.
     std::chrono::microseconds HTTPTimeout;
 
+    /// Maximum total connections (9999 or MAX_CONNECTIONS). Zero disables metric.
+    size_t MaxConnections;
     /// Socket minimum bits per seconds throughput (0). Zero disables metric.
     double MinBytesPerSec;
 
@@ -50,6 +52,7 @@ private:
         : WSPingTimeout(std::chrono::microseconds(2000000))
         , WSPingPeriod(std::chrono::microseconds(3000000))
         , HTTPTimeout(std::chrono::microseconds(30000000))
+        , MaxConnections(9999)
         , MinBytesPerSec(0.0)
         , SocketPollTimeout(std::chrono::microseconds(64000000))
     {
--- a/net/NetUtil.hpp
+++ b/net/NetUtil.hpp
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <chrono>
 #include <functional>
 #include <string>
 #include <memory>
@@ -27,6 +28,43 @@ struct sockaddr;
 
 namespace net
 {
+
+class Defaults
+{
+public:
+    /// WebSocketHandler ping timeout in us (2s default). Zero disables metric.
+    std::chrono::microseconds WSPingTimeout;
+    /// WebSocketHandler ping period in us (3s default), i.e. duration until next ping. Zero disables metric.
+    std::chrono::microseconds WSPingPeriod;
+    /// http::Session timeout in us (30s default). Zero disables metric.
+    std::chrono::microseconds HTTPTimeout;
+
+    /// Socket minimum bits per seconds throughput (0). Zero disables metric.
+    double MinBytesPerSec;
+
+    /// Socket poll timeout in us (64s), useful to increase for debugging.
+    std::chrono::microseconds SocketPollTimeout;
+
+private:
+    Defaults()
+        : WSPingTimeout(std::chrono::microseconds(2000000))
+        , WSPingPeriod(std::chrono::microseconds(3000000))
+        , HTTPTimeout(std::chrono::microseconds(30000000))
+        , MinBytesPerSec(0.0)
+        , SocketPollTimeout(std::chrono::microseconds(64000000))
+    {
+    }
+
+public:
+    Defaults(const Defaults&) = delete;
+    Defaults(Defaults&&) = delete;
+
+    static Defaults& get()
+    {
+        static Defaults def;
+        return def;
+    }
+};
 
 class HostEntry
 {
--- a/net/ServerSocket.hpp
+++ b/net/ServerSocket.hpp
@@ -39,7 +39,7 @@ public:
     }
 
     /// Control access to a bound TCP socket
-    enum Type { Local, Public };
+    STATE_ENUM(Type, Local, Public);
 
     /// Create a new server socket - accepted sockets will be added
     /// to the @clientSockets' poll when created with @factory.
--- a/net/ServerSocket.hpp
+++ b/net/ServerSocket.hpp
@@ -28,8 +28,10 @@ public:
 class ServerSocket : public Socket
 {
 public:
-    ServerSocket(Socket::Type type, SocketPoll& clientPoller, std::shared_ptr<SocketFactory> sockFactory) :
-        Socket(type),
+    ServerSocket(Socket::Type type,
+                 std::chrono::steady_clock::time_point creationTime,
+                 SocketPoll& clientPoller, std::shared_ptr<SocketFactory> sockFactory) :
+        Socket(type, creationTime),
 #if !MOBILEAPP
         _type(type),
 #endif
@@ -43,11 +45,14 @@ public:
 
     /// Create a new server socket - accepted sockets will be added
     /// to the @clientSockets' poll when created with @factory.
-    static std::shared_ptr<ServerSocket> create(ServerSocket::Type type, int port,
-                                                Socket::Type socketType, SocketPoll& clientSocket,
+    static std::shared_ptr<ServerSocket> create(ServerSocket::Type type,
+                                                int port,
+                                                Socket::Type socketType,
+                                                std::chrono::steady_clock::time_point creationTime,
+                                                SocketPoll& clientSocket,
                                                 std::shared_ptr<SocketFactory> factory)
     {
-        auto serverSocket = std::make_shared<ServerSocket>(socketType, clientSocket, std::move(factory));
+        auto serverSocket = std::make_shared<ServerSocket>(socketType, creationTime, clientSocket, std::move(factory));
 
         if (serverSocket && serverSocket->bind(type, port) && serverSocket->listen())
             return serverSocket;
@@ -130,8 +135,9 @@ private:
 class LocalServerSocket : public ServerSocket
 {
 public:
-    LocalServerSocket(SocketPoll& clientPoller, std::shared_ptr<SocketFactory> sockFactory) :
-        ServerSocket(Socket::Type::Unix, clientPoller, std::move(sockFactory))
+    LocalServerSocket(std::chrono::steady_clock::time_point creationTime,
+                      SocketPoll& clientPoller, std::shared_ptr<SocketFactory> sockFactory) :
+        ServerSocket(Socket::Type::Unix, creationTime, clientPoller, std::move(sockFactory))
     {
     }
     ~LocalServerSocket() override;
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -556,6 +556,14 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                 // removed in a callback
                 ++itemsErased;
             }
+            else if (!_pollSockets[i]->isOpen())
+            {
+                // closed socket ..
+                ++itemsErased;
+                LOGA_TRC(Socket, '#' << _pollFds[i].fd << ": Removing socket (at " << i
+                         << " of " << _pollSockets.size() << ") from " << _name);
+                _pollSockets[i] = nullptr;
+            }
             else if (_pollFds[i].fd == _pollSockets[i]->getFD())
             {
                 SocketDisposition disposition(_pollSockets[i]);
@@ -575,7 +583,7 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                     rc = -1;
                 }
 
-                if (!disposition.isContinue())
+                if (!_pollSockets[i]->isOpen() || !disposition.isContinue())
                 {
                     ++itemsErased;
                     LOGA_TRC(Socket, '#' << _pollFds[i].fd << ": Removing socket (at " << i
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -634,6 +634,17 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                          << " of " << _pollSockets.size() << ") from " << _name);
                 _pollSockets[i] = nullptr;
             }
+            else if( _pollSockets[i]->checkRemoval(newNow) )
+            {
+                // timed out socket .. also checks
+                // ProtocolHandlerInterface
+                // - http::Session::checkTimeout() OK
+                // - WebSocketHandler::checkTimeout() OK
+                ++itemsErased;
+                LOGA_TRC(Socket, '#' << _pollFds[i].fd << ": Removing socket (at " << i
+                         << " of " << _pollSockets.size() << ") from " << _name);
+                _pollSockets[i] = nullptr;
+            }
             else if (_pollFds[i].fd == _pollSockets[i]->getFD())
             {
                 SocketDisposition disposition(_pollSockets[i]);
@@ -1003,7 +1014,8 @@ void WebSocketHandler::dumpState(std::ostream& os, const std::string& /*indent*/
 {
     os << (_shuttingDown ? "shutd " : "alive ");
 #if !MOBILEAPP
-    os << std::setw(5) << _pingTimeUs/1000. << "ms ";
+    os << std::setw(5) << _pingMicroS.last()/1000. << "ms, avg "
+       << _pingMicroS.average()/1000. << "ms ";
 #endif
     if (_wsPayload.size() > 0)
         Util::dumpHex(os, _wsPayload, "\t\tws queued payload:\n", "\t\t");
@@ -1403,6 +1415,64 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
         os << clientAddress() << ":" << clientPort();
     }
     return os << "]";
+}
+
+bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
+{
+    if (!isIPType()) // forced removal on outside-facing IPv[46] network connections only
+        return false;
+
+    const auto durTotal =
+        std::chrono::duration_cast<std::chrono::milliseconds>(now - getCreationTime());
+    const auto durLast =
+        std::chrono::duration_cast<std::chrono::milliseconds>(now - getLastSeenTime());
+    const double bytesPerSecIn = durTotal.count() > 0 ? (double)bytesRcvd() / ((double)durTotal.count() / 1000.0) : 0.0;
+    /// TO Criteria: Violate creation time? (invalid passed now)
+    const bool cNow = now < getCreationTime();
+    /// TO Criteria: Violate maximum idle (_pollTimeout default 64s)
+    const bool cIDLE = _pollTimeout > std::chrono::microseconds::zero() &&
+                       durLast > _pollTimeout;
+    /// TO Criteria: Violate minimum bytes-per-sec throughput? (_minBytesPerSec default 0, disabled)
+    const bool cMinThroughput = _minBytesPerSec > std::numeric_limits<double>::epsilon() &&
+                                bytesPerSecIn > std::numeric_limits<double>::epsilon() &&
+                                bytesPerSecIn < _minBytesPerSec;
+    /// TO Criteria: Shall terminate?
+    const bool cTermination = SigUtil::getTerminationFlag();
+    if (cNow || cIDLE || cMinThroughput || cTermination )
+    {
+        LOG_WRN("CheckRemoval: Timeout: {Now " << cNow << ", IDLE " << cIDLE
+                << ", MinThroughput " << cMinThroughput << ", Termination " << cTermination << "}, "
+                << getStatsString(now) << ", "
+                << *this);
+        if (_socketHandler)
+        {
+            _socketHandler->onDisconnect();
+            if( isOpen() ) {
+                // Note: Ensure proper semantics of onDisconnect()
+                LOG_WRN("Socket still open post onDisconnect(), forced shutdown.");
+                shutdown(); // signal
+                closeConnection(); // real -> setClosed()
+                assert(isOpen() == false); // should have issued shutdown
+            }
+        }
+        else
+        {
+            shutdown(); // signal
+            closeConnection(); // real -> setClosed()
+            assert(isOpen() == false); // should have issued shutdown
+        }
+        return true;
+    }
+    else if (_socketHandler && _socketHandler->checkTimeout(now))
+    {
+        assert(isOpen() == false); // should have issued shutdown
+        setClosed();
+        LOG_WRN("CheckRemoval: Timeout: " << getStatsString(now) << ", " << *this);
+        return true;
+    } else {
+        LOG_DBG("CheckRemoval: Test " << getStatsString(now) << ", " << *this);
+    }
+    return false;
 }
 
 #if !MOBILEAPP
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1354,6 +1354,36 @@ LocalServerSocket::~LocalServerSocket()
 #  define LOG_CHUNK(X)
 #endif
 
+#endif // !MOBILEAPP
+
+std::string StreamSocket::toString(WSState t)
+{
+    if( WSState::WS == t )
+    {
+        return "WS";
+    }
+    return "HTTP";
+}
+
+std::ostream& StreamSocket::stream(std::ostream& os) const
+{
+    os << "StreamSocket[#" << getFD()
+       << ", " << toString(_wsState)
+       << ", " << Socket::toString(type())
+       << " @ ";
+    if (Type::IPv6 == type())
+    {
+        os << "[" << clientAddress() << "]:" << clientPort();
+    }
+    else
+    {
+        os << clientAddress() << ":" << clientPort();
+    }
+    return os << "]";
+}
+
+#if !MOBILEAPP
+
 bool StreamSocket::parseHeader(const char *clientName,
                                Poco::MemoryInputStream &message,
                                Poco::Net::HTTPRequest &request,
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -99,6 +99,41 @@ int Socket::createSocket([[maybe_unused]] Socket::Type type)
     return fakeSocketSocket();
 }
 
+std::ostream& Socket::streamStats(std::ostream& os, const std::chrono::steady_clock::time_point &now) const
+{
+    const auto durTotal = std::chrono::duration_cast<std::chrono::milliseconds>(now - _creationTime);
+    const auto durLast = std::chrono::duration_cast<std::chrono::milliseconds>(now - _lastSeenTime);
+
+    float kBpsIn, kBpsOut;
+    if (durTotal.count() > 0)
+    {
+        kBpsIn = (float)_bytesRcvd / (float)durTotal.count();
+        kBpsOut = (float)_bytesSent / (float)durTotal.count();
+    }
+    else
+    {
+        kBpsIn = (float)_bytesRcvd / 1000.0f;
+        kBpsOut = (float)_bytesSent / 1000.0f;
+    }
+
+    const std::streamsize p = os.precision();
+    os.precision(1);
+    os << "Stats[dur[total "
+        << durTotal.count() << "ms, last "
+        << durLast.count() << " ms], kBps[in "
+        << kBpsIn << ", out " << kBpsOut
+        << "]]";
+    os.precision(p);
+    return os;
+}
+
+std::string Socket::getStatsString(const std::chrono::steady_clock::time_point &now) const
+{
+    std::ostringstream oss;
+    streamStats(oss, now);
+    return oss.str();
+}
+
 std::ostream& Socket::streamImpl(std::ostream& os) const
 {
     os << "Socket[#" << getFD()
@@ -986,8 +1021,8 @@ void StreamSocket::dumpState(std::ostream& os)
     const int events = getPollEvents(std::chrono::steady_clock::now(), timeoutMaxMicroS);
     os << '\t' << std::setw(6) << getFD() << "\t0x" << std::hex << events << std::dec << '\t'
        << (ignoringInput() ? "ignore\t" : "process\t") << std::setw(6) << _inBuffer.size() << '\t'
-       << std::setw(6) << _outBuffer.size() << '\t' << " r: " << std::setw(6) << _bytesRecvd
-       << "\t w: " << std::setw(6) << _bytesSent << '\t' << clientAddress() << '\t';
+       << std::setw(6) << _outBuffer.size() << '\t' << " r: " << std::setw(6) << bytesRcvd()
+       << "\t w: " << std::setw(6) << bytesSent() << '\t' << clientAddress() << '\t';
     _socketHandler->dumpState(os);
     if (_inBuffer.size() > 0)
         Util::dumpHex(os, _inBuffer, "\t\tinBuffer:\n", "\t\t");
@@ -1613,7 +1648,7 @@ bool StreamSocket::compactChunks(MessageMap& map)
 bool StreamSocket::sniffSSL() const
 {
     // Only sniffing the first bytes of a socket.
-    if (_bytesSent > 0 || _bytesRecvd != _inBuffer.size() || _bytesRecvd < 6)
+    if (bytesSent() > 0 || bytesRcvd() != _inBuffer.size() || bytesRcvd() < 6)
         return false;
 
     // 0x0000  16 03 01 02 00 01 00 01
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -15,6 +15,7 @@
 #include "TraceEvent.hpp"
 #include "Util.hpp"
 
+#include <atomic>
 #include <chrono>
 #include <cstring>
 #include <cctype>
@@ -60,6 +61,25 @@ std::atomic<bool> SocketPoll::InhibitThreadChecks(false);
 std::atomic<bool> Socket::InhibitThreadChecks(false);
 
 std::unique_ptr<Watchdog> SocketPoll::PollWatchdog;
+
+std::mutex SocketPoll::StatsMutex;
+std::atomic<size_t> SocketPoll::StatsConnectionCount(0);
+
+size_t SocketPoll::StatsConnectionMod(size_t added, size_t removed) {
+    if( added == 0 && removed == 0 ) {
+        return GetStatsConnectionCount();
+    }
+    size_t res, pre;
+    {
+        std::lock_guard<std::mutex> lock(StatsMutex);
+        pre = GetStatsConnectionCount();
+        res = pre + added; // overflow impossible due to MAX_CONNECTIONS
+        res -= removed; // underflow impossible due to MAX_CONNECTIONS
+        StatsConnectionCount.store(res, std::memory_order_seq_cst);
+    }
+    LOG_DBG("SocketPoll::ConnectionCount: " << pre << " +" << added << " -" << removed << " = " << res);
+    return res;
+}
 
 #define SOCKET_ABSTRACT_UNIX_NAME "0coolwsd-"
 
@@ -293,6 +313,8 @@ namespace {
 SocketPoll::SocketPoll(std::string threadName)
     : _name(std::move(threadName)),
       _pollTimeout( net::Defaults::get().SocketPollTimeout ),
+      _limitedConnections( false ),
+      _connectionLimit( 0 ),
       _pollStartIndex(0),
       _stop(false),
       _threadStarted(0),
@@ -494,6 +516,8 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
     // The events to poll on change each spin of the loop.
     setupPollFds(now, timeoutMaxMicroS);
     const size_t size = _pollSockets.size();
+    size_t itemsAdded = 0;
+    size_t itemsErased = 0;
 
     // disable watchdog - it's good to sleep
     disableWatchdog();
@@ -546,7 +570,29 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         std::vector<CallbackFn> invoke;
         {
             std::lock_guard<std::mutex> lock(_mutex);
+            const size_t newConnCount = _newSockets.size();
+            const size_t globCount = GetStatsConnectionCount();
 
+            if( _limitedConnections &&
+                _connectionLimit > 0 &&
+                globCount + newConnCount > _connectionLimit)
+            {
+                // For now we simply drop new connections
+                const size_t overhead = globCount + newConnCount - _connectionLimit;
+                for(size_t i=0; i<overhead; ++i)
+                {
+                    std::shared_ptr<Socket>& socket = _newSockets.back(); // oldest
+                    assert(socket);
+
+                    LOG_WRN("Limiter: #" << socket->getFD() << ": Removing "
+                             << (i+1) << " / " << overhead << " new socket of (pre "
+                             << globCount << " + new " << newConnCount << ") / max " << _connectionLimit
+                             << " from " << _name);
+
+                    socket->resetThreadOwner();
+                    _newSockets.pop_back();
+                }
+            }
             if (!_newSockets.empty())
             {
                 LOGA_TRC(Socket, "Inserting " << _newSockets.size() << " new sockets after the existing "
@@ -558,6 +604,8 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
                 // Copy the new sockets over and clear.
                 _pollSockets.insert(_pollSockets.end(), _newSockets.begin(), _newSockets.end());
+
+                itemsAdded += _newSockets.size();
 
                 _newSockets.clear();
             }
@@ -592,10 +640,15 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         }
     }
 
-    if (_pollSockets.size() != size)
+    if (itemsAdded != _pollSockets.size() - size)
     {
-        LOG_TRC("PollSocket container size has changed from " << size << " to "
-                                                              << _pollSockets.size());
+        // unexpected
+        LOG_WRN("PollSocket container size has changed from " << size
+            << " + " << itemsAdded << " to " << _pollSockets.size());
+    } else if (itemsAdded > 0)
+    {
+        LOG_TRC("PollSocket container size increased from " << size
+            << " + " << itemsAdded << " to " << _pollSockets.size());
     }
 
     // If we had sockets to process.
@@ -612,7 +665,6 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
         if (_pollStartIndex > size - 1)
             _pollStartIndex = 0;
 
-        size_t itemsErased = 0;
         size_t i = _pollStartIndex;
         for (std::size_t j = 0; j < size; ++j)
         {
@@ -691,15 +743,35 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
         if (itemsErased)
         {
-            LOG_TRC("Scanning to removing " << itemsErased << " defunct sockets from "
-                    << _pollSockets.size() << " sockets");
+            const size_t itemsErasedPre = itemsErased;
+            itemsErased = 0; // correcting itemsErased
 
             _pollSockets.erase(
                 std::remove_if(_pollSockets.begin(), _pollSockets.end(),
-                    [](const std::shared_ptr<Socket>& s)->bool
-                    { return !s; }),
+                               [&itemsErased](const std::shared_ptr<Socket>& s) -> bool
+                               {
+                                   if (!s)
+                                   {
+                                       ++itemsErased;
+                                       return true;
+                                   }
+                                   else
+                                   {
+                                       return false;
+                                   }
+                               }),
                 _pollSockets.end());
+
+            LOG_TRC("Removed " << itemsErased
+                    << "(" << itemsErasedPre << ") defunct sockets from "
+                    << _pollSockets.size() << " sockets");
         }
+    }
+    if( _limitedConnections )
+    {
+        // Perform bookkeeping if required
+        // New connections might be dropped if exceeding limits, see _newSockets above.
+        StatsConnectionMod(itemsAdded, itemsErased);
     }
 
     return rc;
@@ -807,9 +879,11 @@ void SocketPoll::createWakeups()
 void SocketPoll::removeSockets()
 {
     LOG_DBG("Removing all " << _pollSockets.size() + _newSockets.size()
-                            << " sockets from SocketPoll thread " << _name);
+                            << " sockets from SocketPoll thread " << _name
+                            << " of " << GetStatsConnectionCount() << " total poll sockets");
     ASSERT_CORRECT_SOCKET_THREAD(this);
 
+    size_t removedPollSockets = 0;
     while (!_pollSockets.empty())
     {
         const std::shared_ptr<Socket>& socket = _pollSockets.back();
@@ -820,6 +894,10 @@ void SocketPoll::removeSockets()
         socket->resetThreadOwner();
 
         _pollSockets.pop_back();
+        ++removedPollSockets;
+    }
+    if( _limitedConnections ) {
+        StatsConnectionMod(0, removedPollSockets);
     }
 
     while (!_newSockets.empty())
@@ -1087,7 +1165,9 @@ void SocketPoll::dumpState(std::ostream& os) const
     const auto pollSockets = _pollSockets;
 
     os << "\n  SocketPoll [" << name() << "] with " << pollSockets.size() << " socket"
-       << (pollSockets.size() == 1 ? "" : "s") << " - wakeup rfd: " << _wakeup[0]
+       << (pollSockets.size() == 1 ? "" : "s")
+       << " of " << GetStatsConnectionCount()
+       << " total - wakeup rfd: " << _wakeup[0]
        << " wfd: " << _wakeup[1] << '\n';
     const auto callbacks = _newCallbacks.size();
     if (callbacks > 0)
@@ -1452,15 +1532,14 @@ bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
                 LOG_WRN("Socket still open post onDisconnect(), forced shutdown.");
                 shutdown(); // signal
                 closeConnection(); // real -> setClosed()
-                assert(isOpen() == false); // should have issued shutdown
             }
         }
         else
         {
             shutdown(); // signal
             closeConnection(); // real -> setClosed()
-            assert(isOpen() == false); // should have issued shutdown
         }
+        assert(isOpen() == false); // should have issued shutdown
         return true;
     }
     else if (_socketHandler && _socketHandler->checkTimeout(now))
@@ -1469,8 +1548,6 @@ bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
         setClosed();
         LOG_WRN("CheckRemoval: Timeout: " << getStatsString(now) << ", " << *this);
         return true;
-    } else {
-        LOG_DBG("CheckRemoval: Test " << getStatsString(now) << ", " << *this);
     }
     return false;
 }
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -56,11 +56,6 @@
 #include <Watchdog.hpp>
 #include <wasm/base64.hpp>
 
-// Bug in pre C++17 where static constexpr must be defined. Fixed in C++17.
-constexpr std::chrono::microseconds SocketPoll::DefaultPollTimeoutMicroS;
-constexpr std::chrono::microseconds WebSocketHandler::InitialPingDelayMicroS;
-constexpr std::chrono::microseconds WebSocketHandler::PingFrequencyMicroS;
-
 std::atomic<bool> SocketPoll::InhibitThreadChecks(false);
 std::atomic<bool> Socket::InhibitThreadChecks(false);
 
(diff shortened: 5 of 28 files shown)


sgothel:
This change set created a linkage error (android), the ['usual' type_info, vtable thing](https://cpci.cbg.collabora.co.uk:8080/job/github_android_app_co-24.04/2725/console). 
Can't figure a missed 'virtual' or even pure virtual '=0' *virtual key* in the code, also reviewed the MOBILEAPP macro.
Hence I need to reproduce and build the android locally here ..  WIP.

net/Socket.cpp
```diff
@@ -1298,6 +1385,52 @@ LocalServerSocket::~LocalServerSocket()
 #  define LOG_CHUNK(X)
 #endif
 
+bool StreamSocket::checkForcedRemoval(std::chrono::steady_clock::time_point now) noexcept
```
caolanm, line 1388: might need to move this impl to after 

"#endif // !MOBILEAPP" (if mobile needs it) and/or put ifdef MOBILEAPP around it in the header if it doesn't or a !MOBILE stub, something like that anyway

sgothel: yeah. Great eyes. Thx!

net/HttpRequest.hpp
```diff
@@ -1201,7 +1201,7 @@ class Session final : public ProtocolHandlerInterface
     /// Returns the default timeout.
     static constexpr std::chrono::milliseconds getDefaultTimeout()
     {
-        return std::chrono::seconds(30);
+        return std::chrono::seconds(30); // FIXME?
```
Ashod, line 1204: This is the default. There is nothing wrong with it. Please see `StorageConnectionManager::getHttpSession()` for how this is overrided.

net/HttpRequest.hpp
```diff
@@ -1692,14 +1695,18 @@ class Session final : public ProtocolHandlerInterface
         net::asyncConnect(_host, _port, isSecure(), shared_from_this(), pushConnectCompleteToPoll);
     }
 
-    void checkTimeout(std::chrono::steady_clock::time_point now) override
+    bool checkTimeout(std::chrono::steady_clock::time_point now) override
```
Ashod, line 1698: Why is this needed at all? This function changes the state of the and fires up the necessary callbacks (see `_response->timeout();` below). We can easily check the state and act accordingly. And anyway, the loop above (line 1435) will be broken.

net/HttpRequest.hpp
```diff
@@ -1432,7 +1432,10 @@ class Session final : public ProtocolHandlerInterface
         while (!_response->done())
         {
             const auto now = std::chrono::steady_clock::now();
-            checkTimeout(now);
+            if (checkTimeout(now))
```
Ashod, line 1435: When we timeout, the state of `_response` changes at line 1715, and the `_response->done()` at line 1432 returns true. If this is not the case, we should fix *that* issue.

Why do we care to break the loop here, is that to avoid the extra poll at 1442?

net/ServerSocket.hpp
```diff
@@ -134,10 +134,10 @@ class LocalServerSocket : public ServerSocket
         ServerSocket(Socket::Type::Unix, clientPoller, std::move(sockFactory))
     {
     }
-    ~LocalServerSocket();
+    ~LocalServerSocket() override;
 
-    virtual bool bind(Type, int) override { assert(false); return false; }
```
Ashod, line 139: I don't think having `virtual` hurts. If anything, it's easier to notice virtual functions. Having `override` of course *helps*, so that's fine.

net/Socket.cpp
```diff
@@ -651,7 +738,7 @@ void SocketPoll::takeSocket(const std::shared_ptr<SocketPoll> &fromPoll,
     ASSERT_CORRECT_THREAD();
 
     // hold a reference during transfer
-    std::shared_ptr<Socket> socket = inSocket;
+    std::shared_ptr<Socket> socket = inSocket; // NOLINT
```
Ashod, line 741: What issue is `NOLINT` fixing?

net/Socket.hpp
```diff
@@ -395,9 +466,17 @@ class Socket
 #endif
     }
 
-private:
     std::string _clientAddress;
+    const Type _type;
+    unsigned int _clientPort;
     const int _fd;
+    /// True if this socket is open.
+    std::atomic_bool _open; // FIXME: cache sync (atomic) required?
```
Ashod, line 474: We already have `_closed`, do we really need `_open` ?

net/WebSocketHandler.hpp
```diff
@@ -221,8 +226,8 @@ class WebSocketHandler : public ProtocolHandlerInterface
 
     void shutdown(bool goingAway, const std::string &statusMessage) override
     {
-        shutdown(goingAway ? WebSocketHandler::StatusCodes::ENDPOINT_GOING_AWAY :
-                 WebSocketHandler::StatusCodes::NORMAL_CLOSE, statusMessage);
+        shutdownImpl(goingAway ? WebSocketHandler::StatusCodes::ENDPOINT_GOING_AWAY :
+                     WebSocketHandler::StatusCodes::NORMAL_CLOSE, statusMessage, false, false);
```
Ashod, line 230: Commented argument-names is always welcome when we have arbitrary arguments like `false`. It helps readability.

net/WebSocketHandler.hpp
```diff
@@ -241,21 +246,52 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
+
+    /// Don't wait for the remote Websocket to handshake with us; go down fast.
+    void shutdownAfterWriting()
+    {
+        shutdownImpl(WebSocketHandler::StatusCodes::NORMAL_CLOSE, std::string(),
+                     true /* hard async shutdown & close */, false);
+    }
+
+    /// Returns true if the underlying socket is connected.
+    bool isConnected() const {
```
Ashod, line 261: Formatting is all out of shape.

net/WebSocketHandler.hpp
```diff
@@ -241,21 +246,52 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
+        std::shared_ptr<StreamSocket> socket = _socket.lock();
+        return nullptr != socket && !socket->isClosed();
```
Ashod, line 263: Please, please, don't do reverse comparisons. It's very unnatural and an eye-sore. 

`return socket && !socket->isClosed();` OR `return socket != nullptr && !socket->isClosed();`

net/WebSocketHandler.hpp
```diff
@@ -241,21 +246,52 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
-            LOGA_TRC(WebSocket, "Shutdown: Closing Connection");
-            if (!_shuttingDown)
+            const bool silent = _shuttingDown || silentShutdown;
+            LOGA_TRC(WebSocket, "Shutdown: Closing Connection " << (silent ? "(silent)" : "(coop)"));
+            if (!silent)
+            {
                 sendCloseFrame(statusCode, statusMessage);
-            socket->closeConnection();
+            }
+            socket->shutdown();
```
Ashod, line 285: This is confusing. What does `closeConnection()` do, and what does `shutdown()` do? Do we really need this distinction?

test/UnitSocketLifecycle.cpp
```diff
@@ -0,0 +1,263 @@
...
+            TST_LOG("Response: " << response->header().toString());
+            TST_LOG("Response size: " << testname << "[" << sockIdx << "]: `" << documentURL << "`: " << response->header().getContentLength());
+            LOK_ASSERT_EQUAL(http::StatusCode::OK, response->statusCode());
+            LOK_ASSERT_EQUAL(true, sessions[sockIdx]->isConnected());
+            LOK_ASSERT(http::Header::ConnectionToken::None ==
+                       response->header().getConnectionToken());
+            LOK_ASSERT(0 < response->header().getContentLength());
+        }
+    }
+    catch (const Poco::Exception& exc)
```
Ashod, line 88: We shouldn't have Poco code here, but perhaps we still do?

test/UnitSocketLifecycle.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+        LOK_ASSERT_FAIL(exc.displayText());
+    }
+    for (std::shared_ptr<http::Session>& session : sessions)
+    {
+        LOK_ASSERT_EQUAL(true, session->isConnected());
+    }
+    TST_LOG("Test: XXX " << testname << " w/ " << sockCount << " connections, sleep...");
+    std::this_thread::sleep_for(std::chrono::seconds(4));
```
Ashod, line 97: Looks like an anti-pattern, especially that we don't check anything afterwards. If something is checked elsewhere, a comment is helpful. Also, the amount of sleep should depend on something that affects the test, so needs to be a variable so we don't have to change this arbitrarily when the test fails or when another parameter changes (at least we shouldn't change it here).

test/WebSocketSession.hpp
```diff
@@ -56,9 +56,9 @@ class WebSocketSession final : public WebSocketHandler
         return result;
     }
 
-    WebSocketSession(const std::string& hostname, Protocol protocolType, int portNumber)
+    WebSocketSession(std::string && hostname, Protocol protocolType, int portNumber)
```
Ashod, line 59: Why not take a straight copy (`std::string hostname`)? At least that's the common pattern I'm familiar with (and used extensively in the code). Is there an advantage to taking r-value reference?

common/Util.cpp
```diff
@@ -727,6 +731,44 @@ namespace Util
 #endif
     }
 
+    std::string vformatString(const char* format, va_list ap) noexcept
```
Ashod, line 734: `noexcept` ? I don't understand.

net/Socket.cpp
```diff
@@ -86,9 +102,57 @@ int Socket::createSocket([[maybe_unused]] Socket::Type type)
...
+    {
+        kBpsIn = (float)_bytesRcvd / (float)durTotal.count();
+        kBpsOut = (float)_bytesSent / (float)durTotal.count();
+    }
+    else
+    {
+        kBpsIn = (float)_bytesRcvd / 1000.0f;
+        kBpsOut = (float)_bytesSent / 1000.0f;
+    }
+    return Util::formatString("Socket[#%d, dur[total %" PRIi64 "ms, last %" PRIi64
```
Ashod, line 135: This seems overkill and in the wrong direction. Surely we don't need (vs)printf-style formatting. In this particular case, we can have a helper that creates this string with the proper formatting using `stringstream` or some such.

What's the motivation behind writing `vformatString()` at all?

Ashod (commented):
Thanks, @sgothel. There are some really good stuff here!

I understand this is a draft, but it reads like a mix of random fixes and improvements (some of it is good and useful, like adding `const` and `= default`, `make_shared`, etc.), re-inventing methods for doing things already supported, and debugging/tracing support for your understanding (to be removed).

It's very hard to read, as there is no clear theme or purpose.

I left various comments, but I didn't dive too deep in a number of areas because it was unclear why we are either replacing an existing approach, or inventing a new one when we can improve on the existing one (where there are shortcomings).

At a minimum, I'd separate the backtrace improvements from the socket improvements, and likely the misc. improvements from new code/features/tests.

Finally, I'm troubled by `vformatString` and the re-working a lot of the socket states and internals with a huge set of new members that change the existing semantics with unclear benefits.

Perhaps there is some context I'm missing. Perhaps this was agreed on and I'm unaware.

sgothel:
Thank you @Ashod for the review.

> I understand this is a draft, but it reads like a mix of random fixes and improvements (some of it is good and useful, like adding `const` and `= default`, `make_shared`, etc.), re-inventing methods for doing things already supported, and debugging/tracing support for your understanding (to be removed).

The draft status was motivated by my intention to discuss the actual semantics and goals, 
i.e. mitigate limited connections - or better, ensure connections are closed as well as
to reduce the risk of running out of sockets/file-handles.

Which brings us to the next point I guess.
> 
> It's very hard to read, as there is no clear theme or purpose.

When reading commit by commit, I believe everything is well split apart semantically.
I believed the purpose was well described in #9833 and the commit https://github.com/CollaboraOnline/online/pull/9916/commits/3ef99c511d168b1a9a9bb6c86d4a857a584e91e4
actually starting to address (matching this drafts' description) was meaningful.

Indeed, I added a few unrelated single commits into this review. Understanding that we allow
reasonable small changes w/o issue report, esp of intrinsic nature, I believe it should be OK.
(admin vs workflow compromise as discussed earlier)

> 
> I left various comments, but I didn't dive too deep in a number of areas because it was unclear why we are either replacing an existing approach, or inventing a new one when we can improve on the existing one (where there are shortcomings).

I am not aware of really replacing anything. I merely moved and added functionality.

> 
> At a minimum, I'd separate the backtrace improvements from the socket improvements, and likely the misc. improvements from new code/features/tests.
> 
As described above, a zero side-effect change in its own commit.

The Backtrace object in particular is useful for all, as we can produce an instance
capturing the current stack, move it around and output where desired.
As discussed, I did not add `addr2line` or `libunwind` dependencies, which I understand ofc.

> Finally, I'm troubled by `vformatString` and the re-working a lot of the socket states and internals with a huge set of new members that change the existing semantics with unclear benefits.

I just like printf formatting as it is more performant and flexible (at least to me).
IMHO I use it in a most reasonable save fashion, while being supported by the compiler to 
remove the risk of wrong format specifiers and stack values. The latter is still a weakness, which I am aware of.

> 
> Perhaps there is some context I'm missing. Perhaps this was agreed on and I'm unaware.

Allow me to continue answering your other remarks below.

net/HttpRequest.hpp
sgothel: Thank you. 
Indeed, I simply marked all timeout values with FIXME to review them.
Goal is to allow us to make them configurable (clumsy word) @ runtime, 
i.e. at least via setter at runtime and/or via a config file (optional).
This is required to allow me to shape the unit tests, triggering timeouts and hence proving functionality.
You mentioned `WopiTestServer::configure()` and I will look into it.

net/HttpRequest.hpp
sgothel: In general I prefer a clear response from a method, here true for timeout in checkTimeout() -> also changed API doc.

Usually this method is called in Socket::handlePoll() where I prefer to bail out quick.
In general I prefer `fail fast` and `bail out early` to help reducing complexity. However, I understand that there might be more side effects and `fail fast` often stands against `keep going`. Hence this review is important, thx!

The particular case of http::Session::syncRequestImpl()  .. se below .. probably needs to change indeed, great catch.

net/HttpRequest.hpp
sgothel: I have to change this particular case in http::Session::syncRequestImpl()  - great catch!

net/ServerSocket.hpp
sgothel: `override` is indeed [important](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html), i.e. signature changes/spelling/.. Not using the (redundant) `virtual` in the `virtual override` is IMHO reasonable and more safe as one could miss the more important `override` specifier itself.

net/Socket.cpp
sgothel: Fixes a warning from `clang-tidy`, noting that a copy might not be needed here (performance, prefer reference).
However, since we probably pass this off into an async case - *and* I didn't not completely review the code, I marked it to silent clang-tidy. 

In case the `socket` reference gets *copied* into the async-callback lambda capture, the shared_ptr counter would be increased as required and we could just use a reference for above `socket`. But as I wrote, I didn't double-check yet.

test/UnitSocketLifecycle.cpp
sgothel: The unit test is the most unfinished part .. 
So indeed, I might have copy-pasta the POCO exception w/o even using it :)

test/UnitSocketLifecycle.cpp
sgothel: See above .. a truly WIP unit test.
When I have timeout configurable .. I will use this to trigger timeouts and then naturally check results.

test/WebSocketSession.hpp
sgothel: Yeah .. this is a specific private ctor only, so moving the string is sufficient.
Also just our test tool. I would have thought longer about it, if public API.
For the latter, I am sure I have seen other occurrences .. and avoiding copy (elision if super fine though) is good.

common/Util.cpp
sgothel: std::string::resize() as used in the implementation is not `noexcept`, i.e. can throw. 
Hence `noexcept` needs to go here indeed. Great catch.

net/Socket.cpp
sgothel: quoting myself ..
> I just like printf formatting as it is more performant and flexible (at least to me).
> IMHO I use it in a most reasonable save fashion, while being supported by the compiler to
> remove the risk of wrong format specifiers and stack values. The latter is still a weakness, which I am aware of.

TBH .. probably more a personal style. If the project rejects the `printf` formatting in general, I will drop it ofc.

net/HttpRequest.hpp
sgothel: On a 2nd look, syncRequestImpl() can only return false if checkTimeout() returns true, i.e. closed the socket.
poller.poll() can't succeed here anymore and remaining should be zero.
Hence leaving the loop and function w/o further side-effects is OK IMHO.

net/Socket.hpp
sgothel: Moved StreamSocket::_closed -> Socket::_open, as described in commit message (besides other properties).
This move helps Socket::checkForcedRemoval() and SocketPoll::poll() has required properties provided.

Granted, SocketPoll::poll() only process StreamSocket objects ..

The isClosed() or isOpen() as well as setClosed() utilization now finds its endpoint in the Socket base class,
where Socket::shutdown() also ensures setClosed().

net/WebSocketHandler.hpp
sgothel: OK. Yes, `bool` in particular is also dangerous (cast, change signature). Glad this is only a private implementation method. 
Adding the comments .. these days I am split about it as toolchains do that. But OK.

net/WebSocketHandler.hpp
sgothel: Yeah, as it was all out of shape (I just moved the methods into same-access-group).
Reminder - for now, while we don't want to break old style necessarily ... and try to be nice formatting wise,
we have no strict rule.

net/WebSocketHandler.hpp
sgothel: While not always doing it, it is a good way to avoid accidental assignment and showing the r-value on the left side. 
So being trained this way, it is natural. However, I also agree that compiler and tools to warn against such accidents. 
Hmm .. but will keep your preference in mind. No guarantees.

net/WebSocketHandler.hpp
sgothel: That was my 1st thought reading the Socket code as well.

However, this change fixes its usage and realigns with other cases,
i.e. overridden StreamSocket::shutdown() only does the async-signalling (intention)
while StreamSocket::closeConnection() does the Socket::Shutdown() ..
I surely don't like it, just played along. 

I am more than willing to propose a cleanup in an upcoming commit post this issue 9833

caolanm:
I think it'll be easier to manage this if its split up so that the mostly uncontroversial misc cleanups of existing stuff where no change in logic is intended, the make_unique, make_shared, privator ctor -> = delete, constifying, expansion of auto& to real type, emplace_back, etc goes into a separate pr.

The backtrace debugging aid looks probably useful to have to me, but can also be its own pr I think (and probably not worth fighting the "CodeQL scanning - action / Analyze" step which apparently can't handle the constexpr use which is blocking ci from passing this pr).

And generally trim this pr down to the core change to make it easier to think about it

net/WebSocketHandler.hpp
Ashod: We write once, read many, many times.

Double checking what we write is preferable to slowing down all readers just so we won't make accidental mistakes. I understand why some make such suggestions to beginners (they want to help them avoid bugs). As a teacher, I rather teach to improve quality and skills, rather than tricks to avoid sloppy coding. That's not helping the beginner to get any better (having fewer bugs doesn't make one better).

But the burden on the reader is huge. It forces the reader to remember what is being compared against before knowing what the subject of the comparison is. So when scanning code (something we do all the time, far more than typing code), we can't quickly trace the variables we are interested in, and we have to stop and reason about `0 != value` and `nullptr != pointer` etc. to decide 1) do I care about this statement, 2) which variable is involved? 3) what is it doing?

There is a reason each language has idiomatic style; it makes it easier to read code quickly and more accurately. That's why we use `!` for negation, instead of `== false`, for example.

Ashod:
Thanks for the responses, @sgothel! We've discussed some offline, so didn't reply here (e.g. printf-style formatting being fast--in fact it's very slow compared to only serialization).

> Indeed, I added a few unrelated single commits into this review. Understanding that we allow reasonable small changes w/o issue report, esp of intrinsic nature, I believe it should be OK. (admin vs workflow compromise as discussed earlier)

In general, agreed. But I'll echo @caolanm's comment here that breaking up into PRs, that can be decided to merge independent of other parts, is always helpful.

I should also point out that this PR is well above a 1000 lines. This is a rare size. Considering that some parts change the interface of a large hierarchy (the sockets), which is a fundamental core part of the whole project and very delicate (can easily break things catastrophically), it requires reading and studying far more code (i.e. the context and consequences) to review properly. An isolated and low-risk change of a few 100 lines is fine. A complex few 100 line change (such as in the sockets) is challenging and time consuming. Sometimes we have no choice because we can't break it up. But a 1000 lines that can be easily broken up is hardly necessary.

Finally, even though we can review individual commits, it's preferable to have a single theme to a PR. Helps the reviewer. In some cases a few misc. changes can go together (comment fixes, cosmetics, formatting, logging, etc.). But mixing functional changes with non-functional changes dilutes the process and can reduce the quality of the review. More so with high-risk changes (i.e. socket code) mixed with debugging code (i.e. backtracing).

Regardless, you have a lot of great improvements here, and I don't want to undermine the importance of your contribution!
Thank you.

sgothel:
Thank you both for your review and kind words.

I have split up single commits into 4 branches, i.e. this pull-request branch and
- #9952 
- #9953 
- #9954 

For this feature work, I need to complete the unit test incl. the timeout configuration.
Looks OK .. shall provide it in 1-2 days.

net/NetUtil.hpp
```diff
@@ -28,6 +32,49 @@ struct sockaddr;
...
+    {
+    }
+
+public:
+    Defaults(const Defaults&) = delete;
+    Defaults(Defaults&&) = delete;
+
+    static Defaults& get()
+    {
+        static Defaults def;
```
Ashod, line 73: Default values are compile-time constants (i.e. `constexpr`). This is a pessimization and overkill for something trivial (there is a hidden lock here, btw... overkill, as I say).

And why would we want to support changing default values?! Best to keep these concepts apart: defaults are compile-time const, and runtime values are class members (variables). Unless the suggestion is to have defaults to initialize these defaults!

Please see common/Common.hpp for other default values and how we use them.

net/Socket.cpp
Ashod: I'd suggest a different take. Let's make the change when/if we check and have an answer. Silencing a tool buries a potentially valid warning. What's the use of that if we haven't checked its validity one way or the other?

Silencing it *is* a decision, and implies someone checked and made a decision. But, if that's not the case, then this is misleading and will make it that much harder to remember and revisit this. Best to leave it alone until we have the time to investigate. Only then we can decide to silence the tool or fix the issue (if it's an issue).

net/WebSocketHandler.hpp
```diff
@@ -599,48 +639,89 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
-            return;
+        {
+            return false;
+        }
 
-        const auto timeSincePingMicroS
-            = std::chrono::duration_cast<std::chrono::microseconds>(now - _lastPingSentTime);
-        if (timeSincePingMicroS >= PingFrequencyMicroS)
+        if (_openPingCount > 0 && _pingTimeout.count() > std::numeric_limits<double>::epsilon() &&
+            _pingMicroS.average() >= _pingTimeout.count())
```
Ashod, line 705: I'm not sure I follow why the ping timeout needs to figure out an average.

Timeouts are absolute. Ping timeout is no different.

If the idea is to have some tolerance, the traditional/typical way of doing that is simple: have a soft limit and a hard limit. Easiest and most common implementation considers the first timeout to be the soft limit, issues a warning, and sends another ping. If that second ping also times out (i.e. it has been 2x timeout-duration since last message from the peer), then the hard limit is reached and the connection is interrupted.

If there is a deeper explanation why we need such complexity, a detailed comment is warranted. But a simple soft/hard limit timeout is sufficient (in fact, one can argue that averages, weighted or not, are very problematic and highly unpredictable).

test/UnitTimeoutNone.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+#include <config.h>
+
+#include <memory>
+#include <string>
+
+#include <HttpRequest.hpp>
+#include <Socket.hpp>
+
+#include <Poco/DOM/DOMParser.h>
```
Ashod, line 20: Safe to assume we don't need XML parsing for ping timeout tests.

Also, this test framework is not recommended for this use-case. See UnitWSD and family for event-based tests.

It is possible to use these tests for this use-case, but they are far harder to debug and fragile in the face of even ordering. Which is why we use UnitBase, UnitWSD, etc.

test/UnitTimeoutSocket.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+#include <config.h>
+
+#include <memory>
+#include <string>
+
+#include <HttpRequest.hpp>
+#include <Socket.hpp>
+
+#include <Poco/DOM/DOMParser.h>
```
Ashod, line 20: These tests are event-based and the preferable approach.

Still, probably no need for XML parsers here.

mmeeks (commented):
As said needs to be split up; lets try to get the less controversial pieces in and merged as separate pull requests - to slim this down =) Can you work on that Sven ?

wsd/COOLWSD.cpp
```diff
@@ -2353,6 +2361,30 @@ void COOLWSD::innerInitialize(Application& self)
     // Allow UT to manipulate before using configuration values.
     UnitWSD::get().configure(conf);
 
+    // net::Defaults
+    {
+        typedef int int_t;
```
Ashod, line 2366: Seem pedantic. If we must, we can use `int64_t` and be done. This covers potentially large values. Other types are virtually impossible to be used with integral config values coming as strings.

wsd/ProxyProtocol.hpp
```diff
@@ -40,10 +40,6 @@ class ProxyProtocolHandler : public ProtocolHandlerInterface
     int getPollEvents(std::chrono::steady_clock::time_point /* now */,
                       int64_t &/* timeoutMaxMs */) override;
 
-    void checkTimeout(std::chrono::steady_clock::time_point /* now */) override
```
Ashod, line 43: Could it be that the author of this intentionally wanted to suppress a default behavior?

I say this because testing the proxy is not commonly done. And unless I can test the proxy, I wouldn't touch it.

sgothel:
> As said needs to be split up; lets try to get the less controversial pieces in and merged as separate pull requests - to slim this down =) Can you work on that Sven ?

will do the split later then (tomorrow) + finishing the tests

wsd/COOLWSD.cpp
sgothel: a left over .. b/c int64_t didn't work

test/UnitTimeoutNone.cpp
sgothel: didn't clean the test up indeed (includes, exception-catch), will do when finishing the test

net/WebSocketHandler.hpp
sgothel: Time weighted values (here for the ping-duration) is most stable here, since we don't like to pull the plug on a 1-2 off hit. It allows to use the integrated value over time.
Also checked for a more simple exponential or mixed (last one with more weight) .. while that is better than just the last value, it still could give a disconnect criteria for 'just a bad few seconds'.
All this assuming that a false disconnect is to be avoided.

I couldn't produce good test values with firefox, but saw a range of 20-120us even locally (6x). 
Ranges we expects are ~1ms - 2s AFAIK.

sgothel:
> @sgothel, there were multiple suggestions to split this large PR up to reduce its size. That was when it had 18 commits and was "only" a 1000 lines. Now it's past 1500 lines and is 1 commit! I'm sensing this isn't going in the direction of the feedback.
> 
> What's the plan, please?

As I wrote, I will split out the config and TimeAverage (~400 lines or so) and put the remainder at the bottom -> 3 commits within this pull request. Perhaps more, w/ a fresh look. 

But splitting this up in multiple pull request (PR) makes would probably makes no sense, as these pieces are required and each PR must be tested. IMHO.

mmeeks:
> As I wrote, I will split out the config and TimeAverage (~400 lines or so) and put the remainder at the bottom -> 3 commits within this pull request. Perhaps more, w/ a fresh look.
> 
> But splitting this up in multiple pull request (PR) makes would probably makes no sense, as these pieces are required and each PR must be tested. IMHO.

Multiple pull requests means we can merge non-controversial stuff fast - and have it being tested in the wild - ideally of course in the red release. It's a definite anti-pattern to build up a huge, blocking change-request that is all or nothing. There is/was lots of other 'stuff' in this that doesn't belong in one CR - IIRC; cosmetic stuff - "remove virtual from lots of functions" - that lower risk stuff should be got in separately, and should not be repeatedly presented to reviewers - it just chews time to skip over. So - please split this into separate CRs that we can merge - non-controversial / pre-requisite pieces, and other bits on top.

Thanks.

sgothel:
I have cut out pull #10012 (which should be merged with this one IMHO).

Also produced smaller commits, which hopefully enables easier review of intend/story.
Since the 4 commits of this pull request (PR) have no real meaning beyond this issue,
allow me to offer these within this PR.

As mentioned Friday, I like to add one more criteria and test it before wrapping things up (-> maxConnections).

wsd/ProxyProtocol.hpp
sgothel: ProxyProtocolHandler -> ProtocolHandlerInterface, and the default is a NOP with return false.

net/Socket.cpp
```diff
@@ -550,6 +619,25 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                 // removed in a callback
                 ++itemsErased;
             }
+            else if (!_pollSockets[i]->isOpen())
```
mmeeks, line 622: Ok - so; this chunk alone seems plausible, but potentially controversial / regression causing. Can it be a separate commit - so it is bisectable; ideally along with a clear rationale for that ?

net/Socket.cpp
```diff
@@ -569,7 +650,7 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                     rc = -1;
                 }
 
-                if (!disposition.isContinue())
+                if (!_pollSockets[i]->isOpen() || !disposition.isContinue())
```
mmeeks, line 660: Given the above comparison - we expect handlePoll to close the socket, but not set the disposition correctly ? what is the use-case for that; should we just assert that if it isContinue then it must be isOpen instead ?

net/Socket.cpp
```diff
@@ -1302,6 +1383,33 @@ LocalServerSocket::~LocalServerSocket()
 #  define LOG_CHUNK(X)
 #endif
 
+#endif // !MOBILEAPP
+
+std::string StreamSocket::toString(WSState t) noexcept
```
mmeeks, line 1396: Would love this non-controversial useful fluff to be in a separate commit to behavioural change. It's really good to improve our state dumping though :-)

net/Socket.cpp
```diff
@@ -1302,6 +1383,33 @@ LocalServerSocket::~LocalServerSocket()
...
+{
+    if( WSState::WS == t )
```
mmeeks, line 1398: Not a fan of vertical space wasteage myself, it's a non-renewable resource; Ash is a huge fan for some reason =) either way - good to fit in with the surrounding code - which (I hope) is an "modern editors help you, but don't make your screen taller" mode ;-)

net/Socket.cpp
```diff
@@ -1543,7 +1651,7 @@ bool StreamSocket::compactChunks(MessageMap& map)
 bool StreamSocket::sniffSSL() const
 {
     // Only sniffing the first bytes of a socket.
-    if (_bytesSent > 0 || _bytesRecvd != _inBuffer.size() || _bytesRecvd < 6)
+    if (getBytesSent() > 0 || getBytesRcvd() != _inBuffer.size() || getBytesRcvd() < 6)
```
mmeeks, line 1715: The accessor-ization is something that should be done in a separate commit - if we want that. Its far from clear to me that it helps readability - prwt. performance _bytesRecvd' I can compile in my head, but an accessor could do anything and lives somewhere else.

Also - it seems we have publicised lots of our previously private members by adding accessors to access them all; is that necessary? not a huge fan of puncturing encapsulation that keeps implementation details visible & local. We can of course make the accessors private too - but then - what value did we add here at what cost ?

net/Socket.hpp
```diff
@@ -362,26 +410,47 @@ class Socket
...
+    /// \return >= 0 for a successfully created socket, -1 on error
+    static int createSocket(Type type);
+
+    std::string toStringImpl() const noexcept;
+
+    void init()
     {
-        if (type != Type::Unix)
+        if (_type != Type::Unix)
+        {
```
mmeeks, line 447: Again changing this in lots of places is not my favourite change =)

net/Socket.hpp
```diff
@@ -362,26 +410,47 @@ class Socket
     /// Construct based on an existing socket fd.
     /// Used by accept() only.
     Socket(const int fd, Type type)
-        : _fd(fd)
+        : _type(type),
+          _clientPort(0),
+          _fd(fd),
+          _open(_fd >= 0),
+          _creationTime(std::chrono::steady_clock::now()),
```
mmeeks, line 418: now() has some cost - we should where possible reduce its usage and pass around the 'now' from the poll that accepted the socket etc.

I'm somewhat nervous that we initialize things here rather than in init. There are two socket constructors, and using 'init' is (cf. the comment above) a way of not duplicating code & state setup between them. Please move as much shared code as possible into init; we could pass 'type' and 'fd' as parameters to init too here nicely.

net/Socket.hpp
```diff
@@ -395,9 +464,17 @@ class Socket
 #endif
     }
 
-private:
     std::string _clientAddress;
+    const Type _type;
+    unsigned int _clientPort;
     const int _fd;
+    /// True if this socket is open.
+    bool _open;
```
mmeeks, line 473: We need to care about size here. Can we use bit-fields to pack _type, _clientPort, and _open into the same space ?

net/WebSocketHandler.hpp
```diff
@@ -241,21 +242,52 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
+            socket->shutdown();
             socket->ignoreInput();
             assert(socket->getInBuffer().empty() &&
                    "Socket buffer must be empty after ignoreInput");
 
-            // force close after writing this message
+            // force close after writing this message (real shutdown)
             if (hardShutdown)
-                socket->shutdown();
+            {
```
mmeeks, line 294: ditto.

mmeeks (changes requested):
So - some good things here; but Socket* is the foundation of our I/O work.
As such - expect lots of scrutiny and review here, far more than elsewhere.

Again I would like to separate the cosmetic re-working of things from functional changes.

I'd like each functional change to be its own separate thing. Eg. silent-shutdown - great feature,non-controversial, we need it - pull it out into its own commit. Lets get that in early.

Pushing down the bytesSent etc. statistics from StreamSocket -> Socket - interesting, not sure of the rationale - perhaps harmless, but would like to know why - where does this simplify something else ?

And of course, other things - not so convinced about ;-)

Thanks for persisting and iterating! - we will get there.

net/Socket.cpp
sgothel: Earmark: Removed the `NOLINT`, however - another patch may remove the explicit copy as it is being copied in the lambda capture. For full replication, the method itself shall receive a copy of the shared_ptr.
But will not touch this here.

net/Socket.cpp
sgothel: `!_pollSockets[i]->isOpen()` is redundant indeed. Iff always `disposition.isContinue() == _pollSockets[i]->isOpen()` (as it probably should be aligned) then the whole block is redundant.
However, I am not sure whether it always gets closed with the disposition in all cases. So I would prefer to keep it in or analyze better..

net/Socket.cpp
sgothel: this is in its single commit, compile clean on its own in this patch series, i.e. https://github.com/CollaboraOnline/online/pull/9916/commits/7db79302b7a4ea310c4defe3009e6818d43c3dc5
hence it is bisectable.

net/Socket.cpp
sgothel: [my personal clang-format](https://jausoft.com/cgit/jaulib.git/tree/.clang-format) looks different indeed, but not here to make enemies on my first weeks (yet) .. so I double the line count and clean my glasses ;-)

net/Socket.cpp
sgothel: only here since these stats moved up to the parent class Socket and I wanted to clarify mutable access, i.e. ease review and having those fields private. Yes, the verb 'get' implies a most simple thing .. in theory. One likes to drop the verb for simple values (like here .. me too).

The discussion of redoing the commits (splitting etc) .. let's do that separately. 
First I like to finish the overall picture / feature .. then we see and balance for the final version.
Me still hacking on things here.

net/Socket.hpp
sgothel: type became a private field (underscore) .. and I wanted to clarify access (see above).
The overall work was to clean up socket, side effects etc .. while reading and understanding the code.
Mixed field access is very often a PIA .. and nowadays our C++ "constexpr" getter or queries are validated and inlined, see e.g. isIPType(). Perhaps a matter of taster.

net/Socket.hpp
sgothel: Aware of the stead monotonic clock now() costs, AFAIK the caller does not have it at this event (but will double check).
- _type is now const
- _creationTime was added as const
- _fd was always const (wan't my idea, but a good one)

The immutable nature clarifies semantics here and well possible if no assignment operator is used/allowed.
Sure, we could live w/o immutable semantics here - as long we have 'em private and use accessor methods as we have now. So .. well, either way. Still, having fields immutable is a good helper :)
Note: Compiler may behave differently .. but that is my least criteria for this code here.
Bottom line: I would stick with the current code as no algorithmic duplication really exist, but will check whether caller of ctor has now() already fetched - if that is OK.

net/Socket.hpp
sgothel: I often use bitfields when having considerable flags indeed.
However, each socket is acompanied by much larger buffers (1-8k) and I doubt 3 x sizeof(int) .. but well, keeping it in mind that its OK for you. An artifical encoding of different types (int + bool) is however .. overkill?
Sure, if we would talk about an object handle, function-ptr or any short lived type in quantities of >= 100k w/o any other side-effects or memory costs otherwise (sure) .. but I don't see it here.
Was thinking about the added memory costs of Socket with added statistics, correct. But as described above .. I doubt this is material here. I could be wrong ofc.

net/WebSocketHandler.hpp
sgothel: safe it is, I know...

net/Socket.cpp
Ashod: Surely I'd put these on one line, @mmeeks! (This doesn't benefit from multiple lines, it suffers.)

`return WSState::WS == t ? "WS" : "HTTP";`

Although, I would call it `state` and return `const char*` (create `string` only when necessary, not by default).

net/Socket.cpp
```diff
@@ -1302,6 +1383,33 @@ LocalServerSocket::~LocalServerSocket()
...
+    {
+        return "WS";
+    }
+    return "HTTP";
+}
+
+std::string StreamSocket::toString() const noexcept
+{
+    std::string s("StreamSocket[#");
+    s.append(std::to_string(getFD()))
```
Ashod, line 1408: If we must use `append`, we better `reserve`. This code will almost certainly re-allocate the string buffer more than once, unnecessarily. Either use `ostringstream` or `string` with reasonable `reserve` up-front to avoid said re-allocations.

net/Socket.hpp
Ashod: Why not lift `_closed` from `StreamSocket` to its parent, here?

At a minimum we should review `StreamSocket::_closed`, now that we have `_open`. Otherwise, horrors when these two go out of sync and the left hand fights the right.

Of course it's also good when adding such common flags to review how we dealt with them for so long, to avoid redundancy and unify the design. Perhaps `_open` was enough and this is redundant. Perhaps not. Worth sharing your analysis either way.

net/Socket.hpp
```diff
@@ -409,6 +486,15 @@ class Socket
     std::thread::id _owner;
 };
 
+// injecting specialization of std::hash to namespace std of Socket
+namespace std
+{
+    template <> struct hash<Socket>
```
Ashod, line 493: What is the motivation for this? I understand that now we can have `Socket` in unordered_set/map, but we didn't need that so far, so what's the use-case now?

(All our Socket instances are shared_ptr and we intentionally use `vector` for performance. Given our strict ownership model, we cannot have duplicate Sockets. So I'm very curious as to why we need new containers and/or a different model.)

net/Socket.hpp
```diff
@@ -1443,21 +1518,13 @@ class StreamSocket : public Socket,
         if (closed)
         {
             LOG_TRC("Closed. Firing onDisconnect.");
-            _closed = true;
             _socketHandler->onDisconnect();
-        }
-
-        if (_closed)
+            setClosed(disposition);
+        } else if (!isOpen()) {
```
Ashod, line 1529: Formatting. (I know, you are still working out the clang-format setup, but pointing out anyway.)

net/WebSocketHandler.hpp
```diff
@@ -241,21 +242,52 @@ class WebSocketHandler : public ProtocolHandlerInterface
     void shutdown(const StatusCodes statusCode = StatusCodes::NORMAL_CLOSE,
                   const std::string& statusMessage = std::string(),
                   bool hardShutdown = false)
+    {
+        shutdownImpl(statusCode, statusMessage, hardShutdown, false);
+    }
+
+    /// Don't wait for the remote Websocket to handshake with us; go down fast.
+    void shutdownAfterWriting()
```
Ashod, line 256: Perhaps `shutdownFast`, `shutdownImmediately`, or `shutdownWithoutWaiting` or something more similar to the comment? 

What has "writing" anything to do with "Don't wait for the remove Websocket" and "go down fast" ? Seems misleading name.

Also, I expect it force-flushes before closing. In that case, it could be called `flushAndShutdown`, or some such.

net/Socket.hpp
sgothel: It was a WIP snippet for a hash-map, but since I don't need such thing I will remove it.
Yup, as mentioned in the hash snippet, its the unique _fd.

net/WebSocketHandler.hpp
sgothel: I just moved `shutdownAfterWriting` method and left the name unchanged.

net/Socket.hpp
sgothel: No, I don't still figuring out my clang-format setup. But yes, I apply it manually for bigger edited code chunks only.

net/Socket.hpp
sgothel: StreamSocket::_closed has been moved to Socket::_open

net/Socket.cpp
mmeeks: great =)

net/Socket.cpp
mmeeks: Fair enough - if you insist on the accessors - lets make sure they have the right scope - protected if necessary; ideally private of course.

net/Socket.hpp
mmeeks: very common failure mode is un-initialized members; the more we share there the safer we are. Lets share all we can =) The 'const' stuff is fine - I guess until we get proper chained constructors in C++.

Pushing 'now' around the code to more places is something that happens a fair bit - and is easy to special-case if it's not needed - just have a default argument for it; and then tweak it in the places that matter I guess. Still sockets are not created that often so prolly not a big deal.

net/NetUtil.hpp
sgothel: Renamed to Config as it is a runtime configuration as required for unit testing of this feature as mentioned.

net/Socket.cpp
sgothel: all commits separated and bisectable now

net/Socket.cpp
sgothel: changed to const char*, thank you

net/Socket.cpp
sgothel: changed to simple noun name (accessor) and moved writer to protected in revised commit.

net/Socket.hpp
sgothel: added ctor arg w/ default, was a PIA and probably not impacting really - but another good reason to walk around the code.

net/Socket.hpp
sgothel: removed, thx!

mmeeks:
> Since I am quite happy now, I move this to 'review'.

Thanks Sven - sounds very promising =) good stuff.

mmeeks:
I'm confused. I still see some patches eg. this one:
https://github.com/CollaboraOnline/online/pull/9916/commits/e5d6dc55a210aab1aa3d22b408e4726fbda96945
that has a combination of totally safe, new debugging methods, as well as much the more risky 'isClosed' change, as well as all this renaming and accessor changes - which need splitting in isolation to be reviewed carefully, 
or am I mis-reading the github UI - TBH I find the UI confusing; perhaps those are old patch fragments from earlier commits or something.
Did you push the split ? and/or am I just not seeing it ?

sgothel:
> I'm confused. I still see some patches eg. this one: [e5d6dc5](https://github.com/CollaboraOnline/online/commit/e5d6dc55a210aab1aa3d22b408e4726fbda96945) that has a combination of totally safe, new debugging methods, as well as much the more risky 'isClosed' change, as well as all this renaming and accessor changes - which need splitting in isolation to be reviewed carefully, or am I mis-reading the github UI - TBH I find the UI confusing; perhaps those are old patch fragments from earlier commits or something. Did you push the split ? and/or am I just not seeing it ?

I didn't remove this one (Socket and WebSocketHandler) as visible by the date + hash.
I reorganized on top of it to separate feature from restructuring code, thought this was the intention
and I do not see anything risky here (flipped a boolean close -> open, OK).

The next parts 2, 3 address single issues and also fix the std::string 'abuse' for osstream,
scoped enums, shorter accessors etc.

All of these are orthogonal to the feature - so could be seen as bisectable ex-feature.

A later Socket change for the passing of time-point, then the 2 feature commits.

Perhaps not all 100% to your preference in regards to sliced down commits .. 
but feature and restructuring is.

If you insist on re-splitting it (i.e. this 1st commit perhaps), advise .. would do it tomorrow then 
incl. clean-build tests at least (as done here).

mmeeks:
Well - CI passes which is a good sign; I'm really not happy at having cosmetic cleanups merged with drastic functional change - today you can see where the functional change is amid the big chunk of non-functional change: how about in two years ? =) We can get this in, but not again for socket code.

My only concern now is what happens during save-on shutdown; I had a nightmare last night (literally) that the lifecycle of ClientSession is essentially owned by the SocketPoll (is it?) - and if that is the last ClientSession that has been hard disconnected from, and we have un-saved data - we need to keep that ClientSession alive until it is saved.

It struck me that this might well be why we have shutdown but not deleted Socket's in the SocketPoll - or did I miss something there?

@Ashod can you do a last quick review for that use-case, and double-check we have a unit test that covers it effectively ? =)

Thanks!

Ashod:
Hi @mmeeks,

> My only concern now is what happens during save-on shutdown; 

It's complicated and one of the most sensitive parts of the code, with a lot of history.

> I had a nightmare last night (literally) that the lifecycle of ClientSession is essentially owned by the SocketPoll (is it?)

Sort of.

SocketPoll does in fact control when we find out that the socket is disconnected (`ClientSession::onDisconnect()`), which lets DocBroker know to start cleaning up (`DocumentBroker::removeSession()`).

> and if that is the last ClientSession that has been hard disconnected from, and we have un-saved data - we need to keep that ClientSession alive until it is saved.

Indeed, as you expect, that's what we do.

We don't blindly remove the `ClientSession`'s reference from DocBroker. Instead, we intentionally keep the _last_ reference to `ClientSession` in DocBroker and start the unload process, if we need to save. We keep this last reference until we successfully save and upload, or we fail and give up.

I have really tried to understand if this PR could affect this in any way, but I can't with any confidence say I found out the answer. On the face of it, nothing stood out. But the sockets are such a foundational part of everything we do, that it's hard to say with a straight face there are no side-effects to worry about.

> It struck me that this might well be why we have shutdown but not deleted Socket's in the SocketPoll - or did I miss something there?

I can't tell. So much has changed, from socket lifetime management, to the socket poll, to refactorings of socket open/closed state, to the timeout management, that I cannot estimate the impact of all of this. (My worry is that we've mangled things badly enough that we either randomly disconnect sockets, or we don't properly handle edge cases that many other functions depend on.)

But I do know that we had far more benign code than this, that had harbored great ills and nasty surprises.

> @Ashod can you do a last quick review for that use-case, and double-check we have a unit test that covers it effectively ? =)

We do have unit-tests for the last-editor disconnecting with modifications to save. Unfortunately, the unit-tests related to save and upload have become very unstable of late, with very high rate of random failure. So that's no consolation.

P.S. This 5th pass was anything but quick. Considering there is ~1000 new lines changed since my last review, at ~2400 lines, I'm surprised it only took me 3 hours. (Actually, no, I'm not surprised; I had to stop.)

Ashod (commented):
Thanks, @sgothel for not making this any larger.

I left some concerns and thoughts. I refrained from repeating myself. My questions aren't rhetorical. Some decisions in this code were discussed, and I had made various suggestions, but they were ignored. 

But perhaps there has been discussions and decisions that I wasn't part of.

At any rate, I leave my comments for your consideration.

Thank you.

mmeeks:
> > I had a nightmare last night (literally) that the lifecycle of ClientSession is essentially owned by the SocketPoll (is it?)
> 
> Sort of.

So I did a bit of reading myself; this is fine: I thought SocketPoll controlled lifecycle of the ClientSession - but in fact the DocumentBroker member:

    /// All session of this DocBroker by ID.
    SessionMap<ClientSession> _sessions;

template<class T>
class SessionMap : public std::map<std::string, std::shared_ptr<T> >

holds a hard reference on it - so not a big deal =)

net/Socket.cpp
```diff
@@ -1302,6 +1383,33 @@ LocalServerSocket::~LocalServerSocket()
...
+std::string StreamSocket::toString(WSState t) noexcept
+{
+    if( WSState::WS == t )
+    {
+        return "WS";
+    }
+    return "HTTP";
+}
+
+std::string StreamSocket::toString() const noexcept
```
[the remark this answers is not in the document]
sgothel: This has been revised to `std::ostream& StreamSocket::stream(std::ostream& os) const` in my latest change set.
Pls note that I also have removed `noexcept` as of late for more complex methods.

net/DelaySocket.cpp
```diff
@@ -23,9 +23,16 @@ std::once_flag Delay::DelayPollOnceFlag;
...
 class DelaySocket : public Socket {
     int _delayMs;
-    enum State { ReadWrite,      // normal socket
-                 EofFlushWrites, // finish up writes and close
-                 Closed };
+    enum class State : uint8_t { ReadWrite,      // normal socket
+                                 EofFlushWrites, // finish up writes and close
+                                 Closed };
+
+    static const char* toString(State s) noexcept {
```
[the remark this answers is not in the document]
sgothel: I had the same thought, but then didn't want to change even more code. So yes, we could reuse our macro .. even if it would render the enum underling bigger (I used uint8_t .. the macro uses int in general, no biggy).

wsd/COOLWSD.cpp
```diff
@@ -2802,15 +2800,24 @@ void COOLWSD::innerInitialize(Application& self)
     {
         conf.setString("feedback.show", "true");
         conf.setString("welcome.enable", "true");
-        COOLWSD::MaxConnections = MAX_CONNECTIONS;
+        COOLWSD::MaxConnections = net::Config::get().MaxConnections; // aligned w/ MAX_CONNECTIONS above
```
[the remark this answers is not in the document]
sgothel: Reason for now net::Config was to have things configurable at least for the unit test, proving functionality.
Then - after your hint - I found your MAX_CONNECTION and realized that it must be considered and used, if not overridden by the optional config layer (unit test).
Hence, if we still want to add this functionality of added cut-off after max-connections - we should test it.
(Now I saw a unit test building up MAX_CONNECTION .. though, so it might indeed not be required)

Considering your MAX_CONNECTION measures and if that is sufficient, this angle of limited connection in this patch series probably shall be removed completely?

net/Socket.hpp
```diff
@@ -136,13 +141,22 @@ class Socket
...
 
-    enum Type { IPv4, IPv6, All, Unix };
+    enum class Type : uint8_t { IPv4, IPv6, All, Unix };
+    static const char* toString(Type t) noexcept;
 
     // NB. see other Socket::Socket by init below.
-    Socket(Type type)
-        : _fd(createSocket(type))
+    Socket(Type type,
+           std::chrono::steady_clock::time_point creationTime = std::chrono::steady_clock::now())
```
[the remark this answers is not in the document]
sgothel: I simply wasn't sure .. and in our discussion (with @mmeeks I guess) a mix of cut-off criteria were mentioned. 
The max-duration criteria, which required knowing the socket age, comes from the old `keep alive` feature (Http 1.0) and I thought it might be helpful here.

In case we deem this not required or perhaps unwarranted, then we surely can drop criteria and this creationTime time-point field. Which removes a lot of this code .. same as with max-connections ofc.

net/Socket.hpp
```diff
@@ -152,40 +166,71 @@ class Socket
...
+    bool isOpen() const noexcept { return _open; }
+    /// Returns true if this socket has been closed, i.e. rejected from polling and potentially shutdown
+    bool isClosed() const noexcept { return !_open; }
 
-    void setClientAddress(const std::string& address)
-    {
-        _clientAddress = address;
-    }
+    constexpr Type type() const noexcept { return _type; }
+    constexpr bool isIPType() const noexcept { return Type::IPv4 == _type || Type::IPv6 == _type; }
```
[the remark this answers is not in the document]
sgothel: I prefer qualification as much as possible, even though knowing a wrong `noexcept` is not ideal (-> disastrous termination in case of an exception). So yes, `noexcept` needs to be used carefully.
A `noexcept` gives benefits of relieving the compiler of creating a potential stack frame and is also required for `constexpr` semantics.

Now `constexpr` is not only useful for using an instance in a compile time manner (as I agree a Socket never will be), but also to
- force inline as much as possible
- force compiler to enhance code validation 
The latter is what interests me most, while the inlining is a very nice to have to allow compiler to remove complexity. This especially for small accessors. Allowing us to be more expressive at zero costs.

net/HttpRequest.hpp
```diff
@@ -1376,7 +1378,7 @@ class Session final : public ProtocolHandlerInterface
     }
 
     /// Returns the socket FD, for logging/informational purposes.
-    int getFD() const { return _fd; }
+    int fd() const { return _fd; }
```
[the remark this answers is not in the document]
sgothel: I have to admit this might be just another misunderstanding here .. my fault.
First I added the accessor for `_fd` with name `getFD()`. Then we had a chat here regarding naming and whether accessors are good/useful. Then I probably wrongly assumed (?) to use simpler naming to imply no processing in such accessor (removing the verb get). 
Yeah, at this stage this added to the confusion and review - bad decision on my behalf. Sorry.

net/SslSocket.hpp
```diff
@@ -65,7 +66,7 @@ class SslStreamSocket final : public StreamSocket
             LOG_TRC("Setting SSL into connect state");
             SSL_set_connect_state(_ssl);
             if (SSL_connect(_ssl) == 0)
-                LOG_DBG("SslStreamSocket connect #" << getFD() << " failed ");
+                LOG_DBG("SslStreamSocket connect #" << fd_in << " failed ");
```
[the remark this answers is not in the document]
sgothel: see above, you are correct.

net/Socket.cpp
```diff
@@ -55,18 +58,61 @@
...
+    }
+    size_t res, pre;
+    {
+        std::lock_guard<std::mutex> lock(StatsMutex);
+        pre = GetStatsConnectionCount();
+        res = pre;
+        if( added <= std::numeric_limits<size_t>::max() - res ) {
+            res += added;
+        } else {
+            // overflow
```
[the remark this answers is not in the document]
sgothel: I know, this was my paranoia code .. instead of forcing w/ a static_cast.

net/WebSocketHandler.hpp
```diff
@@ -393,10 +416,18 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
                         LOG_WRN("Servers should not send pongs, only clients");
 
-                    _pingTimeUs = std::chrono::duration_cast<std::chrono::microseconds>
-                        (std::chrono::steady_clock::now() - _lastPingSentTime).count();
-                    LOGA_TRC(WebSocket, "Pong received: " << _pingTimeUs << " microseconds");
-                    gotPing(code, _pingTimeUs);
+                    const auto now = std::chrono::steady_clock::now();
+                    const int64_t us = std::chrono::duration_cast<std::chrono::microseconds>
+                                          (now - _lastPingSentTime).count();
+                    const double avg_us = _pingMicroS.add(_lastPingSentTime, static_cast<double>(us));
```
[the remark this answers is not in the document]
sgothel: That is a bug of mine, outch .. yes, should pass `static_cast<double>(us)/1000000.0`.  Great catch, thank you!
Note: My unit test didn't catch this either, b/c fluctuations are very low but I should have done a better job in interpretation of the values. So I take the blame here - bad mistake.

Otherwise .. , TimePoint misses denoting dimension of value 'v' in method 'add', yes. Re 'convoluted' .. that really depends whether this functionality is desired and so forth. std:chrono duration in seconds and double could/should  be used for clarity, yup.

sgothel:
> > > I had a nightmare last night (literally) that the lifecycle of ClientSession is essentially owned by the SocketPoll (is it?)
> > 
> ```
...
> template class SessionMap : public std::map<std::string, std::shared_ptr >
> 
> holds a hard reference on it - so not a big deal =)

I was testing (~1 week ago) the socket-close on these criteria of this patch with firefox editing a document. The connection we build-up again and no artifacts were visible.

net/WebSocketHandler.hpp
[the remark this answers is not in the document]
sgothel: > That is a bug of mine, outch .. yes, should pass `static_cast<double>(us)/1000000.0`. Great catch, thank you! Note: My unit test didn't catch this either, b/c fluctuations are very low but I should have done a better job in interpretation of the values. So I take the blame here - bad mistake.

The `us` value in microseconds managed by TimePoint is correct! (and so is the unit test)

Sorry, your comment confused myself .. the value managed by TimePoint is of `arbitrary nature` (and dimension), as mentioned in `TimePoint::add` So here microseconds is used. 

The confusing part is probably that this value itself reflects time duration, however, `TimePoint::duration` is orthogonal, as it only refers over the measured total time and not the semantics of this value (ping response duration in microseconds).
So yes, annotations in TimePoint probably should be more emphasized, only the `add` method does. *sigh*

net/WebSocketHandler.hpp
```diff
@@ -393,10 +416,18 @@ class WebSocketHandler : public ProtocolHandlerInterface
...
+                    if (_openPingCount > 0)
+                    {
+                        --_openPingCount;
+                    }
+                    LOGA_TRC(WebSocket, "Pong received: " << us << "us, avg " << avg_us << "us over "
+                                         << (int)_pingMicroS.duration() << "s, open pings left "
```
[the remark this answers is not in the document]
sgothel: As replied below (my 2nd one) .. TimePoint manages arbitrary values, here ping-micro-seconds. 
The TimePoint::duration() in seconds only gives us a rough idea for how long this average has been taken, where values are added to the average .. like every 3 seconds by default (net::Config::WSPingPeriod).
The granularity in seconds is enough for the logging and manual validation.

net/WebSocketHandler.hpp
```diff
@@ -33,8 +35,11 @@ class WebSocketHandler : public ProtocolHandlerInterface
     std::weak_ptr<StreamSocket> _socket;
 
 #if !MOBILEAPP
+    std::chrono::microseconds _pingPeriod;
+    std::chrono::duration<double, std::micro> _pingTimeout;
     std::chrono::steady_clock::time_point _lastPingSentTime;
-    int _pingTimeUs;
+    size_t _openPingCount;
+    Util::TimeAverage _pingMicroS;
```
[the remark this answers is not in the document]
sgothel: _pingMicroS is a TimeAverage for this value, i.e. averaging out multiple measurements over time, which removes one-offs etc.

_openPingCount is the open ping count, i.e. pings from server w/o client reply. Initially I intended this for debugging only, 1st validation.
Reviewing its use in WebSocketHandler::checkTimeout() is wrongly used IMHO:
>       if (_openPingCount > 0 && _pingTimeout.count() > std::numeric_limits<double>::epsilon() &&
>            _pingMicroS.average() >= _pingTimeout.count())
>        {

instead it should just read 

>       if (_pingTimeout.count() > std::numeric_limits<double>::epsilon() &&
>            _pingMicroS.average() >= _pingTimeout.count())
>        {

So yes, `_openPingCount` should not be material and be removed.

sgothel:
> Again I would like to separate the cosmetic re-working of things from functional changes.
This should be resolved now, all bisectable.

> Pushing down the bytesSent etc. statistics from StreamSocket -> Socket - interesting, not sure of the rationale - perhaps harmless, but would like to know why - where does this simplify something else ?
- to allow dumping the socket stats (incl. throughput)
- to have it in the same level as open state, fd,  - same domain

> 
> And of course, other things - not so convinced about ;-)
Hope most of thes 'claims', incl. @Ashod s, are resolved by now, 
please note `Changes to original PR` in description on the top.

net/DelaySocket.cpp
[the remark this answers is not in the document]
sgothel: Resolved with revised commit `Use scoped enums with efficient types (STATE_ENUM)`, missed it.

wsd/COOLWSD.cpp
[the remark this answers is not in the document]
sgothel: revised version has no more xml config file side-effects, only storing MAX_CONNECTIONS in the net::Defaults - potentially mutated by unit test only.

net/Socket.hpp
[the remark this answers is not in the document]
sgothel: last use case is for stats and minimum throughput criteria (currently disabled by default)

net/HttpRequest.hpp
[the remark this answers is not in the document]
sgothel: reverted

net/SslSocket.hpp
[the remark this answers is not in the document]
sgothel: reverted

sgothel:
Added minor changes addressing @Ashod earlier notes.

sgothel:
Note: Unit tests:   +7 files, 879 insertions(+), remaining changed lines +800, -222

net/Socket.hpp
```diff
@@ -934,33 +1032,29 @@
     }
 };
 
-enum SharedFDType { SMAPS, URPToKit, URPFromKit };
+STATE_ENUM(SharedFDType, SMAPS, URPToKit, URPFromKit);
```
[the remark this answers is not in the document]
sgothel: side effect of using STATE_ENUM macro, injecting such `static const`. Would prefer a `constexpr`.

net/Socket.hpp
```diff
@@ -186,13 +186,17 @@ class Socket
     virtual void shutdown()
     {
         if (_noShutdown)
+        {
+            setClosed();
```
mmeeks, line 222: Why don't we do setClosed at the top of this function just once (?)

net/Socket.cpp
```diff
@@ -1395,6 +1407,60 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
...
+    {
+        return false;
+    }
+    const auto durTotal =
+        std::chrono::duration_cast<std::chrono::milliseconds>(now - getCreationTime());
+    const auto durLast =
+        std::chrono::duration_cast<std::chrono::milliseconds>(now - getLastSeenTime());
+    const double bytesPerSecIn = durTotal.count() > 0 ? (double)bytesRcvd() / ((double)durTotal.count() / 1000.0) : 0.0;
+    const bool c1 = now < getCreationTime();
+    const bool c2 = _pollTimeout > std::chrono::microseconds::zero() &&
```
mmeeks, line 1502: This is very complex. Are there any tests ?
the _pollTimeout I would expect to sometimes be very small - depending on timers - so say it is 1 microsecond, but normally say 5 seconds or somesuch; and the durLast - is the gap to the last seen time can easily be longer than that - so - I would expect c2 to regularly be true, and depend if someone setup a short timeout in the loop - right ?

Can we simplify these - and give them useful symbolic names to aid comprehension ? =) c1 -> c4 don't float my boat :-)

net/Socket.cpp
```diff
@@ -1395,6 +1407,60 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
...
+                    durLast > _pollTimeout;
+    const bool c3 = _minBytesPerSec > std::numeric_limits<double>::epsilon() &&
+                    bytesPerSecIn > std::numeric_limits<double>::epsilon() &&
+                    bytesPerSecIn < _minBytesPerSec;
+    const bool c4 = SigUtil::getTerminationFlag();
```
mmeeks, line 1507: and that can be directly in the if I think =)

net/Socket.cpp
```diff
@@ -1395,6 +1407,60 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
     return os << "]";
 }
 
+bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
+{
+    if (!isIPType()) // forced removal on IPv[46] network connections only
+    {
```
mmeeks, line 1493: those { look redundant to me.

net/Socket.cpp
```diff
@@ -540,7 +564,29 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
...
+                {
+                    std::shared_ptr<Socket>& socket = _newSockets.back(); // oldest
+                    assert(socket);
+
+                    LOG_WRN("Limiter: #" << socket->getFD() << ": Removing "
+                             << (i+1) << " / " << overhead << " new socket of (pre "
+                             << globCount << " + new " << newConnCount << ") / max " << _connectionLimit
+                             << " from " << _name);
+
+                    socket->resetThreadOwner();
```
mmeeks, line 586: Interesting - looks like a useful feature; is there a test for this ? and do we build stats on / log how many dropped connections for admin-console / prometheus metrics ?

mmeeks (approved):
Ok - this is great; thanks Sven & Caolan - much easier to review, can go in at the beginning of the next Red merge window; ignore my comments about unit tests - I see them at the end; other minor cleanups can come later / get tacked on.
Nice work - thanks ! =)

mmeeks:
Actually - I lie the c1,2,3,4 thing really needs fixing or explaining; its not clear to me that it will work =)

net/Socket.cpp
sgothel: I prefer having these criteria flags dedicated for logging, but renamed then.

net/Socket.cpp
sgothel: - c2 (now cIDLE) denotes: `TO Criteria: Violate maximum idle (_pollTimeout default 64s)`
- renamed the criteria names as well as their logging description

net/Socket.hpp
sgothel: good point, fixed

net/Socket.cpp
sgothel: tested in given unit tests.

no properly structured admin-console etc stats produced yet for this feature.

sgothel:
@mmeeks @caolanm addresses Michael's notes
- inlined the changes in rebased commits
- rebased to (Monday's) master

sgothel:
>     * rebased to (Monday's) master

I apologize: Should have pushed with a rebase to master 1st, then another push with the changes.
This would have allowed us to see the push delta on the fixed notes only using this github UI diff.
Now it shows the total change :-(

sgothel:
@caolanm 

Did a full rebase to master a330f9e75717de96c72794d0ac672f75376c9711
- ensuring to pass `git rebase -i --exec "make -j 20"`
  - fixing some mild compilation issues introduced with last rebase merge fixes

sgothel:
CodeQL shows some [setup troubles](https://github.com/CollaboraOnline/online/actions/runs/11358427600/job/31592929976?pr=9916) ...
```
The following packages have unmet dependencies:
 builddeps:libreoffice : Depends: g++ (>= 4:12) but 4:11.2.0-1ubuntu1 is to be installed
                         Depends: gcc (>= 4:12) but 4:11.2.0-1ubuntu1 is to be installed
                         Depends: java-common (>= 0.75) but 0.72build2 is to be installed
                         Depends: libharfbuzz-dev (>= 5.1.0) but 2.7.4-1ubuntu3.1 is to be installed
                         Depends: libmdds-dev (>= 2.1) but 2.0.2-1 is to be installed
                         Depends: liborcus-dev (>= 0.19.1) but 0.17.2-2 is to be installed
                         Depends: librdf0-dev but it is not going to be installed
                         Depends: libxmlsec1-dev (>= 1.2.35) but it is not going to be installed
                         Depends: fonts-crosextra-carlito (>= 20230309) but it is not going to be installed
                         Depends: fonts-liberation (>= 1:2) but 1:1.07.4-11 is to be installed

```

caolanm:
StreamSocket::checkRemoval does something on (cNow || cIDLE || cMinThroughput || cTermination ) where:

cTermination is SigUtil::getTerminationFlag() which looks straightforward

cNow is cNow = now < getCreationTime() which is presumably never in practice

cMinThroughput depends on _minBytesPerSec getting set to non-zero which is not enabled by default (test/UnitTimeoutSocket.cpp uses it) so doesn't (currently) affect normal usage

cIDLE is durLast > _pollTimeout; where _pollTimeout is 64secs (preexisting value as DefaultPollTimeoutMicroS)

PR #9920, opened 2024-08-28
silence error from find during 'make dist'

timar:
It is:
    find: 'node_modules/jquery-ui/ui/i18n': No such file or directory
This directory is created during normal build.

Change-Id: Ie7c2cee129f1f880a22c4a4d8a48788b73570d5b

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -164,7 +164,7 @@ NODE_MODULES_JS =\
 	node_modules/jquery-mousewheel/jquery.mousewheel.js \
 	node_modules/jquery-contextmenu/dist/jquery.contextMenu.js \
 	node_modules/jquery-ui/dist/jquery-ui.js \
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Fine, though I wonder if it would be even better to list those files explicitly instead of a magic 'find' and hope that make dist is done in a clean environment.

I mean, in core.git, external/python3/ExternalPackage_python3.mk has ~900 lines to list everything explicitly, but here we use 'find' to avoid listing 75 files. :-) It's also easier to understand the build system if every source file is listed somewhere explicitly, similar to cpp and ts files. Anyhow, that's just for the long run.

timar:
> Fine, though I wonder if it would be even better to list those files explicitly instead of a magic 'find' and hope that make dist is done in a clean environment.
 I see the point, but in next version they may add new translations that we may miss. With find we add them automatically.

vmiklos:
Fair enough, I didn't consider that problem for external code.

PR #9923, opened 2024-08-28
fix regression from "Use a hidden variable to get the css vars from url"

mcecchetti:
Compiler complains about unknown uint32_t type.

Change-Id: I690d89470bfdd977e7fa71957cdc118d9257b7b9

### DIFF
--- a/wsd/FileServerUtil.cpp
+++ b/wsd/FileServerUtil.cpp
@@ -11,14 +11,14 @@
 
 #include <config.h>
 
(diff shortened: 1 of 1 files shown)


mcecchetti:
Ah so the fix is wrong, I mean locally it works, but it seems not portable

mcecchetti:
New version looks good

gokaysatir (approved):

PR #9924, opened 2024-08-28
added missing header file

lpranam:
Change-Id: I05cc2fe41ba583dfdd24abd2b0c95081b402c85a

* Target version: master 

### Checklist

### DIFF
--- a/wasm/base64.hpp
+++ b/wasm/base64.hpp
@@ -30,6 +30,7 @@
  */
 
 #include <string>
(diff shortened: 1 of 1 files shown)


pedropintosilva:
thanks @lpranam but you forgot to add a reviewer so it seems later on @caolanm has created a PR with the same fix: https://github.com/CollaboraOnline/online/pull/9929 it's already merged

PR #9929, opened 2024-08-29
uint32_t needs <cstdint>

caolanm:
./wasm/base64.hpp:109:7: error: ‘uint32_t’ was not declared in this scope

Change-Id: I9de3054504fc6108b473f99c8b43cc3c7cc567fe

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wasm/base64.hpp
+++ b/wasm/base64.hpp
@@ -29,6 +29,7 @@
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include <cstdint>
 #include <string>
 
 namespace macaron {


caolanm:
trivial build fixer

vmiklos (approved):
Makes sense.

FWIW sometimes I used IWYU on online.git code, like `~/git/include-what-you-use/iwyu_tool.py -p . test/UnitTiffLoad.cpp` (assuming you have a compile_commands.json in the current directory), but that's just for the normal build, not wasm.

caolanm:
FWIW this wasm header started getting used in the normal build since:
https://github.com/CollaboraOnline/online/pull/9830

PR #9932, opened 2024-08-29
lok: sw: restrict delete and modify on comment only mode

JaumePujantell:
On Writer, when in readonly mode with comments activated restrict
the hability to delete and modify comments to only those made by
the same user.

Change-Id: I7787946c7c25d7ad97074ebf6f1c8478fb9a84de


JaumePujantell:
The same checks are made server side on https://gerrit.libreoffice.org/c/core/+/172596

PR #9933, opened 2024-08-29
Detach navigator from sidebar

NickWingate:
* Resolves: # 
* Target version: master 

### Summary
Needs: ~~https://gerrit.libreoffice.org/c/core/+/172594 (24.04)~~ . The gerrit change for 25.04: https://gerrit.libreoffice.org/c/core/+/183296

### TODO

- [ ] scrolling
- [ ] clicking on navigator items leads to relevant part of document
- [ ] set max width for navigator (to deal with long headers)
- [ ] Impress and Calc implementations
    - [ ] Specifically with impresses slide-sorter which occupies the same area of the window

### Checklist


browser/src/control/Control.NavigatorPanel.ts
```diff
@@ -0,0 +1,186 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * JSDialog.NavigatorPanel
+ */
+
+/* global app */
+class NavigatorPanel {
```
eszkadev, line 16: Can we now create base class: `SidebarBase` which will be parent of Sidebar and Navigator.
We can then share almost all of the `onAdd`, `onRemove`, `isVisible`, `onJSUpdate`, `onJSAction`.
We only need to define strings which are different:
```
this.commandId = 'navigator' / 'sidebar';
this.containerWrapperId = 'navigator-dock-wrapper' / 'siebar-dock-wrapper';
this.panelId = 'navigator-panel' / 'sidebar-panel';
```

why do not concatenate even more:   `TYPE + '-dock-wrapper'` ? because it's then easier to `git grep` if id is fully present in the code and css to find related code :)

pedropintosilva:
Hey @NickWingate I have reassigned this to @codewithvk I hope that's ok, it would be sad to don't leverage the nice work you have already done there.

pedropintosilva:
related gerrit change: https://gerrit.libreoffice.org/c/core/+/172594

eszkadev:
please rebase on top or `origin/master`

browser/src/control/Control.UIManager.js
```diff
@@ -261,6 +261,9 @@ L.Control.UIManager = L.Control.extend({
 
 			this.map.sidebar = JSDialog.Sidebar(this.map, {animSpeed: 200});
 
+			this.map.navigator = JSDialog.NavigatorPanel(this.map, { animSpeed: 200 });
+
+			this.map.mention = L.control.mention(this.map);
```
eszkadev, line 266: unwanted change?

browser/src/control/Control.NavigatorPanel.ts
```diff
@@ -0,0 +1,77 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * JSDialog.NavigatorPanel
+ */
+
+/* global app */
+class NavigatorPanel extends JSDialog.SidebarBase {
```
eszkadev, line 16: I think "prefix" `JSDialog` is not needed in .ts files. We will detect class correctly.
It is needed only in JS files.

browser/src/control/Control.Sidebar.ts
```diff
@@ -12,17 +12,8 @@
...
-}
-class Sidebar {
-	options: SidebarOptions;
-
-	map: any;
-
-	container: HTMLDivElement;
-	builder: any;
+// /* global app */
+class Sidebar extends JSDialog.SidebarBase {
```
eszkadev, line 16: `JSDialog.` not needed here I think

browser/src/control/Control.UIManager.js
```diff
@@ -445,13 +448,13 @@ L.Control.UIManager = L.Control.extend({
 					this.map.sidebar.setupTargetDeck('.uno:MasterSlidesPanel');
 				} else if (this.getBooleanDocTypePref('NavigatorDeck', false)) {
 					app.socket.sendMessage('uno .uno:SidebarShow');
-					app.socket.sendMessage('uno .uno:Navigator');
-					this.map.sidebar.setupTargetDeck('.uno:Navigator');
+					//app.socket.sendMessage('uno .uno:Navigator');
```
eszkadev, line 451: Possibly we want to completely remove Navigator option from sidebar context here. To not try to close other decks in sidebar.
We should have separate setting for "left-sidebar". Similar to this one.

Darshan-upadhyay1110:
NOTE:  Need to fix cypress test here then it will be good to review

pedropintosilva:
I have tested the writer/navigator_spec.js and it seems to fail because CI doesn't have the respective gerrit patch (because it's not merged yet) and this the #navigation-sidebar is never found

eszkadev:
good, if only fails will be due to missing core patch, and it passes all locally with it - I will be able to review. Please ping me :)

Merging to be always green on CI:
- (optional) temporary cypress adjustments for old and new navigator passing CI
- core first
- online next
- (optional) remove old cypress checks

@Darshan-upadhyay1110 :
Will we fail tests if just core will be merged?
Is there any option for intermediate step so cypress will be able to get `old or new` navigator?

Darshan-upadhyay1110:
> Will we fail tests if only the core is merged?  

Yes, in the current situation, merging only the core will cause the CI to fail in multiple cases without this patch. I need to skip a few tests to make the CI pass. However, these two patches are closely connected, and they should be merged within a very short time frame.  

> Is there any intermediate step so that Cypress can handle both the old and new navigators?  

No, because the new navigator includes additional HTML elements, such as the "Navigation Header." This causes failures in tests like the navigator height check and verifying element positions. For example, currently, the navigator opens on the right since the CORE 25.05 patch hasn’t been merged, leading to confusion in Cypress when running these tests.

eszkadev:
> > Will we fail tests if only the core is merged?
> 
> Yes, in the current situation, merging only the core will cause the CI to fail in multiple cases without this patch. I need to skip a few tests to make the CI pass. However, these two patches are closely connected, and they should be merged within a very short time frame.
> 
> > Is there any intermediate step so that Cypress can handle both the old and new navigators?
> 
> No, because the new navigator includes additional HTML elements, such as the "Navigation Header." This causes failures in tests like the navigator height check and verifying element positions. For example, currently, the navigator opens on the right since the CORE 25.05 patch hasn’t been merged, leading to confusion in Cypress when running these tests.

I see, positions are calculated from the right edge... let me double check locally too, and we will be able to merge both at the same time possibly

browser/css/cool.css
```diff
@@ -344,11 +345,11 @@ body {
 	}
 }
 
-#sidebar-panel {
+#sidebar-panel, #navigator-panel {
```
eszkadev, line 354: we can make `class` of all of these multiplied rules for `navigator-` and `sidebar-` as `.sidebar-panel` and use in both left and right sidebar. As a followup

browser/src/control/Control.UIManager.js
```diff
@@ -511,6 +514,131 @@ L.Control.UIManager = L.Control.extend({
...
+	},
+
+	// Function to handle tab click
+	switchNavigationTab(tabId) {
+
+		var presentationControlsWrapper = document.getElementById('presentation-controls-wrapper');
+		var navigatorDockWrapper = document.getElementById('navigator-dock-wrapper');
+
+		// Remove 'selected' class from all tabs
+		document.querySelectorAll('.navigation-tabs .tab').forEach(t => t.classList.remove('selected'));
```
eszkadev, line 592: please do not query full document everywhere, we can limit scope to just a `navigatorDockWrapper.query...`

browser/src/control/Control.UIManager.js
```diff
@@ -511,6 +514,131 @@ L.Control.UIManager = L.Control.extend({
 		}
 	},
 
+	initializeNavigator: function (docType) {
```
eszkadev, line 517: can we move this function inside `navigator` file?
then here just call   ` this.navigator.initialize(...);`

also remember inside navigator instances of the containers as memeber variables on initialize.
we do not need to query them all the time

browser/src/control/Control.UIManager.js
```diff
@@ -511,6 +514,131 @@ L.Control.UIManager = L.Control.extend({
...
+		// Toggle visibility based on tabId
+		if (tabId === 'tab-slide-sorter') {
+			presentationControlsWrapper.style.display = 'block';
+			navigatorDockWrapper.style.display = 'none';
+		} else {
+			presentationControlsWrapper.style.display = 'none';
+			navigatorDockWrapper.style.display = 'block';
+		}
+	},
+	createFloatingNavigatorBtn() {
```
eszkadev, line 606: move to navigator itself (related pieces keep together)
we want to split big JS files, not grow them as it is harder to convert to TS later

browser/src/control/Control.Notebookbar.js
```diff
@@ -72,6 +72,9 @@ L.Control.Notebookbar = L.Control.extend({
 
 		$('#toolbar-wrapper').addClass('hasnotebookbar');
 		$('.main-nav').addClass('hasnotebookbar');
+		var FloatingNavIcon = document.querySelector('.navigator-btn-wrapper');
```
eszkadev, line 75: smallCapsFirst -> floatingNavIcon

browser/src/control/Control.Notebookbar.js
```diff
@@ -119,6 +122,9 @@ L.Control.Notebookbar = L.Control.extend({
 		$('.main-nav #document-header').remove();
 		$('.main-nav').removeClass('hasnotebookbar');
 		$('#toolbar-wrapper').removeClass('hasnotebookbar');
+		var FloatingNavIcon = document.querySelector('.navigator-btn-wrapper');
```
eszkadev, line 125: here too

cypress_test/integration_tests/common/impress_helper.js
```diff
@@ -222,7 +222,7 @@ function addSlide(numberOfSlides) {
 		for (let i = 0; i < numberOfSlides; i++) {
 			cy.cGet('#presentation-toolbar #insertpage')
 				.should('not.have.class', 'disabled')
-				.click();
+				.click({force: true});
```
eszkadev, line 225: why does it fail? we should avoid `force`.
was element covered by something?

cypress_test/integration_tests/desktop/writer/navigator_spec.js
```diff
@@ -18,13 +18,13 @@ describe(['tagdesktop'], 'Scroll through document, modify heading', function() {
...
-
-		cy.cGet('#Navigator-button').click();
+		cy.cGet('#navigator-floating-icon').click(); // Ensure it's open
 	});
 
-	it('Navigator visual test', function() {
+	it.skip('Navigator visual test', function() {
 		cy.wait(500); // wait to make fully rendered
-		cy.cGet('#contenttree').compareSnapshot('navigator_writer', 0.06);
+		cy.cGet('#navigator-dock-wrapper').scrollTo(0,0,{ ensureScrollable: false });
```
eszkadev, line 26: why the initial navigator scroll position is at the end? it looks like buggy behavior

cypress_test/integration_tests/common/impress_helper.js
Darshan-upadhyay1110: We can remove `force` as all the cases are now skipped until CORE merges. I think after CORE change merged this will pass without force

browser/src/control/Control.UIManager.js
eszkadev: why this mention thing is added here? I see no change related to mention

cypress_test/integration_tests/desktop/writer/navigator_spec.js
Darshan-upadhyay1110: After we do open this new navigator it automatically scrolls to last position. I have investigated on this but did not got any pointer why this is happening

eszkadev:
now we can rebase on top origin/master ( https://github.com/CollaboraOnline/online/pull/11484 added )
core is merged too

cypress_test/integration_tests/desktop/writer/navigator_spec.js
eszkadev: let's have followup ticket for that

browser/src/control/Control.NavigatorPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+			this.handleFloatingButtonVisibilityOnZoomChange,
+			this,
+		);
+	}
+
+	onRemove() {
+		super.onRemove();
+		this.map.off('navigator');
+		this.map.off('zoomend');
+		this.map.on('doclayerinit');
```
eszkadev, line 49: `off`

browser/src/control/Control.NavigatorPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	onDocLayerInit() {
+		// for presentation show slide sorter navigation panel by default
+		if (app.map.getDocType() === 'presentation' && !window.mode.isMobile())
+			app.map.sendUnoCommand('.uno:Navigator');
```
eszkadev, line 55: I believe this is a costly operation (to setup navigator in the core). And it is done on load.

Can we just show the slide sorter (which doesn't have core dependency) without loading navigator in the core?

browser/src/control/Control.NavigatorPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+			this.navigationPanel.prepend(navHeader);
+		}
+	}
+
+	createFloatingNavigatorBtn() {
+		// Get or create the main wrapper div
+		this.floatingNavIcon.className =
+			'notebookbar unoNavigator unospan-view-navigator unotoolbutton visible';
+		this.floatingNavIcon.setAttribute('tabindex', '-1');
+		this.floatingNavIcon.setAttribute('data-cooltip', 'Navigator');
```
eszkadev, line 146: `_('')` for l18n

browser/src/control/Control.SidebarBase.ts
```diff
@@ -0,0 +1,161 @@
...
+				return true;
+			}
+		}
+
+		return false;
+	}
+	onResize() {
+		var wrapper = document.getElementById(`${this.type}-dock-wrapper`);
+		wrapper.style.maxHeight =
+			document.getElementById('document-container').getBoundingClientRect()
```
eszkadev, line 157: document container is not changed in time, can we store reference in the member variable on initialization? to not query DOM all the time.

browser/src/control/Control.SidebarBase.ts
```diff
@@ -0,0 +1,161 @@
...
+		for (const i in data.children) {
+			if (this.markNavigatorTreeView(data.children[i])) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+	onResize() {
+		var wrapper = document.getElementById(`${this.type}-dock-wrapper`);
```
eszkadev, line 155: this one should be remembered too

eszkadev (changes requested):
I found some small issues - but overall looks good :)

eszkadev:
in desktop:
```
impress/editable_area_spec.js
impress/styatusbar_spec.js
```
seems to fail most

eszkadev:
impress/statusbar_spec.js seems to use zoom 60 not 85

Darshan-upadhyay1110:
> impress/statusbar_spec.js seems to use zoom 60 not 85

Locally i tested both cases and all works good. 

Statusbar_spec.js

image

ediatbel_area_spec.js

image

cypress_test/integration_tests/multiuser/impress/annotation_spec.js
```diff
@@ -28,6 +29,7 @@ describe(['tagmultiuser'], 'Multiuser Annotation Tests', function() {
 
 	it('Insert', function() {
 		cy.cSetActiveFrame('#iframe1');
+		desktopHelper.closeNavigatorSidebar();
```
eszkadev, line 32: we could put that into beforeEach at the end when we are switched to frame 2 :)

eszkadev (approved):
code looks good, let me test quickly and merge

meven:
> Followup task
> 
> #11513 #11514

Add this one :)
https://github.com/CollaboraOnline/online/issues/11518

PR #9934, opened 2024-08-29
chore(eslint): allow ES6 globals

Minion3665:
ES6 is ES2015, and we currently use ES2017. Unfortunately, eslint was configured to deny us globals which were added in ES6, including types (like Set or Map), etc.

To fix this, we can tell eslint to allow us to have ES6 globals, all of which have wide browser support and should have been enabled at the same time as everything else.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665:
I'm a big fan of using Set() and Map() - particularly Map() due to the prototype pitfalls that using objects can get you into, so I noticed this when I was looking at a different task and thought it'd be nice to split this off and send a quick PR now. There's no issue/etc. to refer to

eszkadev (approved):

PR #9938, opened 2024-08-30
Fix JS error: missing sheetgeometry

eszkadev:
we can pass undefined rectangle data to otherViewCursorSection as it is already expecting that case.

To reproduce error:
- session A runs automatic typer in Calc
- session B opens the same sheet and then reloads the page Result: error in session B


gokaysatir (approved):
Thanks for the fix :)

PR #9939, opened 2024-08-30
show thread options only on root comment

JaumePujantell:
Change-Id: Ifb97d0d41ed33ce2b0edb4cd84dc0301d42a8302


gokaysatir (changes requested):
Thanks for noticing this. Previous implementation was buggy in also another way i think.

If you check this, it adds "remove" and "resolve" thread buttons for only non-root comments. I think it needs a "!" at the start. So it can show those buttons only for root comments. Can you please check? Thanks.

gokaysatir:
> Thanks for noticing this. Previous implementation was buggy in also another way i think.
> 
> If you check this, it adds "remove" and "resolve" thread buttons for only non-root comments. I think it needs a "!" at the start. So it can show those buttons only for root comments. Can you please check? Thanks.

Ah sorry my bad, works now :)

gokaysatir (approved):

PR #9941, opened 2024-08-30
Fix Dockerfiles for local docker built

mhoffmann75:
Update Dockerfiles to be able to build local via docker (not via GitHub actions).
- Update to ubuntu 24.04 image to reflect change in gh action
- adapt user-id necessary since Debian 12 / Ubuntu 24.04
- add missing ca-certificates package that prevented WSD starting


timar (changes requested):
Hi @mhoffmann75, thank you very much for your fix. The only problem is that your [commits are not signed-off](https://github.com/CollaboraOnline/online/blob/master/CONTRIBUTING.md#sign-your-work) and do not contain [Change-Id](https://github.com/CollaboraOnline/online/blob/master/.git-hooks/commit-msg#L82) (please install the git hook for that).

timar (approved):

PR #9942, opened 2024-08-30
fix malformed Content-Type header

mhoffmann75:
malformed Content-Typer header prevented admin pages from properly loading with certain browsers (Apple Safari).

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -664,7 +664,7 @@ public:
             size);
     }
 
-    void setBody(const std::string& body, std::string contentType = "text/html charset=UTF-8")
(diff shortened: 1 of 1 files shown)


timar (changes requested):
Hi @mhoffmann75, thank you very much for your fix. The only problem is that your [commit is not signed-off](https://github.com/CollaboraOnline/online/blob/master/CONTRIBUTING.md#sign-your-work) and does not contain [Change-Id](https://github.com/CollaboraOnline/online/blob/master/.git-hooks/commit-msg#L82) (please install the git hook for that).

timar (approved):

timar (approved):

PR #9943, opened 2024-08-30
Fix JS errors in about dialog.

gokaysatir:
Change-Id: If0934089c2dbaa95ddfd345b2c360360c51640e4

* Resolves: # 
* Target version: master 

### Summary
There are JS errors in about dialog. This PR fixes the issues.

### DIFF
--- a/browser/src/control/Control.AboutDialog.ts
+++ b/browser/src/control/Control.AboutDialog.ts
@@ -50,6 +50,14 @@ class AboutDialog {
 			.cloneNode(true) as HTMLElement;
 		content.style.display = 'block';
 
+		if (content.querySelector('#js-dialog')) {
+			(content.querySelector('#js-dialog') as HTMLAnchorElement).onclick =
+				function () {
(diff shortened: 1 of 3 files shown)


browser/src/control/Control.AboutDialog.ts
```diff
@@ -50,6 +50,22 @@ class AboutDialog {
...
+						document.getElementById('modal-dialog-about-dialog-box-yesbutton')
+					)
+						(
+							document.getElementById(
+								'modal-dialog-about-dialog-box-yesbutton',
+							) as HTMLButtonElement
+						).click();
+				};
+		}
+
```
hcvcastro, line 68: Observation  =), invoking a click event inside a click function?.

gokaysatir: When user opens the widget test dialog, "About dialog" stays open. So i wanted to close About dialog for better usability.

But yes it looks suspicious, thanks for the comment :)

hcvcastro (approved):

PR #9944, opened 2024-08-30
browser: accessibility: fix min-content, max-content grid

hcvcastro:
Change-Id: I8a4615ef0c2abdd21d57737708d64a4c79cc134f

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -308,7 +308,7 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 }
 
 .ui-tabs-content.jsdialog > .ui-content.hidden {
-	display: inherit !important;
+	display: none !important;
 	visibility: hidden;
 }
 
--- a/browser/src/control/jsdialog/Widget.Containers.ts
+++ b/browser/src/control/jsdialog/Widget.Containers.ts
@@ -65,8 +65,8 @@ JSDialog.grid = function (
 		rows +
 		', auto); \
 		grid-template-columns: repeat(' +
-		cols +
-		', auto);';
+		(cols - 1) +
+		', max-content) min-content;';
 
 	table.style = gridRowColStyle;
 


eszkadev:
Hi @hcvcastro , can we have description in the commit message what does that fix?
One example, which dialog to open to test the change.

browser/css/jsdialogs.css
```diff
@@ -308,7 +308,7 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 }
 
 .ui-tabs-content.jsdialog > .ui-content.hidden {
-	display: inherit !important;
+	display: none !important;
```
eszkadev, line 311: `visibility: hidden` causes element to be hidden but still occupying the space
`display: none` removes the element completely
This is important inside grids where we have specific order of columns and rows

eszkadev: we did that to keep the same size of every tab in the dialog so "Next" button in the wizard will be at the same place, so user can just click multiple times in the same place

eszkadev:
We decided in the past that dialog should keep the same size when switching tabs if possible. It is important to keep "next" / "previous" button at the same position on the screen in wizards also.
Please see that this change causes a problem here:

[dynamicsize-2024-09-23_08.22.18.webm](https://github.com/user-attachments/assets/e497ad11-9cf6-44f1-adbb-c6d9e02b9bde)

eszkadev:
Before it was:
[dynamicsize-2024-09-23_08.27.20.webm](https://github.com/user-attachments/assets/930d7d0c-9bfe-4b08-82ef-0048644459e7)

pedropintosilva (changes requested):
This was already reviewed by @eszkadev . Additional notes: I appreciate to have the labels near to the fields it solves many layout problems, thanks Henry. Can you please look into trying to have that in while keep one of the sub containers with a static width? What are your thoughts @hcvcastro , @eszkadev ? We could somehow get what is the size required for the bigged tab in every dialog and set it so, no matter the tab the dialog would have the width unchanged.

Here is how paragraph dialog behaves with this PR


https://github.com/user-attachments/assets/7d0e8bf7-d7d0-4dd5-a5f8-5249156efcca



And here is how it behaves if I set #ParagraphPropertiesDialog or the sub vdialog to have 
```css
.ui-dialog-content .jsdialog.vertical[id^=dialog-vbox] {
  width: 624px !important;
}
```


https://github.com/user-attachments/assets/806dbc43-e3df-4d66-9d69-d7662b7dc789

pedropintosilva:
I wonder if @banobepascal could look at this and send an additional fix

eszkadev:
@pedropintosilva I wonder what is the goal here. Currently dialog has the size of biggest tab and it is static. No jumping/rescaling on tab switch. (the only exception is when tab has dynamic content which will appear only after activation)

banobepascal:
> @pedropintosilva I wonder what is the goal here. Currently dialog has the size of biggest tab and it is static. No jumping/rescaling on tab switch. (the only exception is when tab has dynamic content which will appear only after activation)

I think we are trying to have placement of the labels in close proximity to input fields,

eszkadev:
> > @pedropintosilva I wonder what is the goal here. Currently dialog has the size of biggest tab and it is static. No jumping/rescaling on tab switch. (the only exception is when tab has dynamic content which will appear only after activation)
> 
> I think we are trying to have placement of the labels in close proximity to input fields,

Possibly we can use some CSS grid tricks to do that inside inner grids? Without any hard width settings?

banobepascal:
> > > @pedropintosilva I wonder what is the goal here. Currently dialog has the size of biggest tab and it is static. No jumping/rescaling on tab switch. (the only exception is when tab has dynamic content which will appear only after activation)
> > 
> > 
> > I think we are trying to have placement of the labels in close proximity to input fields,
> 
> Possibly we can use some CSS grid tricks to do that inside inner grids? Without any hard width settings?

Agreed.

pedropintosilva:
That would be ideal

browser/src/control/jsdialog/Widget.Containers.ts
```diff
@@ -65,8 +65,8 @@ JSDialog.grid = function (
 		rows +
 		', auto); \
 		grid-template-columns: repeat(' +
-		cols +
```
eszkadev, line 68: why `cols - 1` ? I think it may break grid order eg. in sidebar?

hcvcastro:
Sorry guys I was away for a long time, Thanks for the comments.  Please @banobepascal assign the task, I have a lot pending task to update =)

pedropintosilva:
From 2025-03-11 meeting:
Present: @banobepascal , @pedropintosilva 

        ◦ Wondering to use switch case for setting different grid combinations → but that will not work as we have many places
        ◦ Probably best to attempt to do this on the CSS side only
        ◦ Mainly the case is setting: grid-template-columns: max-content auto max-content !important; 
        ◦ I would suggest to instead of going through every dialog (and add specific rules for specific childreen with a specific ID) accept Henry’s: https://github.com/CollaboraOnline/online/pull/9944/files#diff-3ea8022e4d737184b1a8ebd768c98b9660c2b6b603e00bce2d193ff99cb191ca  and add on top a dialog-s, dialog-m, dialog-l css classes (set in the JS side) where the default would be one of them (-m) and for any other dialog that needs more or less space we would change the class to -s or -l. We would have at the end 3 CSS declarations where we set the width of each one.
        ◦ There are a lot of dialogs: In core [...]LibreOffice/core-24 run:
            ▪ find . -type f -name "*dialog*.ui"

----

What are your thoughts @eszkadev ?

pedropintosilva:
2025-04-01

Present: Banobe, Pedro

Another way would be to either try to see if there are any further attributes that we can reuse from core side and make sure it comes in the json -> we checked, it doesn't look like there are

OR we could:

1.repeat(ncolumns, [100%/ncolumns]) on the very first level where there are grid(s) with more than 1 column

2. remove the width 1--%
```diff
.ui-tabpage.jsdialog {
- width: 100%;
}
```

3. The very first level where there are grid(s) with more than 1 colunm
grid-column-gap: 24px;

4. The very first level where there are grid(s) with more than 1 row
grid-row-gap: 24px;

Darshan-upadhyay1110:
@hcvcastro , can we close this PR? It’s been pending for quite a while and doesn’t seem to be you are working on this currenlty.

If you have a quick fix, we can proceed with that and merge it.

I’m just trying to clean up some of the long-pending PRs.

hcvcastro:
> @hcvcastro , can we close this PR? It’s been pending for quite a while and doesn’t seem to be you are working on this currenlty.
> 
> If you have a quick fix, we can proceed with that and merge it.
> 
> I’m just trying to clean up some of the long-pending PRs.

Yes, it's an old patch — best to close it. I think it was fixed anyway. Thanks.

PR #9946, opened 2024-08-30
Impress: Fix shape selection JS error.

gokaysatir:
Remove inline style from svg html string before using it.

Change-Id: I36d215e63e5e7947695b7da00c53ccf283ab82a9

There is a JS error if we use inline style for SVG elements. We remove it before assigning the SVG content.

To reproduce:
* Add a shape and write some text into it. Then select it.

### DIFF
--- a/browser/src/canvas/sections/ShapeHandlesSection.ts
+++ b/browser/src/canvas/sections/ShapeHandlesSection.ts
@@ -416,14 +416,44 @@ class ShapeHandlesSection extends CanvasSectionObject {
 		}
 	}
 
+	removeTagFromHTML(data: string, startString: string, endString: string): string {
+		let startIndex = data.indexOf(startString);
+		let endIndex = data.indexOf(endString, startIndex + 1);
+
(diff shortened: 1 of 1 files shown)


browser/src/canvas/sections/ShapeHandlesSection.ts
```diff
@@ -411,14 +411,48 @@ class ShapeHandlesSection extends CanvasSectionObject {
...
+		else {
+			return '';
+		}
+	}
+
 	setSVG(data: string) {
 		this.removeSVG();
 
+		/* To get / remove font tag. */
+		//const fontInfo = this.getTagFromHTML(data, '<font ', '</font>');
```
mcecchetti, line 446: what is the role of this comment out code ? is it used for testing ?

gokaysatir:
Thanks for checking :) Removed the commented out code, it looks better this way :)

mcecchetti (approved):
Looks good to me. Thanks

PR #9951, opened 2024-09-01
Calc: Convert print twips to tile before using point info for shape h…

gokaysatir:
…andles.

Change-Id: I59cf06e28f336aaf8f4d32627fd2df9ee4bf4fb8

When there is an image to the bottom rows, selection handles are offset.
This PR ensures they are correctly positioned.


gokaysatir:
The failing "unit-save-torture" test passes locally. Trying again.

mcecchetti (approved):
Looks good to me. Thanks!

PR #9952, opened 2024-09-02
Provide snprintf string formatting.

sgothel:
Provide snprintf string formatting.

Change-Id: I8097727936038d2b0b63a47ec3e195c7c934c638

* Resolves: Add missing convenient snprintf -> std::string, no issue reported
* Target version: master 

### Summary

Adds missing convenient snprintf -> std::string, no issue reported.

Despite snprintf formatting being the `ugly duckling` due to security concern,
at least in a controlled environment it provides an efficient way for string formatting.

Note C++20 `std::format` is not yet available on our current toolchains
and its evaluation has to be completed.

Note On a remotely related work path inspired by the discussion,
please find [argument type checker jau::cfmt](https://github.com/sgothel/jaulib/commit/4b886061fcf166bced932f9a8df8655ec4af2347) which may 
also likes to be integrated into our Util block for security reasons IFF we intend to 
use `snprintf` formatting in the first place.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -727,6 +727,44 @@ namespace Util
 #endif
     }
 
+    std::string vformatString(const char* format, va_list ap)
+    {
+        size_t nchars;
+        std::string str;
+        {
+            const size_t bsz = 1024; // including EOS
+            str.reserve(bsz); // incl. EOS
+            str.resize(bsz - 1); // excl. EOS
+
+            nchars = ::vsnprintf(&str[0], bsz, format,
+                                 ap); // NOLINT(clang-analyzer-valist.Uninitialized): clang-tidy bug
+            if (nchars < bsz)
+            {
+                str.resize(nchars);
+                str.shrink_to_fit();
+                return str;
+            }
+        }
+        {
+            const size_t bsz = std::min<size_t>(nchars + 1, str.max_size() + 1); // limit incl. EOS
+            str.reserve(bsz); // incl. EOS
+            str.resize(bsz - 1); // excl. EOS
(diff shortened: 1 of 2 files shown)


mmeeks (commented):
Looks nice enough to me - although passing arguments via va_list is not terribly type safe, and we're trying to write rather robust code here. What's the use-case?

sgothel:
> Looks nice enough to me - although passing arguments via va_list is not terribly type safe, and we're trying to write rather robust code here. What's the use-case?

It is just a proposal, cut-out of my earlier change-set. Used to simply format a string w/o ostringstream.
The quality goals and security issues of *snprintf() are well known hence the discussion in the description.

For internal formatting it should be fine to use, however, one may also find the save snprintf
compile time statement/snippet mentioned below of interest .. if .. interested. Otherwise, we just throw it away.

> Note On a remotely related work path inspired by the discussion,
> please find [argument type checker jau::cfmt](https://github.com/sgothel/jaulib/commit/4b886061fcf166bced932f9a8df8655ec4af2347) which may
> also likes to be integrated into our Util block for security reasons IFF we intend to
> use snprintf formatting in the first place.

mmeeks:
Well - if we come up with a performance or correctness reason to do this; lets re-open; but having spent a chunk of my past life unwinding va_args pointers I'd suspect that the C++ way is safer - though less arguably less performant - and safety - particularly in coolwsd is paramount. For our logging - we don't hit those code-paths unless that is enabled, so no problem, and we don't translate logging. So lets close this until we need it :-)

PR #9955, opened 2024-09-02
calc: expose "Select Range" and "Refresh Range" buttons

Rash419:
Change-Id: I5e533b7bd128fea77df702c68a0c22d77b2bf432

* Make Select Range dialog async - https://gerrit.libreoffice.org/c/core/+/172626
* Target version: master


gokaysatir (approved):
Thanks :) Works with:
* https://gerrit.libreoffice.org/c/core/+/172626

PR #9960, opened 2024-09-02
remove apparmor profile that depends in abi 4.0 if that is not available

caolanm:
Change-Id: I40e3b4069486141df7e01ba8df480fd9a4820e91

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/debian/coolwsd.postinst.in
+++ b/debian/coolwsd.postinst.in
@@ -3,6 +3,15 @@
 set -e
 
 case "$1" in
+    install|upgrade|configure)
+        # If abi 4.0 is not present, then remove the apparmor profile config
+        if [ ! -e /etc/apparmor.d/abi/4.0 ]; then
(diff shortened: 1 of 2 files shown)


timar (approved):

caolanm:
There doesn't seem to be a particularly good way to query the apparmor abi version, but /etc/apparmor.d/abi/4.0 (and /etc/apparmor.d/abi/3.0) exist on ubuntu 24.04  while debian 12 has just /etc/apparmor.d/abi/3.0, so that looks a practical way to determine if there is apparmor abi 4.0 support.

Given that this is a config file, it persists from the old version even if it wasn't installed by a new version. So explicitly removing it will repair existing installs as well as avoid the issue in new upgrades while retaining it for the Ubuntu 24.04 case that does have 4.0

PR #9969, opened 2024-09-03
config: remove font color value and use fallback 

lpranam:
problem:
font color did not account for the alpha channel
due to not having alpha channel, some conditions would evaluate false in core even if they are same color
i.e: fontColor == COL_AUTO because (COL_AUTO contains alpha channel)

Change-Id: I4c1d62a5203ae4efd03244a4c7fa9d0999a365dd

* Target version: master 

### Checklist

### DIFF
--- a/coolkitconfig-mobile.xcu
+++ b/coolkitconfig-mobile.xcu
@@ -83,11 +83,6 @@
                 <value>true</value>
             </prop>
         </node>
-        <node oor:name="FontColor">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>0</value>
-            </prop>
-        </node>
         <node oor:name="Links">
             <prop oor:name="Color" oor:op="fuse">
                 <value>128</value>
@@ -400,11 +395,6 @@
             </prop>
             <prop oor:name="IsVisible" oor:op="fuse">
                 <value>true</value>
-            </prop>
-        </node>
-        <node oor:name="FontColor">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>0</value>
             </prop>
         </node>
         <node oor:name="Links">
--- a/coolkitconfig.xcu
+++ b/coolkitconfig.xcu
@@ -96,11 +96,6 @@
                 <value>true</value>
             </prop>
         </node>
-        <node oor:name="FontColor">
-            <prop oor:name="Color" oor:op="fuse">
-                <value>0</value>
-            </prop>
-        </node>
         <node oor:name="Links">
             <prop oor:name="Color" oor:op="fuse">
                 <value>128</value>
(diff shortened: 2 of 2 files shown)


caolanm:
I'm not convinced this is the right way to go about it, I feel that at this config level that alpha/transparency isn't really supported and this probably ends up as accidentally equivalent to COL_AUTO in LibreOffice.

What might work better is in core-side at sc/source/ui/app/scmod.cxx where we do aViewRenderingOptions.SetDocColor(aFillColor);
to add a 
aViewRenderingOptions.SetFontColor(); from the current (then correct) value of svtools::FONTCOLOR to save that in the ViewShell RenderingOptions (and add the required member to ScViewRenderingOptions (and accesser's etc)) and then like we do in ScOutputData::DrawSelectiveObjects to get the current shell rendering options for the background color
of ...

const ScTabViewShell* pTabViewShellBg = ... ScTabViewShell::GetActiveViewShell() and
    const ScViewRenderingOptions& rViewRenderingOptions = pTabViewShellBg->GetViewRenderingData();
   = ... rViewRenderingOptions.GetDocColor();

that we try doing similar, except also with that new foreground color, in ScDrawStringsVars::ScDrawStringsVars, so it uses colors from the current ScTabViewShell (if it exists) which should have the desired colors in it, instead of the global svtools::ColorConfig.

caolanm:
This works, but what I think happens in this case is that then svtools::FONTCOLOR has no special configured value so ColorConfig::GetColorValue(svtools::FONTCOLOR, bSmart = true) is called and gets AUTO first, then ColorConfig::GetDefaultColor(svtools::FONTCOLOR) and returns back cAutoColors[FONTCOLOR][dark/light] where cAutoColors has
{ COL_BLACK,        COL_BLACK       }, // FONTCOLOR
so its just always black that is returned

So in ScPatternAttr::fillColor the fontcolor to use on a background then defaults to black unless there is an unambiguous dark background in which case it forces white.

lpranam:
> So in ScPatternAttr::fillColor the fontcolor to use on a background then defaults to black unless there is an unambiguous dark background in which case it forces white.

isn't that what we are looking for? at least that's what I was trying to do...

caolanm:
Yeah, in the specific calc use case we're thinking about we want that result, but I worry about other cases that might have ended up depending on this setting (and I see that there is a coolkitconfig-mobile.xcu which is out of sync with coolkitconfig.xcu for these values too.

Looking at core I think these colors have gotten a little confused and it would make most sense to me that FONTCOLOR should only be used on a DOCCOLOR background where we know that is basically in use because background is COL_AUTO. I think maybe safest in short run to change the calc pattern text color code to basically not use FONTCOLOR at all. And medium term remove the FontColor from both places it is used in this xcu, and the other xcu too.

caolanm (changes requested):
lets also remove the FontColor from the "light" section too for consistency, and I see similar in coolkitconfig-mobile.xcu that look like copy and paste from this original

caolanm:
That looks good to remove the light one too.

But I see there is also a 'coolkitconfig-mobile.xcu' that also has these "FontColor" settings. It probably makes sense to remove these from there too (cc @Minion3665 to be aware of the existing discrepancy of these "FontColor" settings between mobile and non-mobile for the 'dark' case)

caolanm (approved):
looks sane to me, but we'll wait for red window

PR #9973, opened 2024-09-04
Remove inline css from helpdialog

Darshan-upadhyay1110:
- Remove remove css from cool-help.html and replaced with class attrs
- moved all style tag css to file helpdialog.css
- fix typo for icon `fold`

Change-Id: I91076904a58e9d8e2a0952836f5073d744b715fe

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -584,6 +584,22 @@ L.Map.include({
 				this.filterResults(searchTerm, isAnyMatchingContent, id);
 			}
 		}.bind(this));
+
+
+		const onlineHelpContent = document.getElementById('online-help-content');
+		const buttons = onlineHelpContent.querySelectorAll('.scroll-button');
+
+		buttons.forEach((button) => {
+			button.addEventListener('click', () => {
+				const targetId = button.dataset.target;
+				if (targetId) {
+					const targetElement = document.getElementById(`${targetId}`);
+					if (targetElement) {
+						targetElement.scrollIntoView();
+					}
+				}
+			});
+		});
 	},
 
 
--- a/browser/src/control/Toolbar.js
+++ b/browser/src/control/Toolbar.js
@@ -443,22 +443,22 @@ L.Map.include({
 		var i;
 		// Display keyboard shortcut or online help
 		if (id === 'keyboard-shortcuts-content') {
-			document.getElementById('online-help-content').style.display='none';
+			document.getElementById('online-help-content').classList.add('hide');
 			// Display help according to document opened
 			if (map.getDocType() === 'text') {
-				document.getElementById('text-shortcuts').style.display='block';
+				document.getElementById('text-shortcuts').classList.add('show');
 			}
 			else if (map.getDocType() === 'spreadsheet') {
-				document.getElementById('spreadsheet-shortcuts').style.display='block';
+				document.getElementById('spreadsheet-shortcuts').classList.add('show');
 			}
 			else if (map.getDocType() === 'presentation') {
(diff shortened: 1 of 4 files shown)


eszkadev:
@gokaysatir could you review?

gokaysatir (changes requested):
Thanks @Darshan-upadhyay1110 :)

CSS issues seem solved. Can you also add a JS file for help page and bind "onlick" event handlers (and other event handlers if any) inside that JS file?

I think giving "p" tags ids and using those ids on page load to assign event handlers should work. Anyway, implementation is up to you:)

browser/css/toolbar.css
```diff
@@ -1016,7 +1016,7 @@ button.leaflet-control-search-next
 .w2ui-icon.link{ background: url('images/lc_inserthyperlink.svg') no-repeat center; }
 .w2ui-icon.insertsymbol{ background: url('images/lc_insertsymbol.svg') no-repeat center; }
 .w2ui-icon.edit{ background: url('images/lc_editdoc.svg') no-repeat center; }
-.w2ui-icon.fold{ background: url('images/fold.svg') no-repeat center center / 19px; opacity: 0.8; filter: alpha(opacity=80);}
```
pedropintosilva, line 1019: isn't this change already present in https://github.com/CollaboraOnline/online/pull/9900 ? Shouldn't we just merge that first?

browser/css/toolbar.css
```diff
@@ -1305,7 +1305,7 @@ button.leaflet-control-search-next
 [data-theme='dark'] .w2ui-icon.link{ background: url('images/dark/lc_inserthyperlink.svg') no-repeat center; }
 [data-theme='dark'] .w2ui-icon.insertsymbol{ background: url('images/dark/lc_insertsymbol.svg') no-repeat center; }
 [data-theme='dark'] .w2ui-icon.edit{ background: url('images/dark/lc_editdoc.svg') no-repeat center; }
-[data-theme='dark'] .w2ui-icon.fold{ background: url('images/dark/fold.svg') no-repeat center center / 19px; opacity: 0.8; filter: alpha(opacity=80);}
```
pedropintosilva, line 1308: isn't this change already present in https://github.com/CollaboraOnline/online/pull/9900 ? Shouldn't we just merge that first?

Darshan-upadhyay1110: @pedropintosilva oh i think i missed it let's first merge #9900 and then i rebase this

browser/src/control/Toolbar.js
```diff
@@ -584,6 +584,21 @@ L.Map.include({
 				this.filterResults(searchTerm, isAnyMatchingContent, id);
 			}
 		}.bind(this));
+
+
+		const buttons = document.querySelectorAll('.scroll-button');
```
gokaysatir, line 589: Just fyi: We can use something like:
document.getElementById('online-help-content').querySelectorAll(':scope .scroll-button');

So it would pick only the elements inside the 'online-help-content' elements with class '.scroll-button'.

Didn't test this but it should work.

browser/src/control/Toolbar.js
```diff
@@ -584,6 +584,21 @@ L.Map.include({
...
+
+		buttons.forEach((button) => {
+			button.addEventListener('click', () => {
+				const targetId = button.getAttribute('data-target');
```
gokaysatir, line 593: "button.dataset.target" should also work :) That is the strong side of these attributes i guess.

gokaysatir (approved):
Thanks for the fixes :)

PR #9974, opened 2024-09-04
Related: cool#9735 kit: make sure we don't process events in the any input callback

vmiklos:
Commit 930cd616fb6defd8c0de0163bd92e637c7b6b76f (Related: cool#9735 kit:
use the new registerAnyInputCallback() LOK API, 2024-08-26) started to
interrupt Writer layout and other idle tasks when we have pending
events.

The intent was to check if we have pending events in anyInputCallback(),
but it turns out that poll() not only checks if there are events to be
processed but also potentially processes them, which I didn't realize
when I did the above commit.

Fix the problem by only checking the callbacks list and the tile queue
in anyInputCallback(): that is enough to keep the interactivity win and
makes sure to avoid a re-entrancy hazard there.

The 300 pages bugdoc with its numbered list is still laid out in 2 steps
(visible page layout, repaint, rest of the layout) after this.

Change-Id: I7632d8dcfa29ddbf4d09b064bc445bb0ea7e7f13

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2875,19 +2875,31 @@ int pollCallback(void* pData, int timeoutUs)
 bool anyInputCallback(void* data)
 {
     auto kitSocketPoll = reinterpret_cast<KitSocketPoll*>(data);
-    int ret = kitSocketPoll->poll(std::chrono::microseconds(0));
     std::shared_ptr<Document> document = kitSocketPoll->getDocument();
-    if (document)
+    if (!document)
     {
-        std::shared_ptr<KitQueue> queue = document->getQueue();
-        if (!document->hasCallbacks() && queue && queue->getTileQueueSize() == 0)
-        {
-            // Have no pending callbacks and the tile queue is also empty, report that we have no
-            // pending input events.
-            ret = 0;
-        }
+        return false;
     }
-    return ret > 0;
+
+    if (document->hasCallbacks())
+    {
+        return true;
+    }
+
+    std::shared_ptr<KitQueue> queue = document->getQueue();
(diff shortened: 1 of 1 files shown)


vmiklos:

Facepalm for just assuming what poll() does in this context and getting it wrong.

caolanm (approved):

mmeeks:
So - really not as silly as you might thing; AnyInput would typically call XPending / XEventsQueued - which ultimately will do a non-blocking read from the socket and queue the result.

So it's not silly to do a poll() there to slurp from the incoming socket(s) - indeed it could be rather helpful. The problem really is that we want that slurp to be certain not to create any re-entrancy hazards, and just to queue events it reads - which - might also be the case; I just wanted to check that we'd done the code-reading there =)

Either way the change is not terrible either as-is =)

mmeeks:
I fear now we only interrupt idle layout when we've finished it ;-)

vmiklos:
Sorry, your comment went to spam... Checking back my notes I had this for the paste case:

16:51:25.204 global.js:545 1725375085204 OUTGOING: uno .uno:Paste
16:51:25.216 global.js:545 1725375085216 INCOMING: delta: nviewid=1000 part=0 width=256 height=256 tileposx=3840 tileposy=0 tilewidth=3840 tileheight=3840 oldwid=1 wid=46316 ver=872578 imgsize=3141 -> 12ms

And that 12ms was 274ms for me before the AnyInput callback. So I think we're OK here: we get the speedup even without the poll(), and then we can be sure there is no re-entrancy hazard here.

mmeeks:
Nice =)

PR #9975, opened 2024-09-04
refactor(treewide): Enable strict null checking

Minion3665:
Typescript defaults to treating null/undefined in a way that can leak them throughout code, and will also cause type issues (unexpected undefineds/anys) if you use them explicitly

To avoid an unreviewable mess I have either widened types to allow null (where we own the type) or added non-null assertions (where we do not). This is preferable to not discerning nulls at all, and while it still creates a large (by LOC) change, it's hopefully not large (by variation of things) - and should transpile identically to JavaScript.

There is one point where the code does not transpile identically to JavaScript. This is because it directly set something that we don't own which shouldn't be undefined to undefined. This is clearly incorrect, and I've replaced it with setting to an empty string instead.

If it seems like this is bad practice, that's because it is. You shouldn't add non-null assertions all over the place because it makes typescript lie to you. It's of similar badness to adding `as x` everywhere. Unfortunately, worse practice would be mandating that typescript lie to you about nulls and undefineds everywhere, as our tsconfig has been doing. It's far better to limit the lie to these ~500 places where I've had to explicitly tell typescript "the world isn't actually like this - that's OK, carry on anyway".

If you're worried about blames from this change - first: I believe it's worth it, second: this change should only affect single lines, so using --ignore-rev should work very well on it

In the future, probably we should remove some of these assertions, although some changes we'll have to make to remove them are likely either tedious or dangerous.

Change-Id: I1770a1a8be3f45b45bda0fd2bcd57563add25924

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/tsconfig.json
```diff
@@ -12,7 +12,8 @@
       "removeComments": false,
       "preserveConstEnums": false,
       "downlevelIteration": true,
-      "watch": false,
```
Minion3665, line 15: I removed this because it has no effect - we would have to use https://www.typescriptlang.org/docs/handbook/configuring-watch.html#configuring-file-watching-using-a-tsconfigjson if we wanted to configure watch options. Although if this is false (the default) it's doubly-ineffective. It's not strictly related to strictNullChecks

Minion3665:
CI fails because
> ESLint found too many warnings (maximum: 0).

I think I need to set up https://github.com/phenomnomnominal/betterer or something at this point - because I don't really want to set the max warnings to 842 🥲

Minion3665:
@eszkadev - as the CI failure shouldn't require me changing the typescript again, I've requested a review now (I suspect this is both something that may be annoying to continually rebase and something that will take a long time to look through - despite changing very little about the emitted code). I apologize for the size of the diff and all the extra warnings this causes - I think it's preferable to not having them and unfeasible to fix them all at once - but it's still not great.

Minion3665:
I guess the other thing I should mention here is that this was a side-quest on my way to doing #8221 (because I needed to type something as null at one point... and it was leaking an unknown into a place it shouldn't have been which made me realize this option was off)

test/samples/calc-edit.ods
vmiklos, line 1: I believe test/samples/*edit* files should not be added to git.

Minion3665:
Superseded by https://github.com/CollaboraOnline/online/pull/11527

PR #9978, opened 2024-09-04
enable cleanup of namespacing using jails

caolanm:
They used namespace mounts for / so that was both unmounted automatically when the child exited, and not visible to the parent anyway, so there is no "lo" dir to check for.

Change-Id: Ia98032abaae8a33abd57cef38bdc97f6ff4fb0c2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -116,18 +116,20 @@ bool enterUserNS(uid_t uid, gid_t gid)
 #endif
 }
 
-bool coolmount(const std::string& arg, std::string source, std::string target)
+bool coolmount(const std::string& arg, std::string source, std::string target, bool silent = false)
 {
     source = Util::trim(source, '/');
     target = Util::trim(target, '/');
 
     if (isMountNamespacesEnabled())
     {
-        const char *argv[4];
+        const char *argv[5];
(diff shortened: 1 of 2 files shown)


common/JailUtil.cpp
```diff
@@ -267,33 +269,59 @@ void removeAuxFolders(const std::string &root)
     FileUtil::removeFile(Poco::Path(root, "linkable").toString(), true);
 }
 
+/*
+    The tmp dir of a path/<jailid>/tmp is mounted from (or linked to) a
+    path/tmp/cool-<jailid> dir. In a mount namespace case the existence
+    of path/<jailid>/tmp is not visible to the parent process so its
+    contents cannot be removed via the path/<jailid>/tmp view, and
+    in any case the path/<jailid>/tmp should to be removed.
```
mmeeks, line 277: should not be removed ? =)

mmeeks (approved):
Looks plausible; thanks Caolan :-)

caolanm:
"should be removed" was what I had in mind. I see this tmp/cool-jailid dir left behind in either case (but empty for non-namespace mode).

Issue #9730, opened 2024-08-07
localization about chinese-simple（zh-cn） and chinese-Traditional （zh-tw）


pandagxnu:
### Describe the Bug
The match of User interface localization is not suitable.

### Steps to Reproduce

1. apt install collaboraofficebasis-zh-cn   collaboraofficebasis-zh-tw
2. wopiUrl with "&lang=zh-cn"
3. open webbrowser
4. User interface language is Chinese-simple , but styles ( like 'Title' 'Heading' etc ) is still English.

![1](https://github.com/user-attachments/assets/d2910cfe-4a85-4c77-b640-f3154ce9f13a)

6. wopiUrl with "&lang=zh-tw"
7. User interface language is Chinese-traditional, but styles ( like 'Title' 'Heading' etc ) is Chinese-simple.
![2](https://github.com/user-attachments/assets/c249d0b7-191f-4737-ada0-6d5ad6a7c935)

8. After the following operations，and wopiUrl with "&lang=zh-tw" ，I got expected user interface。

\#  cd /opt/collaboraoffice/program/resource
\#  cp -r zh_TW/ zh_CN
\#  systemctl restart coolwsd.service
![3](https://github.com/user-attachments/assets/0e59ca70-cdc0-4c19-80fe-417cd9673b31)

### version
 - **COOLWSD version:** [24.04.5.2]
 - **Collabora version:** [Collabora Office 24.04.6.1]
 - **OS and version:** [Debian GNU/Linux 12 (bookworm)]

timar:
In the zh-CN case I see in the log:
`warn:lok:34566:33687:sfx2/source/view/lokhelper.cxx:203: LANGUAGE mismatch at setView! ... old (wrong) lang:zh-CN new lang:en-US`
In the zh-TW case I do not see such message and I cannot confirm the bug! I see zh-TW text in styles box.
![image](https://github.com/user-attachments/assets/440ed5fd-1b0a-4a98-9819-43878b4b49c5)

timar:
In the zh-CN case the styles box is not the only issue. The sidebar is also partially localized.
![image](https://github.com/user-attachments/assets/935bc4e4-d93d-4ceb-af1f-dadb83817152)

timar:
[...]

PR #9979, opened 2024-09-04
Workaround for broken integrations vs. LOKit language fallback (fixes…

timar:
… #9730)

Change-Id: Ief8fe411e800f2eb0af2d05171a732742044765a

* Resolves: #9730 
* Target version: master

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1606,8 +1606,20 @@ function getInitializerClass() {
 	}
 
 	var lang = global.coolParams.get('lang');
-	if (lang)
+	if (lang) {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
OK, though I wonder: is there a case where you don't want the Y part of X-Y as uppercase?

That would be more general than the few languages you listed here.

timar:
> OK, though I wonder: is there a case where you don't want the Y part of X-Y as uppercase?
> 
> That would be more general than the few languages you listed here.

The list is complete (in terms of supported core languages). I'm not proud of the beauty of this code, but...

PR #9983, opened 2024-09-04
slideshow: improvements

eszkadev:
typescript conversion of engine moved from core, console error fixes, flickering fix


eszkadev:
fails on cpp unit tests which also fail in other PRs...

codewithvk (approved):
Works fine.

PR #9986, opened 2024-09-04
Reduce temporary copies of large buffers

plubius:
Also, use a local autorelease pool to force destruction of remaining temporary buffers immediately after use instead of waiting for the thread's current autorelease pool to drain.

Lastly, just to be safe, include the pretext and posttext lengths when allocating the vector's capacity.


Minion3665 (approved):
looks good and works when tested, thanks very much!

PR #9987, opened 2024-09-05
test: Add betterer

Minion3665:
Betterer is a snapshot test runner which tries to reduce the number of errors. With, for example, eslint, we need to get to 0 errors for the test to pass. This is prohibitive for large changes (e.g. turning on a new eslint option), as you then have to make a lot of changes in the same pull request. For betterer, you only have to avoid making the test fail *more*, so it's OK to fix errors or leave things alone.

---

It runs as a part of check, and outputs (but doesn't block) building.

You can also run it manually, in the same way as prettier or eslint

$ make betterer

And get it to update its snapshots, again in the same way as prettier

$ make betterer-write

---

I avoided adding the betterer precommit, as running eslint takes a long while and this blocked the commit for longer than I considered acceptable. There's probably a way to make this work with eslint-d and a custom test, but even so it's likely custom, brittle and non-ideal.

---

This is a prerequisite for #9975, which is a prerequisite for #8221

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -730,6 +730,12 @@ prettier: browser/node_modules
 prettier-write: browser/node_modules
 	@make -C browser prettier-write
 
+betterer: browser/node_modules
+	@make -C browser betterer
+
+betterer-write: browser/node_modules
+	@make -C browser betterer-write
+
 install-exec-hook:
 	cd $(DESTDIR)$(bindir) && \
 	$(LN_S) coolconfig loolconfig && \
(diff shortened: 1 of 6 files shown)


Minion3665:
Both things this is a prerequisite of have been closed, I'm not sure we need this, and it has various disadvantages (such as requiring a file to be committed with the history...) so closing this for now

PR #9993, opened 2024-09-05
Fix context submeny arrow key color

Darshan-upadhyay1110:
- Context sub-menu arrow key color was same as menu item bg color. Replaced with appropriate color variable

Before:
image

After:
image

Change-Id: I57bff4640156fa52f92ce1678caa364218349b9c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (approved):
Thanks, looks good. We need to do this otherwise the rule that comes from jquery.contextMenu.css `transparent transparent transparent #2f2f2f` is applied and with it the hard coded color

PR #9996, opened 2024-09-05
wip: global.js: convert to typescript

Minion3665:
This supersedes #8221


Minion3665:
We've started doing this more incrementally with global.d.ts, so closing: https://github.com/CollaboraOnline/online/blob/master/browser/src/global.d.ts

PR #9997, opened 2024-09-05
fix(debug): Correct newlines

Minion3665:
Previously we were using \<br/> tags to do newlines in debug mode. In https://github.com/CollaboraOnline/online/pull/9980, we changed from innerHTML to innerText - which is generally a good idea. Unfortunately, it also broke the HTML tags. We can use '\n' for newlines instead.

We also used \<b>, but that wouldn't be possible without introducing extra complexity into the debug overlay.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks :)

PR #9999, opened 2024-09-06
Fix keyboard shortcuts not working outside map region

Darshan-upadhyay1110:
- Save and find events were not considered as global event, which is the reason of Keyboard shortcuts like CTRL + S or CTRL + F are not working outside `map`
- this patch will fix that issue. Now user can get focus on search input from anywhere on the document also document save will also work as expected.
- problems like after CTRL + S sometimes we got web browser save popup will be fixed by this patch

 +++ Problem +++

- CTRL + S and Ctrl + F are not working before this patch for cases like
    - Press alt (to activate the accesskey tooltip) in the notebookbar
    - Place the cursor inside of the document name Press ctrl + S
    - open About dialog
        - Press ctrl + F it triggers in the browser search and not the status bar.
        - Press ctrl + S it triggers in the browser save and not the COOL save action.

Change-Id: I1e1f5aa59b1c47e7e888198568e45e8a3a84843c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/dom/NotebookbarAccessibility.js
+++ b/browser/src/dom/NotebookbarAccessibility.js
@@ -272,6 +272,12 @@ var NotebookbarAccessibility = function() {
 		}
 	};
 
+	this.onInputKeyDown = function(event) {
+		if (event.ctrlKey) {
+			this.resetState();
+		}
+	};
+
 	this.onInputKeyUp = function(event) {
 		var key = event.key.toUpperCase();
 		event.preventDefault();
@@ -395,6 +401,7 @@ var NotebookbarAccessibility = function() {
 		this.accessibilityInputElement.onfocus = this.onInputFocus.bind(this);
 		this.accessibilityInputElement.onblur = this.onInputBlur.bind(this);
 		this.accessibilityInputElement.onkeyup = this.onInputKeyUp.bind(this);
+		this.accessibilityInputElement.onkeydown = this.onInputKeyDown.bind(this);
 		this.accessibilityInputElement.autocomplete = 'off';
 
 		var container = document.createElement('div');
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -397,6 +397,10 @@ L.Map.Keyboard = L.Handler.extend({
 		if (this._map.uiManager.isUIBlocked())
 			return;
 
+		if (window.KeyboardShortcuts.processEvent(app.UI.language.fromURL, ev)) {
+			ev.preventDefault();
+			return;
+		}
 		if (this._map.jsdialog
 			&& (this._map.jsdialog.hasDialogOpened() || this._map.jsdialog.hasSnackbarOpened() || this._map.jsdialog.hasDropdownOpened())
 			&& this._map.jsdialog.handleKeyEvent(ev)) {
@@ -436,23 +440,6 @@ L.Map.Keyboard = L.Handler.extend({
 				this._map._docLayer._preview.partsFocused = false;
 			}
 		}
-		else if (this._isCtrlKey(ev) && !ev.shiftKey && !ev.altKey && ev.keyCode === this.keyCodes.F) {
-			if (app.UI.language.fromURL === 'de' && this._map.getDocType() === 'text') {
-				this._map.sendUnoCommand('.uno:Navigator');
-			}
-			else {
-				if (!this._map.uiManager.isStatusBarVisible()) {
-					this._map.uiManager.showStatusBar();
-				}
-				this._map.fire('focussearch');
-			}
-
-			ev.preventDefault();
-		}
-		else if (this._isCtrlKey(ev) && !ev.shiftKey && ev.altKey && ev.keyCode === this.keyCodes.S && app.UI.language.fromURL === 'de' && this._map.getDocType() === 'text') {
-			this._map.fire('focussearch');
-			ev.preventDefault();
-		}
 		else if (this._isCtrlKey(ev) && ev.keyCode === this.keyCodes.S) {
 			// Save only when not read-only and when HideSaveOption is false.
 			if (!this._map.isReadOnlyMode() && !this._map['wopi'].HideSaveOption) {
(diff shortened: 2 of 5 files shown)


mmeeks:
Have you reviewed all keydown events to ensure that this works as expected ?
Are we sure that eg. in Ctrl-f find popup that key-event / bubbling works properly ?
Are we certain that if the focus is in a toolbar - and we type a key- it doesn't end-up editing in the document ?

I'm suspicious of the simplicity of the fix TBH =) if it's been very heavily tested for un-expected side-effects and there are none then ok but ... =)

Darshan-upadhyay1110:
>  Are we certain that if the focus is in a toolbar - and we type a key- it doesn't end-up editing in the document ?

Oh, I missed this point :( We should not perform any operations like bold, select all using keyboard event when focus is outside the map area.

Thanks, @mmeeks points are really helpful.

/me need to add some conditional things for save and find...

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -837,14 +847,7 @@ L.Map.Keyboard = L.Handler.extend({
...
 		case this.keyCodes.S: // s
-			// Save only when not read-only and when HideSaveOption is false.
-			if (!this._map.isReadOnlyMode() && !this._map['wopi'].HideSaveOption) {
-				this._map.fire('postMessage', {msgId: 'UI_Save', args: { source: 'keyboard' }});
-				if (!this._map._disableDefaultAction['UI_Save']) {
-					this._map.save(false /* An explicit save should terminate cell edit */,
-					               false /* An explicit save should save it again */);
-				}
-			}
+			// Handle save event in globalEventHandler
```
eszkadev, line 850: if we don't handle anything here, shouldn't we just skip this?

eszkadev: true probably means - I handled that

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -436,6 +436,37 @@ L.Map.Keyboard = L.Handler.extend({
 				this._map._docLayer._preview.partsFocused = false;
 			}
 		}
+		else if (this._isCtrlKey(ev) && !ev.shiftKey && !ev.altKey && ev.keyCode === this.keyCodes.F) {
```
eszkadev, line 439: this is example of command we should put into KeyboardShortcuts.ts (in separate commit):
it has easy way to define language specific shortcuts: https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts#L192

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -436,6 +436,37 @@ L.Map.Keyboard = L.Handler.extend({
...
+			else {
+				if (!this._map.uiManager.isStatusBarVisible()) {
+					this._map.uiManager.showStatusBar();
+				}
+				this._map.fire('focussearch');
+			}
+
+			ev.preventDefault();
+		}
+		else if (this._isCtrlKey(ev) && !ev.shiftKey && ev.altKey && ev.keyCode === this.keyCodes.S && app.UI.language.fromURL === 'de' && this._map.getDocType() === 'text') {
```
eszkadev, line 452: another one for keybaord shortcuts

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -436,6 +436,37 @@ L.Map.Keyboard = L.Handler.extend({
...
+			this._map.fire('focussearch');
+			ev.preventDefault();
+		}
+		else if (this._isCtrlKey(ev) && ev.keyCode === this.keyCodes.S) {
```
eszkadev, line 456: this one can be leaved as it is as it is very important one and language-neutral

browser/src/map/handler/Map.Keyboard.js
Darshan-upadhyay1110: https://github.com/CollaboraOnline/online/pull/9999/files/1ccfe139e3d1c29ee565fc6cf72a105216920178#diff-5c4df738e4e71c87f575a954037607d6ffac23b63686d7de65dd029bca95eed6R533

@eszkadev it's a complicated implementaion of this._handleCtrlCommand. if we just skip it then it tries to exucute  rest  and we encounter erros like "User has no permission to save" or frquently we get "Content overwrite by other user overtite the current changes..." so it's a better way if we just return true and let _globalKeyEventHandler to handle "save"

browser/src/dom/NotebookbarAccessibility.js
```diff
@@ -272,6 +272,12 @@ var NotebookbarAccessibility = function() {
 		}
 	};
 
+	this.onInputKeyDown = function(event) {
+		if (event.ctrlKey) {
+			this.resetState();
```
gokaysatir, line 277: Why do we need this and also the event binding?

gokaysatir:
One question, other than it looks good. Thanks :)

browser/src/dom/NotebookbarAccessibility.js
Darshan-upadhyay1110: For cases like Accessibility keys are activated ( Yellow boxes)

- then if use tries to save document ( CTRL + S ) or Find ( CTRL + F) 
- it works fine but there will a view JUMP from Home Tab to other Tab in notebookbar
- And the reason is we disable Accessibility on KeyUp so for some special keys like CTRL we should disbale the state on keyDown because in some cases other operations will execute before our accessibility fuinctions get disabled 

I have made a separate commit on this as well 

https://github.com/CollaboraOnline/online/pull/9999/commits/c3ff6cf5c64a6961f2270660fdb52707fd52ffff

gokaysatir: Thanks for the details :)

gokaysatir:
Please merge in code RED. Just to remind.

gokaysatir (approved):

oleua:
When switching to Ukrainian keyboard layout the shortcut Ctrl-F doesn't work, while others, eg Ctrl-H works fine.

PR #10003, opened 2024-09-06
Fix JS error when opening mobile wizard.

gokaysatir:
There is an error related to inline styles. This commit fixes the error.

Change-Id: I2eb1ca65e829d36c9fb80981dbf82980a8f2d173

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.MobileWizardWindow.js
+++ b/browser/src/control/Control.MobileWizardWindow.js
@@ -98,7 +98,7 @@ L.Control.MobileWizardWindow = L.Control.extend({
 		if (this.isSnackBar)
 			this.mobileWizard.addClass('snackbar');
 		if (this.isBusyPopUp)
-			this.mobileWizard.addClass('busypopup'); 
+			this.mobileWizard.addClass('busypopup');
 	},
 
 	/// resets all classes which can modify the look to the original values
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.MobileWizardWindow.js
```diff
@@ -148,14 +148,27 @@ L.Control.MobileWizardWindow = L.Control.extend({
 		this._applySpecialClasses();
 	},
 
+	_getScrollIndicator() {
```
eszkadev, line 151: please rename `get` to `create`, what better describes what it does

browser/src/control/Control.MobileWizardWindow.js
```diff
@@ -165,8 +178,8 @@ L.Control.MobileWizardWindow = L.Control.extend({
 			var height = wizard.prop('scrollHeight');
 			var contentHeight = wizard.prop('clientHeight');
 			var scrollLeft = height - mWizardContentScroll;
-			if (scrollLeft < contentHeight + 1 || !that.isVisible) { that.scrollIndicator.css('display','none'); }
-			else { that.scrollIndicator.css('display','block'); }
+			if (scrollLeft < contentHeight + 1 || !that.isVisible) { that.scrollIndicator.style.display = 'none'; }
+			else { that.scrollIndicator.style.display  = 'block'; }
```
eszkadev, line 182: two spaces before "="

eszkadev (approved):
code looks ok

PR #10004, opened 2024-09-06
Wrap about dialog elements inside a "dialog" element.

gokaysatir:
Change-Id: Ia0c40a1d094d62af1a1c4471f14ead01cfe470aa

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -180,35 +180,37 @@ m4_ifelse(MOBILEAPP,[true],
     </div>
 
     <!-- Remove if you don't want the About dialog -->
-    <div id="about-dialog" tabIndex="0">
-      <div id="about-dialog-header">
-        <fig id="integrator-logo"></fig>
-        <h1 id="product-name">Collabora Online</h1>
-      </div>
-      <hr/>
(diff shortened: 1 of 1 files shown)


browser/html/cool.html.m4
```diff
@@ -180,35 +180,37 @@ m4_ifelse(MOBILEAPP,[true],
...
-      <div id="about-dialog-header">
-        <fig id="integrator-logo"></fig>
-        <h1 id="product-name">Collabora Online</h1>
-      </div>
-      <hr/>
-      <div id="about-dialog-container">
-        <div id="about-dialog-logos">
-          <fig id="product-logo"></fig>
-          <fig id="lokit-logo"></fig>
+    <dialog>
```
eszkadev, line 183: looks ok, but don't you need to use "open" function with it in the JS? (to actually use the feature fully)

gokaysatir: About dialog code copies the content of the dialog (not the dialog element). We don't need to implement that for now.

eszkadev (approved):
code looks ok

PR #10006, opened 2024-09-06
repair richdocumentscode case

caolanm:
Some handlers, like handleClientProxyRequest, used in the richdocumentscode  Nextcloud 'Built-in CODE Server' case, launch a task that will consume the message contents later, so we shouldn't consume the body in ClientRequestDispatcher::handleIncomingMessage.
    
While some like the file server, have already handled the message and we do need to discard the message contents (cool.html POST) because we expect multiple requests per socket and leaving the residue behind trashes that.

### DIFF
--- a/wsd/ClientRequestDispatcher.cpp
+++ b/wsd/ClientRequestDispatcher.cpp
@@ -667,6 +667,9 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
         return;
 
     LOG_DBG("Handling request: " << request.getURI());
+
+    bool served = false;
+
     try
     {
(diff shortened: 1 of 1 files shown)


sgothel (approved):
Great analysis of this side-effect of handlers.
This hopefully concludes the open connection effects.

Missed the issue in the 1st place (proxy/php/..) and elaborating whether
we properly handle the 'Connection' header properly in the proxy
according to RFC2616 8.1.3, 14.10 etc. 
Meaning, the proxy shall re-evaluate the header tokens according to 
Http 1.0/1.1 connections and only (re-)set them if appropriate.
(just brainstorming, haven't looked into our proxy code)

Ashod:
Good catch, @caolanm!

PR #10007, opened 2024-09-07
signal waiting - always wait for signals before _Exit.

mmeeks:
Centralize waiting in a single, normal exit location. This avoids us loosing and not being able to attach to and debug a crashed child thread when another thread decides to _Exit for some reason while the signal handler is running in the other thread.

Change-Id: I9a788960712506593c61b68305ed6cb90f948ead

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Helped me (at least) debug some unusual unit test crashers =)

caolanm (approved):

PR #10009, opened 2024-09-08
slideshow: fetch correct layer for AnimatedElement

eszkadev:
- fix bug: some animations were not added to the queue
- provide layer to the AnimetedElement

Core side patches:
- https://gerrit.libreoffice.org/c/core/+/173050    (initVisibility property for hiding objects which will appear later)

### DIFF
--- a/browser/src/slideshow/engine/SlideAnimations.ts
+++ b/browser/src/slideshow/engine/SlideAnimations.ts
@@ -164,7 +164,7 @@ class SlideAnimations {
 	private aEventMultiplexer: EventMultiplexer;
 	private bElementsParsed: boolean;
 
-	constructor(aSlideShowContext: SlideShowContext) {
+	constructor(aSlideShowContext: SlideShowContext, metaSlide: MetaSlide) {
 		this.aContext = new NodeContext(aSlideShowContext);
 		this.aSlideShowHandler = aSlideShowContext.aSlideShowHandler;
 		this.aAnimationNodeMap = new Map();
@@ -178,6 +178,7 @@ class SlideAnimations {
 		this.aRootNode = null;
 		this.bElementsParsed = false;
 
+		this.aContext.metaSlide = metaSlide;
 		this.aContext.aAnimationNodeMap = this.aAnimationNodeMap;
 		this.aContext.aAnimatedElementMap = this.aAnimatedElementMap;
 		this.aContext.aSourceEventElementMap = this.aSourceEventElementMap;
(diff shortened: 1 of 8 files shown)


browser/src/slideshow/engine/SlideAnimations.ts
```diff
@@ -40,14 +40,16 @@ function getNodeChildren(aNode: ContainerNodeInfo): Array<AnimationNodeInfo> {
 function createAnimationTree(
 	aAnimationRoot: AnimationNodeInfo,
 	aNodeContext: NodeContext,
+	metaSlide: MetaSlide,
 ): BaseNode {
-	return createAnimationNode(aAnimationRoot, null, aNodeContext);
+	return createAnimationNode(aAnimationRoot, null, aNodeContext, metaSlide);
```
mcecchetti, line 45: I think that would be cleaner, to pass the meta slide or just slide info to SlideAnimation ctor and append it to node context which is already passed down along tree nodes.
Anyway nice job.

browser/src/slideshow/engine/animationnodes/BaseNode.ts
```diff
@@ -425,7 +425,6 @@ abstract class BaseNode {
 
 	public init(): boolean {
 		this.DBG(this.callInfo('init'));
-		if (!this.checkValidNode()) return false;
```
mcecchetti, line 428: To be double check. It can be dangerous.

eszkadev: I added some comment in the commit: https://github.com/CollaboraOnline/online/pull/10009/commits/55ad9ae47ac84d43924b91bf2fa106031ebbde64

do you know maybe what else could be missing instead? it was not allowing to resolve root node in some cases and we were missing animations on click in the slideshow

browser/src/slideshow/engine/SlideAnimations.ts
eszkadev: right, done

browser/src/slideshow/engine/animationnodes/BaseNode.ts
eszkadev: the root problem might be caused by:
```
DBG: AnimationTransformNode( 6, Invalid ).init
global.js:545 createPropertyAnimation: attribute is unknown: transform
global.js:545 AnimationTransformNode.createActivity: failed to create animation.
```
so then we are in the invalid state...

eszkadev: added proper solution now

mcecchetti: Thanks, the problem was caused by capitalized transform name I guess.

mcecchetti (approved):
Now looks good to me.

PR #10011, opened 2024-09-09
Refactor Special Brokers Into New File

Ashod:
Mostly moved code to reduce the ever-growing DocumentBroker.
Also makes configuration values that are static conditional so we can change them in tests (otherwise, tests can't change these config values, conversely, we must extract them over and over at runtime, if they aren't static, which is very costly because Poco is inefficient and throws internal exceptions and whatnot).

- configure: directive for static configuration values
- wsd: move special brokers to a new home


caolanm (approved):
/Users/ci/jenkins/workspace/github_ios_app_co-24.04/wsd/DocumentBroker.cpp:364:5: error: unknown type name 'CONFIG_STATIC'
    CONFIG_STATIC const std::chrono::microseconds migrationMsgTimeout = std::chrono::seconds(

looks like a real build failure, but otherwise uncontroversial I imagine if that can be sorted

caolanm:
I see that there is a ios/config.h.in I wonder if that gets used on ios instead of the normal config.h.in and that's the IOS build error, no CONFIG_STATIC in ios/config.h[.in] ?

Ashod:
Thanks, @caolanm. I've done exactly the same, which is how I found out that there is a different config.h for iOS. But wasn't expecting that the variables weren't applied there in the same way.

Pretty unexpected. I have removed the two CONFIG_STATIC cases from DocumentBroker.cpp. Now that I have a passing build, I'll merge and apply your suggestion separately.

Thanks for your help!

PR #10012, opened 2024-09-09
C++ Misc Cleanup: More override, std-header, std::make_* usage, ..

sgothel:
### Summary

- Use override for virtual function overrides (incl. dtor)
- Use C++ std headers, e.g. cctype instead of ctype.h
- Use global anonym namespace for system functions or std:: system functions, whichever is applicable
- Use std::make_shared<StreamSocket>(..)
- Use std::make_unique<Watchdog>()
- Use 'const' where applicable
- Use pre-incr where applicable
- Use typename where applicable
- Drop redundant type conversions

Change-Id: Id844187c82fce3c6e9bceb512c07180705608ebc

* Prepares for: #9833
* Target version: master 

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -17,12 +17,12 @@
 
 #include <chrono>
 #include <cstring>
-#include <ctype.h>
+#include <cctype>
 #include <iomanip>
 #include <memory>
 #include <ratio>
 #include <sstream>
-#include <stdio.h>
+#include <cstdio>
 #include <string>
 #include <unistd.h>
 #include <sys/stat.h>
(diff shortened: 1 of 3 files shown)


net/Socket.cpp
```diff
@@ -1087,21 +1087,21 @@ std::shared_ptr<Socket> ServerSocket::accept()
             const void *inAddr;
             if (clientInfo.sin6_family == AF_INET)
             {
-                auto ipv4 = (struct sockaddr_in *)&clientInfo;
+                struct sockaddr_in *ipv4 = (struct sockaddr_in *)&clientInfo;
```
caolanm, line 1090: we could probably replace the c-style "struct sockaddrWhatever" with c++-style "sockaddrWhatever", but that's just an observation.

caolanm (approved):

PR #10016, opened 2024-09-09
fix(mobile): Unmangle unicode text messages

Minion3665:
In I66f8edb5f35e2300e3a0407a139e72039bec0d8d I introduced a regression where unicode messages were decoded as ascii, leading to, e.g., accented characters in jsdialogs being mangled.

I've written a function ("b64d") to decode these messages. I picked the name because it's the same length as the original "atob" (which is useful for buffer size compatibility) and stands for "base 64 decode".

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Minion3665:
There's a backport PR here: #10017, which is already positively reviewed

PR #10017, opened 2024-09-09
backport: fix(mobile): Unmangle unicode text messages

Minion3665:
This additionally backports #9986 from @plubius, as the unmangle change doesn't apply cleanly without it


plubius (approved):
Looks good to me. I built with CollaboraOnline master branch plus your "Unmangled unicode" patch (the other 2 patches are already in master). I ran with French (Canada) as my system locale and I can see the correct accents in the spellcheck dialog. If there are any other dialogs that I should test, let me know.

Capture d’écran, le 2024-09-09 à 13 50 40

PR #10019, opened 2024-09-09
kit: show busy popup for exported file until ui is unblocked

lpranam:
allow busy popups with specific ids to run until specific
busy popup is finished

problem:
busy popup used to disappear quickly, exported files may not have completed upload. User may think that exporting is finished and close the doc. Which caused failure in export.

Change-Id: I14bce386a200a4f2deae22b1fbefd597b9347268

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/map/Map.js
+++ b/browser/src/map/Map.js
@@ -1460,10 +1460,10 @@ L.Map = L.Evented.extend({
 			switch (e.statusType)
 			{
 			case 'start':
-				if (e.text) {
-					// e.text translated by Core
-					this.showBusy(e.text);
-				}
+				// e.text translated by Core
+				this.showBusy(e.text ? e.text : _('Please wait!'));
+				if (e.forceid)
+					this._progressBar.forceid = e.forceid;
 				break;
 			case 'setvalue':
 				this._progressBar.setBar(true);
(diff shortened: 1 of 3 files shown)


browser/src/map/Map.js
```diff
@@ -1462,7 +1462,9 @@ L.Map = L.Evented.extend({
 			case 'start':
 				if (e.text) {
 					// e.text translated by Core
-					this.showBusy(e.text);
+					this.showBusy(_(e.text));
```
caolanm, line 1465: Why did this go from e.text to _(e.text) ? Doesn't _() mark it for translation and the comment above suggests that this arrives already translated from code?

lpranam: I have added a new string in kit "Exporting". kit can not translate anything so it needs to be translated in js.

timar: You can write
if(e.text === 'Exporting') {
this.showBusy(_('Exporting'));
}
else {
this.showBusy(e.text);
}

It's lame, but this way we can handle translation.

lpranam: used more generic approach, makes code look better and also maybe helpful in future...

show "please wait" if no text is provided.

kit/ChildSession.hpp
```diff
@@ -100,7 +100,7 @@ class ChildSession final : public Session
         return _docManager->sendFrame(msg.data(), msg.size(), WSOpCode::Binary);
     }
 
-    bool sendProgressFrame(const char* id, const std::string &jsonProps);
+    bool sendProgressFrame(const char* id, const std::string &jsonProps, std::string forcedID = "");
```
caolanm, line 103: you can make this: const std::string& forcedID = "" and "const std::string& forcedID" above to match.

caolanm (approved):
ok, seems reasonable to me

PR #10021, opened 2024-09-10
slideshow: setup bounds for animated layers

eszkadev:


quikee (approved):

browser/src/slideshow/engine/AnimatedElement.ts
mcecchetti, line 1: Changes on AnimatedElement clashes with the patch I'm working on. I'll take care to use getLayerImage, getLayerBounds. Btw getLayerImage return a png image not ImageData even if there is no type error at assigning btw them.

browser/src/slideshow/engine/animationnodes/AnimationBaseNode.ts
mcecchetti, line 1: I've already changed to ctor signature to this in my patch :-)

browser/src/slideshow/engine/AnimatedElement.ts
mcecchetti: As agreed I'll take care to remove what is not needed.

mcecchetti (approved):

Issue #9787, opened 2024-08-12
copy/duplicate sheets


photoalex:
### Describe the Bug
I'm pretty sure I could this last time I used this app

### Steps to Reproduce
Tap and hold sheet tap > copy sheet

### Expected Behavior
Creates a copy of a sheet

### Actual Behavior
No such option but i'm sure I did it at some point. Am I missing something?

### Smartphone
 - android 13
- collabora 23.05.5.2
- 
### Additional Context
Maybe I was dreaming but such an action would seem quite trivial...

Minion3665:
Hello! Thanks for reporting :)

I'm not sure we have ever had this option in the mobile app, but we definitely have it in desktop and tablet views, and I agree that we should make it available for mobile. This sounds trivial and I was as surprised as you to realize that it wasn't there!

---

We have an upcoming Android release - 24.04.5.4 - which it's unfortunately a little too late to make this in time for, however I will aim to do this by the Android release after that.

I'll keep you posted in this issue, thanks very much for your patience!

Minion3665:
Hello! I've just fixed this now, so it should be available in the next Android release - we're expecting that to be around the start of October. I'll let you know when we have a snapshot with this commit in so you can test it out :)

photoalex:
> Hello! I've just fixed this now, so it should be available in the next Android release - we're expecting that to be in early October. I'll let you know when we have a snapshot with this commit in so you can test it out :)

Awesome! Thank you very many!

Minion3665:
Hello @photoalex! The latest snapshot, which you can get [from our website](https://www.collaboraoffice.com/downloads/Collabora-Office-Android-Snapshot/?C=M;O=D) here: (the exact correct link depends on your device, but it's probably [the arm64 build unless you know differently](https://www.collaboraoffice.com/downloads/Collabora-Office-Android-Snapshot/collabora-office-mobile-24.04-snapshot-arm64-v8a-2024-09-15.apk)), contains the option to copy sheets. Please feel free to test it out.

Thanks!

PR #7301, opened 2023-09-25
calc: Show tabs context menu on a tablet's press

Minion3665:
Change-Id: Ibf25b0d76d8487e582ed6a6cf1f82467869453f6

* Resolves: #2567
* Target version: master 

### Checklist

- [ ] Code is properly formatted
- [x] All commits have Change-Id
- [ ] I have run tests with `make check`
- [x] I have issued `make run` and manually verified that everything looks okay
- [x] Documentation (manuals or wiki) has been updated or is not required

PR #10023, opened 2024-09-10
feat(mobile): Show move/copy in calc

Minion3665:
Previously the move/copy button wasn't a jsdialog, so it couldn't be a mobile wizard. After core change I23cef6330a5d5f64ac8b1746ef22516dc76f9d8e, we made it a jsdialog, so we can show it on mobile. Let's do that!

Requires: https://gerrit.libreoffice.org/c/core/+/172975
Fixes: #9787

Change-Id: Idc172ed37ca346d809c3181fdfefbbb5dc0d400b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Tabs.js
+++ b/browser/src/control/Control.Tabs.js
@@ -95,32 +95,27 @@ L.Control.Tabs = L.Control.extend({
 				callback: (this._moveSheetRight).bind(this),
 				visible: areTabsMultiple
 			},
-		};
-
-		if (!window.mode.isMobile()) {
-
-			// no blacklisting available for this context menu so only add when needed
-			this._menuItem['.uno:Move'] = {
+			'.uno:Move': {
 				name: _UNO('.uno:Move', 'spreadsheet', true),
 				callback: (this._moveOrCopySheet).bind(this),
 				visible: areTabsMultiple
-			};
-
-			this._menuItem['.uno:CopyTab'] = {
+			},
+			'.uno:CopyTab': {
 				name: _('Copy Sheet...'),
 				callback: function() {this._map.sendUnoCommand('.uno:Move');}.bind(this),
 				visible: function() {
 					return !areTabsMultiple();
 				}
-			};
-			    if (!this._map.isReadOnlyMode() || window.mode.isTablet()) {
-				L.installContextMenu({
-					selector: '.spreadsheet-tab',
-					className: 'cool-font',
-					items: this._menuItem,
-					zIndex: 1000
-				});
-			}
+			},
+		};
+
+		if ((!window.mode.isMobile() && !this._map.isReadOnlyMode()) || window.mode.isTablet()) {
+			L.installContextMenu({
+				selector: '.spreadsheet-tab',
+				className: 'cool-font',
+				items: this._menuItem,
+				zIndex: 1000
+			});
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.Tabs.js
```diff
@@ -95,32 +95,27 @@ L.Control.Tabs = L.Control.extend({
...
-					selector: '.spreadsheet-tab',
-					className: 'cool-font',
-					items: this._menuItem,
-					zIndex: 1000
-				});
-			}
+			},
+		};
+
+		if ((!window.mode.isMobile() && !this._map.isReadOnlyMode()) || window.mode.isTablet()) {
```
eszkadev, line 112: on tablet do we want to add the menu in readonly?

browser/src/control/Control.Tabs.js
```diff
@@ -180,6 +175,13 @@ L.Control.Tabs = L.Control.extend({
 							'Hide': this._menuItem['.uno:Hide'],
 							'movesheetleft': this._menuItem['movesheetleft'],
 							'movesheetright': this._menuItem['movesheetright'],
+							'Move': this._menuItem['.uno:Move'],
```
eszkadev, line 178: no copy in multi tab?

browser/src/control/Control.Tabs.js
Minion3665: It seemed a little weird to me too, but I applied the principle of Chesterton's fence: that is, don't break it unless you understand it.

The condition before was
```js
if (!window.mode.isMobile()) {
	if (!this._map.isReadOnlyMode() || window.mode.isTablet()) {
```

and the condition after was
```
if ((!window.mode.isMobile() && !this._map.isReadOnlyMode()) || window.mode.isTablet()) {
```

given that $window.mode.isTablet() \implies !window.mode.isMobile()$, these are the same condition

browser/src/control/Control.Tabs.js
Minion3665: > no copy in multi tab?

There's only 1 dialog - the "Move or copy" dialog. When there's only a single tab, we title it "Copy" and disable the move button ... because there's nowhere to move it. You can see some more about these options ~line 100

browser/src/control/Control.Tabs.js
Minion3665: I can take a deeper look into why it was this way if you'd like

caolanm: Amusingly enough git blame suggests this came from https://github.com/CollaboraOnline/online/pull/7301 :-)

"Show tabs context menu on a tablet's press" because "tablets no longer had the context menu registered as they always start in read only mode"

So I guess its fine to retain that fix

caolanm (approved):

browser/src/control/Control.Tabs.js
Minion3665: it feels to me like there should be a better way to do that that registers/unregisters the context menu at the right time ... but I guess it makes sense, desktop doesn't switch from read-only to not whereas tablet does

Minion3665: well, certainly we can keep it for now :)

PR #10026, opened 2024-09-10
Fixes invisible typing issue.

gokaysatir:
_clientSelectedMode is set after the first tiles invalidation.
So this commit sets it beforehand.

See Core side ViewShellBase.cxx -> getEditMode for more info.

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -3213,7 +3213,8 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
             else if (tokens.size() == 3 && tokens.equals(0, "EMPTY"))
             {
                 // with mode:    "EMPTY, <part>, <mode>"
-                const std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
+                // Writer renders everything as part 0.
+                const std::string part = (_docType != "text" ? (_docType == "presentation" ? std::to_string(getLOKitDocument()->getPart()) : tokens[1].c_str()) : "0");
                 const std::string mode = (_docType != "text" ? tokens[2].c_str() : "0"); // Writer is not using mode.
                 sendTextFrame("invalidatetiles: EMPTY, " + part + ", " + mode +
                               " wid=" + std::to_string(getCurrentWireId()));
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -84,7 +84,7 @@ ClientSession::ClientSession(
     _splitX(0),
     _splitY(0),
     _clientSelectedPart(-1),
-    _clientSelectedMode(0),
+    _clientSelectedMode(ViewMode::NORMAL_VIEW),
     _tileWidthPixel(0),
     _tileHeightPixel(0),
     _tileWidthTwips(0),
@@ -1221,9 +1221,9 @@ bool ClientSession::_handleInput(const char *buffer, int length)
             We set the view mode beforehand here.
         */
         if (firstLine == "uno .uno:NormalMultiPaneGUI")
-            _clientSelectedMode = 0;
+            _clientSelectedMode = ViewMode::NORMAL_VIEW;
         else if (firstLine == "uno .uno:NotesMode")
-            _clientSelectedMode = 2;
+            _clientSelectedMode = ViewMode::NOTES_VIEW;
 
         if (!filterMessage(firstLine))
         {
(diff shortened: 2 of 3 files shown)


wsd/ClientSession.cpp
```diff
@@ -2856,7 +2862,7 @@ void ClientSession::handleTileInvalidation(const std::string& message,
     int normalizedViewId = getCanonicalViewId();
 
     std::vector<TileDesc> invalidTiles;
-    if((part == _clientSelectedPart && mode == _clientSelectedMode) || _isTextDocument)
+    if(part == _clientSelectedPart || _isTextDocument)
```
eszkadev, line 2865: are we sure? it was introduced 2 years ago in commit 7043365 together with "mode", some time before our new approach was implemented (wids and dark mode)

it was working fine and condition seems to be correct. I think we did some change recently, would be good to bisect and find real cause

eszkadev: btw we should get rid of `mode` and use the new thing probably in the future :)

wsd/ClientSession.cpp
```diff
@@ -1215,6 +1215,16 @@ bool ClientSession::_handleInput(const char *buffer, int length)
...
 
+        /*
+            When the UI mode is switched, tile invalidation callback is sent before the mode switch callback.
+            This causes the tiles' ui mode to conflict with the view mode.
+            We set the view mode beforehand here.
+        */
+        if (firstLine == "uno .uno:NormalMultiPaneGUI")
+            _clientSelectedMode = 0;
+        else if (firstLine == "uno .uno:NotesMode")
+            _clientSelectedMode = 2;
```
eszkadev, line 1226: we need enum for that magic number :)
1 was master page mode also IIRC

eszkadev: don't we even have enum already?

eszkadev:
we found similar issue with master page as notes, there is some core patch which causes that.
would be good to check if core fix isn't enough as it worked before probably?
But still maybe some parts of this also could be worth merging together.

wsd/ClientSession.cpp
gokaysatir: There wasn't an enumerator for the current values. I added.

kit/ChildSession.cpp
```diff
@@ -3213,7 +3213,8 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
             else if (tokens.size() == 3 && tokens.equals(0, "EMPTY"))
             {
                 // with mode:    "EMPTY, <part>, <mode>"
-                const std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
+                // Writer renders everything as part 0.
+                const std::string part = (_docType != "text" ? (_docType == "presentation" ? std::to_string(getLOKitDocument()->getPart()) : tokens[1].c_str()) : "0");
```
eszkadev, line 3217: hmm, this looks like we want to be smarter than core :)
I think the rule is "core is the only source of truth".
Why the core generates callback with wrong part number?
The proper fix shouldn't be where callback is emited?

gokaysatir: I checked the core side but it's a rabbit hole. This is a workaround until we have more time to check why the pViewShell pointer is null after a mode switch.

gokaysatir: That function still gets the information from core side. But there's a timing issue.

eszkadev: I see..., one more thing: could you check if using this patch can we fix also master page switching?

wsd/ClientSession.cpp
```diff
@@ -1215,6 +1215,16 @@ bool ClientSession::_handleInput(const char *buffer, int length)
...
         }
 
+        /*
+            When the UI mode is switched, tile invalidation callback is sent before the mode switch callback.
+            This causes the tiles' ui mode to conflict with the view mode.
+            We set the view mode beforehand here.
+        */
+        if (firstLine == "uno .uno:NormalMultiPaneGUI")
+            _clientSelectedMode = ViewMode::NORMAL_VIEW;
+        else if (firstLine == "uno .uno:NotesMode")
```
eszkadev, line 1225: do we need maybe, master page mode here?

kit/ChildSession.cpp
gokaysatir: No, unfortunately it doesn't. I checked it.

wsd/ClientSession.cpp
gokaysatir: Ah, that may work. But there may be also other issues with master switch. I think we can check this after this one.

eszkadev (approved):

kit/ChildSession.cpp
eszkadev: @gokaysatir this caused regression in slide previews updates I think, please see:
invalidates

NOW (top) vs OLD (bottom) on slide insertion

Issue #8566, opened 2024-03-18
Improve split button styling


eszkadev:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @pedropintosilva @eszkadev 

### Detailed description and rationale
Split buttons are buttons which have 2 parts: "`<direct command button>` | `<dropdown arrow>`"
Left one applies/runs some command but right part opens dropdown with more options.
It should be marked in the UI as 2 separate parts somehow. This is used for color pickers in notebookbar for example.
Notice that there are MenuButtons which doesn't have different actions for arrow and icon - Paste button in notebookbar.

See how it looks now:
[split-button-2024-03-18_18.22.50.webm](https://github.com/CollaboraOnline/online/assets/5307369/3d9402dd-176b-48c1-8ead-e596df191e00)

Currently on hover we highlight full button. It is not possible to differentiate between split and regular menu button.

### Code pointers
Split button is a kind of MenuButton: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Widget.MenuButton.js#L118

printfdebugging:
I would like to work on this issue.  Don't know much web-programming, but I hope I can come up with something working coming saturday-sunday.

Darshan-upadhyay1110:
Hii @eszkadev after merging solution for this issue can you please add "Hacktoberfest-accepted" label here?

Saurabhyadav0:
@Darshan-upadhyay1110  hello i want to contribute to this repo how can I do that

pedropintosilva:
This was fixed with https://github.com/CollaboraOnline/online/pull/10027 by @maks-w

Darshan-upadhyay1110:
> @Darshan-upadhyay1110 Hi! I'd like to contribute to this repository. How can I get started?

Hi @Saurabhyadav0! Welcome! There are plenty of interesting issues waiting to be solved. You can find them here: [Open Issues](https://github.com/CollaboraOnline/online/issues?q=is%3Aissue+is%3Aopen+label%3AHacktoberfest). 

Feel free to pick any issue, and don't hesitate to reach out if you need any help. Also, it's Hacktoberfest season! If you're interested, you can register and join in here: [Hacktoberfest](https://hacktoberfest.com/). Check out more details on how to participate: [Collabora Online Hacktoberfest](https://collaboraonline.github.io/hacktoberfest/).

Saurabhyadav0:
[...]

PR #10027, opened 2024-09-10
feat(UI): Improve split button styling

maks-w:
Change-Id: I4f0474e02411ada5547c8c2f23dda22cb4433557

* Resolves: #8566
* Target version: master 

### Summary
This is my proposal for UI splitting the buttons in 2 parts.
I hope I didn't take too much liberty with the design ^^

### Checklist

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -10,14 +10,10 @@
 }
 
 /* toolbuttons */
-
-.unotoolbutton .unobutton img {
-	width: var(--btn-size);
-	height: var(--btn-size);
-}
 .unotoolbutton .unobutton.selected + .ui-content.unolabel {
 	color: var(--color-text-dark);
 }
+
 /* toolbuttons non selected */
 #clearFormatting,
 .hasnotebookbar .ui-content.unotoolbutton.has-label,
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -25,17 +25,32 @@
 .hasnotebookbar .ui-content.unotoolbutton:not(.has-label):not(.inline),
 .sidebar.unotoolbutton,
 .jsdialog.unotoolbutton,
+.notebookbar.unotoolbutton,
 #closebutton {
 	background-color: transparent;
 	border: 1px solid transparent;
 	color: var(--color-main-text);
 	border-radius: var(--border-radius);
 }
+
+/* toolbuttons non selected hover */
+.hasnotebookbar .ui-content.unotoolbutton.has-label:hover,
+.hasnotebookbar .ui-content.unotoolbutton.inline:hover,
+.hasnotebookbar .ui-content.unotoolbutton:not(.has-label):not(.inline):hover,
+.sidebar.unotoolbutton:hover,
+.jsdialog.unotoolbutton:hover,
+.notebookbar.unotoolbutton:hover {
+	border: 1px solid var(--color-border-darker);
+	color: var(--color-text-darker);
+	border-radius: var(--border-radius);
+}
+
 /* toolbuttons selected */
 .hasnotebookbar .ui-content.unotoolbutton.selected.has-label,
 .hasnotebookbar .ui-content.unotoolbutton.selected.inline,
 .hasnotebookbar .ui-content.unotoolbutton.selected:not(.has-label):not(.inline),
 .sidebar.unotoolbutton.selected,
+.notebookbar.unotoolbutton.selected,
 .jsdialog.unotoolbutton.selected {
 	background-color: var(--color-background-dark);
 	border: 1px solid var(--color-border-dark);
@@ -50,14 +65,16 @@
 	background-color: var(--color-background-darker);
 	border-color: var(--color-border-darker);
 }
+
 /* toolbuttons selected hover */
+.hasnotebookbar .ui-content.unotoolbutton.has-label:hover,
 .hasnotebookbar .ui-content.unotoolbutton.selected:hover,
 .unotoolbutton.notebookbar:hover,
 .unotoolbutton.jsdialog:hover,
 .hasnotebookbar .ui-content.unotoolbutton.selected:not(.has-label):not(.inline):hover,
 .sidebar.unotoolbutton.selected:hover,
+.notebookbar.unotoolbutton.selected:hover,
 .jsdialog.unotoolbutton.selected:hover {
-	background-color: var(--color-background-darker) !important;
 	border: 1px solid var(--color-border-darker);
 	color: var(--color-text-darker);
 	border-radius: var(--border-radius);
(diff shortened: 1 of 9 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2761,7 +2761,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			if (typeof menubutton === 'object') {
 				L.DomUtil.addClass(menubutton.container, data.class ? data.class + ' has-colorpicker': 'has-colorpicker');
 
-				var valueNode = L.DomUtil.create('div', 'selected-color', menubutton.container);
+				var buttonElem = menubutton.container.querySelector('button');
```
eszkadev, line 2764: we can avoid query here, we should take reference from the menubutton object, probably it will be `menubutton.button`. if not please look into `Widget.MenuButton.ts`

browser/css/btns.css
```diff
@@ -50,14 +65,16 @@
...
 /* toolbuttons selected hover */
+.hasnotebookbar .ui-content.unotoolbutton.has-label:hover,
 .hasnotebookbar .ui-content.unotoolbutton.selected:hover,
 .unotoolbutton.notebookbar:hover,
 .unotoolbutton.jsdialog:hover,
 .hasnotebookbar .ui-content.unotoolbutton.selected:not(.has-label):not(.inline):hover,
 .sidebar.unotoolbutton.selected:hover,
+.notebookbar.unotoolbutton.selected:hover,
 .jsdialog.unotoolbutton.selected:hover {
-	background-color: var(--color-background-darker) !important;
```
maks-w, line 60: We move the responsability of the `background-color` to the children since a button could have 2 parts

browser/css/jsdialogs.css
```diff
@@ -1758,17 +1755,30 @@ input[type='number']:hover::-webkit-outer-spin-button {
...
+	flex-direction: column;
+	padding: 3px;
 	vertical-align: middle;
 	background-color: transparent;
 	border: none;
+	position: relative;
+}
+
+.unoarrow::after,
+.jsdialog .ui-content.unobutton::after {
```
maks-w, line 1769: I extended a little bit the click zone so that the border of the button could be clickable

cypress_test/integration_tests/desktop/impress/delete_objects_spec.js
```diff
@@ -37,6 +37,7 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Delete Objects', function(
 	});
 
 	it('Delete Chart' , function() {
+		cy.cGet('#toolbar-up > .ui-scroll-right').click();
 		cy.cGet('#toolbar-up > .ui-scroll-right').click();
```
maks-w, line 41: The menu is a bit wider now, so we need to click twice for the icon to appear.

browser/src/control/Control.JSDialogBuilder.js
maks-w: done

pedropintosilva (commented):
thanks @maks-w ! It looks much better, well done :clap: 

Before we can merge this it would be good to understand why we are forgoing the 	`--btn-size` CSS var. That was a nice way to not only make sure all icon buttons share the same size but also to allow adjusting them in one single place. With the current changes we are never sure about the resulting size. 

Also the pseudo elements in other buttons are being affected (not only arrows). Example the save icon in the tabbed bar (when it receives the `.savemodified` class. Meaning when the user types and there are changes ready to be saved)

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2760,8 +2760,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 			if (typeof menubutton === 'object') {
 				L.DomUtil.addClass(menubutton.container, data.class ? data.class + ' has-colorpicker': 'has-colorpicker');
-
-				var valueNode = L.DomUtil.create('div', 'selected-color', menubutton.container);
+
```
eszkadev, line 2763: can we remove this white space?

maks-w:
> thanks @maks-w ! It looks much better, well done 👏
> 
> Before we can merge this it would be good to understand why we are forgoing the `--btn-size` CSS var. That was a nice way to not only make sure all icon buttons share the same size but also to allow adjusting them in one single place. With the current changes we are never sure about the resulting size.

I pushed a new commit to restore this behavior. Some changes however, because of the padding, the button and the image are not the same size now, so there is a button size and an image size variable (the image size is computed, so there is still only 1 variable to change to change the size of the button).

I took the liberty to reduce a little bit the size of the buttons, who seem a little too big (but perhaps I took it a step too far). Tell me what you think.

> 
> Also the pseudo elements in other buttons are being affected (not only arrows). Example the save icon in the tabbed bar (when it receives the `.savemodified` class. Meaning when the user types and there are changes ready to be saved)

For the pseudo element, it seems to me that it is a good practice to extend a little bit the click zone of buttons in general; not just for the arrow button. 
Even today we have the rollover actif before we can click on it, which can cause missclicks. 
But maybe I didn't understand correctly the issue?

browser/src/control/Control.JSDialogBuilder.js
maks-w: done

eszkadev:
please rebase, I think tests are failing due to old base

eszkadev:
error in the CI I see is:
`      cy:command ✘  assert	expected **<div#StatePageNumber.jsdialog.ui-badge>** to have text **Page 2 of 4**, but the text was **Pages 2 and 3 of 4**`
this happens when we changed size of the toolbars most often. Are you sure sizing (height) of the split button didn't change?

maks-w:
Should be ok now. I restored the icons to the initial size and adjusted some CSS for the tests to pass.

eszkadev:
solved merge conflict

eszkadev:
looks nice, but requires some more pixel perfection, eg. here I see border on the color bar:
crop

maks-w:
I must be missing something because now on (master) I don't have the selected-color element or the arrow???

image

Am I doing something wrong?

eszkadev:
> I must be missing something because now on (master) I don't have the selected-color element or the arrow???
> 
> image
> 
> Am I doing something wrong?
> 
> I have weird class name like : `unospan-home-back-color:ColorPickerMenu` and i don't have the `has-dropdown` class in the dom

please do  `make clean` and rebuild the app completely, color picker button was moved to separate file and your build seems to not have that built

eszkadev (approved):
beautiful: 
icon

Darshan-upadhyay1110:
hello @maks-w Can you please comment in issue link here i want to assign this task under your username :)
Also if you haven't registered yet for Hacktoberfest please https://hacktoberfest.com/ as you have already solved this issue which has this tag...

PR #10032, opened 2024-09-11
Test Improvements

Ashod:
Adds `loaded:` and `unloaded:` commands to signal the loading/unloading of views (and documents). This avoid the overloading of `status:`, which is issued for a number of other LOK events, and indeed can be issued on demand (via a command), as well.

This fixes a test-case failure due to document size change, which issued `status:` and that triggered the `onDocumentLoaded` callback on the test, which was unexpected and failed the test. This has happened at least 2 times in the past ~week or so.

- wsd: new loaded: and unloaded: commands
- wsd: test: use loaded: to trigger load events

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -154,6 +154,12 @@ void ChildSession::disconnect()
             if (_docManager)
             {
                 _docManager->onUnload(*this);
+
+                // Notify that we've unloaded this view.
+                std::ostringstream oss;
+                oss << "unloaded: viewid=" << _viewId
+                    << " views=" << getLOKitDocument()->getViewsCount();
+                sendTextFrame(oss.str());
             }
         }
         else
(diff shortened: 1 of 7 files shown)


kit/ChildSession.cpp
```diff
@@ -903,13 +909,17 @@ bool ChildSession::loadDocument(const StringVector& tokens)
 
     SigUtil::addActivity(getId(), "load doc: " + getJailedFilePathAnonym());
 
+    // Note: _isDocLoaded is set on our retunt.
```
caolanm, line 912: return?

caolanm (approved):
I like the sound of a test getting more reliable.

kit/ChildSession.cpp
Ashod: Hah, good eye-sight.

mmeeks (approved):
Looks great to me =)

PR #10034, opened 2024-09-11
UI: indicate edit permission in status bar

lpranam:
update edit permission indicator based on WOPI parameters received

Change-Id: I83a743f8aff646e357ce1a0e73abf9d3db99280e

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.StatusBar.js
+++ b/browser/src/control/Control.StatusBar.js
@@ -12,7 +12,7 @@
  * JSDialog.StatusBar - statusbar component
  */
 
-/* global $ app JSDialog _ _UNO */
+/* global $ app JSDialog _ _UNO  getPermissionModeElements */
 class StatusBar extends JSDialog.Toolbar {
 	constructor(map) {
 		super(map, 'toolbar-down');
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.StatusBar.js
```diff
@@ -377,12 +377,19 @@ class StatusBar extends JSDialog.Toolbar {
...
+			permissionContainer.appendChild(getPermissionModeElements(isReadOnlyMode, this.map['wopi'].UserCanWrite));
+		}
+
 		this.builder.updateWidget(this.parentContainer, {
 			id: 'PermissionMode',
 			type: 'htmlcontent',
 			htmlId: 'permissionmode',
 			isReadOnlyMode: isReadOnlyMode,
-			canUserWrite: !app.isReadOnly()
+			canUserWrite: this.map['wopi'].UserCanWrite
```
caolanm, line 392: I wonder if this is ok to do for the android/iOS case? I presume there that this.map['wopi'] would be null, maybe @Minion3665  knows. Would is be safer to use something like:

 this.map['wopi'] ? this.map['wopi'].UserCanWrite : !app.isReadOnly()

Minion3665: A bit worse than that I'm afraid - in the mobile case `this.map['wopi']` *isn't* null but `L.Map.THIS.wopi.UserCanWrite` is always false.

here's some stuff from my chrome debug console...
```
$ L.Map.THIS["wopi"]
NewClass {_map: NewClass, _initHooksCalled: true}

$ L.Map.THIS.wopi.UserCanWrite
false
```

Minion3665 (commented):
Thanks for correcting the mobile issues - this is all good from my end now :)

caolanm (approved):
ok, lets merge then

PR #10035, opened 2024-09-11
slideshow: handle hidden slides

eszkadev:
it makes us aware of hidden slides (previously we didn't get them in the list - so requires core change)
handle both direction of skiping the slide when displaySlide fails

REQUIRES: https://gerrit.libreoffice.org/c/core/+/173218


quikee (approved):
Code looks good to me

PR #10037, opened 2024-09-12
introduce autofill preview tooltip

bayramcicek:
Change-Id: Ice62a3eb891a1653b5ec6ad5b10efe2fac0e914f

* Resolves: # 
* Target version: master 

### Summary

- show cell value previews in a popup while using autofill feature.
- this will allow users to see what value will be filled in the current cell in advance.
- add AutoFillPreviewTooltip class
- requires: https://gerrit.libreoffice.org/c/core/+/172589 (merged - 24.04 - master)
- requires: https://gerrit.libreoffice.org/c/core/+/174502 (merged - 24.04)

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jquery-ui-lightness.css
+++ b/browser/css/jquery-ui-lightness.css
@@ -522,7 +522,7 @@ button.ui-button::-moz-focus-inner {
 .ui-dialog .ui-dialog-content {
 	position: relative;
 	border: 0;
-	padding: 0em 0.2em;
+	padding: .5em 1em;
 	background: none;
 	overflow: auto;
 }
--- a/browser/css/jquery-ui-lightness.css
+++ b/browser/css/jquery-ui-lightness.css
@@ -522,7 +522,7 @@ button.ui-button::-moz-focus-inner {
 .ui-dialog .ui-dialog-content {
 	position: relative;
 	border: 0;
-	padding: .5em 1em;
+	padding: 0em 0.2em;
 	background: none;
 	overflow: auto;
 }
--- a/browser/src/control/Control.AutoFillPreviewTooltip.ts
+++ b/browser/src/control/Control.AutoFillPreviewTooltip.ts
@@ -74,7 +74,6 @@ class AutoFillPreviewTooltip extends L.Control.AutoCompletePopup {
 	}
 
 	openAutoFillPopup(ev: FireEvent): void {
-		const framePos = this.getAutoFillAreaBottomRightLocation(ev); // bottom-right position
 		const entry = ev.data.text;
 		let data: PopupData;
 
@@ -83,20 +82,23 @@ class AutoFillPreviewTooltip extends L.Control.AutoCompletePopup {
 
 			const control = this.getSimpleTextJSON(entry);
 			if (L.DomUtil.get(this.popupId + 'fixedtext')) {
-				data = this.getPopupJSON(control, framePos);
+				data = this.getPopupJSON(control, {
+					x: ev.data.location.cX,
+					y: ev.data.location.cY,
+				});
 				this.sendUpdate(data);
 				return;
 			}
 
 			if (L.DomUtil.get(this.popupId))
 				this.closeMentionPopup({ typingMention: true } as CloseMessageEvent);
-			data = this.newPopupData;
+			data = Object.assign({}, this.newPopupData);
 			data.children[0].children[0] = control;
 		}
 
 		// add position
-		data.posx = framePos.x;
-		data.posy = framePos.y;
+		data.posx = ev.data.location.cX;
+		data.posy = ev.data.location.cY;
 		this.sendJSON(data);
 	}
 }
--- a/browser/src/control/Control.AutoFillPreviewTooltip.ts
+++ b/browser/src/control/Control.AutoFillPreviewTooltip.ts
@@ -1,0 +1,105 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * Control.AutoFillPreviewTooltip - class for tooltip of cell previews during auto fill
+ */
+
+/* global app */
+
+class AutoFillPreviewTooltip extends L.Control.AutoCompletePopup {
+	usageText: string;
+	newPopupData: PopupData;
+
+	constructor(map: ReturnType<typeof L.map>) {
+		super('autoFillPreviewTooltip', map);
+		this.newPopupData = {
+			children: [
+				{
+					id: 'container',
+					type: 'container',
+					enabled: false,
+					children: new Array<WidgetJSON>(),
+					vertical: false,
+				} as any as WidgetJSON,
+			] as Array<WidgetJSON>,
+			jsontype: 'dialog',
+			type: 'modalpopup',
+			cancellable: false,
+			popupParent: '_POPOVER_',
+			clickToClose: '_POPOVER_',
+			id: 'autoFillPreviewTooltip',
+			canHaveFocus: false,
+			noOverlay: true,
+			title: '',
+		} as PopupData;
+	}
+
+	onAdd() {
+		this.newPopupData.isAutoCompletePopup = true;
+		this.map.on('openautofillpopup', this.openAutoFillPopup, this);
+		this.map.on('sendautofilllocation', this.sendAutoFillLocation, this);
+	}
+
+	getAutoFillAreaBottomRightLocation(ev: FireEvent): Point {
+		var currPos = {
+			x: app.calc.cellCursorRectangle.pX2 + ev.data.location.x,
+			y: app.calc.cellCursorRectangle.pY2 + ev.data.location.y,
+		};
+
+		// console.error("--core   : " + JSON.stringify(ev.data.location));
+		// console.error("--currPos: " + JSON.stringify(currPos));
+
+		return new L.Point(currPos.x, currPos.y);
+	}
+
+	sendAutoFillLocation(ev: FireEvent) {
+		this.openAutoFillPopup({ data: ev.data });
+	}
+
+	getSimpleTextJSON(cellValue: string): TextWidget {
+		return {
+			id: this.popupId + 'fixedtext',
+			type: 'fixedtext',
+			text: cellValue,
+			enabled: false,
+		} as TextWidget;
+	}
+
+	openAutoFillPopup(ev: FireEvent): void {
+		const framePos = this.getAutoFillAreaBottomRightLocation(ev); // bottom-right position
+		const entry = ev.data.text;
+		let data: PopupData;
+
+		if (entry.length > 0) {
+			this.closeMentionPopup({ typingMention: true } as CloseMessageEvent);
+
+			const control = this.getSimpleTextJSON(entry);
+			if (L.DomUtil.get(this.popupId + 'fixedtext')) {
+				data = this.getPopupJSON(control, framePos);
+				this.sendUpdate(data);
+				return;
+			}
+
+			if (L.DomUtil.get(this.popupId))
+				this.closeMentionPopup({ typingMention: true } as CloseMessageEvent);
+			data = this.newPopupData;
+			data.children[0].children[0] = control;
+		}
+
+		// add position
+		data.posx = framePos.x;
+		data.posy = framePos.y;
+		this.sendJSON(data);
+	}
+}
+L.control.autofillpreviewtooltip = function (map: any) {
+	return new AutoFillPreviewTooltip(map);
+};
--- a/browser/src/control/Control.AutoFillPreviewTooltip.ts
+++ b/browser/src/control/Control.AutoFillPreviewTooltip.ts
@@ -20,31 +20,14 @@ class AutoFillPreviewTooltip extends L.Control.AutoCompletePopup {
 
 	constructor(map: ReturnType<typeof L.map>) {
 		super('autoFillPreviewTooltip', map);
-		this.newPopupData = {
-			children: [
-				{
-					id: 'container',
-					type: 'container',
-					enabled: false,
-					children: new Array<WidgetJSON>(),
-					vertical: false,
-				} as any as WidgetJSON,
-			] as Array<WidgetJSON>,
-			jsontype: 'dialog',
-			type: 'modalpopup',
-			cancellable: false,
-			popupParent: '_POPOVER_',
-			clickToClose: '_POPOVER_',
-			id: 'autoFillPreviewTooltip',
-			canHaveFocus: false,
-			noOverlay: true,
-			title: '',
-			isAutoFillPreviewTooltip: true,
-		} as PopupData;
 	}
 
 	onAdd() {
-		this.newPopupData.isAutoCompletePopup = true;
+		this.newPopupData.isAutoFillPreviewTooltip = true;
+		this.newPopupData.canHaveFocus = false;
+		this.newPopupData.noOverlay = true;
+		this.newPopupData.id = 'autoFillPreviewTooltip';
+
 		this.map.on(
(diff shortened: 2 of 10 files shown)


bayramcicek:
- currently, popup is blocking the auto-fill

https://github.com/user-attachments/assets/b13b651e-c4b2-49db-8706-892a8ed3e36b

browser/src/control/Control.JSDialog.js
```diff
@@ -635,7 +635,7 @@ L.Control.JSDialog = L.Control.extend({
 		instance.startX = e.data.posx;
 		instance.startY = e.data.posy;
 		instance.updatePos = null;
-		instance.canHaveFocus = !instance.isSnackbar && instance.id !== 'busypopup' && !instance.isAutoCompletePopup;
+		instance.canHaveFocus = !instance.isSnackbar && instance.id !== 'busypopup' && !instance.isAutoCompletePopup && instance.id !== 'autoFillPreviewTooltip';
```
eszkadev, line 638: we want to avoid hardcoding the ids to the engine, instead please put `canHaveFocus` into the JSON :)

browser/src/control/Control.AutoFillPreviewTooltip.ts
```diff
@@ -0,0 +1,129 @@
...
+		// console.error("data: " + ev.data.location);
+		this.openAutoFillPopup({ data: ev.data });
+	}
+
+	getPopupEntries(ev: FireEvent): any[] {
+		const entries: any[] = [{ row: '0', columns: [{ text: ev.data.text }] }];
+		return entries;
+	}
+
+	openAutoFillPopup(ev: FireEvent): void {
```
eszkadev, line 83: isn't that something we can get from the base class AutoCompletePopup? or do we have some changes here?

browser/src/control/Control.AutoFillPreviewTooltip.ts
```diff
@@ -0,0 +1,129 @@
...
+			(data.children[0].children[0] as TreeWidget).entries = entries;
+		}
+
+		// add position
+		data.posx = framePos.x;
+		data.posy = framePos.y;
+		this.sendJSON(data);
+	}
+
+	closeMentionPopup(ev: CloseMessageEvent) {
```
eszkadev, line 104: do we need it? isn't that in the base class?

browser/src/control/Control.AutoFillPreviewTooltip.ts
```diff
@@ -0,0 +1,129 @@
...
+	// private onMouseUp (e: MouseEvent) {
+	//     console.error("yaz: " + e);
+	// }
+
+	getAutoFillAreaBottomRightLocation(ev: FireEvent): Point {
+		// var currPos = ev.data.location;
+		var currPos = {
+			// x: app.file.textCursor.rectangle.cX1,
+			x: app.file.viewedRectangle.cX1,
+			y: app.file.textCursor.rectangle.cY2,
```
eszkadev, line 61: textCursor -> where "caret" is while editing text inside cell
cellCursor -> position of selected cell

please look into `docstate.js`

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: yes, we use `getSimpleTextJSON(entry);` here. So we need to modify `openAutoFillPopup` function here.

browser/src/control/Control.JSDialog.js
bayramcicek: added `canHaveFocus: false,` but it seems we don't need it anymore. 
`instance.noOverlay` seems enough for unblocking the autofill.

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: Done. removed.
Thanks for reviews.

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: I use:
```ts

var currPos = {
	x:
		app.calc.cellCursorRectangle.pWidth +
		app.calc.cellCursorRectangle.pX1 +
		5,
	y:
		app.calc.cellCursorRectangle.pHeight +
		app.calc.cellCursorRectangle.pY1 +
		5,
};

```

### currently we have: 
https://github.com/user-attachments/assets/35591294-fa46-4348-9200-74553ef59100

### issues
- Popup width is too long
- popup location should be updated per autofill selection

browser/src/control/Control.JSDialog.js
```diff
@@ -635,6 +639,7 @@ L.Control.JSDialog = L.Control.extend({
 		instance.startX = e.data.posx;
 		instance.startY = e.data.posy;
 		instance.updatePos = null;
+		instance.noOverlay = instance.id == 'autoFillPreviewTooltip';
```
eszkadev, line 642: Please do not use hardcoded IDs in the Control.JSDialog (it's "framework" for all the dialogs).
Instead please put that `noOverlay` property into JSON of the popup you create :)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1482,6 +1482,23 @@ L.CanvasTileLayer = L.Layer.extend({
...
+			else if (tooltipInfo.type === 'autofillpreviewtooltip') {
+
+				var strTwips = textMsg.match(/\d+/g);
+				if (strTwips != null && this._map.isEditMode()) {
+					var topLeftTwips = new L.Point(parseInt(strTwips[0]), parseInt(strTwips[1]));
+					var offset = new L.Point(parseInt(strTwips[2]), parseInt(strTwips[3]));
+
+					var topLeftPixels = this._twipsToCorePixels(topLeftTwips);
+					var offsetPixels = this._twipsToCorePixels(offset);
+					this._cellAutoFillAreaPixels = L.LOUtil.createRectangle(topLeftPixels.x, topLeftPixels.y, offsetPixels.x, offsetPixels.y);
```
eszkadev, line 1494: can we somehow put this into event's data? in the `map.fire ( ... )`
so we will not need to store it inside canvas layer class

browser/src/control/Control.AutoFillPreviewTooltip.ts
eszkadev: but the position calculation differes than the base one?

bayramcicek: Yes, position calculation on the base one is following the cursor while editing the cell. But we need to get the bottom right location of the cell selected with autofill.

browser/src/control/Control.JSDialog.js
bayramcicek: I added `noOverlay: boolean;` into JSON and use this. I think it's ok :)
Thanks. Done.

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: well, at least we can move the popup right now. I needed to close the popup before update it :)

https://github.com/user-attachments/assets/747dfd19-5705-48e7-b27e-374be7a72f85

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1482,6 +1482,32 @@ L.CanvasTileLayer = L.Layer.extend({
...
+					// var strTwips = textMsg.match(/\d+/g);
+					// var tooltipInfo = JSON.parse(textMsg);
+
+					// tooltipInfo.location = this._map._docLayer._twipsToPixels(topLeftTwips);
+
+					tooltipInfo.location = this._map._docLayer._twipsToPixels(topLeftPixels);;
+
+					// tooltipInfo.location = this._map._docLayer._twipsToLatLng(buttonAreaTwips[1], this._map.getZoom());
+					// tooltipInfo.location = this._cellAutoFillAreaPixels;
+					this._map.fire('sendautofilllocation', { data: tooltipInfo });
```
bayramcicek, line 1507: @eszkadev: But I'm in a big trouble with positioning the popup.  I wonder if the core change (https://gerrit.libreoffice.org/c/core/+/172589) sends the correct location? or did we do some wrong calculations on the online side. I'm not sure. I need assistance on that. (cc: @gokaysatir )

gokaysatir:
@bayramcicek hello, thanks for the PR :)

I sent a small update to fix some issues. But there are still things to look:
* The position has a fixed offset. I hope it is easy to find.
* I added one more special case to JSDialogs along with AutofillMarker case. @eszkadev may have a better idea for this type of popup windows.
* Code may need some beautifications.

I hope this helps, thanks :)

browser/src/control/Control.JSDialog.js
```diff
@@ -710,8 +710,10 @@ L.Control.JSDialog = L.Control.extend({
 				instance.updatePos();
 			}
 
-			if (instance.isAutofilter)
+			if (instance.isAutofilter && instance.id !== 'autoFillPreviewTooltip')
```
eszkadev, line 713: pleae do not use hardcoded ids

gokaysatir: We enter the autofilter case there.

bayramcicek:
> @bayramcicek hello, thanks for the PR :)
> 
> I sent a small update to fix some issues. But there are still things to look:
> 
>     * The position has a fixed offset. I hope it is easy to find.
> 
>     * I added one more special case to JSDialogs along with AutofillMarker case. @eszkadev may have a better idea for this type of popup windows.
> 
>     * Code may need some beautifications.
> 
> 
> I hope this helps, thanks :)

Thank you so much :) I will check them asap.

browser/src/control/Control.JSDialog.js
eszkadev: I see `isAutofilter` is too general now: `instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();` 

Can we check if `instance.isAutoCompletePopup` can be used here?

browser/src/control/Control.JSDialog.js
```diff
@@ -706,8 +710,10 @@ L.Control.JSDialog = L.Control.extend({
...
 				this.calculateAutoFilterPosition(instance);
+			else if (instance.id === 'autoFillPreviewTooltip')
```
eszkadev, line 715: can be `instance.isAutoCompletePopup`?

browser/src/control/Control.AutoFillPreviewTooltip.ts
```diff
@@ -0,0 +1,107 @@
...
+				} as any as WidgetJSON,
+			] as Array<WidgetJSON>,
+			jsontype: 'dialog',
+			type: 'modalpopup',
+			cancellable: false,
+			popupParent: '_POPOVER_',
+			clickToClose: '_POPOVER_',
+			id: 'autoFillPreviewTooltip',
+			canHaveFocus: false,
+			noOverlay: true,
```
eszkadev, line 40: isAutoCompletePopup property might be needed

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: Popup location is now fine. I'm going to fix the Popup width.

browser/src/layer/tile/CanvasTileLayer.js
bayramcicek: I think this is done(?).

browser/src/control/Control.JSDialog.js
bayramcicek: I've used `instance.isAutoFillPreviewTooltip` here. 
Done.

browser/src/control/Control.JSDialog.js
bayramcicek: `instance.isAutoFillPreviewTooltip` is used here.
Done.

bayramcicek:
### issues
- [x] ~~popup location should be updated per autofill selection~~
- [ ] Popup width is too long

requires: https://gerrit.libreoffice.org/c/core/+/174502 (approved)

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -1487,6 +1487,21 @@ L.CanvasTileLayer = L.Layer.extend({
 				var tooltipInfo = JSON.parse(textMsg.substring(textMsg.indexOf('{')));
 				this._map.uiManager.showDocumentTooltip(tooltipInfo);
 			}
+			else if (tooltipInfo.type === 'autofillpreviewtooltip') {
+
+				var strTwips = textMsg.match(/\d+/g);
+				if (strTwips != null && this._map.isEditMode()) {
+
+					var cellRange = this._map._docLayer._parseCellRange(JSON.stringify(tooltipInfo.celladdress));
```
eszkadev, line 1495: All of these calculations don't depend on CanvasTileLayer data.
Can we do them inside Popup code? so here we can just do `_map.fire(....)` and rest will be done inside popup.

browser/src/control/Control.AutoFillPreviewTooltip.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/* global app */
+
+class AutoFillPreviewTooltip extends L.Control.AutoCompletePopup {
+	usageText: string;
+	newPopupData: PopupData;
+
+	constructor(map: ReturnType<typeof L.map>) {
+		super('autoFillPreviewTooltip', map);
+		this.newPopupData = {
```
eszkadev, line 23: This JSON is almost the same as parent one.
Can we just modify different fields from here?

```
this.newPopupData,canHaveFocus = false;
this.newPopupData,noOverlay = true;
this.newPopupData,isAutoFillPreviewTooltip = true;
```

browser/src/layer/tile/CanvasTileLayer.js
bayramcicek: All moved to `AutoFillPreviewTooltip.ts`

Done. Thanks!

browser/src/control/Control.AutoFillPreviewTooltip.ts
bayramcicek: removed `newPopupData` and added only different fields into `onAdd()`.

Done. Thanks!

eszkadev:
is this still WIP? (we are in the RED week, would be good to merge :) )

bayramcicek:
> is this still WIP? (we are in the RED week, would be good to merge :) )

I'll send last the patch today. then we can merge :)

eszkadev:
code looks good :) worth checking with @pedropintosilva if UX is right

eszkadev (approved):

browser/css/jquery-ui-lightness.css
```diff
@@ -522,7 +522,7 @@ button.ui-button::-moz-focus-inner {
 .ui-dialog .ui-dialog-content {
 	position: relative;
 	border: 0;
-	padding: .5em 1em;
+	padding: 0em 0.2em;
```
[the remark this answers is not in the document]
eszkadev: cc: @pedropintosilva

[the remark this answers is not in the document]
bayramcicek: Yes. that's right. let me quickly add this.

bayramcicek:
- Testing with RTL seems good to me.
- Freezing row/columns
  - sometimes, there is no scroll when entering the frozen area -> until mouse reaches to the top (or left) side -> seems unrelated(?) to this patch, and not(?) a blocker I think.

- https://github.com/user-attachments/assets/972f89e9-9502-4b31-9e5e-504d5373e9f4

  - scroll the unfrozen area a bit -> use autofill in the frozen area -> tooltip shown in the wrong location :/
    - core side thinks upper side(frozen side) is also scrolled(moved up) -> possibly... 
 - https://github.com/user-attachments/assets/89757f60-9368-4a3e-b4a9-818680963e50

pedropintosilva (approved):
Awesome! I liked when we had a call and I have checked again. Thanks @bayramcicek

browser/css/jquery-ui-lightness.css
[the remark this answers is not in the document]
bayramcicek: Done.
Thanks.

bayramcicek:
CI is green. Can we merge this @pedropintosilva @eszkadev ?

PR #10045, opened 2024-09-13
Add a page load indicator until all files are loaded.

gokaysatir:
Change-Id: Iad038bf4bca1ff3b8ae7cf52ef50e45ecce59207

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
let's get this in for RED

eszkadev:
but it doesnt fix the issue: we still see some form controls since page load
probably we should hide them directly in the html

Issue #9045, opened 2024-05-13
clipboard race causing cypress instability


mmeeks:
My recent logging commit changed the timing significantly when doing trace logging and seems to reliably break various unit tests.

One race that @vmiklos and I have identified is this:

  https://github.com/CollaboraOnline/online/pull/new/private/mmeeks/clipboardfix

A reasonably reliable fix would be to sleep after sending the .uno:Copy command - and before trying the web download I guess; I assume there used to be one there.

@neilguertincollabora you might be interested; patch still far from complete - I need to unwind the Promise logic more - and it seems we are doing 2x downloads of both text+html each time - at least, AFAICS both promises are executed for each mime type so ... need to unify that into one download I think.

HTH.

mmeeks:
Now merged.

PR #10051, opened 2024-09-16
Private/gokay/copy xx

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -50,8 +50,8 @@ class BrowserProperties {
 		// Firefox has undefined navigator.clipboard.read and navigator.clipboard.write,
 		// unsecure contexts (such as http + non-localhost) has the entire navigator.clipboard
 		// undefined.
-		hasNavigatorClipboardRead = navigator.clipboard && navigator.clipboard.read,
-		hasNavigatorClipboardWrite = navigator.clipboard && navigator.clipboard.write,
+		hasNavigatorClipboardRead = navigator.clipboard && navigator.clipboard.read !== undefined,
+		hasNavigatorClipboardWrite = navigator.clipboard && navigator.clipboard.write !== undefined,
 		webkit    = ua.indexOf('webkit') !== -1,
 		chrome    = ua.indexOf('chrome') !== -1,
 		gecko     = (ua.indexOf('gecko') !== -1 || (cypressTest && 'MozUserFocus' in doc.style))
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -13,7 +13,7 @@
  * local & remote clipboard data.
  */
 
-/* global app _ brandProductName $ ClipboardItem Promise GraphicSelection */
+/* global app _ brandProductName $ ClipboardItem GraphicSelection */
 
 // Get all interesting clipboard related events here, and handle
 // download logic in one place ...
@@ -37,10 +37,6 @@ L.Clipboard = L.Class.extend({
 		this._dummyDiv = div;
 		this._dummyPlainDiv = null;
 		this._dummyClipboard = {};
-
-		// Tracks waiting for UNO commands to complete
-		this._commandCompletion = [];
-		this._map.on('commandresult', this._onCommandResult, this);
 
 		div.setAttribute('id', this._dummyDivName);
 		div.style.userSelect = 'text !important';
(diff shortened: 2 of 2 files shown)


vmiklos (commented):
I assume the "Check clipboard capabilities better." commit is just a style one, I'm fine with that.

I'm afraid the "Refactor some of clipboard.write code." commit message is so brief that I don't understand what the commit does. :-)

Could you please describe in the commit message:

- what is the problem that you are fixing (how to reproduce it, so I can follow those steps when I test your work)
- it seems you more or less revert the race fix from ff8dbe7fdea26087736584f0057e01d168b77777, which was @mmeeks 's fix for #9045 -- it would be good to document the reasoning why this is safe to do
- I assume now your new code handles both the old problem (that was the race) and some new one (that you'll describe above), it would be good to document how this unified solution works
- is any of the problem you fix here something you can cover from Cypress?

Thanks. :-)

gokaysatir:
@vmiklos  thanks for the comment :)

I updated the commit message of the last commit.

Also, i didn't see Cypress issue before your comment. Then CI failed :) Anyway, it's also fixed i think.

vmiklos:
Thanks, glad to see the commit message is improving. :-)

The issues I see unresolved in the last commit are:

1) Am I correct that you implicitly revert ff8dbe7fdea26087736584f0057e01d168b77777 here, as in we no longer wait for the cut/copy to complete and just hope that your HTTP GET of the clipboard will be late enough? If so, could you explain why that logic is no longer needed?

2) The commit message says 'Also, Cypress tests use a dummy copy-paste-container. This commit makes some changes for Cypress tests.', but the commit seems to only modify browser/src/map/Clipboard.js?

3) The commit message says 'Copy some text from Writer and paste it into a desktop office file or a text file.', but if I naively try that use-case, it works for me even without your change -- could you please be more specific what condition is needed to reproduce this?

I wonder, perhaps is it useful to restrict this PR to the first 4 commits, which are less controversial, get that merged and have a separate PR for the last commit?

Also, if you're not comfortable with a step by step reproducer list for the problem in the commit message, maybe file an issue for the problem (that also allows screenshots) and just refer to that in the commit message?

Thanks.

gokaysatir:
> Thanks, glad to see the commit message is improving. :-)
> 
> The issues I see unresolved in the last commit are:
> 
> 1. Am I correct that you implicitly revert [ff8dbe7](https://github.com/CollaboraOnline/online/commit/ff8dbe7fdea26087736584f0057e01d168b77777) here, as in we no longer wait for the cut/copy to complete and just hope that your HTTP GET of the clipboard will be late enough? If so, could you explain why that logic is no longer needed?
> 2. The commit message says 'Also, Cypress tests use a dummy copy-paste-container. This commit makes some changes for Cypress tests.', but the commit seems to only modify browser/src/map/Clipboard.js?
> 3. The commit message says 'Copy some text from Writer and paste it into a desktop office file or a text file.', but if I naively try that use-case, it works for me even without your change -- could you please be more specific what condition is needed to reproduce this?
> 
> I wonder, perhaps is it useful to restrict this PR to the first 4 commits, which are less controversial, get that merged and have a separate PR for the last commit?
> 
> Also, if you're not comfortable with a step by step reproducer list for the problem in the commit message, maybe file an issue for the problem (that also allows screenshots) and just refer to that in the commit message?
> 
> Thanks.

1 * ClientRequestDispatcher: line 862 -> handleClipboardRequest. Socket is moved here in line 1145.
  ClientSession: line 263 -> handleClipboardRequest.
  ClientSession: line 312 -> docBroker->forwardToChild(...getclipboard) The request is forwarded. Socket is pushed into local queue.
  ClientSession: line 2098 -> kit sends the "clipboardcontent" message. Data is sent over the moved socket and the socket queue is cleared.
  Socket is closed (our url fetch ends here).

  Client (browser) side waits until socket sends data, i think there is no timing related issue. Please let me know if i am missing something.

2 * I corrected the commit message about Cypress tests. The reason Cypress tests were not passing (IMHO):
  - We are getting the copied content over a url fetch, the message doesn't touch normal code paths, so the dummyClipboard was not set as before. I used your "app.map._clip._dummyClipboard.write" function to set the clipboard for tests.
  - I also added a check for removing style data. Normally we remove the style data from HTML content for CSP compliance. But some of the Cypress tests check "style" tags' existence. So i kept "style" tags for only Cypress tests.
  - These are the Cypress related changes.
 
3 * I think the function "_navigatorClipboardWrite" was not used for a while. Without that 4th commit, the boolean variable "hasNavigatorClipboardWrite" was always "false" for me.
  Even we omit that check, with latest changes, i think the data format was broken, the pasted text format was unexpected - having HTML content in it. There was also JS errors related to CSP settings.
  caniuse.com shows that most of the modern browsers support "clipboard.write" (https://caniuse.com/mdn-api_clipboard_writetext). This is also an indicator that this "_navigatorClipboardWrite" was not used for sometime. Either it failed or the boolean variable was false and fallback methods were used. And the code makes sure that if it fails once, it is never tried again.
  So, for testing the issue this last commit solves, we may need to:
  - First make sure that "_navigatorClipboardWrite" is called without these changes - some minor modifications like omitting the checks may be enough for this.
  - Clear the browser cache.
  - Copy some content.
  - Then paste the content into desktop Writer app.
  - It should either paste some HTML content (you see the HTML tags) or it says something like "download the content first etc".

By the way, i could reproduce the issue on my local, with and without an integrator running. But before i wasn't able to reproduce, not sure what causes that difference.

vmiklos (commented):
> 1 * ClientRequestDispatcher: line 862 -> handleClipboardRequest. Socket is moved here in line 1145. ClientSession: line 263 -> handleClipboardRequest. ClientSession: line 312 -> docBroker->forwardToChild(...getclipboard) The request is forwarded. Socket is pushed into local queue. ClientSession: line 2098 -> kit sends the "clipboardcontent" message. Data is sent over the moved socket and the socket queue is cleared. Socket is closed (our url fetch ends here).
> 
> Client (browser) side waits until socket sends data, i think there is no timing related issue. Please let me know if i am missing something.

Yes, I think it's wise to re-read the commit message of Michael's race fix. On one hand, we send an uno command (cut or copy) down the websocket, that'll be send to core at some point, in an async way. On the other hand, we fetch the clipboard content using synchronous (as far as the server is concerned) HTTP GET. The timing problem happens if you get the clipboard data before the uno command is processed. The code you removed makes sure the commit goes first, then the HTTP GET. Does that help?

> 2 * I corrected the commit message about Cypress tests. The reason Cypress tests were not passing (IMHO):
> 
>     * We are getting the copied content over a url fetch, the message doesn't touch normal code paths, so the dummyClipboard was not set as before. I used your "app.map._clip._dummyClipboard.write" function to set the clipboard for tests.

I believe this is just a style change: we used to call write() on navigator.clipboard normally, and on this._dummyClipboard in the test case. Now you call app.map._clip._dummyClipboard.write() in the test case, else you call navigator.clipboard.write(). I'm not against the change, but could you please separate this non-functional change from the actual fix? It helps me a lot when it comes to code review.

>     * I also added a check for removing style data. Normally we remove the style data from HTML content for CSP compliance. But some of the Cypress tests check "style" tags' existence. So i kept "style" tags for only Cypress tests.

Did you consider changing those tests to not expect style? The more "if test" code we add, the more likely it is that a real problem is there, but our tests don't see it.

> caniuse.com shows that most of the modern browsers support "clipboard.write" (https://caniuse.com/mdn-api_clipboard_writetext).

Oh, looks like something changed here very recently. https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write says "127 (Released 2024-06-11)" supports this unconditionally, while in the past (this spring, when I worked on the new clipboard code), navigator.clipboard.read and write was undefined on Firefox, unless you used about:config to opt in.

Also note that the site writes "This method must be called within user gesture event handlers.", and that's also true in Chrome/Safari. So you must call navigator.clipboard.read/write in the the magic security context when the event handler is running, calling navigator.clipboard.read/write may or may not work when you're in a callback of an UNO command listener or a fetch() request.

I believe the write case is easier, but Safari still throws an exception when you try to write outside an event listener. (Copy from COOL to an other app.)

So the solution for that in the old code is that you don't call clipboard.navigator.write in the callback, you call it instantly, but instead of using a data parameter, you use a Promise parameter, so the clipboard will resolve your promise and then that works nicely with the security rules.

> By the way, i could reproduce the issue on my local, with and without an integrator running. But before i wasn't able to reproduce, not sure what causes that difference.

I have to confess that even if I don't have anything in local storage (no flag to say "don't even try the new clipboard API"), copying e.g. from share to local Writer works fine for me. That's why I don't yet understand what do you actually fix here. :-)

I hope that once you restore the code to call clipboard.navigator.write in the event handler directly and restore the race fix, the diff will be much smaller and then we'll see what do you actually change that fixes something practical.

To sum up, please:

- [ ] restore the race fix
- [ ] exclude any non-functional changes from the last commit, doing it in a separate PR is fine
- [ ] try to not disable style stripping for tests, adapt tests instead
- [ ] restore the setup that navigator.clipboard.write happens directly in the event handler, not in a callback

Is that doable? Thanks a lot!

gokaysatir:
@vmiklos  i may ping you for the issue tomorrow as you suggested. Since you couldn't reproduce, maybe we can have a call etc. Either way, this PR may be abandoned for a new one and i'm planning to document the solution after we pick one, thanks for the comments :)

gokaysatir:
Closing this in favor of:
* https://github.com/CollaboraOnline/online/pull/10065
* https://github.com/CollaboraOnline/online/pull/10070

PR #10056, opened 2024-09-17
Fix build with feature lock

eszkadev:
When built with additional configure flag: --enable-feature-lock

regression from commit cdbc3af0d764b46458de5ea696b695ade79c662d
cool#10042: Close incoming HTTP Connection if requested (Regression of cool#9621)


sgothel (approved):
Thank you for fixing, my bad!

PR #10057, opened 2024-09-17
Slideshow: use the correct displaySlides_ method for transition

codewithvk:
Change-Id: I8e131a691a5616a8636af48919d290e3c3ea7491

* Resolves: # 
* Target version: master


eszkadev (approved):
code looks ok

PR #10058, opened 2024-09-17
feat(android): Use native copy/paste

Minion3665:
Copy/paste is utterly broken on Android, but Android apps have a really
high level of permission to the clipboard: they are allowed to access
it, without further consent or notification, as long as they are in the
foreground(!!).

This permission does not apply to applications in a webview, they have
the navigator.clipboard object but - it appears - no way to request
permissions for it, leaving them unable to properly use it.

Instead of worrying about that, we can write a native bridge to let us
use the high permissions of the Android app while we are in the webview.
This commit implements that native bridge.

The same issue exists on iOS, but this pull request does not fix it at this time. Another commit (either in this PR or a later one) will be needed to handle it in a similar way to Android

### DIFF
--- a/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
+++ b/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
@@ -959,6 +959,40 @@ public class LOActivity extends AppCompatActivity {
     }
 
     /**
+     * Specialized function for Collabora Online, returns an array of length 2 where the first
+     * element is the clipboard's text content, and the second is the clipboard's HTML content. One
+     * or both of these may be null if the clipboard does not contain it.
+     */
+    @JavascriptInterface
+    public String[] readFromClipboard() {
+        ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+
+        ClipData clipboardData = clipboard.getPrimaryClip();
+
+        if (clipboardData == null) return new String[] { null, null };
+
+        ClipData.Item clipboardItem = clipboardData.getItemAt(0);
+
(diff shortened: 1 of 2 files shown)


Minion3665:
While this works - and might be nice to have in the future (particularly looking to pick the async/await commit, and maybe the other one if we want to tidy up Android clipboard use elsewhere) - there's a lot simpler of a fix here

```
Commit ID: 2e0d966ead3c11ceab172b7774295083a3703422
Change ID: vqwmwrvqxnupkpqlvrrtyorpwoqystvx
Author: Skyler Grey <skyler.grey@collabora.com> (2024-09-18 10:20:43)
Committer: Skyler Grey <skyler.grey@collabora.com> (2024-09-18 11:43:10)

    (no description set)

diff --git a/browser/src/map/Map.js b/browser/src/map/Map.js
index 1dcee2186a..9308a88917 100644
--- a/browser/src/map/Map.js
+++ b/browser/src/map/Map.js
@@ -69,7 +69,9 @@
                        this.options.documentContainer = L.DomUtil.get(this.options.documentContainer);
                }
 
-               this._clip = L.clipboard(this);
+               if (!window.ThisIsAMobileApp) {
+                       this._clip = L.clipboard(this);
+               }
                this._initContainer(id);
                this._initLayout();
```

It seems that the L.clipboard overwrites the previous handling (that works in the Android app) and doesn't correctly exit out when `navigator` can't possibly work for Android - skipping the initialization was probably intended in the first place, given this warning

```
    ======> Assertion failed!? No L.Clipboard object should be needed in a mobile app
```

this patch, however, and the WIP iOS one like it bring us tantalizingly close to being able to remove the bulk of weird clipboard workarounds on Android and iOS - which would be really nice.

Minion3665:
> there's a lot simpler of a fix here

I've quickly realized that this is only half-true, as the patch I posted in here only half-works. Oh. Oops.

PR #10064, opened 2024-09-17
slideshow: do not cache page numbers

eszkadev:
followup for commit b8a951fa25db717724a6ff469b6d5ce3d1a873af slideshow: don't cache fields for now

somehow the previous patch was not working after the merge


quikee (approved):
Looks good to me.

PR #10065, opened 2024-09-17
Private/gokay/copy yy

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -1275,8 +1275,11 @@ L.Clipboard = L.Class.extend({
 
 	_warnLargeCopyPasteAlreadyStarted: function () {
 		this._map.uiManager.showInfoModal('large copy paste started warning');
-		document.getElementById('large copy paste started warning').innerHTML = _('<p>A download due to a large copy/paste operation has already started. ' +
-		'Please, wait for the current download or cancel it before starting a new one</p>');
+		const container = document.getElementById('large copy paste started warning');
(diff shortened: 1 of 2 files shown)


vmiklos (approved):
This reads nicely to me, I like how the new code can't forget to escape HTML. Thanks!

browser/src/map/Clipboard.js
```diff
@@ -1268,7 +1306,31 @@ L.Clipboard = L.Class.extend({
 		var box = document.getElementById(id + '-box');
 		var innerDiv = L.DomUtil.create('div', '', null);
 		box.insertBefore(innerDiv, box.firstChild);
-		innerDiv.innerHTML = msg;
+
```
timar, line 1309: I'm sorry that I joined late to the party. This chunk of the code is sub-optimal from the localization perspective. It is not possible to localize sentences broken in two pieces. 

Part 1: Please press
Part 2: to see more options

E.g. how to translate "Please press", if in some language the sentence translates to "To see more options please press <something>". Impossible.

vmiklos: @gokaysatir would it be reasonable to revert just this part *and* add a comment that content is built this way for l10n reasons, so the next person won't do the same mistake? Thanks.

gokaysatir: Well, this makes the issue complicated :) I will find another solution that satisfies both l10 and CSP sides.

PR #10066, opened 2024-09-18
Prerequisites for Smarter Upload Failure Handling

Ashod:
This is in preparation to re-synchronize the LastModifiedTime after we fail to upload.
_Best to review one commit at a time._

- wsd: wopi: remember the document size we uploaded
- wsd: wopi: track the document size on the server
- wsd: wopi: return the file-size from uploadLocalFileToStorageAsync
- wsd: centralize document conflict handling

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1052,20 +1052,7 @@ bool DocumentBroker::download(
                                  << _storageManager.getLastModifiedTime()
                                  << ", Actual: " << fileInfo.getLastModifiedTime());
 
-            _documentChangedInStorage = true;
-            // Do not reload the document ("close: documentconflict") if there are
-            // any changes in the loaded document, either saved or unsaved.
-            const std::string message =
-                (_lastStorageAttrs.isUserModified() || _currentStorageAttrs.isUserModified() ||
-                 isPossiblyModified())
-                    ? "error: cmd=storage kind=documentconflict"
-                    : "close: documentconflict";
-
-            if (session)
-            {
-                session->sendTextFrame(message);
-                broadcastMessage(message);
-            }
+            handleDocumentConflict();
         }
     }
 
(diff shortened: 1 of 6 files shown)


wsd/Storage.cpp
```diff
@@ -367,12 +365,20 @@ void LocalStorage::uploadLocalFileToStorageAsync(const Authorization& /*auth*/,
...
+
         // update its fileinfo object. This is used later to check if someone else changed the
         // document while we are/were editing it
-        setLastModifiedTime(
-            Util::getIso8601FracformatTime(FileUtil::Stat(path).modifiedTimepoint()));
+        setLastModifiedTime(Util::getIso8601FracformatTime(stat.modifiedTimepoint()));
         LOG_TRC("New FileInfo modified time in storage " << getLastModifiedTime());
         res = UploadResult(UploadResult::Result::OK);
+
+        if (asyncUploadCallback)
```
caolanm, line 376: before this change, if there was an exception and if asyncUploadCallback is present, then we would call:

asyncUploadCallback(... res) with the res of: UploadResult::Result::FAILED, "Internal error" post-catch

but after we no longer do that, so we now have an initially FAILED res which is always overwritten by an OK before use, so no point in having res declared before the try-catch block if that change is intentional?

It feels like we just want to change this to return the size, so maybe leave the call to asyncUploadCallback where it was and just assign stat.size() to a local var that defaults to 0 and return that local on function exit instead?

wsd/DocumentBroker.cpp
```diff
@@ -1052,20 +1052,7 @@ bool DocumentBroker::download(
...
-            // any changes in the loaded document, either saved or unsaved.
-            const std::string message =
-                (_lastStorageAttrs.isUserModified() || _currentStorageAttrs.isUserModified() ||
-                 isPossiblyModified())
-                    ? "error: cmd=storage kind=documentconflict"
-                    : "close: documentconflict";
-
-            if (session)
-            {
-                session->sendTextFrame(message);
```
caolanm, line 1066: I presume that broadcastMessage duplicates the sendTextFrame line here so its loss is ok?

Ashod: Yes. The broadcast already sends to all sessions, including this one. So sending it explicitly here was unnecessary.

wsd/Storage.cpp
Ashod: My concern was that even with an exception we are setting the state as `Complete`, when `FAILED` is more appropriate.

In the new code, with an exception it will be `FAILED` and only when we reached the end do we set it to `Complete`, right before returning the size.

Your suggestion of having a local for size (defaulted to 0) is fine, but regarding what we pass `asyncUploadCallback` in the event of an exception, it's orthogonal.

caolanm (approved):

PR #10067, opened 2024-09-18
slideshow: fix present from current slide

eszkadev:
When we close fullscreen we don't call escape handler as earlier we turn off it's handler so we didn't properly clear status of navigator.

This caused bug when presenting second time. We didn't start from current slide but the last slide of previous fullscreen slideshow.


quikee (approved):
Looks good to me.

PR #10070, opened 2024-09-18
Use a function for fetch instead of a variable.

gokaysatir:
So we create the promise when the function is called. Or fetch starts too early.

Change-Id: I8905b637692fd3bb52262991118740456b7a17bd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/map/Clipboard.js
```diff
@@ -866,15 +866,15 @@ L.Clipboard = L.Class.extend({
 	},
 
 	// Gets status of a copy/paste command from the remote Kit
-        _onCommandResult: function(e) {
-                if (e.commandName === '.uno:Copy' || e.commandName === '.uno:Cut')
+    _onCommandResult: function(e) {
```
vmiklos, line 869: Nit: I think the indent of this line used to match with the `}` at the end of the previous function, and now it doesn't do that anymore.

vmiklos (approved):
Hrm, I though that the point of returning a Promise is that you don't have to produce an actual result (to be async), you only return a Promise. So it confuses me a bit that creating the Promise too early causes fetch() to start too early. But I don't see how this would be problematic (probably just my lack of deep JS knowledge), so I'm not against the PR.

PR #10071, opened 2024-09-18
private/marco/first autoeffect and handle unsupported effects

mcecchetti:
- **animation engine: handle properly first automatic effect**
- **animation engine: do not create animation nodes for unsupported effects**

### DIFF
--- a/browser/src/slideshow/engine/AnimatedElement.ts
+++ b/browser/src/slideshow/engine/AnimatedElement.ts
@@ -263,6 +263,13 @@ class AnimatedElement {
 	private slideRenderer: SlideRenderer;
 	private bIsValid: boolean = false;
 
+	public static readonly SupportedProperties = new Set<string>([
+		'x',
+		'y',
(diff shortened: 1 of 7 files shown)


eszkadev:
nice I think it's better state (no need for N-times-click), but I noticed that some animations which I saw in the regular "hello world" presentation are missing now

mcecchetti:
> nice I think it's better state (no need for N-times-click), but I noticed that some animations which I saw in the regular "hello world" presentation are missing now

I suppose it's because animation nodes involving text targets are no more created. They often don't work, requiring several clicks in order to go on. When developing it's possible to relax this filter by hacking the `isTargetSupported` in `SlideAnimations.ts`.
Moreover after investigating further I found out that hashes involving text shapes are not reliable. On playing a presentation, for the same text shape, sometimes they match the hash included the animations info, sometimes they don't match.

eszkadev (approved):

PR #10073, opened 2024-09-18
Fix: Restrict invert background feature to document content only

Darshan-upadhyay1110:
- Ensure that background inversion applies exclusively to the document page/slide and not the surrounding areas.
- This fixes unintended behavior where inversion affected the non-document regions, which is inconsistent with expected behavior and user experience in most editing environments.

Change-Id: Id3113185b1cd954ad48c9b63df573a919c871049

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -21,6 +21,8 @@
 	--color-stylesview-background: white;
 	--brightness-stylesview: 0.9;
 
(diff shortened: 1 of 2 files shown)


Minion3665 (approved):
This looks sensible, thanks for the patch!

PR #10074, opened 2024-09-18
slideshow: fix start from current slide

eszkadev:
it was not fixed previously after rebase ...

### DIFF
--- a/browser/src/slideshow/SlideShowPresenter.ts
+++ b/browser/src/slideshow/SlideShowPresenter.ts
@@ -145,7 +145,8 @@ class SlideShowPresenter {
 			);
 			this.centerCanvas();
 		} else {
-			this._stopFullScreen();
+			// we need to cleanup current/prev slide
+			this._slideShowNavigator.endPresentation(true);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -146,6 +146,9 @@ class SlideShowPresenter {
 			this.centerCanvas();
 		} else {
 			this._stopFullScreen();
+			// we need to cleanup current/prev slide,
+			// Escape handler is not called as we disabled it above
+			this._slideShowNavigator.endPresentation(false);
```
mcecchetti, line 151: It seems to me that by passing `false` to `SlideShowNavigator.endPresentation` in the endless case we end up invoking `startTimer`: see `SlideShowPresenter.endPresentation`. Moreover that last method already takes care to invoke `_stopFullScreen`. So maybe it could be enough to invoke `SlideShowNavigator.quit()` inside the `else` branch, as it's done when the separate window is closed (see `this._windowCloseInterval = setInterval ...` in `_doInWindowPresentation`).
Anyway I didn't test it.

eszkadev: quit is also fine, done

browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -146,6 +146,8 @@ class SlideShowPresenter {
 			this.centerCanvas();
 		} else {
 			this._stopFullScreen();
```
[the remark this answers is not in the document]
eszkadev: it didn't work (order matters)

mcecchetti (approved):
Thanks for your fix!

PR #10076, opened 2024-09-18
Fix hover text labels in Table menu for consistency and clarity

Darshan-upadhyay1110:
+ Corrected the hover text labels in the Table menu to ensure they accurately describe the actions they represent:
+ Adjustments are done based on which string should be displayed from unocommand.js file( "Context" or "menu")

This resolves inconsistencies and improves user experience by providing clearer descriptions for each menu action.- 

Updated string :

-  "Cell" to "Select Cell"
-  "Table" to "Delete Table" and "Select Table" accordingly
-  "Row" to "Select Row"
-  "Column" to "Select Column"
-  "Columns" and "Rows" to "Delete Selected Columns" and "Delete Selected Rows"

Change-Id: Iec2f04434eb175cbf5a38a5b91dd3e4e896ce7e9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -2357,7 +2357,7 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 							},
 							{
 								'type': 'toolitem',
-								'text': _UNO('.uno:DeleteColumns', 'text'),
+								'text': _UNO('.uno:DeleteColumns', 'text', true),
 								'command': '.uno:DeleteColumns'
 							}
 						]
@@ -2377,7 +2377,7 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 							},
 							{
 								'type': 'toolitem',
-								'text': _UNO('.uno:DeleteRows', 'text'),
+								'text': _UNO('.uno:DeleteRows', 'text', true),
 								'command': '.uno:DeleteRows'
 							}
 						]
@@ -2444,7 +2444,7 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 			},
 			{
 				'type': 'bigtoolitem',
-				'text': _UNO('.uno:EntireCell', 'text'),
+				'text': _UNO('.uno:EntireCell', 'text', true),
 				'command': '.uno:EntireCell'
 			},
 			{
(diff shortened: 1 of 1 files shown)


timar (changes requested):
1st line of unocommands.js:
`// Don't modify, generated using unocommands.py`

browser/src/unocommands.js
```diff
@@ -100,7 +100,7 @@ var unoCommandsArray = {
 	'DeleteAuthor':{global:{menu:_('Delete All Comments by This Author'),},},
 	'DeleteAxis':{global:{menu:_('Delete Axis'),},},
 	'DeleteCell':{spreadsheet:{context:_('De~lete...'),menu:_('Delete C~ells...'),},},
-	'DeleteColumns':{global:{context:_('Delete selected columns'),menu:_('Delete ~Columns'),},text:{context:_('Delete selected columns'),menu:_('~Columns'),},},
+	'DeleteColumns':{global:{context:_('Delete selected columns'),menu:_('Delete ~Columns'),},text:{context:_('Delete selected columns'),menu:_('Delete ~Columns'),},},
```
eszkadev, line 103: I see duplicated `menu` here. Did you use script to generate this? I think it shouldn't be modfied manually.
If we use wrong text somewhere, then might be usefult to add the type specifier so we will use other variant:
`_UNO('.uno:PageDown', 'text', here some variant option)`
https://github.com/CollaboraOnline/online/blob/e07a403d0aa2017d4c6ccf69c05e56d466e9ead5/browser/src/unocommands.js#L528

timar (approved):

Darshan-upadhyay1110:
> The “Column Before” button does “Column After”, and the “Column After” button does “Column Before”, so should these buttons functions be swapped around?

Hii @rpear i just tested this and it does work okay as expected. "Column Before" insert column before in table for writer and same is okay for  "Column After". 

Can you please explain what is wrong here ?

Darshan-upadhyay1110:
> Could the buttons be renamed from “Column Before” to “Insert Column Before,” and “Column After” to “Insert Column After”? 

> Additionally, while “Rows Above” and “Rows Below” function correctly, could we rename these buttons to “Insert Rows Above” and “Insert Rows Below,” respectively? 

Absolutely, we can do that! :)

> I could create a ticket if needed.

Thanks, @rpear! It would be fantastic if you could create one. If you’re interested in tackling this issue, I can provide you with code pointers. It’s a straightforward fix! The PR itself serves as a code pointer; just follow what I did in that commit, and you’ll be able to resolve the issue easily.

rpear:
Hi Darshan,  OK I'd like to have a go at that, I guess there's no urgency.
I'll get back to you.  Rob

rpear:
See https://github.com/CollaboraOnline/online/issues/10175
Fix more hover text labels in Table menu for consistency and clarity #10175

PR #10077, opened 2024-09-18
private/jaume/writer comments ui

JaumePujantell:
- annotations: show parent child realtions in layout
- annotations: nicer scroll on select
- annotations: added option to show a thread in a pop-up


mikekaganski (approved):
Thanks - very nice function!

PR #10078, opened 2024-09-19
Slideshow: Enhance flip tiles transition

codewithvk:
Slideshow: Enhance flip tiles transition
* Added WebGL flags to enable multi-texture rendering for simultaneous display of both slides.
* Fixed a potential GL state issue in the displaySlides function, ensuring proper handling of texture binding and GL state preservation during the transition.

Change-Id: I15d0a7c418225e1f707f16fcb65ade68f4c90ad6

* Resolves: # 
* Target version: master


quikee (approved):
Looks fine to me

PR #10082, opened 2024-09-20
slideshow: allow to exit with 2d only mode

eszkadev:
the last slide with text "click to exit the presentation" requires enabled webgl in the browser. Avoid error if it is missing and just close the slideshow.


quikee (approved):
Looks fine to me

PR #10084, opened 2024-09-20
Document Load Event

Ashod:
- wsd: handle on-document-load events only once

### DIFF
--- a/test/UnitWOPILock.cpp
+++ b/test/UnitWOPILock.cpp
@@ -65,7 +65,7 @@ public:
     {
         const std::string lockToken = request.get("X-WOPI-Lock", std::string());
         const std::string newLockState = request.get("X-WOPI-Override", std::string());
-        LOG_TST("In " << toString(_phase) << ", X-WOPI-Lock: " << lockToken << ", X-WOPI-Override: "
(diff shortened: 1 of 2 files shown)


net/HttpRequest.hpp
```diff
@@ -1397,7 +1397,7 @@ class Session final : public ProtocolHandlerInterface
 
         std::shared_ptr<StreamSocket> socket = _socket.lock();
         if (socket)
-            socket->dumpState(os);
+            os << indent << "socket: #" << socket->getFD();
```
caolanm, line 1400: Should we ditch the workaround of https://github.com/CollaboraOnline/online/commit/7769161afb19f18c484f90931bce0fc00c041dcc as a follow up?

caolanm (approved):

caolanm:
unit-wopi-lock seems to be repeatedly failing here, which is suspicious

net/HttpRequest.hpp
Ashod: Yes. I didn't remove it just yet until we all agree it's safe.

Ashod:
> I thought we merged the loading: stuff already; perhaps I was confused =)

@mmeeks, yes, that was in the tests, this is in wsd. Inefficiencies and repeated "initialization" that I didn't realize we were doing. Now improved.

PR #10086, opened 2024-09-20
browser/html: allow fullscreen for dev pages

meven:
Change-Id: Ieaa9824c7f569965f0ac8ebb9e8f92e8843195c0

Seems bests to have it here, even though it is only necessary for XSS but it spread the knowledge and could prevent issues

cc @hfiguiere

### DIFF
--- a/browser/html/debug.html
+++ b/browser/html/debug.html
@@ -18,7 +18,7 @@
     <form id="form" action enctype="multipart/form-data" target="coolframe" method="post">
          <input name="access_token" value="test" type="hidden"/>
     </form>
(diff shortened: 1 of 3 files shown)


hfiguiere:
Not completely sure as it is not necessary (same origin)

meven:
> Not completely sure as it is not necessary (same origin)

Aware of this, this is to act as a realistic example, users can copy/paste.
I should rephrase the commit message

hfiguiere:
LGTM, but I'd love to see this in the sdk documentation too.

hfiguiere (approved):

PR #10087, opened 2024-09-20
slideshow: do not render invisible element for transition

eszkadev:
if we switch slides with some transition and on the next slide there is animated layer with initial visibility = false then we need to avoid rendering it. without that we had blinking object which was visible during the transition but not when it was finished.

### DIFF
--- a/browser/src/slideshow/LayerDrawing.ts
+++ b/browser/src/slideshow/LayerDrawing.ts
@@ -578,6 +578,7 @@ class LayerDrawing {
 						}
 						return; // no layer means it is not visible
 					}
+					if (!animatedElement.getVisibility()) return;
 				}
 				this.drawBitmap(content.content as ImageInfo);
 			}


browser/src/slideshow/LayerDrawing.ts
```diff
@@ -578,6 +578,7 @@ class LayerDrawing {
 						}
 						return; // no layer means it is not visible
 					}
+					if (!animatedElement.getVisibility()) return;
```
mcecchetti, line 581: So, with this hack when an animated element is not valid the related shape never becomes visible, I remember we agreed that should be avoided.
I'm working on a patch that both refactored `AnimatedElement` and add support for scaling animations. I tested your sample file and I'm not able to reproduce the issue. I think that the reported problem is due to the fact that an animated element is set to valid only on executing `notifySlideStart`, that is, after the transition is over. In fact with the new `AnimatedElement` implementation, an animated element is set to valid as soon as Online receive the related animated layer, that is, before it tries to render that layer. So it shouldn't anymore occur that the animated element is valid but it has not yet been set to valid when the layer is rendered.

eszkadev: oh, good news then. Yes, the problem was that we set it to valid only after transition was finished.
My approach according to my tests also did something similar. Because we get the new value of getVisibility only if layer has arrived and has `initVisibility` set. as default we can setup `true`.
But if you have a patch, we can wait and not merge this one.

mcecchetti: I pushed the patch: https://github.com/CollaboraOnline/online/pull/10097

eszkadev:
merged the other patch from Marco

PR #10099, opened 2024-09-23
Private/gokay/hyperlink

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -13,7 +13,7 @@
  * local & remote clipboard data.
  */
 
-/* global app _ brandProductName $ ClipboardItem Promise GraphicSelection */
+/* global app DocUtil _ brandProductName $ ClipboardItem Promise GraphicSelection */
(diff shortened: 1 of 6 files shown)


eszkadev (approved):
I didn't test but looks like nice cleanup

browser/src/map/Clipboard.js
```diff
@@ -972,6 +932,10 @@ L.Clipboard = L.Class.extend({
 				text = text.substring(idx, text.length);
 			textHtml = text;
 		}
+
+		if (!app.sectionContainer.testing)
```
eszkadev, line 936: what is that `testing` flag?

gokaysatir: It's "true" when Cypress tests is active. We normally remove styles from HTML string but in case of testing, we don't. So we can see the affect of the uno commands inside the incoming HTML.

PR #10105, opened 2024-09-24
private/marco/fix rewind to previous slide

mcecchetti:
- **slideshow: check for hidden slide in SlideShowNavigator.displaySlide**
- **slideshow: anim engine: rewind to prev slide  didn't work as expected**


eszkadev (changes requested):
If we open presentation with hidden slide:
1. select hidden slide
2. start presentation from current slide:
result: black screen, on click - error

mcecchetti:
> If we open presentation with hidden slide:
> 
> 1. select hidden slide
> 2. start presentation from current slide:
>    result: black screen, on click - error

Thanks for catching this. I had forgot to `return` after invoking `displaySlide` with next/prev index. Now it should work.

eszkadev (approved):
thanks, now it works

PR #10108, opened 2024-09-24
We shouldn't split the sentences into multiple parts if they are to b…

gokaysatir:
…e translated.

If we split the sentences, we would have guessed that every language has the splitted parts in the same order. Since this case is not possible, we will try to keep one sentence as a whole string.

This commit fixes a regression that is caused by below PR:
* https://github.com/CollaboraOnline/online/pull/10065

Now the string is not splitted into 2 and the constant text is sent to another paragraph.

Before:

After:


timar (approved):
Thank you, it's a good workaround of the problem.

PR #10109, opened 2024-09-24
Fix : Automatic focus jump during tab navigation

Darshan-upadhyay1110:
Accessibility: Set focus on the first or selected tab in dialogs with tab

- Implemented a recursive search for the 'tabcontrol' type within the dialog object.
- Updated the function to return the target element to focus on, or null if no tabcontrol is found.
- Removed direct focus call from the function, allowing for more flexible handling of the returned element.

To reproduce this issue
-  Open "Table of Content" dialog in ODT file
- then try to move focus using "Right Arrow key"
- you will see focus is inside the Tab content and not on the Tab itself
- Expected should be to have focus on Tab in case of dialog with tab control

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -383,9 +383,19 @@ L.Control.JSDialog = L.Control.extend({
 		if (instance.canHaveFocus && initialFocusElement && initialFocusElement.length)
 			initialFocusElement[0].focus();
 
-		var focusWidget = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
-		if (focusWidget)
-			focusWidget.focus();
+		// pass the current instance and get the tabcontrol object if it exist
+		// this will only search in current instance and not in whole docuemnt
+		const tabControlWidget = this.findTabControl(instance);
+
+		let focusWidget ;
+
+		if (tabControlWidget) {
+			// get DOM element of tabControl from current instance
+			focusWidget = instance.content.querySelector('[id="' + tabControlWidget.id + '"]');
+		} else {
+			focusWidget = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
+		}
+
 		if (focusWidget && document.activeElement !== focusWidget) {
 			var firstFocusable = JSDialog.GetFocusableElements(focusWidget);
 			if (firstFocusable && firstFocusable.length)
@@ -401,6 +411,20 @@ L.Control.JSDialog = L.Control.extend({
 		}
 	},
 
+	 findTabControl: function(obj) {
+		if (obj.type === 'tabcontrol') {
+			// this return the main tabcontrol object
+			return obj;
+		}
+		if (obj.children && obj.children.length > 0) {
+			for (let child of obj.children) {
+				const result = this.findTabControl(child); // Recursively search in children
+				if (result) return result; // If found, return the tabcontrol
+			}
+		}
+		return null; // Return null if tabcontrol is not found
+	},
+	
 	/// if you use updatePos - instance param is binded automatically
 	setPosition: function(instance, updatedPos) {
 		var calculated = false;


browser/src/control/Control.JSDialog.js
```diff
@@ -383,15 +383,22 @@ L.Control.JSDialog = L.Control.extend({
...
-		if (focusWidget)
-			focusWidget.focus();
-		if (focusWidget && document.activeElement !== focusWidget) {
-			var firstFocusable = JSDialog.GetFocusableElements(focusWidget);
-			if (firstFocusable && firstFocusable.length)
-				firstFocusable[0].focus();
-			else
-				console.error('cannot get focus for widget: "' + instance.init_focus_id + '"');
+		// Check if the element with id 'tabcontrol' exists in the DOM
+		const isTabControlWidget = document.getElementById('tabcontrol') ? true : false;
```
eszkadev, line 387: Please do not use hardcoded ids of elements.
This can find any tabcontrol, not in the current dialog too.
If we want, we can scan the tree of widgets (JSON) for widget with `type: tabcontrol`

browser/src/control/Control.JSDialog.js
```diff
@@ -383,15 +383,22 @@ L.Control.JSDialog = L.Control.extend({
...
+		if (isTabControlWidget) {
+			// Find the first focusable tab by its id 'tabcontrol-1'
+			const firstFocusableTab = document.getElementById('tabcontrol-1');
+			firstFocusableTab.focus();
+		}
+		else {
+			const focusWidget = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
```
eszkadev, line 394: Here we have explicitly defined element which should be focused as wanted by the dialog designer.
I think it should be always prefered option.

What is the example of dialog/widget it didn't work?

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2949,7 +2949,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		switch (data.action_type) {
 		case 'grab_focus':
-			control.focus();
+			// We will ignore focus on a specific element of the widget if the widget is tabControl.
+			// Focus should be on Tab and not on any element inside dialog if tabcontrols are present
+			if (!document.getElementById('tabcontrol'))
```
eszkadev, line 2954: Please do not use hardcoded ids, also this looks for any element in the whole webpage, not in the current dialog's container.

eszkadev:
What is the example of wrong behavior and in which dialog? I think we shouldn't modify the core's dialog behavior. Only make sure it is exactly the same.

browser/src/control/Control.JSDialog.js
Darshan-upadhyay1110: Menu "References" ( Dialog "Table of Contents")

Darshan-upadhyay1110:
> What is the example of wrong behavior and in which dialog? I think we shouldn't modify the core's dialog behavior. Only make sure it is exactly the same.

Menu "References" ( Dialog "Table of Contents")

Focus should be on `Tab` and not on the first element of the dialog. Arrow keys won't work if we do focus on first element of dialogs which have tabs.

Darshan-upadhyay1110:
> I think we shouldn't modify the core's dialog behavior. Only make sure it is exactly the same

I think for some exception we need to change behavior here, otherwise the accessibility part will be broken for dialog. User can not move around tabs at intended.

browser/src/control/Control.JSDialogBuilder.js
Darshan-upadhyay1110: Yes i think i can improve this here to only search for `tabcontrol` for open dialogs and not in entire document ?

eszkadev:
> > What is the example of wrong behavior and in which dialog? I think we shouldn't modify the core's dialog behavior. Only make sure it is exactly the same.
> 
> Menu "References" ( Dialog "Table of Contents")
> 
> Focus should be on `Tab` and not on the first element of the dialog. Arrow keys won't work if we do focus on first element of dialogs which have tabs.

Thanks. I see. As I said we should keep behavior consistent  online vs core. I testes that dialog and in LO it opens with focus on a tab control. That means: we failed to focus the right element in online.

Could you check if:
- in the JSON we receive correct init_focus_id ?
- if not, the dialog has default focus widget set in the .ui in the core?
- if we don't have defined default focus in JSON, probably we should take into account tabs as focusable widgets, so our logic will fint them as "first widget in the dialog"

Darshan-upadhyay1110:
> Could you check if:
> 
> * in the JSON we receive correct init_focus_id ?

it's =>  "title" and element will be div#title.ui-edit-container.jsdialog.ui-grid-cell
So basically focus widget is inside first tab and not the `tab` itself

Darshan-upadhyay1110:
This is the patch that is sending info about first focusable element inside dialog ( maybe here we are not considering if tabs are present or not ) https://gerrit.libreoffice.org/c/core/+/126565

> * if we don't have defined default focus in JSON, probably we should take into account tabs as focusable widgets, so our logic will fint them as "first widget in the dialog"

@eszkadev even if there is defined default focus, can we make some change in online to see if tabs are there the force to focus on `Selected tab` ? it is a very basic requirement for accessibility to work (in any case we consider)

This can happen in future is someone adds a new UI file in CORE and messed up with init_focus_id over there, which ends up in online to break this focus thing again...

eszkadev:
> This is the patch that is sending info about first focusable element inside dialog ( maybe here we are not considering if tabs are present or not ) https://gerrit.libreoffice.org/c/core/+/126565
> 
> > * if we don't have defined default focus in JSON, probably we should take into account tabs as focusable widgets, so our logic will fint them as "first widget in the dialog"
> 
> @eszkadev even if there is defined default focus, can we make some change in online to see if tabs are there the force to focus on `Selected tab` ? it is a very basic requirement for accessibility to work (in any case we consider)
> 
> This can happen in future if someone adds a new UI file in CORE and messed up with init_focus_id over there, which ends up in online to break this focus thing again...

ok, then we can do it, but please do not use hardcoded id's. Please search in JSON for widget type (tab control), get it's id and use.

Darshan-upadhyay1110:
> > This is the patch that is sending info about first focusable element inside dialog ( maybe here we are not considering if tabs are present or not ) https://gerrit.libreoffice.org/c/core/+/126565
> > > * if we don't have defined default focus in JSON, probably we should take into account tabs as focusable widgets, so our logic will fint them as "first widget in the dialog"
> > 
> > 
> > @eszkadev even if there is defined default focus, can we make some change in online to see if tabs are there the force to focus on `Selected tab` ? it is a very basic requirement for accessibility to work (in any case we consider)
> > This can happen in future if someone adds a new UI file in CORE and messed up with init_focus_id over there, which ends up in online to break this focus thing again...
> 
> ok, then we can do it, but please do not use hardcoded id's. Please search in JSON for widget type (tab control), get it's id and use.

@eszkadev I have a question about identifying the type of widget at the current position. I have all the necessary information in the `instance` variable, but to find the `tabcontrol`, I need to go deep into the child arrays to check if it's present (reference: https://github.com/CollaboraOnline/online/pull/10109/files#diff-6198f19b22f016e2aec20db3137363605a11450036af7386d590555e2b4ab7d5R387).

Is there any existing function or method to handle this kind of scenario, or would I need to implement a new recursive method to find the tab control?

Darshan-upadhyay1110:
@eszkadev I’m thinking I can use `instance.content`, which gives me only the HTML of the opened JS Dialog. I can check if any `tabcontrol` is present, and based on that, I'll decide whether to perform the focus action.

browser/src/control/Control.JSDialogBuilder.js
Darshan-upadhyay1110: @eszkadev in this type of focus function or action handler function does not have proper widget type information. So what can be other way to see if the opened jsDialog on screen have the tab controls ?

I understand the use of hard-coded id is not ok but to fix this focus issue we need to make some exception here maybe i am wrong please guide me here

browser/src/control/Control.JSDialog.js
```diff
@@ -383,15 +383,24 @@ L.Control.JSDialog = L.Control.extend({
...
-		if (focusWidget)
-			focusWidget.focus();
-		if (focusWidget && document.activeElement !== focusWidget) {
-			var firstFocusable = JSDialog.GetFocusableElements(focusWidget);
-			if (firstFocusable && firstFocusable.length)
-				firstFocusable[0].focus();
-			else
-				console.error('cannot get focus for widget: "' + instance.init_focus_id + '"');
+		const instanceContent = instance.content; // instance.content holds the HTML of the current dialog
+		const tabControl = instanceContent.querySelector('#tabcontrol');
```
eszkadev, line 387: This still uses ID :) please check the JSON, 'type' field. This is misleading a bit instance keeps at the same time JSON and DOM elements here after some rework.
This is important as not always tabcontrol will have such id. Please look for children and traverse the tree of widgets for that type of widget.

browser/src/control/Control.JSDialogBuilder.js
eszkadev: I think we can manage to do better solution - not prevent `grab_focus` anytime we have `tabcontrol` element in the whole document.

eg. better would be to check `control` 's node classes, if they contain class which is set for the root container in the tabcontrol widget.

Darshan-upadhyay1110: >eg. better would be to check control 's node classes, if they contain class which is set for the root container in the tabcontrol widget.

@eszkadev i did not get this point 🤔

eszkadev: `control.classList.contains('ui-some-widget-class')`

Darshan-upadhyay1110:
cc: https://gerrit.libreoffice.org/c/core/+/174423

This patch is for external `grab_focsu` in coming for active page which is not ideal.

CC: @eszkadev

browser/src/control/Control.JSDialog.js
```diff
@@ -401,6 +405,35 @@ L.Control.JSDialog = L.Control.extend({
...
+		return null; // Return null if tabcontrol is not found
+	},
+
+	 focusOnTabControl: function(dialogObject) {
+
+		const tabControl = this.findTabControl(dialogObject);
+	
+		if (tabControl && tabControl.tabs && tabControl.tabs.length > 0) {
+			// Construct the id using 'type' + tabs[0].id
+			const targetId = tabControl.id + '-' + tabControl.tabs[0].id;
```
Darshan-upadhyay1110, line 427: This is is generated as same method we do while creating any tabcontrol inside a tab dialog

eszkadev: This is fragile - if we will change that, eg. due to some a11y rework, we will lost this functionality.
So if we would want this approach it should be done using shared function in both cases.
This is why I don't want using hardcoded ids, it can be easily broken. Also in jsdialogs we sometimes rename elements using some suffix with number if we have duplicated nodes (what can happen in some specific situations - and we don't test all possibledialogs with any other one :) )

Also - I would prefere other method: use helper which takes container and returns first focusable element:
this one https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Util.FocusCycle.js#L117

This makes our code very unified. We find for focusable elements in the same way everywhere.

eszkadev (approved):
looks nice and clean now, thanks @Darshan-upadhyay1110 :)

PR #10110, opened 2024-09-24
In context menu, separators don't have a "text" property.

gokaysatir:
In "_createContextMenuStructure" function, we are checkin the text against certain texts.

We first need to check if there is a "text" property.

To reproduce the issue, one can use the right click menu for spell-checked and redlined text portions.

Change-Id: Ic9a919e453886f752182a0aedc84e46597c13f2e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
thank you for a fix

PR #10112, opened 2024-09-24
slideshow: use clientWidth/clientHeight for the mouse clicks

hfiguiere:
Change-Id: I764fc569d6a56d13c33d4817a0870d221966d997

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):

eszkadev:
@hfiguiere would be good to add some short commit message which will describe what is fixed by the change (like when we do X then Y happens, this fixes it so ... ). We will forget in the long run :)

hfiguiere:
Updated the commit message

PR #10113, opened 2024-09-24
Private/timar/aboutbox

timar:
The idea was to simplify and fix code around the debug info strings in About box, while keeping the feature of sending this info to rating server.


hcvcastro (approved):
Thanks, it looks Ok

PR #10114, opened 2024-09-24
Private/codewithvk/webgl canvas load layer

codewithvk:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mcecchetti (commented):
Code looks good. To try it locally.

mcecchetti:
While testing it I found the following issues:
- shapes with an alpha component are rendered in a wrong color: see the half transparent light blue rectangle in the sample presentation
- shape animations rendering/performance is a lot worse

I reported more details on the related task page

browser/src/slideshow/LayerRenderer.ts
```diff
@@ -103,7 +103,8 @@ class LayerRendererGl implements LayerRenderer {
 		gl.bufferData(gl.ARRAY_BUFFER, texCoords, gl.STATIC_DRAW);
 
 		gl.enable(gl.BLEND);
-		gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
+		gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
+		gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, true);
 	}
```
mcecchetti, line 108: Nice fix. It improved a bit performance, too.

mcecchetti:
I implemented a patch (https://github.com/CollaboraOnline/online/pull/10153) that renders each animation frame directly in the gl canvas provided by `RendererLayerGl` instead of drawing it earlier in a 2d canvas and then pass the layer as a image bitmap to `RendererLayerGl`.
Anyway performances are still less smooth than without the `RendererLayer` patch.
Without `RendererLayer` patch, `Transition2d` (and it's sub-classes) which I used for playing transition filters for a shape are not affected by this problem. So, I wonder if the problem is in `RendererLayerGl`.

codewithvk:
> I implemented a patch (#10153) that renders each animation frame directly in the gl canvas provided by `RendererLayerGl` instead of drawing it earlier in a 2d canvas and then pass the layer as a image bitmap to `RendererLayerGl`. Anyway performances are still less smooth than without the `RendererLayer` patch. Without `RendererLayer` patch, `Transition2d` (and it's sub-classes) which I used for playing transition filters for a shape are not affected by this problem. So, I wonder if the problem is in `RendererLayerGl`.

As we are using offscreen GL canvas performance often depends on the device you are using. TBH it's a bit faster for my device not sure why it's slow for you. Maybe we can discuss and stick to 2d canvas if we want.

mcecchetti (approved):
By the last commit performances are improved. It's fine to merge for me.

PR #10116, opened 2024-09-25
cool#10106 browser, context menu: fix label of the paste special menu item

vmiklos:
Open Writer, copy some text, unselect, right click: a 'more options'
menu item is there for paste special, but the desktop has a clear 'paste
special' menu item instead.

Commit ab896c1c78ed220044ca7fc10b6746d78cdb529a (clipboard: add Paste
Special also in context menu, 2022-04-05) collapsed the desktop context
menu to only have 'paste special' as a menu item instead of a submenu,
which is fine -- but the 'item.text' set is overwritten later based on
the label of the uno command. That 'more options' label is fine, but
only in the context of the 'paste special' context menu, not standalone.

Fix the problem by adding PasteSpecial to the list of UNO commands where
we don't overwrite the label based on the command name.

Change-Id: I27996c48710b43fd2bfaa227be5c632e4d22eb27

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -297,7 +297,9 @@ L.Control.ContextMenu = L.Control.extend({
 				if (commandName == 'None' && !item.text)
 					continue;
 
-				if (hasParam || commandName === 'None' || commandName === 'FontDialogForParagraph' || commandName === 'Delete') {
+				if (hasParam || commandName === 'None' || commandName === 'FontDialogForParagraph' || commandName === 'Delete' || commandName == 'PasteSpecial') {
+					// These commands have a custom item.text, don't overwrite
+					// that with a label based on 'item.command'.
 					itemName = window.removeAccessKey(item.text);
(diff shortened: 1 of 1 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_multi_user/3212/console fails in:

> sheet_operations_spec.js

But `make -C cypress_test check-multi spec=calc/sheet_operations_spec.js` passes locally, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3207/console fails in:

> calc/cell_appearance_spec.js

But `make -C cypress_test check-desktop spec=calc/cell_appearance_spec.js` passes locally, retry.

vmiklos:

The original label made sense for the desktop (with its context), but we rendered the label as-is, without its context, so it was confusing. We even had code to tweak the label, just that got overwritten later.

caolanm (approved):
seems fine to me

PR #10117, opened 2024-09-25
Handle limited open Connections due to keepalive connections (scratch pull draft pre-review, 2nd revision)

sgothel:
* Resolves: # 
* Target version: master 

#### Changes to original PR
- net::Config -> net::Defaults, without any xml config file side-effects
  - only allow net::Defaults to be changed in unit tests via dedicated callback
- removed Socket age criteria via Socket::creationTime and net::defaults::MaxDuration
  - Socket::creationTime still relevant for minimum throughput criteria
  - dropped the net::defaults::MaxDuration 
- max-connections
  - removed under- overflow checks
  - using SC atomic ops (not just relaxed)
- enhanced API doc of Util::TimeAverage (clarification)
- unit tests in one place / commit
- more unification/squashing to single semantic commits
- more sensible order of commits
- removed dead code
- dropped noexcept patch

Further:
- cleaned up the commit messages
- removed the COOLWSD.cpp default config reordering
- reordered the 'Unify socket stats', as it must be upfront the actual connection count limitation
  and before moved _lastSeenHTTPHeader
- fixed another intermediate breakage in the commit chain
- bisectable proof (but no unit test runs ofc) .. via
  - passed `git rebase -i --exec "make -j 20"` for all commits
   
### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1344,6 +1344,36 @@ LocalServerSocket::~LocalServerSocket()
 #  define LOG_CHUNK(X)
 #endif
 
+#endif // !MOBILEAPP
+
+std::string StreamSocket::toString(WSState t)
+{
+    if( WSState::WS == t )
+    {
+        return "WS";
+    }
+    return "HTTP";
+}
+
+std::ostream& StreamSocket::stream(std::ostream& os) const
+{
+    os << "StreamSocket[#" << getFD()
+       << ", " << toString(_wsState)
+       << ", " << Socket::toString(type())
+       << " @ ";
+    if (Type::IPv6 == type())
+    {
+        os << "[" << clientAddress() << "]:" << clientPort();
+    }
+    else
+    {
+        os << clientAddress() << ":" << clientPort();
+    }
+    return os << "]";
+}
+
+#if !MOBILEAPP
+
 bool StreamSocket::parseHeader(const char *clientName,
                                Poco::MemoryInputStream &message,
                                Poco::Net::HTTPRequest &request,
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1409,6 +1409,7 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
 bool StreamSocket::parseHeader(const char *clientName,
                                Poco::MemoryInputStream &message,
                                Poco::Net::HTTPRequest &request,
+                               std::chrono::steady_clock::time_point &lastHTTPHeader,
                                MessageMap& map)
 {
     assert(map._headerSize == 0 && map._messageSize == 0);
@@ -1417,7 +1418,7 @@ bool StreamSocket::parseHeader(const char *clientName,
         std::chrono::duration_cast<std::chrono::milliseconds>(_httpTimeout);
 
     std::chrono::steady_clock::time_point now = std::chrono::steady_clock::now();
-    std::chrono::duration<float, std::milli> delayMs = now - _lastSeenHTTPHeader;
+    std::chrono::duration<float, std::milli> delayMs = now - lastHTTPHeader;
 
     // Find the end of the header, if any.
     static const std::string marker("\r\n\r\n");
@@ -1511,7 +1512,7 @@ bool StreamSocket::parseHeader(const char *clientName,
                 if (chunkLen == 0) // we're complete.
                 {
                     map._messageSize = chunkOffset;
-                    _lastSeenHTTPHeader = now;
+                    lastHTTPHeader = now;
                     return true;
                 }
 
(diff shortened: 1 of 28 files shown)


net/Socket.cpp
```diff
@@ -1298,18 +1495,100 @@
...
+    if (c1 || c2 || c3 || c4 )
+    {
+        LOG_WRN("CheckRemoval: Timeout: {c1 " << c1 << ", c2 " << c2 << ", c3 " << c3 << ", c4 " << c4 << "}, "
+            << getStatsString(now) << ", "
+            << *this);
+        if (_socketHandler)
+        {
+            _socketHandler->onDisconnect();
+            if( isOpen() ) {
+                // FIXME: Ensure proper semantics of onDisconnect()
```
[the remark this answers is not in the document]
sgothel: resolved below

net/WebSocketHandler.hpp
```diff
@@ -599,48 +630,63 @@
...
 
 public:
+    /// Sends a native control-frame ping message
     void sendPing(std::chrono::steady_clock::time_point now,
                   const std::shared_ptr<StreamSocket>& socket)
     {
-//        assert(!_isClient);
+        if (_isClient)
+            LOG_DBG("Clients should not send pings, only servers");
+//        assert(_isClient);
```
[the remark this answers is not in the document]
sgothel: resolved below

net/WebSocketHandler.hpp
```diff
@@ -599,48 +630,63 @@
...
     }
 
+    /// Sends a native control-frame pong message
     void sendPong(std::chrono::steady_clock::time_point now,
                   const char* data, const size_t len,
                   const std::shared_ptr<StreamSocket>& socket)
     {
+        if (!_isClient)
+            LOG_WRN("Servers should not send pongs, only clients");
+//        assert(!_isClient);
```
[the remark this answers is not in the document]
sgothel: resolved below

sgothel:
now (2nd CI run) wopi-save-on-exits fails sporadic again (passed 1st time, known instable test ...)

sgothel:
Reviewed my changes set, added missing setup in COOLWSD.cpp for maxConnection commit
> net::Defaults::get().MaxConnections = std::max<size_t>(3, MAX_CONNECTIONS);

sgothel:
> Reviewed my changes set, added missing setup in COOLWSD.cpp for maxConnection commit
> 
> > net::Defaults::get().MaxConnections = std::max<size_t>(3, MAX_CONNECTIONS);

Added two more typos fixes: 1 comment (Socket.cpp) + 1 LOG message (WebSocketHandler ping/pong)

net/Socket.hpp
```diff
@@ -398,12 +412,19 @@ class Socket
         , _open(_fd >= 0)
         , _creationTime(creationTime)
         , _lastSeenTime(_creationTime)
+        , _bytesSent(0)
```
caolanm, line 415: It seems that the change to move _bytesSent and _bytesRcvd into the base class is effectively only so they are available for Socket::streamStats? So I wonder if that's something we need to do?

wsd/COOLWSD.cpp
```diff
@@ -2081,12 +2082,12 @@ void COOLWSD::innerInitialize(Application& self)
...
         { "net.connection_timeout_secs", "30" },
+        { "net.content_security_policy", "" },
+        { "net.frame_ancestors", "" },
         { "net.listen", "any" },
         { "net.proto", "all" },
-        { "net.service_root", "" },
         { "net.proxy_prefix", "false" },
-        { "net.content_security_policy", "" },
-        { "net.frame_ancestors", "" },
+        { "net.service_root", "" },
```
caolanm, line 2090: seeing as these end up not changed, maybe drop this from this pr and put into a separate one (even though ordering them alphabetically looks a good idea)

caolanm (commented):
Probably need to then update the commit messages so they just relate to the code as it is and not mention previous versions that existed during review because that context won't really apply when merged. I feel changes from earlier drafts can go in the review itself, if it's something that needs to be pointed out.

net/Socket.hpp
sgothel: For stats and offering it as an criteria for checkTimeout / checkRemoval (used in the latter).
When I unified it in the Socket, idea was to push down general properties (like open) etc.

wsd/COOLWSD.cpp
sgothel: yeah, can drop it. reordered when added the other items - now .. either way is OK.

sgothel:
> Probably need to then update the commit messages so they just relate to the code as it is and not mention previous versions that existed during review because that context won't really apply when merged. I feel changes from earlier drafts can go in the review itself, if it's something that needs to be pointed out.

yup, will push another one tmr morning.

net/Socket.hpp
sgothel: reordered the specific commit .. resulting in visible usage and compile clean commits

sgothel:
moved to orig PR

PR #10118, opened 2024-09-25
fix anyInputCallback

caolanm:
fix this low hanging issue anyway

Change-Id: Ib105339bc94e767b1bcfe87a06e5edd7fc932ec4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2898,7 +2898,7 @@ bool anyInputCallback(void* data)
 
     // Have no pending callbacks and the tile queue is also empty, report that we have no
     // pending input events.
-    return true;
+    return false;
 }
 
 /// Called by LOK main-loop


caolanm:
Though I don't find a case of where there are tiles outside the visible area getting requested when layout queries this

mmeeks:
@timar rather an important back-port / ship stopper for code with this functionality but without this fix I think =)

mmeeks:
Hmm - isn't the fundamental problem here that if we do background / idle layout we then (I guess) send lots of document re-size messages (?) and this then causes each of the JS client views to refresh / request more tiles (or does it ?) - perhaps I'm missing something there.

kit/Kit.cpp
```diff
@@ -2893,12 +2899,20 @@ bool anyInputCallback(void* data)
...
     {
+        bool tilesInsideVisibleArea = queue->anyOf([&document](const TileDesc& desc) {
+            return document->isTileDescInsideVisibleArea(desc);
+        });
+        // Pending tiles requests inside visible area -> allow writer layout (querying
+        // if it should quit to allow alternative work to run) to continue.
+        if (tilesInsideVisibleArea)
+            return false;
+        // Quit writer layout to serve tiles.
         return true;
```
vmiklos, line 2910: The logic with this change would be: if we have tiles to be rendered in the queue, then we tell Writer layout to continue, otherwise we interrupt. Don't we want to do the opposite? As in: if we have tiles to be rendered in the queue and those would update the visible area, then interrupt (so we can render those tiles), but otherwise go on (instead of the old behavior of always interrupting).

vmiklos:
> Hmm - isn't the fundamental problem here that if we do background / idle layout we then (I guess) send lots of document re-size messages (?) and this then causes each of the JS client views to refresh / request more tiles (or does it ?) - perhaps I'm missing something there.

Reading https://github.com/libreoffice/core/blob/4418b3e453926a28f96458177273aa151f08e616/sw/source/core/view/viewsh.cxx#L1143-L1156 , it seems to me that we're safe there: if layout calculation is in progress, we already try to not emit LOK callbacks about doc size changes.

vmiklos:
I believe the reason for that list was to try to avoid paint -> invalidation -> paint loops, so adding `LOK_CALLBACK_DOCUMENT_SIZE_CHANGED` there + the first patch from this PR sounds like a better solution than the 2 patches from this PR (at least in its current form).

kit/Kit.cpp
caolanm: yeah, but it seems that on rendering that (pending because interrupted?) layout work gets done and resize due to that don't get seen at the browser side

kit/Kit.cpp
```diff
@@ -2893,12 +2899,20 @@ bool anyInputCallback(void* data)
 
     if (queue->getTileQueueSize() > 0)
     {
+        bool tilesInsideVisibleArea = queue->anyOf([&document](const TileDesc& desc) {
```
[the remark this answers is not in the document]
caolanm: doesn't matter now, its the to-be-dropped 2nd patch

caolanm:
Alright, lets chop this back down to the initial commit then. That's straightforward and go with the code side alt patch for the blank page problem: https://gerrit.libreoffice.org/c/core/+/173936

vmiklos (approved):
Looks good to me now, thanks.

PR #10119, opened 2024-09-25
slideshow: repeated presentation with initial transition

eszkadev:
If we have presentation which has enabled repeated slideshow\with timer, we see at the end counter and presentation starts from the first slide. This was not working correctly as we didn't see first transition when we repeated first slide after seeing counter. Let's use better function which does initial setup too.


mcecchetti (approved):
That's fine.

PR #10120, opened 2024-09-25
Refactor some elements in about dialog box.

gokaysatir:
We want to hide some elements outside of debugging mode, which is activated by triple click on about dialog. That requires some elements to be created hidden by default. This commit removes the "style="display:none"" code from HTML. Because that way of defaulting to hidden raises CSP issues. Instead, we hide them right before we show the about dialog.

Also, there were duplicate IDs of elements in about dialog. Since we clone a dialog which already exists, all the IDs in that dialog are duplicated. This commit adds "-cloned" suffix to 2 elements which have duplicate IDs. We probably need to check also other duplicates.

Change-Id: Idef8e842b2b2fd90a1639db63dd490d7718ff1c9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -201,12 +201,12 @@ m4_ifelse(MOBILEAPP,[true],
               <div class="spacer"></div>
               <div id="lokit-version-label"></div>
               <div class="about-dialog-info-div"><div id="lokit-version" dir="ltr"></div></div>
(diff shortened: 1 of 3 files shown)


timar:
@gokaysatir I tested the patch, and the toggle of visibility works, however, the first time, when the debug mode is off, the debug labels of the About box are visible. 
image

gokaysatir:
> @gokaysatir I tested the patch, and the toggle of visibility works, however, the first time, when the debug mode is off, the debug labels of the About box are visible. image

Aha, let me check, i also tested and it worked, which is weird. Will fix it ASAP.

gokaysatir:
@timar you are right, thanks for checking :) Fixed.

timar (approved):
Nice improvement, thanks!

PR #10121, opened 2024-09-25
slideshow: keep connection when presenting

eszkadev:
don't enter idle mode if presenting, that diconnects us from the websocket and causes to stop the presentation in current point


mcecchetti (approved):
Looks good to me.

PR #10122, opened 2024-09-25
slideshow: fix error on dummy slides click

eszkadev:
after interaction feature was done we expect to always have it's data. but we also have dummy slides without any data for showing the text on blakc brackground (click to exit, etc)

### DIFF
--- a/browser/src/slideshow/engine/SlideShowNavigator.ts
+++ b/browser/src/slideshow/engine/SlideShowNavigator.ts
@@ -294,7 +294,11 @@ class SlideShowNavigator {
 	private clickHandler(aEvent: MouseEvent) {
 		if (aEvent.button === 0) {
 			const slideInfo = this.theMetaPres.getSlideInfoByIndex(this.currentSlide);
-			if (slideInfo.interactions.length == 0) {
+			if (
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/engine/SlideShowNavigator.ts
```diff
@@ -294,7 +294,11 @@ class SlideShowNavigator {
 	private clickHandler(aEvent: MouseEvent) {
 		if (aEvent.button === 0) {
 			const slideInfo = this.theMetaPres.getSlideInfoByIndex(this.currentSlide);
-			if (slideInfo.interactions.length == 0) {
+			if (
+				!slideInfo ||
+				!slideInfo.interactions ||
+				slideInfo.interactions.length == 0
+			) {
```
hfiguiere, line 301: LGTM but can't we use `slideInfo?.interactions?.length == 0`

hfiguiere (approved):

browser/src/slideshow/engine/SlideShowNavigator.ts
eszkadev: I think it is not enabled in the linter...

PR #10123, opened 2024-09-25
slideshow: avoid reccurent calls with wrong number

eszkadev:
global.js:552 Uncaught RangeError: Maximum call stack size exceeded
    at Arguments.slice (<anonymous>)
    at Object.logWithCool [as log] (global.js:552:40)
    at DebugPrinter.print (tools.ts:181:23)
    at SlideShowNavigator.displaySlide (SlideShowNavigator.ts:189:10)
    at SlideShowNavigator.displaySlide (SlideShowNavigator.ts:224:9)
    at SlideShowNavigator.displaySlide (SlideShowNavigator.ts:224:9)

### DIFF
--- a/browser/src/slideshow/engine/SlideShowNavigator.ts
+++ b/browser/src/slideshow/engine/SlideShowNavigator.ts
@@ -194,7 +194,10 @@ class SlideShowNavigator {
 				', bSkipTransition: ' +
 				bSkipTransition,
 		);
-		if (nNewSlide === undefined || nNewSlide < 0) return;
+		if (nNewSlide === undefined || nNewSlide < 0) {
+			NAVDBG.print('SlideShowNavigator.displaySlide: unexpected nNewSlide');
+			return;
+		}
 		if (nNewSlide >= this.theMetaPres.numberOfSlides) {
 			this.currentSlide = nNewSlide;
 			const force = nNewSlide > this.theMetaPres.numberOfSlides;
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/engine/SlideShowNavigator.ts
```diff
@@ -194,8 +194,11 @@ class SlideShowNavigator {
 				', bSkipTransition: ' +
 				bSkipTransition,
 		);
-		if (nNewSlide === undefined || nNewSlide < 0) return;
-		if (nNewSlide >= this.theMetaPres.numberOfSlides) {
+		if (
```
mcecchetti, line 197: I don't think we should quit if user try to go back from first slide.
Anyhow I've not yet the time to understand what's the root cause of the problem.

mcecchetti: `this.currentSlide === nNewSlide

browser/src/slideshow/engine/SlideShowNavigator.ts
```diff
@@ -194,8 +194,11 @@ class SlideShowNavigator {
...
+			nNewSlide === undefined ||
+			nNewSlide < 0 ||
+			nNewSlide >= this.theMetaPres.numberOfSlides
+		) {
 			this.currentSlide = nNewSlide;
 			const force = nNewSlide > this.theMetaPres.numberOfSlides;
 			if (force) this.quit();
```
mcecchetti, line 204: It's annoying that it's not possible to add a comment to any code line.
Anyhow, I think we should change in `if (!slideAvailable)` (~line 225) `this.currentSlide !== undefined` to `this.currentSlide !== undefined && this.currentSlide !== nNewSlide).
I suspect it can cause an infinite recursion when `this.currentSlide === nNewSlide === 0 and the first slide is hidden or no more available.
Or maybe we could check at `displaySlide` beginning and return immediately in case `this.currentSlide === nNewSlide`.

mcecchetti (approved):
Looks good. Thanks!

PR #10128, opened 2024-09-25
UI: indicate write protect status by snackbar

lpranam:
Change-Id: Ic7616a1d5a839a7dadbac765a4f8a469ffd0e946

* Target version: master 

### Checklist


pedropintosilva (approved):

pedropintosilva:
We are on green week but this seems quite safe as it only affects the read-only files.

Issue #10049, opened 2024-09-14
`coolwsd.xml` defaults do not match built-in defaults


xzfc:
### Describe the Bug

[The comment in `COOLWSD.cpp`](https://github.com/CollaboraOnline/online/blob/cp-24.04.7-2/wsd/COOLWSD.cpp#L2036-L2043) states that defaults from `coolwsd.xml` and built-in defaults from the `coolwsd` C++ sources should be the same. However, they are not. Consequently, when `coolwsd` is started with an empty `coolwsd.xml`, the factual configuration would not be the same as if it was started with the default `coolwsd.xml`. 

This cause problems for systems that generate the `coolwsd.xml` config file on its own (e.g. Nix). We can't generate `coolwsd.xml` from scratch with just a few user-specified options, we have to patch the default `coolwsd.xml` instead. This is inconvenient.

Incomplete list of mismatched options:

| Option                               | `coolwsd.xml`                  | built-in                                                              |
| ------------------------------------ | ------------------------------ | --------------------------------------------------------------------- |
| `logging.level`                      | `@COOLWSD_LOGLEVEL@`/`warning` | `trace`                                                               |
| `num_prespawn_children`              | `4`                            | `1`                                                                   |
| `per_document.cleanup[@enable]`      | `true`                         | `false`                                                               |
| `per_view.out_of_focus_timeout_secs` | `300`                          | `120` or `60`                                                         |
| `ssl.termination`                    | `false`                        | `true`                                                                |
| `ssl.cert_file_path`                 | `/etc/coolwsd/cert.pem`        | `COOLWSD_CONFIGDIR "/cert.pem"`/<br>`/usr/local/etc/coolwsd/cert.pem` |
| `storage.ssl.as_scheme`              | `true`                         | `false`                                                               |
| And others                           |                                |                                                                       |

### Steps to Reproduce

1. Create `empty.xml` with this contents: `<config></config>`.
2. First run: run `coolwsd` with the default `coolwsd.xml`.
3. Second run: run `coolwsd --config-file=empty.xml`.

### Expected Behavior

Both runs should behave exactly the same.

### Actual Behavior

They are not. Presumably, the most visible difference is the logging level (see the table above) and paths to SSL certs.

timar:
[...]

PR #10129, opened 2024-09-25
Private/timar/configdefaults

timar:
* Resolves: #10049 
* Target version: master 

### Summary
Synced default values of configuration settings between code and coolwsd.xml.


configure.ac
```diff
@@ -1505,7 +1505,10 @@ AS_IF([test `uname -s` = "FreeBSD"],
 AS_IF([test -n "$with_compiler_plugins"],
       [CPPFLAGS="$CPPFLAGS -Xclang -load -Xclang ${with_compiler_plugins}/compilerplugins/obj/plugin.so -Xclang -add-plugin -Xclang loplugin -Xclang -plugin-arg-loplugin -Xclang --cool-base-path=\${abs_top_srcdir}"])
 
-COOLWSD_CONFIGDIR=${sysconfdir}/${PACKAGE}
+if test "x${prefix}" = "xNONE"; then
+    prefix=/usr/local
+fi
+eval COOLWSD_CONFIGDIR="${sysconfdir}/${PACKAGE}"
```
vmiklos, line 1511: Nit: why do you need eval here? I would expect sysconfdir is /etc and PACKAGE is coolwsd at this stage.

vmiklos (approved):
- OK, I see that coolwsd.xml already uses @COOLWSD_LOGLEVEL@, so this sets the built-in default to the same value, makes sense to me.
- Same for @NUM_PRESPAWN_CHILDREN@, which is now used at more places, OK.
- cleanup: again this was true at a config level, the rest is now adjusted to match, good.
- `out_of_focus_timeout_secs`: same here, good.
- SSL termination: yes, I remember that story, good to have false everywhere.
- configdir: makes sense, hardcoding /etc was not nice.

All in all: looks good to me, thanks.

configure.ac
timar: Without eval the value was literal `${prefix}/etc/coolwsd` when I did not specify sysconfdir separately. The reason is that these paths are evaluated later. Autoconf sets it so, that prefix etc. can be changed even as late as `make install`.

vmiklos: Ah I see, thanks.

PR #10130, opened 2024-09-26
slideshow: stop timer if closing

eszkadev:
If we close "pause screen" with counter before it ends. Timer was still active causing an error in the browser console.


mcecchetti (approved):
Looks good. Thanks!

PR #10131, opened 2024-09-26
Fix keyboard shortcuts not working after row/column selection in Calc

Darshan-upadhyay1110:
- After selecting a row or column, the focus shifts from the 'map' to the 'document'
- Keyboard shortcuts are intended to work specifically within the 'map' area
- This patch ensures that focus returns to the 'map' after selection, allowing keyboard shortcuts to function correctly

This fix resolves the issue where shortcuts would stop working after a row/column selection.

To check this fix :

- Go to ods file
- click on row/column header
- Press CTRL + B 

Without this patch, nothing will happen. Expected behavior should be to make all text bold

Change-Id: Ic5265a7d3eced20e4f005aa1ce50acbc87946b4f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
thanks, looks reasonable :)

Issue #2210, opened 2021-04-29
Styles preview box width depand on window size


Andreas-Kainz:
It would be cool if the width of the previews can be flexible so that you can see 2 - 20 previews.

![Screenshot_20210429_205738](https://user-images.githubusercontent.com/8517736/116603783-9dbd8d80-a92d-11eb-8fb3-79b10e9399b7.png)

thebearon:
Sounds reasonable, probably something to consider in desktop LO as well.

Andreas-Kainz:
I think at COOL this can be done with .css on LO I don't know.

Andreas-Kainz:
@pedropintosilva do you have an idea how to make an flexbox that flexible? Or maybe @eszkadev you have an idea.

pedropintosilva:
yes, similar as I did in some mobile wizard panes (shapes, maybe?) or I think even on the fontwork dialog (you can play with width of #FontworkGalleryDialog in the jsdialog.css and you will see that the grid auto adjusts etc)

PR #10132, opened 2024-09-26
Make Styles preview box width flexible

pedropintosilva:
Resolves Styles preview box width depand on window size #2210

No need to set hard coded width. Better to set it to a portion of the
viewport's width. Also make sure the grid is tidier by actually using
a grid and not inline-flex and setting the proper width and height
of each cell.

Note if more buttons are added to the Home tab then that 35vw should
be updated accordingly

Change-Id: Ifd8696caf0d9b49fcb2c6929ea5f23505d7cff42

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -354,11 +354,13 @@ html[data-theme='dark'] .savemodified.unotoolbutton .unobutton img {
 
 #stylesview {
 	height: 64px;
-	width: 530px;
+	width: 35vw;
 	overflow: auto;
-	display: inline-flex;
-	display: -ms-inline-flexbox;
-	flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


pedropintosilva:
2 cypress specs are failing  ✖  writer/table_operation_spec.js  ✖  writer/top_toolbar_spec.js probably because we are relying on a x amount of clicks in the scroll indicator... I need to check those. Probably best if for those we just scroll horizontally until the scroll indicator is not visible. Or maybe safer to just adjust the number of clicks according to the cypress resolution settings (that way it will not impact any next steps in the test)

pedropintosilva:
so this is a bit sad. I just had time to re-check this... and these failed tests are not related to this PR. It fails in the trying to check an attribute in the #copy-paster-container

Issue #9992, opened 2024-09-05
Digitally sign documents


vmiklos:
### Is your feature request related to a problem?

Once a document is signed digitally, the COOL toolbar has an icon to view these signatures. You can't add or remove a document signature, though.

### Describe the solution you'd like

As a first step, be able to remove document signatures. Later, also be able to add them (sure, that's more complicated).

### Describe alternatives you've considered

An option is to only use COOL for signature verification, and do the signature creation outside COOL with apps like https://apps.nextcloud.com/apps/signlive or https://apps.nextcloud.com/apps/electronicsignatures.

### Additional context

There are lots of unknowns here, but as a first step, let's focus on how the signatures dialog is ready-only: we need a read-write async dialog to start on this.

vmiklos:
For the async dialog conversion, the backtrace of run(), till the async command dispatch:

```
#0  DigitalSignaturesDialog::run() (this=0x74f5200) at /home/vmiklos/git/libreoffice/core/xmlsecurity/source/dialogs/digitalsignaturesdialog.cxx:423
#1  0x00007fffbc31b5c7 in (anonymous namespace)::DocumentDigitalSignatures::ImplViewSignatures(com::sun::star::uno::Reference<com::sun::star::embed::XStorage> const&, com::sun::star::uno::Reference<com::sun::star::io::XStream> const&, DocumentSignatureMode, bool)
    (this=0x76479f0, rxStorage=uno::Reference to (OStorage *) 0x758d788, xSignStream=uno::Reference to (OWriteStream *) 0x7565f78, eMode=DocumentSignatureMode::Content, bReadOnly=false)
    at /home/vmiklos/git/libreoffice/core/xmlsecurity/source/component/documentdigitalsignatures.cxx:449
#2  0x00007fffbc31a80b in (anonymous namespace)::DocumentDigitalSignatures::signDocumentContent(com::sun::star::uno::Reference<com::sun::star::embed::XStorage> const&, com::sun::star::uno::Reference<com::sun::star::io::XStream> const&) (this=0x76479f0, rxStorage=uno::Reference to (OStorage *) 0x758d788, xSignStream=uno::Reference to (OWriteStream *) 0x7565f78)
    at /home/vmiklos/git/libreoffice/core/xmlsecurity/source/component/documentdigitalsignatures.cxx:280
[...]

PR #10135, opened 2024-09-26
cool#9992 doc sign: fix missing enable on the cert chooser description entry

vmiklos:
Open the signatures dialog, add a signature, the certificate chooser
sub-dialog appears. Select a certificate, that enables e.g. the sign
button, but not the description input field for some reason.

Checking the websocket incoming traffic, the message for the working
button and the broken entry is the same:
jsdialog: { "jsontype": "dialog", "action": "action", "id": 10, "data": { "control_id": "viewcert", "action_type": "enable"}}
jsdialog: { "jsontype": "dialog", "action": "action", "id": 10, "data": { "control_id": "description", "action_type": "enable"}}
But the DOM is different:

```html
<button class="ui-pushbutton jsdialog" id="viewcert" accesskey="V" disabled tabindex="0"><u class="access-key">V</u>iew Certificate</button>
<div class="ui-edit-container jsdialog" id="description" tabindex="0"><input class="ui-edit jsdialog" id="description-input" dir="auto" disabled></div>
```

Fix the problem similar to what commit
0d4e4af2cf174aba3435e44cf76aa5aa834ebb20 (jsdialog: find real input on
setText action, 2022-08-24) did for the setText action type: if there is
a child input element, work with that instead.

Probably doing so for the disable case would also make sense, but I this
dialog never disables input fields, so leave that alone for now.

Change-Id: I94a9a3c5fd846be86a5db175ce93ffff136881de

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2975,6 +2975,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			break;
 
 		case 'enable':
+			// E.g. GtkEntry is represented as an <input> inside a <div>, the disabled
+			// attribute is on the <input>, not the <div>.
+			var innerInput = control.querySelector('input');
+			if (innerInput)
+				control = innerInput;
+
 			control.disabled = false;
 			control.removeAttribute('disabled');
 			break;
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1513,8 +1513,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.password === true)
 			edit.type = 'password';
 
-		if (data.enabled === 'false' || data.enabled === false)
+		if (data.enabled === 'false' || data.enabled === false) {
+			container.setAttribute('disabled', 'true');
 			edit.disabled = true;
+		}
 
 		JSDialog.SynchronizeDisabledState(container, [edit]);
 
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -1514,7 +1514,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			edit.type = 'password';
 
 		if (data.enabled === 'false' || data.enabled === false)
-			$(edit).prop('disabled', true);
+			edit.disabled = true;
 
 		JSDialog.SynchronizeDisabledState(container, [edit]);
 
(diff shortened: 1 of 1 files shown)


vmiklos:

This fixes the problem that the cert chooser dialog's cert list enables the sign button on cert select, but the comment of the signature is still disabled after cert select.

CC @eszkadev

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2975,6 +2975,12 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			break;
 
 		case 'enable':
+			// E.g. GtkEntry is represented as an <input> inside a <div>, the disabled
+			// attribute is on the <input>, not the <div>.
+			var innerInput = control.querySelector('input');
```
eszkadev, line 2980: ah, we have better solution for that -> the widget itself doesn't update inner input on root change.
Let me push update here so you can test

eszkadev:
added update, @vmiklos please check if it fixes the issue (I was testing using different example in chart editing sidebar -> title field), if yes -> we can squash probably

vmiklos:
@eszkadev sadly your commit breaks the use-case I have. Let me get everything in for the core side then you can see this in action locally. I'll ping you when I have that.

vmiklos:
@eszkadev OK, finally I hope you can easily reproduce this on your end. Pull on the core side, then cherry-pick https://gerrit.libreoffice.org/c/core/+/174308. That should give you a core build where this certificate chooser dialog is a jsdialog. Then pick the Online side mock patch from https://github.com/CollaboraOnline/online/issues/9992#issuecomment-2367363290 .

Then, steps to reproduce this:

1) Open a hello world document, e.g. Writer.
2) Click on the status bar, where the sign icon would be visible to open the sign dialog.
3) Click "Add"
4) Select the only cert in the cert chooser dialog.
5) The sign button gets enabled, but the description GtkEntry is still disabled.

Please let me know if you need anything else to reproduce this. Thanks!

vmiklos:
@eszkadev I think I understand the problem. So you had a state sync logic from the container to the inner input element, but disabled was only set on the inner input. I pushed a 3rd commit to this PR to also set disabled on the container, that fixes the problem for me. Is the state after the 3rd commit better than the original commit, you think? Thanks.

If so, then you don't have to reproduce this locally, I think. :-)

eszkadev:
> I pushed a 3rd commit to this PR to also set disabled on the container, that fixes the problem for me. Is the state after the 3rd commit better than the original commit, you think? 

exactly, that was missing :) thanks

eszkadev (approved):

vmiklos:
Hm, re CI: https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3326/console fails in writer/table_operation_spec.js, and that 7 failures is something I also see locally, without the changes in this PR. I wonder if a core co-24.04 changes went it that breaks things. https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/ is all red at least.

vmiklos:
FWIW reverting core.git 6a2071cafae41afdfde56b07817d8674482431f6 locally makes

```
make -C cypress_test check-desktop spec=writer/table_operation_spec.js
```

pass, while 7 out of 17 cases were failing in that suite, matching the Jenkins log.

CC @GulsahKose 

Do you think it would be OK to revert the commit for now, make sure it doesn't cause new cypress failures and only get it in again? Thanks.

PR #10136, opened 2024-09-26
Added missing buttons for consistency

Rash419:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -1832,6 +1832,36 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				],
 				'vertical': 'true'
 			},
+			{
+				'type': 'container',
+				'children': [
+					{
+						'type': 'toolbox',
+						'children': [
+							{
+								'id': 'data-add-name',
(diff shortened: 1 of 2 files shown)


eszkadev (approved):
looks good

browser/src/control/Control.NotebookbarCalc.js
```diff
@@ -1832,6 +1832,36 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
...
+				'children': [
+					{
+						'type': 'toolbox',
+						'children': [
+							{
+								'id': 'data-add-name',
+								'type': 'toolitem',
+								'text': _UNO('.uno:AddName', 'spreadsheet'),
+								'command': '.uno:AddName',
+								'accessibility': { focusBack: true,	combination: 'DD', de: null }
```
eszkadev, line 1846: @gokaysatir do we need to always put `de: null` ?

gokaysatir: @eszkadev no we shouldn't need that. I had put them as a placeholder in the first implementation. Because many of them were subject to change.

PR #10137, opened 2024-09-26
make this thread local

caolanm:
Change-Id: Ie0d9e547dc9b4803ea95848c76302601a996c53a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -174,7 +174,7 @@ public:
 
 HostEntry resolveDNS(const std::string& addressToCheck)
 {
-    static DNSResolver resolver;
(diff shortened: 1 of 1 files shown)


caolanm:
This static now looks unsafe, so a thread_local looks a better idea. With a follow up for later of https://github.com/CollaboraOnline/online/pull/10138 to drop some of the remaining sync dns that use it.

vmiklos (approved):
Aha, the resolver has that _queryCache member, we don't protect access to that with a lock, so this can't be static, shared between threads, OK.

PR #10138, opened 2024-09-26
Private/caolan/can elide a dns lookup

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -32,8 +32,6 @@
 namespace net
 {
 
-#if !MOBILEAPP
-
 std::string HostEntry::makeIPAddress(const sockaddr* ai_addr)
 {
     char addrstr[INET6_ADDRSTRLEN];
@@ -125,6 +123,8 @@ HostEntry::HostEntry(const std::string& desc, const char* port)
 }
 
 HostEntry::~HostEntry() = default;
+
+#if !MOBILEAPP
 
 struct DNSCacheEntry
 {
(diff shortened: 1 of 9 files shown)


net/Socket.cpp
```diff
@@ -806,7 +806,7 @@ bool SocketPoll::insertNewUnixSocket(
 
     std::shared_ptr<StreamSocket> socket
         = StreamSocket::create<StreamSocket>(std::string(), fd, Socket::Type::Unix,
-                                             true, websocketHandler);
+                                             true, false, websocketHandler);
```
Ashod, line 809: I worry that multiple boolean arguments would be error prone.

Inline comments with the names would help, though verbose. Ideally, these would be enums.

Ashod (approved):
Thanks, @caolanm. Nice cleanup.

net/Socket.cpp
caolanm: made this an enum now

net/NetUtil.cpp
```diff
@@ -226,17 +226,22 @@ bool isLocalhost(const std::string& targetHost)
...
         }
     }
 
-    LOG_TRC("Host [" << targetHost << "] is not on the same host as the client: \"" << targetAddress
+    LOG_TRC("Host [" << _requestName << "] is not on the same host as the client: \"" << targetAddress
                      << "\".");
     return false;
 }
 
+bool isLocalhost(const std::string& targetHost)
```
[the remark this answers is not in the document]
caolanm: I'll punt this for a later time, Ideally I could remove net::isLocalhost (and net::resolveHostAddress, net::resolveAddresses) which are used by wsd/Storage.cpp with an async DNS solution.

net/Socket.cpp
Ashod: Thanks! Looks much better.

PR #10140, opened 2024-09-27
slideshow: autostart in window

eszkadev:
If we use presentation with autostart (.pps) then
fullscreen request fails as it is not invoked by the user interaction. For now do the fallback -> open in window. If user has disabled popup, he will be asked in the message box to allow the popup.


mcecchetti (approved):
Tested. Works fine, thanks!

PR #10141, opened 2024-09-29
writer: added downloadas html option

Rash419:
Change-Id: I302a46cc5982cee320f8f9496a790ecb7b18e335

* Target version: master


quikee (approved):
looks good

PR #10142, opened 2024-09-30
private/marco/anim engine trans filter for shapes

mcecchetti:
- **slideshow: refactoring: move from SlideInfo to TransitionFilterInfo**
- **slideshow: anim engine: displaySlide, check if we're still presenting**
- **slideshow: initial support for transition filters for shapes**

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -8,7 +8,7 @@
     "@types/jquery.contextmenu": "^1.7.38",
     "@types/jqueryui": "^1.12.21",
     "@types/mocha": "8.2.0",
-    "@types/node": "14.14.25",
+    "@types/node": "18.11.09",
     "@types/offscreencanvas": "^2019.7.3",
     "@types/w2ui": "^1.4.37",
     "@typescript-eslint/eslint-plugin": "^4.21.0",
--- a/browser/src/slideshow/engine/AnimationFactory.ts
+++ b/browser/src/slideshow/engine/AnimationFactory.ts
@@ -284,6 +284,119 @@ class ClippingAnimation extends AnimationBase {
 	}
 }
 
+class TransitionFilterAnimation extends AnimationBase {
+	private nNodeId: number;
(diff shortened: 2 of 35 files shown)


browser/package.json
```diff
@@ -8,7 +8,7 @@
     "@types/jquery.contextmenu": "^1.7.38",
     "@types/jqueryui": "^1.12.21",
     "@types/mocha": "8.2.0",
-    "@types/node": "14.14.25",
+    "@types/node": "18.11.09",
     "@types/offscreencanvas": "^2019.7.3",
```
eszkadev, line 12: Please do not change that if possibleour CI as we experienced in the past. Why is that needed?

browser/src/slideshow/engine/AnimationFactory.ts
```diff
@@ -347,16 +519,22 @@ function createShapeTransition(
...
+				transitionFilterInfo,
+				aAnimatedElement,
 			);
+
+			// const aParametricPolyPolygon = createClipPolyPolygon(
+			// 	eTransitionType,
+			// 	eTransitionSubType,
+			// );
+			// const aClippingAnimation = new ClippingAnimation(
+			// 	aParametricPolyPolygon,
```
eszkadev, line 533: dead comment? or for debbuging later?

browser/package.json
mcecchetti: I upgraded version for structuredClone() function looks nice for cloning arrays, maps.

browser/src/slideshow/engine/AnimationFactory.ts
mcecchetti: I'll remove when work on transition filter is over.

eszkadev (commented):
it looks good, but please bring back   NODE version :)

browser/package.json
eszkadev: ah it's just types :) it'as fine then

eszkadev (approved):

PR #10143, opened 2024-09-30
Only remove apparmor profile during package removal

thebearon:
postrm is also called when upgrading, after the files are unpacked ie. /etc/apparmor.d/usr.bin.coolwsd always got deleted.

Change-Id: I569a1ef6fe8f9346ed164bc1fd7e86e5bc113646

* Target version: master 

### Checklist


thebearon:
Not needed, @timar applied d3235d3121dfa870b476cf794ee498965b1a9752 directly (needs to be ported to master branch).

PR #10144, opened 2024-09-30
Feture: Save status notification

Darshan-upadhyay1110:
- This patch enhances user notification after saving.
- It displays the save progress and document saved status in the main menu bar.
- Added an arrow animation to indicate an ongoing save process.


Darshan-upadhyay1110:
[Screencast from 2024-09-30 16-05-21.webm](https://github.com/user-attachments/assets/8ca040ed-4d6a-438d-8c9a-31ca901580f9)

Darshan-upadhyay1110:
[Screencast from 2024-09-30 17-01-31.webm](https://github.com/user-attachments/assets/262ac5a0-5190-4cee-b759-ff84e705998e)

pedropintosilva (changes requested):
thanks @Darshan-upadhyay1110 ! Here are some things I have spotted immediately upon playing the video:
* Font-family needs fixing: We need to use the same font-family that we use for other UI
* No need to have multiple icons related to the same thing: Save. We already have a blue dot when the document has changes that are yet not save so, we could just replace that dot with a check mark and replace the icon completely while saving.
* In the tooltip (on hover) we should delete the prefix "Status:" it seems unnecessary. We should start with the action so: "Last saved: 38 s ago" . That time info is already being extracted and displayed in  the compact mode : 

image

image

Darshan-upadhyay1110:
> * No need to have multiple icons related to the same thing: Save. We already have a blue dot when the document has changes that are yet not save so, we could just replace that dot with a check mark and replace the icon completely while saving.

@pedropintosilva have you seen the "Saving..." text shown while saving ? Need opinion on that should we display it like that or just icon with animation will be enough ?

pedropintosilva:
> > * No need to have multiple icons related to the same thing: Save. We already have a blue dot when the document has changes that are yet not save so, we could just replace that dot with a check mark and replace the icon completely while saving.
> 
> @pedropintosilva have you seen the "Saving..." text shown while saving ? Need opinion on that should we display it like that or just icon with animation will be enough ?

Yes I think the text makes sense. But I wouldn't add a new icon element. I would instead replace the saving icon with a saving icon with a check etc that's what I was referring to. Is that clear? (Let me read again my original comment and if it's not clear I can try to clarify it more)

Darshan-upadhyay1110:
@pedropintosilva 
> Font-family needs fixing: We need to use the same font-family that we use for other UI

Done

>No need to have multiple icons related to the same thing: Save. We already have a blue dot when the document has changes that are yet not save so, we could just replace that dot with a check mark and replace the icon completely while saving.

Almost completed now i just need this new green check mark icon to replace with blue dot icon after save completed
1. Can you help me to create new save svg with green check ? i tried but i am not expert in this area... 

>In the tooltip (on hover) we should delete the prefix "Status:" it seems unnecessary. We should start with the action so: "Last saved: 38 s ago" . That time info is already being extracted and displayed in the compact mode :
Wroking on it

pedropintosilva:
I have amended the commit 24cec585cdae685cc29b108e6870ed6343f9ca52
Author: Darshan-upadhyay1110 <darshan.upadhyay@collabora.com>
Date:   Mon Sep 30 16:33:37 2024 +0530

    Feture: notify user after manual save.

Better to don't use small check mark (too small size to be understandable). Also better to avoid replace the save icon to a green check mark as it might because confusing (it might be perceived as a button to confirm or even worst on calc since we have other green check marks in the formula). Better to just reuse the save icon and make it green so it's very visible and the same with the label

pedropintosilva (changes requested):
- If I press ctrl+S and then during saving I double press ctrl + S I can make the green floppy icon (save status) to spin
- The save button seems responsive during all stages, it probably shouldn't be possible to save whole `Saving...` 
- Also I wonder if we can do something WRT the movement of the tab buttons when between Save, Saving and Saved states :thinking: 
  - Maybe we could simply set width of the ::after to 0 and then play with box-shadow maybe something like:
```css
width: 0;
box-shadow: 16px 0 0 20px var(--color-main-background), 32px 0 0px 20px var(--color-main-background);
```

Darshan-upadhyay1110:
> * lso I wonder if we can do something WRT the movement of the tab buttons when between Save, Saving and Saved states 🤔
>   
>   * Maybe we could simply set width of the ::after to 0 and then play with box-shadow maybe something like:

@pedropintosilva with 200% zoom level it looks like this

image

For other cases look good but I am concerned it will look odd doe tablet view and with high zoom level

Darshan-upadhyay1110:
> * If I press ctrl+S and then during saving I double press ctrl + S I can make the green floppy icon (save status) to spin

Done , I have added fix for this issue...

pedropintosilva:
> > * lso I wonder if we can do something WRT the movement of the tab buttons when between Save, Saving and Saved states 🤔
> >   
> >   * Maybe we could simply set width of the ::after to 0 and then play with box-shadow maybe something like:
> 
> @pedropintosilva with 200% zoom level it looks like this
> 
> For other cases look good but I am concerned it will look odd for tablet view and with high zoom level

That was just an idea. Anyways, I have fixed by setting the pseudo class with absolute position and under the icon. I have also fixed the something else in the compact mode

pedropintosilva (approved):

Darshan-upadhyay1110:
> > > * lso I wonder if we can do something WRT the movement of the tab buttons when between Save, Saving and Saved states 🤔
> > >   
> > >   * Maybe we could simply set width of the ::after to 0 and then play with box-shadow maybe something like:
> > 
> > 
> > @pedropintosilva with 200% zoom level it looks like this
> > For other cases look good but I am concerned it will look odd for tablet view and with high zoom level
> 
> That was just an idea. Anyways, I have fixed by setting the pseudo class with absolute position and under the icon. I have also fixed the something else in the compact mode

Thanks, Pedro 🙇🏼

timar:
I'm sorry, but why was this PR merged? It was tagged as CODE "red". It adds new strings, and even worse, it adds new unlocalizable strings. "Saving..." and "Saved" are coming from css/menubar.css!

browser/css/menubar.css
```diff
@@ -52,6 +52,62 @@
...
+	background: url('images/lc_saving.svg') no-repeat center;
+}
+
+nav.hasnotebookbar #save::after {
+	position: absolute;
+	top: 18px;
+}
+
+nav.hasnotebookbar #save.saving::after {
+	content: 'Saving';
```
timar, line 92: Not translatable.

browser/css/menubar.css
```diff
@@ -52,6 +52,62 @@
...
+	color: var(--color-main-text);
+}
+
+#save.saved img {
+	background: url('images/lc_saved.svg') no-repeat center;
+	filter: grayscale(0);
+}
+
+nav.hasnotebookbar #save.saved::after {
+	content: 'Saved';
```
timar, line 104: Not translatable.

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2475,8 +2475,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		$(controls.button).on('click', clickFunction);
 		$(controls.label).on('click', clickFunction);
-		$(div).on('mouseenter', mouseEnterFunction);
-		$(div).on('mouseleave', mouseLeaveFunction);
+		// We need a way to also handle the cutome tooltip for any tool button like save in shortcut bar
```
timar, line 2478: cutome -> custom

browser/src/control/Control.UIManager.js
```diff
@@ -1172,6 +1172,40 @@ L.Control.UIManager = L.Control.extend({
...
+		if (window.mode.isMobile())
+			return;
+
+		const saveEle = document.getElementById('save');
+		// Only do saving animation if any content is modified in document
+		if (saveEle.classList.contains('savemodified')) {
+			const saveIconEl = document.querySelector('#save img');
+			saveEle.classList.remove('savemodified');
+			saveEle.classList.add('saving');
+			saveIconEl.classList.add('rotate-icon'); // Stop the icon rotation
```
timar, line 1186: Start the icon rotation.

browser/src/control/Control.UIManager.js
```diff
@@ -1172,6 +1172,40 @@ L.Control.UIManager = L.Control.extend({
...
+		const saveEle = document.getElementById('save');
+		if (saveEle.classList.contains('saving')) {
+			const saveIconEl = document.querySelector('#save img');
+			saveEle.classList.remove('saving');
+			saveIconEl.classList.remove('rotate-icon'); // Stop the icon rotation
+			saveEle.classList.add('saved');
+			// Add some delay to show "saved" status, then hide this info
+			setTimeout(() => {
+				saveEle.classList.remove('saved');
+				saveEle.removeAttribute('disabled');  // Disable the button
```
timar, line 1204: Enable the button

browser/src/map/Map.js
```diff
@@ -458,33 +461,41 @@ L.Map = L.Evented.extend({
...
-				timeout = 60000;
-			} else if (elapsed < 3600000 * 24) {
-				dateValue = _('Last saved:') + ' ' + rtf1.format(-Math.round(elapsed / 3600000), 'hour');
-				timeout = 60000;
-			} else {
-				dateValue = _('Last saved:') + ' ' + dateTime.toLocaleDateString(String.locale,
-					{ year: 'numeric', month: 'short', day: 'numeric', hour: '2-digit', minute: '2-digit' });
-				timeout = 60000;
-			}
+		// just for safty check sometimes this var may not be intialized then in that case better to return empty
```
timar, line 464: safty -> safety

timar (commented):
Please fix these in a followup PR.

pedropintosilva:
@timar is right on all accounts plus one of the commit messages has typos. Please make sure you checked all your changes in every push you do to a PR otherwise, I end up spending more time than I expected reviewing and fixing.

For the content: probably the best/easiest would be to replace all the string sin CSS with a CSS vars and then set those CSS vars in JS encapsulated with the respective known `_('')`

pedropintosilva:
on top of that pretty please turn on your spell checker

pedropintosilva:
Also would be good to add the following to the `/* hide documen-header to free up space when using narrow windows */
@media screen and (max-width: 800px) {` in menubar.css

```css
	.notebookbar-shortcuts-bar {
		padding-inline-start: 5px;
	}
```

this way when the logo disappears we retain the margin

Or even better would be to don't do this ^ and instead remove also the padding-left from the logo while adding to the `.main-nav` the padding-inline-start of 5px

pedropintosilva:
@Darshan-upadhyay1110 also the content is being cropped in chrome and not so much in FF. The reason is probably the line-height. Better to set line-height to 0 since the string will always be one line. With that change the `top` weird value can then be changed to `bottom` and the vlaue can be something more sensible `2px` for example (so it's more understandable that is meant to function - visually - as a margin)

Darshan-upadhyay1110:
@timar followup work is merged : https://github.com/CollaboraOnline/online/pull/10193

Issue #1529, opened 2021-02-24
Support mentioning of contacts in comments


thkoch2001:
**Is your feature request related to a problem? Please describe.**
I asked in Nextcloud for Google Docs-like features of mentioning users in comments:
https://github.com/nextcloud/richdocuments/issues/1380

This feature request is for CollaboraOnline to provide the necessary APIs.

**Describe the solution you'd like**

When a certain symbol gets entered in a comment, e.g. "+" or "@", CollaboraOnline offers an autocomplete of usernames. CO needs to call into the hosting application to get the list of autocompletions.

Once the comment gets posted, CO instructs the hosting application to notify a set of users about the new comment.

thkoch2001:
I believe issue #1673 is related to this one.
I want automatic sharing of comments with mentioned users. #1673 is about getting a link that can be shared manually.

juliushaertl:
After discussing this with @kendy we'd need to have an implementation of this in core first to properly persist the mentions among with the user metadata within the document files

Requirements for a "Mention" element that users would add through the UI:
- Unique user identifier (could be any string, for Nextcloud it would be either the user id or federated cloud id)
- Display name (readable name of the user, as the user id might be a uuid or anything that would not be user friendly)
- E-Mail address (Optional, but could help for providing an option to interact with the user through Collabora Office on the desktop)
- Avatar (Similar to https://github.com/CollaboraOnline/online/issues/4590 in order to be able to display the avatar of the mentioned user)

To give a more concrete example, integrating with a Nextcloud instance on cloud.example.com  if a user would type `@Julius` the interface would fetch the user metadata from Nextcloud and persist that to the file:
- userId: julius@cloud.example.com
- displayName: Julius Haertl
- email: julius@example.com
- avatar: base64 encoded avatar

One additional thing to consider would be that depending on the integrator the userid might not work as a cross reference, between different integrators, for that it would be required to have some way to differentiate:

- namespace with a prefix `nextcloud:julius@cloud.example.com`
- use a url, so that the integrator can check if it is for example a nextcloud instance, this could use the Nextcloud profile page https://cloud.example.com/u/julius
- add an additional metadata for the integrator identifier

Open questions:
- Can this be supported on only specific types (like OpenDocument) or also others like OOXML?

Leaving this up to @kendy for further evaluation of the technical details on the core/cool side as discussed.

juliushaertl:
[...]

PR #10145, opened 2024-09-30
implement comment at-mentions

caolanm:
* Resolves: #1529 
* Target version: master

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -1034,6 +1034,7 @@ export class Comment extends CanvasSectionObject {
 			this.sectionProperties.data.reply = this.sectionProperties.data.text;
 			this.sectionProperties.commentListSection.saveReply(this);
 		} else {
+			this.removeBRTag(this.sectionProperties.nodeReplyText);
 			this.sectionProperties.data.reply = this.sectionProperties.nodeReplyText.textContent;
 			this.sectionProperties.data.html = this.sectionProperties.nodeReplyText.innerHTML;
 			// Assigning an empty string to .innerHTML property in some browsers will convert it to 'null'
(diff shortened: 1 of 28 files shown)


browser/src/canvas/sections/CommentSection.ts
```diff
@@ -508,7 +508,7 @@ export class Comment extends CanvasSectionObject {
 		// Original unlinked text
 		this.sectionProperties.contentText.origText = this.sectionProperties.data.text ? this.sectionProperties.data.text: '';
 		this.sectionProperties.nodeModifyText.textContent = this.sectionProperties.data.text ? this.sectionProperties.data.text: '';
-		this.sectionProperties.nodeModifyText.value = this.sectionProperties.data.text ? this.sectionProperties.data.text: '';
+		this.sectionProperties.nodeModifyText.textContent = this.sectionProperties.data.text ? this.sectionProperties.data.text: '';
```
caolanm, line 506: @Rash419 we end up with the same line here twice, so presumably we don't need one of those?

browser/src/control/AutoCompletePopup.ts
```diff
@@ -136,16 +118,45 @@ abstract class AutoCompletePopup {
 		});
 	}
 
+	getCursorPosition(): Point {
+		const commentSection = app.sectionContainer.getSectionWithName(
+			L.CSections.CommentList.name,
+		);
+
+		if (commentSection?.sectionProperties?.selectedComment?.isEdit()) {
```
caolanm, line 126: (bizarrely) the "selectedComment" appears to actually be the *root* comment of the current comment. So if we insert a comment, and then reply to the comment and "ok" and then edit the reply and use an at-mention. The selectedComment is the root comment, which is not actually being edit and the at-mention popup appears in the main document

browser/src/control/AutoCompletePopup.ts
```diff
@@ -65,11 +65,12 @@ abstract class AutoCompletePopup {
 	abstract onAdd(): void;
 
 	closePopup(): void {
-		var popupExists = L.DomUtil.get(this.popupId);
+		var popupExists = L.DomUtil.get(this.popupId + 'List');
```
caolanm, line 70: @Rash419 After this change, if I type @ followed by something that doesn't mater, e.g. aaaa, then the 'no search results found' popup doesn't disappear on pressing esc

eszkadev (approved):
looks fine

PR #10146, opened 2024-09-30
Fix hyperlink popup position.

gokaysatir:
When the preview of the link is shown, the position of the popup needs to be re-adjusted. Or if the link increases the height of the popup, popup's position overlaps with the current cursor position.

Also added borders to arrow div.


eszkadev (approved):
code looks good

eszkadev:
Fixes #10151

PR #10149, opened 2024-09-30
feat(impress): Add legacy export types

Minion3665:
This patch exposes in online some rare export options that are available in core.

Notably, one of those "download" options is SVG, which poses a problem for how we currently handle slideshows: we need those not to download. We can mitigate this by using the attachment=0 query parameter for slideshows rather than blocking downloads for all SVGs.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
looks reasonable to me

PR #10154, opened 2024-10-01
slideshow: avoid infinite render

eszkadev:
if we went to the last slide saying "click to close" the rendering was retriggered all the time.
that caused issue with going to the previous slides which were covered by the black slide with text


codewithvk (approved):
Nice fix. Thanks!

PR #10165, opened 2024-10-03
Misc Improvements

Ashod:
Larger than I'd have liked due two breaking up upload handling into 3 functions (non-functional refactoring).
Otherwise, _mostly_ trivial changes.

- wsd: minor non-functional improvements to Delta
- wsd: test: better Rename test
- wsd: refactor the handling of successful upload
- wsd: refactor the handling of failed upload
- wsd: end renaming upon upload failure
- wsd: make DocumentBrokerPoll a shared_ptr
- wsd: capture the original storage write Permission
- wsd: setWritePermission on each loaded session
- wsd: add DocumentBroker::onViewLoaded
- wsd: test: better setExpectedPutFile in UploadConflictCommon
- wsd: test: extend the timeouts in UnitProxy
- wsd: log: support on-deman flushing own buffer
- wsd: mark the dump-state better
- wsd: add getFirstAuthorizedSession to DocBroker

### DIFF
--- a/common/Session.cpp
+++ b/common/Session.cpp
@@ -27,24 +27,25 @@ using namespace COOLProtocol;
 
 using Poco::Exception;
 
-Session::Session(const std::shared_ptr<ProtocolHandlerInterface> &protocol,
-                 const std::string& name, const std::string& id, bool readOnly) :
-    MessageHandlerInterface(protocol),
-    _id(id),
-    _name(name),
-    _disconnected(false),
-    _isActive(true),
-    _lastActivityTime(std::chrono::steady_clock::now()),
-    _isCloseFrame(false),
-    _isWritable(!readOnly),
-    _isReadOnly(readOnly),
-    _isAllowChangeComments(false),
-    _haveDocPassword(false),
-    _isDocPasswordProtected(false),
-    _isAdminUser(std::nullopt),
-    _watermarkOpacity(0.2),
-    _accessibilityState(false),
-    _disableVerifyHost(false)
+Session::Session(const std::shared_ptr<ProtocolHandlerInterface>& protocol, const std::string& name,
+                 const std::string& id, bool readOnly)
+    : MessageHandlerInterface(protocol)
+    , _id(id)
+    , _name(name)
+    , _disconnected(false)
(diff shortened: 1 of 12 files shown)


common/Session.cpp
```diff
@@ -318,11 +319,13 @@ void Session::getIOStats(uint64_t &sent, uint64_t &recv)
 
 void Session::dumpState(std::ostream& os)
 {
-    os << "\n\t\tid: " << _id
+    os << std::boolalpha
```
caolanm, line 322: When we have these state changers, should we restore them before leaving scope, or have them set at the stream construction site? There's already plenty of them, so not a new issue though.

wsd/DocumentBroker.cpp
```diff
@@ -2625,6 +2625,22 @@ void DocumentBroker::onViewLoaded(const std::shared_ptr<ClientSession>& session)
     //TODO: Take the WOPI lock, if necessary.
 }
 
+std::shared_ptr<ClientSession> DocumentBroker::getFirstAuthorizedSession() const
```
caolanm, line 2628: I presume there is a plan to build on getFirstAuthorizedSession() later, doesn't seem used in this pr

common/Session.cpp
Ashod: Good question. I've struggled with it myself for the longest time. The approach that I've assumed is to set what we need and assume any arbitrary state. This is because it's really hard to know/remember default states, and it's customary to set states per-preference. Having said that, I think we reset `hex` to `dec`, since it's obvious that we want `dec` by default and `hex` is the exception.

Not ideal, but since we only have a few of those (`boolalpha` and `hex` are by far the most common), it seems to be fine.

But we can change that. If you feel it's cleaner to always reset to a certain "default" (even if by agreement), it might be worth it.

wsd/DocumentBroker.cpp
Ashod: Yes, patches forthcoming!

caolanm (approved):
missing trailing d off "on-deman" in commit message is probably not worth respinning for

PR #10166, opened 2024-10-03
cool#9992 doc sign: add sign button to the home tab of the notebookbar

vmiklos:
If the document has one signature, then the status bar has an icon and
you can add more signatures. But you can't add a signature if the doc
has no signature at all, which is the primary use-case.

There is no icon is the "no sign" case because signing is a relatively
obscure feature, and it's good to not clutter the status bar with an
icon by default.

Fix the problem by adding a button for it on the File tab of the
notebookbar. This has the benefit that it's not visible by default (Home
is the default), and it's also next to the Properties button, matching
the desktop order in the File menu.

Possibly later we'll hide these when we learn that user has no sign
certificates configured, that's not yet done here.

Change-Id: I8383f6f5b14a2770e0fd6a5e804f68ed7c67a215

### DIFF
--- a/browser/src/control/Control.NotebookbarCalc.js
+++ b/browser/src/control/Control.NotebookbarCalc.js
@@ -247,6 +247,18 @@ L.Control.NotebookbarCalc = L.Control.NotebookbarWriter.extend({
 				'type': 'container',
 				'children': [
 					{
+						'id': 'signature',
+						'type': 'bigtoolitem',
+						'text': _('Signature'),
+						'command': '.uno:Signature',
+						'accessibility': { focusBack: true, combination: 'SN' }
+					}
+				]
+			},
+			{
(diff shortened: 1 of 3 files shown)


vmiklos:

This is meant to reduce the amount of my local online changes as I work through the wopi/wsd/kit changes that still needs doing.

caolanm (approved):
I imagine at some point we will have too many buttons in the notebookbar, but a future problem I hope

vmiklos:
Perhaps the first limit we'll hit is that buttons fit, but there are so many buttons that it's hard to find the one you're interested in. Looking at what others do, some apps have sections inside one tab, and only then the buttons -- perhaps sooner or later we could also have something like that?

CC @eszkadev and @pedropintosilva , nothing immediately actionable, just for the long-term.

PR #10169, opened 2024-10-03
Slideshow: Use RenderContextGl for LayerRendererGl

codewithvk:
Refactored LayerRendererGl to utilize RenderContextGl, replacing repeated functions to reduce code duplication and improve maintainability.

Change-Id: I929ca8bb65897e98a10b4efa92c068a257cd9e16

* Resolves: # 
* Target version: master

### DIFF
--- a/browser/src/slideshow/LayerRenderer.ts
+++ b/browser/src/slideshow/LayerRenderer.ts
@@ -34,6 +34,7 @@ class LayerRendererGl implements LayerRenderer {
 	];
 
 	private offscreenCanvas: OffscreenCanvas;
+	private glContext: RenderContextGl;
 	private gl: WebGLRenderingContext;
 	private program: WebGLProgram;
 	private positionBuffer: WebGLBuffer;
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/LayerRenderer.ts
```diff
@@ -47,10 +48,8 @@ class LayerRendererGl implements LayerRenderer {
 
 	constructor(offscreenCanvas: OffscreenCanvas) {
 		this.offscreenCanvas = offscreenCanvas;
-		this.gl = this.offscreenCanvas.getContext('webgl');
-		if (!this.gl) {
-			throw new Error('WebGL not supported');
```
eszkadev, line 52: please leave this exception, I think we need that for detecting when to use 2d fallback

codewithvk: Nice point

codewithvk: @eszkadev I have tested, It works fine without it as we already throwing error from `RenderContextGl` constructor so we don't need to worry about it :)

mcecchetti (approved):
Looks good to me. Thanks!

PR #10170, opened 2024-10-04
wsd: reap the background save process

Ashod:
This reaps zombie child processes that we
couldn't reap after getting disconnected
from them. Very likely because the kernel
hadn't unloaded the process yet.

We reap any child process after spawning
a background save. This way, we would
have at most 1 zombie at any time.

This gives the minimum overhead while
ensuring that we don't leak.

Change-Id: Idd8531ec721744fa31d7ec767b87dc3deb18c193

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -35,6 +35,7 @@
 #include <utime.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include <sys/wait.h>
 #include <sysexits.h>
 
 #include <atomic>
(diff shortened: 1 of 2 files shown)


caolanm (approved):
looks reasonable to me

mmeeks (approved):
Well it's not broken of course - but why does the code in:

void BgSaveParentWebSocketHandler::onDisconnect()

which reaps the children not work nicely ?

And is there some code sharing we could do there ? but still - should work fine.

PR #10176, opened 2024-10-04
cypress: add visual testing example

eszkadev:
snapshots to compare are stored inside:
`cypress_test/integration_tests/snapshots/base`
if you need to generate new one, you can just run test and you will find current results inside:
`cypress_test/integration_tests/snapshots/actual`
what you can copy to base


mmeeks:
Ooh - sounds interesting =)

quikee (approved):
Looks good to me...

Issue #5147, opened 2022-08-12
Holding Ctrl+Shift should preserve the ratio of images & shapes when resizing them


SustReal:
**Describe the bug**
Maybe also rather a request for an enhancement than a bug. People used to work with graphical programs (Adobe, Gimp, Inkscape) are used to hold the key combination "ctrl + shift" to preserve the ration of objects (e.g. images) when resizing them. For probably historic reasons in Collabora Online this is achieved by the "shift lock" key. However, this results in a bad user experience, as users would expect the more widespread key combination to work. It is also a bad experience for users coming from MS 365, as MS already allows both key combinations to preserve the ratio. 

**To Reproduce**
1. Open a document
2. Import an image or create a shape
3. Select the image or shape, Hold "ctrl + shift", an resize the image 
4. See the ratio not being preserved

**Expected behavior**
Holding "ctrl + shift" should preserve the ratio when resizing objects...

**Actual behavior**
Holding "ctrl + shift" does not preserve the ratio...

**Screenshots**
---

**Desktop (please complete the following information)**
 - OS: Ubuntu 22.04
 - Browser: Firefox 103
 - Version: COOL 22.11.05 and 22.05

**Smartphone (please complete the following information)**
-

**Additional context**
I thought preserving the ratio holding "ctrl + shift" had been implemented in earlier versions. This might be a regression...

lefinal:
Would be nice to have this finally fixed as this is really annoying. Thanks guys!

PR #10178, opened 2024-10-04
feat(UI): Holding Ctrl+Shift to preserve the image ratio when resizing

maks-w:
Change-Id: Id340ab65134b09621d2a965c2dc27e4204586f92

* Resolves: #5147 
* Target version: master 

### Summary

Keep the ratio of the image/shape when Ctrl+Shift are pressed. 

### Checklist


gokaysatir (approved):
This is beautiful, thank you:)

PR #10181, opened 2024-10-07
Fewer MOBILEAPP Directive

Ashod:
- wsd: restore performance regression due to isMobileApp
- wsd: remove ifdef MOBILEAPP
- wsd: remove ifdef MOBILEAPP
- wsd: header cleanup
- wsd: remove ifdef MOBILEAPP

### DIFF
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -520,7 +520,7 @@ namespace FileUtil
     {
         assert(!path.empty());
 
-        if (!Util::isMobileApp())
+        if constexpr (!Util::isMobileApp())
         {
             bool hookResult = true;
             if (UnitBase::get().filterCheckDiskSpace(path, hookResult))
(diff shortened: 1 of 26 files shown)


caolanm (commented):
principle looks good, but current compile error on mobile platforms of:

wsd/Storage.cpp:86:13: error: use of undeclared identifier 'HostUtil'
HostUtil::parseWopiHost(app.config());

Ashod:
> principle looks good, but current compile error on mobile platforms of:
> 
> wsd/Storage.cpp:86:13: error: use of undeclared identifier 'HostUtil' HostUtil::parseWopiHost(app.config());

Yes, some of these can't be converted to constexpr conditionals. I'm learning the hard-way and reverting selectively.
Others are impossible to convert altogether (conditionals outside of functions, around includes, etc.).
Still, any reduction of compiler directives seems a win.

caolanm (approved):
Looks good to me

PR #10182, opened 2024-10-07
Dump Memory Usage with SIGUSR1

Ashod:
- wsd: avoid unnecessary temporary strings
- wsd: support reading unknown-sized files
- wsd: dump the smaps summary on USR1

### DIFF
--- a/common/FileUtil.cpp
+++ b/common/FileUtil.cpp
@@ -20,6 +20,7 @@
 #include <pwd.h>
 #include <stdexcept>
 #include <sys/time.h>
+#include <unistd.h>
 #ifdef __linux__
 #include <sys/vfs.h>
 #elif defined IOS
@@ -387,6 +388,7 @@ namespace FileUtil
     std::unique_ptr<std::vector<char>> readFile(const std::string& path, int maxSize)
     {
         auto data = std::make_unique<std::vector<char>>(maxSize);
+        data->resize(0);
         return (readFile(path, *data, maxSize) >= 0) ? std::move(data) : nullptr;
     }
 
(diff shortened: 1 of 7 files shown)


kit/Kit.cpp
```diff
@@ -2494,8 +2494,14 @@ void Document::dumpState(std::ostream& oss)
         << "\n\tduringLoad: " << _duringLoad
         << "\n\tmodified: " << name(_modified)
         << "\n\tbgSaveProc: " << _isBgSaveProcess
-        << "\n\tbgSaveDisabled: "<< _isBgSaveDisabled
-        << "\n";
+        << "\n\tbgSaveDisabled: "<< _isBgSaveDisabled;
+
+    std::string smap;
+    if (const ssize_t size = FileUtil::readFile("/proc/self/smaps_rollup", smap); size <= 0)
+        oss << "\n  smaps_rollup: <unavilable>";
```
caolanm, line 2501: s/unavilable/unavailable FWIW

common/FileUtil.cpp
```diff
@@ -387,6 +388,7 @@ namespace FileUtil
     std::unique_ptr<std::vector<char>> readFile(const std::string& path, int maxSize)
     {
         auto data = std::make_unique<std::vector<char>>(maxSize);
+        data->resize(0);
```
caolanm, line 391: So, before this was actually appending the contents after mazSize zeros? So, e.g. the case in wsd/wopi/WopiProxy.cpp StorageBase::StorageType::FileSystem was broken?

caolanm (approved):

common/FileUtil.cpp
Ashod: Yes, it seems so. But since the response was an empty file (all null), it had no ill-effect.

kit/Kit.cpp
Ashod: I swear it was fixed, evidently in a different branch. Thanks for the catch!

Ashod:
> I wonder - we should prolly also dump the prometheus metrics on USR1 =) that's a nice set of strings that may help us. 

Good idea. Done: https://github.com/CollaboraOnline/online/pull/10182/commits/d7e1ecf86439c82dbb4656974f0307f20119c499

Issue #9989, opened 2024-09-05
perf: de-prioritize rendering non-visible tiles 


mmeeks:
In Kit.cpp (drainQueue) while we batch all the incoming input pieces first we then render all incoming tile requests.

      ```
  if (processInputEnabled() && !isLoadOngoing() &&
            !isBackgroundSaveProcess() && _queue->getTileQueueSize() > 0)
        {
            std::vector<TileCombined> tileRequests = _queue->popWholeTileQueue();

            // Put requests that include tiles in the visible area to the front to handle those first
            std::partition(tileRequests.begin(), tileRequests.end(), [this](const TileCombined& req) {
                return isTileRequestInsideVisibleArea(req); });
            for (auto& tileCombined : tileRequests)
                renderTiles(tileCombined);
     }
```

We should instead while there are events to process:

a) only render & send those that are inside a visible area and
b) re-spin the main-loop  and
c) only render non-visible tiles when there is nothing else to do =)

sgothel:
@mmeeks is it OK for the invisible tiles requests to just vanish, i.e. being ignored iff the main-loop is busy (later on that)? 
In such case .. we just issue `renderTiles` on the `isTileRequestInsideVisibleArea` requests? 
(i.e. don't partition to favor visible tiles, but just for_all and send the visible tiles?)
Otherwise we would need to queue the pending invisible, but AFAIK this would be too much handling outdated data?

Need to check on the 'busy main-loop' criteria, i.e. `nothing else to do`

mmeeks:
> @mmeeks is it OK for the invisible tiles requests to just vanish, i.e. being ignored iff the
> main-loop is busy (later on that)?

I would think c) above was quite clear on that. No it's not ok.

sgothel:
> > @mmeeks is it OK for the invisible tiles requests to just vanish, i.e. being ignored iff the
> > main-loop is busy (later on that)?
> 
> I would think c) above was quite clear on that. No it's not ok.

Maybe a miswording, `vanish` is wrong. Paused until a better idle time as you wrote.
Will make a draft to discuss.

PR #10183, opened 2024-10-07
perf: de-prioritize rendering non-visible tiles

sgothel:
* Resolves: #9989
* Target version: master 

### Summary
Render combined-tiles, timeout limited, invisible last
    
Requested combined-tiles are send until timeout,
at least one item, visible items first.
This approach shall increase responsiveness.

`KitSocketPoll::kitPoll` loops through max-clipped 100ms `poll` slices,
allowing new incoming client requests.
`TileCombined` request are sent as much as remaining time allows,
at least on item while favoring visible items.

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -2377,12 +2377,14 @@ void Document::drainCallbacks()
         _websocketHandler->flush();
 }
 
-void Document::drainQueue()
+DocDrainStats Document::drainQueue(std::chrono::steady_clock::time_point startTime, long timeoutMicroS)
 {
+    DocDrainStats res{ .messages=0, .tilesPre=_queue->getTileQueueSize(),
+                       .tilesSent=0, .tilesPost=0};
     if (UnitKit::get().filterDrainQueue())
     {
         LOG_TRC("Filter disabled drainQueue");
-        return;
+        return res;
     }
 
     try
@@ -2406,7 +2408,7 @@ void Document::drainQueue()
             }
 
             const KitQueue::Payload input = _queue->pop();
-
+            ++res.messages;
             LOG_TRC("Kit handling queue message: " << COOLProtocol::getAbbreviatedMessage(input));
 
             const StringVector tokens = StringVector::tokenize(input.data(), input.size());
@@ -2441,17 +2443,41 @@ void Document::drainQueue()
             }
         }
 
+        const size_t tilesQueued = _queue->getTileQueueSize();
         if (processInputEnabled() && !isLoadOngoing() &&
-            !isBackgroundSaveProcess() && _queue->getTileQueueSize() > 0)
+            !isBackgroundSaveProcess() && tilesQueued > 0)
         {
+            // Render as much tiles until reaching timeout, if timeout>0.
+            // Otherwise render only the first visible TileCombined, but at least one TileCombined (if none are visible).
+            // The latter ensures draining tiles even without timeout>0.
+            const auto drainEnd = startTime + std::chrono::microseconds(timeoutMicroS);
             std::vector<TileCombined> tileRequests = _queue->popWholeTileQueue();
-
-            // Put requests that include tiles in the visible area to the front to handle those first
-            std::partition(tileRequests.begin(), tileRequests.end(), [this](const TileCombined& req) {
-                return isTileRequestInsideVisibleArea(req); });
-            for (auto& tileCombined : tileRequests)
-                renderTiles(tileCombined);
+            std::vector<TileCombined*> earmarkedTiles;
+            for (TileCombined& tileCombined : tileRequests)
+            {
+                if ( ( res.tilesSent==0 && isTileRequestInsideVisibleArea(tileCombined) ) ||
+                     ( timeoutMicroS>0 && std::chrono::steady_clock::now() < drainEnd ) )
+                {
+                    renderTiles(tileCombined);
+                    res.tilesSent += tileCombined.getTiles().size();
+                }
+                else if (res.tilesSent>0)
+                    _queue->pushTileQueue(tileCombined.getTiles());
+                else
+                    earmarkedTiles.push_back(&tileCombined);
+            }
+            for (TileCombined* tileCombined : earmarkedTiles)
+            {
+                if (res.tilesSent==0) // ensure at least one TileCombined gets rendered (if all are invisible)
+                {
+                    renderTiles(*tileCombined);
+                    res.tilesSent += tileCombined->getTiles().size();
+                }
+                else
+                    _queue->pushTileQueue(tileCombined->getTiles());
+            }
         }
+        res.tilesPost = _queue->getTileQueueSize();
     }
     catch (const std::exception& exc)
     {
@@ -2465,6 +2491,7 @@ void Document::drainQueue()
         if constexpr (!Util::isMobileApp())
             flushAndExit(EX_SOFTWARE);
     }
+    return res;
 }
 
 /// Return access to the lok::Document instance.
@@ -2819,12 +2846,13 @@ std::shared_ptr<KitSocketPoll> KitSocketPoll::create() // static
 }
 
 // process pending message-queue events.
-void KitSocketPoll::drainQueue()
+DocDrainStats KitSocketPoll::drainQueue(std::chrono::steady_clock::time_point startTime, long timeoutMicroS)
 {
     SigUtil::checkDumpGlobalState(dump_kit_state);
 
     if (_document)
-        _document->drainQueue();
+        return _document->drainQueue(startTime, timeoutMicroS);
+    return { .messages=0, .tilesPre=0, .tilesSent=0, .tilesPost=0};
 }
 
 // called from inside poll, inside a wakeup
@@ -2874,11 +2902,12 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
     // The maximum number of extra events to process beyond the first.
     int maxExtraEvents = 15;
     int eventsSignalled = 0;
+    int drainedQueueCalls = 0;
 
-    auto startTime = std::chrono::steady_clock::now();
+    const auto startTime = std::chrono::steady_clock::now();
 
     // handle processtoidle waiting optimization
-    bool checkForIdle = ProcessToIdleDeadline >= startTime;
+    const bool checkForIdle = ProcessToIdleDeadline >= startTime;
 
     if (timeoutMicroS < 0)
     {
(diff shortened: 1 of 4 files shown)


sgothel:
The TileCombined's dimension, i.e. its _width and _height, is not fixed in the streaming TileCombined  ctor (line ~404).
Only the 1st TileDesc dimenstion is being stored.
The combined dimension seems to be used at doRender only?
If fixed, i.e. AABBox style grown (-> Rectangle) we simply could use it to determine visibility .. instead of looping through all recreated TileDesc instances.

sgothel:
I need clarification regarding the IDLE criteria, which shall allow sending the invisible tiles.
(timeout  = to)
This patch uses the to_left < time_consumed in drainedQueue, which works if initial to is reasonable reflecting time left.
Testing showed that a to > 0 is only given for IDLE periods?

sgothel:
Force push resolved merge conflict with latest master tip as detected with the bots. I must have pushed it this morning in-between.

common/Rectangle.hpp
```diff
@@ -63,39 +85,49 @@ struct Rectangle
...
+    /// Returns whether this Rectangle intersects (partially contains) given Rectangle.
+    constexpr bool intersects(const Rectangle& rOther) const /*noexcept*/
     {
-        Util::Rectangle intersection;
-        intersection._x1 = std::max(_x1, rOther._x1);
-        intersection._y1 = std::max(_y1, rOther._y1);
-        intersection._x2 = std::min(_x2, rOther._x2);
-        intersection._y2 = std::min(_y2, rOther._y2);
-        return intersection.isValid();
+        const int x1 = std::max(_x1, rOther._x1);
```
mmeeks, line 115: If there are no unit tests for this code; why change it ? If we are changing it in a not totally obvious way - there should be unit tests.

kit/ChildSession.cpp
```diff
@@ -1223,8 +1223,7 @@ bool ChildSession::clientVisibleArea(const StringVector& tokens)
 
 bool ChildSession::isTileInsideVisibleArea(const TileDesc& tile) const
 {
-    return (tile.getTilePosX() >= _clientVisibleArea.getLeft() && tile.getTilePosX() <= _clientVisibleArea.getRight() &&
-        tile.getTilePosY() >= _clientVisibleArea.getTop() && tile.getTilePosY() <= _clientVisibleArea.getBottom());
+    return tile.intersects( _clientVisibleArea );
```
mmeeks, line 1226: Looks prettier - does that exchange four comparisons for four branches with std::min and two comparisons ? or did I miss something ? =)

kit/Kit.cpp
```diff
@@ -2296,7 +2296,7 @@ void Document::drainCallbacks()
         _websocketHandler->flush();
 }
 
-void Document::drainQueue()
+void Document::drainQueue(std::chrono::steady_clock::time_point now, int timeoutMicroS)
```
mmeeks, line 2299: Like the passing of now =)

kit/Kit.cpp
```diff
@@ -2365,11 +2365,40 @@ void Document::drainQueue()
...
-            // Put requests that include tiles in the visible area to the front to handle those first
-            std::partition(tileRequests.begin(), tileRequests.end(), [this](const TileCombined& req) {
-                return isTileRequestInsideVisibleArea(req); });
-            for (auto& tileCombined : tileRequests)
-                renderTiles(tileCombined);
+            const int consumedMicroS = std::chrono::duration_cast<std::chrono::microseconds>(
+                                           std::chrono::steady_clock::now() - now).count();
+            timeoutMicroS -= consumedMicroS;
+
+            if( timeoutMicroS >= consumedMicroS )
```
mmeeks, line 2372: This is curious to me. If we have a huge batch of tiles that will take 100's of milliseconds to render - some of them visible, some of them irrelevant - we will now go render all of them in if it didn't take very long to process the preceeding input events (?)

But we will not re-spin the poll to improve responsiveness or interactivity in this case - right ?

kit/Kit.cpp
```diff
@@ -2365,11 +2365,40 @@ void Document::drainQueue()
...
+            }
+            else
+            {
+                // Limited time left, push-back all invisible tile-requests
+                size_t invisibleCount = 0;
+                for(TileCombined req : tileRequests)
+                {
+                    if( isTileRequestInsideVisibleArea(req) )
+                    {
+                        renderTiles(req);
```
mmeeks, line 2391: Here we seem to render tiles individually one by one - which can create a massive performance problem - vs. rendering groups of them; why do we do that ?

wsd/ClientSession.cpp
```diff
@@ -3012,9 +3011,8 @@ bool ClientSession::isTileInsideVisibleArea(const TileDesc& tile) const
         if (!isSplitPane(panes[i]))
             continue;
 
-        Util::Rectangle paneRect = getNormalizedVisiblePaneArea(panes[i]);
-        if (tile.getTilePosX() >= paneRect.getLeft() && tile.getTilePosX() <= paneRect.getRight() &&
-            tile.getTilePosY() >= paneRect.getTop() && tile.getTilePosY() <= paneRect.getBottom())
+        const Util::Rectangle paneRect = getNormalizedVisiblePaneArea(panes[i]);
+        if( tile.intersects( paneRect ) )
```
mmeeks, line 3015: easier to read I suppose; I wonder why we never did that - I suspect this method is called a lot; what does the history say ?

mmeeks (changes requested):
Thanks, some useful cleanups here perhaps - though the algorithm used doesn't seem to match what was asked for. Let me paste that again:

a) only render & send those that are inside a visible area and
b) re-spin the main-loop  and
c) only render non-visible tiles when there is nothing else to do =)

Was there a problem with detecting nothing else to do in c) - I would expect that is reasonably easy to detect - when we found nothing to read from our remote socket, and had a non-zero timeout to poll for.

"an attempt to reduce bandwidth to client" - this is primarily a way of reducing latency, by not spending time rendering not-yet-seen tiles - and prioritizing visible tiles over non-visible ones. Nothing to do with bandwidth in particular.

mmeeks:
It would be good to cleanup the ordering functionality at the same time though eg.

// FIXME: it's not that clear what good this does for us ...
// we process all previews in the same batch of rendering
void KitQueue::deprioritizePreviews()

Around thumbnailing slides - while ensuring that whatever algorithm we use doesn't create starvation problems.

common/Rectangle.hpp
```diff
@@ -63,39 +85,49 @@ struct Rectangle
             _y2 = rectangle._y2;
     }
 
-    void setLeft(int x1) { _x1 = x1; }
+    constexpr void setLeft(int x1) /*noexcept*/ { _x1 = x1; }
```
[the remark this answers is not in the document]
sgothel: I obeyed the `noexcept` rule, just wanted to annotate for `constexpr`.
Will drop this annotation.

Its is correct, that the compiler very highly likely will optimize all calls to become inline.
However, `constexpr` explicitly does
- enforced inline (`inline` itself is useless)
- enforced check/removal of UB (undefined behavior)
- allow compile time if possible (depending on data source, ...)
  - would require `noexcept` - hence the annotation (to be dropped).

Boils down to the difference of `can` and `must` + perhaps UB.

If you insist of dropping the `must inline` property of `constexpr` in general, I will ofc.

common/Rectangle.hpp
sgothel: WhiteBoxTests.cpp does validate Rectangle intersects after consolidation via TileDesc::rectanglesIntersect

sgothel:
> It would be good to cleanup the ordering functionality at the same time though eg.
> 
> // FIXME: it's not that clear what good this does for us ... // we process all previews in the same batch of rendering void KitQueue::deprioritizePreviews()

Looking at it, it only pushes _leading_ preview-tiles to the end of the queue
and you mentioned in the comment that a tile-req batch either only consist out of 
such preview-tiles or none shall be included(?).

Our IDLE/invisible mechanism act above the produced 'popWholeTileQueue'
on the produced combined tiles, i.e. included this preview reordering.
Not sure if this is the discussion here.

> 
> Around thumbnailing slides - while ensuring that whatever algorithm we use doesn't create starvation problems.

Since out IDLE/invisible mechanism includes the preview-tile order, see above,
this should be well included in the solution.

sgothel:
Offered the optional Tile/TileCombined intersection test patch via
https://github.com/CollaboraOnline/online/pull/10197

sgothel:
COOL 24.04 Cypress (desktop) — FAILURE:
Passed locally: make check-desktop spec=writer/invalidations_spec.js

kit/Kit.cpp
```diff
@@ -2806,20 +2829,32 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
...
+            else // poll error
                 break;
 
-            const auto now = std::chrono::steady_clock::now();
-            drainQueue();
+            drainQueue(isIDLE);
 
-            timeoutMicroS =
-                std::chrono::duration_cast<std::chrono::microseconds>(_pollEnd - now).count();
             ++eventsSignalled;
```
caolanm, line 2858: This ++eventSignalled line wasn't incremented in the before-change case of a poll returning 0, but it is in the after-case (similarly timeoutMicroS is set to 0) and eventSignalled and timeoutMicroS are used after the loop, so that would be a change. Do we want that?

Would there be any real difference in intent, on, say, just passing false to the first drainQueue inside the loop and passing an isIDLE to the trailing drainQueue outside the loop, where isIDLE is set when poll is passed a non-zero timeout and returns 0 ?

sgothel:
rebased to master a330f9e75717de96c72794d0ac672f75376c9711 (no changes)

sgothel:
re CodeQL see https://github.com/CollaboraOnline/online/pull/9916#issuecomment-2415664161

kit/Kit.cpp
```diff
@@ -2300,7 +2300,7 @@ void Document::drainCallbacks()
         _websocketHandler->flush();
 }
 
-void Document::drainQueue()
+void Document::drainQueue(bool isIDLE)
```
mmeeks, line 2303: Why is IDLE capitalized. Parameter makes some sense: 'isIdle' is reasonable.

kit/Kit.cpp
```diff
@@ -2822,15 +2845,24 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
             if (_document && _document->needsQuickPoll())
                 realTimeout = 0;
 
-            if (poll(std::chrono::microseconds(realTimeout)) <= 0)
+            const int rc = poll(std::chrono::microseconds(realTimeout));
```
mmeeks, line 2848: Not convinced by this. What 'idle' are we detecting here. We can tell when the core main-loop is idle from a non-zero timeout; that's fine - how do we detect our own ie. coolforkit idle-ness ? presumably with a very short poll timeout - is that what we do ? confused here. Looks like we conclude that if we waited a long time and timed-out, we are now idle - and now we can do the things we wanted to do but were delayed; but why delay them ? 

Confused.

mmeeks (changes requested):
Should prolly have a call here.

sgothel:
PR #10183 as well as PR #10512 were rebased (forced pushed) twice
- first rebase
  - fixed to the loIdle determination
  - Document::drainQueue returns DocDrainStats for logging details, 
    also reducing code duplication in KitSocketPoll::kitPoll
- second rebase
  - rebase to master

kit/Kit.cpp
```diff
@@ -2441,17 +2443,39 @@ void Document::drainQueue()
             }
         }
 
+        const size_t tilesQueued = _queue->getTileQueueSize();
         if (processInputEnabled() && !isLoadOngoing() &&
-            !isBackgroundSaveProcess() && _queue->getTileQueueSize() > 0)
+            !isBackgroundSaveProcess() && tilesQueued > 0)
         {
+            // Render all tiles if idle.
+            // Otherwise render only the first visible TileCombined, but at least one TileCombined (if non are visible).
```
caolanm, line 2451: trivial nit: non->none

caolanm (commented):
I guess the obvious question then is if https://github.com/CollaboraOnline/online/pull/10512 just "tidies things up" or if this pr on its own isn't sufficient and the other is needed to work properly?

sgothel:
> I guess the obvious question then is if #10512 just "tidies things up" or if this pr on its own isn't sufficient and the other is needed to work properly?

Yes. The idle behavior in discussion may even render idle determination useless. In such case we wouldn't care and just sent the invisible tiles at last and one tile per sliced-loop from KitPoll. 
Otherwise, having the idleHint helps to exclude certain idle-determinations when LO is busy, i.e. claiming high-priority.

sgothel:
Updated based on timeout (remaining time) only, disregarding potential open-ended idle/wakeup case ([note](https://github.com/CollaboraOnline/online/pull/10512#issuecomment-2487631933))
> Due to ambiguous timeout==0 semantics w/o priority (idle or intentional zero timeout),
> we are not able to determine general idleness w/o passing this information down to COOL.
> (Post reviewing LO's Task, Timer and Scheduler, confirmed by vcl/README.scheduler.md)

Updated description accordingly.

mmeeks:
Abandon this for now.

PR #10184, opened 2024-10-07
cool#9992 doc sign: set the view's sign cert from the WOPI UserPrivateInfo

vmiklos:
To be able to sign documents, at least the 'local file' WOPI provider
sets 3 keys under UserPrivateInfo to give the wsd, and then the kit
process the signing certificate.

The other end is the core side, which takes 3 keys as a view options in
initializeForRendering().

What was missing is to connect these two: this way a 'make run' session
can sign documents without hardcoding cert/key/ca PEM files anywhere.

The integration side and cypress tests are not yet done here.

Change-Id: I1f847349e10336c69019c7f1747208e6be954362

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1803,6 +1803,7 @@ std::shared_ptr<lok::Document> Document::load(const std::shared_ptr<ChildSession
     const std::string& macroSecurityLevel = session->getMacroSecurityLevel();
     const bool accessibilityState = session->getAccessibilityState();
     const std::string& userTimezone = session->getTimezone();
+    const std::string& userPrivateInfo = session->getUserPrivateInfo();
 
     if (!Util::isMobileApp())
         consistencyCheckFileExists(uri);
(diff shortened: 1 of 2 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1424/console fails in:

>   │ ✖  writer/file_properties_spec.js           00:25        4        -        1        3        - │

But

```
make -C cypress_test check-desktop spec=writer/file_properties_spec.j
```

passes for me locally, retry.

vmiklos:

With this, the 'make run' case works for me, following the instructions at https://github.com/CollaboraOnline/online/issues/9992#issuecomment-2396115427. An actual integration still needs a bit of work, I see that we have an enable/disable flag for Zotero in the "capabilities", I need to look if that's relevant for signing.

caolanm (approved):
seems reasonable to me

PR #10013, opened 2024-09-09
slideshow rework: presentation info changed on playing slideshow

mcecchetti:
handle presentation info has changed message
request for updated presentation info
clean current status
restart presentation from last displayed slide if it still exists from
first slide otherwise

make SlideSHowNavigator.quit to be invoked on closing the presentation
window through the window close button.

Signed-off-by: Marco Cecchetti <marco.cecchetti@collabora.com>
Change-Id: I72cf065056245b2c98ff5474be7131235e9683da

this needs core patch: https://gerrit.libreoffice.org/c/core/+/173061

PR #10186, opened 2024-10-07
Impress: Fixes invisible typing issue on notes view.

Minion3665:
This is a backport of https://github.com/CollaboraOnline/online/pull/10026, it's only against 24.04-mobile as while it does improve things it also causes a regression - so we can't put it in our fully stable release, but hopefully it should still make things better enough to justify it being in this mobile release. 

_clientSelectedMode is set after the first tiles invalidation.
So this commit sets it beforehand.

See Core side ViewShellBase.cxx -> getEditMode for more info.

Change-Id: Iddd2e0b235cc754f850223f2aff36ba6fc2aeb85* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -3213,7 +3213,8 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
             else if (tokens.size() == 3 && tokens.equals(0, "EMPTY"))
             {
                 // with mode:    "EMPTY, <part>, <mode>"
-                const std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
+                // Writer renders everything as part 0.
+                const std::string part = (_docType != "text" ? (_docType == "presentation" ? std::to_string(getLOKitDocument()->getPart()) : tokens[1].c_str()) : "0");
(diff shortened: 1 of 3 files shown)


Minion3665:
CI failure appears to be as a result of https://github.com/CollaboraOnline/online/pull/10013, which is required by a core patch ... but as it's CODE Red I'm "not keen" to merge it in ... please stand by while I figure out what to do :)

Minion3665:
The solution here is going to be "merge without CI" as any other fix will be, I think, prohibitive... it's a pity, but sometimes it's what you have to do

eszkadev:
@Minion3665 this was causing bug in impress slide preview bar. Please see comment: https://github.com/CollaboraOnline/online/pull/10026#discussion_r1766901202

@gokaysatir did we fix that? I don't remember now..

Minion3665:
> @Minion3665 this was causing bug in impress slide preview bar. Please see comment: https://github.com/CollaboraOnline/online/pull/10026#discussion_r1766901202
> 
> @gokaysatir did we fix that? I don't remember now..

I saw that, my impression from comments around this was that it did cause a regression, but was also an improvement overall ... do we believe this to not be the case?

PR #10189, opened 2024-10-08
cool#9992 doc sign: add a setting to be able to disable this

vmiklos:
Once an integration would provide the signing certificate/key, it's
necessary to configure this on the user settings page of the
integration. Given that document signing is not relevant for many
deployments, the integration may want to hide this.

The way this is done for the Zotero case is that coolwsd has a flag
(enabled by default, the sysadmin can disable it), and then the
/hosting/capabilities JSON informs the integration if this COOL instance
supports document signing or not.

Fix the problem by adding a similar flag for document signing.

At the moment this influences our capabilities JSON & disables the
JS UI buttons; the JS UI button doesn't check if the view actually
provides a signing certificate yet.

Change-Id: Icb7a8b438374d9244e4513e5488fb88bbc46f323

### DIFF
--- a/browser/html/cool.html.m4
+++ b/browser/html/cool.html.m4
@@ -257,6 +257,7 @@ m4_ifelse(MOBILEAPP, [true],
       data-check-file-info-override = "%CHECK_FILE_INFO_OVERRIDE%"
       data-deepl-enabled = "%DEEPL_ENABLED%"
       data-zotero-enabled = "%ZOTERO_ENABLED%"
+      data-document-signing-enabled = "%DOCUMENT_SIGNING_ENABLED%"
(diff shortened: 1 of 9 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1462/console failed in:

>   │ ✖  writer/invalidations_spec.js             00:36        3        2        1        -        - │

but

```
make -C cypress_test check-desktop spec=writer/invalidations_spec.js
```

passes for me locally, retry.

vmiklos:

I expect the richdocuments side can be started once this is in.

caolanm (approved):
looks reasonable to me

Issue #9281, opened 2024-06-13
Copy and Paste functions are broken


gravelfreeman:
# Chomium and Firefox browsers

### Describe the Bug
It's not possible to copy content from within the Text, Calc and Prensentation apps to the clipboard.

### Steps to Reproduce
1. Copy some text in Windows clipboard
2. Paste text in Text, Calc and Prensentation

### Expected Behavior
Text gets copied in Text, Calc and Prensentation.

### Actual Behavior
Text doesn't get copied in Text, Calc and Prensentation.

***

# Chromium browsers only

### Describe the Bug
It's not possible to paste content from the clipboard into Text, Calc and Prensentation apps.

### Steps to Reproduce
1. Copy some text in Windows clipboard
2. Paste text in Text, Calc and Prensentation

### Expected Behavior
Text gets copied from Text, Calc and Prensentation to the clipboard.

### Actual Behavior
Nothing gets added to the clipboard.

***

### Desktop
```
COOLWSD version: 24.04.3.1 (git hash: 594b605 (E))
LOKit version: Collabora Office 24.04.3.1 (git hash: 31b6e2c)
Served by: Debian GNU/Linux 12 (bookworm)
```

### Additional Context
First shared on Nextcloud [forum ](https://help.nextcloud.com/t/copy-paste-issue-with-code-maybe-related-to-http-headers/195011)where I've been told to post my issue here.

mulzhang:
I am also facing this tricky problem now, please solve it as soon as possible. Thank you

HLFH:
I can reproduce the issue on Safari.

timur-g:
There is no general Copy and Paste bug. If it is not working for you, you need to test more to figure out what exactly is not working: (simple text, complex text with "Download to clipboard" button), are there errors in browser console, are you behind some proxy... 
This report was initially Windows. The other user said "me too" wihtout specifics, so we may only assume that is the same. 
Safari may be different issue, we do not know now unless this is pinpointed.

pedropintosilva:
Thank @gravelfreeman for reporting this and @mulzhang . Also thanks @HLFH for posting additional info in https://github.com/nextcloud/richdocuments/issues/3775#issuecomment-2191270588 WRT to version. 

[...]

PR #10190, opened 2024-10-08
Modify ClipboardApi checks.

gokaysatir:
Modify ClipboardApi checks.

Merge read and write permission checks.
Rename the cookie name to "clipboardApiAvailable".

Tested on FF 131 flatpack version.

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -40,32 +40,32 @@ class BrowserProperties {
 	static initiateBrowserProperties(global) {
 		global.L = {};
 
-		var ua = navigator.userAgent.toLowerCase(),
-		uv = navigator.vendor.toLowerCase(),
-		doc = document.documentElement,
+		let ua = navigator.userAgent.toLowerCase();
+		let uv = navigator.vendor.toLowerCase();
+		let doc = document.documentElement;
 
-		ie = 'ActiveXObject' in global,
+		let ie = 'ActiveXObject' in global;
 
-		cypressTest = ua.indexOf('cypress') !== -1,
+		let cypressTest = ua.indexOf('cypress') !== -1;
+
 		// Firefox has undefined navigator.clipboard.read and navigator.clipboard.write,
 		// unsecure contexts (such as http + non-localhost) has the entire navigator.clipboard
 		// undefined.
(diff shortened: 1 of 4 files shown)


gokaysatir:
Possibly fixes: https://github.com/CollaboraOnline/online/issues/9281

browser/js/global.js
```diff
@@ -904,11 +905,6 @@ function getInitializerClass() {
 
 	};
 
-	if (!global.prefs.getBoolean('hasNavigatorClipboardWrite', true)) {
```
vmiklos, line 907: This more or less reverts a161ee30f471c5cd00e17428bd850c2a8604fe32, which was a fix for https://github.com/CollaboraOnline/online/issues/9072, this would mean that e.g. in ownCloud richdocuments (or any other non-updated integration) copy would fail for the first time you edit a document, and only the second copy would work.

The upside would be that in case the integration was broken once, but fixed today, it would start to work.

Do you think that's an acceptable trade-off? It feels a bit brave.

May better to keep the setting, but rename the key name, so we give a one-off chance for updated integrations?

browser/src/map/Clipboard.js
```diff
@@ -911,7 +911,6 @@ L.Clipboard = L.Class.extend({
 			that._warnCopyPaste();
 			// Once broken, always broken.
```
vmiklos, line 912: This comment would be outdated with the code change.

vmiklos (commented):
Can you reproduce the problem you try to fix?

If this "maybe" fixes the new problem but definitely regresses regarding the old use-case, then perhaps this PR is too brave. I'm not totally against it, but let's understand the effect of this change -- is it OK to upset all integrations which were not updated recently? :-)

gokaysatir:
> Can you reproduce the problem you try to fix?
> 
> If this "maybe" fixes the new problem but definitely regresses regarding the old use-case, then perhaps this PR is too brave. I'm not totally against it, but let's understand the effect of this change -- is it OK to upset all integrations which were not updated recently? :-)

I reproduced the issue and this fixes it.

vmiklos (approved):
Yes, this looks like a much better trade-off to me, thanks. :-)

PR #10191, opened 2024-10-08
improve closing the calc with unsaved changes

eszkadev:
Warn on closing when have cells under edit
    
    When we edit a cell and we don't confirm by click outside
    or the Enter key - we can lost the typed text when window
    is closed. To avoid that let's introduce native browser warning
    before leaving the page. We also need to catch when we exit
    using commands as we are embedded in the iframe and not all
    cases trigger the beforeunload event.
    
    Also when we lose focus due to switch to other tab we will
    accept the content.

New behavior: accept content on tab switch, ask when closing the tab

[accept-2024-10-08_11.31.50.webm](https://github.com/user-attachments/assets/41caddc1-4b0b-4351-ba5c-8b5d939e5533)

### DIFF
--- a/browser/src/control/Control.FormulaBarJSDialog.js
+++ b/browser/src/control/Control.FormulaBarJSDialog.js
@@ -241,9 +241,11 @@ class FormulaBar {
 		// in the core we have DrawingArea not TextView
 		if (object.id.indexOf('sc_input_window') === 0) {
 			objectType = 'drawingarea';
-			if (eventType === 'keypress' && data === UNOKey.RETURN || data === UNOKey.ESCAPE)
+			if (eventType === 'keypress' && data === UNOKey.RETURN || data === UNOKey.ESCAPE) {
+				// TODO: not used anymore? now acceptformula dispatch command is executed instead
 				builder.map.focus();
-			else if (eventType === 'grab_focus') {
+				app.allowExit();
+			} else if (eventType === 'grab_focus') {
 				this.focusField();
 				builder.map.onFormulaBarFocus();
 			}
@@ -254,10 +256,12 @@ class FormulaBar {
 
 	focusField() {
 		L.DomUtil.addClass(this.getInputField(), 'focused');
+		app.preventExit();
 	}
 
 	blurField() {
 		L.DomUtil.removeClass(this.getInputField(), 'focused');
+		app.allowExit();
 	}
 
 	enable() {
--- a/browser/src/docdispatcher.ts
+++ b/browser/src/docdispatcher.ts
@@ -41,6 +41,11 @@ class Dispatcher {
 			if ((window as any).ThisIsAMobileApp) {
 				window.postMobileMessage('BYE');
 			} else {
+				if (window.onbeforeunload) {
+					alert(app.getUnsavedCloseMessage());
+					return;
+				}
+
 				app.map.fire('postMessage', {
 					msgId: 'close',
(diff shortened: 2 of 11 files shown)


browser/src/control/Control.FormulaBarJSDialog.js
```diff
@@ -254,10 +256,12 @@ class FormulaBar {
 
 	focusField() {
 		L.DomUtil.addClass(this.getInputField(), 'focused');
+		app.preventExit();
 	}
 
 	blurField() {
 		L.DomUtil.removeClass(this.getInputField(), 'focused');
+		app.allowExit();
 	}
```
mcecchetti, line 265: why do we want to allow exit independently by the event that lead to losing focus ? I mean we don't know if the blurring is due to an accepted/canceled formula.

browser/src/docdispatcher.ts
```diff
@@ -303,13 +324,15 @@ class Dispatcher {
 			app.map.onFormulaBarBlur();
 			app.map.formulabarBlur();
 			app.map.formulabarSetDirty();
+			app.allowExit();
```
mcecchetti, line 327: `allowExit()` is already called by fieldBlur() which is invoked by onFormulaBarBlur()
anyway,  I added comments on the fieldBlur() definition, too

browser/src/docdispatcher.ts
```diff
@@ -303,13 +324,15 @@ class Dispatcher {
...
 		};
 
 		this.actionsMap['cancelformula'] = function () {
 			app.map.sendUnoCommand('.uno:Cancel');
 			app.map.onFormulaBarBlur();
 			app.map.formulabarBlur();
 			app.map.formulabarSetDirty();
+			app.allowExit();
```
mcecchetti, line 335: as above

browser/src/docstatefunctions.js
```diff
@@ -150,3 +150,15 @@ app.isExperimentalMode = function () {
...
 };
+
+app.getUnsavedCloseMessage = function () {
+	return _('You have unsaved changes. Are you sure you want to quit?');
+};
+
+app.preventExit = function () {
+	window.onbeforeunload = app.getUnsavedCloseMessage;
+};
+
```
mcecchetti, line 161: what about just using true/false ? :-)

browser/src/control/Control.FormulaBarJSDialog.js
```diff
@@ -254,10 +256,12 @@ class FormulaBar {
 
 	focusField() {
 		L.DomUtil.addClass(this.getInputField(), 'focused');
+		app.preventExit();
```
mcecchetti, line 259: So if I got it right exiting is prevented even if the user didn't modify the formula bar content ?

browser/src/docstatefunctions.js
eszkadev: `Setting the event object's [returnValue](https://developer.mozilla.org/en-US/docs/Web/API/BeforeUnloadEvent/returnValue) property to a non-empty string value or any other [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value.`

String can be used by some browsers to be shown in the popup.

eszkadev:
I tried various ways here, but it was extremely hard to solve as we didn't keep focus in formulabar for some cases.
I prepared simplified PR with reduced impact and fixing that focus problem instead: https://github.com/CollaboraOnline/online/pull/10196

This PR will be partly reused later

PR #10192, opened 2024-10-08
Reposition the URL PopUp element when it's off the borders.

gokaysatir:
Change-Id: I9f0900609fa3001230898f2056d03b3a2f6831a9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mcecchetti (approved):
it makes sense

gokaysatir:
Thanks for the review :) Passes the tests locally, retrying.

eszkadev:
cypress fail seems to be valid error:
```
Copy Hyperlink from pop-up
```

PR #10193, opened 2024-10-08
Follow up pr for 10144 (Save notification )

Darshan-upadhyay1110:
### Summary

- Fix typos in comment messages & correct the wrong comment message with right meaningful messages
- Dynamically set CSS content for saving and saved states
- Add padding-inline-start to retain margin when logo is removed
- fix: Set line-height to 0 to prevent cropping of "Saving" text

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -89,7 +89,7 @@ nav.hasnotebookbar #save::after {
 }
 
 nav.hasnotebookbar #save.saving::after {
-	content: 'Saving';
+	content: var(--save-state);
 	font-size: var(--default-font-size);
 	font-family: var(--cool-font);
 	color: var(--color-main-text);
@@ -101,7 +101,7 @@ nav.hasnotebookbar #save.saving::after {
 }
 
 nav.hasnotebookbar #save.saved::after {
-	content: 'Saved';
+	content: var(--save-state);
 	color: #309048;
 }
 
--- a/browser/css/menubar.css
+++ b/browser/css/menubar.css
@@ -73,7 +73,7 @@
 	font-family: var(--cool-font);
 	position: relative;
 	background: transparent;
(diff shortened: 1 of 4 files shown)


pedropintosilva:
this PR fixes many things. thanks @Darshan-upadhyay1110 ! 

Additionally I have noticed by looking again to the code and then testing it that sometimes the manual save (either by clicking in the icon or ctrl+s) works and the document is saved but it never changed the icon to saved state because the document didn't have unsaved changes so, it never executed the stuff inside of the following if statement:

`		if (saveEle.classList.contains('saving')) {` https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L1196

Even if there is no unsaved changes we do save the document and thus we should also change the button momentarily to `Saved` status.

Darshan-upadhyay1110:
> this PR fixes many things. thanks @Darshan-upadhyay1110 !
> 
> Additionally I have noticed by looking again to the code and then testing it that sometimes the manual save (either by clicking in the icon or ctrl+s) works and the document is saved but it never changed the icon to saved state because the document didn't have unsaved changes so, it never executed the stuff inside of the following if statement:
> 
> `if (saveEle.classList.contains('savemodified')) {`
> 
> which makes it to never go inside of
> 
> ` if (saveEle.classList.contains('saving')) {` https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.UIManager.js#L1196
> 
> Even if there is no unsaved changes we do save the document and thus we should also change the button momentarily to `Saved` status.

While doing changes for above suggestion I found a new case as well which we need to consider,

- Suppose if "saving"  is on going and at the same time someone changed the document then we should stop saving and make  the icon as save modified state...

Darshan-upadhyay1110:
Oh It is merged :)

PR #10194, opened 2024-10-08
Add a check when getting view name.

gokaysatir:
Or it may cause JS errors if user doesn't have a name.

Guest users may cause this issue. I couldn't reproduce the issue, adding the check reading the logs.

Change-Id: I1b9ab5b90e95b99712d6d5acd8e05b2f15190f99

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mcecchetti (approved):
it makes sense

PR #10195, opened 2024-10-08
Update DOM when saving/saved state is interrupted by document modification

Darshan-upadhyay1110:
- Perform DOM operations to handle cases where the document is modified during a saving or saved state.
- If modifications occur, the save icon is reverted to the "modified" state to reflect the updated document status.

Change-Id: I498128b3a3eee70ef638b472a8fda774d7f6ab92

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -303,6 +303,9 @@ L.Control.Notebookbar = L.Control.extend({
 		}
 
 		this.builder.build(shortcutsBar, shortcutsBarData);
+
+		//create SaveState object after addition of shortcut bar in UI
+		this.map.saveState = new app.definitions.saveState(this.map);
 	},
 
 	reloadShortcutsBar: function() {
--- a/browser/src/control/Control.NotebookbarBuilder.js
+++ b/browser/src/control/Control.NotebookbarBuilder.js
@@ -165,13 +165,8 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 		else if (commandName === '.uno:ModifiedStatus') {
 			const saveEle = document.getElementById('save');
 			if (saveEle) {
-				if (state === 'true') {
-					saveEle.classList.remove('saving');
-					saveEle.classList.remove('saved');
-					const saveIconEl = document.querySelector('#save img');
-					saveIconEl.classList.remove('rotate-icon');
-					saveEle.removeAttribute('disabled');  // Enable the button
-					document.getElementById('save').classList.add('savemodified');
+				if (state === 'true' &&  this.map.saveState) {
+					this.map.saveState.showModifiedStatus();
 					document.getElementById('file-save').classList.add('savemodified');
 				} else {
 					document.getElementById('save').classList.remove('savemodified');
--- a/browser/src/control/Control.SaveState.ts
+++ b/browser/src/control/Control.SaveState.ts
@@ -1,0 +1,72 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+This file defines the SaveState class, which handles the logic for managing the document's save state.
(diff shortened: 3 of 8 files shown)


Darshan-upadhyay1110:
Steps to test patch

- Modify content and save it by keyboard shortcut
- while "Saving" please change the document content
- in this stage we need to revert back to "modified" state instead of make the process continue.

Normal case:

[Screencast from 2024-10-08 21-58-29.webm](https://github.com/user-attachments/assets/05ebfebb-15c9-461f-84a5-9fb46f17a33b)


When we do modify in between state change
[Screencast from 2024-10-08 21-59-42.webm](https://github.com/user-attachments/assets/ade41ed1-ebd6-44eb-b1c2-8ae244cbb80d)


CC: @pedropintosilva @eszkadev

browser/src/control/Control.NotebookbarBuilder.js
```diff
@@ -163,8 +163,14 @@ L.Control.NotebookbarBuilder = L.Control.JSDialogBuilder.extend({
 			$('#applystyle').val(state).trigger('change');
 		}
 		else if (commandName === '.uno:ModifiedStatus') {
-			if (document.getElementById('save')) {
+			const saveEle = document.getElementById('save');
+			if (saveEle) {
 				if (state === 'true') {
+					saveEle.classList.remove('saving');
+					saveEle.classList.remove('saved');
+					const saveIconEl = document.querySelector('#save img');
```
eszkadev, line 171: I see similar code repeated 3 times. Can we do a followup with cleanup - sharing.
Please create new file (in TS), eg. `browser/src/control/Control.SaveIcon.ts`
and define there methods related to the save icon which will be only called from other files.

browser/src/control/Control.UIManager.js
```diff
@@ -1196,7 +1196,7 @@ L.Control.UIManager = L.Control.extend({
 		if (window.mode.isMobile())
 			return;
 		const saveEle = document.getElementById('save');
-		if (saveEle.classList.contains('saving')) {
+		if (!saveEle.classList.contains('savemodified')) {
```
eszkadev, line 1199: to separate file

browser/src/control/Control.Toolbar.js
```diff
@@ -906,11 +906,18 @@ function processStateChangedCommand(commandName, state) {
 		}
 	}
 	else if (commandName === '.uno:ModifiedStatus') {
-		if (document.getElementById('save')) {
-			if (state === 'true')
-				document.getElementById('save').classList.add('savemodified');
+		const saveEle = document.getElementById('save');
```
eszkadev, line 909: to separate file

browser/src/control/Control.NotebookbarBuilder.js
Darshan-upadhyay1110: Thanks, @eszkadev. I am working on it...

Minion3665 (dismissed):

browser/src/control/Control.Notebookbar.js
```diff
@@ -303,6 +303,9 @@ L.Control.Notebookbar = L.Control.extend({
 		}
 
 		this.builder.build(shortcutsBar, shortcutsBarData);
+
+		//create SaveState object after addition of shortcut bar in UI
+		this.map.saveState = new app.definitions.saveState(this.map);
```
eszkadev, line 308: what about compact mode? do we want it there too?
If not - do we check if saveState exists before use? so in case it is missing in compact we don't generate errors

Darshan-upadhyay1110: > what about compact mode? do we want it there too?

Yes we do

> If not - do we check if saveState exists before use?

Still just for a safe check i did add checks for saveState

> so in case it is missing in compact we don't generate errors

I am adding a new initialize for `saveState` in case if app directly opens in compact mode :)

browser/src/control/Control.SaveState.ts
```diff
@@ -0,0 +1,74 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+This file defines the SaveState class, which handles the logic for managing the document's save state.
+It controls the display of the saving status, saved status, and triggers the associated animations and icon changes.
+*/
+
+namespace cool {
```
eszkadev, line 17: please do not use namespaces, it will be hard to use ES modules later

eszkadev (approved):
code looks ok

eszkadev:
please rebase, test was fixed now in master

PR #10196, opened 2024-10-08
calc: save edited cell on exit

eszkadev:
This fixes the main problem which is data loss when we
exit without confirmation of a cell content but also
problem with losing focus from formulabar on some actions.
Especially when we click outside iframe - we lost
focus on formulabar but we were still catching input to
it on the server side.

### DIFF
--- a/browser/src/control/Control.FormulaBarJSDialog.js
+++ b/browser/src/control/Control.FormulaBarJSDialog.js
@@ -283,6 +283,13 @@ class FormulaBar {
 		return L.DomUtil.hasClass(input, 'focused');
 	}
 
+	isInEditMode() {
+		var acceptButton = this.parentContainer.querySelector('#acceptformula');
+		if (acceptButton)
+			return !acceptButton.classList.contains('hidden');
+		return false;
+	}
+
 	showFormulabar() {
 		if (this.parentContainer)
 			this.parentContainer.style.setProperty('display', 'table-cell');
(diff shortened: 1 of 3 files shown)


eszkadev:
I need to fix tests. Focus should be handled somewhere else.

eszkadev:
Now it seems to be fixed. I solved problem with losing focus what was causing issues with detecting when we have editing active...

mcecchetti:
Question: has formula bar focus when a cell is edited in place ?
I'm asking since the patch takes care of the formula bar focus status, only.

eszkadev:
> Question: has formula bar focus when a cell is edited in place ? I'm asking since the patch takes care of the formula bar focus status, only.

@mcecchetti thanks, added also this case

browser/src/control/Control.FormulaBarJSDialog.js
```diff
@@ -283,6 +283,13 @@ class FormulaBar {
 		return L.DomUtil.hasClass(input, 'focused');
 	}
 
+	isInEditMode() {
+		var acceptButton = this.parentContainer.querySelector('#acceptformula');
+		if (acceptButton)
+			return !acceptButton.classList.contains('hidden');
```
mcecchetti, line 289: `return true` ?

mcecchetti (approved):
Okay, seems to work.
Maybe if someone switch to another window/app and then back it could be a bit unexpected that on typing instead of appending new text, it overwrites the current content.

PR #10199, opened 2024-10-09
remove mount_namespaces config option

timar:
During development using of Linux Namespaces was an experimental feature. Therefore it was disabled in production builds, and enabled only in development edition (CODE). Later, as it matured, it was on by default. This change caused problems with upgrade, because when a user had mount_namespaces=false in /etc/coolwsd/coolwsd.xml, it remained false. Probably it is better to get rid of this configuration option, and try to use Linux Namespaces unconditionally. If it's not possible, the code will fall back to use coolforkit-caps and coolmount that require capabilities. And of course if these are not installed (coolwsd-deprecated package), a log entry will be made. It is assumed that this is a rare case. It is also assumed that nobody would want to disable using of Linux Namespaces, when the feature is available on the system.

**If you think this patch makes sense, I'd like to have it in COOL 24.04.8.3, too.**

Change-Id: Iabf23198043baa1c4b8d50e82212ddf9e17ddad0

* Resolves: # 
* Target version: master and distro/collabora/co-24.04


caolanm (approved):
Simple and direct. I like it.

PR #10205, opened 2024-10-10
Share code for close action

eszkadev:


eszkadev:
or better to have already: https://github.com/CollaboraOnline/online/pull/10206

PR #10206, opened 2024-10-10
calc: edited cell on exit of the only user

eszkadev:
calc: edited cell on exit of the only user
    
steps to reproduce:
1. open calc
2. type something in the formulabar but don't accept it (no enter)
3. close the app using "X"

result was: we didn't save cell's content


gokaysatir (approved):
Looks good, thanks.

PR #10212, opened 2024-10-11
jsdialog: transparent color apply

eszkadev:
1. open writer
2. select some default formatted text
3. click on left side of background/highlight button (not arrow)
result: become black
expected: be still transparent


gokaysatir (approved):
Thanks for the fix :)

PR #10214, opened 2024-10-11
Fix search focus when cursor goes into table

eszkadev:
we lost focus when found text was inside table - because context switch was changing the tab inside notebookbar which explicitly focused map


Darshan-upadhyay1110 (approved):
Tested working good. Thanks @eszkadev

Darshan-upadhyay1110:
With this patch working good...

[Screencast from 2024-10-11 21-10-50.webm](https://github.com/user-attachments/assets/6f2f64fc-21f7-4f87-9c02-d639ff8da755)

PR #10215, opened 2024-10-11
sc: change how .uno:FreezePanesColumn/Row get the index

JaumePujantell:
On commit "browser: fix freeze row/column panes"
fe45327ba0e8b330dfa5ea99e6247ed7c9abefad the type of FreezePanesColumn
and FreezePanesRow was changed from Int32 to Point. Since then, calls
trying to set the frozen panes with this commands sending an index fail,
setting the index to 1.

This change allows again a way to set them passing only an index of type
Int32.

Change-Id: I9837586f6c531bf885f24b5913af9a372dc0f5a1

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -915,7 +915,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		var unoName = isSplitCol ? 'FreezePanesColumn' : 'FreezePanesRow';
 		var command = {};
-		command[unoName] = {
+		command['Index'] = {
 			type: 'int32',
 			value: newSplitIndex
 		};


vmiklos (approved):
The change itself works for me, but please next time either file a github issue or include the steps to reproduce the problem in the commit message.

What works for me as a reproduce is:

1) Load the hello world Calc document into COOL, I get up to col R and row 41 on the screen.
2) Go to the top left corner, click on View -> Freeze Rows and Columns on the notebookbar.

Expected result: there is a limit on the the amount of freezed content, about 70% of your display, so you get a freeze at col M and row 29.

Actual result: you get a freeze at col A and row 1.

Also, did you consider writing a cypress test for this? cypress_test/integration_tests/desktop/calc/cell_cursor_spec.js seems to have a related split panes test case already. Would be nice to do that in a follow-up change.

PR #10216, opened 2024-10-11
Minor fixes for iOS build breakages

plubius:
The missing functions are in #if !MOBILEAPP blocks so used the same for iOS. This PR mabye should be tested on Android.

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -225,6 +225,7 @@ public:
 #if !MOBILEAPP
     /// Uses peercreds to get prisoner PID if present or -1
     int getPid() const;
+#endif
 
     /// Sets the kernel socket send buffer in size bytes.
     /// Note: TCP will allocate twice this size for admin purposes,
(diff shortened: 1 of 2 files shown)


net/Socket.hpp
```diff
@@ -390,9 +390,11 @@ class Socket
 #if ENABLE_DEBUG
             if (std::getenv("COOL_ZERO_BUFFER_SIZE"))
             {
+#if !MOBILEAPP
```
Minion3665, line 393: I wonder if we should instead stub out the functions that are causing the issue here... it looks like 8921e19d844da1f4f2c254edc66710120b5a6bfc and 3ba713aabf20279c5d83d039e5a7ac906b79c783 were specifically to try to remove MOBILEAPP directives, replacing them with Util::isMobileApp (also relevant could be 9afde4af160319b407b6ae82212eb84f9d3dac26 which adds the constexprs)

Minion3665 (changes requested):

net/Socket.hpp
plubius: I've reworked this pull request to make the functions stubs. Let me know if you see any additional changes that I should try.

Minion3665 (approved):
Thanks very much! I've tested this on Android too, and it seems good to me

PR #10219, opened 2024-10-14
Private/gokay/notes previews

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/ImpressTileLayer.js
+++ b/browser/src/layer/tile/ImpressTileLayer.js
@@ -61,6 +61,12 @@ L.ImpressTileLayer = L.CanvasTileLayer.extend({
 			else
 				targetElement.classList.remove('selected');
 		}
+
+		if (['DrawPage', 'NotesPage'].includes(e.detail.context)) {
+			this._selectedMode = e.detail.context === 'NotesPage' ? 2 : 0;
+			this._refreshTilesInBackground();
+			this._update();
+		}
 	},
 
 	_isPCWInsideFlex: function () {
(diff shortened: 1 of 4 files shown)


wsd/ClientSession.hpp
```diff
@@ -50,12 +50,6 @@ class ClientSession final : public Session
                WAIT_DISCONNECT // closed and waiting for Kit's disconnected message
     );
 
-    enum ViewMode {
```
eszkadev, line 53: this is useful, can we reintroduce that enum? :)

browser/src/layer/tile/ImpressTileLayer.js
```diff
@@ -61,6 +61,12 @@ L.ImpressTileLayer = L.CanvasTileLayer.extend({
 			else
 				targetElement.classList.remove('selected');
```
eszkadev, line 62: Unrelated but I noticed: an ugly state handler - we could do it automatically with state update for an uno command. Do we send notes command state from core? we should.

eszkadev (approved):

browser/src/layer/tile/ImpressTileLayer.js
gokaysatir: Yeah but i think that's the way we use for toggle buttons. But still, we may consider a better approach as you say. I think it'd take some time so i am passing it for now :)

gokaysatir:
Thanks for checking :)

wsd/ClientSession.hpp
gokaysatir: Ah, yes, let me note this also. I will check these along with the JSON API.

PR #10225, opened 2024-10-14
Add viewMode property to TileLocation to ensure that correct delta is…

gokaysatir:
… replaced with the new one.

Change-Id: I376563989cf58fdbddd65f73315d4a898ceac93c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -196,7 +196,8 @@ namespace RenderTiles
                                                          tileRect.getTop(),
                                                          tileRect.getWidth(),
                                                          tileCombined.getPart(),
-                                                         canonicalViewId
+                                                         canonicalViewId,
+                                                         tileCombined.getEditMode()
                                                          ),
                                                      data, wireId, forceKeyframe, dumpTiles, mode);
                         }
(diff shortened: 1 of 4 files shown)


gokaysatir:
@caolanm hello, can you have a look when you have time?

caolanm (approved):
This looks reasonable to me. How did the problem manifest, wrong tiles on switching in and out of master view in presentations?

gokaysatir:
> This looks reasonable to me. How did the problem manifest, wrong tiles on switching in and out of master view in presentations?

Indeed i worked on a similar issue on Impress. I wanted to keep the previews on "standart" view while the page is at "notes" view. It was experimental. I couldn't succeed. This commit is for something i noticed while trying to understand the root cause.

gokaysatir:
The tile painting request is sent to the core side with the correct info. Core side checks the view mode before painting the tile. If the current view mode is different than the requested tile's view mode, core side temporarily switches the view mode. Everything seems normal and the rendered tile is sent to the browser with the correct view mode. But the content is wrong. Content is from the current view mode, not from the requested one.

I hoped that core side doing right and server has an issue and tried this and some other things. Didn't work but still wanted to prepare this commit in case it may be helpful.

Issue #10228, opened 2024-10-15
Stabilize C++ Unit Tests like unit-wopi-save-on-exit


sgothel:
### Describe the Bug
Certain C++ unit tests (ex Cypress) occasionally fail
and render the uni test framework instable / unreliable.

One of these tests is unit-wopi-save-on-exit.

### Steps to Reproduce
Run unit-wopi-save-on-exit until it fails, i.e. once within 20 - 80 loops.

### Expected Behavior
No error

### Actual Behavior
At least two different failures
- SIGABRT due to assertion within UnitWSD::get()
- Aborted unit test, giving up due to no incoming message

PR #10229, opened 2024-10-15
UT Stability: Resolve UnitBase data races

sgothel:
* Resolves: #10228 

### Summary
- Resolve UnitBase's _socketPoll data race via its socketPoll()
- Resolve UnitBase's UnitWSD::get() data-race

### Checklist

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -402,6 +402,9 @@ int UnitBase::uninit()
 
 std::shared_ptr<SocketPoll> UnitBase::socketPoll()
 {
+    // We could be called from either a UnitWSD::DocBrokerDestroy (prisoner_poll)
+    // or from UnitWSD::invokeTest() (coolwsd main).
+    std::lock_guard<std::mutex> guard(_lockSocketPoll);
     if (!_socketPoll)
         _socketPoll = std::make_shared<SocketPoll>(getTestname());
     return _socketPoll;
@@ -410,8 +413,9 @@ std::shared_ptr<SocketPoll> UnitBase::socketPoll()
 void UnitKit::postFork()
 {
     // Don't drag wakeup pipes into the new process.
-    if (_socketPoll)
-        _socketPoll->closeAllSockets();
+    std::shared_ptr<SocketPoll> socketPoll = getSocketPoll();
+    if (socketPoll)
+        socketPoll->closeAllSockets();
 }
 
 void UnitBase::initialize()
@@ -437,8 +441,9 @@ UnitBase::~UnitBase()
 {
     LOG_TST(getTestname() << ": ~UnitBase: " << (failed() ? "FAILED" : "SUCCESS"));
 
-    if (_socketPoll)
-        _socketPoll->joinThread();
+    std::shared_ptr<SocketPoll> socketPoll = getSocketPoll();
+    if (socketPoll)
+        socketPoll->joinThread();
 }
 
 bool UnitBase::filterLOKitMessage(const std::shared_ptr<Message>& message)
@@ -619,8 +624,9 @@ void UnitBase::returnValue(int& retValue)
 void UnitBase::endTest([[maybe_unused]] const std::string& reason)
 {
     LOG_TST("Ending test by stopping SocketPoll [" << getTestname() << "]: " << reason);
-    if (_socketPoll)
-        _socketPoll->joinThread();
+    std::shared_ptr<SocketPoll> socketPoll = getSocketPoll();
+    if (socketPoll)
+        socketPoll->joinThread();
 
     // tell the timeout thread that the work has finished
     TimeoutConditionVariable.notify_all();
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -32,9 +32,9 @@
 #include <common/SigUtil.hpp>
 #include <common/StringVector.hpp>
 
-UnitKit *GlobalKit = nullptr;
-UnitWSD *GlobalWSD = nullptr;
-UnitTool *GlobalTool = nullptr;
+std::atomic<UnitKit *>GlobalKit = nullptr;
+std::atomic<UnitWSD *>GlobalWSD = nullptr;
+std::atomic<UnitTool *>GlobalTool = nullptr;
 UnitBase** UnitBase::GlobalArray = nullptr;
 int UnitBase::GlobalIndex = -1;
 char* UnitBase::UnitLibPath = nullptr;
@@ -704,8 +704,9 @@ void UnitWSD::DocBrokerDestroy(const std::string& key)
 
                 LOG_TST("Starting test #" << GlobalIndex + 1 << ": "
                                           << GlobalArray[GlobalIndex]->getTestname());
-                if (GlobalWSD)
-                    GlobalWSD->configure(Poco::Util::Application::instance().config());
+                UnitWSD *globalWSD = GlobalWSD;
+                if (globalWSD)
+                    globalWSD->configure(Poco::Util::Application::instance().config());
                 GlobalArray[GlobalIndex]->initialize();
             }
 
@@ -717,8 +718,9 @@ void UnitWSD::DocBrokerDestroy(const std::string& key)
 
 UnitWSD& UnitWSD::get()
 {
-    assert(GlobalWSD);
-    return *GlobalWSD;
+    UnitWSD *globalWSD = GlobalWSD;
+    assert(globalWSD);
+    return *globalWSD;
 }
 
 void UnitWSD::onExitTest(TestResult result, const std::string&)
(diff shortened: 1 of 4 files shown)


caolanm:
std::atomic<std::shared_ptr<>> is a problem for the android/iOS (and whatever config the c++ CodeQL is using) unfortunately

sgothel:
Force push (rebased -i): Replaced atomic<shared_ptr> w/ mutex since neither Android not CodeQL passed.
Note: atomic<shared_ptr> is valid C++20.

sgothel:
> std::atomic<std::shared_ptr<>> is a problem for the android/iOS (and whatever config the c++ CodeQL is using) unfortunately

yup, figured and fixed already :)

sgothel:
failed [desktop/cypress](https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1621/)
- writer/annotation_spec.js
- writer/invalidations_spec.js

but passed locally on my machine

caolanm (approved):
This "writer/annotation_spec.js" is very likely related to comment/annotations changes I made to writer, but I can't reproduce it locally. Odd, lets see if it remains as an issue and I'll retrigger it.

common/Unit.hpp
```diff
@@ -356,6 +362,7 @@ class UnitBase
     UnitType _type;
 
     std::mutex _lock; //< Used to protect cleanup functions.
+    std::mutex _lockSocketPoll; //< Used to sync _socketPoll
```
mmeeks, line 365: Surprised to see a new mutex; fewer mutexes are really a good thing - we could use _lock to protect both the setup and cleanup surely ?

common/Unit.cpp
```diff
@@ -32,9 +32,9 @@
 #include <common/SigUtil.hpp>
 #include <common/StringVector.hpp>
 
-UnitKit *GlobalKit = nullptr;
-UnitWSD *GlobalWSD = nullptr;
-UnitTool *GlobalTool = nullptr;
+std::atomic<UnitKit *>GlobalKit = nullptr;
+std::atomic<UnitWSD *>GlobalWSD = nullptr;
+std::atomic<UnitTool *>GlobalTool = nullptr;
```
mmeeks, line 37: We really need these as std::atomic ? - and GlobalTool too ? I'm surprised at the race this fixes - I would have preferred longer stack traces from the race in the changelog so we can see which one should really be serialized on the other and/or what order they should happen in.

mmeeks (commented):
Merged now I guess; but perhaps the _lock can be cleaned up in a follow up.

common/Unit.hpp
sgothel: Not really a surprise, see my PR offer (flavor, taste) https://github.com/CollaboraOnline/online/pull/10244

common/Unit.cpp
sgothel: Yes of course we need it, as I ran into a data race which was already earmarked in 
`UnitWSD::DocBrokerDestroy` where GlobalWSD etc are set:

> // We could be called from either a SocketPoll (websrv_poll)
> // or from invokeTest (coolwsd main).

Then the static UnitWSD::get() or UnitKit::get() gets called
from e.g. coolwsd with potential old non-sync'ed data.

For this, I only had the assert SIGSEGV stacktrace which I have not saved sadly,
but to no surprise such assertion would occur every now and then w/o atomic.

(though I have stored my _sockePoll logs w/ backtraces to all data races from my earlier triage logs ...)

sgothel: SIGABRT (not SIGSEGV) due to assert as mentioned in the issue description. Sorry.

Ashod: The atomics here are unnecessary and most certainly don't fix anything. The reason is because we are already taking a (new) lock left and right.

We set these synchronously, at startup, when we have but one thread.

Unless something drastically changed that caused a regression, I find it incredible that we went for 9 years without seeing a single segfault until now.

common/Unit.cpp
```diff
@@ -698,8 +704,9 @@ void UnitWSD::DocBrokerDestroy(const std::string& key)
 
                 LOG_TST("Starting test #" << GlobalIndex + 1 << ": "
                                           << GlobalArray[GlobalIndex]->getTestname());
-                if (GlobalWSD)
-                    GlobalWSD->configure(Poco::Util::Application::instance().config());
+                UnitWSD *globalWSD = GlobalWSD;
+                if (globalWSD)
+                    globalWSD->configure(Poco::Util::Application::instance().config());
```
Ashod, line 709: We made `GlobalWSD` an atomic only to take a naked copy?!?

If the atomics are needed, this is code is totally broken and nonsensical. How is taking a copy of an atomic-pointer useful?

Ashod (commented):
Thanks, @sgothel. It sounds like there is a recent regression that we haven't figured out what caused it. This PR reads like a jumbled attempt at random changes in the hope of something would stick. Sorry.

This code doesn't explain why only unit-wopi-save-on-exit would fail in the way it does, and how these changes would fix it. What is so special about unit-wopi-save-on-exit in relation to this very common test objects that would only show up in unit-wopi-save-on-exit, and only in unit-wopi-save-on-exit?

Perhaps I'm missing something obvious, but I hope my expectations are sensible. I expect an explanation of the root cause and a justification for changes that address said root cause. Does that make sense?

sgothel:
> Thanks, @sgothel. It sounds like there is a recent regression that we haven't figured out what caused it. This PR reads like a blind-fix attempt. Sorry.
> 
> This code doesn't explain why only unit-wopi-save-on-exit would fail in the way it does, and how these changes would fix it. What is so special about unit-wopi-save-on-exit in relation to this very common test objects that would only show up in unit-wopi-save-on-exit, and only in unit-wopi-save-on-exit?
> 
> Perhaps I'm missing something obvious, but I hope my expectations are sensible. I expect an explanation of the root cause and a justification for changes that address said root cause. Does that make sense?

I have recreated and documented the issue with this branch [private/sgothel/ut-stability-01_triage0_02_patch1](https://github.com/CollaboraOnline/online/commits/private/sgothel/ut-stability-01_triage0_02_patch1/).
(Since I wasn't sure how to share these triage results, I simply pushed them into that private branch - not to be merged!)

The log [test/logs/ut-stability-01_triage0_02_patch1/detail.txt](https://github.com/CollaboraOnline/online/blob/private/sgothel/ut-stability-01_triage0_02_patch1/test/logs/ut-stability-01_triage0_02_patch1/detail.txt) shows the SIGABRT (assert) only after having the 1st patch applied (_socketPoll lock) as I only encountered it afterwards (AFAIK I have mentioned this in the commit message).

These issues were visible with the mentioned `unit-wopi-save-on-exit` test, 
but indeed should have been triggered with similar multi-serial-test cases I guess
due to the lack of synchronization.

I also mentioned that these patches 'only' resolve the resulting phenomenon,
i.e. fix the issue from the bottom end - call it a band-aid.
This especially, since I am not yet fully familiar or have analyzed the overall calling structure,
the implementation - intention, etc.

For example, creating a global instance off-thread from within a UnitBase instance using a synchronized method but allowing to fetch the same global instance indeed looks sort-of odd.
But then, without fully having analyzed the issue - I refrain to assume or even judge!

Bottom line, this patch series merged was a band-aid to play along with how UnitBase was used,
avoiding the annoying UT failures - nothing more and nothing less.

In referenced branch, you will also find the log failure pre-patch1, which I already posted here.

++++

Now the good news. I have the same branch on current master after Ashod's patch series addressing certain locks/synchronization and I was not able to recreate test failures after running 400 test cycles w/o a patch.
Hence your work seems to also contain fixes for these unit tests, great job.

Maybe you can explain the root reason of 'misused' UnitBase before and why it was fixed?

Thank you.

mmeeks:
I'm not a fan of that detail text file in the source. There needs to be a clear explanation for the code-change, particularly if it's complex in the commit message. That should include tasteful stack-traces: doesn't need to be a gdb output, and if it is needs to go through c++filt so we don't have _ZTV6Foobaa  type stuff in the symbols: and redundant pieces there that are common are not-interesting - lots of frames of libc / libstdc++ implementing abort or whatever adds little value.

The stack trace(s) should highlight what the problem is; if there are two pieces of code that race - we can add printfs to them to show what short stack frame races with what other one, so we are really confident that the problem is fully understood before documenting it, and then fixing it once.

> But then, without fully having analyzed the issue - I refrain to assume or even judge!

I really expect a fully understanding of the issue before the fix is created. We have enough band-aids in the code, and we fight to keep it as beautiful / consistent / minimal as it can be: fewer locks, fewer branches, cleaner code, less duplication etc. otherwise things get worse not better.

If Ash cleaned this up / fixed it some better way - we should revert the band-aid / blind-fix too. Sometimes faced with some random memory corruption we can't reproduce - blind fixes are the only way we can go; we should annotate them as such, for the reviewer and the person following on to fix it properly later =)

Thanks.

PR #10231, opened 2024-10-15
mount an up to date copy of etc if ro systemplate's etc is out of date

caolanm:
same as:

commit 852e8545af3f1807ff6619fe52a922f35bbd9da4
Author:     Caolán McNamara <caolan.mcnamara@collabora.com>
AuthorDate: Mon Jul 22 13:07:17 2024 +0100

    mount an up to date copy of etc if systemplate's etc is out of date

but also do this is the out of date systemplate dynamic files are in a ro systemplate.

reproducible in make run with

rm -f systemplate/etc/hosts && chmod a-w systemplate

where the earlier change only works for

rm -f systemplate/etc/hosts && chmod a-w systemplate/etc

Change-Id: I054d844f15456f189d19cb204fd80541f827dcbf

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/JailUtil.cpp
+++ b/common/JailUtil.cpp
@@ -581,13 +581,13 @@ bool updateDynamicFilesImpl(const std::string& sysTemplate)
 
         if (checkWritableSysTemplate && !FileUtil::isWritable(sysTemplate))
         {
-            disableBindMounting(); // We can't mount from incomplete systemplate that can't be updated.
             LinkDynamicFiles = false;
             LOG_WRN("The systemplate directory ["
                     << sysTemplate << "] is read-only, and at least [" << dstFilename
-                    << "] is out-of-date. Will have to copy sysTemplate to jails. To restore "
-                       "optimal performance, make sure the files in ["
-                    << sysTemplate << "/etc] are up-to-date.");
+                    << "] is out-of-date. Will have to clone dynamic elements of "
+                    << "systemplate to the jails. To restore optimal performance, "
+                    << "make sure the files in [" << sysTemplate << "/etc] "
+                    << "are up-to-date.");
             return false;
         }
 


common/JailUtil.cpp
```diff
@@ -581,13 +581,13 @@ bool updateDynamicFilesImpl(const std::string& sysTemplate)
...
         {
-            disableBindMounting(); // We can't mount from incomplete systemplate that can't be updated.
             LinkDynamicFiles = false;
             LOG_WRN("The systemplate directory ["
                     << sysTemplate << "] is read-only, and at least [" << dstFilename
-                    << "] is out-of-date. Will have to copy sysTemplate to jails. To restore "
-                       "optimal performance, make sure the files in ["
-                    << sysTemplate << "/etc] are up-to-date.");
+                    << "] is out-of-date. Will have to clone dynamic elements of "
+                    << "systemplate to the jails.. To restore optimal performance, "
```
timar, line 588: `..` duplicated.

Is there really a performance penalty?

common/JailUtil.cpp
```diff
@@ -581,13 +581,13 @@ bool updateDynamicFilesImpl(const std::string& sysTemplate)
 
         if (checkWritableSysTemplate && !FileUtil::isWritable(sysTemplate))
         {
-            disableBindMounting(); // We can't mount from incomplete systemplate that can't be updated.
```
Ashod, line 584: This assumes we mount /etc on-top of the mounted sysTemplate? I trust that is working as expected.

Ashod (approved):
Thanks, @caolanm.

common/JailUtil.cpp
caolanm: Well, there is a perf penalty if we end up doing the mega copy of the entire /opt/cool/systemplate in the end. If we only need to make a /etc, then my one is about a 1M so not a big deal I suppose, but it would be better to not do any copy. I just wanted to update this message to reflect what will really happen after this change.

common/JailUtil.cpp
caolanm: I certainly believe so, and it does for me when I try it. [This is the existing bit that does it](https://github.com/CollaboraOnline/online/commit/852e8545af3f1807ff6619fe52a922f35bbd9da4#diff-a835d79df212a8e0dbe2430c2901a1e8dc94086e785f6fd52f269b80009a1d11R3063)

PR #10233, opened 2024-10-15
docs(mobile): Mention co-24.04-mobile in README

Minion3665:
For contributors to mobile, it's nice to mention the process around the co-24.04-mobile branch in relation to master and co-24.04. I've opted not to use the platform-specific READMEs as (1) it applies to both platforms and (2) those aren't very good and probably need to be rewritten more generally

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,9 +71,37 @@ Please consult the README files in the component's directory for more details:
 
 ## iOS and Android apps
 
+### Building
+
 See the corresponding READMEs:
 * **[ios/README](ios/README)**
 * **[android/README](android/README)**
+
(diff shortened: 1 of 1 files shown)


commands.sh
timar, line 1: What is this file?

Minion3665: Ahh, I'm very sorry, that shouldn't be included in this commit, I'll amend to remove it

timar (approved):
OK, but make sure you remove this, when it's obsolete (next year).

PR #10236, opened 2024-10-15
CI: I do not think we need libreoffice build deps

timar:
Change-Id: I73fea30cc75a2a9ff2a3890c4c8e0ae2103c5a81

* Resolves: # 
* Target version: master 

### Summary
It makes the CI a bit faster. libreoffice build deps are huge, and unnecessary. We use pre-built core assets.


vmiklos (approved):
Makes sense to me, thanks.

PR #10239, opened 2024-10-15
let the user copy text from a comment

JaumePujantell:
Right now if a user selects text from a comment and tries to copy it,
the copy command goes to the document instead. This change allows the
user to copy from the content of a comment.

Change-Id: I9dc9198fd28f5a723c454444644f361bc31a4c49

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -568,7 +568,7 @@ L.Clipboard = L.Class.extend({
 		return true; // prevent default
 	},
 
-	_isAnyInputFieldSelected: function() {
+	_isAnyInputFieldSelected: function(forCopy = false) {
 		if ($('#search-input').is(':focus'))
 			return true;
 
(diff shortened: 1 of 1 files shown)


vmiklos (commented):
Hmm, this is indeed a problem, but when I test your patch, it seems to be still not working. Here is what I tried:

1) Invoke 'make run', open the writer-edit.fodt
2) Go to the first comment in the first page, select the word "children" (4th line) by double-clicking on it, a selection is created.
3) Press Ctrl-C to copy it
4) Start e.g. desktop Writer and paste there.

Expected result: "children" gets copied.

Actual result: an empty paragraph gets inserted.

Is this something you can reproduce? Thanks.

JaumePujantell:
> Actual result: an empty paragraph gets inserted.
> 
> Is this something you can reproduce? Thanks.

Right, it doesn't work on Chromium although it works on Firefox, thanks.

vmiklos (changes requested):
OK, please either make it explicit on your commit message that this is Firefox-only for now; or make it also work on Chrome. :-) Thanks.

JaumePujantell:
@vmiklos now it should work for Chrome

vmiklos (approved):
Looks good to me now, thanks.

Could you please look at if you could cover this with a cypress test?

Either in cypress_test/integration_tests/desktop/writer/copy_paste_spec.js or in cypress_test/integration_tests/desktop/writer/annotation_spec.js, for example. Thanks.

PR #10243, opened 2024-10-16
Fix switching from notes view to normal view issue.

gokaysatir:
Reproducer:
* Open an impress file.
* Switch to notes view.
* Edit the note of page.
* Don't remove focus from note object (rectangle).
* Try to switch back to normal view.

We still send the notes view UI command because when user clicked on the note rectangle, context changed again. Now our UI thinks that we are in different mode. We will set our internal selectedMode variable selectively.

Change-Id: Iee73a1bb16aed7c19190ef15592910e39a056beb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/ImpressTileLayer.js
+++ b/browser/src/layer/tile/ImpressTileLayer.js
@@ -50,9 +50,19 @@ L.ImpressTileLayer = L.CanvasTileLayer.extend({
 	},
 
 	_onContextChange(e) {
-		app.impress.notesMode = e.detail.context === 'NotesPage';
+		/*
+			We need to check the context content for now. Because we are using this property for both context and the page kind.
+			When user modifies the content of the notes, the context is changed again. As we use context as a view mode, we shouldn't change our variable in that case.
+			We need to check if the context is something related to view mode or not.
+			For a better solution, we need to send the page kinds along with status messages. Then we will check the page kind and set the notes view toggle accordingly.
+		*/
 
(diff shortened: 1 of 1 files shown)


browser/src/layer/tile/ImpressTileLayer.js
```diff
@@ -50,9 +50,17 @@ L.ImpressTileLayer = L.CanvasTileLayer.extend({
...
-		app.impress.notesMode = e.detail.context === 'NotesPage';
+		/*
+			We need to check the context content for now. Because we are using this property for both context and the page kind.
+			When user modifies the content of the notes, the context is changed again. As we use context as a view mode, we shouldn't change our variable in that case.
+			We need to check if the context is something related to view mode or not.
+			For a better solution, we need to send the page kinds along with status messages. Then we will check the page kind and set the notes view toggle accordingly.
+		*/
 
-		if (app.map.uiManager.getCurrentMode() === 'notebookbar') {
+		if (['DrawPage', 'NotesPage'].includes(e.detail.context))
```
eszkadev, line 60: can we add:
`const isDrawOrNotesPage = ['DrawPage', 'NotesPage'].includes(e.detail.context);`
it's repeated 2 times, here and below

gokaysatir: Thanks for checking. Sent an update.

PR #10244, opened 2024-10-16
UnitBase: No-side-effect usage of AtomicSharedPtr<SocketPoll> instead of single `shared_ptr<T>` and `mutex`

sgothel:
* Resolves: #10228 
* Target version: master 

### Summary
The two seperate single non-recursive mutex _lockSocketPoll and _lock were required due to the recursive call from _lock protected code into _lockSocketPoll protected code.

This PR offers new `AtomicSharedPtr<T>` for thread-sync'ed `shared_ptr<T>` access    
- provides atomic operation for `std::shared_ptr<T>` similar to `std::atomic<std::shared_ptr<T>>`.
- further provides thread-safe `assignIfNull` and `assignIfNotNull` atomic conditional assignment methods.
- implementation uses `std::mutex` to synchronize `std::shared_ptr<T>` access.

Using AtomicSharedPtr<T> for SocketPoll removes any potential unwarranted synchronization side-effects,
similar to `std::atomic<std::shared_ptr<T>>`, while also providing conditional atomic assignment.

### Checklist

### DIFF
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -411,12 +411,6 @@ private:
         return false;
     }
 
-    std::shared_ptr<SocketPoll> getSocketPoll()
-    {
-        std::lock_guard<std::mutex> guard(_lockSocketPoll);
-        return _socketPoll;
-    }
-
     static UnitBase* get(UnitType type);
 
     /// setup global instance for get() method
@@ -440,8 +434,7 @@ private:
     UnitType _type;
 
     std::mutex _lock; //< Used to protect cleanup functions.
-    std::mutex _lockSocketPoll; //< Used to sync _socketPoll
-    std::shared_ptr<SocketPoll> _socketPoll; //< Poll thread for async http comm.
+    AtomicSharedPtr<SocketPoll> _socketPoll; //< Poll thread for async http comm.
 
 protected:
 
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -54,6 +54,83 @@ namespace Poco
 class Session;
 class StorageBase;
 
+/// AtomicSharedPtr provides atomic operation for `std::shared_ptr<T>` similar to `std::atomic<std::shared_ptr<T>>`
+/// It further provides thread-safe `assignIfNull` and `assignIfNotNull` atomic conditional assignment methods.
+/// Implementation uses `std::mutex` to synchronize `std::shared_ptr<T>` access.
+template <class T>
+class AtomicSharedPtr
+{
+    private:
+        mutable std::mutex _lock;
+        std::shared_ptr<T> _sptr;
+    public:
+        AtomicSharedPtr()
+        : _sptr(nullptr)
+        {}
+
+        AtomicSharedPtr(const AtomicSharedPtr& o)
+        : _sptr(nullptr)
+        {
+            std::lock_guard<std::mutex> guard(o._lock);
+            _sptr = o._sptr;
+        }
+
+        AtomicSharedPtr(const std::shared_ptr<T>& sptr)
+        : _sptr(sptr)
+        {}
+
+        AtomicSharedPtr& operator=(const AtomicSharedPtr& o)
+        {
+            std::scoped_lock lock(_lock, o._lock);
+            _sptr = o._sptr;
+            return *this;
+        }
+
+        AtomicSharedPtr& operator=(const std::shared_ptr<T>& sptr)
+        {
+            std::lock_guard<std::mutex> guard(_lock);
+            _sptr = sptr;
+            return *this;
+        }
+        void store(const std::shared_ptr<T>& sptr)
+        {
+            std::lock_guard<std::mutex> guard(_lock);
+            _sptr = sptr;
+        }
+
+        /// Assigns returned `std::shared_ptr<T>` from `ctor` iff this's instance's `std::shared_ptr` is `nullptr`
+        template<class CtorFunc>
+        const std::shared_ptr<T>& assignIfNull(CtorFunc ctor)
+        {
+            std::lock_guard<std::mutex> guard(_lock);
+            if( nullptr == _sptr)
+                _sptr = ctor();
+            return _sptr;
+        }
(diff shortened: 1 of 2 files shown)


mmeeks (changes requested):
Urk; not good. Surely we can have a getSocketPollUnlocked() and just assert that _lock is taken in there; or better just use _lock directly in that section - surely ?

sgothel:
> Urk; not good. Surely we can have a getSocketPollUnlocked() and just assert that _lock is taken in there; or better just use _lock directly in that section - surely ?

This PR is not my favorite, just provided optionally - me happy as-merged.

Since the threading and caller restriction doesn't exist for methods calling into getSocketPoll() etc, 
the initial merged PR is IMHO sufficient and clean using 2 simple mutex.
Note that my initial attempt was to simply using C++20 atomic<shared_ptr>,
but neither CodeQL nor Android/iOS allowed it.

Yes, we could detail and constrain the call of methods calling into getSocketPoll() 
to avoid the lock, i.e. use _socketPoll w/o synchronization.
I usually only do such tricks and 'inner method constraints' above the language level,
for high frequency methods enabled in production. This is not the case here.

I could analyze the recursive nature in more detail and add the constraint to drop the lock
where it currently is already taken. This would indeed require to ensure such methods are not called
otherwise.
Shall I do this?

sgothel:
> In general adding a new lock is a nightmare. We want to reduce the number of locks, and simplify interactions. 

I also would only use synchronization when demanded by a use-case.

> Having a valid reason for having two locks to protect different members of the same class would require some rather complex scenario IMHO - which I don't see here. Similarly recursive locks are bad practice.

Here, _socketPoll is accessed by more than one thread and (being new to this code),
I assume it is intentional and by design. 
Hence the use-case requires the data to be sync'ed, i.e. using atomic or local-lock.

> I don't want our code to sprout an ever increasing number of mutexes. This is in large part why I would like the stack trace of the race / contention - which may indicate some lower level problem. Perhaps you can construct a better description of what was called from where that caused what other thread to race it. 

> I would like to see why we are destroying tests while we are invoking new ones or whatever silly is happening.
> 

Yes, I only added the comment to UnitBase::socketPoll() 
```
    // We could be called from either a UnitWSD::DocBrokerDestroy (prisoner_poll)
    // or from UnitWSD::invokeTest() (coolwsd main).
```
will add stack traces here.

> Ideally with a pair of stack frames. Even better - we can get rid of this extra lock, and re-use the one we already have.
> 

I detailed why it might be risky to avoid having an additional sync-instrument on _socketPoll (constraints of callers).

Will add the stack traces for more clarity and give the constraint another thought. OK!
As mentioned, so far the use-case implies a sync/local-lock is required - but if it turns out
we want to guarantee that certain callers only happen in an already locked state ...
Hence I hesitated to dive to far and imply anything - but will look at it.

sgothel:
Little call-tree analysis 

Called with guarded _lock, safe
- UnitBase::endTest ()
  - caller UnitBase::exitTest

Called outside by diff threads, needs sync for _socketPoll)    
- UnitKit::postFork 
- UnitBase::~UnitBase
- UnitBase::socketPoll()

sgothel:
Best semantic coverage would have been to use C++20 `atomic<shared_ptr> _socketPoll`, 
but not possible due to tooling (see above).
I.e. this would made clear that there are no mutex side-effects even if that atomic internally would use a mutex.

To clarify this (in the orig merged PR) .. we could have a replacement for `atomic<shared_ptr>`
as intended with the `getSocketPoll` method - covering the data + sync.
At least this would clarify intend and remove possible side-effects.

sgothel:
Force pull provides a new proposal for clarification, i.e. no-side-effect usage of synchronized shared_ptr<SocketPoll>,
please read updated description.

If accepted, I would propose to move the new AtomicSharedPtr<T> to online's common
for general use.

sgothel:
One log/trace produced with my instrumented code earlier showing the data-race on _socketPoll.

Compare my marks `JAU 02`  and `JAU 10`  showing Backtraces from different threads both originating from `UnitBase::socketPoll()`.

[unit-wopi-save-on-exit_1.log.txt](https://github.com/user-attachments/files/17413842/unit-wopi-save-on-exit_1.log.txt)

mmeeks (changes requested):
In general std::atomic use is a bad smell for me - perhaps for counters or something it makes sense; but people trying to do cross-thread synchronization properly are likely to be heavily screwed over by std::atomic use - and just make the situation far more complex and problematic.

Having multiple std::atomic shared_ptrs that are checked in a sequence seems (to me) like a recepie for creating complex races.

I suspect while it sounds useful in general, we should avoid all things that create more of this std::atomic mess around the place - except for simple counters and/or conditions that change only once during the lifetime of a process or somesuch.

Thread safety is extremely hard to get right and maintain, and we need to think very, very hard before adding more mutexes.

common/Unit.hpp
```diff
@@ -54,6 +54,83 @@ namespace Poco
 class Session;
 class StorageBase;
 
+/// AtomicSharedPtr provides atomic operation for `std::shared_ptr<T>` similar to `std::atomic<std::shared_ptr<T>>`
+/// It further provides thread-safe `assignIfNull` and `assignIfNotNull` atomic conditional assignment methods.
+/// Implementation uses `std::mutex` to synchronize `std::shared_ptr<T>` access.
+template <class T>
+class AtomicSharedPtr
```
Ashod, line 61: [`std::atomic<std::shared_ptr>`](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2)?

I'm sure this was fun to write, but the standard has this already... assuming we need it at all. I'm not convinced we are fixing the right code.

sgothel: Both, I already elaborated on atomic<shared_ptr> and the overall reasoning.
This AtomicSharedPtr was just offered to sooth acceptance, encapsulating the mutex avoiding abuse
_and_ to offer an atomic easy to use `assingIfNull` method.

Added more context to the conversation in the original branch.
Closing this without merging.

PR #10245, opened 2024-10-16
private/marco/slideshow effects based on transition filters

mcecchetti:
- **slideshow: we not really want to deep clone a TransitionFilterAnimation**
- **slideshow: barwipe transitions were not run**
- **move 3d transition to its own folder**
- **slideshow: refactored transition filter based on clipping**
- **slideshow: CombTransition didn't work properly**
- **slideshow: avoid to create a transparent texture for a shape transition**
- **slideshow: transition filters: handle 'Default' subtype correctly**
- **slideshow: support effects based on transition filters**

**Hint for reviewers: it's easier to review one commit at a time**
**Read commit message can be helpful too**
**The big changes are  in the 4th and the 8th commits**


browser/src/slideshow/transition/BoxTransition.ts
```diff
@@ -13,71 +13,49 @@
...
+                float getMaskValue(vec2 uv, float time) {
                     float progress = time;
 
                     vec2 center = vec2(0.5, 0.5);
 
                     vec2 dist = abs(uv - center);
 
-                    float size = (direction == 1) ? progress * 1.5 : (1.0 - progress * 1.5);
+                    float size = progress * 1.5;
+                    ${isInDir ? 'size = 1.0 - size;' : ''}
```
quikee: Modifying the source is a bit problematic if we want to pre-compile the shaders. It's better to use the shader parameter for such. It is something that we need to think about when we get there...

quikee (approved):
In general it's OK - just uneasy about changing the shader source instead of using shader parameters.

PR #10246, opened 2024-10-17
private/ash/lock

Ashod:
- wsd: test: better scenario transition in WOPIUploadConflictCommon
- wsd: test: add idle-timeout unlock test
- wsd: non-modifying uno commands
- wsd: strict equals comparison in StringVector
- wsd: logging and cosmetics
- wsd: correctly set the lock state
- wsd: unlock before unloading
- wsd: lock only after successful loading
- wsd: remove unused lock helper and dead code


wsd/DocumentBroker.cpp
```diff
@@ -1128,48 +1123,35 @@ bool DocumentBroker::download(
...
-void DocumentBroker::lockIfEditing(const std::shared_ptr<ClientSession>& session,
-                                   const Poco::URI& uriPublic, bool userCanWrite)
+void DocumentBroker::lockIfEditing(const std::shared_ptr<ClientSession>& session)
 {
     if (_lockCtx == nullptr || !_lockCtx->supportsLocks() || _lockCtx->isLocked())
     {
         return; // Nothing to do.
     }
 
-    if (session)
```
caolanm, line 1111: checking lockIfEditing wrt the dropping of "if (session)" I see that its only called from onViewLoaded and that's only called from docBroker->onViewLoaded(client_from_this()); so an empty shared_ptr scenario doesn't arise

caolanm (approved):
I don't see anything outside of the acknowledged string comparison changes to give me pause

PR #10247, opened 2024-10-17
cool#9992 doc sign: hide sign button when user pref doesn't provide cert or key

vmiklos:
Have no sign keys configured, open a document, go to the File tab on the
notebookbar, the signature button clutters the UI but it provides no
actual functionality.

The JS UI already got the private user info in Control.UIManager.js
onUpdateViews(), where we initialize the Zotero UI.

Do something similar for the sign UI: keep it enabled by default, but
once we learn about the private user info and it has no signing
key/cert, then hide the button.

The notebookbar gets re-initialized from time to time, so best to do
this via showButton(), which can remember which buttons are shown/hidden
and gets maintained even after a rebuild of the notebookbar. (The
postmessage API uses the same code to hide a button from an
integration.)

Change-Id: Ifad78220a002ed19a5d822d8affed9aea6073737

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -1035,7 +1035,11 @@ L.Control.UIManager = L.Control.extend({
 		}
 
 		var userPrivateInfo = myViewData.userprivateinfo;
-		if (userPrivateInfo && window.zoteroEnabled) {
+		if (userPrivateInfo === undefined)
+		{
+			return;
+		}
+		if (window.zoteroEnabled) {
 			var apiKey = userPrivateInfo.ZoteroAPIKey;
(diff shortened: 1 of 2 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_mobile/1569/console

ah, we need to hide this button on the notebookbar when we have a notebookbar, which is not the case on mobile, let me fix.

vmiklos:

Hides the button to not clutter the UI if it would not work; also don't send the sensitive data to the client without a reason, we just need to know if this data is configured on the server or not.

caolanm (approved):
looks sensible to me

PR #10248, opened 2024-10-17
add an explicit -Woverloaded-virtual flag

caolanm:
-Wall only enables -Woverloaded-virtual=1 but adding an explicit -Woverloaded-virtual enables level 2

Change-Id: I76ff5d752131a04e8f2c5b4001edf81680510efe

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks, @caolanm. This is similar to what I have. Should prevent regressions.

Issue #10198, opened 2024-10-09
Errors with `startPresentation` parameter


elzody:
### Describe the Bug
When attempting to open a presentation file using the `&startPresentation=true` parameter, I receive either a browser console error or an error directly from Collabora, and the presentation does not load. See screenshot section for images.

### Steps to Reproduce
1. Build Collabora and run it
2. Try with the test presentation given by the link in the terminal
3. Open the link and append `&startPresentation=true` to the end
4. See the error

### Expected Behavior
The expected behavior is that the presentation will automatically launch in presentation mode.

### Actual Behavior
The presentation does not open and displays an error.

### Screenshots
<details>
<summary>Browser error</summary>

![Image](https://github.com/user-attachments/assets/9f531c64-a348-4e23-b503-3ea1d4d38d3c)
</details>

<details>
<summary>Collabora error</summary>

![Image](https://github.com/user-attachments/assets/5d67aa68-c95d-484f-9b16-e5bc7925be0f)
</details>

### Desktop
 - **Collabora version:** 24.04.8.1 (mostly up-to-date master branch)
 - **OS and version:** Arch Linux x86_64 [6.11.1-arch1-1]
 - **Browser and version:** LibreWolf 130.0.1-1 (64-bit)

jazevedo-coll:
I can reproduce this issue.

gokaysatir:
Error raises from Control.UIManager.js file. When the presentation is initiated from URL, UI Manager tries to run it before the page is loaded completely. I think a timer for this case that would try until the page is loaded would help.

juliusknorr:
It looks like there is as regression with this @eszkadev, hope it is fine if I reopen.

For me using the startPresentation parameter just ends in a black screen with a loading spinner with latest online master branch.

PR #10249, opened 2024-10-17
fix: ensure presentation starts only after document is fully loaded

Darshan-upadhyay1110:
- Previously, when attempting to open a presentation file using the `&startPresentation=true` parameter, the presentation would not load, leading to either a browser console error or a direct error from Collabora.

-  This issue was resolved by dispatching the 'presentation' event only after the document has fully loaded, preventing premature attempts to start the presentation.

Change-Id: Ib066a28c0aff0ed4f979b5d6d463bfd900519730

* Resolves: #10198 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks good

PR #10252, opened 2024-10-17
css: fix icons size after split button rework

eszkadev:
it was broken after PR:
commit db10c453e95f732d85c489fe8da8708cec0ad09c
fix(UI): set button size with a CSS variable

related to: https://github.com/CollaboraOnline/online/pull/10027


eszkadev:
but there are visible some small "dots" on the right of buttons with part of old icon? (non replaced original one?)

Issue #7130, opened 2023-08-22
calc comments don't have to be a red rectangle


caolanm:
### Is your feature request related to a problem?

The red rectangles in the spreadsheet that indicate a comment aren't very pretty and don't scale when the spreadsheet is zoomed in

### Describe the solution you'd like

In https://github.com/CollaboraOnline/online/commit/ccb20f1ed329e483f52831858f711b3a0e73502d we moved the rendering of these into the browser, matching how they were doing in core. So just a simple fillRect in red. It might be nicer to at least draw something like a triangle that scales with the zoom.

caolanm:
@pedropintosilva maybe of interest to you

pedropintosilva:
@caolanm : that's great news , thanks for the ping!

pedropintosilva:
## Color

I think it would be good to steer away from red color and instead use yellow or blue. Probably best to go with yellow so to make it stand out (and avoid too similar colors nearby).

## Shape

I would go with something that can signify "open comment, pop out" but still using a basic shape (no icon)
* Paint top right corner of the cell in yellow (Triangle) as it happens in other spreadsheet programs

PR #10254, opened 2024-10-17
Shows calc comment indicator as a triangle.

GulsahKose:
Comment indicator is changed in core d395b42cdf3ef517d215dee905e5ccf2ae73b2b8 We draw that indicator on online side separately. Now we see the same indicator on online too.

Change-Id: I2c640f4850473724e8a14e75fa7d3c00f76d4e9e

* Resolves: #7130
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -1325,13 +1325,18 @@ export class Comment extends CanvasSectionObject {
 				if (cellSize[0] !== 0 && cellSize[1] !== 0) { // don't draw notes in hidden cells
 					// For calc comments (aka postits) draw the same sort of square as ScOutputData::DrawNoteMarks
 					// does for offline
-					var margin = 3;
+					var margin = cellSize[0] * 0.06;
 					var squareDim = 6;
 					// this.size may currently have an artifically wide size if mouseEnter without moveLeave seen
 					// so fetch the real size
 					var x = this.isCalcRTL() ? margin : cellSize[0] - (margin + squareDim);
-					this.context.fillStyle = '#FF0000';
-					this.context.fillRect(x, 0, squareDim, squareDim);
+					this.context.fillStyle = '#BF819E';
+					var region = new Path2D();
+					region.moveTo(x, 0);
+					region.lineTo(cellSize[0], 0);
+					region.lineTo(cellSize[0], cellSize[1]/2);
+					region.closePath();
+					this.context.fill(region);
 				}
 			}
 		}


eszkadev:
Looks nice, but you need to also take into account zoom level I think:
comment-zoom

I inserted comment in view with 100%, then I zoomed in and it is off.

GulsahKose:
@eszkadev @pedropintosilva Position problem is not related with this pr. I've updated core co-24.04 and online latest master. Without this PR it looks as video: 
[Screencast from 2024-10-24 16-04-29.webm](https://github.com/user-attachments/assets/aadc63f5-9e39-42b6-95ad-29091dd3412c)

eszkadev:
@GulsahKose good catch! but can we try to fix it too? It should be really easy.
You do the calculations of the x, y coordinates. It is only required to correctly multiply by the factor based on zoom/dpi.

Some pointers to find numbers: https://github.com/CollaboraOnline/online/blob/master/browser/src/docstate.js#L20
You can look how it is used in other places. In my opinion this issue is serious and very visible.

pedropintosilva:
so it seems Gulsah already tried that but it didn't work, probably because the cell coordinates are already being multiplied and so, I think we shouldn't need to multiple again (? as it seems the triangle is been drawn according to the cell). 

@GulsahKose could you please add here your notes? and also your question around outgoing message `clientVisibleArea`


https://github.com/user-attachments/assets/0362db1f-d24c-486b-a75b-b0493a0fc7b9

GulsahKose:
@eszkadev I've tried to multiply with app.dpiScale but didn't fix the problem. Then I realized I'm using cell coordinates that also multiplied with scale. And the triangle is also grown when we zoom.

And found a good case. When we zoom in then scroll the sheet (with mouse wheel)  the indicator position is fixed.

When i checked the console I see we sent clientvisiblearea message that contains visible area coordinates, width and height. I think after the zoom, this visible area coordinates is wrong so triangle position is wrong too. When the clientvisiblearea updated with scroll position is fixed too.

eszkadev:
Maybe just a "refresh" is needed? @gokaysatir might know

gokaysatir:
> Maybe just a "refresh" is needed? @gokaysatir might know

I am checking the positioning issue now.

pedropintosilva:
_Pressed retry Cypress (desktop) as the failing test is unrelated "integration_tests/mobile/writer/annotation_spec.js "_

gokaysatir:
@eszkadev @GulsahKose i think below commit solves the issue:
* https://github.com/CollaboraOnline/online/pull/10328

pedropintosilva:
_Pressed the retry button on cypress mobile since `writer/annotation_spec.js `  is unrelated_

meven:
There is a typo in the commit message and MR title s/indicatior/indicator

pedropintosilva (approved):

pedropintosilva:
I think we could merge in red window (since it doesn't fix any particular bug) or we could merge this since it seems safe and better than the current situation, what's your opinion @eszkadev 

Furthermore

eszkadev:
this seems to be safe @pedropintosilva

meven:
> There is a typo in the commit message and MR title s/indicatior/indicator

You only changed the PR title, I am more concerned about the commit message.

meven:
> > There is a typo in the commit message and MR title s/indicatior/indicator
> 
> @gokaysatir You only changed the PR title, I am more concerned about the commit message.

Too late to fix now.

PR #10255, opened 2024-10-17
slideshow: fix getting the document node

eszkadev:
avoid infinite reccurency... mistake when solving conflict


mcecchetti (approved):
Fix the problem for me

PR #10260, opened 2024-10-18
impress: fix selection of objects

eszkadev:
it fixes regression introduced in commit a9985a414f9f60299e7ea42e8cee2e0646ff6a25 cypress: run new slideshow in iframe


gokaysatir (approved):
Thanks for the fix :)

Issue #10238, opened 2024-10-15
Function Wizard - Formula Field Overflows Window Borders


Darshan-upadhyay1110:
Potential mentors: @pedropintosilva @Darshan-upadhyay1110 
Here's a concise GitHub issue description for the problem:

**Description**:
1. Open a sheet.
2. Select the Function Wizard.
3. Expand the formula field horizontally.

**Result**:
- The formula field overflows the window borders and drags the Function Wizard with it.
- Currently, the formula field can be resized both horizontally and vertically, but the Function Wizard window only resizes vertically.

**Expected Behavior**:
- The window and the formula field should have matching behavior. We should :
  - Restrict the formula field from expanding horizontally.

[Screencast_from_2024-09-19_16-59-33.webm](https://github.com/user-attachments/assets/7edfdf7f-7dad-459d-889b-f909f812d5d5)

**Suggestion**:
- Make the behavior of the Function Wizard window and the formula field consistent for a better user experience.

PR #10262, opened 2024-10-18
fix: formula field overflow on function wizard window

aarsh21:
Change-Id: Ie034a0c766803a3cd3f1e5ad02f1228393289348

* Resolves: https://github.com/CollaboraOnline/online/issues/10238
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -2252,3 +2252,10 @@ kbd,
 	display: table-caption;
 	height: 16px;
 }
+
+/* text-area overflow on function wizard window */
+
+#FormulaDialog #ed_formula.ui-textarea {
+	max-width: 392px;
+}
+


browser/css/jsdialogs.css
```diff
@@ -1027,6 +1027,7 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	vertical-align: top;
 	width: -webkit-fill-available;
 	width: -moz-available;
+	max-width: 392px;
```
eszkadev, line 1030: This will affect all the textareas in the whole app.
Please add new selector more specific for the case: eg.  `#YOURDIALOGID #TEXTAREAID.ui-textarea`
Best: at the end of a file with a comment above where we can find such field, thanks for a patch anyway!

pedropintosilva (approved):
Thanks! I think we can merge but as a future improvement I would rather find out why we cannot simply set grid-template-columns: 252px auto; to the parent element and also why some of these dialogs' child elements have a hard  coded width such as the text area

PR #10268, opened 2024-10-19
condition format: fixed incorrect parameters in function call

lpranam:
Change-Id: I04c34ab13ca1d45de1ada3ddeb656d00bbc0447c

* Resolves: # 
* Target version: master 

### Checklist


gokaysatir (approved):
Thanks for the catch :)

Issue #10047, opened 2024-09-13
Mention issues: PostMessage is only sent once. Keyboard doesn't work


juliusknorr:
### Describe the Bug

While working on emitting notifications for [user mentions](https://github.com/nextcloud/richdocuments/pull/3960) in the Nextcloud integration I noticed a couple of issues around mentions that are related to the Collabora side of the implementation.

It would be great to address those to make the mentions feature more flawless.

### Steps to Reproduce

1. Use the Nextcloud integration with Collabora
2. Type `@` to insert a mention
3. Start typing and search for a user

- [ ] Currently the search is only triggered once on the first character press `@a`. This is especially a problem for large instances with lots of users where the first result set would not contain all of the users. Instead there should be a debounced fetching of users while typing. You also see that only one postmessage is sent to Nextcloud.
- [ ] Clicking is required to insert a mention, having one match an pressing enter does not work

### Additional Context

Latest master branch of online

timur-g:
#7923 seems related where double @@mention may work.

Rash419:
https://github.com/Rash419/online/commits/debounce-mention-user-query/ I will create a PR once https://github.com/CollaboraOnline/online/pull/10145 is merged : )

Issue #7923, opened 2024-01-03
Mention users in shared documents


MelBr01:
### Describe the Bug
When trying to mention users using @ in a document in a shared folder, I get an empty box with the text "No search results found!".

### Steps to Reproduce
1. Create a document in a shared folder.
2. Type @ and the first letter of a user.
3. The above box appears with no search results found.

### Expected Behavior
A list of users that can be mentioned.

### Actual Behavior
A box with no search results found.

### Screenshots
A user with the first letter R has access to the document. A lower case r produces the same results.
<img width="235" alt="image" src="https://github.com/CollaboraOnline/online/assets/149402562/8d49b9a2-b780-4a26-9ac2-af173d72d20e">

### Desktop
(Please complete the following information)
 - **Collabora version:** 23.05.5.4
 - **OS and version:** Windows 11
 - **Browser and version:** Edge 120.0.2210.91 (Official build) (64-bit)

eszkadev:
Which storage integrator do you use? (eg. Nextcloud? which version)

MelBr01:
Nextcloud 27.1.5

timur-g:
There was a related change for 23.05.6.4, please test if that fixes for you.

MelBr01:
No, unfortunately not. Still the same problem in Libra office and in Word....

timur-g:
I do not understand this for LibreOffice and Word. Mention should be possible using Collabora Online 23.05.6 with NextCloud users.

MelBr01:
Unfortunately it is not. Whenever I try to add a mention the same error occurs: 
<img width="321" alt="image" src="https://github.com/CollaboraOnline/online/assets/149402562/bed7fb75-7e61-49a2-8e3c-82430245741d">

sebastiaanveld:
Tested this with NC 27.1.5 and it does also not work for me with CODE 23.05.6.4.

COOLWSD version: 23.05.6.4 (git hash: aff0ccf (E))
LOKit version: Collabora Office 23.05.6.4 (git hash: 058f614)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: 0c34bb0c

![image](https://github.com/CollaboraOnline/online/assets/37268999/0e0361e7-509c-4fa3-9a23-f259214e4c08)

timur-g:
Please see if double @@mention works for you.

timur-g:
This should work in 24.04.9, please test. If not, please explain.

PR #10273, opened 2024-10-20
Debounce mention user query

Rash419:
* Resolves: #10047 & #7923 
* Target version: master 

I tested the following things thoroughly for the PR manually

# Mention test cases

## Desktop View

1. Document

- [x] - insert @mention in document
- [x] - try to search for user that doesn't exist
- [x] - `Esc` should close the mention popup
- [x] - start typing username and press `Backspace` to improve the mention popup list
- [x] - deleting '@' using `Backspace` should close the mention popup
- [x] - pressing special character should close the mention popup
- [x] - popup should follow the cursor if typing continues
- [x] - popup should be closed on `Space` if there are no matches

2. Comment

- [x] - insert @mention when inserting comment
- [x] - insert @mention when modifying comment
- [x] - insert @mention when replying
- [x] - insert @mention when replying to the reply
- [x] - try to search for user that doesn't exist
- [x] - `Esc` should close the mention popup
- [x] - start typing username and press `Backspace` to improve the mention popup list
- [x] -  `Backspace` should close the mention popup when only '@' remains
- [x] - pressing special character should close the mention popup

## Mobile view

1. Document

- [x] - insert @mention in document
- [x] - try to search for user that doesn't exist
- [x] - `Esc` should close the mention popup
- [x] - start typing username and press `Backspace` to improve the mention popup list
- [x] -  `Backspace` should close the mention popup when only '@' remains
- [x] - pressing special character should close the mention popup

2. Comment

- [x] - insert @mention when inserting comment
- [x] - insert @mention when modifying comment
- [x] - insert @mention when replying
- [x] - try to search for user that doesn't exist
- [x] - `Esc` should close the mention popup
- [x] - start typing username and press `Backspace` to improve the mention popup list
- [x] -  `Backspace` should close the mention popup when only '@' remains
- [x] - pressing special character should close the mention popup

### DIFF
--- a/browser/html/debug.html
+++ b/browser/html/debug.html
@@ -53,7 +53,7 @@
 
         function getMentionUserData() {
           const usernames = ['Alexandra', 'Alison', 'Amanda', 'Amelia', 'Amy', 'Andrea', 'Angela', 'Anna',
-                             'Audrey', 'Ava', 'Bella', 'Bernadette', 'Carol', 'Caroline', 'Carolyn'];
+                             'Audrey', 'Ava', 'Bella', 'Bernadette', 'Carol', 'Caroline', 'Carolyn' , 'John Doe', 'John Smith'];
           const userProfileLink = 'https://github.com/CollaboraOnline/online';
           let data = [];
           usernames.forEach(username => data.push({
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -505,14 +505,15 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		}.bind(this);
 
 		const mentionListCallback = function(objectType: any, eventType: any, object: any, index: number) {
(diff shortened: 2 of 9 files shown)


caolanm (approved):
One thing I now see is that if I type "something @caroline" that the popup moves to the right on every additional key press. Which doesn't feel quite right, I would expect it to stay basically at the original popup position as I add extra letters to it.

We could merge this anyway and consider that in a follow up if you prefer (unless the popup is supposed to move?)

Rash419:
> One thing I now see is that if I type "something @Caroline" that the popup moves to the right on every additional key press. Which doesn't feel quite right, I would expect it to stay basically at the original popup position as I add extra letters to it.
> 
> We could merge this anyway and consider that in a follow up if you prefer (unless the popup is supposed to move?)

I made it move with commit aa91cd48d3b5479bf51f68a522663df91deb1476 maybe I am wrong cause other apps don't move the popups : ). Let me update this patch

PR #10275, opened 2024-10-21
browser: do not scale angle on shape rotation

JaumePujantell:
Angle values do not change with dpi. Right now if a browser has a dpi
different than 1 the shape rotation preview jumps around due to this
extra scaling.

Change-Id: I21d7fb34f6c496235004dc75ee512e23b804b386


gokaysatir (approved):
> Angle values do not change with dpi. 

:) You are right, thanks for the catch :)

PR #10278, opened 2024-10-21
slideshow: fix lost GL context

mcecchetti:
It seems that we create too many webgl context because of running transition
filter animations for shapes.
Now we clear the gl context for animated elements at least when we switch to a
new slide.

Change-Id: I708e13c455d69ea8a63fc1b02b3121c8dfd6d568


codewithvk:
Code looks to me but will Test it Locally.

codewithvk (approved):
Looks good

PR #10279, opened 2024-10-21
minimize data sent for writer rich text comments

caolanm:
if its writer then just send the comments as html, not text and html

Change-Id: Ia3e5a20cb4c82474cff951753a6a5fcc78992f3e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
Looks good!

PR #10281, opened 2024-10-21
writer: when navigator is used, follow the cursor - fixes #10282

### LINKED ISSUE #10282: Navigator: doesn't jump on another try
### Describe the Bug
Cannot move view to the entry selected from Navigator second time.

### Steps to Reproduce
1. open any multi page long document with some headers
2. open navigator and double-click any non visible header (it jumps to it)
3. scroll down/up to not see your cursor
4. try to double-click any other navigator entry

### Expected Behavior
Jump to the selected entry from navigator.

### Actual Behavior
Jump doesn't happen sometimes.

eszkadev:
If any JSDialog action is sent - it means user did interaction with dialog / widget / button what should exit follow other user mode. We are now interested in our area of work.

Fixes #10282 10282

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -279,6 +279,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		window.app.console.debug('control: \'' + objectType + '\' id:\'' + object.id + '\' event: \'' + eventType + '\' state: \'' + data + '\'');
 
+		// if user does action - enter following own cursor mode
+		if (app.viewId)
+			app.setFollowingUser(app.viewId);
+
 		if (builder.wizard.setCurrentScrollPosition)
 			builder.wizard.setCurrentScrollPosition();
 
--- a/cypress_test/integration_tests/desktop/writer/navigator_spec.js
+++ b/cypress_test/integration_tests/desktop/writer/navigator_spec.js
@@ -3,13 +3,12 @@
 var helper = require('../../common/helper');
 var desktopHelper = require('../../common/desktop_helper');
 
-describe.skip(['tagdesktop'], 'Scroll through document, modify heading', function() {
+describe(['tagdesktop'], 'Scroll through document, modify heading', function() {
 
 	beforeEach(function() {
 		helper.setupAndLoadDocument('writer/navigator.odt');
 
-		cy.cGet('#menu-view').click();
-		cy.cGet('#menu-navigator').click();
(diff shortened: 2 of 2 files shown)


gokaysatir (changes requested):
It doesn't seem to work reliably.
* Double click on a page header.
* Click elsewhere other than the header's line.
* Scroll a little down so the cursor is not visible now.
* Double click on a upper page header.
* It sets the cursor position but doesn't jump to that position.

Also, setting the following cursor property looks like a little dangerous to me. It may have some side effects. Like jumping to the cursor when user did something that didn't need jumping.

eszkadev:
Thank you for checking. Probably we have some races / message in different order in some cases. I will check. Anyway it was improving the state for me a lot so I would like to have it in RED. Also enabled some cypress tests to protect the navigator from regressions.

Some explanation wrt approach:
The new following feature works this way that it has new mode "off". So we can follow our own view (follow user, by our own id), or if we scroll away to make some reading elsewhere - we automatically turn it to - off.

That is the reason why this bug happens - we turn it off after scroll, we need to reactivate following to actually jump to our cursor. This is expected as "any user interaction" should stop - "editor following" or "off mode" and set it to follow own cursor. I think it's less dangerous than hacks we had previously: `variable which was set before action and if it was detected in some random place - jump to own cursor`.

There are already various tests which check if we don't jump when not needed - which was introduced with the rework.

gokaysatir:
> Thank you for checking. Probably we have some races / message in different order in some cases. I will check. Anyway it was improving the state for me a lot so I would like to have it in RED. Also enabled some cypress tests to protect the navigator from regressions.
> 
> Some explanation wrt approach: The new following feature works this way that it has new mode "off". So we can follow our own view (follow user, by our own id), or if we scroll away to make some reading elsewhere - we automatically turn it to - off.
> 
> That is the reason why this bug happens - we turn it off after scroll, we need to reactivate following to actually jump to our cursor. This is expected as "any user interaction" should stop - "editor following" or "off mode" and set it to follow own cursor. I think it's less dangerous than hacks we had previously: `variable which was set before action and if it was detected in some random place - jump to own cursor`.
> 
> There are already various tests which check if we don't jump when not needed - which was introduced with the rework.

I see, maybe we can put "follow yourself" code into navigator or we can put it to the core side, like "cursorposition - force follow" etc. Anyway, okay let me approve since this improves things.

eszkadev:
Yes, additional attribute from core in such actions would be very useful :)
Thanks.

eszkadev:
followup https://github.com/CollaboraOnline/online/pull/10297

PR #10283, opened 2024-10-22
fix co-24.04 build

caolanm:
the master commit:

commit 9d49ea31131972bc8e1581a353441230850a78c2
CommitDate: Tue Oct 15 06:51:14 2024 +0300

    Add viewMode property to TileLocation to ensure that correct delta is replaced with the new one.

isn't in co-24, so the test of:

commit 0e7d356c960afdf51f0db4b85eda7352472a54ae
CommitDate: Sat Oct 19 18:50:26 2024 +0100

deltas: test random image - to check we have space allocation right

needs to be adjusted to account for that

Change-Id: If7a54fac69f34ab3fa75fee6fe67aea42d44c669

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/DeltaTests.cpp
+++ b/test/DeltaTests.cpp
@@ -330,13 +330,13 @@ void DeltaTests::testRleRandom()
 
     DeltaGenerator::DeltaData data(
         1, randomImg.data(), 0, 0, 256, 256,
-        TileLocation(9, 9, 9, 0, 1, 0), 256, 256);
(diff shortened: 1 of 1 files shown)


timar (approved):

caolanm:
It will never pass ci against core co-24-04 because that's now only sending 'html' instead of 'text' for writer comments

timar:
OK, I can merge this anyway, if it helps.

PR #10288, opened 2024-10-22
Impress: add minimal Curve inserting

meven:
Change-Id: Ie10752545df5eb3b5baefab22c583bfa72611244

### Summary

Add minimal Curve inserting
 
image

UI is not present when drawing

### Checklist


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/1795/console failed in:

> writer/invalidations_spec.js
> writer/table_operation_spec.js

Retry this as well.

browser/src/unocommands.js
```diff
@@ -251,6 +252,7 @@ var unoCommandsArray = {
 	'IncrementLevel':{text:{menu:_('Promote Outline Level'),},},
 	'IncrementSubLevels':{text:{menu:_('Promote Outline Level with Subpoints'),},},
 	'IndexesMenu':{text:{menu:_('Table of Contents and Inde~x'),},},
+	'InsertAVMedia':{global:{context:_('Insert Audio or Video'),menu:_('Media'),},},
```
hfiguiere, line 255: this change should be removed like its counterpart.

meven: Fair.
The script ought to run again, in a different PR.

hfiguiere: we can fix that in a different PR. 

I wonder why this is out of order....

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/2624/console fails in:

> unit-save-torture

retry.

browser/src/control/Control.NotebookbarImpress.js
```diff
@@ -1461,6 +1461,12 @@ L.Control.NotebookbarImpress = L.Control.NotebookbarWriter.extend({
 					}
 				],
 				'vertical': 'true'
+			},
+			{
```
vmiklos, line 1465: My notebookbar now looks like this:

image

Is it expected that ~all other buttons have a label next to the icon, but not this one? I would expect the new button would look similar to the existing line button, or something like that. Thanks.

meven: This was in my screenshot too.
I noticed, I am looking at it. I don't get why it doesn't show up.

hfiguiere: also it's not in Draw....

meven: Now fixed

vmiklos:
Rebase to be on top of online.git commit 7998ba53a533be83ce8fed60cabc44b6457543c4

vmiklos (approved):
The actual behavior looks a bit confusing to me (even on the desktop), but this does expose the feature in COOL anyway, so look OK to me.

PR #10290, opened 2024-10-22
Private/gokay/impress preview refresh

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Parts.js
+++ b/browser/src/control/Parts.js
@@ -210,49 +210,43 @@ L.Map.include({
 		this._previewQueue.push([part, tileMsg]);
 	},
 
-	getPreview: function (id, index, maxWidth, maxHeight, options) {
-		if (!this._docPreviews) {
-			this._docPreviews = {};
-		}
-		var autoUpdate = options ? !!options.autoUpdate : false;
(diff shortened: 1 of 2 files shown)


caolanm:
The first two commits look good. I'm not sure about the 3rd, I think I'd like to see how to reproduce that one myself. (There is a conflict reported, so I think  a rebase is needed anyay)

caolanm:
I think the third patch is working around a (very odd) problem in core impress that switching from notes to standard doesn't actually change the selected page so there ends up previews getting rendered in "standard" mode that actually render the still selected "note" page so note previews can show up pretending to be standard previews. I believe  https://gerrit.libreoffice.org/c/core/+/176084 should fix that. If that works, then we could drop this 3rd patch and the remainder is nice-to-have tidy ups.

gokaysatir:
> I think the third patch is working around a (very odd) problem in core impress that switching from notes to standard doesn't actually change the selected page so there ends up previews getting rendered in "standard" mode that actually render the still selected "note" page so note previews can show up pretending to be standard previews. I believe https://gerrit.libreoffice.org/c/core/+/176084 should fix that. If that works, then we could drop this 3rd patch and the remainder is nice-to-have tidy ups.

Rebased and dropped the 3rd commit. Thanks for the help :)

caolanm (approved):

PR #10297, opened 2024-10-22
user following: do not turn off when typing

eszkadev:
It was old code before following rework which used unfollowAll - now it it too aggressive. When user
does any interaction we should enter "follow me"
mode not "turn off" the following.

this is followup for commit 153974a35cfbfb79cefd859bf1f3623523dfa229 writer: when navigator is used, follow the cursor


Rash419 (approved):
looks good

eszkadev:
writer/sign_spec.js failed (also in many other PRs, unrelated)

PR #10298, opened 2024-10-23
Fix regression in `Socket::shutdown()` when having `_noShutdown` set

sgothel:
Change-Id: I57d46e0428905137e05d34dad450fea1d6a0f127

* Resolves: #9833
* Target version: master 

### Summary
Regression injected with commit 0579ee81c709865ea420c2d70dcc443bdf69de1a of PR https://github.com/CollaboraOnline/online/pull/9916

Original behavior was to do nothing in Socket::shutdown(), while the commit in question did setClosed() leading to an eventual socket close within SocketPoll::poll().

The latter causes an error within system ::ppoll().

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -218,10 +218,10 @@ public:
     /// TODO: Support separate read/write shutdown.
     virtual void shutdown()
     {
-        setClosed();
         if (!_noShutdown)
         {
             LOG_TRC("Socket shutdown RDWR. " << *this);
+            setClosed();
             if constexpr (!Util::isMobileApp())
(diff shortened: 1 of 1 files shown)


caolanm (approved):
These _noShutdown cases are from the bgsave stuff introduced with 0a9a454c968b8b490f3d513114065ad6b41bd4b1

sgothel:
> These _noShutdown cases are from the bgsave stuff introduced with [0a9a454](https://github.com/CollaboraOnline/online/commit/0a9a454c968b8b490f3d513114065ad6b41bd4b1)

Thank you, saw usage at the paired sockets but was unsure about semantics.

PR #10300, opened 2024-10-23
Slideshow: add mipmaps texture rendering for few transition

codewithvk:
Change-Id: I47e5be505f7304e80e06f1492e13638905811085

* Resolves: # 
* Target version: master 

### Summary
Note: It's hard to see the difference between the old version and the new change, but we can ensure that nothing is broken.

### DIFF
--- a/browser/src/slideshow/engine/SlideShowHandler.ts
+++ b/browser/src/slideshow/engine/SlideShowHandler.ts
@@ -902,13 +902,49 @@ class SlideShowHandler {
 		return this.theMetaPres.getSlideInfoByIndex(nSlideIndex);
 	}
 
-	private getTexture(nSlideIndex: number): WebGLTexture | ImageBitmap | null {
+	private isMipMapsEnable(transitionFilterInfo: TransitionFilterInfo): boolean {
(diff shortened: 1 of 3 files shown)


browser/src/slideshow/engine/SlideShowHandler.ts
```diff
@@ -900,13 +900,49 @@ class SlideShowHandler {
 		return this.theMetaPres.getSlideInfoByIndex(nSlideIndex);
 	}
 
-	private getTexture(nSlideIndex: number): WebGLTexture | ImageBitmap | null {
+	private isMipMapsEnable(transitionFilterInfo: TransitionFilterInfo): boolean {
+		const allowedSubtypes = new Set([
```
mcecchetti, line 904: Maybe it would be better to move the set as a global structure `TransitionsWithMipMapEnabled` or any name that makes sense. So we don't need to create it on each `isMipMapEnabled` invocation

codewithvk: make sense, thanks :)

mcecchetti (approved):
Okay, code looks good to me.

PR #10303, opened 2024-10-23
user following: update on cursor invalidation

eszkadev:
so we don't skip cases when there is no scrolling using eg. mouse. this fixes problem in writer:
1. open long, multi page document
2. search for word which exists on every page
3. search using "next" button until you scroll to other page - works
4. scroll using mouse to the top
5. click before first word to put cursors there
6. click multiple times "next" button

result: cursor no longer shown when going to next page
expected: following the cursor all the time

### DIFF
--- a/browser/src/docstatefunctions.js
+++ b/browser/src/docstatefunctions.js
@@ -134,6 +134,36 @@ app.isFollowingEditor = function () {
 	return app.following.mode === 'editor';
 };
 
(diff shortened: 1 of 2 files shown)


gokaysatir (approved):
Works, thank you for the fix.

gokaysatir:
Addition also looks reasonable.

PR #10310, opened 2024-10-24
fix: paste happens in document instead of combobox input

Rash419:
- now text is pasted in combobox input for example fonts, addressinput filed instead of in the document

Change-Id: Idb2ffa1e6a77d8b14fe43a71e2ed728d4d31a959

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -581,6 +581,9 @@ L.Clipboard = L.Class.extend({
 		if ($('.w2ui-input').is(':focus'))
 			return true;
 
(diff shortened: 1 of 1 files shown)


browser/src/map/Clipboard.js
```diff
@@ -581,6 +581,9 @@ L.Clipboard = L.Class.extend({
 		if ($('.w2ui-input').is(':focus'))
 			return true;
 
+		if ($('input.ui-combobox-content').is(':focus'))
+			return true;
```
eszkadev, line 585: good enough for now, we need to look for our helper to detect inputs and unify in all places later

eszkadev (approved):

PR #10311, opened 2024-10-24
cool#9992 doc sign: move sign key/cert init from doc init to sign dispatch

vmiklos:
core.git commit d48264d51891d81f77fcfd77766e1d34ec17412b (cool#9992 lok
doc sign: allow late-init of the sign cert, 2024-10-22) allows sending
the sign key/cert only when actually signing.

Also commit 7998ba53a533be83ce8fed60cabc44b6457543c4 (fix uno:Signature
cypress test, etc., 2024-10-23) already started to annotate
.uno:Signature with sign parameters.

Complete this move by:

1) Extracting the .uno:Signature handling code to a separate function,
   since ChildSession::_handleInput() is large enough already.

2) Poco::Dynamic::Var::extract() can throw an exception if the user
   private info is not an object (but is e.g. a list), guard against
   that.

3) Stop sending the sign cert/key in Document::makeRenderParams().

Change-Id: I426c0dc8b028eb874f0baf6b99f5c803ad8078d7

### DIFF
--- a/common/JsonUtil.hpp
+++ b/common/JsonUtil.hpp
@@ -267,6 +267,15 @@ inline std::map<std::string, std::string> jsonToMap(const std::string& jsonStrin
     return map;
 }
 
+template <typename T>
+Poco::JSON::Object::Ptr makePropertyValue(const std::string& type, const T& val)
+{
+    Poco::JSON::Object::Ptr obj = new Poco::JSON::Object();
+    obj->set("type", type);
+    obj->set("value", val);
+    return obj;
+}
+
 } // end namespace JsonUtil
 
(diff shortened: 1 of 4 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_multi_user/1842/console fails in:

> sheet_operations_spec.js

retry.

vmiklos:

This does the cleanup you suggested, also gets rid of the manual escaping in the string literal, which is ~fine, but I easily get it wrong myself, so I switched to the json writer to get some safety.

caolanm (approved):

PR #10312, opened 2024-10-24
Don't show spreadsheet-toolbar in other docType

pedropintosilva:
The #spreadsheet-toolbar element shows up in every doc type at
the loading time. This means that the user sees a flash at the bottom
and if the integrator has customized COOL via CSS var etc the color of
the bar or its border color then that flash stands out even more.

Change-Id: Ie325bfb841f5aff2a1cdc9711610fa747b6e1d7b

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -356,6 +356,8 @@ L.Control.UIManager = L.Control.extend({
 			this.sheetsBar = JSDialog.SheetsBar(this.map, isDesktop || window.mode.isTablet());
 
 			let formulabarRow = document.getElementById('formulabar-row');
+			let spreadsheetToolbar = document.getElementById('spreadsheet-toolbar');
+			spreadsheetToolbar.classList.remove('hidden');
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -368,7 +370,6 @@ L.Control.UIManager = L.Control.extend({
 
 		if (this.map.isPresentationOrDrawing()) {
 			// remove unused elements
-			L.DomUtil.remove(L.DomUtil.get('spreadsheet-toolbar'));
```
gokaysatir, line 371: I think these are no harm? This one and the one in `docType === 'text'`.

gokaysatir: Unless we have a good reason, i think it maybe better to remove them when loading different files other than Calc.

pedropintosilva: I thought that if we hide it we don't need to remove it, similar as we are doing for `formulabar-row` but yeah probably leaving the remove is a more conservative way to go about it

gokaysatir (approved):

Issue #10316, opened 2024-10-24
server version mismatch


timar:
I saw this in the log:
```
WRN  Client - server version mismatch, disabling browser cache. Expected: 61e86c9385; Actual URI path with version hash: /browser/61e86c93/logging.html| wsd/FileServer.cpp:564
```
This must be wrong, the two hashes are the same, only the expected is longer by two digits.

PR #10317, opened 2024-10-24
cool#10316 fix server version mismatch warning

timar:
We resized the hash to 8 characters, which incidentally worked when our repo was smaller. Now short hash is 10 chars long.

Change-Id: Ib737e9c011488fbf085cfc2be3dd2a9b6e6daa0b

* Resolves: #10316 
* Target version: master 

**Note:** this bug does not effect official CODE/COOL packages. So the impact is small, fortunately. I checked, and on the machine where we build packages, `git log -1 --format=%h` produces 8 character long hashes. Only on my dev machine I saw 10 character long hashes.

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -368,12 +368,7 @@ namespace Util
 
     std::string getCoolVersion() { return std::string(COOLWSD_VERSION); }
 
-    std::string getCoolVersionHash()
-    {
-        std::string hash(COOLWSD_VERSION_HASH);
-        hash.resize(std::min(8, (int)hash.length()));
-        return hash;
-    }
+    std::string getCoolVersionHash() { return std::string(COOLWSD_VERSION_HASH); }
 
     void getVersionInfo(std::string& version, std::string& hash)
     {
--- a/cypress_test/integration_tests/multiuser/calc/sheet_operations_spec.js
+++ b/cypress_test/integration_tests/multiuser/calc/sheet_operations_spec.js
@@ -103,7 +103,7 @@ describe(['tagmultiuser'], 'Check overlays after tab switching/operations', func
 		//	});
 	});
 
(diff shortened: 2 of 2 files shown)


timar:
@Rash419 I've been struggling with this for 2 days. See https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_multi_user/1887/console
You touched this part last time, and I don't know what's wrong. Content of input field is cleared in theory. But in fact it isn't! Expected 'F6', actual 'D8F6'?! 
I do not think it's related to my change in C++...

timar:
Well, I see that other PRs pass. So it made me wonder. Although the browser cache was forcefully set for cypress (`--forcecaching`) it was not in effect because of the bug that I fixed. 
We set it here: https://github.com/CollaboraOnline/online/blob/d10e7f5471623357ef7282891622865248e68417/wsd/FileServer.cpp#L548
Then we switch it off, if hashes mismatch, later: https://github.com/CollaboraOnline/online/blob/d10e7f5471623357ef7282891622865248e68417/wsd/FileServer.cpp#L567
So now it seems we have to check cypress tests, whether they are stable with enabled browser cache!

Rash419:
FWIW, I tried reproducing cypress failure with your patch but I couldn't reproduce it

timar:
```
      cy:command ✔  clear	
      cy:command ✔  get	@input
      cy:command ✔  assert	expected **<input.ui-combobox-content.addressInput.jsdialog>** to be **empty**
      cy:command ✔  get	@input
      cy:command ✔  type	F6{enter}
      cy:command ✔  get	@input
      cy:command ✘  assert	expected **<input.ui-combobox-content.addressInput.jsdialog>** to have value **F6**, but the value was **D8F6**
```
What is this? Autofill?

PR #10319, opened 2024-10-24
fix More... buttons of Calc conditional format dropdowns

timar:
Change-Id: I9377b7bdbf7673100fababcca52d8048579a931f

### DIFF
--- a/browser/src/control/Control.Toolbar.js
+++ b/browser/src/control/Control.Toolbar.js
@@ -201,7 +201,7 @@ function getConditionalFormatMenuElementsImpl(more, type, count, unoCommand, jsd
 		const button = document.createElement('button');
 		button.id = 'more';
 		button.onclick = function() {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Looks good, I'm not sure how this was supposed to work in the past.

Nit: please be more verbose in the commit message next time. :-) Things to mention:

- what is the current state (is it completely broken? or perhaps an edge case does work?). if you see a problem in practice, how to reproduce that as a user?
- describe the problem / root cause (here: moreConditionalFormat() takes 2 arguments, the uno command string is not meant to be quoted, what was the commit that added this questionable line of code?)
- describe the solution (give orders to the codebase)

Thanks.

timar:
It was a regression after 5aa16c946d24d89a2248b8cbde7f59b95053b9b3
Press to More... button had no effect, because frontend code sent the uno command with invalid syntax.

vmiklos:
Aha, that's useful to know. thanks.

PR #10322, opened 2024-10-25
following user: fixes the state

eszkadev:
various scenarios should trigger following of our own view:
- calc address input
- navigator double click on entry
- search bar interaction


gokaysatir (approved):
Thanks for the improvements.

PR #10326, opened 2024-10-25
StreamSocket::checkRemoval: Signal shutdown only if appropriate

sgothel:
Change-Id: I4e3ba962880b77017443236bc33b03ed20dcdd34

* Resolves: #9833 refining PR #9916
* Target version: master 

### Summary
Signal shutdown only if appropriate, allowing to drain bytes instead of hard disconnect. Only in case of SigUtil::getTerminationFlag(), we force disconnection.

SocketPoll::poll() hence checks isClosed() after checkRemoval() for forced Socket removal. A pending signaled shutdown may occure via handlePoll() after processing more I/O data.

Further, remove useless setClosed() in checkRemoval() after asserting !isOpen().

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -688,8 +688,8 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
             }
             else if( _pollSockets[i]->checkRemoval(newNow) )
             {
-                // timed out socket .. also checks
-                // ProtocolHandlerInterface
+                // Socket to be removed (timeout) w/ immediate shutdown/disconnect,
+                // also checks ProtocolHandlerInterface
                 // - http::Session::checkTimeout() OK
                 // - WebSocketHandler::checkTimeout() OK
                 ++itemsErased;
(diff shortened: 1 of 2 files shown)


net/Socket.cpp
```diff
@@ -686,10 +686,10 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                          << " of " << _pollSockets.size() << ") from " << _name);
                 _pollSockets[i] = nullptr;
             }
-            else if( _pollSockets[i]->checkRemoval(newNow) )
+            else if( _pollSockets[i]->checkRemoval(newNow) && _pollSockets[i]->isClosed() )
```
mmeeks, line 689: Very unclear to me why we are adding even more complexity into the center of our already not-entirely obvious & somewhat complex poll loop. Surely that can be done in 'checkRemoval' (?)

Not clear on the 'isClosed'/isOpen semantic we have here - surely these are still open at the time we want to remove them (?)

sgothel: Yes it could - had it this way earlier but then the method API doc got a bit .. err .. lengthily ..
OK will revise.

sgothel: done

net/Socket.cpp
```diff
@@ -1517,31 +1517,28 @@ bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
...
             if (_socketHandler)
             {
-                _socketHandler->onDisconnect();
-                if( isOpen() ) {
-                    // Note: Ensure proper semantics of onDisconnect()
-                    LOG_WRN("Socket still open post onDisconnect(), forced shutdown.");
-                    shutdown(); // signal
-                    closeConnection(); // real -> setClosed()
+                _socketHandler->onDisconnect(); // Note: Ensure proper semantics of onDisconnect()
+                if( isOpen() )
```
Ashod, line 1521: Spacing could also benefit from fixing up: `if (isOpen())`.

The reason this is more readable (I think) is because language statements are (emphatically) not function calls. Making them stand out helps the eye scan faster and more accurately. I suspect the preference to make them look like function calls is to homogenize the code, but in this case it can be misleading to the reader. Ditto for `while`, `for`, `do`, `return`, etc.

PR #10328, opened 2024-10-25
Fire sheetgeometrychanged event when setTileGeometryData is called.

gokaysatir:
Zoom handler calls the setTileGeometryData function and changes the sheet geometry properties. Since this event is not triggered by core side, we weren't firing an event callback. This causes positions of Calc comments to be wrong until a new position event occurs.

This commit fixes the positioning issue.

Change-Id: Icbea7a6cacbce580b0b91b90701fb698f86cef41

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks ok

PR #10329, opened 2024-10-25
Backport: clang: fix unknown uint32_t error

Darshan-upadhyay1110:
fix build with clang 15 after commit e4b51310e268aa05ad83ea6ed79e92885b783eb6 wsd: header cleanup

In file included from common/SigUtil.cpp:14:
./common/SigUtil.hpp:117:50: error: unknown type name 'uint32_t'
    extern "C" { typedef void (*SigChildHandler)(uint32_t); }
                                                 ^

Change-Id: I42020a2049e69693160eb92f49e7c89f5ba94a96

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/SigUtil.hpp
+++ b/common/SigUtil.hpp
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 
 namespace SigUtil


caolanm (approved):
yeah, this happens with different compiler header files already including cstdint so the lack of an explicit include doesn't get noticed

PR #10333, opened 2024-10-26
Private/mmeeks/pollclean

mmeeks:
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -678,25 +678,6 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                 // removed in a callback
                 ++itemsErased;
             }
-            else if (!_pollSockets[i]->isOpen())
-            {
-                // closed socket ..
-                ++itemsErased;
-                LOGA_TRC(Socket, '#' << _pollFds[i].fd << ": Removing socket (at " << i
-                         << " of " << _pollSockets.size() << ") from " << _name);
-                _pollSockets[i] = nullptr;
-            }
-            else if( _pollSockets[i]->checkRemoval(newNow) )
-            {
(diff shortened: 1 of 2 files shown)


net/Socket.cpp
```diff
@@ -1505,13 +1505,13 @@ bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
         const auto durLast =
             std::chrono::duration_cast<std::chrono::milliseconds>(now - getLastSeenTime());
         /// TO Criteria: Violate maximum idle (_pollTimeout default 64s)
-        const bool isIDLE = _pollTimeout > std::chrono::microseconds::zero() &&
-                           durLast > _pollTimeout;
+        const bool isIdle = _pollTimeout > std::chrono::microseconds::zero() &&
```
Ashod, line 1489: Yes, more than a couple of capitalized letters and it fast becomes obnoxious.

Ashod (approved):
Thanks, @mmeeks. Looks great. Restores the original structure and logic. Much more readable too.

sgothel (approved):
With this re-org `virtual Socket::checkRemoval` can be removed and `StreamSocket::checkRemoval` a plain method.
I can add this on top.

As discussed, if `SigUtil::getTerminationFlag()` checkRemoval 
should indeed `closeConnection()` right away to be save,
as polling might be stopped as in TerminatingPoll or kitPoll().

Looks great.

sgothel:
`Restores exception guards` is very good, as `checkRemoval` also called into `_socketHandler->checkTimeout`.
While the latter is not a custom user callback, its more save this way.

Kicked jenkins ..

PR #10334, opened 2024-10-26
Private/mmeeks/debongmore

mmeeks:
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/WebSocketHandler.hpp
+++ b/net/WebSocketHandler.hpp
@@ -225,7 +225,8 @@ protected:
 
     void shutdown(bool goingAway, const std::string &statusMessage) override
     {
-        shutdownImpl(goingAway ? WebSocketHandler::StatusCodes::ENDPOINT_GOING_AWAY :
+        shutdownImpl(_socket.lock(),
+                     goingAway ? WebSocketHandler::StatusCodes::ENDPOINT_GOING_AWAY :
                      WebSocketHandler::StatusCodes::NORMAL_CLOSE, statusMessage,
                      /*hardShutdown=*/ false, /*silentShutdown=*/ false);
     }
@@ -247,13 +248,15 @@ public:
                   const std::string& statusMessage = std::string(),
                   bool hardShutdown = false)
     {
-        shutdownImpl(statusCode, statusMessage, hardShutdown, false);
+        shutdownImpl(_socket.lock(),
+                     statusCode, statusMessage, hardShutdown, false);
     }
 
     /// Don't wait for the remote Websocket to handshake with us; go down fast.
     void shutdownAfterWriting()
     {
-        shutdownImpl(WebSocketHandler::StatusCodes::NORMAL_CLOSE, std::string(),
+        shutdownImpl(_socket.lock(),
+                     WebSocketHandler::StatusCodes::NORMAL_CLOSE, std::string(),
                      true /* hard async shutdown & close */, false);
     }
 
@@ -265,18 +268,19 @@ public:
     }
 
 private:
-    void shutdownSilent()
+    void shutdownSilent(const std::shared_ptr<StreamSocket>& socket)
     {
-        shutdownImpl(WebSocketHandler::StatusCodes::POLICY_VIOLATION /* ignored */,
+        shutdownImpl(socket,
+                     WebSocketHandler::StatusCodes::POLICY_VIOLATION /* ignored */,
                      std::string(), true /* hard async shutdown & close */, true);
     }
 
-    void shutdownImpl(const StatusCodes statusCode,
+    void shutdownImpl(const std::shared_ptr<StreamSocket>& socket,
+                      const StatusCodes statusCode,
                       const std::string& statusMessage,
                       bool hardShutdown,
                       bool silentShutdown)
     {
-        std::shared_ptr<StreamSocket> socket = _socket.lock();
         if (socket)
         {
             const bool silent = _shuttingDown || silentShutdown;
@@ -671,11 +675,15 @@ public:
         if (_pingTimeout.count() > std::numeric_limits<double>::epsilon() &&
             _pingMicroS.average() >= _pingTimeout.count())
         {
-            LOG_WRN("CheckTimeout: Timeout websocket: Ping: last " << _pingMicroS.last() << "us, avg "
-                    << _pingMicroS.average() << "us >= " << _pingTimeout.count() << "us over "
-                    << (int)_pingMicroS.duration() << "s");
-            shutdownSilent();
-            return true;
+            std::shared_ptr<StreamSocket> socket = _socket.lock();
+            if (socket && socket->isIPType()) // Exclude non-IP local sockets
+            {
+                LOG_WRN("CheckTimeout: Timeout websocket: Ping: last " << _pingMicroS.last() << "us, avg "
+                        << _pingMicroS.average() << "us >= " << _pingTimeout.count() << "us over "
+                        << (int)_pingMicroS.duration() << "s, " << *socket);
+                shutdownSilent(socket);
+                return true;
+            }
         }
         if (!_pingMicroS.initialized() || (_pingPeriod > std::chrono::microseconds::zero() &&
                                            now - _pingMicroS.lastTime() >= _pingPeriod))
--- a/Makefile.am
+++ b/Makefile.am
@@ -483,7 +483,7 @@ else
 # Use coolwsd to cleanup jails, if any. If it fails, we may have a broken/old coolwsd binary, remove it to rebuild.
 # A CI box may switch branches without cleaning up the binaries, if coolwsd from a broken branch is used here
 # it will fail all subsequent builds, until it's rebuilt from the new branch. So removing coolwsd after failing is needed.
-CLEANUP_COMMAND=if test -s ./coolwsd; then echo "Cleaning up..." && ./coolwsd --disable-cool-user-checking --cleanup --o:logging.level=trace || rm -f ./coolwsd; fi
+CLEANUP_COMMAND=if test -s ./coolwsd -a -x ./coolwsd; then echo "Cleaning up..." && ./coolwsd --disable-cool-user-checking --cleanup --o:logging.level=trace || rm -f ./coolwsd; fi
 endif
 
 CLEANUP_COVERAGE=rm -rf ${abs_top_srcdir}/gcov; find . -iname "*.gc??" -delete
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1480,40 +1480,40 @@ std::ostream& StreamSocket::stream(std::ostream& os) const
 
 bool StreamSocket::checkRemoval(std::chrono::steady_clock::time_point now)
 {
-    if( isIPType() )
+    if( !isIPType() )
+        return false;
+
+    // Forced removal on outside-facing IPv[46] network connections only
+    const auto durLast =
+        std::chrono::duration_cast<std::chrono::milliseconds>(now - getLastSeenTime());
+    /// TO Criteria: Violate maximum idle (_pollTimeout default 64s)
+    const bool isIdle = _pollTimeout > std::chrono::microseconds::zero() &&
+        durLast > _pollTimeout;
+    /// TO Criteria: Shall terminate?
+    const bool isTermination = SigUtil::getTerminationFlag();
+    if (isIdle || isTermination )
     {
-        // Forced removal on outside-facing IPv[46] network connections only
-        const auto durLast =
-            std::chrono::duration_cast<std::chrono::milliseconds>(now - getLastSeenTime());
-        /// TO Criteria: Violate maximum idle (_pollTimeout default 64s)
-        const bool isIdle = _pollTimeout > std::chrono::microseconds::zero() &&
-            durLast > _pollTimeout;
-        /// TO Criteria: Shall terminate?
-        const bool isTermination = SigUtil::getTerminationFlag();
-        if (isIdle || isTermination )
+        LOG_WRN("CheckRemoval: Timeout: {Idle " << isIdle
+                << ", Termination " << isTermination << "}, "
+                << getStatsString(now) << ", "
+                << *this);
+        if (_socketHandler)
         {
-            LOG_WRN("CheckRemoval: Timeout: {Idle " << isIdle
-                    << ", Termination " << isTermination << "}, "
-                    << getStatsString(now) << ", "
-                    << *this);
-            if (_socketHandler)
-            {
-                _socketHandler->onDisconnect();
-                if( isOpen() ) {
-                    // Note: Ensure proper semantics of onDisconnect()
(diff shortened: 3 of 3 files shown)


mmeeks:
Wow - the more I read here, the more scared I get; some questions.

* Did we really add this "close websockets" functionality ? - and why was that hidden in this huge unrelated timeout change ?
* where did the 2 second timeout number come from ?
* are/were we really applying that to kit, and background-save process' websockets ?

Then there are more questions:

    /// TO Criteria: Violate maximum idle (_pollTimeout default 64s)
    const bool isIdle = _pollTimeout > std::chrono::microseconds::zero() &&
        durLast > _pollTimeout;
    /// TO Criteria: Shall terminate?
    const bool isTermination = SigUtil::getTerminationFlag();
    if (isIdle || isTermination )

What is this 'isIdle' condition I just renamed ? - why is the _pollTimeout related in any way to when we should close sockets ? perhaps it should have the same value - but the _pollTimeout - is the default amount of time the poll runs when there is no event - it seems totally conceptually unrelated to the amount of time we should wait before closing sockets - luckily it is plausible time.

However as a constant - not a member it should not have an '_' prefix - it looks like an instance variable.

 What does the 'TO Criteria:' mean ? in the comment - can we make that more readable, less shouty, and so on ?

Concerned really.

net/WebSocketHandler.hpp
```diff
@@ -674,7 +674,8 @@ class WebSocketHandler : public ProtocolHandlerInterface
             LOG_WRN("CheckTimeout: Timeout websocket: Ping: last " << _pingMicroS.last() << "us, avg "
                     << _pingMicroS.average() << "us >= " << _pingTimeout.count() << "us over "
                     << (int)_pingMicroS.duration() << "s");
-            shutdownSilent();
+            if( isIPType() )
+                shutdownSilent();
```
sgothel, line 678: Outch, missed the internal kit usage obviously, as also not considered in the modified orig code.
Thank you!

sgothel:
>     * Did we really add this "close websockets" functionality ? - and why was that hidden in this huge unrelated timeout change ?

It was included in the PR #9916 description
as well as in the commit of the patch series
https://github.com/CollaboraOnline/online/pull/9916/commits/994f8f9b2610caa3846fbbac6c4cf612901de0e9

By memory, it was discussed with you and/or Ash .. i.e. 
ping as the only reliable alive-checking instrument we have.
I then explained WS native-ping vs COOL protocol-ping.
We didn't discuss ICMP ping .

> 
>     * where did the 2 second timeout number come from ?

The original code had a period==timeout of 18s.

This code has a period of 3s and a timeout of 2s.
Timing was also described in above mentioned artifacts.

However, this timeout is not triggered in case of a one-off,
but over the average of the total connection duration (-> `TimeAverage`).
It renders the latency tolerance much higher in case of drop-outs,
the longer the connection is well alive. E.g. a sporadic 120s lag after
a 120s good responsive connection duration of 250ms would
result in the average of 1.25s and remains alive.
Only after multiple such extreme failures/lags occur, we would drop the connection.
Hence current 2s on average becomes much more relaxed than original fixed 20s.

[World Ping Test - global ping test](https://www.meter.net/tools/world-ping-test/),
for ICMP ping, demonstrates latency of < 500ms (Sydney w/ ~333ms) at least for me.

Most other utilities I have re-checked now seem to use higher fixed values
ranging between 10s-120min for the timeout - leaning more towards the low-end.
- 5s: [node.js default server.keepAliveTimeout](https://betterstack.com/community/guides/scaling-nodejs/nodejs-timeouts/)
- 18s: Original code had a period==timeout
- 20s: Python WebSockets https://websockets.readthedocs.io/en/latest/topics/keepalive.html
- 20s: grpc https://github.com/grpc/grpc/blob/master/doc/keepalive.md

> 
>     * are/were we really applying that to kit, and background-save process' websockets ?

Originally the WS ping/pong used (fixed 18s)  _without_ disconnecting.
What I understood your commit https://github.com/CollaboraOnline/online/pull/10334/commits/8835ff0d892c15c08d2b994cc00e04201fdd6ee7
resolved this by taking out non-IP based sockets.

I re-reviewed the referenced documentation and the code and will followup on it in a bit.

sgothel:
> What is this 'isIdle' condition I just renamed ? - why is the _pollTimeout related in any way to when we should close sockets ? perhaps it should have the same value - but the _pollTimeout - is the default amount of time the poll runs when there is no event - it seems totally conceptually unrelated to the amount of time we should wait before closing sockets - luckily it is plausible time.

Yes, its about inactivity (badly named idle) and wrongly named _pollTimeout
due to using the same value as SocketPollTimeout (64s).
- isIdle -> isInactive
- _pollTimeout -> inactivityTimeout

Note that this inactivity is disrupted by WS ping, hence also caps the WS ping average timeout to 64s.

> 
> However as a constant - not a member it should not have an '_' prefix - it looks like an instance variable.
An immutable private entity like previous code's `_hostname`.

> 
> What does the 'TO Criteria:' mean ? in the comment - can we make that more readable, less shouty, and so on ?
Sorry, TO is the common abbreviation for timeout - so I should have used lower case `criteria` I guess.

sgothel:
Rebased and adjusted the `WebSocketHandler::checkTimeout` code (compile clean and working w/ product & unit testing).

sgothel:
> I re-reviewed the referenced documentation and the code and will followup on it in a bit.

Considering [Zato's websocket-timeouts](https://zato.io/en/blog/websocket-timeouts.html)
using a missed ping threshold and above mentioned ~20s timeout value,
I prepare a commit considering `threshold * avg-timeout`. 

The `threshold` allows
- to miss `threshold-1` pings, 
- resets the timeout criteria if one pong has been received
- essentially expands the timeout to `threshold * avg-timeout`, i.e. `5 * 2s = 20s`
- goes well along with TimeAverage and the lower `avg-timeout`

Bottom line, should render our WS ping more robust.

sgothel:
> Yes, its about inactivity (badly named idle) and wrongly named _pollTimeout due to using the same value as SocketPollTimeout (64s).
> 
>     * isIdle -> isInactive
> 
>     * _pollTimeout -> inactivityTimeout
> 
> 
> Note that this inactivity is disrupted by WS ping, hence also caps the WS ping average timeout to 64s.

Adding a commit to this PR using above naming scheme, expanding the inactivity timeout to 3600s and adding a unit test.

The fix for WS Ping will go into another PR addressing #10337

sgothel:
> Adding a commit to this PR using above naming scheme, expanding the inactivity timeout to 3600s and adding a unit test.

Resulted in the forced push'ed three commits
- Remove virtual of checkRemoval and cleanup var names _AND_ using 3600s inactivity timeout
- Unit test cleanup/update
- Adding a unit test for inactivity testing

sgothel (approved):
Good as-is with the additional commits re inactivity/checkRemoval.
Would be great to have a re-review by @mmeeks.
I will use this to add the PR for #10337

net/WebSocketHandler.hpp
```diff
@@ -671,11 +675,15 @@ class WebSocketHandler : public ProtocolHandlerInterface
         if (_pingTimeout.count() > std::numeric_limits<double>::epsilon() &&
             _pingMicroS.average() >= _pingTimeout.count())
         {
-            LOG_WRN("CheckTimeout: Timeout websocket: Ping: last " << _pingMicroS.last() << "us, avg "
-                    << _pingMicroS.average() << "us >= " << _pingTimeout.count() << "us over "
-                    << (int)_pingMicroS.duration() << "s");
-            shutdownSilent();
-            return true;
+            std::shared_ptr<StreamSocket> socket = _socket.lock();
```
mmeeks, line 678: Why the lock and the copy when this comes in as a parameter ?

mmeeks: I'm quite disturbed by closing connections on longer average pings too - where does that come from ?

net/Socket.hpp
```diff
@@ -1067,7 +1063,7 @@ class StreamSocket : public Socket,
                  HostType hostType, ReadType readType = ReadType::NormalRead,
                  std::chrono::steady_clock::time_point creationTime = std::chrono::steady_clock::now() ) :
         Socket(fd, type, creationTime),
-        _pollTimeout( net::Defaults::get().SocketPollTimeout ),
+        _inactivityTimeout( net::Defaults::get().InactivityTimeout ),
```
mmeeks, line 1066: Why do we copy this number into a member variable - rather than using it directly in the arithmetic ? surely getting that number from the defaults is far from slow - and it would then be clearer, simpler, fewer lines of code, less memory used for the socket class and so on (?). Please fix that.

sgothel: I have cleaned up these parameter for a WIP for #10337, only querying the net::Defaults::get() once.
Due to the singleton magic nature of net::Defaults::get() (for UTs) it must be queried
to get the thread-safe single instance.

net/WebSocketHandler.hpp
sgothel: > I'm quite disturbed by closing connections on longer average pings too - where does that come from ?

I described the source (and idea) of WS ping TO instrument https://github.com/CollaboraOnline/online/pull/10334#issuecomment-2439712761 as well as wrapped it up in #10337 

I am currently about to finish my draft PR for #10337  and will post its initial version,
currently refining the UTs for it.

sgothel: > Why the lock and the copy when this comes in as a parameter ?

the ` _socket.lock()` has moved up to `checkTimeout` to cover for the `socket->isIPType()` check, which you have correctly introduced with your change request (but not coded all out).

Ashod:
> By memory, it was discussed with you and/or Ash .. i.e. ping as the only reliable alive-checking instrument we have. I then explained WS native-ping vs COOL protocol-ping. We didn't discuss ICMP ping .

We discussed this exact point privately, yes. Let me quote myself:
> Negative. TCP has a perfectly good way for this (with some footnotes). We simply poll the socket handle, and we get back the state, if it is closed. [...]

So it would seem I didn't quite agree that "pings are the only reliable alive-checking instrument." Indeed, we already detect closed sockets, in `SocketPoll`, and remove them.  [Here is another comment on the original PR](https://github.com/CollaboraOnline/online/pull/9916#discussion_r1747041747), also questioning the approach taken.

To repeat what I've been advocating:
- If we want to detect **dead** connections, relying on TCP and poll(2) seems straightforward (as we already do).
- If we need to detect **idle** connections, checking the last data transfer time against a threshold (timeout) would do the trick very reliably (`checkTimeout()` is a good place to do it, and we already do that for HTTP GET).
- If we need to detect **slow** connections, then we need both a _rolling_ average data-rate and a threshold. We might also want to enforce this logic only when we have too many open sockets, or based on some other criteria, but not strictly at all times. Again, `checkTimeout()` is the obvious place to do that.

But most certainly we have to be extremely vigilant _not to close WebSocket_ connections blindly and without justification, which are the lifelines of documents and their life-cycle.
Most critically, the internal connection between a Kit and coolwsd **cannot be _externally_ interrupted** for any reason, whatsoever.

Hope this clarifies where I stand. Happy to discuss where there is a mismatch between my understanding/position and the requirements, of course.

net/WebSocketHandler.hpp
```diff
@@ -674,7 +674,8 @@ class WebSocketHandler : public ProtocolHandlerInterface
             LOG_WRN("CheckTimeout: Timeout websocket: Ping: last " << _pingMicroS.last() << "us, avg "
                     << _pingMicroS.average() << "us >= " << _pingTimeout.count() << "us over "
                     << (int)_pingMicroS.duration() << "s");
-            shutdownSilent();
+            if( isIPType() )
```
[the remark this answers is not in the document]
sgothel: > Would this disconnect the internal connection between coolwsd and kit?

yes, internal connections are of type unix to the LocalServerSocket, validated.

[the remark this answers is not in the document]
sgothel: Re net::Default configuration:
- Its configuration via `configNet` UT callback  was added in parallel since it is beyond the WSD scope,
  where the general WSD configuration resides.
- Lack of thread sync is due to its lifecycle the same reason as WSD configuration, 
  i.e. we call `configNet` pre-creation of sockets etc.
- It is correct that we _had_ duplication of maximum sockets (ahem), which will be  revised due to changed semantics. However, the build global will influence this metric. Not too great, I agree.

sgothel:
> To repeat what I've been advocating:
> 
>     * If we want to detect **dead** connections, relying on TCP and poll(2) seems straightforward (as we already do).

Yup.

> 
>     * If we need to detect **idle** connections, checking the last data transfer time against a threshold (timeout) would do the trick very reliably (`checkTimeout()` is a good place to do it, and we already do that for HTTP GET).

We had this in place for replies.

Now we have general client inactivity (set to 1 hour, with revised commits in this branch).
(This was wrong earlier using just the 64s timeout).

The idle connection criteria was part of the requirement.

This metric is checked in `StreamSocket::checkRemoval` for outside facing connections,
called in parallel to virtual override `checkTimeout`.

> 
>     * If we need to detect **slow** connections, then we need both a _rolling_ average data-rate and a threshold. 
>

In line with proposal #10337 and described here above.
Thank you for mentioning the rolling or moving average characteristic here (I missed to mention that yet).

> We might also want to enforce this logic only when we have too many open sockets, or based on some other criteria, but not strictly at all times.
> 

This is a great addition. Hence we would want an addition property like `LowFreeSockets...`,
a percentage of the used maximum sockets. The latter is the hard-cut-off for new connections but different from build-in maximum session connection, as discussed above

> Again, `checkTimeout()` is the obvious place to do that.

Yes, this is covered with the WS ping timeout within its `checkTimeout` override.

This overall slow connection criteria was part of the requirement.

> 
> But most certainly we have to be extremely vigilant _not to close WebSocket_ connections blindly and without justification, which are the lifelines of documents and their life-cycle. Most critically, the internal connection between a Kit and coolwsd **cannot be _externally_ interrupted** for any reason, whatsoever.
> 
> Hope this clarifies where I stand. Happy to discuss where there is a mismatch between my understanding/position and the requirements, of course.

Yup, thank you.

Revising the work within this branch.

sgothel:
failed desktop cypress test `writer/table_operation_spec.js` not reproduced locally, kicking jenkins.

net/WebSocketHandler.hpp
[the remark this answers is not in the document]
sgothel: WIP: Dropping the net::Defaults instance and `configNet` (using `configure`) as requested.

sgothel:
Rebased and added commit 61f5b7a022cd635a5a6cfd224b298655c4458e8a

mmeeks:
Fixed CI failure mode in unit tests - only execute executable coolwsd ;-) but ... same Cypress test fails locally - interestingly.

sgothel:
My cleanup work continues here https://github.com/CollaboraOnline/online/pull/10355

PR #10335, opened 2024-10-26
Better log message for unexpected css vars

timar:
I've seen for example:
```
ERR  Skipping the token [--nc-light-color-primary-default=] since it does not have '='| wsd/FileServerUtil.cpp:405 
ERR  Skipping the token [--nc-dark-color-primary-default=] since it does not have '='| wsd/FileServerUtil.cpp:405

```
The tokens had '=', so the message was misleading. On the other hand, if we report these as errors, the users may think it's something serious. Better to report them as warnings.

Change-Id: I9f9a31c49ca171f44d52cec9be42ff2953f1b9f4

### DIFF
--- a/wsd/FileServerUtil.cpp
+++ b/wsd/FileServerUtil.cpp
@@ -399,15 +399,21 @@ std::string FileServerRequestHandler::cssVarsToStyle(const std::string& cssVars)
     StringVector tokens(StringVector::tokenize(cssVars, ';'));
     for (const auto& token : tokens)
     {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Looks OK to me, it's easy to confuse the "no value" vs "empty value" cases.

GitHub says you have a conflict, could you please rebase and then get this in? Thanks.

timar:
@Ashod fixed the problem differently in #10399 so it does not make sense to have my patch.

PR #10339, opened 2024-10-27
cypress: remove unused executeCopyFromContextMenu

timar:
It would not work anyway, because vex dialogs are not used any more.

Change-Id: I6bdd1d5b72823106d064b18dafc7802cd8518e9f


gokaysatir (approved):
Thanks for the catch :)

PR #10347, opened 2024-10-28
Option to hide server signature in HTTP response

timar:
Hiding the server signature reduces the likelihood of targeted attacks by making it harder for attackers to identify the server's configuration.

In built-in default configuration security.server_signature is set to false. It is also set to false in coolwsd.xml packaged with production builds. HTTP response header is like:
    Server: COOLWSD HTTP Server
In debug mode security.server_signature is set to true in coolwsd.xml. HTTP response header is like:
    Server: COOLWSD HTTP Server 24.04.9.1

Change-Id: Ie23c2dbc1514e46e2942f558ada2e27d0f108bea

### DIFF
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -47,6 +47,7 @@ namespace
 std::thread TimeoutThread;
 std::mutex TimeoutThreadMutex;
 std::condition_variable TimeoutConditionVariable;
+bool CppunitTesting = false;
 
 } // namespace
 
@@ -352,6 +353,16 @@ void UnitBase::rememberInstance(UnitType type, UnitBase* instance)
         assert(false);
         break;
     }
+}
+
+void UnitBase::setCppunitTesting(bool cppunitTesting)
+{
+    CppunitTesting = cppunitTesting;
+}
+
+bool UnitBase::isCppunitTesting()
+{
+    return CppunitTesting;
 }
 
 int UnitBase::uninit()
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -145,6 +145,9 @@ public:
     /// Returns 0 on success.
     static int uninit();
 
+    static void setCppunitTesting(bool cppunitTesting);
+    static bool isCppunitTesting();
+
     /// Do we have a unit test library hooking things & loaded
     static bool isUnitTesting()
     {
--- a/configure.ac
+++ b/configure.ac
@@ -421,6 +421,7 @@ bundle_msg="using uglified bundled JS and CSS"
 LOK_LOG_ASSERTIONS=0
 log_asserts_msg="disabled"
 SSL_VERIFY="true"
+SERVER_SIGNATURE="false"
 
 # a reasonable default
 NUM_PRESPAWN_CHILDREN=4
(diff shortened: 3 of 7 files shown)


timar:
@Ashod I thought it would be an easy hack, but it turned out that standalone `unithttplib` test neither has access to config, nor the `UnitBase::isUnitTesting()` is usable. I don't want to create an ugly hack. Can you please advise?

vmiklos:
I would suggest introducing a flag that can track if this is a cppunit test, this would be quite similar to how core.git has a flag to track if we're fuzzing (and in that case we don't access the config). Something like this:

```cpp
diff --git a/common/Unit.cpp b/common/Unit.cpp
index 7b83dbdae4..8138466098 100644
--- a/common/Unit.cpp
+++ b/common/Unit.cpp
@@ -47,6 +47,7 @@ namespace
 std::thread TimeoutThread;
 std::mutex TimeoutThreadMutex;
 std::condition_variable TimeoutConditionVariable;
+bool CppunitTesting = false;
 
 } // namespace
 
@@ -354,6 +355,16 @@ void UnitBase::rememberInstance(UnitType type, UnitBase* instance)
     }
 }
 
+void UnitBase::setCppunitTesting(bool cppunitTesting)
+{
+    CppunitTesting = cppunitTesting;
+}
+
+bool UnitBase::isCppunitTesting()
+{
+    return CppunitTesting;
+}
+
 int UnitBase::uninit()
 {
     // Only in debug builds do we support tests.
diff --git a/common/Unit.hpp b/common/Unit.hpp
index 21e0cd7ebe..274d08de8c 100644
--- a/common/Unit.hpp
+++ b/common/Unit.hpp
@@ -145,6 +145,9 @@ public:
     /// Returns 0 on success.
     static int uninit();
 
+    static void setCppunitTesting(bool cppunitTesting);
+    static bool isCppunitTesting();
+
     /// Do we have a unit test library hooking things & loaded
     static bool isUnitTesting()
     {
diff --git a/net/Socket.cpp b/net/Socket.cpp
index 18d629e30e..adcc96f64a 100644
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1736,7 +1736,7 @@ std::string getAgentString() { return "COOLWSD HTTP Agent " + Util::getCoolVersi
 
 std::string getServerString()
 {
-    if (!UnitBase::isUnitTesting())
+    if (!UnitBase::isUnitTesting() && !UnitBase::isCppunitTesting())
         if (config::getBool("security.server_signature", false))
             return "COOLWSD HTTP Server " + Util::getCoolVersion();
     return "COOLWSD HTTP Server";
diff --git a/test/HttpRequestTests.cpp b/test/HttpRequestTests.cpp
index f18eafe305..3de3879e3b 100644
--- a/test/HttpRequestTests.cpp
+++ b/test/HttpRequestTests.cpp
@@ -108,6 +108,7 @@ public:
             = new Poco::Net::Context(Poco::Net::Context::CLIENT_USE, sslParams);
         Poco::Net::SSLManager::instance().initializeClient(nullptr, std::move(invalidCertHandler), std::move(sslContext));
 #endif
+        UnitBase::setCppunitTesting(true);
     }
 
     ~HttpRequestTests()
```

vmiklos (approved):
Looks good to me now, thanks.

PR #10348, opened 2024-10-28
fix: show error to user when online fails to download clipboard on pasting

Rash419:
follow up of e9a0eeb78dcc

Change-Id: I7c3cf0cbb0b0b945b6144d9b2b6a2d8cf02a0a2c

* Target version: master

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -334,9 +334,13 @@ L.Clipboard = L.Class.extend({
 				);
 			}.bind(this),
 			function(progress) { return progress/2; },
-			function() {
+			function(response) {
 				window.app.console.log('failed to download clipboard using fallback html');
 
+				if (response) {
+					app.showAsyncDownloadError(response, _('Failed to download clipboard'));
+					return;
+				}
 				// If it's the stub, avoid pasting.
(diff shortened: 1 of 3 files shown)


Rash419:
Don't merge until I say so I am building docker to test the functionality in cluster environment

eszkadev (approved):

Rash419:
Okay to merge now, it works as expected:
image

browser/src/map/Clipboard.js
```diff
@@ -334,9 +334,20 @@ L.Clipboard = L.Class.extend({
 				);
 			}.bind(this),
 			function(progress) { return progress/2; },
-			function() {
+			function(response) {
 				window.app.console.log('failed to download clipboard using fallback html');
 
+				if (response) {
+					const reader = new FileReader();
```
eszkadev, line 341: this code is duplicated, please create shared function somewhere

eszkadev: eg. in docstatefunctions.js then it is accessible using app.nameOfFunction

eszkadev (changes requested):

eszkadev (approved):

PR #10351, opened 2024-10-28
ContextMenu: add a css class to entries

meven:
Change-Id: I71ce81efcbbb0c6a57d808d5f5d0e691567dfad5

* Resolves: # 
* Target version: master 

### Summary

Context menu entries will now get a class corresponding to the uno
command name associated, in a snake-case format.

I.e InsertAnnotation => insert-annotation

This allows to change style individual context menu entry or even remove them.

### Checklist

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -317,9 +317,13 @@ L.Control.ContextMenu = L.Control.extend({
 					itemName = _UNO(item.command, docType, true);
 				}
 
+				var toSnakeCase = function (text) {
+					return text.replace(/[ _]/gi, '-').replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+				};
+
 				contextMenu[item.command] = {
 					// Using 'click' and <a href='#' is vital for copy/paste security context.
-					name: (window.mode.isMobile() ? _(itemName) : '<a href="#" class="context-menu-link">' +  _(itemName) + '</a'),
+					name: (window.mode.isMobile() ? _(itemName) : '<a href="#" class="context-menu-link ' + toSnakeCase(commandName) + '">' + _(itemName) + '</a'),
 					isHtmlName: true,
 				};
 


meven:
There is the `L.installContextMenu` that it doesn't cover.
I am not sure this is needed.

hfiguiere:
In Control.JSDialogBuilder.js there is `_createIconURL` which turn a UNO command name into and icon name. I think we should share the same logic here to be consistent.

hfiguiere:
the `:has()` pseudo selector might restrict which browsers this is compatible to. It's relatively recent in Firefox.

meven:
> In `Control.JSDialogBuilder.js` there is `_createIconURL` which turn a UNO command name into and icon name. I think we should share the same logic here to be consistent.

The itemName here is not an uno command but some descriptive text from translation:
Uno command: uno:InsertAnnotation
itemName: _Insert_Comment
I wanted to match the end user point of view.

I could work with the uno-command, and use then the "insert-annotation" class instead. It would be consistent with the rest of the logic.

This function reuses same entries for different actions, I don't think it matches well my need:
```
			'insertannotation': 'shownote',
			'insertcomment': 'shownote',
			'objecttitledescription': 'shownote',
			'namegroup': 'shownote',
```

> the `:has()` pseudo selector might restrict which browsers this is compatible to. It's relatively recent in Firefox.

Since `121 (Released 2023-12-19)` and Firefox Mobile `121 (Released 2023-12-19)`. That's the first time I've been using it indeed.
I could try to put the class elsewhere in the dom to simplify the required CSS to hide the entry, but it would complicate more the change.
It is still up to the user to add the CSS.

meven:
Updated to use the more stable uno command as class, i.e InsertAnnotation => insert-annotation

hfiguiere:
LGTM but I think this should have a ack from someone on the dev team.

pedropintosilva:
multiple asserts failing https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2211/consoleFull but all of them seem unrelated. Pressing retry

meven:
@pedropintosilva CI passed (finally), there was spurious build issue
Do you mind approving this ?

pedropintosilva (commented):
thanks this is a very good idea so, we can then if needed emphasize certain context menu entries etc!

The possible issue I see is that I don't follow why we need `toSnakeCase` or at least why we need the replacement of `...`. As far as I understand commandName never has `...`. The item.text on the other hand can have ellipsis but you are anyhow already commandName

browser/src/control/Control.ContextMenu.js
```diff
@@ -309,9 +309,13 @@ L.Control.ContextMenu = L.Control.extend({
 					itemName = _UNO(item.command, docType, true);
 				}
 
+				var toSnakeCase = function (text) {
+					return text.replace(/[ _]/gi, '-').replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase().replace('...', '');
+				};
+
 				contextMenu[item.command] = {
 					// Using 'click' and <a href='#' is vital for copy/paste security context.
-					name: (window.mode.isMobile() ? _(itemName) : '<a href="#" class="context-menu-link">' +  _(itemName) + '</a'),
```
eszkadev, line 314: itemName is translated?  then it makes no sense

eszkadev:
Could we put in the commit message some explanation? what exactly do we want to achieve here?
- sample of "before/after result"
- why that class is needed?
This helps to understand -> review and later debug.

As I assume it is for icon?

browser/src/control/Control.ContextMenu.js
eszkadev: commandName is used -> ok,    I read wrong ;)

browser/src/control/Control.ContextMenu.js
```diff
@@ -309,9 +309,13 @@ L.Control.ContextMenu = L.Control.extend({
 					itemName = _UNO(item.command, docType, true);
 				}
 
+				var toSnakeCase = function (text) {
+					return text.replace(/[ _]/gi, '-').replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase().replace('...', '');
```
eszkadev, line 313: what if we have more than 2 words? or single one?

can't we just do `split` and then `join` ?

meven:
> The possible issue I see is that I don't follow why we need `toSnakeCase` or at least why we need the replacement of `...`. As far as I understand commandName never has `...`. The item.text on the other hand can have ellipsis but you are anyhow already commandName

The '...' is the remnant of previous design.
`toSnakeCase` is used to make uno command more css class names looking.
InsertAnnotation => insert-annotation
So css in the end looks better.

browser/src/control/Control.ContextMenu.js
meven: '/g' ensures it applies on all the string:

toSnakeCase("InserAnnotationTest")
=> "inser-annotation-test"

toSnakeCase("Insert test")
=> "insert-test"

toSnakeCase("Insert")
=> "insert"

meven:
> Could we put in the commit message some explanation? what exactly do we want to achieve here?
> 
>     * sample of "before/after result"
> 
>     * why that class is needed?
>       This helps to understand -> review and later debug.
> 
> 
> As I assume it is for icon?



> Could we put in the commit message some explanation? what exactly do we want to achieve here?
> 
>     * sample of "before/after result"
> 
>     * why that class is needed?
>       This helps to understand -> review and later debug.
> 
> 
> As I assume it is for icon?

I have added to the commit text:
```
Context menu entries will now get a class corresponding to the uno
command name associated, in a snake-case format.

I.e InsertAnnotation => insert-annotation

This allows to change style individual context menu entry or even remove them.
```

The current focus of this is to be able to hide context menu entries through CSS.

Note: We will need to have a filter for contextMenu entries following `Hide_Command` postMessage API.

eszkadev:
> 
> The current focus of this is to be able to hide context menu entries through CSS.
> 
> Note: We will need to have a filter for contextMenu entries following `Hide_Command` postMessage API.

Ok, but later for postmessage I would like to have register of commands which we hide in JS. Like we do for toolbars.
We can also add universal command which will apply that to all components based on uno command if we don't have one.
No need to do it separately for NB, toolbar, menu, context menu probably?

Also it would be usefult to "not create entries" instead of "hiding existing entries".
User can always modify css in the developer tools :)

hfiguiere:
> Ok, but later for postmessage I would like to have register of commands which we hide in JS. Like we do for toolbars.

I mentionned this in discussion that we should hook at least the Show/Hide_Command to the code that build the contextual menus.

PR #10354, opened 2024-10-28
browser: slideshow: add more window feature options

hcvcastro:
Some browsers, such as Firefox, ignore height and width popup windows,
so it is better to specify more options.

Change-Id: I8076fdb4bf6e3ac95a3f18496bfea148d8948d11

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -217,6 +217,45 @@ class PresenterConsole {
 				':' +
 				String(sec).padStart(2, '0');
 		}
+
+		if (this._ticks % 2 == 0 && typeof this._lastIndex !== 'undefined') {
+			setTimeout(
+				L.bind(
+					this._fetchPreview,
+					this,
+					this._lastIndex + 1,
+					this._proxyPresenter.document.querySelector('#next-presentation'),
+				),
+				0,
+			);
+		}
+	}
+
+	_fetchPreview(index, elem) {
+		if (index >= this._presenter._getSlidesCount()) {
+			return;
+		}
+
+		let rect = elem.getBoundingClientRect();
+		if (rect.width === 0 && rect.height === 0) {
+			return;
+		}
+
+		let preview = this._previews[index];
+		if (preview) {
+			this._lastIndex = index;
+			return;
+		}
+
+		let size = this._map.getPreview(2000, 0, rect.width, rect.height, {
+			fetchThumbnail: false,
+			autoUpdate: false,
+		});
+
+		this._map.getPreview(2000, index, size.width, size.height, {
+			autoUpdate: false,
+			slideshow: true,
+		});
 	}
 
 	_onWindowClose() {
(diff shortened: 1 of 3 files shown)


browser/src/slideshow/PresenterConsole.js
```diff
@@ -204,6 +209,46 @@ class PresenterConsole {
 				':' +
 				String(sec).padStart(2, '0');
 		}
+
+		if (this._ticks % 2 == 0) {
+			setTimeout(
+				L.bind(
+					this._fetchPreviews,
+					this,
+					this._proxyPresenter.document.querySelector('#next-presentation'),
```
mcecchetti, line 218: Nice to fetch all thumbnails in advance :-)

browser/src/slideshow/PresenterConsole.js
```diff
@@ -257,6 +304,18 @@ class PresenterConsole {
...
+	_onTransitionStart(e) {
+		if (!this._proxyPresenter) {
+			return;
+		}
+
+		this._currentIndex = e.slide;
+
+		let next =
+			this._proxyPresenter.document.querySelector('#next-presentation');
+		this.drawNext(next);
```
mcecchetti, line 316: Shall you limit `_onTransitionStart` to prefecthing ? It's a bit odd to see the next next slide when the next one has not yet been displayed.

mcecchetti:
Hi Henry, the main issue I see here is that at least for me locally it doesn't fix the reported problem.  The next slide preview is not visible until the presenter console window is resized.
I make a try and it seems that just invoking `_onResize()` at the end `_onPresentInConsole` solve the problem. That is, we simulate a resize :-)

hcvcastro:
Thanks for testing in windows, I did the simulate resize

browser/src/slideshow/PresenterConsole.js
hcvcastro: Changes done. Thanks

browser/src/slideshow/PresenterConsole.js
```diff
@@ -266,6 +316,14 @@ class PresenterConsole {
 		}
 	}
 
+	_onTransitionStart(e) {
+		if (!this._proxyPresenter) {
+			return;
+		}
+
+		this._currentIndex = e.slide;
+	}
```
mcecchetti, line 325: So, now we're not fetching at all, here. My hint was just to avoid to display the preview in advance.
Anyway, if you think that pre-fetching every 2 sec is enough and we don't need anymore to do it here, this method could be completely removed since it does nothing.
On the contrary if you think that it could be helpful to fetch next slide here, you could extract the fetching part from `drawNext` so that we can just fecth the next slide preview without showing it.
Please, do whatever you think it's more appropriate :-)

hcvcastro: Oops, sorry I forgot to include the stage code. Fixed it.

mcecchetti: Okay, I see.
Maybe it would be better to pass the index value to `_fetchPreviews` without modifying `this._lastIndex` value

hcvcastro: It is a good idea, thanks

mcecchetti (approved):
It's fine for me. Thanks Henry

Issue #10337, opened 2024-10-27
Revise WS Ping Timeout Handling to become more robust.


sgothel:
### Summary

WS Ping timeout was described and included in the PR #9916 (Issue #9833), commit 994f8f9b2610caa3846fbbac6c4cf612901de0e9.
(Impact is discussed lately in PR #10334).

[WS native control-frame Ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) is considered the only  reliable alive-checking instrument we have for long-lived WS connections.
This above COOL's protocol-ping and ICMP ping .

### Previous Code
The original code pre commit 994f8f9b2610caa3846fbbac6c4cf612901de0e9 had a fixed one-shot period and timeout (TO) of 18s w/o closing the connection at `checkTimeout`.

### Introduced Disconnect on WS Ping Timeout
Commit 994f8f9b2610caa3846fbbac6c4cf612901de0e9 
uses a period of 3s and an average (avg) timeout (TO) of 2s.

The avg TO is not triggered in case of an one-off,
but over the _cumulative_ average of the total connection duration (-> `TimeAverage`).
It renders the latency tolerance much higher in case of drop-outs,
the longer the connection is well alive. E.g. a sporadic 120s lag after
a 120s good responsive connection duration of 250ms would
result in the average of 1.25s and remains alive.
Only after multiple such extreme failures/lags occur, we would drop the connection.
Hence current 2s on average becomes much more relaxed than original fixed 20s.

### Other Keepalive Timeout Related Instruments
[World Ping Test - global ping test](https://www.meter.net/tools/world-ping-test/),
for ICMP ping, demonstrates latency of < 500ms (Sydney w/ ~333ms) at least for me.

Most other utilities I have re-checked now seem to use higher fixed values
ranging between 10s-120min for the timeout - leaning more towards the low-end.
- 5s: [node.js default server.keepAliveTimeout](https://betterstack.com/community/guides/scaling-nodejs/nodejs-timeouts/)
- 18s: Original code had a period==timeout
- 20s: Python WebSockets https://websockets.readthedocs.io/en/latest/topics/keepalive.html
- 20s: grpc https://github.com/grpc/grpc/blob/master/doc/keepalive.md

KeepAlive Intervals (also considered for timeout)
- 30s: [.Net](https://learn.microsoft.com/en-us/dotnet/api/system.net.websockets.websocket.defaultkeepaliveinterval?view=net-8.0)
[...]

PR #10355, opened 2024-10-28
Revise Handle Limited Open Connections

sgothel:
* Target version: master 

### Summary
This track follows up on PR #10334 and may include PR for issue #10337

#### Functional
- Fixes all review comments in: https://github.com/CollaboraOnline/online/pull/10334 particularly
- Removes WS slow ping connection close feature
  - The added implementation as well as the robust alternative #10337 has been rejected
- Use system maximum TCP connections to limit new connections on `SocketPoll`
  - Drop relation to maximum sessions
  - Use Linux system values min(tcp_max_orphans, net.netfilter.nf_conntrack_max)
    - [tcp_max_orphans](https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html)
    - [net.netfilter.nf_conntrack_max](https://www.kernel.org/doc/html/latest/networking/nf_conntrack-sysctl.html)

#### Style
- Removes immutable member variables on frequently created objects (copies of global settings)
  - In particular `net::Defaults` related
- Fixes (some) naming conventions breakage
- Removes redundant boolean `SocketPoll::_limitedConnections`, use query method instead
- Removes `SocketPoll`'s `itemsErased` accounting while erasing
- Removes redundant `Socket::isClosed()`

### Checklist


common/FileUtil.hpp
```diff
@@ -183,6 +184,39 @@
...
+        if (dataLen <= 1) // no space for data + EOS
+            return defaultValue;
+        errno = 0;  // Flush previous error indicator. Reminder: errno is thread-local
+        const int fd = ::open(path.c_str(), O_RDONLY);
+        if (fd < 0)
+            return defaultValue;
+
+        ::memset(data, 0, dataLen);
+        ssize_t consumed = 0;
+        while ( dataLen-consumed-1 /* ex-EOS */ > 0 )
```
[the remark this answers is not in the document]
sgothel: invalid due to pre-validated dataLen.

sgothel:
All checks have passed .. will rebase to master now .. (no merge yet, duh!)

net/Socket.cpp
```diff
@@ -726,28 +726,14 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
         if (itemsErased)
         {
-            const size_t itemsErasedPre = itemsErased;
-            itemsErased = 0; // correcting itemsErased
+            LOG_TRC("Scanning to removing " << itemsErased << " defunct sockets from "
+                    << _pollSockets.size() << " sockets");
```
mmeeks, line 728: I think what I am looking for here is a revert of the complexity inherent in:

80246f7ac1140257d14162504d3550909d6ba681

Reverting that patch as step #1 in its entirety - and then adding whatever it is we wanted to do back again would make me significantly happier. Lets get the revert in - and then re-think that from scratch - I am suspicious of all the changes to the arithemetic there - the itemsAdded size comparisons etc. etc. if there is a big-fix or benefit from each change to the logic there - it should be split out, reviewed and ideally unit tested separately. We should bin the lock, atomic, etc. etc. there too.

sgothel:
A weirdo merge conflict  occurred via github here, which wasn't existing locally (rebased to master). 
"Fun" ..

mmeeks:
So - there is once again a huge amount of change here - a ton of it is not related to reverting back to the state before we started making huge amounts  of change. Again - I want to restore us to a known-safe state tested on millions of users - and then we take very, very small steps, incrementally, without huge and questionable cosmetic changes on top to improve the code.

Perhaps it is nice to do some stream logging operator improvement - perhaps that makes people feel good - but to do that mixed with functional change, and/or to do it at all now - when the priority is getting us back to a known good state is highly unhelpful.

Perhaps it would be better to have a more simple approach. Revert everything that went in recently - and then go over it again - breaking it down properly, thinking through the design carefully, and so on. At this point I'm basically not interested in any new features - connection limits, trottling, closing websockets, etc. - I want us back where we were.

net/Socket.cpp
sgothel: it was just logging I added. Not required.

sgothel:
rebased again, mmeeks branch (w/ same commits) got merged, hence earlier confusion (collision).

sgothel:
Will be revised to fully revert 2 commits which have been partially reverted and reapply changes on top of vanilla.

PR #10364, opened 2024-10-29
Presenter Console: Add background and text color

pedropintosilva:
Both for light and dark mode. In here I opted to use a dark tint on both and later on - depending on the user feedback - we can switch so something lighter for the light mode

Change-Id: I31eed3b0bac327ac3db365f6fdf79405537058bb

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


codewithvk (commented):
Looks nicer to me :) Thanks!

PR #10367, opened 2024-10-29
Kit versions: log version on start, and also check for known badness.

mmeeks:
The combination of a year+ old core, and a modern coolforkit can initialize and startup, but then fails subsequently in hard to diagnose ways, so look for badness and log it early.

Change-Id: I9b9c94997afcc8faa8174b7339483f832ec87dd6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3517,6 +3517,23 @@ void lokit_main(
 
             SigUtil::setVersionInfo(versionString);
 
+            LOG_INF("Kit core version is " << versionString);
+
+            // Extend the list on new releases
+            static const char *denyVersions[] = {
+                "\"22.05\"", "\"23.05\""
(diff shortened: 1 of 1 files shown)


caolanm (approved):
Fair enough then.

The https://github.com/CollaboraOnline/online/commit/b6aeb5a0a1baad9268f20be03c4fe95760507b2f looks like it would take care of the problem at the package level (though looks like it will need constant manual nvr bumping)

mmeeks:
Sure - of course the packaging thing will mostly help - but - even so we can belt+braces this =)

timar:
It just terminates the kit, coolwsd keeps running after this. But it's true that a FTL error in the log will stand out.

PR #10372, opened 2024-10-30
Private/pedro/presenter console styling temp

pedropintosilva:
Draft... attempting to fix the layout but I'm still getting my head around on how / when do we change the width of the next slide img preview


pedropintosilva:
This already got merge in other PRs

PR #9914, opened 2024-08-27
revert scrollwindow patches breaking function wizard (co-24.04)

eszkadev:
This reverts subset of: https://github.com/CollaboraOnline/online/pull/8011
Impact:
- calc function wizard
- insert symbol
- file properties - custom properties

Related bugs on master:
https://github.com/CollaboraOnline/online/issues/9614
https://github.com/CollaboraOnline/online/issues/9559
https://github.com/CollaboraOnline/online/issues/9770

It seems it fixes all of them, but breaks "polished look" of custom properties. Which is less damage.
cc: @meven @thebearon @timar

PR #8011, opened 2024-01-11
Add Css rules for custom properties fields and adaptative scroll height handling

meven:
Change-Id: I4b57a1c04dc67620f4ee010e17fff4f9e677d0de

### Summary

Improve appearance of the fields  of the custom properties tabs in Document properties.

Requires CORE patch: https://gerrit.libreoffice.org/c/core/+/161892

before

after

### Checklist

- [x] Code is properly formatted
- [x] All commits have Change-Id

PR #10373, opened 2024-10-30
CustomProperties dialog: set a default height

meven:
Change-Id: I83daefea86fbfc86c996c2d015ff1cef5c245900

* Resolves: # 
* Target version: master 

### Summary

Set an appropriate initial size for the CustomProperties scrolledWindow.

This is to improve the appearance of this dialog that degraded after https://github.com/CollaboraOnline/online/pull/9914

Before:
image

After:
image

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -2231,8 +2231,9 @@ kbd,
 	margin: 0;
 }
 
-#DocumentPropertiesDialog #customprops .jsdialog.ui-scrollwindow {
-	max-height: min-content;
+#DocumentPropertiesDialog #customprops #CustomInfoPage > .jsdialog.ui-scrollwindow:not(.formulabar) {
+	height: 540px;
+	max-height: none;
 }
 
 #ServerAuditDialog #ServerAuditDialog-mainbox {


meven:
> I don't see anything obviously problematic with this, and no pushback from @pedropintosilva in a week, so OK.
> 
> I wonder if the root of the problem is that the JSDialog vcl backend doesn't give us enough info on what widget should be expanded to the max available space. Reading the UI files, sfx2/uiconfig/ui/custominfopage.ui is the tabpage and that does set fill to True in properties, so why don't we already try to maximize the size of this widget?

Because contrary to other users of  scrollWindow, this one can start empty without any content, consequently it will have initially a semi-random height.

> 
> But anyway, I can accept this as a short-term workaround till the root of the problem is fixed, if it's correct that there is a deeper problem here. CC @eszkadev

I had a tentative to have height being dynamically computed https://github.com/CollaboraOnline/online/pull/8011 but this wasn't good enough and broke other users of this code, then reverted in https://github.com/CollaboraOnline/online/pull/9914

So here I am trying to make a small change to mitigate the issue instead.

vmiklos:
> So here I am trying to make a small change to mitigate the issue instead.

Makes sense, then indeed this minimal fix is the best. :-)

PR #10374, opened 2024-10-30
calc: fix: autocomplete formula edge case 

Rash419:
- Fixes the following case:
1. Create 3 Named Ranges: testrange1, testrange111, testrange11111
2. Try to autocomplete `=SUM(test`
3. Autocomplete `test` with `testrange1`
4. Now popup will show up because there are `testrange111`, `testrange1111` which can be autocompleted if you select one of those it will autocomplete with `testrange1range111` which is not correct

Change-Id: I81cf0f8f3ebf23e275561f16301ac775e7d00a9f


gokaysatir (approved):
Thanks for the fix :)

PR #10366, opened 2024-10-29
net::Defaults: Revert and re-apply revised code

sgothel:
* Resolves: #9833  refining PR #9916, partially replacing PR #10355
* Target version: master 

### Summary
net::Defaults was a singleton, becoming a global, keeping untouched constexpr in SocketPoll.

- Reverts commit 4499e42b762ef3f96fc20a377cf7d663b48305d2
- SocketPoll::DefaultPollTimeoutMicroS is maintained original
- net::Defaults is a global instance w/ default values
  - inactivityTimeout 3600s
    - instead of using DefaultPollTimeoutMicroS for {idle->inactivity) checkRemoval
  - wsPingAvgTimeout 12s
  - wsPingInterval 18s
  - maxTCPConnections 200000 (arbitrary value)
      - for limiting connections in SocketPoll::poll,
        instead of using the session count

### Checklist

- [ ] I have run `make prettier-write` and formatted the code.
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #10375, opened 2024-10-30
Use WSD-config or system's maximum concurrent TCP connections for net::Defaults.maxTCPConnections

sgothel:
* Resolves: #9833  refining PR #9916 and subsuming PR #10366 and PR #10370
* Target version: master 

### Summary
net::Defaults.maxTCPConnections: Use WSD-config or system's maximum concurrent TCP connections
    
`net::Defaults.maxExtConnections` is set:
- use COOLWSD config `net.max_ext_connections` if >= `COOLWSD::MinConnectedSessions`
- or is disabled (zeroed) if config < 0.
- otherwise set to system-value, see `Util::getMaxConcurrentTCPConnections`.

`Util::getMaxConcurrentTCPConnections` uses Linux kernel values
- /proc/sys/net/ipv4/tcp_max_orphans
  See https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html
- /proc/sys/net/nf_conntrack_max
  See https://www.kernel.org/doc/html/latest/networking/nf_conntrack-sysctl.html
- or returns zero if undefined

The Linux kernel values are memory bound, approximately 4 concurrent TCP connections per MB system memory are provided,
e.g. {4096M -> 16384}, {16384M -> 65536}, {65407M -> 262144}, ...

`COOLWSD::MinConnectedSessions` is a static constexpr with value 3, symbolized for clarity.

### Checklist

### DIFF
--- a/common/Util-mobile.cpp
+++ b/common/Util-mobile.cpp
@@ -21,7 +21,8 @@ int spawnProcess(const std::string& cmd, const StringVector& args) { return 0; }
 
 std::string getHumanizedBytes(unsigned long nBytes) { return std::string(); }
 size_t getTotalSystemMemoryKb() { return 0; }
-std::size_t getFromFile(const char* path) { return 0; }
+size_t getMaxConcurrentTCPConnections() { return 0; }
+std::size_t getFromFile(const char* /*path*/, const size_t /*defaultValue*/) { return 0; }
 std::size_t getCGroupMemLimit() { return 0; }
 std::size_t getCGroupMemSoftLimit() { return 0; }
 size_t getMemoryUsagePSS(const pid_t pid) { return 0; }
--- a/common/Util-server.cpp
+++ b/common/Util-server.cpp
@@ -9,6 +9,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 #include <config.h>
+#include <limits>
 
 #include "Util.hpp"
 
@@ -158,6 +159,38 @@ std::size_t getTotalSystemMemoryKb()
     }
 
     return totalMemKb;
+}
+
+std::size_t getFromFile(const char* path, const size_t defaultValue)
+{
+    FILE* file = fopen(path, "r");
+    if (!file)
+        return defaultValue;
+    char line[4096] = { 0 };
+    // coverity[tainted_data_argument : FALSE] - we trust the kernel-provided data
+    const char *line2 = fgets(line, sizeof(line), file);
+    fclose(file);
+    if(!line2)
+        return defaultValue;
+    const uint64_t num = u64FromString(line, /*def=*/defaultValue).first;
+    return num > std::numeric_limits<size_t>::max() ? defaultValue : size_t(num);
+}
+
+size_t getMaxConcurrentTCPConnections()
+{
+#ifdef __linux__
+    // - 4 tcp_max_orphans sockets per MB
+    // - {4096M -> 16384}, {16384M -> 65536}, {65407M -> 262144}, ...
+    const ssize_t tcp_max_orphans = getFromFile("/proc/sys/net/ipv4/tcp_max_orphans", 0); // ignored if n/a
+    const ssize_t nf_conntrack_max = getFromFile("/proc/sys/net/nf_conntrack_max", 0); // ignored if n/a
+    const size_t res = nf_conntrack_max > 0 ? std::min(tcp_max_orphans, nf_conntrack_max) : tcp_max_orphans;
+    LOG_DBG("MaxConcurrentTCPConnections: min(orphans " << tcp_max_orphans
+            << ", conntrack " << nf_conntrack_max << ") -> "
+            << res);
+    return res;
+#else
+    return 0;
+#endif
 }
 
 std::size_t getFromCGroup(const std::string& group, const std::string& key)
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -72,7 +72,7 @@ std::unique_ptr<Watchdog> SocketPoll::PollWatchdog;
 std::atomic<size_t> StreamSocket::ExternalConnectionCount = 0;
 
 net::DefaultValues net::Defaults = { .inactivityTimeout = std::chrono::seconds(3600),
-                                     .maxExtConnections = 200000 /* arbitrary value to be resolved */ };
+                                     .maxExtConnections = 0 /* disabled by default */};
 
 #define SOCKET_ABSTRACT_UNIX_NAME "0coolwsd-"
 
--- a/test/UnitTimeoutNone.cpp
+++ b/test/UnitTimeoutNone.cpp
@@ -31,9 +31,12 @@ static constexpr size_t ConnectionCount = 9;
 /// Base test suite class for connection limit (no limits) using HTTP and WS sessions.
 class UnitTimeoutNone : public UnitTimeoutBase1
 {
-    void configure(Poco::Util::LayeredConfiguration& /* config */) override
+    void configure(Poco::Util::LayeredConfiguration& config) override
     {
         // Keep original values -> No timeout
+
+        // Disable `net::Defaults.maxExtConnections` instrument, i.e. setting it to zero (skipping system default)
+        config.setInt64("net.max_ext_connections", -1);
     }
 
 public:
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -11,6 +11,8 @@
 
 #include <config.h>
 #include <config_version.h>
+#include <cstdint>
+#include <limits>
 
 #include "COOLWSD.hpp"
 
@@ -1208,6 +1210,8 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
         { "net.proxy_prefix", "false" },
         { "net.content_security_policy", "" },
         { "net.frame_ancestors", "" },
+        // Maximum number of concurrent external TCP connections. See `net::Defaults.maxExtConnections`.
+        { "net.max_ext_connections", "0" },
         { "num_prespawn_children", NUM_PRESPAWN_CHILDREN },
         { "per_document.always_save_on_exit", "false" },
         { "per_document.autosave_duration_secs", "300" },
@@ -1532,6 +1536,27 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
 
     // Allow UT to manipulate before using configuration values.
     UnitWSD::get().configure(conf);
+
+    {
+        // Determine `net::Defaults.maxExtConnections` (maximum external concurrent TCP connections)
+        const size_t maxSessions = std::max<size_t>(COOLWSD::MinConnectedSessions, MAX_CONNECTIONS); // min. ext session count
+        const int64_t confMaxConcurrentTCPConnections = ConfigUtil::getConfigValue<int64_t>(conf, "net.max_ext_connections", 0);
+        const size_t sysMaxConcurrentTCPConnections = Util::getMaxConcurrentTCPConnections();
+        size_t maxConcurrentTCPConnections;
+        if (confMaxConcurrentTCPConnections >= COOLWSD::MinConnectedSessions)
+            maxConcurrentTCPConnections = static_cast<size_t>(confMaxConcurrentTCPConnections); // use config value
+        else if (confMaxConcurrentTCPConnections < 0)
+            maxConcurrentTCPConnections = 0; // ignores instrument
+        else if (sysMaxConcurrentTCPConnections >= maxSessions)
+            maxConcurrentTCPConnections = sysMaxConcurrentTCPConnections; // use system value
+        else
+            maxConcurrentTCPConnections = 0; // ignores instrument
+        net::Defaults.maxExtConnections = maxConcurrentTCPConnections;
+        LOG_DBG("net::Defaults.maxExtConnections: maxSessions " << maxSessions
+                << ", maxExtConnections[conf " << confMaxConcurrentTCPConnections
+                << ", sys " << sysMaxConcurrentTCPConnections << "] -> "
+                << net::Defaults.maxExtConnections);
+    }
 
     // Trace Event Logging.
     EnableTraceEventLogging = ConfigUtil::getConfigValue<bool>(conf, "trace_event[@enable]", false);
@@ -2049,10 +2074,10 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
         }
     }
 
-    if (COOLWSD::MaxConnections < 3)
+    if (COOLWSD::MaxConnections < COOLWSD::MinConnectedSessions)
     {
-        LOG_ERR("MAX_CONNECTIONS must be at least 3");
-        COOLWSD::MaxConnections = 3;
+        LOG_ERR("MAX_CONNECTIONS must be at least " << COOLWSD::MinConnectedSessions);
+        COOLWSD::MaxConnections = COOLWSD::MinConnectedSessions;
     }
 
     if (COOLWSD::MaxDocuments > COOLWSD::MaxConnections)
--- a/common/ConfigUtil.hpp
+++ b/common/ConfigUtil.hpp
@@ -171,6 +171,7 @@ public:
 
     void operator()(int& value) { value = _config.getInt(_name); }
     void operator()(unsigned int& value) { value = _config.getUInt(_name); }
+    void operator()(int64_t& value) { value = _config.getInt64(_name); }
     void operator()(uint64_t& value) { value = _config.getUInt64(_name); }
     void operator()(bool& value) { value = _config.getBool(_name); }
     void operator()(std::string& value) { value = _config.getString(_name); }
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -333,8 +333,12 @@ namespace Util
     /// Returns the total physical memory (in kB) available in the system
     size_t getTotalSystemMemoryKb();
 
+    /// Returns the maximum number of concurrent TCP connections, zero if undefined.
+    /// Value is memory bound. On GNU/Linux approximately 4 concurrent TCP connections per MB system memory are provided.
+    size_t getMaxConcurrentTCPConnections();
+
     /// Returns the numerical content of a file at @path
-    std::size_t getFromFile(const char *path);
+    std::size_t getFromFile(const char *path, const size_t defaultValue=0);
 
     /// Returns the cgroup's memory limit, or 0 if not available in bytes
     std::size_t getCGroupMemLimit();
--- a/net/NetUtil.hpp
+++ b/net/NetUtil.hpp
@@ -36,6 +36,9 @@ public:
     std::chrono::microseconds inactivityTimeout;
 
     /// Maximum number of concurrent external TCP connections. Zero disables instrument.
+    /// Set to COOLWSD config `net.max_ext_connections` if >= `COOLWSD::MinConnectedSessions`
+    /// or is disabled (zeroed) if config < 0.
+    /// Otherwise set to system-value, see `Util::getMaxConcurrentTCPConnections`.
     size_t maxExtConnections;
 };
 extern DefaultValues Defaults;
--- a/test/UnitTimeoutConnections.cpp
+++ b/test/UnitTimeoutConnections.cpp
@@ -11,6 +11,7 @@
 
 #include <config.h>
 
+#include <cstdint>
 #include <string>
 
 #include <HttpRequest.hpp>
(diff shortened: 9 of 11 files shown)


sgothel:
desktop cypress calc/searchbar_spec.js post rebase on a2f3536c1946362ebb030653f633451e64c3f52d, while PR #10370 passed. Hence it got better .. but not fully resolved yet.

caolanm:
I feel readDecimal and dataToDecimal are a bit ornate if we compare to say getTotalSystemMemoryKb or getFromCGroup which just open their /proc file, read a line from it and atoll on that

sgothel:
> I feel readDecimal and dataToDecimal are a bit ornate if we compare to say getTotalSystemMemoryKb or getFromCGroup which just open their /proc file, read a line from it and atoll on that

looked at them ofc and thought about adding this func .. esp due to late reviews :)
problem, the used atoll is not covering all error cases and I wanted a one lines to fetch the value
to have the getTechValue func clean of these things - encapsulation.
If its a blocker, I will revise.

sgothel:
added a tiny fix on calling readDecimal (EOS included), refined its API doc .. kicking CI again (ole cypress failure)

mmeeks (changes requested):
Some good things here that can go in - but a number that need feedback. I -really- don't want review getting lost and ignored in monster review threads of giant collections of commits. It is good to split them up - lets also push them in small batches to different PRs.

checkRemoval -> fine.

Simplify UnitTimeoutBase -> no idea - a huge change with no clear rationale.

Add 'UT' -> the 'UT' is not needed - lets not make up lots of new acronyms in commit messages it is implicit in UnitTestFoo - otherwise good to see a new test -> fine.

is "Remove WebSocketHandler's WS Ping timeout functionality" a nearly straight revert - if so good; lets include the word 'Revert' in the commit message - otherwise fine.

Like the Remove redundant 'Socket::isClosed()' - but I would prefer us to remove 'isOpen' to minimize the end-to-end diff across the code. We don't want non-functional renames of methods isClosed -> isOpen is not a win.

"Add safe ..." don't like creating big, redundant helpers like this. We already have similar code, using other helpers that does this - git grep for & re-use the same code there.

why tcp_max_orphans ?

Thanks!

sgothel:
> Simplify UnitTimeoutBase -> no idea - a huge change with no clear rationale.

Not huge IMHO, but simplifying the unit test more to its point, which pattern is also more reusable for other users.

> 
> Add 'UT' -> the 'UT' is not needed - lets not make up lots of new acronyms in commit messages it is implicit in UnitTestFoo - otherwise good to see a new test -> fine.

Fun fact, I read this abbreviation somewhere else .. and picked it up. OK.

> Like the Remove redundant 'Socket::isClosed()' - but I would prefer us to remove 'isOpen' to minimize the end-to-end diff across the code. We don't want non-functional renames of methods isClosed -> isOpen is not a win.

I counted usage of both, and isOpen was used more often without inversion
while isClosed required inversion - so this is more straight, a win.

> 
> "Add safe ..." don't like creating big, redundant helpers like this. We already have similar code, using other helpers that does this - git grep for & re-use the same code there.

`std::stroll` used here covers all error cases, while existing code does not (using atoll).
Initial motivation to factor it out of the actual evaluation function was to not 
have it littered with ordinary and distracting I/O and conversion details.

(other items were either accepted or discussed in upcoming PRs)

mmeeks (changes requested):
When feedback is incorporated systematically please poke me to review again. Thanks.

sgothel:
> When feedback is incorporated systematically please poke me to review again. Thanks.

Done.

Remark: The requested removal of Socket::isOpen() instead of Socket::isClose() increased the diff as laid out earlier. Regardless, I hope it fits requirements as I also changed the field name to have the accessor match 1:1 w/o inverted logic.

Other changes applied as requested as well.

common/Util-desktop.cpp
```diff
@@ -160,6 +161,56 @@
...
+{
+#ifdef __linux__
+    auto readDecimal = [](const char* path, const size_t defaultValue) -> size_t {
+        const int fd = ::open(path, O_RDONLY);
+        if (fd<0)
+            return defaultValue;
+        char data[1024+1]; // includes EOS
+        ::memset(data, 0, sizeof(data));
+        size_t consumed = 0;
+        while (sizeof(data)-consumed-1 /* ex-EOS */ > 0)
```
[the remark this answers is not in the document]
sgothel: negative value never intended, but `>0` or `!=0` is correct here.

sgothel:
Cleaned up branch rebased tree, i.e. PR #10375 -> #10455 -> #10456

(also refined `getMaxConcurrentTCPConnections` less analytics scarring comparison and borderline `unsigned long long > numeric_limits<size_t>::max()` check)

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
     return totalMemKb;
 }
 
+size_t getMaxConcurrentTCPConnections(const size_t minimum)
+{
+#ifdef __linux__
+    auto readDecimal = [](const char* path, const size_t defaultValue) -> size_t {
+        const int fd = ::open(path, O_RDONLY);
+        if (fd<0)
```
mmeeks, line 170: inconsistent whitespace in your own code here; "if (fd < 0)" surely ? =)

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
     return totalMemKb;
 }
 
+size_t getMaxConcurrentTCPConnections(const size_t minimum)
+{
+#ifdef __linux__
+    auto readDecimal = [](const char* path, const size_t defaultValue) -> size_t {
+        const int fd = ::open(path, O_RDONLY);
```
mmeeks, line 169: This is really grim; please read the other functions in that file - and try to be inspired by them. We have a mix of fopen and fgets - and we generally assume that weirdness will not occur there. Writing and reviewing yet another loop like this not necessary for /proc files I believe.

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
     return totalMemKb;
 }
 
+size_t getMaxConcurrentTCPConnections(const size_t minimum)
+{
+#ifdef __linux__
```
mmeeks, line 167: We assume Linux if not mobile everywhere currently - as you can see all around in this module; lets stay consistent.

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
...
+    auto readDecimal = [](const char* path, const size_t defaultValue) -> size_t {
```
mmeeks, line 168: Sigh - so - we can have a helper function for this - but lets name it something that is consistent with other Util helpers, and use similar code to implement it.

common/Util-mobile.cpp
```diff
@@ -21,6 +21,7 @@ int spawnProcess(const std::string& cmd, const StringVector& args) { return 0; }
 
 std::string getHumanizedBytes(unsigned long nBytes) { return std::string(); }
 size_t getTotalSystemMemoryKb() { return 0; }
+size_t getMaxConcurrentTCPConnections(const size_t /*enabled_minimum*/) { return 0; }
```
mmeeks, line 24: Naming is reasonable.

net/Socket.cpp
```diff
@@ -654,7 +654,7 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                     rc = -1;
                 }
 
-                if (!_pollSockets[i]->isOpen() || !disposition.isContinue())
+                if (_pollSockets[i]->isClosed() || !disposition.isContinue())
```
mmeeks, line 657: Closed stuff is fine - thanks.

wsd/COOLWSD.cpp
```diff
@@ -1514,6 +1514,8 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
     }
     UnitWSD::get().setWSD(this);
 
+    net::Defaults.maxExtConnections = Util::getMaxConcurrentTCPConnections(std::max<size_t>(3, MAX_CONNECTIONS));
```
mmeeks, line 1517: 3 ?

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
...
+        const unsigned long long num = std::strtoull(data, &endptr, 10);
+        if (0 != errno || nullptr == endptr || endptr == data || num > std::numeric_limits<size_t>::max())
+            return defaultValue;
+        return size_t(num);
+    };
+    // - 4 tcp_max_orphans sockets per MB
+    // - {4096M -> 16384}, {16384M -> 65536}, {65407M -> 262144}, ...
+    const ssize_t tcp_max_orphans = readDecimal("/proc/sys/net/ipv4/tcp_max_orphans", 0); // ignored if n/a
+    const ssize_t nf_conntrack_max = readDecimal("/proc/sys/net/nf_conntrack_max", 0); // ignored if n/a
+    const size_t res1 = nf_conntrack_max > 0 ? std::min(tcp_max_orphans, nf_conntrack_max) : tcp_max_orphans;
```
mmeeks, line 202: res1 and res2 ? reminds me of the very un-loved c1, c2, c3 etc.

mmeeks (approved):
Lets get the first two commits in- but not the third - which needs re-working.

common/Util-desktop.cpp
sgothel: Will use FileUtil::readFile then, closes thing we have avoiding fopen/fgets.

common/Util-desktop.cpp
sgothel: Thought about FreeBSD etc and we do have `#ifdef __linux__` within same compilation unit, so this is not clear.

common/Util-desktop.cpp
sgothel: This has been degraded to a local hidden lambda and we also do have `Util::dataToHexString` and `FileUtil::readFile` .. consistent verb `read` and the object to read `decimal`. 
Regardless, using `FileUtil::readFile` for the file ops as mentioned above, OK.

wsd/COOLWSD.cpp
sgothel: the used magic in same compilation unit for the very minimum value of this session limit threshold - but it gets validated at a later stage when used.

common/Util-desktop.cpp
sgothel: res1 -> res_raw (res = result)
res2 -> res_clipped
OK

common/Util-desktop.cpp
```diff
@@ -160,6 +162,56 @@ std::size_t getTotalSystemMemoryKb()
...
+        char* endptr = nullptr;
+        errno = 0;  // Flush previous error indicator. Reminder: errno is thread-local
+        const unsigned long long num = std::strtoull(data, &endptr, 10);
+        if (0 != errno || nullptr == endptr || endptr == data || num > std::numeric_limits<size_t>::max())
+            return defaultValue;
+        return size_t(num);
+    };
+    // - 4 tcp_max_orphans sockets per MB
+    // - {4096M -> 16384}, {16384M -> 65536}, {65407M -> 262144}, ...
+    const ssize_t tcp_max_orphans = readDecimal("/proc/sys/net/ipv4/tcp_max_orphans", 0); // ignored if n/a
```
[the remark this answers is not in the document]
sgothel: [tcp_max_orphans](https://www.man7.org/linux/man-pages/man7/tcp.7.html) is that memory bound freely available socket count, not getting reset. 
> This limit exists only to prevent simple denial-of-service attacks.  Lowering this limit is not recommended. Network conditions might require you to increase the number of orphans allowed, but note that each orphan can eat up to ~64 kB of unswappable memory.  The default initial value is set equal to the kernel parameter NR_FILE.  This initial default is adjusted depending on the memory in the system.

While more sockets can be used, it is a good indicator of a worry free socket count IMHO _at_ application launch.
`nf_conntrack_max` is usually set to `tcp_max_orphans` by default, if the iptables conntrack module is loaded.

Here emphasis should be on _concurrent_tcp_sockets_ and whether we like to use this metric to limit external connections in the first place, I understand this.

It may make sense to also inject an alternative `max_tcp_connections` COOLWSD configuration value.
However cropping such value to `nf_conntrack_max` makes little sense, i.e. it implies the admin has to adjust the `nf_conntrack_max` _and_ (probably) `tcp_max_orphans` limits anyways.
Hence we could use a config COOLWSD  `max_tcp_connections`, use it as-is (if given) but warn if the system `getMaxConcurrentTCPConnections` is lower (-> for the system admin)?

Now if we look at the memory based Linux defaults here, i.e. 4 sockets/MB or 262k sockets for 64G etc, 
it looks to me that these are good values. Whenever this issue is discussed in the web server space, 
it is said to increase the same system limits to achieve a higher concurrent responsiveness while 
(need to) maintain the DoS balance - if that makes sense.

common/Util-desktop.cpp
sgothel: OK, I remember, FileUtil::read is not available to all link targets. So it is fgets/fopen then.

common/Util-desktop.cpp
mmeeks: fair enough.

common/Util-desktop.cpp
```diff
@@ -160,6 +162,43 @@ std::size_t getTotalSystemMemoryKb()
...
+        char line[4096] = { 0 };
+        // coverity[tainted_data_argument : FALSE] - we trust the kernel-provided data
+        const char *line2 = fgets(line, sizeof(line), file);
+        fclose(file);
+        if(line2 == nullptr)
+            return defaultValue;
+        char* endptr = nullptr;
+        errno = 0;  // Flush previous error indicator. Reminder: errno is thread-local
+        const unsigned long long num = std::strtoull(line, &endptr, 10);
+        if (0 != errno || nullptr == endptr || endptr == line || num > std::numeric_limits<size_t>::max())
```
mmeeks, line 181: 0 != errno to errno  != 0 - as previously discussed. We use !ptr - ~everywhere for null pointer checking & readability - as above - please use that idiom - and above twice.

mmeeks: Either way - I don't believe any of this error detection is helpful; if you read systemd you'll see they use sscanf() to parse /proc output - the format and it content is well defined and must be backwards compatible. Lets use stroll as is done in the functions above; and lets split this out into its own readDecimal function instead of having this in-line function - but lets give it a name that doesn't suck - and fits into the existing naming convention - I believe I asked for that before.

common/Util-desktop.cpp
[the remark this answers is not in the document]
mmeeks: The setting is some random other number - for a different use-case. It may often be initialized to something sensible - but it's really for a different purpose AFAICS.

So - fine; lets do this guess-timate thing - but we need to provide a setting to override it, in the case that someone really wants to configure their orphans in some way - and not crop the number of clients connections we can support.

common/Util-desktop.cpp
```diff
@@ -160,6 +162,43 @@ std::size_t getTotalSystemMemoryKb()
...
+    const ssize_t tcp_max_orphans = readDecimal("/proc/sys/net/ipv4/tcp_max_orphans", 0); // ignored if n/a
+    const ssize_t nf_conntrack_max = readDecimal("/proc/sys/net/nf_conntrack_max", 0); // ignored if n/a
+    const size_t res_min = nf_conntrack_max > 0 ? std::min(tcp_max_orphans, nf_conntrack_max) : tcp_max_orphans;
+    const size_t res_clipped = res_min >= cutOffMinimum ? res_min : 0;
+    LOG_DBG("MaxConcurrentTCPConnections: clip(min(orphans " << tcp_max_orphans
+            << ", conntrack " << nf_conntrack_max << ") -> "
+            << res_min
+            << ", cut-off-min " << cutOffMinimum
+            << ") -> " << res_clipped);
+    return res_clipped;
```
mmeeks, line 196: Seems it is very likely that  this function returns zero - is that expected ?

The cutOffMinimum and the way it is injected here is really non-intuitive. Can you add a one-line comments against each line of your ? operator logic above to explain what is going on. We have two metaphores for the same thing 'cut-off' and 'clipped' - please choose something consistent for the variable naming to give the next poor reader a clue.

mmeeks (changes requested):
Minor changes but lets get them done first.

sgothel:
force push: rebase to master only

common/Util-desktop.cpp
sgothel: Moved logic to COOLWSD, simplified getter.

common/Util-desktop.cpp
sgothel: Reanimating the `Util::getFromFile` you introduced last year .. adding a default value.

common/Util-desktop.cpp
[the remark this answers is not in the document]
sgothel: Added WSD config value evaluation.

Ashod (commented):
Thanks, @sgothel. Left some comments/notes.

test/UnitTimeoutNone.cpp
```diff
@@ -31,9 +31,12 @@ static constexpr size_t ConnectionCount = 9;
 /// Base test suite class for connection limit (no limits) using HTTP and WS sessions.
 class UnitTimeoutNone : public UnitTimeoutBase1
 {
-    void configure(Poco::Util::LayeredConfiguration& /* config */) override
+    void configure(Poco::Util::LayeredConfiguration& config) override
     {
         // Keep original values -> No timeout
+
+        // Disable `net::Defaults.maxExtConnections` instrument
+        config.setInt64("net.max_ext_connections", -1);
```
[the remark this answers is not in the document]
sgothel: Explained in NetUtils.hpp for `net::Defaults.maxExtConnections`
>     /// Maximum number of concurrent external TCP connections. Zero disables instrument.
>+    /// Set to COOLWSD config `net.max_ext_connections` if >= `COOLWSD::MinConnectedSessions`
>+    /// or is disabled (zeroed) if config < 0.
>+    /// Otherwise set to system-value, see `Util::getMaxConcurrentTCPConnections`.
>     size_t maxExtConnections;

So the tri-state: <0 disable, [0-2] ignore for system defaults and >=3 is required to fit all needs IMHO.
Differentiation of config value semantics and `net::Defaults.maxExtConnections`

common/Util-server.cpp
```diff
@@ -160,6 +161,42 @@ std::size_t getTotalSystemMemoryKb()
...
+        return defaultValue;
+    char line[4096] = { 0 };
+    // coverity[tainted_data_argument : FALSE] - we trust the kernel-provided data
+    const char *line2 = fgets(line, sizeof(line), file);
+    fclose(file);
+    if(!line2)
+        return defaultValue;
+    char* endptr = nullptr;
+    errno = 0;  // Flush previous error indicator. Reminder: errno is thread-local
+    const unsigned long long num = std::strtoull(line, &endptr, 10);
```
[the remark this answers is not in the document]
sgothel: I can try .. did it with another Util earlier, but linkage was not given for one compilation targets.

test/UnitTimeoutNone.cpp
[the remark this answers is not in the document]
Ashod: The comment on line 38 of test/UnitTimeoutNone.cpp says "Disable" and sets the value to -1. 
The comment on line 75 of net/Socket.cpp says "disable" and sets the value to 0.

Perhaps the issue is the comments?

common/Util-server.cpp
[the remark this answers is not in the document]
Ashod: I'm sure we can resolve the linkage issue. `Util` is included/linked everywhere.

I can try to help, if you share the errors you get. Just ask.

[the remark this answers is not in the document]
sgothel: - Usage of `strto*` is (still) spread across our code, true (also its unsafe application, a double down)
- Changed to your `u64FromString` header code, thx (overseen)
- Wanted to use `FileUtil::*`, but having linker issues as mentioned, so dropped and keeping Michael's version reanimated
- In general I also prefer refactoring and commonly used low-level utils (duh!), such changes are good candidates for cleanup tasks indeed.
- Myself not happy using FILE/fopen/fgets due to added buffering, but obeyed and changed this from low-level FD ops.
I will leave it at that for now.

test/UnitTimeoutNone.cpp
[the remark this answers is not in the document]
sgothel: Yes, I also changed the comment in that use-case, since the config value (tri-state) and the stored value (off or number) in net::Defaults do have different semantics. 
I was indeed teared apart by commenting on this in more detail, but then removed it in favor of the single comment in net::Defaults/DefaultValue. 
`        // Disable `net::Defaults.maxExtConnections` instrument, i.e. setting it to zero (skipping system default)`

mmeeks:
@Ashod - can you do some thinking on whether we want this, and get it in if so ? =)

Ashod:
> @Ashod - can you do some thinking on whether we want this, and get it in if so ? =)

Yes, will do.

PR #10377, opened 2024-10-30
mention: insert extra space after inserting mention

Rash419:
- this makes the behaviour consistent with other applications

Change-Id: I680d8de9b48ebb502ec747e1ef3711fb1d69aa5d

* Target version: master


caolanm (approved):
This seems reasonable to me assuming ci passes

PR #10378, opened 2024-10-30
Accessibility : Enhance Dropdown Accessibility with Arrow Key Navigation 

Darshan-upadhyay1110:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -84,6 +104,28 @@ JSDialog.comboboxEntry = function (parentContainer, data, builder) {
 	return false;
 };
 
+ // Helper function to find the next focusable element using JSDialog.IsFocusable
+ function getNextFocusableEntry(data) {
+	let nextElement;
+	let pos = parseInt(data.pos)
+	do {
+		pos++; // Move to the next position
+		nextElement = document.getElementById(data.comboboxId + '-entry-' + pos);
```
eszkadev, line 113: please try to avoid using `getElementById`, it's fragile on future changes.
Can we maybe use `JSDialog.GetFocusableElements` which we will just iterate?

or can we create helper: `JSDialog.GetNextFocusableElement` which will find element by iterating the DOM?

eszkadev: or maybe even we can just use `JSDialog.KeyboardGridNavigation` ?

eszkadev: we shouldn't implement the same ting few times, ideally it should be simple algorithm which uses "siblings" and not any "ids" and just looks for next one focusable on arrow right, previous on arrow left etc...
this way we can use it in any place without modification

Darshan-upadhyay1110: Yes i can try on that but still i think to get Dropdown container i need to user byID one time.
`const comboboxEntries = document.getElementById(data.id +'-entries');`

Darshan-upadhyay1110: > EDIT: it seems something exists already:
JSDialog.FindFocusableElement(
  potentialCurrentElement.previousElementSibling,
  'previous'
  );
  
  This will not work as it also return the element if current is focusable we need to pass potential next element to this function and this will work...

Darshan-upadhyay1110: So indirectly need to get next element and pass to this function. Or better we just create new helper for it ?

Darshan-upadhyay1110: like this: 
image

Darshan-upadhyay1110: > there is helper for that too: `var dropdownRoot = JSDialog.GetDropdown(data.id);`
> 
> we at least need to encapsulate all these "get by id" so it will be single place to modify in the future (if id generation will be changed)

I see okay then it will work

Darshan-upadhyay1110: Let me do the changes and test, Thanks :)

eszkadev: less duplicated code == better :)

I would try:
1.  `JSDialog.KeyboardGridNavigation` (it works left, right, up, down, might be not good)
2. if above doesn't work, please try to create generic: `JSDialog.KeyboardListNavigation` (up,down only)
3. do custom thing for this one

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -237,11 +237,25 @@ JSDialog.combobox = function (parentContainer, data, builder) {
...
 			}
 		}
+		// check for drop down is in open state or not.
+		// If open then we should make focus in entries field for Arrow key navigation
+		if (builder.map.jsdialog.hasDropdownOpened()) {
+			// Select the parent container
+			const comboboxEntries =  JSDialog.GetDropdown(data.id);
+
+			// Find the child element with the 'selected' class
+			const selectedElement = comboboxEntries.querySelector(".selected");
```
eszkadev, line 252: this is implementation detail - best would be to avoid - to make less fragile

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -237,11 +237,25 @@ JSDialog.combobox = function (parentContainer, data, builder) {
 
 		resetSelection();
 		for (var i in entries) {
-			if (entries[i] == this.value) {
+			if (entries[i].text == this.value) {
```
eszkadev, line 240: we need more typescript :)

eszkadev: this is complicated... I think the "JS-generated entries" have text prop (eg. used for dropdown) but the "core-generated entries" come without it (see example JSON in the header and core impl: https://opengrok.libreoffice.org/xref/core/vcl/source/control/combobox.cxx?r=4b53712c#1570
where it has simple value.

we need to handle both cases

`entries[i] == this.value || entries[i].text == this.value"

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -237,11 +237,25 @@ JSDialog.combobox = function (parentContainer, data, builder) {
...
 				entries[i].selected = true;
 				break;
 			}
 		}
+		// check for drop down is in open state or not.
+		// If open then we should make focus in entries field for Arrow key navigation
+		if (builder.map.jsdialog.hasDropdownOpened()) {
```
eszkadev, line 247: I would put `event.key === 'ArrowDown'` as a first condition, so we will not enter this block when not needed without executing unnecessary code

browser/src/control/jsdialog/Util.KeyboardListNavigation.ts
```diff
@@ -0,0 +1,102 @@
...
+		currentElement.classList.add('selected');
+	}
+}
+
+JSDialog.KeyboardListNavigation = function (container: HTMLElement) {
+	container.addEventListener('keydown', (event: KeyboardEvent) => {
+		const activeElement = document.activeElement as HTMLElement;
+		if (
+			!(
+				(activeElement.tagName === 'INPUT' &&
```
eszkadev, line 94: This code is copy-pated from KeyboardGridNavigation -> please share it.
In Focus cycle we can put new helper: `JSDialog.IsTextField(element: HTMLElement)` for example...

browser/src/control/jsdialog/Util.KeyboardListNavigation.ts
```diff
@@ -0,0 +1,102 @@
...
+
+	while (siblingElement) {
+		if (JSDialog.IsFocusable(siblingElement as HTMLElement)) {
+			(siblingElement as HTMLElement).focus();
+			siblingElement.classList.add('selected');
+			focusableFound = true;
+			break;
+		}
+		siblingElement =
+			direction === 'next'
```
eszkadev, line 77: this is repeated here 2 times and also in FocusCycle in findFocusableElement: move to FocusCycle.js and share `JSDialog.FindNextSibling(element: HTMLElement, direction: string)`

browser/src/control/jsdialog/Widget.Combobox.js
```diff
@@ -237,11 +237,25 @@ JSDialog.combobox = function (parentContainer, data, builder) {
...
 		resetSelection();
 		for (var i in entries) {
-			if (entries[i] == this.value) {
+			if (entries[i].text == this.value) {
 				entries[i].selected = true;
 				break;
 			}
 		}
+		// check for drop down is in open state or not.
+		// If open then we should make focus in entries field for Arrow key navigation
```
eszkadev, line 246: why do we need to add here some additional focusing?
maybe we can just `event.preventDefault()` to stop processing any other callback which will changer focus?

browser/src/control/jsdialog/Widget.Combobox.js
eszkadev: but leave it for now

browser/src/control/jsdialog/Util.Dropdown.js
```diff
@@ -151,6 +154,12 @@ JSDialog.OpenDropdown = function (id, popupParent, entries, innerCallback, popup
 	};
 
 	L.Map.THIS.fire('jsdialog', {data: json, callback: generateCallback(entries)});
+	const listContainer = JSDialog.GetDropdown(id);
+	if(useListKeyboardNavigation && listContainer) {
+		// attach list keybinding for arrow-key navigation in entry list
+		JSDialog.KeyboardListNavigation(listContainer);
```
eszkadev, line 160: I think this can be moved to dropdown handler which doesn't exist currently (it does just default job).
That's why we see in the browser console: `JSDialogBuilder: Unsupported control type: "dropdown"` after we open dropdown.

Please try to add new `JSDialog.Dropdown` handler function which will just do the build based on received JSON and this KeyboardListNavigation if needed.

it should be something like:
```
JSDialog.Dropdown = function(parentContainer, data, builder) {
    builder.build(parentContainer, data.children, data.vertical === true);

   if (!data.noDefaultKeyboardNavigation)
         JSDialog.KeyboardListNavigation(parentContainer);
    return false;
}
```
and regiser it in the JSDialogBuilder: `this._controlHandlers['dropdown'] = JSDialog.Dropdown;`
and colorpicker should have in dropdown JSON property: `noDefaultKeyboardNavigation: true`

then we shouldn't need that code here

Darshan-upadhyay1110: > colorpicker should have in dropdown JSON property: `noDefaultKeyboardNavigation: true`

@eszkadev  I think we do not need this `noDefaultKeyboardNavigation` because color picker already has handler: `this._controlHandlers['colorpicker'] = JSDialog.colorPicker;`

Thanks for the Dropdown handler approach i think it's awesome :)

eszkadev (approved):
Thank you, now looks great! :)

browser/src/control/jsdialog/Widget.Combobox.js
eszkadev: I see, we can still be with focus on the dropdown button or combobox entry, but when using arrow key - we jump to selected entry in the list - which is expected

eszkadev:
could you rebase? it seems searchbar test fails, which was updated recently

PR #10379, opened 2024-10-31
wsd: fix: failed to pass aliases using `--use-env-vars` option

Rash419:
- regression from 1b3218df0466a59fd1e9a8df72d97ede6448425e
- Steps to reproduce/test:
  1. create env variable `export aliasgroup1=http://test.integrator,http://test.integrator1,http://test.integrator2`
  2. Run COOL binary with `--use-env-vars`
  3. You can either search for 'Adding trusted WOPIHost' in coolwsd logs and make sure all the aliases and hosts are allowed or you can add multiple entries to your `/etc/hosts` that points to same WOPIHost instance

Change-Id: I16785458724c81b681b8ebb979c119af5db9b808

* Resolves: #7730

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3270,6 +3270,7 @@ void COOLWSD::initializeEnvOptions()
         bool first = true;
         std::istringstream aliasGroupStream;
         aliasGroupStream.str(aliasGroup);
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks ok for me

PR #10384, opened 2024-10-31
calc: fix: don't update the combobox input if user is editing

Rash419:
- it also fixes the random cypress failure when cypress tries to write in addressinputfield

Change-Id: Id2ca8efb0066506c3bac38d71c04f9e4f5690fcc

* Target version: master


eszkadev (approved):
looks ok

Issue #10332, opened 2024-10-25
SIGABRT after upgrading from 24.04.8-2 to 24.04.9-1


Glandos:
### Describe the Bug
I upgraded today to 24.04.9-1:
```
   code-brand (24.04-22 => 24.04-23)
   collaboraoffice (24.04.8-2 => 24.04.9-1)
   collaboraoffice-ure (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-calc (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-core (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-draw (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-en-us (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-extension-pdf-import (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-graphicfilter (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-images (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-impress (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-math (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-ooofonts (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-ooolinguistic (24.04.8-2 => 24.04.9-1)
   collaboraofficebasis-writer (24.04.8-2 => 24.04.9-1)
   coolwsd (24.04.8.2-1 => 24.04.9.1-1)
```
Server starts correctly, but crashes with SIGABRT when opening a file from Nextcloud

### Expected Behavior
No SIGABRT

### Actual Behavior
```
oct. 26 00:06:43 belette64 coolwsd[710924]: wsd-710924-710945 2024-10-26 00:06:43.668957 +0200 [ websrv_poll ] ERR  Skipping the token [--nc-light-color-primary-default=] since it does not have '='| wsd/FileServerUtil.cpp:405
oct. 26 00:06:43 belette64 coolwsd[710924]: wsd-710924-710945 2024-10-26 00:06:43.669085 +0200 [ websrv_poll ] ERR  Skipping the token [--nc-dark-color-primary-default=] since it does not have '='| wsd/FileServerUtil.cpp:405
oct. 26 00:06:43 belette64 coolwsd[710924]: terminate called after throwing an instance of 'Poco::Net::NetException'
[...]

PR #10385, opened 2024-10-31
NetworkInterface::list can throw

caolanm:
Seen in https://github.com/CollaboraOnline/online/issues/10332

where list has:

if (getifaddrs(&ifaces) < 0)
  throw NetException("cannot get network adapter list");

which seems the likely source of the exception.

Change-Id: I889ae3107253acd69d2abc362523dba2eebfceb9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (approved):
Looks fine to me - thanks Caolan =)

PR #10386, opened 2024-10-31
Robust ServerSocket not throw (EOL) on recoverable accept failure

sgothel:
* Resolves: ...
* Target version: master 

### Summary
- ServerSocket shall not throw (EOL) on accept returning nullptr
- Simplify accept, ending application on unrecoverable error

Have `ServerSocket::accept` determine whether to `Util::forcedExit(EX_SOFTWARE)`
on unrecoverable errors:
- Exit application if `::accept` returns an unrecoverable `errno` (see below)
  - tolerate `Socket` ctor exceptions
  - tolerate `::accept` recoverable errors (see below)
- Note, the following are `no throw`
  - `::close`
  - `::inet_ntop`
    
Also reject exceeding external connections right within `ServerSocket::accept`,
avoiding creation of `Socket` objects - hence saving resources.
    
Recoverable `::accept` `errno` values
following [accept(2)](https://www.man7.org/linux/man-pages/man2/accept.2.html) are 
- EAGAIN         // == EWOULDBLOCK
- ENETDOWN       // man accept(2): to be treated like EAGAIN
- EPROTO         // man accept(2): to be treated like EAGAIN
- ENOPROTOOPT    // man accept(2): to be treated like EAGAIN
- EHOSTDOWN      // man accept(2): to be treated like EAGAIN
- ENONET         // man accept(2): to be treated like EAGAIN (undef on BSD, Apple, ..)
- EHOSTUNREACH   // man accept(2): to be treated like EAGAIN
- EOPNOTSUPP     // man accept(2): to be treated like EAGAIN
- ENETUNREACH    // man accept(2): to be treated like EAGAIN
- ECONNABORTED   // OK
- ETIMEDOUT      // OK, should not happen due ppoll

May consider handling temporary EMFILE, ENFILE, ENOBUFS and ENOMEM errors?

### Unit Testing    
Unit testing hooks for failure injection is provided via
- UnitWSD::simulateExternalAcceptError()
- UnitWSD::simulateExternalSocketCtorException()

and utilized in
- test/UnitServerSocketAcceptFailure1.cpp
- test/UnitStreamSocketCtorFailure1.cpp

### Checklist


net/ServerSocket.hpp
```diff
@@ -103,14 +103,11 @@ class ServerSocket : public Socket
...
-            if (!clientSocket)
+            if (clientSocket)
             {
-                const std::string msg = "Failed to accept. (errno: ";
-                throw std::runtime_error(msg + std::strerror(errno) + ')');
-            }
-
-            LOG_TRC("Accepted client #" << clientSocket->getFD());
-            _clientPoller.insertNewSocket(std::move(clientSocket));
+                LOG_TRC("Accepted client #" << clientSocket->getFD() << ", " << *clientSocket);
```
mmeeks, line 108: LOG_TRCA(Socket, ...) - the socket code should just work and not spam our logs.

net/ServerSocket.hpp
```diff
@@ -103,14 +103,11 @@ class ServerSocket : public Socket
...
+                _clientPoller.insertNewSocket(std::move(clientSocket));
+            } // else continue this ServerSocket as accept() already determined whether to throw (ending it).
```
mmeeks, line 110: Comment reads oddly - if there is a clear warning in the accept() method for this case lets ignore adding this here.

net/ServerSocket.hpp
```diff
@@ -121,6 +118,8 @@ class ServerSocket : public Socket
         return _sockFactory->create(fd, type);
     }
 
+    static constexpr const char * ErrorMessagePrefix = "Failed to accept. (errno: ";
```
mmeeks, line 121: Why pull this far away from the message we create ? there can be no significant performance win - but there is a real readability hit from having that far from the code that it is used in. Please remove.

net/Socket.cpp
```diff
@@ -1239,50 +1239,58 @@ std::shared_ptr<Socket> ServerSocket::accept()
...
+        type = Socket::Type::IPv4;
+    }
+    else
+    {
+        struct sockaddr_in6 *ipv6 = &clientInfo;
+        inAddr = &(ipv6->sin6_addr);
+        type = Socket::Type::IPv6;
+    }
+    ::inet_ntop(clientInfo.sin6_family, inAddr, addrstr, sizeof(addrstr));
+#endif
```
mmeeks, line 1207: Lets have the try{ } inside the non-mobile conditionals - rather than adding more conditionals. The FakeSocket is unlikely to have these problems - and the change is not necessary.

net/Socket.cpp
```diff
@@ -1239,50 +1239,58 @@ std::shared_ptr<Socket> ServerSocket::accept()
...
+        inAddr = &(ipv4->sin_addr);
+        type = Socket::Type::IPv4;
+    }
+    else
+    {
+        struct sockaddr_in6 *ipv6 = &clientInfo;
+        inAddr = &(ipv6->sin6_addr);
+        type = Socket::Type::IPv6;
+    }
+    ::inet_ntop(clientInfo.sin6_family, inAddr, addrstr, sizeof(addrstr));
```
mmeeks, line 1196: Why do we re-order this method from its previous location ? I really don't like shuffling code - it previously occured after "createSocketFromAccept"

net/Socket.cpp
```diff
@@ -1239,50 +1239,58 @@ std::shared_ptr<Socket> ServerSocket::accept()
...
     const int rc = fakeSocketAccept4(getFD());
 #endif
-    LOG_TRC("Accepted socket #" << rc << ", creating socket object.");
-    try
+    if (rc < 0)
     {
-        // Create a socket object using the factory.
-        if (rc != -1)
+        const int cause = errno;
+        if( cause != EINTR && cause != EAGAIN && cause != EWOULDBLOCK )
```
mmeeks, line 1168: It seems to me that leaving coolwsd running without an externally listening socket is a bad move here. 
Can we restore the nesting of the previous 'try' - to avoid all this code change / movement.

And instead do a LOG_FTL() and a Util::forcedExit(EX_SOFTWARE); in the case that we get an unrecoverable error in this case.

Then I expect we can minimize the change, the diff, reduce review time & do something more sensible than before.

net/Socket.cpp
```diff
@@ -1329,11 +1337,21 @@ bool Socket::isLocal() const
 std::shared_ptr<Socket> LocalServerSocket::accept()
 {
     const int rc = ::accept4(getFD(), nullptr, nullptr, SOCK_NONBLOCK | SOCK_CLOEXEC);
+    if (rc < 0)
+    {
+        const int cause = errno;
+        if( cause != EINTR && cause != EAGAIN && cause != EWOULDBLOCK )
+        {
+            // Unrecoverable error, throw, which will close `ServerSocket` and discard it from `AcceptPoll` during `SocketPoll::poll`
```
mmeeks, line 1278: Lets do the same here as above.

mmeeks (changes requested):
the connection counting code is now looking good.
Getting it in is blocked by a) CI - and b) now fixing up the ServerSocket change here to make it much smaller and more helpful I think.
Thanks.

net/ServerSocket.hpp
sgothel: `LOGA_TRC(Socket,  .. ` OK

net/ServerSocket.hpp
sgothel: OK

net/ServerSocket.hpp
sgothel: just the idea from earlier code and reusage .. but sure, making it local

net/Socket.cpp
sgothel: because subsequent patch uses the addrstr resolved before Socket create

net/Socket.cpp
sgothel: OK

sgothel:
This is the ServerSocket fix vanilla branch w/o merged max-connection
https://github.com/CollaboraOnline/online/tree/private/sgothel/sockrepair4_p1

This is the ServerSocket fix branch w/ merged max-connection (this reviewed one)
https://github.com/CollaboraOnline/online/tree/private/sgothel/sockrepair4_p2

I applied all review remarks .. but not fully removed the 'swizzle around' lines
for clarity of the resulting code, please see 
https://github.com/CollaboraOnline/online/blob/private/sgothel/sockrepair4_p2/net/Socket.cpp#L1153

mmeeks (changes requested):
I'd like to see unit tests for this - possibly we need a fault injection hook in UnitBase to throw during SSL socket creation or something - and of course for when we run out of sockets too.
Thanks!

sgothel:
> I'd like to see unit tests for this - possibly we need a fault injection hook in UnitBase to throw during SSL socket creation or something - and of course for when we run out of sockets too. Thanks!

Revised pull request including unit tests injection
- recoverable ::accept errors
- non-recoverable ::accept errors
- recoverable (server) StreamSocket ctor exceptions

Classification of recoverable ::accept errors has been revised and documented.

sgothel:
Unit test logs w/ error injections
- [unit-serversock_accept1.log.txt](https://github.com/user-attachments/files/17648954/unit-serversock_accept1.log.txt)
- [unit-streamsock_ctor1.log.txt](https://github.com/user-attachments/files/17648955/unit-streamsock_ctor1.log.txt)

sgothel:
> ENONET doesn't appear to exist on IOS, presumably an easy fix for that build error

thx, checked on [FreeBSD's accept(2)](https://man.freebsd.org/cgi/man.cgi?accept(2)) as well as on their [errno.h](https://man.freebsd.org/cgi/man.cgi?query=errno&sektion=2&format=html) + Apple's. So it was just this one missing on BSD + Apple.

net/Socket.cpp
```diff
@@ -1145,63 +1158,160 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
 #if !MOBILEAPP
     assert(_type != Socket::Type::Unix);
 
     struct sockaddr_in6 clientInfo;
     socklen_t addrlen = sizeof(clientInfo);
-    const int rc = ::accept4(getFD(), (struct sockaddr *)&clientInfo, &addrlen, SOCK_NONBLOCK | SOCK_CLOEXEC);
+#if TEST_SERVERSOCKET_FAULT_INJECTION
+    const size_t acceptCount = ++TestExternalServerSocketAcceptCount;
+    bool injectedError = false;
+    if (net::Defaults.testExternalServerSocketAcceptSimpleErrorInterval > 0 &&
```
mmeeks, line 1226: please see simulateSocketError for an example of how to do this - could extend the enum paraeter there. We should remove the TEST_SERVERSOCKET... etc.

net/Socket.cpp
```diff
@@ -1145,63 +1158,160 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
+        injectedError = true;
+        LOG_DBG("Injecting recoverable accept failure: EAGAIN: " << acceptCount);
+    }
+    else if (net::Defaults.testExternalServerSocketAcceptFatalErrorInterval > 0 &&
+             acceptCount % net::Defaults.testExternalServerSocketAcceptFatalErrorInterval == 0)
+    {
+        rc = -1;
+        errno = EFAULT; // fatal error
+        injectedError = true;
+        LOG_DBG("Injecting fatal accept failure: EAGAIN: " << acceptCount);
```
mmeeks, line 1240: If this is for unit testing we should use the existing UnitWSD virtual test hooks to ensure we can trigger and test this in a C++ unit test - in the normal way of such things.

net/Socket.cpp
```diff
@@ -1145,63 +1158,160 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
 #endif
 }
 
+bool ServerSocket::isRetryAcceptError(const int cause)
```
mmeeks, line 1161: Really not a fan of this huge list of extremely marginal and/or unusual errors. Previously we'd fail hard in these cases - lets handle EINTR/EAGAIN/EWOULDBLOCK one way, and EMFILE the other in-line, and ERR_SYS and continue with everythign else (?).

net/Socket.cpp
```diff
@@ -1145,63 +1158,160 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
-            LOG_TRC("Accepted socket #" << _socket->getFD() << " has family "
-                                        << clientInfo.sin6_family << ", " << *_socket);
-#else
-            std::shared_ptr<Socket> _socket = createSocketFromAccept(rc, Socket::Type::Unix);
-#endif
-            return _socket;
-        }
-        return std::shared_ptr<Socket>(nullptr);
+    const size_t extConnCount = StreamSocket::getExternalConnectionCount();
+    if( 0 < net::Defaults.maxExtConnections && extConnCount >= net::Defaults.maxExtConnections )
```
mmeeks, line 1289: Whitespace if(<space>foo) doesn't fit with surrounding code - please don't use it.

net/Socket.hpp
```diff
@@ -1056,6 +1056,19 @@ class StreamSocket : public Socket,
         _readType(readType),
         _inputProcessingEnabled(true)
     {
+#if TEST_SERVERSOCKET_FAULT_INJECTION
```
mmeeks, line 1059: We should have a unit test that tests this using a normal unit test hook - two lines at the site, and then more work in test/

mmeeks (changes requested):
Needs some more work; thanks.

net/Socket.cpp
sgothel: OK

net/Socket.cpp
sgothel: Sounds much better, checking.
Right now we only use the net::Defaults in the config hook, right.

net/Socket.cpp
sgothel: Yes, it is huge and determination of recoverable errors depends on whole system esp. for EMFILE/ENFILE and the other 'out of memory' symbols.
The latter memory causes may also be candidates to influence the discussed finer grained limitation behavior at a later stage.
Will refine it.

net/Socket.cpp
sgothel: OK

net/Socket.hpp
sgothel: as discussed above, OK

sgothel: revised.

net/Socket.cpp
```diff
@@ -1251,11 +1309,27 @@ bool Socket::isLocal() const
...
+        constexpr const char * messagePrefix = "Failed to accept. (errno: ";
+        const int cause = errno;
+        if (isRetryAcceptError(cause))
+        {
+            // Recoverable error, ignore to retry
+            LOG_DBG(messagePrefix << std::to_string(cause) << ", " << std::strerror(cause) << ')');
+        }
+        else
+        {
+            // Unrecoverable error, throw, which will close `ServerSocket` and discard it from `AcceptPoll` during `SocketPoll::poll`
```
caolanm, line 1323: The comment here claims it throws, but it actually exits now

sgothel: good catch, thx. yup, we moved to fatal-exit. throwing only inside the unit-test hooks injecting failures (otherwise test can't pass).

net/Socket.cpp
```diff
@@ -1149,63 +1151,119 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
+
 std::shared_ptr<Socket> ServerSocket::accept()
 {
     // Accept a connection (if any) and set it to non-blocking.
     // There still need the client's address to filter request from POST(call from REST) here.
 #if !MOBILEAPP
     assert(_type != Socket::Type::Unix);
 
+#if ENABLE_DEBUG
+    UnitWSD* const unitWsd = UnitWSD::isUnitTesting() ? &UnitWSD::get() : nullptr;
```
mmeeks, line 1186: Much better. But please git grep for UnitWSD::isUnitTesting - and notice the common pattern for whether we need to add ENABLE_DEBUG #ifdefs around - I would expect you to read four other call-sites before using a new method in your code - and then to produce something that fits in. Please remove the #if  block.

net/Socket.cpp
```diff
@@ -1149,63 +1151,119 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
+                << rc << " has family "
+                << clientInfo.sin6_family << ", address " << addrstr << ":" << clientInfo.sin6_port);
+        ::close(rc);
+        return nullptr;
+    }
+    try
+    {
+        // Create a socket object using the factory.
+        std::shared_ptr<Socket> _socket = createSocketFromAccept(rc, type);
+#if ENABLE_DEBUG
```
mmeeks, line 1249: ditto above comment on ifdefs.

mmeeks (changes requested):
Nearly there.

net/Socket.cpp
```diff
@@ -1149,63 +1151,119 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
-                inAddr = &(ipv6->sin6_addr);
-                type = Socket::Type::IPv6;
-            }
+            // Recoverable error, ignore to retry
+            LOG_DBG(messagePrefix << std::to_string(cause) << ", " << std::strerror(cause) << ')');
+        }
+        else
+        {
+            // Unrecoverable error, forceExit instead of throw (closing `ServerSocket` and discard it from `AcceptPoll` during `SocketPoll::poll`)
+            // May allow recovery from temporary EMFILE, ENFILE, ENOBUFS and ENOMEM errors?
```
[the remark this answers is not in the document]
sgothel: The 2nd annotation (comment) was an earmark whether we should handle the mentioned failures, which may be potentially temporary. Removing the note.

The 1st one (a statement) reminds us that if we would not exit the application (forcedExit) but throw an exception, the application would become 'def' no more listening).

net/Socket.cpp
```diff
@@ -1149,63 +1151,119 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
 #endif
 }
 
+bool ServerSocket::isRetryAcceptError(const int cause)
+{
+    switch(cause)
+    {
+        case EINTR:
+        case EAGAIN:        // == EWOULDBLOCK
+        case ENETDOWN:      // man accept(2): to be treated like EAGAIN
```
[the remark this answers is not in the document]
sgothel: OK

net/ServerSocket.hpp
```diff
@@ -29,6 +29,10 @@ class SocketFactory
 class ServerSocket : public Socket
 {
 public:
+    /// Socket ::accept() recoverable retry errors.
+    /// See [accept(2)](https://www.man7.org/linux/man-pages/man2/accept.2.html) errors leading to retry
```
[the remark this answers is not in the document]
sgothel: ok

net/Socket.cpp
```diff
@@ -1149,63 +1151,119 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
+        case EHOSTDOWN:     // man accept(2): to be treated like EAGAIN
+#ifdef ENONET
+        case ENONET:        // man accept(2): to be treated like EAGAIN (undef on BSD, APPLE, ..)
+#endif
+        case EHOSTUNREACH:  // man accept(2): to be treated like EAGAIN
+        case EOPNOTSUPP:    // man accept(2): to be treated like EAGAIN
+        case ENETUNREACH:   // man accept(2): to be treated like EAGAIN
+        case ECONNABORTED:  // OK
+        case ETIMEDOUT:     // OK, should not happen due ppoll
+            return true;
```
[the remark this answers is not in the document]
sgothel: changed to
`if (isRetryAcceptError(cause) || cause == EMFILE || cause == ENFILE || cause == ENOMEM || cause == ENOBUFS)`
which allows to these rare temporary failures not to exit the server application.

Wanted to have discussed these for a later more specific handling of low resources, but very reasonable to try again ofc.

net/Socket.cpp
sgothel: I looked for `StreamSocket::simulateSocketError` as a boiler plate, 
used only `#if ENABLE_DEBUG` removing side-effects.
Yes, `UnitWSD::isUnitTesting()` is unconditionally build-in otherwise (your preference), changing it.
- I have looked at `WebSocketHandler` earlier, etc
- Personally I would have added a `constexpr` `isDebugEnabled()` expression to remove this specific fault injection from the product, while still covering code analysis by the compiler.

Note that the pattern `UnitWSD::get().isUnitTesting()` used is not ideal,
since both `get()` and `isUnitTesting()` are static methods of `UnitWSD`.

net/Socket.cpp
sgothel: OK

mmeeks:
Otherwise - much better of course.

net/Socket.cpp
```diff
@@ -1149,63 +1151,115 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
 #endif
-    LOG_TRC("Accepted socket #" << rc << ", creating socket object.");
-    try
+    if (rc < 0)
     {
-        // Create a socket object using the factory.
-        if (rc != -1)
+        const int cause = errno;
+        constexpr const char * messagePrefix = "Failed to accept. (errno: ";
+        if (isRetryAcceptError(cause) || cause == EMFILE || cause == ENFILE || cause == ENOMEM || cause == ENOBUFS)
```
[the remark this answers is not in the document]
sgothel: The evolution of things ..
First I used an all case covering function, removed due to your request (long case switch list and like pls put the other cases inline - OK).
Now we see that it wasn't too bad (but remove the previously unused check-method) .. with the opposite logic.
So I change `isRetryAcceptError` -> `isUnrecoverableAcceptError` and have it do the logging, so only the clean-cut query and 'fatal' branch remains. Good.

net/Socket.cpp
```diff
@@ -1149,63 +1151,115 @@ bool ServerSocket::bind([[maybe_unused]] Type type, [[maybe_unused]] int port)
...
+    {
+        struct sockaddr_in6 *ipv6 = &clientInfo;
+        inAddr = &(ipv6->sin6_addr);
+        type = Socket::Type::IPv6;
+    }
+    ::inet_ntop(clientInfo.sin6_family, inAddr, addrstr, sizeof(addrstr));
 
-            std::shared_ptr<Socket> _socket = createSocketFromAccept(rc, type);
+    const size_t extConnCount = StreamSocket::getExternalConnectionCount();
+    if (0 < net::Defaults.maxExtConnections && extConnCount >= net::Defaults.maxExtConnections)
```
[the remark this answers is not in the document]
sgothel: OK.
(Last time I talked with my son about this, we realized our reading comes from simple math like `m< i < n`, which we are used to - plus the usual safety. So sometimes I still miss to avoid this for this project.)

sgothel:
missed one `amend'

mmeeks (approved):
Lovely thank you - didn't read the unit tests much; but really glad to see them =)

PR #10391, opened 2024-11-01
Keyboard: allow Ctrl+shift+I shortcut

meven:
Change-Id: Ic57c7f0825522b78f33f23fcebba4b4f8d9ba15b

* Target version: master 

### Summary

Allow to trigger the browser web developper tools with the short Ctrl+Shift+I

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -662,6 +662,11 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// Control + Shift + I, open browser developper tools
+		if (this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.I) {
+			return true;
+		}
+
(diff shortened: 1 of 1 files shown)


Rash419:
@meven can we also do this for `Ctrl+Shift+r` if it doesn't belong to any other shortcuts in application

meven:
> @meven can we also do this for `Ctrl+Shift+r` if it doesn't belong to any other shortcuts in application

This one, I don't think we want. It is known to users. Users using the shortcut willy-nilly would increase the load on wsd while it shouldn't be necessary.

This could be allowed but conditionally though.

vmiklos:
Could you please explain (ideally in the commit message) why is this needed when F12 already opens devtools? Or is this something different? I probably don't understand. :-) Thanks.

hfiguiere:
image

Ctrl-Shift-I is the standard shortcut in Firefox.

meven:
> image
> 
> Ctrl-Shift-I is the standard shortcut in Firefox.

And Chrome as well.
image

vmiklos (approved):
Looks good now, thanks.

meven:
> Could you please explain (ideally in the commit message) why is this needed when F12 already opens devtools? Or is this something different? I probably don't understand. :-) Thanks.

Ctrl+Shift+I is "inspect" what's under the cursor compared to F12 (that I didn't know) that just opens the devtool panel.

PR #10393, opened 2024-11-01
ForKit: log and count killed child processes

meven:
Change-Id: I093c0640216a2346808c7cf84e2ea7b005ef01dc

* Resolves: # 
* Target version: master 

### Summary

cc @timar 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ForKit.cpp
+++ b/kit/ForKit.cpp
@@ -281,25 +281,61 @@ static void cleanupChildren()
     pid_t exitedChildPid;
     int status = 0;
     int segFaultCount = 0;
+    int killedCount = 0;
+    int oomKilledCount = 0;
 
+    siginfo_t info;
     // Reap quickly without doing slow cleanup so WSD can spawn more rapidly.
-    while ((exitedChildPid = waitpid(-1, &status, WUNTRACED | WNOHANG)) > 0)
+    while (waitid(P_ALL, -1, &info, WEXITED | WUNTRACED | WNOHANG) == 0)
     {
+        if (info.si_pid == 0)
+        {
+            // WNOHANG special case
+            break;
+        }
+
+        exitedChildPid = info.si_pid;
+        status = info.si_status;
+
         const auto it = childJails.find(exitedChildPid);
         if (it != childJails.end())
         {
-            if (WIFSIGNALED(status) && (WTERMSIG(status) == SIGSEGV ||
-                                        WTERMSIG(status) == SIGBUS ||
-                                        WTERMSIG(status) == SIGABRT))
+            if (WIFSIGNALED(status))
             {
-                ++segFaultCount;
+                if (WTERMSIG(status) == SIGSEGV || WTERMSIG(status) == SIGBUS ||
+                    WTERMSIG(status) == SIGABRT)
+                {
+                    ++segFaultCount;
 
-                std::string noteCrashFile(it->second + "/tmp/kit-crashed");
-                int noteCrashFD = open(noteCrashFile.c_str(), O_CREAT | O_TRUNC | O_WRONLY, S_IRUSR | S_IWUSR);
-                if (noteCrashFD < 0)
-                    LOG_ERR("Couldn't create file: " << noteCrashFile << " due to error: " << strerror(errno));
+                    std::string noteCrashFile(it->second + "/tmp/kit-crashed");
+                    int noteCrashFD = open(noteCrashFile.c_str(), O_CREAT | O_TRUNC | O_WRONLY,
+                                           S_IRUSR | S_IWUSR);
+                    if (noteCrashFD < 0)
+                        LOG_ERR("Couldn't create file: " << noteCrashFile
+                                                         << " due to error: " << strerror(errno));
+                    else
+                        close(noteCrashFD);
+                }
+                else if (WTERMSIG(status) == SIGKILL)
+                {
+                    // TODO differentiate with docker
+                    int si_code = info.si_code;
+                    if (si_code == SI_KERNEL)
+                    {
+                        ++oomKilledCount;
+                        LOG_WRN("Child " << exitedChildPid << " was killed by OOM, with status"
+                                         << status);
+                    }
+                    else
+                    {
+                        ++killedCount;
+                        LOG_WRN("Child " << exitedChildPid << " was killed, with status" << status);
+                    }
+                }
                 else
-                    close(noteCrashFD);
+                {
+                    LOG_ERR("Child " << exitedChildPid << " has exited, with status" << status);
+                }
             }
 
             LOG_INF("Child " << exitedChildPid << " has exited, will remove its jail [" << it->second << "].");
@@ -313,7 +349,7 @@ static void cleanupChildren()
         }
         else
         {
-            LOG_ERR("Unknown child " << exitedChildPid << " has exited");
+            LOG_ERR("Unknown child " << exitedChildPid << " has exited, with status" << status);
         }
     }
 
@@ -327,12 +363,14 @@ static void cleanupChildren()
                                           << childJails.size() << " left: " << oss.str());
     }
 
-    if (segFaultCount)
+    if (segFaultCount || killedCount || oomKilledCount)
     {
         if (WSHandler)
         {
             std::stringstream stream;
-            stream << "segfaultcount " << segFaultCount << '\n';
+            stream << "segfaultcount=" << segFaultCount << ' ' << "killedcount=" << killedCount
+                    << ' ' << "oomkilledcount=" << oomKilledCount << '\n';
+
             int ret = WSHandler->sendMessage(stream.str());
             if (ret == -1)
             {
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -964,9 +964,11 @@ void Admin::setDocWopiUploadDuration(const std::string& docKey, const std::chron
     addCallback([this, docKey, uploadDuration]{ _model.setDocWopiUploadDuration(docKey, uploadDuration); });
 }
 
-void Admin::addSegFaultCount(unsigned segFaultCount)
+void Admin::addErrorExitCounters(unsigned segFaultCount, unsigned killedCount,
+                                 unsigned oomKilledCount)
 {
-    addCallback([this, segFaultCount]{ _model.addSegFaultCount(segFaultCount); });
+    addCallback([this, segFaultCount, killedCount, oomKilledCount]
+                { _model.addErrorExitCounters(segFaultCount, killedCount, oomKilledCount); });
 }
 
 void Admin::addLostKitsTerminated(unsigned lostKitsTerminated)
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -1017,19 +1017,42 @@ void ForKitProcWSHandler::handleMessage(const std::vector<char> &data)
     const std::string firstLine = COOLProtocol::getFirstLine(&data[0], data.size());
     const StringVector tokens = StringVector::tokenize(firstLine.data(), firstLine.size());
 
-    if (tokens.equals(0, "segfaultcount"))
+    if (tokens.startsWith(0, "segfaultcount"))
     {
-        int count = std::stoi(tokens[1]);
-        if (count >= 0)
+        int segFaultcount = 0;
+        int killedCount = 0;
+        int oomKilledCount = 0;
+        if (COOLProtocol::getNonNegTokenInteger(tokens[0], "segfaultcount", segFaultcount)
+            && COOLProtocol::getNonNegTokenInteger(tokens[1], "killedcount", killedCount)
+            && COOLProtocol::getNonNegTokenInteger(tokens[2], "oomkilledcount", oomKilledCount))
         {
-            Admin::instance().addSegFaultCount(count);
-            LOG_INF(count << " coolkit processes crashed with segmentation fault.");
-            SigUtil::addActivity("coolkit(s) crashed");
-            UnitWSD::get().kitSegfault(count);
+            Admin::instance().addErrorExitCounters(segFaultcount, killedCount, oomKilledCount);
+
+            if (segFaultcount)
+            {
+                LOG_INF(segFaultcount << " coolkit processes crashed with segmentation fault.");
+                SigUtil::addActivity("coolkit(s) crashed");
+                UnitWSD::get().kitSegfault(segFaultcount);
+            }
+
+            if (killedCount)
+            {
+                LOG_INF(killedCount << " coolkit processes killed.");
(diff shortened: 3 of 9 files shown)


kit/ForKit.cpp
```diff
@@ -281,39 +281,56 @@ static void cleanupChildren()
...
+                LOG_INF("Child " << exitedChildPid << " has exited, will remove its jail ["
+                                 << it->second << "].");
+                cleanupJailPaths.emplace_back(it->second);
+                childJails.erase(it);
+                if (childJails.empty() && !SigUtil::getTerminationFlag())
+                {
+                    // We ran out of kits and we aren't terminating.
+                    LOG_WRN("No live Kits exist, and we are not terminating yet.");
+                }
             }
```
mmeeks, line 330: By splitting the if (WIFSIGNALED() check - surely we changed the behavior here. We no longer warn and clean up in the !WIFSiGNALED() case ?

Is that expected, deliberate etc. ?

kit/ForKit.cpp
```diff
@@ -345,6 +362,24 @@ static void cleanupChildren()
         }
     }
 
+    if (killedCount)
+    {
+        if (WSHandler)
+        {
+            std::stringstream stream;
+            stream << "killedcount " << killedCount << '\n';
+            int ret = WSHandler->sendMessage(stream.str());
```
mmeeks, line 371: Why is this new / duplicate code - surely we could share / re-factor the other use-case ? I expect we could send a single message with three counts in it and reduce code duplication.

kit/ForKit.cpp
```diff
@@ -281,39 +281,56 @@ static void cleanupChildren()
...
+                    std::string noteCrashFile(it->second + "/tmp/kit-crashed");
+                    int noteCrashFD = open(noteCrashFile.c_str(), O_CREAT | O_TRUNC | O_WRONLY,
+                                           S_IRUSR | S_IWUSR);
+                    if (noteCrashFD < 0)
+                        LOG_ERR("Couldn't create file: " << noteCrashFile
+                                                         << " due to error: " << strerror(errno));
+                    else
+                        close(noteCrashFD);
+                }
+                else if (WTERMSIG(status) == SIGKILL || WTERMSIG(status) == SIGTERM)
```
mmeeks, line 310: I thought that OOM was detectable from some high-bits set in the status (?) - can we separate known OOM kills from other kills and count / report on that ?

Thanks!

meven: Unfortunately after more research, there is no specific signal.
The oom killer just sends SIGKILL (sometimes SIGTERM first).

There aren't many good sources.
https://www.kernel.org/doc/gorman/html/understand/understand016.html

There seem to be a OOMKilled signal specific to docker.

Regardless logging kill signals is still useful and we can expect OOM to be the main origin.

kit/ForKit.cpp
meven: It was missing.

kit/ForKit.cpp
mmeeks: So - lets detect OOM killers separately since that is/was what was asked; and some code pointers:

https://github.com/torvalds/linux/blob/master/mm/oom_kill.c#L947
	do_send_sig_info(SIGKILL, SEND_SIG_PRIV, victim, PIDTYPE_TGID);
https://github.com/torvalds/linux/blob/master/kernel/signal.c#L1135

Should have siginfo's si_code == SI_KERNEL as a source == OOM. cf. https://man7.org/linux/man-pages/man2/sigaction.2.html

We should also detect the docker OOM and glup that into it which would be with an exit with code 137 and add those two.

Of course counting other types of KILL is fine too - lets report 3 numbers crashes, kills, OOM-kills back in the same method =)

Thanks !

kit/ForKit.cpp
meven: Now using a lambda.

kit/ForKit.cpp
```diff
@@ -327,23 +342,28 @@ static void cleanupChildren()
...
+                else
+                {
+                    LOG_WRN("Successfully sent '" << countName << "' message " << stream.str());
+                }
             }
         }
-    }
+    };
+
+    sendCountMessage( "segfaultcount", segFaultCount);
```
mmeeks, line 365: One message is better than three - I'd pack all three numbers into one message =)  Of course - we would expect some code on the other side in COOLWSD.cpp(?) to read & interpret that - and this would save duplication there too - you'll get to that bit too (?)

Anyhow - code flow looking better :-)

mmeeks (changes requested):

kit/ForKit.cpp
meven: Handled now.

kit/ForKit.cpp
meven: Now use a simple loop to handle this.

wsd/metrics.txt
```diff
@@ -30,6 +30,7 @@ KITS
     kit_assigned_count – number of running kit processes that are assigned to documents (number of currently open documents).
     kit_segfault_count - number of kit processes terminated with SIGSEGV or SIGBUS signals since the start of application.
     kit_lost_terminated_count - number of kit processes that were lost by coolwsd and were terminated by cleanup mechanism.
+    kit_segfault_count - number of kit processes that were killed.
```
timar, line 33: kit_killed_count here

wsd/COOLWSD.cpp
```diff
@@ -1015,27 +1016,57 @@ class AppConfigMap final : public Poco::Util::MapConfiguration
...
 {
     LOG_TRC("ForKitProcWSHandler: handling incoming [" << COOLProtocol::getAbbreviatedMessage(&data[0], data.size()) << "].");
-    const std::string firstLine = COOLProtocol::getFirstLine(&data[0], data.size());
-    const StringVector tokens = StringVector::tokenize(firstLine.data(), firstLine.size());
 
-    if (tokens.equals(0, "segfaultcount"))
-    {
-        int count = std::stoi(tokens[1]);
-        if (count >= 0)
+    std::stringstream ss(std::string(data.cbegin(), data.cend()));
```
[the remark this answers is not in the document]
meven: My current patch would never sent 0 values, and send faultcount values independently at the cost of data of more bytes per-value passed on the Wire.
Grouping packing them makes sense indeed, didn't think about it.

By `get` methods, I see `StringVector::getUInt32` `StringVector::getNameIntegerPair` or `StringVector::getParam` that can't be used to parse "key int int int" or a single uint32_t value. I need to extend this a bit.

kit/ForKit.cpp
mmeeks: Did you read the above Meven ?

wsd/COOLWSD.cpp
```diff
@@ -1017,56 +1016,39 @@ void ForKitProcWSHandler::handleMessage(const std::vector<char> &data)
...
-        auto segFaultcountOpt = parseCount("segfaultcount");
-        if (segFaultcountOpt && segFaultcountOpt.value()) {
-            const auto segFaultcount = segFaultcountOpt.value();
-            Admin::instance().addSegFaultCount(segFaultcount);
-            LOG_INF(segFaultcount << " coolkit processes crashed with segmentation fault.");
-            SigUtil::addActivity("coolkit(s) crashed");
-            UnitWSD::get().kitSegfault(segFaultcount);
-        }
+            if (segFaultcount) {
+                Admin::instance().addSegFaultCount(segFaultcount);
```
mmeeks, line 1034: Can we not really just adapt the method to the Admin::instance to also take three parameters ?

common/StringVector.cpp
```diff
@@ -127,4 +127,11 @@ StringVector StringVector::tokenizeAnyOf(const std::string& s, const char* delim
     return StringVector(s, std::move(tokens));
 }
 
+bool StringVector::getIntParam(std::size_t index, int &value) const
```
mmeeks, line 130: Is there really no way to get an integer parameter from a StringVector already ?
Can you see how the existing method:
    bool getUInt32(std::size_t index, const std::string& key, uint32_t& value) const;
Works - and see how other protocol messages are structured, and re-use the same key/value schema ?

mmeeks (changes requested):
I'd like to see the OOM detection done - by reading the links I provided and have that number separate from the other bits.
Having four commits doing the same thing stacked up is -far- worse than having a single one - can you squash all of these down to one commit for the next iteration.

And a few other comments; thanks !

kit/ForKit.cpp
meven: I did.

But https://linux.die.net/man/2/waitid had the solution.

I found `waitid` to have siginfo.

Latest commit uses it.
Haven't tested the oomKilled counter yet, need to tweak things with cgroup.

wsd/COOLWSD.cpp
```diff
@@ -1015,27 +1016,57 @@ class AppConfigMap final : public Poco::Util::MapConfiguration
...
+        }
+
+        return std::nullopt;
+    };
+
+    while(std::getline(ss, line,'\n'))
+    {
+        tokens = StringVector::tokenize(line.data(), line.size());
+
+        auto segFaultcountOpt = parseCount("segfaultcount");
```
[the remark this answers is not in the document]
meven: This is done.

common/StringVector.cpp
meven: ` bool getUInt32(std::size_t index, const std::string& key, uint32_t& value) const;`
can't be used here given the input format. It can only parse "key intvalue" or "key=intvalue".
But here I have currently "key value value value".
The format can be changed to follow to more conventional: 
```
key1 intvalue
key2 intvalue
...
```
I was using this pattern before sending everything in a single message.

kit/ForKit.cpp
```diff
@@ -281,25 +281,60 @@ static void cleanupChildren()
...
+                                           S_IRUSR | S_IWUSR);
+                    if (noteCrashFD < 0)
+                        LOG_ERR("Couldn't create file: " << noteCrashFile
+                                                         << " due to error: " << strerror(errno));
+                    else
+                        close(noteCrashFD);
+                }
+                else if (WTERMSIG(status) == SIGKILL)
+                {
+                    int si_code = info.si_code;
```
mmeeks, line 321: Looks good - but will that also catch the docker OOM exit state that IIRC had a higher bit set (?)

kit/ForKit.cpp
```diff
@@ -327,23 +362,25 @@ static void cleanupChildren()
...
-            int ret = WSHandler->sendMessage(stream.str());
-            if (ret == -1)
             {
-                LOG_WRN("Could not send 'segfaultcount' message through websocket");
-            }
-            else
-            {
-                LOG_WRN("Successfully sent 'segfaultcount' message " << stream.str());
+                std::stringstream stream;
+                stream << "faultcount " << segFaultCount << ' ' << killedCount << ' '
```
mmeeks, line 370: much better =) though we could use keys for that faultcount foo=3 baa=5 oom=1 - like most of the other commands we use, and then have it be more descriptive and not need the new StringToken method (?) =)

wsd/Admin.cpp
```diff
@@ -974,6 +974,16 @@ void Admin::addLostKitsTerminated(unsigned lostKitsTerminated)
     addCallback([this, lostKitsTerminated]{ _model.addLostKitsTerminated(lostKitsTerminated); });
 }
 
+void Admin::addKilledCount(unsigned killedCount)
```
mmeeks, line 977: Hmm - we have three numbers - surely we can thread them through all this logic with a single method call, single callback etc. ? =) bearing in mind we're trying to add as little code as possible.

wsd/COOLWSD.cpp
mmeeks: Which was this comment I guess =) lets just make a single message to the addActivity with all three counts etc.

mmeeks (changes requested):
Much better - good to see ! =) well found on the waitid call - but please read the diff and get it shorter; this three-fold code duplication around the three counts is still not elegant - read the diff; and think: "how can I evaporate as much of this as possible ?" less code is more maintainable code.

kit/ForKit.cpp
meven: docker OOM, `137` status is just 128 (means status is a signal) + 9 (the signal value, SIGKILL).

It is my understanding that docker leverage Linux cgroup limits to have the kernel handle the oom killing itself https://docs.docker.com/engine/containers/resource_constraints/

kit/ForKit.cpp
meven: Its status is then 137 which mean 128 (status is a signal) + 9 (SIGKILL).
Its si_code should be  SI_KERNEL still as the kernel is the one doing the killing for docker.

kit/ForKit.cpp
mmeeks: Ah ok =) fair enough; I assumed there was some other signalling here.

wsd/COOLWSD.cpp
meven: Done

kit/ForKit.cpp
meven: Done

wsd/Admin.cpp
meven: Done

mmeeks (approved):
Let's get it in like this. Thanks Meven.

timar:
> Let's get it in like this. Thanks Meven.
Providing that CI passes. CodeQL scanning - action / Analyze (cpp) is complaining already.

meven:
The failuer of cypress tests is due to the code changes:
```
wsd-3516065-3516098 2024-11-05 16:13:32.750684 +0000 [ prisoner_poll ] TST  HTTPCrashTest::testBarren [exitTest] (+9ms): ERROR: FAILURE: exitTest: TestResult::Failed: kit killed| common/Unit.cpp:556
```
The patch makes test fail if a child process is killed.
The test might to be updated.

mmeeks:
Hopefully you can get a fix in before the green freeze/tag tonight (?)

meven:
> > Hopefully you can get a fix in before the green freeze/tag tonight (?)
> 
> I have just pushed an update to the test that might fix it.

I need to customise `UnitClient` to igonre kill events that trigger a test fail currently.
Except UnitClient does not have a header and is only linked to files while providing by itself the instantiation its own implementation of `UnitBase *unit_create_wsd`.

I see a few choices:
 - Make UnitClient ignore kill events. I don't like it this weakens test assertions.
 - Add a header for UnitClient and allow to subclass it in `HTTPCrashTest` having its own  `UnitBase *unit_create_wsd`. That would require editing other tests in the process to change how they used `UnitClient` so far.

mmeeks:
Thanks for getting it over the line - lets make sure we improve the tests - ideally testing the new killed counter too =)

PR #10394, opened 2024-11-01
Private/gokay/clipboard 2

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -64,9 +64,6 @@ L.Control.DownloadProgress = L.Control.extend({
 				msg, buttonText, this._onStartDownload.bind(this), this.options.snackbarTimeout);
 
 			this.setupKeyboardShortcutForSnackbar();
-
-			// Snackbars cannot get focus, but we are using it with a button and it requires to be focused. No need to change snackbar imp for now.
-			document.getElementById('button').focus(); // TODO: This "button" id is too generic. It could be something like "snackbar-button" etc.
 		} else {
 			this._map.uiManager.showInfoModal(modalId, this._getDialogTitle(), msg, '',
 				buttonText, this._onStartDownload.bind(this), true, modalId + '-response');
@@ -114,12 +111,14 @@ L.Control.DownloadProgress = L.Control.extend({
 		var buttonText = _('Copy') + L.Util.replaceCtrlAltInMac(' (Ctrl + C)');
 
 		if (inSnackbar) {
-			this._map.uiManager.showProgressBar(snackbarMsg, buttonText,
-				this._onConfirmCopyAction.bind(this), this.options.snackbarTimeout);
-
 			this._map.uiManager.setSnackbarProgress(100);
 
-			this.setupKeyboardShortcutForSnackbar();
+			this._map.uiManager.showSnackbar(
+				snackbarMsg, buttonText, this._onConfirmCopyAction.bind(this), this.options.snackbarTimeout);
+
+			setTimeout(() => {
+				this.setupKeyboardShortcutForSnackbar();
+			}, 100);
 		} else {
 			JSDialog.setMessageInModal(modalId, dialogMsg, '');
 			JSDialog.enableButtonInModal(modalId, modalId + '-response', true);
@@ -154,6 +153,10 @@ L.Control.DownloadProgress = L.Control.extend({
 
 	setupKeyboardShortcutForSnackbar: function () {
 		this._setupKeyboardShortcutForElement('snackbar-container', 'button');
+
+		// Snackbars cannot get focus, but we are using it with a button and it requires to be focused. No need to change snackbar imp for now.
+		if (document.getElementById('button'))
+			document.getElementById('button').focus(); // TODO: This "button" id is too generic. It could be something like "snackbar-button" etc.
 	},
 
 	// isLargeCopy specifies if we are copying and have to explain user the process
--- a/browser/src/control/Control.DownloadProgress.js
+++ b/browser/src/control/Control.DownloadProgress.js
@@ -64,6 +64,9 @@ L.Control.DownloadProgress = L.Control.extend({
 				msg, buttonText, this._onStartDownload.bind(this), this.options.snackbarTimeout);
 
 			this.setupKeyboardShortcutForSnackbar();
+
+			// Snackbars cannot get focus, but we are using it with a button and it requires to be focused. No need to change snackbar imp for now.
+			document.getElementById('button').focus(); // TODO: This "button" id is too generic. It could be something like "snackbar-button" etc.
 		} else {
 			this._map.uiManager.showInfoModal(modalId, this._getDialogTitle(), msg, '',
 				buttonText, this._onStartDownload.bind(this), true, modalId + '-response');
@@ -150,7 +153,7 @@ L.Control.DownloadProgress = L.Control.extend({
 	},
 
 	setupKeyboardShortcutForSnackbar: function () {
-		this._setupKeyboardShortcutForElement('snackbar', 'button');
+		this._setupKeyboardShortcutForElement('snackbar-container', 'button');
 	},
 
 	// isLargeCopy specifies if we are copying and have to explain user the process
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -400,7 +400,7 @@ L.Control.JSDialog = L.Control.extend({
 			focusWidget = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
 		}
 
-		if (focusWidget && document.activeElement !== focusWidget) {
+		if (focusWidget && document.activeElement !== focusWidget && instance.canHaveFocus) {
 			var firstFocusable = JSDialog.GetFocusableElements(focusWidget);
 			if (firstFocusable && firstFocusable.length)
 				firstFocusable[0].focus();
@@ -428,7 +428,7 @@ L.Control.JSDialog = L.Control.extend({
 		}
 		return null; // Return null if tabcontrol is not found
 	},
-	
+
 	/// if you use updatePos - instance param is binded automatically
 	setPosition: function(instance, updatedPos) {
 		var calculated = false;
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -501,12 +501,6 @@ L.Clipboard = L.Class.extend({
 	_getHtmlForClipboard: function() {
 		var text;
 
-		if ($('.ui-edit').is(':focus'))
-			return $('.ui-edit').value();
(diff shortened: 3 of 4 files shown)


browser/src/control/Control.DownloadProgress.js
```diff
@@ -114,12 +111,14 @@ L.Control.DownloadProgress = L.Control.extend({
 		var buttonText = _('Copy') + L.Util.replaceCtrlAltInMac(' (Ctrl + C)');
 
 		if (inSnackbar) {
-			this._map.uiManager.showProgressBar(snackbarMsg, buttonText,
-				this._onConfirmCopyAction.bind(this), this.options.snackbarTimeout);
-
 			this._map.uiManager.setSnackbarProgress(100);
 
-			this.setupKeyboardShortcutForSnackbar();
+			this._map.uiManager.showSnackbar(
```
vmiklos, line 116: You also replace the progressbar with a snackbar here, which is not mentioned in the commit message. But it's probably OK. Could you please describe why do you change it, though?

browser/src/map/Clipboard.js
```diff
@@ -1170,7 +1170,7 @@ L.Clipboard = L.Class.extend({
 		if (!this._downloadProgress || this._downloadProgress.isClosed())
 			return;
 
-		if (this._downloadProgressStatus() === 'downloadButton')
+		if (['downloadButton', 'confirmPasteButton'].includes(this._downloadProgressStatus()))
```
vmiklos, line 1173: Same here: OK to hide the download widget in more cases, but could you describe why is this necessary? What is the case when we get here with a downloadButton case and when do we get confirmPasteButton?

browser/src/control/Control.DownloadProgress.js
```diff
@@ -150,7 +153,7 @@ L.Control.DownloadProgress = L.Control.extend({
 	},
 
 	setupKeyboardShortcutForSnackbar: function () {
-		this._setupKeyboardShortcutForElement('snackbar', 'button');
+		this._setupKeyboardShortcutForElement('snackbar-container', 'button');
```
vmiklos, line 155: OK, but why is this needed? Not clear how we had snackbar here in the past and how snackbar-container is better now.

browser/src/control/Control.DownloadProgress.js
```diff
@@ -64,6 +64,9 @@ L.Control.DownloadProgress = L.Control.extend({
 				msg, buttonText, this._onStartDownload.bind(this), this.options.snackbarTimeout);
 
 			this.setupKeyboardShortcutForSnackbar();
+
+			// Snackbars cannot get focus, but we are using it with a button and it requires to be focused. No need to change snackbar imp for now.
+			document.getElementById('button').focus(); // TODO: This "button" id is too generic. It could be something like "snackbar-button" etc.
```
vmiklos: https://github.com/CollaboraOnline/online/pull/10394/commits/213974e6619b55010008d3aba17f5277201d0f42 has a check here for the case when the getElementById() result is undefined, is it intentional that you don't do the same here? Maybe extract to a function to avoid the duplication?

browser/src/control/Control.JSDialog.js
```diff
@@ -400,7 +400,7 @@ L.Control.JSDialog = L.Control.extend({
 			focusWidget = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
 		}
 
-		if (focusWidget && document.activeElement !== focusWidget) {
+		if (focusWidget && document.activeElement !== focusWidget && instance.canHaveFocus) {
```
vmiklos, line 403: Is it fine to just not do anything in this case? You don't mention in a comment or in the commit message if you assume that the outer element has the focus already at this point.

browser/src/map/Clipboard.js
```diff
@@ -536,33 +530,31 @@ L.Clipboard = L.Class.extend({
 	populateClipboard: function(ev) {
 		this._checkSelection();
 
-		if (this._navigatorClipboardWrite()) {
-			// This is the codepath where the browser initiates the clipboard operation,
```
vmiklos, line 540: I found this comment useful: there are two cases to handle (user clicks on copy button or user presses Ctrl-C) and it's not trivial which one is handled where. Would you consider restoring it?

Also, the early return had the benefit of keeping the intent level small. A usual refactor is to "flatten" a function by adding such early returns. Is there a benefit to do the opposite of that here?

browser/src/map/Clipboard.js
```diff
@@ -501,12 +501,6 @@ L.Clipboard = L.Class.extend({
 	_getHtmlForClipboard: function() {
 		var text;
 
-		if ($('.ui-edit').is(':focus'))
-			return $('.ui-edit').value();
-
-		if ($('.w2ui-input').is(':focus'))
-			return $('.w2ui-input').value();
-
```
vmiklos, line 509: Did you test if copying from comments still work? That kind of info would go to the commit message ideally. :-)

browser/src/map/Clipboard.js
```diff
@@ -857,10 +857,6 @@ L.Clipboard = L.Class.extend({
 	// Executes the navigator.clipboard.write() call, if it's available.
 	_navigatorClipboardWrite: function() {
 		if (!L.Browser.clipboardApiAvailable) {
-			// Show a visible warning, this should not happen in production.
-			this._map.uiManager.showSnackbar(
-				_('The async Clipboard API is not supported by your browser, switching to HTTPS is meant to fix that.')
-			);
```
vmiklos, line 863: OK. I think initially @eszkadev wanted this, because first it was not clear that http:// gives no new/async clipboard, even on the latest Chrome. But if the warning causes more problems than it solves, then let's forget about it.

browser/src/map/Clipboard.js
```diff
@@ -923,7 +923,7 @@ L.Clipboard = L.Class.extend({
 			L.Browser.clipboardApiAvailable = false;
 			window.prefs.set('clipboardApiAvailable', 'false');
 			// Prefetch selection, so next time copy will work with the keyboard.
-			app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');
+			app.socket.sendMessage(that.getMetaURL() + '&MimeType=text/html,text/plain;charset=utf-8');
```
vmiklos, line 914: This one is really strange: the socket messages typically have a syntax where a lowercase command starts the line, then we have various parameters here. Do you really send a raw URL here or do I misread this change?

Also, you don't really describe what is the problem you try to fix here. :-)

wsd/ClientRequestDispatcher.cpp
```diff
@@ -1113,8 +1114,13 @@ bool ClientRequestDispatcher::handleClipboardRequest(const Poco::Net::HTTPReques
             tag = it.second;
         else if (it.first == "MimeType")
             mime = it.second;
+        else if (it.first == "charset")
+            charset = it.second;
     }
 
+    if (!charset.empty())
+        mime += ";charset=" + charset;
```
vmiklos, line 1122: I assume that the charset is then parsed by core, since you only produce this charset info, you don't add code to handle it somewhere. If this is the case, we could have a comment here stating this.

vmiklos (changes requested):
I added a number of inline comments with nits, but those are just small things, this mostly looks OK to me.

One problematic change stands out, though: could you please avoid `sendMessage(that.getMetaURL()...)` e.g. by keeping the gettextselection command there? Thanks. The rest looks good to me.

wsd/ClientRequestDispatcher.cpp
gokaysatir: It's handled below in line ~1167:

` else if (mime == "text/html,text/plain;charset=utf-8")`

browser/src/control/Control.JSDialog.js
gokaysatir: I think if "`instance.canHaveFocus`" is false then we musn't check for focusables. But maybe @eszkadev has a better idea here? This part may need some refactoring.

gokaysatir: Because that if statement conflicts with `instance.canHaveFocus` property.

browser/src/control/Control.DownloadProgress.js
gokaysatir: Narrows down the search. Not vital though :)

browser/src/map/Clipboard.js
gokaysatir: 'downloadButton' doesn't seem to be used, added to the commit message.

browser/src/control/Control.DownloadProgress.js
gokaysatir: Added a note.

browser/src/map/Clipboard.js
gokaysatir: Restored :)

browser/src/map/Clipboard.js
gokaysatir: :D Forgotten debugging thing.

browser/src/map/Clipboard.js
gokaysatir: Yep, tested just before sending the changes.

gokaysatir:
> I added a number of inline comments with nits, but those are just small things, this mostly looks OK to me.
> 
> One problematic change stands out, though: could you please avoid `sendMessage(that.getMetaURL()...)` e.g. by keeping the gettextselection command there? Thanks. The rest looks good to me.


Thanks for checking :) I sent the changes and tested them with traefik software befor sending.

browser/src/map/Clipboard.js
```diff
@@ -334,13 +334,9 @@ L.Clipboard = L.Class.extend({
 				);
 			}.bind(this),
 			function(progress) { return progress/2; },
-			function(response) {
+			function() {
 				window.app.console.log('failed to download clipboard using fallback html');
 
-				if (response) {
```
vmiklos, line 340: Nit: why did you remove this block? This is not something that should happen in practice, so it sounds useful to show a user-visible error in this case?

browser/src/map/Clipboard.js
vmiklos: I see this is gone from the latest version, good.

vmiklos (approved):
Looks good to me now, thanks.

In the future, please try to always have these paragraphs in commit messages:

- what is happening currently (as a user)
- problem description (what you think is the root of the problem, as a developer)

I hope that helps to describe the "why" in commit messages so I nag you less. :-) Thanks.

browser/src/map/Clipboard.js
gokaysatir: There is a warning via "showInfoModal" few lines below. Looks like it will be shown when the operation fails. So i removed the first one.

vmiklos: I see, thanks.

PR #10396, opened 2024-11-01
Various mostly stylistic cleanups

tml1024:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -92,22 +92,21 @@ std::string Socket::toString(Type t)
 
 int Socket::createSocket([[maybe_unused]] Socket::Type type)
 {
-    if constexpr (!Util::isMobileApp())
+#if !MOBILEAPP
+    int domain = AF_UNSPEC;
+    switch (type)
     {
-        int domain = AF_UNSPEC;
-        switch (type)
-        {
-        case Type::IPv4: domain = AF_INET;  break;
-        case Type::IPv6: domain = AF_INET6; break;
-        case Type::All:  domain = AF_INET6; break;
-        case Type::Unix: domain = AF_UNIX;  break;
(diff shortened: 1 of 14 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2415/console failed in:

> writer/annotation_spec.js

retry just that job to see if this is a one-off problem or not.

net/Socket.cpp
```diff
@@ -92,22 +92,21 @@ std::string Socket::toString(Type t)
 
 int Socket::createSocket([[maybe_unused]] Socket::Type type)
 {
-    if constexpr (!Util::isMobileApp())
+#if !MOBILEAPP
```
Ashod, line 95: The idea to deal with the compiler seeing unsupported functions is to stub them. We are trying to avoid compile-time directives because the code behind them can rot (at least that's my favorite reason, but there are others, including making code harder to read).

Ashod (approved):
Thanks, @tml1024. Nice cleanups.

Only the `#ifdef MOBILEAPP` worry me. I left comment as to how/why we're moving away from them (although for headers and other outside-functions code, we need them).

It's not a blocker for me, but best if @mmeeks can give his input.

net/Socket.cpp
tml1024: Yes, that is how I will do it in the future. But this PR has commits that are almost a month old by now... before I came to the conclusion that what you describe indeed is the best way. Ifdefs will be removed in later commits.

mmeeks (approved):

mmeeks:
Thanks Tor =)

PR #10397, opened 2024-11-01
feat: Allow drag and drop of A/V files

Minion3665:
In I881818277edb69ee1e98ba1bd3454e8d29593889, we allowed inserting multimedia into the document. We didn't, however, add any provision for dragging/dropping it into the document, so only images remained drag/droppable.

This commit is a preliminary change to enable drag/drop for audio/video files. It doesn't use the same copy/paste mechanism as inserting images (as this fails, which appears to be for the same reasons we need to specify a size manually on insert...). Instead, we use the normal insert pathway, as if the user had clicked the insert button and uploaded their file. This comes with a disadvantage by way of insert position: the video is always uploaded at the top left of the document. The way to fix this is probably to change the way we determine insert position, however, as the default behavior is poor for the insert media button too...

Change-Id: I3f3fae9cc4de55552d8f5d27077c141d37539b10

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -374,7 +374,7 @@ L.Clipboard = L.Class.extend({
 		);
 	},
 
-	_onFileLoadFunc: function(file) {
+	_onImageLoadFunc: function (file) {
 		var that = this;
 		return function(e) {
 			that._pasteTypedBlob(file.type, e.target.result);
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
Thanks, looks good :)

PR #10403, opened 2024-11-02
prevent double-dispatch of uno events bound to function keys

caolanm:
this is reproducible with F1 and a breakpoint on

app.map.showHelp('online-help-content');

in browser/src/docdispatcher.ts

Change-Id: Id140878d3a5d19450fe8141977c3324f97541cea

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -397,6 +397,11 @@ L.Map.Keyboard = L.Handler.extend({
 		if (this._map.uiManager.isUIBlocked())
 			return;
 
+		// if we press F1/F2 then the keyboard event is handled twice, use that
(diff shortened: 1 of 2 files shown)


browser/src/map/handler/Map.Keyboard.js
```diff
@@ -397,6 +397,11 @@ L.Map.Keyboard = L.Handler.extend({
 		if (this._map.uiManager.isUIBlocked())
 			return;
 
+		// if we press F1/F2 then the keyboard event is handled twice, use that
+		// the defaultPrevented is set the first time to avoid the double dispatch
+		if (ev.defaultPrevented)
```
eszkadev, line 402: interesting, probably we should check also why it is called 2 times and do it once

eszkadev (approved):

PR #10407, opened 2024-11-04
dumpState Improvements

Ashod:
Better formatting of `dumpState()` and more memory stats included. Also, we now forward SIGUSR1 to the child processes, to get their memory stats and other info.

- wsd: better dumpState formatting
- wsd: dump malloc-info with USR1
- wsd: refactor forwarding a signal to our children
- wsd: forward SIGUSR1 to children
- wsd: unify dump_malloc_state with getMallocInfo
- wsd: dump malloc_info in the children

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -232,6 +232,7 @@ void requestShutdown()
         signalLog(buf + i + 1);
     }
 
+#endif // !MOBILEAPP
 
     const char *signalName(const int signo)
     {
@@ -294,6 +295,8 @@ void requestShutdown()
         }
         // LCOV_EXCL_STOP Coverage for these is not very useful.
     }
+
+#if !MOBILEAPP
 
     static
     void handleTerminationSignal(const int signal)
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -44,6 +44,10 @@
 #include <fcntl.h>
 #include <spawn.h>
 
+#if defined __GLIBC__
+#include <malloc.h>
+#endif
+
 #include <atomic>
 #include <cassert>
 #include <chrono>
@@ -804,6 +808,31 @@ namespace Util
         SigUtil::SigHandlerTrap::wait();
 
         std::_Exit(code);
+    }
+
+    std::string getMallocInfo()
+    {
+        std::string info;
+
+#if defined __GLIBC__
+        size_t size = 0;
+        char* p = nullptr;
+        FILE* f = open_memstream(&p, &size);
+        if (f)
+        {
+            // Dump malloc internal structures.
+            malloc_info(0, f);
+            fclose(f);
+
+            if (p && size)
+            {
+                info = std::string(p, size);
+                free(p);
+            }
+        }
+#endif // __GLIBC__
+
+        return info;
     }
 
     bool matchRegex(const std::set<std::string>& set, const std::string& subject)
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -755,9 +755,9 @@ public:
     void dumpState(std::ostream& os, const std::string& indent = "\n  ") const
     {
         os << indent << "http::Request: " << _version << ' ' << _verb << ' ' << _url;
-        os << indent << "stage: " << name(_stage);
-        os << indent << "headers: ";
-        _header.serialize(os);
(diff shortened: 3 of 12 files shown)


net/HttpRequest.hpp
```diff
@@ -755,9 +755,9 @@ class Request final
     void dumpState(std::ostream& os, const std::string& indent = "\n  ") const
     {
         os << indent << "http::Request: " << _version << ' ' << _verb << ' ' << _url;
-        os << indent << "stage: " << name(_stage);
-        os << indent << "headers: ";
-        _header.serialize(os);
+        os << indent << "\tstage: " << name(_stage);
```
mmeeks, line 758: Curious as to why we're adding another indent to indent in-line here, rather than just bumping that in the parent caller ?

common/Util.cpp
```diff
@@ -806,6 +807,29 @@ namespace Util
         std::_Exit(code);
     }
 
+    std::string getMallocInfo()
```
mmeeks, line 813: Like it =)

wsd/COOLWSD.cpp
```diff
@@ -4976,20 +4978,30 @@ void forwardSigUsr2()
     Util::assertIsLocked(DocBrokersMutex);
     std::lock_guard<std::mutex> newChildLock(NewChildrenMutex);
 
+    forwardSignal(SIGUSR2);
+}
+
+void forwardSignal(const int signum)
```
mmeeks, line 4988: Nice minimal patch; good stuff.

wsd/COOLWSD.cpp
```diff
@@ -4959,6 +4959,10 @@ void dump_state()
 #if !MOBILEAPP
     Admin::dumpMetrics();
 #endif
+
+    std::lock_guard<std::mutex> docBrokerLock(DocBrokersMutex);
```
mmeeks, line 4963: I assume and hope that this lock ordering is consistent everywhere =)

mmeeks (changes requested):
Few cleanups / squashes here needed - but looks good. Thanks Ash =)

common/SigUtil.cpp
```diff
@@ -159,6 +159,68 @@ void requestShutdown()
         UnattendedRun = true;
     }
 
+    const char *signalName(const int signo)
```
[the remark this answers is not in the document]
Ashod: I didn't want to add `#if !MOBILEAPP`. That's the only reason. I really hated moving the large code (and the 8-year-old git blame). So happy to replace it with ifdefs.

wsd/COOLWSD.cpp
Ashod: I maintained the same order, yes. And it seems these signal-triggered functions are the only place we need to take both locks. I didn't revise the logic, though not a fan of multi-locks.

net/HttpRequest.hpp
Ashod: It makes it more readable to have the indents:
```
http::server::Session: #123 
    connected: ...
    startTime: ...
```

Rather than have them all on the same level. It's just a matter of indenting the members of the object vs it's title, for readability (these dumps do get very long and tiring/confusing to read).

Ashod:
Thanks, @mmeeks, for the feedback. Updated per request and added a new patch to include buffer/cache capacities (in addition to sizes), to gauge the memory footprint of these buffers/caches.

net/HttpRequest.hpp
mmeeks: Not a fan of the large number of lines changed; but hey.

mmeeks (approved):
Looks ok - but what are all these horror MOBILEAPP conditionals doing? surely we can have no-op methods inside if we want that return dummy values and will be GC'd out by the linker on mobile - without the conditionals (?)

Otherwise lets get this in then for green then fix it further.

common/SigUtil.cpp
```diff
@@ -232,6 +232,7 @@ void requestShutdown()
         signalLog(buf + i + 1);
     }
 
+#endif // !MOBILEAPP
```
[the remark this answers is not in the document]
Ashod: But that's what moving this function tried to avoid. The function is blocked in `#if !MOBILEAPP`, and we need to call on mobile, so we either move it out, or we exclude it, like I do here.

PR #10408, opened 2024-11-04
wsd: use the 'flush' logging property to disable buffering

Ashod:
In some cases it's benefitial to disable buffered-
logging and flush after each entry is generated.

We already have an unbuffered logger and a
property, called 'flush', already exists and
is used by Poco to also disable buffering its
file-logger.

We leverage this property to disable buffering
in both the console logger (this patch), as well
as in Poco's file-logger (existing functionality).

Change-Id: I5e3b75e66f82a8a5df57901ef4f6c8264aee3565

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -622,7 +622,17 @@ namespace Log
         }
         else
         {
-            channel = static_cast<Poco::Channel*>(new Log::BufferedConsoleChannel());
+            const auto it = config.find("flush");
+            if (it != config.end() && Util::toLower(it->second) != "false")
+            {
+                // Buffered logging, reduces number of write(2) syscalls.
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks fine - but do we document this in the configuration ? =)

mmeeks (approved):
Thanks =)

Ashod:
> Looks fine - but do we document this in the configuration ? =)

Yes, it reads:
> Enable/disable flushing after logging each line. May harm performance. Note that without flushing after each line, the log lines from the different processes will not appear in chronological order.

Though in fairness it's under logging.file, so not the most obvious place to imply it affects both console and file logging. Still, seems less hidden than an envar.

PR #10410, opened 2024-11-04
When forkit installs fonts, sparekits that already exist don't get them

caolanm:
at least the first spare kit created by forkit_main will exist without the fonts which arrive later

Change-Id: Idd778a40791266b0e2ff7ec6ba02dca6a19d0f23

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -1195,6 +1195,14 @@ private:
     std::string _expectedKind;
 };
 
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Not a fan of the NewChildrenMutex I think we should be doing all of that via message-passing to the PrisonerPoll and bin that thing; but ... it fits with what we have I guess.

PR #10412, opened 2024-11-05
net: keep the socket handler instance alive a little longer

vmiklos:
`make -C cypress_test check-desktop spec=calc/annotation_spec.js`
sometimes failed due to coolwsd crashing in the background.

Checking with sanitizers, a memory corruption always happens on the
websrv_poll thread. First we call handleIncomingMessage() on a socket
handler:
    #13 0x5632a3dbf888 in StreamSocket::setHandler(std::shared_ptr<ProtocolHandlerInterface>) /home/vmiklos/git/collaboraonline/online-san/./net/Socket.hpp:1335:24
    #14 0x5632a3d6d44e in AdminSocketHandler::handleInitialRequest(std::weak_ptr<StreamSocket> const&, Poco::Net::HTTPRequest const&) /home/vmiklos/git/collaboraonline/online-san/wsd/Admin.cpp:499:17
    #15 0x5632a4291333 in ClientRequestDispatcher::handleIncomingMessage(SocketDisposition&) /home/vmiklos/git/collaboraonline/online-san/wsd/ClientRequestDispatcher.cpp:797:17
    #16 0x5632a4c9da3e in StreamSocket::handlePoll(SocketDisposition&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l>>>, int) /home/vmiklos/git/collaboraonline/online-san/./net/Socket.hpp:1501:33
Note how StreamSocket::handlePoll() invokes a member function of
_socketHandler, which calls back and resets _socketHandler. Then a
little bit later, still in
ClientRequestDispatcher::handleIncomingMessage() we try to continue
using `this`:
    #0 0x5632a3db1485 in ProtocolHandlerInterface::logPrefix(std::ostream&) const /home/vmiklos/git/collaboraonline/online-san/./net/Socket.hpp:517:66
    #1 0x5632a4298f32 in ClientRequestDispatcher::handleIncomingMessage(SocketDisposition&) /home/vmiklos/git/collaboraonline/online-san/wsd/ClientRequestDispatcher.cpp:969:9
    #2 0x5632a4c9da3e in StreamSocket::handlePoll(SocketDisposition&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l>>>, int) /home/vmiklos/git/collaboraonline/online-san/./net/Socket.hpp:1501:33
And that's a problem, since setHandler() already deleted the
ClientRequestDispatcher by that time.

Fix the problem by keeping the socket handler alive for the duration of
the handleIncomingMessage() call.

All tests in the `make -C cypress_test check-desktop
spec=calc/annotation_spec.js` suite passes with sanitizers after this.

Change-Id: I2360e71f7420ae96486c0a95fff77640c4bef29e

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -1512,6 +1512,9 @@ public:
 
             try
             {
+                // Keep the current handler alive, while the incoming message is handled.
+                std::shared_ptr<ProtocolHandlerInterface> socketHandler(_socketHandler);
+
                 _socketHandler->handleIncomingMessage(disposition);
             }
             catch (const std::exception& exception)


vmiklos:

This is just something I found when trying to see what happens when I try to run a cypress test in the san env of LODE. The picked test is just a random one. Years ago it didn't work at all (somehow chrome didn't even start in that env), and now seems to work mostly, except that we always go to the admin console at the end of tests to see if the doc count drops to 0 (or something like that) and then we hit this heap-use-after-free. It looks minor (if I read correctly, only happens when the log level is debug or higher), but still a memory corruption in the wsd process.

caolanm (approved):

PR #10414, opened 2024-11-06
Dump the Config

Ashod:
- wsd: better hex data formatting in dumpState
- wsd: simplify config loading and capture the path
- wsd: dump the config

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -178,6 +178,7 @@ static std::atomic<int> OutstandingForks(0);
 std::map<std::string, std::shared_ptr<DocumentBroker>> DocBrokers;
 std::mutex DocBrokersMutex;
 static Poco::AutoPtr<Poco::Util::XMLConfiguration> KitXmlConfig;
+static std::string LoggableConfigEntries;
 
 extern "C"
 {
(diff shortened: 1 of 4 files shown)


wsd/COOLWSD.cpp
```diff
@@ -2061,18 +2062,18 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
...
+        { "logging.file.property[3][@name]", "compress" },
         { "logging.file.property[4]", "10 days" },
         { "logging.file.property[4][@name]", "purgeAge" },
         { "logging.file.property[5]", "10" },
         { "logging.file.property[5][@name]", "purgeCount" },
         { "logging.file.property[6]", "true" },
         { "logging.file.property[6][@name]", "rotateOnOpen" },
         { "logging.file.property[7]", "false" },
-        { "logging.file.property[7][@name]", "archive" },
+        { "logging.file.property[7][@name]", "flush" },
```
caolanm, line 2076: ah, this pair of hunks just reindexes these into the order they appear in coolwsd.xml. I was confused because the default I see in coolwsd.xml for archive is "timestamp" not "true" of "false", but it was like that anyway.

caolanm (approved):

PR #10415, opened 2024-11-06
private/ash/misc

Ashod:
- wsd: avoid repeated config lookup
- wsd: test: initialize the config subsystem in unit-tests
- wsd: simplify getServerString()


caolanm (approved):
looks fine to me

timar (approved):
Thanks!

PR #10416, opened 2024-11-06
cool#9992 doc sign: add menu item for the compact UI

vmiklos:
Open Writer, switch to the compact (menu-based) UI, notice the file menu
has no entry to add a signature.

This is primarily interesting as notebookbar is our default, but PDF
files are still opened in compact & read-only mode by default.

Fix the problem by adding a menu entry to all 4 components in the menu
bar, at a position similar to the desktop.

Also the notebookbar button was missing for the Draw case, fix that.

Change-Id: I9af4e42d11f4717a3ab6c5e25b237dd35c407410


vmiklos:

PDF sign is still impossible because PDF is opened in a read-only Draw view, but this at least sorts out all combinations of Writer/Calc/Draw/Impress and notebookbar/compact UI.

caolanm (approved):

PR #10419, opened 2024-11-06
writer: fix: don't show mentionpopup when typing email address

Rash419:
Change-Id: I1833bfdebf09342b25eb88c4e63a7b02288b4c19

* Target version: master


caolanm (approved):
Did we always have this problem, or is it new since the recent work on extending these to comments?

Rash419:
> Did we always have this problem, or is it new since the recent work on extending these to comments?

Already had this problem. It is not regression

PR #10420, opened 2024-11-06
feat(touch): Determine last-used input method

Minion3665:
Previously, when we wanted to determine if some touch UI should be shown we would check if there were any touchscreens on the system. This is still used in some places, but it turns out that this is poor in many cases (a key example would be the cursor droplet, which showed on touchscreen laptops even when the touchscreen was not being used, for example).

You might think that this problem is easily solved: "Just check if the thing that clicked into the document was a touchscreen, and only show the droplet then". We can't, however, trivially detect if the input that caused the cursor to show was from a touchscreen as we roundtrip through core before showing the cursor. This sort of property (hard to get a triggering event) holds true for the other events we were using hasAnyTouchscreen for previously.

Instead, we can listen to *all* mouse or tap events, and change our behavior based on whatever was triggered last. This isn't quite the "triggering event" behavior (e.g. if we close a jsdialog with a touchscreen we will get a droplet cursor, when arguably we might not want that) but it'll be the same a vast majority of the time (and I am undecided on whether this behavior is in-fact preferable in the cases where it differs)

Change-Id: I88f7936495415ad04012fa998b1bd242b3bd8f5b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks @Minion3665 :) I guess only "touchstart", "pointerdown" and "mousedown" event listeners can also be sufficient. no big deal though.

gokaysatir:
> > I guess only "touchstart", "pointerdown" and "mousedown" event listeners can also be sufficient
> 
> I considered this, but I think there would be some weird circumstances if someone tried to use both at the same time which would be solved by listening to the `*up` events too - given that things generally happen on clicks that are directly after the `*up` events. As there's no huge drawback to including the extra few registrations, I opted to do so

I see, thanks for the comment:)

PR #10423, opened 2024-11-06
setWopiFileInfo takes ownership of its arg

caolanm:
make this more clear with a more typical explicit std::move pattern

Change-Id: I8dc7da1af59c6aa09c608d3cbfcd2c9167eab418

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/ClientSession.hpp
+++ b/wsd/ClientSession.hpp
@@ -178,7 +178,7 @@ public:
     }
 
     /// Set WOPI fileinfo object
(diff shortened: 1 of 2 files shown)


caolanm:
I was just looking to test ci with a sample change, but this tripped me up as trying to get something from wopiFileInfo after setWopiFileInfo was called (at the end of this function) and an explicit std::move would have made this less obscure.

mmeeks (approved):
Seems reasonable to me if Ash is happy =)

Ashod (approved):
Thanks, @caolanm. Indeed you can read my intentions better than I can write them!

PR #10424, opened 2024-11-06
Uno: Enable DeleteContentControl command in context menu

codewithvk:
Uno: Enable DeleteContentControl command in context menu

Adds 'Delete Content Control' option to context menu, allowing deletion of content controls without removing their content.

Note: Requires running unocommands.py before release.

Change-Id: I662409f84dc17547cd82c75484c4b06e2eb8c66d

Core changes need:
https://gerrit.libreoffice.org/c/core/+/175927


vmiklos (approved):
Looks good. Nit: please try to wait for CI next time, and only request review once the PR passed already. Life is short, we should first let machines have a go. :-) Thanks.

PR #10431, opened 2024-11-07
cool#9992 doc sign: add UI to sign PDF files

vmiklos:
Open a PDF file, it is loaded into a read-only Draw view, so the File
menu has no signing menu item to sign the document.

There are some menu items, but all of those are actions, and one of
those are inserting a comment: so probably the intent is to avoid not
necessary modifications to the PDF file (like delete the only bitmap on
the draw page), than to actually avoid modifying the document.

Similar to how we have an allowlist for actions, add an allowlist for
UNO commands which are OK to dispatch in read-only mode: just have the
signature command there for now.

In the future, we may want to revisit this to also have a more strict
read-only mode without commenting/signing, but treating signing similar
to commenting is probably good enough for now.

Change-Id: Ib3b396f110cd2d99dbb5d4704f5733bb135f891e

### DIFF
--- a/cypress_test/integration_tests/desktop/writer/file_properties_spec.js
+++ b/cypress_test/integration_tests/desktop/writer/file_properties_spec.js
@@ -13,25 +13,21 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'File Property Tests', func
 
 	it('Add File Description.', function() {
 		writerHelper.openFileProperties();
+		// Too early click on the 'Description' button would have no effect.
+		cy.wait(500);
 		cy.cGet('#tabcontrol-2').click();
 		helper.waitUntilIdle('#title-input.ui-edit');
 		cy.cGet('#title-input.ui-edit').type('New Title');
 		// sometimes it doesn't finish typing
 		helper.waitUntilIdle('#title-input.ui-edit');
 
-		// Fixme: type now char by char because we receive update messages
-		//        can be reverted after core update
-		cy.cGet('#comments.ui-textarea').type('N');
-		cy.wait(500);
-		cy.cGet('#comments.ui-textarea').type('e');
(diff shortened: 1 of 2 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2287/console fails in:

> writer/file_properties_spec.js

but

```
make -C cypress_test check-desktop spec=writer/file_properties_spec.js
```

passes for me locally (failed once, but then passed..), retry.

vmiklos:

With this, interactive PDF signing works for me in COOL. The 2nd commit tries to improve the stability of that cypress test -- I don't see any change to the stability with or without the 1st commit.

cypress_test/integration_tests/desktop/writer/file_properties_spec.js
```diff
@@ -13,25 +13,21 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'File Property Tests', func
 
 	it('Add File Description.', function() {
 		writerHelper.openFileProperties();
+		// Too early click on the 'Description' button would have no effect.
+		cy.wait(500);
 		cy.cGet('#tabcontrol-2').click();
```
caolanm, line 18: I imagine its not as straightforward as the button is disabled until scrolled into place and async enabled just after cypress clicks on it.

caolanm (approved):

cypress_test/integration_tests/desktop/writer/file_properties_spec.js
vmiklos: Hm, possible, I'll keep that in mind next time we run into something like this. I know click() loudly fails when you try to click on something not visible, but possibly it's different for the "visible, but disabled" case.

PR #10433, opened 2024-11-07
Use 10 characters for short git hash

timar:
In release build only 8 character long hash was generated, and it turned out that it was too short. For example the link in About box did not work:
- https://github.com/CollaboraOnline/online/commits/778b9f6 
But a longer hash would have worked:
- https://github.com/CollaboraOnline/online/commits/778b9f619

Change-Id: I5b24d7dcc4ef30acac49b7c8d72f2eca55880aed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -434,10 +434,10 @@ noinst_HEADERS = $(wsd_headers) $(shared_headers) $(kit_headers) \
                  test/helpers.hpp
 
 GIT_BRANCH := $(shell git symbolic-ref --short HEAD)
-GIT_HASH := $(shell git log -1 --format=%h)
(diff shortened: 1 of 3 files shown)


vmiklos (approved):
OK, but I think in the long run we would rather want the full hash, and only turn this to a link in the about dialog that has the first 10 chars in the link text and the link URL would contain the full hash.

Or is there any place where changing from 7 to 10 is OK, but changing to the full hash would break things? (And again, we can still abbreviate to 10 on the UI.)

timar:
@vmiklos Skyler (@Minion3665) also asked this. We use the hash not only in About box, but e.g. in URLs, like /browser/<hash>/<whatever>. See use cases of Util::getCoolVersionHash(). I don't mind, if someone reworks it.

PR #10441, opened 2024-11-07
cool#10440 - perf: don't touch comment's innerText unless it changes.

mmeeks:
Previously CanvasSectionContainer.onMouseWheel - spent 30% of its time setting comment properties to the same value.

Change-Id: I0e50de9bae65db762fd6637618002485514c13b6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -1540,14 +1540,18 @@ export class Comment extends CanvasSectionObject {
 		if (this.sectionProperties.docLayer._docType === 'spreadsheet')
 			return;
 
+		var innerText;
 		if (this.isEdit())
-			this.sectionProperties.collapsedInfoNode.innerText = '!';
+			innerText = '!';
 		else if (replycount === '!' || typeof replycount === "number" && replycount > 0)
-			this.sectionProperties.collapsedInfoNode.innerText = replycount;
+			innerText = replycount;
 		else
-			this.sectionProperties.collapsedInfoNode.innerText = '';
+			innerText = '';
 
-		if (this.sectionProperties.collapsedInfoNode.innerText === '' || this.isContainerVisible())
+		if (this.sectionProperties.collapsedInfoNode.innerText != innerText)
+			this.sectionProperties.collapsedInfoNode.innerText = innerText;
+
+		if (innerText === '' || this.isContainerVisible())
 			this.sectionProperties.collapsedInfoNode.style.display = 'none';
 		else if ((!this.isContainerVisible() && this.sectionProperties.collapsedInfoNode.innerText !== ''))
 			this.sectionProperties.collapsedInfoNode.style.display = '';


mmeeks:
@timar if some sensible reviewer things it's safe might be useful on the cool branch =)

gokaysatir (approved):
Thanks :)

browser/src/canvas/sections/CommentSection.ts
```diff
@@ -1540,14 +1540,18 @@ export class Comment extends CanvasSectionObject {
...
-			this.sectionProperties.collapsedInfoNode.innerText = '';
+			innerText = '';
 
-		if (this.sectionProperties.collapsedInfoNode.innerText === '' || this.isContainerVisible())
+		if (this.sectionProperties.collapsedInfoNode.innerText != innerText)
+			this.sectionProperties.collapsedInfoNode.innerText = innerText;
+
+		if (innerText === '' || this.isContainerVisible())
 			this.sectionProperties.collapsedInfoNode.style.display = 'none';
 		else if ((!this.isContainerVisible() && this.sectionProperties.collapsedInfoNode.innerText !== ''))
```
Ashod, line 1556: Extra parens can be removed. But better to remove the (now redundant) conditional altogether.

browser/src/canvas/sections/CommentSection.ts
```diff
@@ -1540,14 +1540,18 @@ export class Comment extends CanvasSectionObject {
...
+			innerText = replycount;
 		else
-			this.sectionProperties.collapsedInfoNode.innerText = '';
+			innerText = '';
 
-		if (this.sectionProperties.collapsedInfoNode.innerText === '' || this.isContainerVisible())
+		if (this.sectionProperties.collapsedInfoNode.innerText != innerText)
+			this.sectionProperties.collapsedInfoNode.innerText = innerText;
+
+		if (innerText === '' || this.isContainerVisible())
```
Ashod, line 1554: I suppose we don't need to check this twice; this check is the logical inverse of the `else` case below. Best to make the one on line 1556 an unconditional `else`.

browser/src/canvas/sections/CommentSection.ts
```diff
@@ -1540,14 +1540,18 @@ export class Comment extends CanvasSectionObject {
...
+			innerText = '!';
 		else if (replycount === '!' || typeof replycount === "number" && replycount > 0)
-			this.sectionProperties.collapsedInfoNode.innerText = replycount;
+			innerText = replycount;
 		else
-			this.sectionProperties.collapsedInfoNode.innerText = '';
+			innerText = '';
 
-		if (this.sectionProperties.collapsedInfoNode.innerText === '' || this.isContainerVisible())
+		if (this.sectionProperties.collapsedInfoNode.innerText != innerText)
```
Ashod, line 1551: Minor point: comparisons with the `innerText` in this function are without coercion (i.e. `===` and `!==`). Shouldn't this too do the same?

Incidentally, `===` and `!==` are slightly faster than `==` and `!=` (for the obvious reason that type coercion needs to be considered/tried for the latter, where the former will short-circuit if the types aren't the same).

Ashod (approved):
Thanks, @mmeeks. Looks good. Left some suggestions.

browser/src/canvas/sections/CommentSection.ts
mmeeks: sure - it can be improved; but I left what was there there =)

browser/src/canvas/sections/CommentSection.ts
mmeeks: sure - follow-on cleanup possible; again not changing existing logic.

browser/src/canvas/sections/CommentSection.ts
mmeeks: Sounds reasonable - but doesn't work: we inject numbers into innerText which are then converted to strings internally, and when we get them back we get to do a sloppy compare here I think. Either way - it's a big win for rather a minimal change.

If you'd like to do a cleanup, and re-validate, feel free but ... =)

PR #10444, opened 2024-11-07
Hiding server signature

timar:
We were advised to hide not only version number but the 'COOLWSD HTTP Server' string, too.

Change-Id: I5f737405c4f44c18cc033ef634ce363e5c9a74ea

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -1742,7 +1742,7 @@ std::string getServerString()
     if (sig)
         return "COOLWSD HTTP Server " + Util::getCoolVersion();
 
-    return "COOLWSD HTTP Server";
+    return " ";
 }
 }
 


net/Socket.cpp
```diff
@@ -1742,7 +1742,7 @@ std::string getServerString()
     if (sig)
         return "COOLWSD HTTP Server " + Util::getCoolVersion();
 
-    return "COOLWSD HTTP Server";
+    return " ";
```
Ashod, line 1745: Intentional to return a single space? Why not a blank string?

timar: The reporter hinted at Apache's `SecServerSignature " "` setting. I was not sure, whether an empty string would break something.

PR #10447, opened 2024-11-08
kit: fix div by zero in ChildSession::renderWindow()

vmiklos:
Running 'make -C cypress_test check-desktop' in a san environment and
checking cypress_test/cypress/wsd_logs/coolwsd_output.log after the run
contains:

kit/ChildSession.cpp:2482:5: runtime error: division by zero

If the rendering is fast enough that we round the cost of it to 0, then
avoid the division by zero.

This typically doesn't affect production setups, since the log is only
for trace level.

Change-Id: I19285a8b21f5b5594d7e9a2df6ae19e2c1e43f8a

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -2482,7 +2482,7 @@ bool ChildSession::renderWindow(const StringVector& tokens)
     const double elapsedMics = elapsedMs.count() * 1000.; // Need MPixels/second, use Pixels/mics.
     LOG_TRC("paintWindow for " << winId << " returned " << width << 'X' << height << "@(" << startX
                                << ',' << startY << ',' << " with dpi scale: " << dpiScale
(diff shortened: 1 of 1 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2323/console fails in:

> calc/searchbar_spec.js
> writer/searchbar_spec.js

retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2324/console fails in:

> calc/sheet_switch_spec.js
> writer/searchbar_spec.js

hmm.

vmiklos:
@caolanm this is just to clear out an obvious problem, so at least the desktop part of the cypress suite is free from san errors. Some tests still fail, but that's probably about timing.

caolanm (approved):

Issue #10442, opened 2024-11-07
perf: VRuler._fixOffset - should be more efficient.


mmeeks:
Quick profile on scrolling up and down in writer in our sample document using the touch-pad.

![image](https://github.com/user-attachments/assets/99f1b59d-041d-468b-b6a0-3dc4784a56de)

some somewhat amazing things: in browser/src/control/VRuler.ts

It seems that we do a lot of work here even when the vertical ruler is not showing (as it is not for me by default in writer ;-) - so:

* lets not do expensive things for invisible vrulers.

* I believe we shouldn't be doing: 		const computedStyle = window.getComputedStyle(mapPane); - but @Cwiiis is the expert - since it forces synchronous layout (or something) - and surely that's not necessary anyway to render a vertical ruler (?) =)

Seems this is another between 1/3rd and 2/3rds (oddly it differs) of the time it takes to process a scroll wheel event - we should fix that !

mmeeks:
@Darshan-upadhyay1110 might be worth fixing this yourself for back-port to a release branch if its easy - quite a chunk of cost there :-) and a nice win.

PR #10450, opened 2024-11-08
Optimize VRuler performance by avoiding layout calculations when hidden

Darshan-upadhyay1110:
- Avoid calculating offset if the vertical ruler is hidden.
 - Skip updating paragraph indentation when the ruler is not visible.
 - Optimize performance by directly extracting Y translation value from mapPane style

Change-Id: I1953d680b6e132c1d48f9b435fde8e06735ed48d

* Resolves: #10442
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/VRuler.ts
+++ b/browser/src/control/VRuler.ts
@@ -433,12 +433,19 @@ class VRuler {
 		const firstTileXTranslate = topLeft.y;
 
 		let mapPaneYTranslate: number = 0;
-		const computedStyle = window.getComputedStyle(mapPane);
-		const transformValue = computedStyle.getPropertyValue('transform');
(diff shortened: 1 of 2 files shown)


browser/src/control/VRuler.ts
```diff
@@ -231,6 +232,19 @@ class VRuler {
 		return this._rWrapper;
 	}
 
+	_onRulerChanged() {
+		// update show ruler state on rulerChange event
+		this.options.showruler = this._map.uiManager.getBooleanDocTypePref(
```
eszkadev, line 237: this method looks exactly like in the horizontal ruler -> could be shared
we need to remember about it when converting to TS -> I see no point doing base class in Leaflet style now...

eszkadev (approved):
looks ok

PR #10451, opened 2024-11-08
fix(android): Fix progress bar with b64d messages

Minion3665:
In days gone by, every message was sent to the Android/iOS fake websocket as unencoded text. Unfortunately, some of the text was untrusted and this allowed XSS attacks to abuse sent messages like so...

message: 'command: {untrustedContent}' where {untrustedContent} is ' + maliciousJS() + ' would lead to the message executing maliciousJS()

...to prevent this, we encoded all messages in base64, and decoded them on the JavaScript side, leading to messages looking like this:

message: b64d('Y29tbWFuZDogJyArIG1hbGljaW91c0pTKCkgKyAn')

This appeared to work great - and in-fact did work great most of the time! Unfortunately, we failed to realize that sometimes we intercepted these messages in native Android code before sending them on to online...

...A specific case where we intercepted them was the progress bar messages, which looked something like this:

message: 'progress: { "id": "setvalue", "value": 15 }'

We detected the message by looking for the "'progress" text at the start of the message, and parsed out the JSON to get the value

after base64 encoding everything, these messages looked like this instead...

message: b64d('cHJvZ3Jlc3M6IHsgImlkIjogInNldHZhbHVlIiwgInZhbHVlIjogMTUgfQ==')

...which completely broke our detection.

...This wasn't noticed at the time as there were also various other issues with the progress bar, which made it not properly work before this regression for separate reasons...

This patch adds some functions to detect the start of messages despite base64, and then decode the message so we can parse out the JSON as normal

I've also standardized on using b64d for messages that come directly from Java, which we probably should've been doing anyway ... there's no immediate technical need to do this, but the new functions will not work with anything still using atob. The reason atob and b64d are different is because b64d supports unicode, so we can't use the built-in atob everywhere.

Change-Id: I8983d6f89aee417124d8124f0495780f0fc498f1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
+++ b/android/lib/src/main/java/org/libreoffice/androidlib/LOActivity.java
@@ -964,7 +964,7 @@ public class LOActivity extends AppCompatActivity {
     void callFakeWebsocketOnMessage(final String message) {
         String base64Message = Base64.getEncoder().encodeToString(message.getBytes());
 
-        rawCallFakeWebsocketOnMessage("window.atob('" + base64Message + "')");
+        rawCallFakeWebsocketOnMessage("b64d('" + base64Message + "')");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


caolanm (approved):
looks fine to me

PR #10452, opened 2024-11-08
Don't update the render state if the payload is empty.

gokaysatir:
When switching between views, new pages may be created and old ones may get deleted. In this case, some pointers are null. Check "DrawViewShell not available!" warning on the core side. This null pointer causes render state payload to be empty. Empty payload is not correct value for client side renderState parameter.

Change-Id: I08af6208807e0dc38538b85539dd992bb166695c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1179,8 +1179,11 @@ void Document::trimAfterInactivity()
             std::shared_ptr<ChildSession> session = document->findSessionByViewId(descriptor->getViewId());
             if (session)
             {
-                session->setViewRenderState(payload);
-                document->invalidateCanonicalId(session->getId());
+                if (!payload.empty())
+                {
+                    session->setViewRenderState(payload);
+                    document->invalidateCanonicalId(session->getId());
+                }
             }
             else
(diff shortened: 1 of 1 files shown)


caolanm:
Is there a backtrace of where that warning comes from in core when this happens? I wonder if core (impress) should not send such an empty LOK_CALLBACK_VIEW_RENDER_STATE but I guess that might depend on if core sent this unprompted, or if it was asked to send one and this was the best it could do.

gokaysatir:
> Is there a backtrace of where that warning comes from in core when this happens? I wonder if core (impress) should not send such an empty LOK_CALLBACK_VIEW_RENDER_STATE but I guess that might depend on if core sent this unprompted, or if it was asked to send one and this was the best it could do.

Yes, when switching to / from notes view, that warning should be seen.

caolanm (approved):
I think this is safe, looking at where we generate the string, there are no legal empty string cases. Still be nice a get a backtrace example of where one of these gets emitted.

PR #10453, opened 2024-11-08
browser: do not update split panes with outdated SheetGeometry

JaumePujantell:
We shouldn't update the split panes position if we still have an
outdated sheet geometry data form another sheet.

Change-Id: Iadbf6c5c6cd1f5188530943ba2ac97dafe91a0fc


vmiklos (approved):
Looks OK in general, fixes the mentioned use-case.

Nits:

- it's ideal to first wait for CI and only then ask for review, machine time is unlimited, our time is limited :-) also you may need to rebase and there was some back & forth with the codeql checks
- did you consider what happens when the sheet geometry for the current sheet arrives later? who will call setSplitPosFromCell() in that case, now that the sheet geometry's part and the selected part matches?

JaumePujantell:
> it's ideal to first wait for CI
Yes, sorry, I misunderstood the 'Expected — Waiting for status to be reported ' as it'll be ok soon, only later I saw the whole back and forth about codeql

> what happens when the sheet geometry for the current sheet arrives later?
setSplitPosFromCell is always called when the sheet geometry updates (browser/src/layer/tile/CalcTileLayer.js:757)

PR #10457, opened 2024-11-10
Misc small fixes to docker build from source

timar:
* Target version: master 

- docker: updated docker/from-source/README.md
- docker: inotifytools and procps are not dependencies
- docker: use theme names instead of magic numbers
- docker: install sass, if it's not installed
- docker: when building from source, --disable-tests saves build time
- docker: git clone --depth=1 saves some time, disk space and bandwidth
- docker: use poco version 1.12.5p2 (the same version is used in production)
- docker: wrong default, distro/collabora/co-24.05 never existed
- docker: we don't have to check for sudo without asking a password


caolanm (approved):
looks good to me

juliusknorr (approved):

PR #10463, opened 2024-11-11
Improve impress templates

Rash419:
* Resolves: # 
* Target version: master 

### TODO

- [x] tested using https://github.com/Rash419/RemoteConfigSever


kit/Kit.cpp
```diff
@@ -3311,6 +3315,34 @@ void lokit_main(
                     return false;
                 }
 
+                // copy default tempates from 'common' dir to shared templates dir
+                // TODO: maybe I shouldn't copy if whole point to mounting is that we don't require copying.
```
Rash419, line 3319: I am not sure. How can we improve here except of copying default templates ?

kit/Kit.cpp
```diff
@@ -3311,6 +3315,34 @@ void lokit_main(
...
+                        LOG_WRN("Failed to copy default impress template from ["
+                                << sourcePath << "] to [" << sharedTemplate << ']');
+                    }
+                }
+
+                // mount the shared templates over the lo shared templates' 'common' dir
+                if (!JailUtil::bind(sharedTemplate, loJailDestImpressTemplatePath) ||
+                    !JailUtil::remountReadonly(sharedTemplate, loJailDestImpressTemplatePath))
+                {
+                    // TODO: actually do this link on failure
```
Rash419, line 3340: If I disable mounting default templates templates doesn't show up at all.

kit/Kit.cpp
```diff
@@ -3313,6 +3317,37 @@ void lokit_main(
                     return false;
                 }
 
+                if (!FileUtil::isEmptyDirectory(loJailDestImpressTemplatePath)) // make sure template directory exist
```
caolanm, line 3320: Maybe you don't need this "isEmptyDirectory" condition at all if https://gerrit.libreoffice.org/c/core/+/177353 is merged ?

But if you do, maybe it should be something like:

const FileUtil::Stat templateStat(loJailDestImpressTemplatePath);
if (templateStat.good() && templateStat.isDirectory())  // make sure template directory exist

Because if the directory exists, but it is empty, then isEmptyDirectory is true and this code isn't run, but an empty target dir here is perfectly ok as long as it exists.

After that (with my core patch above applied), if I then remove the explicit copy of "copy default tempates from 'common' dir" I think it works ok without it.

The mount namespaces can get a bit confusing, but FWIW given the pid of the kit, e.g. from a log message like: kit-2421928-2421928 2024-11-26 17:08:44.948975 +0000 [ kit_spare_002 ] DBG  Mounting...

you can:

$ sudo nsenter -m -t 2421928

and then mount and ls will report what pid 2421928 sees.

So for me mount reports:

/dev/nvme0n1p3 on /home/caolan/LibreOffice/collabora-online/jails/2421849-b13c8f67/5pFZQab2t34k3MK1/lo/share/template/common/presnt ...

so I can see something got mounted on the lo/share/template/common/presnt dir and

ls /home/caolan/LibreOffice/collabora-online/jails/2421849-b13c8f67/5pFZQab2t34k3MK1/lo/share/template/common/presnt

lists

template1.otp  template2.otp  template3.otp

So that looked like it worked.

We do want this "copy" fallback though for the case that bindMounts fail, I think that should go further down after the "linkOrCopy(loTemplate, loJailDestPath" lines which is where we copy the libreoffice install when we can't mount so this could go after that.

kit/Kit.cpp
```diff
@@ -360,16 +360,12 @@ namespace
...
-                    strstr(path, "libssl3") ||
-                    strstr(path, "libsoftokn3") ||
-                    strstr(path, "libsqlite3") ||
-                    strstr(path, "libfreeblpriv3"))
+                if (strstr(path, "libnspr4") || strstr(path, "libplds4") ||
+                    strstr(path, "libplc4") || strstr(path, "libnss3") ||
+                    strstr(path, "libnssckbi") || strstr(path, "libnsutil3") ||
+                    strstr(path, "libssl3") || strstr(path, "libsoftokn3") ||
+                    strstr(path, "libsqlite3") || strstr(path, "libfreeblpriv3") ||
+                    strstr(path, "libucphier1"))
```
caolanm, line 368: This bit strikes me as suspicious, because everything else listed here is belongs to nss (or nss links to it) and we don't list any other of LibreOffice's own shared libraries here.

IIUC we "preload" a bunch of libraries and I suspect that this has a relation so the problem you are seeing. Looking at core I see a twisty set of things in lo_initialize in desktop/source/lib/init.cxx but from there then in "cppuhelper/source/servicemanager.cxx" I see:

/// Allow external callers & testers to disable certain components
const char *pDisable = getenv("UNODISABLELIBRARY");

and I see in kit/Kit.cpp we have:

// Disable problematic components that may be present from a
// desktop or developer's install if env. var not set.
::setenv("UNODISABLELIBRARY",
         ... ucpgio1 ucphier1 ucpimage ...
             0 /* no overwrite */);

Which I think is intended to disable libraries that might happen to be in a developers local build of core, that don't exist in the real/release configuration core build. So... I wonder if removing "ucphier1" from that other list instead of adding libucphier1 to this list also fixes the problem?

If that also works, I expect that is the preferred solution. (And, assuming it works, worth a separate pr which we can run by Michael to see if ucphier1 is just accidentally added in there or if there is a reason I'm unaware of why that is explicitly disabled in that UNODISABLELIBRARY list)

Rash419: Yes it works. I have created https://github.com/CollaboraOnline/online/pull/10618

Rash419:
> Am I right that "cypress: attempt to fix "Delete Chart" test" is for an existing test failure that isn't a new problem in this pr? Maybe we should get that in a separate pr that we could merge first?

Removed the cypress commit. This test were failing due to linking issue : )

Rash419:
@caolanm can you approve we are in red cycle now ?

PR #10464, opened 2024-11-11
browser: slideshow: add effect down/up button

hcvcastro:
Change-Id: I6f9f5a21bfa7d76e8d8f86d8c0b2419b722a34b8

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -327,6 +327,8 @@ class PresenterConsole {
 		elem.style.bottom = 0;
 		elem.style.width = '100%';
 		elem.addEventListener('click', L.bind(this._onToolbarClick, this));
+		elem.addEventListener('mousedown', L.bind(this._onToolbarMouseDown, this));
+		elem.addEventListener('mouseup', L.bind(this._onToolbarMouseUp, this));
 
 		let list = this._proxyPresenter.document.querySelectorAll('button');
 		for (elem = 0; elem < list.length; elem++) {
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/PresenterConsole.js
```diff
@@ -382,18 +384,34 @@ class PresenterConsole {
...
+
+		return target;
+	}
+
+	_onToolbarMouseDown(e) {
+		let target = this._getButton(e.target);
 		if (!target) {
 			return;
 		}
+		target.style.transform = 'translateY(5px)';
```
Darshan-upadhyay1110, line 400: Hello @hcvcastro code looks good but can we add a method for adding transform animation ?

Like a method `xyz(Element , add/remove:boolean)` this can help to add animation to other elements as well without repeating the code

Any thoughts ?

Darshan-upadhyay1110: Oh i see it works for all toolbar button. Please ignore my previous comment.

Darshan-upadhyay1110 (commented):
@hcvcastro can we not add animation and just change background color like same we do in our main document page?

I will be better if we do same to make consistent

For ex:
[Screencast from 2024-11-13 15-14-37.webm](https://github.com/user-attachments/assets/2d5ac0dd-7918-416b-81bc-036de1d6d42b)

CC: @pedropintosilva thoughts

hcvcastro:
oh, no problem, I just added a simple effect, Please assign the task yourself I see you have design ui improvements to do =)

PR #10472, opened 2024-11-12
fix(keyboard): Allow using Ctrl+Shift+Z for redo

Minion3665:
I found this on an Apple device, as Cmd+Z and Cmd+Shift+Z were both undoing. On every other app I could find, Cmd+Shift+Z was bound to redo, however here we hadn't set it. Without setting this shortcut, it instead also activates undo, causing trouble for people who are used to using it as redo.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -734,14 +734,17 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
-		if (this._isCtrlKey(e) && e.keyCode === this.keyCodes.Z) {
-			app.socket.sendMessage('uno .uno:Undo');
+		if (
+			(this._isCtrlKey(e) && e.keyCode === this.keyCodes.Y) ||
+			(this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.Z)
+		) {
+			app.socket.sendMessage('uno .uno:Redo');
 			e.preventDefault();
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


Minion3665:
GitHub's diff view is kinda grim on this one, here's difftastic which is (imo) easier to understand
image

eszkadev:
we have this new flexible shortcuts table: https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts#L161

Would it be possible to move this hander there? or it has some additional requirements?
We can call uno command or create `dispatch` command to trigger on shortcut.

eszkadev:
> or it has some additional requirements? 

Asking because I see the method has comment: `// Given a DOM keyboard event that happened while the Control key was depressed,`

Minion3665:
> we have this new flexible shortcuts table: [`master`/browser/src/map/handler/Map.KeyboardShortcuts.ts#L161](https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts?rgh-link-date=2024-11-14T14%3A11%3A51Z#L161)
> 
> Would it be possible to move this hander there? or it has some additional requirements? We can call uno command or create `dispatch` command to trigger on shortcut.

> Asking because I see the method has comment: `// Given a DOM keyboard event that happened while the Control key was depressed,`

I don't know of any additional requirements, but I'm not super experienced in the clipboard shortcuts code so I might not expect to. If we're going to change this, it's certainly worth following up on that comment to avoid Chesterton's fencing ourselves

eszkadev (approved):

Issue #8205, opened 2024-02-06
Menu key should trigger right click menu


pedropintosilva:
### Describe the Bug
Menu key does nothing when pressed. Here is what that key might look like in your keyboard (it can also just say "Menu"):

![image](https://github.com/CollaboraOnline/online/assets/65948705/42e60ef1-2596-43a2-91db-183bc542eb24)

### Steps to Reproduce
1. Open a document
2. Select a piece of text
3. Press the menu key on your keyboard
4. Nothing happens

### Expected Behavior
To trigger the right-click menu

### Actual Behavior
Does nothing

PR #9825, opened 2024-08-15
Enable context menu key

NickWingate:
* Resolves: #8205 
* Target version: master 

### Summary
Catch and pass context menu key to core (respective core patch: https://gerrit.libreoffice.org/c/core/+/171914)
Once rendered context menu, focus it so we can continue navigating it with the keyboard

### TODO

- [ ] ...

### Checklist

- [ ] I have run `make prettier-write` and formatted the code.
- [ ] All commits have Change-Id
- [ ] I have run tests with `make check`
- [ ] I have issued `make run` and manually verified that everything looks okay
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #10473, opened 2024-11-12
Enable context menu key

codewithvk:
* Resolves: https://github.com/CollaboraOnline/online/issues/8205
* Target version: master 

Cherry-pick two commits from: https://github.com/CollaboraOnline/online/pull/9825

Core side patch : https://gerrit.libreoffice.org/c/core/+/171914

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -632,6 +632,9 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 }
 
 /* context menu */
+.context-menu-root:focus {
+	outline: none;
+}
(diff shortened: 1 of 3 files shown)


eszkadev:
Thanks for improvement, at first look seems to work. I noticed that we always trigger context menu for the "CELL" in calc, even if we selected part of the text inside EDIT mode. Might be good improvment in the followup.

[contextmenu-2024-11-13_11.04.24.webm](https://github.com/user-attachments/assets/2593d792-e22c-4794-8abc-22b78a9a98ac)

eszkadev (approved):

pedropintosilva (approved):
Thanks!

PR #10478, opened 2024-11-13
kit: don't remove original file after save

vmiklos:
```
The bugreport was about an ODS file which was edited to have a model
where ODS export fails, but XLS(X) export works. No steps to reliably
reproduce the problem are known at the moment.

Investigating that problem, it turns out that we had logic in place, so
that once we get the UNO command result for an .uno:Save command, we
rename the result to "<original path>.upload". This has the benefit that
in case we would start uploading to storage but an other save starts in
the meantime, then we never upload a half-finished export result. The
downside is that some code in core.git expects the original file to be
around while editing, e.g. (at least) PDF signing won't work without it.
It's also hard to audit all save codepath to be sure that the lack of
original path doesn't break other corner-cases.

Fix the problem by still uploading "<original path>.upload" instead of
"<orignal path>", but copy the content to the new path, don't rename.
```

Change-Id: I1110580981f64e9cadf6e075861d4f83ef38d119

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -863,21 +863,21 @@ namespace {
      * DocBroker will have to decide to upload or skip.
      */
     [[maybe_unused]]
-    void renameForUpload(const std::string& url)
+    void copyForUpload(const std::string& url)
     {
         const std::string oldName = Poco::URI(url).getPath();
         const std::string newName = oldName + TO_UPLOAD_SUFFIX;
-        if (rename(oldName.c_str(), newName.c_str()) < 0)
+        if (!FileUtil::copyAtomic(oldName, newName, /*preserveTimestamps=*/true))
         {
-            // It's not an error if there was no file to rename, when the document isn't modified.
+            // It's not an error if there was no file to copy, when the document isn't modified.
             const auto onrre = errno;
-            LOG_TRC("Failed to rename [" << oldName << "] to [" << newName << "] ("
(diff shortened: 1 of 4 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2476/console fails in:

> calc/searchbar_spec.js

but that passed for the parent commit, so doesn't look like a real failure, retry.

vmiklos:
@caolanm could you please review this?

The original bugreport is still lacking repro steps, but this "open file, then delete it while core has it opened" is very unusual from the core code point of view, so is related. By accident, there is a signing-related use-case here: if you open a pdf, add a comment, save it, then without this PR you can't sign it, as signing needs the original file in place. And who knows what else depends on this.

caolanm (approved):
In https://gerrit.libreoffice.org/c/core/+/165769 for file properties we ended up sending an additional filesize arg because of this renaming issue. I was very reluctant to change the move/rename to a copy in case there is some subtlety that depends on the original not existing anymore and it was clearly quite an old decision to rename.

But I don't have a known argument against copying it if that also works.

vmiklos:
Oh well, this feels less risky than the anyInput() event reordering at least. :-) And indeed, perhaps let's give this some time, but if we find no problems, then the above core change can be also cleaned up. Thanks for pointing that out.

PR #10479, opened 2024-11-13
Generate CanvasTileWorker.js in DIST_FOLDER, not srcdir

Cwiiis:
Change-Id: I336eadd5f890acf4f42fd1419d2330ce2d2a609c


browser/Makefile.am
```diff
@@ -1111,7 +1110,7 @@ pot:
 		src/app/DocEvents.ts \
 		src/app/GraphicSelectionMiddleware.ts \
 		src/errormessages.js \
-		src/layer/tile/CanvasTileWorker.js \
+		src/layer/tile/CanvasTileWorker.js.m4 \
```
timar, line 1113: We do not need neither src/layer/tile/CanvasTileWorker.js.m4 nor src/layer/tile/CanvasTileWorker.js in the pot section, because they do not contain localizable strings.

timar (changes requested):

caolanm (approved):

timar (approved):

PR #10482, opened 2024-11-13
Replace TileLocation with TileDescription.

gokaysatir:
Two structures are very similar.

Also replaced "throw" error with "log_err" or behchmark.cpp would complain about it.

Change-Id: I411a0b7f7763578ccdc9cee6fd35c8cec59a89ec

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/RenderTiles.hpp
+++ b/common/RenderTiles.hpp
@@ -71,7 +71,7 @@ namespace RenderTiles
                                  size_t pixmapHeight, int pixelWidth, int pixelHeight,
                                  LibreOfficeKitTileMode mode)>& blendWatermark,
         const std::function<void(const char* buffer, size_t length)>& outputMessage,
-        [[maybe_unused]] unsigned mobileAppDocId, int canonicalViewId, bool dumpTiles)
+        [[maybe_unused]] unsigned mobileAppDocId, bool dumpTiles)
     {
         const auto& tiles = tileCombined.getTiles();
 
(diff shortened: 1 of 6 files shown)


kit/Delta.hpp
```diff
@@ -467,7 +433,7 @@ class DeltaGenerator {
             assert(wasInUse && "Error: delta was already un-used by another thread");
         }
 
-        TileLocation _loc;
+        TileDesc _loc;
     private:
         std::atomic<bool> _inUse; // thread debugging check.
         TileWireId _wid;
```
mmeeks, line 439: If we go to the expense and memory weight of having a full TileDesc - then we should at least try to re-use the WireId in there surely ? Then again _id - is included into the == comparison operator we use below for DeltaCompare - so probably we do have to waste that.

common/RenderTiles.hpp
```diff
@@ -191,14 +191,7 @@ namespace RenderTiles
...
                                                      pixmapWidth, pixmapHeight,
-                                                     TileLocation(
-                                                         tileRect.getLeft(),
-                                                         tileRect.getTop(),
-                                                         tileRect.getWidth(),
-                                                         tileCombined.getPart(),
-                                                         canonicalViewId,
-                                                         tileCombined.getEditMode()
-                                                         ),
+                                                     tiles[tileIndex],
```
mmeeks, line 194: This looks very dangerous to me. TileDesc may easily have a wire-id set here in the desc._id. Most likely it will be different to the last usage.
In this case - we not only use more memory to store this - but we also completely disable the Delta cache, Deltas - burn client bandwidth and worse.

I would like to see a unit test with wire-ids set to different numbers to ensure this does not happen if we do this.

I would want to see a custom comparison function here.

Or - better, we could split the TileLocation concept out of TileDesc - and whack it in a base-class of that - and just continue to use TileLocation :-)

mmeeks (changes requested):
Not convinced.

gokaysatir:
@mmeeks thanks for checking :) Let me consider this refactoring later.

PR #10483, opened 2024-11-13
Fix combined keyframe+delta tile updates

Cwiiis:
### Summary

I'd made the bad assumption that keyframe and delta tile updates always arrived separately, but this is very much not the case, especially outside of Writer apparently. This change fixes that assumption and un-breaks the very obvious breakage in Impress when editing.

### Checklist


caolanm (approved):

Cwiiis:
> As a reminder to myself, this change I think now makes separate keyframe/delta storage pointless, so I'll tidy that up tomorrow morning.

This is #10489

PR #10484, opened 2024-11-13
feat: unify mobile clipboard with online

Minion3665:
- There is a dependency in the iOS parts with https://gerrit.libreoffice.org/q/topic:%22private/skyler/ios-clipboard%22
- On Android specifically, this introduces several regressions, of which I have found the following
	- This pastes formulas by value, rather than by reference
	- This cannot copy images or graphics

	On iOS these are not regressions, not because they don't exist but because copy/paste is already so broken that it doesn't matter. This is strictly an improvement.

---

I'm going to spend some time investigating these regressions, but if I'm not able to fix them within a timely fashion then I think the solution will be to split I9352a4473be7ab38acc6187a93a5b63e22208815 into iOS/Android copies, and then merge the following in a separate PR
- I64a268c1bdb1412ea2002882389fd49a96bc97d8
- Ia8bbe65f4cd4b31b42f2a8bb24cefe668196dedc
- I40e65150bd41ea99f42c61f75118f3e1aecb3220
- I9352a4473be7ab38acc6187a93a5b63e22208815 (iOS half of split only)

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -439,6 +439,19 @@ L.Clipboard = L.Class.extend({
 		return true;
 	},
 
+	_sendToInternalClipboard: async function (content) {
+		if (window.ThisIsTheiOSApp) {
+			await window.webkit.messageHandlers.clipboard.postMessage(`sendToInternal ${await content.text()}`); // no need to base64 in this direction...
+		} else {
+			var formData = new FormData();
+			formData.append('file', content);
+
+			return await this._doAsyncDownload('POST', this.getMetaURL(), formData, false,
+				function(progress) { return progress; }
+			);
+		}
+	},
+
 	dataTransferToDocumentFallback: async function(dataTransfer, htmlText, usePasteKeyEvent) {
 		var content;
 		if (dataTransfer) {
@@ -484,21 +497,18 @@ L.Clipboard = L.Class.extend({
 			return;
 		}
 
-		if (content != null) {
-			window.app.console.log('Normal HTML, so smart paste not possible');
-
-			var formData = new FormData();
-			formData.append('file', content);
-
-			var that = this;
-			await this._doAsyncDownload('POST', this.getMetaURL(), formData, false,
-				function(progress) { return progress; }
-			);
-			window.app.console.log('Posted ' + content.size + ' bytes successfully');
-			that._doInternalPaste(that._map, usePasteKeyEvent);
-		} else {
+		if (content == null) {
 			window.app.console.log('Nothing we can paste on the clipboard');
+			return;
 		}
+
+		window.app.console.log('Normal HTML, so smart paste not possible');
+
+		await this._sendToInternalClipboard(content);
+
+		window.app.console.log('clipboard: Sent ' + content.size + ' bytes successfully');
+
+		this._doInternalPaste(this._map, usePasteKeyEvent);
 	},
 
 	_checkSelection: function() {
@@ -698,7 +708,8 @@ L.Clipboard = L.Class.extend({
 			return;
 		}
 
-		if (document.execCommand(operation) &&
+		if (!window.ThisIsTheiOSApp && // in mobile apps, we want to drop straight to navigatorClipboardRead as execCommand will require user interaction...
+			document.execCommand(operation) &&
 			serial !== this._clipboardSerial) {
 			window.app.console.log('copied successfully');
 			this._unoCommandForCopyCutPaste = null;
@@ -853,7 +864,7 @@ L.Clipboard = L.Class.extend({
 
 	// Executes the navigator.clipboard.write() call, if it's available.
 	_navigatorClipboardWrite: function() {
-		if (!L.Browser.clipboardApiAvailable) {
+		if (!L.Browser.clipboardApiAvailable && !window.ThisIsTheiOSApp) {
 			return false;
 		}
 
@@ -874,33 +885,37 @@ L.Clipboard = L.Class.extend({
 		// that command so we are sure the clipboard is set before
 		// fetching it.
 
-		const url = this.getMetaURL() + '&MimeType=text/html,text/plain;charset=utf-8';
+		if (window.ThisIsTheiOSApp) {
+			await window.webkit.messageHandlers.clipboard.postMessage(`write`);
+		} else {
+			const url = this.getMetaURL() + '&MimeType=text/html,text/plain;charset=utf-8';
 
-		var result = await fetch(url);
-		var text = await result.text();
+			var result = await fetch(url);
+			var text = await result.text();
 
-		const clipboardItem = new ClipboardItem({
-			'text/html': this._parseClipboardFetchResult(text, 'text/html', 'html'),
-			'text/plain': this._parseClipboardFetchResult(text, 'text/plain', 'plain')
-		});
-		let clipboard = navigator.clipboard;
-		if (L.Browser.cypressTest) {
-			clipboard = this._dummyClipboard;
-		}
+			const clipboardItem = new ClipboardItem({
+				'text/html': this._parseClipboardFetchResult(text, 'text/html', 'html'),
+				'text/plain': this._parseClipboardFetchResult(text, 'text/plain', 'plain')
+			});
+			let clipboard = navigator.clipboard;
+			if (L.Browser.cypressTest) {
+				clipboard = this._dummyClipboard;
+			}
 
-		try {
-			await clipboard.write([clipboardItem]);
-		} catch (error) {
-			window.app.console.log('navigator.clipboard.write() failed: ' + error.message);
+			try {
+				await clipboard.write([clipboardItem]);
+			} catch (error) {
+				window.app.console.log('navigator.clipboard.write() failed: ' + error.message);
 
-			// Warn that the copy failed.
-			this._warnCopyPaste();
-			// Once broken, always broken.
-			L.Browser.clipboardApiAvailable = false;
-			window.prefs.set('clipboardApiAvailable', false);
-			// Prefetch selection, so next time copy will work with the keyboard.
-			app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');
-		}
+				// Warn that the copy failed.
+				this._warnCopyPaste();
+				// Once broken, always broken.
+				L.Browser.clipboardApiAvailable = false;
+				window.prefs.set('clipboardApiAvailable', false);
+				// Prefetch selection, so next time copy will work with the keyboard.
+				app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');
+			}
+			}
 	},
 
 	// Parses the result from the clipboard endpoint into HTML and plain text.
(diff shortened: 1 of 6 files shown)


Minion3665:
I've made this iOS only, as mentioned in the PR description. The Android code is available at 3a648b7b9c5c299ffc57ead708e11acb18d4c67c, up until the time I make another pull request

browser/src/map/Clipboard.js
```diff
@@ -245,51 +248,67 @@
...
+		optionalFormData,
+		forClipboard,
+		progressFn,
+	) {
+		if (window.ThisIsAMobileApp)
+			throw new Error(
+				'Attempted to call _doAsyncDownload on a mobile platform, this will never work so should never happen...',
+			);
+
+		var request = new XMLHttpRequest();
```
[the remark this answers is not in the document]
Minion3665: My understanding here is that this probably doesn't matter. It's not executed on a server, and is quite limited (the path should have a /clipboard route appended to it, for instance).

Further, it was here before: this commit refactored the code around here which triggered the warning so I presume someone has analyzed the risk here before me... however still best to get someone else to look over so I'm not dismissing this off-hand. Certainly it's a credible warning: we really do make a request to a server specified on the clipboard, and that is core to the functionality here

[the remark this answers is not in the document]
eszkadev: It will always come from the user.
I think domains can be filtered by optional sanitizer used below: `url = window.processCoolUrl({ url: url, type: 'clipboard' });`
or by blocking some requests using headers defined inside CheckFileInfo.

browser/src/map/Clipboard.js
```diff
@@ -876,55 +879,60 @@ L.Clipboard = L.Class.extend({
...
+			window.app.console.log('New ' + command + ' promise');
+			// FIXME: add a timeout cleanup too ...
+			this._commandCompletion.push({
+				resolve: resolve,
+				reject: reject,
 			});
-		};
+		});
+
+		await commandCompletePromise;
```
eszkadev, line 921: why it is called here, and no params?

browser/src/map/Clipboard.js
```diff
@@ -876,55 +879,60 @@ L.Clipboard = L.Class.extend({
...
+		} catch (error) {
 			window.app.console.log('navigator.clipboard.write() failed: ' + error.message);
 
 			// Warn that the copy failed.
-			that._warnCopyPaste();
+			this._warnCopyPaste();
 			// Once broken, always broken.
 			L.Browser.clipboardApiAvailable = false;
-			window.prefs.set('clipboardApiAvailable', 'false');
+			window.prefs.set('clipboardApiAvailable', false);
```
eszkadev: previously it was string not boolean (`false`)

browser/src/map/Clipboard.js
```diff
@@ -956,32 +964,75 @@ L.Clipboard = L.Class.extend({
...
 			}
-		});
+			return;
+		}
+
+		if (isSpecial) {
+			this._navigatorClipboardPasteSpecial = true;
+		}
+
+		if (clipboardContents.length < 1) {
```
eszkadev, line 1052: I think it was better as it was (in separate method: `_navigatorClipboardReadCallback`) - > shorter function to read.

browser/src/map/Clipboard.js
```diff
@@ -977,7 +1005,32 @@ L.Clipboard = L.Class.extend({
 		let clipboardContents;
 
 		try {
-			clipboardContents = await clipboard.read();
+			if (window.ThisIsTheiOSApp) {
+				const encodedClipboardData = await window.webkit.messageHandlers.clipboard.postMessage('read');
```
eszkadev, line 1009: can we put that special iOS handling into separate method? Will be easier to read

eszkadev:
I read the online part. From code reading it looks like it wouldn't work in regular desktop case. But I might be confused as the diff is hard :)

So I left few comments to check for now

browser/src/map/Clipboard.js
Minion3665: Correct. I wrote the prefs API and these do the same thing (we convert everything that goes in to a string), but since as we are getting booleans later I feel it is clearer to do this ... it's also more consistent with other uses of the prefs API for storing booleans (e.g. see some of the stuff on the dark mode). As I was refactoring this code anyway, I decided to change the usage at the same time

image

Minion3665:
> From code reading it looks like it wouldn't work in regular desktop case. But I might be confused as the diff is hard :)

That's fair... my belief is that the hardest part is the convert to async/await patch (the first one), since as it affects a lot and isn't obviously limited to only things that would affect mobile. I would definitely recommend reading patch-by-patch if you aren't already doing so.

That said, I can try to make that a bit easier to read, but a lot of things are Just Quite Different when you async something and make the changes there to align it with more idiomatic async/await patterns ... for example functions that were callbacks will be pulled underneath the code that would previously have called them ... personally I find this easier to read as there's less jumping about to find what I need, but I can see the argument that functions are getting longer... I can possibly split up that patch further, but I suspect this will lead to a fair amount of garbage (much thenning and such) in intermediate patches. Maybe I should just do that, though...

browser/src/map/Clipboard.js
Minion3665: Sure, that makes sense, let me do that...

browser/src/map/Clipboard.js
Minion3665: I could, I guess, change the way this worked to look a little more conventional if I made a function that sent the command and returned a promise like this at the same time... maybe that would be easier to grok

eszkadev:
> > From code reading it looks like it wouldn't work in regular desktop case. But I might be confused as the diff is hard :)
> 
> That's fair... my belief is that the hardest part is the convert to async/await patch (the first one), since as it affects a lot and isn't obviously limited to only things that would affect mobile. I would definitely recommend reading patch-by-patch if you aren't already doing so.
> 
> That said, I can try to make that a bit easier to read, but a lot of things are Just Quite Different when you async something and make the changes there to align it with more idiomatic async/await patterns ... for example functions that were callbacks will be pulled underneath the code that would previously have called them ... personally I find this easier to read as there's less jumping about to find what I need, but I can see the argument that functions are getting longer... I can possibly split up that patch further, but I suspect this will lead to a fair amount of garbage (much thenning and such) in intermediate patches. Maybe I should just do that, though...

yes, I agree it's easier to read, but quite big diff.

For example I prefer to do such things in small steps:
Like one commit == 1 or 2 small callbacks moved to separate functions.
Then all is compact -> do a commit which converts into new flow. As you basically move single lines.
It's then easy to review from correctness point of view as scope is limited in every commit.

eszkadev:
`calc/clipboard_spec.js ` failed <- looks like true positive

browser/src/map/Clipboard.js
eszkadev: ok fine, I just note things I found different so you can check if they are intentional :)

Minion3665: all good, I see it as a non-functional change but I agree it does look like it could behave differently :)

vmiklos (commented):
I'm not against moving to async/await, but my understanding is that as soon as you have a single async in a function, all the caller functions have to be async as well, is that correct? If so, you would need to convert things from the outside to inside (can call sync from async, not the other way around), incrementally. The first commit itself is quite large (I know, this is subjective) with its "284 insertions(+), 227 deletions(-)". I understand doing it in steps is more work in the short term, but in case it breaks, the next person who will have to fix things with bisect will thank you for working incrementally. :-) It also helps not to reformat the code you move around, just fix the indent, so the 'git show -w' output is short. In short, I'm not against the change, but you do multiple things here, which is then hard to review.
I would suggest to move the first commit to a separate PR, and at a minimum make sure that no single commit touches more than 100 lines. And try to just do the async conversion (as mechanical as possible), so a review is possible for mere mortals like me. :-)
It also helps if you ask for review only after CI passed, since I'm not good at catching things a machine can easily notice.

I can't comment on the iOS part, but code-reading it, I see nothing obviously problematic with it.

For the getMetaBase commit: that "meta" marker has two purposes: once we write to the clipboard in HTML format, we want to make sure that the reading side can cheat by internal copying or ODF import when they read the HTML. The ODF part is not interesting on mobile, since the receiving end won't be able to connect to the mobile COOL anyway, I agree. But detecting internal copy is interesting: that's necessary if you want to be able to correctly copy footnotes in Writer or copy formulas in Calc (where Calc rewrites your formulas on paste). I'm just commenting about that here, in case it would sound to use that the meta marker in the HTML output is not interesting on mobile at all.

I don't see anything obviously problematic with the last commit. Thanks!

Minion3665:
> I'm not against moving to async/await, but my understanding is that as soon as you have a single async in a function, all the caller functions have to be async as well, is that correct? 

Not quite ... so after Szymon's review I'm splitting that commit up a lot and here's one of the changes I have to make to support converting the `_doAsyncDownload` function

image

Thankfully, in JavaScript async/await is really syntactic sugar around promises, which are fairly similar to callbacks and require trivial conversion (moving the callback functions from arguments into `.then` and `.catch`). Happily this means that, as long as you're OK with shuffling your callbacks a little, you can convert `async` functions from the inside out if you like.

> The first commit itself is quite large (I know, this is subjective) with its "284 insertions(+), 227 deletions(-)".

Yes, don't worry, I also think it's large - I'm working on splitting it now ... I'll also move it out into another PR as you suggest

Minion3665:
Patches to convert to async/await are [being added when I've split them off] here: https://github.com/CollaboraOnline/online/pull/10498

pedropintosilva:
all the failed tests were not related to this

browser/src/map/Clipboard.js
```diff
@@ -929,22 +947,49 @@ L.Clipboard = L.Class.extend({
...
+		const clipboardData = Array.from(
+			encodedClipboardData.split(' '),
+		).map((encoded) =>
+			(encoded === '(null)' ? '' : window.b64d(encoded)),
+		);
+
+		const dataByMimeType = {};
+
+		if (clipboardData[0]) {
+			dataByMimeType['text/plain'] = new Blob([clipboardData[0]]);
```
eszkadev, line 969: are we sure it always appears in that order?    plain in index 0, html in index 1 ?

eszkadev (commented):
+1 , I read the code and looks fine - but I have no setup to test

browser/src/map/Clipboard.js
Minion3665: Yep! The native implementation of this function is later down, and always returns in that order
```objc
            NSString * encodedPlainPayload = [plainPayload base64EncodedStringWithOptions:0];
            NSString * encodedHtmlPayload = [htmlPayload base64EncodedStringWithOptions:0];

            replyHandler([NSString stringWithFormat:@"%@ %@", encodedPlainPayload, encodedHtmlPayload], nil);
```

vmiklos (commented):
Apart from what Szymon wrote, I don't see anything problemtic with this anymore. Thanks.

vmiklos (approved):
Ah, I didn't see the +1 from Szymon, then I'm OK with this.

Minion3665:
awesome, thanks @vmiklos!

Would you mind also taking a look at [the gerrit patches](https://gerrit.libreoffice.org/q/topic:%22private/skyler/ios-clipboard%22) since as they need to be merged at the same time

vmiklos:
Yes, I just went through that.

PR #10490, opened 2024-11-14
sometimes this test has 'Pages 1 and 2 of 6'

caolanm:
instead of 'Page 2 of 6'

similar ambiguity exists in statusbar test

Change-Id: I436a749c663d35684140373189c8ff2a88799c3f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (approved):
Thanks, it was very annoying.

PR #10492, opened 2024-11-14
private/marco/slideshow for review 2

mcecchetti:
- **slideshow: repeat initial transition when go back from first slide**
- **slideshow: save/resore animated element state on effect start/rewind**
- **slideshow: when parsing a hsl/rgb string values can be float**
- **slideshow: ignore some invalid attribute instead of remove whole effect**
- **slideshow: skip slide transitions in the 2d context case**


quikee (approved):
Looks fine to me.

PR #10494, opened 2024-11-14
enable/disable presenter console

hcvcastro:
- **notebookbar: add listener to disable button**
- **browser: slideshow: enable/disable presenter console**

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -61,6 +61,7 @@ L.Control.Notebookbar = L.Control.extend({
 		this.map.on('darkmodechanged', this.onDarkModeToggleChange, this);
 		this.map.on('showannotationschanged', this.onShowAnnotationsChange, this);
 		this.map.on('a11ystatechanged', this.onAccessibilityToggleChange, this);
+		this.map.on('enablecommand', this.disableNotebookbarControl, this);
 		if (docType === 'presentation') {
 			this.map.on('updateparts', this.onSlideHideToggle, this);
 			this.map.on('toggleslidehide', this.onSlideHideToggle, this);
@@ -339,6 +340,14 @@ L.Control.Notebookbar = L.Control.extend({
 		);
 
 		this.reloadShortcutsBar();
+	},
+
+	disableNotebookbarControl: function(e) {
+		var toolbar = L.DomUtil.get('toolbar-up');
+		this.builder.executeAction(toolbar, {
+			control_id: e.controlId,
+			action_type: e.enable ? 'enable' : 'disable',
+		});
 	},
 
 	showNotebookbarButton: function(buttonId, show) {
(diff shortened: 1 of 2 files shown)


Darshan-upadhyay1110 (approved):

browser/src/control/Control.Notebookbar.js
```diff
@@ -341,6 +342,24 @@ L.Control.Notebookbar = L.Control.extend({
...
+		var toolbar = L.DomUtil.get('toolbar-up');
+		var container = toolbar.querySelector(e.buttonId);
+		var button = toolbar.querySelector(e.buttonId + '-button');
+		if (!container || !button) {
+			return;
+		}
+
+		var disabled = Boolean(e.disabled);
+		if (disabled) {
+			L.DomUtil.addClass(container, 'disabled');
```
eszkadev, line 355: please do not do it manually but over JSDialog API,
in other case it might to stop working with some events at some point...

eszkadev (changes requested):

browser/src/control/Control.Notebookbar.js
eszkadev: this method: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Component.Toolbar.ts#L115

on this component: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.TopToolbar.js

eszkadev: can be found in `this.map.topToolbar = JSDialog.TopToolbar(this.map);`

eszkadev: ah it's for notebookbar, I confused `toolbar-up`...
above pointers are for compact toolbars, so please do something similar here:
```
this.builder.executeAction(this.parentContainer, {
	control_id: command,
	action_type: enable ? 'enable' : 'disable',
});
```

hcvcastro: Ok, that is the reason I could not find the disable function

browser/src/control/Control.Notebookbar.js
```diff
@@ -341,6 +342,14 @@ L.Control.Notebookbar = L.Control.extend({
 		this.reloadShortcutsBar();
 	},
 
+	disableNotebookbarControl: function(e) {
+		var toolbar = L.DomUtil.get('toolbar-up');
+		this.builder.executeAction(toolbar, {
```
eszkadev, line 347: just `this.parentContainer` can be used

eszkadev (approved):
thanks, looks good now

browser/src/control/Control.Notebookbar.js
hcvcastro: It does not work, It looks like this.parentContainer was not updated when detach childs,
but it is another issue to address.

PR #10506, opened 2024-11-15
fix(clipboard): set command in _doCopyCut

Minion3665:
_doCopyCut has the ability to call functions that depend on _unoCommandForCopyCutPaste, but despite having access to the required uno command here it never sets the variable.

That causes copy/paste commands that go through this route to use the last command executed, or if no commands were executed before this to throw an error and fail entirely.

I noticed this while working on #10498, but decided to avoid including it there as that is a pure refactor

Change-Id: I00549aa872fee8f51b5c0426e5d1eb1569210378

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -1040,6 +1040,7 @@ L.Clipboard = L.Class.extend({
 				this._clipboardSerial++;
 			}
 		} else {
+			this._unoCommandForCopyCutPaste = `.uno:${unoName}`;
 			preventDefault = this.populateClipboard(ev);
 		}
 


vmiklos (approved):
Looks good to me. We have two codepaths here:

- _execCopyCutPaste() is when the user clicks on a button and an UNO command gets dispatched and our JS copying is a side effect of that
- _doCopyCut() is when the user presses a key, and the browser initiates the copying

_execCopyCutPaste() already writes this._unoCommandForCopyCutPaste, so it's valid to requre that populateClipboard() writes it, but only its caller _doCopyCut() has the info, so it looks correct to save the action there.

You can also see the difference in the console. Before:

```
15:34:04.272 global.js:564 1731681244272 OUTGOING: uno null
15:34:04.272 global.js:564 New null promise
```

after:

```
15:38:42.692 global.js:564 1731681522692 OUTGOING: uno .uno:Copy
15:38:42.693 global.js:564 New .uno:Copy promise
```

When testing from Chrome, some error handling later this made copying work in practice, but this definitely looks like an improvement.

Nit: next time it would be ideal to 1) wait for CI before you request code review and 2) either have a matching github issue or have the steps to reproduce the problem you see in practice, since that's not clear from the commit message to me. Thanks. :-)

Minion3665:
RE 2: Sure, you're right, that is confusing if you don't already know what _doCopyCut is ... let me reword it

PR #10507, opened 2024-11-15
optimize draw next frame

hcvcastro:
- **browser: slideshow: fix request animation frame call**
- **browser: slideshow: optimize draw next frame**


mcecchetti (approved):
Looks fine

PR #10511, opened 2024-11-17
private/marco/presenter console to be reviewed 1

mcecchetti:
- **presenter console: on Firefox current slide canvas has unexpected size**
- **presenter console: current slide view should preserve aspect ratio**


quikee (approved):
Looks good to me

PR #10512, opened 2024-11-18
Apply TaskPriority to KitSocketPoll::kitPoll, help to determine idle-ness

sgothel:
* Resolves: # 
* Target version: master 

### Summary
This enhancement of [#10183](https://github.com/CollaboraOnline/online/pull/10183)
require the [LO TaskPriority patch](https://gerrit.libreoffice.org/c/core/+/176701).

### Checklist


kit/Kit.cpp
```diff
@@ -2383,16 +2383,36 @@ void Document::drainQueue()
...
-            // Put requests that include tiles in the visible area to the front to handle those first
-            std::partition(tileRequests.begin(), tileRequests.end(), [this](const TileCombined& req) {
-                return isTileRequestInsideVisibleArea(req); });
-            for (auto& tileCombined : tileRequests)
-                renderTiles(tileCombined);
+            std::vector<TileCombined*> earmarkedTiles;
+            size_t renderedTiles = 0;
+            for(TileCombined& tileCombined : tileRequests)
+            {
+                if (isIdle || (renderedTiles==0 && isTileRequestInsideVisibleArea(tileCombined)))
```
caolanm, line 2395: If we are idle then render everything, that seems straight forward. But if we are not idle, and there are requests in the visible area, then we just render the first of those requests here? Why not just render all the visible area tiles?

kit/Kit.cpp
```diff
@@ -2830,30 +2851,57 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
...
+                                (taskPriority == 100 && timeoutMicroS > 2500000); // LO-IdleHandler's SC_IDLE_MAX = 3s, allowing a 500ms lag
         if (checkForIdle)
             timeoutMicroS = 0;
 
+        auto getTileQueueSize = [&]() -> size_t {
+            return _document ? _document->getTileQueueSize() : 0;
+        };
         // Flush at most maxEvents+1, or return when nothing left.
         _pollEnd = startTime + std::chrono::microseconds(timeoutMicroS);
+        int slice = 0;
```
caolanm, line 2867: so, if we want to poll until a given timeout, and we know from core that it is "idle" then we split the timeout to do a quick initial poll to see if there are any events, and if there are none we can drain the queue of pending non-priority rendering requests, right?

Does that mean there is no point in realTimeout = std::min(100000, timeoutMicroS), I don't get what difference it makes, but I'm probably missing something. Maybe some comments in here as to the intention would help.

kit/Kit.cpp
```diff
@@ -2830,30 +2851,57 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
     }
     else
     {
+        // taskPriority: 100==TaskPriority::DEFAULT, 110==TaskPriority::DEFAULT_IDLE
+        const bool loIdleDefault = (taskPriority > 100 && taskPriority <= 110) ||
+                                   (taskPriority == 100 && timeoutMicroS == 0);
+        const bool loIdleLong = taskPriority > 110 ||
+                                (taskPriority == 100 && timeoutMicroS > 2500000); // LO-IdleHandler's SC_IDLE_MAX = 3s, allowing a 500ms lag
```
caolanm, line 2858: Was there some known problem that prompts the use of SC_IDLE_MAX here? The special casing and then there's a distinction between idle and very idle?

caolanm:
I think we could split the first constizing commit into its own pr and get that in right away.

kit/Kit.cpp
```diff
@@ -2383,16 +2383,36 @@ void Document::drainQueue()
...
+                {
+                    renderTiles(tileCombined);
+                    renderedTiles += tileCombined.getTiles().size();
+                }
+                else if(renderedTiles>0)
+                    _queue->pushTileQueue(tileCombined.getTiles());
+                else
+                    earmarkedTiles.push_back(&tileCombined);
+            }
+            for(TileCombined* tileCombined : earmarkedTiles)
```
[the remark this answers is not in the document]
sgothel: yes, we could end up missing rendering tiles if not becoming idle, i.e. the ` unit-tiletest` (`TileCacheTests::testClientPartCalc`) failed otherwise waiting for one invisible tile.

kit/Kit.cpp
sgothel: The IdleHandler didn't set taskPriority higher (sadly), so I identify it this way. Besides, there is an 'Await' (spellchecker etc) task which also used the magic 3s clearly denoting a long idle period (w/o changing the priority).

kit/Kit.cpp
sgothel: Additionally in non-idle mode, we would also render one invisible tile if all tiles are invisible.

In non-idle mode we assume to have limited time and only render one tile, then continue with the poll+drain loop.

sgothel: added comment

kit/Kit.cpp
```diff
@@ -2830,30 +2851,57 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
...
+                << ", to[" << timeoutMicroSPre/1000 << "ms, " << realTimeout/1000 << "ms]: rc " << rc
+                << ", t# " << tilesInQueue - getTileQueueSize() << "/" << tilesInQueue << ", dqc " << drainedQueueCalls
+                << ", used " << std::chrono::duration_cast<std::chrono::milliseconds>(now - startTime).count()
+                << "ms, left " << timeoutMicroS/1000 << "ms");
+            ++slice;
         } while (timeoutMicroS > 0 && !SigUtil::getTerminationFlag() && maxExtraEvents-- > 0);
     }
 
-    if (_document && checkForIdle && eventsSignalled == 0 && timeoutMicroS > 0 &&
+    if (_document && checkForIdle && eventsSignalled == 0 && timeoutMicroS > 0 && // bug: never: checkForIdle && timeoutMicroS > 0
```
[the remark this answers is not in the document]
sgothel: removed - and it is marked here at least

kit/Kit.cpp
sgothel: added note to comment

kit/Kit.cpp
sgothel: in the revised version using LibreOfficeKitIdleHintType it could be clarified better. 
However, the SC_IDLE_MAX case on default priority still needs to be detected.

kit/Kit.cpp
```diff
@@ -2441,17 +2444,39 @@ void Document::drainQueue()
...
+                    res.tilesSent += tileCombined.getTiles().size();
+                }
+                else if(res.tilesSent>0)
+                    _queue->pushTileQueue(tileCombined.getTiles());
+                else
+                    earmarkedTiles.push_back(&tileCombined);
+            }
+            for(TileCombined* tileCombined : earmarkedTiles)
+            {
+                if(res.tilesSent==0) // ensure at least one TileCombined gets rendered (if all are invisible)
```
mmeeks, line 2470: consistent white-space.

kit/Kit.cpp
```diff
@@ -2441,17 +2444,39 @@ void Document::drainQueue()
...
+            // The latter ensures draining tiles even without idle-ness.
             std::vector<TileCombined> tileRequests = _queue->popWholeTileQueue();
-
-            // Put requests that include tiles in the visible area to the front to handle those first
-            std::partition(tileRequests.begin(), tileRequests.end(), [this](const TileCombined& req) {
-                return isTileRequestInsideVisibleArea(req); });
-            for (auto& tileCombined : tileRequests)
-                renderTiles(tileCombined);
+            std::vector<TileCombined*> earmarkedTiles;
+            for(TileCombined& tileCombined : tileRequests)
```
mmeeks, line 2456: consistent white-space needed.

kit/Kit.cpp
```diff
@@ -2888,26 +2916,51 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
...
+            const size_t tilesPrePoll = _document ? _document->getTileQueueSize() : 0;
+            const bool quickPoll = _document && _document->needsQuickPoll();
+            int realTimeout;
+            if (quickPoll || (slice==0 && (loIdleCond || loIdleLong)))
                 realTimeout = 0;
+            else
+                realTimeout = std::min(100000, timeoutMicroS); // 10Hz 100ms maximum slice
 
-            if (poll(std::chrono::microseconds(realTimeout)) <= 0)
+            const int rc = poll(std::chrono::microseconds(realTimeout));
```
mmeeks, line 2941: Surely we know the poll 'did' something - when we get more events queued to process in drainQueue - there is no need to check the result of the poll (?)

kit/Kit.cpp
```diff
@@ -2888,26 +2916,51 @@ int KitSocketPoll::kitPoll(int timeoutMicroS)
...
+            if (rc<0) // poll error
+                break;
+            else if (rc>0) // poll event
+                ++eventsSignalled;
+            else if(realTimeout>0 && timeoutMicroS-realTimeout<=0) // && rc==0: poll timeout on last slice
```
mmeeks, line 2946: Surely we should not have a >0 realTimeout if we have idle work to do anyway - waiting any real time to determine if we are idle doesn't seem helpful.

kit/Kit.cpp
```diff
@@ -2441,17 +2444,39 @@ void Document::drainQueue()
             }
         }
 
+        const size_t tilesQueued = _queue->getTileQueueSize();
         if (processInputEnabled() && !isLoadOngoing() &&
-            !isBackgroundSaveProcess() && _queue->getTileQueueSize() > 0)
+            !isBackgroundSaveProcess() && tilesQueued > 0)
         {
+            // Render all tiles if idle.
```
mmeeks, line 2451: All pending tiles could be a huge queue of work that could take perhaps seconds to do - if we've done a big scroll. Why would we choose to block all users for seconds when they may have more immediate things they want doing in order to render all the work.

This approach seems to be essentially back where we were before we last discussed this at some length.

mmeeks (changes requested):
I find it hard to reconcile this with the call and discussion we had; perhaps I'm missing something.

PR #10514, opened 2024-11-18
wsd: test: quarantine tests

Ashod:
A new crash test and improvements to an old one.


caolanm (approved):
test only changes, so I imagine uncontroversial

PR #10515, opened 2024-11-18
cool#4250 - dpiscale: propagate from the client to the view via new l…

mmeeks:
…ok API.

In order to get the zoom level correct during save, it is necessary to get the dpiscale from the browser to the client. Use a new LOK setViewOption API to do this.

Change-Id: Ib32d4765fffefaf9b6b52f6a789834914ae58b6f

This will need: https://gerrit.libreoffice.org/c/core/+/176715 cool#4250 - add API to set per view options
merging before it will build.


mmeeks:
Would love a rubber stamp here in green; the real bug can then be fixed with the information that gets to core.

wsd/ClientSession.hpp
```diff
@@ -376,6 +376,7 @@ class ClientSession final : public Session
     int _tileHeightPixel;
     int _tileWidthTwips;
     int _tileHeightTwips;
+    double _dpiScale;
```
caolanm, line 379: This _dpiScale member isn't used (and same for local var dpiScaleStr in ClientSession.cpp). I imagine we don't actually need those if we are just passing through the message from CanvasTileLayer.js?

caolanm (approved):

PR #10516, opened 2024-11-18
sidebar: fix TypeError on switch

eszkadev:
This fixes the error on switching the decks in Writer. Regression from commit b5eab2c988515c904f8c4b6c975a25e1f777e028 browser: treeview: remove global var hack


gokaysatir (approved):
Couldn't repdoruce but looks reasonable. Thanks :)

PR #10522, opened 2024-11-18
redact a few more sensitive config items in the log

timar:
Change-Id: I9ce1b4d0e53d29b95bc2023604b1682cc197f604


vmiklos (approved):
Makes sense -- if these still matter when debugging a problem, we can always just ask for the actual config.

PR #10523, opened 2024-11-18
stateChangeHandler: EditDoc now uses boolean or disabled payload

lpranam:
now EditDoc may have true, false or disabled values, handle it accordingly

Change-Id: I52668290007cc07d3f1bae015d533d50fc538c76

* Target version: master 

### Checklist


lpranam:
requires: https://gerrit.libreoffice.org/c/core/+/176736

PR #10527, opened 2024-11-19
Fix: Notes placeholder string gets mixed up with "Slide of" string

Darshan-upadhyay1110:
- Because we same variable is the reason that notes text gets into #slides container text
- this patch will get notes content container and set inner text by default "no notes"
- we will assign text only if notes are defined in current slide

Before:

image

After:

image

Change-Id: I32dd646ca4cbabce94842cbc3d34cda636ec45b1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -900,13 +900,10 @@ class PresenterConsole {
 
 		if (this._notes) {
 			let notes = this._presenter.getNotes(e.slide);
-			if (notes) elem = this._notes.querySelector('#notes');
-			if (elem) {
-				elem.innerText =
-					notes && notes.toLowerCase() !== 'click to add notes'.toLowerCase()
-						? notes
-						: _('No Notes');
-			}
+			let notesContentElem = this._notes.querySelector('#notes');
+			notesContentElem.innerText = _('No Notes');
+			if (notes && notes.toLowerCase() !== 'click to add notes'.toLowerCase())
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/PresenterConsole.js
```diff
@@ -900,13 +900,13 @@ class PresenterConsole {
...
-				elem.innerText =
-					notes && notes.toLowerCase() !== 'click to add notes'.toLowerCase()
-						? notes
-						: _('No Notes');
-			}
+			let notesContentElem = this._notes.querySelector('#notes');
+			notesContentElem.innerText = _('No Notes');
+			if (
+				notes &&
+				notes.toLowerCase() !== _('click to add notes').toLowerCase()
```
eszkadev, line 907: this string is not set anywhere: "click to add notes"
does it come from core? we shouldn't compare translated versions as core and online might be different...

Darshan-upadhyay1110: @eszkadev so should we just do normal string comparison ?

Darshan-upadhyay1110: https://github.com/CollaboraOnline/online/pull/10521#issuecomment-2484080473

eszkadev: we should use some not fragile approach, which will work if core will change translation

eszkadev (approved):

PR #10528, opened 2024-11-19
Implement remote multimedia insertion support

mikekaganski:
Similar to the existing "Insert Image", this turns multimedia insertion
button into a menu, which is optionally expanded to include the remote
item. It depends on explicit integration support (controlled by present
"EnableInsertRemoteMultimedia" file info response field), and requests
the integration to show a "pick an asset" UI with given MIME types (by
sending 'UI_InsertFile' request), and to reply Action_InsertMultimedia,
that expects an URL.

Change-Id: Ide6eefa6a55213f4ec0d1e47e35b5e4bf4a109be


mikekaganski:
richdocuments integration support:
https://github.com/mikekaganski/richdocuments/commit/3d4874f3d5fafff917d362c52c91e0d4a291b3e3

vmiklos (approved):
I don't see anything obviously problematic with this. Thanks.

PR #10530, opened 2024-11-19
cool#9992 doc electronic sign: allow passing sign params from JS

vmiklos:
Try to dispatch the .uno:Signature command from the JS console with
parameters, the parameters get lost by the time they hit the LOK API
boundary.

This is because kit/ code tries to annotate .uno:Signature with the
signature cert/key PEM data, but it won't append this data the
parameters, just replace the original parameters.

Given that there is no use-case to combine the parameters, just avoid
annotating when there are parameters already: electronic signing /
external signatures will have all the info as parameters in JS.

With this, a code snippet like:
app.socket.sendMessage('uno .uno:Signature {"SignatureTime":{"type":"string","value":"..."},"SignatureValue":{"type":"string","value":"..."}}');
works as expected, similar to how it works in core in a cppunit test.

Change-Id: I42df9bf864fe0f7a2854429652d15e4e61efda7a

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -676,7 +676,8 @@ bool ChildSession::_handleInput(const char *buffer, int length)
             }
             else if (tokens[1].find(".uno:Signature") != std::string::npos)
             {
-                if (unoSignatureCommand())
+                // See if the command has parameters: if not, annotate with sign cert/key.
+                if (tokens.size() == 2 && unoSignatureCommand())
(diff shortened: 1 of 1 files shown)


vmiklos:

For the eIDEasy based signing, we'll learn about the signature value in a browser-based interface, so the JS will have the value and will forward it to the server (and no annotations are needed there), so the idea is to just not do the annotation when it's useless & problematic.

PR #10532, opened 2024-11-19
Log UI commands related to undo

aszucs3:
Log several kind of commands, and Undo count change. This may helps to find what user try to undo at most cases.

We can set what kind of commands we want to log at:
    std::set<std::string> aCmdToLog = {
        "uno", "key", "mouse", "textinput", "removetextcontext",
        "paste", "insertfile", "commandvalues" };

We can set what commands should be merged at:
    std::set<std::string> aCmdToMergeLog = {
        "mouse", "textinput", "removetextcontext" };
Merged means, if the same command is called again, it wont be logged again, but the next time before a different command is called, the merged command's count will be logged.

Some commands merge may need some more deep logic. at "mouse" command only the move is merged.
"key" could be more complex as it can be an insert space, arrow keys, or pageup/down. now all of them are logged but sure we can merge some of them if needed

It is logged into session.log file.. that exist only while the file is opened. When Childsession is destructed this file is copied, appended into /tmp/coolwsd-ui-cmd.log This file is configurable similarly as coolwsd.log

added a python script to make actual statistics

Multi user case may confuse things, as undo may used on non-last undo change. maybe it could be handed too.

sample LOGs:

CMD:textinput id=0 text=f
UNDO_CHG:(4 -> 5):textinput id=0 text=f
CMD_REPATED:5
CMD:mouse type=move x=4485 y=7080 count=1 buttons=0 modifier=0 CMD_REPATED:2
CMD:mouse type=buttondown x=4530 y=7125 count=1 buttons=1 modifier=0 CMD:mouse type=buttonup x=4530 y=7125 count=1 buttons=1 modifier=0 CMD:mouse type=move x=4665 y=7305 count=1 buttons=0 modifier=0 CMD_REPATED:1
CMD:key type=input char=32 key=0
UNDO_CHG:(5 -> 6):key type=input char=32 key=0
CMD:uno .uno:Undo
UNDO_CHG:(6 -> 5):uno .uno:Undo

CMD:uno .uno:ToolbarMode?Mode:string=notebookbar_online.ui CMD:commandvalues command=.uno:LanguageStatus
CMD:commandvalues command=.uno:ViewAnnotations
CMD:uno .uno:SidebarShow

probably some more info may needed at some places.. maybe at document open/close...

Change-Id: I276b8fd42062e45f271f60644c1b79509bc60689

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -423,6 +423,21 @@ namespace Log
 
     } Static;
 
+    static struct StaticUIHelper: StaticHelper
+    {
+    private:
+        bool _mergeCmd = false;
+        bool _logTimeEndOfMergedCmd = false;
+    public:
+        void setLogMergeInfo(bool mergeCmd, bool logTimeEndOfMergedCmd)
+        {
+            _mergeCmd = mergeCmd;
+            _logTimeEndOfMergedCmd = logTimeEndOfMergedCmd;
+        }
+        bool getMergeCmd() const { return _mergeCmd; }
+        bool getLogTimeEndOfMergedCmd() const { return _logTimeEndOfMergedCmd; }
+    } StaticUILog;
+
     thread_local GenericLogger* StaticHelper::_threadLocalLogger = nullptr;
 
     bool IsShutdown = false;
@@ -601,7 +616,9 @@ namespace Log
                     const std::string& logLevel,
                     const bool withColor,
                     const bool logToFile,
-                    const std::map<std::string, std::string>& config)
+                    const std::map<std::string, std::string>& config,
+                    const bool logToFileUICmd,
+                    const std::map<std::string, std::string>& configUICmd)
     {
         Static.setName(name);
         std::ostringstream oss;
@@ -669,6 +686,30 @@ namespace Log
         LOG_INF("Initializing " << name << ". Local time: "
                                 << std::put_time(localtime_r(&t, &tm), "%a %F %T %z")
                                 << ". Log level is [" << logger->getLevel() << ']');
+
+        StaticUILog.setName(name+"_ui");
+        AutoPtr<Channel> channelUILog;
+        if (logToFileUICmd)
+        {
+            channelUILog = static_cast<Poco::Channel*>(new Poco::FileChannel("coolwsd-ui-cmd.log"));
+            for (const auto& pair : configUICmd)
+            {
+                channelUILog->setProperty(pair.first, pair.second);
+            }
+
+            channelUILog = static_cast<Poco::Channel*>(new Poco::FileChannel("/tmp/coolwsd-ui-cmd.log"));
+            channelUILog->open();
+            try
+            {
+                auto& loggerUILog = GenericLogger::create(StaticUILog.getName(), std::move(channelUILog), Poco::Message::PRIO_TRACE);
+                StaticUILog.setLogger(&loggerUILog);
+            }
+            catch (ExistsException&)
+            {
+                auto loggerUILog = static_cast<GenericLogger *>(&Poco::Logger::get(StaticUILog.getName()));
+                StaticUILog.setLogger(loggerUILog);
+            }
+        }
     }
 
     GenericLogger& logger()
@@ -681,6 +722,46 @@ namespace Log
         return pLogger ? *pLogger
             : *static_cast<GenericLogger *>(
                 &GenericLogger::get(Static.getInited() ? Static.getName() : std::string()));
+    }
+
+    GenericLogger& loggerUI()
+    {
+        GenericLogger* logger = StaticUILog.getThreadLocalLogger();
+        if (logger != nullptr)
+            return *logger;
+
+        logger = StaticUILog.getLogger();
+        return logger ? *logger
+            : *static_cast<GenericLogger *>(
+                &GenericLogger::get(StaticUILog.getInited() ? StaticUILog.getName() : std::string()));
+    }
+
+    bool isLogUIEnabled()
+    {
+        if (StaticUILog.getThreadLocalLogger() || StaticUILog.getLogger())
+            return true;
+        return false;
+    }
+
+    void logUI(Level l, const std::string &text)
+    {
+        if (isLogUIEnabled())
+            Log::loggerUI().doLog(l, text);
+    }
+
+    bool isLogUIMerged()
+    {
+        return StaticUILog.getMergeCmd();
+    }
+
+    bool isLogUITimeEnd()
+    {
+        return StaticUILog.getLogTimeEndOfMergedCmd();
+    }
+
+    void setUILogMergeInfo(bool mergeCmd, bool logTimeEndOfMergedCmd)
+    {
+        StaticUILog.setLogMergeInfo(mergeCmd, logTimeEndOfMergedCmd);
     }
 
     bool isEnabled(Level l, Area a)
--- a/configure.ac
+++ b/configure.ac
@@ -412,6 +412,8 @@ COOLWSD_LOG_TO_FILE="false"
 COOLWSD_LOGFILE="/var/log/coolwsd.log"
 COOLWSD_TRACEEVENTFILE="/var/log/coolwsd.trace.json"
 COOLWSD_ANONYMIZE_USER_DATA=false
+COOLWSD_LOG_UICMD_TO_FILE="false"
+COOLWSD_LOGFILE_UICMD="/var/log/coolwsd-ui-cmd.log"
 BROWSER_LOGGING="false"
 debug_msg="secure mode: product build"
 anonym_msg=""
@@ -438,6 +440,8 @@ if test "$enable_debug" = "yes"; then
    COOLWSD_LOGFILE="/tmp/coolwsd.log"
    COOLWSD_TRACEEVENTFILE="/tmp/coolwsd.trace.json"
    COOLWSD_ANONYMIZE_USER_DATA=false
+   COOLWSD_LOG_UICMD_TO_FILE="true"
+   COOLWSD_LOGFILE_UICMD="/tmp/coolwsd-ui-cmd.log"
    BROWSER_LOGGING="true"
    debug_msg="low security debugging mode"
    SSL_VERIFY="false"
@@ -472,6 +476,7 @@ AC_SUBST(ENABLE_DEBUG)
 AC_SUBST(ENABLE_BUNDLE)
 AC_SUBST(COOLWSD_LOGLEVEL)
 AC_SUBST(COOLWSD_LOG_TO_FILE)
+AC_SUBST(COOLWSD_LOG_UICMD_TO_FILE)
 AC_SUBST(BROWSER_LOGGING)
 AC_SUBST(NUM_PRESPAWN_CHILDREN)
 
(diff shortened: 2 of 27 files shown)


wsd/COOLWSD.cpp
```diff
@@ -1199,6 +1199,23 @@ void COOLWSD::innerInitialize(Poco::Util::Application& self)
         { "logging.docstats", "false" },
         { "logging.userstats", "false" },
         { "logging.disable_server_audit", "false" },
+        { "logging_ui_cmd.file.property[0]", "coolwsd-ui-cmd.log" },
+        { "logging_ui_cmd.file.property[0][@name]", "path" },
+        { "logging_ui_cmd.file.property[1]", "never" },
```
caolanm, line 1204: I think all these options like rotation and archive etc aren't needed, they are complicated things that the "main log" has to compress and backup these logs etc, and we probably don't need them for this log and it would makes things smaller and simpler to not have them. We could add them if we need them later.

configure.ac
```diff
@@ -412,6 +412,8 @@ COOLWSD_LOG_TO_FILE="false"
 COOLWSD_LOGFILE="/var/log/coolwsd.log"
 COOLWSD_TRACEEVENTFILE="/var/log/coolwsd.trace.json"
 COOLWSD_ANONYMIZE_USER_DATA=false
+COOLWSD_LOG_UICMD_TO_FILE="false"
```
caolanm, line 415: line 479 of COOLWSD_LOG_TO_FILE_UICMD doesn't match here, this 415 line should likely be fixed to match the other

wsd/COOLWSD.cpp
aszucs3: I deleted some from coolwsd.xml.in, but i left default values here.
If user set them, Poco will use them.

I also added 2 non Poco setting .. so it can make unümerged logs if needed.. and it can write time intervall if needed.

configure.ac
aszucs3: Thx, i fixed the name problem :)

scripts/undoStatistics.py
```diff
@@ -0,0 +1,169 @@
...
+        usageAndExit()
+
+    documents = []
+    users = {}
+
+    # dict of (command - count of undo)
+    undoedCommands = {}
+    totalUndoedCommands = {}
+
+    # Process all Document relatd, and undo related calculations
```
aszucs3, line 52: typo relatd -> related

kit/Kit.hpp
```diff
@@ -186,6 +186,43 @@ class KitSocketPoll final : public SocketPoll
...
+    void processLogFile();
+    bool loadNextLine();
+    int findNextUserLine(int lineId);
+    bool mergeIfMouseClick(int lineId);
+    bool removeIfFalseUndoDecrease(int lineId);
+private:
+    // A temporary file for logging ui commands
+    std::fstream aFileStreamUICommands;
+    std::time_t aKitStartTimeSec;
+    std::string aKitStartTimeStr;
```
aszucs3, line 218: Probably It could be optimized out by calculating it from aKitStartTimeSec somehow, but this way it was simple to do.

kit/ChildSession.hpp
```diff
@@ -29,6 +29,23 @@
 class Document;
 class ChildSession;
 
+class LOG_UICOMMANDS {
```
mmeeks, line 32: This is the first SHOUTING_NAMES_WITH_UNDERSCORES I have ever seen across this code-base, or the LibreOffice core code. Please rename this to fit in with the rest of the code.

kit/Kit.hpp
```diff
@@ -186,6 +186,43 @@ class KitSocketPoll final : public SocketPoll
 #endif
 };
 
+class LogUiCmd
```
mmeeks, line 189: Can we put this UI command logging stuff in it's own header somewhere ?

kit/Kit.cpp
```diff
@@ -3983,4 +4001,225 @@ void dump_kit_state()
...
+        }
+    }
+    else
+    {
+        // Process the UI command log file. merge what we can
+        // and copy it to a bigger permanent file
+        nLoadedLines = 0; //Loaded lined count
+        int nProcessedLines = 0;
+        bFileNotEnd = true;
+        //std::vector<lineOfUiLogFile> aLoadedLines;
```
mmeeks, line 4120: This all looks like core code: we don't use these hungarian notation in online - please fit in with the rest of the code re: coding style.

What is this commented out aLoadedLines ?

kit/Kit.cpp
```diff
@@ -3983,4 +4001,225 @@ void dump_kit_state()
...
+                            if (!(aLoadedLines[nNextLine].bMerged
+                                || (aLoadedLines[nProcessedLines].aCmd.starts_with("cmd")
+                                && aLoadedLines[nNextLine].aCmd.starts_with("undo-count-change"))))
+                            {
+                                if (aLoadedLines[nNextLine].aCmd == aLoadedLines[nProcessedLines].aCmd
+                                    && aLoadedLines[nNextLine].aSubCmd == aLoadedLines[nProcessedLines].aSubCmd)
+                                {
+                                    //found a mergeable .. continue to search more
+                                    nMergeCount++;
+                                    aLoadedLines[nNextLine].bMerged = true;
```
mmeeks, line 4170: I'm pretty unhappy with arithmetic on these aLoadedLines - which should be _loadedLines - but why do we load lines ?

Surely we can do event de-duplication and merging efficiently as we stream these to the log file.

kit/ChildSession.hpp
```diff
@@ -29,6 +29,23 @@
...
+public:
+    ChildSession* aSession;
+    int nLastUndoCount = 0;
+    const StringVector* aTokens;
```
mmeeks, line 36: Glad we're using the StringVector etc. here.

kit/ChildSession.hpp
aszucs3: I now renemed it to LogUiCommands.
It was as a macro like "LOG_WRN" .. but later i had to make it a class... and i forgot to rename it.

mmeeks:
Hi Attila,

> > 'struct lineOfUiLogFile' and 'class I_AM_SHOUTING_HARD' - in the same diff - why is that ?
> 
> Sorry, i forgot this ... it started as a Macro like LOG_WRN .. and i never renamed it later

Thanks =) please do read your git diff before pushing to check for consistency =)

> I thought that the commands order are a "must keep" thing .. that is why i asked about it at my last emails. 

Fair enough; a reasonable goal - but not at expense of complexity. We should have a strong ordering per-view  - ie. if you filter on a view-id, it should be correct of course; but we can't easily do that between views. Post-sorting on time-stamp will achieve an ordering.

We really need to optimize for reduced complexity, reduced risk (particularly in coolwsd), readability & maintainability.

> But sure i know it is still more calculation as if we merge at real-time ...

State wise we only need to keep the last action per-view in memory I hope.

Thanks for iterating & polishing ! =)

kit/Kit.hpp
aszucs3: moved LogUiCmd from Kit.hpp/.cpp to logUI.hpp/.cpp

kit/Kit.cpp
aszucs3: Changed variable names notations.
the commented code was removed as it was in the post-process

kit/Kit.cpp
aszucs3: removed it as it was part of the post-process.
loading line is still present but now it is only a "std::getline(_fileStreamUICommands, line)"

loadedLines was a cache of lines preloaded from the temp file, before they saved into the permanent file.
I kept them in memory as merge needed some preloaded lines, but i made sure that every line loaded only once (and only when that was needed).. and kept memory usage at minimum possible for the merge.

aszucs3:
Hi Michael!

> Thanks =) please do read your git diff before pushing to check for consistency =)
I do, i will do a bit more times then.. unfortunatelly, im not the perfect reviewer.. as i tend to not spot some of them.. :(

> State wise we only need to keep the last action per-view in memory I hope.
Yes, mostly... 
I had to keep 2 last commands at some cases .. because of the mouse-click+mouse-down+mouse-up -> 2*mouse-click case
And i have to keep in memory the undo-state-changes... and all their datas... 
But it is ok to do.
I not saved the times of undo-state-changes ... instead, used its commands times ... to keep it a bit more simple,.. and probably it it not a big problem... and maybe it will be easier to sort them this way. (if we will want to do it at a later time)

kit/ChildSession.cpp
```diff
@@ -3791,4 +3797,218 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
...
+            {
+                if (undoAct - _lastUndoCount > 0)
+                    undoChg = 1;
+                else
+                    undoChg = -1;
+            }
+            if (commandHandled && undoChg != 0)
+            {
+                // Now, possible only if a buttonup become click
+                _session->_lastUiCmdLinesLogged[0]._undoChange += undoChg;
```
mmeeks, line 3957: This is pretty ugly; as a general thought we should not be grouping inside another object's member (at offset [0]) 's private members like this. It looks like a factoring anti-pattern on several levels.

At a very minimum we should have some shared reference for this _session->_lastUIFoo[0] and use it in this method instead of constantly duplicating this de-reference - surely ?

We are ultimately trying to generate less code fewer changes, easier to review code, with higher entropy.

common/Log.cpp
```diff
@@ -407,6 +409,14 @@ namespace Log
 
         const std::string& getLevel() const { return _logLevel; }
 
+        void setUILogMergeInfo(bool mergeUiCmd, bool logTimeEndOfMergedUiCmd)
```
mmeeks, line 412: Not entirely positive about having this functionality glupped into a set of methods with UI mangled in their name into the Logging class. Can we aggregate those as a sub-object of Log if we need to - and fetch that before calling methods that don't have UI in their name ? =) getUILog().setLogMergInfo(...) or whatever ?

mmeeks (changes requested):
Getting better; still really not happy about all these [0] array indexes. It seems the use-case for >1 item in there is rather limited - can we use some simpler approach to catch that case ? And other comments there.

Anyhow - improving =)

kit/ChildSession.cpp
```diff
@@ -3791,4 +3797,218 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
...
+                }
+            }
+            else
+            {
+                actCmd = (*_tokens)[0];
+            }
+
+            // Here we are sure we want to log this command sometime...
+            struct timeval tv;
+            gettimeofday(&tv, NULL);
```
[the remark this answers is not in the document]
aszucs3: changed them to std::chrono::steady_clock::time_point.

(i got this gettimeofday() from the normal log)

kit/ChildSession.cpp
aszucs3: thx, and sorry for it, i made some references for this long things, to solve this issue.
I left the array anyway... [1] / [i] is rarely used.. but if later some more complex log change will be needed, then we may will need more of them.

I was tought the compiler optimize them out... but sure that not helps other reviewers / developers...

common/Log.cpp
aszucs3: i made a new class inherited from the old:
static struct StaticUIHelper: StaticHelper
and there i renamed variables/methods to not have UI in it.
I hope this is what you wanted.

mmeeks (approved):
Looks so much better; would love to see it in on red & tweaked later if needed =)

PR #10534, opened 2024-11-19
Prevent fetching data for non-existing parts.

gokaysatir:
Change-Id: I107c39d35a33b9e5e95589964904a7c8f4690411

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1072,13 +1072,18 @@ L.CanvasTileLayer = L.Layer.extend({
 	},
 
 	_initPreFetchPartTiles: function() {
+		const targetPart = this._selectedPart + this._map._partsDirection;
+
+		if (targetPart < 0 || targetPart >= this._parts)
+			return;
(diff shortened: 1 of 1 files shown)


caolanm (approved):
This seems safe and fairly reasonable, but do we know how this comes about? Are things already wrong at the time of _initPreFetchPartTiles or are they ok at that launch time and then wrong/out-of-date at the time of running the timer callback?

gokaysatir:
> This seems safe and fairly reasonable, but do we know how this comes about? Are things already wrong at the time of _initPreFetchPartTiles or are they ok at that launch time and then wrong/out-of-date at the time of running the timer callback?

Yep, it's "_partsDirection" variable that is set upon key press. So it's a raw value and we need to check it somewhere. But i see, i think we could add the check higher in the call stack. Before calling the function. Let me check that.

gokaysatir:
Thanks for checking :) I moved the check into "_initPreFetchPartTiles" and tested again. Will merge when it passes CI.

PR #10535, opened 2024-11-19
Localization fixes

timar:
* Do not assemble localizable string from multiple parts
* SI prefixes are not for localization
* Enable localization of Presenter Console

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -1046,6 +1046,7 @@ pot:
         src/slideshow/engine/animationnodes/NodeContext.ts \
         src/slideshow/engine/animationnodes/PropertyAnimationNode.ts \
         src/slideshow/engine/AnimatedElement.ts \
+        src/slideshow/PresenterConsole.js \
         src/slideshow/TransitionsInfo.ts \
         src/slideshow/Transition2d.ts \
         src/slideshow/transition/ClippingTransition.ts \
--- a/browser/admin/src/Util.js
+++ b/browser/admin/src/Util.js
@@ -55,7 +55,7 @@ var Util = Base.extend({
             }
         }
         var unit = 1000;
-        var units = [_(''), ('k'), _('M'), _('G'), _('T'), _('P'), _('E'), _('Z'), _('Y'), _('B')];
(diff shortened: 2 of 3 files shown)


browser/Makefile.am
```diff
@@ -1046,6 +1046,7 @@ pot:
         src/slideshow/engine/animationnodes/NodeContext.ts \
         src/slideshow/engine/animationnodes/PropertyAnimationNode.ts \
         src/slideshow/engine/AnimatedElement.ts \
+        src/slideshow/PresenterConsole.js \
```
vmiklos, line 1049: Good catch. Do you have a bash oneliner to try to find files which use `_(...)` and are not listed in the pot target? If not, perhaps it would be good to have one, maybe this is not the only missing file.

browser/admin/src/Util.js
```diff
@@ -55,7 +55,7 @@ var Util = Base.extend({
             }
         }
         var unit = 1000;
-        var units = [_(''), ('k'), _('M'), _('G'), _('T'), _('P'), _('E'), _('Z'), _('Y'), _('B')];
+        var units = ['', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y', 'B'];
```
vmiklos, line 58: Quick research suggest that some languages would still translate these? See e.g. <https://en.wikipedia.org/wiki/Chinese_numerals#SI_prefixes>. But you probably know what you're doing here. :-)

vmiklos (approved):
Makes sense. See inline comments for nits.

browser/admin/src/Util.js
timar: `_('')` and `('k')` were wrong anyway. Thanks for the link though. I think leaving them untranslated is better that having them mistranslated by mistake.

browser/Makefile.am
timar: You are right. There is more to do, I just sent a quick fix last night to react to a bug report. I compared the list of files that should be here to the list of files that are actually here, and I detected issues. I'll clean up the list in a follow-up PR. For example many files in this list do not contain localizable strings, so they are there unnecessarily. I found some more files to add, too.

PR #10536, opened 2024-11-19
Presenter console UI: remove btn borders and update nav and time btn

pedropintosilva:
Before this commit the button radius looked odd because the size of
the nav btns or not a sqr
 - Better to don't rely on the container for visuals and instead make
 it bultin in the svg. Plus we can then remove the extra lines

Additionally update timer related buttons.

Change-Id: I6058b85f440be87a774a5bee8d0667af1d38d82d


Darshan-upadhyay1110 (approved):
Looks beautiful, Thanks Pedro :)

PR #10537, opened 2024-11-19
forward-port: fix(ios): better approximate our version format

Minion3665:
This is a forward-port of https://github.com/CollaboraOnline/online/pull/10164

Previously we were using versions like "24.04.7" for release "24.04.7.3". Unfortunately, this works poorly if we want to have a patch release, as we can't reuse codes so we have to immediately go to "24.04.8", even when releasing from a "24.04.7" tag.

Merging our year and major release into Apple's major release number makes more sense with how we expect releases to change over time...

We also need to change how we generate the CFBundleVersion, as  otherwise it will be invalidly-long (the maximum length being 18 characters and "2404.7.202410021739" being a frustrating 19 characters, for example

Change-Id: I4aa9ab53b3a4d5786e04a90ab4c7c759d9606706

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/ios/Mobile/Info.plist.in
+++ b/ios/Mobile/Info.plist.in
@@ -247,9 +247,13 @@
 
 	     Also, currently the version in configure.ac is four numbers. This
 	     CFBundleShortVersionString must, however, be three numbers.
+
+	     To avoid us getting as out-of-sync, we merge the year and major into Apple's "major"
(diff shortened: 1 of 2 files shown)


timar (approved):
I approved this on a branch, so I approve it here, too. It's ugly, but it's not very visible on the UI.

PR #10538, opened 2024-11-20
Move the Anonymizer to Own Home

Ashod:
Non-functional refactoring to move the Anonymizer from multiple files into its own home and simplifies includes greatly. Similarly, moves the Process wrappers to a new home.
This reduces clutter in overgrown files.

Improvements to better manage the memory footprint of the Anonymizer will follow.

- wsd: include cleanup
- wsd: move process wrappers into their own home
- wsd: header cleanup
- wsd: move anonymization implementation to own home


caolanm (approved):
I imagine common/Anonymizer.hpp and wsd/Process.hpp need to get listed in Makefile.am but otherwise looks alright to me

PR #10540, opened 2024-11-20
cool#9992 doc electronic sign: allow more user private info settings with make run

vmiklos:
Digital signing already supported .ca.pem / .cert.pem / .key.pem files
next to documents in the no-integration case, but electronic signing
would now bring 3 more similar settings, and this isn't really scaling.

FileServer.cpp handleWopiRequest() already builds a JSON dictionary from
these files, so it would be natural to allow any other UserPrivateInfo
keys as well in files next to local documents.

Fix the problem by reading a single .user-private-info.json file
instead, which means that setting any user private info keys is now
possible without explicit support by the local file WOPI backend.

Adapt the only cypress test that used the old format accordingly.

Change-Id: Ic833821074e622d12933c42a127e4193ade46ca8


vmiklos:

https://github.com/CollaboraOnline/online/issues/9992#issuecomment-2487771968 has a sample JSON to show how this looks like in practice, so we have something more generic and not increasing 3 specific settings to 6 ones.

caolanm (approved):

PR #10541, opened 2024-11-20
kit, BackgroundSaveWatchdog: fix memory management in the ctor

vmiklos:
By the time the thread runs, the ctor's stack is gone, so capture the
doc ID by value.

A problem since commit f27f2e297eadcf06171bfeb86ce99d717f6c73fc (wsd:
bgsave watchdog thread, 2024-11-14).

Change-Id: Iec98aeff059104b1ba321442bff98198dce5cc61

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -184,7 +184,8 @@ public:
     BackgroundSaveWatchdog(unsigned mobileAppDocId)
         : _saveCompleted(false)
         , _watchdogThread(
-              [&]()
+            // mobileAppDocId is on the stack, so capture it by value.
+              [mobileAppDocId, this]()
               {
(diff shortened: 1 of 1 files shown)


tml1024:
How doesn't the compiler notice this? Is there some static analysis tool that would have caught it?

vmiklos:
> How doesn't the compiler notice this? Is there some static analysis tool that would have caught it?

Not sure, I found this in a '-fsanitize=address -fsanitize=undefined' build that flags this at runtime; luckily we have good enough test coverage so that 'make check' hits that.

I guess this is the same root cause than what we started to see in crashreports.

caolanm (approved):
Is the commit this fixes just in master, or did it escape into co-24.04 already

PR #10544, opened 2024-11-20
Presenter console:Add tooltip to all btns

Darshan-upadhyay1110:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


hcvcastro:
Please simplify code using class src/control/Control.Tooltip.js

hcvcastro:
to follow up patch, It looks Ok

Darshan-upadhyay1110:
> Please simplify code using class src/control/Control.Tooltip.js

@hcvcastro i tried but it seems all tooltip shows on main screen and not on presenter console so for a quick thing i have to do this patch

hcvcastro (approved):

hcvcastro:
I see, perhaps you can add options to improve the use case

hcvcastro:
anyway let's focus on priority, we can improve later

Darshan-upadhyay1110:
> anyway let's focus on priority, we can improve later

Sure thanks for the review

PR #10545, opened 2024-11-20
Private/timar/l10nfix

timar:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/layer/marker/TextInput.js
```diff
@@ -439,8 +439,8 @@ L.TextInput = L.Layer.extend({
 				_('Screen reader support for text content is disabled. ') +
 				_('You need to enable it both at server level and in the UI. ') +
 				_('Look for the accessibility section in coolwsd.xml for server setting. ') +
-				_('Also check the voice over toggle under %parentControl.').replace(
-					'%parentControl',
+				_('Also check the voice over toggle under {parentControl}.').replace(
+					'{parentControl}',
```
vmiklos, line 443: Fine, but what's the difference between %foo and {foo} in this case?

If we want to be consistent, perhaps just go with {0}?

vmiklos (approved):

browser/src/layer/marker/TextInput.js
timar: PO processing tools choked on %foo, we had complaints in forums. Somehow they were mistreated as printf format strings.
We have a lot of non-{0} style placeholders in the code base, in this case I did not change it, because {parentControl} makes more sense to translators than {0}.

browser/src/layer/marker/A11yTextInput.js
```diff
@@ -226,29 +226,32 @@ L.A11yTextInput = L.TextInput.extend({
...
 		}
 		if (this._lastRowIndex !== row || this._lastRowSpan !== rowSpan) {
 			this._lastRowIndex = row;
-			eventDescription += _('Row') + ' ' + row;
 			if (this._lastRowSpan !== rowSpan) {
 				if (rowSpan > 1) {
-					eventDescription += ' ' + _('through') + ' ' + (row + rowSpan - 1) ;
+					eventDescription += _('Row {0} through {1}').replace('{0}', row).replace('{1}', row + rowSpan - 1);
 				}
 				this._lastRowSpan = rowSpan;
```
thebearon, line 237: When the code ends up here in case of:
- `this._lastRowSpan !== rowSpan` is true
- `rowSpan > 1` is false

...in the previous version of the code, `_('Row') + ' ' + row` is already appended to `eventDescription`, but not in the changed code.

browser/src/layer/marker/A11yTextInput.js
```diff
@@ -226,29 +226,32 @@ L.A11yTextInput = L.TextInput.extend({
...
 		}
 		if (this._lastColIndex !== col || this._lastColSpan !== colSpan) {
 			this._lastColIndex = col;
-			eventDescription += _('Column') + ' ' + col;
 			if (this._lastColSpan !== colSpan) {
 				if (colSpan > 1) {
-					eventDescription += ' ' + _('through') + ' ' + (col + colSpan - 1);
+					eventDescription += _('Column {0} through {1}').replace('{0}', col).replace('{1}', col + colSpan - 1);
 				}
 				this._lastColSpan = colSpan;
```
thebearon, line 250: Same as above.

thebearon (changes requested):

timar:
the iOS CI failure is unrelated

browser/src/layer/marker/A11yTextInput.js
```diff
@@ -226,29 +226,32 @@ L.A11yTextInput = L.TextInput.extend({
...
 		if (this._lastRowIndex !== row || this._lastRowSpan !== rowSpan) {
 			this._lastRowIndex = row;
-			eventDescription += _('Row') + ' ' + row;
-			if (this._lastRowSpan !== rowSpan) {
-				if (rowSpan > 1) {
-					eventDescription += ' ' + _('through') + ' ' + (row + rowSpan - 1) ;
-				}
+			if (this._lastRowSpan !== rowSpan && rowSpan > 1) {
+				eventDescription += _('Row {0} through {1}').replace('{0}', row).replace('{1}', row + rowSpan - 1);
 				this._lastRowSpan = rowSpan;
```
thebearon, line 235: On second thought, I would move
- `this._lastRowSpan = rowSpan;` - line 239 / 235
- `if (this._lastRowSpan !== rowSpan) this._lastRowSpan = rowSpan;` - line 239-240 +

...after line 241 + as unconditional `this._lastRowSpan = rowSpan;`

browser/src/layer/marker/A11yTextInput.js
```diff
@@ -226,29 +226,32 @@ L.A11yTextInput = L.TextInput.extend({
...
 		if (this._lastColIndex !== col || this._lastColSpan !== colSpan) {
 			this._lastColIndex = col;
-			eventDescription += _('Column') + ' ' + col;
-			if (this._lastColSpan !== colSpan) {
-				if (colSpan > 1) {
-					eventDescription += ' ' + _('through') + ' ' + (col + colSpan - 1);
-				}
+			if (this._lastColSpan !== colSpan && colSpan > 1) {
+				eventDescription += _('Column {0} through {1}').replace('{0}', col).replace('{1}', col + colSpan - 1);
 				this._lastColSpan = colSpan;
```
thebearon, line 248: Same here.

thebearon (changes requested):

thebearon (approved):
Looks good, thanks!

Issue #10436, opened 2024-11-07
Black flash on initial load of document


Cwiiis:
Similar to #10434, there is a black flash when loading a document sometimes. #10435 fixes #10434 but doesn't fix the root cause and so doesn't fix this entirely.

The issue, I think, is the document being redrawn (due to resizing in the sidebar case, for example) while event emission has paused drawing (which happens during tile updates). Likelihood there needs to be a check-and-defer somewhere in the resize/redraw chain while drawing is paused, or something along those lines. I'll be looking into this as the flash is ugly when observed.

There's the possibility this bug may be observed in other situations, but this is the only one I can easily observe right now. If it's observed elsewhere, please add details to this bug.

Cwiiis:
Our behaviour shouldn't change between workers/non-workers, that it flashes black is just a bug - I have some ideas about what's causing it, the flash shouldn't be there at all (and already in the case of no content, it should be white). I think because of this bug, we're occasionally flashing black even when there's content available when resizing the canvas.

PR #10548, opened 2024-11-20
Fix black flashing when resizing the canvas

Cwiiis:
* Resolves: #10436
* Target version: master 

### Summary

With workers enabled, drawing is sometimes not synchronous. This causes a problem when resizing, which clears the canvas entirely and expects tile rehydration to happen synchronously afterwards. Work around this by redrawing the old canvas contents when resizing.

I considered more comprehensive fixes, but they all added a lot of complexity and used more memory than this. I think there's probably a better possible fix here of doing less work when resizing, but how to do that wasn't immediately obvious.

I've also bundled a much smaller fix which is the more common case, where we were dumping the old canvas contents and redrawing everything even when the canvas size wasn't changed (for example, copy/paste from another open sheet causes two null resizes to happen in both sheets).

### Checklist

### DIFF
--- a/browser/src/canvas/CanvasSectionContainer.ts
+++ b/browser/src/canvas/CanvasSectionContainer.ts
@@ -1333,6 +1333,9 @@ class CanvasSectionContainer {
 		newWidth = Math.floor(newWidth * app.dpiScale);
 		newHeight = Math.floor(newHeight * app.dpiScale);
 
+		if (this.right === newWidth && this.bottom === newHeight)
+			return;
+
 		this.canvas.width = newWidth;
 		this.canvas.height = newHeight;
 
--- a/browser/src/canvas/CanvasSectionContainer.ts
+++ b/browser/src/canvas/CanvasSectionContainer.ts
@@ -1336,6 +1336,12 @@ class CanvasSectionContainer {
 		if (this.right === newWidth && this.bottom === newHeight)
 			return;
 
+		// Drawing may happen asynchronously so backup the old contents to avoid
+		// showing a blank canvas.
+		var oldImageData: ImageData = null;
+		if (this.paintedEver)
+			oldImageData = this.context.getImageData(0, 0, this.canvas.width, this.canvas.height);
+
 		this.canvas.width = newWidth;
 		this.canvas.height = newHeight;
 
@@ -1348,9 +1354,10 @@ class CanvasSectionContainer {
 		app.canvasSize.pX = newWidth;
 		app.canvasSize.pY = newHeight;
 
-		// Avoid black default background if canvas was never painted
-		// since construction.
-		if (!this.paintedEver)
+		// Avoid black default background.
+		if (oldImageData)
+			this.context.putImageData(oldImageData, 0, 0);
+		else
 			this.clearCanvas();
 
 		this.clearMousePositions();
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -757,6 +757,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		this.setSplitPosFromCell();
 		this.dontSendSplitPosToCore = false;
 
+		app.sectionContainer.reNewAllSections(true);
 		this._syncTileContainerSize();
 
 		this._map.fire('sheetgeometrychanged');


caolanm (approved):

caolanm:
I suspect the mocha_tests/CanvasSectionContainer.test.js failure in https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3334/console is triggered by this change?

Cwiiis:
> I suspect the mocha_tests/CanvasSectionContainer.test.js failure in https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3334/console is triggered by this change?

Yes, the setup helper is using onResize as an initialisation function and I guess it's calling it with the default canvas width/height so it isn't doing the things that it expects to be done. I think I probably need an initialised flag or some such...

Cwiiis:
I've changed the check to use `this.right` and `this.bottom`, which I assume are initialised to zero, instead of `this.canvas.width` and `this.canvas.height` - hopefully this will fix the tests.

Cwiiis:
Looks like the alteration fixed the problem, the test failures now look unrelated to this change.

caolanm:
I find I can reproduce (after a side journey of discovering my cypress always failed on --enable-debug https://github.com/CollaboraOnline/online/pull/10570) that the cypress failure here is triggered by the first patch.

I find that:
cd cypress_test && make check-desktop spec=calc/cell_cursor_spec.js

fails reproducibly but that moving the "this.clearMousePositions" in onResize before the early return make it pass

Cwiiis:
> I find I can reproduce (after a side journey of discovering my cypress always failed on --enable-debug #10570) that the cypress failure here is triggered by the first patch.
> 
> I find that: cd cypress_test && make check-desktop spec=calc/cell_cursor_spec.js
> 
> fails reproducibly but something like
> 
> ```
>                 if (this.right === newWidth && this.bottom === newHeight) {
>                         this.clearMousePositions();
>                         this.reNewAllSections();
>                         return;
>                 }
> ```
> 
> passes, though probably negates the point of the patch then?

Yes, certainly the `reNewAllSections()` does a lot of work - I'll take a look at the test and see what's going on. If things are relying on a null resize just refreshing a bunch of unrelated stuff, that definitely sounds like a bad thing (and if they need that refresh, they should just call refresh - which at least wouldn't unnecessarily wipe the canvas contents).

Cwiiis:
From what I can tell, the only difference with that small optimisation is now when `CalcTileLayer._handleSheetGeometryDataMsg` calls `_syncTileContainerSize`, which in turn calls `CanvasSectionContainer.onResize`, that last `onResize` gets short-circuited because the size hasn't changed. When I watch the test run, I don't see a problem, but then it's a 2-pixel difference, so I probably wouldn't... I kind of wonder if the test is well-crafted and this isn't some kind of race, but I have to assume it is and that it's a real problem. One thing I notice is that the other calls to `_syncTileContainerSize` in CalcTileLayer are preceded by a call to reNewAllSections - possibly this is just an ommission and that will fix this?

Cwiiis:
> From what I can tell, the only difference with that small optimisation is now when `CalcTileLayer._handleSheetGeometryDataMsg` calls `_syncTileContainerSize`, which in turn calls `CanvasSectionContainer.onResize`, that last `onResize` gets short-circuited because the size hasn't changed. When I watch the test run, I don't see a problem, but then it's a 2-pixel difference, so I probably wouldn't... I kind of wonder if the test is well-crafted and this isn't some kind of race, but I have to assume it is and that it's a real problem. One thing I notice is that the other calls to `_syncTileContainerSize` in CalcTileLayer are preceded by a call to reNewAllSections - possibly this is just an ommission and that will fix this?

Yes, this appears to be the case, at least locally.

Cwiiis:
I can reproduce the writer/searchbar_spec.js failure locally too, looking into it now. I hope/expect it'll be something similar.

Cwiiis:
Just debugging this test - if I short-circuit the test so that it just loads the document and types 'sit' in the search field, for whatever reason, it doesn't highlight the first match, it goes to the third. On the other hand, if short-circuit the test just before it searches and search for 'sit' myself, it does go to the first. Maybe something up with `searchHelper.typeIntoSearchField`? Or some kind of race condition on load?

Cwiiis:
I think this failure is likely a pre-existing race condition on load that the second patch in this PR exacerbates... I could be completely wrong of course :) Still digging...

caolanm:
yeah, I've seen this test fail in other prs so I believe it is preexisting.

PR #10552, opened 2024-11-20
last arg to findOrCreateDocBroker is always null

caolanm:
Change-Id: I9f542a0d4093a102c09d77f3be3cccb635d7a721

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/fuzzer/ClientSession.cpp
+++ b/fuzzer/ClientSession.cpp
@@ -23,7 +23,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
     Poco::URI uriPublic;
     std::string docKey = "/fuzz/fuzz.odt";
     auto docBroker = std::make_shared<DocumentBroker>(DocumentBroker::ChildType::Interactive, uri,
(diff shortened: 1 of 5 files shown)


caolanm:
@Ashod just pulling a loose thread here, I saw that the last arg to findOrCreateDocBroker was always null in practice and I wonder if this is some work in progress to be completed sometime or something that can be cleaned up like so.

Ashod:
> @Ashod just pulling a loose thread here, I saw that the last arg to findOrCreateDocBroker was always null in practice and I wonder if this is some work in progress to be completed sometime or something that can be cleaned up like so.

Interesting observation! Thanks for bringing to my attention. `downloadAdvance` was working. Looks like something was changed that disabled it, effectively. I'll look into it, but of course it's best to download in parallel to serving the pages, to reduce the load latency visible to the user.

mmeeks:
Curious about this - would be good to rip that out to avoid confusion if so - and to get this closed - perhaps in 25.04 ? =)

PR #10553, opened 2024-11-20
Add an exception for default prevented boolean.

gokaysatir:
Or it also prevents doocument save action.

Change-Id: I2561f274f656b5b31093227b145adf1517bc866c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
The original change of b2b637e153ac2d73c272c8284ed85e5fcffa3e16 was perhaps too naive then. Lets get this in then for now.

PR #10556, opened 2024-11-20
feat: Add user label to mention

elzody:
Change-Id: `Id2484d5dee155a7755e0cca55a97d7191ec83985`

* Target version: master 

### Summary
In Nextcloud, we are working on improving our mention support, and it would be really nice to also be able to pass a display name (label) along with a username (unique identifier) and the profile link. Having this would allow us to determine the user quicker and easier on our side, and I don't see the harm in having it there for other integrations as well, should they want it.

This is a work-in-progress PR, and it's my first contribution to this repository, so please bear with me and provide any constructive feedback you may have :)

### Checklist

### DIFF
--- a/browser/src/control/Control.Mention.ts
+++ b/browser/src/control/Control.Mention.ts
@@ -17,6 +17,7 @@
 interface MentionUserData {
 	username: string;
 	profile: string;
+	label?: string;
 }
 
 class Mention extends L.Control.AutoCompletePopup {
@@ -70,8 +71,10 @@ class Mention extends L.Control.AutoCompletePopup {
 		// filterout the users from list according to the text
 		if (text.length > 1) {
 			this.filteredUsers = users.filter((element: any) => {
+				const uid = element.label ?? element.username;
+
 				// case insensitive
-				return element.username.toLowerCase().includes(text.toLowerCase());
+				return uid.toLowerCase().includes(text.toLowerCase());
 			});
 		} else {
 			this.filteredUsers = users;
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.Mention.ts
```diff
@@ -80,10 +81,10 @@ class Mention extends L.Control.AutoCompletePopup {
 		if (this.filteredUsers.length !== 0) {
 			for (const i in this.filteredUsers) {
 				const entry = {
-					text: this.filteredUsers[i].username,
+					text: this.filteredUsers[i].label,
```
juliusknorr, line 84: Just from reading the code: Does it make sense to fallback to the username here in case the integrator does not send a label?

browser/src/control/Control.Mention.ts
```diff
@@ -15,6 +15,7 @@
 /* global app */
 
 interface MentionUserData {
+	label: string;
```
juliusknorr, line 18: As with the comment below, maybe for backward compatibility this could be nullable

browser/src/control/Control.Mention.ts
elzody: Good point! I was thinking of that at the time, but seems I missed it actually

Rash419 (approved):
Looks good, thank you. I have created SDK MR, once this patch is released in the new version. I will merge the SDK MR so it gets updated

Rash419 (changes requested):
I noticed in mobile comment we don't use ``label`` when available https://github.com/CollaboraOnline/online/blob/6af2cecd5d094a81389bc07ad32569e26abf59a2/browser/src/canvas/sections/CommentListSection.ts#L526. Otherwise, looks good : )

Rash419 (approved):
Thanks : )

PR #10562, opened 2024-11-21
Remove duplicate sending of copy and cut commands.

gokaysatir:
_doCopyCut function is already sending the relevant command.

Also saw that _sendCommandAndWaitForCompletion returns a promise that is not needed anymore (after Skyler's refactorings).

So we can remove that and use a boolean value for pursuing the state of cut/copy commands.

This fixes (Calc):
* Cut text.
* Paste it.
* It pastes empty text - because of double issuing of cut command, second time it cuts empty text.

Change-Id: I0d4a2b0637892234a7dab68638f26cacadd8bb11

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -827,12 +827,13 @@ L.Clipboard = L.Class.extend({
 
 	_sendCommandAndWaitForCompletion: function(command) {
 		if (command !== '.uno:Copy' && command !== '.uno:Cut') {
-			throw `_sendCommandAndWaitForCompletion was called with '${command}', but anything except Copy or Cut will never complete`;
+			console.error(`_sendCommandAndWaitForCompletion was called with '${command}', but anything except Copy or Cut will never complete`);
+			return null;
 		}
 
 		if (this._commandCompletion.length > 0) {
-			throw 'Already have ' + this._commandCompletion.length +
-				' pending clipboard command(s)';
+			console.warn('Already have ' + this._commandCompletion.length + ' pending clipboard command(s)');
+			return null;
 		}
 
 		app.socket.sendMessage('uno ' + command);
@@ -870,7 +871,10 @@ L.Clipboard = L.Class.extend({
 
 	_asyncAttemptNavigatorClipboardWrite: async function() {
 		const command = this._unoCommandForCopyCutPaste;
-		await this._sendCommandAndWaitForCompletion(command);
+		const check_ = await this._sendCommandAndWaitForCompletion(command);
+
+		if (check_ === null)
+			return; // Either wrong command or a pending event.
 
 		// This is sent down the websocket URL which can race with the
 		// web fetch - so first step is to wait for the result of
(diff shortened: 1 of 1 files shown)


browser/src/map/Clipboard.js
```diff
@@ -877,8 +851,10 @@ L.Clipboard = L.Class.extend({
 	},
 
 	_asyncAttemptNavigatorClipboardWrite: async function() {
-		const command = this._unoCommandForCopyCutPaste;
-		await this._sendCommandAndWaitForCompletion(command);
+		if (this._commandCompletion)
+			throw 'Already have ' + this._commandCompletion.length + ' pending clipboard command(s)';
+		else
+			this._commandCompletion = true;
```
Minion3665, line 857: do you ever send the command here? It looks like you remove `_sendCommandAndWaitForCompletion`, possibly without realizing that I moved the `app.socket.sendMessage` line into it...

browser/src/map/Clipboard.js
```diff
@@ -877,8 +851,10 @@ L.Clipboard = L.Class.extend({
 	},
 
 	_asyncAttemptNavigatorClipboardWrite: async function() {
-		const command = this._unoCommandForCopyCutPaste;
-		await this._sendCommandAndWaitForCompletion(command);
```
Minion3665, line 881: We definitely do need to wait for the command to complete somehow or we come across some data races... I don't mind if you do it through some other promise, but it looks like you're not going to be waiting at all after this PR?

From your PR description I suspect you believed the promise was usused here as there's no value resolved into it? That's not true: the promise is awaited to make sure the side-effects of running the uno command are completed... see also https://github.com/CollaboraOnline/online/commit/1b3c5bfd1be0ad77882452a4ddddd8e4ef5fe807 which ran into this issue and moved (and duplicated) fetches to avoid it before the refactor

browser/src/map/Clipboard.js
```diff
@@ -38,8 +38,8 @@ L.Clipboard = L.Class.extend({
 		this._dummyPlainDiv = null;
 		this._dummyClipboard = {};
 
-		// Tracks waiting for UNO commands to complete
-		this._commandCompletion = [];
+		// Is there an awaiting command.
+		this._commandCompletion = false;
```
Minion3665, line 42: This is a fair idea (stopping this from being an array as it can only have one item), but if you want to have it as a promise then you need to save the resolve/reject somewhere ... and you do need it to be a promise or similar because otherwise you can't wait for it ... and you do need to wait for it (at least on mobile) because otherwise you run into data races that lead to data being miscopied

Minion3665 (changes requested):

browser/src/map/Clipboard.js
Minion3665: From your PR description you mention
> _doCopyCut function is already sending the relevant command

but I think the function is also run from `_execCopyCutPaste`, where I think the command is only run here... so I guess this eliminates a duplicate for the _doCopyCut case but I'm worried it might break the _execCopyCutPaste case in the process

gokaysatir:
There is a duplicate call to cut command.

There is also a tricky part for handling the command response. I will fix also that tricky code path.

Thanks for checking.

Minion3665 (approved):
I believe this breaks mobile copy/paste again, but we should have time to fix mobile copy/paste before the next release... and it certainly makes the desktop side better. Don't backport this to 24.04-mobile.

PR #10564, opened 2024-11-21
fix(mobile): Style comment property on word count

Minion3665:
In core change Ib94c8db5beb88d0b71035081680ce25a74774d08, we are adding support for the comments property in the word count dialog. Not styling this shows the property name and its value on separate lines

Change-Id: I12e7e07f59a5f30e51b65066adee5c90394a3991

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -1277,7 +1277,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #mobile-wizard #wordslabel2,
 #mobile-wizard #inclspaceslabel2,
 #mobile-wizard #exclspaceslabel2,
-#mobile-wizard #cjkcharsft2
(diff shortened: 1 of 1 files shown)


browser/css/jsdialogs.css
```diff
@@ -1277,7 +1277,8 @@ input[type='number']:hover::-webkit-outer-spin-button {
 #mobile-wizard #wordslabel2,
 #mobile-wizard #inclspaceslabel2,
```
eszkadev, line 1278: don't we have already "id/name of the dialog" in the mobilewizard structure? (I indoriduced containers for separate "windows" at some point)

then we could just create a generic rule here

eszkadev (approved):
looks good, we can do a followup if my coment is true

PR #10567, opened 2024-11-21
Revert "prevent double-dispatch of uno events bound to function keys"

caolanm:
and its follow up fix for ctrl+s. Which is a bit of a bodge. So revert the two together and have another round at this later.

The goal was to enable "F3" to do something which isn't needed yet.

This reverts commit b2b637e153ac2d73c272c8284ed85e5fcffa3e16.

Change-Id: I79869dc3e07c86edf9f99bb6100b9e6dcb5b5698

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -400,11 +400,6 @@ L.Map.Keyboard = L.Handler.extend({
 		if (this._map.uiManager.isUIBlocked())
 			return;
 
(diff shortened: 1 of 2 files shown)


caolanm:
Given the problem that my original patch causes, it looks safer to go back to the way it was before that.

I don't have a pressing need for my fix, which was to experiment with using F3. And I can have a second look as to why it causes the original problem.

gokaysatir (approved):
Thanks :)

PR #10571, opened 2024-11-22
Consolidate Reaping Helper and Reap on SIGCHLD

Ashod:
This adds logic to reap zombie children on SIGCHLD, which doesn't leave zombies around until the next background save (which would leak if bg-save got disabled).

To avoid adding yet another reaping logic, we consolidate the implementations into a neat helper that also logs.

- wsd: sigaction's sa_handler takes int
- wsd: extract reapZombieChild and reuse
- wsd: reap bg-save on SIGCHLD

### DIFF
--- a/common/SigUtil.cpp
+++ b/common/SigUtil.cpp
@@ -18,11 +18,12 @@
 #  include <execinfo.h>
 #endif
 #include <csignal>
-#include <sys/types.h>
-#include <sys/stat.h>
 #include <fcntl.h>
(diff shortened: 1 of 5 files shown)


mmeeks (approved):
Looks reasonable for a green release - lets not port it to COOL yet though.

Thanks Ash. Reading this again - it looks like we don't count SEGV's of our bgsave processes: perhaps that's a good thing - they will eventually crash the main process I suppose.

PR #10575, opened 2024-11-22
fix(calc): Improve grid colors in dark mode

juliusknorr:
Change-Id: Ie379efb16f5e82de8d53203b88fdccad15e8da37

* Resolves: # 
* Target version: master 

### Summary

This allow to adjust the calc grid line color through css variables. It also sets a different value for dark mode to make the contrast ratio with the background similar to light mode.

Change might be subtle but makes calc in dark mode look much nicer as discussed at some point already with @pedropintosilva and @jancborchardt

One issue seems still that a reload is required, on switching to dark mode the grid color is not picked up again. I'm not sure what the best way would be to update the sectionProperties values in that case. Maybe someone has a pointer there?

| Before | After |
|---|---|
| image | image |
| image | image |

### Checklist

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -50,6 +50,8 @@
 	--color-box-shadow: rgba(0, 0, 0, 0.5);
 	--color-hyperlink: #1d99f3;
 
+	--color-calc-grid: #474747;
+
 	--color-grid-helper-line-solid: #e6e6e6;
 	--color-grid-helper-line-dashed: #191919;
 	--color-smart-guides-helper-line: #f36d4f;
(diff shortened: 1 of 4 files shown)


pedropintosilva (approved):

pedropintosilva:
Pressed Retry test

juliusknorr:
Could be a follow up but that would still be interesting:

> One issue seems still that a reload is required, on switching to dark mode the grid color is not picked up again. I'm not sure what the best way would be to update the sectionProperties values in that case. Maybe someone has a pointer there?

gokaysatir (approved):
Looks beautiful, thanks for the improvements :)

gokaysatir:
> Could be a follow up but that would still be interesting:
> 
> > One issue seems still that a reload is required, on switching to dark mode the grid color is not picked up again.

I added one commit for the follow up, i hope that's ok:)

Re-triggered the tests.

juliusknorr:
Great, thanks for that 👍

PR #10580, opened 2024-11-22
treeview: missing columns

eszkadev:
jsdialog: treeview: put dummy columns at correct position
    
    It was a regression from commit aee68f732003d02bdf7c9b0768df8ca0a3880123
    browser: treeview: align listbox with headers using subgrid
    
    If we used autofilter in calc which has icons at the end of an entry,
    we got dummy column at the beginning of the row...
    
    This adds preprocessing of columns so we always have all of them
    unified (the same number of cols in every row) with dummy ones at
    the correct position. Exception is separator which has to be only
    one in the row to be full width.


eszkadev:
passed but change id check didn't work

gokaysatir (approved):
Thanks for the fix.

PR #10584, opened 2024-11-22
WIP scroll on selecting cells with keyboard

JaumePujantell:
When selecting multiple cells with shift + down arrow, if the selection
goes out of the view area it should scroll to keep the newly selected
cells into view.

Change-Id: I30aac5e3595482c7e43adff189cc47fed4310b59


vmiklos (approved):
The lack of vertical scrolling is indeed fixed with this, thanks.

But could you please explain why replacing the "pY1, then pY2" check with a "pY2, then pY1" check fixes the problem? Ideally this kind of info would go to a comment or to the commit message (just a tip for next time).

Also, there is now an unexpected horizontal scrolling, could you please investigate that? See this video:


https://github.com/user-attachments/assets/4315c560-d961-495e-944f-ed4cee6839af

Once you fix that, could you could also assert "no change on the horizontal scrollbar" in your testcase.

All in all I guess this is an improvement, but I would recommend fixing that followup horizontal scroll problem before moving on to a different topic. Thanks.

JaumePujantell:
> But could you please explain why replacing the "pY1, then pY2" check with a "pY2, then pY1" check fixes the problem? 

The problem was introduced when changing LatLong to rectangles in commit cfeba70dab1a1012285c880183fe64548120d380 getNorth was changed to pY1 and getSouth to pY2 and it should have been the other way around.

> there is now an unexpected horizontal scrolling

Yes, I'm looking into it, it doesn't seem to have to do with this changed comparisons so it will be in another PR.

vmiklos:
Ah, thanks for documenting the "why", makes sense. Next time please add that to the commit message and it'll be perfect. :-)

PR #10587, opened 2024-11-22
Private/timar/config

timar:
Some cleanup wrt. welcome and home_mode.


vmiklos (approved):
I don't see anything problematic with these changes, thanks.

PR #10592, opened 2024-11-25
jsdialog: treeview: simplify missing columns injection

eszkadev:
it was possible to skip some column before due to index calculation, code is also simpler now


gokaysatir (approved):
Thanks for the improvement :)

PR #10593, opened 2024-11-25
cool#9992 doc electronic sign: send the hash to be signed

vmiklos:
Once the hash is extracted from the document, we need to send it to the
3rd-party for signing.

Given that we don't want to send the entire document, our signature
container should not be an entire PDF file, we just want this for the
signature binary itself (PKCS#7 blob), set the file type to "cades",
which does exactly this.

This requires passing the API URL, and credentials to
L.Control.ESignature.

Finally allow communicating with these hosts in the CSP headers.
Initially I attempted to take the URL from CheckFileInfo, but we seem to
decide this when serving cool.html, so that order doesn't allow making
the CSP headers depend on the CheckFileInfo response. In practice only 2
public servers implement this protocol (production and test), so just
allow them for now, similar to Zotero. If there is a need for a custom
server later, it can be always added to net.content_security_policy in
coolwsd.xml.

Change-Id: If1dfa50defb37205e54b2f3708f7bdaeab22afce

### DIFF
--- a/browser/src/control/Control.ESignature.ts
+++ b/browser/src/control/Control.ESignature.ts
@@ -19,12 +19,30 @@ namespace cool {
 		commandValues: SignatureResponse;
 	}
 
+	export interface HashSendResponse {
+		doc_id: string;
+	}
+
 	/**
 	 * Provides electronic signing with document hashes for PDF files.
 	 */
 	export class ESignature {
-		constructor() {
-			app.map.on('commandvalues', this.onCommandValues);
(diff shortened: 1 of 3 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3436/console failed in:

> unit-save-torture

but

```
./run-wsdunit unit-save-torture
```

passes for me, retry.

caolanm (approved):

PR #10594, opened 2024-11-25
ci(docker): Remove cache from nightly container build

juliusknorr:
Otherwise the cache would also contain the sources we build online from so only updates once the cache is invalidated.

Build time is still reasonable with 12 minutes as only ran once a day

Signed-off-by: Julius Knorr <jus@bitgrid.net>
Change-Id: I96d57cb75e48f2560ceae6ebdd5213478b03efee

* Target version: master 

### Checklist


juliusknorr:
@pedropintosilva Should address that we sometimes have a slightly outdated build on the daily instance we use for note taking

timar (approved):

PR #10596, opened 2024-11-25
slideshow: on key down handler was not removed on exit

mcecchetti:
the problemed occurred in fullscreen for the nextcloud integration

Change-Id: Iebaa76757c047dc3bd1f7badb4118e67c59e0ae7


eszkadev (approved):
looks ok

PR #10599, opened 2024-11-25
Fix show all comment button problems

GulsahKose:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/control/Control.Menubar.js
```diff
@@ -1848,10 +1849,19 @@ L.Control.Menubar = L.Control.extend({
 						} else {
 							$(aItem).removeClass('disabled');
 						}
+					} else if (id === 'showannotations') {
+						var section = app.sectionContainer.getSectionWithName(L.CSections.CommentList.name);
+						if (section) {
```
eszkadev, line 1854: can we receive the state just from the stateHandler as in other places? (one way to do the same thing):

like: https://github.com/CollaboraOnline/online/blob/dc930ffa6d93c42148ce5aa47af01faed7826e0b/browser/src/docdispatcher.ts#L611

eszkadev (approved):
looks ok

PR #10600, opened 2024-11-25
Fix incorrect order of message arguments

mikekaganski:
The 'pictureurl' message is handled in ChildSession::contentControlEvent,
which relies on the order of the arguments.
In commit 60d9a9c20bf69b3f1d8591b2d8400c4a453970ab (Implement remote
multimedia insertion support, 2024-11-19), I incorrectly modified the
order, trying to unify it between nearby calls. Fix it here.

Change-Id: I612973e1b8aa245a601d3053e516b67007482a69

### DIFF
--- a/browser/src/map/handler/Map.FileInserter.js
+++ b/browser/src/map/handler/Map.FileInserter.js
@@ -210,7 +210,8 @@ L.Map.FileInserter = L.Handler.extend({
 		}
 
 		if (e.urltype == "graphicurl" && section && section.sectionProperties.picturePicker) {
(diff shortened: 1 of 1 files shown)


vmiklos (approved):
Looks OK, thanks.

Nits:

- Did you consider covering this change with a test? I see that e.g. dropdowns on mobile have a test at cypress_test/integration_tests/mobile/writer/insert_content_control_spec.js, perhaps you could add an image test there?
- This bug is somewhat self-explaining (possibly only to me, since I worked on content controls earlier), but in general please either file a github issue and refer to that in the commit message or include the steps to reproduce the problem in the commit message.

Thanks. :-)

PR #10603, opened 2024-11-26
cool#9992 doc electronic sign: actually sign the hash in a popup

vmiklos:
Open an PDF in Draw, Insert -> Electronic signature, the doc hash is
sent to ESignatureBaseUrl, but we don't attempt to sign it.

The actual signing happens using a HTML page provided by eIDEasy, so
open that in a popup.

We can avoid polling for the signature result by listening for
postmessages: the signing popups sends us a postmessage on completion.

This is the same mechanism that's used by the SignatureJS npm library,
developed by eIDEasy (but don't use that here, since it would pull in a
number of dependencies since it doesn't use the fetch() API).

Change-Id: I10902ec6868a64f250fe5afdd0cf9ba36a2f63e6

### DIFF
--- a/browser/src/control/Control.ESignature.ts
+++ b/browser/src/control/Control.ESignature.ts
@@ -23,6 +23,11 @@ namespace cool {
 		doc_id: string;
 	}
 
+	export interface SignedResponse {
+		type: string;
+		error: string;
+	}
+
 	/**
 	 * Provides electronic signing with document hashes for PDF files.
 	 */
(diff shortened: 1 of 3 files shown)


vmiklos:

If you want to try it, see https://github.com/CollaboraOnline/online/issues/9992#issuecomment-2499927954 for a test config. Initially I thought I need to poll the server if we have a result or need to wait more, but turns out there is a postmessage we can listen to, which I consider nice.

caolanm (approved):

PR #10605, opened 2024-11-26
slideshow: prevent slide sorter to receive key events on fullscreen mode

mcecchetti:
Apart from changing the current slide in the editor,
this causes the onUpdateParts to be triggered which in turn request presentation
info again even if that's not needed.

Change-Id: I15efd3d205c7b03a76d04ff0e8f4f22b07bac00e


eszkadev (approved):
looks ok

PR #10609, opened 2024-11-27
cool#9992 doc electronic sign: fetch the created signature

vmiklos:
Sign a hash with electronic signing, we get a postmessage that the
signing is done on the external site, but we still don't have the
signed hash.

There is a /api/signatures/download-signed-file endpoint to fetch the
PKCS#7 blob, but that is only meant to succeed once we know the signing
happened.

Handle this by fetching the signature once we got a postmessage that
the signing was fine.

The blob we get still needs integrating, though.

Change-Id: Id0dcdf5a6518c6b71ab02f1f77a3736038546821

### DIFF
--- a/browser/src/control/Control.ESignature.ts
+++ b/browser/src/control/Control.ESignature.ts
@@ -28,6 +28,11 @@ namespace cool {
 		error: string;
 	}
 
+	export interface ReceiveSignatureResponse {
+		status: string;
+		signed_file_contents: string;
+	}
+
 	/**
 	 * Provides electronic signing with document hashes for PDF files.
 	 */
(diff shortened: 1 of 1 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2919/console fails in:

> calc/searchbar_spec.js

but a local

```
make -C cypress_test -sr check-desktop spec=calc/searchbar_spec.js
```

passes, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2922/console again fails in:

> calc/searchbar_spec.js

but turns out my local baseline was newer when I ran cypress (vs the one I used for pushing), rebase.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/2924/console still fails in:

> calc/searchbar_spec.js

hmm.

vmiklos:

If the above cypress problem starts to be a pattern, we should probably investigate, but it's not something I see locally. Apart from that, only a last step is meant to be missing for the "have 1 working scenario" of this e-sign thing.

caolanm (approved):
I have seen calc searchbar a few times, chris attempted a fix recently I think

PR #10612, opened 2024-11-27
Check canvas size before copy in CanvasSectionContainer.onResize

Cwiiis:
Don't try to call getImageData if the canvas width or height are zero, as this will result in an IndexSizeError.


mmeeks (approved):
Thanks Chris =)

timur-g:
Ref: related to  https://github.com/CollaboraOnline/online/pull/10548.

PR #10613, opened 2024-11-27
Clipboard write: Don't consider "Document is not focused." as an error.

gokaysatir:
It causes the process to think that clipboard API has failed. Then we don't try again with the clipboard API. This commit logs a warning and skips the fallback part.

Change-Id: I4f04feb3bf77ea986453733a6c8976ef3d0ac2a4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -901,15 +901,20 @@ L.Clipboard = L.Class.extend({
 			try {
 				await clipboard.write([clipboardItem]);
 			} catch (error) {
-				window.app.console.log('navigator.clipboard.write() failed: ' + error.message);
-
-				// Warn that the copy failed.
-				this._warnCopyPaste();
-				// Once broken, always broken.
-				L.Browser.clipboardApiAvailable = false;
-				window.prefs.set('clipboardApiAvailable', false);
-				// Prefetch selection, so next time copy will work with the keyboard.
-				app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');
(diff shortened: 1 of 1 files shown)


browser/src/map/Clipboard.js
```diff
@@ -901,15 +901,20 @@ L.Clipboard = L.Class.extend({
...
-
-				// Warn that the copy failed.
-				this._warnCopyPaste();
-				// Once broken, always broken.
-				L.Browser.clipboardApiAvailable = false;
-				window.prefs.set('clipboardApiAvailable', false);
-				// Prefetch selection, so next time copy will work with the keyboard.
-				app.socket.sendMessage('gettextselection mimetype=text/html,text/plain;charset=utf-8');
+				// "Document is not focused." is an error but it shouldn't stop the usage of clipboard API.
+				if (error.message.includes('Document is not focused.')) {
```
eszkadev, line 905: hmm, it seems to be DOMException: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#notallowederror

I assume thete is only that generic code available, and no other "identifier". We have to compare message?

eszkadev: Did you check different browsers? Do they have the same message? or it is some specific case?

eszkadev: or maybe we can detect `lack of focus` ?

gokaysatir: Yep, checking if document has focus is a better way. Thanks for checking.

eszkadev (approved):
looks more stable now :)

PR #10614, opened 2024-11-27
Calc: Set spreasheet header colors from CSS vars

pedropintosilva:
Note: this commit doesn't touch or fix the problem with the corner
header being always white even if on dark mode. That is outside of the
scope of this patch.

Better to create dedicated css vars instead of rely on other generic
CSS vars such as background-color-darker etc. this way we avoid
calc regressions when someone changes those more generic values.

Also avoid setting the fill of header-row and header-column-selected
to the primary color. If the integrator wants to set that they are
free to overwrite that but better to default to neutral color that
looks ok on both light and dark mode.

Change-Id: I3c6d0b7a8e58ae8c239bda97d35454cdc70826fe

### DIFF
--- a/browser/css/color-palette-dark.css
+++ b/browser/css/color-palette-dark.css
@@ -51,6 +51,11 @@
 	--color-hyperlink: #1d99f3;
 
 	--color-calc-grid: #474747;
+	--color-calc-header: #0a0a0a;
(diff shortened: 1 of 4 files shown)


gokaysatir (approved):
Thanks for the improvements.

PR #10621, opened 2024-11-28
private/jaume/calc scroll drift

JaumePujantell:
- fix typo in geometry.ts
- fix scroll drift on cell selection


cypress_test/integration_tests/desktop/calc/scrolling_spec.js
```diff
@@ -31,6 +31,7 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Scroll through document',
 
 	it('Scroll while selecting vertically', function() {
 		desktopHelper.assertScrollbarPosition('vertical', 19, 21);
+		desktopHelper.assertScrollbarPosition('horizontal', 48, 50);
```
vmiklos, line 34: Nit: it would be nice to have a way to just remember the scrollbar position & later assert that it didn't change without hardcoding the values here; but perhaps that's not easy, because the current approach has this auto-retry that only works if you state the expected value explicitly.

vmiklos (approved):
Works fine for me, thanks. :-)

PR #10383, opened 2024-10-31
Comment popup: Use Ctrl+Enter shortcut to save and close

meven:
Change-Id: I8da6e1208b0cbba6b51da97c6926d29da19a47a4

### Summary

Use Shift+Enter to save and close the comment popup for convenience.
This is a common shortcut for such functionnality.

### Checklist

- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #10622, opened 2024-11-28
calc: allow to use shortcut ctrl+Enter in comments

meven:
this.map.mention can be unset in particular in calc.

Amends  https://github.com/CollaboraOnline/online/pull/10383

Change-Id: I8dba5d36f20d6fe3bf0df9b9fa15a98b70b51eaa

* Resolves: # 
* Target version: master 

### Summary

In calc > insert a  comment > type some text > hit Ctrl+Enter

Expected:
The popup closes, the comment is saved.

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -436,7 +436,7 @@ export class Comment extends CanvasSectionObject {
 	private textAreaKeyDown (ev: any): void {
 		if (ev && ev.ctrlKey && ev.key === "Enter") {
 			ev.preventDefault();
-			this.map.mention.closeMentionPopup(false);
+			this.map.mention?.closeMentionPopup(false);
 			this.onSaveComment(ev);
 			return;
 		}


hfiguiere (approved):

vmiklos (approved):
Looks good, thanks.

Nits:

- it would be good to be more verbose in the commit message. Ideally you would describe what's the purpose of this 'map.mention' (that's the at-completion in Writer, correct?) in Writer. Also describe how this failed (I assume this is always undefined in Calc/Impress?). Finally describe how the new state is meant to keep the old use-case fixed, while fixing the new use-case.
- Would it be possible to cover this fix with a cypress test?

meven:
> it would be good to be more verbose in the commit message. Ideally you would describe what's the purpose of this 'map.mention' (that's the at-completion in Writer, correct?) in Writer. Also describe how this failed (I assume this is always undefined in Calc/Impress?). Finally describe how the new state is meant to keep the old use-case fixed, while fixing the new use-case.

Too late to complete now.

> Would it be possible to cover this fix with a cypress test?

I think so. Especially easy if inserting comment is already tested.

PR #10631, opened 2024-11-29
Reset autofill mark on close autofill context menu

GulsahKose:
When we click outside of the autofill context menu, menu is closed but marked area doesn't reset.

We fix the problem with the patch

Change-Id: I819d1b0393d1f062570c5b1d4776c32ae90aa887

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -204,7 +204,11 @@ L.Control.ContextMenu = L.Control.extend({
 						var $menu = opt.$menu;
 						$menu.attr('tabindex', 0); // Make the context menu focusable
 					},
-					hide: function() { map.focus(); }
+					hide: function() {
+						if(autoFillContextMenu)
+							app.map._docLayer._resetReferencesMarks();
+						map.focus();
+					}
 				}
 			});
 			if (autoFillContextMenu)
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -3027,6 +3027,12 @@ L.CanvasTileLayer = L.Layer.extend({
 		this._references.clear();
 	},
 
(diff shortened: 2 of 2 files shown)


browser/src/control/Control.ContextMenu.js
```diff
@@ -214,6 +215,14 @@ L.Control.ContextMenu = L.Control.extend({
 			$('.context-menu-root').focus();
 			this.hasContextMenu = true;
 		}
+		const contextMenuLayer = document.getElementById('context-menu-layer');
+		contextMenuLayer.addEventListener('mousedown', this._onOutsideMouseDown, false);
```
eszkadev, line 219: we add new listeners but never remove them - it might stack events with time

browser/src/control/Control.ContextMenu.js
```diff
@@ -214,6 +215,14 @@ L.Control.ContextMenu = L.Control.extend({
 			$('.context-menu-root').focus();
 			this.hasContextMenu = true;
 		}
+		const contextMenuLayer = document.getElementById('context-menu-layer');
```
eszkadev, line 218: Please try to avoid using direct usage of generated DOM elements by some libraries.
I think context menu has own mechanics to add "on click outside" handler - when we create it

eszkadev: [ browser/src/control/Control.ContextMenu.js ](https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.ContextMenu.js#L207)

eszkadev: did you check that? hide thing?

GulsahKose: I missed that. Thank you Szymon. I updated the patch

eszkadev (approved):
looks good for me, thanks!

PR #10632, opened 2024-11-29
Revert "browser: fix PageDown cursor position"

eszkadev:
This reverts commit a074a6f199c0594afd173fe3638b72d1d649b3ff. browser: fix PageDown cursor position

This causes randomm jumps on tile rendering when we need to move view to render. It sends additional curosr updates which are not correct (like 0,0,0,0 and scroll: true).

We already have logic here to scroll if needed, we shouldn't need any information from core.

see: https://gerrit.libreoffice.org/c/core/+/177520


gokaysatir (approved):

gokaysatir:
@hcvcastro fyi you might want to recheck the page down issue :)

Issue #10634, opened 2024-12-02
Error on state change for tracked change


eszkadev:
### Describe the Bug
There was an error in the online reported by user:
![image](https://github.com/user-attachments/assets/628b9de7-61b0-4123-b2ac-2ca828da44e1)

But it still seemed to work.

### Steps to reproduce
1. open writer
2. review tab -> activate "record"
3. do some changes in document
4. turn off "record"
5. review tab -> "next" / "previous"

I quickly checked the code and it fils in: https://github.com/CollaboraOnline/online/blob/9663a3f75e992dbd09776ca52e46dd6192837cab/browser/src/map/handler/Map.StateChanges.js#L50

Interesting is that `_annotation` is not set enywhere. Quick grep shows:
```
browser/src/layer/tile/CanvasTileLayer.js:              if (this._annotations) {
browser/src/layer/tile/CanvasTileLayer.js:                      var annotations = this._annotations;
browser/src/layer/tile/CanvasTileLayer.js:              if (this._docType === 'spreadsheet' && this._annotations !== undefined) {
browser/src/map/Map.js:         if (this._docLayer && this._docLayer._annotations && this._docLayer._annotations._items) {
browser/src/map/Map.js:                 for (var idxAnno in this._docLayer._annotations._items) {
browser/src/map/Map.js:                         var annotation = this._docLayer._annotations._items[idxAnno];
```

I guess it should be replaced with proper get-section thing:
```
const annotations = app.sectionContainer.getSectionWithName(L.CSections.CommentList.name);
if (annotations) annotations.doSomething()
```

Probably we forgot to remove few usages of that `_annotations` ?

eszkadev:
cc: @gokaysatir @lpranam you might know

lpranam:
let me have a look

Ezinnem:
Not reproducible in:
COOLWSD version: 24.04.12.3(git hash: 733474e0d4 (E))
LOKit version: Collabora Office 24.04.12.3(git hash: 65a518b5f0)
Served by: Ubuntu 22.04.5 LTS
Server ID: 1773a1a3
WOPI host: nextcloud 30.0.0.14 - richdocuments 8.5.3

Pehaps the patch from [#10635](https://github.com/CollaboraOnline/online/pull/10635) fixed it

PR #10635, opened 2024-12-02
Avoid undefined in CurrentTrackedChangeId

eszkadev:
Related to issue #10634


gokaysatir (commented):
Yes it seems _annotations is not set. It'd be better if we can check its other usages. It seems to be used in a few more places. Do we need them or can we remove them?

We can have a follow up for other usages or continue with this PR if you wish.

eszkadev:
> Yes it seems _annotations is not set. It'd be better if we can check its other usages. It seems to be used in a few more places. Do we need them or can we remove them?
> 
> We can have a follow up for other usages or continue with this PR if you wish.

I reported it in https://github.com/CollaboraOnline/online/issues/10634
we need to scan these and do a followup

gokaysatir (approved):
Okay thank you for the fix.

PR #10637, opened 2024-12-02
cool#9992 doc electronic sign: select the provider interactively

vmiklos:
- **cool#9992 doc electronic sign: select the provider interactively**
- **browser: try to speed up incremental builds**


vmiklos:

The first completes the provider selector UI, so it's now part of the signing flow. The second fixes a problem that somehow an incremental build took 2 minutes here when touching a TS file, which looked very odd.

caolanm (approved):

Issue #10167, opened 2024-10-03
Following user pop up should automatically close


pedropintosilva:
### Describe the Bug
It should automatically close when the user clicks in one the user names

### Steps to Reproduce
1. Users Clicks in the avatars
2. Then clicks in one of the users
3. pop up doesn't close and stays in the way of the document **while the focus is actually on document**

### Expected Behavior
It should close automatically

### Version

COOLWSD version: 24.04.8.1(git hash: 53851545 (E))
LOKit version: Collabora Office 24.04.8.1(git hash: 4baad22)
Served by: openSUSE Leap 15.5
Server ID: f2379670
WOPI host: nextcloud 30.0.0.14 - richdocuments 8.5.0

pedropintosilva:
To test this you need multiple users (you can also open the same document in different web browsers tabs, private tabs etc)

![image](https://github.com/user-attachments/assets/a4a79487-3f47-4028-bdaf-88d5c81e2d5d)

Pressing Anna should automatically set that and close the popup

pedropintosilva:
The userlist/following UI control is located at browser/src/control/Control.UserList.ts

pedropintosilva:
for any mentoring please ping @szyklos he can help you with any further code pointers

PR #10638, opened 2024-12-02
Ensure popup closes automatically on user selection

banobepascal:
Change-Id: I4e7a06bab413dad0b5f2360c4bef729db7d5e0d0

* Resolves: #10167 
* Target version: master 

### Summary
Fixes a UX bug that disrupts user workflow by leaving the selection popup open after a user is chosen, blocking underlying content.

### TODO
Users Clicks in the avatars
Pop up closes when one of the users is clicked 

### Checklist

- ✅ I have run `make prettier-write` and formatted the code.
- ✅  All commits have Change-Id
- ✅  I have run tests with `make check`
- ✅ I have issued `make run` and manually verified that everything looks okay

### DIFF
--- a/browser/src/control/Control.UserList.ts
+++ b/browser/src/control/Control.UserList.ts
@@ -520,6 +520,7 @@ class UserList extends L.Control {
 			listItem.appendChild(userLabelContainer);
 			listItem.addEventListener('click', () => {
 				this.followUser(viewId);
+				JSDialog.CloseDropdown('userlist');
 			});
 
 			return listItem;
--- a/cypress_test/integration_tests/multiuser/calc/cell_cursor_spec.js
+++ b/cypress_test/integration_tests/multiuser/calc/cell_cursor_spec.js
@@ -16,7 +16,7 @@ describe(['tagmultiuser'], 'Check cell cursor and view behavior', function() {
 		cy.cSetActiveFrame('#iframe2');
 		cy.cGet('#userListHeader').click();
 		cy.cGet('.user-list-item').eq(1).click();
-		cy.cGet('.jsdialog-overlay').click({force: true});
+		cy.cGet('.jsdialog-overlay').should('not.exist');
 
 		// first view goes somewhere in the middle of a sheet: A400
 		cy.cSetActiveFrame('#iframe1');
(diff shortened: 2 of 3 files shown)


pedropintosilva:
Thanks @banobepascal @eszkadev and I will take a look. Meanwhile could you please fix the DCO check but following the instructions at: image

banobepascal:
> Thanks @banobepascal @eszkadev and I will take a look. Meanwhile could you please fix the DCO check but following the instructions at: image

Hey @pedropintosilva, I'm working on this now

browser/src/control/Control.UserList.ts
```diff
@@ -520,6 +520,7 @@ class UserList extends L.Control {
 			listItem.appendChild(userLabelContainer);
 			listItem.addEventListener('click', () => {
 				this.followUser(viewId);
+				JSDialog.CloseDropdown('userlist');
```
eszkadev, line 523: looks good :)

eszkadev (approved):

pedropintosilva:
it seems the failing test is integration_tests/multiuser/calc/cell_cursor_spec.js which might be a false negative . Ideally you would re-run the test locally to see if it really fails due to your change or if it's the test that is unstable (if it's unstable we can just press the button retry by: clicking the details url and then log in with your Collabora credentials)    

paste from the CI logs:

```javascxript

**Comment by pedropintosilva** [comment]
meanwhile I have retried the CI to see if it keeps failing, but please @banobepascal check the test locally

**Comment by banobepascal** [comment]
> * make -C cypress_test run-multi spec=calc/cell_cursor_spec.js

@pedropintosilva I'm running the tests locally, I'm going to provide an update

**Comment by eszkadev** [comment]
This test seems to be related `.jsdialog-overlay` is used for popups. Here we close popup which wasn;t closed before. It might be expected by the test (what should be modified now)

**Comment by banobepascal** [comment]
> This test seems to be related `.jsdialog-overlay` is used for popups. Here we close popup which wasn;t closed before. It might be expected by the test (what should be modified now)

We can add an explicit wait or assertion to ensure the popup is visible before attempting to close it as this will give the application time to render the popup before the test interacts with it.

**Comment by pedropintosilva** [comment]
So, that line 
```javascript
		cy.cGet('.jsdialog-overlay').click({force: true});
```

that is in that test twice should just be removed.

Alternatively we would assert that '.jsdialog-overlay' doesn't exist thus, introducing a test for your improvement via `.should('not.exist');`

PR #10642, opened 2024-12-02
Private/caolan/consistent var naming

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
This changes some variable names from core-style semi-hungarian notation to the scheme typically used in online for consistency.

vmiklos (approved):
Makes sense, we don't do that sort of thing here in Online. FWIW there is https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html if you want to find these systematically, I haven't tried it myself, though.

PR #10643, opened 2024-12-02
Revert "UI: replace conditional format dialog with conditiona manager"

lpranam:
This reverts commit 1e096ee31200c4f9e815f59034e6e721466f7f4f.

Change-Id: I173b6b88b7621bc46df4579a5de0816d6942e821

* Target version: master 

Core patches are reverted corresponding to this commit

### Checklist


eszkadev:
In case of revert best to also write in the message why it is needed, what problems were introduced by wrong patch

eszkadev (approved):
I didn't test but code looks ok and it's just a revert

PR #10644, opened 2024-12-02
Add cypress tests for shortcut to save comments

meven:
Using Ctrl+Enter in calc and writer.

See dce46718f7be40d5266743547fd88b26fa485e4b

Context: https://github.com/CollaboraOnline/online/pull/10622#pullrequestreview-2471953765

Change-Id: Ic7e947623aac6345e4ad9c92a9454dfb5b2a4c2e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


meven:
> you make my poor live a bit easier if you only request review when CI passed (and not before), so I can also merge after approval :-)

Sorry, since CI is very slow, this is a better practice indeed.

> good to partition your test code to given (setup), when (here: do the ctrl-enter), then (assert) parts, so when it fails, it's more obvious to the next person if the interesting part of the test failed as intended (to catch badness) or it's something else

I followed the test existing construction but indeed, Given, When Then makes things a lot easier to maintain, and show an example to improve the other tests in those files.

Thanks

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3010/console failed in calc/annotation_spec.js, seems that include your new test, if you search for "Modify and Save using shortcut Ctrl+Enter". Do you have a guess what's the problem there?

meven:
> https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3010/console failed in calc/annotation_spec.js, seems that include your new test, if you search for "Modify and Save using shortcut Ctrl+Enter". Do you have a guess what's the problem there?

Not yet, I have been investigating.

Writer is the same.
The weird part is it worked before, and I didn't change code, so it seems spurious, still I want it stable. 

I have been investigating:
```
                    
                    Timed out retrying after 20000ms: Expected to find element: `#document-canvas`, but never found it. Queried from:
                    
                                  > cy.get(#coolframe, [object Object]).its(0.contentDocument, [object Object])
                    
                    /home/collabora/jenkins/workspace/github_online_master_debug_vs_co-24.04_cypress_desktop/cypress_test/support/index.js:181:1
                      179 | 		return cy.get(cy.cActiveFrame, {log: false})
                      180 | 			.its('0.contentDocument', {log: false})
                    > 181 | 			.find(selector, optionsWithLogFalse);
                          | ^
                      182 | 	} else {
                      183 | 		return cy.get(cy.cActiveFrame, optionsWithLogFalse)
                      184 | 			.its('0.contentDocument', {log: false});
          cy:log ✱  Finishing test: integration_tests/desktop/calc/annotation_spec.js / Annotation Tests / Modify and Save using shortcut Ctrl+Enter
```

eszkadev:
Requires some fixes, cypress faisl seems to be related.
Please also rebase as now cypress is more stable in general.

meven:
I have https://github.com/cypress-io/cypress/issues/20891 making it harder on me to make progress on this.

vmiklos:
@meven  I'm a bit confused, I expect a number of other people like @caolanm or @quikee are already on wayland, without issues regarding running cypress tests locally (I'm still on x11). What is the specific problem you're running into?

PR #10645, opened 2024-12-02
Generate calc sample file calc-edit.fods from a light template.

mmeeks:
Hopefully this makes it easy to maintain in git, without bloating the repository with a large volume of data, while providing flexibility to change things in the future.

Change-Id: I474ead8aa90cea85cf38999c986631b557dcd06d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/scripts/build-samples.py
+++ b/scripts/build-samples.py
@@ -11,7 +11,9 @@
 #
 
 import os
+import re
 import sys
+import random
 import shutil
 
 def usageAndExit():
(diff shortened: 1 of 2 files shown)


mmeeks:
@thebearon you might be interested in this - it is not perfect; but it is /way/ better than nothing. Should be easy  to iterate too without bloating the repo. Sample overview needs a few more shapes, some validation bits - etc. but we don't need huge blocks of data, and tons of sheets to test lots of features I think.

caolanm (approved):

PR #10648, opened 2024-12-03
Return true if filterExecCopyPaste succeeds for pasteSpecial.

gokaysatir:
Or it opens 2 instances of the popup.

Change-Id: Id091775cef6b7f162c9ffd99d119a3d7c64085da

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/Clipboard.js
+++ b/browser/src/map/Clipboard.js
@@ -1061,7 +1061,7 @@ L.Clipboard = L.Class.extend({
 			this._execCopyCutPaste('paste', cmd);
 		} else if (cmd === '.uno:PasteSpecial') {
 			if (this._navigatorClipboardRead(true)) {
-				return;
+				return true;
 			}
 			this._openPasteSpecialPopup();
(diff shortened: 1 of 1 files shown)


vmiklos:
@gokaysatir could you please make sure CI passes first? Maybe you just need to retry the desktop and multi-user tests. Thanks.

vmiklos (approved):
Looks OK. Please next time have a reference to a github issue or describe in the commit message what is the problem you're fixing. It turns out that the steps to reproduce this are:

1. Open Calc, go to a cell with a formula
2. Right click, Copy
3. Go to an other cell
4. Right click, Paste special

Actual result: 2 paste special dialogs are shown, which is visible if you drag away the appearing paste special dialog a little.
Expected result: only one paste special dialog appears.

Also, do you think you could add a test for this in cypress_test/integration_tests/desktop/calc/clipboard_spec.js? Thanks.

PR #10650, opened 2024-12-03
Impress: Don't activate draggin if text cursor is visible.

gokaysatir:
When selecting a portion of text, ShapeHandlesSection's dragging is activated. This causes the shape to move while selecting a portion of text. Check the visibility of the cursor and don't start dragging event if cursor is visible.

Change-Id: I9a7c8917125cc3c32ffd14c81ab07bf491b7158d

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected

PR #10653, opened 2024-12-04
Disable "Presenter Console" option when "Present in Window" is clicked

Darshan-upadhyay1110:
- Fire "state change command" to disable toolbar option for "Present Console" when we already in "Present in Window".

This ensures the "Presenter Console" option is disabled while the presentation is in window mode.

Change-Id: I65d37187be901a073a83cfcdd987fbb2e14bc896

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -62,7 +62,6 @@ L.Control.Notebookbar = L.Control.extend({
 		this.map.on('darkmodechanged', this.onDarkModeToggleChange, this);
 		this.map.on('showannotationschanged', this.onShowAnnotationsChange, this);
 		this.map.on('a11ystatechanged', this.onAccessibilityToggleChange, this);
-		this.map.on('enablecommand', this.disableNotebookbarControl, this);
 		if (docType === 'presentation') {
 			this.map.on('updateparts', this.onSlideHideToggle, this);
 			this.map.on('toggleslidehide', this.onSlideHideToggle, this);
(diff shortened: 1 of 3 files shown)


eszkadev:
I think this approach if half-done - we block the action, but does that show the status in the UI (greyed-out icon) ?
I propose to use our `app.map['stateChangeHandler']` which will store state for command. Then JSDialog automatically will disable action and grey out the icon everywhere for the given command.

example: `this.fire('commandstatechanged', {commandName : 'showannotations', state : on ? 'true': 'false'});`

commands are: `presenterconsole` and `presentinwindow`

Could you check?

Darshan-upadhyay1110:
> I think this approach if half-done - we block the action, but does that show the status in the UI (greyed-out icon) ? I propose to use our `app.map['stateChangeHandler']` which will store state for command. Then JSDialog automatically will disable action and grey out the icon everywhere for the given command.
> 
> example: `this.fire('commandstatechanged', {commandName : 'showannotations', state : on ? 'true': 'false'});`
> 
> commands are: `presenterconsole` and `presentinwindow`
> 
> Could you check?

[Screencast from 2024-12-04 19-29-36.webm](https://github.com/user-attachments/assets/808f1a3e-4ca8-4e97-ae31-92de2fff8047)

Done :)

browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -721,7 +727,12 @@ class SlideShowPresenter {
 		if (!this._onPrepareScreen(true))
 			// opens full screen, has to be on user interaction
 			return;
-
+		// disable present in console if on present in window executed
+		this._map.fire('commandstatechanged', {
```
eszkadev, line 731: this code is duplicated, can we create method: `enablePresenterConsole(enable: boolean)` ?

browser/src/control/Control.Notebookbar.js
```diff
@@ -349,14 +348,6 @@ L.Control.Notebookbar = L.Control.extend({
 		this.reloadShortcutsBar();
 	},
 
-	disableNotebookbarControl: function(e) {
```
eszkadev, line 352: good catch!

browser/src/slideshow/SlideShowPresenter.ts
Darshan-upadhyay1110: Sure

eszkadev (approved):
looks good :)

PR #10654, opened 2024-12-04
cool#10630 doc electronic sign: add /cool/signature endpoint

vmiklos:
- **cool#10630 doc electronic sign: more human-readable names for providers**
- **cool#10630 doc electronic sign: add /cool/signature endpoint**
- **kit: fix clang 18 -Werror,-Wvla-cxx-extension**


vmiklos:

Naming is hard, I went with /cool/signature to be consistent with e.g. /cool/clipboard, which looked better than something complicated like /cool/after-signing-finished-successfully, which becomes a problem when you want to reuse the same endpoint for a few more situations inside signing in the future. For the clang19 fixes, I'm aware of two more problems, then we should be warning-free (the fuzzer job uses --disable-werror at the moment).

kit/Delta.hpp
```diff
@@ -242,7 +242,7 @@ class DeltaGenerator {
 
         void initRow(const uint32_t *from, unsigned int width)
         {
-            uint32_t scratch[width];
+            auto scratch = static_cast<uint32_t*>(alloca(sizeof(uint32_t) * width));
```
caolanm, line 245: Or I guess we could turn the warning off?

caolanm (approved):

kit/Delta.hpp
vmiklos: That's also an option, though searching github, it seems some other big projects just adapt similarly, https://github.com/dotnet/diagnostics/commit/d592e0da173e520ef42d4f9cdd9d7b138a9896fd is a sample .NET commit e.g.

PR #10655, opened 2024-12-04
Add a cypress test for paste-special command.

gokaysatir:
Also add one more class name to context menu in order to be able to distinguish it amongst other pre-defined context menus.

Change-Id: I592e60c828cd516bcd56d558be97d4e1090ac0f6

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -179,7 +179,7 @@ L.Control.ContextMenu = L.Control.extend({
 		} else {
 			L.installContextMenu({
 				selector: '.leaflet-layer',
-				className: 'cool-font',
+				className: 'cool-font on-the-fly-context-menu',
 				trigger: 'none',
 				zIndex: 1500,
 				build: function() {
--- a/cypress_test/integration_tests/desktop/calc/clipboard_spec.js
+++ b/cypress_test/integration_tests/desktop/calc/clipboard_spec.js
@@ -187,4 +187,36 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Calc clipboard tests.', fu
 
 		cy.cGet('#copy-paste-container').should('have.text', url);
 	});
+
+	it('Paste-Special', function () {
+		helper.setDummyClipboardForCopy();
+
+		calcHelper.clickOnFirstCell();
+		helper.typeIntoDocument('Something to copy paste.');
+		helper.typeIntoDocument('{enter}');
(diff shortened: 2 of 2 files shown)


gokaysatir:
This is a follow up for: https://github.com/CollaboraOnline/online/pull/10648

browser/src/control/Control.ContextMenu.js
```diff
@@ -179,7 +179,7 @@ L.Control.ContextMenu = L.Control.extend({
 		} else {
 			L.installContextMenu({
 				selector: '.leaflet-layer',
-				className: 'cool-font',
+				className: 'cool-font on-the-fly-context-menu',
```
vmiklos, line 182: Nit: maybe "dynamic-context-menu" or "automatic-context-menu"? But sure, naming is hard. :-)

cypress_test/integration_tests/desktop/calc/clipboard_spec.js
```diff
@@ -187,4 +187,36 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Calc clipboard tests.', fu
...
+
+		cy.cGet('#modal-dialog-paste_special_dialog-box').should('be.visible');
+
+		cy.cGet('#modal-dialog-paste_special_dialog-box-yesbutton').should('be.visible');
+		cy.cGet('#modal-dialog-paste_special_dialog-box-yesbutton').click();
+
+		cy.cGet('#PasteSpecial').should('be.visible');
+		cy.cGet('#ok').click();
+
+		cy.cGet('#formulabar').should('contain.text', 'Something to copy paste.');
```
vmiklos, line 220: Maybe assert here that #PasteSpecial is no longer visible? That was the primary problem visible to the user, that we had 2 dialogs, so clicking OK still left the user with a dialog.

vmiklos (approved):
Looks good in general, thanks.

In your future tests, it's useful to set up the code to given (setup code), when (do the action that went wrong in the past) and then (do your asserts) parts; which is most useful when the test fails and we want to understand if the test caught the badness it has to (the "when" part fails, so it's dangerous to just "update" the test), vs the other cases where the setup or assert has to be done differently after some refactor.

cypress_test/integration_tests/desktop/calc/clipboard_spec.js
gokaysatir: Ah thanks, let me add this soon, thanks for the review :)

PR #10656, opened 2024-12-04
private/pedro/status bar permission mode mobile vs desktop

pedropintosilva:
- **Desktop Status bar and Mobile top bar: clarify Edit mode label**
- **Mobile: badges should be subtle**
- **Mobile: Do not display permission mode label when in edit mode**

### DIFF
--- a/browser/src/control/jsdialog/Widget.HTMLContent.ts
+++ b/browser/src/control/jsdialog/Widget.HTMLContent.ts
@@ -50,7 +50,7 @@ function getPermissionModeElements(
 	} else {
 		permissionModeDiv.classList.add('status-edit-mode');
 		permissionModeDiv.title = _('Permission Mode');
-		permissionModeDiv.textContent = _('Edit');
+		permissionModeDiv.textContent = _('Edit mode');
 	}
 
 	return permissionModeDiv;
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -48,6 +48,11 @@ html[dir='rtl'] .sub-menu-arrow {
 	background: url('images/lc_listitem-selected.svg') no-repeat center / 28px !important;
 }
 
+/* Permisson mode status */
+.mobile .ui-badge {
+	background-color: transparent;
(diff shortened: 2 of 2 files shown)


browser/src/control/jsdialog/Widget.HTMLContent.ts
```diff
@@ -50,7 +50,7 @@ function getPermissionModeElements(
 	} else {
 		permissionModeDiv.classList.add('status-edit-mode');
 		permissionModeDiv.title = _('Permission Mode');
-		permissionModeDiv.textContent = _('Edit');
+		permissionModeDiv.textContent = _('Edit mode');
```
Minion3665, line 53: Can we make this say something that makes it clear this is the state of the document not an action? I'm worried that "Edit mode" might still sound too much like "Enter edit mode" whereas something like "Editable" or "Can be edited" would not suffer from this problem

Minion3665 (changes requested):

browser/src/control/jsdialog/Widget.HTMLContent.ts
Minion3665: ahh, wait, this is hidden in edit mode anyway? Maybe I misunderstand this PR - why are we changing the text?

Minion3665: I realized this works on desktop too, where "Edit mode" is probably fine, apologies for the noise

pedropintosilva: Ah I see you already realized that. But even in the hypothetical case that we would show that in mobile (which we don't with this PR):  "Editable" seems quite not the term to use because it conveys that you can edit but not really the mode you are in and that's what we are trying to communicate with that label. That together with the fact that is less familiar jargon to the user (and more dev/machine jargon) + that is a passive term makes it really not the best option at least in this context for COOL, I think.

gokaysatir (approved):
Code looks good.

PR #10658, opened 2024-12-04
fix automake warnings

timar:
test/Makefile.am:362: warning: 'TEST_EXTENSIONS' cannot have conditional contents
- I removed the conditional from the test/Makefile.am and added conditional whether to descend into test/ subdirectory to the main Makefile.am. I did the same for cypress_test/Makefile.am, although that one did not cause warning.

test/Makefile.am:217: warning: variable 'unit_admin_la_SOURCES' is defined but no program or test/Makefile.am:217: library has 'unit_admin_la' as canonical name (possible typo) test/Makefile.am:218: warning: variable 'unit_admin_la_LIBADD' is defined but no program or test/Makefile.am:218: library has 'unit_admin_la' as canonical name (possible typo)
- unit_admin_la is commented out, so I commented out these lines, too

test/Makefile.am:305: warning: variable 'unit_synthetic_lok__la_LIBADD' is defined but no program or test/Makefile.am:305: library has 'unit_synthetic_lok__la' as canonical name (possible typo)
- This one was a typo indeed, double underscore '_' character.

Change-Id: Ia166363e7e725570881fb7d7ca94d534ce2f8286

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,7 +14,15 @@ endif
 
 else
 
-SUBDIRS = . browser test cypress_test
+SUBDIRS = . browser
+
+if ENABLE_TESTS
+SUBDIRS += test
+endif
+
+if ENABLE_CYPRESS
+SUBDIRS += cypress_test
+endif
 
 export ENABLE_DEBUG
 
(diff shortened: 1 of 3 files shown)


vmiklos (approved):
Makes sense; I imagine we didn't use any cppunit assert in unit-synthetic-lok, that's why the typo wasn't discovered.

PR #10662, opened 2024-12-05
Private/gokay/hyperlink 4 12

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/URLPopUpSection.ts
+++ b/browser/src/canvas/sections/URLPopUpSection.ts
@@ -102,7 +102,11 @@ class URLPopUpSection extends HTMLObjectSection {
 		}
 
 		document.getElementById(this.copyButtonId).onclick = () => {
-			app.map.sendUnoCommand('.uno:CopyHyperlinkLocation', params);
+			// If _navigatorClipboardWrite is available, use it.
+			if (L.Browser.clipboardApiAvailable || window.ThisIsTheiOSApp)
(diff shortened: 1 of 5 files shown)


eszkadev:
rebased after cypress fix - let's try now

browser/src/canvas/sections/URLPopUpSection.ts
```diff
@@ -102,7 +102,11 @@ class URLPopUpSection extends HTMLObjectSection {
 		}
 
 		document.getElementById(this.copyButtonId).onclick = () => {
-			app.map.sendUnoCommand('.uno:CopyHyperlinkLocation', params);
+			// If _navigatorClipboardWrite is available, use it.
+			if (L.Browser.clipboardApiAvailable || window.ThisIsTheiOSApp)
+				app.map._clip.filterExecCopyPaste('.uno:CopyHyperlinkLocation', params);
+			else // Or use previous method.
+				app.map.sendUnoCommand('.uno:CopyHyperlinkLocation', params);
```
eszkadev, line 109: can't we move that `if` into `sendUnoCommand`?
That will make us sure we will not forget it when we add new call to that uno command.
Also makes less code needed :)

https://github.com/CollaboraOnline/online/blob/8add6a52b5c850e27d4545494e17d8fec617fe4e/browser/src/control/Toolbar.js#L357

eszkadev (approved):
Thanks for fix, now copy the link from popup works

PR #10668, opened 2024-12-05
Misc Improvements

Ashod:
- wsd: test: assert correct access-token
- wsd: do map lookup only once
- wsd: const and deduplication in HostUtil
- wsd: logging alias group mapping
- wsd: consistent logging
- wsd: maybe_unused in verifyWOPISrc

### DIFF
--- a/wsd/HostUtil.cpp
+++ b/wsd/HostUtil.cpp
@@ -25,7 +25,7 @@ std::set<std::string> HostUtil::hostList;
 std::string HostUtil::FirstHost;
 bool HostUtil::WopiEnabled;
 
-void HostUtil::parseWopiHost(Poco::Util::LayeredConfiguration& conf)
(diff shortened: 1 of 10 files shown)


caolanm (approved):

wsd/HostUtil.cpp
```diff
@@ -141,15 +143,14 @@ void HostUtil::parseAliases(Poco::Util::LayeredConfiguration& conf)
                 {
                     continue;
                 }
-                const std::string host = aliasUri.getHost();
 
-                std::vector<std::string> strVec = Util::splitStringToVector(host, '|');
-                const Poco::URI realUri(uri);
-                for (auto& x : strVec)
+                for (const std::string& x : Util::splitStringToVector(aliasUri.getHost(), '|'))
```
meven, line 147: `x` is not great as a variable name, even for a loop temporary.

PR #10671, opened 2024-12-05
debug frame: Limit the messages element in size

hfiguiere:
Change-Id: I1f43d4839341e9d1acac617cc6a6a96e2118a51f

* Resolves: # 
* Target version: master 

### Summary

In the debug frame, if you let it run it would push the COOL frame down the page as the post message grew.

This does two things:
- add an overscroll limiting the height of the logs
- add a clear button to clear it up.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/html/framed.doc.html
+++ b/browser/html/framed.doc.html
@@ -63,6 +63,16 @@
           margin: 0px;
       }
 
+      #messages {
+          height: 5vh;
+          overflow: scroll;
+      }
+
+      #messages > p {
+              margin: 0;
+              font-family: monospace;
+      }
+
     </style>
 
     <script>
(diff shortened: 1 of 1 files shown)


browser/html/framed.doc.html
```diff
@@ -300,7 +318,9 @@
         }
 
         let messageOut = document.getElementById("messages");
-        messageOut.textContent = messageOut.textContent + JSON.stringify(msg) + "\n";
+        let line = document.createElement("p");
```
meven, line 321:
```suggestion
        let line = document.createElement("pre");
```
Using <pre> should allow not to use `font-family: monospace`

eszkadev:
please rebase to pass CI

browser/html/framed.doc.html
hfiguiere: but using pre would make everything overflow impossible to wrap.

meven: Using p with monospace is fine.

hfiguiere:
CI is green now.

meven (approved):

PR #10672, opened 2024-12-06
cool#10630 doc electronic sign: send the hash on the server

vmiklos:
- **cool#10630 doc electronic sign: expose the name of the document**
- **wsd: fix clang 18 -Werror,-Wvla-cxx-extension**
- **cool#10630 doc electronic sign: send the hash on the server**

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -428,6 +428,90 @@ void ClientSession::onTileProcessed(TileWireId wireId)
         LOG_INF("Tileprocessed message with an unknown wire-id '" << wireId << "' from session " << getId());
 }
 
+#if !MOBILEAPP
+namespace
+{
+std::shared_ptr<http::Session> makePrepareSignatureSession(const std::shared_ptr<ClientSession> clientSession, const std::string& requestUrl)
+{
+    // Create the session and set a finished callback
+    std::shared_ptr<http::Session> httpSession = http::Session::create(requestUrl);
+    if (!httpSession)
+    {
+        LOG_WRN("PrepareSignature: failed to create HTTP session");
+        return nullptr;
+    }
+
+    http::Session::FinishedCallback finishedCallback = [clientSession](const std::shared_ptr<http::Session>& session)
+    {
+        const std::shared_ptr<const http::Response> httpResponse = session->response();
+        Poco::JSON::Object::Ptr resultArguments = new Poco::JSON::Object();
+        resultArguments->set("commandName", ".uno:PrepareSignature");
+
+        bool ok = httpResponse->statusLine().statusCode() == http::StatusCode::OK;
+        resultArguments->set("success", ok);
+
+        const std::string& responseBody = httpResponse->getBody();
+        Poco::JSON::Object::Ptr responseBodyObject = new Poco::JSON::Object();
+        if (!JsonUtil::parseJSON(responseBody, responseBodyObject))
+        {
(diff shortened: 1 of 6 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_android_app_co-24.04/2810/console fails because I forgot that it seems our mobile apps don't do any networking, so just put that inside a mobile ifdef for now. Perhaps as a next step it'll be useful to add a wsd/ClientSession-$PLATFORM.cpp similar to e.g. common/Util-mobile.cpp, to avoid the ifdefs.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3661/consoleFull failed in unit-storage and unit-storage but both pass for me locally, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3662/consoleFull hmm, same failure here, retry.

vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3112/console had 5 failures: impress/undo_redo_spec.js, writer/annotation_spec.js, writer/navigator_spec.js, writer/scrolling_spec.js and writer/searchbar_spec.js, retry to see if this is persistent.

vmiklos:
Ah, I could not reproduce the failure locally because my core was too old: https://gerrit.libreoffice.org/c/core/+/177957 is meant to help here.

vmiklos:
After rebase, https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04/3691/console fails due to a timeout, hmm.

vmiklos:

The point is that the credentials are a pair of client ID + secret, so secret is needed to initiate a sign (which has costs; results in a "document ID"), but the browser can do the signing without the secret now. This will also allow undoing the CSP rules for that specific domain, which I always disliked.

The CI is still in a bad shape, but this one at least passed now.

wsd/ClientSession.cpp
```diff
@@ -428,6 +428,90 @@ void ClientSession::onTileProcessed(TileWireId wireId)
...
+
+    // Make the request: this requires a JSON body, where we set the secret
+    std::string commandArguments = tokens.cat(' ', 2);
+    Poco::JSON::Object::Ptr commandArgumentsObject;
+    if (!JsonUtil::parseJSON(commandArguments, commandArgumentsObject))
+    {
+        LOG_WRN("PrepareSignature: failed to parse arguments as JSON");
+        return false;
+    }
+    auto requestBodyObject = commandArgumentsObject->get("body").extract<Poco::JSON::Object::Ptr>();
```
caolanm, line 498: FWIW I think (didn't verify though) that get + extract will throw if "body" doesn't exist, while getObject("body") will return null to allow an explicit test for its existence. Not sure if that unlikely case matters here, and if it does then maybe that throw is perfectly acceptable.

caolanm (approved):

wsd/ClientSession.cpp
vmiklos: This is something we control, so I would say it's OK if this throws in the unlikely case that our JS code at browser/src/control/Control.ESignature.ts:230 produces an unexpected JSON.

But you have a point here, I'll see if I can rework this to use JsonUtil::findJSONValue() instead, which won't throw, even if the key is missing. Thanks.

PR #10674, opened 2024-12-06
Private/caolan/use data method

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
Given the issue in https://github.com/CollaboraOnline/online/commit/5ad65b0e1977c9c3321bf8a24752ceb493c483e4 I'd like to change &foo[0] generally to foo.data() to make that the common pattern even when a specific site is guaranteed to be non-empty

vmiklos (approved):
Makes sense to me, thanks.

PR #10675, opened 2024-12-06
On presenter console window reload close the slideshow

Darshan-upadhyay1110:
Thanks. It looks good

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -139,9 +139,11 @@ class PresenterConsole {
 		this._map.on('transitionend', this._onTransitionEnd, this);
 		this._map.on('tilepreview', this._onTilePreview, this);
 
-		this._computeCanvas(
-			this._proxyPresenter.document.querySelector('#current-presentation'),
-		);
+		// safe check for current-presentation element
+		const currentPresentationCanvas =
+			this._proxyPresenter.document.querySelector('#current-presentation');
+		if (!currentPresentationCanvas) return;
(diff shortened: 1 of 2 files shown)


browser/src/slideshow/PresenterConsole.js
```diff
@@ -1056,7 +1058,6 @@ class PresenterConsole {
 		);
 		this._proxyPresenter.clearInterval(this._timer);
 
-		delete this._proxyPresenter;
 		delete this._currentIndex;
```
mcecchetti, line 1061: Not deleting `_proxyPresenter` on window close causes problems. 
[presenter-console-error.webm](https://github.com/user-attachments/assets/e4690a18-36ac-4b94-b52f-8ba2c811e3fb)

The point is that even once closed `PresenterConsole` still receives presentationinfo msg.
Giving a glance to `PresenterConsole._onPresentationInfo` is clear that the patch changes this method behavior. We end up setting listeners as the one for `_onTransitionStart` where `_previews` is initialized but `_previews` has been deleted in `_onConsoleClose`.

I see that the problem here is that `unload` doesn't mean `closed` so I appended the following code fragment at the beginning of `_onConsoleClose` which does the job without the need to preserve `_proxyPresenter` instance btw a presenter console invocation and the next one.
```
		if (!this._proxyPresenter.closed) {
			this._proxyPresenter.close(); // not trigger _onConsoleClose again
			setTimeout(this._onConsoleClose.bind(this), 100);
			return;
		}

```

hcvcastro (approved):

mcecchetti (approved):
Seems to work now, thanks

PR #10676, opened 2024-12-06
fix(tooltips): Don't display on touch events

Minion3665:
When we used w2-ui, we hid tooltips on mobile devices and tablets. When
we moved to using JQuery's tooltips we didn't carry this over, causing
persistent tooltips when people were touching mobile devices. This was
not ideal.

Rather than hiding these on mobile devices we can show them only on
mouse events, since as you can use mobile devices without a touchscreen
or have touchscreen desktop devices (commonly touchscreen laptops).
We've just added support for touch event detection of the mouseenter
events that we need for this. We don't want to lock touchleave events to
mouse, as we want to close the tooltip if we start tapping on other
things.

### DIFF
--- a/browser/css/device-mobile.css
+++ b/browser/css/device-mobile.css
@@ -430,11 +430,6 @@ textarea.cool-annotation-textarea {
 	overflow-y: auto !important;
 }
 
-/* Disable tooltips when on touch devices */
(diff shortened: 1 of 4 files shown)


eszkadev (approved):
Works nice, tested with touch screen laptop

PR #10680, opened 2024-12-06
Check LayerRenderer is disposed before drawBitmap

GulsahKose:
When user stopped presentation early, online still tries to draw slides because of delayed "slidelayer" messages from core. That is the cause of following error

TypeError: Cannot read properties of null (reading 'useProgram')

So we shold check if the user exited from presentation before drawing slides.

Change-Id: If3b3cb340cfc15632de16d0fed1c730ab4d9d763

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
Please rebase, tests were fixed

mcecchetti (approved):
I suspect more work is needed here, anyway it's surely an improvement. Thanks :-)

PR #10681, opened 2024-12-07
anyInput - should only depend on incoming messages from coolwsd.

mmeeks:
Not on outgoing messages that have come from core.

We should not treat rendering / tile requests as input.

Change-Id: I0ced5ee0f07796e8aae015f31c91cb7845443ea8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3024,34 +3024,14 @@ int pollCallback(void* data, int timeoutUs)
 #endif
 }
 
+// Do we have any pending input events from coolwsd ?
+// FIXME: we could helpfully poll our incoming socket too here.
 bool anyInputCallback(void* data)
 {
     auto kitSocketPoll = reinterpret_cast<KitSocketPoll*>(data);
     std::shared_ptr<Document> document = kitSocketPoll->getDocument();
-    if (!document)
-    {
-        return false;
-    }
 
-    if (document->hasCallbacks())
-    {
-        return true;
-    }
-
-    std::shared_ptr<KitQueue> queue = document->getQueue();
-    if (!queue)
-    {
-        return false;
-    }
-
-    if (queue->getTileQueueSize() > 0)
-    {
-        return true;
-    }
-
-    // Have no pending callbacks and the tile queue is also empty, report that we have no
-    // pending input events.
-    return false;
+    return document && document->hasQueueItems();
 }
 
 /// Called by LOK main-loop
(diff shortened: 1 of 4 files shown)


vmiklos:
I think I added this because the order of events were like this before the change: layout visible area, invalidate visible area, layout the rest, release SolarMutex and let COOL render a tile; so with a large enough document and large enough invalidation we laid out the entire document before trying to paint the visible area.
After the change, the order of events were: layout visible area, invalidate visible area, paint the tiles for the visible area, to the rest of the layout.
So I think the reason to handle "pending tile requests" as input was to give the visible area a priority. But I'll update my optimized build tree and see if indeed the 500ms hang after paste is back with this or my concern is not relevant in practice.
Otherwise the change makes sense to me, just trying to avoid unwanted side-effects.

vmiklos (approved):
Looks good; I tested that the "let's render the visible area before the full doc layout" case continues to work with this:

before:
13:39:35.820 global.js:564 1733747975820 OUTGOING: uno .uno:Paste 
...
13:39:35.829 global.js:564 1733747975829 INCOMING: delta: nviewid=1000 part=0 width=256 height=256 tileposx=7680 tileposy=0 tilewidth=3840 tileheight=3840 oldwid=1 wid=58 ver=58 imgsize=1
=> 9 ms

after:
13:45:14.641 global.js:564 1733748314641 OUTGOING: uno .uno:Paste
...
13:45:14.652 global.js:564 1733748314652 INCOMING: delta: nviewid=1000 part=0 width=256 height=256 tileposx=7680 tileposy=0 tilewidth=3840 tileheight=3840 oldwid=1 wid=89 ver=93 imgsize=1
=> 11 ms

So the cost is the ~same, the old 500ms cost is not returning.

kit/Kit.cpp
```diff
@@ -748,7 +748,7 @@ Document::Document(const std::shared_ptr<lok::Office>& loKit, const std::string&
     , _docId(docId)
     , _url(url)
     , _obfuscatedFileId(Uri::getFilenameFromURL(docKey))
-    , _queue(std::make_shared<KitQueue>())
+    , _queue(new KitQueue())
```
Ashod, line 751: If we always create a `KitQueue` and it seems never reset or re-assign it, perhaps we don't need a pointer at all. This removes an indirection and the many `_queue &&`.

Ashod (approved):
Thanks, @mmeeks. One suggestion, but looks good otherwise.

eszkadev:
rebased

PR #10683, opened 2024-12-07
Private/mmeeks/priority

mmeeks:
Not ready - a WIP - but trying to de-bong various prioritization bugs I discovered when reading the code here =)

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1252,17 +1252,10 @@ float ChildSession::getTilePriority(const TileDesc& tile) const
     // FIXME: recent interactivity - should bump priority
     // FIXME: proximity to an editing cursor should bump priority
     // FIXME: proximity to a viewing area should bump priority
+
+    // FIXME: if tile.isPreview() -> lower priority ...
+
     return tile.intersects(_clientVisibleArea) ? 1.0 : 0.0;
-}
-
-bool ChildSession::isTileInsideVisibleArea(const TileDesc& tile) const
-{
-    return tile.intersects( _clientVisibleArea );
-}
-
-bool ChildSession::isTileInsideVisibleArea(const TileCombined& tileCombined) const
-{
-    return tileCombined.toAABBox().intersects( _clientVisibleArea );
 }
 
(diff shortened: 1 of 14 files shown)


kit/ChildSession.cpp
```diff
@@ -1247,14 +1249,38 @@ bool ChildSession::clientVisibleArea(const StringVector& tokens)
...
+        // interacted with the keyboard/mouse recently ?
+        if (getInactivityMS(now) < 200 /* ms */) // typing etc.
+            score *= 2.0;
+
+        // pre-loading near the viewing area is also more important than far away
+        Util::Rectangle r = tile.toAABBox();
+        // grow in each direction
+        Util::Rectangle enlarged(r.getLeft() - r.getWidth(), r.getTop() - r.getHeight(),
+                           r.getRight() + r.getWidth(), r.getBottom() + r.getHeight());
+        if (enlarged.intersects(_clientVisibleArea))
```
caolanm, line 1271: I think the 4 arg Util::Rectangle ctor is: int x, int y, int width, int height, while there is a ::create that takes x1, y1, x2, y2 which is what this looks like it wants.

mmeeks: Grief; I was fooled by reading the rather odd variable initialization in the constructor somehow
    Rectangle(int x, int y, int width, int height)
        : _x1(x)
        , _y1(y)
        , _x2(x)
        , _y2(y)

mmeeks: fixed; thanks =)

mmeeks:
@eszkadev @gokaysatir thoughts on the Typescript changes appreciated too - would really like to get this into RED - seems to improve things for me somewhat when scrolling in calc; and provides some more diagnostic tools for monitoring pre-loading.

gokaysatir (approved):
Preload map looks beautiful. JS code looks ok to me.

We are hitting an assertion, visible in JS console: "finalQueue length... ...exceeded maxTilesToFetch"

caolanm:
assert is 27 tiles vs maxTilesToFetch of 26.25

PR #10687, opened 2024-12-09
cypress: make clickOnFirstCell stable

eszkadev:
- remove timeout which showed us problem with click on first cell
- be sure we are in first cell before double click


vmiklos (approved):
Looks good to me, but please wait for the mobile suite, if I understand correctly then common/ code may break a multi-user test, still.

PR #10688, opened 2024-12-09
cool#10630 doc electronic sign: get the signature on the server

vmiklos:
- **cool#10630 doc electronic sign: get the signature on the server**
- **cool#10630 doc electronic sign: stop sending the secret to the client**
- **cool#10630 doc electronic sign: remove CSP rules**

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1711,23 +1711,20 @@ void Document::notifyViewInfo()
                         Parser parser;
                         Poco::Dynamic::Var var = parser.parse(userPrivateInfo);
                         Object::Ptr userPrivateInfoObj = var.extract<Object::Ptr>();
-                        std::string signatureCert;
-                        JsonUtil::findJSONValue(userPrivateInfoObj, "SignatureCert", signatureCert);
-                        if (!signatureCert.empty())
+                        std::initializer_list<std::string> keys = {
+                            "SignatureCert",
+                            "SignatureKey",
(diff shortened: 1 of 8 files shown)


vmiklos:
Rebase to be on top of latest master + switch to non-throwing when the 'body' produced by JS is not there.

kit/Kit.cpp
```diff
@@ -1711,23 +1711,20 @@ void Document::notifyViewInfo()
...
                         Poco::Dynamic::Var var = parser.parse(userPrivateInfo);
                         Object::Ptr userPrivateInfoObj = var.extract<Object::Ptr>();
-                        std::string signatureCert;
-                        JsonUtil::findJSONValue(userPrivateInfoObj, "SignatureCert", signatureCert);
-                        if (!signatureCert.empty())
+                        std::initializer_list<std::string> keys = {
+                            "SignatureCert",
+                            "SignatureKey",
+                            "SignatureCa",
+                            "ESignatureSecret",
```
vmiklos, line 1718: This is just adding the ESignatureSecret as a key to be filtered + refactor to avoid further copy & paste.

vmiklos:

The second half of this "move secret to the server" story. I got this wrong initially as following Zotero as an example looked like a good idea; but eideasy is more nuanced with its split clientid + secret token pair.

caolanm (approved):
looks sane to me

PR #10689, opened 2024-12-09
npm: install -> ci

vmiklos:
To see if this matters; in theory 'npm install' can get the last
compatible version of a depdendency, while 'npm ci' will try to install
exactly the same version, so it may be more predictable.

Change-Id: I7f242ea56c813f586bb84092933e31fdc96fdca8


vmiklos:
Ah, we have no package-lock.json; try to lock down each version instead, then.

vmiklos:
OK, this is not working; and the idea was that perhaps tinderbox has old/good versions (due to incremental build) and the PR check does from-scratch builds, but both do from-scratch builds, so this is not leading anywhere, closing.

PR #10690, opened 2024-12-09
Private/gokay/hyperlink position

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/URLPopUpSection.ts
+++ b/browser/src/canvas/sections/URLPopUpSection.ts
@@ -10,6 +10,10 @@ class URLPopUpSection extends HTMLObjectSection {
 	removeButtonId = 'hyperlink-pop-up-remove';
 	arrowDiv: HTMLDivElement;
 
+	static arrowHalfWidth = 10;
+	static horizontalPadding = 6;
+	static popupVerticalMargin = 20;
+
 	constructor(url: string, documentPosition: cool.SimplePoint, linkPosition?: cool.SimplePoint) {
         super(URLPopUpSection.sectionName, null, null, documentPosition, URLPopUpSection.cssClass);
 
@@ -116,6 +120,7 @@ class URLPopUpSection extends HTMLObjectSection {
 
 		document.getElementById(this.removeButtonId).onclick = () => {
 			app.map.sendUnoCommand('.uno:RemoveHyperlink', params);
+			URLPopUpSection.closeURLPopUp();
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


browser/src/canvas/sections/URLPopUpSection.ts
```diff
@@ -111,18 +116,47 @@ class URLPopUpSection extends HTMLObjectSection {
...
+			We will check if we reposition the popup.
+			If the arrow position falls outside of the popup, we will put it on the edge.
+		*/
+		const clientRect = this.getPopUpBoundingRectangle();
+		const arrowHalfWidth = 10;
+		let arrowCSSLeft = this.sectionProperties.documentPosition.pX - this.documentTopLeft[0] + this.containerObject.getDocumentAnchor()[0] - arrowHalfWidth;
+		arrowCSSLeft /= app.dpiScale;
+		arrowCSSLeft += document.getElementById('canvas-container').getBoundingClientRect().left; // Add this in case there is something on its left.
+		arrowCSSLeft -= clientRect.left;
+		this.arrowDiv.style.left = (arrowCSSLeft > 6 ? arrowCSSLeft : 6) + 'px'; // 6: padding of the popup.
```
eszkadev, line 138: Instead of a comment, can we put 6 as const variable? it gives a label and share value in all places.
const popupPadding = 6;

browser/src/canvas/sections/URLPopUpSection.ts
```diff
@@ -111,18 +116,47 @@ class URLPopUpSection extends HTMLObjectSection {
...
-		const left = section.sectionProperties.documentPosition.pX - section.getPopUpWidth() * 0.5 * app.dpiScale;
-		const top = section.sectionProperties.documentPosition.pY - (section.getPopUpHeight() + 20) * app.dpiScale;
+		let left = section.sectionProperties.documentPosition.pX - section.getPopUpWidth() * 0.5 * app.dpiScale;
+		let top = section.sectionProperties.documentPosition.pY - (section.getPopUpHeight() + 20) * app.dpiScale;
 
 		if (section) {
-			section.setPosition((left >= 0 ? left : 10), (top >= 0 ? top : 10));
+			let arrowAtTop = false;
+			if (top < 0) {
+				top = section.sectionProperties.documentPosition.pY + (40 * app.dpiScale);
```
eszkadev, line 151: I would put all the magic numbers in single place here as memebers of a class and reuse everywhere:
```
const popupHorizontalPagging = 6;
const popupVerticalPadding = 20;
```
then we have single value to change all of them in case designers will decide that

PR #10691, opened 2024-12-09
Private/caolan/coverity2024 12 08

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Util.cpp
+++ b/common/Util.cpp
@@ -728,11 +728,10 @@ namespace Util
             malloc_info(0, f);
             fclose(f);
 
-            if (p && size)
-            {
+            if (size)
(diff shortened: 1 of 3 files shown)


eszkadev:
rebased after cypress fix - let's try now

caolanm:
open_memstream size arg reports size without trailing null, so if there is a FILE* there was an allocation saved to *ptr and if nothing written size will claim 0, but there still needs to be a deallocation. Practically probably not a problem.

vmiklos (approved):
Aha, first time I see, open_memstream(), makes sense.

PR #10692, opened 2024-12-09
increase UnitTimeoutInactive inactivityTimeout

caolanm:
fails for me locally with --enable-debug 'CFLAGS=-O0 -ggdb' 'CXXFLAGS=-O0 -ggdb'

Change-Id: I8cd0cb1d187bf3adab1f81345c1cd83b5180eee4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitTimeoutInactive.cpp
+++ b/test/UnitTimeoutInactive.cpp
@@ -37,7 +37,7 @@ class UnitTimeoutInactivity : public UnitTimeoutBase0
     void configure(Poco::Util::LayeredConfiguration& /* config */) override
     {
         // net::Defaults.inactivityTimeout = std::chrono::seconds(3600);
-        net::Defaults.inactivityTimeout = std::chrono::milliseconds(100);
+        net::Defaults.inactivityTimeout = std::chrono::milliseconds(150);
         //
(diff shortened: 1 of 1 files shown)


caolanm:
This is fairly edge case I guess, but it looks harmless to me to increase this.

test/UnitTimeoutInactive.cpp
```diff
@@ -37,7 +37,7 @@ class UnitTimeoutInactivity : public UnitTimeoutBase0
     void configure(Poco::Util::LayeredConfiguration& /* config */) override
     {
         // net::Defaults.inactivityTimeout = std::chrono::seconds(3600);
-        net::Defaults.inactivityTimeout = std::chrono::milliseconds(100);
+        net::Defaults.inactivityTimeout = std::chrono::milliseconds(150);
```
vmiklos, line 40: Aha, the default is 3600 seconds and we decrease this drastically to have a faster test, but then with 100ms we turn a stream socket into an inactive one so fast that it happens when no timeout should happen in a debug build, and increasing that decreased value makes the test more stable, while still not having to wait for 1h for the inactivity to happen. Makes sense after reading this about 3 times. :-)

vmiklos (approved):

PR #10694, opened 2024-12-09
Test also we have no open paste-special popup windows after completin…

gokaysatir:
…g the test.

Below PR solved the issue and this addition is to ensure it doesn't regress.

https://github.com/CollaboraOnline/online/pull/10648

Change-Id: I68a61d3ba81cad4f8790b1950ac591c7f5c37890


vmiklos (approved):
Makes sense, thanks. :-)

PR #10696, opened 2024-12-09
Bin two pointless #defines

tml1024:
They are used only in one place, so expand them there.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
rebased

vmiklos (approved):
Makes sense, I assume this was originally copied from a location when e.g. the suffix wasn't always ".so" and so on.

PR #10697, opened 2024-12-10
jsdialog: added missing 'disabled' attribute to listbox

lpranam:
problem:
if listbox was initially disabled main container had missing disabled attribute, and because of that MutationObserver was never triggered to enable it

Change-Id: Ibc1686eb03f4099066ff5b1537c90747432aee66

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -376,7 +376,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 		if (data.enabled === false) {
 			div.disabled = true;
-			spinfield.setAttribute('disabled', '');
+			spinfield.setAttribute('disabled', 'true');
 		}
 
 		JSDialog.SynchronizeDisabledState(div, [spinfield]);
@@ -1683,12 +1683,13 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		listboxArrow.id = 'listbox-arrow-' + data.id;
 
 
-		JSDialog.SynchronizeDisabledState(container, [listbox]);
-
 		if (data.enabled === false) {
 			container.disabled = true;
 			listbox.disabled = true;
+			container.setAttribute('disabled', 'true');
 		}
+
+		JSDialog.SynchronizeDisabledState(container, [listbox]);
 
 		$(listbox).change(() => {
 			if ($(listbox).val())
@@ -2388,7 +2389,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 					}
 
 					if (disabled)
-						div.setAttribute('disabled', '');
+						div.setAttribute('disabled', 'true');
 					else
 						div.removeAttribute('disabled');
 				};
@@ -2408,7 +2409,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			}
 
 			if (disabled) {
-				div.setAttribute('disabled', '');
+				div.setAttribute('disabled', 'true');
 			}
 
 			var selectFn = function() {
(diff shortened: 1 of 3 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1688,6 +1688,7 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === false) {
 			container.disabled = true;
 			listbox.disabled = true;
+			container.setAttribute('disabled', '');
```
eszkadev, line 1691: why? this is `data.enabled === false` branch

lpranam: Yes, that means it's disabled, that's where attributes are set. Similar thing happens in spinfield, unobutton , etc...

eszkadev: ah right, it adds empty but present attr

eszkadev (approved):

eszkadev:
please rabase, cypress was fixed

eszkadev:
@lpranam please check failing jsdialog_spec.js test - seems to be related

eszkadev:
again ` impress/jsdialog_spec.js` failed so looks like real problem

browser/src/control/jsdialog/Util.StateChange.ts
```diff
@@ -21,7 +21,7 @@ function onStateChange(element: Element, callback: StateChangeCallback) {
 	var enabledCallback = function (mutations: Array<MutationRecord>) {
 		for (var i in mutations) {
 			if (mutations[i].attributeName === 'disabled') {
-				var enable = mutations[i].oldValue !== null;
+				var enable = mutations[i].oldValue !== null && mutations[i].oldValue !== '';
```
eszkadev, line 24: https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord/oldValue
`For attributes, it is the value of the changed attribute before the change.`

so if the `oldValue` was `''` (empty string) and it meant to be "disabled" -> now we want to enable widget
then the `enable` value should be `true` so we will do `removeAttribute('disabled')` inside synchronizeDisabledState

or I missed something?

eszkadev: Then best would be to do `setAttribute('disabled', 'true')` not empty string for disabled case.
Possibly we should change all the places we do the same...

eg in edit we use true: https://github.com/CollaboraOnline/online/blob/9a9ff2663aefa5c794b39c111cf0d6293363f365/browser/src/control/Control.JSDialogBuilder.js#L1523
in unobutton not: https://github.com/CollaboraOnline/online/blob/9a9ff2663aefa5c794b39c111cf0d6293363f365/browser/src/control/Control.JSDialogBuilder.js#L2504

eszkadev: That will be less confusing

lpranam: This approach is not making sense to me based on `oldValue`. In case of re-triggering same state, if previous state was set to disabled and we get message to disable it again, it will be enabled incorrectly. 

And so far I am not able to find the way to get the new value which we are trying to set in MutationObserver.

eszkadev:
```[MutationRecord.target](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord/target) Read only

    The node the mutation affected, depending on the MutationRecord.type.

        For attributes, it is the element whose attribute changed.```
        
isn't the node already changed when that code is called? it means you can check value of target

File: browser/src/layer/marker/TextInput.js [hunk 3]
```diff
@@ -159,7 +159,7 @@ L.TextInput = L.Layer.extend({
 	},
 
 	disable: function () {
-		this._textArea.setAttribute('disabled', true);
```
eszkadev, line 162: please do not modify TextInput.js file, it's unrelated to JSDialog, its about keyboard handling

eszkadev:
@lpranam could you add cypress test for your case?

PR #10702, opened 2024-12-10
private/marco/slideshow skip process layers after exit

mcecchetti:
- **slideshow: not process left slide layers after quitting from slideshow**
- **Test also we have no open paste-special popup windows after completing the test.**


mcecchetti:
bad PR

PR #10704, opened 2024-12-10
HTTPCrashTest::testCrashForkit can timeout while forkit is still rest…

caolanm:
…arting

under high load with a -O0 debug online build and dbgutil core build

Change-Id: Id740772ac421f6c29594a5ded26573d71f8d39b3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/httpcrashtest.cpp
+++ b/test/httpcrashtest.cpp
@@ -39,6 +39,10 @@
 
 using namespace helpers;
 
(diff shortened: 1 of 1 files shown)


Ashod (approved):
Thanks, @caolanm. Looks like we were getting (un)lucky and didn't notice this until now.

PR #10706, opened 2024-12-11
cool#10630 doc electronic sign: configure the language of the popup

vmiklos:
- **automake silent rules: handle tsc**
- **cool#10630 doc electronic sign: configure the language of the popup**

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -29,13 +29,17 @@ $(L10N_IOS_ALL_JS) : $(wildcard $(srcdir)/po/ui-*.po) $(shell find $(srcdir)/l10
 	done
 endif
 
+QUIET_TSC = $(QUIET_TSC_$(V))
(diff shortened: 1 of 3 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/3327/console fails in `writer/annotation_spec.js`, but a local `make -C cypress_test check-desktop spec=writer/annotation_spec.js` passes, retry.

vmiklos:

The first part is just an aside, we have a mix of "@command" and "command" invocations in our makefiles, while automake has a consistent way to have silent vs verbose rules (both by default and then by asking for verbose in a silent-by-default case), this is just a first step to do it consistently for tsc invocations.

The second is about routing info about our UI language to the popup's external HTML, e.g. German UI works consistently with this.

caolanm (approved):
code change looks straight forward.

In core we should make V=1 do the same as VERBOSE=true, that's bugged me for a while

PR #10709, opened 2024-12-11
Use better checks for determining if a comment is focused or not.

gokaysatir:
Some browsers don't support events like focusin focusout, also browsers may have different behaviour about parent of a focused element.

The issue: commentHasFocus variable is not updated correctly. This prevents cut/paste operations. Because when a comment is focused, cut/paste from document is prevented.
Fix: It's better to check if a modify / reply field is focused via document.activeElement.

This changes the code introduced in below commit:
* https://github.com/CollaboraOnline/online/pull/7634/files

So the fix will be re-checked.

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -242,8 +242,6 @@ export class Comment extends CanvasSectionObject {
 		var isRTL = document.documentElement.dir === 'rtl';
 		this.sectionProperties.container = L.DomUtil.create('div', 'cool-annotation' + (isRTL ? ' rtl' : ''));
 		this.sectionProperties.container.id = 'comment-container-' + this.sectionProperties.data.id;
-		this.sectionProperties.container.addEventListener('focusin', this.onContainerGotFocus.bind(this));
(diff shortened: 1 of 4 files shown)


browser/src/canvas/sections/CommentSection.ts
```diff
@@ -1192,6 +1181,10 @@ export class Comment extends CanvasSectionObject {
 		return null;
 	}
 
+	public static isAnyFocus(): boolean {
+		return document.activeElement.id.startsWith('annotation-modify-textarea') || document.activeElement.id.startsWith('annotation-reply-textarea');
```
eszkadev, line 1185: This is fragile, if someone will change id - we will have a bug.
Can we bind that by some shared variable or method, with the place where we create textarea?

gokaysatir: Thanks for checking. It should be safer now.

eszkadev (approved):
thanks, now is good :)

mmeeks:
@timar should we back-port to latest COOL release ? - would love a fix for Peter =)

PR #10712, opened 2024-12-11
Do not put Makefile.am and Makefile.in in browser/dist/l10n/ directory

timar:
Change-Id: I99da267352b1caaa0f44ce032aa9131abdc2e62e

see e.g. https://share.collaboraonline.com/browser/dist/l10n/Makefile.am – totally unnecessary.


vmiklos (approved):
Yes, this is probably my fault from 2bd3317463ee62188e701f1ef06d8f30914e509b.

Diffing the old vs new output, indeed just makefiles and directories were non-JSON files, so this looks good to me, thanks.

PR #10719, opened 2024-12-12
cool#10630 doc electronic sign: add a country selector to the dialog

vmiklos:
- **automake silent rules: handle eslint**
- **cool#10630 doc electronic sign: add a country selector to the dialog**
- **cool#10630 doc electronic sign: add ISO country names**
- **cool#10630 doc electronic sign: send selected country to the server**


vmiklos:

The confusing flag is now gone:

image

There are smaller things remaining (sort the list by UI name, reduce the providers based on country selection), but I suggest to not grow this PR further.

browser/src/control/Control.ESignature.ts
```diff
@@ -100,12 +110,96 @@ namespace cool {
...
 		constructor(url: string, clientId: string) {
 			this.url = url;
 			this.clientId = clientId;
 
 			app.map.on('commandvalues', this.onCommandValues.bind(this));
 			app.map.on('commandresult', this.onCommandResult.bind(this));
+
+			if (!ESignature.countryNames) {
+				ESignature.countryNames = {
+					// Generated by scripts/countries.py
```
vmiklos, line 125: I asked for input from @timar here: either we bundle an npm package (it has an additional dependency) that has these strings is painful external JSON files (but then no duplication here) or we have this here and we avoid the bloat of a new dependency. His suggestion was to have this inline, but do it with a script. I like this approach, because then this avoids the political issues with country names: we just take the names as-is from the ISO data.

caolanm: I think i18n-iso-countries has done some (sensible) processing of the ISO data, because notoriously 3166 has a short(!) name of "United Kingdom of Great Britain and Northern Ireland" for the UK :-)

"AX" is unusual choice for inclusion, but I guess that's eID's decision

caolanm (approved):

browser/src/control/Control.ESignature.ts
vmiklos: There are a few others, like https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#TR says "TR" is Türkiye, but this package sticks to Turkey.

PR #10721, opened 2024-12-12
Fix close when map is not present

eszkadev:
fixes regression from commit 662a707ec87d0a1855ca81102fa4dd512bece91a Share code for close action

if we open broken file and cancel repair - we end with closed document map is not initialized and dispatcher too, we need to use dummy one to allow close button to close view


gokaysatir (approved):
Looks good, thanks.

PR #10722, opened 2024-12-12
private/marco/slideshow motion animations

mcecchetti:
- **slideshow: provide support for motion animations along a path**
- **slideshow: shape position can be wrong when running some animations**


quikee (approved):
looks good tome

PR #10723, opened 2024-12-12
Issue: There is a JS error when we scroll while mouse pointer is abov…

gokaysatir:
…e a comment.

Fix: Simplify event propagation from comment container to CanvasSectionContainer.

Now we call sectionContainer's event from comment container.

Why we need it: Comment container is an HTML object and we want to selectively propagate events from it to canvas.
Note: The coordinate of action is probably wrong but ScrollSection is not using the mouse coordinates, so it's fine.

Change-Id: If2819146c9d055b419944156cef12798b8974de8


browser/src/canvas/sections/CommentSection.ts
```diff
@@ -170,8 +170,14 @@ export class Comment extends CanvasSectionObject {
...
-		L.DomEvent.on(this.sectionProperties.container, 'wheel', app.sectionContainer.onMouseWheel, app.sectionContainer);
-		L.DomEvent.on(this.sectionProperties.contentNode, 'wheel', this.onMouseWheel, this);
+
+		this.sectionProperties.container.onwheel = function(e: WheelEvent) {
+			// Don't scroll the document if mouse is over comment content. Scrolling the comment content is priority.
+			if (!this.sectionProperties.contentNode.matches(':hover')) {
+				e.preventDefault();
+				app.sectionContainer.onMouseWheel(e);
+			}
+		}.bind(this);
```
eszkadev, line 180: we can use arrow functions to have `this` automatically: () => {}

Issue #10725, opened 2024-12-12
Checkbox label in user popup is not vertically centered correctly


banobepascal:
### Describe the bug
<img width="306" alt="Screenshot 2024-12-12 at 13 51 02" src="https://github.com/user-attachments/assets/5ca52d33-b541-415e-9131-0a725b9e7c08" />

### Steps to reproduce
 - Open writer/calc/impress with multiple users
 - Click the user avatars and a popup will appear 
 - Notice the alignment of the text "Always follow the editor"

PR #10726, opened 2024-12-12
Align checkbox label vertically and centered in user popup

banobepascal:
Change-Id: Ic7857910b9321a087798ebfa83cca81aed622845

* Resolves: #10725 
* Target version: master 

### Summary
Vertical alignment of the checkbox label in the user popup, ensuring it is properly centered for better visual consistency and user experience.

### TODO
Screenshot 2024-12-12 at 17 02 51

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -1727,11 +1727,13 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
 
 #follow-editor {
 	padding: 8px;
+	display: flex;
+	align-items: center;
 }
 
 #follow-editor .follow-editor-checkbox {
-	margin-right: 12px;
-	margin-left: 7px;
+	margin-inline-end: 11px;
+	margin-inline-start: 6px;
 	border-radius: 1px;
 	height: 16px;
 	width: 16px;


browser/css/toolbar.css
```diff
@@ -1727,11 +1727,13 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
 
 #follow-editor {
 	padding: 8px;
+	display: flex;
+	align-items: center;
```
pedropintosilva, line 1731: great!

browser/css/toolbar.css
```diff
@@ -1727,11 +1727,13 @@ menu-entry-with-icon.padding-left + menu-entry-icon.width */
...
 }
 
 #follow-editor .follow-editor-checkbox {
-	margin-right: 12px;
-	margin-left: 7px;
+	margin-right: 11px;
+	margin-left: 6px;
```
pedropintosilva, line 1736: wouldn't it be better to drop the margin-left? I think we can do that by making sure that the checkbox has the same size as the avatars (28px). Lastly, it would be good to avoid using -left or -right and instead use margin-inline-start/end  in this case https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end this way we are sure that the margin will be correctly applied for property depending on the values defined for [writing-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode), [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction), and [text-orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/text-orientation).

pedropintosilva (changes requested):

browser/css/toolbar.css
banobepascal: Screenshot from 2024-12-20 12-47-08

Using margin-inline-end/start is better, but adjusting the checkbox to 28px is not better for UX

pedropintosilva: agreed, I thought somehow that the check box itself would not change the size that would only occupy more space but that graphically would be exact the same perceived. you are correct, thanks

pedropintosilva:
please change the margin to inline so RTL etc doesn't get badly affected and we can m,erge this

pedropintosilva:
I have pressed re-try since the failed test "calc/searchbar_spec.js" from COOL 24.04 Cypress (desktop) is unrelated

pedropintosilva:
we should ignore the iOS failed test for now

pedropintosilva (approved):
Thank you!

PR #10727, opened 2024-12-12
temp

gokaysatir:
Change-Id: I604dc30871aa4debb60b8a8139c80cd3e65645e9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -3376,9 +3376,13 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
             else if (tokens.size() == 3 && tokens.equals(0, "EMPTY"))
             {
                 // with mode:    "EMPTY, <part>, <mode>"
-                const std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
-                const std::string mode = (_docType != "text" ? tokens[2].c_str() : "0"); // Writer is not using mode.
-                sendTextFrame("invalidatetiles: EMPTY, " + part + ", " + mode +
+                std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
+                std::string mode = (_docType != "text" ? tokens[2].c_str() : "0"); // Writer is not using mode.
+
+                int part_ = getLOKitDocument()->getPart();
+                int mode_ = getLOKitDocument()->getEditMode();
+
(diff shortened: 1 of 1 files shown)


kit/ChildSession.cpp
```diff
@@ -3417,8 +3417,9 @@ void ChildSession::loKitCallback(const int type, const std::string& payload)
             else if (tokens.size() == 3 && tokens.equals(0, "EMPTY"))
             {
                 // with mode:    "EMPTY, <part>, <mode>"
-                const std::string part = (_docType != "text" ? tokens[1].c_str() : "0"); // Writer renders everything as part 0.
-                const std::string mode = (_docType != "text" ? tokens[2].c_str() : "0"); // Writer is not using mode.
+                std::string part = (_docType != "text" ? std::to_string(getLOKitDocument()->getPart()) : "0"); // Writer renders everything as part 0.
+                std::string mode = (_docType != "text" ? std::to_string(getLOKitDocument()->getEditMode()) : "0"); // Writer is not using mode.
+
                 sendTextFrame("invalidatetiles: EMPTY, " + part + ", " + mode +
                               " wid=" + std::to_string(getCurrentWireId()));
```
eszkadev, line 3424: does that mean we ingore what was requested? (part)

gokaysatir: Callback is sent from the core side. I think we confirm the request's part and mode rather than ignoring :) And this case is relevant when core side sends mode and part numbers with an "EMPTY" string. I think this case occurs mostly when we switch modes in Impress.

The reason for confirming is that while core side sends the callback, the window pointer is empty in Impress. That null pointer causes core side to send "0" instead of the actual part and mode. This happens if user switches from normal view to notes view.

Other than above scenario, getPart and getEditMode should fetch the same values.

eszkadev: what if `EMPTY` will be invoked for other page ? eg. we do massive invalidation like changed the page format/size ? isn't that a case here?

gokaysatir: Now i tested it and this more and it seems this is not a good solution. I need to narrow-down the cases. Moving slides or pages can cause unwanted behaviour. I will re-check a core side solution again or will add a corner case for the notes switch.

Thanks for the comments.

eszkadev (changes requested):
marked as draft and request for change so it will not appear in ready to merge queue

Issue #3077, opened 2021-08-13
ctrl-space should select a column ...


mmeeks:
**This is an [Easy Hack](https://collaboraonline.github.io/post/easyhacks/).**
Potential mentors: @merttumer 

### Detailed description and rationale

calc: ctrl-space should select a column ...

### Code pointers

This works in the PC / desktop app. You'll want to tripple click in help->about, look in the browser's console to see what events are going out on the wire for this (I guess), and see whether the browser swallows the ctrl up/down, or if the 'space' event (which is quite a special event) is mis-transmitted somehow.

Shift-space selects the row - and this appears to work; unclear why =)

Might be an nightmare =) but perhaps it is easy enough; lets see.

Mathusela:
Hello, is this issue still open?
I'm new to open source, but I would like to take a look at this if it's still relevant.
Thanks.

mmeeks:
Hi Mathusela - feel free to start on the ticket =) please do the tripple click as above, turn on protocol logging, and see what happens in the protocol log for a <ctrl>-<space> - it may be that we can't get this out of the browser. Drop your findings back into the ticket ... Thanks !

Mathusela:
Can't see anything immediately out of place in the protocol logs, shift+space (working event) and ctrl+space (broken event) appear consistent.
It's worth noting that a short time after ctrl+space is used (or after using the shortcut and moving the selected cell) the row headers on the left of the screen change colour as if a column is selected; this is not true selection, deleting at this point will delete only the selected cell not the column.
This does however seem to suggest that the event is being processed and the issue is elsewhere.

![image](https://user-images.githubusercontent.com/46897785/233461104-d59e8113-ef3e-4ee0-abd4-abf14469e38f.png)

Mathusela:
Hi sorry for the late follow up:
Here's the log (just saved from the chrome console):
[collabora.log](https://github.com/CollaboraOnline/online/files/11326414/collabora.log)

mmeeks:
Its odd - it seems we have rather few 'keyup' events happening - I would expect us to have an 

OUTGOING: key type=up 

sort of event; then again - <ctrl>-<space> seems to work immediately on the key-down in the desktop version; so that's not obviously the problem.

It is particularly amazing that the action seems to happen after some time delay - if you press ctrl-space and leave it. I'd look at the logs on the coolwsd side to hunt down what oddness is going on in that gap - and that suddenly makes us refresh the row selections.

Rather an interesting one =)

[...]

PR #10729, opened 2024-12-12
calc: fix Ctrl+Space to select a column

rparth07:
Change-Id: Icc13560fbb6754ce4c5fc73cbacf2f4dcc41e829

* Resolves: #3077
* Target version: master 

### Summary
This resolves the Ctrl+Space issue and improves consistency across platforms.

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.KeyboardShortcuts.ts
+++ b/browser/src/map/handler/Map.KeyboardShortcuts.ts
@@ -171,7 +171,7 @@ keyboardShortcuts.definitions.set('default', new Array<ShortcutDescriptor>(
     new ShortcutDescriptor('spreadsheet', 'keydown', Mod.CTRL | Mod.SHIFT, 'PageUp', undefined, undefined),
     new ShortcutDescriptor('spreadsheet', 'keydown', Mod.CTRL | Mod.SHIFT, 'PageDown', undefined, undefined),
     new ShortcutDescriptor('spreadsheet', 'keydown', 0, 'F5', null, null, null),
-
+    new ShortcutDescriptor('spreadsheet', 'keydown', Mod.CTRL, ' ', '.uno:SelectColumn', null, null),
 
     new ShortcutDescriptor('text', 'keydown', 0, 'F2', null, null, null),
     new ShortcutDescriptor('text', 'keydown', 0, 'F5', null, null, null),


Darshan-upadhyay1110:
@rparth07 have you tested the normal master branch ?

I see without this patch i can still user `Ctrl + Space` and it works as expected. 

[Screencast from 2024-12-12 22-30-11.webm](https://github.com/user-attachments/assets/8f5c7bcb-0cac-45ef-82b4-802a0f5fc2a9)

Darshan-upadhyay1110:
@eszkadev although this is very strange that this functionality not working for staging.

rparth07:
> @rparth07 have you tested the normal master branch ?
> 
> I see without this patch i can still user `Ctrl + Space` and it works as expected.
> 
>  Screencast.from.2024-12-12.22-30-11.webm 
> https://github.com/CollaboraOnline/online/blob/master/browser/src/layer/tile/CanvasTileLayer.js#L3116

Strange! In my local, I checked on latest master commit but it is not working.

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -863,6 +863,12 @@ L.Map.Keyboard = L.Handler.extend({
 		case this.keyCodes.PERIOD: // .
 			app.socket.sendMessage('uno .uno:SuperScript');
 			return true;
+		case this.keyCodes.SPACE:
```
eszkadev, line 866: We have new array for adding shortcuts in more unified and verified way.
https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts#L161

Could you try to add that one there?

gokaysatir (changes requested):
Whole row and column selections are handled by the core side. Client only sends the events. I think we need a good reproducer before having a solution. Otherwise we may duplicate the events.

browser/src/map/handler/Map.Keyboard.js
rparth07: Thanks @eszkadev for the suggestion! I've updated the implementation and it works as expected.

rparth07:
> Whole row and column selections are handled by the core side. Client only sends the events. I think we need a good reproducer before having a solution. Otherwise we may duplicate the events.

@gokaysatir, could you please suggest an approach for creating a reliable reproducer for this issue? I've attached the logs for both key events (Shift + Space for row selection and Ctrl + Space for column selection) below. These show the incoming and outgoing messages around the events.

### Log for Shift + Space (row selection):
```
==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265271378,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
global.js:564 1734265271966 OUTGOING: key type=input char=0 key=5380
 
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265271964,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
global.js:564 1734265271973 INCOMING: calcfunctionlist: hidetip 
global.js:564 1734265271974 INCOMING: cellselectionarea: 0, 256, 1279, 255 
global.js:564 1734265271974 INCOMING: window: {"id":"7","action":"invalidate","rectangle":"2, 2, 161, 24"} 
global.js:564 1734265271974 INCOMING: celladdress: A2:XFD2 
global.js:564 1734265271975 INCOMING: cellselectionarea: 0, 256, 20971519, 255 
global.js:564 1734265271975 INCOMING: textselection: 0, 256, 20971519, 255 
global.js:564 1734265271977 INCOMING: cellautofillarea: 20889555, 465, 75, 75 
global.js:564 1734265271977 INCOMING: jsdialog: { "jsontype": "addressinputfield", "action": "action", "id": 0, "data": { "control_id": "pos_window", "text": "A2:XFD2", "action_type": "setText"}} 
global.js:564 1734265272064 OUTGOING: key type=up char=0 key=5380
 
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265272062,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265272140,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
global.js:564 1734265272291 INCOMING: statechanged: .uno:InsertColumnsAfter=disabled 
global.js:564 1734265272291 INCOMING: statechanged: .uno:InsertColumnsBefore=disabled 
global.js:564 1734265272292 INCOMING: statechanged: {
    "state": {
        "vertical": "false",
        "horizontal": "false"
    },
    "commandName": ".uno:BorderInner"
}
 
global.js:564 1734265272292 INCOMING: statechanged: { "commandName": ".uno:RowColSelCount", "locale": "en-US", "state": "Selected: 1 row, 16,384 columns"} 
global.js:564 1734265272293 INCOMING: statechanged: .uno:MergeCells=enabled 
global.js:564 1734265272293 INCOMING: statechanged: .uno:ToggleMergeCells=false 
```


### Log for Ctrl + Space (column selection):
```
==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265187777,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
global.js:564 1734265188441 OUTGOING: key type=input char=0 key=9476
 
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265188438,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265188597,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
debug.html?file_path=/home/rparth07/open-source/Collabora/collabora-online/test/samples/calc-edit.ods:87 ==== debug.html receiveMessage: {"MessageId":"FollowUser_Changed","SendTime":1734265188685,"Values":{"FollowedViewId":4,"IsFollowUser":true,"IsFollowEditor":false}}
```

It seems like the events are being transmitted, but I suspect there may be an issue with how they are processed by the core. Let me know your thoughts on the next steps for creating a reproducible case! 

Thanks in advance for your help!

gokaysatir:
@rparth07 thank you for sharing logs. When i compare the logs with my console, i see that the action is performed and we receive the results correctly. If you check below lines, they all show a huge selection:

```
global.js:564 1734265271975 INCOMING: cellselectionarea: 0, 256, 20971519, 255 
global.js:564 1734265271975 INCOMING: textselection: 0, 256, 20971519, 255 
global.js:564 1734265271977 INCOMING: cellautofillarea: 20889555, 465, 75, 75 
global.js:564 1734265271977 INCOMING: jsdialog: { "jsontype": "addressinputfield", "action": "action", "id": 0, "data": { "control_id": "pos_window", "text": "A2:XFD2", "action_type": "setText"}} 
```

Now as we know that the core side thinks that the cells are selected, there are 2 possibilities i can think of:
* We deselect the rows after selecting them for some reason.
* Rows are still selected but we don't paint the selected cells.

For checking if the rows are still selected, we can try below methods:
* After selecting whole row: Try pressing "enter" key. It should move to next cell to the right, not to the next row because of the selection.
* After selecting whole column: Try pressing "tab" key. It should move to next row, not to the right cell because of the selection.

If above methods show that core side keeps the selected cells, we will need to check why don't we paint them.

If methods show that cells are not selected anymore, we will need to check why we cancel the selection.

I hope this helps to find the root cause. Thank you.

rparth07:
Thank you @gokaysatir for the detailed review and feedback.

Based on the tests conducted with the provided methods, here are the observations:

**After selecting the whole row using using shortcut and pressing the "Enter" key:**

This is working as expected in both cases, with or without the solution in this PR. The "Enter" key moves to the next cell to the right as intended, which indicates that the row selection is being properly maintained.

**After selecting the whole column using shortcut and pressing the "Tab" key:**

- **Without the solution in this PR:** After selecting the whole column, pressing the "Tab" key does not move to the next row, but instead moves to the right cell. This indicates that the column selection is not being handled correctly, causing the focus to behave differently.
- **With the solution in this PR:** After selecting the whole column, pressing the "Tab" key correctly moves to the next row as expected. This suggests that the column selection is being correctly recognized, and the focus shift to the next row works as intended.

From these observations, it appears that the changes in this PR address the issue with column selection behavior. Please let me know if you think any additional checks or tests are needed, or if there are any other aspects we should explore further.

Thank you once again for your valuable feedback!

gokaysatir:
Thank you too @rparth07. It'd be good to have a fix because other users are also affected by this issue. I can't reproduce but i guess we will get there.

I checked the CTRL+Space logs again. It seems to send the correct event to core side.

* Can you check again if there is a log in JS console something like (CTRL+Space):
    `INCOMING: cellselectionarea: 8960, 0, 1279, 268435455`?

* I guess there isn't, since columns don't act like they are selected, but still good to check. If there is, it means we are deselecting the cells after selecting them.

It seems that we have 2 things ahead:
* We will need to check why row selections are not painted on Online side.
* But first, can you check if desktop app selects the column with CTRL+Space? Maybe desktop app is not working as expected.
   * If desktop app also doesn't select the column with CTRL+Space, we will need to check some settings there or some forums to see if it is a settings-related issue. If it is not a settings-related issue but a bug, we will need to pursue the fix on the core side.
   * If desktop app does select the column with CTRL+Space, then we will need to check Online side to see why selection is not happening. Maybe it's selecting and deselecting the column, JS console logs can light that point.

Let me explain why i don't want to send an uno command to the server:
* We are letting core side to handle events when possible. This ensures that the event handling is aligned with core side. For example, for now while editing a cell, you can use CTRL+Space and select whole column. What if that behaviour changes and while editing a cell, column selection is prevented (on the core side): That would cause us to do extra work on the Online side, adding a corner case to the event handling to not send uno command while editing a cell (in order to align with core side behaviour).
* We need to prevent the default behaviour if we send uno command ourself, instead of letting core to handle events. This prevention needs to be checked and guarded with a test. So we will add one more thing to maintain.
* Of course we can add things to maintain, but first please let's be sure that we don't have any other solutions. This bug may be OS-related, browser related or something else related because it doesn't occur for every user.

By the way, my OS is Linux-Mint and tried with FireFox and Chrome. Selections worked.

Thank you.

rparth07:
Thank you @gokaysatir for the detailed steps and suggestions! 

Here's an update on the tasks:

> Can you check again if there is a log in JS console something like (CTRL+Space):
INCOMING: cellselectionarea: 8960, 0, 1279, 268435455?

I confirmed that there are no logs in the JS console resembling `INCOMING: cellselectionarea: 8960, 0, 1279, 268435455`. This aligns with your suspicion that the cells may not be getting selected at all.

> But first, can you check if desktop app selects the column with CTRL+Space? 

I tested this functionality on the desktop app, and the column selection with `CTRL+Space` works as expected. This indicates that the core side is capable of handling the event correctly.

> Let me explain why i don't want to send an uno command to the server:
> ...

I completely agree with the need to avoid directly sending a `.uno` command unless absolutely necessary. Letting the core side handle events ensures alignment and reduces maintenance overhead.

> By the way, my OS is Linux-Mint and tried with FireFox and Chrome. Selections worked.

I am using Ubuntu 24.04 LTS and have tested the behavior on Chrome, Firefox, and Brave browsers. Unfortunately, the issue persists across all of them! :(

Regarding painting row selections on the Online side, I’m unsure of the best debugging approach. Could you suggest specific methods or areas we should investigate to identify the cause?

gokaysatir:
I also can see the issue on our test server. I have an _attempt_ to fix it:
* https://gerrit.libreoffice.org/c/core/+/178720

Or we will need to check the issue further.

For row selections, test server works and paints the selected cells.

But on your system selections were not painted. I think you can check "`CCellSelection`" class inside "`CRectangle.ts`" file.

I hope that helps and thank you for pursuing the fix. This issue is a nasty one :)

timar:
Let me close this PR, because the bug has been fixed in core.

PR #10730, opened 2024-12-12
[WIP] feat: insert text and images generated with nextcloud assistant

elzody:
* Part of: https://github.com/nextcloud/richdocuments/issues/3934
* Target version: master 

### Summary
These changes are needed from the Nextcloud side to support integrating the Nextcloud Assistant with Collabora. There is currently no way to add buttons to the notebook bar from other apps, so we would need to hardcode it just like we did with the Smart Picker.

The related pull request in richdocuments is: https://github.com/nextcloud/richdocuments/pull/4333

### Checklist


pedropintosilva:
I think this seems quite ready to review so, I have press the button "ready to review" . @elzody , if this is ready for review please rename the PR's title

elzody:
@pedropintosilva I went ahead and did a rebase onto master, it seems to have resolved some of the failing test, but some remain, and I am unsure how to interpret them to tell if they're related.

eszkadev:
I tried to run tests again, sometimes we have false failures

eszkadev:
Please rebase again, API was changed and it fails now

browser/src/control/Control.Menubar.js
```diff
@@ -190,6 +190,7 @@ L.Control.Menubar = L.Control.extend({
 				{type: 'separator'},
 				{name: _UNO('.uno:HyperlinkDialog'), id: 'inserthyperlink', type: 'action'},
 				{name: _('Smart Picker'), id: 'remotelink', type: 'action'},
+				{name: _('AI Assistant'), id: 'remoteaicontent', type: 'action'},
```
hfiguiere, line 193: what happen if WOPI `EnableRemoteAIContent` isn't set? Shouldn't this menu item be hidden?

elzody: You're right, didn't take the compact view menu into account. Will adjust.

browser/src/docdispatcher.ts
```diff
@@ -91,6 +91,9 @@ class Dispatcher {
 		this.actionsMap['remotelink'] = function () {
 			app.map.fire('postMessage', { msgId: 'UI_PickLink' });
 		};
+		this.actionsMap['remoteaicontent'] = function () {
+			app.map.fire('postMessage', { msgId: 'UI_InsertAIContent' });
```
hfiguiere, line 95: so how does this work? What is the expected response to this message?

We'll need to update the documentation afterwards.

juliusknorr: We don't expect any response, the integrator can handle it with whatever they want. Our integration is inserting text for example: https://github.com/nextcloud/richdocuments/pull/4333

@pedropintosilva The smart picker is also not documented.

pedropintosilva:
Good to have this merged for CODE Red

eszkadev (approved):
looks ok

browser/src/docdispatcher.ts
hfiguiere: it's on my list of things to document.

PR #10731, opened 2024-12-13
Fix Autofilter dialog position on browser zoom

Darshan-upadhyay1110:
- Change autofilter from canvas to whole document
- now on browser zoom because we change the parent autofilter dialog will go beyond the canvas and user can see content
- content will be visible for high browser zoom level
- as we now consider autofilter dialog for whole document and not canvas, i have adjusted the some corner case values to be sure at edge of document dialog stays visible

Change-Id: Ib467c3605ff55bf3d5df33979a923bcc31fe4583

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/AutoCompletePopup.ts
+++ b/browser/src/control/AutoCompletePopup.ts
@@ -162,8 +162,21 @@ abstract class AutoCompletePopup {
 		data.children[0].children[0] = control;
 		(data.children[0].children[0] as TreeWidget).entries = entries;
 
-		data.posx = cursorPos.x;
-		data.posy = cursorPos.y;
+		const isSpreadsheetRTL = this.map._docLayer.isCalcRTL();
+		const canvasEl = this.map._docLayer._canvas.getBoundingClientRect();
+		const offsetX = isSpreadsheetRTL
+			? 0
+			: app.sectionContainer.getSectionWithName(L.CSections.RowHeader.name)
+					.size[0];
+		const offsetY = app.sectionContainer.getSectionWithName(
+			L.CSections.ColumnHeader.name,
+		).size[1];
+
+		if (isSpreadsheetRTL) cursorPos.x = this.map._size.x - cursorPos.x;
+
+		data.posx = cursorPos.x + offsetX + canvasEl.left;
+		data.posy = cursorPos.y + offsetY + canvasEl.top;
+
 		this.sendJSON(data);
 	}
 
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -552,6 +552,10 @@ L.Control.JSDialog = L.Control.extend({
 			return;
 		}
 		this.parentAutofilter = instance.form;
+		const devicePixelRatio = app.dpiScale;
+		// Convert the server-provided px values to screen coordinates
+		instance.posx = instance.posx / devicePixelRatio;
+		instance.posy = instance.posy / devicePixelRatio;
 		var left = parseInt(instance.posx) * scale;
 		var top = parseInt(instance.posy) * scale;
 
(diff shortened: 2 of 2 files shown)


gokaysatir (commented):
I think we could use `app.dpiScale` here, instead of `devicePixelRatio`.

Other than that, @eszkadev knows this part better.

browser/src/control/Control.JSDialog.js
```diff
@@ -572,20 +576,22 @@ L.Control.JSDialog = L.Control.extend({
 		if (isSpreadsheetRTL)
 			left = this.map._size.x - left;
 
+		var canvasEl = this.map._docLayer._canvas.getBoundingClientRect();
+
 		instance.posx = left + offsetX;
```
eszkadev, line 581: why we don't add also the canvas position for X ? We have a plan to add left sidebar at some point, then it will not be 0.

browser/src/control/Control.JSDialog.js
```diff
@@ -669,8 +675,8 @@ L.Control.JSDialog = L.Control.extend({
 		instance.canHaveFocus = !instance.isSnackbar && instance.id !== 'busypopup' && !instance.isAutoCompletePopup;
 		instance.isDocumentAreaPopup = instance.popupParent === '_POPOVER_' && instance.posx !== undefined && instance.posy !== undefined;
 		instance.isPopup = instance.isModalPopUp || instance.isDocumentAreaPopup || instance.isSnackbar;
-		instance.containerParent = instance.isDocumentAreaPopup ? document.getElementById('document-container'): document.body;
 		instance.isAutofilter = instance.isDocumentAreaPopup && this.map._docLayer.isCalc();
+		instance.containerParent = (instance.isDocumentAreaPopup && !instance.isAutofilter) ? document.getElementById('document-container'): document.body;
```
eszkadev, line 679: above isAutofilter is not perfect name - it is not only autofilter but also AutoFill popup too - https://github.com/CollaboraOnline/online/blob/master/browser/src/control/AutoCompletePopup.ts

could you check if auto fill popup still works as expected?
in spreadhseet with split panes (freezed rows and columns) at some distant cell - like Z500

Darshan-upadhyay1110:
@eszkadev The autofill popup seems to be broken. I think it would be best to add an `isAutofilter` flag from the core, so we can adjust the parent element accordingly. Additionally, I believe renaming the `isAutofilter` variable to something more appropriate would be a good idea.

Please do not merge this i will do changes in core and ping you back soon

Darshan-upadhyay1110:
> @eszkadev The autofill popup seems to be broken. I think it would be best to add an `isAutofilter` flag from the core, so we can adjust the parent element accordingly. Additionally, I believe renaming the `isAutofilter` variable to something more appropriate would be a good idea.
> 
> Issue:
> 
>  Peek.2024-12-16.16-45.mp4 
> Please do not merge this i will do changes in core and ping you back soon

Did many changes to fix all cases :)

Ready to review and test

CC: @eszkadev @gokaysatir

browser/src/control/Control.JSDialog.js
```diff
@@ -850,6 +855,17 @@ L.Control.JSDialog = L.Control.extend({
 	},
 
 	updatePosition: function (target, newX, newY) {
+		var width = target.getBoundingClientRect().width;
```
eszkadev, line 858: This function is a setter (only applies the calculated position) - is used not only in autofilter - see `onPan` just above.

Can we separate these two cases? And use new calculation only for auto popups, but onPan and all other old cases will use only setter?
```
updatePosition() {
    here your caclulation code...
    setNewPosition(target, x, y);
}

setNewPosition(target, x, y) {
    here the old code....
}
```

browser/src/control/AutoCompletePopup.ts
```diff
@@ -162,8 +162,21 @@ abstract class AutoCompletePopup {
 		data.children[0].children[0] = control;
 		(data.children[0].children[0] as TreeWidget).entries = entries;
 
-		data.posx = cursorPos.x;
-		data.posy = cursorPos.y;
+		const isSpreadsheetRTL = this.map._docLayer.isCalcRTL();
```
eszkadev, line 165: can't this be handled by the Control.JSDialog code?

I think it should work the same as other cases, we adjust RTL and limit to screen edge there.

eszkadev (approved):

PR #10734, opened 2024-12-16
cool#10630 doc electronic sign: moved esign settings out of UserPrivateInfo

vmiklos:
- **cool#10630 doc electronic sign: moved esign settings out of UserPrivateInfo**
- **cool#10630 doc electronic sign: support per-server private info during 'make run'**


vmiklos:

Julius suggests to not have the esign settings in UserPrivateInfo, since those are "nextcloud instance-wide" settings, configured by admins, which is fair enough. I don't want to route each & every per-server setting through the various laters, so I suggest to go with a ServerPrivateInfo key to host such per-server credentials. [MS-WOPI] doesn't seem to have something similar, but at the same time ServerPrivateInfo is then not a name they would use, so it's probably safe to take.

PR #10735, opened 2024-12-16
private/marco/slideshow dim color

mcecchetti:
- **slideshow: provide support for final dimming to a given color**
- **slideshow: animation color: improving color substitution**


quikee (approved):
Looks good to me.

PR #10738, opened 2024-12-16
slideshow: reuse full-in-window mode from cypress

eszkadev:
We don't want to provide bad experience with low fps when main COOL tab is hidden in the browser (what is not controlled by us). So just don't open new window. Always use main window as a presentation. Only presenter console will be separate window. This way we are sure if someone will want to present - will have main window shown.


quikee (approved):
Looks fine to me.

PR #10740, opened 2024-12-17
Config Improvements

Ashod:
- wsd: config: privatize XmlConfig
- wsd: config: clang-tidy fixes
- wsd: config: hoist invariant string concatenation
- wsd: config: ConfigValueGetter -> ConfigRawValueGetter
- wsd: config: getSafeConfig -> getRawConfig
- wsd: config: simplify
- wsd: config: consistent default values
- wsd: config: use AbstractConfiguration instead of LayeredConfiguration
- wsd: config: extract defaulted values
- wsd: config: take reference instead of pointer
- wsd: config: redact before stringifying
- coolwsd.xml: set default boolean values explicitly
- coolwsd.xml: defaults
- wsd: config: add default config values
- wsd: config: add types
- wsd: config: update the ignore list
- wsd: config: typo
- wsd: config: correct logging_ui_cmd.file defaults
- wsd: config: omit '[0]' for the first entry in an array


common/ConfigUtil.cpp
```diff
@@ -104,7 +104,7 @@ static const std::map<std::string, std::string> DefAppConfig = {
     { "logging.file.property[0][@name]", "path" },
     { "logging.file.property[1]", "never" },
     { "logging.file.property[1][@name]", "rotation" },
-    { "logging.file.property[2]", "false" },
+    { "logging.file.property[2]", "timestamp" },
```
caolanm, line 114: as an aside, is it possible to reorder these to have the name property before its value, reading it backwards "timestamp" as value for "archive" is a pain and I'm not the only one who has occasionally associated the value with the preceding name  :-)

caolanm (approved):

common/ConfigUtil.cpp
Ashod: Yes, agreed. It annoys me as well. Of course I did it with a tool, so it's lexicographical sorting.

Will reverse them. Thanks.

PR #10741, opened 2024-12-17
Misc Improvements

Ashod:
- wsd: string_view in Util::split
- wsd: string_view in Util::trim
- wsd: single-character string literal -> char
- wsd: cosmetic better logging
- wsd: avoid string copies in LogUiCmd::logUiCmdLine()
- wsd: cosmetics in StorageConnectionManager
- wsd: cosmetic

### DIFF
--- a/common/Util.hpp
+++ b/common/Util.hpp
@@ -604,19 +604,16 @@ namespace Util
 
     size_t findInVector(const std::vector<char>& tokens, const char *cstring, std::size_t offset = 0);
 
-    inline std::string& trim(std::string& s, const char ch)
+    /// Trim trailing characters (on the right).
+    inline std::string_view trim(const std::string_view s, const char ch)
     {
         const size_t last = s.find_last_not_of(ch);
         if (last != std::string::npos)
         {
-            s = s.substr(0, last + 1);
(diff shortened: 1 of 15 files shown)


common/Util.hpp
```diff
@@ -604,19 +604,16 @@ namespace Util
 
     size_t findInVector(const std::vector<char>& tokens, const char *cstring, std::size_t offset = 0);
 
-    inline std::string& trim(std::string& s, const char ch)
+    /// Trim trailing characters (on the right).
+    inline std::string_view trim(const std::string_view s, const char ch)
```
caolanm, line 608: After this, we now have two trims. This one where with a 2nd arg which doesn't modify the 1st arg, while the other one with no 2nd arg does modify its string arg.

Checking, it looks like all current uses of this 2 arg trim already assign the return to what was passed as 1st arg (two in common/JailUtil.cpp) so no change in practice, but the discrepancy is annoying

caolanm (approved):

common/Util.hpp
Ashod: Good point. Let me check and eliminate.

PR #10743, opened 2024-12-17
cool#10630 doc electronic sign: limit the offered providers based on the selected country

vmiklos:
- **cool#10630 doc electronic sign: default to a redirect-based provider**
- **cool#10630 doc electronic sign: set defaults without an index**
- **cool#10630 doc electronic sign: return dialog result without an index**
- **cool#10630 doc electronic sign: limit the offered providers based on the selected country**

### DIFF
--- a/browser/src/control/Control.ESignature.ts
+++ b/browser/src/control/Control.ESignature.ts
@@ -365,8 +365,10 @@ namespace cool {
 
 		// Turns a list of provider IDs into a list of signature providers
 		createProviders(providerIds: Array<string>): Array<cool.SignatureProvider> {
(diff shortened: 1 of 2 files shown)


vmiklos:

The first 3 commits are just ground work, and then the last commit makes it possible to have this case that we start up with Germany / d-trust and smart-id not available (in a positive way, it would be broken); and once you switch to e.g. Estonia then smart-id shows up as expected. So the provider list is not a long one with mostly broken entries, but it lists e.g. 4 options for Germany which at least claim to work in that country.

caolanm (approved):
Looks reasonable to me at least

PR #10746, opened 2024-12-17
Synchronise comment blocks with document position

Cwiiis:
Don't postpone and don't animate comment block position changes when scrolling or resizing the document.

Change-Id: I95c1ae1eb732c6dc682018b85debb6133d85ce8c

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -337,9 +337,10 @@ export class Comment extends CanvasSectionObject {
 			lastPosY = childPositions[i].posY + 24;
 		}
 		if (i < this.sectionProperties.childLines.length) {
-			for (let j = i; j < this.sectionProperties.childLines.length; j++)
+			for (let j = i; j < this.sectionProperties.childLines.length; j++) {
 				this.sectionProperties.childLinesNode.removeChild(this.sectionProperties.childLines[i]);
-			this.sectionProperties.childLines.splice(i);
+				this.sectionProperties.childLines.splice(i);
+			}
 		}
 
 	}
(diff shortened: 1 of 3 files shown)


Cwiiis:
Looks like this is a legitimate test failure - there's some conflict that the previous setTimeout on scroll I think was working around when expanding comments where updateChildLines in CommentSection tries to remove a node that isn't attached to the comment.

Ideally, scrolling would only change the position of comments and not do an entire relayout (which is something that could happen in a setTimeout if we want to reposition comments to fill a space - though I don't see this happening without this patch, so probably an aside), so I'm going to take a look at reworking this to do less work. This would have the up-shot of also fixing the potential bad performance of scrolling with comments.

browser/src/canvas/sections/CommentSection.ts
```diff
@@ -337,9 +337,10 @@ export class Comment extends CanvasSectionObject {
 			lastPosY = childPositions[i].posY + 24;
 		}
 		if (i < this.sectionProperties.childLines.length) {
-			for (let j = i; j < this.sectionProperties.childLines.length; j++)
+			for (let j = i; j < this.sectionProperties.childLines.length; j++) {
```
Cwiiis, line 340: For what it's worth, I think the failure was this patch exposing this pre-existing bug.

Cwiiis:
Also for what it's worth, though I think some rearchitecting here would hugely benefit performance and simplify the code, I'm taking an incremental approach - I'll hopefully have follow-up after this patch to make layout cheaper and do it less.

Cwiiis:
I'm closing this in favour of a branch that includes this and the follow-up work to optimise comment layout.

PR #10750, opened 2024-12-17
add in all tiles of equal priority

caolanm:
Change-Id: I2d9120467256ec9b77559a57d0954f4c3fd3f2ee

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/KitQueue.cpp
+++ b/kit/KitQueue.cpp
@@ -433,7 +433,7 @@ TileCombined KitQueue::popTileQueue(float &priority)
     float prioritySoFar = -1000.0;
     for (size_t i = 0; i < _tileQueue.size(); ++i)
     {
-        auto& prio = _tileQueue[i];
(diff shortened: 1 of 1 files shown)


caolanm:
I don't know if we want to do this or not, but we could.

e.g. first page of tiles on connect where there is no specific priority tile currently picks basically the tile at 0,0 and gets the first two rows handled first, maybe we should munge together everything of the same prio?

mmeeks (changes requested):
Seems extremely dangerous - those tiles could be far away - and create a huge bbox to render. Possibly it makes sense to have some deeper banding - perhaps two rows high or somesuch =)

I guess pre-fetching should also be taught about tile-combining in the JS - so we fetch wider strips - since (I guess) they are at least as fast as rendering just one tile but ... not sure about this at all.

PR #10751, opened 2024-12-17
slideshow: when presenter window is resized, slide is not

mcecchetti:
Now that the presenter window is an iframe, its no more able to trigger the
resize event so we need to attach it to the main window.

Change-Id: I208aa90fd52bfa956bd30eebd6e8cd03538776a1


eszkadev (approved):
looks ok

PR #10752, opened 2024-12-17
add missing files to the install dir

timar:
We need
- src/layer/tile/CanvasTileUtils.js (included from TileWorker.js) and
- src/layer/tile/TileWorker.js (because we have `this._worker = new Worker('src/layer/tile/TileWorker.js');` in bundle.js

These files have to be present separately in install dir. Whether it's a permanent solution, or a temporary hack, I don't know.

Change-Id: I9d0d21463be0cef10ff95a0cd80226a5fc29fa75


timar:
Let's try this in red CODE first. Unfortunately, because of this bug, the TileWorker optimization patches have never been tested anywhere in production, only on developer's computer.

PR #10754, opened 2024-12-18
fix for log UI Commands

aszucs3:
Save documentID to the log, and rewrote reporting script to reorder the file. (because looks like some cases more kit processes wrote into the permanent file at the same time, and their log mixed)

fixed a configuration path problem
fixed a click merge problem

Change-Id: Ibf1281b560d94c7ea9a97614dd1648aaa10f00f7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -697,7 +697,6 @@ namespace Log
                 channelUILog->setProperty(pair.first, pair.second);
             }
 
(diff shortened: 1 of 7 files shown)


caolanm (approved):
looks good, should fix the misplaced log file problem

PR #10758, opened 2024-12-18
feat(iOS): Improve clipboard handling

Minion3665:
The iOS clipboard handling code has up-until-now not supported much stuff that needed rich context. This patch allows us to store the whole internal clipboard on the system clipboard, letting us copy/paste things like images, or formulas where all referenced cells are included in the copy. This still isn't quite an internal paste, so it doesn't allow things like copying formulas within the same spreadsheet that don't include all their references.


tml1024 (approved):
Looks fine to me, and the commit messages show that you have investigated this well enough.

PR #10764, opened 2024-12-18
fix(iOS): Allow moving shapes

Minion3665:
Since the fix for tdf#123874, we've stopped sending "shapeselectioncontent" to iOS. That message provided an SVG which, elsewhere, is used to preview the movement of the shape.

When modifying the preview, we correctly gated it in a check that the SVG was set. Unfortunately, we also introduced a dependency on this SVG ƒor other things: critical pieces for moving shapes such as preventing panning when a shape is being dragged. Instead, it's better to restrict the dependency on the SVG to only where it's needed, as that allows iOS to move shapes even without the SVG existing.

Change-Id: I69c9ea25011dedd31c492e56bf0ef4aca5cfb21f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/ShapeHandlesSection.ts
+++ b/browser/src/canvas/sections/ShapeHandlesSection.ts
@@ -897,12 +897,14 @@ class ShapeHandlesSection extends CanvasSectionObject {
 	}
 
 	onMouseMove(position: number[], dragDistance: number[]) {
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks good, but I cannot test

PR #10768, opened 2024-12-19
wopi: add UserCanOnlyComment option

JaumePujantell:
If UserCanWrite and UserCanOnlyComment are set to true the document is
opened in comment only mode. If UserCanWrite is false,
UserCanOnlyComment is ingnored.

Change-Id: I3417aadc428e079bb1533e2432ea60a1e62d03fb


hfiguiere:
My only concern here is that `UserCanOnlyComment` is presented as a higher permission of `UserCanWrite`. So if the integration return `UserCanOnlyComment` (along with `UserCanWrite`) to a version of COOL that doesn't understand it, it will basically grant the write permission, while I think in that case it should revert to read only.

I believe that the proper thing would be to set `UserCanOnlyComment` to allow commenting, without `UserCanWrite` being set to true.

hfiguiere (approved):
LGTM

PR #10769, opened 2024-12-19
COOL_JS.m4 constantly grows in size

caolanm:
and gets multiple duplicates of "script src=" lines in it

and on switching branches, etc there can be entries in it that doesn't exist, which then causes unit-integration to fail on the Fileserver not able to provide all files referenced in cool.html

presumably since:

commit e30367d6344a87787b816a9efd92294711e8c5ba
CommitDate: Wed Aug 28 12:44:57 2024 +0200

    input file to m4 macro through intermediate file, not parameter

Change-Id: Id48ff3b420649532740958e7fed86daf9fa64e99

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -769,10 +769,9 @@ $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist: $(COOL_JS_DST)
 	if cmp -s $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist; then rm $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp; else mv $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist; fi
 
 $(INTERMEDIATE_DIR)/COOL_JS.m4: $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist
-	@rm -f COOL_JS.m4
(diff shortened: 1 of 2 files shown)


browser/Makefile.am
```diff
@@ -769,10 +769,9 @@ $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist: $(COOL_JS_DST)
 	if cmp -s $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist; then rm $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp; else mv $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist.tmp $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist; fi
 
 $(INTERMEDIATE_DIR)/COOL_JS.m4: $(INTERMEDIATE_DIR)/COOL_JS.m4.jslist
-	@rm -f COOL_JS.m4
-	$(foreach file,$(NODE_MODULES_JS) $(COOL_LIBS_JS) $(patsubst %.ts,%.js,$(COOL_JS_WEBORDER)), $(shell echo -n $(file), >> $(INTERMEDIATE_DIR)/COOL_JS.m4))
-# remove last ","
-	@truncate -s -1 $(INTERMEDIATE_DIR)/COOL_JS.m4
+	$(foreach file,$(NODE_MODULES_JS) $(COOL_LIBS_JS) $(patsubst %.ts,%.js,$(COOL_JS_WEBORDER)), $(shell echo -n $(file), >> $(INTERMEDIATE_DIR)/COOL_JS.m4.tmp))
```
vmiklos, line 772: Aha, Tor was also complaining about this, that the path for 'rm' and 'echo' is not the same. I was confused by what 'truncate -s -1' does, but I see it removes the last byte of the file.

One day we should really go to some more mainstream way of budling (e.g. richdocuments uses webpack), but this fixes the immediate problem.

vmiklos (approved):
Looks good, thanks for investigating this.

Issue #10762, opened 2024-12-18
Design bits - New comment in cells look odd when the cell is extended


pedropintosilva:
As stated in https://github.com/CollaboraOnline/online/pull/10254 and as discussed in the COOL meeting: the cell's comment indicator need polishing.

When the cell is at the default size, the comment looks OK
![image](https://github.com/user-attachments/assets/a1152797-a93f-421d-afa7-a0b24241d4fc)

 When the cell is extended 

![image](https://github.com/user-attachments/assets/f014116b-bff1-4d8d-97e4-48bb1227b17f)

I would like to see the 3 following enhancements/follow ups:
* The triangle not being drawn on top of the cell border (see the screenshot it is above the right cell border)
* The right edge of the triangle should have the same length as the top edge of the triangle
* The color should come from the palette or dark palette css instead (css var)

And to reiterate what I said beforehand: the ratio should never change (point 2)

PR #10771, opened 2024-12-19
Adjust comment styling for extended cells to ensure consistent appearance

banobepascal:
Change-Id: Ia3b6ac17ad1f4677e61d9128806bac90a819bfb3

* Resolves: #10762 
* Target version: master 

### Summary
This fix addresses an issue where comments in cells appeared misaligned or visually inconsistent when the cell was extended

### Visual Representation
Screenshot from 2024-12-25 11-43-11
Screenshot from 2024-12-25 11-43-37

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -1350,18 +1350,25 @@ export class Comment extends CanvasSectionObject {
 
 				var cellSize = this.calcCellSize();
 				if (cellSize[0] !== 0 && cellSize[1] !== 0) { // don't draw notes in hidden cells
-					// For calc comments (aka postits) draw the same sort of square as ScOutputData::DrawNoteMarks
-					// does for offline
-					var margin = cellSize[0] * 0.06;
-					var squareDim = 6;
+					// `zoom` represents the current zoom level of the map, retrieved from `this.map.getZoom()`.
+					// `baseSize` is a constant that defines the base size of the square at the initial zoom level.
+					// `squareDim` calculates the dimension of the square, which dynamically adjusts based on the current zoom level.
+					// The dimension increases proportionally to the zoom level by adding `zoom` to `baseSize`.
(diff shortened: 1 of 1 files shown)


banobepascal:
@eszkadev please review

eszkadev (changes requested):
I think it requires fix for RTL sheet `layout->RTL` (not confuse with RTL UI `<URL>&lang=ar`)

rtl

Also it is now not aligned to the cell border (has some gap), is it intended?

banobepascal:
> I think it requires fix for RTL sheet `layout->RTL` (not confuse with RTL UI `<URL>&lang=ar`)
> 
> rtl
> 
> Also it is now not aligned to the cell border (has some gap), is it intended?

Removed gap and also fixed RTL 

Screenshot from 2024-12-25 11-43-37
Screenshot from 2024-12-25 11-43-11

banobepascal:
@eszkadev please review the changes

eszkadev:
I can confirm now it is better and RTL works.
I found that with lower zoom we still have the same size of triangle as with large zoom.
So with 50% it looks bigger than a cell and with 200% small:
zoom

Also in RTL there is 1px gap on the left edge if we want to be pixel-perfect.

banobepascal:
> I can confirm now it is better and RTL works. I found that with lower zoom we still have the same size of triangle as with large zoom. So with 50% it looks bigger than a cell and with 200% small: zoom
> 
> Also in RTL there is 1px gap on the left edge if we want to be pixel-perfect.

The triangle is now adjusting according to the zoom level
Screenshot from 2024-12-26 17-24-53
Screenshot from 2024-12-26 17-24-23


The reason for the RTL margin 1px is because of the column borders are not straight with the top bar so if the column or row borders are adjusted this creates a misalignment.
Screenshot from 2024-12-26 17-22-20

eszkadev (approved):
works now :)

PR #10774, opened 2024-12-19
ci(iOS): bump build version

Minion3665:
Due to an expired xcode login session on our CI, I noticed that the version of xcode we were running was incompatible with the version of MacOS we were running.

To rectify this, I updated xcode. Unfortunately, the latest versions of xcode don't seem to have support for iOS 14.5 builds. Updating the deploy target to the earliest supported version (16.3) fixes the issue.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
Lets get it in now so ci can be unblocked in the meantime I guess

PR #10775, opened 2024-12-19
slideshow: followup: exit  from notes view before starting

mcecchetti:
This is a follow up since previous patch has a problem when presenter console is
started in Firefox when Impress is in notes view mode.
Oddly, it seems that presentation information are requested but the reply with
the info never reaches CanvasTileLayer._onMessage.
Anyway I can see that the reply is sent by the core.
Since I'm not able to understand the root cause, I implemented a simple
workaround by delay the presenter console start by 500ms.
In order to be safer I decided to apply the same delay also when we present
without the console.

Change-Id: I941419b3d33095608973af18aab9df2b767a02f4

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -193,7 +193,7 @@ class PresenterConsole {
 		if (this._waitForExitingNotesMode && e.mode === 0) {
 			this._waitForExitingNotesMode = false;
 			this._map.off('impressmodechanged', this._onImpressModeChanged, this);
-			this._onPresentInConsole();
+			setTimeout(this._onPresentInConsole.bind(this), 500);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


hcvcastro:
Looks good

hcvcastro (approved):

hcvcastro:
almost forgot, minor change .. proxyWindow.setTimeout

mcecchetti:
Thanks Henry for reviewing

mcecchetti:
> almost forgot, minor change .. proxyWindow.setTimeout

I guess, you mean `_proxyPresenter`. Anyhow, iIt has not yet been created. ;-)

Darshan-upadhyay1110 (approved):

mcecchetti:
> @mcecchetti better to rerun single failing job, not all using close/reopen. I see unstable tests in at least 3 jobs so it could be very hard to get them all green at once :)

it makes sense, but I'm not able to access Jenkins

PR #10776, opened 2024-12-19
Fix mousewheel scrolling sometimes not working in one direction

Cwiiis:
Scroll extents were not being recalculated when attempting to scroll in ScrollSection, causing scrolling initiated with the mouse not to work sometimes. This was especially noticeable with Impress notes view as scroll extents weren't otherwise recalculated elsewhere.


gokaysatir (approved):
Seems reasonable, thank you for the fix.

PR #10777, opened 2024-12-19
don't forward signals to pid -1

caolanm:
Change-Id: I22386110225bb9a0c0052196fb0fc4059245f872

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks, @caolanm. I checked other `kill(2)` calls for the same oversight, and this seems to be the only one.

Issue #10770, opened 2024-12-19
Cell handle appears offset while scrolling in Calc


Cwiiis:
If you scroll with the mouse-wheel, the little square that appears on the bottom-right of the selected cell range appears offset during the scroll animation. Once the animation finishes, it appears in the correct position. This seems to be a regression from #10509, but I expect is the symptom of a pre-existing problem.

Cwiiis:
This is fixed by #10783

PR #10778, opened 2024-12-19
Fix the calc cursor handle being offset during animated scroll

Cwiiis:
Introduce an onAnimationStep callback for animating canvas objects so they can run animation logic outside of the drawing callgraph. Use this for scroll animations in ScrollSection.

* Resolves: #10770
* Target version: master

### DIFF
--- a/browser/src/canvas/CanvasSectionContainer.ts
+++ b/browser/src/canvas/CanvasSectionContainer.ts
@@ -2082,13 +2082,16 @@ class CanvasSectionContainer {
 			this.continueAnimating = false;
 		}
 
+		const section: CanvasSectionObject = this.getSectionWithName(this.getAnimatingSectionName());
+		if (section)
+			section.onAnimationStep(this.frameCount, this.elapsedTime);
+
 		if (this.continueAnimating) {
 			this.drawSections(this.frameCount, this.elapsedTime);
 			this.frameCount++;
 			requestAnimationFrame(this.animate.bind(this));
 		}
 		else {
-			var section: CanvasSectionObject = this.getSectionWithName(this.getAnimatingSectionName());
 			if (section) {
 				section.isAnimating = false;
 				section.onAnimationEnded(this.frameCount, this.elapsedTime);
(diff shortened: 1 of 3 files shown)


browser/src/canvas/CanvasSectionContainer.ts
```diff
@@ -2082,13 +2082,16 @@ class CanvasSectionContainer {
 			this.continueAnimating = false;
 		}
 
+		const section: CanvasSectionObject = this.getSectionWithName(this.getAnimatingSectionName());
```
gokaysatir, line 2085: I think we don't need to add this. onDraw function is called on animation steps. It looks like moving the last if block to the top of onDraw function would do the same thing (in ScrollSection.ts)? Maybe we can find a way withouth adding this function. Thanks for checking the issue:)

gokaysatir (changes requested):

browser/src/canvas/CanvasSectionContainer.ts
Cwiiis: This wouldn't do the same thing - this is moving the altering of scroll position outside of the middle of the drawing chain. The problem isn't fixed by altering the scroll position before drawing the scrollbars. The scroll position has external dependencies and needs to be changed before drawing begins, not in the middle of it.

Cwiiis: The other thing we could do is add an `onBeforeDraw` function, I suppose - the result would be the same, although the impact would be slightly greater (it's negligible either way, but this way, only animated containers are affected, if we add an `onBeforeDraw`, every container will be affected, animating or not).

gokaysatir:
@Cwiiis i guess this issue appeared after recent works on scroll section. I'd expect to see this issue earlier even before improvements on scroll section though.

Anyway, this is about drawing orders of sections, can we use something like in below PR for now, if it has no downsides?
* https://github.com/CollaboraOnline/online/pull/10783

Then we can of course add animation step if we encounter something we can't solve other way.

gokaysatir (changes requested):
We had a solution playing with the drawing and processing orders. Let's keep this out for now. And thanks for investigating the issue :)

eszkadev:
the other PR was merged to solve issue, can we close this one?

PR #10779, opened 2024-12-20
wsd: do not complain if process is reaped already

Ashod:
When a process is already reaped, waitpid(3) will
fail with errno == 0 (success). Instead of blindly
warning, we for success in errno and therefore
have nothing to warn about.

Change-Id: I4d2d35a58eb3cfe6b70b0d2593e62f4ee1015581


caolanm (approved):
looks sensible to me

PR #10783, opened 2024-12-20
Re-adjust z-index of Scroll section.

gokaysatir:
Because other sections need to read the up-to-date data that is updated by scroll section, we need to set its z-index to a higher value.

Issue: AutoFill marker section drawing lags behind while scrolling.

Also refactored some of the things in the file, without changing the logic.

Change-Id: Ia2305efaaa0bbca5bd59881c5e246e7894bc5043

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/CanvasSectionProps.js
+++ b/browser/src/canvas/CanvasSectionProps.js
@@ -15,6 +15,14 @@
  * So we have this file, to manage their orders easily. Define them here, globally. Then you can use from everywhere.
  * Refer to CanvasSectionContainer.ts for definitions of processingOrder, drawingOrder and zIndex.
  */
+
+/*
+    z-index        : Higher zIndex will be drawn on top.
+    Processing and drawing orders are meaningful between sections with the same zIndex.
+    Processing order	: Important for locations and sizes of sections. Lowest processing order will be processed first.
+    Drawing order	: Highest with the same zIndex will be drawn on top.
+*/
+
 /* global L */
 
 L.CSections = {};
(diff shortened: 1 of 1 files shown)


eszkadev:
@gokaysatir it's RED now, do we want ot include this?

browser/src/canvas/CanvasSectionProps.js
```diff
@@ -26,32 +34,32 @@ L.CSections.Tiles = 				{ name: 'tiles'				, zIndex: 5 };
 L.CSections.Overlays =				{ name: 'overlay'			, zIndex: 5 };
 L.CSections.CalcGrid = 				{ name: 'calc grid'			, zIndex: 5 };
 L.CSections.Debug.Splits = 			{ name: 'splits'			, zIndex: 5 };
-L.CSections.Debug.TilePixelGrid = 	        { name: 'tile pixel grid'	, zIndex: 5 };
-L.CSections.Debug.PreloadMap = 	                { name: 'preload map'	        , zIndex: 5 };
-
+L.CSections.Debug.TilePixelGrid = 	{ name: 'tile pixel grid'	, zIndex: 5 };
```
eszkadev, line 37: looks like alignment is worse now

gokaysatir: I think it looks like that here. In code editor, it's normal.

gokaysatir:
> @gokaysatir it's RED now, do we want ot include this?

Yes i guess we can have it in. This is bound to 2 different issues.

eszkadev (approved):
nice, fixes the issue with "auto fill marker" no in sync while scrolling/zooming

PR #10784, opened 2024-12-20
sc autofill: make cells selectable after first click - if autofill context menu is open

bayramcicek:
- disable mouse clicks for context-menu-layer
- close popups if a dropdown menu is opened -> this only applies to the autofill context-menu. other context menus should not be affected.
- close all popups when scrolling and reset selection with _resetReferencesMarks() if autofill context menu is open
- close all popups if a row/column header is selected

Change-Id: Iaf7b5c09003291f27155ae9ab6787f2f871ac466

* Resolves: # 
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/canvas/sections/ScrollSection.ts
+++ b/browser/src/canvas/sections/ScrollSection.ts
@@ -1148,6 +1148,8 @@ export class ScrollSection extends CanvasSectionObject {
 	public onMouseWheel (point: Array<number>, delta: Array<number>, e: MouseEvent): void {
 		if (e.ctrlKey) return;
 
+		this.map.fire('closepopups'); // close all popups when scrolling
+
 		let hscroll = 0, vscroll = 0;
 		if (Math.abs(delta[1]) > Math.abs(delta[0])) {
 			if (e.shiftKey)
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -109,6 +109,7 @@ L.Control.ContextMenu = L.Control.extend({
 
 	onAdd: function (map) {
 		this._prevMousePos = null;
+		this._autoFillContextMenu = false;
 
 		map._contextMenu = this;
 		map.on('locontextmenu', this._onContextMenu, this);
@@ -119,6 +120,12 @@ L.Control.ContextMenu = L.Control.extend({
 	},
 
 	_onClosePopup: function () {
+
+		if (this._autoFillContextMenu) {
+			this._autoFillContextMenu = false;
+			app.map._docLayer._resetReferencesMarks();
+		}
+
 		$.contextMenu('destroy', '.leaflet-layer');
 		this.hasContextMenu = false;
 	},
@@ -168,7 +175,7 @@ L.Control.ContextMenu = L.Control.extend({
 			} else if (menuItem.indexOf('.uno:AutoFill') !== -1) {
 				// we should close the autofill preview popup before open autofill context menu
 				map.fire('closeautofillpreviewpopup');
-				autoFillContextMenu = true;
+				this._autoFillContextMenu = autoFillContextMenu = true;
 				break;
 			}
 		}
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -204,6 +204,12 @@ L.Control.ContextMenu = L.Control.extend({
 						var $menu = opt.$menu;
 						$menu.attr('tabindex', 0); // Make the context menu focusable
 					},
+					activated: function (opt) {
+						if (autoFillContextMenu) {
+							var $layer = opt.$layer;
+							$layer.css('pointer-events', 'none'); // disable mouse clicks for the layer
+						}
+					},
 					hide: function() {
 						if(autoFillContextMenu)
 							app.map._docLayer._resetReferencesMarks();
--- a/browser/src/control/jsdialog/Util.Dropdown.js
+++ b/browser/src/control/jsdialog/Util.Dropdown.js
@@ -154,6 +154,9 @@ JSDialog.OpenDropdown = function (id, popupParent, entries, innerCallback, popup
 		};
 	};
 	L.Map.THIS.fire('jsdialog', {data: json, callback: generateCallback(entries)});
+
+	L.Map.THIS.fire('closepopups'); // close popups if a dropdown menu is opened
+	L.Map.THIS._docLayer._resetReferencesMarks();
 };
 
 JSDialog.CloseDropdown = function (id) {
--- a/browser/src/control/jsdialog/Util.Dropdown.js
+++ b/browser/src/control/jsdialog/Util.Dropdown.js
@@ -153,10 +153,8 @@ JSDialog.OpenDropdown = function (id, popupParent, entries, innerCallback, popup
 				JSDialog.CloseDropdown(id);
 		};
 	};
-	L.Map.THIS.fire('jsdialog', {data: json, callback: generateCallback(entries)});
-
 	L.Map.THIS.fire('closepopups'); // close popups if a dropdown menu is opened
-	L.Map.THIS._docLayer._resetReferencesMarks();
+	L.Map.THIS.fire('jsdialog', {data: json, callback: generateCallback(entries)});
 };
 
 JSDialog.CloseDropdown = function (id) {
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -635,6 +635,8 @@ export class Header extends app.definitions.canvasSectionObject {
 		L.DomUtil.enableImageDrag();
 		L.DomUtil.enableTextSelection();
 
+		this._map.fire('closepopups'); // close all popups if a row/column header is selected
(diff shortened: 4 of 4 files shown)


bayramcicek:
- this fixes the double click issue on autofill contextmenu. but prevents the contextmenu from closing itself when scrolling and opening another popups...

- a working example:

https://github.com/user-attachments/assets/1ddbd17c-8722-4a16-9fc9-073ce4163fcc

eszkadev:
is it finished? or still draft/wip ?

bayramcicek:
> is it finished? or still draft/wip ?

still WIP.  I'm trying to fix some issues.

bayramcicek:
### before: (master)

https://github.com/user-attachments/assets/2a9e057b-f63b-4cbd-ba23-d7e3951a7cd7


### after:



https://github.com/user-attachments/assets/c0cee4e6-cbf7-49a6-9709-049ce9dae3d7

browser/src/control/Control.ContextMenu.js
```diff
@@ -131,6 +132,17 @@ L.Control.ContextMenu = L.Control.extend({
...
+		if (this._autoFillContextMenu) {
+			this._autoFillContextMenu = false;
+
+			var latlng = e.latlng;
+			var mousePos = this._map._docLayer._latLngToTwips(latlng);
+
+			// generate a left mouse click so that a single click can select a cell
+			// after closing the autofill context menu
+			this._map._docLayer._postMouseEvent('buttondown', mousePos.x, mousePos.y, 1, 1, 0);
+			this._map._docLayer._postMouseEvent('buttonup', mousePos.x, mousePos.y, 1, 1, 0);
```
eszkadev, line 144: hmm, isn't the autofill menu generated in online?
or do we pass it from core? (I don't remember)

eszkadev:
did you test that with multiple users?
A: open autofill
B: tries to click to move cursor somewhere

does the autofil from A view have impact on B?

browser/src/control/Control.ContextMenu.js
bayramcicek: autofill contextmenu is generated in online.

bayramcicek:
> did you test that with multiple users? A: open autofill B: tries to click to move cursor somewhere
> 
> does the autofil from A view have impact on B?

my experience with multiple users seem work as expected. (or should it behave different?) 

https://github.com/user-attachments/assets/237a7345-4cfc-4c7f-870b-f7c978422c0a

browser/src/control/Control.ContextMenu.js
bayramcicek: > but does it take any data from core to create that popup? if NOT use core at all to invoke popup -> this approach is possibly wrong -> why we would need to do additional click in core while popup was online only? or I missed something?

yes, that's right. Is there any way to do additional click on the online? I thought `_postMouseEvent` is on the online.

eszkadev: Then - adding a click feels like workaround. It would be good to be able to answer: `why we need additional click?`

Do we have present `.jsdialog-overlay` on the screen maybe? Or the `map/canvas` is not in focus?
Do you have some logs from the browser - what happens just after the first click?

bayramcicek: > Then - adding a click feels like workaround. 
- yes. we can say that it is a workaround.

> It would be good to be able to answer: `why we need additional click?`
- after opening the autofill contexmenu there is a  id="context-menu-layer" that blocks scroll and clicking on the cells.
- if we click on a cell -to select it-, it does not click the cell but removes the layer. So if we add an additional click at this point, then we can click on the cell we are on - after that layer was removed - without need to double click. 

> Do we have present `.jsdialog-overlay` on the screen maybe? 
- we have `id="context-menu-layer"`

Screenshot_20250103_163658

> Or the `map/canvas` is not in focus? 
gets focus after contextmenu is hidden.

`Control.ContextMenu.js:219`
```js
					hide: function() {
						if(autoFillContextMenu)
							app.map._docLayer._resetReferencesMarks();
						map.focus();
					}
```

Do you have some logs from the browser - what happens just after the first click?

`hide: function()` ^^^ executes and `id="context-menu-layer"` is removed. then auto contextmenu is closed.

eszkadev: I see. what about something like:
```
diff --git a/browser/src/control/Control.ContextMenu.js b/browser/src/control/Control.ContextMenu.js
index aecb8a6fe7..c7a13d264c 100644
--- a/browser/src/control/Control.ContextMenu.js
+++ b/browser/src/control/Control.ContextMenu.js
@@ -204,9 +204,22 @@ L.Control.ContextMenu = L.Control.extend({
                                                var $menu = opt.$menu;
                                                $menu.attr('tabindex', 0); // Make the context menu focusable
                                        },
+                                       activated: function (opt) {
+                                               console.error('activated');
+                                               console.error(opt);
+
+                                               var $layer = opt.$layer;
+                                               $layer.css('pointer-events', 'none'); // layer doesn't get mouse clicks
+
+                                               $layer.on('mousedown', function () {
+                                                       console.error('This click handler will not be fired');
+                                               });
+                                       },
                                        hide: function() {
-                                               if(autoFillContextMenu)
+                                               if(autoFillContextMenu) {
                                                        app.map._docLayer._resetReferencesMarks();
+                                                       console.error('but this will be still fired thanks to closing context menu on map click');
+                                               }
                                                map.focus();
```

`activated` event is called when context menu is built and ready to use.
We can just remove the layer on show but.. perfect would be to just not handle clicks to avoid errors inside jQuery.

you can read also internals of jQuery: `browser/node_modules/jquery-contextmenu/src/jquery.contextMenu.js` for interesting details

bayramcicek: > I see. what about something like:
> 
> ...
> 
> `activated` event is called when context menu is built and ready to use. We can just remove the layer on show but.. perfect would be to just not handle clicks to avoid errors inside jQuery.

Yes. This is better than removing the `context-menu-layer` on `show()`. However there are some issues like: when we scroll down or open the e.g. color palette, the autofill context menu is still there:


https://github.com/user-attachments/assets/5f6f6b98-baf7-4ac8-af91-a989e86c87fb

I will try to detect if user clicks on the `class="leaflet-layer"` - without removing the `context-menu-layer`.

bayramcicek: - [x] scrolling on the class="leaflet-layer" does not close the autofill context menu. -> Done
- [ ] selecting a row/column does not close the autofill context menu. -> trying...

browser/src/canvas/sections/ScrollSection.ts
```diff
@@ -1148,6 +1148,9 @@ export class ScrollSection extends CanvasSectionObject {
 	public onMouseWheel (point: Array<number>, delta: Array<number>, e: MouseEvent): void {
 		if (e.ctrlKey) return;
 
+		this.map.fire('closepopups'); // close all popups when scrolling
+		this.map._docLayer._resetReferencesMarks();
```
eszkadev, line 1152: does that function exists aslo outside of calc?

browser/src/control/jsdialog/Util.Dropdown.js
```diff
@@ -154,6 +154,9 @@ JSDialog.OpenDropdown = function (id, popupParent, entries, innerCallback, popup
 		};
 	};
 	L.Map.THIS.fire('jsdialog', {data: json, callback: generateCallback(entries)});
+
+	L.Map.THIS.fire('closepopups'); // close popups if a dropdown menu is opened
+	L.Map.THIS._docLayer._resetReferencesMarks();
```
eszkadev, line 159: this should be done earlier (before jsdialog message which will create potential popup)

browser/src/canvas/sections/ScrollSection.ts
eszkadev: isn't that causing that even if autofill popup is not opened, but we did some selection eg. for doing formula - it will disappear when scrolling?

type: `=SUM(`, select some range, scroll

eszkadev: I think we want to just close popups -> then inside autofill implementation, it should reset references if is closing

bayramcicek: I think this is done with the new update.
Thanks for pointing out the issue on formula popup -> fixed.

browser/src/control/jsdialog/Util.Dropdown.js
bayramcicek: I just put that before the `L.Map.THIS.fire('jsdialog', ....` line. I hope it is ok.

browser/src/control/Control.ContextMenu.js
bayramcicek: So, now I'm going to close the popup when selecting a row/column header...

eszkadev (approved):
looks good now, thanks :)

browser/src/control/Control.ContextMenu.js
bayramcicek: selecting a row/column does not close the autofill context menu. -> Done with the latest update.

PR #10789, opened 2024-12-21
Focus inside input field on show show input modal

Darshan-upadhyay1110:
Change-Id: I5a29ddf7b4562221fc8a6c574a775d2a2e9180a8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -397,22 +397,25 @@ L.Control.JSDialog = L.Control.extend({
 		// this will only search in current instance and not in whole docuemnt
 		const tabControlWidget = this.findTabControl(instance);
 
-		let focusWidget ;
+		let focusWidget, firstFocusableElement ;
 
 		if (tabControlWidget) {
 			// get DOM element of tabControl from current instance
 			focusWidget = instance.content.querySelector('[id="' + tabControlWidget.id + '"]');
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -1570,6 +1570,9 @@ L.Control.UIManager = L.Control.extend({
 				that.closeModal(dialogId);
 			}}
 		]);
+
+		// focus on input modal on show
+		document.getElementById('input-modal-input').focus();
```
eszkadev, line 1575: We should try to avoid `getElementById` where possible.
Also: why not put this inside `showModal` function? then in all places where we show modal we will focus it, not only in this case

eszkadev: if we have to, better to query inside container: `container.querySelector('#id')`
that makes us sure we will not get element from some other place than our current component/dialog

Darshan-upadhyay1110: @eszkadev found better way :)

It seems it was a regression from my commit when i did fix for tabcontrol focus https://github.com/CollaboraOnline/online/pull/10795

eszkadev (approved):
looks ok

eszkadev:
can we have some cypress test which will check that initial focus?

PR #10793, opened 2024-12-23
Add missing enum `MissingSSLError` and fix typo

Darshan-upadhyay1110:
Change-Id: Ibe3a8417b7ea54cfaa00857a5ed37bbf6f25e1dd

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -389,7 +389,7 @@ asyncConnect(const std::string& host, const std::string& port, const bool isSSL,
     if (isSSL)
     {
         LOG_ERR("Error: isSSL socket requested but SSL is not compiled in.");
-        asyncCb(nullptr, asyncConnectResult::MissingSSLError);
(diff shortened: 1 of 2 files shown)


meven (approved):
I don't understand how this could have passed CI.

meven:
> I don't understand how this could have passed CI.

I see this was in the !SSL ifdef

Darshan-upadhyay1110:
> > I don't understand how this could have passed CI.
> 
> I see this was in the !SSL ifdef

Yes ,  should we add a test case for non-ssl flow ?

meven:
> > > I don't understand how this could have passed CI.
> > 
> > 
> > I see this was in the !SSL ifdef
> 
> Yes , should we add a test case for non-ssl flow ?

That would be valuable to prevent those kinds of issues but that's not simple to put together, would need a separate CI build maybe.

bayramcicek (commented):
`make run` works well after this change. 
Thanks.

PR #10800, opened 2024-12-24
Refocus comment when needed automatically

lpranam:
* Target version: master 

### Checklist


gokaysatir (approved):
It feels like we need to set "needFocus" to null in more placees. But i couldn't find an edge case. This is a useful feature, thanks for the improvements.

PR #10803, opened 2024-12-25
Fix toggle status of show all comments button

GulsahKose:
Before initial state was always true.
Now we decide to initial state according to document have comments or not. If document have comments, default initial state is comments on, if not comments off.

Change-Id: If3e8eb5eeff29c8a72658d0afafc7379444d62ab

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -2359,9 +2359,8 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 			this.checkSize();
 			this.update();
 		}
-
-		var show = this.map.stateChangeHandler.getItemValue('showannotations');
-		this.setView(show === true || show === 'true');
+		var initalShowAnnotations = commentList.length > 0;
+		app.map.showComments(initalShowAnnotations);
 
 		var showResolved = this.map.stateChangeHandler.getItemValue('ShowResolvedAnnotations');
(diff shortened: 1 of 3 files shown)


browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -2359,9 +2359,8 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 			this.checkSize();
 			this.update();
 		}
-
-		var show = this.map.stateChangeHandler.getItemValue('showannotations');
-		this.setView(show === true || show === 'true');
+		var initalShowAnnotations = commentList.length > 0;
+		app.map.showComments(initalShowAnnotations);
```
eszkadev, line 2363: This makes it out of sync when inserting a comment to document which didn't have comments before.
Then you have to click 2 times "show comments" to acctually toggle.

I think it should stay as it was (default is on) - less problems.

As improvement we can probably remember the last state which user had in the local browser settings?  cc: @pedropintosilva 
Or maybe it is saved in the document? Could you try with odt and docx, if they remember the last status after we save and reopen document in LibreOffice?

eszkadev (changes requested):
comment above

eszkadev:
can we close PR as we talked on community meeting it's not needed? or the other fix will be done?
cc @pedropintosilva

PR #10806, opened 2024-12-25
Fix regression from c0e53f04741987b70

timar:
The intent was to avoid VLA, but later in the code the sizeof operator returned 16 and the length of token[4] was 256. This mismatch caused a regression, admin console's secure password was not accepted.
I decided to use std::vector instead of alloca. It's more readable and C++-like to me.

Change-Id: I3cbbce5c4ac568735b2b44dad2fb20e143bf4c75


mmeeks (approved):
Looks sensible to me.

vmiklos:
Oh, this was my fault, thanks for fixing.

PR #10808, opened 2024-12-26
Fix width of Thesaurus left arrow button

timar:
SVG images may not have an intrinsic width and height defined. This can cause max-content to behave unexpectedly since it relies on the size of the content, and an SVG without dimensions may not contribute to the content's width.

We set width explicitely in this case.

Change-Id: I5ebed738bd4f3a12a96596ddac07341097e43bc3

Before:
image

After:
image

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -170,6 +170,11 @@ button.has-img {
 	min-width: auto !important;
 }
 
+/* Thesaurus <- button */
+button#left.has-img > img {
+	width: 16px;
+}
+
 /* Calc Auto-Filter buttons coming from core */
 [data-theme='dark'] button#select_current.has-img > img {
 	filter: invert(.9)


browser/css/btns.css
```diff
@@ -170,6 +170,11 @@ button.has-img {
 	min-width: auto !important;
 }
 
+/* Thesaurus <- button */
+button#left.has-img > img {
```
eszkadev, line 174: better to add ID of the dialog before to target only specific dialog ("left" is popular name probably)
But shouldn't break anything as icons inside buttons will be similar size.

eszkadev (approved):

eszkadev:
cc: @pedropintosilva @banobepascal
can we have some generic fix for all the svgs with the similar problem?

pedropintosilva:
> cc: @pedropintosilva @banobepascal
> can we have some generic fix for all the svgs with the similar problem?

I'm not sure if that would be a good idea (I think we have the same SVG for different btn sizes. Thus a generic solution might not be ideal)

timar:
FWIW I did not find any other example than this (i.e. SVG image on a dialog button). I found many examples when the image was an inline data stream and in PNG format.

PR #10810, opened 2024-12-27
wsd: wopiAccessCheck, remove unused CheckStatus, hook NotHttpSuccess

meven:
Change-Id: Ic022614625bfa3aa0a5b819fda1adcf50c7465d1

* Resolves: # 
* Target version: master 

### Summary

Amends #9202

### TODO

- [ ] ...

### Checklist


hfiguiere:
what happens if the TCP connection fails? it can't always be DNS.

meven:
> what happens if the TCP connection fails? it can't always be DNS.

It is caught in `CheckStatus::ConnectionAborted`.

hfiguiere (approved):

PR #10817, opened 2024-12-28
wsd: suppress Ctrl+q

Ashod:
The Ctrl+q shortcut exits Core when
forwarded. There is no use to having
this shortcut processed in Core, so
we suppress it.

If we should process it by saving and
closing the document, it needs to be
handled in a special way (ideally by
sending 'closedocument' from the
client instead).

Change-Id: I6b3280342490a013645ae73888188f20cd25168d


caolanm (approved):
In core this is .uno:Quit as Q_MOD1 in officecfg/registry/data/org/openoffice/Office/Accelerators.xcu. All Q_MOD1 are uno:Quit, so nothing seems to have rebound ctrl+q to something different in a specific translation so that potential problem doesn't arise.

PR #10821, opened 2024-12-30
calc: improve vertical scroll calculation with split panes

eszkadev:
With 0f461ed3129250f6ef5fe003e6d543b3612f8da7
calc: scroll to the cell so it is visible

Vertical scrolling in case of split panes was designed to put
cell at the center of free pane. That doesn't take into account
cells higher than free pane. Let's do it in 2 steps, first
scroll to the top of a cell, then make sure bottom is visible.
That is similar behavior like in the desktop app.
Also don't use as a margin the height of a current cell but rather
as const value.

make cypress test spreadsheet more complex


eszkadev:
Improved search field to make calc/searchbar_spec.js stable as it is most failing test.

lpranam (approved):

PR #10823, opened 2024-12-30
For undefined init focus id we should focus on first focus able element in container

Darshan-upadhyay1110:
- in current case we did not have any fallback for cases like undefined init focus
- it is simple that if no focus-able id defined then we will add focus on first focus-able element

Change-Id: Ib46a1c7a3366933e8b01ca831798d92bf96797e0

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -413,7 +413,8 @@ L.Control.JSDialog = L.Control.extend({
 
 		} else {
 			// will directly set element of focusable element based on init focus id
-			firstFocusableElement = instance.init_focus_id ? instance.container.querySelector('[id=\'' + instance.init_focus_id + '\']') : null;
+			// If init_id is not defined, select the first focusable element from the container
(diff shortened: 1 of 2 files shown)


Darshan-upadhyay1110:
I am adding some cypress test for focus element case like on "Sheet rename", on open "Background color", on language drop down open etc etc....

Please do not merge this now

bayramcicek (commented):
Thanks for the quick fix. 
I tested and it works well. No errors displayed on the console.


https://github.com/user-attachments/assets/40dc24c3-2097-4f43-8d69-48137b430b56

Darshan-upadhyay1110:
https://github.com/CollaboraOnline/online/pull/10825#issuecomment-2566324846

@eszkadev i think your comment should be here you pasted in other PR

eszkadev:
no, please see: https://github.com/CollaboraOnline/online/pull/10825/commits/794f30aa324b2ff3834fe0b93bbda9a49f224438

comment says about this change but code is different :)

eszkadev (approved):

PR #10825, opened 2024-12-31
Accessibility: Navigator use an empty alt tag for the icons

Darshan-upadhyay1110:
Missing alt Attribute for Icons
The Icons do not have alt attributes.

Solution : In this case, it is advisable to use an empty alt tag for the icons, as the information of the funktion is available in text form

Change-Id: Ibdb1c7833477b9a020c1e77e5da6fe08a0f90336

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.JSDialog.js
+++ b/browser/src/control/Control.JSDialog.js
@@ -406,7 +406,8 @@ L.Control.JSDialog = L.Control.extend({
 
 		} else {
 			// will directly set element of focusable element based on init focus id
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
looks good

eszkadev:
commit message was wrong now I noticed :) but seems a good fix for initial focus - it's good to have first widget focused if not specified any desired custom widget to be first

PR #10827, opened 2024-12-31
jsdialog: no padding in treeview context menu

eszkadev:
do not add additional margin or padding to context menu also don't round the borders in it's entries
example menu to test can be found in style sidebar Writer -> Format -> Style sidebar (experimental for now)

result:
contextmenu


banobepascal:
Works fine :)

banobepascal (approved):
Works fine :)

PR #10838, opened 2025-01-02
Fix memory handling in Session::asyncConnect()

vmiklos:
`result` is allocated on the stack, don't capture its address in a
lambda.

Change-Id: I2d6dad1f76002559e400c57cd5b4ffa0b6e0f0d6

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1731,7 +1731,7 @@ private:
         _socket.reset(); // Reset to make sure we are disconnected.
 
         auto pushConnectCompleteToPoll = [this, &poll](std::shared_ptr<StreamSocket> socket, net::AsyncConnectResult result ) {
-            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket), &result]() {
+            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket), result]() {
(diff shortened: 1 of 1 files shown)


meven:
+1 `result` should be passed by value.

mmeeks (approved):
Wow - is there some more paranoid compiler option, or clang plugin we can use to catch this class of problems at compile time ? =) assuming that all lambdas are long-lived is probably a good idea in coolwsd.

vmiklos:
The closest I could find with brief research is https://github.com/llvm/llvm-project/issues/84679, but that is about returning a lambda that would return the address of a local variable, not about long-living lambdas. I guess it would be possible to write a clang AST matcher / plugin that would flag cases where function parameters are passed to lambdas by reference, that is a subset of what you point out (local variables are not considered), but it would find this problem and also 19861a2d0bdc2db9825f9d4b953ce3e0a7a3539b was also a parameter, I expect.

mmeeks:
@grandinj I guess we don't have clang-plugins in 'online' yet, but - this one would be rather useful in the modern lambda world I think if you're interested =)

vmiklos:
> @grandinj I guess we don't have clang-plugins in 'online' yet, but - this one would be rather useful in the modern lambda world I think if you're interested =)

#10963 adds an initial "coplugin" that catches 2 past issues (similar type), let's see.

PR #10842, opened 2025-01-02
jsdialog: allow to open links inside treeview

eszkadev:
This is used in Audit dialog to open SDK help


gokaysatir (approved):
Makes sense, thank you.

Issue #10814, opened 2024-12-27
RTL sheet: no col/row headers when loaded


eszkadev:
### Describe the Bug
When we have RTL sheet which is opened on spreadsheet load - we have no col/row headers (some race?).
It generated error which I fixed in https://github.com/CollaboraOnline/online/pull/10812
But it doesn't solve the problem.

### Steps to Reproduce
1. Open attached file: 
[rtl-load.ods](https://github.com/user-attachments/files/18260548/rtl-load.ods)

Result: no headers on right side of sheet:
![noheaderbefore](https://github.com/user-attachments/assets/fac8bac7-dc83-4805-8a5c-8c66b8e62c30)

Expected:
![nobeadersafter](https://github.com/user-attachments/assets/1d9ca0e7-18fa-4bcd-ae3f-3b7019e96049)

I noticed that switching the sheets to some LTR then again to RTL fixes it.
So something is not available on time on load.

eszkadev:
cc: @gokaysatir related to sections

PR #10843, opened 2025-01-02
Delay setting layoutIsRtl to ensure correct section positioning

get-state:
Prevent premature setting of layoutIsRtl, which can cause some sections to remain in incorrect positions if they are not defined before calling _adjustCanvasSectionsForLayoutChange. Fix this by ensuring all sections are adjusted before setting layoutIsRtl.

Change-Id: I13f9e18c185e61265503d413a77a2a8151bb6857

* Resolves: #10814 
* Target version: master 

### Summary

Right now, an early call to _adjustCanvasSectionsForLayoutChange will set the sheet's layoutIsRTL to true before adjusting the anchor position of all of the sections. 

I'm not sure if this is the correct fix.  What I can think of is if "status" messages and w/e calls into _adjustCanvasSectionsForLayoutChange should be delayed further (set is_ready after we render). Any comments on this would be appreciated.

### Checklist

- [x] I have run tests with `make check` (This fails on unit-synthetic-lok, which seems unrelated)

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -687,8 +687,12 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 			this._setAnchor('tilesSection', tilesSection,
 				[[L.CSections.ColumnHeader.name, 'bottom', 'top'], [L.CSections.RowHeader.name, '-left', 'right']]);
-
-			this._layoutIsRTL = true;
+			
+			// Do not set layoutIsRtl to true prematurely. If set before all sections are defined 
+			// (e.g., during load with onStatusMsg), some sections may not update to their correct positions. 
+			// Ensure all sections are adjusted first, then set layoutIsRtl to true to show that they have moved.
+			if (rowHeaderSection)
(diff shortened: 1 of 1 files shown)


get-state:
From my understanding, the process works as follows:

* Initial Status Message: A status message is sent first, which triggers the onStatus function.
    * This calls _adjustCanvasSectionsForLayoutChange to set the layout to RTL (Right-To-Left), which happens before the header entry section is created, but with the toolbar.
- Row Headers Creation: After the initial status message, the row headers are created through a commandsvalue msg.
    - Subsequent Adjustment: Another call to _adjustCanvasSectionsForLayoutChange is made after the row headers are created. However, since the document is already set to RTL from the first call, this second adjustment has no effect.

The pr currently assumes the above order is 'valid' and makes it work under that assumption.

gokaysatir (approved):
This is a good solution. We could do better but async nature of server communication makes the timing hard. Checking if the sections are initialized in that function looks good to me.

Thank you for checking the issue :)

eszkadev:
this one fixes: https://github.com/CollaboraOnline/online/issues/10814

PR #10844, opened 2025-01-02
jsdialog: added more cypress visual tests

eszkadev:
- combobox
- edit field


cypress_test/integration_tests/snapshots/base/integration_tests/desktop/writer/jsdialog_widgets_spec.js/edit_enabled.png
eszkadev, line 1: interesting that we have margin/padding inside widget, not outside <- TO FIX LATER

vmiklos (approved):
Looks OK to me. I have some local failure from visual tests like this for slideshow (to be investigated), but

```
make -C cypress_test check-desktop spec=writer/jsdialog_widgets_spec.js
```

passes here fine even with these changes.

PR #10850, opened 2025-01-03
tooltip improvements

hcvcastro:
- **browser: tooltip: handle 'ESC' key to hide tooltip**
- **browser: tooltip: add option to handle different windows**
- **browser: tooltip: add 'container' option**
- **browser: create initial tooltip for presenter console**

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -561,6 +561,11 @@ class PresenterConsole {
 
 		mainContentContainer.append(this.tooltip);
 
+		this._tooltip = L.control.tooltip({
+			window: this._proxyPresenter,
+			container: this.tooltip,
+		});
+
 		let allPresenterConsoleButtons =
 			this._proxyPresenter.document.querySelectorAll('button');
 		this._attachTooltips(allPresenterConsoleButtons);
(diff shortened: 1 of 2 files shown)


Darshan-upadhyay1110 (approved):
Tested and it works but i am confused why we need to use 
`capture and passive ` pros here in event listeners ?

Ned +2 on this PR 

CC: @eszkadev

browser/src/slideshow/PresenterConsole.js
```diff
@@ -561,6 +561,11 @@ class PresenterConsole {
 
 		mainContentContainer.append(this.tooltip);
 
+		this._tooltip = L.control.tooltip({
```
eszkadev, line 564: why do we create new tooltip here while we set the style above?
Shouldn't that be mooved above styling?

hcvcastro: I have setup an initial code to simplify tooltip, @Darshan-upadhyay1110 should modify according to his needs =)

hcvcastro:
> Tested and it works but i am confused why we need to use
> capture and passive props here in event listeners ?

You can check the API documentation, capture before dispatching the target element, passive, prevent default is ignored

browser/src/slideshow/PresenterConsole.js
eszkadev: ah so it's todo, ok

PR #10858, opened 2025-01-06
SSL Improvements (1)

Ashod:
Non-functional SSL error handling and logging improvements.
Functional changes will follow separately.

- **wsd: net: breakout handleSslError() from handleSslState()**
- **wsd: net: handle SSL_ERROR_NONE**
- **wsd: net: hoist getBioError() and call it once**
- **wsd: net: assert and error on unsupported async encryption**
- **wsd: net: document the error conditions**
- **wsd: net: unify SSL error logging**
- **wsd: net: remove duplicate logging of rc**

### DIFF
--- a/net/SslSocket.hpp
+++ b/net/SslSocket.hpp
@@ -358,7 +358,7 @@ private:
                   << Util::symbolicErrno(last_errno) << ": " << std::strerror(last_errno) << ")"
 
         // Handle non-fatal cases first.
-        const std::string bioErrStr = getBioError(rc);
+        const std::string bioErrStr = getBioError();
         switch (sslError)
         {
             // Not an error; should be handled elsewhere.
@@ -530,13 +530,13 @@ private:
         return rc;
     }
 
-    std::string getBioError(const int rc) const
+    std::string getBioError() const
     {
         // The error is coming from BIO. Find out what happened.
         const long bioError = ERR_peek_error();
 
         std::ostringstream oss;
-        oss << "BIO error: " << bioError << ", rc: " << rc;
+        oss << "BIO error: " << bioError;
 
         char buf[512];
         ERR_error_string_n(bioError, buf, sizeof(buf));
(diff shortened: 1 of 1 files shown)


net/SslSocket.hpp
```diff
@@ -414,11 +414,19 @@ class SslStreamSocket final : public StreamSocket
                                           << bioErrStr);
 #if OPENSSL_VERSION_NUMBER > 0x10100000L
                 else if (sslError == SSL_ERROR_WANT_ASYNC)
-                    LOG_TRC("SSL error (" << context << "): WANT_ASYNC (" << sslError << ") "
+                {
+                    LOG_WRN("SSL error (" << context << "): WANT_ASYNC (" << sslError << ") "
                                           << bioErrStr);
+                    LOG_ASSERT_MSG(sslError != SSL_ERROR_WANT_ASYNC,
```
mmeeks, line 466: Not a very obvious way of firing an assertion in this path ;-) != the value we know it is.

net/SslSocket.hpp
```diff
@@ -357,26 +362,37 @@ class SslStreamSocket final : public StreamSocket
...
                                       << "): " << bioErrStr);
                 errno = last_errno; // Restore errno.
                 return rc;
 
-            case SSL_ERROR_SYSCALL:
+            // Unexpected: happens only with SSL_CTX_set_client_cert_cb().
+            case SSL_ERROR_WANT_CLIENT_HELLO_CB: // 11
+                LOG_TRC("SSL error (" << context << "): WANT_X509_LOOKUP (" << sslError
+                                      << "): " << bioErrStr);
+                errno = last_errno; // Restore errno.
```
mmeeks, line 423: Was this the functional change / fix ? =)

mmeeks (approved):
Lots of change and cleanup, but looks sensible enough I guess. Lets get it into green if we can.

Thanks Ash =)

net/SslSocket.hpp
Ashod: It's here for completeness.

net/SslSocket.hpp
Ashod: Changed in the follow-up.

Issue #10859, opened 2025-01-06
Font size of the Push button is not uniform with other buttons


banobepascal:
### Steps to Reproduce
1. Go to 'Writer|Calc|Draw|Impress'
2. Click on 'Layout'
3. Click on 'Page Style|Format Page|Page Properties|Slide Propeties'
4. Notice how the font size of the Push button is not uniform with other buttons

### Expected Behavior
All button font-size should be the same

### Actual Behavior
The font size of the Push button is not uniform with other buttons

### Screenshots
![Screenshot from 2025-01-06 19-34-55](https://github.com/user-attachments/assets/55314e3e-dc0b-467b-99d0-5aab1395e7b2)

PR #10865, opened 2025-01-07
Add font size to the jsdialog Push button to ensure it is consistent with other buttons 

banobepascal:
Change-Id: I2b03d7e241618a75b97e8712c01c94e5c264b2ba

* Resolves: #10859 
* Target version: master 

### Summary
Fixes the font size of the "Push" button to ensure it is consistent with other buttons in the dialog

### Screenshots
Screenshot from 2025-01-07 14-38-37

### Checklist

### DIFF
--- a/browser/css/btns.css
+++ b/browser/css/btns.css
@@ -132,6 +132,7 @@ button.jsdialog img {
 
 .ui-button-box .jsdialog.ui-pushbutton:not(.hidden) {
 	display: initial !important;
+	font-size: var(--default-font-size);
 }
 
 .ui-pushbutton[disabled] {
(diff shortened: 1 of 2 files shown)


banobepascal:
@pedropintosilva While working on this I noticed that all button-primary don't have a cursor: pointer while being hovered. 

Is that by design with this commit
https://github.com/CollaboraOnline/online/commit/2e863c37ee7bf97e790b43a3c16e4a964f843c49

browser/css/btns.css
```diff
@@ -132,6 +132,7 @@ button.jsdialog img {
 
 .ui-button-box .jsdialog.ui-pushbutton:not(.hidden) {
 	display: initial !important;
+	font-size: 12px;
```
eszkadev, line 135: don't we have variable for that?

banobepascal: @eszkadev resolved this to use --default-font-size from cool.css

banobepascal:
> Yes, please : )

I have pushed the change

PR #10870, opened 2025-01-08
jsdialog: treeview improvements

eszkadev:
- move sorting and cell creation to separate function
- enable/disable sorting functionality depending on "sortable" property from JSON
- requires: https://gerrit.libreoffice.org/c/core/+/179949 to be merged first


gokaysatir (approved):

PR #10871, opened 2025-01-08
Improve URL popup positioning.

gokaysatir:
In Writer, it was not performing well. Document position is also taken into account.

Issue:
* Open a Writer file.
* Add some text and a hyperlink.
* Scroll down the document and let the hyperlink be the first visible row.
* If you click on the text part with the link, hyperlink popup will be shown in the invisible area.

Change-Id: Idfd01a0486bf0ce0f072472c3ca6dfe26d1f6014

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/URLPopUpSection.ts
+++ b/browser/src/canvas/sections/URLPopUpSection.ts
@@ -146,26 +146,27 @@ class URLPopUpSection extends HTMLObjectSection {
 	}
 
 	public static resetPosition(section: URLPopUpSection) {
-		if (!section)
-			section = app.sectionContainer.getSectionWithName(URLPopUpSection.sectionName);
+		if (!section) section = app.sectionContainer.getSectionWithName(URLPopUpSection.sectionName);
+		if (!section) return;
 
 		let left = section.sectionProperties.documentPosition.pX - section.getPopUpWidth() * 0.5 * app.dpiScale;
 		let top = section.sectionProperties.documentPosition.pY - (section.getPopUpHeight() + URLPopUpSection.popupVerticalMargin) * app.dpiScale;
 
-		if (section) {
-			let arrowAtTop = false;
-			if (top < 0) {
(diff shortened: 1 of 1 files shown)


browser/src/canvas/sections/URLPopUpSection.ts
```diff
@@ -146,26 +146,27 @@ class URLPopUpSection extends HTMLObjectSection {
...
 		let left = section.sectionProperties.documentPosition.pX - section.getPopUpWidth() * 0.5 * app.dpiScale;
 		let top = section.sectionProperties.documentPosition.pY - (section.getPopUpHeight() + URLPopUpSection.popupVerticalMargin) * app.dpiScale;
 
-		if (section) {
-			let arrowAtTop = false;
-			if (top < 0) {
-				top = section.sectionProperties.documentPosition.pY + (URLPopUpSection.popupVerticalMargin * 2 * app.dpiScale);
-				arrowAtTop = true;
-			}
+		const checkLeft = (section.sectionProperties.documentPosition.pX - section.containerObject.getDocumentTopLeft()[0]) - section.getPopUpWidth() * 0.5 * app.dpiScale;
```
eszkadev, line 155: isn't that just `checkLeft = left - section.containerObject.getDocumentTopLeft()[0])` ?

can we simplify?

gokaysatir: Yes but that is the check. Indeed that was above the "left "variable but i moved it below to make the review process easier (fewer changes). Let me send an update there. We can move cheks to top then we can use some maths there. It'll be better in terms of design i think.

gokaysatir: Okay done. Thank you for checking.

eszkadev (approved):
thanks, now it't easier to read

PR #10873, opened 2025-01-08
a11y: treeview get current entry focused

mcecchetti:
This hack removes the need to press the `tab` key in order to get a tree entry
focused after clicking on it.

That improves usability for the Navigation pane and also for gridview such as
the one for the 'font' tab in the Format cell dialog.

Change-Id: I1be820367f935c8460ab5a60ef74cc648230ac64


Darshan-upadhyay1110:
Hii @mcecchetti how do i test this as i have tried following thing still i need to use `tab`.

1. click on navigation sidebar
2. it opens the sidebar
3. i clicked on first option
4. then i tried to navigate using Arrow keys but it's not working
5. then i see that i have to press `tab` to use arrow keys

eszkadev (approved):
looks ok

PR #10874, opened 2025-01-08
Safer Socket Closed State

Ashod:
This combines the _closed member of Socket
with the _fd, such that we invalidate the
_fd and don't need a separate flag. This
is safer and reduces duplication.

This fixes a possible bug where we
inadvertently close _another_ socket
when we call shutdown(2) on the already
closed FD after the FD gets recycled.

For example, StreamSocket::checkRemoval()
has a case where we unconditionally call
Socket::shutdown(), via closeConnection(),
which calls shutdown(2). There could
easily be other cases where we close a
socket that can later be reused by
accept(2) before we call shutdown(2) on
this recylced FD again, closing the
recently accepted connection.

If this has been happening in the wild,
it probably is rather rare, otherwise
we should have seen more issues related
to closed sockets. Still, there have
been numerous cases over the years of
unexplained reports with closed sockets
(discounting those that were proven to be
due to external components).

It is the prudent thing to invalidate any
and all FD's post closing.

- **wsd: replace _closed member with invalidating the FD**
- **wsd: invalidate socket FD by negating**
- **wsd: simplify shutting down in checkRemoval**

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -217,7 +217,7 @@ public:
     /// TODO: Support separate read/write shutdown.
     virtual void shutdown()
     {
-        if (!_noShutdown)
+        if (!_noShutdown && !isClosed())
         {
             LOG_TRC("Socket shutdown RDWR. " << *this);
             if constexpr (!Util::isMobileApp())
@@ -436,7 +436,14 @@ protected:
     void setNoShutdown() { _noShutdown = true; }
 
     /// Explicitly marks this socket closed, i.e. rejected from polling and potentially shutdown
-    void setClosed() { _fd = -1; }
+    /// Note: to preserve the original FD post closing (f.e. in logs and debugger), we negate it.
+    void setClosed()
+    {
+        if (_fd > 0)
+            _fd = -_fd;
+        else if (_fd == 0) // Unlikely, but technically possible.
+            _fd = -1;
+    }
 
     /// Explicitly marks this socket and the given SocketDisposition closed
     void setClosed(SocketDisposition &disposition) { setClosed(); disposition.setClosed(); }
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -147,7 +147,6 @@ public:
         : _type(type)
         , _clientPort(0)
         , _fd(createSocket(type))
-        , _closed(_fd < 0)
         , _creationTime(creationTime)
         , _lastSeenTime(_creationTime)
         , _bytesSent(0)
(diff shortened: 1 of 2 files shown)


net/Socket.hpp
```diff
@@ -218,14 +217,14 @@ class Socket
     /// TODO: Support separate read/write shutdown.
     virtual void shutdown()
     {
-        if (!_noShutdown)
+        if (!_noShutdown && !isClosed())
         {
             LOG_TRC("Socket shutdown RDWR. " << *this);
-            setClosed();
             if constexpr (!Util::isMobileApp())
                 ::shutdown(_fd, SHUT_RDWR);
```
Ashod, line 224: Unconditionally calling `shutdown(2)` is highly problematic.

caolanm (approved):
ok, this looks reasonable to me

vmiklos:
@Ashod recently my local 'make check' started to fail, the relevant test is this:

```
$ (cd test; CPPUNIT_TEST_NAME=HttpRequestTests::test500GetStatuses ./unithttplib)
...
##Failure Location unknown## : Error
Test name: HttpRequestTests::test500GetStatuses
uncaught exception of type std::exception (or derived).
- I/O error

Failures !!!
Run: 1   Failure total: 1   Failures: 0   Errors: 1
```

I had this passing a week ago, a git bisect on dd9b0df23c22187a137f4c1f125171063eb93cce..bc58df26801cacf28be049e9263596acfad3227e flags af747c19fc96ae2024a5660034a23b7936620b25 as the first bad commit, i.e. this PR.

Interestingly 'ulimit -n' reports 1024 here, if I bump that to 2048, then all is well, but we didn't need that in the past, it sounds like we start to leak fds after this commit for some reason. One thing I tried is to assert that we don't try to close an already closed socket, that did find one bad caller, but at the end the test still fails. Here is what I tried:

```
diff --git a/net/Socket.hpp b/net/Socket.hpp
index 1a6e1e99e4..17fb0f2007 100644
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -439,6 +439,7 @@ protected:
     /// Note: to preserve the original FD post closing (f.e. in logs and debugger), we negate it.
     void setClosed()
     {
+        assert(_fd >= 0);
         if (_fd > 0)
             _fd = -_fd;
         else if (_fd == 0) // Unlikely, but technically possible.
@@ -1583,7 +1584,14 @@ public:
         {
             LOG_TRC("Closed. Firing onDisconnect.");
             _socketHandler->onDisconnect();
-            setClosed(disposition);
+            if (!isClosed())
+            {
+                setClosed(disposition);
+            }
+            else
+            {
+                disposition.setClosed();
+            }
         }
         else if (isClosed())
             disposition.setClosed();
```

I wonder if at this stage it would make sense for you to take this over, if you can reproduce this regression in an 'ulimit -n 1024' environment. What do you think? Thanks.

Ashod:
> I wonder if at this stage it would make sense for you to take this over, if you can reproduce this regression in an 'ulimit -n 1024' environment. What do you think? Thanks.

@vmiklos, thanks, much appreciated the report. Will follow up.

PR #10876, opened 2025-01-08
calc: follow before search result is selected

eszkadev:
we need to turn on the following with direct jump so we will always jump to result

remove old hack which remembers if we just searched.. we know that we received search result so it's not needed also we switch to correct page when user switched in the meantime

example testing scenario:
- select cell A1 and remember some text in cell A2
- scroll to A1000
- search for text inside A2 using search in statusbar expected: jump to A2
result: selection, but no jump was done


gokaysatir:
I checked the code and all makes sense to me.

But i couldn't reproduce the issue that is preventing the jump to search results. In my local, it jumps to result.
* Select A1.
* Put "2" into B1.
* Scroll down to ~1000 rows or below.
* CTRL+F and search for "2".
* And it jumps.

But to note, i don't have a touchpad so i scroll with either mouse wheel or via the scroll bars.

There is also the issue that the reverted commit was aiming to fix. Did we also test that it doesn't regress?

Can a QA member check the previous and current issues and confirm both are fixed? Then we can relax and get this in :)

gokaysatir (approved):
So this seems safe to approve if we get this in, in RED code. We still need a person who can reproduce both issues and confirms the fixes.

Thank you.

eszkadev:
Thank you for checking. That issue doesn't happen first time when I try, only later. I will try to add followup tomorrow for the reverted part.

PR #10881, opened 2025-01-09
cypress: check search with following off

eszkadev:
Followup for https://github.com/CollaboraOnline/online/pull/10876:
- cell not fully visible if we scroll to it after search (thanks to [@jaumep](https://phabricator.collabora.com/p/jaumep/) )
- fix for turning off following ( regression )
- jump to found result always ( if have text cursor or cell cursor visible)
- added cypress tests

### DIFF
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -2507,6 +2507,10 @@ L.CanvasTileLayer = L.Layer.extend({
 			{
 				var _fillColor = '#CCCCCC';
 				var strTwips = result.twipsRectangles.match(/\d+/g);
+
+				if (!strTwips || !strTwips.length)
+					continue;
+
 				var rectangles = [];
 				for (var i = 0; i < strTwips.length; i += 4) {
 					var topLeftTwips = new L.Point(parseInt(strTwips[i]), parseInt(strTwips[i + 1]));
--- a/browser/src/layer/tile/CanvasTileLayer.js
+++ b/browser/src/layer/tile/CanvasTileLayer.js
@@ -1009,7 +1009,7 @@ L.CanvasTileLayer = L.Layer.extend({
 
 	_isLatLngInView: function (position) {
 		var centerOffset = this._map._getCenterOffset(position);
-		var viewHalf = this._map._getCenterLayerPoint();
+		var viewHalf = this._map.getSize()._divideBy(2);
 		var positionInView =
 			centerOffset.x > -viewHalf.x && centerOffset.x < viewHalf.x &&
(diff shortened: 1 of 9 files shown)


cypress_test/integration_tests/desktop/calc/sheet_switch_spec.js
```diff
@@ -27,15 +27,15 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Sheet switching tests', fu
 		desktopHelper.assertScrollbarPosition('vertical', 15, 25);
 	});
 
-	it('Check view position on repeated selection of currently selected sheet', function() {
+	it.only('Check view position on repeated selection of currently selected sheet', function() {
```
JaumePujantell, line 30: Is this 'only' intentional?

eszkadev: done

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2474,7 +2474,15 @@ L.CanvasTileLayer = L.Layer.extend({
...
 		this._map.fire('search', {originalPhrase: originalPhrase, count: count, highlightAll: highlightAll, results: results});
-		app.setFollowingUser(this._viewId, /* instant jump */ true);
+
+		app.setFollowingUser(this._viewId);
+
+		// always jump to search result - we already received cell / text cursor before so we need
+		// to force it in case we had following OFF
+		if (app.file.textCursor.visible)
+			this._onUpdateCursor(/* scroll */ true);
+		if (app.calc.cellCursorVisible)
```
gokaysatir, line 2484: I guess we could use an else-if here.
```suggestion
		if (app.calc.cellCursorVisible)
```

gokaysatir (commented):
There is a slight difference in behaviour:
* Select CELL A2
* Scroll down without selecting any other cell.
* Press arrow up key to go up.
* It doesn't hit the top, there is a slight gap on top.

browser/src/layer/tile/CanvasTileLayer.js
eszkadev: done

eszkadev:
> There is a slight difference in behaviour:
> 
>     * Select CELL A2
> 
>     * Scroll down without selecting any other cell.
> 
>     * Press arrow up key to go up.
> 
>     * It doesn't hit the top, there is a slight gap on top.

I see: cell
But it was like that even before and even when just selecting the first row - not related to this PR I think

eszkadev:
checked multiple times on CI and:
- calc/searchbar_spec.js
- calc/sheet_switch_spec.js
are stable
- writer/searchbar_spec.js now is fixed

gokaysatir (approved):
Thanks :)

PR #10883, opened 2025-01-09
CommentList: importComments function is already clearing the list. No…

gokaysatir:
… need to duplicate the call.

Change-Id: Ie0b92e10c129f8d8a9ac2fa7dce6de53a449bb3e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks reasonable, also in import it might happen we will return early due to dialog - so list will be left empty

PR #10890, opened 2025-01-10
a11y: treeview: make widget focusable when no entry is selected

mcecchetti:
Change-Id: Id88daf456a02d563a96a4989ad46712364823ce3


browser/src/control/jsdialog/Widget.TreeView.ts
```diff
@@ -1161,6 +1166,10 @@ class TreeViewControl {
 		}
 
 		if (entries && entries.length === 0) this.makeEmptyList(data, builder);
+
+		// we need to provide a way for making the treeview control focusable
+		// when no entry is selected
+		if (level === 1 && !hasSelectedEntry) this._container.tabIndex = 0;
```
eszkadev, line 1172: can we create method for that:
```
makeTreeViewFocusable(focus: boolean) {
if (focus)
    this._container.tabIndex = 0;
else
    this._container.removeAttribute('tabindex');
}
```

and use here and in the `selectEntry()` - that will show the conection between the two instructions

mcecchetti: Done

eszkadev (approved):
thanks, looks good :)

eszkadev:
can we also get the cypress test for that case?

eszkadev:
visual regreession was detected, you can check locally in cypress_tests/snapshots/

PR #10892, opened 2025-01-10
Use twips and core pixels instead of latlng.

gokaysatir:
Change-Id: I4d399d5d00ad40f38015d10f1f8faefd56dbcf30

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mcecchetti (approved):
Thanks!

PR #10899, opened 2025-01-10
scripts/unocommands.py: expand path and check dirs params exist 

meven:
### Summary

I have tried multiple times to run `./scripts/unocommands.py --update . ../libreoffice/core/`.
Except it failed silently, causing headaches and frustration.

This will make this command work with relative path or fail explicitly when parameters are not directories.


hfiguiere (approved):
LGTM

timar (approved):

PR #10900, opened 2025-01-10
use innerText to retain newlines in comments

caolanm:
instead of textContent so newlines entered into calc comments round trip


caolanm:
Does this make sense.

Rash419 (approved):
> Does this make sense.
Yes, thank you. textContent just gets raw text but in contentEditable div the new lines are entered using html tag <br>/<p> so if we use textContent it removes html tag and we end up with no new lines.

Issue #10887, opened 2025-01-10
Misaligned and Incorrectly Styled Replies Indicator in Comments Section


banobepascal:
### Steps to Reproduce
1. Go to 'Writer'
2. Click on 'Insert'
3. Click on '''Comment and add replies to the comment"

### Expected Behavior
The number of replies indicator should align side by side with the page, the circle should be positioned next to the document page rather than above it, the circle should be fully visible and styled appropriately, the number of replies indicator should be sized to fit seamlessly within the layout, and the entire element should include a hover state to provide user feedback.

### Actual Behavior
The number of replies indicator is misaligned and not side by side with the page, the circle is positioned above the document page instead of next to it, the circle is transparent and difficult to notice, the number of replies indicator is excessively large and disrupts the layout, and the element does not include a hover state, reducing interactivity.

### Screenshots
![image](https://github.com/user-attachments/assets/1a61f58c-51a3-43d7-8fb3-59dd3e36559f)

PR #10901, opened 2025-01-10
Align comments indicator and add hover state to comments avatar

banobepascal:
Change-Id: I4746de9d38ce016e0fdef2d0ef15d810474276b6

* Resolves: #10887 
* Target version: master 

### Summary
Fixes the alignment, size, and interactivity issues with the number of replies indicator in the comments section, resized to fit seamlessly within the circle, and includes a hover state

### Screenshot
[Screencast from 10-01-25 15:38:57.webm](https://github.com/user-attachments/assets/bed079eb-32fe-44e8-8846-5308a80c413a)

### DIFF
--- a/browser/css/cool.css
+++ b/browser/css/cool.css
@@ -589,20 +589,22 @@ body {
 }
 
 .cool-annotation-info-collapsed {
-	text-align: center;
+	display: flex;
+	justify-content: center;
+	align-items: center;
 	margin: 0;
 	padding: 0;
 	background-color: var(--color-primary);
 	color: var(--color-primary-text);
 	font-weight: bold;
 	font-family: var(--cool-font);
-	font-size: 20px;
+	font-size: var(--default-font-size);
 	border: 1px solid var(--color-primary);
 	border-radius: 50%;
 	z-index: 10;
 	position: relative;
-	width: var(--btn-size);
-	height: var(--btn-size);
+	width: var(--btn-size-s);
+	height: var(--btn-size-s);
 	left: -6px;
 	top: -15px;
 }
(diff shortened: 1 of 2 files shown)


banobepascal:
@pedropintosilva will be waiting for your feedback

pedropintosilva (changes requested):
Thanks @banobepascal , the sizes look much better. However - you can see this by yourself when testing with an integrator or simply checking staging perf - the border color is overwritten later in https://github.com/CollaboraOnline/online/blob/ac743d3e3d51321663d04cf8eb8547d40ca764ed/browser/src/canvas/sections/CommentListSection.ts#L575 so, each comment avatar/bubble gets the same border color as its author. Thus, changing the border color on hover is not the best for UX, maybe better would be to:
* Hover: Add a subtle box-shadow + opacity to 1
* Normal state: Set element with opacity 0.9

This way we would have avatars/bubbles more discreet as default with a hover state that would be visible.

Note: It would be good if these changes are only apply for when there is no comment cards ( only bubbles floating due to space needs)

banobepascal:
> so, each comment avatar/bubble gets the same border color as its author. Thus, changing the border color on hover is not the best for UX, maybe better would be to:
> * Hover: Add a subtle box-shadow + opacity to 1
> * Normal state: Set element with opacity 0.9
> 
> This way we would have avatars/bubbles more discreet as default with a hover state that would be visible.
> 
> Note: It would be good if these changes are only apply for when there is no comment cards ( only bubbles floating due to space needs)

New improvements
[Screencast from 17-01-25 00:20:07.webm](https://github.com/user-attachments/assets/3b93c261-9a5f-4c90-8840-0fc7cf52f223)

banobepascal:

pedropintosilva (approved):
Thanks, looks good, I would like to have and extra +1 on the commentListSection from Gokay

gokaysatir (approved):
Thanks, looks beautiful. Can we merge this after below one?
* https://github.com/CollaboraOnline/online/pull/10921

There may be merge conflicts with my PR and i would like to have conflicts here instead of in 300 lines of change.

I hope that is not selfish :(

pedropintosilva:
Sorry @gokaysatir I have merged this, in my GitHub web interface your comment didn't  show up until I have pressed the merge button. But as far as I see now your PR doesn't seem to have any conflicts

gokaysatir:
Yep, np thanks :)

PR #10908, opened 2025-01-10
kill saving thread with SIGABRT

caolanm:
so the the correct/main thread gets the signal so we can get a backtrace generated.

Change-Id: I64f8f374bad92964f42ba34e23011bf827d87bf9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -175,11 +175,11 @@ static LokHookFunction2* initFunction = nullptr;
 class BackgroundSaveWatchdog
 {
 public:
-    BackgroundSaveWatchdog(unsigned mobileAppDocId)
+    BackgroundSaveWatchdog(unsigned mobileAppDocId, int savingTid)
         : _saveCompleted(false)
         , _watchdogThread(
             // mobileAppDocId is on the stack, so capture it by value.
-              [mobileAppDocId, this]()
+              [mobileAppDocId, savingTid, this]()
               {
                   Util::setThreadName("kitbgsv_" + Util::encodeId(mobileAppDocId, 3) + "_wdg");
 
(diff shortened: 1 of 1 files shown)


caolanm:
We already have a killThreadById(int tid, int signal which is ::syscall(SYS_tgkill, getpid(), tid, signal);

With this patch and a long enough sleep in core to make this happens, then bt is the wanted thread:

```
kit-2007983-2008056 2025-01-13 08:50:43.332141 +0000 [ kitbgsv_002_001 ] SIG   Fatal signal received: SIGABRT code: 18446744073709551610 for address: 0x3e8001ea3f8
...
Backtrace 2008056 - kit startup of 24.04.12.0 87a2a7f3c2:
/home/caolan/LibreOffice/collabora-online/coolforkit-ns(_ZN7SigUtil13dumpBacktraceEv+0x71)[0x5377f1]
/home/caolan/LibreOffice/collabora-online/coolforkit-ns[0x537c25]
/lib64/libc.so.6(+0x40d00)[0x7f69e724fd00]
/lib64/libc.so.6(clock_nanosleep+0x63)[0x7f69e72f2c13]
/lib64/libc.so.6(nanosleep+0x17)[0x7f69e72fec47]
/lib64/libc.so.6(sleep+0x41)[0x7f69e73112b1]
/home/caolan/LibreOffice/co-24.04/instdir/program/libmergedlo.so(+0x31d4c64)[0x7f69de5d4c64]
```

mmeeks (approved):
Beautiful; thanks Caolan @timar can we get this back-ported into a green update and get it running on various servers ? =) would like to get the the root cause of bgsave problems =)

PR #10909, opened 2025-01-11
js: fix: comment doesn't presists newlines

Rash419:
- happens on firefox when you make the comment autosave and then finally save the comment

Change-Id: Ida59e546577a2c55197568225a1277340ef94367


caolanm (approved):

eszkadev:
@Rash419 can we have cypress test for that issue?

PR #10912, opened 2025-01-12
Private/caolan/coverity2025 01 12

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


wsd/FileServer.cpp
```diff
@@ -914,6 +914,8 @@ void FileServerRequestHandler::readDirToHash(const std::string &basePath, const
                 // Add the uncompressed version; it's better to serve uncompressed than nothing at all.
                 FileHash.emplace(prefix + relPath,
                                  std::make_pair(std::move(uncompressedFile), std::string()));
+
+                deflateEnd(&strm);
```
vmiklos, line 918: Nit: we may want to have an RTTI class for this, so you don't have to end manually.

vmiklos (approved):

PR #10918, opened 2025-01-13
widget: on demand rendering also sends separators

lpranam:
problem:
not including separators in messes up indexing and misplaces the inconview widget

Change-Id: I663b01271cd375177a6a573bf77cb9b90f772cc5

* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
Makes sense to me

PR #10919, opened 2025-01-13
CheckFileInfo: disconnect callbacks in dtor

meven:
* Resolves: # 
* Target version: master 

### Summary

Ensures the CheckFileInfo lambda using `this` can't be called after being destructed.

### Checklist


caolanm:
Did you find any way to reproduce this, e.g. injecting an error or timeout in something in RequestVettingStation.cpp or similar?

meven:
> Did you find any way to reproduce this, e.g. injecting an error or timeout in something in RequestVettingStation.cpp or similar?

I haven't.
Will give it a try.

meven:
Replaced by https://github.com/CollaboraOnline/online/pull/10919#issuecomment-2587127566

PR #10920, opened 2025-01-13
Clipboard Response Is Not Cacheable

Ashod:
- **wsd: string -> string_view in HttpHelper**
- **wsd: reuse HttpHelper::sendErrorAndShutdown**
- **wsd: test: simplify UnitCopyPaste**
- **wsd: send clipboard response with no-cache header**

### DIFF
--- a/net/HttpHelper.cpp
+++ b/net/HttpHelper.cpp
@@ -28,7 +28,7 @@
 namespace HttpHelper
 {
 void sendError(http::StatusCode errorCode, const std::shared_ptr<StreamSocket>& socket,
-               const std::string& body, const std::string& extraHeader)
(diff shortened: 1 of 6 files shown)


caolanm (approved):
Looks reasonable to me. Is there a known difference in practice with a non-cached clipboard response? Outdated content pasted?

Ashod:
> Looks reasonable to me. Is there a known difference in practice with a non-cached clipboard response? Outdated content pasted?

The difference is that without the no-cache header, proxies (and browsers, though not in practice) can pass outdated contents. It seems there is at least one such known case (though I'm not sure if the pasted contents were outdated or simply broken).

PR #10922, opened 2025-01-13
UnitKitSyntheticLok can pass and finish earlier than UnitSyntheticLok…

caolanm:
… expects

It seems that UnitKitSyntheticLok regularly finishes successfully, setting TestResult::Ok, before UnitSyntheticLok even gets to see that the document is loaded.

UnitKitSyntheticLok flags UnitSyntheticLok that the test passed, so presumably if we see that the test passed, then we are done don't need to do anything else except exit.

Change-Id: I1111a195943789c5e31da814c8713e2b7b0070b3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitPerf.cpp
+++ b/test/UnitPerf.cpp
@@ -29,7 +29,6 @@
 #include <string>
 #include <thread>
 
-/// Save torture testcase.
 class UnitPerf : public UnitWSD
 {
     void testPerf(std::string testType, std::string fileType, std::string tracesStr);
--- a/test/UnitSyntheticLok.cpp
+++ b/test/UnitSyntheticLok.cpp
@@ -33,7 +33,8 @@ namespace {
     }
 }
 
-/// Save torture testcase.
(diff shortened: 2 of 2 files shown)


caolanm:
This test is a regular failure in ci. As far as I can tell, the test does initially "pass",  UnitKitSyntheticLok passes and flags UnitSyntheticLok that info, so that one "passes" too but often UnitKitSyntheticLok completes and exits before UnitSyntheticLok's loadDocAndGetSession gets "progress: finished", which won't happen as UnitKitSyntheticLok closed the document.

This looks like the easiest thing to do, pass the "don't assert" arg to loadDocAndGetSession and if it exits because the overall test succeeded and the connection went away, then that's just fine.

vmiklos (commented):
Makes sense to me: there was a race here where we started listening too late and then this timeout out.

vmiklos (approved):

PR #10923, opened 2025-01-13
Private/caolan/weak ptr checkfileinfo

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/wopi/CheckFileInfo.cpp
+++ b/wsd/wopi/CheckFileInfo.cpp
@@ -39,8 +39,12 @@ void CheckFileInfo::checkFileInfo(int redirectLimit)
                                                            << httpRequest.header());
 
     http::Session::FinishedCallback finishedCallback =
-        [this, startTime, uriAnonym, redirectLimit](const std::shared_ptr<http::Session>& session)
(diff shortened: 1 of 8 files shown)


wsd/wopi/CheckFileInfo.cpp
```diff
@@ -145,8 +149,12 @@ void CheckFileInfo::checkFileInfo(int redirectLimit)
     _httpSession->setFinishedHandler(std::move(finishedCallback));
 
     http::Session::ConnectFailCallback connectFailCallback =
-        [this](const std::shared_ptr<http::Session>& /* httpSession */)
+        [selfWeak = weak_from_this(), this](const std::shared_ptr<http::Session>& /* httpSession */)
```
Ashod, line 152: Clever. Non-obvious, but will work.

Ashod (approved):
Thanks, @caolanm. This should do it. Though I prefer a cleaner design that avoids shared_/weak_ptr altogether (I didn't spend time to find out why this case breaks when we have other similar cases that work fine), for now we need something reliable yet minimal.

caolanm:
Lets go with this then, @meven FYI

Issue #10841, opened 2025-01-02
Audit dialog has unwanted grey dot icons and wrong text alignment


eszkadev:
### Describe the Bug
After treeview rework CSS in Audit Dialog produces additional icons.

### Steps to Reproduce
1. Go to Help -> Server Audit

### Expected Behavior
Only first column has grey dot icon

### Actual Behavior
All columns hav grey dot icon

### Screenshots
![auditicons](https://github.com/user-attachments/assets/06b21fc3-22dc-4608-ba05-2f082d4c2afe)

banobepascal:
@eszkadev I think these dots were mean't to be up/down icons since "Status" and "Help" are sorting options. 
Maybe when we are to also remove the sorting experience then the right approach would be to remove the dot icons.

Whats your take on the above?

cc @pedropintosilva

Ezinnem:
Seen on:
COOLWSD version: 24.04.10.1(git hash: 3799c094da (E))
LOKit version: Collabora Office 24.04.10.1(git hash: 565bfb5)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: ce791f39
WOPI host: nextcloud 30.0.0.14 - richdocuments 8.5.3

![image](https://github.com/user-attachments/assets/baf824f8-5f6f-4429-bb78-8a38e5025060)


OK in:
COOLWSD version: 24.04.8.1(git hash: 8475197 (E))
LOKit version: Collabora Office 24.04.8.1(git hash: 5c1b808)
Served by: Debian GNU/Linux 12 (bookworm)
Server ID: 07ed0e59
WOPI host: nextcloud 30.0.0.14 - richdocuments 8.5.3


![image](https://github.com/user-attachments/assets/88aed1c5-48ec-471e-a14b-1d52112a3b62)

pedropintosilva:
Thanks @Ezinnem , yes we should at least in a first step aim to reach that old OK status:

![image](https://github.com/user-attachments/assets/d36393ae-f16e-4071-9b9f-833a4197e8f1)

However there are a couple of problems that we should try to fix:

1. Ideally the status icon would be part of the Status column and thus no gray icon as heading would be needed.
  * But for that to happen the treeview widget needs to be improved/expanded so it allows more than one element per column (icon + text)
2. All text should be left aligned 
3. "No issues found" is also quite out of place maybe we could move it to other place (tile bar) or at the end with different format , what do you think @banobepascal 
[...]

PR #10925, opened 2025-01-14
Correct text and icon alignments in server audit dialog

banobepascal:
Change-Id: Id4b45e9097b8ba77df5280bedf1635693fce72fc

* Resolves: #10841 
* Target version: master 

### Summary
Improvements in alignment of icons and text in the server audit dialog

###  IN (COOLWSD version: 24.04.8.1(git hash: 8475197 (E))
image (2)

### AFTER: https://github.com/CollaboraOnline/online/pull/10870
Screenshot from 2025-01-14 03-29-37

### CURRENT FIX AND IMPROVEMENT
Screenshot from 2025-01-13 16-05-13

Screenshot from 2025-01-14 02-20-31

### Checklist

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -541,6 +541,12 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	white-space: nowrap;
 }
 
+#ServerAuditDialog .ui-treeview-entry {
+	display: grid;
+	grid-template-columns: var(--btn-img-size-m) 6fr 3fr;
+	white-space: normal;
+}
+
 .ui-treeview-expanded-content {
 	display: grid;
 }
@@ -566,6 +572,21 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	display: grid;
 	grid-auto-flow: column;
 	/* grid-column: '1 / ' + (this._columns + 1) - set inside JS */
+}
+
+#ServerAuditDialog .ui-treeview-headers {
+	display: grid;
+	grid-column-end: 4 !important;
+	grid-template-columns: 0fr 6fr 3fr;
+	margin-inline-start: 4px;
+}
+
+#ServerAuditDialog .ui-treeview-headers div.ui-treeview-icon-column {
+	width: 0px /* prevent display of ui-treeview-icon-column in ui-treeview-header */
+}
+
+#ServerAuditDialog .ui-treeview-headers > div:nth-last-child(1) {
+	margin-inline-start: 7px;
 }
 
 .ui-treeview-header {
@@ -2209,11 +2230,12 @@ kbd,
 }
 
 #ServerAuditDialog #ServerAuditDialog-mainbox {
-	min-width: calc(min(800px, 80vw));
+	min-width: calc(min(600px, 80vw));
 }
 
 #ServerAuditDialog .ui-treeview {
 	border-color: transparent;
+	max-height: max-content;
 }
 #ServerAuditDialog .ui-treeview td[role='gridcell'] {
 	padding-inline-start: 0;


banobepascal:
Screenshot from 2025-01-20 00-37-53

Screenshot from 2025-01-20 00-38-34


@pedropintosilva advise on the darkmode icon.

pedropintosilva:
> Screenshot from 2025-01-20 00-37-53

Nice, Maybe the indicator could look more line one element (distinct from the others rows and compound by icon and text) if the text would be set with the same dark green as the outer circle of the new icon. Another option would be to keep the text black but set a subtle green background that encompasses both the icon and the text. You need to see what it's better in terms of being distinct enough to be perceived as an indicator of the overall status and subtle enough to don't be confused by a button or other component that requires an active action from the user via the UI.

> 
> Screenshot from 2025-01-20 00-38-34

Please create another icon for dark mode purposes that should be uploaded to the `dark` subfolder under `images`. The current icon doesn't look very readable (specially with the white inner path vs green outer shapes vs dark bakcground)
> 
> @pedropintosilva advise on the darkmode icon.

Also, I didn't see changes WRT how it looks when there is an error, please attach a screenshot if you can

pedropintosilva (changes requested):
^

pedropintosilva:
I also see that there is only one commit, maybe it would be better to have one single commit (PR) that fixes the issue (linked in the description) so we can merge it as soon as it passes CI and then another PR with further UX improvements

banobepascal:
> I also see that there is only one commit, maybe it would be better to have one single commit (PR) that fixes the issue (linked in the description) so we can merge it as soon as it passes CI and then another PR with further UX improvements

Screenshot from 2025-01-21 09-41-43

Alignments of Icons, and Text in the server dialog

Will be creating another PR for UX improvements on the Status

pedropintosilva (changes requested):
image
 It seems that now the dialog can become too narrow resulting in this ^

Also we could make sure that the text wraps if needed

banobepascal:
[Screencast from 21-01-25 23:22:14.webm](https://github.com/user-attachments/assets/f8287a3c-4d54-40d9-a999-65d61e437dc6)

banobepascal:

browser/css/jsdialogs.css
```diff
@@ -541,6 +541,11 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	white-space: nowrap;
 }
 
+#ServerAuditDialog .ui-treeview-entry {
+	display: grid;
+	grid-template-columns: auto 1fr 180px;
+}
```
pedropintosilva, line 547: I see 2 problems here:

* I would prefer to avoid hard coded pixels and instead use ratios
* Where I'm missing a pixel value is on the initial column (the icon)

```css
grid-template-columns: var(--btn-img-size-s) 6fr 3fr;
```

browser/css/jsdialogs.css
```diff
@@ -541,6 +541,11 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	white-space: nowrap;
 }
 
+#ServerAuditDialog .ui-treeview-entry {
```
pedropintosilva, line 544: This element should be set in a way that allows word wrapping as I said previously so that would be:

```css
white-space: normal;
```

browser/css/jsdialogs.css
```diff
@@ -568,6 +573,25 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	/* grid-column: '1 / ' + (this._columns + 1) - set inside JS */
 }
 
+#ServerAuditDialog .ui-treeview-headers {
+	display: grid;
+	grid-template-columns: auto 1fr 170px;
+}
```
pedropintosilva, line 579: I see 2 problems here:

* I would prefer to avoid hard coded pixels and instead use ratios
* Where I'm missing a pixel value is on the initial column (the icon)

```css
grid-template-columns: var(--btn-img-size-s) 6fr 3fr;
```

browser/css/jsdialogs.css
```diff
@@ -568,6 +573,25 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 	/* grid-column: '1 / ' + (this._columns + 1) - set inside JS */
 }
 
+#ServerAuditDialog .ui-treeview-headers {
```
pedropintosilva, line 576: I have noticed that somehow the .ui-treeview-headers are being set from JS side with a grid-column: 1 / 5; which means that headers are being set in terms of grid-column-end with one extra column (5) instead of (4) as we have in the .ui-treeview-entry.

So, we either try to understand why that is happening on the JS side or for we fix it for this dialog byt adding the following via css into this `#ServerAuditDialog .ui-treeview-headers`:

```css
grid-column-end: 4 !important;
```

browser/css/jsdialogs.css
```diff
@@ -568,6 +573,25 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
...
+	display: grid;
+	grid-template-columns: auto 1fr 170px;
+}
+
+#ServerAuditDialog .ui-treeview-icon {
+	margin-inline-end: 2px;
+}
```
pedropintosilva, line 583: We don't need margins if we use column-gap as in:

```css
.ui-treeview-headers,
+.ui-treeview-entry {
+	column-gap: 8px;
+}
```

browser/css/jsdialogs.css
```diff
@@ -568,6 +573,25 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
...
+
+#ServerAuditDialog .ui-treeview-headers div.ui-treeview-icon-column {
+	margin-inline-end: 4px;
```
pedropintosilva, line 586: We don't need margins if we use column-gap as in:

```css
.ui-treeview-headers,
+.ui-treeview-entry {
+	column-gap: 8px;
+}
```

browser/css/jsdialogs.css
```diff
@@ -568,6 +573,25 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
...
+	width: 0px /* prevent display of ui-treeview-icon-column in ui-treeview-header */
+}
+
+#ServerAuditDialog .ui-treeview-entry > div:nth-last-child(1) {
+	margin-inline-start: 8px;
+}
```
pedropintosilva, line 592: We don't need margins if we use column-gap as in:

```css
.ui-treeview-headers,
+.ui-treeview-entry {
+	column-gap: 8px;
+}
```

browser/css/jsdialogs.css
```diff
@@ -2214,6 +2238,7 @@ kbd,
 
 #ServerAuditDialog .ui-treeview {
 	border-color: transparent;
+	max-width: fit-content;
```
pedropintosilva, line 2241: If previous comments I added we don't need this.

pedropintosilva (changes requested):
Additionally it seems there is a max-height being set elsewhere and the min-width of the #ServerAuditDialog-mainbox doesn't need to be so wide by default. Bellow is a the diff patch that I ended up having locally while reviewing your changes. Feel free to apply it with `git apply pedroChangesOnTopOf10925pr.txt`

[pedroChangesOnTopOf10925pr.txt](https://github.com/user-attachments/files/18504517/pedroChangesOnTopOf10925pr.txt)

browser/css/jsdialogs.css
banobepascal: Changed from pixel to ratios and used --btn-img-size-m as the ratio for the first column

browser/css/jsdialogs.css
banobepascal: Removed max-width, but added max-height to max-content as the max-height was capped at 150px 

```

.jsdialog:not(.sidebar).ui-treeview {
   max-height: 150px; 
}
```

browser/css/jsdialogs.css
banobepascal: Removed margin-inline-end

browser/css/jsdialogs.css
banobepascal: removed margin-inline-end

browser/css/jsdialogs.css
banobepascal: Used `0fr` to remove the unused icon

browser/css/jsdialogs.css
banobepascal: Added whitespace: normal ui-treeview-entry

browser/css/jsdialogs.css
banobepascal: Using `column-gap: 8px` on `ui-treeview-headers`  adds a gap between all three grid items, but we want to avoid adding extra space between the icon column and the status.

pedropintosilva (approved):

PR #10926, opened 2025-01-14
Require OpenSSL 3.x

Ashod:
This bumps the minimum OpenSSL version requirements to 3.x.

I believe gitpod uses 1.1.1 and will need to be updated before we merge this, lest we break it.
CC: @Darshan-upadhyay1110

- **configure: require OpenSSL 3.0.11 or newer**
- **wsd: logging**

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -52,9 +52,6 @@
 /* User feedback URL. */
 #undef FEEDBACK_URL
 
-/* Whether OpenSSL has PKCS5_PBKDF2_HMAC() */
-#undef HAVE_PKCS5_PBKDF2_HMAC
-
 /* Define to 1 if you have the `ppoll' function. */
 #define HAVE_PPOLL 0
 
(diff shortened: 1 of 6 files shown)


timar:
Why 3.0.11, why not 3.0.0? E.g. in my Ubuntu the version of libssl-dev is 3.0.2. The API/ABI does not change betwwen micro versions. Ubuntu (Debian) backports the security fixes.

Ashod:
> Why 3.0.11, why not 3.0.0? E.g. in my Ubuntu the version of libssl-dev is 3.0.2. The API/ABI does not change between micro versions. Ubuntu (Debian) backports the security fixes.

We can change it. I just noticed that the earliest minor updates had a number of CVE fixes. .11 was the first that was old enough that didn't have major CVE's. Debian 12 comes with .15, btw. So I assumed most other distros would be more up to date than that. .15 is released in Sep 2024, while .11 is released a year earlier.

But, to your point, perhaps we should care about the API/ABI, and not force minor versions.

Do we agree on 3.0.2, or should it be 3.0 ?

timar (approved):

PR #10927, opened 2025-01-14
cool#10630 doc electronic sign: handle visual signatures

vmiklos:
- **auto make silent rules: handle echo**
- **cool#10630 doc electronic sign: handle visual signatures**

### DIFF
--- a/browser/Makefile.am
+++ b/browser/Makefile.am
@@ -45,6 +45,10 @@ QUIET_M4 = $(QUIET_M4_$(V))
 QUIET_M4_ = $(QUIET_M4_$(AM_DEFAULT_VERBOSITY))
 QUIET_M4_0 = @echo "  M4      " $@;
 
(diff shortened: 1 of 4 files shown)


vmiklos:
https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_multi_user/3444/console failed in invalidations_spec.js, retry.

vmiklos:

The 1st patch just continues the cosmetic effort to have sane output with silent rules. The 2nd patch is what actually matters, the COOL UI now uses this new "external" mode when inserting a signature line, which makes it work with esign (you still can't move/resize in the esign case, but I'll fix that later). The 3rd patch makes the multi-user suite a bit more stable, at least I had several failures in a row and it now passed on first try.

caolanm (approved):

PR #10931, opened 2025-01-14
Private/caolan/coverity2025 01 14

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/test/UnitBadDocLoad.cpp
+++ b/test/UnitBadDocLoad.cpp
@@ -73,7 +73,7 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
 
         // Extract all json entries into a map.
         items = JsonUtil::jsonToMap(dialog.substr(sizeof("jsdialog:")));
-        auto firstId = items["id"];
+        const std::string firstId = items["id"];
 
         // Click "Yes" in a dialog
         helpers::sendTextFrame(socket, "dialogevent " + firstId + " {\"id\":\"yes\", \"cmd\": \"click\", \"data\": \"2\", \"type\": \"responsebutton\"}", testname);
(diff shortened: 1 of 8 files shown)


tools/WebSocketDump.cpp
```diff
@@ -224,6 +222,8 @@ int main (int argc, char **argv)
 {
     (void) argc; (void) argv;
 
+    SocketPoll DumpSocketPoll("websocket");
```
vmiklos, line 225: I assume this is minor, WebSocketDump sounds like some debug tool.

tools/Replay.hpp
```diff
@@ -178,6 +178,7 @@ struct Stats {
                 std::stringstream ss(line);
                 std::string key;
                 size_t value;
+                // coverity[tainted_data_argument : FALSE] - we trust the kernel-provided data
```
vmiklos, line 181: Similar here, but otherwise why not.

tools/ConfigMigrationAssistant.cpp
```diff
@@ -108,12 +108,12 @@ void MigrateLevel(const XMLConfiguration &sourceConfig, XMLConfiguration &target
                 // Keep record of these configs for post processing
                 if (commonKeyPart == NET_POST_ALLOW_HOST)
                 {
-                    netPostAllow.push_back(sourceElement);
+                    netPostAllow.push_back(std::move(sourceElement));
```
vmiklos, line 111: Aha, and no emplace_back() here, since the element is constructed already.

test/UnitBadDocLoad.cpp
```diff
@@ -73,7 +73,7 @@ UnitBase::TestResult UnitBadDocLoad::testBadDocLoadFail()
 
         // Extract all json entries into a map.
         items = JsonUtil::jsonToMap(dialog.substr(sizeof("jsdialog:")));
-        auto firstId = items["id"];
+        const std::string firstId = items["id"];
```
vmiklos, line 76: Also a good idea in general, README.CONTRIBUTING.md recommends against auto in this case.

vmiklos (approved):
Looks OK, inline comments are more about just noting things as I read this through.

(GH swaps the order of the last 2 comments depending on if you view individual commits or all changes on one page.)

PR #10932, opened 2025-01-14
refactor: allow rescaling without projecting

Minion3665:
Previously, in order to change the scale of a coordinate we roundtripped through lat/lng. The conversion here, unproject with the old zoom and then re-project with the new zoom, is valid with all point types (core pixels, css pixels and twips) but the intermediate step (a lat/lng from the unproject) isn't valid for anything except CSS pixels. The new rescale function is a simplification of the project(unproject(x)), as a single step and with duplicate terms cancelled.

Change-Id: I65fec84d2160299d5edb2d5264bc3b8ce0c1d29f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/geo/crs/CRS.js
+++ b/browser/src/geo/crs/CRS.js
@@ -38,6 +38,16 @@ L.CRS = {
 		return Math.pow(1.2, zoom);
 	},
 
+	// equivalent to doing an unproject with oldZoom then a project with newZoom
+	// except that unproject is technically invalid (so possibly confusing) for any non-css-pixel
+	// but this function will work with any scaling (including twips or core pixels)
+	rescale: function (point, oldZoom, newZoom) {
+		return L.point(
+			point.x * this.scale(newZoom - oldZoom),
+			point.y * this.scale(newZoom - oldZoom),
+		);
+	},
+
 	distance: function (latlng1, latlng2) {
 		var dx = latlng2.lng - latlng1.lng,
 		    dy = latlng2.lat - latlng1.lat;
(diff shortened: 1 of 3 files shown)


browser/src/geo/crs/CRS.js
```diff
@@ -38,6 +38,16 @@ L.CRS = {
 		return Math.pow(1.2, zoom);
 	},
 
+	// equivalent to doing an unproject with oldZoom then a project with newZoom
+	// except that unproject is technically invalid (so possibly confusing) for any non-css-pixel
+	// but this function will work with any scaling (including twips or core pixels)
+	rescale: function (point, oldZoom, newZoom) {
+		return L.point(
+			point.x * this.scale(newZoom - oldZoom),
```
gokaysatir, line 46: I will check the commit in detail now but, do we need to add a `* 0.5` here?

Minion3665: I don't... believe so? I got to this point by taking the project(unproject()) functions, writing down what they did and cancelling terms... it's possible I missed something but I'd find it weird if it were a factor of 0.5

Minion3665: I also checked this locally in a repl to see that it outputted the same thing ... I would write a unit test but as far as I know we don't have anything like [vitest](https://vitest.dev/) to test pure JavaScript so I'd have to write something in cypress ... maybe that's a thing for next hack week

gokaysatir (approved):

PR #10933, opened 2025-01-14
wip: private/caolanm/kit presets

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
This is unfortunately quite a lot of stuff wrt review. The general idea is if wopi server provides apis like https://github.com/nextcloud/richdocuments/issues/4041 and can support "shared" and "user" config/presets then there are "subforkits" that inherit from the default forkit which have the "shared" presets/config installed, and each user's document has their specific overrides installs as another layer. Equivalent to the desktop LibreOffice's shared/user distinction.

There is a cache of these presets implemented, keyed similar to how quarantine keys documents.

There is an iframe which can appear in the integrators admin (for shared settings) and user (for user settings) settings page, which can be used to upload/delete/download the presets.

Presets include "autotext" (F3 is bound to this, make run has "COLD" and "FLU" as samples) and "wordbook" which are user dictionaries.

wsd/wopi/WopiStorage.cpp
```diff
@@ -237,6 +237,11 @@ WopiStorage::WOPIFileInfo::WOPIFileInfo(const FileInfo& fileInfo, Poco::JSON::Ob
     JsonUtil::findJSONValue(object, "UserExtraInfo", _userExtraInfo);
     JsonUtil::findJSONValue(object, "UserPrivateInfo", _userPrivateInfo);
     JsonUtil::findJSONValue(object, "ServerPrivateInfo", _serverPrivateInfo);
+    if (auto settingsJSON = object->getObject("UserSettings"))
+        JsonUtil::findJSONValue(settingsJSON, "uri", _userSettingsUri);
+
+    JsonUtil::findJSONValue(object, "UserSettingsUri", _userSettingsUri);
```
Ashod, line 243: We are overwriting `_userSettingsUri` ? Doesn't look right. 

Is the idea to use `UserSettings.uri` if available, but always prefer `UserSettingsUri`, if set ?

A comment is warranted, if this is indeed intentional.

wsd/wopi/CheckFileInfo.cpp
```diff
@@ -87,6 +87,8 @@ void CheckFileInfo::checkFileInfo(int redirectLimit)
         const std::string& wopiResponse = httpResponse->getBody();
         const bool failed = (httpResponse->statusLine().statusCode() != http::StatusCode::OK);
 
+        // here we get wopiResponse
```
Ashod, line 90: Well, actually, we got the response 3 lines above. "here" we process and parse it.

wsd/SpecialBrokers.hpp
```diff
@@ -33,7 +33,7 @@ class StatelessBatchBroker : public DocumentBroker
 public:
     StatelessBatchBroker(const std::string& uri, const Poco::URI& uriPublic,
                          const std::string& docKey)
-        : DocumentBroker(ChildType::Batch, uri, uriPublic, docKey)
+        : DocumentBroker(ChildType::Batch, uri, uriPublic, docKey, /*TODO*/ "")
```
Ashod, line 36: `, /*configId=*/std::string())`

wsd/RequestVettingStation.cpp
```diff
@@ -141,6 +144,71 @@ void RequestVettingStation::sendUnauthorizedErrorAndShutdown()
...
+
+            std::string stamp;
+            JsonUtil::findJSONValue(settingsJSON, "stamp", stamp);
+            if (!stamp.empty())
+                _configId.append("-").append(stamp);
+        }
+    }
+
+    std::string _uri;
+    std::string _configId;
```
Ashod, line 169: Best to hide these before their use proliferates.

wsd/RequestVettingStation.cpp
```diff
@@ -141,6 +144,71 @@ void RequestVettingStation::sendUnauthorizedErrorAndShutdown()
...
+
+    auto finishedCallback = [selfWeak = weak_from_this(), this, configId](bool success)
+    {
+        std::shared_ptr<RequestVettingStation> selfLifecycle = selfWeak.lock();
+        if (!selfLifecycle)
+            return;
+
+        if (!success)
+        {
+            sendErrorAndShutdown(_ws, "shared config install failed",
```
Ashod, line 195: Worth checking that we have a valid `_ws` at this point. It's not a given.

wsd/DocumentBroker.cpp
```diff
@@ -1457,6 +1495,388 @@ DocumentBroker::updateSessionWithWopiInfo(const std::shared_ptr<ClientSession>&
...
+        {
+            std::string searchDir = presetsPath;
+            searchDir.append("wordbook");
+            const auto fileNames = FileUtil::getDirEntries(searchDir);
+            for (auto& fileName : fileNames)
+            {
+                std::string filePath = searchDir;
+                filePath.append("/");
+                filePath.append(fileName);
+                _presetTimestamp[fileName] = FileUtil::getLastModificationTimestamp(filePath);
```
Ashod, line 1668: Should we skip inserting/updating the map if `getLastModificationTimestamp()` fails (file doesn't exist or some other error) ?

wsd/RequestVettingStation.cpp
```diff
@@ -141,6 +144,71 @@ void RequestVettingStation::sendUnauthorizedErrorAndShutdown()
...
+                                 WebSocketHandler::StatusCodes::UNEXPECTED_CONDITION);
+        }
+        else
+        {
+            COOLWSD::ensureSubForKit(configId);
```
Ashod, line 200: The implication of the comment above the `DocumentBroker::asyncInstallPresets()` call (line 206) is that here we "launch docbroker," but it's unclear to me how this is related.
Perhaps I need to understand the workflow better.

I suspect these continuations here, in the Vetting Station, now have a large number of combinations of conditions that need testing. I'd mark this file for better review scrutiny and testing.

Ashod (commented):
Thanks everyone. This is a huge feature. I haven't finished reviewing, but I'm sharing the comments I have so far. I'll need another session (at least).

EDIT: Lest I forget, we should somehow cap the number of subforkits, for good resource management.

wsd/wopi/WopiStorage.cpp
codewithvk: I added that for testing and forgot to remove : ) 

fixed : https://github.com/CollaboraOnline/online/pull/10933/commits/438500f319e3969c16a16dd04bad6ba2187cda65. Thanks for pointing out :)

Ashod (approved):
I haven't completely finished, but I think getting it in and getting some cycles for testing is more valuable at this point. I didn't see any blockers, but I did highlight some potential problems (resource consumption is one, but there are others sprinkled in my comments).

Please see the notes/suggestions/comments inline. Some are cosmetics, others are more serious.

I'll continue to read/review offline over the next few days, and will leave comments here if necessary, or communicate offline.

caolanm:
Wrt old configs and coolwsd.xml, the idea is that packages/docker-images have:

ExecStart=/usr/bin/coolwsd --version ... --o:cache_files.path=/opt/cool/cache ...

in e.g. coolwsd.service

to set a default, and if for some reason we end up with the empty cache path, then the idea is that we simply don't cache as the ultimate fallback.

wsd/FileServer.cpp
```diff
@@ -516,6 +558,283 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request, http:
...
+        configInfo->set("wordbook", configDictionaries);
+
+        if (serveBrowserSetttings)
+        {
+            assert(kind == "user");
+            const std::string& browserSettingPath = "test/data/presets/user/browsersettings.json";
+            if (FileUtil::Stat(Poco::Path(COOLWSD::FileServerRoot, browserSettingPath).toString())
+                    .exists())
+            {
+                auto ss = std::ostringstream{};
```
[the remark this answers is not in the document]
Rash419: https://github.com/CollaboraOnline/online/pull/11001

kit/ForKit.cpp
```diff
@@ -435,90 +517,150 @@ static int createLibreOfficeKit(const std::string& childRoot,
...
+    pid_t childPid = 0;
+
+    auto childFunc = [childRoot, sysTemplate, loTemplate,
+                      subForKitIdent, useMountNamespaces]()
+    {
+        // reset parent of this subforkit to its forkit parent, main loop
+        // detects a parentPid != getppid() as a cue to exit
+        parentPid = getppid();
+
+        ForKitIdent = subForKitIdent;
```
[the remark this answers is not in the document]
caolanm: Yeah, ForKitIdent is used to identify the current forkit process, so it's empty for the primordial case. If a subforkit is requested then createSubForKit forks the process, and the new subforkit child process changes the global ForKitIdent for the process to the requested ident and announces itself as that ident to coolwsd. Later forks to generate kits from a subforkit are passed their parent ForKits ForKitIdent to use to announce themselves as kits that belong to that parent forkit.

caolanm:
https://github.com/CollaboraOnline/online/pull/11001 and https://github.com/CollaboraOnline/online/pull/11000 for the majority of the low hanging issues. Some more to follow on sendBrowserSettingsSync, etc.

kit/ForKit.cpp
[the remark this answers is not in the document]
Ashod: Thanks. That was my assumption, but it's really hard to review to verify. It should also be equally hard to maintain that relationship in the future, since it's not obvious. Can't we return it from the lambda and pass it around explicitly?

wsd/FileServer.cpp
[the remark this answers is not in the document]
Ashod: Thank you, @Rash419!

Ashod:
> Wrt old configs and coolwsd.xml, the idea is that packages/docker-images have:
> 
> ExecStart=/usr/bin/coolwsd --version ... --o:cache_files.path=/opt/cool/cache ...

@caolanm, but this wouldn't be the "default", if empty. This will **override** whatever is in the config!

The command-line takes precedence over the config (test it locally to prove me wrong).

caolanm:
Yeah, poor choice of words maybe, but I meant, that "by default" /opt/cool/cache will be set by coolwsd.service (or the equivalent docker launchers). This is what I see already for sys_template_path and child_root_path to get those to use /opt/cool. Do you think I should change something here?

wsd/DocumentBroker.cpp
```diff
@@ -3549,6 +3979,131 @@ void DocumentBroker::alertAllUsers(const std::string& msg)
...
+    for (auto& it : _sessions)
+    {
+        if (it.second->getUserId() != userId)
+            continue;
+
+        it.second->updateBrowserSettingsJSON(key, value);
+        if (!upload)
+        {
+            uploadBrowserSettingsToWopiHost(it.second);
+            upload = true;
```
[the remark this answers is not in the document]
Rash419: > We shouldn't make assumptions about the user's device/browser/setup. We should just update all sessions with the userId and upload to all of them. 

@Ashod still unclear to me why would we need to upload the same json multiple times. 
Lets consider following scenario:
1. userA with userId=1 opens the writer document
2. userA with userId=1 open presentation document in different browser
3. userA makes changes in writer session, `syncBrowserSettings` syncs browsersetting json between all the session with userId=1 in this case it will update the browser setting in writer session and presentation session
4. Now all the session with userId=1 has same browser setting json 

my question: why we need to upload the same browser setting json to wopihost twice in this case ?

PR #10939, opened 2025-01-15
Private/caolan/coverity2025 01 15

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
I see nothing problematic here.

PR #10940, opened 2025-01-15
Add "Print" option in menubar for readonly mode in all app

Darshan-upadhyay1110:
- Current case only writer had the print option available in menubar
- added print option for all apps by adding this to `AllowedView Mode action` list

Change-Id: I0884fe0809bb1de5eb12c141a9ee9ce499096212

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Thanks for the improvement :)

PR #10943, opened 2025-01-15
Avoid redundant `getpresentationinfo` call when `startPresentation` parameter is used

Darshan-upadhyay1110:
- Added a `_startingPresentation` flag to track the state of the presentation startup.
- Modified the `onUpdateParts` method to prevent redundant `getpresentationinfo` calls during initialization.
- Updated the presentation start logic to set the flag when the presentation begins, ensuring the request is sent only once.

* Resolves: #10198 

Change-Id: Ib1edc8b57a3ec879db93cffbd649b5b96d0004f4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/slideshow/SlideShowPresenter.ts
+++ b/browser/src/slideshow/SlideShowPresenter.ts
@@ -129,6 +129,7 @@ class SlideShowPresenter {
 	private _cypressSVGPresentationTest: boolean = false;
 	private _onKeyDownHandler: (e: KeyboardEvent) => void;
 	private _onImpressModeChanged: any = null;
+	private _startingPresentation: boolean = false;
 
 	constructor(map: any) {
 		this._cypressSVGPresentationTest =
@@ -172,7 +173,8 @@ class SlideShowPresenter {
 	}
 
 	private onUpdateParts() {
-		if (this._checkAlreadyPresenting()) this.onSlideShowInfoChanged();
+		if (this._checkAlreadyPresenting() && !this._startingPresentation)
+			this.onSlideShowInfoChanged();
 	}
 
 	public getNavigator() {
@@ -407,6 +409,7 @@ class SlideShowPresenter {
 		}
 
 		this._canvasLoader.startLoader();
+		this._startingPresentation = false;
 	}
 
 	public stopLoader(): void {
@@ -705,7 +708,7 @@ class SlideShowPresenter {
 			return;
 		// disable slide sorter or it will receive key events
 		this._map._docLayer._preview.partsFocused = false;
-
+		this._startingPresentation = true;
 		app.socket.sendMessage('getpresentationinfo');
 	}
 
@@ -717,6 +720,7 @@ class SlideShowPresenter {
 			return;
 		// disable present in console onStartInWindow
 		this._enablePresenterConsole(true);
+		this._startingPresentation = true;
 		app.socket.sendMessage('getpresentationinfo');
 	}
 


browser/src/slideshow/SlideShowPresenter.ts
```diff
@@ -813,7 +813,16 @@ class SlideShowPresenter {
...
+		// Do not call getpresentationinfo if the presentation starts directly via the URL parameter.
+		// Reason: In the _onStart handler, slide info is already requested once.
+		// Calling getpresentationinfo again would be redundant in this case.
+		const startPresentationGet: string =
+			this._map.isPresentationOrDrawing() &&
+			window.coolParams.get('startPresentation');
+		const isPresentationStartParam: boolean =
+			startPresentationGet === 'true' || startPresentationGet === '1';
+		if (!isPresentationStartParam)
+			app.socket.sendMessage('getpresentationinfo');
```
meven, line 825: What I would do instead of a hard condition with "&startPresentation", you could have a boolean guard to prevent `app.socket.sendMessage('getpresentationinfo');` to be called more than necessary.

The boolean guard should be reset sometimes, but I can't tell you when.

Because there is the case of autostarting files, ppsx/pps files. (https://gerrit.libreoffice.org/c/core/+/163652)

meven (changes requested):

browser/src/slideshow/SlideShowPresenter.ts
Darshan-upadhyay1110: @meven, I understand your point, but in this case, `getpresentationinfo` is called twice only here. If I set a global variable, it would only be used in this context, so I don’t see the need for storing it with a broader scope.

This is happening because of slide gets loaded on main screen and i am bit concern about `		this._map.on('updateparts', this.onUpdateParts, this);`

If i do set and reset of a variable if will break many things to get slide data from CORE.

Any thoughts ?

Darshan-upadhyay1110: The current issue arises because the process is too quick, and the presentation takes some time to start. As a result, `_checkAlreadyPresenting` does not return the expected value during the initial load. This triggers the normal routine of updating slide info, even though the presentation hasn't fully started yet.

```
private onUpdateParts() {
    if (this._checkAlreadyPresenting()) this.onSlideShowInfoChanged();
}
```

Here, `_checkAlreadyPresenting` isn't set in time because of the numerous calls and processes happening during the first load.

meven: > @meven, I understand your point, but in this case, getpresentationinfo is called twice only here. If I set a global variable, it would only be used in this context, so I don’t see the need for storing it with a broader scope.

I didn't mean a broader var, I was suggesting just an internal guard for SlideShowPresenter.
Having a generic guard preventing mishaps, is better than having a hard condition, then used in _start and onSlideShowInfoChanged. We could use `onSlideShowInfo` to disable the state back.

SlidePresenter should be able to work without paying attention to outside factors IMO

The fact that `dispatch('presentation');` isn't succeeding is the issue.
Or the call site must be corrected, or pass-in a parameter.

I testd ppsx like `sd/qa/unit/tiledrendering/data/test.ppsx` in core and it doesn't exhibit the issue.

A senior reviewer would probably have a opinion perhaps @eszkadev

meven: I have another suggestion, I tested it:
```
diff --git browser/src/slideshow/SlideShowPresenter.ts browser/src/slideshow/SlideShowPresenter.ts
index 64e880fa3a..17c26df55f 100644
--- browser/src/slideshow/SlideShowPresenter.ts
+++ browser/src/slideshow/SlideShowPresenter.ts
@@ -129,6 +129,7 @@ class SlideShowPresenter {
        private _cypressSVGPresentationTest: boolean = false;
        private _onKeyDownHandler: (e: KeyboardEvent) => void;
        private _onImpressModeChanged: any = null;
+       private _startingPresentation: boolean = false;
 
        constructor(map: any) {
                this._cypressSVGPresentationTest =
@@ -172,7 +173,7 @@ class SlideShowPresenter {
        }
 
        private onUpdateParts() {
-               if (this._checkAlreadyPresenting()) this.onSlideShowInfoChanged();
+               if (this._checkAlreadyPresenting() && !this._startingPresentation) this.onSlideShowInfoChanged();
        }
 
        public getNavigator() {
@@ -407,6 +408,7 @@ class SlideShowPresenter {
                }
 
                this._canvasLoader.startLoader();
+               this._startingPresentation = false;
        }
 
        public stopLoader(): void {
@@ -706,6 +708,7 @@ class SlideShowPresenter {
                // disable slide sorter or it will receive key events
                this._map._docLayer._preview.partsFocused = false;
 
+               this._startingPresentation = true;
                app.socket.sendMessage('getpresentationinfo');
        }
 
@@ -717,6 +720,7 @@ class SlideShowPresenter {
                        return;
                // disable present in console onStartInWindow
                this._enablePresenterConsole(true);
+               this._startingPresentation = true;
                app.socket.sendMessage('getpresentationinfo');
        }
```

Darshan-upadhyay1110: @meven awesome hack to do. I also tested the cases it works well. Thanks for the suggestion :)

Darshan-upadhyay1110:
Nice idea form @meven 

Todo for me : A enum state, initial, loading, started, presenting, finished for instance would have been great

meven (approved):

meven:
> Nice idea from @meven
> 
> Todo for me : A enum state, initial, loading, started, presenting, finished for instance would have been great

And bonus point for a cypress test.

We can thank typescript for adding enums \o/

quikee (approved):
Look fine to me

Issue #10942, opened 2025-01-15
Scroll issues with toolbar and statusbar on LTR and RTL docs


bayramcicek:
### Describe the Bug
**LTR docs:**
- Scrolling toolbar and statusbar with `Shift` + `mouseWheelUp/Down` does not hide/show the scroll right/left buttons correctly.
- Therefore right/left buttons overlap with other elements

**RTL docs:**
- all LTR issues exist.
- scrolling with clicking on the buttons sometimes does not show/hide buttons. (this works well on LTR docs.)
  - the reason: `this._RTL` always returns `undefined`. So we couldn't know if the doc is RTL or LTR.
- scroll-right buttons' arrows are not aligned well. 

### LTR:

https://github.com/user-attachments/assets/17d2552b-68f6-4d9c-a3cf-42c1af40089b

- no misaligned scroll arrows - good

![Screenshot_20250115_140000](https://github.com/user-attachments/assets/6dcfaf46-0e92-4d39-a49b-812ae3987dd3)

---

### RTL:

https://github.com/user-attachments/assets/0a17edc8-1a25-4f9e-bbc1-5c6c31fee9bd

- misaligned scroll-right arrows - bad

![Screenshot_20250115_140012](https://github.com/user-attachments/assets/53f69460-b360-4b70-96f7-ca8fc58260d8)

### Online
COOLWSD version: 24.04.12.0(git hash: a72bfd0b2c)
LOKit version: Collabora Office 24.04.12.0(git hash: 1c18da8)
Served by: Manjaro Linux
Server ID: b77a286d
WOPI host:

PR #10944, opened 2025-01-15
Calc toolbar/statusbar: fix scroll button issues on horizontal scroll

bayramcicek:
- fixed "this._RTL always returns undefined" issue. scroll buttons are now correctly visible/hidden when scrolling horizontally with right/left scroll buttons.

- aligned scroll-right button arrows on RTL docs
    - scroll-right button arrows of statusbar and toolbar were misaligned to the left. This is fixed with 'left: 0px'

-  Calc: show/hide scroll buttons correctly with shift + mouse wheel
    - Fixed: "Scrolling toolbar and statusbar with Shift + mouseWheelUp/Down does not hide/show the scroll right/left buttons correctly. Therefore, right/left buttons overlap with other elements." 

Change-Id: Ieeb124df61ba0e2dea94ea256ec1daa2f87b220b

* Resolves: #10942 
* Target version: master 

### TODO

**LTR docs:**
- [x] Scrolling toolbar and statusbar with `Shift` + `mouseWheelUp/Down` does not hide/show the scroll right/left buttons correctly. Therefore right/left buttons overlap with other elements

**RTL docs:**
- [x] all LTR issues exist.
- [x] scrolling with clicking on the buttons sometimes does not show/hide buttons. (this works well on LTR docs.)
  - the reason: `this._RTL` always returns `undefined`. So we couldn't know if the doc is RTL or LTR.
- [x] scroll-right buttons' arrows are not aligned well. 

### Checklist

### DIFF
--- a/browser/src/control/jsdialog/Util.ScrollableBar.ts
+++ b/browser/src/control/jsdialog/Util.ScrollableBar.ts
@@ -48,7 +48,8 @@ function showArrow(arrow: HTMLElement, show: boolean) {
 function setupResizeHandler(container: Element, scrollable: Element) {
 	const left = container.querySelector('.ui-scroll-left') as HTMLElement;
 	const right = container.querySelector('.ui-scroll-right') as HTMLElement;
-	var isRTL = document.documentElement.dir === 'rtl';
+	var isRTL: boolean = document.documentElement.dir === 'rtl';
+	var timer: any; // for shift + mouse wheel up/down
 
 	const handler = function () {
 		const rootContainer = scrollable.querySelector('div');
@@ -77,8 +78,24 @@ function setupResizeHandler(container: Element, scrollable: Element) {
 		}
 	}.bind(this);
 
+	// handler for toolbar and statusbar
+	// runs if shift + mouse wheel up/down are used
+	const shiftHandler = function (e: MouseEvent) {
+		const rootContainer = scrollable.querySelector('div');
+		if (!rootContainer || !e.shiftKey) return;
+
+		clearTimeout(timer);
+		// wait until mouse wheel stops scrolling
+		timer = setTimeout(function () {
+			JSDialog.RefreshScrollables();
+		}, 350);
+	}.bind(this);
+
 	window.addEventListener('resize', handler);
 	window.addEventListener('scroll', handler);
+	document
+		.querySelector('.ui-scrollable-content')
+		.addEventListener('wheel', shiftHandler);
 }
 
 JSDialog.MakeScrollable = function (parent: Element, scrollable: Element) {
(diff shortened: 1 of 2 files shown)


browser/src/control/jsdialog/Util.ScrollableBar.ts
```diff
@@ -48,35 +48,54 @@ function showArrow(arrow: HTMLElement, show: boolean) {
...
+		timer = setTimeout(function () {
+			JSDialog.RefreshScrollables();
+		}, 350);
+	}.bind(this);
+
 	window.addEventListener('resize', handler);
 	window.addEventListener('scroll', handler);
+	document
+		.querySelector('.ui-scrollable-content')
+		.addEventListener('wheel', shiftHandler);
```
eszkadev, line 98: if `setupResizeHandler` is called many times, this will add multiple 'wheel' handlers to the same objects. (every time we add to all of them). I think it should add only to the current one.

we should avoid at all cost querying `document`, please query only inside current component

browser/src/control/jsdialog/Util.ScrollableBar.ts
```diff
@@ -48,35 +48,54 @@ function showArrow(arrow: HTMLElement, show: boolean) {
...
+	const shiftHandler = function (e: MouseEvent) {
+		const rootContainer = scrollable.querySelector('div');
+		if (!rootContainer || !e.shiftKey) return;
+
+		clearTimeout(timer);
+		// wait until mouse wheel stops scrolling
+		timer = setTimeout(function () {
+			JSDialog.RefreshScrollables();
+		}, 350);
+	}.bind(this);
```
eszkadev, line 92: tip: in TS arrow functions ("lambda" `(param) => {code}`) automatically bind this, please try to use just `(e: MouseEvent) => { ... code ... }` instead  :)

bayramcicek: Thanks for the tip :)
DONE.

browser/src/control/jsdialog/Util.ScrollableBar.ts
bayramcicek: What I understand with the `query only inside current component` is that the `container: Element` parameter of the `setupResizeHandler` function. So I used it in the last commit. I hope it is OK.
Thanks.

browser/src/control/jsdialog/Util.ScrollableBar.ts
```diff
@@ -48,35 +48,55 @@ function showArrow(arrow: HTMLElement, show: boolean) {
...
+		clearTimeout(timer);
+		// wait until mouse wheel stops scrolling
+		timer = setTimeout(function () {
+			JSDialog.RefreshScrollables();
+		}, 350);
+	};
+
 	window.addEventListener('resize', handler);
 	window.addEventListener('scroll', handler);
+	const scrollableContent = container.querySelector(
```
eszkadev, line 96: don't you get the same instance as `scrollable` passed as an argument?
see `MakeScrollable`

eszkadev (approved):
Thanks, code looks good now :)
Nice small logically separated commits btw

PR #10950, opened 2025-01-16
Add tiny scrollbar for stylepreview in notebookbar

Darshan-upadhyay1110:
- A small visible scrollbar should also be included in stylepreview.

Before:
image

After:

image

Change-Id: I73131ac6a6753cd8cc2907f7c8397f50948637ba

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


pedropintosilva (changes requested):
Thanks! Do you think you can investigate a way to prevent scrollbar to be clipped (meaning: right now - on chrome - it appears above the border thus making the whole stylesview component not round in the right edge)

Darshan-upadhyay1110:
Added Other PR which will have the option with scroll buttons


https://github.com/CollaboraOnline/online/pull/11039

PR #10951, opened 2025-01-16
Private/caolan/coverity2025 01 16

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Message.hpp
+++ b/common/Message.hpp
@@ -199,7 +199,9 @@ private:
     std::string getForwardToken(const char* buffer, int length)
     {
         std::string forward = COOLProtocol::getFirstToken(buffer, length);
-        return (forward.find('-') != std::string::npos ? forward : std::string());
+        if (forward.find('-') != std::string::npos)
+           return forward;
+        return std::string();
     }
 
     std::vector<char> copyDataAfterOffset(const char *p, size_t len, size_t fromOffset)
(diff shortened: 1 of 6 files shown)


caolanm:
a few more little coverity warning fixes, use the three arg wait_for variant to guard against the "spurious wakeup" that coverity is warning about there

common/Message.hpp
```diff
@@ -199,7 +199,9 @@ class Message
     std::string getForwardToken(const char* buffer, int length)
     {
         std::string forward = COOLProtocol::getFirstToken(buffer, length);
-        return (forward.find('-') != std::string::npos ? forward : std::string());
+        if (forward.find('-') != std::string::npos)
```
vmiklos, line 202: Nit: while I like the new version as it's more obvious to step through it in gdb, is this here to work around some Coverity bug or I miss something and indeed the new version does less copying?

caolanm: I don't think its a coverity bug, in the sense that coverity basically recommends in its warning to have:

return (forward.find('-') != std::string::npos ? std::move(forward) : std::string());

which fits with e.g. https://fekir.info/post/guaranteed-copy-elision/ "the ternary operator forces a copy in a return statement (return expr ? a : b), unless both a and b are temporaries"

But adding explicit std::move to returns feels like something to be avoided: https://pvs-studio.com/en/blog/terms/6516/

So I'm drawn to "An if statement seems to be a superior alternative."

vmiklos: I see now, thanks.

PR #10954, opened 2025-01-17
annotation: refocus comments on tab getting focus

lpranam:
problem:
if user was editing comment and switched tab/window, on returning to tab document will gain focus instead of comment and user accidentally type in doc instead of comment

Change-Id: I356f311461010fc191620ac489f6b3bc65434e2b

* Target version: master 

### Checklist


lpranam:
is it possible to change tab/window in cypress?

PR #10958, opened 2025-01-17
Adding some features and fixes to undoStatistics.py

elliotfreebairn1:
* Resolves: # 
* Target version: master 

### Summary

- Headings and text columns are bold now
- chart script combined into undoStats (removed the charting script file)
- Tables are all sorted now - for instance
- Viewer_Editors, sorted by number of editors then number of viewers
- Added 3 more charts - Total Viewers Per Doc, Total Editors Per Doc, Sliced version of command matrix
- Changed Viewer_Editors, so the column Editor | Viewer is the column/key
- Each table now has an x axis title (if needed)
- Changed to snake Case.
- Allowed charts to chart more than 26 columns
- command matrix is now sorted by hot pairs into the top left.
- Added heat map effect to the command matrix. As of now, the full matrix has heat map off on default and the sliced matrix has it on.
- Added the ability to rotate top row of text by 90 degrees. However, It's off as default due to it not looking great.
-Reduced the sizes of columns as they were too large

- [ ] ...

### Checklist


codewithvk (approved):
Thanks @elliotfreebairn1. Code Looks good.

codewithvk:
restarted CI

Issue #10589, opened 2024-11-23
coolwsd.xml defaults vs. built-in defaults


timar:
### Describe the Bug
In #10049 it was requested that coolwsd.xml defaults should match built-in defaults, and some progress was made. Now we have a systematic approach. 
`coolwsd` logs non-default values from coolwsd.xml at startup. 

For example:
`podman run -i -t --name collabora_code -p 9980:9980 -e "extra_params=--o:ssl.enable=false --o:ssl.termination=false" docker.io/collabora/code:latest`

```
wsd-00001-00001 2024-11-23 11:15:35.736304 +0000 [ coolwsd ] INF  Loaded config file [/etc/coolwsd/coolwsd.xml] (non-default values):
        admin_console.enable: true
        admin_console.password: <redacted>
        admin_console.username: <redacted>
        fonts_missing.handling: log
        indirection_endpoint.geolocation_setup: false
        indirection_endpoint.migration_timeout_secs:
        languagetool.enabled:
        languagetool.ssl_verification:
        logging.anonymize.anonymization_salt: <redacted>
        logging.anonymize.anonymize_user_data: false
        logging.color: false
        logging.disabled_areas: Socket,WebSocket,Admin,Pixel
        logging.file.property: /var/log/coolwsd.log
        logging.file.property[0]: /var/log/coolwsd.log
        logging.file.property[2]: timestamp
        logging.least_verbose_level_settable_from_client: fatal
        logging.most_verbose_level_settable_from_client: notice
        net.connection_timeout_secs:
        net.lok_allow.host: 192\.168\.[0-9]{1,3}\.[0-9]{1,3}
        net.lok_allow.host[10]: ::ffff:172\.3[01]\.[0-9]{1,3}\.[0-9]{1,3}
        net.lok_allow.host[11]: 10\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}
        net.lok_allow.host[12]: ::ffff:10\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}
        net.lok_allow.host[13]: localhost
        net.lok_allow.host[1]: ::ffff:192\.168\.[0-9]{1,3}\.[0-9]{1,3}
        net.lok_allow.host[2]: 127\.0\.0\.1
        net.lok_allow.host[3]: ::ffff:127\.0\.0\.1
        net.lok_allow.host[4]: ::1
[...]

PR #10959, opened 2025-01-17
Configuration defaults and documentation

timar:
* Resolves: #10589 
* Target version: master


ios/Mobile.xcodeproj/project.pbxproj
```diff
@@ -3863,7 +3863,6 @@
 					"DISABLE_DYNLOADING=1",
 					"ENABLE_DEBUG=1",
 					"COOLWSD_CONFIGDIR='\".\"'",
-					"COOLWSD_LOGLEVEL='\"trace\"'",
```
vmiklos, line 3866: If this is truly unused, remove it also from android/lib/src/main/cpp/CMakeLists.txt.in?

I'm but exactly sure how this works, but seems DefAppConfig in common/ConfigUtil.cpp still uses COOLWSD_LOGLEVEL, which seems to be outside any "if not on mobile" ifdefs, so I guess this was then indeed redundant and you're right.

vmiklos (approved):
Looks OK in general, see inline comments for some thoughts.

PR #10960, opened 2025-01-17
Private/caolan/coverity2025 01 17

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -470,7 +470,7 @@ private:
         if constexpr (!Util::isMobileApp())
         {
 #if ENABLE_DEBUG
-            if (std::getenv("COOL_ZERO_BUFFER_SIZE"))
+            if (std::getenv("COOL_ZERO_BUFFER_SIZE") && _fd >= 0)
             {
                 const int oldSize = getSocketBufferSize();
                 setSocketBufferSize(0);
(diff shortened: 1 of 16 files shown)


caolanm:
Add some markup for dumpState to disable warnings for the known thread-unsafe dumping we do in those. Rework some read loops into a pattern that coverity is happy with but are logically unchanged. And some minor std::move, reference passing tweaks.

net/Socket.hpp
```diff
@@ -470,7 +470,7 @@ class Socket
         if constexpr (!Util::isMobileApp())
         {
 #if ENABLE_DEBUG
-            if (std::getenv("COOL_ZERO_BUFFER_SIZE"))
+            if (std::getenv("COOL_ZERO_BUFFER_SIZE") && _fd >= 0)
```
vmiklos, line 473: Hmm, but then if isClosed() called twice, you'll get a positive & invalid _fd. Still, this looks OK, I complained about the broader issue at https://github.com/CollaboraOnline/online/pull/10874#issuecomment-2592410559, hopefully soon @Ashod will have time to take a look.

vmiklos (approved):
Looks OK in general, see inline comments for some nits.

PR #10963, opened 2025-01-20
Add an initial COOL-specific clang plugin

vmiklos:
As suggested at
<https://github.com/CollaboraOnline/online/pull/10838#pullrequestreview-2527339319>,
detect some cases where capturing variables by reference in a lambda is
not safe, assuming that most lambdas are long-lived in coolwsd.

It would fail without commit
19861a2d0bdc2db9825f9d4b953ce3e0a7a3539b (kit, BackgroundSaveWatchdog:
fix memory management in the ctor, 2024-11-20) with the following
output:

kit/Kit.cpp:183:67: error: function parameter captured by reference, capture by value instead
                  Util::setThreadName("kitbgsv_" + Util::encodeId(mobileAppDocId, 3) + "_wdg");
                                                                  ^~~~~~~~~~~~~~
1 error generated.

And it would fail without commit
92efdebaeb3dae25eedb02801ac395ad6abf95dd (Fix memory handling in
Session::asyncConnect(), 2025-01-03) with the folliwng output:

./net/HttpRequest.hpp:1734:107: error: function parameter captured by reference, capture by value instead
            poll.addCallback([selfLifecycle = shared_from_this(), this, &poll, socket=std::move(socket), &result]() {
                                                                                                          ^~~~~~
1 error generated.

The clang API is nominally unstable, but this simple plugin seems to
work with openSUSE clang 15 and Fedora clang 19 without any ifdefs so
far.

Change-Id: Ic5b40198f9c6c192603395fec6d6522951cb535e

### DIFF
--- a/clang/Makefile
+++ b/clang/Makefile
@@ -1,0 +1,18 @@
+# Build this without automake, so that the plugin is not used when building the plugin itself.
+
+CLANG_CXXFLAGS = $(shell llvm-config --cxxflags)
+
+plugin.so: plugin.o Makefile
+	clang++ -shared plugin.o -o plugin.so
+
+plugin.o: plugin.cpp Makefile
+	clang++ -O2 -g $(CLANG_CXXFLAGS) -Wall -Wextra -Werror plugin.cpp -fPIC -c -o plugin.o
+
+run:
+	clang++ -Werror -Xclang -load -Xclang ./plugin.so -Xclang -add-plugin -Xclang coplugin -c -o test/capture.o test/capture.cpp
+
+check:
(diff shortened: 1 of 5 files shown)


vmiklos:

I guess this is rather safe, since we don't have a CI job that would enforce we don't have warnings from this new "coplugin". Otherwise just a minimal checker that does something useful. One notable difference from core.git loplugin is that logplugin seems to use the older "recursive AST visitor" API to do the checks and it seems Mozilla/Google gave up on such manual visiting of the complex AST and they recommend "matchers" instead. So let's try that approach.

caolanm (approved):
I don't have any particular clang-plugin knowledge, but it looks reasonable to me

PR #10965, opened 2025-01-20
Comment section: Use int for part hash.

gokaysatir:
Check container existence before setting edit mode.

Fixes adding comment to wrong part in Impress.

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -2366,17 +2366,6 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		this.disableLayoutAnimation = true;
 		var comment;
 		if (Comment.isAnyEdit()) {
-			this.map.uiManager.showConfirmModal(
-				'comments-update',
-				_('Comments Updated'),
-				_('Another user has updated comments. Would you like to proceed updating?'),
-				_('OK'),
-				() => {
-					CommentSection.pendingImport = false;
-					this.clearList();
-					this.importComments(commentList);
-				}
-			);
 			CommentSection.pendingImport = true;
 			return;
 		}


gokaysatir:
@lpranam can you have a look when you have time? This fixes the issue you found earlier. cc @eszkadev

browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -1720,7 +1724,7 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 	private adjustCommentFileBasedView (comment: any): void {
 		// Below calculations are the same with the ones we do while drawing tiles in fileBasedView.
 		var partHeightTwips = app.map._docLayer._partHeightTwips + app.map._docLayer._spaceBetweenParts;
-		var index = app.impress.getIndexFromSlideHash(comment.parthash);
+		var index = app.impress.getIndexFromSlideHash(parseInt(comment.parthash));
```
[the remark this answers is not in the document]
gokaysatir: Oh, so it can be a string. I didn't see while testing. But it makes sense, let me update it.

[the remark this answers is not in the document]
gokaysatir: I have just checked. It's an integer. impress.partList is using integer values for part hashes.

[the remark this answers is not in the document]
gokaysatir: Here it is sent from server as integer values:
`app.impress.partList = Object.assign([], statusJSON.parts);`

lpranam:
When I click somewhere in empty area and then try to insert comment with keyboard shortcut I am getting `A comment is being edited` dialog

browser/src/canvas/sections/CommentSection.ts
```diff
@@ -120,7 +120,7 @@ export class Comment extends CanvasSectionObject {
 		this.sectionProperties.showSelectedCoordinate = true; // Writer.
 
 		if (app.map._docLayer._docType === 'presentation' || app.map._docLayer._docType === 'drawing') {
-			this.sectionProperties.parthash = this.sectionProperties.data.parthash;
+			this.sectionProperties.parthash = parseInt(this.sectionProperties.data.parthash);
```
eszkadev, line 123: if `parthash` inside `data` should be always a number, shouldn't we do it once when we receive that data and assign to `sectionProperties`? so we don't have to convert again.
Can we have `number` type assigned to it so TS compiler will warn about incorrect usage?

eszkadev:
Can we have some cypress test here too?

browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -2366,17 +2366,6 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		this.disableLayoutAnimation = true;
 		var comment;
 		if (Comment.isAnyEdit()) {
-			this.map.uiManager.showConfirmModal(
```
eszkadev, line 2365: So probably we should detect when we initiate import or not (as optional param?)
Then we can show the dialog only when needed

lpranam:
comments are working as expected now

browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -1305,8 +1306,11 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 	}
 
 	public add (comment: any): cool.Comment {
-		if (!comment.sectionProperties)
-			comment = new cool.Comment(comment, comment.id === 'new' ? {noMenu: true} : {}, this);
+		if (!comment.sectionProperties) {
+			const temp = new cool.Comment(comment, comment.id === 'new' ? {noMenu: true} : {}, this);
+			temp.sectionProperties.data = comment;
```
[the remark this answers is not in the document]
gokaysatir: It's a special override. I have refactored the code but it's still a bit confusing. Will probably be better with the next cycle.

browser/src/canvas/sections/CommentListSection.ts
gokaysatir: We have below lines:
```
			if (refreshAnnotation)
				app.socket.sendMessage('commandvalues command=.uno:ViewAnnotations');
```

I guess that re-import breaks things. My take is to keep this modal out for now. We have onACKComment and it also seems to do the job.

eszkadev: If it is only for now, then would be good to keep strings so we will not lost translations already done.
Could you move it to separate method (not used) - if that's the case. If not we can delete.

browser/src/canvas/sections/CommentSection.ts
gokaysatir: I think this is the right place. This is constructor and sectionProperties.data is an input parameter.

browser/src/canvas/sections/CommentListSection.ts
gokaysatir: In fact, i think refreshAnnotation boolean is somehow needed and it re-imports all the comments. In that case, "importComments" is not something that is called at initialization. All the comments are re-imported when that function is called. Including the one we just created and still editing. So i think we can't have that confirmation modal there.

We need to remove it. Then we can have a better look at the function calls and add something else i guess.

gokaysatir:
> Can we have some cypress test here too?

I will check and hopefully add a Cypress test soon.

eszkadev:
added minimal version with updated commit message with bug reproducer: https://github.com/CollaboraOnline/online/pull/11094

eszkadev:
rebased

eszkadev:
do we need this still? there is already merged the   string -> int for hash

gokaysatir:
Closed in favor of: https://github.com/CollaboraOnline/online/pull/11168

PR #10966, opened 2025-01-20
focusable element

hcvcastro:
- **browser: jsbuilder: check if child has focusable element**
- **browser: jsdialog: fix keys event handlers**


eszkadev (approved):
looks ok, can we have some cypress test for that too?

hcvcastro:
sure

PR #10971, opened 2025-01-21
wsd: shutdown and close sockets early

Ashod:
We now close socket FDs when we shutdown
the underlying connection, rather than
waiting until the Socket object is destroyed.

This fixes a recent regression where we leaked
FDs because we invalidated before closing.
This was introduced in af747c1.

In addition, we unify the logic to shutdown
and close and localize all in setClosed().

Change-Id: Ib3cb84394fca53f41a64aba9201c5779ee536dcf

### DIFF
--- a/net/Socket.hpp
+++ b/net/Socket.hpp
@@ -159,16 +159,7 @@ public:
     {
         LOG_TRC("Socket dtor");
 
-        // Doesn't block on sockets; no error handling needed.
-        if constexpr (!Util::isMobileApp())
-        {
-            ::close(_fd);
-            LOG_DBG("Closed socket " << toStringImpl());
-        }
-        else
-        {
-            fakeSocketClose(_fd);
-        }
+        setClosed();
     }
 
     /// Returns true iff this socket has been closed or is invalid.
@@ -217,15 +208,7 @@ public:
     /// TODO: Support separate read/write shutdown.
     virtual void shutdown()
     {
-        if (!_noShutdown && !isClosed())
-        {
-            LOG_TRC("Socket shutdown RDWR. " << *this);
-            if constexpr (!Util::isMobileApp())
-                ::shutdown(_fd, SHUT_RDWR);
-            else
-                fakeSocketShutdown(_fd);
-            setClosed(); // Invalidate the FD.
-        }
+        setClosed(); // Shutdown and close if not closed.
     }
 
     /// Prepare our poll record; adjust @timeoutMaxMs downwards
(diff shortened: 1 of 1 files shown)


vmiklos (commented):
At least the reported

```
(cd test; CPPUNIT_TEST_NAME=HttpRequestTests::test500GetStatuses ./unithttplib)
```

now passes for me with the fd limit reduced back to 1024, while previously it required 2048. Thanks.

net/Socket.hpp
```diff
@@ -435,18 +418,42 @@ class Socket
...
+                if constexpr (!Util::isMobileApp())
+                    ::shutdown(fd, SHUT_RDWR);
+                else
+                    fakeSocketShutdown(fd);
+            }
+
+            // Doesn't block on sockets; no error handling needed.
+            if constexpr (!Util::isMobileApp())
+            {
+                ::close(fd);
```
mmeeks, line 448: Well - things have changed a lot here. 'shutdown' now calls 'close' on file-descriptors that it used to call 'shutdown' on - that's quite a difference. Is that thought through ? why do we have a 'shutdown' method and not a 'close()' method instead of 'setClosed()' - why is that semantic now different to what it was before.

All looks very risky to me. But then we should get it in in red. We should badly go back and understand why we did this shutdown RD_RW - was that so that we can still 'poll' on a socket that is shutdown but not closed without aborting or ? ... I'd really like some detailed thought on this *and* cleaning this up - we should have a clean semantic here. We should name 'setClosed()' to 'close()' - and remove the 'shutdown' method, and touch all call sites of that to clean this up and we should read all the log entries on 'shutdown'.

Surprised we managed to merge a socket leak here - clearly there was not enough thought before - now we should do more and more carefully; and I'm not convinced that happened.

mmeeks (approved):
Lets get it in for red - but @Ashod this is increases risk over reverting and re-understanding the previous change that caused the leak.

net/Socket.hpp
Ashod: Fair points. Thanks for the feedback, @mmeeks.

I suspect we missed the leak because of the shutdown vs close distinction.

Agreed that `setClosed()` needs renaming. I didn't do it in this patch to minimize changes.

Regarding the design/approach, the fundamental issue being addressed (in the previous patch and this one) is that we should not have a `close(2)` call more than once on a given FD, lest we call close on a recycled FD, closing a valid socket/file elsewhere. This is why I moved the `close(2)` call to the shutdown member function. 

Technically, we could still have `close(2)` called only in the destructor. That would be closer to the original approach. But since we don't have a separate flag to track logically closed FDs vs actually closed FDs (i.e. having called `close(2)` on them), there would be no reliable way to guarantee that `close(2)` is called once and only once.

This approach puts `shutdown(2)` and `close(2)` in the same member, and we only apply them on a valid FD, prior to invalidating.

This is the better design. Will review the `virtual shutdown()` calls, but there shouldn't be any code that relies on the FD post `shutdown(2)`. Conversely, we need the `virtual shutdown()` to hard-close sockets immediately as necessary (f.e. on invalid external requests), earlier than destroying the socket.

PR #10974, opened 2025-01-21
fd leak: just revert instead of an invasive fix for now

vmiklos:
- **Revert "wsd: shutdown and close sockets early"**
- **Revert "wsd: invalidate socket FD by negating"**
- **Revert "wsd: replace _closed member with invalidating the FD"**

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -619,7 +619,7 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
                 // re-entrancy hazard
                 LOG_DBG("Unexpected socket poll resize");
             }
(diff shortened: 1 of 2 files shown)


vmiklos:

Just a set of reverts, which feels safer than the invasive fix: so this goes to COOL and the original fix goes to red CODE.

CC @Ashod

caolanm (approved):
brings us to where we were before

Ashod:
> Just a set of reverts, which feels safer than the invasive fix: so this goes to COOL and the original fix goes to red CODE.

These patches were intended for red CODE. This is good approach. Thanks for reverting.

Thank you.

PR #10978, opened 2025-01-21
calc: drag select outside the map

JaumePujantell:
In calc, when handling auto-scroll outside the map, the selected range
was updated using the mouse down position. This causes the selection to
shrink and expand when the mouse leaves and enters the map area.

Change-Id: I307b9e30f8a06333a034f18914b1322146358512


vmiklos:
Could you please describe what are the steps to reproduce the problem you are fixing here? A list of steps and the old actual result and the new expected result would be helpful when I try to test your changes to see if they do what's intended.

Thanks!

JaumePujantell:
Select a range of cells by clicking and dragging from A2  to D2 and then to below outside the document.
The selected range keeps growing but, while the mouse is inside the document the selected range is A2:Dn, and once the mouse leaves the document it shrinks in columns to something like A2:An. 
It should always end in the D column if you are moving straight down.

vmiklos (approved):
Looks good to me know, thanks!

PR #10982, opened 2025-01-21
cypress focus unit test

hcvcastro:
check map object if it is not assigned

Change-Id: I467d4a6ed901151bda4e867088affe09a83bebb1

### DIFF
--- a/cypress_test/integration_tests/desktop/calc/jsdialog_spec.js
+++ b/cypress_test/integration_tests/desktop/calc/jsdialog_spec.js
@@ -36,6 +36,60 @@ describe(['tagdesktop'], 'JSDialog unit test', function() {
 			});
 	});
 
+	it('JSDialog child focus', function() {
+		cy.getFrameWindow().then(function(win) {
+			var smile = win.document.querySelector('meta[name="previewSmile"]').content;
+			var jsonDialog = {
+				id: 'testfocus',
+				type: 'dialog',
+				text: 'Focus test',
+				children: [{
+					id: 'tabcontrol',
+					type: 'tabcontrol',
+					selected: 1,
+					tabs: [{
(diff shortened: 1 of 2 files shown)


cypress_test/integration_tests/desktop/calc/jsdialog_spec.js
```diff
@@ -36,6 +36,60 @@ describe(['tagdesktop'], 'JSDialog unit test', function() {
 			});
 	});
 
+	it('JSDialog child focus', function() {
+		cy.getFrameWindow().then(function(win) {
+			var smile = win.document.querySelector('meta[name="previewSmile"]').content;
```
eszkadev, line 41: This is rather syntethic unit test - also fine, but can we do the integration test on real dialog from core?
(then if JSON will change format we don't have to change it here - we test behavior, not implementation)

I guess you tested that on some real dialog, right?

hcvcastro: This is a JSDialog class unit test, the issue is the focus in client side. If the JSON change it will change the implementation JSDialog builder too.

eszkadev: I see. In general cypress is integration test framework.
For unit tests we have `mocha`, see here: https://github.com/CollaboraOnline/online/tree/master/browser/mocha_tests

Maybe next time please try to look there (might have better fit for your needs and not require full app running).

eszkadev (approved):

cypress_test/integration_tests/desktop/calc/jsdialog_spec.js
hcvcastro: Thanks, I will look the mocha test, I did not notice.

PR #10983, opened 2025-01-22
tile: avoid rendering beyond document boundary

lpranam:
problem:
i.e: if calc sheet has most of row hidden and remaining rows don't cover the canvas area then
1. canvas center was incorrectly calculated based on available area instead the required doc area
2. grids were drawn beyond the last visible raw

Change-Id: I1f43e37e104c693b8f070271dac620385c0a2dd0

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Header.ts
+++ b/browser/src/control/Control.Header.ts
@@ -548,6 +548,7 @@ export class Header extends app.definitions.canvasSectionObject {
 			this._prevMouseOverEntry = this._mouseOverEntry;
 			this._mouseOverEntry = result.entry;
 		}
+		else return;
 
 		if (!this.containerObject.isDraggingSomething()) { // If we are not dragging anything.
 			this._dragDistance = null;
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -384,6 +384,103 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		}
 	},
 
+	syncTileContainerSize: function() {
+		if (!this._map) return;
+
+		var tileContainer = this._container;
+		if (tileContainer) {
+			// Document container size is up to date as of now.
+			var documentContainerSize = document.getElementById('document-container');
+			documentContainerSize = documentContainerSize.getBoundingClientRect();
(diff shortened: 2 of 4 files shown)


lpranam:
before: 
Screenshot_20250122_054530

after:
Screenshot_20250122_054659

gokaysatir (commented):
This is a really good approach to the issue.

I have 2 thoughts on this:
* Can you also shrink the size of the canvas when there aren't enough rows or columns to show? Map and canvas elements should have the same size. We are planning the remove map element but until then, we'll need to keep them aligned.
    Also, we need to check shape operations, text selections and the like, after shrinking the elements' sizes. Because the UI may not comply with the changing sizes.

* Indeed, maybe we can even play with the (id="document-container") element's size. Since map and canvas will probably resize accordingly, everything may "just work". We can do so by adding margins to the "document-container" element.

These are initial thoughts and of course need trying. But changing the size of the elements looks to me like a good approach.

Then i think we can get this in, in the red code.

Thank you.

gokaysatir:
By the way, this resizing method may be tricky when user zooms in/out the document. If this method proves to be not useful, we may need to work on SheetGeometry handling. My first thought was making the data handling better. That would require to check ColumnHeader.ts, RowHeader.ts and other related files. That way, we may not need to modify the map's and canvas's sizes.

eszkadev:
could you add cypress test too?
if only way is visual comparison: we have that, examples:  cypress_test/integration_tests/desktop/writer/jsdialog_widgets_spec.js

lpranam:
1. Map and canvas usually have different sizes in calc.
2. Tried "document-container" resizing but it did not work
3. Added a new commit to take care of zooming and syncing tiles in that case.

I'll added cypress too

gokaysatir:
Yes i remember now, canvas has column and row headers so that makes a small difference. But we still need to align the canvas's size. Map will be removed. It means this issue will regress when we remove it. I think we need to set also canvas's size. We are mirroring events from map to canvas. Dividing them would look like a workaround rather than a solution.

I also tested previous version and if you move mouse cursor where map doesn't exist but canvas does, row headers rise JS errors. This JS errors is not the main reason for this comment, main reason is canvas and map should stay coupled.

Thanks for checking the zoom issue :) I will check & test this again.

gokaysatir:
I think we can now abandon this commit.

PR #10986, opened 2025-01-22
browser: jsbuilder: replace button label for span element

hcvcastro:
Accessibility request to remove "label for" elements for button

Change-Id: Ief83bb73f6befbe39097de3e6aa550d58fd04de5

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -2094,11 +2094,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 			controls['button'] = button;
 			if (builder.options.noLabelsForUnoButtons !== true) {
-				var label = L.DomUtil.create('label', 'ui-content unolabel', button);
-				label.htmlFor = buttonId;
-				label.textContent = builder._cleanText(data.text);
-				builder._stressAccessKey(label, button.accessKey);
-				controls['label'] = label;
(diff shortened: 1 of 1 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -2303,11 +2303,10 @@ L.Control.JSDialogBuilder = L.Control.extend({
 
 			controls['button'] = button;
 			if (builder.options.noLabelsForUnoButtons !== true) {
-				var label = L.DomUtil.create('label', 'ui-content unolabel', button);
-				label.htmlFor = buttonId;
-				label.textContent = builder._cleanText(data.text);
-				builder._stressAccessKey(label, button.accessKey);
-				controls['label'] = label;
+				var span = L.DomUtil.create('span', 'ui-content unolabel', button);
```
eszkadev, line 2306: Are we sure we want to remove native "label" type?
It provides automatic information for text-to-speach software what to read and which input filed it labels

eszkadev: I see it's not for button element: https://www.w3schools.com/tags/tag_label.asp

eszkadev: but don't we need to add 'labelled-by' or 'labelFor' then?

hcvcastro: It does not specify the attribute change in the task description. Please ask Lilly for more details

PR #10987, opened 2025-01-22
Add null safety check for controlPosition initialization

Darshan-upadhyay1110:
- for undefined element we should add safe check here for controlPosition
- we are already showing error about absece of contolId element in `executeAction`` method (browser/src/control/Control.JSDialogBuilder.js#L2778)
- so no need to show JsError here.

Change-Id: I557e2c386408f0b9fe197fcf7a931152b31f9286

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
looks good, thanks for fix

PR #10988, opened 2025-01-22
SBOM (part #1)

timar:
This PR creates and installs SBOM for the online part of COOL based on a template, that we can extend in the future. For now, it has all dependencies coolwsd/coolforkit/etc, C++ binaries, and the Javascript UI of COOL.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,6 +57,7 @@ dist_doc_DATA = wsd/README.md \
                 wsd/README.vars \
                 wsd/protocol.txt \
                 wsd/metrics.txt \
+                collabora-online-sbom.spdx.json \
                 THIRDPARTYLICENSES
 
 coolwsddatadir = @COOLWSD_DATADIR@
@@ -462,6 +463,7 @@ dist-hook:
 
 EXTRA_DIST = discovery.xml \
              favicon.ico \
+             cool-sbom-template.spdx.json \
              coolkitconfig.xcu \
              coolwsd.spec \
              coolwsd.xml.in \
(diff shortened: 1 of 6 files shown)


Makefile.am
```diff
@@ -808,3 +812,6 @@ coverage-clean:
 
 czech: check
 	@echo "This should do something much cooler"
+
+collabora-online-sbom.spdx.json: coolforkit-ns
```
vmiklos, line 816: Nit: also depend on the python script to help re-running this when you touch it?

cool-sbom-template.spdx.json
```diff
@@ -0,0 +1,369 @@
...
+    "creators": [
+      "Tool: Custom Script"
+    ],
+    "created": "2024-12-15T12:00:00Z"
+  },
+  "packages": [
+    {
+      "SPDXID": "SPDXRef-coolwsd",
+      "name": "coolwsd",
+      "versionInfo": "24.04.10.3",
```
vmiklos, line 17: Nit: you process this template with a script, so I assume this value is a placeholder, but then it would be useful to set it to something like @VERSION@ to make it clear it will be replaced.

cool-sbom-template.spdx.json
```diff
@@ -0,0 +1,369 @@
...
+          "referenceCategory": "SECURITY",
+          "referenceType": "vulnerability",
+          "referenceLocator": "https://github.com/CollaboraOnline/online/security/advisories"
+        }
+      ]
+    },
+    {
+      "SPDXID": "SPDXRef-openssl",
+      "name": "OpenSSL",
+      "versionInfo": "3.0.15",
```
vmiklos, line 32: And same for the other libs we link statically.

vmiklos (approved):
Looks OK in general, see inline comments for some nits.

PR #10990, opened 2025-01-22
treeview: handle custom previews

eszkadev:
This enabled rendering of the previews inside style sidebar.
Check format -> Styles list, in the sidebar you can watch style before use.
In dark mode we want it to be as on a printed paper - white background (real colors).

Snapshot_2025-07-09_16-21-34


eszkadev:
it still might be blurry, HiDPI improvements are planned as followup

pedropintosilva (changes requested):
It looks good and works but the context menu is now badly positioned :  
image

eszkadev:
> It looks good and works but the context menu is now badly positioned : image

done :)

pedropintosilva (approved):

PR #10991, opened 2025-01-22
jsdialog: fix findEntryWithRow in TreeView

eszkadev:
We need to check row property, not index in the array
Also  setup correct state initially and keep it if we
just expand row. We want to change state of checkbox
only on checkbox click.

JSON looks like:
0: {entry... row: 0}
       - 0: {child1... row: 15}
       - 1: {child2... row: 200}
1: {entry.... row: 1}
2: {entry.... row: 2}

So inside tree we would not find anything as we compare just with index


eszkadev:
correct tree on initial load of sample from test is:
tree

vmiklos (commented):
I see nothing obviously wrong with this, but it's not really an area, so in general I would defer to Tomaz.

quikee (approved):
Looks fine to me

PR #10992, opened 2025-01-23
cool#10630 doc electronic sign: fix no graphic selection on 2nd page

vmiklos:
Open a 2 page PDF, scroll to page 2, Insert -> Signature, the graphic
selection doesn't appear for the signature line.

Seems the graphic selection always assumed a single part, but the text
selection doesn't have that problem since commit
a9c4aca4f8be5b3943cef777ac306aca99d78bbb (PDF View: Update search
feature to make it work with PDF view., 2021-07-27).

GraphicSelection.onMessage() seems to be the place where we parse the
graphic selection, so shift down the graphic selection there, as early
as possible, so nobody can have the info without the page offset.

The SVG inside the signature line & the drag of it is still broken.

Change-Id: Ia699c80ac94b9541fde2ea1ad1642e6150307cf7

### DIFF
--- a/browser/src/app/GraphicSelectionMiddleware.ts
+++ b/browser/src/app/GraphicSelectionMiddleware.ts
@@ -149,6 +149,35 @@ class GraphicSelection {
 			this.convertObjectRectangleTwipsToPixels();
 	}
 
+	/// Push down the graphic selection on non-first pages of scrolling PDF view.
+	static transformGraphicSelection(messageJSON: any) {
+		if (!app.file.fileBasedView) {
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


vmiklos:

It fixes the graphic selection handles on non-first PDF pages; though I'm aware of 2 more related problems, I'm still looking at those. Likely a similar style of fix will be needed there as well, "just" need to find the right place. :-)

caolanm (approved):
Can this get called in a typical calc session? Is there a chance that a calc graphic selection, if we are on say, tab 3, gets misadjusted vertically?

PR #11003, opened 2025-01-24
UI log: small fixes, and extra addons

aszucs3:
delete ?params from the end of uno commands, before logging The script delete them too... to be sure... (if we have old log)

Added char/sec type speed to statistics.
Fixed a bug with uno command connection... (a list was not deleted) found a problem: some cases undo-count-change not happens therefore some undo may not connectable to an undo-count-change made a new "unknown" command .. and count these wrong cases.

Made other statistics a bit better:
-added 45 degree rotattion
-added logarithmical scaling to uno commands
(maybe soem other charts will need it too)
-wrote text to A1 ..and fixed the code to not use that for charts -added new groups: "Single viewer", "Single Editor"

Change-Id: I6bfa40e3505971bd6fb7f77da618f76410ebf87a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -3946,6 +3946,10 @@ LogUiCommands::~LogUiCommands()
                     return;
                 actCmd = (*_tokens)[0];
                 actSubCmd = (*_tokens)[1];
+                std::size_t pos = actSubCmd.find_first_of ('?');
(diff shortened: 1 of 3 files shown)


caolanm (approved):
looks good to me

PR #11004, opened 2025-01-24
comment: autosave only when entire comment container loses focus

lpranam:
avoid triggering autosave when textbox loses focus so user can navigate buttons using tab

Change-Id: Ibe0348ab7160c383f40e97328c5d844921a461c3

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -151,8 +151,6 @@ export class Comment extends CanvasSectionObject {
 			return;
 
 		var button = L.DomUtil.create('div', 'annotation-btns-container', this.sectionProperties.nodeModify);
-		L.DomEvent.on(this.sectionProperties.nodeModifyText, 'blur', this.onLostFocus, this);
-		L.DomEvent.on(this.sectionProperties.nodeReplyText, 'blur', this.onLostFocusReply, this);
 		L.DomEvent.on(this.sectionProperties.nodeModifyText, 'input', this.textAreaInput, this);
 		L.DomEvent.on(this.sectionProperties.nodeReplyText, 'input', this.textAreaInput, this);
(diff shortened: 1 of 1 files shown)


browser/src/canvas/sections/CommentSection.ts
```diff
@@ -1115,6 +1114,14 @@ export class Comment extends CanvasSectionObject {
 
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 	public onLostFocus (e: any): void {
+
+		if (!this.isEdit() || this.sectionProperties.container.contains(e.relatedTarget))
+			return;
+		if (this.sectionProperties.nodeReply.contains(e.srcElement)) {
```
eszkadev, line 1120: https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement
`Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards`

is that the same as `e.relatedTarget` ?

eszkadev: can we add comment above this function documenting why, when it is called and what it does ?
so next person will not have to read the wall of text

eszkadev: and for `onLostFocusReply` the same

eszkadev: or better: in both onLostFocus and onLostFocusReply move duplicated chunk of code which is  hard to read - so it will get name and explain itself :)

eszkadev (approved):

eszkadev:
I found bug:
- start adding comment,
- scroll so it is invisible,
- switch to other tab,
- switch back to COOL,
- start typing
result: UI broken
expected: scroll document to the comment

[comment-2025-01-27_13.46.09.webm](https://github.com/user-attachments/assets/e8aee0eb-610a-473a-8d7b-265e683e8ef1)

PR #11006, opened 2025-01-24
Update document and presentation samples

thebearon:
Change-Id: I54e0424fd59b8d6fab61280390c9f687d6fc343a

* Target version: master 

### Checklist


timar (approved):
Looks good to me at first glance. You did not add a changelog, but considering the time you spent on this, I'm pretty sure that these templates are much better then their previous versions.

PR #11007, opened 2025-01-24
cool#11002 kit: fix memory corruption in Document::handleSaveMessage()

vmiklos:
- **wsd: fix clang-15 build**
- **cool#11002 kit: fix memory corruption in Document::handleSaveMessage()**

### DIFF
--- a/wsd/CacheUtil.hpp
+++ b/wsd/CacheUtil.hpp
@@ -23,6 +23,13 @@ struct CacheQuery
     std::string _uri;
     std::string _stamp;
     std::string _dest;
+
+    CacheQuery(const std::string& uri, const std::string& stamp, const std::string& dest)
+        : _uri(uri)
+        , _stamp(stamp)
+        , _dest(dest)
+    {
+    }
 };
 
 class Cache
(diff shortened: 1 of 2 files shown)


vmiklos:

We had the same patter in core unit tests in the past: given that "on idle" jobs may send LOK callbacks, it's useful to unregister the callback function before we delete the COOL side. I wasn't entirely happy with the 4x sfx2 band-aid fixes, I think this is finally fixing a root cause.

I promise this is my last review request for today. :-)

wsd/CacheUtil.hpp
```diff
@@ -23,6 +23,13 @@ struct CacheQuery
     std::string _uri;
     std::string _stamp;
     std::string _dest;
+
+    CacheQuery(const std::string& uri, const std::string stamp, const std::string dest)
```
caolanm, line 27: would it be better to be all "const std::string&" here?

caolanm (approved):
This looks reasonable to me.

wsd/CacheUtil.hpp
vmiklos: Ah yes, let me fix. I just happened to have clang-15 here and after pulling, I ran into this where old enough compilers need an explicit ctor for some reason.

PR #11008, opened 2025-01-24
mocha: store the compiled js files in a separate folder away from the typescript files

dennisfrancis:
Store the compiled mocha js files in a separate folder away from the source typescript files.

Change-Id: Ibc1ecd263f53b2e6db736d9c6f1330998d4e2bf8

* Target version: master 

### Summary
Store the compiled mocha js files in a separate folder away from the source typescript files.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/.gitignore
+++ b/browser/.gitignore
@@ -9,7 +9,7 @@ bower.json
 component.json
 
 mocha.log
-mocha_tests/*.js
+mocha_tests/workdir/
 tscompile.done
 duplication.log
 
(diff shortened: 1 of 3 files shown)


vmiklos (commented):
For cypress tests, such generated files are under cypress_test/workdir/, would it make sense to go with the same "workdir" name here for consistency?

Ignoring a dir named "js" is a bit strange when some of our actual code is also in JS. What do you think?

vmiklos (approved):
Looks good to me now, thanks!

Nit: in general, feel free to ask such code reviews from @quikee , we try to distribute the workload . :-)

PR #11010, opened 2025-01-24
wsd: presets: browsersetting is sent as a file not in userconfig json…

Rash419:
… itself

Change-Id: If32b33e74612d31abde107f4b568adb444d7d2a6

* Target version: master


caolanm (approved):
looks good to me if it passes ci

PR #11012, opened 2025-01-24
iframe: improve admin settings iframe ui

codewithvk:
Done
	- Refactor entier iframe ui 
	- accept css vars from integrator
	- fix button upload problem with uploadWordbookButton

Todo: 
	- append branding css
    	- try to replace svg to css svg if we have icons?

debug mode on ss: 
nextcloud local_index php_settings_user_richdocuments (3)

Change-Id: Ifbd34567ea716d334e2b277eba1abc255c523f69

### DIFF
--- a/browser/admin/adminIntegratorSettings.html
+++ b/browser/admin/adminIntegratorSettings.html
@@ -1,35 +1,74 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-
-<head>
+  <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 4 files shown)


pedropintosilva (approved):
This seems ok, would be probably good to have another +1 but it seems quite safe (since outside of the feature it doesn't affect the normal state of things)

caolanm (approved):

pedropintosilva:
all checks passed with the exception of mobile cypress tests. These are completely unrelated to this PR and moreover these changes do not impact cypress mobile tests in any way. Also, I have notice one of those cypress flaky mobile tests has already failed : `writer/annotation_spec.js`, 

With all that said, I'm merging this now

PR #11014, opened 2025-01-24
Calc: Implement column row highlight feature

GulsahKose:
Change-Id: I985b2fa7e4f59c44ca8cba00666959ff8c18d236

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
I commented in the core patch. It should be done more on online side (core only sends on/off).
We draw here the highlight. @gokaysatir could give some pointers how to best implement that.

I guess something from this folder would help: https://github.com/CollaboraOnline/online/blob/master/browser/src/canvas/sections/OtherViewCellCursorSection.ts

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2065,6 +2065,61 @@ L.CanvasTileLayer = L.Layer.extend({
 		}
 	},
 
+    _highlightColAndRow: function (textMsg, part) {
+        this._resetReferencesMarks();
+        var strTwips = textMsg.match(/\d+/g);
+        var references = [];
+        this._referencesAll = [];
+        var rectangles = [];
+        var strColor = 'ccffff';
```
eszkadev, line 2074: would be good to get css value from variable so we can theme it, @pedropintosilva can help which one to use
example: https://github.com/CollaboraOnline/online/blob/e19aa5c9c3f944cf6b3a21c03bbff183d6a1242f/browser/src/map/handler/Map.Feedback.js#L101

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -1045,6 +1049,10 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this._update();
 			this.enableDrawing();
 		}
+        if(this.isHighlightEnabled)
+            L.CanvasTileLayer.prototype._highlightColAndRow.call(this, textMsg, this._selectedPart);
```
eszkadev, line 1053: why not `this._highlightColAndRow(textMsg)` ?
`this._selectedPart` doesn't have to be passed as you can read it inside

eszkadev:
Does that work after we used reference marks for formula selection?
- activate highlight and test if it works
- edit some cell, write `=SUM(`, now select few cells (we will use reference marks)
- after formula was created, accept the cell value, go to some other cells, test if highlighting still works

I fear we need some update after we switch sheet, change cell selected (but I only looked at code)

browser/src/layer/tile/CanvasTileLayer.js
GulsahKose: Done

GulsahKose:
[Screencast from 2025-02-03 16-50-56.webm](https://github.com/user-attachments/assets/6f6626d5-46c0-434d-bf64-dbbeb97542ef)

browser/src/layer/tile/CalcTileLayer.js
eszkadev: can't we do this?

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -878,6 +879,9 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 				}
 			}
 		}
+        else if (e.commandName === '.uno:ViewColumnRowHighlighting') {
```
eszkadev, line 882: maybe we can use online-only setting for that as Gokay noticed?

example with user preferences which can be remembered in the browser: https://github.com/CollaboraOnline/online/blob/875c3b0e81f2f16d639d07051f8f6f570b38ae13/browser/src/map/handler/Map.Feedback.js#L48

GulsahKose:
@eszkadev Now core patch is discarded. We can keep the status on browser. But Toggled view is not working now. Is the only was editing css?. Is there any better way to keep button toggled?

eszkadev:
we can also load old value on app load, this can be done in UIManager, similar to other things: https://github.com/CollaboraOnline/online/blob/25dfda07cf5414923501e51b3027f422fd2dc315/browser/src/control/Control.UIManager.js#L83

browser/src/control/Control.UIManager.js
```diff
@@ -376,6 +384,9 @@ L.Control.UIManager = L.Control.extend({
 			// remove unused elements
 			L.DomUtil.remove(L.DomUtil.get('presentation-controls-wrapper'));
 			document.getElementById('selectbackground').parentNode.removeChild(document.getElementById('selectbackground'));
+
+			if (this.getHighlightMode()) $('#colrowhighlight').addClass('selected');
+			else $('#colrowhighlight').removeClass('selected');
```
GulsahKose, line 389: @eszkadev Here i planned to remember toggle state for the case on notebookbar
enable highlight
refresh the document
expecting to see button toggled on
actually toggled off.

What is wrong here?

browser/src/layer/tile/CalcTileLayer.js
eszkadev: line can be simplified:
`L.CanvasTileLayer.prototype._highlightColAndRow.call(this, textMsg, this._selectedPart);`
to
`this._highlightColAndRow(this, textMsg);`

and inside `_highlightColAndRow` function you can get your `this._selectedPart` without passing as argument

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -1045,6 +1045,10 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this._update();
 			this.enableDrawing();
 		}
+        if (this._map.uiManager.getHighlightMode())
+            L.CanvasTileLayer.prototype._highlightColAndRow.call(this, textMsg, this._selectedPart);
+        else
+            L.CanvasTileLayer.prototype._resetReferencesMarks.call(this);
```
eszkadev, line 1051: also `this._resetReferenceMarks()`

browser/src/control/Control.UIManager.js
eszkadev: `map['stateChangeHandler'].getItemValue('YOUR COMMAND')` should get the correct value from preferences saved in the browser.

because toolbutton use it to fetch the state. command must match what you have defined for the button's `command` in JSON.

see: https://github.com/CollaboraOnline/online/blob/25dfda07cf5414923501e51b3027f422fd2dc315/browser/src/control/Control.JSDialogBuilder.js#L2343

browser/src/docdispatcher.ts
```diff
@@ -456,6 +456,12 @@ class Dispatcher {
 			// https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft
 			L.DomUtil.get('spreadsheet-tab-scroll').scrollLeft = 100000;
 		};
+		this.actionsMap['.uno:ViewColumnRowHighlighting'] = function () {
+			var newState = !window.prefs.getBoolean('ColumnRowHighlightEnabled');
+			window.prefs.set('ColumnRowHighlightEnabled', newState);
+			if (newState) $('#colrowhighlight').addClass('selected');
+			else $('#colrowhighlight').removeClass('selected');
```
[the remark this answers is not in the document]
eszkadev: here we need to save browser preference: `window.prefs...`

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2065,6 +2065,61 @@ L.CanvasTileLayer = L.Layer.extend({
 		}
 	},
 
+    _highlightColAndRow: function (textMsg, part) {
```
eszkadev, line 2068: this function is needed only in Calc, please move to `CalcTileLayer` so we can do simplification I mentioned in `CanvasTileLayer`

browser/src/docdispatcher.ts
```diff
@@ -456,6 +456,12 @@ class Dispatcher {
 			// https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft
 			L.DomUtil.get('spreadsheet-tab-scroll').scrollLeft = 100000;
 		};
+		this.actionsMap['columnrowhighlight'] = function () {
+			var newState = !app.map.uiManager.getHighlightMode();
+			app.map.uiManager.setHighlightMode(newState);
+			if (newState) $('#colrowhighlight').addClass('selected');
```
eszkadev, line 462: these 2 lines shouldn't be needed (adding css class)
setHighlightMode should fire commandstatechange which will inform button about state change

eszkadev (approved):
code looks good :) thanks

eszkadev (approved):

PR #11015, opened 2025-01-24
dev_notes: added notes for locale decimal separators

lpranam:
Change-Id: I69bdb5de8a51ef04e15dcb3696a48855472b19b8

* Target version: master 

### Checklist


caolanm:
Looks like the sort of thing we want, but move it a dir higher I think, stick "dev-notes" into the root dir. If you think there should be a browser subdir, then dev-notes/browser/whatever, but just dev-notes/locale.md is probably fine. If we have lots of stuff, then we can put into subdirs of dev-notes afterwards I support.

caolanm (approved):
looks good, I think we can skip ci for docs

PR #11017, opened 2025-01-24
Fix Presenter Console gets close in case of repeat slideshow after x sec

Darshan-upadhyay1110:
- handled the case where we do enable the repeat slide show & PC should not get closed
- isEndless will make sure we do not terminate the window

Change-Id: I6e7f5bb11e784195acc2e8dbb4918875c829e7f3

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/slideshow/PresenterConsole.js
+++ b/browser/src/slideshow/PresenterConsole.js
@@ -788,7 +788,8 @@ class PresenterConsole {
 			}
 			case 'next': {
 				this._presenter.getNavigator().dispatchEffect();
-				if (isLastSlide) {
+				// if repeat after sec is set then do not close on last slide
+				if (isLastSlide && !this._presenter._presentationInfo.isEndless) {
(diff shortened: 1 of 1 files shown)


browser/src/slideshow/PresenterConsole.js
```diff
@@ -788,7 +791,8 @@ class PresenterConsole {
 			}
 			case 'next': {
 				this._presenter.getNavigator().dispatchEffect();
-				if (isLastSlide) {
+				// if repeat after sec is set then do not close on last slide
+				if (isLastSlide && !this._presenterInfo.isEndless) {
```
eszkadev, line 795: Can't we  use `this._presenter._presentationInfo.isEndless`   here?
No sense to add new memeber variable for single use if can be avoided.
It adds possibility to go out of sync with 2 different instances of info when we change `_presenter` in the menatime for example

PR #11018, opened 2025-01-25
doc_memory_used_bytes should be in bytes not in kilobytes (fixes #10984)

### LINKED ISSUE #10984: Prometheus exporter: exported per-document metrics have wrong units
### Describe the Bug

coolwsd has a built-in exporter for Prometheus metrics. I have discovered two issues with the per-document metrics (`doc_memory_used_bytes`) of this exporter:

- The unit naming is wrong. The name and [documentation](https://github.com/CollaboraOnline/online/blob/master/wsd/metrics.txt) state that `doc_memory_used_bytes` is "bytes of memory". However in practice the metric appears to be kilobytes.
- Even when assuming that `doc_memory_used_bytes` is kilobytes, the value is slightly off compared to the value displayed in the admin console. It turns out that if you divide `doc_memory_used_bytes` by 1000 it matches the value from the admin console. So either the admin console is using "1000" as a divisor, where "1024" should be used or it's the Prometheus metrics which are off.

### Steps to Reproduce
1. Open any document and keep it open. Don't touch it anymore.
2. Open the admin console (http://coolwsd-host.domain/browser/dist/admin/admin.html)
3. Open the metrics (http://coolwsd-host.domain/cool/getMetrics)
4. Compare `doc_memory_used_bytes`  from the metrics with the value displayed in the admin console

Values for an example document:

* Value reported by admin console: **18.8 MB**	
* Value reported by  `doc_memory_used_bytes` metric: **18780**

### Expected Behavior

* The `doc_memory_used_bytes` should actually be bytes, as documented

### Actual Behavior

* The `doc_memory_used_bytes` is not bytes. It is Kilobytes and it still does not match the value displayed on the admin console.

timar:
Change-Id: I562a7eacdb04bae7745c5f92502c21ef88bf9627

* Resolves: #10984
* Target version: master

### DIFF
--- a/browser/admin/src/Util.js
+++ b/browser/admin/src/Util.js
@@ -22,7 +22,7 @@ var Util = Base.extend({
 }, { // class interface
 
 	humanizeMem: function (kbytes) {
-		var unit = 1000;
+		var unit = 1024;
 		var units = [_('kB'), _('MB'), _('GB'), _('TB'), _('PB'), _('EB'), _('ZB'), _('YB'), _('BB')];
 		for (var i = 0; Math.abs(kbytes) >= unit && i < units.length; i++) {
 			kbytes /= unit;
--- a/wsd/AdminModel.cpp
+++ b/wsd/AdminModel.cpp
@@ -1251,7 +1251,7 @@ void AdminModel::getMetrics(std::ostringstream &oss)
         oss << "doc_views" << suffix << doc.getViews().size() << "\n";
         oss << "doc_views_active" << suffix << doc.getActiveViews() << "\n";
         oss << "doc_is_modified" << suffix << doc.getModifiedStatus() << "\n";
-        oss << "doc_memory_used_bytes" << suffix << doc.getMemoryDirty() << "\n";
+        oss << "doc_memory_used_bytes" << suffix << doc.getMemoryDirty() * 1024 << "\n";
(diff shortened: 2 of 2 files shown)


browser/admin/src/Util.js
```diff
@@ -22,7 +22,7 @@ var Util = Base.extend({
 }, { // class interface
 
 	humanizeMem: function (kbytes) {
-		var unit = 1000;
+		var unit = 1024;
 		var units = [_('kB'), _('MB'), _('GB'), _('TB'), _('PB'), _('EB'), _('ZB'), _('YB'), _('BB')];
```
vmiklos, line 26: I'm always a bit confused here, but isn't it like "KB" is 1000 as a multiplier and "KiB" is 1024 as a multiplier? See the table at https://en.wikipedia.org/wiki/Gigabyte

If I'm right, then 1000 was correct here.

vmiklos: OK, let's be consistent an use 1024 everywhere. If people complain, we can reiterate.

vmiklos (approved):

browser/admin/src/Util.js
klada: I think people expect memory sizes in `MiB` and `GiB`, as RAM memory is marketed with binary units (in contrast to other computer memory). It's also what most operating system GUIs use for "file size".

So IMHO `1024` is the correct divisor here. To be more precice, `units` could be changed to `KiB`, `MiB`, ...

PR #11025, opened 2025-01-27
wsd: fix assert fail in DocumentBroker::getJailRoot()

vmiklos:
This can happen if the websocket incoming traffic has a load command,
followed by a removesession command, so asserting a non-empty jail ID
seems to be going too far.

Backtrace:

    #9 0x555e36ada9b8 in DocumentBroker::getJailRoot[abi:cxx11]() const wsd/DocumentBroker.cpp:3646:5
    #10 0x555e36b49ae3 in DocumentBroker::uploadPresetsToWopiHost(Authorization const&) wsd/DocumentBroker.cpp:4094:41
    #11 0x555e36b444c9 in DocumentBroker::removeSession(std::shared_ptr<ClientSession> const&) wsd/DocumentBroker.cpp:3772:9
    #12 0x555e369bb709 in ClientSession::_handleInput(char const*, int) ClientSession.cpp:1088:24
    #13 0x555e3703957b in Session::handleMessage(std::vector<char, std::allocator<char>> const&) common/Session.cpp

Change-Id: Ib9cfd4856838dfa6ba304888770898964e922260


vmiklos:

Found by a fuzzer, we assumed that a doc is always loaded (downloaded) before the session removal message arrives, but the added sample shows this is not necessarily the case. I assume it's OK to downgrade this assert to a warning, at least the caller in the backtrace just iterates the files of the return dir, so it copes with the empty result. The assert was there for a long time, but the new ~autotext code now uses it more, I guess that's how this was found just now.

caolanm (approved):
I imagine this is just a matter of the fuzzer finally hitting on this, rather than a specific recent change that allows hitting this path

PR #11027, opened 2025-01-27
fix: custom logo doesn't persist sometimes

Rash419:
- when UI is refreshed old html is deleted which consist of overwritten css background-image
- this patch makes sure to update document header after '#document-header' is re-created

Change-Id: I6b2723ada4319f9139be8ff8d29b463401540969

### DIFF
--- a/browser/src/control/Control.UIManager.js
+++ b/browser/src/control/Control.UIManager.js
@@ -410,9 +410,7 @@ L.Control.UIManager = L.Control.extend({
 
 		this.map.on('changeuimode', this.onChangeUIMode, this);
 
-		if (typeof window.initializedUI === 'function') {
-			window.initializedUI();
-		}
+		this.refreshTheme();
 
 		var startPresentationGet = this.map.isPresentationOrDrawing() && window.coolParams.get('startPresentation');
 		// check for "presentation" dispatch event only after document gets fully loaded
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.UIManager.js
```diff
@@ -1059,6 +1059,10 @@ L.Control.UIManager = L.Control.extend({
 			this.refreshNotebookbar();
 		else
 			this.refreshMenubar();
+
+		if (typeof window.initializedUI === 'function') {
+			window.initializedUI();
```
eszkadev, line 1064: I see it is repeated in Socket.js in reconnecting. Can we put it in a method here and share (name also will tell others what it does). Like `refreshTheme()` to be unified with other `refresh` name usage .

eszkadev: also repeated https://github.com/CollaboraOnline/online/blob/3e8375319361f68d08bd4089536030719183b5fb/browser/src/control/Control.UIManager.js#L645

eszkadev: and https://github.com/CollaboraOnline/online/blob/3e8375319361f68d08bd4089536030719183b5fb/browser/src/control/Control.UIManager.js#L414

eszkadev (approved):
looks good :)

PR #11028, opened 2025-01-27
Drop the config url query params from identifier

caolanm:
Unescaped and used in subforkit announce they get parsed as additional arguments. Escaped, they end up doubly unescaped elsewhere. Just drop them entirely and disambiguate where necessary.

Change-Id: Ia57f14f7cb3aa7c4f77f2dc536996bdcaa38db78

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
LGTM, I haven't tested

PR #11029, opened 2025-01-27
Add xml parameters brandProductName and brandproductURL

meven:
In user_interface section.

Change-Id: If5033a3221de985cf82eace8dc7c28fec220c419

* Resolves: # 
* Target version: master 

### Summary

This allows to set a product name and product url.

The option is visible only in COOL.

### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -989,6 +989,7 @@ WELCOME_CONFIG_FRAGMENT=
 HOME_MODE_CONFIGURATION=
 ENABLE_WELCOME_MESSAGE=0
 WELCOME_URL=
+CONFIG_INTERFACE_FRAGMENT=
 if test -n "$with_welcome_url"; then
     WELCOME_URL=$with_welcome_url
     ENABLE_WELCOME_MESSAGE=1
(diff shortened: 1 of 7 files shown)


mmeeks (commented):
Looks reasonable to me I'd prefer @timar to review though. Thanks.

configure.ac
```diff
@@ -999,12 +1000,16 @@ else
     WELCOME_CONFIG_FRAGMENT="<welcome>
       <enable type=\"bool\" desc=\"Controls whether the welcome screen should be shown to the users on new install and updates. When enabled, your custom welcome screen must be created at /usr/share/coolwsd/browser/dist/welcome/welcome.html.\" default=\"false\">false</enable>
     </welcome>"
+    CONFIG_INTERFACE_FRAGMENT="<brandProductName desc=\"Change the displayed name the application will have for the user.\" type=\"string\"></brandProductName>
+    <brandProductURL desck=\"Change the url of the , empty makes disable the link\" type=\"string\">https://www.collaboraonline.com</brandProductURL>"
```
timar, line 1004: typo: `desck`
Incomplete sentence in description

timar (changes requested):
It will not work with brand package, because settings in branding.js will overwrite the settings in global.js. You have to put them into a branding_override.js, and load it after branding.js, but only if the product name and URL are actually overridden. So the settings could be `brandProductNameOverride` and `brandProductURLOverride` to begin with, and their default value should be an empty string.
When you add a new config setting to coolwsd.xml, please add the default value to DefAppConfig map.

configure.ac
meven: > It will not work with brand package, because settings in branding.js will overwrite the settings in global.js.

It is meant to be used when no branding is set.

> branding_override.js, and load it after branding.js

I'd like to avoid adding a new request, plus this file has to be dynamic since it depends on settings, so that would adding some file generation logic and serving. It seems to me way too complicated for what the feature I am adding.

Having a different name with Override postfix would make things compatible, but do we really need it ?

timar (approved):

timar:
It is OK with the accompanying change in branding repo. It is OK the merge in green CODE, because it has no effect in CODE. We have to be careful, and test it in CODE once again, before release.

PR #11031, opened 2025-01-27
Add new sections to setting iframe

codewithvk:
* Resolves: # 
* Target version: master 

### Summary
- Generate a configuration section using JavaScript to reduce code repetition.
- Add a browser settings section with the following conditions: 
  - Only one file can be uploaded. If a file is already uploaded, the upload button should be disabled.
  - Applicable only for user configurations.
- Add an XCU section similar to browser settings:
  - Only one file can be uploaded at a time.

nextcloud local_index php_settings_user_richdocuments (5)
nextcloud local_index php_settings_user_richdocuments (4)

### DIFF
--- a/browser/admin/adminIntegratorSettings.html
+++ b/browser/admin/adminIntegratorSettings.html
@@ -28,36 +28,7 @@
   </head>
   <body>
     <div id="fileControls">
-      <div id="sharedConfigs">
-        <div class="section">
-          <h3>Autotext</h3>
-          <ul id="autotextList"></ul>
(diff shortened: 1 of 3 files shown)


Rash419 (approved):
looks good :)

caolanm (approved):
xcu is "Registry Update Format" https://people.freedesktop.org/~vmiklos/2013/oor-document-format.html the libreoffice file format used to configure the options of the kit, I don't know if it would be helpful to use some string like that, but I think we can merge this now.

PR #11035, opened 2025-01-28
Hide scaling handles of shapes when text cursor is visible.

gokaysatir:
Issue:
* Open an Impress file and start editing a textbox.
* See that handles are visible.
* Try to re-scale the selected shape.
* It doesn't work but the preview is shown.
* When text editing is active, scaling is not enabled.
* We need to hide the handles in that case.

Change-Id: I0ebf20270fc08f8553c1985456836711369407e4

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir:
Tests are passing on local. Rebased.

eszkadev (approved):

PR #11036, opened 2025-01-28
UX iprovements on server audit dialog

banobepascal:
Change-Id: I6c91fb915d6abd6dfd93536fdde6a2835091b6f5

* Improves: https://github.com/CollaboraOnline/online/pull/10925#issuecomment-2601894287
* Target version: master 

### Summary
- Added new stylings to the indicator to make it distinct
- Added both cases for no issues and errors plus count of issues
- Added new check-mark icon for audit success

### Screenshots
Screenshot from 2025-01-28 09-58-39
Screenshot from 2025-01-28 09-58-21

With no Issues
Screenshot from 2025-01-28 13-52-51
Screenshot from 2025-01-28 14-18-06

### Checklist

### DIFF
--- a/browser/src/control/Control.ServerAuditDialog.ts
+++ b/browser/src/control/Control.ServerAuditDialog.ts
@@ -176,8 +176,18 @@ class ServerAuditDialog {
 		return hasErrors;
 	}
 
+	private countErrors(): number {
+		return (
+			(app.serverAudit?.filter((entry: AuditEntry) => entry.status !== 'ok')
+				.length ?? 0) +
+			(app.clientAudit?.filter((entry: AuditEntry) => entry.status !== 'ok')
+				.length ?? 0)
+		);
+	}
+
 	private getJSON(entries: Array<any>): JSDialogJSON {
 		const hasErrors = this.hasErrors();
+		const countErrors = this.countErrors();
 
 		return {
 			id: this.id,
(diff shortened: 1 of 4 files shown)


browser/src/control/Control.ServerAuditDialog.ts
```diff
@@ -215,6 +218,17 @@ class ServerAuditDialog {
...
+						!hasErrors
+							? {
+									id: 'auditsuccess',
+									type: 'fixedtext',
+									text: _('No issues found'),
+								}
+							: {
+									id: 'auditerror',
+									type: 'fixedtext',
+									text: `${countErrors} issue${countErrors > 1 ? 's' : ''} found`,
```
eszkadev, line 230: We need translatable string, we do that by using special wrapping function: `_('string to translate')`.
It has to be static text, any dynamic content should be provided by replacement of a variable mark like eg. `%COUNT%`.
If you need both single and plural form, please use 2 separate strings, not combine that in JS as it will not work with other languages.

example:
```
if (count === 1)
      return _('1 issue found');
else
      return _('%COUNT issues found').replaceAll('%COUNT%', number);
```

I'm not sure if it is exactly %var% maybe different style, please check within a code for examples

banobepascal: Please checkout the new changes

banobepascal:

pedropintosilva (approved):
I didn't see any functional problem and since this introduces a new string I would like to see it merge so, I'm approving it. However I still don't think the following is good:

image

* The indicator protrudes from the UI and it might look like a button
* There is border on border situation going on there
* I'm not sure about using an x that is white

By making it a bit more subtle it would be already an improvement :

image

Still, probably improving the icon would be a welcome change

banobepascal:
New Improvements
Screenshot from 2025-02-05 01-03-00

Screenshot from 2025-02-05 00-58-14


Screenshot from 2025-02-05 01-04-33

Screenshot from 2025-02-05 01-05-20

PR #11038, opened 2025-01-28
fix: ensure documentAnchor is set in onResize

Minion3665:
in I6df1f36d6100857373686c6b57daaa805917df02, we early-exit from onResize to avoid it being called unnecessarily. Unfortunately, it also sets `documentAnchor`, which we missed when checking whether we needed to run through it. This caused an error on mobile web, leading to a failure to open documents in, say, the Nextcloud app.

Change-Id: I4938ade7f03be95b0f0ff3edd4e5805302f22036

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Cwiiis (approved):
Looks good to me - it'd be great if this was something that was guaranteed to be set whenever reNewAllSections was called, but this is a pretty benign change in the grand scheme of things :)

PR #11041, opened 2025-01-29
wsd: cosmetic: remove unnecessary todos

Rash419:
Change-Id: Ib0ec66b257dd173fd3b6279b2fa8ae5d47afde56

* Resolves: # 
* Target version: master

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -744,7 +744,6 @@ bool FileServerRequestHandler::isAdminLoggedIn(const HTTPRequest& request, http:
             std::string filePath;
             for (const auto& param : params)
             {
-                // TODO: replace fileId with filePath
                 if (param.first == "fileId")
                     filePath = param.second;
             }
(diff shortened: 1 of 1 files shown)


wsd/FileServer.cpp
```diff
@@ -2306,11 +2301,8 @@ void FileServerRequestHandler::preprocessIntegratorAdminFile(const HTTPRequest&
     csp.appendDirective("frame-src", "'self'");
     csp.appendDirective("frame-src", "blob:"); // Equivalent to unsafe-eval!
     csp.appendDirective("connect-src", "'self'");
-    // TODO: we have inline script tag and css in admin panel. can't have following CSPs
-    // csp.appendDirective("default-src", "'none'");
-    // csp.appendDirective("script-src", "'self'");
-    // csp.appendDirective("script-src", "'unsafe-eval'");
-    // csp.appendDirective("style-src", "'self'");
+    csp.appendDirective("script-src", "'self'");
+    csp.appendDirective("script-src", "'unsafe-eval'");
```
caolanm, line 2305: if this was commented out before, do we really need to add it now?

Rash419: not really but I assume to safe to add these directive. before it was comment because we were using inline script tag in html and we removed so I added back the directivies.

caolanm (approved):

Issue #10578, opened 2024-11-22
perf: Expanding the sidebar causes several resizes of the document


Cwiiis:
### Describe the Bug
The animation of expanding the sidebar causes the document view to be resized many times. Resizing this view is very expensive, the entire canvas is resized (which is essentially destroying it) and every section is redrawn. I expect this is why the animation isn't smooth. Ideally we'd delay the document resize until after the sidebar animation is finished.

This doesn't affect closing the sidebar as closing the sidebar isn't animated (it would also be nice if this animation was symmetric, in my opinion).

### Steps to Reproduce
1. Instrument `CanvasSectionContainer.onResize` to output when resizes happen
2. Open a document and close the sidebar if it's already open
3. Open the browser console and clear it
4. Expand the sidebar

### Expected Behavior
A single resize event.

### Actual Behavior
Several resize events, depending on the power of the machine and the pixel size of the document.

Cwiiis:
The root of this problem is essentially using jQuery to show the sidebar - this just animates width/height/opacity and of course that means constant relayouts and resize events. The ideal thing would be to replace these with CSS transitions on transforms.

PR #11044, opened 2025-01-29
Replace jQuery show animation with CSS animation in sidebar

Cwiiis:
The sidebar-showing animation was relying on jQuery, which just animates the width/height properties (absolutely killer for performance). Replace this with a CSS animation. Note, the animation isn't identical; previously the document size would animate, but this now just snaps to the destination size. Animating document size changes is unlikely to be viable any time soon, but we may want to consider some kind of 2-part move + resize animation.

Change-Id: I68eb16e9eff9a3a9601f3f903f9edb25b97a56cd

* Resolves: #10578
* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Sidebar.ts
+++ b/browser/src/control/Control.Sidebar.ts
@@ -63,11 +63,11 @@ class Sidebar {
 	}
 
 	isVisible(): boolean {
-		return $('#sidebar-dock-wrapper').is(':visible');
+		return $('#sidebar-dock-wrapper').hasClass('visible');
 	}
 
 	closeSidebar() {
-		$('#sidebar-dock-wrapper').hide();
+		$('#sidebar-dock-wrapper').removeClass('visible');
 		this.map._onResize();
 
 		if (!this.map.editorHasFocus()) {
@@ -236,8 +236,7 @@ class Sidebar {
 				}
 
 				this.builder.build(this.container, [sidebarData]);
-				if (!this.isVisible())
-					$('#sidebar-dock-wrapper').show(this.options.animSpeed);
+				if (!this.isVisible()) $('#sidebar-dock-wrapper').addClass('visible');
 
 				this.map.uiManager.setDocTypePref('ShowSidebar', true);
 			} else {
--- a/cypress_test/integration_tests/desktop/writer/editable_area_spec.js
+++ b/cypress_test/integration_tests/desktop/writer/editable_area_spec.js
@@ -211,12 +211,14 @@ describe(['taga11yenabled'], 'Editable area - Basic typing and caret moving', fu
 		ceHelper.type('{enter}');
 		ceHelper.type('green red');
 		// Need to wait here after typing paragraph in
-		// order for caret position to stick later
-		cy.wait(200);
+		// order for caret position to stick later.
+		// FIXME: We shouldn't need to wait.
+		cy.wait(1000);
 		ceHelper.checkPlainContent('green red');
 		ceHelper.moveCaret('left', '', 4);
 		ceHelper.checkCaretPosition(5);
 		helper.getBlinkingCursorPosition('P2');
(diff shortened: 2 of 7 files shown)


browser/src/control/Control.Sidebar.ts
```diff
@@ -208,41 +215,30 @@ class Sidebar {
...
-						// just to be sure chack with other method
-						if (isCurrent === 'false' || !isCurrent)
-							isCurrent =
-								this.targetDeckCommand === this.commandForDeck(currentDeck);
-						if (this.targetDeckCommand && (isCurrent === 'false' || !isCurrent))
-							this.changeDeck(this.targetDeckCommand);
-					} else {
+			this.onResize();
+
+			if (sidebarData.children[0].id) {
```
eszkadev, line 220: I would leave these additional checks, core seems to send weird JSON sometimes

browser/src/control/Control.Sidebar.ts
```diff
@@ -187,10 +193,11 @@ class Sidebar {
 		if (
 			sidebarData.action === 'close' ||
 			window.app.file.disableSidebar ||
-			this.map.isReadOnlyMode()
+			this.map.isReadOnlyMode() ||
+			!(sidebarData.children && sidebarData.children.length)
 		) {
 			this.closeSidebar();
-		} else if (sidebarData.children) {
+		} else {
```
eszkadev, line 200: best to leave as it was (sidebarData.children check)
not sure we filtered all the actions which will not have that property

Cwiiis: Agreed, I've pushed the minimal changeset. This file could really use a refactor, but this isn't the commit to do it in (or rather, half-heartedly do a tiny amount in :))

Cwiiis:
Looking into each failure - I observe correct behaviour, I expect some tests are breaking because they ran and completed while the sidebar was still expanding, which would change horizontal scroll positions. The sidebar now doesn't cause the document to go through a series of resizes, so some adjustments need to be made - making them now.

Cwiiis:
I've fixed the related failures, there's one more failure here that I believe is an intermittent ("Moving between paragraphs" in `desktop/writer/editable_area_spec.js`), probably caused by the same issue we had with an intermittent search highlighting failure - If you move up and down quickly, the cursor is in the wrong place, if you leave time before moving down, it's in the correct place. I expect this is due to not waiting for a server acknowledge of the first cursor move and I expect this may have been made worse by this patch as the jank caused by the expanding sidebar may have slowed this test down.

cypress_test/integration_tests/desktop/writer/editable_area_spec.js
```diff
@@ -211,12 +211,14 @@ describe(['taga11yenabled'], 'Editable area - Basic typing and caret moving', fu
...
-		// order for caret position to stick later
-		cy.wait(200);
+		// order for caret position to stick later.
+		// FIXME: We shouldn't need to wait.
+		cy.wait(1000);
 		ceHelper.checkPlainContent('green red');
 		ceHelper.moveCaret('left', '', 4);
 		ceHelper.checkCaretPosition(5);
 		helper.getBlinkingCursorPosition('P2');
+		cy.wait(800);
```
Cwiiis, line 221: Not happy about the size of these waits, but these seemed to be the minimum values to get this test to pass consistently locally. Both of these are masking a real bug, I can easily replicate the failure with manual inputs and it doesn't mirror the behaviour in desktop Writer. This seems very similar to #10585 and I wouldn't be surprised if the underlying reason is basically the same.

gokaysatir (approved):
This works and i guess increases the rendering performance.

For the visual changes, i would also inform @pedropintosilva :) It'd be good to have his opinion.

Thank you for the improvement.

pedropintosilva (approved):

PR #11048, opened 2025-01-29
Make preset upload, fetch and delete async

Rash419:
* Target version: master


caolanm (approved):
This seems reasonable to me

PR #11049, opened 2025-01-29
fix(mobileWizard): stop highlighting on close

Minion3665:
Way-back-when, we deselected the toolbar button when the mobile wizard was closed. We stopped doing this during the leaflet migration because the functions we previously used to do it no longer existed (I652a163ea0c831b47ce7725ffd0f9f190a98d9bf). Instead, we can use whether the 'selected' class is on our elements in the DOM to see if they should be unchecked.

That change had far more '.uncheck(...)' and '.check(...)' than I've set up here, and some of those may be necessary, but this was the bug I knew about and I didn't want to poke around with things that I wasn't completely sure where I was changing.

Change-Id: I737693f6db96c8298c07b195aa374a1bbb670108

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.MobileWizard.js
+++ b/browser/src/control/Control.MobileWizard.js
@@ -132,17 +132,17 @@ L.Control.MobileWizard = L.Control.extend({
 	},
 
 	_updateToolbarItemStateByClose: function() {
-		var toolbar = app.map.mobileTopBar;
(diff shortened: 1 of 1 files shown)


eszkadev (approved):
thanks for the fix!

PR #11052, opened 2025-01-30
Buttons in presenter console does not have border radius until be we hover on it

Darshan-upadhyay1110:
- all the btn (ex: Notes, Slider sorter etc) should have border radius on initialization
- this patch will fix the inconsistency of border radius around btn in PresenterConsole

Change-Id: I880159a69cf5c20ac914785ca12c826c257b6c96

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


hcvcastro (approved):

hcvcastro:
It looks good

PR #11055, opened 2025-01-30
Seccomp: filter more syscalls

caolanm:
Change-Id: Ia17360035035418ada519cf758e38be35c01a177 (cherry picked from commit 32f30437e6fe6185190cef37fd666756f6fe88cb)

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (approved):
Looks good, I don't recall a case off the top of my head where the kit process would need to exec new binaries. The poppler-based PDF import would be one such case, but we use pdfium instead.

PR #11059, opened 2025-01-30
Fix: Prevent cursor position reference issue

Darshan-upadhyay1110:
- Created a separate copy of cursorPosAtStart to avoid modifying the original values when adjusting `cursorPos.y`.
- this will fix the popup moves when user types mention name
- popup should be at start and do not move

Change-Id: I09fad05d742cec7f735648e4f0a5e5217d0d2434

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Rash419 (approved):
LGTM!

PR #11061, opened 2025-01-30
js: disable autotext dialog shortcut

Rash419:
Change-Id: I40cf7f0d0fcff4270a9172b6b50ffc4aa3da899e

* Target version: master

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -671,6 +671,11 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// Control + F3, ignore Autotext dialog
+		if (this._isCtrlKey(e) && e.keyCode === this.keyCodes.F3) {
+			return true;
+		}
+
 		// Control + Shift + I, open browser developper tools
 		if (this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.I) {
 			return true;


caolanm (approved):
Yeah, I wanted the F3 autotext expansion, but I was unaware of ctrl+F3 to get the dialogs, which we don't want. At this time at least.

browser/src/map/handler/Map.Keyboard.js
```diff
@@ -671,6 +671,11 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// Control + F3, ignore Autotext dialog
+		if (this._isCtrlKey(e) && e.keyCode === this.keyCodes.F3) {
```
eszkadev, line 675: I see it is merged already, but I have some comment for that:
- if you need to add/modify any keyboard shortcut - please add it into https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts#L161
- it allows deterministic, easily auditable way to define shortcuts and overwrites for different languages
- adding more ifs to Map.Keyboard makes it harder to maintain and we had many regressions due to not easy to spot connections in the past

Rash419: @eszkadev thank you for review. Let me fix it : )

Rash419: Done: https://github.com/CollaboraOnline/online/pull/11069

PR #11065, opened 2025-01-31
cool#11064 kit: check for callbacks & wsd socket in the anyInput callback

vmiklos:
Open the 300 pages bugdoc, go to the end of page 1 (before a page
break), press enter, a page 2 is inserted and we get new tiles after a
delay:
15:08:33.803 global.js:566 1738073313803 OUTGOING: key type=up char=0 key=1280
...
15:08:34.249 global.js:566 1738073314249 INCOMING: delta: nviewid=1000 part=0 ...
-> 446 ms.

The reason this happens is because our anyInput callback tells core to
finish the idle Writer layout first and only then render the tiles for
the visible area, which is a problem similar to what commit
930cd616fb6defd8c0de0163bd92e637c7b6b76f (Related: cool#9735 kit: use
the new registerAnyInputCallback() LOK API, 2024-08-26) fixed.

Fix this problem by improving the anyInput callbacks to take 2 more
"inputs" into account:
1) LOK_CALLBACK_INVALIDATE_TILES went to KitQueue::_callbacks and this
   was only processed in the next main loop iteration, so checks for
  this in the anyInput callback, which partly reverts commit
  2823933a2111136e14a8605cf5698dff40835e72 (anyInput - should only depend
  on incoming messages from coolwsd., 2024-12-10).
2) Now that the invalidation is processed, it's sent to the wsd process
   and we get a tile request on the wsd -> kit socket, in the kit
  process. Fix this problem by doing a poll() on that socket as the
  already existing FIXME comment suggested.

With this, we paint tiles for the visible area before calculating pages
3..300 in Writer:
10:15:47.441 global.js:566 1738314947441 OUTGOING: key type=input char=13 key=1280
...
10:15:47.476 global.js:566 1738314947476 INCOMING: delta: nviewid=1000 part=0 ...
-> 24ms

Change-Id: I668d2ebf341b628841e0a637fee9e476eeddd179

### DIFF
--- a/net/Socket.cpp
+++ b/net/Socket.cpp
@@ -475,7 +475,7 @@ void SocketPoll::enableWatchdog()
     _watchdogTime = Watchdog::getTimestamp();
 }
 
-int SocketPoll::poll(int64_t timeoutMaxMicroS)
+int SocketPoll::poll(int64_t timeoutMaxMicroS, bool justPoll)
 {
     if (_runOnClientThread)
         checkAndReThread();
@@ -525,6 +525,22 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
 
     // from now we want to race back to sleep.
     enableWatchdog();
+
+    if (justPoll)
+    {
+        // Done with the poll(), don't process anything.
+        bool ret = false;
+        // Run through the poll entries (except the wakeup poll), and combine them into an answer.
+        for (size_t i = 0; i < size; ++i)
+        {
+            if (_pollFds[i].revents)
+            {
+                ret = true;
+                break;
+            }
+        }
+        return ret;
+    }
 
     // First process the wakeup pipe (always the last entry).
     if (_pollFds[size].revents)
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3016,13 +3016,33 @@ int pollCallback(void* data, int timeoutUs)
 }
 
 // Do we have any pending input events from coolwsd ?
-// FIXME: we could helpfully poll our incoming socket too here.
 bool anyInputCallback(void* data)
 {
     auto kitSocketPoll = reinterpret_cast<KitSocketPoll*>(data);
     const std::shared_ptr<Document>& document = kitSocketPoll->getDocument();
 
-    return document && document->hasQueueItems();
+    if (document)
+    {
+        if (document->hasCallbacks())
+        {
+            // Have pending LOK callbacks from core.
+            return true;
+        }
+
+        // Poll our incoming socket from wsd.
+        int ret = kitSocketPoll->poll(std::chrono::microseconds(0), /*justPoll=*/true);
+        if (ret)
+        {
+            return true;
+        }
+
+        if (document->hasQueueItems())
+        {
+            return true;
+        }
+    }
+
+    return false;
 }
 
 /// Called by LOK main-loop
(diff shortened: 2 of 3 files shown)


vmiklos:

Please don't merge it, though; I believe this belongs to red, re-arranging events is always tricky. Also I have some idea on how to test this, but I need to build infrastructure for that, so probably that'll need some core side changes and also a follow-up COOL PR.

CC @mmeeks

net/Socket.cpp
```diff
@@ -526,6 +526,12 @@ int SocketPoll::poll(int64_t timeoutMaxMicroS)
     // from now we want to race back to sleep.
     enableWatchdog();
 
+    if (justPoll && size > 0)
+    {
+        // Done with the poll(), don't process anything.
+        return _pollFds[0].revents;
```
caolanm, line 532: And we definitely just want to look at the first one, not all of then (or all of minus trailing wakeup) ?

caolanm (approved):

net/Socket.cpp
vmiklos: This is for the kit case where (I think) it's guaranteed that we have exactly one incoming socket (from the wsd), but let me assert that, so it's more explicit.

In theory a "just poll" mode with multiple sockets would be interesting, but I don't see a practical need for that yet.

vmiklos: > This is for the kit case where (I think) it's guaranteed that we have exactly one incoming socket (from the wsd)

I was wrong, Document::forkToSave() calls insertNewSocket() for the child socket, so we can have multiple sockets for the background save, but it seems the wsd socket is always the same in the array, seeing that https://github.com/collaboraonline/online/blob/c26e006e16a5110dccbb8ca63d3a5f32dbe1f834/net/Socket.cpp#L559 inserts at the end, I think the `_pollFds[0].revents` reading is fine at the end.

> Typically such pieces of code parse & prioritize the incoming messages and stick them into a queue - rather than acting on them immediately

I think I had that normal poll() before 7ca2677e8ef37effa13147611c38e20329f03bab and I removed it because you were concerned about we may do too much processing there and https://github.com/CollaboraOnline/online/pull/9974#issuecomment-2332219771 suggests we may have a re-entrancy hazard if we do "poll + queue" instead of just "poll".

Am I correct that then this "just poll" way is correct or did I miss something? Thanks.

mmeeks: Lets just run through all the poll entries (except perhaps the wakeup poll if that turns out to be a bad idea (?)) - and OR those all together into an answer.

Otherwise the API change looks fine - thanks =)

vmiklos: > Lets just run through all the poll entries (except perhaps the wakeup poll if that turns out to be a bad idea (?)) - and OR those all together into an answer.

Aha, we may have multiple interesting poll entries, e.g. the background safe process, OK.

Including the wakeup poll in the result looks like making everything slower. Rough times for loading this big document:

1. just poll the first socket: 5.06 s (stopwatch from opening the URL till layout settles and shows the 300 pages in the page count statusbar)
2. poll all sockets: 12.32 s
3. poll all-1 sockets (i.e. exclude the wakeup pipe): 5.12 s

(Not very precise measurement, I would consider 1. and 3. equivalent.)

So let me exclude the wakeup pipe for now, we can revisit that later if needed.

mmeeks: Makes sense - I would expect the wakeup pipe to be nearly always full when we're busy in the core anyway =) Thanks !

PR #11069, opened 2025-02-03
js: move disabling AutoText dialog shortcut logic to Map.KeyboardShortcuts.ts

Rash419:
Change-Id: I28e1583b0ba3cb3e0e2bf85bfd482392e3ea04d8

* Target version: master


eszkadev (approved):
didn't test, but looks good :)

browser/src/map/handler/Map.KeyboardShortcuts.ts
```diff
@@ -175,6 +175,7 @@ keyboardShortcuts.definitions.set('default', new Array<ShortcutDescriptor>(
 
     new ShortcutDescriptor('text', 'keydown', 0, 'F2', null, null, null),
     new ShortcutDescriptor('text', 'keydown', 0, 'F3', '.uno:ExpandGlossary', null, null),
+    new ShortcutDescriptor('text', 'keydown', Mod.CTRL, 'F3', null, null, null),
```
eszkadev, line 178: shouldn't be before the plain F3 ? so in case we have also ctrl we will use that rule before

PR #11078, opened 2025-02-04
Private/darshan/add accessbox for sidebaricons

Darshan-upadhyay1110:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -559,12 +559,14 @@ L.Control.Notebookbar = L.Control.extend({
 			{
 				'type': 'toolitem',
 				'text': _UNO('.uno:Sidebar', '', true),
-				'command': '.uno:SidebarDeck.PropertyDeck'
+				'command': '.uno:SidebarDeck.PropertyDeck',
+				'accessibility': { focusBack: false,	combination: 'ZB', de: null }
 			},
 			{
 				'type': 'toolitem',
 				'text': _UNO('.uno:Navigator'),
-				'command': '.uno:Navigator'
+				'command': '.uno:Navigator',
+				'accessibility': { focusBack: false, combination: 'ZN', de: 'V' }
 			},
 			{
 				'type': 'toolitem',
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -455,10 +455,6 @@ L.Control.Notebookbar = L.Control.extend({
 						'type': 'toolitem',
 						'text': _UNO('.uno:SidebarDeck.ElementsDeck', '', true),
 						'command': '.uno:SidebarDeck.ElementsDeck'
-					},
-					{
-						'type': 'toolitem',
-						// dummy node to avoid creating labels
 					}
 				];
 			}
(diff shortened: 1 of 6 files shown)


eszkadev:
I remember for German community shortcuts used were important so they don't have to change habits. Worth checking if any change is good/allowed. More info can have: @gokaysatir or @pedropintosilva

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -55,13 +55,13 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 				'text': _('Layout'),
 				'id': layoutTabName + '-tab-label',
 				'name': layoutTabName,
-				'accessibility': { focusBack: true, combination: 'P', de: 'S' }
+				'accessibility': { focusBack: true, combination: 'L', de: 'L' }
```
pedropintosilva, line 58: discard change

browser/src/control/Control.NotebookbarWriter.js
```diff
@@ -55,13 +55,13 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
...
 			},
 			{
 				'text': _('References'),
 				'id': referencesTabName + '-tab-label',
 				'name': referencesTabName,
-				'accessibility': { focusBack: true, combination: 'S', de: 'C' }
+				'accessibility': { focusBack: true, combination: 'C', de: 'C' }
```
pedropintosilva, line 64: discard changes

browser/src/control/Control.Notebookbar.js
```diff
@@ -555,22 +555,31 @@ L.Control.Notebookbar = L.Control.extend({
...
+				'command': '.uno:Navigator',
+				'accessibility': { focusBack: false, combination: 'K', de: 'V' }
 			},
 			{
-				'type': 'toolitem',
-				// dummy node to avoid creating labels
+				'type': 'customtoolitem',
+				'text': _('Share'),
+				'command': 'shareas',
+				'accessibility': { focusBack: false, combination: 'SH', de: null },
```
pedropintosilva, line 579: I would prefer to group all sidebar buttons in terms of access-key by assign the same first letter and the second letter should be based on the label and not the uno command. So, in this case would be : `ZS`

gokaysatir:
> I remember for German community shortcuts used were important so they don't have to change habits. Worth checking if any change is good/allowed. More info can have: @gokaysatir or @pedropintosilva

We had a chat about this with @Darshan-upadhyay1110 and my suggestion is to keep "S" for layout in German UI. But Pedro knows best.

browser/src/dom/NotebookbarAccessibilityDefinitions.js
```diff
@@ -133,6 +133,21 @@ var NotebookbarAccessibilityDefinitions = function() {
 				}
 			}
 		}
+
+		// add optionstoolbox accesskey information to selected definitions
+
+		const optionsToolSectionData = app.map.uiManager.notebookbar.getDefaultToolItems();
```
gokaysatir, line 139: I think we need a function for this. "checkIntegratorButtons" doesn't seem like a good location to add this functionality.

browser/src/dom/NotebookbarAccessibilityDefinitions.js
```diff
@@ -108,15 +108,15 @@ var NotebookbarAccessibilityDefinitions = function() {
...
+		var containerList = ['save', 'userListHeader', 'shortcutstoolbox', 'closebuttonwrapper'];
 		for (var i = 0; i < containerList.length; i++) {
 			var container = document.getElementById(containerList[i]);
 
 			if (container) {
-				// All the buttons inside the container.
-				var buttonList = container.querySelectorAll('button');
-
+				// All the buttons inside the container which are visible on screen.
+				var buttonList = Array.from(container.querySelectorAll('button')).filter(btn => {
```
gokaysatir, line 117: Why do we need this change? Can you please write how to reproduce the issue?

gokaysatir (changes requested):

browser/src/dom/NotebookbarAccessibilityDefinitions.js
Darshan-upadhyay1110: Sure i will do that improvement with other requested changes ...

browser/src/dom/NotebookbarAccessibilityDefinitions.js
gokaysatir: Thank you :) I will check this soon. Meanwhile, i would most probably check it dynamically before creating the shortcut rectangles (those yellow ones). So code can cope with dynamic changes.

gokaysatir: Can you please check this again? After moving the visibility check into the "addTabAccelerators" function, i guess we don't need this visibility check here anymore? If we still do need, then i'd still remove this from here and add it to a position that checks the visibilities every time.

gokaysatir (changes requested):
I have one comment other than that, it looks good, thanks :)

Also, i guess we'll need a Cypress test for the accelerator buttons.

browser/src/dom/NotebookbarAccessibilityDefinitions.js
Darshan-upadhyay1110: Yes correct no need to add it again in check Integrator buttons

eszkadev (approved):
code looks good for me, if Gokay is also happy we can merge & test

eszkadev:
please rebase, cypress was fixed in the meantime

gokaysatir (approved):
Cool, thanks @Darshan-upadhyay1110 :)

PR #11079, opened 2025-02-04
FileServer: replace needed template variables

meven:
* Resolves: # 
* Target version: master 

### Summary

replace needed template variables
And prevent bad values to be set to brandProductName and brandProductURL.

### Checklist

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -281,8 +281,14 @@ class InitializerBase {
 			document.getElementById('content-keeper').remove();
 		}, false);
 
-		window.brandProductName = document.getElementById("init-product-branding-name").value;
-		window.brandProductURL = document.getElementById("init-product-branding-url").value;
+		let productName = document.getElementById("init-product-branding-name").value;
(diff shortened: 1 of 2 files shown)


timar (approved):

timar:
@meven CI won't pass, until you fix the error: maybe `std::string()` instead of `""` would work.

meven:
> @meven CI won't pass, until you fix the error: maybe `std::string()` instead of `""` would work.

Thanks for the tip, indeed the rest of the code there uses std::string and we don't use automatic char[] to string, I wasn't aware of.

Fixed now.

meven:
```
      cy:command ✘  assert	expected **15** to be below **14**
                    Actual: 	15
                    Expected: 	14
      cy:command ✔  fail:	
                    Test failed: integration_tests/multiuser/writer/invalidations_spec.js / Joining a document should not trigger an invalidation / Join after document save and modify
```

Ci is constantly failing at the same test.
I highly doubt this code could impact this.

PR #11081, opened 2025-02-04
Fuzzing and Misc Improvements

Ashod:
- **fuzz: include cleanup in HttpEcho.cpp**
- **wsd: constexpr Util::isFuzzing()**
- **wsd: short-circuit isKitInProcess() where possible**
- **clang-tidy: ignore static-assert**
- **fuzz: move CHECK to Common.hpp**
- **fuzz: no error logging while fuzzing**
- **fuzz: disable some assertions during fuzzing**
- **fuzz: disable precondition assertion during fuzzing**
- **fuzz: disable handling of some commands**
- **fuzz: disable asserting for valid commands**
- **fuzz: never terminate**
- **wsd: use the sanitized LogLevel**
- **wsd: avoid copying shared_ptr**
- **wsd: optimize resolveDNS eviction**

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -140,12 +140,6 @@ static HostEntry resolveDNS(const std::string& addressToCheck,
 {
     const auto now = std::chrono::steady_clock::now();
 
-    // remove entries >= 20 seconds old
-    std::erase_if(querycache, [now](const auto& entry)->bool {
-                                 auto ageMS = std::chrono::duration_cast<std::chrono::milliseconds>(now - entry.lookupTime).count();
-                                 return ageMS > 20000;
-                              });
-
     // search for hit
     auto findIt = std::find_if(querycache.begin(), querycache.end(),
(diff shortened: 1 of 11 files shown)


net/NetUtil.cpp
```diff
@@ -140,24 +140,28 @@ static HostEntry resolveDNS(const std::string& addressToCheck,
 {
     const auto now = std::chrono::steady_clock::now();
 
-    // remove entries >= 20 seconds old
-    std::erase_if(querycache, [now](const auto& entry)->bool {
```
Ashod, line 144: It's true that this would remove _all_ expired references, not the one we lookup.

This means that entries that are stale, will remain in the cache unless/until there is a lookup, accumulating entries.

I suspect this is highly unlikely, but possible. As a workaround, we can revise this to do `erase_if` when we have an expired entry. That is, the first entry that is expired, also prompts us to erase any _other_ entries that might be expired as well. This way, we don't get the double-lookup hit on each lookup.

caolanm (approved):
fine by me

PR #11082, opened 2025-02-04
browser: accessibility: add 'aria-disabled' attribute

hcvcastro:
Change-Id: Ic71df7415075523c841194902b0f7f1a6177b6b8

### DIFF
--- a/browser/src/control/Control.JSDialogBuilder.js
+++ b/browser/src/control/Control.JSDialogBuilder.js
@@ -357,13 +357,14 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			$(spinfield).attr('step', step);
 		}
 
-		if (data.enabled === false) {
+		const isDisabled = data.enabled === false;
+
+		if (isDisabled) {
 			div.disabled = true;
 			spinfield.setAttribute('disabled', 'true');
 		}
 
-		var enabled = Boolean(data.enabled);
-		spinfield.setAttribute('aria-disabled', !enabled);
+		spinfield.setAttribute('aria-disabled', isDisabled);
 
 		JSDialog.SynchronizeDisabledState(div, [spinfield]);
 
@@ -374,9 +375,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			$(spinfield).hide();
 
 		spinfield.addEventListener('change', function() {
-			var isDisabled = div.hasAttribute('disabled');
+			const isCurrentlyDisabled = div.hasAttribute('disabled');
 			var isValid = this.checkValidity();
-			if (!isDisabled && isValid) {
+			if (!isCurrentlyDisabled && isValid) {
 				if (customCallback)
 					customCallback('spinfield', 'change', div, this.value, builder);
 				else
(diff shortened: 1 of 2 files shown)


browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -381,6 +381,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			spinfield.setAttribute('disabled', 'true');
 		}
 
+		spinfield.setAttribute('aria-disabled', Boolean(data.enabled));
```
eszkadev, line 384: wrong logic:     should be `data.enabled === false` (if not present it means it is enabled too)

eszkadev (changes requested):

browser/src/control/Control.JSDialogBuilder.js
hcvcastro: Hmm, it is a core property of every widget, it should always exist. Can you give some example to investigate?

eszkadev: disabled != enabled

please reread carefully, the logic is inverted here ;)

hcvcastro: jaja,  I see now, . Thanks

vmiklos:
https://github.com/CollaboraOnline/online/pull/11193 has to go in first, then please rebase. Sorry about the trouble.

browser/src/control/jsdialog/Widget.MenuButton.js
```diff
@@ -158,6 +158,8 @@ function _menubuttonControl (parentContainer, data, builder) {
 			}
 		});
 
+		button.setAttribute('aria-disabled', Boolean(data.enabled));
```
eszkadev, line 161: still wrong logic

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1328,6 +1332,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			checkbox.disabled = true;
 		}
 
+		checkbox.setAttribute('aria-disabled', Boolean(data.enabled));
```
eszkadev, line 1335: wrong logic

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -1285,6 +1287,8 @@ L.Control.JSDialogBuilder = L.Control.extend({
 		if (data.enabled === 'false' || data.enabled === false)
 			$(radiobutton).attr('disabled', 'disabled');
 
+		radiobutton.setAttribute('aria-disabled', Boolean(data.enabled));
```
eszkadev, line 1290: wrong logic

hcvcastro:
Oh my gosh, I forgot to push the new changes. Not a good day =(

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -382,6 +382,9 @@ L.Control.JSDialogBuilder = L.Control.extend({
 			spinfield.setAttribute('disabled', 'true');
 		}
 
+		var enabled = Boolean(data.enabled);
```
eszkadev, line 385: In case of missing `enabled` property it will be `undefined` (it is optional by default it is enabled).
Then `!enabled` will be `true` and will disable widget which is not good.

Please use explicit check for `const disabled = data.enabled === false;`

eszkadev: We use that just above in the if condition. Would be good to have single variable and reuse it.

eszkadev:
desktop passed (only known fail was present)

browser/src/control/Control.JSDialogBuilder.js
eszkadev: @hcvcastro could you please adjust code to above corner-case?
apart of that is ready to merge

eszkadev (approved):

PR #11087, opened 2025-02-05
test, UnitTimeoutInactive: increase the timeout further

vmiklos:
Encouraged by commit 91a83bcacb7d1d181bb6e37a26e69920dfde912a (increase
UnitTimeoutInactive inactivityTimeout, 2025-01-30), looking at why a
core.git dbgutil + online.git debug build still fails here, it seems it
now passes fine with a bit more relaxed timeout.

The new value is still 10k times smaller than the default, so is meant
to be still minimal.

Change-Id: I04cc2e111ac86b4bb49e03ab73635796078f0a1a


vmiklos:

I put 'run make check locally' aside in the recent past as I had no time to investigate the failures, but seeing your patch, this fixes the c++ part I see locally.

caolanm (approved):
sounds good to me

PR #11088, opened 2025-02-05
config: support hasSettingIframeSupport capability

codewithvk:
Change-Id: I8f4d3bb5c8629a469fbf377c2f1a9a55a3801198

* Resolves: # 
* Target version: master

### DIFF
--- a/common/ConfigUtil.cpp
+++ b/common/ConfigUtil.cpp
@@ -297,6 +297,7 @@ static const std::unordered_map<std::string, std::string> DefAppConfig = {
     { "watermark.text", "" },
     { "welcome.enable", "false" },
     { "zotero.enable", "true" },
+    { "setting_iframe.enable", "true" },
 };
 
 void initialize(const Poco::Util::AbstractConfiguration* config)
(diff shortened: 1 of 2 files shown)


caolanm (approved):
Any reason to not just make it:
capabilities->set("hasSettingIframeSupport", true)
But maybe it is desirable to be able to override this via the config?

eszkadev:
> Any reason to not just make it: capabilities->set("hasSettingIframeSupport", true) But maybe it is desirable to be able to override this via the config?

+1, but I think we can merge it now as it will be useful for testing

eszkadev (approved):

caolanm:
yeah, agreed

codewithvk:
> Any reason to not just make it: capabilities->set("hasSettingIframeSupport", true) But maybe it is desirable to be able to override this via the config?

Actually, I was thinking of adding a config condition inside the file server where we serve the routes of the iframe.(https://github.com/CollaboraOnline/online/blob/master/wsd/FileServer.cpp#L971) This would make it easier to identify that this config is responsible for the setting iframe, and in the future, someone else can find it easily. Since we are in a hurry, I thought, let’s just make this change and push it. We can do the improved version later.

let me know your thoughts?

caolanm:
If you foresee a need to be able to toggle it then it looks sensible to have a config setting for it.

PR #11089, opened 2025-02-05
Use current cell's decimal separator when decimal separator button on…

gokaysatir:
… keyboard is pressed.

Issue: When a language that uses a decimal separator symbol other than current keyboard's layout's decimal separator symbol, the decimal separator button in numPad sends the wrong char.

To prevent this, we send the decimal separator of the current cell from core side. Then we watch the NumpadDecimal keycode. If it is active, we send current cell's decimal separator instead of keyboard layout's.

Change-Id: I1be81bcd5cc5f19016961eccdac180803e47df89

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


vmiklos (commented):
Looks promising, thanks. Did you consider writing a test for this? At least the part that reads app.calc.decimalSeparator could be tested from cypress, I believe. setDummyClipboardForCopy() has code that sets a variable inside the 'app' structure, so that looks doable with reasonable amount of effort. Could you try that, please? Thanks.

gokaysatir:
Thank you for checking. I added a test as the second commit.

cypress_test/integration_tests/desktop/calc/cell_cursor_spec.js
```diff
@@ -76,3 +76,30 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Test jumping on large cell
...
+		cy.cGet('#sidebar').click({force: true});
+	});
+
+	it('Check different decimal separators', function() {
+		helper.typeIntoInputField(helper.addressInputSelector, 'A1');
+		cy.wait(400);
+
+		cy.window().then(win => {
+			var app = win['0'].app;
+			cy.expect(app.calc.decimalSeparator).to.be.equal('.');
```
vmiklos, line 94: Nice, I wasn't aware cypress can assert something that's not a DOM node.

vmiklos (approved):
Looks good to me, thanks.

PR #11090, opened 2025-02-05
feat: crop image in writer

Rash419:
Change-Id: I441f2812fe3caad4c6301e31c609621e907971f9

* Target version: master 

* TODO:
  - [x] Maybe add cypress tests
  - [x] Bug: Rotating images messes up the crop handles
  - [x] need to disable crop button(s) when something other than an image is selected.
  - [x] expose crop button in mobileview
  - [x] ~~is this writer only feature ?~~. ~~I will add crop in separate PR as it requires more work on core side. No point blocking this PR : )~~ https://github.com/Rash419/online/tree/editengine/crop/image I will create PR once this is merged as it is dependent 

* requires: https://gerrit.libreoffice.org/c/core/+/181160

image


gokaysatir:
Thanks @Rash419 :) Code looks good and it also works beautifully. Some issues you already noticed:
* We need to disable crop button(s) when something other than an image is selected.
* Rotated image cropping is slow for some reason and it is also buggy (JS errors).


I will check this again once you inform, thanks :)

gokaysatir:
Few questions:
* Is this Writer only? It seems that we can enable this for Impress, Calc and Draw.
* It seems to work with different dpiScales. I guess we can also add mobile menu buttons, under "Edit" menu etc.

browser/src/canvas/sections/ShapeHandlesSection.ts
```diff
@@ -553,13 +553,18 @@ class ShapeHandlesSection extends CanvasSectionObject {
...
+				GraphicSelection.extraInfo.isCropMode
 			);
 			return newSubSection;
 		}
 		else {
 			newSubSection.sectionProperties.ownInfo = handle.info;
 			newSubSection.setPosition(handle.point.pX, handle.point.pY);
+			if (GraphicSelection.extraInfo.isCropMode && !newSubSection.sectionProperties.cropModeEnabled) {
+				newSubSection.sectionProperties.cropModeEnabled = GraphicSelection.extraInfo.isCropMode;
+				newSubSection.containerObject.requestReDraw();
```
gokaysatir, line 566: I am not sure if we need to request a redraw here? It looks like `updateGraphicSelection` already does it.

gokaysatir (approved):
All looks good to me. Thanks @Rash419 :)

There is only one comment regarding a redraw request. I think you can have a look at it while you check the failing Cypress tests.

Thanks :)

browser/src/canvas/sections/ShapeHandlesSection.ts
Rash419: Done, Thanks.

PR #11094, opened 2025-02-06
Comment section: Use int for part hash.

eszkadev:
Check container existence before setting edit mode. This fixes a bug:
- open Impress
- insert comment on second slide
- delete the comment
- insert again
- notice it was inserted on the first slide not second as expected

change from @gokaysatir


gokaysatir (approved):
Thank you.

PR #11096, opened 2025-02-06
UserList: do not render when not changed

eszkadev:
We seems to rerender avatars (in DOM) on every key press. Do it only if selected user was changed.

see:
Map.Keyboard.js     : _handleKeyEvent
Control.UserList.ts : selectUser


caolanm (approved):
This looks reasonable to me, but e.g. this._map.userList.followUser in browser/src/map/handler/Map.Keyboard.js is there since mid 2024, and similar for the other touched places. Do we have an idea why this seems to be noticeable only recently?

mmeeks:
@timar would love to get this into a re-spin of the stable release too if possible =) still - I expect there is much more we can catch here.

PR #11098, opened 2025-02-06
Update highlight mode without waiting cursor invalidation

GulsahKose:
Pressing Column Row Highlight button wasn't switching the highlight mode quickly. We had to wait cursor invalidation to see highlight. Now pressing button triggers the mode quickly.

Change-Id: Ic5c84e625a3f96f6cc9858dfb84d14474bd9e773

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -1053,9 +1053,29 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			this._resetReferencesMarks();
 	},
 
+	updateHighlight: function () {
+		if ( this._map) {
+			if (this._map.uiManager.getHighlightMode()) {
+				var updateMsg = 'updateHighlight';
+				this._highlightColAndRow(updateMsg);
+			}
+			else
+				this._resetReferencesMarks();
+		}
+	},
+
 	_highlightColAndRow: function (textMsg) {
+		var strTwips = [];
+		if(textMsg.startsWith('updateHighlight')) {
+			strTwips[0] = app.calc.cellCursorTopLeftTwips.x;
+			strTwips[1] = app.calc.cellCursorTopLeftTwips.y;
+			strTwips[2] = app.calc.cellCursorOffset.x;
+			strTwips[3] = app.calc.cellCursorOffset.y;
+		}
+		else
+			strTwips = textMsg.match(/\d+/g);
+
 		this._resetReferencesMarks();
-		var strTwips = textMsg.match(/\d+/g);
 		var references = [];
 		this._referencesAll = [];
 		var rectangles = [];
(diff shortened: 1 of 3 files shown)


GulsahKose:
Here is the video of the current state: 
[Uploading Screencast from 2025-02-06 16-54-17.webm…]()

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -1077,7 +1099,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 			} else {
 				// Row rectangle
 				topLeftTwips = new L.Point(parseInt(0), parseInt(strTwips[1]));
-				offset = new L.Point(parseInt(maxRow), parseInt(strTwips[4]));
+				offset = new L.Point(parseInt(maxRow), parseInt(strTwips[3]));
```
eszkadev, line 1102: this changes behavior of: `!textMsg.startsWith('updateHighlight')` case

is that valid?

GulsahKose: Not a mistake. Planned. It was wrong before, i fixed it in that commit.

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -2080,13 +2080,14 @@ L.CanvasTileLayer = L.Layer.extend({
...
 			var strTwips = textMsg.match(/\d+/g);
-
 			var topLeftTwips = new L.Point(parseInt(strTwips[0]), parseInt(strTwips[1]));
 			var offset = new L.Point(parseInt(strTwips[2]), parseInt(strTwips[3]));
 			var bottomRightTwips = topLeftTwips.add(offset);
 			let _cellCursorTwips = this._convertToTileTwipsSheetArea(new L.Bounds(topLeftTwips, bottomRightTwips));
 
 			app.calc.cellAddress = new app.definitions.simplePoint(parseInt(strTwips[4]), parseInt(strTwips[5]));
+			app.calc.cellCursorTopLeftTwips = topLeftTwips;
+			app.calc.cellCursorOffset = offset;
```
eszkadev, line 2090: what does the above cause? Also cc: @gokaysatir , he knows best that code

GulsahKose: I need to know where is the cursor in twips. We have app.calc.cellCursorRectangle but it is a converted from Twips to SheetArea. If i use this converted value higlight is shown wrong.

GulsahKose: But maybe i can reverse convert with getTileTwipsPointFromPrint let me try o/

eszkadev: I don't say it's wrong, I just say it's not obvious for me what it does :)

GulsahKose: I tried but conversion causes rounding error.  I think this is the most safest way to get exact cursor position and offset in printTwips.

eszkadev:
code looks fine for me, if @gokaysatir will say     cursor-related changes are good - we can merge :)

gokaysatir (approved):
Looks good to me too, thanks :)

I would also add the highlight button to mobile view menu, if it is not there yet.

Also, a Cypress test would be good to have :)

gokaysatir:
Test can press the button and check if the highlight is on, that'd be enough to guard the feature i think.

Btw, it is a  beautiful UI improvement.

GulsahKose:
Thanks I can add cypress test and mobile view improvements as followup commit.

PR #11101, opened 2025-02-07
cypress: check if we are fully initialized before start

eszkadev:
I hope this helps to be sure we are ready to start testing.
Excluded from new check:
- multi
- draw
as are not working yet


caolanm (approved):
This looks encouraging

PR #11109, opened 2025-02-08
More localizable issue count on Server Audit dialog

timar:
Some languages have more than 1 plural forms, which we cannot handle in current l10n framework.

Change-Id: Ifc5cb44e667ea8649b75d2983198e655b36728d0


banobepascal (commented):
The word `count` sounds abit more technical and the number itself already implies counting, how about `Issues found:`

For example: `Issues found: 5` or even just `Issues: 5`

What do you think @pedropintosilva

banobepascal (approved):

PR #11112, opened 2025-02-10
Do not fail install when /opt/cool is not fully deletable

timar:
Backstory:
Our client has set up an EDR on their servers.
Recently Collabora could not be updated.

On the configuration step of the coolwsd package, the script tries to remove "/opt/cool/systemplate/etc/hosts" . Because it's a hard-link of the "/etc/hosts", the EDR disallow any modification, including removal.

The client doesn't want to set an exception to allow a package to modify a sensible file like "hosts". Can Collabora avoid a removal on this file ?

--

I hope this helps...

Change-Id: Id4c1d72b705dee460a709a5a22312bc3d692a5a6


caolanm (approved):
Hopefully it won't go on to fail at a later step

PR #11117, opened 2025-02-10
private/pedro/insert tabs string improvements

pedropintosilva:
- **Insert Tab: Remove "Insert" from "Insert Table"**
- **Writer: Insert Tab: Remove "Object" from "Formula Object"**


timar (commented):
OK for red CODE.

PR #11118, opened 2025-02-10
zotero: request items in async chunks using paging for speed

lpranam:
Change-Id: I6c7d76d08e305cd67e714abe5cf36000b8fc05c0

* Target version: master 

### Checklist

### DIFF
--- a/browser/src/control/Control.Zotero.js
+++ b/browser/src/control/Control.Zotero.js
@@ -393,31 +393,33 @@ L.Control.Zotero = L.Control.extend({
 		return this;
 	},
 
-	enableDialogOKButton: function() {
+	updateControl: function (controlJSON) {
 		this.map.fire('jsdialogupdate', {
 			data: {
 				jsontype: 'dialog',
 				action: 'update',
 				id: 'ZoteroDialog',
-				control: {
-					id: 'ok',
-					type: 'pushbutton',
-					text: _('OK'),
-					'has_default': true
-				},
+				control: controlJSON,
 			},
 			callback: this._onAction.bind(this)
 		});
 	},
 
+	enableDialogOKButton: function() {
+		this.updateControl(
+			{
+				id: 'ok',
+				type: 'pushbutton',
+				text: _('OK'),
+				'has_default': true
+			}
+		);
+	},
+
 	enableDialogLanguageCombobox: function(locale) {
 		var that = this;
-		this.map.fire('jsdialogupdate', {
-			data: {
-				jsontype: 'dialog',
-				action: 'update',
-				id: 'ZoteroDialog',
-				control: {
+		this.updateControl(
+			{
 					id: 'zoterolocale',
 					type: 'combobox',
 					entries: Array.from(Object.keys(this.availableLanguages), function(langCode) {return that.availableLanguages[langCode][0];}),
@@ -425,10 +427,7 @@ L.Control.Zotero = L.Control.extend({
 					selectedEntries: [
 						Object.keys(this.availableLanguages).indexOf(locale)
 					],
-				},
-			},
-			callback: this._onAction.bind(this)
-		});
+				});
 	},
 
 	enableDialogFieldTypeCombobox: function(citationFormat) {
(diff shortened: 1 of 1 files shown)


lpranam:
todo:
1. pagination inside the js dialog (according to profile, after data arrived from Zotero most time is spent creating and rendering entries)
2. cache the item entries with the index to open the dialog with the same entry sequence

eszkadev:
>     1. pagination inside the js dialog (according to profile, after data arrived from Zotero most time is spent creating and rendering entries)

1. do we want paginated view? will the sorting/filtering be possible on full collection?
or you will use API for that, not the cached entries?

2. is the showing tree time long? (I mean rendering entries takes many seconds, or it's just high % of short time?)
If it is ~instant - no need to optimize probably... Could you upload video?

I know about some optimization which can be done, we traverse the tree a lot when creating entries what can be avoided.

lpranam:
it takes almost 7 seconds to load approx 600 entries. rendering tree is most time 94%.

eszkadev:
could you provide profile? what is the biggest thing? https://github.com/CollaboraOnline/online/blob/master/browser/src/control/jsdialog/Widget.TreeView.ts#L151C2-L151C18 ?

eszkadev:
what I see is: https://github.com/CollaboraOnline/online/blob/d4a368ff54f60fcfd2199eca43ffa109e90c00dd/browser/src/control/Control.Zotero.js#L1086C2-L1086C17

which rebuilds the tree just when chunk of data arrives: ` that.map.fire('jsdialogupdate', dialogUpdateEvent);`

I think you want to split: fetching and UI update.

1. load initial small data
2. build initial tree
3. (if it is full "page" already, not fetched all data):
3a. show progressbar and in the background fetch the rest of data, page by page
3b. update progressbar, prev page, next pzage buttons when chunk of data is fetched (not the tree), progressbar and buttons are separate widgets
4. after all data is downloaded: refresh tree UI

lpranam:
Currently, PR has one issue, if there is already one snake bar showing, the progress bar is delayed which then causes problems in showing progress or closing it.

browser/src/control/Control.Zotero.js
```diff
@@ -27,6 +27,9 @@ L.Control.Zotero = L.Control.extend({
 			});
 		} else {
 			return fetch(url).then(function (response) {
+				if (!response.ok)
+					return that.getCachedOrFetch(url);
```
eszkadev, line 31: what it does? when request fails it will repeat again and again I assume?
Possibly we want to stop after single retry?

browser/src/control/Control.Zotero.js
```diff
@@ -426,6 +457,55 @@ L.Control.Zotero = L.Control.extend({
 		});
 	},
 
+	enableDisableNextPrevious: function(hide) {
+		this.map.fire('jsdialogupdate', {
+			data: {
+				jsontype: 'dialog',
+				action: 'update',
+				id: 'ZoteroDialog',
+				control: {
```
eszkadev, line 466: can we create helper function: `updateControl(controlJSON)` ?
all the time we repeat few lines, you need to pass only internal "control"

eszkadev: (can be used in many places in this file)

browser/src/control/Control.Zotero.js
```diff
@@ -736,7 +822,35 @@ L.Control.Zotero = L.Control.extend({
...
+			this.totalItems = -1;
+			this.totalPages = -1;
+			this._currentPage = 0;
+			fetch('https://api.zotero.org/users/'+ this.userID +'/items/top?v=3&key=' + this.apiKey + '&limit=1') //find total number of entries
+			.then(function (response) {
+				that.map.uiManager.showProgressBar(_("Loading references"),);
+				that.showItemsForUrl('https://api.zotero.org/users/' + that.userID + '/items/top' + that.getZoteroItemQuery(15, 0), true);
+
+				that.totalItems = parseInt(response.headers.get('Total-Results'));
+				that.totalPages = Math.ceil(that.totalItems/15);
```
eszkadev, line 834: can we use variable for that `15` which is used in many places? so if we want to change number of rows it will be single place

eszkadev:
looks good, added some code style proposals

PR #11121, opened 2025-02-10
each subforkit presets dir should be separate

caolanm:
Change-Id: I501ebbfd75379f35ee33d4b3044d79f98b614058

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
we need to have a separate dir for each subforkit config, we ended up sharing one between them

Rash419 (approved):
Looks good!
Thank you

PR #11122, opened 2025-02-10
fix: reset file input after upload to allow re-upload of the same file

codewithvk:
Change-Id: I341de2f1b14aec29eb0ba4c73415ea7b35929fbf

* Resolves: # 
* Target version: master


caolanm (approved):
excellent, thanks

PR #11126, opened 2025-02-10
we might have a HostEntry in the dns cache already

caolanm:
Change-Id: I2a08bd77f837ba3b8a2a57e90e79ce5bb86305e8

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/NetUtil.cpp
+++ b/net/NetUtil.cpp
@@ -89,7 +89,7 @@ std::string HostEntry::errorMessage() const
     return std::string("[" + _requestName + "]: " + errmsg);
 }
 
-HostEntry::HostEntry(const std::string& desc, const char* port)
+HostEntry::HostEntry(const std::string& desc)
     : _requestName(desc)
     , _saved_errno(0)
     , _eaino(0)
@@ -99,7 +99,7 @@ HostEntry::HostEntry(const std::string& desc, const char* port)
     hints.ai_flags = AI_CANONNAME | AI_ADDRCONFIG;
 
     addrinfo* ainfo = nullptr;
-    int rc = getaddrinfo(desc.c_str(), port, &hints, &ainfo);
+    int rc = getaddrinfo(desc.c_str(), nullptr, &hints, &ainfo);
     if (rc != 0)
     {
         setEAI(rc);
@@ -127,13 +127,11 @@ HostEntry::~HostEntry() = default;
 struct DNSCacheEntry
 {
     std::string queryAddress;
-    std::string queryPort;
     HostEntry hostEntry;
     std::chrono::steady_clock::time_point lookupTime;
 
-    DNSCacheEntry(const std::string& address, const std::string& port, const HostEntry& entry, const std::chrono::steady_clock::time_point& time)
+    DNSCacheEntry(const std::string& address, const HostEntry& entry, const std::chrono::steady_clock::time_point& time)
     : queryAddress(address)
-    , queryPort(port)
     , hostEntry(entry)
     , lookupTime(time)
     {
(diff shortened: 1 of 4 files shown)


net/HttpRequest.hpp
```diff
@@ -1426,33 +1426,37 @@ class Session final : public ProtocolHandlerInterface
...
+        ConnectFailCallback origOnConnectFail = _onConnectFail;
+
         if (!isConnected())
         {
-            std::shared_ptr<StreamSocket> socket = connect();
-            if (!socket)
-            {
+            // Intercept onConnectFail to detect connection failure, call the original (if any),
+            // and finish the response to end the poller loop.
+            setConnectFailHandler([this, origOnConnectFail](const std::shared_ptr<http::Session>& rSession) {
```
Ashod, line 1438: If we have to do this, something is likely wrong with the design. (Also see below.)

If we need this to call `_response->finish();`, then the question is: why isn't it called when the connection fails anyway?

I think we should call `onDisconnect()` (or its equivalent) from `callOnConnectFail()`.

The original design is to always call the 'onFinished' handler. We promise this for the `setFinishedHandler()`:

```
    /// Set a callback to handle onFinished events from this session.
    /// onFinished is triggered whenever a request has finished,
    /// regardless of the reason (error, timeout, completion).
    void setFinishedHandler(FinishedCallback onFinished) { _onFinished = std::move(onFinished); }
```

So it's a good idea to do so when we fail to connect as well. We can extend the `Response::State` enum to include `ConnectionError` (we can rename the current `Error` to `ProtocolError`), and always invoke the same `onDisconnect()` and `onFinished`, to let the client do whatever cleanup and/or continuation.

(I'm not sure we ever needed a 'connectFailHandler', as such, rather than `AsyncConnectResult` + onFinished. Perhaps what's missing is the new State above?)

net/HttpRequest.hpp
```diff
@@ -1426,33 +1426,37 @@ class Session final : public ProtocolHandlerInterface
...
             if (checkTimeout(now))
-                return false;
+                break;
 
             const auto remaining =
                 std::chrono::duration_cast<std::chrono::microseconds>(deadline - now);
             poller.poll(remaining);
         }
 
+        setConnectFailHandler(origOnConnectFail);
```
Ashod, line 1459: This isn't exception-safe. Yet another problem with the need to switch handlers, which we should avoid by having a more homogeneous design.

Ashod (commented):
Thanks, @caolanm. I like the direction, but have concerns with switching the callback. I think we should work towards a better design that handles this case seamlessly.

caolanm:
My assumption here with "routing syncRequestImpl connect through asyncDNS" was that the remaining "sync" dns use in wsd/Storage.cpp StorageBase::validate might be amenable to being split up into something where the result in an async callback would be usable so all the remaining sync DNS uses could be dropped. That looks plausible for the uses in wsd/RequestVettingStation.cpp and wsd/wopi/WopiProxy.cpp, but the StorageBase::create use of validate looks much more uncertain to achieve successfully.

Trying an alternative approach of passing a HostEntry to "validate" and bubbling that down from the initial places the uri can come from becomes unwieldy pretty fast and terminates in a bunch of places where have the same issue.

So lets try another alternative.

Ashod:
> So lets try another alternative.

Fair enough. I like the idea of making sure all DNS lookups go through the same cache, which needs to be async. As always, the devil is in the details.

net/NetUtil.cpp
```diff
@@ -184,10 +184,41 @@ class DNSResolver
...
+        result = std::make_shared<HostEntry>(hostEntry);
+        lock.unlock();
+        cv.notify_one();
+    };
+
+    std::unique_lock<std::mutex> lock(mutex);
+
+    AsyncDNS::lookup(addressToCheck, callback, dumpState);
+
+    cv.wait(lock, [&result]{ return static_cast<bool>(result); });
```
Ashod, line 215: No need to cap this with timeout?

Ashod (approved):
Thanks, @caolanm. Looks good.

net/NetUtil.cpp
caolanm: I don't think so.  Maybe I'll be shown wrong here.

PR #11131, opened 2025-02-11
UILog save/load statistics

aszucs3:
load / save is logged with time duration, filesize, and extension like this:

kit=001 time=0.594 dur=0.534 user=4 rep=1 cmd:load size=10377 ext=docx

there are some separarate "save" for different saving: "save"
"savebg" - save in background
"saveas"
"exportas"

reporting script is not final, it just list all the load/save It will need a better statistics with better charts, but that can be done later.

Change-Id: I67ab4bf4591aca4ec03d9167a26b13cbc6b6aa07

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm (approved):
I think that looks reasonable

Issue #10980, opened 2025-01-21
perf: Double rendering


mmeeks:
Open a spreadsheet, hit page-down a few times - notice some tiles have been rendered twice.

We should save that double rendering. I believe this is because we send a tilecombine that is interpreted as being a request to force rendering of a tile - and the JS client shoudl not have that 'power' - it should be sent a cached version.

Meanwhile DocumentBroker needs a way to force a new key-frame, and I tihnk the JS does this by accident, and sometimes these race unhelpfully.

So the 'oldwid=' thing needs adapting / filtering from the JS side to avoid this in the Kit/ code =)

Thanks !

eszkadev:
I have initial PR for that: https://github.com/CollaboraOnline/online/pull/11134
seems to not hit 3 or more tiles on page down as before in manual testing :)

mmeeks:
Looks good - approved; but would be good to have Caolan's comment closed in there =) thanks !

Darshan-upadhyay1110:
Fixed by PR #11134 'tiles: duplicated rendering' (merged 2025-02-12)

PR #11134, opened 2025-02-11
tiles: duplicated rendering

eszkadev:
Related to #10980 
Check if we have pending rendering before requesting new one for the same tile.
Add some useful logging for rare cases in TRC.
Client only "can" request rendering keyframe by wid=0 when tile is not in cache.
Client cannot force it when we already have correct one.

Manual test:
1. open calc
2. open debug tools with tile overlays and invalidations
3. hold page down until reaching missing tiles (red color)
Check what we received, ideally would be to just get single tile as it wasn't edited.

5-after-page-down

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -4405,9 +4405,7 @@ void DocumentBroker::handleTileCombinedRequest(TileCombined& tileCombined, bool 
         {
             if (!cachedTile || tooLarge)
                 tile.forceKeyframe();
-            tilesNeedsRendering.push_back(tile);
-            _debugRenderedTileCount++;
-            tileCache().subscribeToTileRendering(tile, session, now);
+            requestTileRendering(tile, /*forceKeyFrame*/ true, now, tilesNeedsRendering, session);
         }
     }
     if (hasOldWireId)
@@ -4569,6 +4567,30 @@ void DocumentBroker::handleMediaRequest(std::string range,
     }
 }
 
+bool DocumentBroker::requestTileRendering(TileDesc& tile, bool forceKeyframe,
+                                          const std::chrono::steady_clock::time_point &now,
+                                          std::vector<TileDesc>& tilesNeedsRendering,
+                                          const std::shared_ptr<ClientSession>& session)
+{
+    bool allSamePartAndSize = true;
+    if (!tileCache().hasTileBeingRendered(tile, &now) || // There is no in progress rendering of the given tile
+        tileCache().getTileBeingRenderedVersion(tile) < tile.getVersion()) // We need a newer version
+    {
+        tile.setVersion(++_tileVersion);
+        if (forceKeyframe)
+        {
+            LOG_TRC("Forcing keyframe for tile was oldwid " << tile.getOldWireId());
+            tile.setOldWireId(0);
+        }
+        allSamePartAndSize &= tilesNeedsRendering.empty() || tile.sameTileCombineParams(tilesNeedsRendering.back());
+        tilesNeedsRendering.push_back(tile);
+        _debugRenderedTileCount++;
+    }
+
+    tileCache().subscribeToTileRendering(tile, session, now);
+    return allSamePartAndSize;
+}
+
 void DocumentBroker::sendRequestedTiles(const std::shared_ptr<ClientSession>& session)
 {
     ASSERT_CORRECT_THREAD();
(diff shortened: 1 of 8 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -4567,6 +4570,30 @@ void DocumentBroker::handleMediaRequest(std::string range,
...
+{
+    bool allSamePartAndSize = true;
+    if (!tileCache().hasTileBeingRendered(tile, &now) || // There is no in progress rendering of the given tile
+        tileCache().getTileBeingRenderedVersion(tile) < tile.getVersion()) // We need a newer version
+    {
+        tile.setVersion(++_tileVersion);
+        if (forceKeyframe)
+        {
+            LOG_TRC("Forcing keyframe for tile was oldwid " << tile.getOldWireId());
+            tile.setOldWireId(0);
```
caolanm, line 4586: forceKeyframe() is just { setOldWireId(0); }, so if we replace setOldWireId(0) with that alias, then it becomes more clear that in the call from DocumentBroker::handleTileCombinedRequest that the forceKeyFrame argument to requestTileRendering doesn't really matter, that tile is already forceKeyframe()-ed by handleTileCombinedRequest if it passes forceKeyframe true to here

wsd/DocumentBroker.cpp
```diff
@@ -4401,11 +4401,14 @@ void DocumentBroker::handleTileCombinedRequest(TileCombined& tileCombined, bool
         bool tooLarge = cachedTile && cachedTile->tooLarge();
         if(!cachedTile || !cachedTile->isValid() || tooLarge)
         {
+            bool forceKeyFrame = false;
             if (!cachedTile || tooLarge)
+            {
+                forceKeyFrame = true;
```
mmeeks, line 4407: Looks sensible to me =) I was worried about the race between the kit and coolwsd but we request this tile anyway I guess.

wsd/DocumentBroker.cpp
```diff
@@ -4567,6 +4570,30 @@ void DocumentBroker::handleMediaRequest(std::string range,
...
 }
 
+bool DocumentBroker::requestTileRendering(TileDesc& tile, bool forceKeyframe,
+                                          const std::chrono::steady_clock::time_point &now,
+                                          std::vector<TileDesc>& tilesNeedsRendering,
+                                          const std::shared_ptr<ClientSession>& session)
+{
+    bool allSamePartAndSize = true;
+    if (!tileCache().hasTileBeingRendered(tile, &now) || // There is no in progress rendering of the given tile
+        tileCache().getTileBeingRenderedVersion(tile) < tile.getVersion()) // We need a newer version
```
mmeeks, line 4580: I'm not so sure that version is the right thing here - but it can't do that much harm I think. Looks like we will re-render but send an empty delta.

wsd/DocumentBroker.cpp
mmeeks: yes =)

wsd/DocumentBroker.cpp
```diff
@@ -4610,20 +4637,7 @@ void DocumentBroker::sendRequestedTiles(const std::shared_ptr<ClientSession>& se
...
-                    {
-                        LOG_TRC("Forcing keyframe for tile was oldwid " << tile.getOldWireId());
-                        tile.setOldWireId(0);
-                    }
-                    allSamePartAndSize &= tilesNeedsRendering.empty() || tile.sameTileCombineParams(tilesNeedsRendering.back());
-                    tilesNeedsRendering.push_back(tile);
-                    _debugRenderedTileCount++;
-                }
-                tileCache().subscribeToTileRendering(tile, session, now);
+                allSamePartAndSize &= requestTileRendering(tile, !cachedTile, now, tilesNeedsRendering, session);
```
mmeeks, line 4640: nice to split out =)

mmeeks (approved):
Looks fine for me on red - would be good to tweak re: Caolan's comment.

wsd/DocumentBroker.cpp
eszkadev: This is directly moved from `DocumentBroker::sendRequestedTiles`.

I assume it was thought to bump version (line below) so `getTileBeingRenderedVersion` will be equal with the next potential request while this one is pending.

That will ignore it and not allow to render the same twice. Then we still subscribe to it - so after finished render we will send it.

mmeeks: Yep all good =)

mmeeks:
Looking good to me for an early merge on red - thanks Szymon =) !

wsd/DocumentBroker.cpp
eszkadev: I checked the version thing and it was duplicated as I missed previous bump.
Followup: https://github.com/CollaboraOnline/online/pull/11171

PR #11136, opened 2025-02-11
wsd: better CheckFileInfo validation

Ashod:
We now validate the CheckFileInfo response
immediately after receiving it, and we mark
it as Fail if it doesn't pass.

This results in the prompt closing of the
client connection, with an equally immediate
"Document loading failed" message to the user
with "Unauthorized" in the message (though the
message refers to the host).

The logs show:
```
ERR  #30: BaseFileName should be the name of the file without a path, but is: ../../about21.odt
ERR  #30: WOPI::CheckFileInfo (24ms) failed or no valid JSON payload returned. Access denied.
DBG  #31: WOPI::CheckFileInfo failed, sending error and closing connection now
```

And the connection is closed via sendUnauthorizedErrorAndShutdown().

Change-Id: I1e706c7d12e66452476a90631b1521fc93d44e63

### DIFF
--- a/wsd/wopi/CheckFileInfo.cpp
+++ b/wsd/wopi/CheckFileInfo.cpp
@@ -121,7 +121,7 @@ void CheckFileInfo::checkFileInfo(int redirectLimit)
             return;
         }
 
-        if (JsonUtil::parseJSON(wopiResponse, _wopiInfo))
+        if (parseResponseAndValidate(wopiResponse))
         {
             LOG_DBG("WOPI::CheckFileInfo ("
                     << callDurationMs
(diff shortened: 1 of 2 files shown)


caolanm:
is a blank BaseFileName allowed or abnormal?

Ashod:
> is a blank BaseFileName allowed or abnormal?

We consider it invalid. 

`StorageBase::FileInfo` has this:

```
        bool isValid() const
        {
            // 0-byte files are valid; LO will open them as new docs.
            return !_filename.empty();
        }
```

caolanm (approved):
Yeah, I'm just checking if a new behaviour of a fail on an empty BaseFileName is desirable, and it appears so.

PR #11138, opened 2025-02-11
Update all custom button style rules during insert new one

GulsahKose:
When we add a new custom button it resets all top toolbar and builds again. So we lost the style changes of the old custom buttons and have only last inserted one.

To fix that we have to update old custom button style too.

Change-Id: I26188ff8af99dfe07484fd4c8b00edeed81c562e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
I didn't test but code looks good

PR #11140, opened 2025-02-11
cool#11064 kit: send the client's visible area during file load

vmiklos:
Large enough documents load significantly faster in desktop Writer than
in COOL for some reason.

Investigating the reason, it turns out that the layout mechanism to
layout the visible area (first page, typically) of the document as part
of doc load goes wrong in the LOK case, where nominally the entire
document is visible.

Fix the problem by sending the approximate client visible area (in
twips, assuming 100% zoom) to core as a file load parameter.

This helps a lot for a 300 pages long document.
lok::Office::documentLoad() cost for the bugdoc:
- before: 540 ms
- after: 112 ms

Change-Id: I10acbc571ee5119f6e69b5e8012cb2d422a8de4b

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -1684,6 +1684,14 @@ function getInitializerClass() {
 		return global.makeDocAndWopiSrcUrl(httpURI, docUrlParams, suffix, wopiSrcParam);
 	};
 
+	global.makeClientVisibleArea = function() {
+		// An approximation till we don't yet have CanvasTileLayer, which would properly use
+		// map.getPixelBounds() and pixelsToTwips().
+		const width = window.innerWidth * 15;
+		const height = window.innerHeight * 15;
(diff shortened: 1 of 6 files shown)


vmiklos:

The first commit is the COOL side of the "set view before load" work; turns out we have 2 places to load the doc, so update both to do do this. We don't have a map object with all the usual functionality, but an approximation of the view looks doable.

The second is a fix for a CI problem to get it green again. I'm not too happy about just increasing the tolerance, but otherwise this would block all PRs, which is worse. CC @eszkadev for that part.

vmiklos:
Ah hm, cypress desktop was green at https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/4571/, but an earlier try finished later, so we're now red, let me retry, sorry for the too early ping.

PR #11141, opened 2025-02-11
Private/mmeeks/harderkill2

mmeeks:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -198,7 +198,7 @@ public:
                   else
                   {
                       // Failed!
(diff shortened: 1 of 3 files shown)


mmeeks:
So - this should take us down in the pathological case of socket confusion and non-closing of sockets - and at least stop the processes stacking up and killing us :-)

But it is not a fix - it's a backup; we need a proper fix.

Ashod (approved):
Thanks, @mmeeks. One not related to the new log entry.

PR #11142, opened 2025-02-12
wsd: exit the bgsave process once saved

Ashod:
Once saving is complete, there little
point for the bgsave process to stick
around. Best to exit promptly.

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1395,7 +1395,8 @@ void Document::handleSaveMessage(const std::string &)
         // cleanup any lingering file-system pieces
         _loKitDocument.reset();
 
-        // Next step in the chain is BgSaveChildWebSocketHandler::onDisconnect
+        UnitKit::get().preBackgroundSaveExit();
+        Util::forcedExit(EX_OK);
     }
 }
 


mmeeks (changes requested):
The socket content needs to be flushed - we need the SocketPoll to run for that - or we get stuck saves - surely.

Ashod:
> The socket content needs to be flushed - we need the SocketPoll to run for that - or we get stuck saves - surely.

I do see it flushed, _reliably_. At least in my local tests:
```
[ kitbgsv_001_007 ] TRC  postMessage forwarding to parent of save process: client-002 unocommandresult: { "commandName": ".uno:Save", "success": true, "result": { "type": "boolean"}, "wasModified": false, "startUnixTimeMics": 542323680131, "saveDurationMics": 205057}| kit/Kit.cpp:810
[ kitbgsv_001_007 ] TRC  Check save message| kit/Kit.cpp:1356
[ kitbgsv_001_007 ] TRC  BgSave completed| kit/Kit.cpp:1361
[ kitbgsv_001_007 ] TRC  Shutting down bgsv child's socket to parent kit post save| kit/Kit.cpp:1370
...
[ kitbroker_001 ] DBG  #26: Closed but will drain incoming data per POLLIN| net/Socket.hpp:1504
[ kitbroker_001 ] DBG  #26: bgsv_kit_ws: recv from parent [client-002 unocommandresult: { "commandName": ".uno:Save", "success": true, "result": { "type": "boolean"}, "wasModified": false, "startUnixTimeMics": 542323680131, "saveDurationMics": 205057}| kit/KitWebSocket.cpp:261 
...
[ kitbgsv_001_007 ] INF  Forced Exit with code: 0| common/Util.cpp:723
```

But, I can restore the two-phase closing, with shutdown, and resolve the other inconsistencies.
It should give us more confidence that flushing is certainly done.

PR #11144, opened 2025-02-12
wsd: better CheckFileInfo validation

caolanm:
We now validate the CheckFileInfo response
immediately after receiving it, and we mark
it as Fail if it doesn't pass.

This results in the prompt closing of the
client connection, with an equally immediate
"Document loading failed" message to the user
with "Unauthorized" in the message (though the
message refers to the host).

Change-Id: I1e706c7d12e66452476a90631b1521fc93d44e63

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/Storage.cpp
+++ b/wsd/Storage.cpp
@@ -624,6 +624,26 @@ http::Request WopiStorage::initHttpRequest(const Poco::URI& uri, const Authoriza
     return httpRequest;
 }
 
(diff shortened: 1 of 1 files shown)


timar:
I don't know why CI fails. I assume it's not related to this patch. I can't wait to drop this old branch in May.

PR #11145, opened 2025-02-12
impress: fix navigator sidebar icon missing width and height

banobepascal:
Change-Id: I84239134f959e62fb6cfe7bea62bba782593bf29

* Target version: master 

### Summary
These properties were missing making the branding icon invisible in light mode

### Before
Screenshot from 2025-02-12 12-47-28

### After change
Screenshot from 2025-02-12 13-10-37

### Checklist

### DIFF
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -321,7 +321,7 @@
 	background-position: right;
 	background-repeat: no-repeat;
 }
-.unospan-optionstoolboxdown .unobutton > img {
+#optionstoolboxdown .unotoolbutton .unobutton > img {
 	width: var(--btn-img-size);
 	height: var(--btn-img-size);
 }


browser/css/toolbar.css
```diff
@@ -321,6 +321,7 @@
 	background-position: right;
 	background-repeat: no-repeat;
 }
+.unospan-options-navigator .unobutton > img,
 .unospan-optionstoolboxdown .unobutton > img {
```
eszkadev, line 325: maybe we can prepare some generic rule? not target every element 1 by 1?

eszkadev:
please do not use merge commits (remove it and rebase instead)

browser/css/toolbar.css
banobepascal: The notebookbar `impress` was created differently from `writer|calc|draw`
Check 

https://github.com/CollaboraOnline/online/blob/cf23fdc2851732a45f94fe6f37f37af967b34738/browser/src/control/Control.Notebookbar.js#L548

https://github.com/CollaboraOnline/online/blob/cf23fdc2851732a45f94fe6f37f37af967b34738/browser/src/control/Control.NotebookbarImpress.js#L66

eszkadev: both start with `options`
`[id^="options"]`

banobepascal: Used `id="optionstoolboxdown"` since it is the container of the toolbox that encompases the navigator icon across all

eszkadev (approved):

PR #11146, opened 2025-02-12
Private/caolan/close foo

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks, @caolanm. We had agreed to revert the shutdown->close change, but keep the fd invalidation logic. This is good, we can have the fd invalidation done on top of this, separately.

mmeeks:
@timar I suspect this needs back-porting too.

PR #11147, opened 2025-02-12
browser: replace the 'title' attribute with 'cool-tips'.

hcvcastro:
Change-Id: I4e7259ba81665c603bd6e00f9e959d869653fb59

### DIFF
--- a/browser/src/control/Control.StatusBar.js
+++ b/browser/src/control/Control.StatusBar.js
@@ -419,7 +419,7 @@ class StatusBar extends JSDialog.Toolbar {
 		if (permissionContainer) {
 			while (permissionContainer.firstChild)
 				permissionContainer.removeChild(permissionContainer.firstChild);
-			permissionContainer.appendChild(getPermissionModeElements(isReadOnlyMode, canUserWrite));
+			permissionContainer.appendChild(getPermissionModeElements(isReadOnlyMode, canUserWrite, this.map));
 		}
 
 		this.builder.updateWidget(this.parentContainer, {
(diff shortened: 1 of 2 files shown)


gokaysatir (commented):
I think we can use the functions without sending the map pointer. We could use `app.map` pointer i guess?

If you agree that we don't need to add map parameter to functions: We can prepare a follow up or modify this commit?

Other than that, code looks good to me. I also tested.

hcvcastro:
> I think we can use the functions without sending the map pointer. We could use `app.map` pointer i guess?
> 
> If you agree that we don't need to add map parameter to functions: We can prepare a follow up or modify this commit?
> 
> Other than that, code looks good to me. I also tested.
> 
> Thanks for the improvements :)

Yes, I know, The code was adapted according to builder object to JSDialog(Builder). Now this code, it does not have
a builder object. Using app.map it is good alternative. It guess, it is a global singleton. I do not like to a class depend
on a global singleton, also attaching mouse events, but hopefully it is a very finite control elements.

eszkadev (commented):
looks good to me

gokaysatir:
> > I think we can use the functions without sending the map pointer. We could use `app.map` pointer i guess?
> > If you agree that we don't need to add map parameter to functions: We can prepare a follow up or modify this commit?
> > Other than that, code looks good to me. I also tested.
> > Thanks for the improvements :)
> 
> Yes, I know, The code was adapted according to builder object to JSDialog(Builder). Now this code, it does not have a builder object. Using app.map it is good alternative. I guess, it is a global singleton. I do not like to a class depend on a global singleton, also attaching mouse events, but hopefully it is a very finite control elements.

Thanks for the point of view :)

gokaysatir (approved):
Thank you for the improvements.

vmiklos:
https://github.com/CollaboraOnline/online/pull/11193 has to go in first, then please rebase. Sorry about the trouble.

PR #11148, opened 2025-02-12
use template url correctly on document load

JaumePujantell:
When creating an ODF file from an OOXML template loading fails. This
happens because the ODF url is given to LOKit instead of the template
url and it deduces the wrong loading format from it.

Change-Id: I4f30376f1c53548307f429b8091201656c3e059a


kit/Kit.cpp
```diff
@@ -1939,13 +1940,26 @@ std::shared_ptr<lok::Document> Document::load(const std::shared_ptr<ChildSession
             | LOK_FEATURE_VIEWID_IN_VISCURSOR_INVALIDATION_CALLBACK;
         _loKit->setOptionalFeatures(flags);
 
+        std::string loadUri = uri;
+
+        if (!docTemplate.empty())
```
vmiklos, line 1945: Could you please add a comment here on how this works? Here is how I understand it (but probably I'm wrong):

1) 'uri' is the new document, say doc.pptx and 'docTemplate' is the template, say template.potx
2) now you tweak newPath to be be doc.potx instead of doc.pptx.

Shouldn't this be the other way around? Thanks.

vmiklos (commented):
This may be OK, but I'm not sure I fully understand your changes. See the inline comment. :-)

kit/Kit.cpp
vmiklos: This is done now, thanks.

kit/Kit.cpp
```diff
@@ -1939,13 +1940,30 @@ std::shared_ptr<lok::Document> Document::load(const std::shared_ptr<ChildSession
...
+        {
+            // The template has been downloaded to 'uri'
+            // But since the template might have a different format we temporarily
+            // change the url to have the correct extension
+            // It will be saved back to 'uri' in ChildSession once loaded
```
vmiklos, line 1950: Aha, the template is already copied to 'uri' at this point, OK.

kit/Kit.cpp
```diff
@@ -1939,13 +1940,30 @@ std::shared_ptr<lok::Document> Document::load(const std::shared_ptr<ChildSession
...
+            rename(pocoUri.getPath().c_str(), newPath.toString().c_str());
+            pocoUri.setPath(newPath.toString());
+            loadUri = pocoUri.toString();
+        }
+
         // Save the provided password with us and the jailed url
         _haveDocPassword = haveDocPassword;
         _docPassword = docPassword;
-        _jailedUrl = uri;
+        _jailedUrl = loadUri;
```
vmiklos, line 1963: And seeing that ChildSession always uses the _jailedUrl when it calls copyForUpload(), the above rename() should be no problem.

vmiklos (approved):
OK, now with the comment the code is a bit confusing, but it does make sense. :-) Please merge it yourself when CI is green. Thanks.

PR #11155, opened 2025-02-13
Fix an issue in LOUtil._doRectanglesIntersect() caught by mocha

dennisfrancis:
* Target version: master 

### Summary
Fix an issue in LOUtil._doRectanglesIntersect() caught by mocha. Also introduce unit tests for LOUtil._getIntersectionRectangle().

### Checklist


eszkadev (approved):
good catch!

PR #11156, opened 2025-02-13
cypress: make calc/clipboard_spec.js more stable

eszkadev:


cypress_test/integration_tests/desktop/calc/clipboard_spec.js
```diff
@@ -56,19 +60,21 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Calc clipboard tests.', fu
...
 		helper.typeIntoInputField(helper.addressInputSelector, 'C1');
+		cy.cGet(helper.addressInputSelector).should('have.prop', 'value', 'C1');
 		cy.window().then(win => {
 			var app = win['0'].app;
 			app.socket.sendMessage('uno .uno:Copy');
 		});
 		var html = '<div id="meta-origin" data-coolorigin="%META_URL%">ignored</div>';
 		setDummyClipboard('text/html', html);
 
+		cy.wait(200);
```
vmiklos, line 73: Is this a wait so that .uno:Copy is processed by the time we're post-wait?

If so, I wonder if we could have a command dispatch listener here instead, to avoid the unconditional wait.

But sure, probably I introduced a race here (which was not a problem in practice before), so this is still a net improvement. :-)

vmiklos (approved):
Looks good in general, see inline comment for a nit, it would be ideal to listen for command completion instead of a hardcoded wait, I suspect.

cypress_test/integration_tests/desktop/calc/clipboard_spec.js
eszkadev: yes, I also think we need some listeners. Not sure if thst is the issue as cannot reproduce locally.
I want to check if this test will not appear anymore as failed in next days.

PR #11157, opened 2025-02-13
Introduce an event delay watchdog debugging feature

Cwiiis:
Add the ability to start a watchdog timer on event delays when processing events to track when large, unexpected event delays may happen.

Change-Id: I7a73069fade52b3eca6f8793e43ee52bf9087c9a

### DIFF
--- a/browser/src/core/Debug.js
+++ b/browser/src/core/Debug.js
@@ -374,6 +374,28 @@ L.DebugManager = L.Class.extend({
 		});
 
 		this._addDebugTool({
+			name: 'Event delay watchdog',
+			category: 'Logging',
+			startsOn: true,
+			onAdd: function () {
+				self.eventDelayWatchdog = true;
+				self._eventDelayTimeout = null;
+				self._lastEventDelayTime = 0;
+				self._lastEventDelay = 0;
+			},
+			onRemove: function () {
+				self.eventDelayWatchdog = false;
+				self.clearOverlayMessage('eventDelayTime');
+
+				if (self._eventDelayTimeout) clearTimeout(self._eventDelayTimeout);
+				delete self._eventDelayTimeout;
+				delete self._eventDelayWatchStart;
+				delete self._lastEventDelayTime;
+				delete self._lastEventDelay;
+			},
+		});
+
+		this._addDebugTool({
 			name: 'Typer',
 			category: 'Functionality',
 			startsOn: false,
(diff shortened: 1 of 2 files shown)


Cwiiis:
This is somewhat what @mmeeks, @caolanm, @eszkadev and I were discussing last week - a timeout that gets set when processing events to keep track of how long it takes us to process events. I don't think this is exactly what we were talking about though, or its final form necessarily.

This adds a watchdog timer when processing events and sending messages in the debug features that defaults to on. The event processing delay is displayed at the top-left - if that delay is above a threshold (250ms), it will hold that value there for 3 seconds so that it isn't easily missed (though larger delays are allowed to replace it).

- This doesn't pop up a modal dialog - the more I thought about it, the more this doesn't seem like a great idea... But I do think it needs to be more than just a nondescript message (at least slow delays should be made more visible somehow)
- This doesn't continuously run a timeout, but only runs one on event processing. Perhaps there should be X amount of time on event processing that we continuously run this? I don't know that that necessarily makes sense, I don't feel it does.

This does seem to work - if I trigger a bunch of tile events, you can see the event handling delay grow, and when tiles are on the client, it falls back down again. 250ms might be too gracious a slow-event barrier - anything over 16ms is going to mean guaranteed dropped frames, anything over 50ms is probably going to feel quite sluggish... 250ms is something's-gone-wrong territory, but then maybe that's what we want.

Perhaps 50ms for a slow-event categorisation, but 250ms for a "pop up a modal dialog that something's gone wrong" category?

Anyway, feedback welcome and appreciated.

mmeeks:
Looks great - 50ms is good - I'd like this not to be something that has to be enabled - but is always on in a debug build; I think there is a setting in cool.html.m4 around that - but lets get this in.

Cwiiis:
This will do, I think - it is actually quite useful and gets triggered when I expect it to. I changed the threshold for slow handling to be 50ms and now catastrophically slow (>250ms) will pop up a modal dialog. I couldn't immediately see how to have this enabled by default and didn't want to just hack something in if there's some existing/expected way of doing that - so any advice there welcome.

I think we should also consider lowering that catastrophic threshold to 150ms or even 100ms. I can trigger the 250ms one in Firefox on a 4k screen (where there is a serious performance issue), but even in that case it isn't easy. However, it would be if the trigger was 100ms, which still feels like an unacceptable level of performance. I suppose the trick will be to see if we have any spikes around these thresholds, but I'd consider any blocking for that amount of time to be a bug, personally.

mmeeks:
Awesome 100ms sounds great =) and seems it found useful things for Szymon already we can see on profiles.
Szymon - can you get it in if you're happy - longer it's in there the more fun we find I think.

Would love to have the end-user able to see particular slow-ness particularly of the network as a 2nd thing.

eszkadev (approved):
looks good :)

PR #11158, opened 2025-02-13
add missing "Screen Reading" menu

hcvcastro:
- **wsd: accessbility: fix the compact user interface settings.**
- **browser: accessibility: add missing "Screen Reading" menu for calc**

### DIFF
--- a/wsd/FileServer.cpp
+++ b/wsd/FileServer.cpp
@@ -1769,7 +1769,7 @@ FileServerRequestHandler::ResourceAccessDetails FileServerRequestHandler::prepro
 
     // default to the notebookbar if the value is "default" or whatever
     // nonsensical
-    if (enableAccessibility == "true" || (userInterfaceMode != "classic" && userInterfaceMode != "notebookbar"))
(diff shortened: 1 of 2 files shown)


wsd/FileServer.cpp
```diff
@@ -1753,7 +1753,7 @@ FileServerRequestHandler::ResourceAccessDetails FileServerRequestHandler::prepro
 
     // default to the notebookbar if the value is "default" or whatever
     // nonsensical
-    if (enableAccessibility == "true" || (userInterfaceMode != "classic" && userInterfaceMode != "notebookbar"))
+    if (enableAccessibility == "true" && userInterfaceMode != "classic" && userInterfaceMode != "notebookbar")
```
timar, line 1756: I don't understand this change. You changed the logic, but why?

hcvcastro: If enableAccessbility is true, no way to change the user interface to 'compact', it will always force to interface .

Darshan-upadhyay1110 (commented):
Hii @hcastro Screen reading option is related to accessibility and we do not offer that in compact mode as compact mode is not suppose to bind with accessibility

 maybe the PR is not needed as compact mode is good without the option

PR #11159, opened 2025-02-13
Fix wrong custom button image URL

GulsahKose:
Button image URL was like:
"http://localhost:9980/browser/b776c533b4/images/images/lc_save.svg" That causes the missing icon when new inserted custom buttons After fix, new URL will be like
"http://localhost:9980/browser/b776c533b4/images/lc_save.svg"

Change-Id: Ide8f6df68018fcac98ce95f1a5c4a9317cff3473

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Notebookbar.js
+++ b/browser/src/control/Control.Notebookbar.js
@@ -336,15 +336,18 @@ L.Control.Notebookbar = L.Control.extend({
 		}
 
 		var isUnoCommand = button.unoCommand && button.unoCommand.indexOf('.uno:') >= 0;
+		var forceToUseDefaultIcon = false;
 		if (button.unoCommand && !isUnoCommand)
 			button.unoCommand = '.uno:' + button.unoCommand;
+		if (isUnoCommand && button.imgurl.indexOf('http') === -1)
+			forceToUseDefaultIcon = true;
 
 		this.additionalShortcutButtons.push(
 			{
 				id: button.id,
 				type: 'toolitem',
 				text: button.label ? button.label : (button.hint ? _(button.hint) : ' '),
-				icon: button.imgurl,
+				icon: forceToUseDefaultIcon ? undefined : button.imgurl,
 				command: button.unoCommand,
 				accessKey: button.accessKey ? button.accessKey: null,
 				postmessage: button.unoCommand ? undefined : true,


browser/src/control/Control.Notebookbar.js
```diff
@@ -332,12 +332,14 @@ L.Control.Notebookbar = L.Control.extend({
...
 
+		var imgUrl = button.imgurl.split('/')[1];
+
 		this.additionalShortcutButtons.push(
 			{
 				id: button.id,
 				type: 'toolitem',
 				text: button.label ? button.label : (button.hint ? _(button.hint) : ' '),
-				icon: button.imgurl,
+				icon: imgUrl,
```
eszkadev, line 342: I think this is for custom icons which can come from outside COOL (integrator).
Then we need full URL.

In case of just standard COOL icons: we will use `command` to find icon -> so `icon:` can be empty.
If we want to use standard COOL icon without using `command`: client should provide good input in the PostMessage I think... 

Logic for that is here I think: https://github.com/CollaboraOnline/online/blob/master/browser/src/control/Control.JSDialogBuilder.js#L2287

browser/src/control/Control.Notebookbar.js
```diff
@@ -332,12 +332,14 @@ L.Control.Notebookbar = L.Control.extend({
 		if (button.unoCommand && !isUnoCommand)
 			button.unoCommand = '.uno:' + button.unoCommand;
 
+		var imgUrl = button.imgurl.split('/')[1];
```
eszkadev, line 335: this will break custom icons from integrator

GulsahKose: You are right this is wrong. Most probably we send wrong image url via postMessage frame here:
https://github.com/CollaboraOnline/online/blob/500df64292e53fe07c92d492511ed82bfcb8b546/browser/html/framed.doc.html#L253

We just send lc_save.svg  instead of images/lc_save.svg or leave empty. If it is empty it automatically uses the uno command's default icon.

browser/src/control/Control.Notebookbar.js
GulsahKose: If button.imgurl is undefined it automatically uses the uno command's default icon. So no need to fix anything here.

browser/src/control/Control.Notebookbar.js
eszkadev: Great, could you fix it then?

eszkadev: and check if SDK tell something about it? do we have any wrong information there?

GulsahKose: I'm testting Szymon. Let me write down my detections.

We don't let the handle post message without imgurl here https://github.com/CollaboraOnline/online/blob/3d6065175382afd3b83c4f5e9221677b225942e4/browser/src/map/handler/Map.WOPI.js#L445 
so it never be undefined in this case.

If I change the images/lc_save.svg  to lc_save.svg  tabbed view is ok but compact view is nok ok. 
In the compact view we should see img style="background: url(&quot;images/lc_save.svg&quot;) center center no-repeat;" 

So I'll check SDK then change frame.html, compact or tabbed view url handling...

GulsahKose: @eszkadev I solved it. I didn't touch the compact view and frame.doc.html 

I just forced the tabbed view to use default button icon, if command is an uno command and url is an internal path. So It won't break if icon url is absolute path "http...."

PR #11160, opened 2025-02-13
cool#11064 any input: add page insert testcase

vmiklos:
- **cool#11064 any input: add page insert testcase**


vmiklos:

This is just a test for an earlier fix, as adding this new type of test took a bit of time, so I didn't do it together with the fix.

caolanm (approved):
adding more tests is good

PR #11161, opened 2025-02-13
CheckFileInfo created without std::shared_ptr

caolanm:
and since:

commit e0ca4a82328a6750c5ab99dde64b3000b4e11f17
CommitDate: Tue Jan 14 08:31:11 2025 +0000

    use std::weak_ptr's of CheckFileInfo to detect callbacks after end of life

we rely upon creation via std::shared_ptr to detect callbacks that outlive CheckFileInfo

and so richdocumentscode didn't work

Change-Id: I1feb7859cc9d078ac70833a230c49ed964ef43e7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
We need to create CheckFileInfo via std::shared_ptr to make the referenced commit work, I missed this one as I was focused on changing std::unique to std::shared and missed the create on stack option used here that ends up being needed from richdocumentscode

timar (approved):

PR #11164, opened 2025-02-14
net: minimize certif verif enum size

meven:
### Summary

Noticed by clang-tidy:

```
[clang-tidy] (performance-enum-size) Enum 'CertificateVerification' uses a larger base type ('int', size: 4 bytes) than necessary for its value set, consider using 'std::uint8_t' (1 byte) as the base type to reduce its size
```
Saves 3 bytes.


mmeeks (commented):
@grandinj would be a much better reviewer - and I think has some clang plugins that warn about struct/class ordering and packing which is likely to be much more useful than ad-hoc one-line patches to do this sort of thing I think. Noel - can you review this one and/or advise Meven on this ? =) I looked at 'StreamSocket' and there are a load of badly arranged booleans that we should fix in 25.04 I think.

We tend not to use std:: for uint types I think,

grandinj (approved):
LGTM

PR #11165, opened 2025-02-14
coplugin: add refcounting check

vmiklos:
Finds:

  CXX      wsd/DocumentBroker.o
wsd/DocumentBroker.cpp:1035:27: error: instance allocated on the stack, create it with std::shared_ptr [coplugin:refcounting]
            CheckFileInfo checkFileInfo(poller, session->getPublicUri(), [](CheckFileInfo&) {});
                          ^~~~~~~~~~~~~
1 error generated.

With 67504c45efeb36a8d91b1ce85b27af8cf257e051 (CheckFileInfo created
without std::shared_ptr, 2025-02-13) reverted.

Change-Id: Ic702d5ab3cd08c0de54df5edc61db8d5c50834fb

### DIFF
--- a/clang/Makefile
+++ b/clang/Makefile
@@ -12,7 +12,8 @@ run:
 	clang++ -Werror -Xclang -load -Xclang ./plugin.so -Xclang -add-plugin -Xclang coplugin -c -o test/capture.o test/capture.cpp
 
 check:
-	clang++ -Werror -Xclang -verify -Xclang -load -Xclang ./plugin.so -Xclang -add-plugin -Xclang coplugin -c -o test/capture.o test/capture.cpp
+	clang++ -Werror -Xclang -verify -Xclang -load -Xclang ./plugin.so -Xclang -add-plugin -Xclang coplugin -c test/capture.cpp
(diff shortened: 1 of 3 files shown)


vmiklos:
@caolanm could you please review this when you're back? Thanks.

I was primarily curious if there are any other instances of the same problem in the code, and luckily we don't have anything like that. Then misused this opportunity to add a 2nd sub-plugin, so it's a bit easier to see what does it take to add a new one.

grandinj:
We should copy rtl::Reference and SimpleReferenceBase into COOL, std::shared_ptr/weak_ptr is abysmally slow.

grandinj:
Also, if we use rtl::Reference and SimpleReferenceBase, we can extend this check to anything that subclasses SimpleReferenceBase, and the coverage will automatically extend to new things that we are managing with ref-counting.

vmiklos:
Are you interested in adding a base class (template class) that does this and use it in one place (where it helps) as an example?

I don't know much about performance and it's news to me that shared_ptr is slow for our use-cases. Thanks.

PR #11167, opened 2025-02-14
following: stop when user clicks

eszkadev:
Regression from commit 8fa9d5521333e52e6da4e162a42f3508451def07 UserList: do not render when not changed

Bug:
1. open calc with 2 sessions
2. user A starts following user B
3. user A clicks on a cell
4. user A types Result: following chip didn't disappear in the top toolbar Expected: following chip dissapears

Bug caused by using the function which only setups a state of following. We need to notify the user list too.

For now it's not a good approach... we need to be careful. We should move all the following code to single place which will notify the UI automatically so client code doesn't need to remember it.


vmiklos (approved):
Looks good to me.

PR #11168, opened 2025-02-14
Impress: Return silently if a comment is being edited.

gokaysatir:
While importing comments, confirmation popup causes false alerts. Issue: Because when user's comment is autosaved, import comments is activated and model pops up.

Change-Id: I61d75d4ee5c9343f14edb86630c9b5e49b29a7ec

### DIFF
--- a/browser/src/canvas/sections/CommentListSection.ts
+++ b/browser/src/canvas/sections/CommentListSection.ts
@@ -2366,17 +2366,6 @@ export class CommentSection extends app.definitions.canvasSectionObject {
 		this.disableLayoutAnimation = true;
 		var comment;
 		if (Comment.isAnyEdit()) {
-			this.map.uiManager.showConfirmModal(
-				'comments-update',
-				_('Comments Updated'),
-				_('Another user has updated comments. Would you like to proceed updating?'),
-				_('OK'),
-				() => {
-					CommentSection.pendingImport = false;
-					this.clearList();
-					this.importComments(commentList);
-				}
-			);
 			CommentSection.pendingImport = true;
 			return;
 		}


browser/src/canvas/sections/CommentListSection.ts
```diff
@@ -2366,17 +2366,6 @@ export class CommentSection extends app.definitions.canvasSectionObject {
...
-				_('Comments Updated'),
-				_('Another user has updated comments. Would you like to proceed updating?'),
-				_('OK'),
-				() => {
-					CommentSection.pendingImport = false;
-					this.clearList();
-					this.importComments(commentList);
-				}
-			);
 			CommentSection.pendingImport = true;
```
eszkadev, line 2369: but it will not import if it will be like that, right?
It will just return without importing.

I think the reason for that dialog was - if we edit a comment - we want to be able to copy content before reload.
And the reload happened in the callback. So we didn't want to continue.

I assume `pendingImport` is also not needed without that dialog?

gokaysatir: With this fix:
     If a comment is being edited at the time of re-import, importing is postponed. After the comment edit is done, there will be a comment notification from core side. With that notification "pendingImport" will work and will re-import all the comments.

Without the fix:
     User's own comment's notification is falsely handled if we show the popup.

eszkadev: I see. It would be nice to put here comment which will describle what happens next (which message arrices, which function will be called). Now it's very not obvious.
Also why we do this.

eszkadev (approved):
I didn't test this
cc @lpranam

lpranam:
I am not able to reproduce the issue

PR #11171, opened 2025-02-15
tiles: avoid duplicated version bump

eszkadev:
Related to issue #10980
Followup for 6ae718e
tiles: client cannot force keyframe

Version can protect us from rendering the same tile twice. The increase was introduced in commit 5260eae05f89056d4ffc1bdfad1927c3c5f9706d Avoid rendering / sending the same tile twice


mmeeks (approved):
Looks good to me - nice to keep the versions lower =)

PR #11172, opened 2025-02-15
setting: allow wordbook to edit and remove words from wordbook

codewithvk:
Change-Id: I68b70ee8c140e8e71f312b6120ae880fab46a5d1

* Resolves: # 
* Target version: master 

### Description

This PR enables the edit button for WordBook, which opens a popup dialog allowing you to add or delete words from the dictionary.

It also updates the .dic file upload flow:
	-  Previously, .dic files were uploaded directly without processing, which caused issues.
	-  Now, if a user uploads a file without headers, a default header is appended; if headers are present, they are removed and replaced with our own.

### TODO

- [x] Blocker: error handling for non parsed file - how we should handle it? 
- [x] Blocker: Upload dic file separately? We shouldn't allow to upload headers
- [x] Blocker: Test for all case
- [x] Non-Blocker - Move wordbook to separate class/module
- [ ] Follow-Up - when we load a large file - we should add some kind of loader. 
- [ ] Follow-Up - optimization: cache the file? Maybe an async call to fetch too?

### DIFF
--- a/browser/admin/src/integrator/WordBook.ts
+++ b/browser/admin/src/integrator/WordBook.ts
@@ -1,0 +1,251 @@
+/* eslint-disable */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+interface DicFile {
+	headerType: string; // eg. "OOoUserDict1"
+	language: string; // the value after "lang:" (eg. "<none>")
+	dictType: string; // the value after "type:" (eg. "positive")
+	words: string[]; // list of dictionary words
+}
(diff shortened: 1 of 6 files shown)


browser/admin/src/integrator/WordBook.ts
```diff
@@ -0,0 +1,251 @@
...
+
+interface DicFile {
+	headerType: string; // eg. "OOoUserDict1"
+	language: string; // the value after "lang:" (eg. "<none>")
+	dictType: string; // the value after "type:" (eg. "positive")
+	words: string[]; // list of dictionary words
+}
+
+class WordBook {
+	openDicEditor(fileName: string, dic: DicFile): void {
```
codewithvk, line 20: Let's use a wordbook instead of a dic or dictionary for core standards.

codewithvk: Commit : https://github.com/CollaboraOnline/online/pull/11172/commits/ec8bb9ca1ab0b729be8ad3a84bc0e96f87bb6308

caolanm (approved):
looks reasonable to me

PR #11174, opened 2025-02-16
feat: crop images in impress, calc and draw

Rash419:
* Target version: master 
* requires: https://gerrit.libreoffice.org/c/core/+/181705


gokaysatir (approved):
Thanks @Rash419 :)

PR #11176, opened 2025-02-17
browsersetting: fix: client defaults to use darkTheme with inverted background

Rash419:
- Steps to reproduce:
1. Open document and toggle darkmode
2. Reload the document

- Actual: Notice inverted document background with darkmode
- Expected: Both background and UI should be in darkmode

Change-Id: I22dc8cc45adc0baf98a95947242f927826c0437c


quikee (approved):
Makes sense

PR #11183, opened 2025-02-17
following: don't stop if using scrollbar

eszkadev:
Scrollbar section is on top of document, if we click there - events are propagated. If we follow someone we don't want to stop just by scrolling. Let's not allow to pass events to the document if user clicks on a scrollbar.

```
(anonymous) (Map.Mouse.js:133)
_executeMouseEvents (Map.Mouse.js:196)
setTimeout
(anonymous) (Map.Mouse.js:136)
(anonymous) (global.js:954)
fire (Events.js:153)
_fireDOMEvent (Map.js:1602)
_handleDOMEvent (Map.js:1572)
(anonymous) (global.js:954)
handler (DomEvent.js:52)
```

### DIFF
--- a/browser/src/map/handler/Map.Mouse.js
+++ b/browser/src/map/handler/Map.Mouse.js
@@ -97,6 +97,10 @@ L.Map.Mouse = L.Handler.extend({
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


gokaysatir (approved):
Looks good to me, thanks for the fix.

Issue #11170, opened 2025-02-14
Status bar: Hide on/off Indicators if they have default values


pedropintosilva:
### Describe the Bug

Some indicators are not that important specially if their value has not been changed (is the default). We should hide those in these cases.

#### Writer
![Image](https://github.com/user-attachments/assets/8a2e9dc1-41a4-481e-b34a-13103b04f433)

* Hide:
  * `InsertMode` when the value is `Insert`
  * `showcomments` when the value is `Comments: On`
  * `permissionmode` when the value is `Edit mode`
  * `StatusSelectionMode` when the value is `Standard Selection` (in the future - when we make it a dropdown as in LO - we can make it visible at all times)

PR #11186, opened 2025-02-17
status bar: hide on/off indicators if they have default values

banobepascal:
Change-Id: Iec53e53cc820fe14c3a5657b80e59b5d01b26a5a

* Resolves: #11170 
* Target version: master 

### Summary
Add data-default attribute to status elements to hide indicators when they are in default state:
 - InsertMode when the value is 'Insert'
 - showcomments when the value is 'Comments: On'
 - permissionmode when the value is 'Edit mode'
 - StatusSelectionMode when the value is 'Standard selection'

### Previews
Screenshot from 2025-02-17 22-29-36

### DIFF
--- a/browser/src/control/jsdialog/Widget.HTMLContent.ts
+++ b/browser/src/control/jsdialog/Widget.HTMLContent.ts
@@ -50,18 +50,28 @@ function getPermissionModeElements(
 	} else {
 		permissionModeDiv.classList.add('status-edit-mode');
 		permissionModeDiv.title = _('Permission Mode');
+		permissionModeDiv.setAttribute('default-state', 'true');
 		permissionModeDiv.textContent = _('Edit mode');
 	}
 
 	return permissionModeDiv;
 }
 
-function getStatusbarItemElements(id: string, title: string, text: string) {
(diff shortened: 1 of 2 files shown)


browser/src/control/jsdialog/Widget.HTMLContent.ts
```diff
@@ -83,14 +93,20 @@ function getStatusDocPosElements(text: string) {
 }
 
 function getInsertModeElements(text: string) {
-	return getStatusbarItemElements('InsertMode', _('Entering text mode'), text);
+	return getStatusbarItemElements(
+		'InsertMode',
+		_('Entering text mode'),
+		text,
+		text === 'Insert',
```
eszkadev, line 100: It will not work. We get translated strings

pedropintosilva (commented):
Looks awesome @banobepascal !

But did you check if there is performance differences? I have noticed that some status indicators take too long to appear, but maybe it's unrelated to this patch

gokaysatir (commented):
Looks good but i couldn't see the "readonly" label anywhere. Maybe that check should be done elsewhere?

PR #11187, opened 2025-02-17
Use DOMPurify 3.2.4

timar:
DOMPurify before 3.2.4 has an incorrect template literal regular expression, sometimes leading to mutation
cross-site scripting (mXSS).

Change-Id: If69bb5b06a0519c56c13019f7d19d7f58856814e


gokaysatir (approved):
Thanks for the update.

PR #11190, opened 2025-02-18
wsd/wopiAccessCheck: handle self-signed certificate

### LINKED ISSUE #11162: ssl_verification=false is not respected in wopiAccessCheck
### Describe the Bug

When making a request to the wopiAccessCheck endpoint while having ssl_verification disabled in coolwsd.xml i still get a CertificateValidation response. 

This would block setting up an instance (mostly for development purposes with using this access check in Nextcloud

### Steps to Reproduce

1. Disable ssl verification in coolwsd.xml
2. Send a request to the wopiAccessCheck that uses a callbackUrl with self-signed certificates that are not in the system trust store

### Expected Behavior
I should get an Ok response as the verification is disabled

### Actual Behavior

I see a CertificateValidation error

cc @meven

meven:
* Resolves: # 
* Target version: master 

### Summary

Add new CheckStatus SelfSignedCertificate and ExpiredCertificate along the way.

Fixes #11162

Context: https://github.com/CollaboraOnline/online/pull/9202

### Checklist

### DIFF
--- a/net/HttpRequest.hpp
+++ b/net/HttpRequest.hpp
@@ -1361,6 +1361,18 @@ public:
 #endif
     }
 
+    long getSslVerifyResult()
+    {
+#if ENABLE_SSL
+        std::shared_ptr<StreamSocket> socket = _socket.lock();
+        if (socket)
+            return socket->getSslVerifyResult();
+        return _handshakeSslVerifyFailure;
+#else
+        return 0; // X509_V_OK
+#endif
+    }
+
     std::string getSslCert(std::string& subjectHash)
     {
 #if ENABLE_SSL
--- a/wsd/ClientRequestDispatcher.cpp
+++ b/wsd/ClientRequestDispatcher.cpp
@@ -1120,6 +1120,8 @@ STATE_ENUM(CheckStatus,
     UnspecifiedError,
     ConnectionAborted,
     CertificateValidation,
+    SelfSignedCertificate,
+    ExpiredCertificate,
     SslHandshakeFail,
     MissingSsl,
     NotHttps,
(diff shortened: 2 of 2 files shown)


wsd/ClientRequestDispatcher.cpp
```diff
@@ -1284,9 +1285,12 @@ bool ClientRequestDispatcher::handleWopiAccessCheckRequest(const Poco::Net::HTTP
 
             if (!probeSession->getSslVerifyMessage().empty())
             {
-                status = CheckStatus::CertificateValidation;
-
-                LOG_DBG("Result ssl: " << probeSession->getSslVerifyMessage());
+                if (probeSession->getSslVerifyMessage() == "self-signed certificate") {
```
hfiguiere, line 1288: this problem here is that this checks the result of `X509_verify_cert_error_string()` which is "human readable".

1. are we sure the text is never localized
2. are we sure it never changes?

Otherwise we'd better use the result of `SSL_get_verify_result(_ssl)`

meven: The API would need to expose properly errors code indeed, it would be better.
This `getSslVerifyMessage` is currently used only for logging, here I have a different use-case.

We can expect it to be always english (https://linux.die.net/man/3/x509_verify_cert_error_string) and its string to be stable, as they even are documented.

meven: I switched to use the ssl result.

net/HttpRequest.hpp
```diff
@@ -1361,6 +1361,18 @@ class Session final : public ProtocolHandlerInterface
 #endif
     }
 
+    int getSslVerifyResult()
```
hfiguiere, line 1364: The signature on `Socket::` is `long getSslVerifyResult()`. So this should match.
`int` and `long` are two different types.

hfiguiere: Also `_handshakeSslVerifyFailure` is a long

hfiguiere:
```suggestion
    long getSslVerifyResult()
```

meven: Well spotted, missed it.

It seems we don't use -Werror :/ I imagine this would have been spotted then.

hfiguiere: `-Werror` wouldn't do anything if there is no warning emitted.

hfiguiere (approved):

PR #11191, opened 2025-02-18
Allow to handle custom uno commands without imgurl

GulsahKose:
Before imgurl was essential while inserting custom buttons. Now we find the button icon via uno command name even if the imgurl doesn't send by user.

Change-Id: Ide8f6df68018fcac98ce95f1a5c4a9317cff3473

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/app/LOUtil.ts
+++ b/browser/src/app/LOUtil.ts
@@ -22,6 +22,10 @@ interface RectangleLike {
 	right: number;
 }
 
+interface IconNameMap {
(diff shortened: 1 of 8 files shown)


eszkadev (approved):
looks good

GulsahKose:
CI core should be updated. That is reason of CodeQL failing. Other failing cypress tests are not related with patch. I testes locally without my change. They still failing

PR #11195, opened 2025-02-18
Preset: Add support for impress templates

Rash419:
Change-Id: I288f38c54c0b3339f2f6ffafc2408b55c5012428

* Target version: master


Rash419:
I plan to revert most of the remoteassetpoll commits that were added in https://github.com/CollaboraOnline/online/pull/10463

caolanm (approved):
code looks good, presumably presentation templates will only exist if listed in the presets config

PR #11200, opened 2025-02-19
cool#11064 any input: conditionally consider priority of core tasks

vmiklos:
Since commit 500df64292e53fe07c92d492511ed82bfcb8b546 (cool#11064 kit:
check for callbacks & wsd socket in the anyInput callback, 2025-02-17),
we interrupt core more frequently when w.e. the wsd socket of the kit
process has some incoming traffic.

This is wanted, but it may lead to the other extreme that now COOL
interrupts even high priority jobs of core, so we hit the other problem
that we execute the COOL tasks early and it would be faster to wait a
bit and batch-process multiple tasks together.

Solve this problem by considering the priority of the most urgent core
task: if it's in the "high priority" region, then don't interrupt. Do
this conditionally, as only core.git co-25.04 has commit
4f15a47c7faf55997e776af738106d3da229c7d2 (cool#11064 Set priority of
some idles to TaskPriority::DEFAULT_IDLE, 2025-02-18), and without that
a number of core-side idles would be high priority. The environment
variable can be dropped once online.git master doesn't have to work with
core co-24.04 anymore.

Note that this is primarily a theoretical concern: while it does happen
from time to time that e.g. UNO command async dispatch activates this
mechanism, there isn't a known use-case that would be quite bad without
this change. At the same time, knowing the priority will allow making
more informed decisions in COOL in the future.

Change-Id: I7d4fe9a77daa40280ee63e364cf649f93e029cf6

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3034,13 +3034,25 @@ int pollCallback(void* data, int timeoutUs)
 }
 
 // Do we have any pending input events from coolwsd ?
-bool anyInputCallback(void* data, int /*priority*/)
+bool anyInputCallback(void* data, int mostUrgentPriority)
 {
     auto kitSocketPoll = reinterpret_cast<KitSocketPoll*>(data);
     const std::shared_ptr<Document>& document = kitSocketPoll->getDocument();
 
     if (document)
     {
+        static bool considerPriority = std::getenv("COOL_ANYINPUT_CONSIDER_PRIORITY");
+        if (considerPriority && document->isLoaded())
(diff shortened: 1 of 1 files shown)


vmiklos:

The first commit is conditional for now, it can be enabled by default once online.git is OK to not work with core.git co-24.04 anymore. The second is just a CI improvement, as I had to retry an unusual amount of times, so fix one of the problems that I could reproduce locally.

caolanm (approved):
ok, and not disruptive for right now

PR #11202, opened 2025-02-19
Private/gokay/search

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev:
I also looked into regressing commit and I found which line does that: https://github.com/CollaboraOnline/online/pull/11206

cypress_test/integration_tests/desktop/writer/searchbar_spec.js
```diff
@@ -142,4 +142,20 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Searching via search bar'
 		desktopHelper.assertScrollbarPosition('vertical', 200, 305);
 		desktopHelper.assertVisiblePage(3, 4, 6);
 	});
+
+	it.only('Search input should keep the focus after a part change', function() {
```
eszkadev, line 146: forgotten `only`

gokaysatir: Ah, thank you. Do we still need this commit? I think you have another fix?

eszkadev: I added your test to my fix. I think better to not check the hardcoded "id". But would be good if you look if  my patch is ok.

PR #11205, opened 2025-02-19
wopi: Hide Rename button if WOPI doesn't support or allow

hfiguiere:
If either _supportsRename() or UserCanRename are false (default value) the rename document button is hidden. So is the menu item.
    
_supportRename() checks for both SupportsRename and not UserCanNotWriteRelative
    

Change-Id: I1543b6e02358ee6307a6c184440bd2a850a4b848

* Resolves: # 
* Target version: master 

### Summary

Hide the rename document button / menu if rename isn't supported or allowed.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/control/Control.Menubar.js
+++ b/browser/src/control/Control.Menubar.js
@@ -2306,6 +2306,9 @@ L.Control.Menubar = L.Control.extend({
 		if (menuItem.id === 'insertgraphicremote' && !this._map['wopi'].EnableInsertRemoteImage)
 			return false;
 
+		if (menuItem.id === 'renamedocument' && !(this._map['wopi']._supportsRename() && this._map['wopi'].UserCanRename))
+			return false;
+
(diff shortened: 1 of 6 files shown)


hfiguiere:
Well, there is that:

https://github.com/CollaboraOnline/online/blob/f12ee1ffc2ba65c9cadba4a829bb6a612649229c/browser/src/control/Control.DocumentNameInput.js#L32

tl;dr it wants to SaveAs if it can't rename. but that doesn't necessarily look right.

marking this as Draft for now, until I figure it out.

vmiklos:
Possibly the right action is to tell the user that rename is not supported and don't try to do something else, assuming we know better.

It would be fine for me to do that in a separate PR. Unless you have some cosmetic changes, the easiest to review PRs are the ones with 1 commit in it, gerrit-style. :-)

hfiguiere:
I have extended the check to have `_suportsRename()` to combine both `SupportsRename` and `!UserCanNotWriteRelative` that is used as a work around.

PR #11206, opened 2025-02-19
search: don't lose focus in footer

eszkadev:
Fixes regression from commit ec4858b8dfdc7b1f1d694b87ccdc5098f61356a9 Use statusJSON for Impress.

In writer document I searched for text present in the footer with hyperlink. It was loosing focus which landed inside document.


gokaysatir (approved):
Seems that part is not set for Writer. That's really weird. I checked other parts and couldn't find an issue.

We could also prevent the function call here - by checking if it's a Writer document:
`this._map.setPart(results[0].part); // go to first result.`

This "no part setting for Writer" doesn't sound good to me. But that is outside the scope of this change.

Thanks for the fix.

eszkadev:
@gokaysatir Writer is known to have all as single "page" in core

PR #11207, opened 2025-02-19
fix coolmap crash

caolanm:
stl_vector.h:1149: Assertion '__n < this->size()' failed.

Change-Id: Ie9a081fd6c797be181b0075c70b54b5bed8f4a1b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
This is to fix a crash as above that I see on using this coolmap utility to measure unshared memory of a kit process

mmeeks (approved):
Thanks =)

PR #11210, opened 2025-02-20
cool#11064 test: support creating a large Writer document with many pages

vmiklos:
- **tools: fix clang-15 build**
- **tools: prefix members of PerfMetricInfo**
- **automake silent rules: handle cmp**
- **cool#11064 test: support creating a large Writer document with many pages**


vmiklos:

The first 3 are meant to be uninteresting; the last one is something Michael wanted so we have an easy way of seeing how we work when the Writer page count is large. In a dbgutil build, you'll get the first 40 pages on load and it then slowly increases till 300, which is what we want for manual testing.

caolanm (approved):

mmeeks (commented):
Nice to see the file auto-generation stuff being useful - thanks Miklos ! =)

Issue #11149, opened 2025-02-12
ODF Spreadsheet: wrong date displayed in cell comment (language / interface bug?)


marscat5:
### Describe the Bug
There seems to be a mixup in date processing for cells comments (2025-12-02 vs. 2025-02-12) with German display language (maybe also other languages).

### Steps to Reproduce
Create a new ODF-Spreadsheet. Language of the Interface is German.
Add a comment to a cell. I did that today (Feb 12 2025) and the Comment is displayed as it was created at Dec 2 2025.

### Expected Behavior
I expect to see the correct (in my example the current) date / today in the comment of the cell displayed in the correct language. In my example there should be "Mi, 12. Feb. 2025".

### Actual Behavior
I see as date for the comment "Di., 2. Dez. 2025".

### Screenshots
see below

### Desktop
 - **Collabora version:** 24.04.11.2
 - **OS and version:** Windows 11
 - **Browser and version:** Chrome 133.0.6943.60

### Smartphone
- **Device:** Samsung Galaxy S22
 - **OS:** Android 14
 - **Browser and version:** Samsung Internet 27.0.0.79

Darshan-upadhyay1110:
I am not able to reproduce this. @Ezinnem can you please confirm this ? seems to be a important issue if it present...
CC: @annanazarayan

marscat5:
> I am not able to reproduce this. [@Ezinnem](https://github.com/Ezinnem) can you please confirm this ? seems to be a important issue if it present... CC: [@annanazarayan](https://github.com/annanazarayan)

I did a retest today and there was no error (see screenshot). I suspect it might only happen if the day can also be a month. As you can see in my initial bug report I observed it on Feb. 12 (02-12) where the numbers can be switched to get another valid date: Dec. 2 (12-02). When you tried reproducing this bug did you also set the date of your system to Feb. 12?

It would also be interesting to see, if Feb. 11 (02-11) would result in a change to Nov. 2 (11-02).

![Image](https://github.com/user-attachments/assets/2f1c961e-7971-45f3-a4e6-4e45d1a087aa)

marscat5:
> [@marscat5](https://github.com/marscat5) i did change the system date but still no issue. works fine for me.
> 
> ![Image](https://github.com/user-attachments/assets/94b405ca-19b6-46fc-a829-8bdd7f66cde4)
> 
> is it still same for you after changing the system date ?

I can see in your screenshot that you did not change the language. Can you change it to German and try again?
[...]

PR #11212, opened 2025-02-20
Fix incorrect date display in cell comments

Darshan-upadhyay1110:
- Fixed an issue where day and month were swapped in cell comments (e.g., "12.02.2025" interpreted as "02.12.2025").
- Implemented proper date parsing to correctly interpret DD.MM.YYYY HH:MM format.
- Ensured comments display the expected localized date format across different languages.( Specially in german)

Change-Id: Id6d44792bf005992ea22e62b27934e16fa6e3732

* Resolves: #11149
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/canvas/sections/CommentSection.ts
+++ b/browser/src/canvas/sections/CommentSection.ts
@@ -475,7 +475,8 @@ export class Comment extends CanvasSectionObject {
 			this.sectionProperties.authorAvatartdImg.style.borderColor = color;
 		}
 
-		var d = new Date(this.sectionProperties.data.dateTime.replace(/,.*/, 'Z'));
+		// dateTime already in UTC so we will not append Z that wil create issues while converting date
+		var d = new Date(this.sectionProperties.data.dateTime.replace(/,.*/, ''));
 		var dateOptions: any = { weekday: 'short', year: 'numeric', month: 'short', day: 'numeric' };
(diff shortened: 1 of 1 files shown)


browser/src/canvas/sections/CommentSection.ts
```diff
@@ -475,15 +475,31 @@ export class Comment extends CanvasSectionObject {
...
 	}
 
+	private getFormattedDate(dateString: any): Date {
+		const match = dateString.match(/^(\d{2})\.(\d{2})\.(\d{4}) (\d{2}):(\d{2})$/);
+		
+		if (!match) {
+			return new Date(dateString); // Return original input as a date if format doesn't match
+		}
+	
+		// Extract values properly
```
gokaysatir, line 492: Do we ever hit this line?

gokaysatir (changes requested):
I used Turkish UI on a German Calc file, i guess to make things more complicated :)

Anyway, in my case, it says "Invalid Date".

I think this issue deserves more attention. If we don't use the date anywhere else, if we only show that date to the user - not sending back to core side for some reason: We can simplify the comment's date on the core side.

I would check if it is possible to send the date from the core side always in only one specific format.

If it is possible, then we can play with it on the client side. We can show it according to the UI language.

Thanks :)

Darshan-upadhyay1110:
For this online patch we need to have this CORE patch : https://gerrit.libreoffice.org/c/core/+/182093

CC: @gokaysatir

gokaysatir (approved):
Thanks @Darshan-upadhyay1110 :) There is one more thing to improve but it is on the core side. Let me approve this.

PR #11213, opened 2025-02-20
asyncRequest doesn't close the connection

caolanm:
so the associated fds accumulate

Change-Id: Ic47a676447dc102bcae9698c0aaea68541b6963c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
--fixes my problem with cypress, thanks--
no, it only delayed the problem. still I can see the: `Failed to accept. (errno: EMFILE, Too many open files)| net/Socket.cpp:1184`

Ashod (approved):
Thanks, @caolanm. Good catch.

PR #11215, opened 2025-02-21
private/vmiklos/master

vmiklos:
- **make run: don't build all the C++ tests**
- **wsd: clean up nested redundant #if in ClientRequestDispatcher**


vmiklos:

Have c++ test code in the "check" category as autotools would suggest ideally and then add some hack (similar to core) to be able to still build everything if that's needed for some reason.

caolanm (approved):

PR #11219, opened 2025-02-21
chore(iOS): Allow debugging slideshow webview

Minion3665:
Our main webview is debuggable in debug mode but we run slideshows in a separate webview - one which wasn't marked as debuggable.

That omission prevented it from showing up in Safari's remote debugger

Change-Id: I91ae6318e0a2f735c4cd0b7faf9a9d2b1d00c68a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


timar (approved):
You have to rebase to pass CI, otherwise the patch is OK.

PR #11222, opened 2025-02-24
ui-treeview: reduce margins to fit headings in treeview sidebar

banobepascal:
Change-Id: I0ce4621c9db23ec429fcd8304f3133850f89495b

* Target version: master 

### Adjusted margins to fit the headings to fit in ui-treeview sidebar when Highlight styles is activated
Style list -> Hierarchical -> Default paragraph->  Body text -> List

### Before
image

### Now
Screenshot from 2025-02-24 18-09-54

### Added text ellipsis on longer text and tooltips for style list
[Screencast from 27-02-25 15:03:18.webm](https://github.com/user-attachments/assets/19d9d889-90c8-47c5-ada5-ce5c7c922162)

-

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -693,19 +693,19 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
 }
 
 .ui-treeview div[aria-level='2'] + .ui-treeview-expanded-content {
-	margin-left: 5ch;
+	margin-left: 3.5ch;
 }
 
 .ui-treeview div[aria-level='3'] + .ui-treeview-expanded-content {
-	margin-left: 7.5ch;
+	margin-left: 4.5ch;
 }
 
 .ui-treeview div[aria-level='4'] + .ui-treeview-expanded-content {
-	margin-left: 10ch;
+	margin-left: 5.5ch;
 }
 
 .ui-treeview div[aria-level='5'] + .ui-treeview-expanded-content {
-	margin-left: 12.5ch;
+	margin-left: 6.5ch;
 }
 
 .ui-treeview div.ui-treeview-expander {
--- a/browser/css/jssidebar.css
+++ b/browser/css/jssidebar.css
@@ -343,6 +343,11 @@ button#button2.ui-pushbutton.jsdialog.sidebar {
 	font-size: 14px;
 	white-space: break-spaces;
 }
+
+#StyleListPanelPanelExpander .ui-treeview-cell-text {
+	white-space: break-spaces;
+}
+
 #NavigatorDeck .ui-treeview-cell {
 	/* Move this whole block away and fix in the main control */
 	display: flex;


browser/src/control/jsdialog/Widget.TreeView.ts
```diff
@@ -374,6 +374,23 @@ class TreeViewControl {
...
+			2: 33,
+			3: 25,
+			4: 19,
+			5: 10,
+		};
+
+		// Get text limit for the level (or default to 20)
+		const textLimit = textLengthAndLevel[level] || 20;
+
+		if (data.id === 'treeview' && entry.text && entry.text.length > textLimit) {
```
eszkadev, line 389: we shouldn't rely on any 'id's, please remove that condition - I think this elipsis feature can be universal to all treeviews

banobepascal: @eszkadev 
How about on such information, I don't think its better to preview it in a tooltip

https://github.com/CollaboraOnline/online/pull/11222#issuecomment-2687793922

browser/css/jsdialogs.css
```diff
@@ -693,19 +693,35 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
...
 }
 
 .ui-treeview div[aria-level='3'] + .ui-treeview-expanded-content {
-	margin-left: 7.5ch;
+	margin-left: 4.5ch;
 }
 
 .ui-treeview div[aria-level='4'] + .ui-treeview-expanded-content {
-	margin-left: 10ch;
+	margin-left: 5.5ch;
```
pedropintosilva, line 704: Looks good

browser/css/jsdialogs.css
```diff
@@ -693,19 +693,35 @@ input[type='checkbox']:checked.autofilter, .jsdialog input[type='checkbox']:chec
...
+#StyleListPanelPanelExpander .ui-treeview-expanded-content .ui-treeview-entry[aria-level='3'] .ui-treeview-cell-text > .ui-treeview-cell-text {
+	max-width: 180px;
+}
+
+#StyleListPanelPanelExpander .ui-treeview-expanded-content .ui-treeview-entry[aria-level='4'] .ui-treeview-cell-text > .ui-treeview-cell-text {
+	max-width: 150px;
+}
+
+#StyleListPanelPanelExpander .ui-treeview-expanded-content .ui-treeview-entry[aria-level='5'] .ui-treeview-cell-text > .ui-treeview-cell-text {
+	max-width: 120px;
```
pedropintosilva, line 724: Looks good but I wonder if we could simplify it. But maybe that's ok for now

browser/src/control/jsdialog/Widget.TreeView.ts
pedropintosilva: Initially I thought comments were the only special case where we should allow multi line (no text-overflow: ellipsis). However I now see - and as pointed out by Banobe - that there are many other places (if not all) where they should be multi line (hyperlinks, fields, comments, headings, etc) and probably the only place we could have the treeview items to don't occupy multiple lines would be in the styles view but even in that case -> I'm not sure. 

Probably best, to keep it simply and avoid extra mouse cursor movements (user would need to move the mouse cursor to every cropped style just to read what is it) and just allow every treeview item both on Navigator and Styles view to wrap

pedropintosilva (changes requested):

pedropintosilva (approved):
Thanks, looks good

eszkadev (approved):

PR #11223, opened 2025-02-24
Use timeout instead of interval for tile prefetching.

gokaysatir:
Issue: interval calls itself multiple times.

Change-Id: I5ef46145debb02c2dd0d716b0adb0dc715449885

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -665,7 +665,7 @@ class TileManager {
 
 	private static clearTilesPreFetcher() {
 		if (this._tilesPreFetcher !== undefined) {
-			clearInterval(this._tilesPreFetcher);
+			clearTimeout(this._tilesPreFetcher);
 			this._tilesPreFetcher = undefined;
 		}
 	}
@@ -1469,7 +1469,7 @@ class TileManager {
 		this._preFetchMode = this._docLayer._selectedMode;
 		this._preFetchIdle = setTimeout(
 			L.bind(function () {
-				this._tilesPreFetcher = setInterval(
+				this._tilesPreFetcher = setTimeout(
 					L.bind(this.preFetchTiles, this),
 					interval,
 				);
--- a/cypress_test/integration_tests/common/helper.js
+++ b/cypress_test/integration_tests/common/helper.js
@@ -1074,16 +1074,17 @@ function typeIntoInputField(selector, text, clearBefore = true)
 {
 	cy.log('>> typeIntoInputField - start');
 
-	cy.cGet(selector).as('input');
-	cy.get('@input').focus();
-	cy.get('@input').should('have.focus');
 	if (clearBefore) {
-		cy.get('@input').invoke('val', '');
-		cy.get('@input').should('have.value', '');
+		cy.wait(300);
+		cy.cGet(selector).type('{selectall}{backspace}{selectall}{backspace}');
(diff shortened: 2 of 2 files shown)


caolanm:
I believe this started to be a problem since commit 66f2449f3e4f00031e4bea86e1a28921a65e9de5 so we probably should revisit what changed there as there wasn't intended to be any change in behavior IIUC

gokaysatir:
> I believe this started to be a problem since commit [66f2449](https://github.com/CollaboraOnline/online/commit/66f2449f3e4f00031e4bea86e1a28921a65e9de5) so we probably should revisit what changed there as there wasn't intended to be any change in behavior IIUC

I only removed:
`this._previewInvalidations.push(invalidBounds);`

from Writer and Calc tile layers. Because they don't have previews. That code was not needed in those places. That may have exposed the issue.

mmeeks:
Oh! I shortened the timeouts for pre-fetching left and right. Hmm - is it the case that we can multiple pre-fetch the same tiles if they have not been received yet ? if so - "that would be bad"(TM). Gokay- did you test this with the preload map enabled - to ensure that we continue to pre-load properly ? =) - particularly in calc.

gokaysatir:
> Oh! I shortened the timeouts for pre-fetching left and right. Hmm - is it the case that we can multiple pre-fetch the same tiles if they have not been received yet ? if so - "that would be bad"(TM). Gokay- did you test this with the preload map enabled - to ensure that we continue to pre-load properly ? =) - particularly in calc.

Yes i tested it now. I think this is a bug that we didn't notice before.

caolanm:
I added some comments in https://github.com/CollaboraOnline/online/pull/11211 I think I see some discrepancies in the conversion there that look worth clarifying.

gokaysatir:
> I added some comments in #11211 I think I see some discrepancies in the conversion there that look worth clarifying.

Thanks again: You are right, there were 2 issues, fixed now:
* https://github.com/CollaboraOnline/online/pull/11231

I think that we can replace interval with timer, since we already reset prefetcing via various events like scroll and move. I think a position change (move) can be enough for being sure that we prefetch at the right time.

I think this is correct that we re-send the tile request until they arrive:

> Hmm - is it the case that we can multiple pre-fetch the same tiles if they have not been received yet ? 

And it can slow down a bit more when a large document is viewed.

mmeeks:
> I think this is correct that we re-send the tile requests until they arrive:

I suspect that is a bug - particularly if we don't have Szymon's patch that stops us re-rendering the tiles as they are requested each time from a tilecombine =) ie. we badly need:

commit 4c85039d7d85424c98c9aeb8e6c3809beed17d4d
Author: Szymon Kłos <szymon.klos@collabora.com>
Date:   Mon Feb 10 16:08:49 2025 +0100

    tiles: do not duplicate rendering if tilecombine requested

in addition there I think.

gokaysatir:
I think we will get this in, can we approve the request?

gokaysatir:
> Please be careful. This breaks pre-loading.
> 
> image
> 
> Please turn on the feature I asked for - and see what happens to incremental tile pre-loading around the display area - which is an iterative process.

I see, i think i mis-read the expected behaviour and couldn't see the regression. Let me abandon this PR. I will cherry-pick the cypress fix.

PR #11230, opened 2025-02-25
browser: remove unused nodejs canvas pkg

dennisfrancis:
Change-Id: I548db86fd4e426312f3d0404f8557c9275633895

* Resolves: # 
* Target version: master 

### Summary
Remove unused nodejs canvas pkg

### Checklist


mmeeks (approved):
Thanks Dennis ! =)
We do require the canvas in 'make perf-test' I think - but no doubt that test needs some maintenance and fails in interesting ways ;-)

PR #11232, opened 2025-02-25
use higher uid/gid for cool in docker images

caolanm:
Change-Id: If690af3ad746f6ecc6f4afcfed4cd6d627d589a7

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/docker/from-packages/Dockerfile
+++ b/docker/from-packages/Dockerfile
@@ -118,6 +118,10 @@ RUN --mount=type=secret,id=secret_key \
             apt-get -y install code-brand; \
         fi; \
     fi && \
+# Change cool uid/gid to 1001/1001
+    find /opt/cool -user cool -exec chown -h 1001:1001 {} \; && \
+    usermod -u 1001 cool && groupmod -g 1001 cool && \
+    chown cool:cool /etc/coolwsd/coolwsd.xml && \
 # Fix ownership of config directory that will be modified on start of the container by cool user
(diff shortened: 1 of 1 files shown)


caolanm:
Some policy scanners apparently report warnings like:

> runAsUser > 1000 & runAsGroup > 1000:
> Containers must run with a user who has a UserID > 1000 and a GroupID > 1000.

for the docker image. Which just uses whatever debian gives from:
adduser --quiet --system --group --home /opt/cool cool
in debian/coolwsd.postinst.in, which is 100/101 for the current docker image.

@Rash419 @meven: FYI, maybe this is useful

docker/from-packages/Dockerfile
```diff
@@ -118,6 +118,10 @@ RUN --mount=type=secret,id=secret_key \
             apt-get -y install code-brand; \
         fi; \
     fi && \
+# Change cool uid/gid to 10000/10001
+    find /opt/cool -user cool -exec chown -h 10000:10001 {} \; && \
```
meven, line 122: 1001:1001 should be sufficient.

```suggestion
     # Change cool uid/gid to 1001/1001
    find /opt/cool -user cool -exec chown -h 1000:1001 {} \; && \
```

https://stackoverflow.com/a/73359838 seems relevant here. At least I learned a thing or two.

timar (approved):

meven (approved):

PR #11233, opened 2025-02-25
Mocha: Mock necessary canvas context API functions

dennisfrancis:
* Target version: master 

### Summary
Mocha: Mock necessary canvas context API functions and remove canvas dependency from npm-shrinkwrap.json.in too.

### Checklist


timar (approved):

mmeeks (approved):
what a beautiful canvas implementation =) we should re-use it in the perf-test code :-) I learned only with this patch that these no parameter functions were valid to have more stuff passed to them - fun.

PR #11238, opened 2025-02-26
Private/caolan/pahole

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (commented):
Ooh - it's a nice change. It will break quite a lot of back-porting though - We should think about when best to merge this. To what degree is this auto-generated ? =)

caolanm:
What can be packed was autogenerated with included incantation, but the ordering was manual, albeit mechanical by largest first and e.g. fancy std::atomic<bool>s grouped before bools

mmeeks:
Heh - conflicts already sadly; we should get this into red for 25.04 and prolly back-port it soon afterwards to the stable branch I think.

mmeeks (approved):
Lets get it in =) @timar would be nice to have this in 24.04 too soonish to avoid cross-cherry-picking problems.

PR #11239, opened 2025-02-27
Misc Improvements

Ashod:
- **wsd: better logging**
- **wsd: comments**
- **wsd: remove const to allow moving**
- **wsd: catch exceptions from uploadPresetsToWopiHost()**


Rash419 (approved):
Thanks

PR #11240, opened 2025-02-27
Misc Improvement

Rash419:
Change-Id: I1297d71308bcd3650e3bc6c0faf770bd5ed0bb1e

* Target version: master

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -4030,32 +4030,33 @@ void DocumentBroker::syncBrowserSettings(const std::string& userId, const std::s
                                                  << "] for all sessions with userId [" << userId
                                                  << ']');
 
-    bool upload = false;
+    // update browsersetting for all the matching session
+    std::shared_ptr<ClientSession> sessionForUpload;
     for (auto& it : _sessions)
     {
         if (it.second->getUserId() != userId)
             continue;
 
         it.second->updateBrowserSettingsJSON(key, value);
-        if (!upload)
-        {
-            try
-            {
-                uploadBrowserSettingsToWopiHost(it.second);
-                upload = true;
-            }
-            catch (const std::exception& exc)
-            {
-                LOG_WRN("Failed to upload browsersetting json for session ["
-                        << it.second->getId() << "]: " << exc.what());
-            }
-        }
+        sessionForUpload = it.second;
+    }
+
+    // upload browsersetting only once if we found matching session
+    if (!sessionForUpload)
+        return;
+
+    try
+    {
+        uploadBrowserSettingsToWopiHost(sessionForUpload);
+    }
+    catch (const std::exception& exc)
+    {
+        LOG_WRN("Failed to upload browsersetting json for session [" << sessionForUpload->getId());
     }
 }
 
-Poco::URI DocumentBroker::getPresetUploadBaseUrl(const Poco::URI& uri)
+Poco::URI DocumentBroker::getPresetUploadBaseUrl(Poco::URI uriObject)
 {
-    Poco::URI uriObject(uri);
     std::string path = uriObject.getPath();
     size_t pos = path.find("/files/");
     if (pos != std::string::npos)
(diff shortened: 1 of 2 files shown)


wsd/DocumentBroker.cpp
```diff
@@ -4039,17 +4039,23 @@ void DocumentBroker::syncBrowserSettings(const std::string& userId, const std::s
         it.second->updateBrowserSettingsJSON(key, value);
         if (!upload)
```
Ashod, line 4040: Is the assumption that a user might have multiple sessions (which is true), and we want to `updateBrowserSettingsJSON()` for all sessions, but `uploadBrowserSettingsToWopiHost()` only once (for the current user) ?

If yes, can we please add a comment? If no, can we please add a comment with the correct logic ?

Ashod, line 4040: This is reverse logic: if upload is true, we _don't_ upload, but if it's false we _do_ upload.

Better to either rename this to 'uploaded' (past tense) or 'needUpload' (and invert the logic everywhere).

wsd/DocumentBroker.cpp
```diff
@@ -4064,8 +4070,7 @@ void DocumentBroker::uploadPresetsToWopiHost(const Authorization& auth)
     const std::string& jailPresetsPath = FileUtil::buildLocalPathToJail(
         COOLWSD::EnableMountNamespaces, getJailRoot(), JAILED_CONFIG_ROOT);
 
-    Poco::URI uriObject = DocumentBroker::getPresetUploadBaseUrl(_docKey);
-
+    Poco::URI uriObject = DocumentBroker::getPresetUploadBaseUrl(_uriOrig);
```
Ashod, line 4073: Did we look at the differences between `_uriOrig` and `_uriPublic` to decide which one to use?

I would expect we want `_uriPublic`. Please read the code that sets these two values (hint: it's in `RequestVettingStation::handleRequest()`). Looking at the difference between them should make it clear why we want `_uriPublic`.

In fact, a quick grep shows that `_uriOrig` is only used in one case (argument to `COOLWSD::dumpNewSessionTrace()`), while `_uriPublic` is used _everywhere_ we need the URI of the document (grep for 'getPublicUri'). So, again, I'm not sure why we're not using the more sensible value. (We should check what we use and look at existing usage, otherwise ask when it's still not obvious.)

Ashod (commented):
Thanks, @Rash419. Please see comments.

wsd/DocumentBroker.cpp
Rash419: I ended taking a bit different route. Is it okay now ?

wsd/DocumentBroker.cpp
Rash419: Thanks

wsd/DocumentBroker.cpp
```diff
@@ -4030,26 +4030,34 @@ void DocumentBroker::syncBrowserSettings(const std::string& userId, const std::s
                                                  << "] for all sessions with userId [" << userId
                                                  << ']');
 
-    bool upload = false;
+    // update browsersetting for all the session
+    std::shared_ptr<ClientSession> sessionForUpload = nullptr;
```
Ashod, line 4034: This initialization is not necessary, and it's actually better not to do what by default is already done. The following is sufficient, efficient, and clear.
```
std::shared_ptr<ClientSession> sessionForUpload;
```

wsd/DocumentBroker.cpp
```diff
@@ -4030,26 +4030,34 @@ void DocumentBroker::syncBrowserSettings(const std::string& userId, const std::s
...
     }
 }
 
-Poco::URI DocumentBroker::getPresetUploadBaseUrl(const std::string& docKey)
+Poco::URI DocumentBroker::getPresetUploadBaseUrl(const Poco::URI& uri)
 {
-    Poco::URI uriObject(Uri::decode(docKey));
-
-    // extract base wopiUri from docKey
+    Poco::URI uriObject(uri);
```
Ashod, line 4060: We can delete this line and name the parameter from `uri` to `uriObject`.

Ashod (approved):
Thanks, @Rash419. It's much better now.

caolanm (approved):
so lets merge then

PR #11246, opened 2025-02-27
lokit: fix fail to mount shared preset template directory

Rash419:
- happens when templates are not passed as a part of sharedconfig.json

Change-Id: I5ca53e35a6db3cef84d18ad8004ceeed4cc839ac

* Resolves: # 
* Target version: master


caolanm (approved):

caolanm:
top_toolbar is an existing common problem in ci at the moment

PR #11253, opened 2025-02-28
Compact mode: Fix toolbar-row height

pedropintosilva:
In compact mode, the toolbar-row (the second bar up top) was being set
with a hard coded pixel (42px) which was resulting in extra space and
the misaligned icon row.

The initial diff comes from:
  e88d9340a0067c9fb267fe3a3653e42c0298ee71
  Fix toolbar up section after w2ui rework

Maybe there is something I missed, but I think we can now safely
remove that compact mode related CSS declaration while making sure
the one that is intended to affect notebookbar only does that.

Change-Id: Ic50c8a89578bb1dc7c2acf5f631fdea81a1d235d

### DIFF
--- a/browser/css/notebookbar.css
+++ b/browser/css/notebookbar.css
@@ -176,7 +176,7 @@ html[data-theme='dark'] .savemodified.unotoolbutton .unobutton img {
 	z-index: 11;
 }
 
-#toolbar-row {
+.hasnotebookbar > #toolbar-row {
 	height: 74px;
 }
 
--- a/browser/css/toolbar.css
+++ b/browser/css/toolbar.css
@@ -245,10 +245,6 @@
 	top: 3px;
 }
 
-.main-nav:not(.hasnotebookbar) ~ #toolbar-wrapper .toolbar-row {
-	height: 42px;
-}
-
 /* center the toolbar */
 #tb_presentation-toolbar_item_left {
 	width: 50%;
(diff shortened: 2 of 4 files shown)


Darshan-upadhyay1110 (approved):
Tested and seems good to me, Thanks :)

Darshan-upadhyay1110:
I do not recall but the fix height change was something that was need after w2ui removal. But anyhow this patch works fine and doesn't break anything on toolbar....

Thanks Pedro

Darshan-upadhyay1110:
I did re-trigger CI 2 times it's failing. I will check the cypress test....

Maybe because of change in height some cypress test is not passing.

Darshan-upadhyay1110:
We have total 3 cypress test cases are failing 

Desktop

1. writer/scrolling_spec.js 
2. calc/autofilter_spec.js 

=> I have some fix for this test cases which i will push 

Mobile

3. calc/formulabar_spec.js : I tested locally and it **passed ** failing case which is integration_tests/mobile/calc/formulabar_spec.js

image

I do not know why after 4 retry this case is still failed in CI but work fine in locally for mobile case :(

pedropintosilva:
Looks good, thanks for the cypress update. I unable to approve because I'm the PR author, adding @gokaysatir

pedropintosilva:
> Mobile
> 
>     3. calc/formulabar_spec.js : I tested locally and it **passed ** failing case which is integration_tests/mobile/calc/formulabar_spec.js

Same here, it passed multiple times but I did manage to get 1 in ~12 failing but it is unrelated to this PR. I have checked the dimension in #toolbar-row, #formulabar-row and their parent #toolbar-wrapper on this particular spec (calc/formulabar_spec.js) and they are the same with or without this PR.

gokaysatir (approved):
Looks good, thank you.

eszkadev:
> I did re-trigger CI 2 times it's failing. I will check the cypress test....
> 
> Maybe because of change in height some cypress test is not passing.

It is known that changing the canvas size (eg. by toolbar size) is causing tests to fail.
We check scrollbar positions, visible page number etc.

PR #11256, opened 2025-02-28
tiles: avoid sending tilecombines too fast.

mmeeks:
We should give coolwsd plenty of time to return a complete keyframe if we are pre-loading it, so annotate a tilecombine with when the last tilecombined was sent on it.

For tiles in the current view, we should get pushed tile content anyway almost always; special-case keyframe fetching when we have cleared our cache early.

Change-Id: Id6bfe630ac276a7cc3d7eb7a748c769bddd95a80

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


gokaysatir (approved):
Checked & tested, thanks for the improvement :)

pedropintosilva:
Also pressed the re-try button for the cypress desktop tests at it was failing to assert the value of the #addressInput input (which I have seen multiple times failing out side of this PR)

PR #11259, opened 2025-03-01
WIP - cool#11255 - don't free tile canvases, but queue them for re-use.

mmeeks:
A rather unfinished first-cut at this, we need to free up our tiles as/when we go idle to save GPU/CPU memory, and altogether optimise this further.

Change-Id: I12f55207feed684b7fee0188fb7ce914a9f5929f

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks:
Abandoning this - Chris has a better version in https://github.com/CollaboraOnline/online/pull/11304

PR #11264, opened 2025-03-03
Use correct command for style sidebar

eszkadev:
Remove old CSS rule to hide style sidebar shortcut. We have it now in calc. Unify menubar command with other places. EditStyle doesn't work second time when used.

Menubar changes:
    - put "Open the Styles Deck" in Calc -> Format (like in Writer)
    - add "Styles Deck" in Calc -> View
    - moved Edit -> View in Writer for "Styles Deck" (similar to Navigator and general properties Sidebar decks)

### DIFF
--- a/browser/css/spreadsheet.css
+++ b/browser/css/spreadsheet.css
@@ -208,14 +208,6 @@
 	z-index: -1;
 }
 
-/* Sidebar: Style: Hide dialog button
-   - There is no paragraph style dialog in calc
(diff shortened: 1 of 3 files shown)


eszkadev:
Added entry in calc which was missing.

styles

Possibly we would prefer better woding. cc: @pedropintosilva @banobepascal 
`Style sidebar` or what should it be?

pedropintosilva:
The same as we use in Writer's Format Tab , I would say

PR #11269, opened 2025-03-04
avoid excessively wide tile combines

caolanm:
* Resolves: # 
* Target version: master 

### Summary

Try and create more sensible tile combines. Combine as many horizontal tiles as sensible with priority tile, working from both sides of it, favoring nearer tiles until reasonable span width is exhausted. Combine adjacent rows with this span. Distribute any "unused" available span width on comparing adjacent rows to pick up relatively cheap additions to the combine.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/TileDesc.hpp
+++ b/wsd/TileDesc.hpp
@@ -181,6 +181,16 @@ public:
         return a ^ b;
     }
 
+    TileDesc makeTileForGridPos(int gridX, int gridY) const
+    {
+        TileDesc tile(*this);
+
+        tile._tilePosX = gridX * tile._tileWidth;
+        tile._tilePosY = gridY * tile._tileHeight;
+
+        return tile;
+    }
+
     /// Returns the tile's AABBox, i.e. tile-position + tile-extend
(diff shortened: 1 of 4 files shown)


wsd/TileDesc.hpp
```diff
@@ -181,6 +181,16 @@ class TileDesc final
         return a ^ b;
     }
 
+    TileDesc makeTileForGridPos(int gridX, int gridY) const
```
mmeeks, line 184: Algorithm reads nicely - thanks =)
This gives me the heebie jeebies though ;-) the TileDesc brings so much important meta-data around, that having new / naked ones being created - and that they might slip into some list/tile-combine etc. worries me somewhat. Could we instead have a throwaway X,Y struct or even parameters we pass down instead of the anchorTile using this ?

mmeeks (approved):
Looks great - thanks, would be nice to get a follow-on re: the naming / re-use of TileDesc as an X,Y position =) and ideally get more unit tests alive but ... nice work.

mmeeks (approved):
thanks ! =)

PR #11270, opened 2025-03-05
Unload While Loading

Ashod:
**wsd: handle disconnection while loading**
When the client disconnects while we are
loading the document, we risk getting
in an invalid state.

When we get disconnected while loading,
we have logic to terminate the kit.
Unfortunately, we don't account for
the fact that the session isn't yet
removed from the container, and fail
to detect this case. Furtheremore,
we should stop the DocBroker poll,
which we didn't previously.

Also, we fix an invalid state transition:

**wsd: avoid transitioning to LIVE from WAIT_DISCONNECT**
wsd: avoid transitioning to LIVE from WAIT_DISCONNECT
In the event that we gave up on loading,
we should never change the state back to LIVE
once we are in WAIT_DISCONNECT.

This was caught with: WRN  ToClient-002: Unusual race - attempts to
transition from SessionState::WAIT_DISCONNECT to SessionState::LIVE|
wsd/ClientSession.cpp:151

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -2472,24 +2472,28 @@ bool ClientSession::handleKitToClientMessage(const std::shared_ptr<Message>& pay
         }
         else if (tokens.equals(0, "loaded:"))
         {
-            setState(ClientSession::SessionState::LIVE);
-
-            if (firstLine.find("isfirst=true") != std::string::npos)
+            // We expect to be in the Loading state, as set in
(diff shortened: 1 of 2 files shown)


Ashod:
> Looks very sensible - but we badly need to catch all of these corner cases which are easy to forget in future. Hopefully now we understand it it shouldn't be that hard to do. I would like a C++ unit test for this - perhaps we need to add new hooks for that - can you prioritize that ? =) Good work.

Will add a test. Thanks.

PR #11272, opened 2025-03-05
unit-tiletest, testUnresponsiveClient: don't assume invalidate is sent without having tiles first

vmiklos:
- **run-wsdunit: build the test if necessary**
- **unit-tiletest, testUnresponsiveClient: don't assume invalidate is sent without having tiles first**
- **unit-tiletest: disable remaining broken tests in this suite**

### DIFF
--- a/scripts/run-wsdunit
+++ b/scripts/run-wsdunit
@@ -15,6 +15,11 @@ if [ -z "$name" ]; then
 fi
 shift # Eat the first argument.
 
+# Make sure to not run an outdated binary.
+if ! make -C test $name.la; then
+    exit 1
+fi
+
 # Maximum number of runs.
 limit=1
 if [ $# -ne 0 ];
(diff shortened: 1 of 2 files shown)


vmiklos:

Unfortunately https://gerrit.libreoffice.org/c/core/+/182524 went in sooner than I expected, so the last commit in the PR is just a minimal fix to get C++ tests pass again without investigation. (I.e. everything just fails till this in in, annoyingly.)

Cypress desktop: there is a failure in writer/image_operation_spec.js, it passes for me locally, so if we try once more, it would pass, I expect.

Cypress mobile at https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_mobile/4189/console fails in calc/delete_objects_spec.js & calc/formulabar_spec.js, these 2 pass for me locally, this PR improves C++ tests, mobile tests are not required: I suggest to ignore the failure here and get this in.

caolanm (approved):
Got it, adds something new that should work, temp disable the others that stopped working :-)

Issue #11124, opened 2025-02-10
perf: touch styles/DOM less ...


mmeeks:
It seems we have a fair bit of work done in touching the DOM as people type:

![Image](https://github.com/user-attachments/assets/1e526b18-fdff-4154-a1d5-abb259472ab8)

updateCursorAndOverlay seems to do lots of blue/purple stuff that seems un-necessary for moving a cursor - the isOpen and isAnyInputFocused look like things we could do without jQuery =)

showCursor (I guess) constantly touches a CSS opacity setting.

Canvas-rendering looks like it would benefit from Caolan's patch to reduce the number of tiles rendered too - to only the ones that changed (I guess we had a patch for that somewhere).

Of course - whether these problems are 'real' is unclear to me, but would be worth nailing if they are easy to fix I think =)

eszkadev:
Mostly CSS style computation which is expensive -> offsetWidth on cursor update.
Probably it's about checking if we should move the view (jump to cursor, scroll the page as we went outside).

I bet we can do some clever optimization if we know position of view on the document.
Probably we have enough information inside canvas sections? cc: @gokaysatir

gokaysatir:
Maybe we can exit early from "isAnyInputFocused" function if map is focued. This can work if map loses focus when an input dialog has it.

gokaysatir:
Yes there may be unnecessary calls to "setOpacity" function. I don't know if it causes performance drop.  I would move cursor to a section but a comment there says: "This is used as the text-cursor(in and out of document)."

So i didn't touch it. I think there may be some room there for improvement.

mmeeks:
As a general rule of thumb changing any DOM attribute - even to the same value causes a performance drop; we should just avoid any writes we possibly can there.

mmeeks:
I guess we should defer that DOM touching until an animationframe - which might help us get an associated tile to be synchronized with the cursor movement too which would be good. It is hard to avoid the DOM touch caused by the keyboard input into this invisible text entry which (IIRC) needs to be thought to be visible to function properly (but perhaps that is worth re-visiting).

eszkadev:
Comments also need some squashing :

![Image](https://github.com/user-attachments/assets/1daf66b3-91b5-4879-b482-817305ae679c)

mmeeks:
@eszkadev nice are you working to fix that ? or can we give that to say @codewithvk ? =)

eszkadev:
[...]

PR #11275, opened 2025-03-05
jsdialogs: slurp also widget updates

eszkadev:
When we open big dialogs like eg. Format -> Character we first get the full dialog and then series of updates. We touch DOM a lot by replacing some nodes and layouting the content.

Let's create a queue which will collect them and flush at once.

Use DocumentFragment when we build widget to not insert directly into real Document.

Setup number formatting only once, not on every builder creation.

BEFORE (open Character dialog):
beforequeue

Related to issue  #11124

All the layouting happens now in the seaparate task. JS for building dialog is shorter.

### DIFF
--- a/browser/src/app/LayoutingService.ts
+++ b/browser/src/app/LayoutingService.ts
@@ -1,0 +1,58 @@
+/* -*- js-indent-level: 8; fill-column: 100 -*- */
+
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * This file contains service which will coordinate intensive DOM modifications
+ * which may impact browser rendering performance.
+ */
+
+type LayoutingTask = () => void;
+
+class LayoutingService {
+	private _layoutTasks: Array<LayoutingTask> = [];
+	private _layoutTaskFlush: ReturnType<typeof setTimeout> | null = null;
+
+	private _flushLayoutingQueue() {
+		if (this._layoutTasks.length) {
+			window.requestAnimationFrame(() => {
+				const start = performance.now();
+				let task = this._layoutTasks.shift();
+				let duration = 0;
+				while (task) {
+					task.call(this);
+					duration = performance.now() - start;
+					if (duration > 10) {
+						this.scheduleLayouting();
+						break;
+					}
+					task = this._layoutTasks.shift();
+				}
+			});
+		}
+
+		this._layoutTaskFlush = null;
+	}
+
+	public appendLayoutingTask(task: LayoutingTask) {
+		this._layoutTasks.push(task);
+	}
+
+	public scheduleLayouting() {
+		if (this._layoutTaskFlush) return;
+
+		// get something around 60 fps
+		this._layoutTaskFlush = setTimeout(
+			this._flushLayoutingQueue.bind(this),
+			10,
+		);
+	}
+}
--- a/browser/src/app/LocaleService.ts
+++ b/browser/src/app/LocaleService.ts
@@ -24,8 +24,8 @@ class LocaleService {
 		this._initialized = true;
 
 		if (typeof Intl !== 'undefined') {
-			let formatter, lang;
 			try {
+				let formatter;
 				if (app.UI.language.fromURL && app.UI.language.fromURL !== '')
 					formatter = new Intl.NumberFormat(app.UI.language.fromURL);
 				else formatter = new Intl.NumberFormat(L.Browser.lang);
@@ -41,7 +41,7 @@ class LocaleService {
 					}
 				});
 			} catch (e) {
-				window.app.console.log('Exception parsing lang ' + lang + ' ' + e);
+				window.app.console.log('Exception parsing lang ' + e);
 			}
 		}
 	}
--- a/browser/src/app/LocaleService.ts
+++ b/browser/src/app/LocaleService.ts
@@ -1,0 +1,56 @@
+/* -*- js-indent-level: 8; fill-column: 100 -*- */
+
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * This file holds locale dependent options.
+ */
+
+class LocaleService {
+	private _globalDecimal: string = '.';
+	private _globalMinusSign: string = '-';
+	private _initialized: boolean = false;
+
+	public initializeNumberFormatting() {
+		if (this._initialized) return;
+		this._initialized = true;
+
+		if (typeof Intl !== 'undefined') {
+			let formatter, lang;
+			try {
+				if (app.UI.language.fromURL && app.UI.language.fromURL !== '')
+					formatter = new Intl.NumberFormat(app.UI.language.fromURL);
+				else formatter = new Intl.NumberFormat(L.Browser.lang);
+
+				formatter.formatToParts(-11.1).map((item) => {
+					switch (item.type) {
+						case 'decimal':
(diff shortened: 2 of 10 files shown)


eszkadev:
@Cwiiis would be nice if you can check this at some point

browser/src/control/Control.JSDialogBuilder.js
```diff
@@ -16,6 +16,32 @@
...
+				break;
+				case 'minusSign':
+					_globalMinusSign = item.value;
+				break;
+			}
+		});
+	} catch(e) {
+		window.app.console.log('Exception parsing lang ' + lang + ' ' + e);
+	}
+}
```
Cwiiis, line 43: This is a good optimisation, but I wonder if there's a suitable object in the hierarchy that we can hang this off of? It feels a bit messy to just stick this in the global scope like this.

browser/src/control/Parts.js
```diff
@@ -491,24 +471,54 @@ L.Map.include({
...
+		}
+	},
+
+	showPageModalImpl: function(dialogId) {
+		const modal = document.getElementById(dialogId);
+
+		if (!modal) {
+			setTimeout(() => { this.showPageModalImpl(dialogId) }, 10);
+			return;
+		}
```
Cwiiis, line 484: Is there a possibility that this gets called multiple times before this timeout can be serviced? Is there also a chance that the `dialogId` element could never be created and this would just create an infinite (non-blocking) loop?

Assuming this is now asynchronous because of the newly introduced layouting task, I would suggest that this also uses a layouting task (assuming they're serviced in order) and fails if `dialogId` doesn't exist. It feels risky to add something like this and also makes the code harder to understand and debug if things fail.

browser/src/docstatefunctions.js
```diff
@@ -335,3 +335,23 @@ app.impress.isSlideSelected = function (index) {
...
+	if (app.layoutTasks.length) {
+		window.requestAnimationFrame(() => {
+			const task = app.layoutTasks.shift();
+			if (task) {
+				task.call(this);
+				app.scheduleLayouting();
+			}
+		});
+	}
+};
```
Cwiiis, line 349: This does flush the queue, but only at a rate of 1 task/10ms... Is this intentional? It feels like either we'd want to service the entire queue here, or this should be called something else. Maybe `flushLayoutingQueueAsync` or `drainLayoutingQueue`.

browser/src/docstatefunctions.js
```diff
@@ -335,3 +335,23 @@ app.impress.isSlideSelected = function (index) {
...
+
+app.appendLayoutingTask = function (task) {
+	app.layoutTasks.push(task);
+};
+
+app.scheduleLayouting = function () {
+	setTimeout(app._flushLayoutingQueue, 10);
```
Cwiiis, line 356: See above - it seems like the benefit of grouping DOM changes like this is partially lost if we don't actually flush the queue here.

browser/src/map/Clipboard.js
```diff
@@ -1264,7 +1264,18 @@ L.Clipboard = L.Class.extend({
...
+
+	_warnCopyPasteImpl: function (id) {
 		var box = document.getElementById(id + '-box');
+
+		// TODO: do it JSDialog native...
+		if (!box) {
+			setTimeout(() => { this._warnCopyPasteImpl(id) }, 10);
+			return;
+		}
+
```
Cwiiis, line 1278: Same comment as before, this should probably be a layouting task if it relies on elements that get added via layouting tasks. The pattern of "this thing isn't here now, just try again infinitely every 10ms" feels like a problem waiting to happen.

browser/src/map/Clipboard.js
```diff
@@ -1351,7 +1362,19 @@ L.Clipboard = L.Class.extend({
...
+	_openPasteSpecialPopupImpl: function (id) {
 		var box = document.getElementById(id + '-box');
+
+		// TODO: do it JSDialog native...
+		if (!box) {
+			setTimeout(() => { this._openPasteSpecialPopupImpl(id) }, 10);
+			return;
+		}
+
```
Cwiiis, line 1377: Same here. Note that also using a layouting task would also mean we don't need to add an extra class method, it could just be a lambda.

Cwiiis (changes requested):
This looks really good, but I think we need to remove the "spin infinitely at 10Hz until things exist" pattern contained within.

browser/src/map/Clipboard.js
eszkadev: That thing was converted from old "build from scratch with some library" to at least share our JSDialog window.
So now it creates window and then hacks it's inner content. It should be later moved to "native" JSDialog think which will not have that problem to solve (waiting for window).

browser/src/docstatefunctions.js
eszkadev: omg, I forgot to remember the timer and cancel old one :)

eszkadev:
new profile after I fixed the "flush" timeout which now squashes the execution:
unifiedqueue5

browser/src/control/Control.JSDialog.js
```diff
@@ -748,32 +748,48 @@ L.Control.JSDialog = L.Control.extend({
...
 
-			// Special case for nonModal dialogues. Core side doesn't send their initial coordinates. We need to center them.
-			if (instance.nonModal && !(instance.startX && instance.startY)) {
-				this.centerDialogPosition(instance);
-			} else {
-				instance.updatePos();
-			}
+			app.appendLayoutingTask(() => {
+				// dialog built - add to DOM now
+				const existingNode = dialogDomParent.querySelector('[id="' + instance.container.id + '"]');
```
Cwiiis, line 762: Any reason not to use `getElementById` here?

eszkadev: I'm not sure if will be unique... (it comes form the core) so wants to search inside container we worked on

eszkadev:
Fresh profiles in Writer -> format -> character dialog.

Chrome shows rendering reduction 354 ms -> 99ms in total.

BEFORE:
beforech
beforeff


AFTER:
afterch
afterff

browser/src/app/LocaleService.ts
```diff
@@ -0,0 +1,56 @@
...
+						case 'decimal':
+							this._globalDecimal = item.value;
+							break;
+						case 'minusSign':
+							this._globalMinusSign = item.value;
+							break;
+					}
+				});
+			} catch (e) {
+				window.app.console.log('Exception parsing lang ' + lang + ' ' + e);
```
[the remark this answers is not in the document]
eszkadev: done

eszkadev:
[COOL 25.04 unit tests w/o cypress](https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-25.04/807/) - PASSED but didn't report (false fail)
[COOL 25.04 Cypress (mobile)](https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-25.04_cypress_mobile/747/) - PASSED (only known annotation test failed)

browser/src/app/LayoutingService.ts
```diff
@@ -0,0 +1,58 @@
...
+					task = this._layoutTasks.shift();
+				}
+			});
+		}
+
+		this._layoutTaskFlush = null;
+	}
+
+	public appendLayoutingTask(task: LayoutingTask) {
+		this._layoutTasks.push(task);
```
mmeeks, line 46: Is there a reason we separate scheduleLayouting from this ? seems a pain to have to remember to schedule that ;-) perhaps there is a rare case here but then scope for leakage.

Could we merge that in a follow-up ? =)

browser/src/control/Control.JSDialog.js
```diff
@@ -748,32 +748,42 @@ L.Control.JSDialog = L.Control.extend({
 			if (this.map)
 				this.map._progressBar.end();
 
-			this.createContainer(instance, instance.overlay ? instance.overlay: instance.containerParent);
+			const dialogDomParent = instance.overlay ? instance.overlay: instance.containerParent;
+			const documentFragment = new DocumentFragment(); // do not modify dom yet
```
mmeeks, line 752: Like constructing this separately =)

mmeeks (approved):
Looks great to me - lets get it in =) Some final comments there though.

I expect we need one of these queues coupled to the existing canvas rendering pipeline next.

browser/src/app/LayoutingService.ts
eszkadev: done in https://github.com/CollaboraOnline/online/pull/11644/files

PR #11286, opened 2025-03-06
browser/socket: print warning in case of error

meven:
Change-Id: I62b0dbaf046a8640fab14cc531a42266848f806c

* Resolves: # 
* Target version: master 

### Summary

Use the event to log the error information when a socket error happens:
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event

### Checklist

### DIFF
--- a/browser/src/core/Socket.js
+++ b/browser/src/core/Socket.js
@@ -1645,8 +1645,8 @@ app.definitions.Socket = L.Class.extend({
 		this._map.fire('hyperlinkclicked', {url: link, coordinates: coords});
 	},
 
-	_onSocketError: function () {
-		window.app.console.debug('_onSocketError:');
+	_onSocketError: function (event) {
+		window.app.console.warn('_onSocketError:', event);
 		this._map.hideBusy();
 		// Let onclose (_onSocketClose) report errors.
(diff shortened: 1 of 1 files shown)


hfiguiere:
please rebase

browser/src/core/Socket.js
```diff
@@ -1649,8 +1649,8 @@ app.definitions.Socket = L.Class.extend({
 		this._map.fire('hyperlinkclicked', {url: link, coordinates: coords});
 	},
 
-	_onSocketError: function () {
-		window.app.console.debug('_onSocketError:');
+	_onSocketError: function (event) {
+		window.app.console.warning('_onSocketError:', event);
```
eszkadev, line 1653: shouldn't be: `window.app.console.warn` ?
example: https://github.com/CollaboraOnline/online/blob/3b7a26e15f866fa53fe3e49c01432b3d653a38f1/browser/src/core/Socket.js#L994

hfiguiere: Indeed. I always trip on this

https://developer.mozilla.org/en-US/docs/Web/API/console/warn_static

PR #11291, opened 2025-03-07
output from writerand is unused

caolanm:
so doesn't need to be written.

See man 1 openssl for "Random State Options", -rand option reads the output from -writerand and that's unused here and apparently unnecessary since OpenSSL 1.1.1

Change-Id: Ia3709f993b71644be8f0b816a88212efc7b76092

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
I don't think this line is doing anything and I don't think we need it.

timar (approved):

PR #11214, opened 2025-02-20
feat(iOS): Support loading video

Minion3665:
Previously, as videos are normally loaded via a /media/ endpoint on the
server, mobile apps did not support loading it. That's because mobile
apps lack a server.

To get around either making a server (opening up HTTP just to serve user
media) or embedding the content directly into the page (as data:, losing
control of when the content was loaded and losing the ability to update
to partial loading - essential for long videos), I've opted to use a
custom protocol handler. Effectively, requests to "cool:/cool/media" are
treated as iOS media requests. Sadly this technique won't immediately
work on Android as it doesn't seem to support custom protocol handlers
so I'll possibly have to make up something else over there (it'll
probably be best to use a similar technique though).

I've also implemented support for the Range header, which will allow us
to only load part of the video (à la https://github.com/CollaboraOnline/online/pull/7367)

Signed-off-by: Skyler Grey <skyler.grey@collabora.com>

PR #11292, opened 2025-03-07
feat(slideshow): Add touch gesture support

Minion3665:
Previously the only way to control a presentation from a touchscreen was
by tapping to go forward - one slide at a time until you got to the end
where you could tap to exit. This is "non-ideal" for web touchscreen
users, and a blocker on adding support for the new presentation code to
the mobile apps.

I've added some code using hammer.js to detect swipes and act on them,
mirroring some of the keyboard actions. I've done my best to replicate
existing expectations (swipe forwards to go forwards, swipe backwards to
go backwards, swipe up-or-down to exit) from the previous presentation
code.

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -11,6 +11,8 @@ window.app = {
 	console: {}
 };
 
+// For typings (including the global object), please see browser/src/global.d.ts
+
 // This function may look unused, but it's needed in WASM and Android to send data through the fake websocket. Please
 // don't remove it without first grepping for 'Base64ToArrayBuffer' in the C++ code
 // eslint-disable-next-line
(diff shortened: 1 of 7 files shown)


eszkadev:
I see, this is extracted part needed for: https://github.com/CollaboraOnline/online/pull/11214

Minion3665:
@eszkadev do you believe this patch to be dangerous enough it needs to be tagged with CODE "red" - or is the retag more because it's an extracted part of something that is CODE "red"?

While it was written *for* #11214, there is no dependency on it and it is also a benefit to *any* touchscreen users, not just those using the new video in the iOS app: the webview always uses new presentations, including on various tablets, phones and 2-in-1s

eszkadev:
The general rule is to have only bugfixes and important tested improvements in GREEN.
As it will be tested with mentioned feature - let's keep them together in RED.

eszkadev (approved):
looks ok

PR #11293, opened 2025-03-07
assuming a readonly root, if we gen certs, don't overwrite the originals

caolanm:
Change-Id: Id1077ff63203f74f4802af087558a6ae652fbcf1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/docker/from-packages/scripts/start-collabora-online.sh
+++ b/docker/from-packages/scripts/start-collabora-online.sh
@@ -20,10 +20,11 @@ else
 openssl req -key certs/servers/localhost/privkey.pem -new -sha256 -out certs/tmp/localhost.csr.pem -subj "/C=DE/ST=BW/L=Stuttgart/O=Dummy Authority/CN=${cert_domain}"
 fi
 openssl x509 -req -in certs/tmp/localhost.csr.pem -CA certs/ca/root.crt.pem -CAkey certs/ca/root.key.pem -CAcreateserial -out certs/servers/localhost/cert.pem -days 9131
-mv -f certs/servers/localhost/privkey.pem /etc/coolwsd/key.pem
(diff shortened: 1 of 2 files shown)


caolanm:
If we do this, then we can get a step closer to a running under docker --read-only

timar:
> If we do this, then we can get a step closer to a running under docker --read-only

Is this because /tmp is in memory so it does not count?
But I still don't understand. With your patch we do not overwrite the cert files in /etc/coolwsd, but we effectively override them, because command line settings take precedence over /etc/coolwsd/coolwsd.xml setting. Maybe this scenario makes no sense (i.e. when we have something in /etc/coolwsd and we generate cert. So maybe it's OK...

As a side note, I always wanted to get rid of this startup shell script. (I expect next time we will be asked to start coolwsd directly, remove shell as an attack vector etc.)

caolanm:
> Is this because /tmp is in memory so it does not count?

With docker --read-only (from man docker-run) apparently you get an implied additional option that is "When running --read-only containers, mount a read-write tmpfs on /dev, /dev/shm, /run, /tmp, and /var/tmp. The default is true". So /tmp is still writable unless there is an additional --read-only-tmpfs=false at which point "In this mode writable directories need to be added via external volumes or mounts".  The report, so far, has been a permission failure on the cp to the final dest, not the initial generation in /tmp, so I think its safe to assume a writeable /tmp (and /var/tmp), but not a writeable /etc

> With your patch we do not overwrite the cert files in /etc/coolwsd, but we effectively override them, because command line settings take precedence over /etc/coolwsd/coolwsd.xml setting.

That's my thinking here.

timar (approved):

PR #11295, opened 2025-03-07
host child_root_path in /var/tmp/cool instead of /opt/cool

caolanm:
for readonly docker

Change-Id: Iabee2bbe40b27ee33e69827f44a0d2cde238278c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
recommend -v to bind /opt/cool/child-roots/ and /opt/cool/cache to something writable instead if using --read-only to have less difference between docker image and a normal install

PR #11296, opened 2025-03-07
ci: stability fixes

eszkadev:
- reduces races
- better checks to be sure we are ready for the next actions


caolanm (approved):
Anything that might help is worth it :-)

PR #11299, opened 2025-03-10
wsd: do not enqueue client messages if disconnected

Ashod:
When enqueueing messages to send to a
client, it's not enough to check for
isCloseFrame(). It is possible
that the socket is shutdown, but the
close frame is not set.

isCloseFrame() is set onDisconnect(),
so it should implicitly be equivalent
to isDisconnected(). However, to be
extra safe and avoid edge cases where
the socket is disconnected but
isCloseFrame() returns false, we
don't add more messages in the queue.

### DIFF
--- a/wsd/ClientSession.cpp
+++ b/wsd/ClientSession.cpp
@@ -2826,7 +2826,7 @@ bool ClientSession::forwardToClient(const std::shared_ptr<Message>& payload)
 
 void ClientSession::enqueueSendMessage(const std::shared_ptr<Message>& data)
 {
-    if (isCloseFrame())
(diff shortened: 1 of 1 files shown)


caolanm (approved):
This sounds reasonable to me

mmeeks (changes requested):
Stabbing in the dark - the problematic log has:

disconnected: false

in the debug output - along with 655k duplicate items in the queue. We should do more code reading of the paths that generate this I think.

We should also update the debugging not to overwhelm the logs - and to show the first 20 and last 20 of the queue items  I guess instead ...

PR #11302, opened 2025-03-10
Log: Allow to set the logToFileUiPath

meven:
Change-Id: Ib4b7cc71db1fc7ddb0a9726962d493780a43fbd8

* Resolves: # 
* Target version: master 

### Summary

Currently logging_ui_cmd.file.property.path setting values is not used 

It is used then used to save data from `LogUiCmd` class.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/common/Log.cpp
+++ b/common/Log.cpp
@@ -691,7 +691,14 @@ namespace Log
         AutoPtr<Channel> channelUILog;
         if (logToFileUICmd)
         {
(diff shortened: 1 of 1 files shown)


hfiguiere:
how does this work?

mmeeks:
Looked at merging this just before green - but no documentation, no changes in the config file (which would be normal) and so on -> un-merge-able. Cryptic one line commit message what is this "logToFileUiPath" and how would anyone reading the patch know what it is ? Please fix it.

meven:
I am sure this is correct and I don't have time to spend more on it.

PR #11303, opened 2025-03-10
browsersetting: don't fail to load if browsersetting json is broken

Rash419:
Change-Id: I2e9e1b36c34f349fe84f7e9240606e93b87fe7b1

* Target version: master


caolanm (approved):
Is this a problem you've seen happening or reports of?

Issue #11255, opened 2025-02-28
perf: re-allocate canvas' from a pool of canvases ...


mmeeks:
The new warning dialog that we didn't get back the the mainloop is really useful, helped me catch this beastie. @Cwiiis I think you warned about this:

![Image](https://github.com/user-attachments/assets/8ccfa17a-8421-449f-b6d8-0981d2df771c)

900+ms freeing canvas' it seems - which we could pool and just re-use; of course we could free them when they're no longer needed I guess but ... easy to reproduce after scrolling around a calc sheet a bit and then zooming.

@Minion3665 this could easily be one of the problems around pinch-to-zoom on mobile I guess; I expect it's also not horribly hard to solve - instead of the high/low watermark thing - we should

* allocate on demand up to <N> canvas' (the high watermark I guess) taking from the pool if there is one (and wiping it ?)
* free by returning tile canvases to the pool down the low watermark - I expect the "which tiles to evict" algorithm is worth re-using
* actually free re-size free canvases from the pool when we idle / after some minute of inactivity or whatever - and wind that number right down - perhaps just to what is visible on the screen and a small rectangle around it.

Is that do-able ? =)

mmeeks:
WIP patch here - https://github.com/CollaboraOnline/online/compare/private/mmeeks/tilereuse?expand=1 or - https://github.com/CollaboraOnline/online/pull/11259

Would appreciate help taking it to completion :-)

Minion3665:
heya!

sure, I'll take a look once I finish out #11214 - thanks for the mention 😄!

Cwiiis:
I've updated the patch - to be honest, I think it was already most of the way there to a good first implementation, I haven't changed much.

Mainly, I added trim/clear methods to the cache and made use of those in the garbage collection function. I reused the existing low/high watermarks, so theoretically, we've doubled the amount of canvas memory we're happy to keep around... But to be honest, the numbers seem very low anyway so that probably needs a more detailed review based on real-world testing. The cache gets entirely cleared in the places we do more drastic things to free memory.

For kicks, I added a call to pruneTiles when the document goes idle - I don't think this is exactly what you meant, but it seems like a reasonable thing to do anyway. I don't know that it really makes sense to start evicting from the cache on a more generic idle - if anything, coming back from a short-term idle is a situation we want to feel responsive, so caches should be maintained and ready. It does make sense to evict on document idle to me though (which is a more long-term idle).

[...]

PR #11304, opened 2025-03-10
cool#11255 - don't free tile canvases, but queue them for re-use

Cwiiis:
* Resolves: #11255 
* Target version: master 

### Summary

Return canvas elements to a pool for reuse instead of always creating them on-demand and destroying them immediately.

### Checklist

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -130,6 +130,7 @@ class Tile {
 	coords: TileCoordData;
 	current: boolean = true; // is this currently visible
 	canvas: any = null; // canvas ready to render
+	ctx: CanvasRenderingContext2D | null = null; // canvas context to render onto
 	imgDataCache: any = null; // flat byte array of canvas data
 	rawDeltas: any = null; // deltas ready to decompress
 	deltaCount: number = 0; // how many deltas on top of the keyframe
@@ -195,6 +196,81 @@ class Tile {
 	}
 }
 
+class CanvasItem {
+	canvas: HTMLCanvasElement | null = null;
+	ctx: CanvasRenderingContext2D | null = null;
+}
+
+// Allocating and freeing canvas' is surprisingly expensive and
+// can block rendering for long periods, so re-use canvas' instead.
+class CanvasCache {
+	private _tileSize: number;
+
+	private _canvasList: CanvasItem[] = [];
+
+	constructor(tileSize: number) {
+		this._tileSize = tileSize;
+	}
+
+	get size() {
+		return this._canvasList.length;
+	}
+
+	acquireCanvas(tile: Tile): CanvasRenderingContext2D | null {
+		let item: CanvasItem;
+
+		if (this._canvasList.length === 0) {
+			item = new CanvasItem();
+
+			// This allocation is usually cheap and reliable,
+			// getting the canvas context, not so much.
+			item.canvas = document.createElement('canvas');
+			item.canvas.width = this._tileSize;
+			item.canvas.height = this._tileSize;
+
+			// So we need to grab the context too ...
+			item.ctx = item.canvas.getContext('2d');
+			// handle null item.ctx higher up
+		} else item = this._canvasList.pop();
+
+		tile.canvas = item.canvas;
+		tile.ctx = item.ctx;
+
+		return item.ctx;
+	}
+
+	releaseCanvas(tile: Tile) {
+		var item: CanvasItem = new CanvasItem();
+		item.canvas = tile.canvas;
+		item.ctx = tile.ctx;
+		tile.canvas = null;
+		tile.ctx = null;
+		this._canvasList.push(item);
+
+		tile.imgDataCache = null;
+	}
+
+	trim(limit: number) {
+		while (this._canvasList.length > limit) {
+			const item = this._canvasList.pop();
+
+			// Fix for cool#5876 allow immediate reuse of canvas context memory
+			// WKWebView has a hard limit on the number of bytes of canvas
+			// context memory that can be allocated. Reducing the canvas
+			// size to zero is a way to reduce the number of bytes counted
+			// against this limit.
+			item.canvas.width = 0;
+			item.canvas.height = 0;
+
+			delete item.canvas;
+		}
+	}
+
+	clear() {
+		this.trim(0);
+	}
+}
+
 class TileManager {
 	private static _docLayer: any;
 	private static _zoom: number;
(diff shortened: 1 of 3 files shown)


browser/src/app/TilesMiddleware.ts
```diff
@@ -331,24 +406,25 @@ class TileManager {
 				if (!tile.current) this.removeTile(keys[i]);
 			}
 		}
+
```
mmeeks, line 409: Hmm - I was really trying to avoid any 'real' freeing of canvases during active use. We should 'free' back onto the free list and presumably the code will stop us from allocating new canvases > than that limit (?)

browser/src/core/Socket.js
```diff
@@ -1001,6 +1001,7 @@ app.definitions.Socket = L.Class.extend({
 
 			if (textMsg === 'idle' || textMsg === 'oom') {
 				app.idleHandler._dim();
+				TileManager.pruneTiles();
```
mmeeks, line 1004: Does this do a real free ? - I think we should have something expensive and grim that really blows the cache - but I fear we do 'pruneTiles' on all sorts of places where we don't want a 900ms sleep ;-) for example on sheet switching ( can you audit all of those - I don't know what pruneTiles was supposed to do - but massive-delay-method was not the intention ;-) it should never actually set a canvas.width=0 which is the stop-world-method it seems :-)

I expect we would need a new method for the real "freeAllBackingCanvases" or something that is sensible; and quite possibly we should do that in some timeout anyway a few at time to be a good citizen and not kill the browser somewhere else (?) ;-)

mmeeks (changes requested):
Not totally convinced. I think we need to read through the pruneTiles semantic and make sure that we call 'trim' and 'clear' at suitable places etc.

the 'garbageCollect' path should probably not be overloaded to try to cleanup and free all canvas/tile memory - since it's so expensive if some well-intentioned person pushes that into the hot path by mistake.

browser/src/core/Socket.js
mmeeks: When I last read it - garbageCollect should not then be actually freeing canvases - that's surely much of the point of the patch & the change ? =) so we avoid the stalling - it is far better to re-use a canvas than to spend time freeing it - so pushing it to the free list and re-allocating it to a more useful user.

Against that we do need a real mechanism to actually free canvas memory when we are truly idle (ie. not when when we switch sheets etc.). So - again - "I expect we would need a new method for the real ..." (as above). :-) I'd really like not to have 250 (80Mb), 125 (40Mb) or whatever sitting around in idle / unused tabs: that seems likely to make other tabs fail to be able to get the texture memory they use surely.

Cwiiis: hmm, testing this though, `pruneTiles` is called sometimes when you drag the scroll handle... Which I don't think should be happening, so looking into that!

browser/src/layer/tile/CanvasTileLayer.js
```diff
@@ -814,7 +814,8 @@ L.CanvasTileLayer = L.Layer.extend({
 			if (!L.Browser.mobileWebkit)
 				TileManager.update(this._map.getCenter(), tileZoom);
 
-			TileManager.pruneTiles();
+			if (tileZoomChanged)
+				TileManager.pruneTiles();
```
Cwiiis, line 818: This change fixes `pruneTiles` being called during the course of normal document navigation.

browser/src/app/TilesMiddleware.ts
```diff
@@ -259,6 +334,10 @@ class TileManager {
 		if (this.debugDeltas)
 			window.app.console.log('Garbage collect! iter: ' + this.gcCounter);
 
+		// In case garbage collection was forced outside of maybeGarbageCollect, make sure future
+		// garbage collection doesn't happen prematurely.
+		this.gcCounter += (53 - (this.gcCounter % 53)) % 53;
```
Cwiiis, line 339: This change stops garbage collection happening prematurely in the case that it gets forced. It still happens periodically during scrolling of the document, but that's probably ok.

browser/src/app/TilesMiddleware.ts
Cwiiis: I'm not keen on this either, but with the right numbers, this ought to do nothing. Presumably, the right number is likely `highNumCanvases - lowNumCanvases` (the amount of canvases regularly being returned to the pool due to garbage collection), maybe with some safety buffer... Given that `highNumCanvases` is currently double `lowNumCanvases`, this means this is probably about right I think, though a comment explaining this is probably warranted (and explicit testing).

browser/src/core/Socket.js
Cwiiis: Fixed, probably :)

Cwiiis: > When I last read it - garbageCollect should not then be actually freeing canvases - that's surely much of the point of the patch & the change ? =) so we avoid the stalling - it is far better to re-use a canvas than to spend time freeing it - so pushing it to the free list and re-allocating it to a more useful user.
> 
> Against that we do need a real mechanism to actually free canvas memory when we are truly idle (ie. not when when we switch sheets etc.). So - again - "I expect we would need a new method for the real ..." (as above). :-) I'd really like not to have 250 (80Mb), 125 (40Mb) or whatever sitting around in idle / unused tabs: that seems likely to make other tabs fail to be able to get the texture memory they use surely.

Missed this, but currently making changes that I think at least partially address it.

Cwiiis: I've added a `discardAllCache` method to TilesMiddleware that is the same as pruneTiles, except it forces the garbage collection limits to zero so *all* non-current tiles, canvases and deltas are discarded. I don't know that we really want to do that last one necessarily, maybe it'd make more sense to just trim it significantly, but this ought to now vastly reduce memory use (both GPU and main memory) when a document goes idle.

This currently piggy-backs on the Socket/server idle - whether we want to introduce another idle signal that can call pruneTiles to do a less aggressive version of this is a good question, but I don't have any intuition on what ought to be considered 'idle' and we are essentially periodically doing this as we draw the document anyway.

browser/src/app/TilesMiddleware.ts
Cwiiis: Verified this and left a comment in the code - under normal situations (which I also briefly tested), no canvases will be freed. I don't see a path in which cached canvases will get freed during normal use except in some kind of exceptional circumstance where we suddenly allocate a bunch of tiles considerably over the high canvas limit (and in that situation, I think having a failsafe is a good idea).

Cwiiis: I did also change the values used to explicitly work under this assumption - so now we keep the canvas cache to `highNumCanvases - lowNumCanvases` and we trim it when it reaches above a multiple of that (I arbitrarily picked 1.5, that seemed reasonable).

mmeeks:
Hopefully it helps with tab switching etc. @eszkadev is CI working or can we expect these failing tests ? =)

PR #11306, opened 2025-03-10
browsersetting: send multiple preference to server at once

Rash419:
- There are multiple sidebar setting/preference are set by client when user opens the sidebar.
- This patches avoid to send preference individually to each time.
- Ths will also the reduce the amount of upload request to wopihost.

Change-Id: I62d3000493f9d175b9190dffe7fefff79b8f976e

* Target version: master


caolanm (approved):
looks ok to me

PR #11308, opened 2025-03-10
fix(MacOS): Allow Mac users more system keybinds

Minion3665:
Online doesn't know what keybinds it has. Instead, it leaves knowing that up to core. This means we preventDefault() on keybinds relatively indiscriminately, having explicit carve-outs for keybinds we want to execute - such as Ctrl+Shift+I to open the browser devtools, and sending everything to core.

Sometimes, we may have keybinds which don't have a core equivalent - for example Cmd+Ctrl+Space. In this case the keybind is ignored in core and Everything Is Fine... mostly. The problem is when these bindings collide with system bindings. Cmd+Ctrl+Space happens to be the binding on MacOS to open the emoji picker... which previously never opened in Collabora Online. Another example is the MacOS minimize keybind, Cmd+M, which appeared to do nothing but was preventing a MacOS system shortcut in some browsers.

Similarly, sometimes there are keybinds which, while interpreted as some action in core, have a well-understood meaning for the system. An example is Cmd+R or Cmd+M (refresh tab, used in LibreOffice as "justify text right"). This is such an obviously-bad thing to bind over that Safari won't let you!.. but Chrome and Firefox both will, causing some nasty browser inconsistencies and breaking users' expectations of how the web works.

Don't do that.

Change-Id: I3b5c3390e5eb88e8e38e552fa424cedd946f65e9

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.Keyboard.js
+++ b/browser/src/map/handler/Map.Keyboard.js
@@ -406,7 +406,6 @@ L.Map.Keyboard = L.Handler.extend({
 		}
 
 		if (window.KeyboardShortcuts.processEvent(app.UI.language.fromURL, ev)) {
-			ev.preventDefault();
 			return;
 		}
 		if (this._map.jsdialog
@@ -493,7 +492,6 @@ L.Map.Keyboard = L.Handler.extend({
 
 		if (window.KeyboardShortcuts.processEvent(app.UI.language.fromURL, ev)) {
 			ev.shortCutActivated = true;
-			ev.preventDefault();
 			return;
 		}
 		if (!keyEventFn && docLayer && docLayer.postKeyboardEvent) {
@@ -672,11 +670,6 @@ L.Map.Keyboard = L.Handler.extend({
 
 		// Control + INSERT
 		if (this._isCtrlKey(e) && e.keyCode === this.keyCodes.INSERT) {
-			return true;
-		}
-
-		// Control + Shift + I, open browser developper tools
-		if (this._isCtrlKey(e) && e.shiftKey && e.keyCode === this.keyCodes.I) {
 			return true;
 		}
 
--- a/browser/src/map/handler/Map.KeyboardShortcuts.ts
+++ b/browser/src/map/handler/Map.KeyboardShortcuts.ts
@@ -42,6 +42,7 @@ class ShortcutDescriptor {
     unoAction: string;
     dispatchAction: string;
     viewType: ViewType;
+    preventDefault: boolean;
 
     constructor({
         docType = null,
@@ -52,6 +53,7 @@ class ShortcutDescriptor {
         unoAction = null,
         dispatchAction = null,
         viewType = null,
+        preventDefault = true,
     }: {
         docType?: string,
         eventType: string,
@@ -61,6 +63,7 @@ class ShortcutDescriptor {
         unoAction?: string,
         dispatchAction?: string,
         viewType?: ViewType,
+        preventDefault?: boolean,
     }) {
         app.console.assert(keyCode !== null || key !== null, 'registering a keyboard shortcut without specifying either a key or a keyCode - this will result in an untriggerable shortcut');
 
@@ -72,6 +75,7 @@ class ShortcutDescriptor {
         this.unoAction = unoAction;
         this.dispatchAction = dispatchAction;
         this.viewType = viewType;
+        this.preventDefault = preventDefault;
     }
 }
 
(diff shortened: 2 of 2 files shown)


browser/src/map/handler/Map.Keyboard.js
```diff
@@ -680,6 +680,24 @@ L.Map.Keyboard = L.Handler.extend({
 			return true;
 		}
 
+		// Cmd + Ctrl + Space on a Mac/iPad, open system emoji picker
+		if (e.keyCode === this.keyCodes.SPACE
```
eszkadev, line 684: can we use new approach for it: https://github.com/CollaboraOnline/online/blob/master/browser/src/map/handler/Map.KeyboardShortcuts.ts

Minion3665:
emoji picker is still capital-w Weird... it opens on Chrome and Firefox but not Safari, but it doesn't look like the keybind is being preventdefaulted... and then Chrome can insert emojis properly but Firefox needs you to type additional characters...?

:(

---

that'll probably need more fixes even beyond the keybind work here...

browser/src/map/handler/Map.KeyboardShortcuts.ts
```diff
@@ -88,11 +94,14 @@ class KeyboardShortcuts {
         const viewType = this.map.isEditMode() ? ViewType.Edit : ViewType.ReadOnly;
 
         const shortcuts = descriptors.filter((descriptor: ShortcutDescriptor) => {
+            const keyMatches = descriptor.key === key;
+            const keyCodeMatches = Array.isArray(descriptor.keyCode) ? descriptor.keyCode.includes(keyCode) : descriptor.keyCode === keyCode;
```
eszkadev, line 175: keyCode cannot be array, not needed?

Minion3665: keyCode must be an array to get some keys correctly - such as how any number on the keypad is treated as KP_9 rather than 9, say. Previously I was defining the keyCodes as an object as they were in Map.Keyboard.js, so it was obviously needed due to the typing system, so I added it

At present it isn't needed and I guess I could drop it and someone could extend it later if they want it again

browser/src/map/handler/Map.KeyboardShortcuts.ts
```diff
@@ -267,6 +267,12 @@ keyboardShortcuts.definitions.set('default', new Array<ShortcutDescriptor>(
     // Passthrough some system shortcuts
     new ShortcutDescriptor({ eventType: 'keydown', modifier: Mod.CTRL | Mod.SHIFT, key: 'I', preventDefault: false, platform: Platform.WINDOWS | Platform.LINUX }), // Open browser developer tools on Non-MacOS - shift means the I here is capital
     new ShortcutDescriptor({ eventType: 'keydown', modifier: Mod.CTRL | Mod.ALT, keyCode: 73 /* keyCode('I') === 73 */, preventDefault: false, platform: Platform.MAC }), // Open browser developer tools on MacOS - registered with keyCode as alt+i triggers a dead key on MacOS
+    new ShortcutDescriptor({ eventType: 'keydown', modifier: Mod.CTRL, key: 'r', preventDefault: false }), // Refresh browser tab
```
eszkadev, line 326: Do we want to allow refreshing the window?

eszkadev:
fails ` cy:command ✘  uncaught exception	TypeError: Cannot read properties of undefined (reading 'events')`

Minion3665:
turns out CI was failing because cypress uses duplicate checking code, which was inadvertently calling the find code with nulls sometimse - [tsconfig option strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) could have saved her

eszkadev:
> turns out CI was failing because cypress uses duplicate checking code, which was inadvertently calling the find code with nulls sometimes - [tsconfig option strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) could have saved her

can we do this in master now as we switched base to 25 already?

Best - without any big rework of all the files. Just a subset which will be extended like we did with prettier.
I hope we can use this trick: https://stackoverflow.com/questions/44676944/how-to-compile-a-specific-file-with-tsc-using-the-paths-compiler-option#44748041
to do such list with more strict params?

Minion3665:
I tried doing this before but unilaterally, and it failed pretty spectacularly - too big of a change to sensibly maintain... the "I hope we can use this trick" trick is a very good shout :)

eszkadev (approved):
looks good, thanks

Minion3665:
> I hope we can use this trick: [stackoverflow.com/questions/44676944/how-to-compile-a-specific-file-with-tsc-using-the-paths-compiler-option#44748041](https://stackoverflow.com/questions/44676944/how-to-compile-a-specific-file-with-tsc-using-the-paths-compiler-option#44748041)
> to do such list with more strict params?

A PR doing a similar thing is open here: #11527, although it is not possible to use this trick to do so and I have instead used a 3rd-party compiler plugin.

PR #11310, opened 2025-03-11
cool#11226 sw per-view redline on: add tri-state UI in the notebookbar

vmiklos:
Have 2 views, both are typing. View 1 turns on recording of tracked
changes, now surprisingly this is also turned on in view 2.

If we record track changes were part of the model previously, now this
is optionally part of the view.

Fix the problem by adding explicit UI to have recording 'off', or 'on'
for a single view / all views.

Requires core.git co-24.04 commit
a67519e2053fae13945df64c959e975bb1298833 (cool#11226 sw per-view redline
on: support this-view <-> all-views transition, 2025-03-10).

Change-Id: I1d5f1079e3b705b6844516a6938f54b446c0c261

### DIFF
--- a/browser/src/control/Control.NotebookbarWriter.js
+++ b/browser/src/control/Control.NotebookbarWriter.js
@@ -2263,8 +2263,8 @@ L.Control.NotebookbarWriter = L.Control.Notebookbar.extend({
 				'vertical': 'true'
 			},
 			{
-				'id': 'review-track-changes',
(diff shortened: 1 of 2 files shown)


vmiklos:

This just turns a toggle button into a menu, so you can switch between these 3 states, added recently in core.

caolanm (approved):
looks ok to me

PR #11313, opened 2025-03-11
Better OOM Handling

Ashod:
Mostly logging and state-dumping improvements, with only one functional change where we now upload when previously we had skipped it. The commit log (and comments) explain the case in question, which is rather rare but gets us stuck.

- **wsd: better admin OOM closing**
- **wsd: dump Storage state**
- **wsd: log the URI that we failed to connect to**
- **wsd: detect inconsistent lastModifiedTime in the managers**
- **wsd: allow uploading when otherwise there might be no need**

### DIFF
--- a/wsd/Admin.cpp
+++ b/wsd/Admin.cpp
@@ -13,6 +13,7 @@
 
 #include <chrono>
 #include <csignal>
+#include <cstdint>
 #include <iomanip>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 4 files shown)


wsd/Admin.cpp
```diff
@@ -1027,35 +1028,43 @@ void Admin::triggerMemoryCleanup(const size_t totalMem)
             memLimit << "% (" << static_cast<size_t>(_totalSysMemKb * memLimit / 100.) << " KB).");
 
     const double memToFreePercentage = (totalMem / static_cast<double>(_totalSysMemKb)) - memLimit / 100.;
-    int memToFreeKb = clamp<double>(memToFreePercentage * _totalSysMemKb, 0, std::numeric_limits<int>::max());
-    // Don't kill documents to save a KB or two.
-    if (memToFreeKb > 1024)
+    int64_t memToFreeKb =
+        clamp<double>(memToFreePercentage * _totalSysMemKb, 0, std::numeric_limits<int>::max());
```
mmeeks, line 1032: Quite a lot of fairly pointless churn here that took most of the time to review... hey-ho. If it could be split out into a pure re-factor commit in the 1st-case that just changes the indent that would be nice - with logic changes in a follow-on that would be good.

mmeeks (approved):
Looks plausible - I would really like a unit test for the changed code-path around saving on up-load failure; surely that could be tacked onto an existing test ?

Would be nice to have in green @timar

wsd/DocumentBroker.cpp
```diff
@@ -2321,6 +2321,19 @@ bool DocumentBroker::isStorageOutdated() const
             << " and the last uploaded file was modified at " << lastModifiedTime << ", which are "
             << (currentModifiedTime == lastModifiedTime ? "identical" : "different"));
 
+#ifdef ENABLE_DEBUG
+    if (_storageManager.getLastUploadedFileModifiedTime() != _saveManager.getLastModifiedTime())
+    {
+        const std::chrono::steady_clock::time_point now = std::chrono::steady_clock::now();
+        LOG_ERR("StorageManager's lastModifiedTime ["
```
caolanm, line 2328: This seems to fire for me everytime I open and close a document in nextcloud.

e.g.: StorageManager's lastModifiedTime [Wed Mar 19 20:14:08.765 2025 (6s 920ms ago)] doesn't match that of SaveManager's [Thu Jan 01 01:00:00.000 1970 (1742415255s 686ms ago)]. File lastModifiedTime: [Wed Mar 19 20:14:15.677 2025 (0s 8ms ago)]| wsd/DocumentBroker.cpp:2328

Ashod: Yes, this is fixed recently. The issue was `#ifdef` vs `#if`.

Issue #11273, opened 2025-03-05
Paragraph style dialog not reachable from sidebar properties


pedropintosilva:
### Describe the Bug

![Image](https://github.com/user-attachments/assets/15641e6f-c7f4-46de-9ee1-b15939718e56)

Recently, there was a change that made the that more button of the current style to open the Styles list (Styles Deck) instead of open the Paragraph Style dialog of the that particular style. 

I'm not sure if this was on purpose, but I'm afraid it's a bug:
* there is no other top level way to access that dialog
  * The format tab > Paragraph triggers the Paragraph dialog not the Paragraph Style dialog
  * The remaining way to open the Paragraph Style dialog of the current style is now less obvious and with more clicks when starting from sidebar properties: Click more button; Find/scroll selected style; Assume you can right click to edit; Right Click; Click `Edit Style`
* all the other sidebar containers have a similar button that open the respective dialog (example character style dialog)

Note: not all more buttons are created equal. The more button in the Home > Styles Preview triggers the Style list and that is expected as it not only is the behaviour found elsewhere but also it conveys that to see more of that list you press that (that's why the button is placed at the widget scroll bar level and not in each preview). 

cc: @eszkadev @banobepascal

eszkadev:
It is a side effect of https://cgit.freedesktop.org/libreoffice/core/commit/?id=90a63a2325e45b435762570ac9aecd50fcb5b676 I assume.

It was for calc (where such option didn't exist before). Maybe we can separate Writer and Calc cases.

eszkadev:
fix in core: https://gerrit.libreoffice.org/c/core/+/182763

PR #11314, opened 2025-03-11
styles list: allow second activation in Calc

eszkadev:
in sidebar there is shortcut icon in Calc to activate style list - use correct command for it to remember selection adn allow to activate it not only first time when clicked


eszkadev:
fixes #11273

PR #11322, opened 2025-03-12
wsd: delay the socket shutdown if request require async operation

Rash419:
- usually happens when we have preset fetch, upload, delete request from admin integrator panel

Change-Id: I4b8124e7302d6352cb542b2d85ded246a9e7c24e

* Resolves: # 
* Target version: master 

### Summary
Tested adding a `sleep(3)` on richdocuments side for wopi/setting path.


wsd/ClientRequestDispatcher.cpp
```diff
@@ -802,11 +802,7 @@ void ClientRequestDispatcher::handleIncomingMessage(SocketDisposition& dispositi
                     launchAsyncCheckFileInfo(_id, accessDetails, RequestVettingStations,
                                              RvsHighWatermark);
                 }
-                servedSync = true;
             }
-
-            if (!servedSync)
```
caolanm, line 808: I think this "(!servedSync) test and setErrorAndShutdown" pair of lines should probably stay, but get moved to before line 789, so it appears at the end of the "if (pos != std::string::npos)" block so whatever used to happen still happens.

caolanm (approved):
That looks good, all error paths seem to be using sendError which do HttpHelper::sendErrorAndShutdown

caolanm (changes requested):
probably need to restore the HttpHelper::sendErrorAndShutdown use for the mentioned case

wsd/ClientRequestDispatcher.cpp
Rash419: Makes sense!

caolanm (approved):
looks good, ci seems to a little confused about the tarball vs what dir it will create

Rash419:
backported: https://github.com/CollaboraOnline/online/pull/11324

PR #11326, opened 2025-03-12
call onFinishCallback on early error cases too

caolanm:
Change-Id: Id4687f3914b26bdaba62e385276e854811a4d432

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


Ashod (approved):
Thanks, @caolanm. Looks reasonable.

PR #11333, opened 2025-03-13
cool#11064 any input: enable COOL_ANYINPUT_CONSIDER_PRIORITY by default

vmiklos:
This wasn't possible previously, because the core side was missing
commit 4f15a47c7faf55997e776af738106d3da229c7d2 (cool#11064 Set priority
of some idles to TaskPriority::DEFAULT_IDLE, 2025-02-18), but now
co-25.04 has it.

The benefit is that now we don't interrupt high priority core jobs now,
and only start painting tiles after the high priority tasks are done,
but still before idle tasks.

Change-Id: I5556fdbc43841a2fbf3e2ea084199c17e04439d6

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -3082,8 +3082,7 @@ bool anyInputCallback(void* data, int mostUrgentPriority)
 
     if (document)
     {
-        static bool considerPriority = std::getenv("COOL_ANYINPUT_CONSIDER_PRIORITY");
-        if (considerPriority && document->isLoaded())
+        if (document->isLoaded())
         {
             // Check if core has high-priority tasks in which case we don't interrupt.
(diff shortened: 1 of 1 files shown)


vmiklos:

In short, COOL used to interrupt core as soon as there were some input/messages to be processed, now we do this in a more careful way: if core has high prio tasks, then we don't interrupt. This is meant to be a middle way: still interrupting before core idles run, but give core enough time to produce all its events from high prio tasks (so we can deduplicate those, etc). This was not done in co-24.04 previously because lots of idle-style tasks actually had high priorities (and fixing that was risky), so blocking on high prio tasks broke all previous optimizations. Now partial layout on doc load or on page insert works even if we consider these priorities on co-25.04.

caolanm (approved):
Lets see how it goes

PR #11334, opened 2025-03-13
status bar: remove right and left arrows

banobepascal:
Change-Id: Ied57f580bc7c1ea5d488e935f5f906f0935bd3cf

* Target version: master 

### Previews
[Screencast from 25-03-25 04:40:39.webm](https://github.com/user-attachments/assets/199992d7-fbc2-4032-97a5-abb1f9bc7544)

### Summary
- Removes the left and right arrows from the status bar to address inconsistencies after removing default states.
- The arrows also block the Zoom "+" which is frustrating to users as they would need to double click the arrow to remove it and click zoom +

### DIFF
--- a/browser/css/jsdialogs.css
+++ b/browser/css/jsdialogs.css
@@ -291,7 +291,7 @@ td.jsdialog > [id^='table-box']:not(.sidebar) {
 #toolbar-down .ui-badge,
 #toolbar-down .unolabel {
 	font-family: var(--cool-font);
-	font-size: var(--default-font-size);
+	font-size: var(--default-font-size) !important;
 	color: var(--color-status-badge) !important;
 }
 
@@ -2024,6 +2024,11 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	background-color: transparent !important;
 	border-color: transparent;
 }
+
+#toolbar-down  .jsdialog .ui-edit-container{
+	place-content: center;
+}
+
 /* Formula Wizard */
 #FormulaDialog .ui-tabs-root {
 	background-color: var(--color-background-tabs-group);
@@ -2295,6 +2300,11 @@ kbd,
 #showcomments-container[default-state='true'],
 #insertmode-container[default-state='true'],
 #permissionmode-container[default-state='true'],
+#documentstatus-container[default-state='true'],
 #statusselectionmode-container[default-state='true'] {
+	display: none !important;
+}
+
+.status-hidden {
 	display: none !important;
 }
--- a/browser/src/control/Control.StatusBar.js
+++ b/browser/src/control/Control.StatusBar.js
@@ -181,11 +181,11 @@ class StatusBar extends JSDialog.Toolbar {
 		this.updateHtmlItem('ShowComments', state ? statemsg : ' ');
 	}
 
-	_generateHtmlItem(id) {
+	_generateHtmlItem(id, dataPriority) {
 		var isReadOnlyMode = app.map ? app.map.isReadOnlyMode() : true;
 		var canUserWrite = !app.isReadOnly();
 
-		return {
+		const item = {
 			type: 'container',
 			id: id + '-container',
 			children: [
@@ -194,7 +194,13 @@ class StatusBar extends JSDialog.Toolbar {
 			],
 			vertical: false,
 			visible: false
-		};
+		}
+
+		if (dataPriority) {
+			item.dataPriority = dataPriority;
+		}
+
+		return item;
 	}
 
 	_generateStateMenuEntry(id, text, selection) {
@@ -216,7 +222,7 @@ class StatusBar extends JSDialog.Toolbar {
 			.filter((item) => { return item.selected; })
 			.map((item) => { return item.text; });
 		var text = selected.length ? selected.join('; ') : _('None');
-		return {type: 'menubutton', id: 'StateTableCellMenu', text: text, image: false, menu: submenu, visible: visible};
+		return {type: 'menubutton', id: 'StateTableCellMenu', text: text, image: false, menu: submenu, visible: visible, dataPriority: 5};
 	}
 
 	_generateZoomItems() {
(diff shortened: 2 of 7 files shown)


eszkadev:
I'm not sure about it. Maybe we should improve detection of case when to hide the arrows?
With the patch in the small windows user cannot access zooming without touch screen.
(it requires shift and scroll wheel - not known by many people)

[resize-2025-03-19_08.23.42.webm](https://github.com/user-attachments/assets/054f99af-50fd-4434-8f71-9ea3e41177db)

eszkadev:
BTW also search bar seems to be misaligned now

banobepascal:
> I'm not sure about it. Maybe we should improve detection of case when to hide the arrows? With the patch in the small windows user cannot access zooming without touch screen. (it requires shift and scroll wheel - not known by many people)
> 
>  resize-2025-03-19_08.23.42.webm

I think  when the window becomes smaller we need to hide some information that might not be so crucial to the user or come up with a different implementation to show the information and with this implementation, we shall not need to scroll the status bar and also the arrows will not be needed.

pedropintosilva:
now testing, I think I wrote somewhere in an issue and to Banobe. But let me write it here, the goal is:

* Make the status bar never need to scroll. Reason: Status bar is not like top toolbar with a series of buttons it's rather a discreet bottom bar that should show the most important status without additional clicks
  * The available controls on the edges should be always visible
  * Remove the scroll indicators
  * Statuses in the middle get hidden depending on the window's width
    * We need to make sure that we agree on a priority, if the elements are already positioned correctly (left most import) we don't need to do extra work, we simply hide it from right to left.
* Follow up enhancements
  * Add button that appears when there are hidden status. That button triggers a new menu widget that shows all those status.

pedropintosilva (changes requested):
Thanks @banobe for removing the right and left arrows and updating the cypress.

For this PR to be ready we need to address the following

>     * Make the status bar never need to scroll. Reason: Status bar is not like top toolbar with a series of buttons it's rather a discreet bottom bar that should show the most important status without additional clicks
>       
>       * The available controls on the edges should be always visible
 >       * Remove the scroll indicators
>       * Statuses in the middle get hidden depending on the window's width
>         
>         * We need to make sure that we agree on a priority, if the elements are already positioned correctly (left most import) we don't need to do extra work, we simply hide it from right to left.

eszkadev:
@pedropintosilva I see, then I'm good with the change.

For priority: possibly we want new property for a toolbar item in the JSON.
No priority - show always.
Priority - number

We did "make scrollable" in the past. There is logic for overflow detection. We can reuse it.
In that case we should do similar thing: hide lowest priority, check overflow, until no-overflow or no more prioritized widgets.

So this way there is no hardcoded id needed :)

pedropintosilva (changes requested):
It looks awesome! However I feel that we cannot hide the search in any case because otherwise the ctrl + F will do nothing,. Furthermore the non-power user will probably seek that search at the bottom (as it is one of the few things that are interactive in the status bar).

* On the left: both search (reasons above) and page number should never disappear (as the user relies on that to know where is at)
  * additionally the search input field has a static width of 200px that should decrease at least to half instead of disappearing completely 
* I like that if in read-only mode that status never disappears, good choice :+1: 
* Also, I noticed that it seems some of the elements (jsdialog menubutton in the status bar) are inheriting some odd font-size which is different that the rest
  * we would probably benefit from using the existing css var --default-font-size
* And please fix the cypress spec

banobepascal:
@pedropintosilva @eszkadev please review

pedropintosilva (commented):
- [x] On the left: both search (reasons above) and page number should never disappear (as the user relies on that to know where is at)
- [x] the search input field has a static width of 200px that should decrease at least to half instead of disappearing completely
- [x] Read-only mode status should never disappear when on read-only mode
- Also, I noticed that it seems some of the elements (jsdialog menubutton in the status bar) are inheriting some odd font-size which is different that the rest
  - [x] we would probably benefit from using the existing css var --default-font-size
- Please fix cypress spec
  - [x] Removed the scroll test
  - Does it fail for you locally?

pedropintosilva (changes requested):
cypress tests need update, for example:

```javascript
 cy:command ✔  cGet	#StateTableCellMenu .arrowbackground
      cy:command ✘  click	
      cy:command ✔  fail:	
                    Test failed: integration_tests/desktop/calc/statusbar_spec.js / Statubar tests. / Selected data summary.
                    
                    Timed out retrying after 10050ms: `cy.click()` failed because this element is not visible:
                    
                    `<div class="arrowbackground">...</div>`
                    
                    This element `<div.arrowbackground>` is not visible because its parent `<div#StateTableCellMenu.unotoolbutton.jsdialog.ui-content.unospan-StateTableCellMenu.StateTableCellMenu.has-label.has-dropdown.menubutton.status-hidden>` has CSS property: `display: none`
```

pedropintosilva (approved):

browser/src/control/Control.StatusBar.js
```diff
@@ -194,7 +194,13 @@ class StatusBar extends JSDialog.Toolbar {
 			],
 			vertical: false,
 			visible: false
-		};
+		}
+
+		if (dataPriority) {
+			item['data-priority'] = dataPriority;
```
eszkadev, line 200: can we avoid `-` ? (camelCase instead)
It will be less problematic later, eg when using as `item.dataPriority`

browser/src/control/Control.StatusBar.js
```diff
@@ -216,7 +222,7 @@ class StatusBar extends JSDialog.Toolbar {
 			.filter((item) => { return item.selected; })
 			.map((item) => { return item.text; });
 		var text = selected.length ? selected.join('; ') : _('None');
-		return {type: 'menubutton', id: 'StateTableCellMenu', text: text, image: false, menu: submenu, visible: visible};
+		return {type: 'menubutton', id: 'StateTableCellMenu', text: text, image: false, menu: submenu, visible: visible, 'data-priority': '5'};
```
eszkadev, line 225: `dataPriority: 5` would be nicer. 5 should be a number

cypress_test/integration_tests/desktop/calc/bottom_bar_spec.js
```diff
@@ -12,8 +12,14 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Calc bottom bar tests.', f
 	it('Bottom tool bar.', function() {
 		cy.cGet('#map').focus();
 		calcHelper.clickOnFirstCell();
-		cy.cGet('#StateTableCellMenu').click();
-		// If it clicks, it passes.
-		cy.cGet('body').contains('.ui-combobox-entry', 'CountA').click();
+		cy.cGet('#StateTableCellMenu').then(($tableCellMenu) => {
+			if ($tableCellMenu.is(':visible')) {
```
eszkadev, line 16: Please avoid conditional checks in cypress, it makes it unstable.
Please to explicit checks for visible element and explicit check if something is hidden or missing.
`cy.cGet('#StateTableCellMenu:visible').click();` <- does what that code does. If it will fail - cypress will print that it was invisible.

We want strict checks, which can be repeated automatically.

cypress_test/integration_tests/desktop/calc/statusbar_spec.js
```diff
@@ -51,10 +51,16 @@ describe(['tagdesktop', 'tagnextcloud', 'tagproxy'], 'Statubar tests.', function
 		cy.cGet('#StateTableCell').should('have.text', 'Average: 10; Sum: 10');
 
 		desktopHelper.makeZoomItemsVisible();
-		cy.cGet('#StateTableCellMenu .unolabel').contains('Average; Sum');
-		cy.cGet('#StateTableCellMenu .arrowbackground').click();
-		cy.cGet('.jsdialog-overlay').should('exist');
-		cy.cGet('.ui-combobox-entry.selected').contains(/Average|Sum/g);
+		cy.cGet('#StateTableCellMenu').then(($tableCellMenu) => {
+			if ($tableCellMenu.is(':visible')) {
```
eszkadev, line 55: no conditional checks

eszkadev (changes requested):
would be nice to address things I pointed to make it more stable for CI and less  error prone

browser/css/jsdialogs.css
```diff
@@ -2024,6 +2024,11 @@ input[type='number']:hover::-webkit-outer-spin-button {
 	background-color: transparent !important;
 	border-color: transparent;
 }
+
+#toolbar-down  .jsdialog .ui-edit-container{
+	place-content: center;
```
eszkadev, line 2029: possibly that should be the default for any `.ui-edit-container`, but for now is fine

PR #11340, opened 2025-03-13
Impress: Fix drag & drop slide pages for reordering.

gokaysatir:
Sometimes the child of the expected element fires the event. Check also parentNode when searching for clicked part.

Issue:
* Cannot re-position a slide by dragging without first selecting it.

Change-Id: I884fc2908ef6523b0f2ad58e00b3793fcda1de2e

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


eszkadev (approved):
works as expected, thanks for fix!

PR #11342, opened 2025-03-13
ensure threads are re-started on a failed joinThreads.

mmeeks:
If we got a:

[ kitbroker_002 ] WRN  Failed to ensure we have just one thread for bgsave, we have: 2 synchronously saving| kit/Kit.cpp:1451

Then we would bail out of forkToSave without re-starting the _deltaPool.

So guard against that from all return paths more simply.

Change-Id: I0751b68613832da3b3a2bb96235300d638119035

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/kit/Kit.cpp
+++ b/kit/Kit.cpp
@@ -1427,7 +1427,8 @@ bool Document::forkToSave(const std::function<void()> &childSave, int viewId)
         return false;
     }
 
-    if (!joinThreads())
+    ThreadDropper threadGuard;
+    if (!threadGuard.dropThreads(this))
     {
         LOG_WRN("Failed to join threads before async save");
         return false;
@@ -1505,6 +1506,8 @@ bool Document::forkToSave(const std::function<void()> &childSave, int viewId)
         SigUtil::addActivity("forked background save process: " +
                              std::to_string(pid));
 
+        threadGuard.clear();
+
         SigUtil::dieOnParentDeath();
 
         childSocket.reset();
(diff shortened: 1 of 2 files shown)


caolanm:
The comment here mentions "without re-starting the _deltaPool", but Document::joinThreads can only return false if getLOKit()->joinThreads() fails, in which case it returns early before it does _deltaPool.stop() ?

If that joinThreads returns false, then presumably _deltaPool and the possible SocketPoll::PollWatchdog are still running.  With this change we will always call _deltaPool.start() even if we haven't called _deltaPool.stop() and they both look like they assume they are not running if requested to start

mmeeks:
> The comment here mentions "without re-starting the _deltaPool", but Document::joinThreads can only return false if getLOKit()->joinThreads() fails, in which case it returns early before it does _deltaPool.stop() ?

It's not joinThreads() returning false that is the problem - it is returning true - but then us later detecting one of several early exit conditions from Document::forkToSave - that don't get around to re-starting the critical threads.

> If that joinThreads returns false, then presumably _deltaPool and the possible SocketPoll::PollWatchdog are still running. With this change we will always call _deltaPool.start() even if we haven't called _deltaPool.stop() and they both look like they assume they are not running if requested to start

_deltaPool.start() is fine assuming both start and stop are called in the same thread - which is the casel
PollWatchdog->startThread() is a good catch - with the same caveat it could use some protection from re-starting the same thread un-necessarily. It is possible that the core hook to re-start threads is also not happy with being called back to back.

Quite probably we should just not attempt a re-start if the core 'joinThreads' fails - which is an easy tweak =)

caolanm (approved):
Yeah, got it now. Content.

mmeeks:
I did a quick code-read of the core code around re-starting threads on failure - and it looks safe to stop/start at various points in the failure path there.

PR #11346, opened 2025-03-14
cypress: make click on first cell more stable

eszkadev:
this is most failing piece, it seems we don't wait for correct update on formulabar or address input


caolanm (approved):
I think I see the writer one in 24.04 as well, e.g. https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-24.04_cypress_desktop/5202/console can we get a backport to 24.04 for that too?

I imagine that one might be new since https://github.com/CollaboraOnline/online/pull/11304

PR #11350, opened 2025-03-14
Private/caolan/nullptr crash24

caolanm:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


caolanm:
backport of the fix for the frequent null-deref we see from this DBG_WRN

vmiklos (approved):
Aha, this crashed when the first iteration caught an exception and sessionForUpload was still nullptr, makes sense.

PR #11351, opened 2025-03-14
cypress: tiles middleware 24.04

eszkadev:
backport + additional fix for canvas is null exception


mmeeks (approved):
Looks very sensible to me @Cwiiis ?

Cwiiis:
> Looks very sensible to me @Cwiiis ?

Seems like a reasonable precaution, but are we calling `releaseCanvas` on tiles with a null canvas? I guess I should take a look at that :)

PR #11352, opened 2025-03-14
cypress: fix exception in TilesMiddleware.ts

eszkadev:
additional check which was needed for 24.04 (https://github.com/CollaboraOnline/online/pull/11351)

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -260,8 +260,10 @@ class CanvasCache {
 			// context memory that can be allocated. Reducing the canvas
 			// size to zero is a way to reduce the number of bytes counted
 			// against this limit.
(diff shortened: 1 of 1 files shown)


caolanm (commented):
I wonder now in retrospect, if the "if (!item)" check isn't needed and it was just the item.canvas deref that was the issue. I mean, there's just the one canvasList.push() and the thing pushed there isn't null

eszkadev:
> I wonder now in retrospect, if the "if (!item)" check isn't needed and it was just the item.canvas deref that was the issue. I mean, there's just the one canvasList.push() and the thing pushed there isn't null

might be, I overlooked line number and after running it wasn't reproducible in my local env so didn't dig more...

caolanm:
unit test w/o cypress is a bogus fail after make succeeds

PR #11353, opened 2025-03-14
wsd: fix: browsersetting is uploaded multiple times

Rash419:
- Steps to reproduce:
1. Open 2 document for same user
2. Do a single change like for e.g toggle sidebar
3. Notice in logs browsersetting is uploaded twice

- It happens because syncBrowserSettings was getting called for each docbroker and in syncBrowserSetting we were uploading the browsersetting if we find a session with same userId.

- For above mentioned case we uploaded twice

Change-Id: I5bb1f3a6830bc3bc944e74e4f718c9962958db54

* Target version: master


caolanm (approved):
Makes sense to me.

PR #11358, opened 2025-03-14
Calc: Fix view size issue.

gokaysatir:
Issue: When file/view size is changed, we also check the size of the map. If the file is smaller than the view, we also re-size map and canvas. While switching between small size (many rows are hidden to bottom) / large size views, we need to check the map's and canvas's sizes. Or the small size is kept for large views.

Change-Id: Ie6ebbdda8e5b2c78799cae49b627fa476cea8ae2

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -258,6 +258,8 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		if (limitWidth || limitHeight || extendedLimit)
 			app.sectionContainer.requestReDraw();
+
+		this._syncTileContainerSize();
```
eszkadev, line 262: shoudln't be done only if any change done?
do we redraw first (one line above) then resize again? is that ok?
didn't trigger two renderings?

gokaysatir: Thanks for the catch @eszkadev. I added a boolean variable to avoid duplicate calls to redraw.

gokaysatir:
All desktop tests seem to pass but the result is marked as a failure.

eszkadev:
You are right :) some anomaly https://cpci.cbg.collabora.co.uk:8080/job/github_online_master_debug_vs_co-25.04_cypress_desktop/84/console

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -515,6 +518,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		var scrollSection = app.sectionContainer.getSectionWithName(L.CSections.Scroll.name);
 		scrollSection.stepByStepScrolling = true;
+		this._syncTileContainerSize();
```
eszkadev, line 521: Status is sent very often in calc, mostly it doesn't change the sheet (where I belive issue appears).
Can we add some guard here to not touch sizing too much?

eszkadev:
added also @Cwiiis to double check the code too as it might have impact on rendering/perf

browser/src/layer/tile/CalcTileLayer.js
eszkadev: example: someone else added new sheet, will it force redraw our view? or similar case

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -255,7 +255,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		this._map.fire('scrolllimits', newSizePx.clone());
 
-		if (limitWidth || limitHeight || extendedLimit)
+		if (!this._syncTileContainerSize() && (limitWidth || limitHeight || extendedLimit))
```
Cwiiis, line 258: This doesn't seem right to me, but perhaps I've misunderstood. Shouldn't these be reversed? It sounds like `_syncTileContainerSize` only needs to be called in the case that the unchanged clause is true - but this will now always call it.

Cwiiis (changes requested):
This seems reasonable in theory, but comments need to be addressed. This may be affected by #11371 once that's finished.

browser/src/layer/tile/CalcTileLayer.js
gokaysatir: It returns true if it calls requestReDraw. So if it false, we can reDraw.

browser/src/layer/tile/CalcTileLayer.js
```diff
@@ -424,7 +424,10 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 		if (heightIncreased || widthIncreased) {
 			app.sectionContainer.requestReDraw();
 			this._map.fire('sizeincreased');
+			return true;
 		}
+
+		return false;
```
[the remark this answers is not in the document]
gokaysatir: Added false.

browser/src/layer/tile/CalcTileLayer.js
gokaysatir: It turns out that we don't need the call there.

gokaysatir:
Thanks for the reviews :) I hope we have a better fix now.

browser/src/layer/tile/CalcTileLayer.js
Cwiiis: That isn't quite what I meant - I don't think we should be calling `sync_tileContainerSize` unconditionally here, surely it only makes sense to call it when `limitWidth || limitHeight || extendedLimit` is true? It feels to me like the correct if statement here is `if ((limitWidth || limitHeight || extendedLimit) && !this._syncTileContainerSize())`. If it isn't, I'd like to know why you think differently so I can improve my understanding of what's happening here.

Cwiiis: I'd even go as far as to say actually that relying on the side-effect of a multi-clause if statement is bad practice and the if statement should be unchanged, but the contents of the block should be `if (!this._syncTileContainerSize()) app.sectionContainer.requestReDraw();`

gokaysatir: Restricted document size may result in different view size. That's why we need to call syncTilecontainersize there. restrictDocumentSize calls requestReDraw if the allowed borders change. syncTileContainerSize calls requestReDraw if the size of the HTML elements change. We need those 2 checks i guess. Multi-clause if statement may or may not be a bad practice. That's arguable. I assume you want a rafactoring here, that's not the intention of this commit and i don't think it would be straightforward. I guess we can work on it later.

Cwiiis: That's still not quite what I mean - if I understand this function correctly (which I may not :)), the view size is only changed if `limitWidth || limitHeight || extendedLimit` is `true`. If all of those are false, the view hasn't changed size, so there should be no need to call `_syncTileContainerSize` (which is expensive) - is this not the case?

gokaysatir: newDocWidth is re-set regardless of the value of  "limitWidth" and same is true for newDocHeight. That code got a little chaotic in time i think. My brain stopped functioning there and just called the function.

There is a "shouldRestrict" variable and we can assume that if the size is changed, syncTileContainerSize will request a redraw. But it requests reDraw if only the size is higher than before.

So we turn back to first state and rely on "limitWidth ||...." etc checks.

This code is really hard to follow. Really sorry if i am insisting unnecessarily. I think we strongly need to refactor that part.

gokaysatir: I think we need to refactor syncTileContainer size and shouldn't let it set the CSS variables unnecessarily. I see your point, that's expensive. Then i guess we can add some checks into that function.

Cwiiis: Ok, in that case I think I was misunderstanding this function - thanks for the clarification! You're right, I think some refactoring (and commenting) would really help here, but that's a separate issue. This is fine with me :)

gokaysatir: Thanks for sparing time for the review :)

eszkadev:
IIRC restrict was only used if "we scrolled far away triggering the document area to grow", "deleted/added most far-away cell in the sheet"

PR #11361, opened 2025-03-17
Handle Save Failure While Unloading

Ashod:
This better handles an edge case where saving fails while uploading, as described in the commit log.

- **wsd: more aggressive saving and uploading while unloading**
- **wsd: report how much more time before issuing save/upload**
- **wsd: force saving before unloading if the last save had failed**

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -3506,11 +3506,14 @@ void DocumentBroker::autoSaveAndStop(const std::string& reason)
     {
         // Stop if there is nothing to save.
         const bool possiblyModified = isPossiblyModified();
+        const bool lastSaveSuccessful = _saveManager.lastSaveSuccessful();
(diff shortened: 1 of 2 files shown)


mmeeks (approved):
Looks plausible to me - do we really need to wrappers next to each other timeToNextSave and timeToNextUpload ? =) that do the same thing ?

Ashod:
> Looks plausible to me - do we really need to wrappers next to each other timeToNextSave and timeToNextUpload ? =) that do the same thing ?

The wrappers were introduced (many years ago) because using the base class's members quickly becomes confusing and hard to follow. For example, searching for `lastModifiedTime` that shows up across different context (saving, uploading, file in jail, etc.) becomes unhelpful. The wrappers make the context explicit and grep-friendly. I'm following this approach here when adding new members.

mmeeks:
Fair enough =)

PR #11362, opened 2025-03-17
Better Valgrind Support

Ashod:
Various improvements to getting a cleaner valgrind memcheck report.

- **wsd: don't process logging_ui_cmd config if disabled**
- **wsd: don't process logging config if disabled**
- **wsd: clean up remote font config thread before exiting**
- **wsd: unique_ptr needs the size of the pointee**
- **wsd: addStorageDebugCookie in StorageConnectionManager**
- **wsd: replace WopiStorage::initHttpRequest() with StorageConnectionManager**
- **wsd: Admin class cosmetics**
- **wsd: overwrite coolwsd exit code with unit-test result**
- **wsd: clean up as much as possible before forced exit**
- **wsd: add SigUtil::uninitialize()**
- **wsd: string_view in TileDesc**
- **wsd: privatize ClientDeltaTracker**
- **wsd: log active ClientSession in DocBroker dtor**
- **wsd: uninitialize ClientRequestDispatcher**
- **wsd: check the value of ENABLE_DEBUG**

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4055,6 +4055,8 @@ void COOLWSD::cleanup([[maybe_unused]] int returnValue)
         FileRequestHandler.reset();
         JWTAuth::cleanup();
 
+        Util::forcedExit(returnValue);
+
         TraceDumper.reset();
 
         Socket::InhibitThreadChecks = true;
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -4039,7 +4039,7 @@ std::shared_ptr<TerminatingPoll> COOLWSD:: getWebServerPoll ()
     return WebServerPoll;
 }
 
-void COOLWSD::cleanup()
+void COOLWSD::cleanup([[maybe_unused]] int returnValue)
 {
     try
     {
@@ -4088,7 +4088,7 @@ int COOLWSD::main(const std::vector<std::string>& /*args*/)
     SigUtil::resetTerminationFlags();
 #endif
 
-    int returnValue;
+    int returnValue = EXIT_SOFTWARE;
 
     try {
         returnValue = innerMain();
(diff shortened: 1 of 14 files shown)


caolanm:
"COOL 25.04 unit tests w/o cypress" all fail with "double free or corruption" so there's something not quite right with this pr yet.

Ashod:
> "COOL 25.04 unit tests w/o cypress" all fail with "double free or corruption" so there's something not quite right with this pr yet.

The issue is observed with 25.04 only, not 24.04. Should be fixed.

wsd/ClientRequestDispatcher.hpp
```diff
@@ -25,15 +25,21 @@
...
     {
         StaticFileContentCache["discovery.xml"] = getDiscoveryXML();
     }
 
     typedef std::function<void(bool)> AsyncFn;
 
+    /// Uninitialize static data.
+    static void uninitialize()
+    {
+        StaticFileContentCache.clear();
```
caolanm, line 38: as an aside, it looks like StaticFileContentCache is a std::map with just one entry in it :-)

wsd/COOLWSD.cpp
```diff
@@ -4030,10 +4030,6 @@ int COOLWSD::innerMain()
 
     SigUtil::addActivity("finished with status " + std::to_string(returnValue));
 
-    // At least on centos7, Poco deadlocks while
-    // cleaning up its SSL context singleton.
-    Util::forcedExit(returnValue);
```
caolanm, line 4024: ok, innerMain is only called from main, which always calls cleanup, so move this force exit into cleanup instead and release everything unproblematic and then just skip the ssl deinitialization.

Sounds good, but I guess that means this cleanup path was mostly unused before this change then.

wsd/COOLWSD.cpp
```diff
@@ -4055,6 +4055,8 @@ void COOLWSD::cleanup([[maybe_unused]] int returnValue)
         FileRequestHandler.reset();
         JWTAuth::cleanup();
 
+        Util::forcedExit(returnValue);
```
caolanm, line 4058: ah, and then there is some (currently unknown) glitch with the cleanup path so bail out early for now

caolanm (approved):
I imagine the overall theme is trying to use valgrind for leak checking

wsd/COOLWSD.cpp
Ashod: Exactly. I'm unable to reproduce the double-free reported (on jenkins) with asan. I had found and fixed one (accessing `getWakeupsArray()`, which returns a reference to a static in `SocketPoll::removeFromWakeupArray()` after said static had been destroyed during exit), which wasn't reproducible with 24.04 but easily was with 25.05 (as odd as that is).

wsd/COOLWSD.cpp
Ashod: Yes, was unused for a long time, which is why when I enabled it we saw the double-free errors in jenkins (it had bit-rot). As the comment says, at one point we noticed on CentOS Poco SSL uninitialization was getting deadlocked, so we forcefully exited.

But now I needed to chase leaks, and this is in the way.

wsd/ClientRequestDispatcher.hpp
Ashod: Yes, it was provisioned for other statics as well.

PR #11363, opened 2025-03-17
cool#11347 cypress desktop: restore the last test in writer/scrolling_spec.js

vmiklos:
This was removed in commit 0acb3d515fe4d2a020c91ed833db1c3a72a09cfb
(cypres: fix broken test after switching to 25.04 core branch,
2025-03-14), but the underlying problem is fixed with core.git commit
2490d8fc97b9ec9819755a3b81a6e464431c7cdb (cool#11347 sw lok: avoid view
jump on showing formatting marks, 2025-03-14), so this can be restored.

Change-Id: Ifdabf3a4f7f193dc257568df3bb1e6658422efff


vmiklos:
@caolanm could you please review this when you're back? Thanks.

This restores the lost test, now that the root cause on the core side is fixed.

PR #11364, opened 2025-03-18
Don't cache null canvases in TilesMiddleware.CanvasCache

Cwiiis:
### Summary

`releaseCanvas` can be called on tiles with no canvas, make sure to check a canvas and context are non-null before adding them to the cache. This is follow-up from #11351 

### Checklist


caolanm (approved):
makes sense to me, it didn't feel right that there was null entries in _canvasList in the first place

PR #11366, opened 2025-03-18
fix(multimedia): Insert multimedia at correct size

Minion3665:
Previously we were not getting the size of multimedia we were inserting as we can't use gstreamer to do this job in core and we had no process for doing this in online. That meant that all videos were inserted with a fallback size, which happened to be small and square. As not all videos are small and square, this led to some videos having a box with a considerably different aspect ratio than the video itself.

This patch is probably not quite right - for instance if we insert a very large video it will end up much, much bigger than the slide - but it's probably a good jumping-off point to do nicer things.

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/map/handler/Map.FileInserter.js
+++ b/browser/src/map/handler/Map.FileInserter.js
@@ -109,11 +109,48 @@ L.Map.FileInserter = L.Handler.extend({
 		this._toInsertBackground = {};
 	},
 
-	_sendFile: function (name, file, type) {
+	_sendFile: async function (name, file, type) {
 		var socket = app.socket;
 		var map = this._map;
 		var sectionContainer = app.sectionContainer;
 		var url = this.getWopiUrl(map);
+
+		var size;
+
+		if (type === 'multimedia') {
+			const videoElement = document.createElement('video');
+			const objectURL = window.URL.createObjectURL(file);
+			videoElement.src = objectURL;
+
+			const videoLoadPromise = new Promise((resolve, reject) => {
+				videoElement.addEventListener("loadedmetadata", resolve);
+				videoElement.addEventListener("error", reject);
+			});
+
+			videoElement.load();
+
+			let videoLoaded = false;
+
+			try {
+				await videoLoadPromise;
+				videoLoaded = true;
+			} catch (_error) {
+				size = {
+					width: 0,
+					height: 0,
+				}; // 0, 0 will make core pick the minimum size - which was the behavior before we checked size like this
+			}
+
+			if (videoLoaded) {
+				size = {
+					width: videoElement.videoWidth,
+					height: videoElement.videoHeight,
+				};
+			}
+
+			videoElement.src = undefined;
+			window.URL.revokeObjectURL(objectURL);
+		}
 
 		if ('processCoolUrl' in window) {
 			url = window.processCoolUrl({ url: url, type: 'insertfile' });
@@ -140,8 +177,15 @@ L.Map.FileInserter = L.Handler.extend({
 					for (var i = 0; i < byteBuffer.length; i++) {
 						strBytes += String.fromCharCode(byteBuffer[i]);
 					}
-					window.postMobileMessage('insertfile name=' + aFile.name + ' type=' + type +
-										       ' data=' + window.btoa(strBytes));
+
+					if (type === 'multimedia') {
+						window.postMobileMessage('insertfile name=' + aFile.name + ' type=' + type +
+											       ' data=' + window.btoa(strBytes) +
+											       ' width=' + size.width + ' height=' + size.height);
+					} else {
+						window.postMobileMessage('insertfile name=' + aFile.name + ' type=' + type +
+											       ' data=' + window.btoa(strBytes));
+					}
 				};
 			})(file);
 			reader.onerror = function(e) {
@@ -165,6 +209,8 @@ L.Map.FileInserter = L.Handler.extend({
 						}
 						if (section && section.sectionProperties.picturePicker && type === 'graphic') {
 							socket.sendMessage('contentcontrolevent type=picture' + ' name=' + name);
+						} else if (type === 'multimedia') {
+							socket.sendMessage('insertfile name=' + name + ' type=' + type + ' width=' + size.width + ' height=' + size.height);
 						} else {
 							socket.sendMessage('insertfile name=' + name + ' type=' + type);
 						}
--- a/browser/src/map/handler/Map.FileInserter.js
+++ b/browser/src/map/handler/Map.FileInserter.js
@@ -146,6 +146,20 @@ L.Map.FileInserter = L.Handler.extend({
 					width: videoElement.videoWidth,
 					height: videoElement.videoHeight,
 				};
+
+				const maxSize = {
+					width: app.file.size.cX * map.getZoomScale(10),
+					height: app.file.size.cY * map.getZoomScale(10),
+				};
+
+				const shrinkToFitFactor = Math.min(
+					1,
+					maxSize.width / size.width,
+					maxSize.height / size.height
+				);
+
+				size.width *= shrinkToFitFactor;
+				size.height *= shrinkToFitFactor;
 			}
 
 			videoElement.src = undefined;
--- a/browser/src/canvas/sections/ShapeHandlesSection.ts
+++ b/browser/src/canvas/sections/ShapeHandlesSection.ts
@@ -442,22 +442,6 @@ class ShapeHandlesSection extends CanvasSectionObject {
 		source[0].src = decodeURIComponent(source[0].src);
 
 		this.addVideoSupportHandlers(videos);
-
-		function _fixSVGPos() {
-			var mat = this.sectionProperties.svg.getScreenCTM();
-			var boundingBox = this.context.canvas.getBoundingClientRect();
-			videoContainer.style.transform = 'matrix(' + [mat.a, mat.b, mat.c, mat.d, mat.e - boundingBox.x, mat.f - boundingBox.y].join(', ') + ')';
-		}
-		var fixSVGPos = _fixSVGPos.bind(this);
-
-		if (L.Browser.safari) {
-			fixSVGPos();
-			var observer = new MutationObserver(fixSVGPos);
-
-			observer.observe(this.context.canvas, {
-				attributes: true
-			});
-		}
 	}
 
 	removeTagFromHTML(data: string, startString: string, endString: string): string {
--- a/kit/ChildSession.cpp
+++ b/kit/ChildSession.cpp
@@ -1721,21 +1721,45 @@ bool ChildSession::paste(const char* buffer, int length, const StringVector& tok
 bool ChildSession::insertFile(const StringVector& tokens)
 {
     std::string name, type, data;
+    int multimedia_width = 0;
+    int multimedia_height = 0;
 
     if constexpr (!Util::isMobileApp())
     {
-        if (tokens.size() != 3 || !getTokenString(tokens[1], "name", name) ||
+        if (tokens.size() < 3 || !getTokenString(tokens[1], "name", name) ||
             !getTokenString(tokens[2], "type", type))
         {
+            sendTextFrameAndLogError("error: cmd=insertfile kind=syntax");
+            return false;
+        }
+
+        if (type == "multimedia") {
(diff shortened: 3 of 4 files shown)


Minion3665:
So, this isn't working properly on my iPad - it seems to think the slide size is much larger than it actually is... not sure why, though

Minion3665:
found that dividing by app.dpiScale fixes the issue on iOS - but can inconsistently cause videos to be inserted too-small on desktop... which is weird because that's not the piece I'd expect anything like data races/etc. in

Minion3665:
so, it looks like the browser code is sending off the sizes correctly - and they're being ... ignored, maybe? by core/something else inbetween

Minion3665:
> so, it looks like the browser code is sending off the sizes correctly - and they're being ... ignored, maybe? by core/something else inbetween

I've looked at why this is happening, and it seems that sometimes we're erroneously getting a 'pPickObject" (which I assume without proof is an empty object that we can fill in with a video - I'm thinking like the half of that one common slide type which has 2 images side by side). This particular object seems to be an empty, invisible object in the middle of the slide judging by where the video seems to be inserted... regardless, the following patch is a workaround:

```diff
minion@apx-vso-pico:~/Code/Collabora/CollaboraOnline/standard/core$ jj show --git
Commit ID: b92cc6c4aa7c1afa3a9331b56671c2e894a7e5bf
Change ID: svqwsplzktwyxopqsukxutlzwkytsnrs
Author   : Skyler Grey <skyler.grey@collabora.com> (2025-04-01 15:45:03)
Committer: Skyler Grey <skyler.grey@collabora.com> (2025-04-01 15:45:44)

    (no description set)

diff --git a/sd/source/ui/view/sdview4.cxx b/sd/source/ui/view/sdview4.cxx
index f46c876dac..634668d11a 100644
--- a/sd/source/ui/view/sdview4.cxx
+++ b/sd/source/ui/view/sdview4.cxx
@@ -330,7 +330,7 @@
 
     rtl::Reference<SdrMediaObj> pNewMediaObj;
     SdrPageView*    pPV = GetSdrPageView();
-    SdrObject*      pPickObj = GetEmptyPresentationObject( PresObjKind::Media );
+    SdrObject*      pPickObj = nullptr; // GetEmptyPresentationObject( PresObjKind::Media );
 
     if(pPV && dynamic_cast<const ::sd::slidesorter::view::SlideSorterView* >(this) )
     {
```

Minion3665:
After a bit more poking about I'm probably going to do this - since as having any empty media object be filled as you insert videos is not necessarily desirable behavior(?) -- will need to talk over the consequences of this with someone who knows a bit more around the GetEmptyPresentationObject bits though:

```diff
minion@apx-vso-pico:~/Code/Collabora/CollaboraOnline/standard/core$ jj show --git 
Commit ID: c765e553657fbaf59eeac7d73bc3b0e84f216fd3
Change ID: svqwsplzktwyxopqsukxutlzwkytsnrs
Author   : Skyler Grey <skyler.grey@collabora.com> (2025-04-01 15:45:03)
Committer: Skyler Grey <skyler.grey@collabora.com> (2025-04-01 17:06:40)

    (no description set)

diff --git a/sd/source/ui/inc/View.hxx b/sd/source/ui/inc/View.hxx
index 342bf7b625..91c7621f34 100644
--- a/sd/source/ui/inc/View.hxx
+++ b/sd/source/ui/inc/View.hxx
@@ -244,7 +244,7 @@
     void SetResizePropAllowed( bool bSet ) { m_bResizePropAllowed = bSet; }
     void SetResizeProtected( bool bSet ) { m_bResizeProtect = bSet; }
 
-    SdrObject* GetEmptyPresentationObject( PresObjKind eKind );
+    SdrObject* GetEmptyPresentationObject( PresObjKind eKind, bool bSelectedOnly = false );
     SdPage* GetPage();
     SdrObject* GetSelectedSingleObject(SdPage const * pPage);
     void SetAuthor(const OUString& rAuthor) { m_sAuthor = rAuthor; }
diff --git a/sd/source/ui/view/sdview4.cxx b/sd/source/ui/view/sdview4.cxx
index f46c876dac..be5da06ab0 100644
--- a/sd/source/ui/view/sdview4.cxx
+++ b/sd/source/ui/view/sdview4.cxx
@@ -330,7 +330,7 @@
 
     rtl::Reference<SdrMediaObj> pNewMediaObj;
     SdrPageView*    pPV = GetSdrPageView();
-    SdrObject*      pPickObj = GetEmptyPresentationObject( PresObjKind::Media );
+    SdrObject*      pPickObj = GetEmptyPresentationObject( PresObjKind::Media, true );
 
     if(pPV && dynamic_cast<const ::sd::slidesorter::view::SlideSorterView* >(this) )
     {
diff --git a/sd/source/ui/view/sdview5.cxx b/sd/source/ui/view/sdview5.cxx
index c3ac066bc0..bf48488494 100644
--- a/sd/source/ui/view/sdview5.cxx
+++ b/sd/source/ui/view/sdview5.cxx
@@ -76,7 +76,7 @@
     return pRet;
 }
 
-SdrObject* View::GetEmptyPresentationObject( PresObjKind eKind )
+SdrObject* View::GetEmptyPresentationObject( PresObjKind eKind, bool bSelectedOnly )
 {
     SdPage* pPage = GetPage();
     SdrObject* pEmptyObj = nullptr;
@@ -87,6 +87,11 @@
         if( pObj && pObj->IsEmptyPresObj() && implIsMultiPresObj( pPage->GetPresObjKind(pObj) ) )
             pEmptyObj = pObj;
 
+        if (bSelectedOnly)
+        {
+            return pEmptyObj; // No further chances to find an empty object, so no need to check if we found one
+        }
+
         // try to find empty pres obj of same type
         if( !pEmptyObj )
         {
```

Minion3665:
I realized the issue - I wasn't handling the case where the video is unable to load, which seems to be the case in cypress ... I'm not sure why this fails in cypress, but I'm willing to chalk it up to a difference in environment (perhaps cypress on the CI doesn't have the correct libraries to fully load video or somesuch(?)). Regardless, we can still fallback to default if the video does fail to load.

```diff
minion@apx-vso-pico:~/Code/Collabora/CollaboraOnline/standard/online$ jj diff --from 9c72203e636d84175c261abf5c04faf2127e80fd --to 1ae34cd8cfd75bffe02cd4d69fd6725cbbaef244 --git --no-pager
diff --git a/browser/src/map/handler/Map.FileInserter.js b/browser/src/map/handler/Map.FileInserter.js
index a843fe06d0..fc78c21eeb 100644
--- a/browser/src/map/handler/Map.FileInserter.js
+++ b/browser/src/map/handler/Map.FileInserter.js
@@ -122,31 +122,45 @@
 			const objectURL = window.URL.createObjectURL(file);
 			videoElement.src = objectURL;
 
-			const videoLoadPromise = new Promise((resolve) => {
+			const videoLoadPromise = new Promise((resolve, reject) => {
 				videoElement.addEventListener("loadedmetadata", resolve);
+				videoElement.addEventListener("error", reject);
 			});
 
 			videoElement.load();
-			await videoLoadPromise;
-
-			size = {
-				width: videoElement.videoWidth,
-				height: videoElement.videoHeight,
-			};
-
-			const maxSize = {
-				width: app.file.size.cX * map.getZoomScale(10),
-				height: app.file.size.cY * map.getZoomScale(10),
-			};
-
-			const shrinkToFitFactor = Math.min(
-				1,
-				maxSize.width / size.width,
-				maxSize.height / size.height
-			);
-
-			size.width *= shrinkToFitFactor;
-			size.height *= shrinkToFitFactor;
+
+			let videoLoaded = false;
+
+			try {
+				await videoLoadPromise;
+				videoLoaded = true;
+			} catch (_error) {
+				size = {
+					width: 0,
+					height: 0,
+				}; // 0, 0 will make core pick the minimum size - which was the behavior before we checked size like this
+			}
+
+			if (videoLoaded) {
+				size = {
+					width: videoElement.videoWidth,
+					height: videoElement.videoHeight,
+				};
+
+				const maxSize = {
+					width: app.file.size.cX * map.getZoomScale(10),
+					height: app.file.size.cY * map.getZoomScale(10),
+				};
+
+				const shrinkToFitFactor = Math.min(
+					1,
+					maxSize.width / size.width,
+					maxSize.height / size.height
+				);
+
+				size.width *= shrinkToFitFactor;
+				size.height *= shrinkToFitFactor;
+			}
 		}
 
 		if ('processCoolUrl' in window) {
```

image

Here's the diff of that, since as GitHub as usual makes it rather challenging to follow through the forcepushes. I've also included a [difftastic](https://difftastic.wilfred.me.uk/) diff screenshot since as part of it was a reindent

browser/src/map/handler/Map.FileInserter.js
```diff
@@ -140,8 +188,15 @@ L.Map.FileInserter = L.Handler.extend({
 					for (var i = 0; i < byteBuffer.length; i++) {
 						strBytes += String.fromCharCode(byteBuffer[i]);
 					}
-					window.postMobileMessage('insertfile name=' + aFile.name + ' type=' + type +
-										       ' data=' + window.btoa(strBytes));
+
+					if (type === 'multimedia') {
+						window.postMobileMessage('insertfile name=' + aFile.name + ' type=' + type +
+											       ' data=' + window.btoa(strBytes) +
```
eszkadev, line 194: hmm, might be a big payload. Are there any limitations for a size?
isn't there any option for sending HTTP POST request to the "server" in the mobile ?

browser/src/map/handler/Map.FileInserter.js
```diff
@@ -109,12 +109,60 @@ L.Map.FileInserter = L.Handler.extend({
...
 		var socket = app.socket;
 		var map = this._map;
 		var sectionContainer = app.sectionContainer;
 		var url = this.getWopiUrl(map);
 
+		var size;
+
+		if (type === 'multimedia') {
+			const videoElement = document.createElement('video');
+			const objectURL = window.URL.createObjectURL(file);
```
eszkadev, line 122: we need to release memory: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#memory_management

also there is a note about function beeing deprecated soon: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#using_object_urls_for_media_streams

Minion3665: > we need to release memory

good catch, I'll do that and update the PR...

> also there is a note about function beeing deprecated soon

unluckily, as is often the case with the web, the replacements for the soon-deprecated features are as-of-yet unsuitable: https://caniuse.com/?search=srcObject

luckily, as is often the case with the web, I have doubts about any modern browser removing support for createObjectURL any time soon

Minion3665: (on the caniuse page there, the partial-support options aren't enough - since as we don't have a `MediaStream`. Therefore, we would be supporting effectively only Safari with this feature)

browser/src/map/handler/Map.FileInserter.js
eszkadev: I understand. My concern is: previously we had image now video/media which might be significantly bigger in average case. HTTP transport is used for clipboard which also could have bigger content. I see it as more reliable in such cases as there is always some streaming support built-in

browser/src/map/handler/Map.FileInserter.js
eszkadev: thanks

eszkadev (approved):
it works, but the inserted video is centered at (0, 0) position, while should be moved to the center of a slide I think:

insert

Minion3665:
> it works, but the inserted video is centered at (0, 0) position, while should be moved to the center of a slide I think:

yeah, indeed

I think we should probably put some more effort into media positioning across the board - stuff should be centered at 0,0 unless we drag/drop it on in which case it should end up where our cursor is... but that's a separate task :)

eszkadev:
> > it works, but the inserted video is centered at (0, 0) position, while should be moved to the center of a slide I think:
> 
> yeah, indeed
> 
> I think we should probably put some more effort into media positioning across the board - stuff should be centered at 0,0 unless we drag/drop it on in which case it should end up where our cursor is... but that's a separate task :)

I think it should be easy in core. We have something like "DirectInsert" or similar name for some insert commands. Then in core it by default puts it in the center and doesn't require user to "paint" the shape or point the place.

PR #11367, opened 2025-03-18
make run mode fd leak

caolanm:
To reproduce: open a document through a debug link and close it to establish a baseline.

On each subsequent open and close:
$ ls /proc/`pidof coolwsd`/fd|wc
increments by 2

in ClientRequestDispatcher::handleIncomingMessage closeConnection is the default 'false' for this request and nothing otherwise explictly closes the socket on that side.

Anything similar is not seen in release mode with, e.g. a nextcloud instance.

Change-Id: I927138ba77f54ff6ae86a7c10f003abd89549653

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


mmeeks (commented):
Looks sensible enough to me - thanks.

Ashod (approved):
Thanks, @caolanm. It looks like there is an ownership issue. Either the session or the socket-poll should do this cleanup, automatically.

PR #11370, opened 2025-03-19
Private/gokay/calc invalidation

gokaysatir:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -1554,11 +1554,10 @@ class TileManager {
 	}
 
 	private static coordsToTileBounds(coords: TileCoordData): number[] {
-		var zoomFactor = app.map.zoomToFactor(coords.z);
-		const x = (coords.x * app.pixelsToTwips) / zoomFactor;
-		const y = (coords.y * app.pixelsToTwips) / zoomFactor;
-		const width = app.tile.size.x / zoomFactor;
-		const height = app.tile.size.y / zoomFactor;
+		const x = coords.x * app.pixelsToTwips;
+		const y = coords.y * app.pixelsToTwips;
+		const width = app.tile.size.x;
+		const height = app.tile.size.y;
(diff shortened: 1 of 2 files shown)


browser/src/app/TilesMiddleware.ts
```diff
@@ -1554,11 +1554,10 @@ class TileManager {
 	}
 
 	private static coordsToTileBounds(coords: TileCoordData): number[] {
-		var zoomFactor = app.map.zoomToFactor(coords.z);
-		const x = (coords.x * app.pixelsToTwips) / zoomFactor;
-		const y = (coords.y * app.pixelsToTwips) / zoomFactor;
-		const width = app.tile.size.x / zoomFactor;
-		const height = app.tile.size.y / zoomFactor;
+		const x = coords.x * app.pixelsToTwips;
```
vmiklos, line 1557: Looks OK, though given that you return just a list of numbers here, would it make sense to make it more explicit that you take pixels here and you return twips? Perhaps rename the function to `pixelCoordsToTwipTileBounds`?

At least to me it's not too trivial that app.tile.size is in twips, but TileCoorData is in pixels.

gokaysatir: That's right, it confused also me when i first read the code :) Renamed.

vmiklos:
One more thought:  I wonder if it's possible to write a test to make sure that the 3 intersect methods of that SimpleRectangle class stay correct?

gokaysatir:
> One more thought: I wonder if it's possible to write a test to make sure that the 3 intersect methods of that SimpleRectangle class stay correct? Fine to do it in a separate PR if you want.

Of course but there is LOUtil.test.ts. And now we are using that functionality. Let me note this and at some point i think we need tests for geometry.ts functions.

gokaysatir:
All tests are passing locally.

PR #11371, opened 2025-03-19
Fix double-resize on CanvasSectionContainer

Cwiiis:
### Summary

When a resize is handled by CanvasTileLayer, two resizes may end up being triggered on CanvasSectionContainer. In this situation, the first happens prematurely and causes drawing to happen before the map is updated, while the second is ignored because the canvas size change was already handled.

### Checklist


Cwiiis:
Ok, this very much breaks things, perhaps the ordering is important :)

gokaysatir (approved):
Makes sense and seems not to break anything, checked quickly. Thanks.

PR #11374, opened 2025-03-19
mobile: do not use notebookbar

eszkadev:
on mobile there is no notebookbar but we still tried to use it in some cases - detected in cypress


eszkadev:
we need https://gerrit.libreoffice.org/c/core/+/183132 to fix mobile-wizard fails

PR #11376, opened 2025-03-19
fd leak with --with-infobar-url

caolanm:
If I munge fetchUpdateCheck to make this happen once a second instead of the default every 10 hours I can see this rapidly accumulates hundreds of open fds: ls /proc/`pidof coolwsd`/fd|wc

Unlikely to be very significant in a normal setup.

Change-Id: Ib5041835d2bbd10a52aa98997cbc806a25f06e3a

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/wsd/COOLWSD.cpp
+++ b/wsd/COOLWSD.cpp
@@ -3536,6 +3536,8 @@ void COOLWSD::processFetchUpdate(SocketPoll& poll)
         request.add("Accept", "application/json");
 
         FetchHttpSession->setFinishedHandler([](const std::shared_ptr<http::Session>& httpSession) {
+            httpSession->asyncShutdown();
+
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Looks very sensible =) thanks.

mmeeks:
Should go into 24.04 too I think - I expect things like 'remote_config' which is polled every second and can be rather large is something we should look at in a similar vein (?) =)

caolanm:
I looked at RemoteJSONPoll before looking at this, and even if hacked to retry every second, both with a working remote_config_url and a non-existing one (or with injected invalid JSON), then coolwsd RSS is stable. RemoteJSONPoll uses http::Session::syncRequest so this particular asyncRequest ambiguity doesn't apply there.

PR #11377, opened 2025-03-19
Convert L.Log and L.DebugManager to typescript classes

dennisfrancis:
* Resolves: # 
* Target version: master 

### Summary
Convert L.Log and L.DebugManager to typescript classes

### Checklist


mmeeks:
Sorry if I created conflicts with you Dennis =) can you fix ?

browser/src/app/Log.ts
```diff
@@ -0,0 +1,109 @@
...
+			if (a.time < b.time) {
+				return -1;
+			}
+			if (a.time > b.time) {
+				return 1;
+			}
+			return 0;
+		});
+		let data: string = '';
+		for (var i = 0; i < this._logs.length; i++) {
```
hfiguiere, line 59:
```suggestion
		for (let i = 0; i < this._logs.length; i++) {
```

Let's stop using `var`

browser/src/app/Log.ts
```diff
@@ -0,0 +1,109 @@
...
+	private _logs: LogMsg[];
+	private startTime: number;
+
+	constructor() {
+		this._logs = [];
+		this.startTime = null;
+	}
+
+	public log(msg: string, direction: Direction): void {
+		var time = Date.now();
```
hfiguiere, line 31: here and below, let's use `let` instead of `var`.

Also maybe new line after the `if ()` or `while ()`. This would make things more readable.

browser/src/app/Log.ts
dennisfrancis: done

browser/src/app/Log.ts
hfiguiere: :-(

eszkadev (approved):
let's merge to avoid conflicts, thanks

Issue #11287, opened 2025-03-06
Issues with HTTP Status Moved (Permanently) HTTP 301  Nextcloud, Collabora & Nginx Reverse Proxies


ksaadDE:
### Describe the Bug
When having everything properly configured (including nextcloud), the collabora system tries to grab the preset from an url, when opening a document in nextcloud.

If the preset pull response status is 301 (HTTP STATUS 301 / PERMANENTLY MOVED) e.g. due to one or multiple nginx reverse proxies, it will automatically break and report failure.

### Steps to Reproduce
1. Install nextcloud latest & collabora/code latest via docker (configure everything, including the nginx reverse proxies etc)
3. Go in the nextcloud and load a document (.odt)
4. `docker logs collabora` (shows the `Failed to get settings json from [http://http_nextcloud.domain/index.php/apps/richdocuments/wopi/settings?type=systemconfig&access_token=xxxxx&fileId=-1] with status[Moved Permanently]|wsd/DocumentBroker.cpp:1820`)
5. Install older tag e.g. `collabora/code:24.04.10.1.1` (this one shall work)

### Expected Behavior
Installation just loads and allows editing of the documents via nextcloud using collabora/code container behind nginx reverse proxies

### Actual Behavior
The loading of the preset fails, due to Permanently Moved, as only HTTP STATUS OK is being checked for, thus the whole execution is aborted. See suggested patch below.

### Additional Context
please read https://github.com/CollaboraOnline/online/commit/3ed4b490fc84868c6dab7a2aa62eac6b3182286#r153381554

Not sure if this belongs to online but it has the bug in the same code base so

ksaadDE:
@Rash419

ksaadDE:
Short summary of the troublesome line as mentioned in the commit & commentary: https://github.com/CollaboraOnline/online/commit/3ed4b490fc84868c6dab7a2aa62eac6b3182286#r153381554
```cpp
if (presetHttpResponse->statusLine().statusCode() != http::StatusCode::OK) 
{
// ...
}
```

The required patch could be:
```cpp
if (presetHttpResponse->statusLine().statusCode() != http::StatusCode::OK && 
presetHttpResponse->statusLine().statusCode() != http::StatusCode::MovedPermanently) 
{
//...
}
```
Surely, there is a more smart and dynamic alternative (C++ is not my main business) to exclude various HTTP codes that shall not negatively affect the continuation of the processing!

[...]

PR #11380, opened 2025-03-20
wsd: preset: handle redirection response

Rash419:
- Patch handles the following scenario:
- sometimes due to misconfiguration, even in SSL environment  WOPIHost might send preset urls with `http` protocol and when reverse proxy is configured with `force SSL` rule. Proxy redirects requests with `http` protocol to `https` protocol.

Change-Id: Ia8ea40636feab41248ef995a1dd440f171b293e5

* Resolves: #11287 
* Target version: master 

### Testing
I have tested this with https://github.com/Rash419/CollaboraTestSetups/tree/main/collabora-nextcloud-nginx. Additionally I have added a rule where nginx redirects all the http request to https https://github.com/Rash419/CollaboraTestSetups/blob/f6fcb1aa91fe9714de712b7ec17d0b7537e435eb/collabora-nextcloud-nginx/nginx/nginx.conf#L7

### DIFF
--- a/wsd/DocumentBroker.cpp
+++ b/wsd/DocumentBroker.cpp
@@ -1478,7 +1478,7 @@ void PresetsInstallTask::asyncInstall(const std::string& uri, const std::string&
     installPresetStarted(id);
 
     DocumentBroker::asyncInstallPreset(_poll, _configId, uri, stamp, fileName, id,
-                                       presetInstallFinished, session);
+                                       presetInstallFinished, session, HTTP_REDIRECTION_LIMIT);
 }
 
 void PresetsInstallTask::installPresetStarted(const std::string& id)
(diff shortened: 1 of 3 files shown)


caolanm (approved):
This makes sense to me. I had wondered if we needed to handle redirects, so it turns out we do.

Rash419:
@caolanm It should pass the CI now. The problem was passing that I was passing `_asyncPresetTask` as reference to lambda. I have other 3 question just to be sure:
1. poll is passed as reference to lamba here in https://github.com/CollaboraOnline/online/pull/11380/files#diff-1346662577e72b9ac23f0977699d096e506badf1ce7d2cf56a84ab8769d9be66R1768. I am assuming it is webserv poll which should exist during the lambda's life time, right ?
2. clang warns about recursive call here https://github.com/CollaboraOnline/online/pull/11380/files#diff-1346662577e72b9ac23f0977699d096e506badf1ce7d2cf56a84ab8769d9be66R1645 as I am calling itself when the request get redirected. Is there any better way to do it ? 
3. Redirected URL we get with redirection response do we need sanitize it anyway ?

caolanm:
We've reworked this area to make the lifecycle of polls themselves more robust/clear. So that's conflicted with the pr unfortunately, so this needs to be rebased, but it should at least make it safer to pass the poll around as it is presented to asyncInstallPresets as a shared_ptr. The "websrv_poll" is long lived, but user presets get installed using the specific document poll which just lives as long as the document and there are corner cases where that may have terminated when returning from asyncConnect from the DNS thread

ksaadDE:
I read today's code as well, looks like it should work. Yet I do not see where `HTTP_REDIRECTION_LIMIT` is defined, and what the limit is. 

Thank you very much for that patch, it looks great - I am going back to backseating and watching you guys rock! 🚀

caolanm:
be good to get this rebased I think

PR #11381, opened 2025-03-20
mobile: do not send invalid jsdialog message

eszkadev:
wo no longer use UI testing module in the core so we don't need to refresh whole sidebar, welding sends messages when needed


eszkadev:
doesn't work: writer/mobile_wizard_state_spec.js fails

mmeeks (approved):
Looks rather sensible - hopefully fixes the navigator issues ? =)

Issue #11297, opened 2025-03-07
SaveAsMode=group no longer working


juliusknorr:
### Describe the Bug

This is a regression that our cypress tests in richdocuments catched

We set the ui defaults SaveAsMode to group as described in https://sdk.collaboraonline.com/docs/theming.html#user-interface-modifications

### Steps to Reproduce

1. Set SaveAsMode=group 

### Expected Behavior

Save as button has a submenu 

<img width="336" alt="Image" src="https://github.com/user-attachments/assets/41afa3c5-d79f-4790-882f-5e3cce055b83" />

### Actual Behavior
Save as button is not grouped (single button
<img width="227" alt="Image" src="https://github.com/user-attachments/assets/02430c80-8d9f-4bcb-9285-e23f3ab0cfa8" />

### Additional Context

I can see the value be set in the js context of online:

<img width="1505" alt="Image" src="https://github.com/user-attachments/assets/90858834-6106-4ef4-a845-e0f32d7faa7c" />

juliusknorr:
Looking at CI jobs this tarted somewhere after February 6th  https://github.com/nextcloud/richdocuments/actions/runs/13170096670 

Due to other ci issues it is hard to narrow it down, did not bisect yet

jazevedo-coll:
I also could not reproduce.

pedropintosilva:
I could reproduce it  with NC 31, I think. And definitely I saw a screenshare from Julius from a daily rebuild instance and it was reproducible:

 OC.getCapabilities().richdocuments.version
12:01:12.362 "9.0.0-dev.0"
12:01:19.742 oc_config.version
12:01:19.761 "32.0.0.0" 

And Collabora Online daily image (following master)

jazevedo-coll:
This regression was introduced somewhere in this merge request: 

https://github.com/nextcloud/richdocuments/pull/4373/commits

Had to skip two of the commits, but all 4 commits are from the same merge reques

|Status| Commit|
|----------|----------|
| Not OK  | 59a2192e82e3a7d92ee64cef38f01b6c04d907d1  |
| Skipped | 0e79c206f2dba487f77002627f24fbf479e586ea |
| Skipped | 82bb7a890b697f064518bf7d048225beed25975a |
| OK         |dbf9d3aea27b2257a27e7cbc2ad379d7e2ea22da |

jazevedo-coll:
Tested on; 

[...]

PR #11382, opened 2025-03-20
browsersettting: fallback to localStorage/uiDefault for undefined setting

codewithvk:
browsersettting: fallback to localStorage/uiDefault for undefined setting

When a browser setting is not defined, use localStorage/uiDefault as a fallback. This fixes the regression related to the saveAsMode issue. bug: https://github.com/CollaboraOnline/online/issues/11297

Change-Id: Ic035f3a5af95f903aff4934e998dd0a38675ccdc

* Resolves: #11297 
* Target version: master

### DIFF
--- a/browser/js/global.js
+++ b/browser/js/global.js
@@ -722,8 +722,11 @@ function getInitializerClass() {
 				let val = defaultValue;
 				if (Object.prototype.hasOwnProperty.call(global.prefs._userBrowserSetting, key))
 					val = global.prefs._userBrowserSetting[key];
-				global.prefs._localStorageCache[key] = val;
-				return val;
+
+				if(val !== undefined) {
+					global.prefs._localStorageCache[key] = val;
+					return val;
+				}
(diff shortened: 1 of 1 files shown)


browser/js/global.js
```diff
@@ -720,8 +720,11 @@ function getInitializerClass() {
 				let val = defaultValue;
 				if (Object.prototype.hasOwnProperty.call(global.prefs._userBrowserSetting, key))
 					val = global.prefs._userBrowserSetting[key];
-				global.prefs._localStorageCache[key] = val;
-				return val;
+
+				if(val != undefined) {
```
Rash419, line 724: !== ?

Rash419 (changes requested):

browser/js/global.js
codewithvk: done :)

Rash419 (approved):
Thanks

PR #11384, opened 2025-03-20
Introduce DocumentBase and ViewLayoutBase

gokaysatir:
Moves viewedRectangle and sendClientVisibleArea into ViewLayoutBase.

Others are mostly renames for the documentTopLeft and the like.

### DIFF
--- a/browser/src/app/ViewLayout.ts
+++ b/browser/src/app/ViewLayout.ts
@@ -1,0 +1,87 @@
+/* -*- js-indent-level: 8 -*- */
+/*
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+class ViewLayoutBase {
+	viewedRectangle: cool.SimpleRectangle;
+	lastViewedRectangle: cool.SimpleRectangle;
+	private clientVisibleAreaCommand: string = '';
+
+	constructor() {
+		this.viewedRectangle = new cool.SimpleRectangle(0, 0, 0, 0);
+		this.lastViewedRectangle = new cool.SimpleRectangle(0, 0, 0, 0);
+	}
+
+	public resetClientVisibleArea(): void {
+		this.lastViewedRectangle = new cool.SimpleRectangle(0, 0, 0, 0);
+	}
+
+	public sendClientVisibleArea(forceUpdate: boolean = false): void {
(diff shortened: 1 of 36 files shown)


eszkadev:
@gokaysatir what this is about? is it still usefull to keep? or some leftover we can close?

gokaysatir:
> @gokaysatir what this is about? is it still usefull to keep? or some leftover we can close?

This is for multi-page view of Writer files. I will continue to work on this. Couldn't have time for some time.

browser/src/app/ViewLayout.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Copyright the Collabora Online contributors.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+namespace cool {
```
eszkadev: can we NOT use namespaces?
Names will be shorter, we don't have conflicting class names.
Later we could use ES modules and simply "import from" what will be enough.
It doesn't work great at the same time with ES modules I heard.

browser/src/canvas/CanvasSectionContainer.ts
```diff
@@ -1011,6 +957,37 @@ class CanvasSectionContainer {
 		}
 	}
 
+	public onNewDocumentTopLeft() {
```
eszkadev, line 961: what it does? not obvious from the name, can we have comment above?

eszkadev:
It seems you need to fix cypress desktop tests, fail on scrolling assertions.
Possibly the calculations are changed and thumb position is at different place.

eszkadev:
It doesn't detect end of the view (bottom) in calc, so when we move cursor with arrow down it doesn't scroll:
[RecApp-2025-08-04-08:25:36.webm](https://github.com/user-attachments/assets/0f6f5dc4-2bde-4d06-bbe8-490ed67def03)

browser/src/app/ViewLayout.ts
gokaysatir: I see, i am also indecisive, i also don't want to use namespaces but they are safer. Anyway, let's go without them for now. If we need, we can add namespaces any time. It's just a find & replace anyway.

browser/src/canvas/CanvasSectionContainer.ts
gokaysatir: Now we need to separate the view management from CanvasSectionContainer. Because multi-page view and possible future others need to manage the coordinates themselves.

Views need to set the document top left. In that case, CanvasSectionContainer needs to know when the document is scrolled. I think name is self-explaining but i will also add a comment there.

PR #11387, opened 2025-03-20
Calc: Return early from _syncTileContainerSize if size doesn't need t…

gokaysatir:
…o change.

Previously we were using _syncTileContainerSize after resizing event. Now we need to check it after restricDocumentSize function. If there is no need to resize the canvas and map elements, return early.

Change-Id: Ic56f4358327680fc52c9a788e48fe1750deeca9c

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -401,6 +401,11 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		const mapElement = document.getElementById('map'); // map's size = tiles section's size.
 		const oldMapSize = [mapElement.clientWidth, mapElement.clientHeight];
+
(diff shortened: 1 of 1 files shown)


gokaysatir:
This is a follow up for: https://github.com/CollaboraOnline/online/pull/11358

@Cwiiis can you have a look when you have time? I checked and confirm that our previous fix still works. This is for not resizing the canvas and map elements with the same dimensions. Thank you for your previous review.

gokaysatir:
By the way, in case it may help. For reproducing the previous issue which was fixed in the mentioned PR:
* Open a Calc file.
* Zoom out.
* Create a new sheet.
* The size of the canvas and map elements should cover the available space.

Cwiiis (approved):
Assuming tests pass, this looks pretty benign to me :)

PR #11388, opened 2025-03-20
notebookbar: correctly setup toolbox after update

eszkadev:
Command for dropdown in menubutton inserted into toolbox require correct parent id to work, let's reuse toolbox implementation as everywhere else in the notebookbar.

It fixes not working currency dropdown in the notebookbar when tried second time. Regression from commit 7e5075fea733d68c8167ac33e5ce34d89b50a6a9 jsdialog: toolbar set_item_active method

Requires update of  toolbox in JSON (without we replace content with only currency button, not all 3 as before dropdown usage). Initial toolbox is just a JSON from online, not core.


eszkadev:
will add test in followup

gokaysatir (approved):
Thanks for the fix :)

PR #11389, opened 2025-03-20
Fix TileManager.garbageCollect

Cwiiis:
* Target version: master 

### Summary
This appears to have been comprehensively broken for a while. This commit fixes these issues:
- Tile.current should reflect whether the tile is in the visible area
- Various tile update functions were unnecessarily redrawing the screen
- Tile.hasPendingDelta and Tile.hasPendingKeyframe were not initialised
- Tile.lastRendered was not stored or updated efficiently
- TileManager.garbageCollect was not sorting tiles correctly
- TileManager.garbageCollect was not respecting highNumCanvases correctly
- Tile preload map was not visualising current or canvas-GC'd tiles
- TileManager.updateLayoutView was typo'd as 'udpateLayoutView'
    
Fixing garbage collection has the knock-on effect of markedly enhancing performance on 4k screens and possibly fixing some OOM situations.

### Checklist

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -150,7 +150,7 @@ class Tile {
 	viewId: number = 0; // canonical view id
 	wireId: number = 0; // monotonic timestamp for optimizing fetch
 	invalidFrom: number = 0; // a wireId - for avoiding races on invalidation
-	lastRendered: Date = new Date();
+	lastRendered: number = performance.timeOrigin;
 	private lastRequestTime: Date = undefined; // when did we last do a tilecombine request.
 	hasPendingDelta: number = 0;
 	hasPendingKeyframe: number = 0;
@@ -366,14 +366,10 @@ class TileManager {
 		   highDeltaMemory = 1024*1024; lowDeltaMemory = 1024*128;
 		   highTileCount = 100; lowTileCount = 50; */
 
-		var keys: Array<string> = [];
-		for (const key in this.tiles) // no .keys() method.
-			keys.push(key);
-
 		// FIXME: should we sort by wireId - which is monotonic server ~time
 		// sort by oldest
-		keys.sort(function (a: any, b: any) {
-			return b.lastRendered - a.lastRendered;
+		var keys = Object.keys(this.tiles).sort((a: any, b: any) => {
+			return this.tiles[a].lastRendered - this.tiles[b].lastRendered;
 		});
 
 		var canvasKeys = [];
@@ -448,14 +444,13 @@ class TileManager {
 		this.maybeGarbageCollect();
 
 		// important this is after the garbagecollect
-		if (!tile.canvas) this.ensureCanvas(tile, null, false);
-
+		if (!tile.canvas) this.ensureCanvas(tile, false, false);
 		if (tile.ctx) return tile.ctx;
 
 		// Not a good result - we ran out of canvas memory
 		this.garbageCollect();
 
-		if (!tile.canvas) this.ensureCanvas(tile, null, false);
+		if (!tile.canvas) this.ensureCanvas(tile, false, false);
 		if (tile.ctx) return tile.ctx;
 
 		// Free non-current canvas' and start again.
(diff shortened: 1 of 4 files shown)


browser/src/app/TilesMiddleware.ts
```diff
@@ -1373,13 +1376,21 @@ class TileManager {
...
+	private static getMissingTiles(
+		pixelBounds: any,
+		zoom: number,
+		updateCurrent: boolean,
+	) {
 		var tileRanges = this.pxBoundsToTileRanges(pixelBounds);
 		var queue = [];
 
+		if (updateCurrent) {
+			for (var key in this.tiles) this.tiles[key].current = false;
```
mmeeks, line 1388: I'm suspicious of this. We can have multiple tile ranges that are current in split pane cases I would expect (?) =) So we need dis-contiguous regions of live / current tiles all rendered on the screen.

Otherwise this looks overall great - we need Gokay's review of his bits =)

Cwiiis: Yes, I suspect this is broken too - I wonder if we can just get rid of current and calculate it as necessary instead - that would really reduce complexity a lot... I suppose it's only really needed for garbage collection and that doesn't happen so frequently that we need to cache this value I don't think(?)

mmeeks: We could try calculating it on-demand; but keeping that in-sync is an issue - and we shouldn't need to be perfect. One thing we could do that is simpler is just sort by the last-rendered timestamp (I guess) and free the oldest - presumably dumping a 'now' into a member is fast and easy as we whack the canvas on screen. Its not clear to me that we can safely assume we have all the canvases we need in-memory to render to the screen anyway.

Cwiiis: Turns out it isn't broken - `getMissingTiles` is called with a bounds list, not a rect - so split panes work fine with this patch :) I think it would be nice to get rid of `current` and I think it'd be a relatively small change on top of this, but perhaps one thing at a time. I'm going to start splitting this up - test results also looking encouragingly positive.

Cwiiis:
@gokaysatir if you get a spare moment, it'd be great if you could give this a quick test - it seems to work well for me, but I may well be snow-blind...

browser/src/app/TilesMiddleware.ts
mmeeks: If you're happy lets get it in =)

mmeeks:
If it fixes the leak regression - great; we need that in 24.04 soon I think; we can get rid of '.current' of course - quite possibly we should have a nice method that gets the current visible tile-bounds anyway - seems a useful thing to have around :-)

browser/src/app/TilesMiddleware.ts
Cwiiis: I wouldn't be surprised if there are edge-cases where this does the wrong thing - I haven't hit one yet (been messing about in writer/calc/impress, and with split panes), but I definitely don't know how to exercise this code as well as others :) - All the more reason to get rid of Tile.current though, which I'm currently (hah) looking at. It's a bigger change than I thought, so definitely subsequent work, but it should significantly simplify what's happening here.

browser/src/app/TilesMiddleware.ts
```diff
@@ -2373,14 +2380,18 @@ class TileManager {
 
 	// Ensure we have a renderable canvas for a given tile
 	// Use this immediately before drawing a tile, pass in the time.
-	public static ensureCanvas(tile: Tile, now: any, forPrefetch: any) {
+	public static ensureCanvas(
+		tile: Tile,
+		forPrefetch: any,
```
gokaysatir, line 2385: I think this is a boolean with default "false".

gokaysatir (approved):
Looks good, checked the code and tested roughly. Thanks :)

There are 2 small things but one is related to an extra call to a function.

mmeeks:
Lets get it in & iterate to polish =) thanks Chris. @Cwiiis can you back-port to 24.04 and submit too ? =)

PR #11392, opened 2025-03-21
Use tile's own scale when getting bounds of a tile.

gokaysatir:
Also, when a tile is comes, compare its scale with the document's scale before setting the view dirty.

Issue: On Calc:
* Change background color of a cell.
* Zoom out.
* See that the background color of the cell is not updated.
* Because tile was not invalidated due to different zoom level and wrong calculation of bounds.

Change-Id: I5d9317b8a94b08fdc468016a0c2e4c20c34d2303

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/app/TilesMiddleware.ts
+++ b/browser/src/app/TilesMiddleware.ts
@@ -47,7 +47,7 @@ class TileCoordData {
 		this.part = part !== null ? part : app.map._docLayer._selectedPart;
 		this.mode = mode !== undefined ? mode : 0;
 
-		this.scale = Math.round(Math.pow(1.2, this.z - 10) * 1000) / 1000;
+		this.scale = Math.pow(1.2, this.z - 10);
 	}
 
 	getPos() {
@@ -889,10 +889,14 @@ class TileManager {
 
 		// Don't paint the tile, only dirty the sectionsContainer if it is in the visible area.
 		// _emitSlurpedTileEvents() will repaint canvas (if it is dirty).
+		const tileVisible = app.isRectangleVisibleInTheDisplayedArea(
+			this.pixelCoordsToTwipTileBounds(coords),
+		);
+
+		// Also check the scale because incoming tile may not be in the same zoom level.
 		if (
-			app.isRectangleVisibleInTheDisplayedArea(
-				this.pixelCoordsToTwipTileBounds(coords),
-			)
+			tileVisible &&
+			Math.round(coords.scale * 1000) === Math.round(app.getScale() * 1000)
 		)
 			app.sectionContainer.setDirty(coords);
 	}
@@ -1555,10 +1559,12 @@ class TileManager {
 	}
 
 	private static pixelCoordsToTwipTileBounds(coords: TileCoordData): number[] {
-		const x = coords.x * app.pixelsToTwips;
-		const y = coords.y * app.pixelsToTwips;
-		const width = app.tile.size.x;
-		const height = app.tile.size.y;
+		// We need to calculate pixelsToTwips for the scale of this tile. 15 is the ratio between pixels and twips when the scale is 1.
+		const pixelsToTwipsForTile = 15 / coords.scale;
+		const x = coords.x * pixelsToTwipsForTile;
+		const y = coords.y * pixelsToTwipsForTile;
+		const width = app.tile.size.pX * pixelsToTwipsForTile;
+		const height = app.tile.size.pY * pixelsToTwipsForTile;
 
 		return [x, y, width, height];
 	}
@@ -1571,6 +1577,8 @@ class TileManager {
 		textMsg: string,
 	) {
 		let needsNewTiles = false;
+		const calc = app.map._docLayer.isCalc();
+		const scale = app.getScale();
 
 		for (const key in this.tiles) {
 			const coords: TileCoordData = this.tiles[key].coords;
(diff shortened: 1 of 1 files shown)


browser/src/app/TilesMiddleware.ts
```diff
@@ -1555,10 +1556,11 @@ class TileManager {
 	}
 
 	private static pixelCoordsToTwipTileBounds(coords: TileCoordData): number[] {
-		const x = coords.x * app.pixelsToTwips;
-		const y = coords.y * app.pixelsToTwips;
-		const width = app.tile.size.x;
-		const height = app.tile.size.y;
+		const pixelsToTwipsForTile = 15 / coords.scale;
```
Cwiiis, line 1559: This is too arcane to leave here with no comment. Ideally it'd be a utility function and involve `app.pixelsToTwips`.

browser/src/app/TilesMiddleware.ts
```diff
@@ -889,11 +889,12 @@ class TileManager {
...
-			app.isRectangleVisibleInTheDisplayedArea(
-				this.pixelCoordsToTwipTileBounds(coords),
-			)
-		)
+		const tileVisible = app.isRectangleVisibleInTheDisplayedArea(
+			this.pixelCoordsToTwipTileBounds(coords),
+		);
+
+		// Also check the scale because incoming tile may not be in the same zoom level.
+		if (tileVisible && coords.scale === app.getScale())
```
Cwiiis, line 897: I'm not keen on the floating point comparison here (I realise it ought to work given the rest of the changes, but we're relying on external, undocumented properties never changing) - I'd rather we didn't change `app.getScale` (which seems prone to break other things/cause interesting to track errors), multiplied the scale out here and compared the integer values.

browser/src/app/TilesMiddleware.ts
```diff
@@ -47,7 +47,7 @@ class TileCoordData {
 		this.part = part !== null ? part : app.map._docLayer._selectedPart;
 		this.mode = mode !== undefined ? mode : 0;
 
-		this.scale = Math.round(Math.pow(1.2, this.z - 10) * 1000) / 1000;
+		this.scale = Math.round(Math.pow(1.2, this.z - 10) * 10000) / 10000;
```
Cwiiis, line 50: I'm dubious as to why this was here in the first place. Personally, I'd rather the opposite change - remove the squashing of the decimals here and add a comparison utility function when we only need to know that something approximately equals this. This feels like it'd end up causing all sorts of oddities with regards to accuracy that would be a pain to track down if you weren't aware this happened.

browser/src/app/TilesMiddleware.ts
```diff
@@ -1579,7 +1583,8 @@ class TileManager {
 			if (
 				coords.part === part &&
 				coords.mode === mode &&
-				invalidatedRectangle.intersectsRectangle(tileRectangle)
+				(invalidatedRectangle.intersectsRectangle(tileRectangle) ||
+					(calc && coords.scale !== scale)) // In calc, we invalidate all tiles with different zoom levels.
```
Cwiiis, line 1587: I'm confused as to why this problem would be specific to calc?

browser/src/docstatefunctions.js
```diff
@@ -79,7 +79,7 @@ app.isReadOnly = function () {
 };
 
 app.getScale = function () {
-	return (app.tile.size.pX / app.tile.size.x) * 15;
+	return Math.round((app.tile.size.pX / app.tile.size.x) * 15 * 10000) / 10000;
```
Cwiiis, line 82: Presumably there are other users of `app.getScale` and this removes a bunch of precision. I don't think this is a good idea.

Cwiiis (changes requested):
I think what this is doing is fine, but I have reservations about the method. I think it would be a better idea to add an `approximatelyEquals` function so that you can compare two float values at a given precision without having to store them at that lower precision.

browser/src/app/TilesMiddleware.ts
gokaysatir: We cannot use app.pixelsToTwips there. Added a comment.

browser/src/app/TilesMiddleware.ts
gokaysatir: That's right, somehow i picked the other way but this is better. Improved.

browser/src/app/TilesMiddleware.ts
gokaysatir: Because we adjust the twips position for Calc. We cannot use the twips (from core side) directly. But for other files, we can calculate the twips' position for different scales. Please see "getTileTwipsSheetAreaFromPrint" function and related things for more info.

gokaysatir: We can also do the calculation for Calc but it would be expensive for this case.

gokaysatir:
Thanks for checking @Cwiiis.

browser/src/app/TilesMiddleware.ts
Cwiiis: I'll take your word for it :) I might be nice to have a comment here briefly explaining why this is done for calc alone.

browser/src/app/TilesMiddleware.ts
gokaysatir: Let me note this, i am working on these things and will do it later.

browser/src/app/TilesMiddleware.ts
```diff
@@ -47,7 +47,7 @@ class TileCoordData {
 		this.part = part !== null ? part : app.map._docLayer._selectedPart;
 		this.mode = mode !== undefined ? mode : 0;
 
-		this.scale = Math.round(Math.pow(1.2, this.z - 10) * 1000) / 1000;
+		this.scale = Math.pow(1.2, this.z - 10);
```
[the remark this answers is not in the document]
gokaysatir: I think it's not used at all :) I recently added it.

PR #11393, opened 2025-03-21
reset underlying capacity of previously large buffers

caolanm:
large capacity but ~empty buffers can be reproduced by inserting a large jpg into a blank writer document and copying it to clipboard a few times and then: kill -SIGUSR 1 `pidof coolwsd`

```
rbuffered   rcapacity   wbuffered   wcapacity
        0        1993           0    19881378
```

Change-Id: Idc4020ffef5e3d4bc4801792412e2d1f706e2ef1

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/net/Buffer.hpp
+++ b/net/Buffer.hpp
@@ -28,6 +28,14 @@ class Buffer
     std::size_t _offset;  /// offset into _buffer of data
     std::vector<char> _buffer;
 
+    void resetOffset()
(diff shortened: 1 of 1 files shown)


mmeeks (approved):
Thanks Caolan - can we have in 24.04 too ? =)

PR #11396, opened 2025-03-21
Calc: Call _syncTileContainerSize after changing document size.

gokaysatir:
_syncTileContainerSize returns true if it calls requestRedraw. In that case "_restrictDocumentSize" will not call it again.

Issue: Adding a new sheet while zoomed out or switching sheets may result in a smaller view than expected.

Change-Id: Id3663e360b57b433ffede2213b32cb4d4bcd492b

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/browser/src/layer/tile/CalcTileLayer.js
+++ b/browser/src/layer/tile/CalcTileLayer.js
@@ -256,7 +256,7 @@ L.CalcTileLayer = L.CanvasTileLayer.extend({
 
 		this._map.fire('scrolllimits', newSizePx.clone());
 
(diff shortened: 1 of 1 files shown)


gokaysatir:
Backport of the fix below:
* https://github.com/CollaboraOnline/online/pull/11358/files

@Cwiiis can you have a look when you have time?

Cwiiis (approved):
I'd prefer if this and #11395 were in the same PR, but this looks good to me.

